PR #41383, opened 2023-02-14
ddl: ignore reorg elem error when the job is cancelling

### LINKED ISSUE #41381: Unexpected error "DDL reorg element does not exist" when adding index
## Bug Report

```sql
create table t(id int primary key, b int);
insert into t values (1, 1);
alter table t add unique index idx(b);

insert into t values (2, 1); -- write only
admin cancel ddl jobs X; -- the job must be rolling back because of the duplicated entry
```

It is not easy to reproduce with MySQL client because the state changes are usually fast.

```
Cancelled DDL job
```

or

```
Duplicate entry '1' for key 't.idx'
```

```
ERROR 8235 (HY000): DDL job rollback, error msg: DDL reorg element does not exist
```

tangenta:

Issue Number: close #41381

Problem Summary:

The root cause is the write-conflict between DDL job's transaction and `ADMIN CANCEL` transaction. The latter wins and aborts the DDL job's transaction. As a result, the "duplicate entry" error is lost.

DDL worker starts to handle the cancelling case. It finds that the job is in write-reorg state, and tries to pick up the work by getting reorg handle. However, the reorg handle is cleaned by another previous transaction.

This PR ignores the "reorg element not exist" error when the job is cancelling.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


wjhuang2016 (approved):

ddl/ddl_worker.go
```diff
@@ -770,6 +770,10 @@ func (w *worker) HandleDDLJobTable(d *ddlCtx, job *model.Job) (int64, error) {
 	// and retry later if the job is not cancelled.
 	schemaVer, runJobErr = w.runDDLJob(d, t, job)
 
+	d.mu.RLock()
```
Benjamin2037, line 773: Does this code only for test？

tangenta: Yes.

ddl/column.go
```diff
@@ -822,7 +822,7 @@ func doReorgWorkForModifyColumn(w *worker, d *ddlCtx, t *meta.Meta, job *model.J
 		return false, ver, errors.Trace(err)
 	}
 	reorgInfo, err := getReorgInfo(d.jobContext(job.ID), d, rh, job, dbInfo, tbl, BuildElements(changingCol, changingIdxs), false)
-	if err != nil || reorgInfo.first {
+	if err != nil || reorgInfo == nil || reorgInfo.first {
```
Benjamin2037, line 825: Why modify column need this logic？

tangenta: Because `reorgInfo` is possible to be `nil` after the changes in `getReorgInfo`. `reorgInfo == nil` means the job is cancelling.

ddl/ddl_worker.go
Benjamin2037: Do we have a way to execute these three lines codes only in test?

tangenta: The overhead is negligible. 

Please also check the other methods of this interface. All of them are written in this style.

Benjamin2037 (approved):
LGTM

ddl/ddl_worker.go
Benjamin2037: I know the lock and a nil function call and unlock overhead is very small. Just want to know if there is a way like c "debug micro" that only debug mode has this code.

PR #41394, opened 2023-02-14
executor: fix case-sensitive when grant & revoke

### LINKED ISSUE #41048: case-sensitive when grant & revoke
## Bug Report

mysql> show tables;
Empty set (0.00 sec)

mysql> create table public.TEST(id int);
Query OK, 0 rows affected (0.16 sec)

mysql> show tables;
+------------------+
| Tables_in_public |
+------------------+
| TEST             |
+------------------+
1 row in set (0.01 sec)

mysql> GRANT SELECT ON public.test TO 'root'@'%';
Query OK, 0 rows affected (0.03 sec)

mysql> revoke SELECT ON public.test from 'root'@'%';       
ERROR 1105 (HY000): get table privilege fail for root % public TEST
mysql> 
mysql> revoke SELECT ON public.TEST from 'root'@'%';    
ERROR 1105 (HY000): There is no such grant defined for user 'root' on host '%' on table public.TEST
mysql> 

case-insensitive when grant & revoke
case-sensitive when grant & revoke
mysql> select version();
+--------------------+
| version()          |
+--------------------+
| 5.7.25-TiDB-v6.5.0 |
+--------------------+
1 row in set (0.00 sec)

mysql>

fanrenhoo:
Signed-off-by: fanrenhoo <fanrenhoo@sina.com>

Issue Number: close #41048 

Problem Summary: the grant & revoke syntax did not make case insensitive logic correct for table 

Made the SQL syntax to SELECT、UPDATE mysql.Table_Priv‘s comparation of Table_Name all with lowercase

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
I tested, now it works

MySQL [mysql]> GRANT SELECT ON test.test TO 'root'@'%';
MySQL [mysql]> revoke SELECT ON test.TEST from 'root'@'%';
Query OK, 0 rows affected (0.03 sec)

- [ ] No code

Side effects
No

Documentation
No
- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Make the grant and revoke for table name case insensitive
```

### DIFF
--- a/executor/grant.go
+++ b/executor/grant.go
@@ -548,7 +548,7 @@ func (e *GrantExec) grantTableLevel(priv *ast.PrivElem, user *ast.UserSpec, inte
 	if err != nil {
 		return err
 	}
-	sqlexec.MustFormatSQL(sql, " WHERE User=%? AND Host=%? AND DB=%? AND lower(Table_name)=%?", user.User.Username, user.User.Hostname, dbName, strings.ToLower(tblName))
+	sqlexec.MustFormatSQL(sql, " WHERE User=%? AND Host=%? AND lower(DB)=%? AND lower(Table_name)=%?", user.User.Username, user.User.Hostname, strings.ToLower(dbName), strings.ToLower(tblName))
 
 	ctx := kv.WithInternalSourceType(context.Background(), kv.InternalTxnPrivilege)
 	_, err = internalSession.(sqlexec.SQLExecutor).ExecuteInternal(ctx, sql.String())
@@ -574,7 +574,7 @@ func (e *GrantExec) grantColumnLevel(priv *ast.PrivElem, user *ast.UserSpec, int
 		if err != nil {
 			return err
 		}
-		sqlexec.MustFormatSQL(sql, " WHERE User=%? AND Host=%? AND DB=%? AND lower(Table_name)=%? AND Column_name=%?", user.User.Username, user.User.Hostname, dbName, strings.ToLower(tbl.Meta().Name.O), col.Name.O)
+		sqlexec.MustFormatSQL(sql, " WHERE User=%? AND Host=%? AND lower(DB)=%? AND lower(Table_name)=%? AND lower(Column_name)=%?", user.User.Username, user.User.Hostname, strings.ToLower(dbName), tbl.Meta().Name.L, col.Name.L)
 
 		ctx := kv.WithInternalSourceType(context.Background(), kv.InternalTxnPrivilege)
 		_, err = internalSession.(sqlexec.SQLExecutor).ExecuteInternal(ctx, sql.String())
@@ -698,19 +698,19 @@ func dbUserExists(ctx sessionctx.Context, name string, host string, db string) (
 
 // tableUserExists checks if there is an entry with key user-host-db-tbl in mysql.Tables_priv.
 func tableUserExists(ctx sessionctx.Context, name string, host string, db string, tbl string) (bool, error) {
-	return recordExists(ctx, `SELECT * FROM %n.%n WHERE User=%? AND Host=%? AND DB=%? AND lower(Table_name)=%?;`, mysql.SystemDB, mysql.TablePrivTable, name, host, db, strings.ToLower(tbl))
+	return recordExists(ctx, `SELECT * FROM %n.%n WHERE User=%? AND Host=%? AND lower(DB)=%? AND lower(Table_name)=%?;`, mysql.SystemDB, mysql.TablePrivTable, name, host, strings.ToLower(db), strings.ToLower(tbl))
 }
 
 // columnPrivEntryExists checks if there is an entry with key user-host-db-tbl-col in mysql.Columns_priv.
 func columnPrivEntryExists(ctx sessionctx.Context, name string, host string, db string, tbl string, col string) (bool, error) {
-	return recordExists(ctx, `SELECT * FROM %n.%n WHERE User=%? AND Host=%? AND DB=%? AND lower(Table_name)=%? AND Column_name=%?;`, mysql.SystemDB, mysql.ColumnPrivTable, name, host, db, strings.ToLower(tbl), col)
+	return recordExists(ctx, `SELECT * FROM %n.%n WHERE User=%? AND Host=%? AND lower(DB)=%? AND lower(Table_name)=%? AND lower(Column_name)=%?;`, mysql.SystemDB, mysql.ColumnPrivTable, name, host, strings.ToLower(db), strings.ToLower(tbl), strings.ToLower(col))
 }
 
 // getTablePriv gets current table scope privilege set from mysql.Tables_priv.
 // Return Table_priv and Column_priv.
 func getTablePriv(sctx sessionctx.Context, name string, host string, db string, tbl string) (string, string, error) {
 	ctx := kv.WithInternalSourceType(context.Background(), kv.InternalTxnPrivilege)
-	rs, err := sctx.(sqlexec.SQLExecutor).ExecuteInternal(ctx, `SELECT Table_priv, Column_priv FROM %n.%n WHERE User=%? AND Host=%? AND DB=%? AND lower(Table_name)=%?`, mysql.SystemDB, mysql.TablePrivTable, name, host, db, strings.ToLower(tbl))
+	rs, err := sctx.(sqlexec.SQLExecutor).ExecuteInternal(ctx, `SELECT Table_priv, Column_priv FROM %n.%n WHERE User=%? AND Host=%? AND lower(DB)=%? AND lower(Table_name)=%?`, mysql.SystemDB, mysql.TablePrivTable, name, host, strings.ToLower(db), strings.ToLower(tbl))
 	if err != nil {
 		return "", "", err
 	}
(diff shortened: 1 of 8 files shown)


executor/grant.go
```diff
@@ -580,7 +580,7 @@ func (e *GrantExec) grantColumnLevel(priv *ast.PrivElem, user *ast.UserSpec, int
 		if err != nil {
 			return err
 		}
-		sqlexec.MustFormatSQL(sql, " WHERE User=%? AND Host=%? AND DB=%? AND Table_name=%? AND Column_name=%?", user.User.Username, user.User.Hostname, dbName, tbl.Meta().Name.O, col.Name.O)
+		sqlexec.MustFormatSQL(sql, " WHERE User=%? AND Host=%? AND DB=%? AND lower(Table_name)=%? AND Column_name=%?", user.User.Username, user.User.Hostname, dbName, strings.ToLower(tbl.Meta().Name.O), col.Name.O)
```
lance6716, line 583: does database name and column name also need this `lower`?

BTW, TiDB should behave like lower_case_table_names = 2. ref https://dev.mysql.com/doc/refman/5.6/en/identifier-case-sensitivity.html

executor/revoke.go
```diff
@@ -309,7 +309,7 @@ func (e *RevokeExec) revokeColumnPriv(internalSession sessionctx.Context, priv *
 		if err != nil {
 			return err
 		}
-		sqlexec.MustFormatSQL(sql, " WHERE User=%? AND Host=%? AND DB=%? AND Table_name=%? AND Column_name=%?", user, host, dbName, tbl.Meta().Name.O, col.Name.O)
+		sqlexec.MustFormatSQL(sql, " WHERE User=%? AND Host=%? AND DB=%? AND lower(Table_name)=%? AND Column_name=%?", user, host, dbName, strings.ToLower(tbl.Meta().Name.O), col.Name.O)
```
lance6716, line 312:
```suggestion
		sqlexec.MustFormatSQL(sql, " WHERE User=%? AND Host=%? AND DB=%? AND lower(Table_name)=%? AND Column_name=%?", user, host, dbName, tbl.Meta().Name.L, col.Name.O)
```

For the type `CIStr`, you can use `CIStr.L`

fanrenhoo: THKS MUCH!!!

executor/grant.go
fanrenhoo: Done, please check! THKS

executor/grant.go
```diff
@@ -554,7 +554,7 @@ func (e *GrantExec) grantTableLevel(priv *ast.PrivElem, user *ast.UserSpec, inte
 	if err != nil {
 		return err
 	}
-	sqlexec.MustFormatSQL(sql, " WHERE User=%? AND Host=%? AND DB=%? AND Table_name=%?", user.User.Username, user.User.Hostname, dbName, tblName)
+	sqlexec.MustFormatSQL(sql, " WHERE User=%? AND Host=%? AND lower(DB)=%? AND lower(Table_name)=%?", user.User.Username, user.User.Hostname, strings.ToLower(dbName), strings.ToLower(tblName))
```
lance6716, line 557: Do you want to use `lower` to handle existing data? That will cause table `mysql.tables_priv` can not be accessed by index

https://github.com/pingcap/tidb/blob/8d91a142503da041a80d8d21c877590a43f0db54/session/bootstrap.go#L147

Maybe we should normalize all data to lowercase during an upgrading? @tiancaiamao

fanrenhoo: Thks very much. Yes that's a problem, let me think more. What I worry about is: if we make all data stored lowercase, will make the promption not same with table/db creators initial?

fanrenhoo: can we create index wiith lower(db)...?

lance6716: > can we create index wiith lower(db)...?

Yes, let me clarify it. From my view we now have two problems:

1. There may have both uppercase and lowercase data for one privilege record. For example, user have created a SELECT privilege on db.TBL, and UPDATE privieleg on db.tbl. We should know what's the current behaviour (user will see SELECT or UPDATE or both?), and after your PR this behaviour is not changed.
2. Since the privilege is stored in a table, we should make use of primary key or index to access the table.

CbcWestwolf (changes requested):
In the current TiDB, there will be duplicate records in `mysql.tables_priv` if we grant the same privilege to a user with a different case of `Table_name`:
```sql
tidb0> create table T (id int);
tidb0> grant select on T to root;
tidb0> grant select on t to root;
tidb0> select * from mysql.tables_priv where User = 'root'
+------+------+------+------------+---------+---------------------+------------+-------------+
| Host | DB   | User | Table_name | Grantor | Timestamp           | Table_priv | Column_priv |
+------+------+------+------------+---------+---------------------+------------+-------------+
| %    | test | root | T          | root@%  | 2023-03-03 14:25:09 | Select     | Select      |
| %    | test | root | t          | root@%  | 2023-03-03 16:22:41 | Select     | Select      |
+------+------+------+------------+---------+---------------------+------------+-------------+
```

However, MySQL only keeps the lowercase of `Table_name` in the table:
```sql
MySQL> select * from mysql.tables_priv where user = 'root'
+-----------+------+------+------------+----------------+---------------------+------------+-------------+
| Host      | Db   | User | Table_name | Grantor        | Timestamp           | Table_priv | Column_priv |
+-----------+------+------+------------+----------------+---------------------+------------+-------------+
| localhost | test | root | t          | root@localhost | 2023-03-03 16:23:10 | Select     |             |
+-----------+------+------+------------+----------------+---------------------+------------+-------------+
```

So I think we need to:
1. Use `lower()` when recording table name in `mysql.tables_priv`
2. Remove the historical duplicate records  when upgrading. You can take `upgradeToVerXXXX` in `session/bootstrap.go` as an example.

fanrenhoo:
> In the current TiDB, there will be duplicate records in `mysql.tables_priv` if we grant the same privilege to a user with a different case of `Table_name`:
> 
> ```sql
> tidb0> create table T (id int);
> tidb0> grant select on T to root;
> tidb0> grant select on t to root;
> tidb0> select * from mysql.tables_priv where User = 'root'
> +------+------+------+------------+---------+---------------------+------------+-------------+
> | Host | DB   | User | Table_name | Grantor | Timestamp           | Table_priv | Column_priv |
> +------+------+------+------------+---------+---------------------+------------+-------------+
> | %    | test | root | T          | root@%  | 2023-03-03 14:25:09 | Select     | Select      |
> | %    | test | root | t          | root@%  | 2023-03-03 16:22:41 | Select     | Select      |
> +------+------+------+------------+---------+---------------------+------------+-------------+
> ```
> 
> However, MySQL only keeps the lowercase of `Table_name` in the table:
> 
> ```sql
> MySQL> select * from mysql.tables_priv where user = 'root'
> +-----------+------+------+------------+----------------+---------------------+------------+-------------+
> | Host      | Db   | User | Table_name | Grantor        | Timestamp           | Table_priv | Column_priv |
> +-----------+------+------+------------+----------------+---------------------+------------+-------------+
> | localhost | test | root | t          | root@localhost | 2023-03-03 16:23:10 | Select     |             |
> +-----------+------+------+------------+----------------+---------------------+------------+-------------+
> ```
> 
> So I think we need to:
> 
> 1. Use `lower()` when recording table name in `mysql.tables_priv`
> 2. Remove the historical duplicate records  when upgrading. You can take `upgradeToVerXXXX` in `session/bootstrap.go` as an example.

fanrenhoo:
Hi，I fixed with your suggestion, but not stored all with lowercase, just keep the CASE when first time used, and delete the duplicate when upgrade

session/bootstrap.go
```diff
@@ -163,6 +163,7 @@ const (
 		Column_name	CHAR(64),
 		Timestamp	TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
 		Column_priv	SET('Select','Insert','Update','References'),
+		KEY             IDX(Host, lower(DB), User, lower(Table_name), lower(Column_name)),
```
lance6716, line 166: Can you try to convert existing data to lowercase and use primary key + lowercased data to access the table, rather than add this index? Index is not the best way in this scenario

bb7133:
Hi @fanrenhoo thanks for your contribution.

> Remove the historical duplicate records when upgrading. You can take upgradeToVerXXXX in session/bootstrap.go as an example.

IMHO the challenging part here is how we deal with the 'duplicate records', I can imagine it's dangerous to remove the privileged data correctly.

So my suggestion is, can we just change the collation `DB` and `Table_name` to `utf8mb4_unicode_ci`? By doing so, the they will be case-insensitive by definition:

```
// CreateTablePrivTable is the SQL statement creates table scope privilege table in system db.
CreateTablePrivTable = `CREATE TABLE IF NOT EXISTS mysql.tables_priv (
	Host		CHAR(255),
	DB		CHAR(64) CHARSET utf8mb4 COLLATE utf8mb4_unicode_ci,
	User		CHAR(32) ,
	Table_name	CHAR(64) CHARSET utf8mb4 COLLATE utf8mb4_unicode_ci,
	Grantor		CHAR(77),
	Timestamp	TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
	Table_priv	SET('Select','Insert','Update','Delete','Create','Drop','Grant','Index','Alter','Create View','Show View','Trigger','References'),
	Column_priv	SET('Select','Insert','Update','References'),
	PRIMARY KEY (Host, DB, User, Table_name));`
```

mjonss (commented):
Would it be possible to always use the actual table name (i.e. as created), so when stored in different tables like `mysql.tables_priv`, it is matching the actual case of the original table?
And do a lookup for the existing table name before checking the grants etc.

session/bootstrap.go
fanrenhoo: Done please check, THKS

fanrenhoo:
Hi, One question: I saw the related issue fixed by another "https://github.com/pingcap/tidb/pull/43443". So should I close this one, and do need my fix above anymore?

mjonss:
> Hi, One question: I saw the related issue fixed by another "#43443". So should I close this one, and do need my fix above anymore?

Hi @fanrenhoo and sorry for the very late reply. #43443 will fix this issue for clusters created with v7.2.0 or later by changing the collation (how to compare and thus order strings) to utf8mb4_generic_ci for Schema, Table and Column names.

I will leave it up to you if you want to close this, or to add handling of possible duplicates names etc. during upgrades.

PR #41209, opened 2023-02-08
Optimizer: derive TopN from filter on row number

ghazalfamilyusa:
Issue Number: Ref #39792

### Problem Summary:
This PR is planner support for deriving TopN from "filter on row number". The PR inserts a TopN node above the window function that computes row number and discards the filter since it is not needed. There is no execution engine support needed for this change since it is limited to "no partition by". In this case, the current TopN operator is sufficient to handle this logic. Next PR will add partition by fields as extension to TopN. Execution support for this extension is here https://github.com/tikv/tikv/pull/14116

The code has sufficient comments about the pattern and action. For completeness here is the description:
Patter
- Look for filter over row number window function data source "datasource->window->filter"
        - Filter is simple condition of row_number < value or row_number <= value
        - The window function is a simple row number
        - With default frame: rows between current row and current row. Check is not necessary since
          current row is only frame applicable to row number
        - No partition
        - Child of window is a data source.
 

### Tests
Unit tests

### Side effects 
None

### Documentation
None

### Release note

```release-note
None
```

Issue #39792, opened 2022-12-09
filter on `row_number()` can be used on optimize `window_sort` to `window_topN`


windtalker:
## Enhancement
```
mysql> desc test;
+-------+---------+------+------+---------+-------+
| Field | Type    | Null | Key  | Default | Extra |
+-------+---------+------+------+---------+-------+
| id    | int(11) | YES  |      | NULL    |       |
| value | date    | YES  |      | NULL    |       |
+-------+---------+------+------+---------+-------+
2 rows in set (0.00 sec)

mysql> explain select * from (select *, row_number() over (order by value) as rn from test) d where rn < 10;
+------------------------------------+---------+--------------+---------------+------------------------------------------------------------------------------------------------+
| id                                 | estRows | task         | access object | operator info                                                                                  |
+------------------------------------+---------+--------------+---------------+------------------------------------------------------------------------------------------------+
| TableReader_30                     | 1.60    | root         |               | data:ExchangeSender_29                                                                         |
| └─ExchangeSender_29                | 1.60    | mpp[tiflash] |               | ExchangeType: PassThrough                                                                      |
|   └─Selection_28                   | 1.60    | mpp[tiflash] |               | lt(Column#5, 10)                                                                               |
|     └─Window_25                    | 2.00    | mpp[tiflash] |               | row_number()->Column#5 over(order by test.test.value rows between current row and current row) |
|       └─Sort_17                    | 2.00    | mpp[tiflash] |               | test.test.value                                                                                |
|         └─ExchangeReceiver_16      | 2.00    | mpp[tiflash] |               |                                                                                                |
|           └─ExchangeSender_15      | 2.00    | mpp[tiflash] |               | ExchangeType: PassThrough                                                                      |
|             └─TableFullScan_14     | 2.00    | mpp[tiflash] | table:test    | keep order:false, stats:pseudo                                                                 |
+------------------------------------+---------+--------------+---------------+------------------------------------------------------------------------------------------------+
8 rows in set (0.00 sec)
```
As shown above, for window function row_number(), there is a common usage that user only wants to get the first N rows, so user will add a filter on the results of row_number().
However, TiDB's plan for this kind of query requires global sort on all data set, if the orignal data set is big, the performance is bad. 
A potential optimization is the optimizer push `where rn < 100` to the sort operation, and convert `sort` to `topn`

fixdb:
[...]

PR #41469, opened 2023-02-16
Optimizer: Add partition by support for derived TopN(filter on row_nu…

ghazalfamilyusa:
Issue Number: Ref #39792

### Problem Summary:
This is a continuation of https://github.com/pingcap/tidb/pull/41209 which introduced deriving top N from "row_number as RN ... where RN <= N". This PR adds support for partition by in the window function "row_number as RN over (partition by X) ... where RN <= N" and restricted if the partition by is a prefix of the PK. Execution support for this extension is here https://github.com/tikv/tikv/pull/14116

**Design**

This optimization is a logical rewrite that (1) finds the desired pattern and (2) rewrites the query by inserting a TopN node that is expected to be pushed down to storage later by the physical planner.

- **Pattern**
The logical rewrite looks like for the pattern selection(filter)->window->data-source
with the following restrictions: 
          - Filter is simple condition of row_number < value or row_number <= value
          - The window function is a simple row number with default frame. If the row_number has partition by then the partition
          - by fields should be a prefix of clustered index. This is reflects the limitation in the executor that works only 
             for this case. 
          - Data source has no tiflash option.
- **Rewrite**
Add topN above (down stream) of data source. The write change selection(filter)->window->data-source to selection(filter)->window->topN->data-source. TopN node is extended to store the partition by of row_number if any.  The physical planner may change TopN to Limit if sort is met at the source. For this reason, we also extended PhysicalLimit to include partition by as well. Another change made to physical planner is to only apply TopN/Limit to TiKv since it does not help to apply it at the root. Applying TopN at the root with partition by also requires changes to executor which is not needed. 

- **Dependencies**
The change in this PR address only the planner enhancements. The end to end solution requires: (a) extending the interface between physical planner and execution to include the partition by fields for both Limit and TopN (b) enhancing the execution engine code for Limit and TopN to support partitioning when data is fully/partially ordered on partition key.

**Scope and user interface**
The change adds a session/global variable to enable/disable this optimization called "tidb_opt_derive_topn" and it is off by default to reduce the risk. The change also limits the optimization to TikV and rule not applied when TiFlash exists. This is the case since TiFllash execution does not support the extended TopN (with partition). Also, TiFlash can push down window functions and probably does not need this optimization.

**Customer use case**
The original user case that motivates this change is tested and work as expected in the planner. See planner/core/casetest/testdata/derive_topn_from_window_out.json in this PR. 
```
        "SQL": "select * from (select *, row_number() over (partition by primary_key, secondary_key order by c_timestamp) as rownum from customer where primary_key = 0x002 and secondary_key >= 0x001 and c_timestamp >= 1661883508511000000) as nested where rownum <= 10 order by secondary_key desc;",
        "Plan": [
          "Sort 0.89 root  test.customer.secondary_key:desc",
          "└─Selection 0.89 root  le(Column#6, 10)",
          "  └─Window 1.11 root  row_number()->Column#6 over(partition by test.customer.primary_key, test.customer.secondary_key order by test.customer.c_timestamp rows between current row and current row)",
          "    └─Sort 1.11 root  test.customer.primary_key, test.customer.secondary_key, test.customer.c_timestamp",
          "      └─TableReader 1.11 root  data:TopN",
          "        └─TopN 1.11 cop[tikv]  partition by test.customer.primary_key, test.customer.secondary_key order by test.customer.c_timestamp, offset:0, count:10",
          "          └─Selection 1.11 cop[tikv]  ge(test.customer.c_timestamp, 1661883508511000000)",
          "            └─TableRangeScan 33.33 cop[tikv] table:customer range:[0x0002 0x0001,0x0002 +inf], keep order:false, stats:pseudo"
```

### Tests
Unit tests

### Side effects
None

### Documentation
None

### Release note

```release-note
None
```


planner/core/casetest/testdata/derive_topn_from_window_out.json
```diff
@@ -196,18 +331,98 @@
...
+      },
+      {
+        "SQL": "select * from (select *, row_number() over (partition by primary_key, secondary_key order by c_timestamp) as rownum from customer where primary_key = 0x002 and secondary_key >= 0x001 and c_timestamp >= 1661883508511000000) as nested where rownum <= 10 order by secondary_key desc;",
+        "Plan": [
+          "Sort 1.11 root  test.customer.secondary_key:desc",
+          "└─Window 1.11 root  row_number()->Column#6 over(partition by test.customer.primary_key, test.customer.secondary_key order by test.customer.c_timestamp rows between current row and current row)",
+          "  └─Sort 1.11 root  test.customer.primary_key, test.customer.secondary_key, test.customer.c_timestamp",
+          "    └─TopN 1.11 root  test.customer.c_timestamp, offset:0, count:10",
+          "      └─TableReader 1.11 root  data:TopN",
+          "        └─TopN 1.11 cop[tikv]  test.customer.c_timestamp, offset:0, count:10",
```
LittleFall, line 421: can we add partition by column to explain the result?
like "TopN 1.11 cop[tikv]  order_by:XXX desc, partition_by: XXX, offset: X, count: X"

ghazalfamilyusa: Good point and will do in next update.

planner/core/rule_derive_topn_from_window.go
```diff
@@ -71,12 +96,21 @@ func windowIsTopN(p *LogicalSelection) (bool, uint64) {
...
 	if !isDataSource {
 		return false, 0
 	}
-	if len(child.WindowFuncDescs) == 1 && child.WindowFuncDescs[0].Name == "row_number" && len(child.PartitionBy) == 0 &&
-		child.Frame.Type == ast.Rows && child.Frame.Start.Type == ast.CurrentRow && child.Frame.End.Type == ast.CurrentRow {
+
+	// Give up if TiFlash is one possible access path. Pushing down window aggregation is good enough in this case.
+	for _, path := range dataSource.possibleAccessPaths {
+		if path.StoreType == kv.TiFlash {
+			return false, 0
```
chrysan, line 107: Do you mean we will not compare the cost of mpp window aggregation to the topn pushdown in this PR?

planner/core/rule_derive_topn_from_window.go
```diff
@@ -38,14 +39,38 @@ func appendDerivedTopNTrace(topN LogicalPlan, opt *logicalOptimizeOp) {
...
 
+// checkPartitionBy mainly checks if partition by of window function is a prefix of
+// data order (clustered index) of the data source. TiFlash is allowed only for empty partition by.
+func checkPartitionBy(p *LogicalWindow, d *DataSource) bool {
+	// No window partition by. We are OK.
+	if len(p.PartitionBy) == 0 {
+		return true
+	}
+
+	// Table not clustered and window has partition by. Can not do the TopN piush down.
```
chrysan, line 50:
```suggestion
	// Table not clustered and window has partition by. Can not do the TopN push down.
```
A typo here.

planner/core/rule_derive_topn_from_window.go
ghazalfamilyusa: Yes. Our goal in this PR is to allow KiTv TopN push down for now with a flag. We need a more fundamental way of choosing between mpp and TiKV later on.

planner/core/rule_derive_topn_from_window.go
ghazalfamilyusa: Got it and will be fixed in next update.

planner/core/casetest/rule_derive_topn_from_window_test.go
```diff
@@ -82,3 +138,28 @@ func TestPushDerivedTopnPositive(t *testing.T) {
...
 }
+
+// Negative test when tidb_opt_derive_topn is off
+func TestPushDerivedTopnFlagOff(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("set tidb_opt_derive_topn=0")
+	tk.MustExec("use test")
+	tk.MustExec("drop table if exists t")
+	tk.MustExec("create table t(a int, b int, primary key(b,a))")
```
chrysan, line 149: Can you add cases to test what will happen to pk with generated columns? Like `create table t (a int, b int as (a+1) stored, primary key(b,a));`

ghazalfamilyusa: What is the secanrio you intend to test? is it partition by a,b or is it partition by a,a1+1?

chrysan: Maybe both? Just to check it works well too with generated columns.

ghazalfamilyusa: @chrysan this is done along with previous comments in latest update.

chrysan (approved):
should have some tests to cover result correctness? Rest LGTM.

sessionctx/variable/tidb_vars.go
```diff
@@ -49,6 +49,9 @@ const (
 	// TiDBOptAggPushDown is used to enable/disable the optimizer rule of aggregation push down.
 	TiDBOptAggPushDown = "tidb_opt_agg_push_down"
 
+	// TiDBOptDeriveTopN is used to enable/disable the optimizer rule of deriving topN.
```
LittleFall, line 52: Should we control this optimizer rule by the blocklist, instead of a new variable?


ref: https://docs.pingcap.com/tidb/stable/blocklist-control-plan#important-optimization-rules

ghazalfamilyusa: I think the use case is different for system variables and blocklist. Blocklist to me seems a configuration which lasts longer. Session/global variables can be used to control a more granular level on  query level as a workaround to force using or not using an optimization.

qw4990: Yes, actually the optimizer rule blocklist is rarely used now, so a system variable seems better here.

qw4990 (approved):

PR #41586, opened 2023-02-20
ttl: limit total count of running tasks across the cluster

### LINKED ISSUE #41585: the concurrency of running TTL tasks should be limited
The TTL tasks are not limited, so all workers (at least 1 * count of TiDB) will join to work on the TTL tasks, which make it hard to control the whole resource consumed by the TTL.

We could add a new variable `tidb_ttl_running_task` to control the whole concurrency of TTL tasks. All workers must check the `running` tasks cannot exceed this value.

YangKeao:

Issue Number: close #41585

Problem Summary:

A new system varaible `tidb_ttl_running_task` is introduced to limit the concurrency of TTL tasks. The default value of it is 0, which means 3 times the count of TiKV. If it's set to -1, it means no limitation.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

### Release note

```release-note
A new variable `tidb_ttl_running_task` is added to limit the concurrency of TTL tasks across the cluster.
```


sessionctx/variable/sysvar.go
```diff
@@ -2368,6 +2368,16 @@ var defaultSysVars = []*SysVar{
 			return BoolToOnOff(s.EnableINLJoinInnerMultiPattern), nil
 		},
 	},
+	{Scope: ScopeGlobal, Name: TiDBTTLRunningTask, Value: strconv.Itoa(DefTiDBTTLRunningTask), Type: TypeInt, MinValue: -1, MaxValue: 256, SetGlobal: func(ctx context.Context, vars *SessionVars, s string) error {
```
lcwangchao, line 2371: It's better to use the pre-defined `MaxConfigurableConcurrency` as the upper bound of the value. BTW, I found other similar variables have a range `[1, MaxConfigurableConcurrency]` , should we follow the similar design? I think "unlimited" value is meaningless (zero value can be used as auto).

sessionctx/variable/tidb_vars.go
```diff
@@ -929,6 +929,9 @@ const (
 	TiDBStmtSummaryFileMaxSize = "tidb_stmt_summary_file_max_size"
 	// TiDBStmtSummaryFileMaxBackups indicates the maximum number of files written by stmtsummary.
 	TiDBStmtSummaryFileMaxBackups = "tidb_stmt_summary_file_max_backups"
+	// TiDBTTLRunningTask limits the count of running ttl tasks. Default to 0, means 3 times the count of TiKV (or no
+	// limitation, if the storage is not TiKV). If set to -1, means there is no limitation.
+	TiDBTTLRunningTask = "tidb_ttl_running_task"
```
lcwangchao, line 934: `tidb_ttl_running_tasks` ?

ttl/ttlworker/task_manager.go
```diff
@@ -328,12 +348,26 @@ func (m *taskManager) lockScanTask(se session.Session, task *cache.TTLTask, now
 	err := se.RunInTxn(ctx, func() error {
 		var err error
 
+		// The `for update` here ensures two things:
+		// 1. The instance which will update this row has committed, and we can get the newest information.
+		// 2. There is only one successful transaction concurrently.
+		//    If it's a `for update nowait`, we cannot avoid the situation that multiple transactions fetched the tasks
+		//    at the same time, and the task limitation doesn't work.
```
lcwangchao, line 355: I think these comments should be moved into `syncTaskFromTable`

lcwangchao: tidb does not support gapped lock, is it possible to avoid the task count to exceed the limit?

ttl/ttlworker/task_manager.go
```diff
@@ -269,14 +281,22 @@ func (m *taskManager) rescheduleTasks(se session.Session, now time.Time) {
 		return
 	}
 
+loop:
 	for _, t := range tasks {
```
lcwangchao, line 285: Do we need to count the running tasks before locking a task to reduce some unnecessary queries?

YangKeao: I don't think it's necessary.

There is only one extra transaction (count the running tasks also requires one query), becuase we'll `break loop` after encoutering `errTooManyRunningTasks`.

sessionctx/variable/tidb_vars.go
YangKeao: Good!

sessionctx/variable/sysvar.go
YangKeao: I don't have opinion on `MaxConfigurableConcurrency`. I don't know the "normal" size of a TiDB cluster. Could you give any advice on the value of `MaxConfigurableConcurrency`?

ttl/ttlworker/task_manager.go
YangKeao: > I think these comments should be moved into `syncTaskFromTable`

There comments are here to describe how we make the limitation work, and the limitation is checked and considered in `lockScanTask`, so I kept these comments in the `lockScanTask`.

> tidb does not support gapped lock, is it possible to avoid the task count to exceed the limit?

How is it connected with gapped lock? This is only a point get (with `for update`) :thinking:

sessionctx/variable/sysvar.go
lcwangchao: MaxConfigurableConcurrency has already been defined to 256: 	https://github.com/pingcap/tidb/blob/796d5b9bfad25249fa96a8627c41449576686074/sessionctx/variable/tidb_vars.go#L949

ttl/ttlworker/task_manager.go
YangKeao: Discussed with lcwangchao. Resolved.

lcwangchao (approved):

bb7133 (approved):
LGTM

PR #41706, opened 2023-02-23
ttl: ttl resource limit without delete

YangKeao:
This is only a pull request to build TiDB binary with TTL resource group, but don't actually delete anything, to test the performance impact of scanning in TTL.

Don't need to review or merge.


YangKeao:
CI will not run for a draft PR :facepalm:

Issue #38825, opened 2022-11-02
Global Resource Quota Control 


nolouch:
# Feature Request
## Background
A Global Resource Control mechanism that limits multiple applications' resource usage in a shared TiDB cluster, avoids interference between applications.

## Design
RFC: https://github.com/pingcap/tidb/blob/master/docs/design/2022-11-25-global-resource-control.md

###  Interface and Configuration Part
Here a tiny example of the SQL interface.
```
CREATE RESOURCE GROUP `group_name` RRU_PER_SECOND=10000, WRU_PER_SECOND=10000.
ALTER RESOURCE GROUP `group_name` RRU_PER_SECOND= 20000;
/* bind resource group to user*/
ALTER USER user_name RESOURCE GROUP resouce_group_name;
/* reset to default systemn resource group for the current session*/
SET CURRENT RESOURCE GROUP DEFAULT;
```
Above the example shown, it's easy to set up for users or tenants. In theory, we can measure at the requested level,  it is able to aggregate at different levels, so it is possible to bind the resource unit limit to the user or the session to a group.

### SQL Interface
- [ ] Supports Resource Group Syntax
    - [x] Supports Create/Alter Syntax https://github.com/pingcap/tidb/pull/39546  @nolouch 
- [ ] DDL Part
     - [x] Supports Create/Alter/Delete Resource Group https://github.com/pingcap/tidb/pull/40188 @nolouch 
     - [x] Supports Bind Resource Group for Users  https://github.com/pingcap/tidb/pull/39561
     - [x] https://github.com/pingcap/tidb/issues/40380
- [ ] Usability
    - [x] Show Resource Groups with SQL interface
    - [x] #39781
    - [ ] Estimate total RU in a cluster @BornChanger  @glorv 
    - [ ] History RU Metrics @JmPotato 
        - [x] https://github.com/tikv/pd/issues/5854 
        - [x] https://github.com/pingcap/tidb/issues/40526
- [x] Privilege check & constraint syntax check.  https://github.com/pingcap/tidb/pull/40632 @tiancaiamao  
- [ ] Internal Job priority push to tikv/tiflash, like DDL, Analyze  @tiancaiamao @glorv 
- [ ] Test&Compatibility
    - [x] https://github.com/pingcap/tidb/issues/40371
    - [x] https://github.com/pingcap/tidb/issues/40370

### Proto
- [x] Protobuf relative
    - [x] https://github.com/pingcap/kvproto/pull/1019
[...]

PR #41722, opened 2023-02-24
session: support set current session's resource group name

glorv:

Issue Number: ref #38825 

Problem Summary:

Add following two sql:
- `SET RESOURCE GROUP <name>;`  to change the resource group of the current session
- `SELECT current_resource_group();` to return the resource group of the current session
Support set resource group name for the current session.

NOTE: currently, we don't do privilege for this statement. Will do it in the future version.

### Check List

Tests 

- [ ] Unit test

Side effects

Documentation

- [x] Contains syntax changes

### Release note

```release-note
Support change and return the resource group name of current session.
```


nolouch (approved):
lgtm

nolouch:
ptal @tiancaiamao @BornChanger

parser/parser.y
```diff
@@ -10067,6 +10067,10 @@ SetStmt:
 	{
 		$$ = &ast.SetSessionStatesStmt{SessionStates: $3}
 	}
+|	"SET" "RESOURCE" "GROUP" ResourceGroupName
```
BornChanger, line 10070: please also provide the sql command like `show current resource group` to show the resource group for current session.

executor/simple_test.go
```diff
@@ -133,3 +133,17 @@ func TestUserAttributes(t *testing.T) {
 	rootTK.MustExec("alter user usr1 resource group rg1")
 	rootTK.MustQuery("select user_attributes from mysql.user where user = 'usr1'").Check(testkit.Rows(`{"metadata": {"comment": "comment1"}, "resource_group": "rg1"}`))
 }
+
+func TestSetResourceGroup(t *testing.T) {
```
BornChanger, line 137: Please add case to verify hint overrides session setting and session setting override use binding.

glorv: done

glorv:
@nolouch I have added support for `SELECT CURRENT_RESOURCE_GROUP();`, so PTAL again, thanks~

parser/parser.y
glorv: done

nolouch (commented):
lgtm~

nolouch:
ptal @JmPotato @BornChanger

JmPotato (approved):

glorv:
@tiancaiamao  PTAL, as this PR has introduced a new builtin function.

parser/parser.y
```diff
@@ -74,222 +74,223 @@ import (
...
+	convert              "CONVERT"
+	create               "CREATE"
+	cross                "CROSS"
+	cumeDist             "CUME_DIST"
+	currentDate          "CURRENT_DATE"
+	currentTime          "CURRENT_TIME"
+	currentTs            "CURRENT_TIMESTAMP"
+	currentUser          "CURRENT_USER"
+	currentRole          "CURRENT_ROLE"
+	currentResourceGroup "CURRENT_RESOURCE_GROUP"
```
tiancaiamao, line 110: MySQL does not have `CURRENT_RESOURCE_GROUP` as a reserved keyword https://dev.mysql.com/doc/refman/8.0/en/keywords.html

Maybe we can make it non-keyword...
The different is that you can't `create table CURRENT_RESOURCE_GROUP (...)` after this change.

But the current way is still OK to me, since it follows the `CURRENT_XXX` as reserved keyword convention in MySQL

tiancaiamao (approved):

parser/parser.y
glorv: Oh, yes, I think we can make it a non-keyword, so the compatibility should be better.

PR #41726, opened 2023-02-24
adapter, txn: Move aggressive locking controlling logic to handlePessimisticXXX functions to avoid some potential misuses

### LINKED ISSUE #41724: Some tests panics when running all tests with tidb_pessimistic_txn_aggressive_locking globally enabled
## Bug Report

Change the default value of variable `tidb_pessimistic_txn_aggressive_locking` to true in the code, then run all unit tests

Everything goes well.

Some test cases fails:

```
Error 1105: Trying to start aggressive locking while it's already started
```

After some investigation, it's found to be caused by not exiting aggressive locking after executing the previous statement in the transaction.

For example, `TestRcTSOCmdCountForPrepareExecuteExtra` panics because TiDB's `ast.IsReadOnly` regards a select statement with a select-for-update subquery to be readonly (which is wrong). This causes `StmtCommit` or `StmtRollback` (which are responsible to exit aggressive locking if it's active) not invoked after finishing the statement.

`show_profile` in mysql-test panics in a similar way, but it's not successfully reproduced.

This means exiting aggressive locking in `StmtCommit` or `StmtRollback` is a bad idea since TiDB has complicated and even maybe incorrect logic to determine whether `StmtCommit` or `StmtRollback` needs to be called after finishing a statement. We'd better exit aggressive locking within `handlePessimisticDML` and `handlePessimisticSelectForUpdate`, which is also the place that enters aggressive locking. 

master

MyonKeminta:

Issue Number: close #41724 

Problem Summary:

This PR moves invokations of aggressive locking starting/retrying/exiting to single functions (`handlePessimisticDML` and `handlePessimisticSelectForUpdate`). These two functions will enter aggressive locking if needed, and exit aggressive locking when it returns. Previously we exits aggressive locking in `StmtCommit` and `StmtRollback`. However in issue #41724 we founds that sometimes TiDB doesn't evaluate the condition to call `StmtCommit`/`StmtRollback` correctly, causing the aggressive locking state leak to the next statement. This PR avoids this leaking by exiting aggressive locking within the function that starts the aggressive locking.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that sometimes query fails when `tidb_pessimistic_txn_aggressive_locking` is enabled and prepare-execute is used on `select ... (select ... for update)` statements (select statements with select-for-update subquery).
```

### DIFF
--- a/sessiontxn/interface.go
+++ b/sessiontxn/interface.go
@@ -136,12 +136,12 @@ type TxnContextProvider interface {
 	OnInitialize(ctx context.Context, enterNewTxnType EnterNewTxnType) error
 	// OnStmtStart is the hook that should be called when a new statement started
 	OnStmtStart(ctx context.Context, node ast.StmtNode) error
-	// OnHandlePessimisticStmtStart is the hook that should be called when starts handling a pessimistic DML or
+	// OnPessimisticStmtStart is the hook that should be called when starts handling a pessimistic DML or
 	// a pessimistic select-for-update statement.
-	OnHandlePessimisticStmtStart(ctx context.Context) error
-	// OnHandlePessimisticStmtEnd is the hook that should be called when finishes handling a pessimistic DML or
+	OnPessimisticStmtStart(ctx context.Context) error
+	// OnPessimisticStmtEnd is the hook that should be called when finishes handling a pessimistic DML or
 	// select-for-update statement.
-	OnHandlePessimisticStmtEnd(ctx context.Context, isSuccessful bool) error
+	OnPessimisticStmtEnd(ctx context.Context, isSuccessful bool) error
 	// OnStmtErrorForNextAction is the hook that should be called when a new statement get an error
 	OnStmtErrorForNextAction(ctx context.Context, point StmtErrorHandlePoint, err error) (StmtErrorAction, error)
 	// OnStmtRetry is the hook that should be called when a statement is retried internally.
@@ -188,12 +188,12 @@ type TxnManager interface {
 	OnTxnEnd()
 	// OnStmtStart is the hook that should be called when a new statement started
 	OnStmtStart(ctx context.Context, node ast.StmtNode) error
-	// OnHandlePessimisticStmtStart is the hook that should be called when starts handling a pessimistic DML or
+	// OnPessimisticStmtStart is the hook that should be called when starts handling a pessimistic DML or
 	// a pessimistic select-for-update statement.
-	OnHandlePessimisticStmtStart(ctx context.Context) error
-	// OnHandlePessimisticStmtEnd is the hook that should be called when finishes handling a pessimistic DML or
+	OnPessimisticStmtStart(ctx context.Context) error
+	// OnPessimisticStmtEnd is the hook that should be called when finishes handling a pessimistic DML or
 	// select-for-update statement.
-	OnHandlePessimisticStmtEnd(ctx context.Context, isSuccessful bool) error
+	OnPessimisticStmtEnd(ctx context.Context, isSuccessful bool) error
 	// OnStmtErrorForNextAction is the hook that should be called when a new statement get an error
 	// This method is not required to be called for every error in the statement,
 	// it is only required to be called for some errors handled in some specified points given by the parameter `point`.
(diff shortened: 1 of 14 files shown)


sessiontxn/interface.go
```diff
@@ -137,8 +137,14 @@ type TxnContextProvider interface {
 	// OnStmtStart is the hook that should be called when a new statement started
 	OnStmtStart(ctx context.Context, node ast.StmtNode) error
 	// OnHandlePessimisticStmtStart is the hook that should be called when starts handling a pessimistic DML or
-	// a pessimistic select-for-update statements.
+	// a pessimistic select-for-update statement.
 	OnHandlePessimisticStmtStart(ctx context.Context) error
+	// OnPessimisticLockError is the hook that should be called when lock error happens during a pessimistic DML or
+	// select-for-update statement.
+	OnPessimisticLockError(ctx context.Context, lockErr error) error
```
cfzjywxk, line 144: `OnStmtErrorForNextAction` has some logic about handling `StmtErrAfterPessimisticLock`, could we merge these logic together so we could avoid adding a new interface.

ekexium (approved):

sessiontxn/interface.go
MyonKeminta: That makes sense and looks safe to use. I didn't notice that.

sessiontxn/interface.go
```diff
@@ -137,8 +137,14 @@ type TxnContextProvider interface {
...
+	// OnHandlePessimisticStmtEnd is the hook that should be called when finishes handling a pessimistic DML or
+	// select-for-update statement.
+	OnHandlePessimisticStmtEnd(ctx context.Context, isSuccessful bool) error
```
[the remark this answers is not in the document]
MyonKeminta: I actually want to emphasize their difference with `OnStmtStart` etc, but, ok

sessiontxn/isolation/repeatable_read.go
```diff
@@ -234,7 +234,11 @@ func notNeedGetLatestTSFromPD(plan plannercore.Plan, inLockOrWriteStmt bool) boo
 	return false
 }
 
-func (p *PessimisticRRTxnContextProvider) handleAfterPessimisticLockError(lockErr error) (sessiontxn.StmtErrorAction, error) {
+func (p *PessimisticRRTxnContextProvider) handleAfterPessimisticLockError(ctx context.Context, lockErr error) (sessiontxn.StmtErrorAction, error) {
+	if err := p.retryAggressiveLockingIfNeeded(ctx); err != nil {
+		return sessiontxn.ErrorAction(err)
+	}
```
cfzjywxk, line 240: Is `txn.retryAggressiveLockingIfNeeded` necessary for the other two pathes that error would be returned? If not should we move this part to L255-L268?

MyonKeminta: Ah yes, I missed that. It's only needed in retrying cases. And I just noticed that for single-statement-deadlock cases, we should exit aggressive locking mode, otherwise it will still keep the lock that's blocking the other transaction.

sessiontxn/isolation/readcommitted.go
```diff
@@ -237,6 +242,9 @@ func (p *PessimisticRCTxnContextProvider) handleAfterPessimisticLockError(lockEr
 	}
```
cfzjywxk, line 242: Should we call the `CancelAggressiveLocking` for the last else branch?
```
if deadlock, ok := errors.Cause(lockErr).(*tikverr.ErrDeadlock); ok && deadlock.IsRetryable {
} else if terror.ErrorEqual(kv.ErrWriteConflict, lockErr) {
} else {
  // This branch.
}
```

MyonKeminta: The final `else` branch will not set `retryable = true` thus the function will finally return error. Cancelling here looks not necessary.

cfzjywxk (approved):

MyonKeminta:
@ekexium There's some change to the code. Could you take a look again?

MyonKeminta:
emmmm
/merge

ekexium:
> @ekexium There's some change to the code. Could you take a look again?

LGTM

Issue #37119, opened 2022-08-16
Support distributed processing of DDL reorg phase


zimulala:
## Feature Request

**Is your feature request related to a problem? Please describe:**

TiDB currently supports the parallel processing of DDL jobs on the owner. However, the resources of a single TiDB are limited. Even if the parallel framework is supported, the speed of DDL execution is limited, and the competition for resources will affect the daily operations of TiDB, such as TPS.

**Describe the feature you'd like:**

At present, the time-consuming and resource-consuming phase is the reorg phase. Considering the problems that can significantly improve DDL performance and TiDB resource utilization, we will implement distributed processing in the DDL reorg phase.

Feature branch: [distribute-reorg](https://github.com/pingcap/tidb/tree/feature/distribute-reorg)

##### Tech design
- [x] #37129
##### Development Plan
- Step 1（sprint 11-12） 
  - [x] #37120
  - [x] #37122
  - [x] #37124
  - [x]  Do demo and simple performance test

- Step 2 （sprint 13）
  - [x] #37123
  - [x] #37125
    - [x] Lighting accelerates Add Index adaptation
  - [ ] #37172
  - [x] #37127
  - [x] #41002
  - [x] #41144
  - [x] #41266
  - [ ] Test-related work
  
- Step 3（sprint 14）
  - [ ] #40298
  - [ ] #40933
  - [ ] Support distributed processing `LitMerge` in the `BackfillStateMerging`
  - [ ] Compatible with `admin cancel ddl jobs`
  - [ ] Compatible with the correct error and warning messages for DDL
  - [ ] Improved sync mode of txn-merge and lit-merge
  - [ ] Compatible with DDL retry methods (control retries based on `tidb_ddl_error_count_limit`)
  - [ ] Specification dist-reorg related log
  - [ ] Performance tuning
  - [ ] #37128
  - [ ] #38686
 
 - Step 4
  - Backfill processes the Backfill job scheduling policy, `Admin show DDL jobs` supports reorg progress display, get rid of the DDL owner design, etc.

- Related Issues
  - [x] #41106
  - [x] #41208
  - [x] https://github.com/pingcap/tidb/issues/41371
  - [x] #41576
  - [x] #42176
  - [x] #42199
- Unified table structure related issues
  - [x] https://github.com/pingcap/tidb/issues/41579
  - [x] https://github.com/pingcap/tidb/issues/42175

***Note***
[...]

PR #41787, opened 2023-02-28
br/lightning: change `KvPair`'s row ID type from `int64` to `[]bytes`

tangenta:

Issue Number: ref #37119

Problem Summary:

In distributed execution framework, adding index job is split into multiple sub-tasks. Different TiDB instance may pick different sub-tasks up and execute them. Some of the sub-tasks may be rescheduled to another TiDB instance if the previous one halted due to some reasons.

In the perspective of the TiDB instance, it must be able to solve the reentrancy problem for the same range. Otherwise, an unexpected "duplicate key" error is reported even if there is no duplicate row actually.

To detect the duplicate key, lightning provides a field `RowID(int64)` to identify the row in the data source, appends it to the key slice and stores the key slice to the local engine. Before importing the data to TiKV, the iterator reads the keys,  strips their row IDs, and compares the keys to decide whether they are duplicate. 

Thus, it is the caller responsibility to guarantee that the same index key has the same `RowID`. For lightning, it is natural to use the "file offset" as the `RowID`. However, `int64` is not enough to identify a row for adding index.

This PR changes the `RowID` type to `[]byte` so that the handles(including int handles and common handles) can be used as the `RowID`.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


br/pkg/lightning/backend/kv/sql2kv.go
```diff
@@ -433,7 +434,7 @@ func (kvcodec *tableKVEncoder) Encode(
 	}
 	kvPairs := kvcodec.se.takeKvPairs()
 	for i := 0; i < len(kvPairs.pairs); i++ {
-		kvPairs.pairs[i].RowID = rowID
+		kvPairs.pairs[i].RowID = kv.IntHandle(rowID).Encoded()
```
lance6716, line 437: maybe directly encode a [8]byte from int64. This is a hot path, performance may matter

br/pkg/lightning/backend/local/key_adapter.go
```diff
@@ -46,36 +47,41 @@ func reallocBytes(b []byte, n int) []byte {
...
 type dupDetectKeyAdapter struct{}
 
-func (dupDetectKeyAdapter) Encode(dst []byte, key []byte, rowID int64) []byte {
+func (dupDetectKeyAdapter) Encode(dst []byte, key []byte, rowID []byte) []byte {
 	dst = codec.EncodeBytes(dst, key)
-	dst = reallocBytes(dst, 8)
-	n := len(dst)
-	dst = dst[:n+8]
-	binary.BigEndian.PutUint64(dst[n:n+8], codec.EncodeIntToCmpUint(rowID))
+	dst = reallocBytes(dst, len(rowID)+2)
```
lance6716, line 68: what's the length of longest handle?

tangenta: It's max value of uint16(0xFFFF):

https://github.com/pingcap/tidb/blob/bc95a4f02270e10ccd0dba39b1c0513528bb2eba/tablecodec/tablecodec.go#L1635-L1641

br/pkg/lightning/backend/kv/sql2kv.go
tangenta: Updated.

br/pkg/lightning/backend/kv/sql2kv.go
```diff
@@ -433,7 +434,8 @@ func (kvcodec *tableKVEncoder) Encode(
 	}
 	kvPairs := kvcodec.se.takeKvPairs()
 	for i := 0; i < len(kvPairs.pairs); i++ {
-		kvPairs.pairs[i].RowID = rowID
+		var encoded [8]byte
+		kvPairs.pairs[i].RowID = codec.EncodeInt(encoded[:0], rowID)
```
gozssky, line 438: Is there any memcomparable varint-encoded format of the int64 rowID?  Since we will add extra two bytes in `keyAdapter`, it would be better if we can save some bytes here.

br/pkg/lightning/backend/local/key_adapter.go
```diff
@@ -46,36 +47,41 @@ func reallocBytes(b []byte, n int) []byte {
...
 	return len(key)
 }
 
 var _ KeyAdapter = noopKeyAdapter{}
 
 type dupDetectKeyAdapter struct{}
 
-func (dupDetectKeyAdapter) Encode(dst []byte, key []byte, rowID int64) []byte {
+func (dupDetectKeyAdapter) Encode(dst []byte, key []byte, rowID []byte) []byte {
 	dst = codec.EncodeBytes(dst, key)
```
gozssky, line 67: I wonder if we can skip encoding the key first, since in all cases, the key is already memcomparable.

br/pkg/lightning/backend/kv/sql2kv.go
gozssky: Seems `codec.EncodeComparableVarint` is okay for this.

tangenta: Updated

br/pkg/lightning/backend/local/key_adapter.go
tangenta: Updated

lance6716: Did you forget to remove `dst = codec.EncodeBytes(dst, key)`?

br/pkg/lightning/backend/kv/sql2kv.go
```diff
@@ -433,7 +433,8 @@ func (kvcodec *tableKVEncoder) Encode(
 	}
 	kvPairs := kvcodec.se.takeKvPairs()
 	for i := 0; i < len(kvPairs.pairs); i++ {
-		kvPairs.pairs[i].RowID = rowID
+		var encoded [8]byte
```
lance6716, line 436:
```suggestion
		var encoded [9]byte
```

lance6716 (commented):
rest lgtm

br/pkg/lightning/backend/local/key_adapter.go
tangenta: No, I reverted this change because a lot of tests failed and @gozssky is not quite sure about if this can be removed.

lance6716: Seems we can introduce `type EncodedKey []byte` in future, then we will have more confidence to know if it's encoded already

lance6716 (approved):

gozssky (approved):

Issue #41471, opened 2023-02-16
txn: record long running transaction and session information for diagnosis


cfzjywxk:
## Enhancement

The `safepoint` calculation is the first step for [garbage collection](https://docs.pingcap.com/tidb/stable/garbage-collection-overview) in TiDB, data older than the `safepoint` could be safely reclaimed. The `safepoint` must be less than the global minimal `start_ts` among all the tidb sessions, so a session with long-running transactions may block the advancement of `safepoint`. For example, if the user starts a transaction but never commits, the `safepoint` could not be pushed and data reclamation is blocked. To make the diagnosis of such issues easier, consider the following improvements in TiDB:

1. Record the long-running transaction and session information in the expensive logs. If the `safepoint` is blocked by some ongoing sessions, the `start_ts` could be used to locate the blocking session.
2. Record ongoing transaction duration in the metrics with a certain flushing interval, so the long-running transitions could be viewed directly on the grafana panels.

PR #41864, opened 2023-03-02
txn: Slow txn log

### LINKED ISSUE #41863: Log slow transactions
## Enhancement
We already have slow logs for queries. It makes sense to log slow transactions to help identify and diagnose slow transactions, especially those due to long idle time.

ekexium:

Issue Number: close #41863 

Problem Summary:

Add a session variable `tidb_slow_txn_log_threshold`. Log transaction events observed by the transaction manager of any transaction that takes longer than the threshold. Setting it to 0 means disabling the log.

The log looks like
```log
[2023/03/02 16:12:19.644 +08:00] [INFO] [txnmanager.go:194] ["slow transaction"] [duration=1.5955ms] [conn=1] [txnStartTS=439810717028974592] [events="[\"{event=\\\"enter txn\\\",gap=0s}\",\"{event=\\\"update pessimistic set v = 2 where v = 1\\\",gap=62.458µs}\",\"{event=\\\"stmt commit\\\",gap=1.157125ms}\",\"{event=\\\"txn end\\\",gap=375.167µs}\"]"]
```
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

- [x] Performance regression: Consumes more CPU

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Add a session variable `tidb_slow_txn_log_threshold` to log slow transactions.
```


cfzjywxk:
Is this change intended to resolve issues like https://github.com/pingcap/tidb/issues/41471, if so we may need find out a way to show the long-running transactions like `show processlist` does.

ekexium:
> Is this change intended to resolve issues like #41471, if so we may need find out a way to show the long-running transactions like `show processlist` does.

No. I'd rather treat them as different problems. This PR is a counterpart of slow queries. I intend to use it as a tool to diagnose tail latencies. Though finished long-running transaction could logged by it as well.

ekexium:
> How to diagnose tail latencies using the long-txn log, I don't quite get it. Could you give an example of it?

For example we see high tail latencies for transactions and high max idle durations in metrics. But we cannot guarantee that the tail latency is caused by idle duration.
With slow txn log like 
```
,\"{event=\\\"stmt end\\\",gap=18.89µs}\",\"{event=\\\"insert bs_accum_lmt_detail (haveAccumCnt,haveAccumLmt,lastLmtDt,lmtNo,lmtObj,lmtPeriodType,lmtSubSeqNo,rgstBrc,rgstDt,rgstOperr,rgstTm,transSrlNo) values (?,?,?,?,?,?,?,?,?,?,?,?)\\\",gap=218.345562ms}\",
```
We can tell that there is an idle duration of 218ms before the INSERT stmt comes. And it might help further diagnosis in client side.

session/session.go
```diff
@@ -2175,6 +2175,7 @@ func (s *session) ExecuteStmt(ctx context.Context, stmtNode ast.StmtNode) (sqlex
 	cmd32 := atomic.LoadUint32(&s.GetSessionVars().CommandValue)
 	s.SetProcessInfo(stmtNode.Text(), time.Now(), byte(cmd32), 0)
 	s.txn.onStmtStart(digest.String())
+	defer sessiontxn.GetTxnManager(s).OnStmtEnd()
```
cfzjywxk, line 2178: I tried to merge the two `OnStmtEnd` into one https://github.com/pingcap/tidb/pull/41122 but it looks difficult..

util/logutil/log.go
```diff
@@ -38,6 +38,8 @@ const (
 	DefaultLogFormat = "text"
 	// DefaultSlowThreshold is the default slow log threshold in millisecond.
 	DefaultSlowThreshold = 300
+	// DefaultSlowTxnThreshold is the default slow txn log threshold in ms.
+	DefaultSlowTxnThreshold = 0
```
cfzjywxk, line 42: How to choose a default-enabled value?

session/txnmanager.go
```diff
@@ -153,12 +174,29 @@ func (m *txnManager) EnterNewTxn(ctx context.Context, r *sessiontxn.EnterNewTxnR
...
 func (m *txnManager) OnTxnEnd() {
 	m.ctxProvider = nil
 	m.stmtNode = nil
+
+	m.events = append(m.events, event{event: "txn end", duration: time.Since(m.lastInstant)})
+
+	duration := time.Since(m.enterTxnInstant)
+	threshold := m.sctx.GetSessionVars().SlowTxnThreshold
+	if threshold > 0 && uint64(duration.Milliseconds()) >= threshold {
+		logutil.BgLogger().Info(
```
cfzjywxk, line 191: Print connection id if possible is better as `BgLogger` is used.

session/txnmanager.go
```diff
@@ -222,14 +275,33 @@ func (m *txnManager) OnStmtCommit(ctx context.Context) error {
...
+	m.lastInstant = time.Now()
+}
+
+func (m *txnManager) resetEvents() {
+	if m.events == nil {
+		m.events = make([]event, 0, 10)
+	} else {
+		m.events = m.events[:0]
+	}
+	m.enterTxnInstant = m.lastInstant
```
zyguan, line 296: Why set `enterTxnInstant` as `lastInstant`? Should it be set as `time.Now()` on entering new txn?

session/txnmanager.go
```diff
@@ -52,6 +57,22 @@ type txnManager struct {
 
 	// We always reuse the same OptimisticTxnContextProvider in one session to reduce memory allocation cost for every new txn.
 	reservedOptimisticProviders [2]isolation.OptimisticTxnContextProvider
+
+	// used for slow transaction logs
+	events          []event
+	lastInstant     time.Time
+	enterTxnInstant time.Time
```
zyguan, line 64: Maybe `lastInstant` is unnecessary, we can record the start instant and the duration (since the start instant) of each event, then the gaps can be inferred from these info.

zyguan (commented):
Note that query_dur > stmt_dur, eg. {event="select ...",gap=100ms} may be caused by slow parsing, we cannot confirm that it's caused by client idle time without the slow query log of that query.

util/logutil/log.go
ekexium: Default to 0. It's totally dependent on the application logic to determine what is "slow". And defaulting to 0 will prevent noisy logs from inner transactions.

session/txnmanager.go
ekexium: Connection id has been included

session/txnmanager.go
```diff
@@ -172,9 +210,24 @@ func (m *txnManager) OnStmtStart(ctx context.Context, node ast.StmtNode) error {
...
 	}
+
+	var sql string
+	if node != nil {
+		sql = node.OriginalText()
+		if m.sctx.GetSessionVars().EnableRedactLog {
+			sql = parser.Normalize(sql)
+		}
+	}
+	m.resetEvents()
```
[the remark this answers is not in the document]
ekexium: Oh my mistake.. It should be in `EnterNewTxn`

session/txnmanager.go
ekexium: Yeah I've considered the alternative. I think they are basically equivalent and either is fine. If you prefer this style I can change it

ekexium:
> Note that query_dur > stmt_dur, that is `{event="select ...",gap=100ms}` may be caused by slow parsing, we cannot confirm that it's caused by client idle time without the slow query log of that query.

That's true. And the stat doesn't count the time spent in TiDB _after_ `OnTxnEnd`. Precisely it's counting from the "txn manager's view". But that helps, too. It's better than nothing. What do you think?

In real cases we should use it together with slow query logs and stmt info. For things like `{event="select ...",gap=100ms}` we can almost assure it's an idle duration since long parsing/compilation is kinda rare.

cfzjywxk:
BTW I think we've added some "event tracking" things in the consistency check enhancement, should we also include the `inconsistency` events and merge them into a single event tracking framework.

session/txnmanager.go
zyguan: I think it’s fine. A field of type time.Time here is not expensive at all.

zyguan (approved):

cfzjywxk (approved):

cfzjywxk:
@ekexium 
Need to resolve the conflicts.

PR #41866, opened 2023-03-02
ttl: fix upgrade from 6.5 to 6.6 configuration default value

### LINKED ISSUE #41865: ttl upgrade from 6.5 to 6.6. will give unexpected job interval
If the user create an TTL table in 6.5, and upgrade the cluster to 6.6. The TTL job will continue running because the `TTL_JOB_INTERVAL` was initialized to 0.

YangKeao:

Issue Number: close #41865

If the value of `TTL_JOB_INTERVAL` is not set, use `1h` instead. 

Actually the `TTL_JOB_INTERVAL` will be set to `1h` during the DDL. However, if the user created the table at 6.5 and upgrade the cluster to 6.6, the attribute is empty.

I'm not sure whether this is the best (and most suggested) way to fix this problem. If you have more suggestion, feel free to point out.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

### Release note

```release-note
Fix the issue that an TTL table upgraded from old version could have 0 job interval and the TTl job will keep running.
```

### DIFF
--- a/executor/show.go
+++ b/executor/show.go
@@ -1274,7 +1274,11 @@ func ConstructResultOfShowCreateTable(ctx sessionctx.Context, tableInfo *model.T
 		err = restoreCtx.WriteWithSpecialComments(tidb.FeatureIDTTL, func() error {
 			restoreCtx.WriteKeyWord("TTL_JOB_INTERVAL")
 			restoreCtx.WritePlain("=")
-			restoreCtx.WriteString(tableInfo.TTLInfo.JobInterval)
+			if len(tableInfo.TTLInfo.JobInterval) == 0 {
+				restoreCtx.WriteString(model.DefaultJobInterval.String())
+			} else {
+				restoreCtx.WriteString(tableInfo.TTLInfo.JobInterval)
+			}
 			return nil
 		})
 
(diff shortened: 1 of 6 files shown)


executor/show.go
```diff
@@ -1274,7 +1274,11 @@ func ConstructResultOfShowCreateTable(ctx sessionctx.Context, tableInfo *model.T
 		err = restoreCtx.WriteWithSpecialComments(tidb.FeatureIDTTL, func() error {
 			restoreCtx.WriteKeyWord("TTL_JOB_INTERVAL")
 			restoreCtx.WritePlain("=")
-			restoreCtx.WriteString(tableInfo.TTLInfo.JobInterval)
+			if len(tableInfo.TTLInfo.JobInterval) == 0 {
+				restoreCtx.WriteString("1h")
```
bb7133, line 1278: Can we introduce something like `DefaultTTLJobInterval`?

YangKeao: Good!

YangKeao: I declare this const value near the `TTLInfo` struct in `model` :+1:

lcwangchao (approved):

hawkingrei (approved):

YangKeao:
Really weird :scream:

Issue #41720, opened 2023-02-24
  spool: a simple goroutine pool 


hawkingrei:
## Enhancement

Spool is the simplest pool. It is used to control concurrency.

Here is the main interface.
```
type Pool interface {
    // adjust concurrency
    Tuner(concurrency int)        
    
    // Run a job if the submission exceeds the configuration, 
    // if the block is true, it will block and wait for resources. 
    // otherwise, it will report an error
    Run(fn func()) error 
    RunWithConcurrency(fn func(), workerCnt cnt) error
    
    // Release all worker and wait for all exiting.
    ReleaseAndWait()
}
```

But it is not a goroutine pool. so you can not get better performance from it.

PR #41908, opened 2023-03-03
resourcemange: spool support task manage

hawkingrei:

Issue Number: ref https://github.com/pingcap/tidb/issues/41720

Problem Summary:

Spool is the simplest pool. It is used to control concurrency. But it cannot reuse the goroutine.

Here is the main interface.
```
type Pool interface {
    // adjust concurrency
    Tuner(concurrency int)        
    
    // Run a job if the submission exceeds the configuration, 
    // if the block is true, it will block and wait for resources. 
    // otherwise, it will report an error
    Run(fn func()) error 
    RunWithConcurrency(fn func(), workerCnt cnt) error
    
    // Release all worker and wait for all exiting.
    ReleaseAndWait()
}
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


resourcemanager/pool/spool/spool.go
```diff
@@ -95,7 +107,6 @@ func (p *Pool) Run(fn func()) error {
 	if !run {
 		return pool.ErrPoolOverload
 	}
-	p.taskManager.RegisterTask(p.GenTaskID(), 1)
```
hawkingrei, line 98: a simple task will not be controlled the concurrency.

resourcemanager/poolmanager/task_manager_iterator.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+	}
+	return nil, false
+}
+
+func canBoost(m *Meta, min time.Time) (result *Meta, isBreak bool) {
+	if m.running.Load() < m.initialConcurrency {
+		box := m
+		if box != nil {
+			return box, true
```
you06, line 95: Why not just return `m`?
```go
if m != nil {
	return m
}
```

resourcemanager/poolmanager/task_manager_iterator.go
```diff
@@ -0,0 +1,105 @@
...
+					result = newResult
+					tid = id
+					compareTS = stats.createTS
+					return true
+				}
+				if newResult != nil {
+					result = newResult
+					tid = id
+					compareTS = stats.createTS
+				}
```
you06, line 66: Can be simplified.

```suggestion
				if newResult != nil {
					result = newResult
					tid = id
					compareTS = stats.createTS
				}
				if pauseFind {
					return true
				}
```

resourcemanager/pool/spool/spool.go
```diff
@@ -79,8 +79,20 @@ func (p *Pool) Tune(size int32) {
...
+		_, task := p.taskManager.Overclock()
+		if task != nil {
+			p.running.Add(1)
+			p.run(func() {
+				runTask(task)
+			})
+		}
+		return
+	}
+	p.taskManager.Downclock()
```
you06, line 95: A question about overclock and downclock here, when calling `p.taskManager.Overclock()` or `p.taskManager.Downclock()`, it adds or reduces one task, but if $\left|old - size\right| > 1$,  should we call them multi times?

resourcemanager/pool/spool/spool.go
```diff
@@ -79,8 +79,20 @@ func (p *Pool) Tune(size int32) {
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	p.SetLastTuneTs(time.Now())
+	old := p.capacity
 	p.capacity = size
 	p.concurrencyMetrics.Set(float64(size))
+	if old < size {
```
you06, line 85: If `old == size`, then do nothing?

resourcemanager/poolmanager/task_manager_iterator.go
hawkingrei: some cases have more high priority. we need to exit in advance.

for example, when we need to overclock, if we find a task has downclocked the concurrency before, we will exit in advance and overclock for it.

resourcemanager/poolmanager/task_manager_iterator.go
hawkingrei: Done

resourcemanager/pool/spool/spool.go
hawkingrei: the scheduler is a simple goroutine so we cannot call them multi-time at some times. Overclock is to create an new goroutine and downclock is to decrease one goroutine. | old- size| > 1 is impossible, all the concurrency is scheduled by resource management.

resourcemanager/pool/spool/spool.go
hawkingrei: Yes， I refactor it. Now if old == size, return directly.

resourcemanager/poolmanager/task_manager_iterator.go
you06: What I mean is just simplifying the code.

you06:
This PR is generally LGTM, I have another question about the invoke of the Tune function.

When `ResourceManager.schedule` is called, it iterates over the shared pool maps and decides to overclock or downclock by the CPU usage. There are 8 shards in TiDB, which means in most cases, 8 tasks are being boosted or paused at once, I'm not sure whether it's under design.

hawkingrei:
> There are 8 shards in TiDB, which means in most cases, 8 tasks are being boosted

we decide to use a shard map for managing the data. It can avoid frequent read-write races. Because we iter it while new task need to insert into this map.

you06 (approved):

resourcemanager/pool/spool/spool.go
wjhuang2016: | old- size| > 1 is possible， the Tune() may be not called from the resource management.

resourcemanager/poolmanager/task_manager_iterator.go
```diff
@@ -0,0 +1,102 @@
...
+		box := m
+		if box != nil {
+			return box, true
+		}
+	}
+	if m.createTS.After(min) {
+		box := m
+		if box != nil {
+			return box, false
+		}
```
tangenta, line 99: What does the "box" mean?

hawkingrei: It is a name. It looks like Box in Rust. It is just a container to save copy object.

resourcemanager/poolmanager/task_manager_iterator.go
```diff
@@ -0,0 +1,102 @@
...
+			for id, stats := range t.task[i].stats {
+				if result == nil {
+					if stats.running.Load() != 0 {
+						result = stats
+					}
+					tid = id
+					compareTS = stats.createTS
+					continue
+				}
+				newResult, pauseFind := fn(stats, compareTS)
```
lcwangchao, line 55: Is that make sense that  the `newResult != stats` while `stats != nil`. If not I think fn returns `bool, bool` is better

resourcemanager/poolmanager/task_manager_iterator.go
```diff
@@ -0,0 +1,102 @@
...
+			return false
+		}(shard)
+		if breakFind {
+			break
+		}
+	}
+	return tid, result
+}
+
+func canPause(m *Meta, max time.Time) (result *Meta, isBreak bool) {
```
lcwangchao, line 74: Is the second argument means min or max? I think it filters the meta with the min createTS

resourcemanager/poolmanager/task_manager_iterator.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	if m.createTS.Before(max) {
+		if m.running.Load() != 0 {
+			return m, true
+		}
+	}
+	return nil, false
+}
+
+func canBoost(m *Meta, min time.Time) (result *Meta, isBreak bool) {
```
lcwangchao, line 88: ditto

resourcemanager/poolmanager/task_manager_iterator.go
```diff
@@ -0,0 +1,102 @@
...
+		if breakFind {
+			break
+		}
+	}
+	return tid, result
+}
+
+func canPause(m *Meta, max time.Time) (result *Meta, isBreak bool) {
+	if m.initialConcurrency < m.running.Load() {
+		if m.running.Load() != 0 {
```
lcwangchao, line 76: I think it is always true when `m. initialConcurrency  >= 0`

resourcemanager/pool/spool/spool.go
```diff
@@ -79,8 +79,23 @@ func (p *Pool) Tune(size int32) {
...
+		_, task := p.taskManager.Overclock()
+		if task != nil {
+			p.running.Add(1)
+			p.run(func() {
+				runTask(task)
+			})
+		}
+		return
+	}
+	p.taskManager.Downclock()
```
lcwangchao, line 98: Another question is do we need to call `Downclock` async because it can be blocked by channel and what is worse, `p.mu` is locked when it happens.

resourcemanager/pool/spool/spool.go
```diff
@@ -197,3 +210,21 @@ func (p *Pool) ReleaseAndWait() {
...
+	exitCh := task.GetExitCh()
+	task.IncTask()
+	defer task.DecTask()
+	for {
+		select {
+		case f, ok := <-taskCh:
+			if !ok {
+				return
+			}
+			f()
```
lcwangchao, line 225: Seems if `f()` runs a long time, the goroutine's shrink will be slow because it will waiting for `f()` exit. Is it better to pass a `Context` to f and then the goroutine is about to exit, the function can be notified.

resourcemanager/poolmanager/task_manager_iterator.go
hawkingrei: You are right. It is unnecessary to copy it.

resourcemanager/poolmanager/task_manager_iterator.go
hawkingrei: CanPause should be min. The less timestamp, the more possible to pause.
CanBoost should be max. The more timestamp, the more possible to boost.

Fix.

resourcemanager/poolmanager/task_manager_iterator.go
```diff
@@ -0,0 +1,102 @@
...
+			return m, true
+		}
+	}
+	if m.createTS.Before(max) {
+		if m.running.Load() != 0 {
+			return m, true
```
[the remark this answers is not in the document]
hawkingrei: Done

resourcemanager/pool/spool/spool.go
hawkingrei: so It is the reason that it has a shard map to avoid frequent lock races. and His lock fine-grained is very small. it doesn't take too much time on it.

resourcemanager/poolmanager/task_manager_iterator.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	return false, false
+}
+
+func canBoost(m *Meta, max time.Time) (find bool, toBreakFind bool) {
+	if m.running.Load() < m.initialConcurrency {
+		if m != nil {
+			return true, true
+		}
+	}
```
tangenta, line 93: `m` is never equal to `nil` because you have used `m.running`.

resourcemanager/poolmanager/task_manager_iterator.go
hawkingrei: Done

resourcemanager/pool/spool/spool.go
hawkingrei: I agree with you. We can solve it in the next PR.

resourcemanager/pool/spool/spool.go
hawkingrei: I know. If I find that it is blocked from sending a msg to exitChan, I will skip it.

resourcemanager/poolmanager/task_manager_iterator.go
hawkingrei: so Downclock or Overclock can deal with a goroutine every time to avoid too much influence for this case

resourcemanager/pool/spool/spool.go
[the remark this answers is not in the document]
hawkingrei: No, I add some conditional judgments to avoid this situation.

resourcemanager/poolmanager/task_manager_iterator.go
hawkingrei: removed.

hawkingrei:
@lcwangchao @tangenta PTAL

lcwangchao (approved):

hawkingrei:
@tangenta PTAL

resourcemanager/pool/spool/spool_test.go
```diff
@@ -192,3 +188,30 @@ func TestRunWithNotEnough2(t *testing.T) {
...
+	time.Sleep(100 * time.Microsecond)
+	require.Equal(t, int32(1), p.Running())
+
+	// increase the concurrency
+	p.Tune(2)
+	time.Sleep(100 * time.Microsecond)
+	require.Equal(t, int32(2), p.Running())
+	p.Tune(3)
+	time.Sleep(100 * time.Microsecond)
+	require.Equal(t, int32(3), p.Running())
```
tangenta, line 207: Will this introduce unstable test?

hawkingrei: Maybe, So I use ```require.Eventually``` to replace.

resourcemanager/poolmanager/task_manager.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// GetExitCh is to get the exit channel.
+func (m *Meta) GetExitCh() chan struct{} {
+	return m.exitCh
+}
+
+// TaskStatusContainer is a container that can control or watch the pool.
+type TaskStatusContainer struct {
+	stats map[uint64]*Meta
```
zimulala, line 83: s/stats/metas ?

resourcemanager/poolmanager/task_manager.go
```diff
@@ -0,0 +1,135 @@
...
+// DeleteTask delete a task from the manager.
+func (t *TaskManager) DeleteTask(taskID uint64) {
+	shardID := getShardID(taskID)
+	t.task[shardID].rw.Lock()
+	delete(t.task[shardID].stats, taskID)
+	t.task[shardID].rw.Unlock()
+}
+
+// hasTask check if the task is in the manager.
+func (t *TaskManager) hasTask(taskID uint64) bool {
```
zimulala, line 124: We don't use this function, could we remove it?

PR #41992, opened 2023-03-07
executor: fix global index scan, index join, and query filtering reorg partition for global index

### LINKED ISSUE #41991: executor: global index scan returns wrong result when specify partition
## Bug Report

premise:
set enable-global-index = true  in config/config.toml.example

SQL: 
```
use test
drop table if exists p;
create table p (id int, c int) partition by range (c) (
    partition p0 values less than (4),
    partition p1 values less than (7),
    partition p2 values less than (10));
alter table p add unique idx(id);
insert into p values (1,3), (3,4), (5,6), (7,9);
select id from p partition(p0) use index (idx);
```

1

1
3
5
7
image

----+
| Release Version: v6.7.0-alpha-139-g6c1674cb1-dirty
Edition: Community
Git Commit Hash: 6c1674cb1e75eb23876efae6b43be7f3fa149c9c
Git Branch: local/fix_truncate_drop_partition_for_global_index
UTC Build Time: 2023-03-02 03:19:49
GoVersion: go1.19.4
Race Enabled: false
TiKV Min Version: 6.2.0-alpha
Store: unistore |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

### LINKED ISSUE #42065: executor: global index join returns wrong result
## Bug Report

set `enable-global-index=true` in config/config.toml.example

```
create table t1 (id int, c int) partition by range (c) (
partition p0 values less than (4),
partition p1 values less than (7),
partition p2 values less than (10));
alter table t1 add unique idx(id);
insert into t1 values (1,3), (3,4), (5,6), (7,9);
create table t2 (id int, c int);
insert into t2 values (1, 3);
select /*+ INL_JOIN(t1, t2) */ * from t1 partition(p0) inner join t2 on t1.id = t2.id;
select /*+ INL_JOIN(t1, t2) */ t1.id from t1 partition(p0) inner join t2 on t1.id = t2.id;
```

The first select result: 1 3 1 3
The second select result:  1

The first select result: empty set
The second select result:  empty set
image

                                                                |
+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Release Version: v6.7.0-alpha-205-g29adb0aa8-dirty
Edition: Community
Git Commit Hash: 29adb0aa86705d4728e1c9fd28a3afdc705c969a
Git Branch: HEAD
UTC Build Time: 2023-03-09 09:40:00
GoVersion: go1.19.4
Race Enabled: false
TiKV Min Version: 6.2.0-alpha
Store: unistore |
+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

L-maple:

Issue Number: close #41991
Issue Number: close #42065

Problem Summary:

- fix global index scan when specify partition
- fix global index join
- fix drop/truncate partition reorg for global index

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/ddl/db_partition_test.go
+++ b/ddl/db_partition_test.go
@@ -2053,6 +2053,79 @@ func TestGlobalIndexUpdateInDropPartition(t *testing.T) {
 	tk.MustQuery("select * from test_global use index(idx_b) order by a").Check(testkit.Rows("2 11 11", "12 12 12"))
 }
 
+func TestGlobalIndexReaderInDropPartition(t *testing.T) {
+	defer config.RestoreFunc()()
+	config.UpdateGlobal(func(conf *config.Config) {
+		conf.EnableGlobalIndex = true
+	})
+	store, dom := testkit.CreateMockStoreAndDomain(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("drop table if exists test_global")
+	tk.MustExec(`create table test_global ( a int, b int, c int)
+	partition by range( a ) (
+		partition p1 values less than (10),
+		partition p2 values less than (20),
+		partition p3 values less than (30)
+	);`)
+	tk.MustExec("alter table test_global add unique index idx_b (b);")
+	tk.MustExec("insert into test_global values (1, 1, 1), (8, 8, 8), (11, 11, 11), (12, 12, 12);")
+
+	var indexScanResult *testkit.Result
+	hook := &callback.TestDDLCallback{Do: dom}
+	hook.OnJobRunBeforeExported = func(job *model.Job) {
+		assert.Equal(t, model.ActionDropTablePartition, job.Type)
+		if job.SchemaState == model.StateDeleteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+
+			indexScanResult = tk1.MustQuery("select b from test_global use index(idx_b)").Sort()
+		}
+	}
+	dom.DDL().SetHook(hook)
+
+	tk.MustExec("alter table test_global drop partition p1")
+
+	indexScanResult.Check(testkit.Rows("11", "12"))
+}
+
+func TestGlobalIndexLookUpInDropPartition(t *testing.T) {
+	defer config.RestoreFunc()()
+	config.UpdateGlobal(func(conf *config.Config) {
+		conf.EnableGlobalIndex = true
+	})
+
+	store, dom := testkit.CreateMockStoreAndDomain(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("drop table if exists test_global")
+	tk.MustExec(`create table test_global ( a int, b int, c int)
+	partition by range( a ) (
+		partition p1 values less than (10),
+		partition p2 values less than (20),
+		partition p3 values less than (30)
+	);`)
+	tk.MustExec("alter table test_global add unique index idx_b (b);")
+	tk.MustExec("insert into test_global values (1, 1, 1), (8, 8, 8), (11, 11, 11), (12, 12, 12);")
+
+	var indexLookupResult *testkit.Result
+	hook := &callback.TestDDLCallback{Do: dom}
+	hook.OnJobRunBeforeExported = func(job *model.Job) {
+		assert.Equal(t, model.ActionDropTablePartition, job.Type)
+		if job.SchemaState == model.StateDeleteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+			tk1.MustExec("analyze table test_global")
+			indexLookupResult = tk1.MustQuery("select * from test_global use index(idx_b)").Sort()
+		}
+	}
+	dom.DDL().SetHook(hook)
+
+	tk.MustExec("alter table test_global drop partition p1")
+
+	indexLookupResult.Check(testkit.Rows("11 11 11", "12 12 12"))
+}
+
 func TestAlterTableExchangePartition(t *testing.T) {
 	store := testkit.CreateMockStore(t)
 	tk := testkit.NewTestKit(t, store)
--- a/executor/builder.go
+++ b/executor/builder.go
@@ -3765,6 +3765,21 @@ func (b *executorBuilder) buildIndexReader(v *plannercore.PhysicalIndexReader) E
 	}
 
 	if is.Index.Global {
+		tmp, ok := b.is.TableByID(ret.table.Meta().ID)
+		if !ok {
+			b.err = infoschema.ErrTableNotExists
+			return nil
+		}
+		tbl, ok := tmp.(table.PartitionedTable)
+		if !ok {
+			b.err = exeerrors.ErrBuildExecutor
+			return nil
+		}
+		ret.partitionIDMap, err = getPartitionIdsAfterPruning(b.ctx, tbl, &v.PartitionInfo)
+		if err != nil {
+			b.err = err
+			return nil
+		}
 		return ret
 	}
 
@@ -3964,8 +3979,8 @@ func (b *executorBuilder) buildIndexLookUpReader(v *plannercore.PhysicalIndexLoo
 			b.err = err
 			return nil
 		}
-		tbl, ok1 := tmp.(table.PartitionedTable)
-		if !ok1 {
+		tbl, ok := tmp.(table.PartitionedTable)
+		if !ok {
 			b.err = exeerrors.ErrBuildExecutor
 			return nil
 		}
@@ -4495,6 +4510,30 @@ func (builder *dataReaderBuilder) buildIndexReaderForIndexJoin(ctx context.Conte
 		return e, err
 	}
 
+	is := v.IndexPlans[0].(*plannercore.PhysicalIndexScan)
+	if is.Index.Global {
+		tmp, ok := builder.is.TableByID(tbInfo.ID)
+		if !ok {
+			return nil, infoschema.ErrTableNotExists
+		}
+		tbl, ok := tmp.(table.PartitionedTable)
+		if !ok {
+			return nil, exeerrors.ErrBuildExecutor
+		}
+		e.partitionIDMap, err = getPartitionIdsAfterPruning(builder.ctx, tbl, &v.PartitionInfo)
+		if err != nil {
+			return nil, err
+		}
+
+		if e.ranges, err = buildRangesForIndexJoin(e.ctx, lookUpContents, indexRanges, keyOff2IdxOff, cwc); err != nil {
+			return nil, err
+		}
+		if err := e.Open(ctx); err != nil {
+			return nil, err
+		}
+		return e, nil
+	}
+
 	tbl, _ := builder.executorBuilder.is.TableByID(tbInfo.ID)
 	usedPartition, canPrune, contentPos, err := builder.prunePartitionForInnerExecutor(tbl, e.Schema(), &v.PartitionInfo, lookUpContents)
 	if err != nil {
@@ -4544,6 +4583,32 @@ func (builder *dataReaderBuilder) buildIndexLookUpReaderForIndexJoin(ctx context
 		err = e.open(ctx)
 		return e, err
 	}
+
+	is := v.IndexPlans[0].(*plannercore.PhysicalIndexScan)
+	ts := v.TablePlans[0].(*plannercore.PhysicalTableScan)
+	if is.Index.Global {
+		tmp, ok := builder.is.TableByID(ts.Table.ID)
+		if !ok {
+			return nil, infoschema.ErrTableNotExists
+		}
+		tbl, ok := tmp.(table.PartitionedTable)
+		if !ok {
+			return nil, exeerrors.ErrBuildExecutor
+		}
+		e.partitionIDMap, err = getPartitionIdsAfterPruning(builder.ctx, tbl, &v.PartitionInfo)
+		if err != nil {
+			return nil, err
+		}
+		e.ranges, err = buildRangesForIndexJoin(e.ctx, lookUpContents, indexRanges, keyOff2IdxOff, cwc)
+		if err != nil {
+			return nil, err
+		}
+		if err := e.Open(ctx); err != nil {
+			return nil, err
+		}
+		return e, err
+	}
+
 	tbl, _ := builder.executorBuilder.is.TableByID(tbInfo.ID)
 	usedPartition, canPrune, contentPos, err := builder.prunePartitionForInnerExecutor(tbl, e.Schema(), &v.PartitionInfo, lookUpContents)
 	if err != nil {
--- a/executor/distsql.go
+++ b/executor/distsql.go
@@ -30,6 +30,7 @@ import (
 	"github.com/pingcap/tidb/distsql"
 	"github.com/pingcap/tidb/expression"
 	"github.com/pingcap/tidb/kv"
+	"github.com/pingcap/tidb/parser/ast"
 	"github.com/pingcap/tidb/parser/charset"
 	"github.com/pingcap/tidb/parser/model"
 	"github.com/pingcap/tidb/parser/mysql"
@@ -173,6 +174,7 @@ type IndexReaderExecutor struct {
 	ranges          []*ranger.Range
 	partitions      []table.PhysicalTable
 	partRangeMap    map[int64][]*ranger.Range // each partition may have different ranges
+	partitionIDMap  map[int64]struct{}        // partitionIDs that global index access
 
 	// kvRanges are only used for union scan.
 	kvRanges         []kv.KeyRange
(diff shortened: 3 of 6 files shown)


Defined2014:
Does the PR ready to be reviewed? Maybe you could remove `[WIP]` in title.

L-maple:
> WIP



> Does the PR ready to be reviewed? Maybe you could remove `[WIP]` in title.

@Defined2014 Hi, could you run /ok-to-test? 
I will contribute a littile more code for drop partition in this PR, which also has bug.

executor/distsql.go
```diff
@@ -239,10 +240,46 @@ func (e *IndexReaderExecutor) Next(ctx context.Context, req *chunk.Chunk) error
...
+	// Global index has the last partitionID column, filter out needed rows and prune the last column.
+	if e.index.Global {
+		req.Reset()
+		retTps := append(e.base().retFieldTypes, types.NewFieldType(mysql.TypeLonglong))
+		if err := e.result.ResetFieldTypes(retTps); err != nil {
+			return err
+		}
+
+		// TODO(L-Maple): chk should be reused
+		chk := chunk.NewChunkWithCapacity(retTps, req.Capacity())
```
mjonss, line 252: Would this be needed if the column was already in the output columns?
Any reason for not including the PartitionID (physical table id) in the output, to avoid yet another pruning call if need to read columns not included in the index?

I assume if we add the ExtraPidColID column to the v.OutputColumns that we also would need a column pruning step in the plan?

executor/distsql.go
```diff
@@ -239,10 +240,46 @@ func (e *IndexReaderExecutor) Next(ctx context.Context, req *chunk.Chunk) error
...
+		}
+
+		for i := 0; i < chk.NumRows(); i++ {
+			row := chk.GetRow(i)
+			pidOffset := row.Len() - 1
+			pid := row.GetInt64(pidOffset)
+			if _, exist := e.partitionIDMap[pid]; !exist {
+				continue
+			}
+
```
mjonss, line 266: Is this the reason for the partitionIDMap, using it for pruning when reading the index, to avoid extra processing?

Could we not also push down that filtering, i.e. to avoid retrieving it from TiKV?

executor/partition_table_test.go
```diff
@@ -3142,6 +3142,28 @@ partition p2 values less than (10))`)
 	tk.MustQuery("select id from p use index (idx) order by id").Check(testkit.Rows("1", "3", "5", "7"))
 }
 
+func TestAggWithGlobalIndex(t *testing.T) {
+	failpoint.Enable("github.com/pingcap/tidb/planner/core/forceDynamicPrune", `return(true)`)
+	defer failpoint.Disable("github.com/pingcap/tidb/planner/core/forceDynamicPrune")
```
mjonss, line 3147: I would prefer to not use these failpoints in new test, but rather control the execution with analyze table etc.

planner/core/physical_plans.go
```diff
@@ -329,6 +330,7 @@ func (p *PhysicalIndexReader) Clone() (PhysicalPlan, error) {
 		return nil, err
 	}
 	cloned.OutputColumns = util.CloneCols(p.OutputColumns)
+	cloned.CommonHandleCols = util.CloneCols(p.CommonHandleCols)
```
mjonss, line 333: Should `PhysicalIndexLookUpReader.Clone()` be similar?

planner/core/exhaust_physical_plans.go
```diff
@@ -2677,6 +2677,11 @@ func (p *baseLogicalPlan) canPushToCopImpl(storeTp kv.StoreType, considerDual bo
 				if len(path.PartialIndexPaths) > 0 {
 					considerIndexMerge = true
 				}
+				// Global index may be filtered for selected partitions in TiDB, so cannot be pushed down now.
+				// TODO(L-Maple): add cop support for global index.
```
mjonss, line 2681: Thanks, relates to a previous question I had above.

statistics/testdata/integration_suite_out.json
```diff
@@ -7,19 +7,19 @@
...
+        "└─Selection_6 2.00 cop[tikv]  eq(test.exp_backoff.c, 1)",
+        "  └─TableFullScan_5 5.00 cop[tikv] table:exp_backoff keep order:false"
       ],
       [
-        "IndexReader_10 3.00 root  index:Selection_9",
-        "└─Selection_9 3.00 cop[tikv]  ge(test.exp_backoff.d, 3), le(test.exp_backoff.d, 5)",
-        "  └─IndexFullScan_8 5.00 cop[tikv] table:exp_backoff, index:idx(a, b, c, d) keep order:false"
+        "TableReader_7 3.00 root  data:Selection_6",
+        "└─Selection_6 3.00 cop[tikv]  ge(test.exp_backoff.d, 3), le(test.exp_backoff.d, 5)",
+        "  └─TableFullScan_5 5.00 cop[tikv] table:exp_backoff keep order:false"
```
mjonss, line 22: Why are these needing to change?

ddl/db_partition_test.go
```diff
@@ -1952,6 +1953,119 @@ func TestDropPartitionWithGlobalIndex(t *testing.T) {
...
+	);`)
+	tk.MustExec("alter table test_global add unique index idx_b (b);")
+	tk.MustExec("insert into test_global values (1, 1, 1), (8, 8, 8), (11, 11, 11), (12, 12, 12);")
+	return store, dom
+}
+
+func TestGlobalIndexReaderInDropPartition(t *testing.T) {
+	t.Skip("not support yet")
+	failpoint.Enable("github.com/pingcap/tidb/planner/core/forceDynamicPrune", `return(true)`)
+	defer failpoint.Disable("github.com/pingcap/tidb/planner/core/forceDynamicPrune")
```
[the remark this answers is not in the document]
L-maple: resolved.

executor/builder.go
```diff
@@ -3695,6 +3695,16 @@ func buildNoRangeIndexReader(b *executorBuilder, v *plannercore.PhysicalIndexRea
...
+	// Global index needs an offset for ExtraPidColID.
+	if e.index.Global {
+		var handleLen int
+		if len(v.CommonHandleCols) != 0 {
+			handleLen = len(v.CommonHandleCols)
+		} else {
+			handleLen = 1
+		}
+		dagReq.OutputOffsets = append(dagReq.OutputOffsets, uint32(len(v.OutputColumns)+handleLen))
+	}
```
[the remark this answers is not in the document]
L-maple: by following your advice, I push down filter logic to tikv, so this change is removed now

executor/partition_table_test.go
L-maple: resolved

planner/core/exhaust_physical_plans.go
L-maple: by following your advice, this logic is removed now.

statistics/testdata/integration_suite_out.json
L-maple: removed. If we push down partitionIds，cost will not change  and modifications above is unnecessary now.

planner/core/physical_plans.go
L-maple: this logic is removed

planner/core/find_best_task.go
```diff
@@ -1588,27 +1590,25 @@ func (is *PhysicalIndexScan) initSchema(idxExprCols []*expression.Column, isDoub
 		}
 	}
 
-	if isDoubleRead {
```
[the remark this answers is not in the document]
L-maple: Now this is necessary. The origin reason: if we have to filter the partitions in TiDB, we have to tell tikv to return ExtraPid. So that we can filter by the returned Pid.

distsql/select_result.go
```diff
@@ -72,6 +72,8 @@ type SelectResult interface {
 	Next(context.Context, *chunk.Chunk) error
 	// Close closes the iterator.
 	Close() error
+	// ResetFieldTypes reset the field types
+	ResetFieldTypes([]*types.FieldType) error
```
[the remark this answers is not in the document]
L-maple: removed.

executor/distsql.go
L-maple: You are right. I found that if we handle filter logic in TiDB, there are at least two problems: 1) performance; 2) lots of tests have to be modified, as cost is changed. 
By following your advice, I propose the tipb PR: https://github.com/pingcap/tipb/pull/297?w=1. Please review it.

go.mod
```diff
@@ -277,5 +277,6 @@ replace (
 	// fix potential security issue(CVE-2020-26160) introduced by indirect dependency.
 	github.com/dgrijalva/jwt-go => github.com/form3tech-oss/jwt-go v3.2.6-0.20210809144907-32ab6a8243d7+incompatible
 	github.com/pingcap/tidb/parser => ./parser
+	github.com/pingcap/tipb => github.com/L-maple/tipb v0.0.0-20230313092931-abd3eac30f1a
```
L-maple, line 280: tipb: https://github.com/pingcap/tipb/pull/297?w=1 have to be merged first.

L-maple: @Defined2014 @mjonss  Please help me review the tipb PR. If you have any questions, please comment and I would reply as quickly as possible, thanks~

ddl/db_partition_test.go
```diff
@@ -1952,6 +1953,119 @@ func TestDropPartitionWithGlobalIndex(t *testing.T) {
...
+	tk.MustExec("drop table if exists test_global")
+	tk.MustExec(`create table test_global ( a int, b int, c int)
+	partition by range( a ) (
+		partition p1 values less than (10),
+		partition p2 values less than (20),
+		partition p3 values less than (30)
+	);`)
+	tk.MustExec("alter table test_global add unique index idx_b (b);")
+	tk.MustExec("insert into test_global values (1, 1, 1), (8, 8, 8), (11, 11, 11), (12, 12, 12);")
+	return store, dom
```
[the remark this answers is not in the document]
L-maple: resolved

executor/distsql.go
L-maple: this logic is removed now, as I push down the filter partitionIds.

L-maple:
@mjonss @Defined2014 This PR can be reviewed. This PR depends on tipb PR: https://github.com/pingcap/tipb/pull/297?w=1.

L-maple:
> Does the PR ready to be reviewed? Maybe you could remove `[WIP]` in title.

Done

executor/distsql.go
Defined2014: Why not filter out unnecessary pids when creating kvrange? Then we don't need change TiKV and TiPB any more.

executor/distsql.go
```diff
@@ -302,6 +303,24 @@ func (e *IndexReaderExecutor) open(ctx context.Context, kvRanges []kv.KeyRange)
...
+		}
+
+		idxScanExec := e.dagPB.Executors[0]
+		idxScanExec.IdxScan.IsGlobal = &e.index.Global
+		idxScanExec.IdxScan.PartitionIds = make([]int64, 0, len(e.partitionIDMap))
+		for pid := range e.partitionIDMap {
+			if _, exist := visiblePartitionIDMap[pid]; exist {
+				idxScanExec.IdxScan.PartitionIds = append(idxScanExec.IdxScan.PartitionIds, pid)
+			}
+		}
```
mjonss, line 322: Could this be a helper function and used both here and in `(e *IndexLookUpExecutor) open()` to avoid code duplication?

go.mod
mjonss: I think we split the optimization of pushing down the IndexFilter / Selection into a future PR.

L-maple: @mjonss For global index, pushing down `partitionIds` is not just an optimization. If `partitionIds` is not pushed down, there are wrong query cases.  I mention my questions(or say, awkward situation) in tipb PR.

executor/distsql.go
L-maple: resolved

go.mod
```diff
@@ -277,5 +277,6 @@ replace (
 	// fix potential security issue(CVE-2020-26160) introduced by indirect dependency.
```
hawkingrei, line 277: remember to run ```make bazel_prepare```.

store/mockstore/unistore/cophandler/mpp.go
```diff
@@ -168,6 +168,14 @@ func (b *mppExecBuilder) buildIdxScan(pb *tipb.IndexScan) (*indexScanExec, error
 	if b.dagReq.GetCollectRangeCounts() {
 		prevVals = make([][]byte, numIdxCols)
 	}
+	var partitionIDMap map[int64]struct{}
+	isGlobal := pb.IsGlobal != nil && *pb.IsGlobal
```
mjonss, line 172: If we go with extending the tipb protocol, is `isGlobal` really needed? is not `PartitionIds` enough? I.e. if `PartitionIds` is nil or of zero length, it already has the same practical meaning as of `isGlobal`?

L-maple: I understand `isGlobal` cannot be removed. 
w.g., If a table has only a partition, and it's in reorg. The query goes global index but the PartitionIds is empty now.

ddl/db_partition_test.go
```diff
@@ -1952,6 +1953,88 @@ func TestDropPartitionWithGlobalIndex(t *testing.T) {
 	require.Equal(t, 2, cnt)
 }
 
+func TestGlobalIndexReaderInDropPartition(t *testing.T) {
+	t.Skip("not support yet")
```
L-maple, line 1957: `partitionIds` for global index will be pushed down in a graceful manner in future.

L-maple:
@mjonss I implement the Selection push down for global index `partitionIds`. You can review the code again, and confirm whether the modification is right.

L-maple:
> Details

@CbcWestwolf  Hi, I find the ut runs timeout. What should I do?
image

mjonss (approved):
LGTM with some questions/suggestions.

mjonss:
@L-maple please update/merge with upstream (pingcap/master) and resolve the `ddl/db_partition_test.go` conflict.

executor/builder.go
```diff
@@ -4520,6 +4559,32 @@ func (builder *dataReaderBuilder) buildIndexLookUpReaderForIndexJoin(ctx context
...
+
+	is := v.IndexPlans[0].(*plannercore.PhysicalIndexScan)
+	ts := v.TablePlans[0].(*plannercore.PhysicalTableScan)
+	if is.Index.Global {
+		tmp, ok := builder.is.TableByID(ts.Table.ID)
+		if !ok {
+			return nil, err
+		}
+		tbl, ok1 := tmp.(table.PartitionedTable)
+		if !ok1 {
```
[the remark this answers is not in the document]
L-maple: resolved

executor/distsql.go
```diff
@@ -325,6 +327,41 @@ func (e *IndexReaderExecutor) open(ctx context.Context, kvRanges []kv.KeyRange)
...
+		idxScanExec := e.dagPB.Executors[0]
+		conditions := make([]expression.Expression, 0, len(e.partitionIDMap))
+		for pid := range e.partitionIDMap {
+			column := &expression.Column{
+				UniqueID: model.ExtraPidColID,
+				RetType:  types.NewFieldType(mysql.TypeLonglong),
+				Index:    len(idxScanExec.IdxScan.Columns) - 1,
+				OrigName: model.ExtraPartitionIdName.L,
+			}
+			fc, err := expression.NewFunction(e.ctx, ast.EQ, types.NewFieldType(mysql.TypeLonglong), column, expression.NewInt64Const(pid))
```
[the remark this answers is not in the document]
L-maple: Yes. I tried and it works. `ast.In` is actually more simple.

L-maple:
> @L-maple please update/merge with upstream (pingcap/master) and resolve the `ddl/db_partition_test.go` conflict.

@mjonss Hi, I resolved the conflict. But later I find building failed, and I cannot understand what it means and what I should do.
image

executor/distsql.go
```diff
@@ -325,6 +327,39 @@ func (e *IndexReaderExecutor) open(ctx context.Context, kvRanges []kv.KeyRange)
...
+			RetType:  types.NewFieldType(mysql.TypeLonglong),
+			Index:    len(idxScanExec.IdxScan.Columns) - 1,
+			OrigName: model.ExtraPartitionIdName.L,
+		}
+		args = append(args, column)
+		for pid := range e.partitionIDMap {
+			args = append(args, expression.NewInt64Const(pid))
+		}
+
+		inCondition, err := expression.NewFunction(e.ctx, ast.In, types.NewFieldType(mysql.TypeLonglong), args...)
```
mjonss, line 344:
```suggestion
		inCondition, err := expression.NewFunction(e.ctx, ast.In, types.NewFieldType(mysql.TypeLonglong), args...)
		if err != nil {
			return err
		}
```

L-maple: resolved, thanks.

L-maple:
@Defined2014  please help me review this PR, thanks~

executor/builder.go
```diff
@@ -3765,6 +3765,21 @@ func (b *executorBuilder) buildIndexReader(v *plannercore.PhysicalIndexReader) E
...
+		tmp, ok := b.is.TableByID(ret.table.Meta().ID)
+		if !ok {
+			b.err = err
+			return nil
+		}
+		tbl, ok := tmp.(table.PartitionedTable)
+		if !ok {
+			b.err = exeerrors.ErrBuildExecutor
+			return nil
+		}
```
Defined2014, line 3777: maybe could remove those codes and take L3786 - L3787 to L3766

executor/builder.go
```diff
@@ -4544,6 +4583,32 @@ func (builder *dataReaderBuilder) buildIndexLookUpReaderForIndexJoin(ctx context
...
 		return e, err
 	}
+
+	is := v.IndexPlans[0].(*plannercore.PhysicalIndexScan)
+	ts := v.TablePlans[0].(*plannercore.PhysicalTableScan)
+	if is.Index.Global {
+		tmp, ok := builder.is.TableByID(ts.Table.ID)
+		if !ok {
+			return nil, err
+		}
```
Defined2014, line 4593: Forgot set err here. By the way, I think no need check ok here, just like L4612.

Also please fix other similar codes.

planner/core/find_best_task.go
```diff
@@ -1591,9 +1591,10 @@ func (is *PhysicalIndexScan) initSchema(idxExprCols []*expression.Column, isDoub
...
+	if isDoubleRead || is.Index.Global {
```
Defined2014, line 1594: Does it mean if not an isDoubleRead plan with global index also need extraHandleID here? Could you breif explain it?

executor/builder.go
L-maple: @mjonss told me I should check explicitly before. You can talk about it and I will follow your common advice.

executor/builder.go
L-maple: Yes👍🏻，You are right. But I find other places in this method also not set err. When the builder.err should be set?

planner/core/find_best_task.go
L-maple: If IndexScan is global, it should also decode `ExtraPidColID`  from values in TiKV. So that we can filter the rows from needed partitions.
e.g., select id from t partition(p0) use index(global_idx);
In this example, We should tell TiKV to decode `ExtraPidColID` in index value, and check whether we need the row.
If the IndexScan is single read, the `ExtraPidColID` would not return to SQL. Because we don't add its offset in DAG.

L-maple: You are right. For a singleRead, ExtraHandleID is actually useless. 
But unistore/TiKV's decode method has limit for it. I notice if handle is not needed, `decodeIndexKvGeneral` in unisore will return directly and `ExtraPidColID` is not decoded.
image

executor/builder.go
Defined2014: I think you could change it to `return nil, infoschema.ErrTableNotExists` here.

Defined2014: The err without set maybe always equal to `nil`.

L-maple: resolved

Defined2014 (approved):

executor/builder.go
L-maple: I handle the err (ErrTableNotExists) and still check the type transform explicitly here. @Defined2014

Issue #31316, opened 2022-01-05
Improve row count estimation when scaling estimated row count


time-and-fate:
## Enhancement
1. When there is `LIMIT` in the SQL, usually we will scale the estimated table/index scan row count according to the limit count.
1.  For the inner side of the index join, we will also estimate the table/index scan row count by scaling the inner side eventual row count according to the filter selectivity.

But it turns out that sometimes we have large estimation errors in these cases, so we may need to experiment and adjust our estimation strategy.

---

### UPDATE
This issue has been too broad for a specific problem. The description above actually covers a series of related estimation issues.
For the overestimation problem for the inner side of the index join, please refer to https://github.com/pingcap/tidb/issues/44855

time-and-fate:
Close this issue now. For the specific problem, please refer to https://github.com/pingcap/tidb/issues/42060 and https://github.com/pingcap/tidb/issues/44855.
We'll keep improving related estimations in the future.

PR #41996, opened 2023-03-07
planner: add an upper bound for estimated row count of inner side of index join

time-and-fate:

Issue Number: ref #31316

Problem Summary:

Because of our current implementation of index join, the row count for its inner side might be severely overestimated.

Add a reasonable upper bound for the inner children of index join to prevent very severe estimation errors.

Specifically, the average row count for the inner side `IndexScan` that corresponds to each row from the outer side should be no larger than (total row count / NDV of join key columns).

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


statistics/integration_test.go
```diff
@@ -693,3 +695,49 @@ func TestSingleColumnIndexNDV(t *testing.T) {
...
+			"IndexJoin 1000000.00 root  inner join, inner:IndexLookUp, outer key:test.t.a, inner key:test.t.b, equal cond:eq(test.t.a, test.t.b)",
+			"├─TableReader(Build) 1000.00 root  data:Selection",
+			"│ └─Selection 1000.00 cop[tikv]  lt(test.t.a, 1), not(isnull(test.t.a))",
+			"│   └─TableFullScan 500000.00 cop[tikv] table:t keep order:false, stats:pseudo",
+			"└─IndexLookUp(Probe) 1000000.00 root  ",
+			"  ├─Selection(Build) 1000000.00 cop[tikv]  not(isnull(test.t.b))",
+			"  │ └─IndexRangeScan 1000000.00 cop[tikv] table:t2, index:idx(b) range: decided by [eq(test.t.b, test.t.a)], keep order:false, stats:pseudo",
+			"  └─Selection(Probe) 1000000.00 cop[tikv]  eq(test.t.a, 0)",
+			"    └─TableRowIDScan 1000000.00 cop[tikv] table:t2 keep order:false, stats:pseudo",
+		))
```
time-and-fate, line 742: As a comparison, on current master branch, the execution plan is:
```
IndexJoin 1000000.00 root  inner join, inner:IndexLookUp, outer key:test.t.a, inner key:test.t.b, equal cond:eq(test.t.a, test.t.b)
├─TableReader(Build) 1000.00 root  data:Selection
│ └─Selection 1000.00 cop[tikv]  lt(test.t.a, 1), not(isnull(test.t.a))
│   └─TableFullScan 500000.00 cop[tikv] table:t keep order:false, stats:pseudo
└─IndexLookUp(Probe) 1000000.00 root  
  ├─Selection(Build) 500000000.00 cop[tikv]  not(isnull(test.t.b))
  │ └─IndexRangeScan 500000000.00 cop[tikv] table:t2, index:idx(b) range: decided by [eq(test.t.b, test.t.a)], keep order:false, stats:pseudo
  └─Selection(Probe) 1000000.00 cop[tikv]  eq(test.t.a, 0)
    └─TableRowIDScan 500000000.00 cop[tikv] table:t2 keep order:false, stats:pseudo
```

planner/core/exhaust_physical_plans.go
```diff
@@ -1146,13 +1146,53 @@ func (p *LogicalJoin) constructInnerUnionScan(us *LogicalUnionScan, reader Physi
...
+		copy(orderedIdxCols, idxCols)
+		slices.Sort(orderedIdxCols)
+		if !slices.Equal(idxCols, colUIDs) {
+			continue
+		}
+		if idxStats, ok := histColl.Indices[idxID]; ok && idxStats != nil {
+			return idxStats.NDV
+		}
+	}
+	return -1
```
winoros, line 1186: return max{single col ndv} instead?

time-and-fate: I think it should be min.

time-and-fate: Updated.

planner/core/exhaust_physical_plans.go
```diff
@@ -1146,13 +1146,70 @@ func (p *LogicalJoin) constructInnerUnionScan(us *LogicalUnionScan, reader Physi
...
+		slices.Sort(orderedIdxCols)
+		if !slices.Equal(idxCols, colUIDs) {
+			continue
+		}
+		if idxStats, ok := histColl.Indices[idxID]; ok && idxStats != nil {
+			return idxStats.NDV
+		}
+	}
+
+	// 3. If we still haven't got an NDV, we use the minimal NDV in the column stats as a lower bound.
```
AilinKid, line 1187: prefer a more case 3.5， once the index prefix columns satisfied the keys，we can use the index‘s DNV / the accent column’s DNV （thought they are completely independent distributions， and result after division is quite small than the real value， leading a higher （not that highest from direct dividing lowest col-ndv） estimations row count is which we wanna get from here

AilinKid (commented):
Rest LGTM

planner/core/exhaust_physical_plans.go
time-and-fate: That's a clever one!
It's not that obvious but indeed a correct lower bound for the NDV. And I think it doesn't need to be prefix columns, it's enough if the index contains the columns we want.

time-and-fate: But if we add this, there needs to be a strategy for choosing from multiple possible indexes, and we need to make sure the result is stable, we also need to care about some edge cases like prefix index and generated columns, and we also need to construct corresponding test cases, that's too much for this sprint.
Besides, this PR would be cherry-picked for a customer, safety is more important, and less change is preferred.
So I tend to add a TODO comment here and leave it to the future.

AilinKid (approved):

planner/core/exhaust_physical_plans.go
AilinKid: > That's a clever one! It's not that obvious but indeed a correct lower bound for the NDV. And I think it doesn't need to be prefix columns, it's enough if the index contains the columns we want.

make sense，subset of index column is enough

winoros (approved):

PR #41606, opened 2023-02-20
*: visible default resource group

BornChanger:
### What problem does this PR solve?

Issue Number: close #40875

Problem Summary:

### What is changed and how it works?

Initialize the default resource group to resource manager, and let it readable and changeable  but not droppable. 

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)
Simulate the upgrade from 6.6 to nightly to verify default resource group will be created.
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #42080, opened 2023-03-09
builder: support EXPLAIN ANALYZE to display RU statistical information

### LINKED ISSUE #41972: Make EXPLAIN ANALYZE show RU statistics info
Part of #38825.

To help users better know the actual RU cost for a particular SQL, it's very reasonable to introduce the RU statistics info by executing `EXPLAIN ANALYZE`. This is the tracking issue for this feature development.

JmPotato:

Issue Number: close #41972, #42087.

Problem Summary:

- Support `EXPLAIN ANALYZE` to display RU statistical information.
- Update the PD and client-go versions.

See https://github.com/tikv/client-go/pull/732 for more details. 

### Check List

Tests 

- [x] Unit test
- [x] Integration test

图片

图片

### Release note

```release-note
None
```


JmPotato:
/cc @glorv @tiancaiamao @nolouch 

PTAL, thx.

JmPotato:
> maybe we no need to make it visible by default. How about control it by a session variable? such as mysql has a variable like :

Upon examining the `EXPLAIN` [synopsis](https://docs.pingcap.com/tidb/dev/sql-statement-explain-analyze), it appears that `EXPLAIN ANALYZE` is unable to utilize `FORMAT = "xxx"` for specifying the output format. I am uncertain if implementing a new session variable to regulate this would be appropriate. Nevertheless, given that resource control is currently enabled by default, displaying RU information at all times seems favorable to me. Perhaps we could restrict showing the information only when the resource group is not empty or set as "default". What are your thoughts?

tiancaiamao (approved):
LGTM, please update go.mod when the dependency is ready.

JmPotato:
Wait for #42140 to be merged first.

executor/builder.go
```diff
@@ -1197,6 +1198,25 @@ func (b *executorBuilder) buildExplain(v *plannercore.Explain) Executor {
...
+		// If the resource group name is not empty or "default", we could collect and display the RU
+		// runtime stats for analyze executor.
+		resourceGroupName := b.ctx.GetSessionVars().ResourceGroupName
+		if len(resourceGroupName) > 0 && resourceGroupName != variable.DefaultResourceGroupName {
+			// Try to register the RU runtime stats for analyze executor.
+			store, ok := b.ctx.GetStore().(interface {
+				CreateRURuntimeStats(uint64) *clientutil.RURuntimeStats
+			})
+			if ok {
+				// StartTS will be used to identify this SQL, so that the runtime stats could
```
Connor1996, line 1210:
```suggestion
			if store, ok := b.ctx.GetStore().(interface {
				CreateRURuntimeStats(uint64) *clientutil.RURuntimeStats
			}); ok {
				// StartTS will be used to identify this SQL, so that the runtime stats could
```

executor/builder.go
```diff
@@ -1197,6 +1198,25 @@ func (b *executorBuilder) buildExplain(v *plannercore.Explain) Executor {
 		if b.ctx.GetSessionVars().StmtCtx.RuntimeStatsColl == nil {
 			b.ctx.GetSessionVars().StmtCtx.RuntimeStatsColl = execdetails.NewRuntimeStatsColl(nil)
 		}
+		// If the resource group name is not empty or "default", we could collect and display the RU
```
Connor1996, line 1201: why not display for default group>

JmPotato: Because it won't calculate the RU due to https://github.com/tikv/client-go/blob/9d950905d743258e72894c9fb09dd6e453e0b9af/internal/client/client_interceptor.go#L96.

Connor1996: Then why don't we calculate the RU for the default group?

JmPotato: It's an expected behavior decided before. According to the [documentation](https://docs.pingcap.com/tidb/dev/tidb-resource-control#how-to-use-resource-control):

> If a user is not bound to a resource group or is bound to a default resource group, the user's requests are not subject to the flow control restrictions of TiDB. The default resource group is currently not visible to the user and cannot be created or modified. You cannot view it with SHOW CREATE RESOURCE GROUP or SELECT * FROM information_schema.resource_groups. But you can view it through the mysql.user table.

nolouch: we will make default group visible，ref https://github.com/pingcap/tidb/pull/41606

nolouch (commented):
rest lgtm

executor/builder.go
```diff
@@ -1197,6 +1198,24 @@ func (b *executorBuilder) buildExplain(v *plannercore.Explain) Executor {
...
+		// runtime stats for analyze executor.
+		resourceGroupName := b.ctx.GetSessionVars().ResourceGroupName
+		if len(resourceGroupName) > 0 && resourceGroupName != ddl.DefaultResourceGroupName {
```
nolouch, line 1204: BTW, need check the variable`tidb_enable_resource_control`?

JmPotato: If the switch is not on, the user won't be able to bind the resource group. So I think checking the resource group name is enough.

nolouch (approved):
lgtm

glorv (approved):

PR #42083, opened 2023-03-09
schema cache: record the ddl timestamp and associate schema cache with the timestamp for stale read

### LINKED ISSUE #40740: avoid excessive PointGet request to a single tikv node for stale read when DDL happens
## Enhancement

currently, to avoid to get the schema version for a `snapshotTs`, the `snapshotTs` of the stale read needs to be greater than the update timestamp of the current schema version. https://github.com/pingcap/tidb/blob/master/infoschema/cache.go#L111

this behavior causes significant load issue on the single tikv node stores the schema version when DDL happens. usually the stale read will request the data from a few second up to minutes before current timestamp. this mean, when DDL happens with `maxUpdatedSnapshotTS  = t0`, all of the stale read requests happens from `t0` to `t0 + t_staleness` will query the data of `snapshotTs` before `t0`, thus requires to get the schema version at that `snapshotTs`.

one solution i could think of is to change the change the schema version request from `GetRequest` to `VerGetRequest` to obtain the corresponding timestamp for each schema version change. in this way, tidb could maintain a local map of the schema version and its commit timestamp, so that it could completely avoid the point get request for schema version on stale read code path.

hihihuhu:

Issue Number: close #40740

Problem Summary:

when handle ddl, record the start ts of the transaction as the timestamp of the schema change. then when populate the schema cache, associate each schema version with a timestamp, and search the schema by timestamp for stale read.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


hihihuhu:
@xhebox could you help to take a look when you have a chance? thanks!

ddl/ddl.go
```diff
@@ -394,6 +394,12 @@ func (sv *schemaVersionManager) setSchemaVersion(job *model.Job, store kv.Storag
 		var err error
 		m := meta.NewMeta(txn)
 		schemaVersion, err = m.GenSchemaVersion()
+		if err != nil {
+			return err
+		}
+
+		logutil.BgLogger().Info("DDL create schema with timestamp", zap.Int64("version", schemaVersion), zap.Uint64("start", txn.StartTS()))
+		err = m.SetTimestampForSchemaVersion(int64(txn.StartTS()), schemaVersion)
```
xhebox, line 402: I am afraid that you need `commitTS`.  Check `GetSchemaVersionWithNonEmptyDiff`. You need a timestamp where schemadiff is actually set.

This is a sub txn updating schema version only. That means the actual schemadiff and metadata are not writtent in this txn. I think the correct txn is at https://github.com/pingcap/tidb/blob/541ccd5013d2303a4c3e0a1793c6f16b73c25e30/ddl/ddl_worker.go#L813, where `w.sess.commit` is successfully invoked.

There is one RPC avaiable for that. Check https://github.com/pingcap/tidb/blob/d6530113280f5b019814642b7587689e90fd8bf5/docs/tidb_http_api.md, `mvcc/txn/{startTS}/{db}/{table}`.  Also this:

https://github.com/pingcap/tidb/blob/d6530113280f5b019814642b7587689e90fd8bf5/store/helper/helper.go#L127

domain/domain.go
```diff
@@ -199,6 +199,12 @@ func (do *Domain) loadInfoSchema(startTS uint64) (infoschema.InfoSchema, bool, i
 	if err != nil {
 		return nil, false, 0, nil, err
 	}
+	schemaTs, err := m.GetTimestampForSchemaVersion(neededSchemaVersion)
```
xhebox, line 202: I think another method is that. we can just store the needed info, `startTS` or `commitTS` or whatever, into schemadiff. Check `GetSchemaVersionWithNonEmptyDiff()`, you are able to retrive the correct schema diff.

infoschema/cache.go
```diff
@@ -98,48 +139,70 @@ func (h *InfoCache) GetBySnapshotTS(snapshotTS uint64) InfoSchema {
...
 	version := is.SchemaMetaVersion()
+
+	// assume this is the timestamp order as well
 	i := sort.Search(len(h.cache), func(i int) bool {
-		return h.cache[i].SchemaMetaVersion() <= version
+		return h.cache[i].infoschema.SchemaMetaVersion() <= version
 	})
 
+	// keep maxUpdatedSnapshotTS for backward compatibility
 	if h.maxUpdatedSnapshotTS < snapshotTS {
```
xhebox, line 165: We don't need this, if we really give accurate timestamp. Your solution will be an improvement over `maxUpdatedSnapshotTS`.

xhebox (commented):
Maybe we should rename `loadschema(startTS uint64)` to `loadschema(timestamp uint64)`. This function will accept a timestamp that **starts with the desired schema and schemadiff/metadata**. It does not mean that `startTS` is the timestamp set by `updateSchemaVersion`.

domain/domain.go
```diff
@@ -199,6 +200,20 @@ func (do *Domain) loadInfoSchema(startTS uint64) (infoschema.InfoSchema, bool, i
 	if err != nil {
 		return nil, false, 0, nil, err
 	}
+	// fetch the commit timestamp of the schema diff
+	schemaTs := int64(0)
+	tikvStore, ok := do.Store().(helper.Storage)
+	if ok {
+		helper := helper.NewHelper(tikvStore)
```
hihihuhu, line 207: @xhebox thanks for the suggestion! i feel like it is perhaps better to use the mvcc commit ts of the schema diff as the timestamp, because add a new field in schema diff will need to deal with backward compatible that some schema may not have the timestamp while some other does. however, in order to fetch the mvcc, i am constructing a new `helper` like this, is this ok or it could / should be done better? thanks!

meta/meta.go
```diff
@@ -427,6 +430,19 @@ func (m *Meta) GetSchemaVersionWithNonEmptyDiff() (int64, error) {
 	return v, err
 }
 
+// Returns the timestamp of a schema version, which is the commit timestamp of the schema diff
+func (m *Meta) GetTimestampForSchemaVersion(helper *helper.Helper, version int64) (int64, error) {
+	diffKey := m.schemaDiffKey(version)
+	data, err := helper.GetMvccByEncodedKey(m.txn.EncodeStringDataKey(diffKey))
```
xhebox, line 436: I don't recommend to insert the logic here, maybe in `domain.go`

domain/domain.go
xhebox: > because add a new field in schema diff will need to deal with backward compatible that some schema may not have the timestamp while some other does. 

It is perfectly fine. We use `json`, that means additional fields won't break old clusters.

**And honestly. your current solution will also fail on old clusters, since old cluster have no corresponding meta info.**

You need to solve the problem that old clusters does not have a infoschema. It is only safe to execute the code after all tidb fetched the latest infoschema.

That is there should be no DDL while upgrading, which is not required by the current tidb. You either need a global system variable, or just do not cache infoschema with invalid TS, except the `maxSnapshotTS` one.

> however, in order to fetch the mvcc, i am constructing a new helper like this, is this ok or it could / should be done better

LGTM, this looks like the correct solution however it is backward compatibility breaker now.

hihihuhu: thanks for the review! i am addressing other comments, but have some confusion on the compatibility issue you mentioned? the timestamp used here is the mvcc commit timestamp of schema diff, so your concern is maybe for some historical version, the commit timestamp could not be available?

xhebox: > so your concern is maybe for some historical version, the commit timestamp could not be available?

Exactly, it is fine on TiDB cloud cluster. Our manager can handle that. But for self-managed clusters. it will be an unexpected compatibility breaker.

It is better to have a system variable to control, I think.

xhebox: > the timestamp used here is the mvcc commit timestamp of schema diff,

I am saying you can store `startTS` into schema diff, and get `commit TS` in `domain.go`, eliminating the need of changing `meta info`. `Meta info` will not be cleared, it will accumulated.

hihihuhu:
thanks! i updated to store startTs in schemadiff and use it as the timestamp of a schema version, i haven't tested it yet but want to get your opinion first to see if this is what you expect?

parser/model/ddl.go
```diff
@@ -902,6 +902,9 @@ type SchemaDiff struct {
 	RegenerateSchemaMap bool `json:"regenerate_schema_map"`
 
 	AffectedOpts []*AffectedOption `json:"affected_options"`
+
+	// The start ts of the transaction that writes the SchemaDiff, used as the timestamp that the schema starts to take effect
```
xhebox, line 906: Not exactly. The timestamp that actually "starts to take effect", is the commitTS corresponding to this `startTS`. Try to query `get_mvcc_by_ts/{starTS}/{m.schemaDiffKey(schemaVersion)}`, see if you can get the desired information.

xhebox:
> thanks! i updated to store startTs in schemadiff and use it as the timestamp of a schema version, i haven't tested it yet but want to get your opinion first to see if this is what you expect?

You still need to use `commitTS`. Otherwise LGTM.

parser/model/ddl.go
hihihuhu: > > thanks! i updated to store startTs in schemadiff and use it as the timestamp of a schema version, i haven't tested it yet but want to get your opinion first to see if this is what you expect?
> 
> You still need to use `commitTS`. Otherwise LGTM.

try to get some clarity, do you suggest to use https://github.com/pingcap/tidb/pull/42083/commits/c169d462e80c104b7965bcb8c30e60beeb82234f ? that version uses the commit ts of the schema diff, and don't need to handle backward compatibility because the commit ts is already available

xhebox: OK, OK.

I think I overlooked the commit message. I thought you are still using startTS to retrive timestamp.

Yes, that commit looks fine. I'll take a deeper look tomorrow.

xhebox: @hihihuhu OK, sorry. :(
I did overlooked the commit. That commit is fine. Please change to that commit.

structure/string.go
```diff
@@ -79,3 +79,8 @@ func (t *TxStructure) Clear(key []byte) error {
 	}
 	return errors.Trace(err)
 }
+
+// Expose how to encode a key to fetch the MVCC info
+func (t *TxStructure) EncodeStringDataKey(key []byte) kv.Key {
+	return t.encodeStringDataKey(key)
```
xhebox, line 85: Maybe we can just make  `t.encodeStringDataKey` public.

meta/meta.go
xhebox: Move this function to `domain.go` and avoid import `github.com/pingcap/tidb/store/helper` in this library.

domain/domain.go
```diff
@@ -199,6 +200,20 @@ func (do *Domain) loadInfoSchema(startTS uint64) (infoschema.InfoSchema, bool, i
...
+		schemaTs, err = m.GetTimestampForSchemaVersion(helper, neededSchemaVersion)
+		if err != nil {
+			logutil.BgLogger().Warn("failed to get schema version", zap.Error(err), zap.Int64("version", neededSchemaVersion))
+			return nil, false, 0, nil, err
+		}
+	} else {
+		logutil.BgLogger().Warn("cannot get store from domain")
+		return nil, false, 0, nil, errors.Errorf("cannot get store from domain")
```
xhebox, line 215: In case that they are not `helper.Storage`, we should not fail, just warn.

domain/domain.go
```diff
@@ -199,6 +200,20 @@ func (do *Domain) loadInfoSchema(startTS uint64) (infoschema.InfoSchema, bool, i
 	if err != nil {
 		return nil, false, 0, nil, err
 	}
+	// fetch the commit timestamp of the schema diff
+	schemaTs := int64(0)
+	tikvStore, ok := do.Store().(helper.Storage)
+	if ok {
+		helper := helper.NewHelper(tikvStore)
+		schemaTs, err = m.GetTimestampForSchemaVersion(helper, neededSchemaVersion)
```
xhebox, line 208: Rename `GetTimestampForSchemaVersion`  to `GetTimestampForSchemaVersionWithNonEmptyDiff` and coment that `GetTimestampForSchemaVersionWithNonEmptyDiff` should not fail if `GetSchemaVersionWithNonEmptyDiff` did not fail.

infoschema/cache.go
```diff
@@ -98,48 +124,53 @@ func (h *InfoCache) GetBySnapshotTS(snapshotTS uint64) InfoSchema {
...
 	}
 	return nil
 }
 
 // Insert will **TRY** to insert the infoschema into the cache.
 // It only promised to cache the newest infoschema.
 // It returns 'true' if it is cached, 'false' otherwise.
-func (h *InfoCache) Insert(is InfoSchema, snapshotTS uint64) bool {
+// snapshotTS is the timestap of the schema update, schemaTS is the timestamp of the schema taking effect
+func (h *InfoCache) Insert(is InfoSchema, snapshotTS, schemaTS uint64) bool {
```
xhebox, line 138: We can remove `snapshotTS` now. It is not used anywhere.

Change comments of `GetBySnapshotTS`, too: `snapshotTS` was `startTS`. But in `GetBySnapshotTS`, we mean that, give any random TS, we should return the correct schema or not.

meta/meta.go
```diff
@@ -43,6 +44,7 @@ var (
 // Meta structure:
 //	NextGlobalID -> int64
 //	SchemaVersion -> int64
+//  SchemaTimestampKey -> int64
```
xhebox, line 47: Useless comment now.

infoschema/cache_test.go
```diff
@@ -116,16 +116,46 @@ func TestGetLatest(t *testing.T) {
...
 	require.Equal(t, is2, ic.GetLatest())
 }
+
+func TestGetByTimestamp(t *testing.T) {
+	ic := infoschema.NewCache(16)
+	require.NotNil(t, ic)
+	require.Nil(t, ic.GetLatest())
+
+	is1 := infoschema.MockInfoSchemaWithSchemaVer(nil, 1)
+	ic.Insert(is1, 2, 1)
```
xhebox, line 139: It is impossible that a schema is updated at `t2`, but commited diff at `t1`. It suggest remove `snapshotTS` completely as suggested.

xhebox (commented):
Generally LGTM now,  please address the following comments. And I will find another reviewer for you. @hihihuhu

structure/string.go
hihihuhu: updated

infoschema/cache.go
```diff
@@ -134,9 +142,9 @@ func (h *InfoCache) GetBySnapshotTS(snapshotTS uint64) InfoSchema {
 // Insert will **TRY** to insert the infoschema into the cache.
 // It only promised to cache the newest infoschema.
 // It returns 'true' if it is cached, 'false' otherwise.
-// snapshotTS is the timestap of the schema update, schemaTS is the timestamp of the schema taking effect
-func (h *InfoCache) Insert(is InfoSchema, snapshotTS, schemaTS uint64) bool {
-	logutil.BgLogger().Debug("INSERT SCHEMA", zap.Uint64("snapshot ts", snapshotTS), zap.Uint64("schema ts", schemaTS))
+// schemaTs is the startTs of the txn creates the schema diff, which indicates since when the schema version is taking effect
```
xhebox, line 145: It is `commitTS`.

infoschema/cache.go
```diff
@@ -117,8 +125,8 @@ func (h *InfoCache) getByVersionNoLock(version int64) InfoSchema {
 }
 
 // GetBySnapshotTS gets the information schema based on snapshotTS.
-// If the snapshotTS is new than maxUpdatedSnapshotTS, that's mean it can directly use
-// the latest infoschema. otherwise, will return nil.
+// It searches the schema cache and find the schema with max schema ts that equals or smaller than given snapshot ts
+// Where the schema ts is the start ts of the txn creates the schema diff
```
xhebox, line 129: Same, it is `commit ts` of the txn.

infoschema/cache.go
```diff
@@ -51,18 +59,44 @@ func (h *InfoCache) GetLatest() InfoSchema {
...
+	// this is ok because the size of h.cache is small (currently set to 16)
+	// moreover, the most likely hit element in the array is the first one in steady mode
+	// thus it may have better performance than binary search
+	for _, is := range h.cache {
+		if is.timestamp == 0 {
+			// the schema version doesn't have a timestamp
+			// ignore all the schema cache equals or less than this version in search by timestamp
+			break
+		}
+		if ts >= uint64(is.timestamp) {
```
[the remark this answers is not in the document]
hihihuhu: good catch! updated to ignore the schema cache if there is a gap in the schema version between the searched schema and the immediate one after it in the cache.

infoschema/cache.go
```diff
@@ -51,18 +59,48 @@ func (h *InfoCache) GetLatest() InfoSchema {
 	infoschema_metrics.GetLatestCounter.Inc()
 	if len(h.cache) > 0 {
 		infoschema_metrics.HitLatestCounter.Inc()
-		return h.cache[0]
+		return h.cache[0].infoschema
 	}
 	return nil
 }
 
+func (h *InfoCache) getSchemaByTimestampNoLock(ts uint64) (InfoSchema, error) {
```
lcwangchao, line 67: Maybe change the return values' type to `(InfoSchema, bool)` is better. `errors.New` will generate a stack which is not so efficient and in semantic return a bool value is the same design with `map` if a item not exist

lcwangchao (commented):
Rest LGTM

infoschema/cache.go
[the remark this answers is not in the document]
xhebox: The current check only works if `cache` hits. What about moving `L80-L83` to `L78`, making it a general check like `is.timestamp == 0`?

That is, we trust if and only if `is.timestamp != 0` and no gap.

tiancaiamao (approved):

meta/meta.go
```diff
@@ -441,6 +441,12 @@ func (m *Meta) GetSchemaVersionWithNonEmptyDiff() (int64, error) {
 	return v, err
 }
 
+// Returns the raw kv key for a schema diff
```
CbcWestwolf, line 444:
```suggestion
// EncodeSchemaDiffKey returns the raw kv key for a schema diff
```

domain/domain.go
```diff
@@ -259,10 +266,28 @@ func (do *Domain) loadInfoSchema(startTS uint64) (infoschema.InfoSchema, bool, i
...
+		if err != nil {
+			return 0, err
+		}
+		if len(data.Info.Writes) == 0 {
+			return 0, errors.Errorf("There is no Write MVCC info for the schema version")
+		}
+		return int64(data.Info.Writes[0].CommitTs), nil
+	} else {
+		return 0, errors.Errorf("cannot get store from domain")
+	}
```
CbcWestwolf, line 288:
```suggestion
	}
	return 0, errors.Errorf("cannot get store from domain")
```

hihihuhu:
@xhebox consider this is a fix of performance issue, is it possible to merge this into release-6.5? thanks!

PR #42089, opened 2023-03-10
Zimuxia/dist dispatcher x

zimulala:

Issue Number: close https://github.com/pingcap/tidb/issues/41748

Problem Summary:
Implement the simplest logical flow(run through demo normally) for the dispatcher

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


disttask/framework/dispatcher/BUILD.bazel
```diff
@@ -0,0 +1,39 @@
...
+        "//domain/infosync",
+        "//util",
+        "//util/logutil",
+        "@com_github_pingcap_errors//:errors",
+        "@org_uber_go_zap//:zap",
+    ],
+)
+
+go_test(
+    name = "dispatcher_test",
```
hawkingrei, line 23: Please add 

```
    timeout = "short",
    flaky = True,
```

disttask/framework/proto/task.go
```diff
@@ -78,9 +79,11 @@ type Subtask struct {
 // Each subtask is divided into multiple minimal tasks by scheduler.
 type MinimalTask interface{}
 
-// TaskTypeExample is TaskType of Example.
 const (
+	// TaskTypeExample is TaskType of Example.
 	TaskTypeExample = "Example"
+	// TaskTypeCreateIndex is TaskType of "add index" operation.
```
GMHDBJD, line 85: Remove it or add the Type2Int/Int2Type function

disttask/framework/dispatcher/register.go
```diff
@@ -0,0 +1,39 @@
...
+package dispatcher
+
+import (
+	"github.com/pingcap/tidb/disttask/framework/proto"
+)
+
+// TaskFlowHandle is used to control the process operations for each global task.
+type TaskFlowHandle interface {
+	ProcessNormalFlow(d Dispatch, gTask *proto.Task, fromPending bool) (finished bool, subtasks []*proto.Subtask, err error)
+	ProcessErrFlow(d Dispatch, gTask *proto.Task, receive string) (meta []byte, err error)
```
GMHDBJD, line 24: prefer pass the TaskMeta to user instead of proto.Task

disttask/framework/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,410 @@
...
+	"github.com/pingcap/tidb/disttask/framework/proto"
+	"github.com/pingcap/tidb/disttask/framework/storage"
+	"github.com/pingcap/tidb/domain/infosync"
+	tidbutil "github.com/pingcap/tidb/util"
+	"github.com/pingcap/tidb/util/logutil"
+	"go.uber.org/zap"
+)
+
+const (
+	DefaultConcurrency        = 8
```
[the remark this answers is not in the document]
zimulala: This is used to control the parallelism of the global task.

disttask/framework/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,410 @@
...
+					break
+				}
+				// There are currently no global tasks to work on.
+				if gTask == nil {
+					break
+				}
+				if d.isRunningGlobalTask(gTask) {
+					break
+				}
+				if gTask.State == proto.TaskStateReverting {
```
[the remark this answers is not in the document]
zimulala: If the dispatcher down, it might get the task 'TaskStateReverting'

disttask/framework/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,410 @@
...
+				gTask, err := d.gTaskMgr.GetNewTask()
+				if err != nil {
+					logutil.BgLogger().Warn("get new task failed", zap.Error(err))
+					break
+				}
+				// There are currently no global tasks to work on.
+				if gTask == nil {
+					break
+				}
+				if d.isRunningGlobalTask(gTask) {
```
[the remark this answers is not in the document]
zimulala: I don't understand. It's all running after calling `processNormalFlow`, so I put it in `setRunningGlobalTasks`, and it is running.

disttask/framework/dispatcher/register.go
```diff
@@ -0,0 +1,39 @@
...
+type TaskFlowHandle interface {
+	ProcessNormalFlow(d Dispatch, gTask *proto.Task, fromPending bool) (finished bool, metas [][]byte, err error)
+	ProcessErrFlow(d Dispatch, gTask *proto.Task, receive string) (meta []byte, err error)
+}
+
+var taskDispatcherHandleMap = make(map[string]TaskFlowHandle)
+
+// RegisterTaskFlowHandle is used to register the global task handle.
+func RegisterTaskFlowHandle(taskType string, dispatcherHandle TaskFlowHandle) {
+	taskDispatcherHandleMap[taskType] = dispatcherHandle
```
lcwangchao, line 31: It's better to use `sync.Mutex` to make sure this operation is thread safe.

BTW, cloud we avoid using static variables which can make some troubles in testing. I think it's better to introduce a `Node` object to manage all these registers. 

```
type Node interface {
    RegisterTaskFlowHandle()
    RegisterExecutor()
    Start()
}
```

disttask/framework/dispatcher/dispatcher.go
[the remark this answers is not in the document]
zimulala: Added comments.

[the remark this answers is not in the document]
wjhuang2016: Please use a better name.

disttask/framework/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,412 @@
...
+	// TODO: UpdateTask and addSubtasks in a txn.
+	// Write the global task meta into the storage.
+	err = d.gTaskMgr.UpdateTask(gTask)
+	if err != nil {
+		logutil.BgLogger().Warn("update global task failed", zap.Error(err))
+		return err
+	}
+
+	if finished {
+		return nil
```
lcwangchao, line 342: When the task state is `Pending` and the user's `ProcessNormalFlow` returns `finished=true`, it seems the task status is `Running` without submitting any tasks ?

disttask/framework/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,412 @@
...
+	}
+
+	// Special handling for the new tasks.
+	// About TaskStateSucceed gTask, we consider keeping records first.
+	if gTask.State == proto.TaskStatePending {
+		// TODO: Consider using TS.
+		gTask.StartTime = time.Now().UTC()
+		gTask.State = proto.TaskStateRunning
+	} else {
+		gTask.StateUpdateTime = time.Now().UTC()
```
lcwangchao, line 328: When a task has serval steps, the `StateUpdateTime` also updates while step change, but the `State` does not change (Running => Running), it is a little strange

disttask/framework/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,412 @@
...
+					cnt++
+					continue
+				}
+
+				err = d.processNormalFlow(gTask, true)
+				if err != nil {
+					d.delRunningGlobalTasks(gTask.ID)
+				}
+				d.setRunningGlobalTasks(gTask)
+				cnt++
```
lcwangchao, line 155: Why `delRunningGlobalTasks` and then `setRunningGlobalTasks` if `err != nil` ?

disttask/framework/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,412 @@
...
+		select {
+		case <-d.ctx.Done():
+			logutil.BgLogger().Info("dispatch task loop exits", zap.Error(d.ctx.Err()))
+			return
+		case <-ticker.C:
+			cnt := len(d.getRunningGlobalTasks())
+
+			for cnt < DefaultConcurrency {
+				// TODO: Consider get all unfinished tasks.
+				gTask, err := d.gTaskMgr.GetNewTask()
```
lcwangchao, line 132: What if the dispatch crashes when a task's state changes to "running" ? `GetNewTask` only returns a task with state `Pending`

disttask/framework/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,412 @@
...
+	// Write the global task meta into the storage.
+	err = d.gTaskMgr.UpdateTask(gTask)
+	if err != nil {
+		logutil.BgLogger().Warn("update global task failed", zap.Error(err))
+		return err
+	}
+
+	// New rollback subtasks and write into the storage.
+	for _, id := range instanceIDs {
+		subtask := proto.NewSubtask(gTask.ID, gTask.Type, id, meta)
```
lcwangchao, line 286: Should we waiting a tasks from the previous step exit before submitting new tasks?

disttask/framework/dispatcher/dispatcher.go
zimulala: We only update the global task in line335.

disttask/framework/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,412 @@
...
+			}
+		}
+	}
+}
+
+func (d *dispatcher) processFlow(gTask *proto.Task, errStr string) {
+	var err error
+	if errStr != "" {
+		// Found an error when task is running.
+		logutil.BgLogger().Info("detection, handle an error", zap.Int64("taskID", gTask.ID))
```
lcwangchao, line 229: When `errStr != "" && gTask.State == proto.TaskStateReverting` , it seems the dispatcher will keep adding new revert tasks. Is it by design?

disttask/framework/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,412 @@
...
+
+// Start implements Dispatch.Start interface.
+func (d *dispatcher) Start() {
+	d.wg.Run(func() {
+		d.DispatchTaskLoop()
+	})
+	d.wg.Run(func() {
+		d.DetectTaskLoop()
+	})
+}
```
hawkingrei, line 109:
```suggestion
func (d *dispatcher) Start() {
	d.wg.Run(d.DispatchTaskLoop)
	d.wg.Run(d.DetectTaskLoop)
}
```

disttask/framework/dispatcher/dispatcher.go
zimulala: This item does not have an interface on the storage side at present, and will be changed to get unfinished tasks later

disttask/framework/dispatcher/register.go
```diff
@@ -0,0 +1,39 @@
...
+
+package dispatcher
+
+import (
+	"github.com/pingcap/tidb/disttask/framework/proto"
+)
+
+// TaskFlowHandle is used to control the process operations for each global task.
+type TaskFlowHandle interface {
+	ProcessNormalFlow(d Dispatch, gTask *proto.Task, fromPending bool) (finished bool, metas [][]byte, err error)
```
lcwangchao, line 23: It's better `metas [][]byte` => `[]*NewSubTaskRequest` . Introducing `NewSubTaskRequest` can wrap all information needed by creating a new sub task.

Another thing is could we remove `finished` return value and dispatcher can check whether the returned subtasks is empty. If it is empty, it can be dedicated the task will be finished.

disttask/framework/dispatcher/register.go
```diff
@@ -0,0 +1,39 @@
...
+	ProcessErrFlow(d Dispatch, gTask *proto.Task, receive string) (meta []byte, err error)
```
lcwangchao, line 24: ditto, I think returning a named struct `RollbackRequest` is better

disttask/framework/dispatcher/register.go
lcwangchao: Also, clould we remove `fromPending` from paramter? `gTask.State` contains this information

disttask/framework/dispatcher/dispatcher.go
zimulala: Yes.

disttask/framework/dispatcher/dispatcher.go
zimulala: The next step of step1-running should be step1-finish or reverting

disttask/framework/dispatcher/register.go
zimulala: I think the "Node object" suggestion is a good one

disttask/framework/dispatcher/register.go
zimulala: Don't do `metas [][]byte` => `[]*NewSubTaskRequest`, because `[]byte` is enough, and use `NewSubtask` directly.

disttask/framework/dispatcher/register.go
zimulala: Ditto.

disttask/framework/dispatcher/dispatcher.go
lcwangchao: If we do not waiting previous tasks stop, we will have two problems.

1. The user should consider that some subtasks from previous step are still running when rollbacking
2. If error happens when reverting, `processErrFlow` may be called frequently because we cannot seperate different error tasks from different steps in reverting state ?

zimulala: 1. schedule will handle it, and we can assume it hasn't received the cancellation notice yet
2. after the first revert, the gTask state is reverting, we won't get the error msg because `error happens` like the first revert.

wjhuang2016 (approved):

GMHDBJD (approved):
LGTM

PR #42099, opened 2023-03-10
extension: support audit log

### LINKED ISSUE #42188: Support audit log based on the extension framework

CbcWestwolf:

Issue Number: close #42188

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Support new audit log in tidb enterprise version
```


server/conn.go
```diff
@@ -2454,7 +2456,7 @@ func (cc *clientConn) handleResetConnection(ctx context.Context) error {
 		return errors.New("Could not reset connection")
 	}
 	if cc.dbname != "" { // Restore the current DB
-		err = cc.useDB(context.Background(), cc.dbname)
+		_, err = cc.useDB(context.Background(), cc.dbname)
```
lcwangchao, line 2459: Do we need to audit the use stmt here?

CbcWestwolf: Yes, I think we'd better audit the `use db` stmt.

lcwangchao: If so, I think it's better to move the `onStmtEnd` call into `useDB` ?

CbcWestwolf:
The timeout mechanism is also needed for writing audit log like https://github.com/pingcap/tidb/pull/42212

server/conn.go
CbcWestwolf: I misunderstood what you mean here. 
I think that we need to audit the query statement that users input, including `use db`. 
But the `use test` caused by non-query, like changing user, would not be audited, since users can get the visited db name by the `CURRENT_DB` field in the log entry.
So we only audit the `use db` [here](https://github.com/pingcap/tidb/pull/42099/files#diff-412615e1c1d4c6ae11a7885a11c879778bf401e0967c9c13c4b9dfd1f13ed29aR1321).

bb7133 (approved):
LGTM

.gitmodules
Mahmeoud19993, line 1: @lcwangchao

Mahmeoud19993 (commented):
@ti-chi-bot

PR #42116, opened 2023-03-10
planner, sessionctx: add variable to control index selection for "WHERE b ORDER BY a LIMIT n"

### LINKED ISSUE #42060: Add variables to control index selection for "WHERE b ORDER BY a LIMIT n" query
## The problem

Consider this case:
The table has indexes on column a and b respectively. And the query contains LIMIT n, WHERE condition on column b, and ORDER BY column a.

In this case, we have two indexes to choose between:
1. If we choose index on column b, we can make use of the WHERE condition and only need to read rowids that satisfy the filter from the index. Then, we use the rowids to read data from the table and do a TopN to get the final result.
2. If we choose index on column a, we can make use of its order. We just read rowids from index(a) from the beginning, and use them to read data from the table while keeping the order. Then, we use the WHERE condition to filter the data. Once we read n rows after the filter, it's finished.

The execution plan is as follows respectively:
```sql
> explain select * from t1 use index (ib) where b > 900000 order by a limit 1;
+----------------------------------+-----------+-----------+-----------------------+---------------------------------------+
| id                               | estRows   | task      | access object         | operator info                         |
+----------------------------------+-----------+-----------+-----------------------+---------------------------------------+
| TopN_9                           | 1.00      | root      |                       | test.t1.a, offset:0, count:1          |
| └─IndexLookUp_16                 | 1.00      | root      |                       |                                       |
|   ├─IndexRangeScan_13(Build)     | 100284.60 | cop[tikv] | table:t1, index:ib(b) | range:(900000,+inf], keep order:false |
|   └─TopN_15(Probe)               | 1.00      | cop[tikv] |                       | test.t1.a, offset:0, count:1          |
|     └─TableRowIDScan_14          | 100284.60 | cop[tikv] | table:t1              | keep order:false                      |
+----------------------------------+-----------+-----------+-----------------------+---------------------------------------+

> explain select * from t1 use index (ia) where b > 900000 order by a limit 1;
+-----------------------------------+---------+-----------+-----------------------+-----------------------+
| id                                | estRows | task      | access object         | operator info         |
+-----------------------------------+---------+-----------+-----------------------+-----------------------+
| Limit_12                          | 1.00    | root      |                       | offset:0, count:1     |
| └─Projection_22                   | 1.00    | root      |                       | test.t1.a, test.t1.b  |
|   └─IndexLookUp_21                | 1.00    | root      |                       |                       |
|     ├─IndexFullScan_18(Build)     | 9.97    | cop[tikv] | table:t1, index:ia(a) | keep order:true       |
|     └─Selection_20(Probe)         | 1.00    | cop[tikv] |                       | gt(test.t1.b, 900000) |
|       └─TableRowIDScan_19         | 9.97    | cop[tikv] | table:t1              | keep order:false      |
+-----------------------------------+---------+-----------+-----------------------+-----------------------+
```

The essential problem here is that: when we choose index(a), we usually can't accurately estimate how many rows we need to read on index(a) to get n rows satisfying the condition.

This makes tidb easily chooses the wrong index. Especially when the row count for index(a) is underestimated and tidb wrongly chooses index(a), this plan is usually disastrous.

Therefore we plan to add a system variable to add some control over the behavior.

## Behavior

Variable name: `tidb_opt_ordering_index_selectivity_threshold`

SESSION | GLOBAL level

Default value: 0, means the behavior is not changed.
If there exists one filter index (the index(b)) whose estimated selectivity is smaller than the variable, the optimizer won't consider the ordering index (the index(a)).

## Implementation

1. Add a new system variable as stated above.
2. In `(*DataSource).DeriveStats()`, after we generated the access paths and estimated the row count for them, calculate the minimal selectivity among the access paths.
3. In `(*DataSource).getOriginalPhysicalIndexScan()`, if we find the variable is larger than the minimal selectivity, and there is Selection on the IndexScan, we won't use the ExpectedCnt to adjust the estimated row count of the IndexScan.

Note that we don't directly drop the ordering index access path, instead, we just make the most pessimistic estimation for the ordering index. I think this behavior is acceptable and reasonable. Especially when the ordering index also has an `IndexRangeScan` or `Selection`. If we choose to directly drop the ordering index access path, we might wrongly drop an obviously better access path.

## Test
1. In a query with `ORDER BY`, `LIMIT`, and `WHERE`, the optimizer chooses the `ORDER BY` index by default:
    1. If there is another index path whose selectivity is less than 0.1. After setting the variable to 0.1, the optimizer chooses the index path by default.
    2. If there is another table path whose selectivity is less than 0.1. After setting the variable to 0.1, the optimizer chooses the table path by default.
    3. If there is another index merge path whose selectivity is less than 0.1. After setting the variable to 0.1, the optimizer chooses the index merge path by default.
2. In a query with `ORDER BY`, `LIMIT`, and `WHERE`, there are two index paths, one of them only satisfies `WHERE` conditions, and the other one satisfies `ORDER BY` and some other `WHERE` conditions. The selectivities of both indexes are less than 0.1. After setting the variable to 0.1, the optimizer is able to choose both indexes.

time-and-fate:

Issue Number: close #42060

Please see the issue for details.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


planner/core/logical_plans.go
```diff
@@ -1236,6 +1236,8 @@ type DataSource struct {
 	// colsRequiringFullLen is the columns that must be fetched with full length.
 	// It is used to decide whether single scan is enough when reading from an index.
 	colsRequiringFullLen []*expression.Column
+
+	accessPathMinSelectivity float64
```
qw4990, line 1240: Better to add some comments here since `DataSource` is a large structure with lots of fields...

time-and-fate: Updated~

qw4990 (approved):
LGTM

planner/core/logical_plans.go
```diff
@@ -1582,6 +1587,8 @@ func (ds *DataSource) deriveIndexPathStats(path *util.AccessPath, _ []expression
 		} else {
 			path.CountAfterIndex = math.Max(path.CountAfterAccess*selectivity, ds.stats.RowCount)
 		}
+	} else {
+		path.CountAfterIndex = path.CountAfterAccess
```
xuyifangreeneyes, line 1591: We don't set `path.CountAfterIndex` when `path.IndexFilters == nil` before the PR. Would that cause some potential wrong cardinality estimation? (since I thought `CountAfterIndex` would always be properly set before...

xuyifangreeneyes (approved):

planner/core/logical_plans.go
time-and-fate: That also surprised me...
It looks like there's no big problem most times, but I'm not sure about some edge cases, like when there are no index conditions here, but there are index conditions caused by virtual columns in `addPushedDownSelection`.

PR #42158, opened 2023-03-13
br/restore: fail fast when meeting some error during restoration

### LINKED ISSUE #42157: Restore: BR should fast fail when encountering some fatal errors
## Enhancement
For now, BR would stall for a long time when the thread pool is heavy loaded. The principle just be like [this playground](https://go.dev/play/p/zpEWQZKDSaI).

YuJuncen:

Issue Number: close #42157

Problem Summary:
As the issue has described, after encountered an error, we still try to fire new requests to the thread pool.

This PR would give up issuing new requests to the thread pool when there are failed tasks.

### Check List

Tests 

- [ ] (Almost) No code
This PR is really hard to test: perhaps we need to figure out a effective way to mock the `restore.Client`.
If you want to verify these changes, you may need to make a environment where download fail very slow. In our case, that is all S3 requests from TiKV has been rejected by a firewall just DROP TCP packets(also not send some RST packet). 

### Release note

```release-note
Fixed a problem that made BR looks like be stuck when there are some network issues.
```

### DIFF
--- a/br/pkg/restore/client.go
+++ b/br/pkg/restore/client.go
@@ -1187,7 +1187,12 @@ func (rc *Client) RestoreSSTFiles(
 	for rangeFiles, leftFiles = drainFilesByRange(files, rc.fileImporter.supportMultiIngest); len(rangeFiles) != 0; rangeFiles, leftFiles = drainFilesByRange(leftFiles, rc.fileImporter.supportMultiIngest) {
 		filesReplica := rangeFiles
 		if ectx.Err() != nil {
-			log.Warn("Restoring encountered error and already failed, give up remained files.", zap.Int("remained", len(leftFiles)))
+			log.Warn("Restoring encountered error and already stopped, give up remained files.",
+				zap.Int("remained", len(leftFiles)),
+				logutil.ShortError(ectx.Err()))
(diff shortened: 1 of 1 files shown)


Leavrth (approved):

br/pkg/restore/client.go
```diff
@@ -1206,6 +1215,12 @@ func (rc *Client) RestoreSSTFiles(
 		)
 		return errors.Trace(err)
 	}
+	// Once the parent context canceled and there is no task running in the errgroup,
+	// we may break the for loop without error in the errgroup. (Will this happen?)
+	// At that time, return the error in the context here.
+	if err := ectx.Err(); err != nil {
```
3pointer, line 1221: I'm not understand when will this error happen. what if we just return nil?

YuJuncen: > what if we just return nil?
The user may get some confusing error message.

3pointer (approved):

YuJuncen:
tests/realtikvtest/brietest/backup_restore_test.go:43
[schema:1050]Table 'br.t1' already exists
🤔

YuJuncen:
It seems the two failed BRIE tests covers all cases... For saving time, let merge it.
/merge

PR #42165, opened 2023-03-13
resource_control: support calibrate resource

glorv:

Issue Number: ref #38825

Problem Summary:

This PR add a new statement `calibrate resource` to estimate the total Request-Units(RU) of the current cluster.
Because the total ru usage is related to workload resource consuming, so the maximum RU can be different with different workload. Thus, the maximum RU estimated by this PR is based on a given workload -- TPC-C, and we may support other workload(e.g. sysbench) in the future.

In general, the bottle of a cluster can be one of TiDB CPU, TiKV CPU, TiKV IO Bandwidth. Currently, we can get the exact IO bandwidth and for most workload, io is unlikely to be the bottleneck. So here, we only consider TiDB CPU or TiKV CPU as bottleneck.

For a specified workload, the resource consuming is linear co-related with each other. So this PR use pre-benchmarked data of each resource dimension to calculate the ru cost per 1 tikv cpu. So if tikv cpu is the bottleneck, then Max RU = max_ru_per_1_kv_cpu * Total_TiKV_CPU; if tidb cpu is the bottleneck, then we just decrease the total kv cpu with a certain portaion. 

The PR calculate the RU cost of different resource dimension separated so we can support calculate total ru with custom ru config and the expected RU capacity can reflect the RU config change. 

The current SQL UI is as follows(We may add more information in the future version):
```
mysql> calibrate resource;
+-------+
| QUOTA |
+-------+
| 68569 |
+-------+
1 row in set (0.18 sec)
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [x] Contains syntax changes

### Release note

```release-note
Support estimate cluster total request unit with `calibrate resource`
```


glorv:
@nolouch @BornChanger PTAL

nolouch (commented):
Overall LGTM.

executor/calibrate_resource.go
```diff
@@ -0,0 +1,183 @@
...
+// workloadBaseRUCostMap contains the base resource cost per 1 kv cpu,
+// the data is calculated by the benchmark result.
+var workloadBaseRUCostMap = map[string]*baseResourceCost{
+	"tpcc": {
+		tidbCPU:       0.6,
+		kvCPU:         0.15,
+		readBytes:     units.MiB / 2, // 0.5MiB
+		writeBytes:    units.MiB,     // 1MiB
+		readReqCount:  300,
+		writeReqCount: 1750,
```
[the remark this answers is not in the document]
glorv: Yes. It is based on benchmark result.

nolouch (approved):

executor/calibrate_resource.go
```diff
@@ -0,0 +1,191 @@
...
+	totalKVCPUQuota, err := getTiKVTotalCPUQuota(ctx, exec)
+	if err != nil {
+		return err
+	}
+	totalTiDBCPU, err := getTiDBTotalCPUQuota(ctx, exec)
+	if err != nil {
+		return err
+	}
+
+	workload := "tpcc"
```
JmPotato, line 98: What about using a const or defined type?

glorv: done

executor/calibrate_resource.go
```diff
@@ -0,0 +1,199 @@
...
+		return nil, errors.Trace(err)
+	}
+	if len(rows) == 0 {
+		return nil, errors.New("PD request-unit config not found")
+	}
+	var nameIdx, valueIdx int
+	for i, f := range fields {
+		switch f.ColumnAsName.L {
+		case "instance":
+			//instanceIdx = i
```
Connor1996, line 146: please clean it

JmPotato (approved):

tiancaiamao (approved):
The result could be inaccurate.
It has many hypothesis, like the bottlenect is TiDB | TiKV CPU, like the workload assumption, like the performance on different hardware...

glorv:
> The result could be inaccurate. It has many hypothesis, like the bottlenect is TiDB | TiKV CPU, like the workload assumption, like the performance on different hardware...

Yes. This is the restriction of the current implementation. We plan to expand this command to support estimating the RU capacity based on user's workload dynamically, this should be more useful for the user.

Issue #42121, opened 2023-03-12
REPLACE statement has unexpected behavior in concurrent transactions


sayJason:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```sql
/* init */ CREATE TABLE t0(c0 DOUBLE UNIQUE);
/* init */ INSERT INTO t0(c0) VALUES (9.1);

/* tx1 */ BEGIN;
/* tx1 */ INSERT INTO t0(c0) VALUES (2.0);
/* tx1 */ REPLACE INTO t0(c0) VALUES (9.1);
/* tx2 */ BEGIN;
/* tx2 */ DELETE FROM t0;
/* tx2 */ COMMIT;
/* tx1 */ COMMIT;
/* tx1 */ SELECT * FROM t0;
```

### 2. What did you expect to see? (Required)

DELETE statement is blocked (in MySQL), or the final SELECT statement returns [(2.0), (9.1)].

### 3. What did you see instead (Required)

The final SELECT statement returns [(2.0)].

### 4. What is your TiDB version? (Required)

```sql
Release Version: v6.6.0
Edition: Community
Git Commit Hash: f4ca0821fb96a2bdd37d2fb97eb26c07fc58d4e4
Git Branch: heads/refs/tags/v6.6.0
UTC Build Time: 2023-02-17 14:49:02
GoVersion: go1.19.5
Race Enabled: false
TiKV Min Version: 6.2.0-alpha
Check Table Before Drop: false
Store: tikv
```

zyguan:
Thanks for your report, @sayJason . We forget to mark unchanged keys to be locked in `replace`, which leads to this bug.

Issue #25659, opened 2021-06-22
Frequently updating non-index column causing too many Lock records in the unique index key


MyonKeminta:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

This problem occurs if a table has:
* Never / rarely updated unique index
* Another column(s) that's frequently updated with point get in pessimistic transaction.

Example table definition: (without clustered index so that the primary key is a separated unique index)

```sql
create table t (id varchar(128), v int, primary key (id));
```

And there's frequent update operations like:

```sql
begin pessimistic;
update t set v = ? where id = ?;
commit;
```

### 2. What did you expect to see? (Required)

Everything goes well

### 3. What did you see instead (Required)

The performance decreases and it can be seen that `next` operation is very high some requests like `acquire_pessimistic_lock`  and `kv_get`.

Since this  locks the unique index key, it will produce a `WriteType::Lock` record in the write cf on the unique index key. When reading value on such a key, it will need to iterate over all these `Lock` records to find the latest `Put` or `Delete` record. Every time it iterates a Lock record, it performs a `next` operation. In some bad cases, reading each key may produce over 100k+ next operations.

### 4. What is your TiDB version? (Required)

4.0.10+, 5.x, master

cfzjywxk:
Since https://github.com/pingcap/tidb/pull/21229, the unique index key will be pessimistically locked, this change make the lock behaviour more reasonable but increased the possibility encountering the performance issue of the `LOCK` record at the tikv side.
So we may need to:
- prepare a switch controling the lock behaviour for the unique index read, also a hotfix patch is needed in case some users may encouter the performance issues using old versions.
- optimize the issue related to the `LOCK` record, there is already some plans for it.

sticnarf:
@cfzjywxk To solve exactly the problem introduced by #21229, I think we can just use `PUT` instead `LOCK` to write the unchanged index.

The index record should be small, so it does not bring too much overhead to put the index record again. With clustered index, the index can be larger, but typically it won't.

cfzjywxk:
> @cfzjywxk To solve exactly the problem introduced by #21229, I think we can just use `PUT` instead `LOCK` to write the unchanged index.
> 
> The index record should be small, so it does not bring too much overhead to put the index record again. With clustered index, the index can be larger, but typically it won't.

Agree,  this seems to be the plan with less changing risk, the TODOs are:
[...]

PR #42210, opened 2023-03-14
executor: lock duplicated keys on insert-ignore & replace-nothing

zyguan:

Issue Number: close #42121 

Problem Summary: When doing insert-ingore or replace without changes, tidb doesn't lock conflict keys properly.

Lock keys even when insert-ingnore meets duplicated rows or replace without changes.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

**The change may introduce more lock records, which may further lead performance regression like https://github.com/pingcap/tidb/issues/25659 .**

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/executor/insert_common.go
+++ b/executor/insert_common.go
@@ -1208,7 +1208,7 @@ func (e *InsertValues) batchCheckAndInsert(ctx context.Context, rows [][]types.D
 				} else {
 					e.ctx.GetSessionVars().StmtCtx.AppendWarning(r.handleKey.dupErr)
 					if txnCtx := e.ctx.GetSessionVars().TxnCtx; txnCtx.IsPessimistic {
-						// lock duplicated row key on insert-ingore
+						// lock duplicated row key on insert-ignore
 						txnCtx.AddUnchangedRowKey(r.handleKey.newKey)
 					}
 					continue
--- a/executor/insert_common.go
+++ b/executor/insert_common.go
@@ -1207,6 +1207,10 @@ func (e *InsertValues) batchCheckAndInsert(ctx context.Context, rows [][]types.D
 					}
 				} else {
 					e.ctx.GetSessionVars().StmtCtx.AppendWarning(r.handleKey.dupErr)
+					if txnCtx := e.ctx.GetSessionVars().TxnCtx; txnCtx.IsPessimistic {
+						// lock duplicated row key on insert-ingore
+						txnCtx.AddUnchangedRowKey(r.handleKey.newKey)
+					}
 					continue
 				}
 			} else if !kv.IsErrNotFound(err) {
@@ -1239,6 +1243,10 @@ func (e *InsertValues) batchCheckAndInsert(ctx context.Context, rows [][]types.D
 				} else {
 					// If duplicate keys were found in BatchGet, mark row = nil.
 					e.ctx.GetSessionVars().StmtCtx.AppendWarning(uk.dupErr)
+					if txnCtx := e.ctx.GetSessionVars().TxnCtx; txnCtx.IsPessimistic {
+						// lock duplicated unique key on insert-ingore
+						txnCtx.AddUnchangedRowKey(uk.newKey)
+					}
 					skip = true
 					break
 				}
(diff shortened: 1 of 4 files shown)


executor/insert_common.go
```diff
@@ -1239,6 +1243,10 @@ func (e *InsertValues) batchCheckAndInsert(ctx context.Context, rows [][]types.D
 				} else {
 					// If duplicate keys were found in BatchGet, mark row = nil.
 					e.ctx.GetSessionVars().StmtCtx.AppendWarning(uk.dupErr)
+					if txnCtx := e.ctx.GetSessionVars().TxnCtx; txnCtx.IsPessimistic {
+						// lock duplicated unique key on insert-ingore
+						txnCtx.AddUnchangedRowKey(uk.newKey)
```
zyguan, line 1248: It seems mysql only locks unique keys as well.
```sql
/* init */ select version();
-- init >> +-----------+
-- init    | version() |
-- init    +-----------+
-- init    | 8.0.28    |
-- init    +-----------+
/* init */ drop table if exists t;
-- init >> 0 rows affected
/* init */ create table t (a int primary key, b int unique key, c int);
-- init >> 0 rows affected
/* init */ insert into t values (1, 1, 1);
-- init >> 1 rows affected
/* t1 */ begin;
-- t1 >> 0 rows affected
/* t2 */ begin;
-- t2 >> 0 rows affected
/* t1 */ insert ignore into t values (2, 1, 1);
-- t1 >> 0 rows affected
/* t2 */ select b from t where a = 1 for update;
-- t2 >> +---+
-- t2    | b |
-- t2    +---+
-- t2    | 1 |
-- t2    +---+
/* t2 */ select a from t where b = 1 for update;
-- t2 >> blocked
/* t1 */ commit;
-- t1 >> 0 rows affected
-- t2 >> resumed
-- t2 >> +---+
-- t2    | a |
-- t2    +---+
-- t2    | 1 |
-- t2    +---+
/* t2 */ commit;
-- t2 >> 0 rows affected
```

executor/insert_common.go
```diff
@@ -1207,6 +1207,10 @@ func (e *InsertValues) batchCheckAndInsert(ctx context.Context, rows [][]types.D
 					}
 				} else {
 					e.ctx.GetSessionVars().StmtCtx.AppendWarning(r.handleKey.dupErr)
+					if txnCtx := e.ctx.GetSessionVars().TxnCtx; txnCtx.IsPessimistic {
+						// lock duplicated row key on insert-ingore
```
you06, line 1211:
```suggestion
						// lock duplicated row key on insert-ignore
```

executor/insert_common.go
cfzjywxk: The behaviors of this part are not described in the documents yet, we could add them there.

cfzjywxk (approved):
LGTM

tests/realtikvtest/pessimistictest/pessimistic_test.go
```diff
@@ -539,7 +539,7 @@ func TestOptimisticConflicts(t *testing.T) {
 	tk.MustExec("begin pessimistic")
 	// This SQL use BatchGet and cache data in the txn snapshot.
 	// It can be changed to other SQLs that use BatchGet.
-	tk.MustExec("insert ignore into conflict values (1, 2)")
+	tk.MustExec("select * from conflict where id in (1, 2, 3)")
```
you06, line 542: I think this query can be removed.

you06 (approved):

tests/realtikvtest/pessimistictest/pessimistic_test.go
zyguan: May I know why the query is unnecessary? IMO, it mean to test the [snapshot cache](https://github.com/tikv/client-go/blob/master/txnkv/txnsnapshot/snapshot.go#L132). We cache (1, 2) here and don't want to see the following select-for-update (L547) returns the cached row. Although, it won't read from the cache even when we forget to invalidate the cache because it will read from lock cache firstly now (also, we won't reuse the snapshot with for-update-ts now). I think it's ok to keep the query since the test point still makes sense.

PR #42229, opened 2023-03-14
Optimizer: Fix Merge In and Not Equal ranges #39676

### LINKED ISSUE #39676: Merge In and Not Equal ranges
## Bug Report

create table t(id bigint primary key, a varchar(256), b int, index(a), index(b));
explain select * from t where b in (1, 2) and b <> 1;

scan only b = 2

scans b in [1,2] and then filter out 1.

image

The actual problem is `1` has millions of record and I can't remove it from in expression.

5.7.25-TiDB-v6.3.0-serverless

ghazalfamilyusa:

Issue Number: ref #39676

Problem Summary:

### Background

TiDB optimizer has limited predicate simplifications and does not support many cases. One example is filtering redundant conditions across <> and In for same column. For example, a <> 1 and a in (1,2,3) can be simplified to a in (2,3). Also, a <> 5 and a in (1,2,3) can be simplified to a in (1,2,3) 

### Solution 
We added a new place holder logical rewrite optimization for predicate simplification. The general solution is 
1- Process all predicates as a set of ranges for all columns 
2- build equivalence classes for columns based col = col 
3- Merge all ranges of each equivalence class 
4- Build range predicates for each column in each equivalence class
This solution is on the optimizer roadmap and similar to what Teradata did https://docs.teradata.com/r/8mHBBLGP88~HK9Auie2QvQ/dinTQRLUmuWh_KUBr7ffdg

In this PR, we only do <> and In list merge to address the customer issue.  The algorithm is simple and applied only to table scan predicates. The main idea is for each pair of conditions of a <> constant_0 and a in (constant_1, ... constant_n) we 
- Remove the a <> constant_0 since it is always reduntant
- Remove constant_0 from the in list if any. We keep at least one element in the in list and postpone the <dualtable> (no table access) for later on. 

Tests 

- [ ] Unit test

Side effects
N/A

Documentation

N/A

### Release note

```release-note
None
```

### DIFF
--- a/planner/core/casetest/testdata/predicate_simplification_out.json
+++ b/planner/core/casetest/testdata/predicate_simplification_out.json
@@ -94,6 +94,14 @@
         ]
       },
       {
+        "SQL": "select f from t use index() where f <> 4 and f in (1,2,3) -- No intersection but <> is redundant",
+        "Plan": [
+          "TableReader 30.00 root  data:Selection",
+          "└─Selection 30.00 cop[tikv]  in(test.t.f, 1, 2, 3)",
+          "  └─TableFullScan 10000.00 cop[tikv] table:t keep order:false, stats:pseudo"
+        ]
+      },
+      {
         "SQL": "select f from t use index() where f in (1,2,3) and f <> 2 -- Simple case with different order. Test case for issue 39676",
         "Plan": [
           "TableReader 20.00 root  data:Selection",
@@ -128,8 +136,8 @@
       {
         "SQL": "select f from t use index() where f <> 3 and f in (1,2,3) and f <> 1 and f <> 2 -- Multiple <> values and cover whole inlist. We keep at least one in inlist",
         "Plan": [
-          "TableReader 10.00 root  data:Selection",
-          "└─Selection 10.00 cop[tikv]  in(test.t.f, 2)",
+          "TableReader 0.00 root  data:Selection",
+          "└─Selection 0.00 cop[tikv]  in(test.t.f, 2), ne(test.t.f, 2)",
           "  └─TableFullScan 10000.00 cop[tikv] table:t keep order:false, stats:pseudo"
         ]
       },
--- a/planner/core/casetest/testdata/predicate_simplification_out.json
+++ b/planner/core/casetest/testdata/predicate_simplification_out.json
@@ -1,0 +1,183 @@
+[
+  {
+    "Name": "TestRemoveRedundantPredicates",
+    "Cases": [
+      {
+        "SQL": "select f from t use index() where f = 1 and f = 1 -- simple redundancy of exact condition",
+        "Plan": [
+          "TableReader 10.00 root  data:Selection",
+          "└─Selection 10.00 cop[tikv]  eq(test.t.f, 1)",
+          "  └─TableFullScan 10000.00 cop[tikv] table:t keep order:false, stats:pseudo"
+        ]
+      },
+      {
+        "SQL": "select f from t use index() where f = 1 and f = 2 -- unsatisfiable condition",
+        "Plan": [
+          "TableDual 0.00 root  rows:0"
+        ]
+      },
+      {
+        "SQL": "select f from t use index() where f = 1 and f in (1,2,3)  -- intersection of in and =",
+        "Plan": [
+          "TableReader 10.00 root  data:Selection",
+          "└─Selection 10.00 cop[tikv]  eq(test.t.f, 1)",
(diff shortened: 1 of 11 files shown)


planner/core/rule_predicate_simplification.go
```diff
@@ -0,0 +1,157 @@
...
+	// which can be optimized with LogicalDual. But, this ia already done. TODO: the false
+	// optimization and its propagation through query tree will be added part of predicate simplification.
+	if len(newValues) < 2 {
+		newValues = append(newValues, lastValue)
+	}
+	newPred := expression.NewFunctionInternal(v.GetCtx(), v.FuncName.L, v.RetType, newValues...)
+	return newPred
+}
+
+func indexInSlice(index int, list []int) bool {
```
fixdb, line 108: You can use `slices.Contains` by importing `golang.org/x/exp/slices`.

fixdb (commented):
Other than the comments, looks good to me.

planner/core/rule_predicate_simplification.go
```diff
@@ -0,0 +1,157 @@
...
+// in-list and not equal list intersection.
+type predicateSimplification struct {
+}
+
+type predicateType = byte
+
+const (
+	inListPredicate   predicateType = 0x00
+	notEqualPredicate predicateType = 0x01
+	otherPredicate    predicateType = 0x02
```
winoros, line 34:
```suggestion
	inListPredicate   predicateType = iota
	notEqualPredicate
	otherPredicate 
```
In golang, this syntactic sugar should work

planner/core/rule_predicate_simplification.go
```diff
@@ -0,0 +1,157 @@
...
+		args := v.GetArgs()
+		col, colOk := args[0].(*expression.Column)
+		if !colOk {
+			return nil, otherPredicate
+		}
+		if v.FuncName.L == ast.NE {
+			if _, ok := args[1].(*expression.Constant); !ok {
+				return nil, otherPredicate
+			}
+			return col, notEqualPredicate
```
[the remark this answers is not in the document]
ghazalfamilyusa: The check is for more than the function name like col in (list of constants) and col <> constant.

planner/core/rule_predicate_simplification.go
ghazalfamilyusa: Sounds good. Will be done in next round.

planner/core/rule_predicate_simplification.go
ghazalfamilyusa: Sounds good. Will be done in next round.

fixdb (approved):
+1

planner/core/rule_predicate_simplification.go
```diff
@@ -0,0 +1,149 @@
...
+		redudantValue := valueOK && value.Equal(v.GetCtx(), notEQValue)
+		if !redudantValue {
+			newValues = append(newValues, element)
+		}
+		if valueOK {
+			lastValue = value
+		}
+	}
+	// Special case if all IN list values are prunned. Ideally, this is False condition
+	// which can be optimized with LogicalDual. But, this ia already done. TODO: the false
```
winoros, line 100:
```suggestion
	// which can be optimized with LogicalDual. But, this is already done. TODO: the false
```

planner/core/casetest/testdata/predicate_simplification_out.json
```diff
@@ -0,0 +1,183 @@
...
+        ]
+      },
+      {
+        "SQL": "select f from t use index() where f <> 3 and f in (1,2,3) and f <> 1 and f <> 2 -- Multiple <> values and cover whole inlist. We keep at least one in inlist",
+        "Plan": [
+          "TableReader 10.00 root  data:Selection",
+          "└─Selection 10.00 cop[tikv]  in(test.t.f, 2)",
+          "  └─TableFullScan 10000.00 cop[tikv] table:t keep order:false, stats:pseudo"
+        ]
+      },
```
winoros, line 135: We get the wrong result in this case? If we keep `f in (2)`, we should also keep `f != 2`.

ghazalfamilyusa: You are right and we need to keep both. Good catch. Fixed.

planner/core/rule_predicate_simplification.go
ghazalfamilyusa: Fixed

winoros (approved):

Issue #42245, opened 2023-03-15
Can't set TiFlash replica in disaggregated mode.


hehechen:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
![img_v2_fd5646c9-a7d0-4637-95d0-28543782cc8g](https://user-images.githubusercontent.com/14118780/225222187-84afa18c-b8b8-4a72-824b-fe8608be19cc.jpg)

### 2. What did you expect to see? (Required)
Set TiFlash replica successfully.
### 3. What did you see instead (Required)
Failed because check serverType failed.
### 4. What is your TiDB version? (Required)
6bc625c36e0c186e9500372bbd5d9ec4f9544877

PR #41998, opened 2023-03-07
*: display tiflash_write in cluster_info

hehechen:
### What problem does this PR solve?

Issue Number: close #42064

Problem Summary:
https://github.com/pingcap/tiflash/issues/7007
When the label of TiFlash contains `engine_role`:`write`, it means that this TiFlash is a `tiflash_write` node. So `tiflash_write` should be displayed in `information_schema.cluster_info`.
### What is changed and how it works?
Check the label and display `tiflash_write`.
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)
image
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #42325, opened 2023-03-16
*:  Fix TiFlash serverType check in disaggregated mode

hehechen:

Issue Number: ref #42245

Problem Summary:
In https://github.com/pingcap/tidb/pull/41998, I modified `serverType` in `GetStoreServerInfo`, but there are several problems such as #42245. So I think it's better to only modify `serverType` in `dataForTiDBClusterInfo`.
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
select * from information_schema.cluster_info;
alter table test.t1 set tiflash replica 1;
select count(*) from TIFLASH_TABLES;
select count(*) from TIFLASH_SEGMENTS;
alter table test.t1 compact tiflash replica ;
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/ddl/ddl_api.go
+++ b/ddl/ddl_api.go
@@ -5741,7 +5741,7 @@ func isTableTiFlashSupported(schema *model.DBInfo, tb table.Table) error {
 
 func checkTiFlashReplicaCount(ctx sessionctx.Context, replicaCount uint64) error {
 	// Check the tiflash replica count should be less than the total tiflash stores.
-	tiflashStoreCnt, err := infoschema.GetTiFlashWriteStoreCount(ctx)
+	tiflashStoreCnt, err := infoschema.GetTiFlashStoreCount(ctx)
 	if err != nil {
 		return errors.Trace(err)
 	}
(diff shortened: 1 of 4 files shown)


guo-shaoge (approved):

breezewish (approved):

ddl/ddl_api.go
```diff
@@ -5741,7 +5741,7 @@ func isTableTiFlashSupported(schema *model.DBInfo, tb table.Table) error {
 
 func checkTiFlashReplicaCount(ctx sessionctx.Context, replicaCount uint64) error {
 	// Check the tiflash replica count should be less than the total tiflash stores.
-	tiflashStoreCnt, err := infoschema.GetTiFlashWriteStoreCount(ctx)
+	tiflashStoreCnt, err := infoschema.GetTiFlashStoreCount(ctx)
```
breezewish, line 5744: Actually I believe we still need to only count write nodes here. But we could fix it (along with elsewhere that requires write nodes) later.

hehechen: > Actually I believe we still need to only count write nodes here. But we could fix it (along with elsewhere that requires write nodes) later.

We don't modify `serverType` now, the `serverType` of `tiflash_write` is `tiflash`, so `tiflashStoreCnt` includes the count of `tiflash_write` and non-disaggregated TiFlash.

Issue #40499, opened 2023-01-11
integrate lightning with LOAD DATA to support more file types and storage


lance6716:
## Enhancement

LOAD DATA can load csv/parquet/sql type data files, with some compression, from s3/gcs...

more details will be posted later

forsaken628:
#40489 @lance6716

hey,
csv escape double quote by preceding it with another double quote, not \" 

https://www.rfc-editor.org/rfc/rfc4180

lance6716:
> csv escape double quote by preceding it with another double quote

we also support it

> Introduce the "format" key word 

Yes this is in the plan

> some settings for single sql as function args

Can you explain which settings do you want?

forsaken628:
Just that test case on #40489 make me confusing，no more.

lance6716:
> Just that test case on #40489 make me confusing，no more.

Hi, I have updated the issue to include the first step to change the parser syntax.

PR #42330, opened 2023-03-16
*: LOAD DATA returns multiple errors in one error message

lance6716:

Issue Number: ref #40499

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

```
MySQL [(none)]> use test

Database changed
MySQL [test]> load data local infile '/tmp/123' into table ttt with aaa = 1, detached;
ERROR 8173 (HY000): Found multiple errors in precheck:
  ERROR 1146 (42S02): Table 'test.ttt' doesn't exist
  ERROR 8163 (HY000): Unknown option aaa
  ERROR 8172 (HY000): The job can not be DETACHED when LOAD DATA LOCAL INFILE
```

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


D3Hunter (commented):
i think those errors're normal errors before precheck, we should return them once when we met them, most other sql do this in the same way.

the precheck should be like whether tikv left capacity is enough / table is empty.....

lance6716:
> i think those errors're normal errors before precheck, we should return them once when we met them, most other sql do this in the same way.
> 
> the precheck should be like whether tikv left capacity is enough / table is empty.....

Yes I also think it's weird. Let's ask PM in the meeting.

lance6716:
will refine user interaction later

PR #42366, opened 2023-03-17
br: Skip verify region if the provider is ceph or minio

### LINKED ISSUE #42033: BR - S3 Endpoint is not set when get bucket informations
## Bug Report

Hi,

We get following error when we try to backup our TiDB on s3 :

`Error: s3 bucket and region are not matched, bucket=ringover-eu-2023, input region=fr-par, real region=us-east-1`

After checking the code, i think it's because the endpoint is not set when calling `GetBucketRegionWithClient` here : https://github.com/pingcap/tidb/blob/master/br/pkg/storage/s3.go#L374

My guess is that you need a `WithEndpoint` when creating `aws.Config` there : https://github.com/pingcap/tidb/blob/master/br/pkg/storage/s3.go#L357

You can check this comment in AWS SDK : https://github.com/aws/aws-sdk-go/blob/main/service/s3/s3manager/bucket_region.go#L91

### Use BR to backup DB

Try to backup with BR and specify a endpoint : 

`tiup br backup full --pd "$PDIP:$PDPORT"  --s3.endpoint "https://s3.fr-par.scw.cloud/" -s "s3://bucket1/TEST1" --s3.region "fr-par" --send-credentials-to-tikv=true --ca "$CAFILE" --key "$KEYFILE" --cert "$CERTFILE"`

This is the cli we use, the "bucket1" exists and the region `fr-par` is the good one, you can see the result of an aws cli call : 
 
`$ aws s3api get-bucket-location --bucket bucket1 --profile scw
{
    "LocationConstraint": "fr-par"
}`

### The backup fails

We get following : 
`Error: s3 bucket and region are not matched, bucket=ringover-eu-2023, input region=fr-par, real region=us-east-1`

The region returned is not the same than the provided in cli because the endpoint is not set

### TiDB 6.5

Everything works well before 6.5

regards

Fife86:

Issue Number: close #42033

Problem Summary:

When we use br with our infra, our endpoint didn't provide the correct region to our infra. In fact, it send us no region at all. The region test was failing, so we decided to skip this check for our ceph and minio providers.

We skip the region check in br if the provider is ceph or minio

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/storage/s3.go
+++ b/br/pkg/storage/s3.go
@@ -177,6 +177,7 @@ func (options *S3BackendOptions) Apply(s3 *backuppb.S3) error {
 	}
 
 	s3.Endpoint = strings.TrimSuffix(options.Endpoint, "/")
+	s3.Provider = options.Provider
 	s3.Region = options.Region
 	// StorageClass, SSE and ACL are acceptable to be empty
 	s3.StorageClass = options.StorageClass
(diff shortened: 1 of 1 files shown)


br/pkg/storage/s3.go
```diff
@@ -358,38 +359,45 @@ func NewS3Storage(ctx context.Context, backend *backuppb.S3, opts *ExternalStora
...
 	c := s3.New(ses, s3CliConfigs...)
-	confCred := ses.Config.Credentials
-	setCredOpt := func(req *request.Request) {
-		// s3manager.GetBucketRegionWithClient will set credential anonymous, which works with s3.
-		// we need reassign credential to be compatible with minio authentication.
-		if confCred != nil {
-			req.Config.Credentials = confCred
+
+	// s3manager.GetBucketRegionWithClient failed to get region from client if the client is ceph or minio
+	// It defaut back to us-west-1 which break the code if a region is already set
```
dveeden, line 364:
```suggestion
	// It defaults back to us-west-1 which breaks the code if a region is already set
```

br/pkg/storage/s3.go
```diff
@@ -358,38 +359,45 @@ func NewS3Storage(ctx context.Context, backend *backuppb.S3, opts *ExternalStora
...
+	if qs.Provider != "ceph" && qs.Provider != "minio" {
```
dveeden, line 365: 1. Not sure if excluding ceph and minio is the right solution, maybe we should check if `qs.Provider == "s3" ` (or GCS?)
2. Maybe we should separate the provider config from the logic here. And check if `ProviderConfig.check_bucket_with_region == true` and then elsewhere store the config for each provider.

These are just ideas, I have no objections on merging this as-is.

br/pkg/storage/s3.go
Fife86: Thanks my mistakes ^^"

br/pkg/storage/s3.go
Fife86: 1 . I'm not sure about the list. We add ceph and minio because it's what we are working with and had troubles, but the issues may be wider.
2 Would it be simpler to add a parameter like qs.SkipBucketRegionCheck and a tag --skip-bucket-region-check when we launch br ? ( the wording is a bit long ^^")

dveeden:
Could you add a release note?

dveeden:
@Fife86 Looks like author email that you used for 224237091966350c8c53c2c7efff5a4f6c5247ea isn't linked to your GitHub account causing the CLA checks to fail.

Please try to add that email address to your GitHub account or squash the commits and set the author to be the email address you use for GitHub.

br/pkg/storage/s3.go
```diff
@@ -358,38 +359,45 @@ func NewS3Storage(ctx context.Context, backend *backuppb.S3, opts *ExternalStora
...
-	setCredOpt := func(req *request.Request) {
-		// s3manager.GetBucketRegionWithClient will set credential anonymous, which works with s3.
-		// we need reassign credential to be compatible with minio authentication.
-		if confCred != nil {
-			req.Config.Credentials = confCred
+
+	// s3manager.GetBucketRegionWithClient failed to get region from client if the client is ceph or minio
+	// It defaults back to us-west-1 which breaks the code if a region is already set
+	if qs.Provider != "ceph" && qs.Provider != "minio" {
+
```
dveeden, line 366:
```suggestion
```

From the CI checks:
```
br/pkg/storage/s3.go:365:53: empty-lines: extra empty line at the start of a block (all_revive)
```

Leavrth (approved):

Leavrth:
I have added the default `release note` into the first comment, like this:
``` release-note
None
```

you can modify the `None` if in need, refer to https://github.com/pingcap/tidb/pull/42366#issuecomment-1480722645

dveeden:
Thank you for the pull request, issue and all the work you did on this. Let's close this as a very similar PR has been merged that should fix the issue. Feel free to add a note here or on the issue if testing shows that this isn't completely fixed or if there are any other concerns. Thank you for choosing TiDB!

PR #42432, opened 2023-03-21
bootstrap: use BatchCreateTableWithInfo to speed up

lance6716:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/session/bootstrap.go
+++ b/session/bootstrap.go
@@ -40,6 +40,7 @@ import (
 	"github.com/pingcap/tidb/meta"
 	"github.com/pingcap/tidb/owner"
 	"github.com/pingcap/tidb/parser"
+	"github.com/pingcap/tidb/parser/ast"
 	"github.com/pingcap/tidb/parser/auth"
 	"github.com/pingcap/tidb/parser/model"
 	"github.com/pingcap/tidb/parser/mysql"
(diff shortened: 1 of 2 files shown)


executor/executor_test.go
```diff
@@ -5663,9 +5663,10 @@ func TestAdmin(t *testing.T) {
 	tk.MustExec("use test")
 	tk.MustExec("drop table if exists admin_test2")
 	tk.MustExec("create table admin_test2 (c1 int, c2 int, c3 int default 1, index (c1))")
+	// we cn only see last 10 queries
```
lance6716, line 5666:
```suggestion
	// we can only see last 10 queries
```

Due to BatchCreateTableWithInfo, we have less history DDL jobs

tiancaiamao (commented):
This affects not only the test logic, but also the real bootstrap processing.
I have another idea: can we use a bootstraped store as the image for testing, so the bootstrap in testing can be totally avoided?

lance6716:
> This affects not only the test logic, but also the real bootstrap processing. I have another idea: can we use a bootstraped store as the image for testing, so the bootstrap in testing can be totally avoided?

Yes that seems safer but I don't know how to do it. I guess I should prepare a bootstrapped KV store in some `init()`, and copy all KV to newly created test store in https://github.com/pingcap/tidb/blob/273763b51e8d887168d0888e4c7b1b3a5f884e10/store/mockstore/mockstore.go#L165 . Can you help me checking this idea?

tiancaiamao:
> > This affects not only the test logic, but also the real bootstrap processing. I have another idea: can we use a bootstraped store as the image for testing, so the bootstrap in testing can be totally avoided?
> 
> Yes that seems safer but I don't know how to do it. I guess I should prepare a bootstrapped KV store in some `init()`, and copy all KV to newly created test store in
> 
> https://github.com/pingcap/tidb/blob/273763b51e8d887168d0888e4c7b1b3a5f884e10/store/mockstore/mockstore.go#L165
> 
> . Can you help me check this idea? (For example, can all KV be copied?)

OK, let me have a try.

lance6716:
can be replaced by https://github.com/pingcap/tidb/pull/42521

PR #42433, opened 2023-03-21
ddl: fix truncate partition for global index

### LINKED ISSUE #42435: ddl: truncate partition not handled correctly for global index
## Bug Report

A unit test: 
```golang
func TestTruncatePartitionWithGlobalIndex(t *testing.T) {
	defer config.RestoreFunc()()
	config.UpdateGlobal(func(conf *config.Config) {
		conf.EnableGlobalIndex = true
	})
	tk.MustExec("use test")
	tk.MustExec("drop table if exists test_global")
	tk.MustExec(`create table test_global ( a int, b int, c int)
	partition by range( a ) (
		partition p1 values less than (10),
		partition p2 values less than (20)
	);`)
	tt := external.GetTableByName(t, tk, "test", "test_global")
	pid := tt.Meta().Partition.Definitions[1].ID

	tk.MustExec("Alter Table test_global Add Unique Index idx_b (b);")
	tk.MustExec("Alter Table test_global Add Unique Index idx_c (c);")
	tk.MustExec(`INSERT INTO test_global VALUES (1, 1, 1), (2, 2, 2), (11, 3, 3), (12, 4, 4)`)

	tk.MustExec("alter table test_global truncate partition p2;")
	result := tk.MustQuery("select * from test_global;")
	result.Sort().Check(testkit.Rows(`1 1 1`, `2 2 2`))

	tt = external.GetTableByName(t, tk, "test", "test_global")
	idxInfo := tt.Meta().FindIndexByName("idx_b")
	require.NotNil(t, idxInfo)
	cnt := checkGlobalIndexCleanUpDone(t, tk.Session(), tt.Meta(), idxInfo, pid)
	require.Equal(t, 2, cnt)

	idxInfo = tt.Meta().FindIndexByName("idx_c")
	require.NotNil(t, idxInfo)
	cnt = checkGlobalIndexCleanUpDone(t, tk.Session(), tt.Meta(), idxInfo, pid)
	require.Equal(t, 2, cnt)
}
```

pass
failed, because global index rows was not cleaned.
image

L-maple:

Issue Number: close #42435

Problem Summary:

add DeleteOnly and DeleteReorg for truncate partition ddl, so that global index can be handled correctly.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/ddl/db_partition_test.go
+++ b/ddl/db_partition_test.go
@@ -2129,6 +2129,222 @@ func TestGlobalIndexUpdateInDropPartition(t *testing.T) {
 	tk.MustQuery("select * from test_global use index(idx_b) order by a").Check(testkit.Rows("2 11 11", "12 12 12"))
 }
 
+func TestTruncatePartitionWithGlobalIndex(t *testing.T) {
+	defer config.RestoreFunc()()
+	config.UpdateGlobal(func(conf *config.Config) {
+		conf.EnableGlobalIndex = true
+	})
+
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("drop table if exists test_global")
+	tk.MustExec(`create table test_global ( a int, b int, c int)
+	partition by range( a ) (
+		partition p1 values less than (10),
+		partition p2 values less than (20)
+	);`)
+	tt := external.GetTableByName(t, tk, "test", "test_global")
+	pid := tt.Meta().Partition.Definitions[1].ID
+
+	tk.MustExec("Alter Table test_global Add Unique Index idx_b (b);")
+	tk.MustExec("Alter Table test_global Add Unique Index idx_c (c);")
+	tk.MustExec(`INSERT INTO test_global VALUES (1, 1, 1), (2, 2, 2), (11, 3, 3), (12, 4, 4), (15, 15, 15)`)
+
+	tk2 := testkit.NewTestKit(t, store)
+	tk2.MustExec(`use test`)
+	tk2.MustExec(`begin`)
+	tk2.MustExec(`insert into test_global values (5,5,5)`)
+	syncChan := make(chan bool)
+	go func() {
+		tk.MustExec("alter table test_global truncate partition p2;")
+		syncChan <- true
+	}()
+	waitFor := func(i int, s string) {
+		for {
+			tk4 := testkit.NewTestKit(t, store)
+			tk4.MustExec(`use test`)
+			res := tk4.MustQuery(`admin show ddl jobs where db_name = 'test' and table_name = 'test_global' and job_type = 'truncate partition'`).Rows()
+			if len(res) == 1 && res[0][i] == s {
+				break
+			}
+			time.Sleep(10 * time.Millisecond)
+		}
+	}
+	waitFor(4, "delete only")
+	tk3 := testkit.NewTestKit(t, store)
+	tk3.MustExec(`begin`)
+	tk3.MustExec(`use test`)
+	tk3.MustQuery(`select b from test_global use index(idx_b) where b = 15`).Check(testkit.Rows())
+	tk3.MustQuery(`select c from test_global use index(idx_c) where c = 15`).Check(testkit.Rows())
+	// Here it will fail with
+	// the partition is not in public.
+	err := tk3.ExecToErr(`insert into test_global values (15,15,15)`)
+	assert.NotNil(t, err)
+	tk2.MustExec(`commit`)
+	tk3.MustExec(`commit`)
+	<-syncChan
+	result := tk.MustQuery("select * from test_global;")
+	result.Sort().Check(testkit.Rows(`1 1 1`, `2 2 2`, `5 5 5`))
+
+	tt = external.GetTableByName(t, tk, "test", "test_global")
+	idxInfo := tt.Meta().FindIndexByName("idx_b")
+	require.NotNil(t, idxInfo)
+	cnt := checkGlobalIndexCleanUpDone(t, tk.Session(), tt.Meta(), idxInfo, pid)
+	require.Equal(t, 3, cnt)
+
+	idxInfo = tt.Meta().FindIndexByName("idx_c")
+	require.NotNil(t, idxInfo)
+	cnt = checkGlobalIndexCleanUpDone(t, tk.Session(), tt.Meta(), idxInfo, pid)
+	require.Equal(t, 3, cnt)
+}
+
+func TestGlobalIndexUpdateInTruncatePartition(t *testing.T) {
+	defer config.RestoreFunc()()
+	config.UpdateGlobal(func(conf *config.Config) {
+		conf.EnableGlobalIndex = true
+	})
+	store, dom := testkit.CreateMockStoreAndDomain(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("set @@tidb_partition_prune_mode='dynamic'")
+	tk.MustExec("set @@session.tidb_analyze_version=2")
+	tk.MustExec("drop table if exists test_global")
+	tk.MustExec(`create table test_global ( a int, b int, c int)
+	partition by range( a ) (
+		partition p1 values less than (10),
+		partition p2 values less than (20),
+		partition p3 values less than (30)
+	);`)
+	tk.MustExec("alter table test_global add unique index idx_b (b);")
+	tk.MustExec("insert into test_global values (1, 1, 1), (8, 8, 8), (11, 11, 11), (12, 12, 12);")
+	tk.MustExec("analyze table test_global")
+
+	originalHook := dom.DDL().GetHook()
+	defer dom.DDL().SetHook(originalHook)
+
+	var err error
+	hook := &callback.TestDDLCallback{Do: dom}
+	hook.OnJobRunBeforeExported = func(job *model.Job) {
+		assert.Equal(t, model.ActionTruncateTablePartition, job.Type)
+		if job.SchemaState == model.StateDeleteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+			err = tk1.ExecToErr("update test_global set a = 2 where a = 11")
+			assert.NotNil(t, err)
+		}
+	}
+	dom.DDL().SetHook(hook)
+
+	tk.MustExec("alter table test_global truncate partition p1")
+}
+
+func TestGlobalIndexUpdateInTruncatePartition4Hash(t *testing.T) {
+	defer config.RestoreFunc()()
+	config.UpdateGlobal(func(conf *config.Config) {
+		conf.EnableGlobalIndex = true
+	})
+	store, dom := testkit.CreateMockStoreAndDomain(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("set @@tidb_partition_prune_mode='dynamic'")
+	tk.MustExec("set @@session.tidb_analyze_version=2")
+	tk.MustExec("drop table if exists test_global")
+	tk.MustExec(`create table test_global ( a int, b int, c int)
+	partition by hash(a) partitions 4;`)
+	tk.MustExec("alter table test_global add unique index idx_b (b);")
+	tk.MustExec("insert into test_global values (1, 1, 1), (8, 8, 8), (11, 11, 11), (12, 12, 12);")
+	tk.MustExec("analyze table test_global")
+
+	originalHook := dom.DDL().GetHook()
+	defer dom.DDL().SetHook(originalHook)
+
+	var err error
+	hook := &callback.TestDDLCallback{Do: dom}
+	hook.OnJobRunBeforeExported = func(job *model.Job) {
+		assert.Equal(t, model.ActionTruncateTablePartition, job.Type)
+		if job.SchemaState == model.StateDeleteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+			err = tk1.ExecToErr("update test_global set a = 1 where a = 12")
+			assert.NotNil(t, err)
+		}
+	}
+	dom.DDL().SetHook(hook)
+
+	tk.MustExec("alter table test_global truncate partition p1")
+}
+
+func TestGlobalIndexReaderInTruncatePartition(t *testing.T) {
+	defer config.RestoreFunc()()
+	config.UpdateGlobal(func(conf *config.Config) {
+		conf.EnableGlobalIndex = true
+	})
+	store, dom := testkit.CreateMockStoreAndDomain(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("drop table if exists test_global")
+	tk.MustExec(`create table test_global ( a int, b int, c int)
+	partition by range( a ) (
+		partition p1 values less than (10),
+		partition p2 values less than (20),
+		partition p3 values less than (30)
+	);`)
+	tk.MustExec("alter table test_global add unique index idx_b (b);")
+	tk.MustExec("insert into test_global values (1, 1, 1), (8, 8, 8), (11, 11, 11), (12, 12, 12);")
+	tk.MustExec("analyze table test_global")
+
+	hook := &callback.TestDDLCallback{Do: dom}
+	hook.OnJobRunBeforeExported = func(job *model.Job) {
+		assert.Equal(t, model.ActionTruncateTablePartition, job.Type)
+		if job.SchemaState == model.StateDeleteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+
+			tk1.MustQuery("select b from test_global use index(idx_b)").Sort().Check(testkit.Rows("11", "12"))
+		}
+	}
+	dom.DDL().SetHook(hook)
+
+	tk.MustExec("alter table test_global truncate partition p1")
+}
+
+func TestGlobalIndexInsertInTruncatePartition(t *testing.T) {
+	config.UpdateGlobal(func(conf *config.Config) {
+		conf.EnableGlobalIndex = true
+	})
+	store, dom := testkit.CreateMockStoreAndDomain(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("set @@tidb_partition_prune_mode='dynamic'")
+	tk.MustExec("set @@session.tidb_analyze_version=2")
+	tk.MustExec("drop table if exists test_global")
+	tk.MustExec(`create table test_global ( a int, b int, c int)
+	partition by range( a ) (
+		partition p1 values less than (10),
+		partition p2 values less than (20),
+		partition p3 values less than (30)
+	);`)
+	tk.MustExec("alter table test_global add unique index idx_b (b);")
+	tk.MustExec("insert into test_global values (1, 1, 1), (8, 8, 8), (11, 11, 11), (12, 12, 12);")
+	tk.MustExec("analyze table test_global")
+
+	var err error
+	hook := &callback.TestDDLCallback{Do: dom}
+	hook.OnJobRunBeforeExported = func(job *model.Job) {
+		assert.Equal(t, model.ActionTruncateTablePartition, job.Type)
+		if job.SchemaState == model.StateDeleteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+			err = tk1.ExecToErr("insert into test_global values(2, 2, 2)")
+			assert.NotNil(t, err)
+		}
+	}
+	dom.DDL().SetHook(hook)
+
+	tk.MustExec("alter table test_global truncate partition p1")
+}
+
 func TestGlobalIndexReaderInDropPartition(t *testing.T) {
 	defer config.RestoreFunc()()
 	config.UpdateGlobal(func(conf *config.Config) {
--- a/ddl/partition.go
+++ b/ddl/partition.go
@@ -1765,6 +1765,34 @@ func getTableInfoWithDroppingPartitions(t *model.TableInfo) *model.TableInfo {
 	return nt
 }
 
+// getTableInfoWithOriginalPartitions builds oldTableInfo including truncating partitions, only used by onTruncateTablePartition.
+func getTableInfoWithOriginalPartitions(t *model.TableInfo, oldIDs []int64, newIDs []int64) *model.TableInfo {
+	nt := t.Clone()
+	np := nt.Partition
+
+	// reconstruct original definitions
+	for _, oldDef := range np.DroppingDefinitions {
+		var newID int64
+		for i := range newIDs {
+			if oldDef.ID == oldIDs[i] {
+				newID = newIDs[i]
+				break
+			}
+		}
+		for i := range np.Definitions {
+			newDef := &np.Definitions[i]
+			if newDef.ID == newID {
+				newDef.ID = oldDef.ID
+				break
+			}
+		}
+	}
+
+	np.DroppingDefinitions = nil
+	np.NewPartitionIDs = nil
+	return nt
+}
+
 func dropLabelRules(_ *ddlCtx, schemaName, tableName string, partNames []string) error {
 	deleteRules := make([]string, 0, len(partNames))
 	for _, partName := range partNames {
@@ -1965,12 +1993,16 @@ func removeTiFlashAvailablePartitionIDs(tblInfo *model.TableInfo, pids []int64) 
 }
 
 // onTruncateTablePartition truncates old partition meta.
-func onTruncateTablePartition(d *ddlCtx, t *meta.Meta, job *model.Job) (int64, error) {
+func (w *worker) onTruncateTablePartition(d *ddlCtx, t *meta.Meta, job *model.Job) (int64, error) {
 	var ver int64
-	var oldIDs []int64
-	if err := job.DecodeArgs(&oldIDs); err != nil {
+	var oldIDs, newIDs []int64
+	if err := job.DecodeArgs(&oldIDs, &newIDs); err != nil {
 		job.State = model.JobStateCancelled
 		return ver, errors.Trace(err)
+	}
+	if len(oldIDs) != len(newIDs) {
+		job.State = model.JobStateCancelled
+		return ver, errors.Trace(errors.New("len(oldIDs) must be the same as len(newIDs)"))
 	}
 	tblInfo, err := GetTableInfoAndCancelFaultJob(t, job, job.SchemaID)
 	if err != nil {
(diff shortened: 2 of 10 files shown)


mjonss (commented):
Hi, can you give a high level description of how this will work?

I'm not sure this or the current handling of global index reorganization works properly, but I think you are moving it closer to a working solution.

One thing I don't think works currently is in the following scenario:
During the global index 'cleanup', what will happen to the sessions reading it?
I would assume that if the metadata has the old partition ids, then the index will be inconsistent with existing rows/partitions (since they are being deleted during the partitions are readable)
I would also assume that if the metadata has the new partition ids, then the index will still have records not yet deleted to partitions that no-longer exists.

Note that we allow currentSchemaVersion and currentSchemaVersion-1 as valid schema versions!

ddl/partition.go
```diff
@@ -1885,27 +1885,129 @@ func onTruncateTablePartition(d *ddlCtx, t *meta.Meta, job *model.Job) (int64, e
...
+	originalState := job.SchemaState
+	switch job.SchemaState {
+	case model.StatePublic:
+		truncatingPartitionIDs := make(map[int64]struct{}, len(oldIDs))
+		for _, oldID := range oldIDs {
+			truncatingPartitionIDs[oldID] = struct{}{}
+		}
+		pi.TruncatingPartitionIDs = truncatingPartitionIDs
+		job.SchemaState = model.StateDeleteOnly
+		ver, err = updateVersionAndTableInfo(d, t, job, tblInfo, originalState != job.SchemaState)
```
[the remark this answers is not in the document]
L-maple: resolved

L-maple:
> Hi, can you give a high level description of how this will work?
> 
> I'm not sure this or the current handling of global index reorganization works properly, but I think you are moving it closer to a working solution.
> 
> One thing I don't think works currently is in the following scenario: During the global index 'cleanup', what will happen to the sessions reading it? I would assume that if the metadata has the old partition ids, then the index will be inconsistent with existing rows/partitions (since they are being deleted during the partitions are readable) I would also assume that if the metadata has the new partition ids, then the index will still have records not yet deleted to partitions that no-longer exists.
> 
> Note that we allow currentSchemaVersion and currentSchemaVersion-1 as valid schema versions!

@mjonss Please review this PR again.
I solved the problem by saving the original Definitions in the new data memeber `OriginalDeifnitions`.
- Read: Because Definitions has no truncating definitions, so read is OK.
- Write: We can add check to ensure no record can be inserted into truncating definitions. How should we do it? We can locate partition in both `Defintions` and `OriginalDefinitions` in the metadata. If the result pids are different, we can ensure the write should be forbided.

L-maple:
I am in holiday now. I will finish it and request your review. Much thanks！



---- Replied Message ----
| From | Mattias ***@***.***> |
| Date | 05/01/2023 23:05 |
| To | ***@***.***> |
| Cc | Changjie ***@***.***>***@***.***> |
| Subject | Re: [pingcap/tidb] ddl: fix truncate partition for global index (PR #42433) |

/retest

L-maple:
@mjonss You can review now, thanks!

L-maple:
@mjonss @Defined2014 Hi, this PR can be reviewed~
I have a little question: What else should I do before global index can be a GA feature?

Defined2014:
> @mjonss @Defined2014 Hi, this PR can be reviewed~ I have a little question: What else should I do before global index can be a GA feature?

IMO, we should accomplish the following
1. Provide a complete test report (including all related parts, like all executors, dynamic pruning/static pruning, etc.)
2. Close all bugs above or equals to major severity
3. Do some paper works (internal)

Do you have anything to add? @mjonss @bb7133

ddl/partition.go
```diff
@@ -1911,27 +1911,157 @@ func onTruncateTablePartition(d *ddlCtx, t *meta.Meta, job *model.Job) (int64, e
...
 
-	newPartitions := make([]model.PartitionDefinition, 0, len(oldIDs))
-	for _, oldID := range oldIDs {
-		for i := 0; i < len(pi.Definitions); i++ {
-			def := &pi.Definitions[i]
-			if def.ID == oldID {
-				pid, err1 := t.GenGlobalID()
-				if err1 != nil {
-					return ver, errors.Trace(err1)
+	if job.State == model.JobStateRunning && job.SchemaState == model.StateNone {
```
Defined2014, line 1914: Should move it to
`ddl_api.go: func (d *ddl) TruncateTablePartition(ctx sessionctx.Context, ident ast.Ident, spec *ast.AlterTableSpec) error `

Defined2014: And also should consider `IsRollbackable` function. Do we need to support cancel `ActionTruncateTablePartition` job?

ddl/partition.go
```diff
@@ -1911,27 +1911,157 @@ func onTruncateTablePartition(d *ddlCtx, t *meta.Meta, job *model.Job) (int64, e
...
+				}
+			}
+		}
+		if len(newPartitions) == 0 {
+			job.State = model.JobStateCancelled
+			return ver, table.ErrUnknownPartition.GenWithStackByArgs(fmt.Sprintf("pid:%v", oldIDs), tblInfo.Name.O)
+		}
+
+		if err = clearTruncatePartitionTiflashStatus(tblInfo, newPartitions, oldIDs); err != nil {
+			job.State = model.JobStateCancelled
```
Defined2014, line 2019: If we canceled `ActionTruncateTablePartition` job, we should consider how to rollback it.

L-maple:
> > @mjonss @Defined2014 Hi, this PR can be reviewed~ I have a little question: What else should I do before global index can be a GA feature?
> 
> IMO, we should accomplish the following
> 
> 1. Provide a complete test report (including all related parts, like all executors, dynamic pruning/static pruning, etc.)
> 2. Close all bugs above or equals to major severity
> 3. Do some paper works (internal)
> 
> Do you have anything to add? @mjonss @bb7133

Ok, I will provide a test report and do the paper work. Could I send the report+paper to you? @Defined2014 
Besides, 
- global index will only work only in dynamic prunning.
- I checked all related parts except `indexMerge` and `reorganizePartition`.

ddl/partition.go
L-maple: OK, I will check the `IsRollbackable` function and complete the logic.

Defined2014:
> > > @mjonss @Defined2014 Hi, this PR can be reviewed~ I have a little question: What else should I do before global index can be a GA feature?
> > 
> > 
> > IMO, we should accomplish the following
> > 
> > 1. Provide a complete test report (including all related parts, like all executors, dynamic pruning/static pruning, etc.)
> > 2. Close all bugs above or equals to major severity
> > 3. Do some paper works (internal)
> > 
> > Do you have anything to add? @mjonss @bb7133
> 
> Ok, I will provide a test report and do the paper work. Could I send the report+paper to you? @Defined2014 Besides,
> 
> * global index will only work only in dynamic prunning.
> * I checked all related parts except `indexMerge` and `reorganizePartition`.
> * related issues have been handled except `indexMerge`, which is in developing.

Sounds good to me. I will ask PM and release team next week to find out what paper work is needed.

ddl/partition.go
```diff
@@ -1911,27 +1911,152 @@ func onTruncateTablePartition(d *ddlCtx, t *meta.Meta, job *model.Job) (int64, e
...
-				def.ID = pid
-				// Shallow copy only use the def.ID in event handle.
-				newPartitions = append(newPartitions, *def)
-				break
+			}
+			if !found {
+				newDefs = append(newDefs, pi.Definitions[i])
 			}
 		}
+		pi.Definitions = newDefs
```
mjonss, line 1930: Would not this mean that the partitions that are being truncated can neither be read or written, i.e. unaccessible and possibly give errors on insert/updates?

table/tables/partition.go
```diff
@@ -117,6 +121,15 @@ func newPartitionedTable(tbl *TableCommon, tblInfo *model.TableInfo) (table.Part
 		return nil, errors.Trace(err)
 	}
 	ret.partitionExpr = partitionExpr
+
+	if pi.OriginalDefinitions != nil {
```
mjonss, line 125:
```suggestion
	if len(pi.OriginalDefinitions) > 0 {
```

parser/model/model.go
```diff
@@ -1160,6 +1160,9 @@ type PartitionInfo struct {
 	AddingDefinitions []PartitionDefinition `json:"adding_definitions"`
 	// DroppingDefinitions is filled when dropping a partition that is in the mid state.
 	DroppingDefinitions []PartitionDefinition `json:"dropping_definitions"`
+	// OriginalDefinitions is filled when truncating a partition that is in the mid state.
+	// original definitions used to ensure no record inserted into truncating partitions in reorg state.
+	OriginalDefinitions []PartitionDefinition `json:"original_definitions"`
```
mjonss, line 1165: Why is this needed? I.e. are the full PartitionDefinition's needed or just ids?

ddl/partition.go
```diff
@@ -1911,27 +1911,152 @@ func onTruncateTablePartition(d *ddlCtx, t *meta.Meta, job *model.Job) (int64, e
...
-	newPartitions := make([]model.PartitionDefinition, 0, len(oldIDs))
-	for _, oldID := range oldIDs {
-		for i := 0; i < len(pi.Definitions); i++ {
-			def := &pi.Definitions[i]
-			if def.ID == oldID {
-				pid, err1 := t.GenGlobalID()
-				if err1 != nil {
-					return ver, errors.Trace(err1)
+	switch job.SchemaState {
+	case model.StatePublic:
```
mjonss, line 1915: Should we simplify this if there are no global index and only have a single state change where the partitions will have the new ids? Or do we still need to go through the different stages?

mjonss:
> > Hi, can you give a high level description of how this will work?
> > I'm not sure this or the current handling of global index reorganization works properly, but I think you are moving it closer to a working solution.
> > One thing I don't think works currently is in the following scenario: During the global index 'cleanup', what will happen to the sessions reading it? I would assume that if the metadata has the old partition ids, then the index will be inconsistent with existing rows/partitions (since they are being deleted during the partitions are readable) I would also assume that if the metadata has the new partition ids, then the index will still have records not yet deleted to partitions that no-longer exists.
> > Note that we allow currentSchemaVersion and currentSchemaVersion-1 as valid schema versions!
> 
> @mjonss Please review this PR again. I solved the problem by saving the original Definitions in a new data memeber `OriginalDeifnitions`.
> 
> * Read: `Definitions` in the metadata doesn't contain truncating definitions, so read is OK.
> * Write: We can add check to ensure no record can be inserted into truncating definitions. How should we do it? We can locate partition in both `Defintions` and `OriginalDefinitions` in the metadata. If the result pids are different, we can ensure the write should be forbided.
>   e.g.,
>   `create table test_global (a int) partition by range( a ) ( partition p1 values less than (10), partition p2 values less than (20) );`
>   When execute `alter table test_global truncate partition p1;`, we should forbid `insert into test_global values (1);` stmt be executed.
>   For `Definitions` in the metadata, (1) would be located in p2, while for `OriginalDefinitions` in the metadata, (1) would be located in p1. So this check enables no record can be inserted into truncating partitions.

Will reading still work with named partitions, like `SELECT * FROM t PARTITION (pTruncated)`?
I don't think it is acceptable to start getting errors, or blocking writes, in an application that are reading and writing to a table, due to a concurrent `truncate partition`.

I would expect that either a session/transaction sees the partitions as before (if the Schema version is before the effective `truncate partition`) or sees the partitions as after (if the Schema version is after the effective `truncate partition`).

The trick would be to do the reorganization / update / cleaning of the global index transparent to both the before and the after Schema versions. I.e. that the sessions sees the before version, would have full access to the older partitions but not see any of the new partition ids, and the sessions that sees the after version does not see any of the old truncated ids, only the new ones. So during the in-place reorganization of the global index, there would need to be some kind of filter in place.

table/tables/partition.go
```diff
@@ -91,8 +91,11 @@ func (t *partitionedTable) GetPartitionedTable() table.PartitionedTable {
 // partitionedTable is a table, it contains many Partitions.
 type partitionedTable struct {
 	TableCommon
-	partitionExpr   *PartitionExpr
-	partitions      map[int64]*partition
+	partitionExpr *PartitionExpr
+	partitions    map[int64]*partition
+	// originalPartitionExpr is only used by truncating table partition.
+	originalPartitionExpr *PartitionExpr
```
mjonss, line 97: Why is this needed, just because the number of partitions have changed?

mjonss (commented):
Some questions and comments.

Also how will this work with HASH/KEY partitioning only using number of partitions?

ddl/partition.go
L-maple: resolved

ddl/partition.go
L-maple: you are right, the previous logic does not allow concurrent read/writes for truncating partitions. I solve the problem in the new PR.

table/tables/partition.go
L-maple: resolved

parser/model/model.go
L-maple: here you can review the new PR.

ddl/partition.go
L-maple: `onDropTablePartition` also use Public->DeleteOnly->DeleteReorg 3 stages. Here I just stay the same. If you think it is necessary, I will improve the design.

table/tables/partition.go
L-maple: This data member is removed in the new PR.

L-maple:
> Some questions and comments.
> 
> Also how will this work with HASH/KEY partitioning only using number of partitions?

I think the new logic is OK with hash/key partitioning.

ddl/rollingback.go
```diff
@@ -401,6 +401,8 @@ func convertJob2RollbackJob(w *worker, d *ddlCtx, t *meta.Meta, job *model.Job)
 		err = rollingbackDropTableOrView(t, job)
 	case model.ActionDropTablePartition:
 		ver, err = rollingbackDropTablePartition(t, job)
+	case model.ActionTruncateTablePartition:
```
Defined2014, line 404: Could combine with `model.ActionDropForeignKey`

ddl/db_partition_test.go
```diff
@@ -2087,6 +2087,193 @@ func TestGlobalIndexUpdateInDropPartition(t *testing.T) {
...
+		if job.SchemaState == model.StateDeleteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+			err = tk1.ExecToErr("update test_global set a = 2 where a = 11")
+		}
+	}
+	dom.DDL().SetHook(hook)
+
+	tk.MustExec("alter table test_global truncate partition p1")
+	assert.NotNil(t, err)
```
Defined2014, line 2165: I think we could move it into hook which is easy to understand.

ddl/db_partition_test.go
```diff
@@ -2087,6 +2087,193 @@ func TestGlobalIndexUpdateInDropPartition(t *testing.T) {
...
+		if job.SchemaState == model.StateDeleteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+			err = tk1.ExecToErr("update test_global set a = 1 where a = 12")
+		}
+	}
+	dom.DDL().SetHook(hook)
+
+	tk.MustExec("alter table test_global truncate partition p1")
+	assert.NotNil(t, err)
```
Defined2014, line 2201: ditto

ddl/db_partition_test.go
```diff
@@ -2087,6 +2087,193 @@ func TestGlobalIndexUpdateInDropPartition(t *testing.T) {
...
+		if job.SchemaState == model.StateDeleteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+			err = tk1.ExecToErr("insert into test_global values(2, 2, 2)")
+		}
+	}
+	dom.DDL().SetHook(hook)
+
+	tk.MustExec("alter table test_global truncate partition p1")
+	assert.NotNil(t, err)
```
Defined2014, line 2274: ditto

ddl/db_partition_test.go
```diff
@@ -2087,6 +2087,193 @@ func TestGlobalIndexUpdateInDropPartition(t *testing.T) {
...
+			tk1.MustExec("use test")
+
+			indexScanResult = tk1.MustQuery("select b from test_global use index(idx_b)").Sort()
+		}
+	}
+	dom.DDL().SetHook(hook)
+
+	tk.MustExec("alter table test_global truncate partition p1")
+
+	indexScanResult.Check(testkit.Rows("11", "12"))
```
Defined2014, line 2238: ditto

ddl/db_partition_test.go
```diff
@@ -2087,6 +2087,190 @@ func TestGlobalIndexUpdateInDropPartition(t *testing.T) {
...
+
+	var err error
+	hook := &callback.TestDDLCallback{Do: dom}
+	hook.OnJobRunBeforeExported = func(job *model.Job) {
+		assert.Equal(t, model.ActionTruncateTablePartition, job.Type)
+		if job.SchemaState == model.StateDeleteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+			err = tk1.ExecToErr("update test_global set a = 2 where a = 11")
+			assert.NotNil(t, err)
```
mjonss, line 2160: Why do we want an error here? (also it will fail even if there are no global index!)
I don't think this is expected by users.

ddl/db_partition_test.go
```diff
@@ -2087,6 +2087,190 @@ func TestGlobalIndexUpdateInDropPartition(t *testing.T) {
...
+
+	var err error
+	hook := &callback.TestDDLCallback{Do: dom}
+	hook.OnJobRunBeforeExported = func(job *model.Job) {
+		assert.Equal(t, model.ActionTruncateTablePartition, job.Type)
+		if job.SchemaState == model.StateDeleteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+			err = tk1.ExecToErr("insert into test_global values(2, 2, 2)")
+			assert.NotNil(t, err)
```
mjonss, line 2266: Same here, I don't think it is expected by the users to get an error here.

ddl/db_partition_test.go
```diff
@@ -2087,6 +2087,190 @@ func TestGlobalIndexUpdateInDropPartition(t *testing.T) {
...
+
+	var err error
+	hook := &callback.TestDDLCallback{Do: dom}
+	hook.OnJobRunBeforeExported = func(job *model.Job) {
+		assert.Equal(t, model.ActionTruncateTablePartition, job.Type)
+		if job.SchemaState == model.StateDeleteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+			err = tk1.ExecToErr("update test_global set a = 1 where a = 12")
+			assert.NotNil(t, err)
```
mjonss, line 2196: Same here, why expecting an error, I would expect it to succeed, from a users point of view.

L-maple:
The truncating partition is in delete only state，so write is not allowed.



---- Replied Message ----
| From | Mattias ***@***.***> |
| Date | 05/17/2023 20:11 |
| To | ***@***.***> |
| Cc | Changjie ***@***.***>***@***.***> |
| Subject | Re: [pingcap/tidb] ddl: fix truncate partition for global index (PR #42433) |

@mjonss commented on this pull request.

In ddl/db_partition_test.go:

> +			err = tk1.ExecToErr("update test_global set a = 1 where a = 12")
+			assert.NotNil(t, err)


Same here, why expecting an error, I would expect it to succeed, from a users point of view.

mjonss:
> The truncating partition is in delete only state，so write is not allowed.

From the implementation view, it is correct. But from a user perspective the implementation just introduced a new error, where there were no error before (even for partitioned tables without global index).

One of the design goals with the distributed DDL by states, is to not break or block concurrent transactions, by allowing transactions to work with both current schema version and one version older schema.

L-maple:
> > The truncating partition is in delete only state，so write is not allowed.
> 
> From the implementation view, it is correct. But from a user perspective the implementation just introduced a new error, where there were no error before (even for partitioned tables without global index).
> 
> One of the design goals with the distributed DDL by states, is to not break or block concurrent transactions, by allowing transactions to work with both current schema version and one version older schema.

@mjonss  Oh, I start to understand what you means. Okkk, I will remove the write check.

L-maple:
@mjonss @Defined2014 You can review the PR again~

ddl/db_partition_test.go
```diff
@@ -2087,6 +2087,190 @@ func TestGlobalIndexUpdateInDropPartition(t *testing.T) {
...
+		if job.SchemaState == model.StateDeleteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+			err = tk1.ExecToErr("insert into test_global values(2, 2, 2)")
+			assert.Nil(t, err)
+		}
+	}
+	dom.DDL().SetHook(hook)
+
+	tk.MustExec("alter table test_global truncate partition p1")
```
Defined2014, line 2271: It will see `(2, 2, 2)` when truncate finished. This is not only inconsistent with the semantics of `delete only` state, and also inconsistent with comments `NewPartitionIDs is used to ensure no records can be inserted into new partitions when reorg.`

ddl/partition.go
```diff
@@ -1940,27 +1953,139 @@ func onTruncateTablePartition(d *ddlCtx, t *meta.Meta, job *model.Job) (int64, e
...
-		return ver, table.ErrUnknownPartition.GenWithStackByArgs(fmt.Sprintf("pid:%v", oldIDs), tblInfo.Name.O)
+		pi.OriginalDefinitions = pi.Definitions
+		pi.Definitions = newDefinitions
+
+		job.SchemaState = model.StateDeleteOnly
+		ver, err = updateVersionAndTableInfo(d, t, job, tblInfo, true)
+	case model.StateDeleteOnly:
+		// This state is not a real 'DeleteOnly' state, because tidb does not maintaining the state check in partitionDefinition.
+		// Insert this state to confirm all servers can not see the old partitions when reorg is running,
+		// so that no new data will be inserted into old partitions when reorganizing.
```
Defined2014, line 1982: Could add a comment here to indicates that the state of `DeleteOnly` is 'fake', and the insert into new partition can still be successful in this state.

ddl/partition.go
```diff
@@ -1940,27 +1953,139 @@ func onTruncateTablePartition(d *ddlCtx, t *meta.Meta, job *model.Job) (int64, e
...
 		for i := 0; i < len(pi.Definitions); i++ {
-			def := &pi.Definitions[i]
-			if def.ID == oldID {
-				pid, err1 := t.GenGlobalID()
-				if err1 != nil {
-					return ver, errors.Trace(err1)
+			newDef := pi.Definitions[i].Clone()
+			for _, oldID := range oldIDs {
+				if newDef.ID == oldID {
+					pid, err1 := t.GenGlobalID()
```
mjonss, line 1964: Maybe allocate the new global ids already in the alter table client session, like [this](https://github.com/pingcap/tidb/blob/a085eff3da5c5c195985a029113d3e0c872a7192/ddl/ddl_api.go#L6269-L6276)? (Recently fixed https://github.com/pingcap/tidb/issues/42940)

ddl/db_partition_test.go
L-maple: NewPartitionIDs is removed now.

ddl/partition.go
L-maple: OK, I add it.

ddl/partition_test.go
```diff
@@ -124,11 +124,12 @@ func testDropPartition(t *testing.T, ctx sessionctx.Context, d ddl.DDL, dbInfo *
...
-		TableID:    tblInfo.ID,
-		Type:       model.ActionTruncateTablePartition,
-		BinlogInfo: &model.HistoryInfo{},
-		Args:       []interface{}{pids},
+		SchemaID:    dbInfo.ID,
+		TableID:     tblInfo.ID,
+		Type:        model.ActionTruncateTablePartition,
+		SchemaState: model.StatePublic,
+		BinlogInfo:  &model.HistoryInfo{},
+		Args:        []interface{}{pids},
```
Defined2014, line 132: miss genIDs here

Defined2014 (approved):
LGTM

If you are still interested in GA global index, you can send an email to `mohangjie1995@gmail`, and I can help you with some internal works.

L-maple:
> 

OK, I will contact him/her next week~

L-maple:
> LGTM
> 
> If you are still interested in GA global index, you can send an email to `mohangjie1995@gmail`, and I can help you with some internal works.


OK, I will contact him/her next week~

ddl/db_partition_test.go
```diff
@@ -2087,6 +2087,190 @@ func TestGlobalIndexUpdateInDropPartition(t *testing.T) {
...
+	tt := external.GetTableByName(t, tk, "test", "test_global")
+	pid := tt.Meta().Partition.Definitions[1].ID
+
+	tk.MustExec("Alter Table test_global Add Unique Index idx_b (b);")
+	tk.MustExec("Alter Table test_global Add Unique Index idx_c (c);")
+	tk.MustExec(`INSERT INTO test_global VALUES (1, 1, 1), (2, 2, 2), (11, 3, 3), (12, 4, 4)`)
+
+	tk.MustExec("alter table test_global truncate partition p2;")
+	result := tk.MustQuery("select * from test_global;")
+	result.Sort().Check(testkit.Rows(`1 1 1`, `2 2 2`))
```
mjonss, line 2114: If one adds to this test, then in a new transaction in 'delete only' state, one cannot find the truncated "15 15 15" but it gives a duplicate error on insert.
```suggestion
       tk.MustExec(`INSERT INTO test_global VALUES (1, 1, 1), (2, 2, 2), (11, 3, 3), (12, 4, 4), (15, 15, 15)`)
 
       tk2 := testkit.NewTestKit(t, store)
       tk2.MustExec(`use test`)
       tk2.MustExec(`begin`)
       tk2.MustExec(`insert into test_global values (5,5,5)`)
       syncChan := make(chan bool)
       go func() {
               tk.MustExec("alter table test_global truncate partition p2;")
               syncChan <- true
       }()
       waitFor := func(i int, s string) {
               for {
                       tk4 := testkit.NewTestKit(t, store)
                       tk4.MustExec(`use test`)
                       res := tk4.MustQuery(`admin show ddl jobs where db_name = 'test' and table_name = 'test_global' and job_type = 'truncate partition'`).Rows()
                       if len(res) == 1 && res[0][i] == s {
                               break
                       }
                       time.Sleep(10 * time.Millisecond)
               }
       }
       waitFor(4, "delete only")
       tk3 := testkit.NewTestKit(t, store)
       tk3.MustExec(`begin`)
       tk3.MustExec(`use test`)
       tk3.MustQuery(`select b from test_global use index(idx_b) where b = 15`).Check(testkit.Rows())
       tk3.MustQuery(`select c from test_global use index(idx_c) where c = 15`).Check(testkit.Rows())
       // Here it will fail with
       // [kv:1062]Duplicate entry '15' for key 'test_global.idx_b'
       tk3.MustExec(`insert into test_global values (15,15,15)`)
       tk2.MustExec(`commit`)
       tk3.MustExec(`commit`)
       syncChan <- true
       result := tk.MustQuery("select * from test_global;")
       result.Sort().Check(testkit.Rows(`1 1 1`, `2 2 2`, `15 15 15`))
```

It can be quite confusing for a user if the row cannot be found in a transaction, while it still is not possible to insert.

parser/model/model.go
```diff
@@ -1163,8 +1163,11 @@ type PartitionInfo struct {
 	AddingDefinitions []PartitionDefinition `json:"adding_definitions"`
 	// DroppingDefinitions is filled when dropping a partition that is in the mid state.
 	DroppingDefinitions []PartitionDefinition `json:"dropping_definitions"`
-	States              []PartitionState      `json:"states"`
-	Num                 uint64                `json:"num"`
+	// OriginalDefinitions is filled when truncating partitions that is in the mid state.
+	OriginalDefinitions []PartitionDefinition `json:"new_definitions"`
```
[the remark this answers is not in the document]
L-maple: I considered it, but maybe not graceful. 
Not only the dropping definitions is used in other circumstances,  but I have to store truncated definitions' positions in the old definitions, so that I can joint the original definitions when deleting global index.

ddl/db_partition_test.go
L-maple: You are right. Because the global index row still exists when the state is in deleteOnly.

Maybe I should add write check for the truncating definitions, otherwise it will cause data inconsistency(imagine the global index records we just add will be deleted later). 

@mjonss What's your advice?

L-maple: @mjonss  I add write check when truncate partition ddl state is not in public.

L-maple:
@mjonss Do you have any advice further? Maybe this PR can be merged.

L-maple:
@Defined2014 Sadly, I am busy recently and have few time for global index paper work. Maybe you can do it further. Anyway,  when I have free time, I will contribute more code for TiDB.

Defined2014:
> Sadly, I am busy recently and have few time for global index paper work. Maybe you can do it further. Anyway, when I have free time, I will contribute more code for TiDB.

No problem, I hope you can contribute more to TiDB when you have time in the future.

parser/model/model.go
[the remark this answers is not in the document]
mjonss: Why would you need the positions in the original definitions, is not Physical Table IDs the only thing that is needed from each truncated partition?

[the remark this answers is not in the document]
L-maple: @mjonss OK, I understand what you means. I reuse the DroppingDefinitions and remove the OriginalDefinitions.

mjonss (approved):
Please add the check for global index in the StatePublic in onTruncateTablePartition and keep the single state change if no global index, so that this PR does not affect partitioned tables without global index. After adding this, I will merge this (once there is a second approval)

And sorry for the delay of my reviews, my major concern has not been the code change but that Truncate Table/Partition is not compatible with the distributed DDL framework (where every state change is compatible with the previous state) since Truncate will remove rows. So I think we simply need to document this (I will create a separate Docs pull request for this).

L-maple:
> Please add the check for global index in the StatePublic in onTruncateTablePartition and keep the single state change if no global index, so that this PR does not affect partitioned tables without global index. After adding this, I will merge this (once there is a second approval)
> 
> And sorry for the delay of my reviews, my major concern has not been the code change but that Truncate Table/Partition is not compatible with the distributed DDL framework (where every state change is compatible with the previous state) since Truncate will remove rows. So I think we simply need to document this (I will create a separate Docs pull request for this).

@mjonss resolved.

Defined2014 (approved):

PR #42465, opened 2023-03-22
util/sqlexec: EscapeSQL support resolve some underlying type

lance6716:

Issue Number: close #xxx

Problem Summary:

Support use customized type whose underlying type is intXXX, uintXXX, floatXX, bool, string in `ExecuteSQL`. Typically this can make ease of using customized enum type when `ExecuteSQL`.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

benchmark

at 2137ff3974d9bcf3a90683f72f682d7423cdb3c0

```
goos: darwin
goarch: arm64
pkg: github.com/pingcap/tidb/util/sqlexec
BenchmarkEscapeString
BenchmarkEscapeString-8       	 4011201	       292.8 ns/op
BenchmarkUnderlyingString
BenchmarkUnderlyingString-8   	 3403500	       351.3 ns/op
BenchmarkEscapeInt
BenchmarkEscapeInt-8          	 6085053	       196.5 ns/op
BenchmarkUnderlyingInt
BenchmarkUnderlyingInt-8      	 4736172	       253.8 ns/op
```

at master

```
goos: darwin
goarch: arm64
pkg: github.com/pingcap/tidb/util/sqlexec
BenchmarkEscapeString
BenchmarkEscapeString-8   	 4089615	       294.3 ns/op
BenchmarkEscapeInt
BenchmarkEscapeInt-8   	 	 6049938	       198.4 ns/op
```

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/util/sqlexec/utils.go
+++ b/util/sqlexec/utils.go
@@ -93,10 +93,6 @@ func escapeStringBackslash(buf []byte, v string) []byte {
 	return escapeBytesBackslash(buf, hack.Slice(v))
 }
 
-var (
-	reflectTpTime = reflect.TypeOf(time.Time{})
-)
-
 // escapeSQL is the internal impl of EscapeSQL and FormatSQL.
 func escapeSQL(sql string, args ...interface{}) ([]byte, error) {
 	buf := make([]byte, 0, len(sql))
--- a/util/sqlexec/utils.go
+++ b/util/sqlexec/utils.go
@@ -94,122 +94,22 @@ func escapeStringBackslash(buf []byte, v string) []byte {
 }
 
 var (
-	reflectTpInt      = reflect.TypeOf(int(0))
-	reflectTpInt8     = reflect.TypeOf(int8(0))
-	reflectTpInt16    = reflect.TypeOf(int16(0))
-	reflectTpInt32    = reflect.TypeOf(int32(0))
-	reflectTpInt64    = reflect.TypeOf(int64(0))
-	reflectTpUint     = reflect.TypeOf(uint(0))
-	reflectTpUint8    = reflect.TypeOf(uint8(0))
-	reflectTpUint16   = reflect.TypeOf(uint16(0))
-	reflectTpUint32   = reflect.TypeOf(uint32(0))
-	reflectTpUint64   = reflect.TypeOf(uint64(0))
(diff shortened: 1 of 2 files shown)


util/sqlexec/utils.go
```diff
@@ -135,86 +250,20 @@ func escapeSQL(sql string, args ...interface{}) ([]byte, error) {
...
-				case string:
-					buf = append(buf, '\'')
-					buf = escapeStringBackslash(buf, v)
-					buf = append(buf, '\'')
-				case []string:
-					for i, k := range v {
-						if i > 0 {
-							buf = append(buf, ',')
+				if !handleDynTpArg(arg) {
+					reflectTp := reflect.TypeOf(arg)
```
xhebox, line 254: Try `switch reflect.TypeOf(arg).Kind()`. You don't need to `Convert()`.

lance6716: I'm not sure about your suggestion, how could the type be changed without `Convert`? And for `Struct` kind I think we must use other function to tell if they are same

https://github.com/golang/go/blob/f57f02fcd519a86683c47a444e0e24a086fea8e0/src/reflect/type.go#L1681

And I think I can improve current code for basic types by `switch kind` to avoid looping one-by-one to check ConvertibleTo

xhebox: > I know they have same _kind_, but what do you suggest to change the _type_ without `Convert`? https://go.dev/play/p/jw9FCSekc2s

You don't need to change the reflection type. Check https://go.dev/play/p/oh00SrayMmq.

lance6716: updated, I will try to handle customized types whose underlying type is slice later. And I noticed now the performance is a bit lower, will run benchmark later as well.

xhebox: Because you always call `reflect.TypeOf`. We could do  same as before, first we match by `v.(type)`, then we try `reflect.TypeOf`.

lance6716: fixed, and updated the benchmark in PR description. PTAL

xhebox (approved):

GMHDBJD (approved):
LGTM

PR #42472, opened 2023-03-22
ddl: decouple the index writer from range task

tangenta:

Problem Summary:

Since this is the refactoring, let's skip the issue check.

image

This PR removes the red part in the above image.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

I made a simple performance test on 10M sysbench dataset.

Before this PR:
```
mysql> alter table sbtest1 add index idx(c);
Query OK, 0 rows affected (3 min 35.43 sec)

mysql> alter table sbtest1 add index idx(c);
Query OK, 0 rows affected (3 min 12.41 sec)

mysql> alter table sbtest1 add index idx(c);
Query OK, 0 rows affected (3 min 42.54 sec)

mysql> alter table sbtest1 add index idx(c);
Query OK, 0 rows affected (3 min 35.92 sec)
```

After this PR:
```
mysql> alter table sbtest1 add index idx(c);
Query OK, 0 rows affected (3 min 17.92 sec)

mysql> alter table sbtest1 add index idx(c);
Query OK, 0 rows affected (3 min 37.46 sec)

mysql> alter table sbtest1 add index idx(c);
Query OK, 0 rows affected (3 min 45.83 sec)

mysql> alter table sbtest1 add index idx(c);
Query OK, 0 rows affected (3 min 16.46 sec)
```

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


ddl/backfilling.go
```diff
@@ -520,107 +520,144 @@ func splitTableRanges(t table.PhysicalTable, store kv.Storage, startKey, endKey
...
-			}
-		}
-	}
-	return keeper.nextKey, addedCount, errors.Trace(firstErr)
+type resultConsumer struct {
+	dc      *ddlCtx
+	wg      *sync.WaitGroup
+	taskCnt *atomic.Int64
+	doneCh  chan struct{}
+	err     *atomic.Value
```
hawkingrei, line 528: atomic.Value is replaced by atomic.Pointer to avoid type convert.

ddl/backfilling.go
```diff
@@ -520,107 +520,144 @@ func splitTableRanges(t table.PhysicalTable, store kv.Storage, startKey, endKey
...
-	logutil.BgLogger().Info("[ddl] backfill workers successfully processed batch",
-		zap.Stringer("element", reorgInfo.currElement),
-		zap.Int64("total added count", *totalAddedCount),
-		zap.String("start key", hex.EncodeToString(startKey)),
-		zap.String("next key", hex.EncodeToString(nextKey)),
-		zap.Int64("batch added count", taskAddedCount),
-		zap.String("take time", elapsedTime.String()),
-		zap.NamedError("updateHandleError", err1))
-	return nil
+func drainTasks(taskCh chan *reorgBackfillTask) int {
```
hawkingrei, line 654: I suggest that the private function or object can be moved into ```ddl/internal``` and make public. because it is in the ```internal```. It is only public for ```ddl``` package and all subpackage of ```ddl```. It is an effective way to avoid too large ```ddl``` package

ddl/index.go
```diff
@@ -1756,31 +1760,38 @@ func (*addIndexIngestWorker) String() string {
 }
 
 // BackfillData will ingest index records through lightning engine.
-func (w *addIndexIngestWorker) BackfillData(handleRange reorgBackfillTask) (backfillTaskContext, error) {
-	var taskCtx backfillTaskContext
+func (w *addIndexIngestWorker) BackfillData(handleRange reorgBackfillTask) (taskCtx backfillTaskContext, err error) {
+	taskCtx.nextKey = handleRange.startKey
 	oprStartTime := time.Now()
 	defer func() {
 		logSlowOperations(time.Since(oprStartTime), "writeIndexKVsToLocal", 3000)
```
wjhuang2016, line 1767: It may always `log slow` since it handles many chunks now

ddl/backfilling.go
tangenta: Removed.

ddl/backfilling.go
tangenta: Removed.

wjhuang2016 (approved):

ddl/backfilling.go
```diff
@@ -520,106 +522,111 @@ func splitTableRanges(t table.PhysicalTable, store kv.Storage, startKey, endKey
...
+			zap.Error(result.err))
+		// Drain tasks to make it quit early.
+		for len(scheduler.taskCh) > 0 {
+			<-scheduler.taskCh
+		}
+		return result.err
+	}
+	*totalAddedCount += int64(result.addedCount)
+	keeper.updateNextKey(result.taskID, result.nextKey)
+	if taskSeq%scheduler.workerSize()*4 == 0 {
```
Benjamin2037, line 601: what is "taskSeq%scheduler.workerSize()*4" means? also here the 4 meaning?

tangenta: For `txn` and `txn-merge` mode, it gives a chance to adjust the worker size according to `tidb_ddl_reorg_worker_cnt` for every 4 tasks processed.

For the `ingest` mode, it doesn't affect anything, because when the second task returns, all tasks have been completed.

`4` is a magic number. Any suggestion?

ddl/index_cop.go
```diff
@@ -213,26 +187,21 @@ func (c *copReqSenderPool) adjustSize(n int) {
...
-	logutil.BgLogger().Info("[ddl-ingest] close cop-request sender pool", zap.Int("results not handled", len(c.results.Keys())))
+func (c *copReqSenderPool) close(force bool) {
+	if c.closed {
+		return
+	}
+	logutil.BgLogger().Info("[ddl-ingest] close cop-request sender pool")
 	close(c.tasksCh)
-	for _, w := range c.senders {
-		w.cancel()
+	if force {
```
Benjamin2037, line 196: I saw int backfilling.go:870 normal process will close(false),  and in backfilling.go:811 -> defer scheduler.close(true). if this will cause there is no chance to close closeBackfillWorkers(b.workers)?

tangenta: Yes. Maybe `closeBackfillWorkers` should be renamed into `cancelBackfillWorkers`?

All it does is notify workers to shutdown.

ddl/backfilling.go
Benjamin2037: should it be like this taskSeq%(scheduler.workerSize()*4), it is 4 times worker cnt?

tangenta: Updated.

Benjamin2037 (approved):
LGTM

hawkingrei (approved):

Issue #38116, opened 2022-09-23
PreparedStatement parameter related error causing TiDB server down when using cursor fetch


Qiuchi0918:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

The bug can be reproduced using JDBC

install dependency

```XML
<dependency>
  <groupId>mysql</groupId>
  <artifactId>mysql-connector-java</artifactId>
  <version>8.0.15</version>
</dependency>
```

create test table

```SQL
create table test_tab
(
    idx_col bigint auto_increment
        primary key
);
```

execute following codes

```JAVA
public static void reproduceParameterError() throws SQLException {

    String url = "jdbc:mysql://${ip-address}:${port}/${database}?useCursorFetch=true";
    String user = "${user}";
    String password = "${password}";

    Connection conn = DriverManager.getConnection(url, user, password);

    conn.setAutoCommit(false); // must set

    PreparedStatement statement1 = conn.prepareStatement("" +
            "select * " +
            "from " +
            "test_tab " +
            "where (idx_col = ?)"); // must be wrapped in parentheses

    statement1.setFetchSize(500); // must set

    statement1.setBigDecimal(1, BigDecimal.valueOf(123456)); // must set BigDecimal instead of long or integer

    statement1.executeQuery();
    conn.commit();
    conn.close();
}
```

check tidb server status

### 2. What did you expect to see? (Required)
nothing unexpected

### 3. What did you see instead (Required)
tidb server down with following log in tidb_stderr.log

```
panic: runtime error: index out of range [0] with length 0

goroutine 3429 [running]:
github.com/pingcap/tidb/expression.(*ParamMarker).GetUserVar(...)
	/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/expression/constant.go:91
github.com/pingcap/tidb/expression.(*Constant).getLazyDatum(0x6?, {0x0?, 0x0?})
	/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/expression/constant.go:187 +0x26a
github.com/pingcap/tidb/expression.(*Constant).Eval(0xc000b3acc0, {0x0?, 0xc009c3d0e0?})
	/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/expression/constant.go:197 +0x98
github.com/pingcap/tidb/expression.(*Constant).ExplainInfo(0x427b380?)
[...]

PR #39950, opened 2022-12-14
*: prevent cursor read from being cancelled by GC

zyguan:
Signed-off-by: zyguan <zhongyangguan@gmail.com>

### What problem does this PR solve?

Issue Number: close #39447

Problem Summary: 

Cursor read might be cancelled by GC since min-start-ts is calculated by `min([s.ProcessInfo.CurTxnStartTS for s in all sessions])` however the `CurTxnStartTS` will be reset once fetch returned.

### What is changed and how it works?

1. Let each session hold a list of pretected timestamps.
2. Let session manager calculate the min-start-ts (considering protected timestamps).
3. When start to do cursor read, hold the start-ts until the result set gets close.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #42473, opened 2023-03-22
session, com_stmt: fetch all rows during EXECUTE command

### LINKED ISSUE #41891: If jdbc fetch and udv are used, the udv value is changed after fetch, and subsequent fetch results are incorrect.
## Bug Report

## java code 
String sql_select = "select id from user where id < @a ";
stmt = conn.createStatement();
stmt.execute("set @a = 30");
pstmt_select.setFetchSize(2);
ResultSet res = pstmt_select.executeQuery();
//stmt.execute("set @a = 70");
while(res.next()) {
  int id = res.getInt("id") ;
  System.out.println("线程ID"+num +" id:" + id);
  stmt.execute("set @a = 70");
}
res.close();

The console prints values less than 30 .

The console prints values less than 70 .

v6.5.0

| Release Version: v6.5.0
Edition: Community
Git Commit Hash: 706c3fa3c526cdba5b3e9f066b1a568fb96c56e3
Git Branch: heads/refs/tags/v6.5.0
UTC Build Time: 2022-12-27 03:50:44
GoVersion: go1.19.3
Race Enabled: false
TiKV Min Version: 6.2.0-alpha
Store: tikv |

### LINKED ISSUE #42424: params may have changed during the executing of prepared statements with cursor fetch
The params may have changed while the executors' worker running background, so it's dangerous to allow statements running concurrently.

YangKeao:

Issue Number: close #42424, close #41891

1. Fetch all rows and store them in the `fetchedRows` in the `ResultSet`.
2. Close the  result set in `EXECUTE` command, so we don't need to close it again in the following commands (`FETCH`, `RESET` and `CLOSE`)

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Test with the following a little bigger table:

```java
import java.math.BigDecimal;
import java.sql.*;

public class Example {
    public static void main(String[] args) throws SQLException, InterruptedException, Exception {
        Connection conn = DriverManager.getConnection("jdbc:mysql://127.0.0.1:4000/test?useCursorFetch=true&useServerPrepStmts=true&useSSL=false", "root", "");

        conn.setAutoCommit(false); // must set

        conn.prepareStatement("drop table if exists t").execute();
        conn.prepareStatement("create table t (id int primary key, id_2 int)").execute();

        for (int i = 0; i < 10000; i++) {
            PreparedStatement stmt = conn.prepareStatement("insert into t values (?, ?)");
            stmt.setInt(1, i);
            stmt.setInt(2, i);
            stmt.execute();
        }

        // submit a statement with more arguments:
        PreparedStatement statement1 = conn.prepareStatement("" +
                "select * " +
                "from " +
                "t " +
                "where id > ? and id_2 > ?");
        statement1.setFetchSize(500);
        statement1.setBigDecimal(1, new BigDecimal(2));
        statement1.setBigDecimal(2, new BigDecimal(2));
        ResultSet rs_1 = statement1.executeQuery();

        // submit a statement with less arguments:
        PreparedStatement statement2 = conn.prepareStatement("" +
                "select * " +
                "from " +
                "t " +
                "where (id > ?)");
        statement2.setFetchSize(500);
        statement2.setBigDecimal(1, new BigDecimal(1));
        ResultSet rs_2 = statement2.executeQuery();


        // fetch the result from the first argument
        int iter = 3;
        while (rs_1.next()) {
            int result = rs_1.getInt(1);

            if (result != iter++ ) {
                throw new Exception("");
            }
        }
        System.out.println(iter);
        iter = 2;
        while (rs_2.next()) {
            int result = rs_2.getInt(1);

            if (result != iter++ ) {
                throw new Exception("");
            }
        }
        System.out.println(iter);

        conn.commit();
        statement1.close();
        statement2.close();
        conn.close();
    }
}

```

Side effects

It stores all rows of a cursor fetch statement locally, while execuing the `EXECUTE` command, so it will eat up more memory (related with the size of query).

### Release note

```release-note
Previously, TiDB loads data in `FETCH` command with the `cursorFetch` enabled. Now, TiDB will load all data into the memory in `EXECUTE` command. It'll use more memory according to the size of result. And also, TiDB will never report write conflict in `FETCH` command with read-committed isolation level.
```


zyguan:
cc @jackysp , the other issue of cursor read, we need to tell users that it may cost more memory.

server/conn_stmt.go
```diff
@@ -297,24 +298,44 @@ func (cc *clientConn) executePreparedStmtAndWriteResult(ctx context.Context, stm
...
-		// fix https://github.com/pingcap/tidb/issues/39447. we need to hold the start-ts here because the process info
-		// will be set to sleep after fetch returned.
-		if pi := cc.ctx.ShowProcess(); pi != nil && pi.ProtectedTSList != nil && pi.CurTxnStartTS > 0 {
-			unhold := pi.HoldTS(pi.CurTxnStartTS)
-			rs = &rsWithHooks{ResultSet: rs, onClosed: unhold}
+		// fetch all results of the resultSet, and stored them locally, so that the future `FETCH` command can read
+		// the rows directly to avoid running executor and accessing shared params/variables in the session
+		// NOTE: chunk should not be allocated from the connection allocator, which will reset after executing this command
+		// but the rows are still needed in the following FETCH command.
+		chk := rs.NewChunk(nil)
```
qw4990, line 305: Please leave a TODO comment here to remind us to trace memory usage here in the future...

YangKeao: Good.

qw4990 (approved):

server/conn_stmt_test.go
```diff
@@ -399,93 +312,6 @@ func TestCursorExistsFlag(t *testing.T) {
 	require.False(t, mysql.HasCursorExistsFlag(getLastStatus()))
 }
 
-func TestCursorReadWithRCCheckTS(t *testing.T) {
```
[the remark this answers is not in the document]
cfzjywxk: I remember there are quite some issues with the cursor fetch mode, for example, the `statement context` would be reset processing the next `fetch` command and the information in it could be lost processing.
For this test case I think it's ok to fetch all the results at the same time, the thing I'm worried about is that if there would be a memory problem if all the results are stored on the tidb-server side, usually when the cursor mode is used the result set would be large, it may increase the OOM risk.

cfzjywxk:
@bob34007 
If all the query results would be fetched processing the `execute` statement at the same time and stored in the tidb-server side, it seems the `tidb_rc_read_check_ts` optimization could be used together with the cursor fetch mode.

YangKeao:
> If all the query results would be fetched processing the execute statement at the same time and stored in the tidb-server side, it seems the tidb_rc_read_check_ts optimization could be used together with the cursor fetch mode.

I'll try to allow this optimization for cursor fetch, and add a test for it.

server/conn_stmt_test.go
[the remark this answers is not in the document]
YangKeao: > I remember there are quite some issues with the cursor fetch mode, for example, the statement context would be reset processing the next fetch command and the information in it could be lost.

Yes. This PR is trying to solve this problem :thinking:. These context are not needed in the `FETCH` command now.

> I'm worried about is that there would be a memory problem if all the results are stored on the tidb-server side, usually when the cursor mode is used the result set would be large, it may increase the OOM risk.

Can't agree more :crying_cat_face:

ddl/db_test.go
```diff
@@ -1353,21 +1355,13 @@ func TestReportingMinStartTimestamp(t *testing.T) {
...
-		{CurTxnStartTS: math.MaxUint64, ProtectedTSList: &se.GetSessionVars().ProtectedTSList},
-		{CurTxnStartTS: lowerLimit, ProtectedTSList: &se.GetSessionVars().ProtectedTSList},
-		{CurTxnStartTS: validTS, ProtectedTSList: &se.GetSessionVars().ProtectedTSList},
+		{CurTxnStartTS: 0},
+		{CurTxnStartTS: math.MaxUint64},
+		{CurTxnStartTS: lowerLimit},
+		{CurTxnStartTS: validTS},
 	}
-	infoSyncer.ReportMinStartTS(dom.Store())
-	require.Equal(t, validTS, infoSyncer.GetMinStartTS())
```
tiancaiamao, line 1364: What's the cause of those failures?

server/server.go
```diff
@@ -1012,37 +1012,3 @@ func (s *Server) KillNonFlashbackClusterConn() {
 		s.Kill(id, false)
 	}
 }
-
-// GetMinStartTS implements SessionManager interface.
```
tiancaiamao, line 1016: Why this change? seems unrelated

bob34007:
> @bob34007 If all the query results would be fetched processing the `execute` statement at the same time and stored in the tidb-server side, it seems the `tidb_rc_read_check_ts` optimization could be used together with the cursor fetch mode.

If the result set is too large, oom may be triggered. In the scenario of using a cursor, the business cannot determine the size of the result set.

tiancaiamao (approved):
Memory usage is a concern.
But this change only affect the cursor read, and the old cursor read is buggy.
Correctness vs OOM risk, I think it's acceptable, let's make it correct first.

server/server.go
tiancaiamao: Revert of https://github.com/pingcap/tidb/pull/39950

zyguan (approved):

YangKeao:
Every time, different test failed :cry: . None of them can be reproduced on my local machine.

YangKeao:
rebase the master. @hawkingrei seems to have made some optimization on test speed.

tiancaiamao:
close #38116

PR #42538, opened 2023-03-23
parser: allocate a new array for parser result every time

### LINKED ISSUE #42230: Parser should always allocate a new array for result
## Enhancement
In parser.Parser.ParseSQL, the result slice is truncated and the underlying array is reused.
https://github.com/pingcap/tidb/blob/master/parser/yy_parser.go#L158
`parser.result = parser.result[:0]`

This makes it slightly more difficult to reuse the parser object in some cases. For example
`
q1 := parser.ParseSQL(query1)
q2 := parser.ParseSQL(query2)
// do something about q1 and q2
`
The code would not work because the second call is going to overwrite the result of the first call.

We are supposed to reuse the parser object since it is not lightweight but parse object is not goroutine safe. So one use case is to use the parser object synchronously but process parsing results in different goroutines. However, due to reusing of the result array, one need to make a copy of the result before processing it in goroutine. It would be better if the parse can allocate a new array for result every time.

yujiarista:
close pingcap/tidb#42230

Issue Number: close #42230 

Problem Summary: 
User has to make a copy of parser result before using it in goroutine to avoid data race which is inconvenient and can be missed if not familiar with how the parser reuses the underlying array for results.

Allocate a new array for parser result before parsing, instead of just truncate the slice while reusing the underlying array.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/parser/yy_parser.go
+++ b/parser/yy_parser.go
@@ -155,7 +155,7 @@ func (parser *Parser) ParseSQL(sql string, params ...ParseParam) (stmt []ast.Stm
 		}
 	}
 	parser.src = sql
-	parser.result = parser.result[:0]
+	parser.result = []ast.StmtNode{}
 
 	var l yyLexer = &parser.lexer
 	yyParse(l, parser)


parser/yy_parser.go
```diff
@@ -155,7 +155,7 @@ func (parser *Parser) ParseSQL(sql string, params ...ParseParam) (stmt []ast.Stm
 		}
 	}
 	parser.src = sql
-	parser.result = parser.result[:0]
+	parser.result = []ast.StmtNode{}
```
jfrabaute, line 158: maybe just set to `nil`.
In case of an error returned with no result, this avoids one allocation.

lance6716:
Hi, can you add a unit test to explain why you need this change?
/cc @xhebox

parser/yy_parser.go
xhebox: This will affect behaviors of TiDB, however.

What about adding a flag in `SetParserConfig`, and use flags to control whether we will or not allocate the new slice?

jfrabaute: How is this affecting the behavior? (and which part? Setting to an empty slice, or setting to nil?)
From what we've seen, right now, there is a bug of the same parser is reused because one parsing can override the entry of another parsing (either sequential or in parallel), which seems to be a pretty severe bug.

xhebox: > How is this affecting the behavior? (and which part? Setting to an empty slice, or setting to nil?)

Because we expects to reuse before. It is hard for any reviewer to ensure that changes of default behavior will not involve bad changes, .e.g. performance regression in case of this PR.

> From what we've seen, right now, there is a bug if the same parser is reused because one parsing can override the entry of another parsing (either sequential or in parallel), which seems to be a pretty severe bug.

Since we may expect to reuse the array in TiDB. And we may clone the result array somewhere to avoid race. Or we will just use it for one-time.

It is our bad that this library tightly coupled with TiDB, it is indeed a problem for other users of `parser`.

I'd like to see an option controlling this behavoir. You could enable **safe mode** by yourselves. And we may eventually migrate to that, and enable **safe mode** by default, deprecating the current behavior.


Does that makes sense to you?

jfrabaute: Yes.

I might have been confused by the Notes in this doc: https://github.com/pingcap/tidb/blob/master/parser/docs/quickstart.md

`The instantiated parser object is not goroutine safe. It is better to keep it in a single goroutine.`

`The instantiated parser object is not lightweight. It is better to reuse it if possible.`

I wrongly assumed it meant that what was returned by `parser.Parse` was safe to be reused concurrently, which is not the case, and the doc does not state this explicitely indeed.
So far, we've just copied the slice we receive from `parser.Parse` to make it threadsafe.

If you're worried about perf implications, maybe it might just be better to keep it the way it is right now and update the doc to tell other users to just copy the slice received from `parser.Parse`.

Or you can add the *safe mode* option you propose as well, if you prefer.

yujiarista: I have created a PR to update the doc here https://github.com/pingcap/tidb/pull/42572
@xhebox Can you add the safe mode option? It is probably much easier for a contributor who is familiar with the repo.

xhebox:
Closed by 42572.

PR #34753, opened 2022-05-17
server: support deprecation of EOF packets

dveeden:
### What problem does this PR solve?

Adds support for `CLIENT_DEPRECATE_EOF`

Issue Number: close #32118

Note that  #32119 has a different solution to the same problem

Related: https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_response_packets.html

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
Support for the CLIENT_DEPRECATE_EOF feature of the MySQL protocol was added.
```

We could also say something about improved support for "MySQL Connector/J", but this may impact more than just the connector for Java.

PR #42601, opened 2023-03-27
session, com_stmt: fetch all rows during EXECUTE command (#42473, hotfix)

YangKeao:
This is a carefully cherry-pick of https://github.com/pingcap/tidb/pull/42473 on hotfix branch.

```release-note
Previously, TiDB loads data in `FETCH` command with the `cursorFetch` enabled. Now, TiDB will load all data into the memory in `EXECUTE` command. It'll use more memory according to the size of result. And also, TiDB will never report write conflict in `FETCH` command with read-committed isolation level.
```


server/conn_stmt_test.go
```diff
@@ -248,3 +252,122 @@ func TestParseStmtFetchCmd(t *testing.T) {
...
+	tk := testkit.NewTestKitWithSession(t, store, c.Context().Session)
+	tk.MustExec("use test")
+	tk.MustExec("drop table if exists t")
+	tk.MustExec("create table t(a int primary key)")
+	tk.MustExec("insert into t values (1), (2), (3), (4), (5), (6), (7), (8)")
+	tk.MustQuery("select count(*) from t").Check(testkit.Rows("8"))
+
+	getLastStatus := func() uint16 {
+		raw := out.Bytes()
+		return binary.LittleEndian.Uint16(raw[len(raw)-2 : len(raw)])
```
YangKeao, line 278: This test is a little different from the original one, because 6.1 doesn't support DEPRECATE EOF: see https://github.com/pingcap/tidb/pull/34753

tiancaiamao (approved):

Issue #36598, opened 2022-07-27
Support general plan cache


Reminiscent:
## Feature Request

**Is your feature request related to a problem? Please describe:**

TP queries can get lots of performance benefits from Plan Cache, but the current [TiDB Plan Cache](https://docs.pingcap.com/tidb/dev/sql-prepared-plan-cache) can only support [Prepare](https://docs.pingcap.com/tidb/dev/sql-statement-prepare)/[Execute](https://docs.pingcap.com/tidb/v5.0/sql-statement-execute) Statements. Most of our users are still using general queries to access TiDB instead of Prepare/Execute. So if Plan Cache can support general queries, these users can gain some performance benefits as well.

**Describe the feature you'd like:**

When we execute the general query, we can skip the parser and optimizer phase to directly get the physical plan from the plan cache. That can save more time for the TP queries.

**Teachability, Documentation, Adoption, Migration Strategy:**

The detail design will be posted later.

PR #42606, opened 2023-03-27
planner: fix incompatibility issues between plan-cache and expr-blacklist

qw4990:

Issue Number: ref #36598

Problem Summary: planner: add more test cases for non-prep plan cache

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


winoros (approved):

executor/reload_expr_pushdown_blacklist.go
```diff
@@ -67,10 +67,28 @@ func LoadExprPushdownBlacklist(sctx sessionctx.Context) (err error) {
 		}
 		newBlocklist[name] = value
 	}
+	if isSameExprPushDownBlackList(newBlocklist, expression.DefaultExprPushDownBlacklist.Load().(map[string]uint32)) {
+		return nil
+	}
+	expression.ExprPushDownBlackListVer.Add(1)
```
xuyifangreeneyes, line 73: The semantic of `isSameExprPushDownBlackList` and `ExprPushDownBlackListVer` seems not very consistent. For example, one tidb node adds one item into the blacklist and then deletes it from the blacklist, and for another tidb node, it sees the same blacklist before the two two operations and after the two operations and doesn't change `ExprPushDownBlackListVer`. However, in the normal semantic of version, it should change in the case. However, I haven't thought about any bad case and the correctness seems to be ensured.

xuyifangreeneyes (approved):
rest LGTM

executor/reload_expr_pushdown_blacklist.go
xuyifangreeneyes: The name is ok for me. I think it's fine if we add some comment to illustrate.

Issue #26166, opened 2021-07-13
Push down sort property to range partition table


lysu:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```
mysql> create table t1(id int primary key, v int) partition by range(id) (partition p1 values less than(10000), partition p2 values less than(20000));
Query OK, 0 rows affected (0.01 sec)

mysql> create table t2(id int primary key, v int);
Query OK, 0 rows affected (0.01 sec)

mysql> explain select * from t1 where id > 20001 order by id limit 1;
+-----------------------------+---------+-----------+----------------+----------------------------------------------------+
| id                          | estRows | task      | access object  | operator info                                      |
+-----------------------------+---------+-----------+----------------+----------------------------------------------------+
| TopN_8                      | 1.00    | root      |                | test.t1.id, offset:0, count:1                      |
| └─TableReader_17            | 1.00    | root      | partition:dual | data:TopN_16                                       |
|   └─TopN_16                 | 1.00    | cop[tikv] |                | test.t1.id, offset:0, count:1                      |
|     └─TableRangeScan_15     | 3333.33 | cop[tikv] | table:t1       | range:(20001,+inf], keep order:false, stats:pseudo |
+-----------------------------+---------+-----------+----------------+----------------------------------------------------+
4 rows in set (0.00 sec)

mysql> explain select * from t2 where id > 20001 order by id limit 1;
+-----------------------------+---------+-----------+---------------+---------------------------------------------------+
| id                          | estRows | task      | access object | operator info                                     |
+-----------------------------+---------+-----------+---------------+---------------------------------------------------+
| Limit_12                    | 1.00    | root      |               | offset:0, count:1                                 |
| └─TableReader_22            | 1.00    | root      |               | data:Limit_21                                     |
|   └─Limit_21                | 1.00    | cop[tikv] |               | offset:0, count:1                                 |
|     └─TableRangeScan_20     | 1.00    | cop[tikv] | table:t2      | range:(20001,+inf], keep order:true, stats:pseudo |
+-----------------------------+---------+-----------+---------------+---------------------------------------------------+
4 rows in set (0.01 sec)

mysql> set tidb_partition_prune_mode='dynamic';
Query OK, 0 rows affected (0.00 sec)

mysql> explain select * from t1 where id > 20001 order by id limit 1;
+-----------------------------+---------+-----------+----------------+----------------------------------------------------+
[...]

PR #42694, opened 2023-03-29
planner: open fully order prop push down for partition table's table scan and single index scan

winoros:

Issue Number: ref https://github.com/pingcap/tidb/issues/26166

Problem Summary:

We've implemented merge operation between partitions of table scan and single index scan.
So we don't need the former quick path.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


planner/core/casetest/integration_test.go
```diff
@@ -929,6 +929,8 @@ func TestKeepOrderHint(t *testing.T) {
 	tk.MustExec("create table t(a int, b int, primary key(a));")
 	tk.MustExec("create table t1(a int, b int, index idx_a(a));")
 	tk.MustExec("create table th (a int, key(a)) partition by hash(a) partitions 4;")
+	tk.MustExec("create table thp (a int, primary key(a)) partition by hash(a) partitions 4;")
```
Defined2014, line 932: Do we use table `thp`?

Defined2014: Got it

Defined2014 (approved):

qw4990 (approved):

planner/core/find_best_task.go
```diff
@@ -1520,13 +1520,28 @@ func (ds *DataSource) convertToIndexScan(prop *property.PhysicalProperty,
...
 			col, isNew := cop.tablePlan.(*PhysicalTableScan).appendExtraHandleCol(ds)
 			cop.extraHandleCol = col
+			if ds.tableInfo.GetPartitionInfo() != nil && !is.Index.Global {
+				is.Columns = append(is.Columns, model.NewExtraPhysTblIDColInfo())
+				is.schema.Append(&expression.Column{
+					RetType:  types.NewFieldType(mysql.TypeLonglong),
+					UniqueID: ds.ctx.GetSessionVars().AllocPlanColumnID(),
+					ID:       model.ExtraPhysTblID,
+				})
+			}
```
time-and-fate, line 1544: Why do we need this? Does this conflict with `(is *PhysicalIndexScan) initSchema`?

planner/core/task.go
```diff
@@ -1262,120 +1224,7 @@ func (p *PhysicalTopN) pushPartialTopNDownToCop(copTsk *copTask) (task, bool) {
...
-		rootLimit := PhysicalLimit{
-			Count:       p.Count,
-			Offset:      p.Offset,
-			PartitionBy: newPartitionBy,
-		}.Init(p.SCtx(), stats, p.SelectBlockOffset())
-		rootLimit.SetSchema(rootTask.plan().Schema())
-		return attachPlan2Task(rootLimit, rootTask), true
 	} else {
 		return nil, false
 	}
```
time-and-fate, line 1230: This entire if-else branch can be simplified.

winoros: This func will be removed in the next pr since there's fully support for IndexMerge

planner/core/find_best_task.go
winoros: No, that case is to add a partition id for the global index(It's decoded from the value of the global index). Here we add an extra partition id for non-global index which is decoded from the key instead of the value.

Defined2014: `ExtraPhysTblID` will let TiKV return PID in chunk. IndexWorker will use the pid to create a partitionHandle for tableWorker found the handle belongs to which PID.

PR #42769, opened 2023-04-03
ddl: support checkpoint for ingest mode

tangenta:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


ddl/backfilling.go
```diff
@@ -596,7 +597,11 @@ func handleOneResult(result *backfillResult, scheduler backfillScheduler, consum
 		scheduler.drainTasks() // Make it quit early.
 		return result.err
 	}
-	*totalAddedCount += int64(result.addedCount)
+	if result.totalCount > 0 {
```
Benjamin2037, line 600: If this means totalCount is zero or negtive number, we treat it as init value and it should be zero?

ddl/ingest/checkpoint.go
```diff
@@ -0,0 +1,380 @@
...
+	"github.com/pingcap/tidb/kv"
+	"github.com/pingcap/tidb/meta"
+	"github.com/pingcap/tidb/util/intest"
+	"github.com/pingcap/tidb/util/logutil"
+	"go.uber.org/zap"
+)
+
+// CheckpointManager is an interface to manage checkpoints.
+type CheckpointManager interface {
+	IsComplete(taskID int, start, end kv.Key) bool
```
Benjamin2037, line 39: l28y07L7gj
Why not put explaination of interface method in code?

ddl/backfilling_scheduler.go
```diff
@@ -262,24 +266,31 @@ type ingestBackfillScheduler struct {
...
 		reorgInfo: info,
+		sessPool:  sessPool,
 		tbl:       tbl,
 		taskCh:    make(chan *reorgBackfillTask, backfillTaskChanSize),
 		resultCh:  make(chan *backfillResult, backfillTaskChanSize),
 		poolErr:   make(chan error),
 	}
 }
 
+const checkpointUpdateInterval = 5 * time.Second
```
Benjamin2037, line 292: Since it is a checkpoint update interval, should we set it longer? for example 10 minutes?  it seems to redo 10 mins will not cause a very big impact on the user.

tangenta: This is used for integration test. I think it is reasonable to be 10 min.

ddl/backfilling.go
```diff
@@ -596,7 +597,11 @@ func handleOneResult(result *backfillResult, scheduler backfillScheduler, consum
...
+		*totalAddedCount = int64(result.totalCount)
+	} else {
```
Benjamin2037, line 602: So the result.totalCount only will be zero or positive int, so could we remove else part of this if statement?

ddl/backfilling_scheduler.go
Benjamin2037: what I mean is checkpoint could be a little bit bigger time period.

ddl/backfilling_scheduler.go
```diff
@@ -311,8 +328,21 @@ func (b *ingestBackfillScheduler) close(force bool) {
 		return
 	}
 	close(b.taskCh)
-	b.copReqSenderPool.close(force)
-	b.writerPool.ReleaseAndWait()
+	if b.copReqSenderPool != nil {
```
Benjamin2037, line 331: if we first cancel writerPool， then colse readerPool more safe？

ddl/backfilling_scheduler.go
```diff
@@ -439,30 +472,24 @@ func (w *addIndexIngestWorker) HandleTask(rs idxRecResult) {
...
-				result.err = err
-				w.resultCh <- result
-			}
-		}
 		return
 	}
-	result.scanCount = count
-	result.addedCount = count
+	cnt, nextKey := w.checkpointMgr.Status()
+	result.totalCount = cnt
```
Benjamin2037, line 486: Whether line 486 should be put after line 492? the count should be added first then  set to result.totalCount?

ddl/backfilling_scheduler.go
tangenta: Why? If we cancel the `writerPool` first, the chunk read by the reader will be blocked.

ddl/backfilling_scheduler.go
tangenta: The effect is exactly the same since no one reads the result until it is sent back to the channel(`w.resultCh`).

ddl/ingest/checkpoint.go
```diff
@@ -0,0 +1,383 @@
...
+	sess "github.com/pingcap/tidb/ddl/internal/session"
+	"github.com/pingcap/tidb/kv"
+	"github.com/pingcap/tidb/meta"
+	"github.com/pingcap/tidb/util/intest"
+	"github.com/pingcap/tidb/util/logutil"
+	"go.uber.org/zap"
+)
+
+// CheckpointManager is an interface to manage checkpoints.
+type CheckpointManager interface {
```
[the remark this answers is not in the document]
tangenta: Removed.

Benjamin2037 (approved):
LTGM

ddl/backfilling_scheduler.go
```diff
@@ -110,6 +111,8 @@ func (b *txnBackfillScheduler) setupWorkers() error {
 }
 
 func (b *txnBackfillScheduler) sendTask(task *reorgBackfillTask) {
+	b.taskMaxID++
```
wjhuang2016, line 114: Why reallocate the task ID?

ddl/backfilling_scheduler.go
```diff
@@ -288,6 +299,12 @@ func (b *ingestBackfillScheduler) setupWorkers() error {
 		return errors.Trace(errors.New("cannot get lightning backend"))
 	}
 	b.backendCtx = bc
+	mgr, err := ingest.NewCheckpointManager(b.ctx, bc, b.sessPool, job.ID,
```
wjhuang2016, line 302: It shouldn't set the manager in distribute case.

ddl/backfilling_scheduler.go
tangenta: Because it needs to be unique during the lifetime of the DDL job, instead of a task batch.

ddl/backfilling_scheduler.go
tangenta: Updated.

wjhuang2016 (approved):

PR #42780, opened 2023-04-03
lightning: generate and send region job more smoothly

### LINKED ISSUE #42456: ddl: compared with v6.6.0, v7.0.0 has 10% - 13% performance regression
## Bug Report

1. deploy a TiDB cluster and restore the data of ossinsight workload
2. add index `alter table github_events add index test_idx(repo_id,type,action,closed_at,actor_login)`

No performance regression
image

v7.0.0

lance6716:

Issue Number: close #42456

Problem Summary:

1. in the old PR we try to generate all region jobs at once, then process jobs, and generate jobs for unfinished range again, then process again... In order to reduce the lag caused by generating, we now start processing jobs once a key range has generated job, and start generating job once the job need to be generated again, like the region is changed. Also, these work are done concurrently.
2. `unfinishedRange` is removed because we don't have a barrier to stop all processing before use `unfinishedRange` to generate job. Now the after the job is initially generated, the key range of job will be totally finished by worker, or shrank by partitially ingestion. The job may also be splitted into multiple job whose total key range equals to the original job.
3. In order to let the workers schedule the backoffing job for later processing, a regionJobRetryer is introduced. Now the workflow becomes
```
	/*
	   [prepareAndSendJob]-----jobToWorkerCh--->[workers]
	                        ^                       |
	                        |                jobFromWorkerCh
	                        |                       |
	                        |                       v
	               [regionJobRetryer]<--[dispatchJobGoroutine]-->done
	*/
```
Reviews please help check if there's deadlock and if there's a better way to implement it.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


br/pkg/lightning/backend/local/local.go
```diff
@@ -1328,28 +1380,74 @@ func (local *local) ImportEngine(ctx context.Context, engineUUID uuid.UUID, regi
...
+
+func (e *firstError) load() error {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	return e.err
+}
+
+func (local *local) doImport(ctx context.Context, engine *Engine, regionRanges []Range, regionSplitSize, regionSplitKeys int64) error {
+	/*
+	   [prepareAndSendJob]-----jobToWorkerCh--->[workers]
```
lance6716, line 1413: @gozssky Can you give some suggestions on the pattern? Currently regionJob has different stages, multiple workers hold the resouces (pd client, ...) that can advance the stage, and producer/consumer are communicated by channel

lance6716: I can think of two design patterns

one is using goroutines to acquire "worker resource" and then start job, when the job need to sleep some period to retry, release the worker resource. like

```go
for _, j := range jobs {
  w := workerPool.Acquire()
  go func() {j.run(w); w.Release()}
}
```

and

```go
// when meet error
j.w.Release()
sleep()
j.w := workerPool.Acquire()
```

In this method, I want to rely on golang's goroutine scheduling to keep N system thread where N is worker count. But I'm afraid we will spawn too many sleeping goroutines and affect the runtime. 

In the refactoring PR I want to do manually scheduling, where try to keep the N worker goroutines always busy and put the job back when the job needs retry. To solve https://github.com/pingcap/tidb/issues/42456, I should put every time consuming work in worker to make use of concurrency, and remove the barrier in the single-thread producer goroutine. Do you have any suggestions? @gozssky @D3Hunter

br/pkg/lightning/backend/local/region_job.go
```diff
@@ -619,10 +661,141 @@ func (j *regionJob) fixIngestError(
...
+		putBackCh: putBackCh,
+		reload:    make(chan struct{}, 1),
+		done:      make(chan struct{}),
+	}
+}
+
+func (q *regionJobRetryer) run(ctx context.Context) {
+	var (
+		front     *regionJob
+		toPutBack *regionJob
```
gozssky, line 725: What's the difference between `front` and `toPutBack`?

br/pkg/lightning/backend/local/region_job.go
```diff
@@ -619,10 +661,141 @@ func (j *regionJob) fixIngestError(
...
+			front = nil
+		}
+		q.qMu.Unlock()
+
+		switch {
+		case toPutBack != nil:
+			select {
+			case <-ctx.Done():
+				return
+			case <-q.done:
```
D3Hunter, line 742: if we done `close`, but `toPutBack` is taken out, so we might miss one jobWg.Done

D3Hunter (commented):
rest lgtm

br/pkg/lightning/backend/local/local.go
```diff
@@ -1349,28 +1412,83 @@ func (local *Local) ImportEngine(ctx context.Context, engineUUID uuid.UUID, regi
...
+	err = local.doImport(ctx, lf, regionRanges, regionSplitSize, regionSplitKeys)
+	if err == nil {
+		log.FromContext(ctx).Info("import engine success", zap.Stringer("uuid", engineUUID),
+			zap.Int64("size", lfTotalSize), zap.Int64("kvs", lfLength),
+			zap.Int64("importedSize", lf.importedKVSize.Load()), zap.Int64("importedCount", lf.importedKVCount.Load()))
+	}
+	return err
+}
+
+type firstError struct {
```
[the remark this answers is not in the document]
lance6716: https://github.com/pingcap/tidb/pull/42780/commits/75ce8f1f69cc455d713c4dd6e6e8a41189546afe

br/pkg/lightning/backend/local/region_job.go
```diff
@@ -58,45 +73,64 @@ func (j jobStageTp) String() string {
...
+
+	// injected is used in test to set the behaviour
+	injected []injectedBehaviour
 }
 
 type tikvWriteResult struct {
-	sstMeta    []*sst.SSTMeta
-	rangeStats rangeStats
+	sstMeta           []*sst.SSTMeta
+	rangeStats        rangeStats
```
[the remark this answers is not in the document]
lance6716: I prefer we don't embed it because I can't search the usage of embedded fields in IDE 😂

br/pkg/lightning/backend/local/region_job.go
```diff
@@ -41,12 +42,26 @@ import (
 
 type jobStageTp string
 
-// nil -> regionScanned: create a new region job
-// regionScanned -> wrote: write the data to TiKV
-// wrote -> ingested: ingest the data to TiKV
-// ingested -> nil: finish the job
-// regionScanned / wrote -> needRescan: need to rescan the data, maybe region is expanded.
-// needRescan -> nil: discard the job. caller will create a new job from unfinishedRanges.
+// nil -> regionScanned: create a new region job initially
```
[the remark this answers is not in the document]
lance6716: https://github.com/pingcap/tidb/pull/42780/commits/d96e4bca5bfe89298ddeaf98082197a352da6388

br/pkg/lightning/backend/local/region_job.go
```diff
@@ -259,16 +305,13 @@ func (j *regionJob) writeToTiKV(
 		if totalSize >= regionMaxSize || totalCount >= j.regionSplitKeys {
 			// we will shrink the key range of this job to real written range
 			if iter.Valid() && iter.Next() {
```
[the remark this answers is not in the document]
lance6716: Seems we **only** need to call iter.Next()? iter.Valid() must be true in this for-loop, and the return value of iter.Next() means if the new KV is valid.

@gozssky please double check

br/pkg/lightning/backend/local/region_job.go
```diff
@@ -619,10 +661,141 @@ func (j *regionJob) fixIngestError(
...
+		switch {
+		case toPutBack != nil:
+			select {
+			case <-ctx.Done():
+				return
+			case <-q.done:
+				return
+			case q.putBackCh <- toPutBack:
+				toPutBack = nil
+			}
```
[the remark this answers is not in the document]
lance6716: there are two ways to write this pattern:

```
select {
case ...:
  select {
  ...
  }
}
```

```
if stateA {
  select {}
} else if stateB {
  select {}
}
```

I slightly prefer the latter. do you have some suggestions? @gozssky

br/pkg/lightning/backend/local/region_job.go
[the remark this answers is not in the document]
gozssky: Yes, only `Next` is enough.

br/pkg/lightning/backend/local/region_job.go
lance6716: https://github.com/pingcap/tidb/pull/42780/commits/20fc8f5d064dcd70714373f08ca83bea43ada28f

br/pkg/lightning/backend/local/region_job.go
```diff
@@ -41,12 +42,39 @@ import (
...
++-----+ingested+---------+
+
+	+---+----+
+	    |
+	    v
+
+above diagram shows the state transition of a region job, here are some special
+cases:
+  - regionScanned can directly jump to ingested if the keyRange has no data
+  - regionScanned can only transit to wrote. TODO: check if it should be transited
```
lance6716, line 71: @gozssky in release-6.5, if lightning meet "region is changed" error, like region split, `WriteToTiKV` will return a !isRetryableImportTiKVError error, and it will break the whole import procedure? Will TiKV side return "region is changed" error during ImportSstService.write?

br/pkg/lightning/backend/local/region_job.go
gozssky: I think we can get and check `front` in each loop, it's unnecessary to have a new variable.

br/pkg/lightning/backend/local/region_job.go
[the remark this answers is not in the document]
lance6716: https://github.com/pingcap/tidb/pull/42780/commits/0b42bd29ef7ea87c54995941b3225e10091cc2c7

br/pkg/lightning/backend/local/region_job.go
gozssky: The workflow is:
1. Get `front`.
2. Check until time and determine whether need to sleep.
3. After sleep, pop `front` and add it to the queue.

lance6716: > I think we can get and check `front` in each loop, it's unnecessary to have a new variable.

I don't understand your suggesion, do you mean

```
if q.toPutBack != nil {
  select {
  case can-put-back
  case waiting-for-q.q[0]
  }
} else {
  select {
  case waiting-for-q.q[0]
  ...
  }
}
```

q.q should be proetced by mutex, so I must use `front`

br/pkg/lightning/backend/local/region_job.go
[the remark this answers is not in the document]
gozssky: I prefer to remove `toPutBack` and move this logic after `case <-time.After(time.Until(front.waitUntil))`.

br/pkg/lightning/backend/local/region_job.go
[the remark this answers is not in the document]
D3Hunter: i mean inline struct `rangeStats` into this one, seems only used here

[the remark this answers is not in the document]
lance6716: https://github.com/pingcap/tidb/pull/42780/commits/3c2b1c4e3e44a92ff880d1cad725a8ed93909278

br/pkg/lightning/backend/local/region_job.go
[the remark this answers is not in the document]
lance6716: https://github.com/pingcap/tidb/pull/42780/commits/8836a0271f86df42700769a0c83a2fb952d74c5c

D3Hunter (approved):

lance6716:
/lgtm cancel

fix a problem that sending to `putBackCh` will block `retryer.push` https://github.com/pingcap/tidb/pull/42780/commits/45b208d7f5043e4a240417039f0601017c432cc6

because `retryer.push` must access the inner queue, sending to `putBackCh` must not access the inner queue to avoid blocking the former. So we should use a temporary `toPutBack` to reduce the locking scope. @D3Hunter @gozssky

lance6716:
/hold

need to fix https://github.com/pingcap/tidb/pull/42780#discussion_r1162623760

lance6716:
> /lgtm cancel
> 
> fix a problem that sending to `putBackCh` will block `retryer.push` [45b208d](https://github.com/pingcap/tidb/commit/45b208d7f5043e4a240417039f0601017c432cc6)
> 
> because `retryer.push` must access the inner queue, sending to `putBackCh` must not access the inner queue to avoid blocking the former. So we should use a temporary `toPutBack` to reduce the locking scope. @D3Hunter @gozssky

@D3Hunter @gozssky PTAL

lance6716:
lightning integration test passed https://ci.pingcap.net/blue/organizations/jenkins/br_ghpr_unit_and_integration_test/detail/br_ghpr_unit_and_integration_test/15798/pipeline

D3Hunter (approved):

br/pkg/lightning/backend/local/region_job.go
```diff
@@ -619,10 +666,169 @@ func (j *regionJob) fixIngestError(
...
+	close(q.done)
+
+	// don't need to lock them in fact because of the requirement of calling close
+	q.protectedQueue.mu.Lock()
+	defer q.protectedQueue.mu.Unlock()
+	q.protectedToPutBack.mu.Lock()
+	defer q.protectedToPutBack.mu.Unlock()
+	ret := len(q.protectedQueue.q)
+	if q.protectedToPutBack.toPutBack != nil {
+		ret++
```
gozssky, line 831: The `ret` may be changed after this function returns if `run` is about to send a new job to `q.putBackCh` after line 758. Although you close `q.done`, it doesn't mean it must enter the `case <-q.done:` branch on line 771.

br/pkg/lightning/backend/local/region_job.go
```diff
@@ -619,10 +666,169 @@ func (j *regionJob) fixIngestError(
...
+	select {
+	case q.reload <- struct{}{}:
+	default:
+	}
+	return true
+}
+
+// close will return the number of jobs that are not put back when first called.
+// close MUST be called when caller is sure that regionJobRetryer holds no job
+// or regionJobRetryer.run has exited.
```
gozssky, line 813: If `regionJobRetryer.run` has exited, what is `q.done` is used for?

br/pkg/lightning/backend/local/region_job.go
lance6716: Yes, so I add a comment when call `close` `run` should be exited or all jobs are done

```
// close MUST be called when caller is sure that regionJobRetryer holds no job
// or regionJobRetryer.run has exited.
```

Maybe I should let retryer owns the jobWg, so close can be called at any time and `run` routine `Done` the jobWg 🤔

br/pkg/lightning/backend/local/region_job.go
lance6716: the other case is all job is finished, caller use `retryer.close()` to notify `retryer.run()`

gozssky: The semantics is so confusing. I suggest using `ctx` to notify `run` to exit.

gozssky (approved):

Issue #42160, opened 2023-03-13
A better way to init stats


xuyifangreeneyes:
## Enhancement

Currently, when a tidb-server starts, one thing to do is to init stats. It loads stats meta and indexes' Histogram and TopN(It doesn't load columns' Histogram and TopN to reduce stats mem usage. Columns' Histogram and TopN are loaded when they are needed by the optimizer). If there are many tables and partitions, it may take minutes even one hour to init stats, especially when the cluster is under high pressure and the tidb-server restarts. During the period of init stats, no table stats can be used by the optimizer. The optimizer uses pseudo stats to generate plans, which are probably inefficient and put more pressure to the cluster.

PR #42876, opened 2023-04-07
statistics: remove statistics.Column.Count

xuyifangreeneyes:

Issue Number: ref #42160

Problem Summary:

Remove `statistics.Column.Count`. 

Before the PR, in order to maintain `statistics.Column.Count`, we need to read `mysql.stats_top_n` and `mysql.stats_buckets` for each column when init stats, which is time-consuming. On the other hand, the usage of `statistics.Column.Count` is limited. We change `(*DataSource).getColumnNDV` and add `(*Column).StatsCollected` to get rid of `statistics.Column.Count`.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/planner/core/stats.go
+++ b/planner/core/stats.go
@@ -163,9 +163,13 @@ func (p *baseLogicalPlan) DeriveStats(childStats []*property.StatsInfo, selfSche
 // histogram of `DataSource` which is retrieved from storage(not the derived one).
 func (ds *DataSource) getColumnNDV(colID int64) (ndv float64) {
 	hist, ok := ds.statisticTable.Columns[colID]
-	if ok && hist.Count > 0 {
-		factor := float64(ds.statisticTable.RealtimeCount) / float64(hist.Count)
-		ndv = float64(hist.Histogram.NDV) * factor
+	if ok && hist.IsStatsInitialized() {
+		ndv = float64(hist.Histogram.NDV)
+		// TODO: a better way to get the row count derived from the last analyze.
+		if hist.IsFullLoad() {
+			factor := float64(ds.statisticTable.RealtimeCount) / hist.TotalRowCount()
+			ndv *= factor
+		}
 	} else {
 		ndv = float64(ds.statisticTable.RealtimeCount) * distinctFactor
 	}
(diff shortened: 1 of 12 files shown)


planner/core/stats.go
```diff
@@ -163,9 +163,13 @@ func (p *baseLogicalPlan) DeriveStats(childStats []*property.StatsInfo, selfSche
...
-	if ok && hist.Count > 0 {
-		factor := float64(ds.statisticTable.RealtimeCount) / float64(hist.Count)
-		ndv = float64(hist.Histogram.NDV) * factor
+	if ok && hist.IsStatsInitialized() {
+		ndv = float64(hist.Histogram.NDV)
+		// TODO: a better way to get the row count derived from the last analyze.
+		if hist.IsFullLoad() {
+			factor := float64(ds.statisticTable.RealtimeCount) / hist.TotalRowCount()
+			ndv *= factor
+		}
```
xuyifangreeneyes, line 172: Here the way to calculate column NDV is changed. We assume that the change is ok for most cases.

statistics/handle/update.go
```diff
@@ -998,12 +998,12 @@ var AutoAnalyzeMinCnt int64 = 1000
 // TableAnalyzed checks if the table is analyzed.
 func TableAnalyzed(tbl *statistics.Table) bool {
 	for _, col := range tbl.Columns {
-		if col.Count > 0 {
+		if col.StatsVer != statistics.Version0 {
 			return true
 		}
 	}
 	for _, idx := range tbl.Indices {
-		if idx.Histogram.Len() > 0 {
```
xuyifangreeneyes, line 1006: The condition is wrong since `idx.Histogram` may be evicted.

statistics/handle/update.go
```diff
@@ -998,12 +998,12 @@ var AutoAnalyzeMinCnt int64 = 1000
 // TableAnalyzed checks if the table is analyzed.
 func TableAnalyzed(tbl *statistics.Table) bool {
 	for _, col := range tbl.Columns {
-		if col.Count > 0 {
+		if col.StatsVer != statistics.Version0 {
```
qw4990, line 1001: Why not use `col.StatsCollected`?

xuyifangreeneyes: If the column is added and its stats are generated according to its default value, `StatsCollected` returns true but the column is not analyzed. I will add another function called `IsAnalyzed` for better encapsulation.

Issue #39892, opened 2022-12-13
`InitLogger` before `CreateMockStore` in tests to avoid data race


time-and-fate:
## Bug Report

In tests related to slow log, we would use `InitLogger` to create a separate slow log file.

In `CreateMockStore`, it will run queries in a separate goroutine and access the slow query logger. The `InitLogger` would also modify the slow query logger.

So there might be a data race if we first `CreateMockStore` and then `InitLogger`.

Ref: #39746

PR #42899, opened 2023-04-10
planner, runtimefilter: Runtime filter generator, Support IN and Local RF

### LINKED ISSUE #40220: Support Runtime Filter in TiDB
# Goal

Improve the **efficiency of Join** execution in OLAP scenarios.

# Background

PhysicalHashJoin completes the Join by constructing a HashTable from the data in the right table, and the data in the left table keeps probing the HashTable. If the Join Key value in part of the Probe process cannot Hit HashTable, then this part of the data in the description does not exist in the right table and will not appear in the last Join result.

In other words, this part of the data from PhysicalTableScan to send to PhysicalHashJoin is meaningless. If you can **filter out this part of the Join Key data in advance**, it will reduce scanning time and network overhead, thereby greatly **improving Join efficiency** .

# Runtime Filter

Runtime Filter is a predicate that is generated during query planning **The predicate that takes values dynamically**. This predicate has the same function as other predicates in Selection. It is applied to filter rows in TableScan that do not meet the predicate condition. 

The only difference is that **the parameter values** of the Runtime Filter predicate are **constructed in HashJoin.**

The following is an example:
T1 is a fact table with 100w data rows and T2 is a dimension table with 100 data rows.
```select * from T1, T2 where T1.id=T2.id```
The RF execution method is to scan the data of T2 first, PhysicalHashJoin calculates a filter based on the data of T2, such as the maximum and minimum values of the T2 data. Then send this filter to TableFullScan waiting to scan T1. T1 then applies this filter and gives the filtered data to PhysicalHashJoin, thereby reducing the rows of probe hash tables and network overhead.

```
                         2. build RF values
            +-------->+-------------------+
            |         |PhysicalHashJoin   |<-----+
            |    +----+                   |      |
4.after RF  |    |    +-------------------+      | 1. scan T2
    5000    |    |3. send RF                     |      100
            |    | filter data                   |
            |    |                               |
      +-----+----v------+                +-------+--------+
      |  TableFullScan  |                | TabelFullScan  |
      |      T1         |                |      T2        |
      +-----------------+                +----------------+
```

**In summary, Runtime Filters are generated in the planner, built in PhysicalHashJoin, and applied in TableScan.**

## Type of predicate
1. IN, it is suitable for the case where the Join Key column has a small number of values. The predicate is of the form T1.id in (1, 5, 7, 8, 10)
2. MIN MAX, suitable for columns where the Join Key column is a range value. T1.id > 1 and T1.id < 10
3. Bloom Filter, suitable for situations where there are many values in the Join Key column.

## TiFlash or TiKV
There are two storage engine in TiDB: TiFlash and TiKV. To achieve the best efficiency, RF necessary to combine the "predicate push down" and "delayed materialization" of the storage engine. However, TiKV does not have the function of delayed materialization because it is line storage engine, so RF applied to TiKV is not ideal.
**Only TiFlash supports RF.**

# Precondition
- [ ] Delayed materialization (in design), @Lloyd-Pottiger https://github.com/pingcap/tidb/pull/39654/files
- [ ] Bloom Filter Index and Bloom Filter Expression @windtalker
- [ ] Asynchronous read @windtalker

# Stage goal

1. The first phase supports Broadcast local, IN the situation.
2. The second phase supports Shuffle Global's RF, as well as Bloom Filter, Min Max, BF/IN.

# Design
Runtime Filter needs to be planned in the Planner and executed in Compute Engine. So the design is mainly divided into two parts:
1. Planner: Based on the matching pattern, generate RF and assign to the correct target node. ( TiDB )
2. Compute Engine: HashJoinNode builds RF and applies to target node ( TiFlash )
This article does not cover the design of Global's RF.

## Planner design
The planner part of the RF is mainly used to identify HashJoin nodes through the structure of Tree, and plan the RF that meet the ```Patterns``` to the corresponding target node. According to the above logic, the Planner can be split into two core processes:
1. Generate Filter: Responsible for matching HashJoinNode that meets the ```Patterns```, and combining EqPredicates on HashJoinNode to construct RF.
2. Assign Filter: Find the target node according to the constructed RF.

**Process**
Preorder traversal PhysicalPlanTree. Generate Filter when encountering PhysicalHashJoin node, Assign Filter when encountering TableFullScan.

### Generate Filter
Patterns are following: 
1. PhysicalHashJoin :  exclude Left outer，full outer，anti join 
2. Predicate
  + eq predicate 
  + The left child is only column expr, or cast (column expr).
  + Right child types: numeric type, bool, character type. (Other json, blob, binary, etc. are not supported)
  + Evaluate selectivity

### Assign Filter
The relationship between the Hash Join node and the Target Node node is shown in the figure below, and in real cases, one or more nodes may be interspersed between the two.
image

If TableScan contains the same Column as RF (the **same unique id**), then this TableScan will be assigned the current RF.
+ Principle: RF column can be pushed down as long as it has not undergone any other transformation, and the untransformed RF column is the same as the column unique id in TargetNode.

## Compute Engine design
The main work of the Runtime Filter at the execution layer is as follows:
1. Compile: Translate the pb structure into executable RF, and associate the Build and Probe sides through Ref.
2. Generator Values: During the execution of HashBuild, the values of the required columns in each Block are continuously collected and RF expressions are constructed
3. Apply RF: When RF is ready, push RF down to the storage engine as a normal expression.

### Compile

As shown in the figure, if a RF is encountered in the Compile stage, the corresponding vector < RuntimeFilter > is constructed RF and the reference is stored in HashBuild and Unordered respectively. The two are held together, one is responsible for constructing and the other is responsible for using.
image

### Generator Values
Generator Values continuously obtain new Block data through the read method of HashBuild and obtain the data in the Block according to the column name required by the RF and increment to update the RuntimeFilter.

**RF cost control**
If IN values are too many, it may cause the RF to run too inefficiently. So we need to add a runtime_filter_max_in_values variable here to control that if the maximum value is exceeded, RF is discarded.

### Apply RF
The role of the Apply RF is mainly to RF application to the Scan thread, and its location is more important in the whole process. The following are two prerequisites:
1. After RF ready: Occurs after RF construction. Since RF must wait for the Probe to complete the collection of Join Values before it can be applied by the Target Node. Therefore, the TargetNode's read must occur after RF Read.
    + is_ready: HashJoin and TargetNode jointly hold is_ready variable. TargetNode will not enable scan out of the box until is_ready variable is set to true.
2. Before scan thread: Occurs before the scan thread starts. If the scan of the storage layer has already started, even after the RF is built, it cannot be pushed down to the storage layer for filtering. It cannot reduce the IO effect.
    + runtime_filter_wait_time_ms: Maximum waiting time for the runtime filter. If this waiting time is exceeded, no RF is performed.
image

# Added session variables
1. runtime_filter_type：in，min_max, bf，in/bf。
2. runtime_filter_mode：local，global
3. runtime_filter_wait_time_ms: the most time spent waiting for the target node, more than this time does not wait for RF. (for TiFlash for)
4. runtime_filter_in_max_values: The number of values used to control IN this RF, if it exceeds RF will not be run. (for TiFlash for)

# Observability
1. Explain: RF id, RF producer, PF consumer, reference column. Used to check the correctness of the planner layer RF.
```
MySQL [ml_test]> explain select * from  t1, t2 where t1.k2=t2.k2;
+------------------------------+----------+-----------+---------------+---------------------------------------------------------------------------+
| id                           | estRows  | task      | access object | operator info                                                             |
+------------------------------+----------+-----------+---------------+---------------------------------------------------------------------------+
| HashJoin_8                   | 12487.50 | root      |               | inner join, equal:[eq(ml_test.t1.k2, ml_test.t2.k2)], rf_1: producer t2.k2|
| ├─TableReader_15(Build)      | 9990.00  | root      |               | data:Selection_14                                                        |
| │ └─Selection_14             | 9990.00  | cop[tikv] |               | not(isnull(ml_test.t2.k2))                                              |
| │   └─TableFullScan_13       | 10000.00 | cop[tikv] | table:t2      | keep order:false, stats:pseudo                                          |
| └─TableReader_12(Probe)      | 9990.00  | root      |               | data:Selection_11                                                        |
|   └─Selection_11             | 9990.00  | cop[tikv] |               | not(isnull(ml_test.t1.k2))                                               |
|     └─TableFullScan_10       | 10000.00 | cop[tikv] | table:t1      | keep order:false, stats:pseudo , rf_1: consumer t1.k2                    |
+------------------------------+----------+-----------+---------------+---------------------------------------------------------------------------+
```
2. Profile display for RF performance tuning and execution data presentation.
      + RF waiting time
      + RF construction time
      + Changes in data volume before and after RF
      + etc

### PR DESCRIPTION
**Pull request by elsa0520**

Issue Number: close #40220 

Problem Summary:

Complete the Runtime Filter Local mode, and implement the planner layer of the IN type.

Process：
The preorder traverses the physical plan tree, generates RF, and assigns it to the corresponding scan node.
1. Generates RF: only hash join node and eq predicate will generate RF.
2. Assign RF: When the corresponding RF column is found, this RF will be assigned to the current scan node. Only [TiFlash] TableScan will be assigned RF.
3. Add selection upper then target scan node: for later logic.

Session Variables:
1. enable_runtime_filter: default is false. It is used to control enable runtime filter or not.
2. runtime_filter_mode: default is local. Now, we only support local mode.
4. runtime_filter_type:default is IN. It means the filter type is IN predicate. Now, we only support IN type.

WIP:
I changed tidb_enforce_mpp and tidb_isolation_read_engines to global for test easier. I will change it back later. 

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


planner/core/physical_plans.go
```diff
@@ -2037,6 +2077,11 @@ type PhysicalSelection struct {
 	// The flag is only used by cost model for compatibility and will be removed later.
 	// Please see https://github.com/pingcap/tidb/issues/36243 for more details.
 	fromDataSource bool
+
+	// The flag indicates whether this Selection is used for RuntimeFilter
+	// True: Used for RuntimeFilter
+	// False: Only for normal conditions
+	hasRFConditions bool
```
winoros, line 2084:
```suggestion
	isRFConditions bool
```
will be better?

planner/core/runtime_filter_generator.go
```diff
@@ -0,0 +1,233 @@
...
+		if generator.matchEQPredicate(eqPredicate, hashJoinPlan.RightIsBuildSide()) {
+			// construct runtime filter
+			newRFList, targetColumnUniqueId := NewRuntimeFilter(generator.rfIdGenerator, eqPredicate, hashJoinPlan)
+			// update generator rf list
+			rfList := generator.columnUniqueIdToRF[targetColumnUniqueId]
+			if rfList == nil {
+				generator.columnUniqueIdToRF[targetColumnUniqueId] = newRFList
+			} else {
+				// todo how to create a ref for a list
+				//rfList = append(rfList, newRFList...)
```
winoros, line 88: What does this TODO mean?

planner/core/runtime_filter_generator_test.go
```diff
@@ -0,0 +1,92 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package core_test
+
+import (
+	"fmt"
```
winoros, line 18: split to two import group

planner/core/runtime_filter_generator_test.go
```diff
@@ -0,0 +1,92 @@
...
+		testdata.OnRecord(func() {
+			output[i].SQL = ts
+			output[i].Plan = testdata.ConvertRowsToStrings(tk.MustQuery("explain " + ts).Rows())
+		})
+		tk.MustQuery("explain " + ts).Check(testkit.Rows(output[i].Plan...))
+	}
+	//rows := tk.MustQuery("explain select * from t1, t2 where t1.k1=t2.k1 and t2.k2 = 1").Rows()
+	//fmt.Println(testdata.ConvertRowsToStrings(rows))
+	//rows = tk.MustQuery("explain select * from t1, t2 where t1.k1=t2.k1 and t2.k2 = 1").Rows()
+	//fmt.Println(testdata.ConvertRowsToStrings(rows))
```
winoros, line 91: remove?

planner/core/runtime_filter_generator.go
```diff
@@ -0,0 +1,233 @@
...
+		return false
+	default:
+		for _, childNode := range currentNode.Children() {
+			if generator.belongsToSameFragment(childNode, targetNode) {
+				return true
+			}
+		}
+		return false
+	}
+}
```
winoros, line 233: This func is not sufficient enough. We can move this to the place we cut fragments?

planner/core/runtime_filter_generator.go
elsa0520: Sorry, I will remove it later

planner/core/runtime_filter_generator.go
elsa0520: If the target node and source node are belongs to different fragment, the RF mode should be global. Now, we need to judge if RF is global or local. If RF is global, we doesn't support now, so we need to remove it.
So it cannot be executed in the place where the fragment is split later.

planner/core/resolve_indices.go
```diff
@@ -653,6 +653,14 @@ func (p *PhysicalApply) ResolveIndices() (err error) {
 	return
 }
 
+func (p *PhysicalTableScan) ResolveIndices() (err error) {
+	p.physicalSchemaProducer.ResolveIndices()
```
time-and-fate, line 657: Unhandled error.

planner/core/resolve_indices.go
```diff
@@ -653,6 +653,14 @@ func (p *PhysicalApply) ResolveIndices() (err error) {
...
+	for i, column := range p.schema.Columns {
+		column.Index = i
+	}
```
time-and-fate, line 660: Why do we need this?

planner/core/physical_plans.go
```diff
@@ -2058,6 +2098,13 @@ type PhysicalSelection struct {
 	// The flag is only used by cost model for compatibility and will be removed later.
 	// Please see https://github.com/pingcap/tidb/issues/36243 for more details.
 	fromDataSource bool
+
+	// todo Since the feature of adding filter operators has not yet been implemented,
+	// the following code for this function will not be used for now.
+	// The flag indicates whether this Selection is used for RuntimeFilter
+	// True: Used for RuntimeFilter
+	// False: Only for normal conditions
+	// hasRFConditions bool
```
time-and-fate, line 2107: Just curious, what's this for?

planner/core/explain.go
```diff
@@ -246,6 +248,15 @@ func (p *PhysicalTableScan) OperatorInfo(normalized bool) string {
...
 	}
+	if len(p.runtimeFilterList) > 0 {
+		buffer.WriteString(", runtime filter:")
+		for i, runtimeFilter := range p.runtimeFilterList {
+			if i != 0 {
+				buffer.WriteString(", ")
+			}
+			buffer.WriteString(runtimeFilter.ExplainInfo(false))
+		}
+	}
```
time-and-fate, line 259: Just to remind, any changes in the content will cause a different plan digest and therefore be considered as a different execution plan by modules like stmt summary.

sessionctx/variable/sysvar.go
```diff
@@ -2518,6 +2520,59 @@ var defaultSysVars = []*SysVar{
...
+	{Scope: ScopeGlobal | ScopeSession, Name: EnableRuntimeFilter, Value: BoolToOnOff(DefEnableRuntimeFilter), Type: TypeBool,
+		SetSession: func(s *SessionVars, val string) error {
+			s.enableRuntimeFilter = TiDBOptOn(val)
+			return nil
+		},
+		GetSession: func(s *SessionVars) (string, error) {
+			return BoolToOnOff(s.enableRuntimeFilter), nil
+		},
+	},
+	{Scope: ScopeGlobal | ScopeSession, Name: RuntimeFilterTypeName, Value: DefRuntimeFilterType, Type: TypeStr,
```
time-and-fate, line 2533: Not a big problem and I think it's OK to remain them unchanged.
Just to mention, usually, we don't implement `GetSession` and in that case, tidb will return what the user inputs.

sessionctx/variable/tidb_vars.go
```diff
@@ -989,6 +989,12 @@ const (
 	// TiDBTTLRunningTasks limits the count of running ttl tasks. Default to 0, means 3 times the count of TiKV (or no
 	// limitation, if the storage is not TiKV).
 	TiDBTTLRunningTasks = "tidb_ttl_running_tasks"
+	// EnableRuntimeFilter control runtime filter, true: runtime filter open.
+	EnableRuntimeFilter = "enable_runtime_filter"
+	// RuntimeFilterTypeName the value of is string, a runtime filter type list split by ",", such as: "IN,MIN_MAX"
+	RuntimeFilterTypeName = "runtime_filter_type"
+	// RuntimeFilterModeName the mode of runtime filter, such as "LOCAL"
+	RuntimeFilterModeName = "runtime_filter_mode"
```
time-and-fate, line 997: TiDB-specific variables almost always need a "tidb_" prefix.
And I'm not sure but maybe we need to ask PM whether we need a "tidb_opt_" prefix here.

planner/core/runtime_filter_generator.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+	}
+}
+
+func (generator *RuntimeFilterGenerator) assignRuntimeFilter(plan PhysicalPlan) {
+	// only full scan
+	physicalTableScan := plan.(*PhysicalTableScan)
+	if !physicalTableScan.isFullScan() {
+		return
+	}
```
time-and-fate, line 114: Is this a necessary check, or just in case something unexpected happens?

planner/core/physical_plans.go
elsa0520: has been removed

planner/core/explain.go
```diff
@@ -171,7 +171,9 @@ func (p *PhysicalTableScan) TP() string {
 
 // ExplainInfo implements Plan interface.
 func (p *PhysicalTableScan) ExplainInfo() string {
-	return p.AccessObject().String() + ", " + p.OperatorInfo(false)
+	var buffer strings.Builder
+	buffer.WriteString(p.AccessObject().String() + ", " + p.OperatorInfo(false))
+	return buffer.String()
```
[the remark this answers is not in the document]
elsa0520: Because I changed the implementation of operatorinfo, I feel that it will be more beautiful to use stringbuilder. But it doesn't matter and you can leave it alone.

planner/core/explain.go
elsa0520: Are there any other effects of different execution plan? With rf and without rf itself is a different planning plan

planner/core/physical_plans.go
elsa0520: There are two filter conditions on selection, one is runtime filter and the other is common filter. 
This field is mainly used to identify that the current selection contains predicates of type runtime filter. 
After marking, selection will wait for the RF status to become ready before executing.

planner/core/runtime_filter_generator.go
elsa0520: It is a necessary check. If table scan is not a full scan, the runtime filter will not be assign

sessionctx/variable/sysvar.go
elsa0520: RuntimeFilter has different type "IN,MIN_MAX,BLOOMFILTER". Different types of queries are suitable for different types. So we need to change the type in session var.

sessionctx/variable/tidb_vars.go
elsa0520: RuntimeFilter is not only a tidb function but also a tiflash function. So should we add this prefix ?

sessionctx/variable/sysvar.go
elsa0520: Also, I removed get session function

planner/core/physical_plans.go
```diff
@@ -2058,6 +2092,13 @@ type PhysicalSelection struct {
...
```
winoros, line 2101: Why commented out this variable?

planner/core/resolve_indices.go
winoros: +1

sessionctx/variable/tidb_vars.go
time-and-fate: I thought we should, because we have `tidb_enable_mpp`, `tidb_mpp_store_fail_ttl`, and `tidb_max_tiflash_threads`, but then I found now we also have `mpp_exchange_compression_mode` and `mpp_version`.
So now I'm also not 100% sure 😂

sessionctx/variable/session.go
```diff
@@ -3367,3 +3376,107 @@ func (s *SessionVars) GetRelatedTableForMDL() *sync.Map {
...
+	Local RuntimeFilterMode = iota + 1
+	Global
+)
+
+func (rfMode RuntimeFilterMode) String() string {
+	switch rfMode {
+	case Local:
+		return "LOCAL"
+	case Global:
+		return "Global"
```
time-and-fate, line 3464: Inconsistent case.

planner/core/physical_plans.go
time-and-fate: A bit confused.
So do we need this field to make the runtime filter work correctly?
I thought the newly added fields in `PhysicalTableScan` is enough.

planner/core/runtime_filter_generator.go
time-and-fate: I don't quite understand this. Would it affect correctness?

planner/core/runtime_filter_generator.go
```diff
@@ -0,0 +1,247 @@
...
+		srcColumn = eqPredicate.GetArgs()[1].(*expression.Column)
+	} else {
+		targetColumn = eqPredicate.GetArgs()[1].(*expression.Column)
+		srcColumn = eqPredicate.GetArgs()[0].(*expression.Column)
+	}
+	// match target column
+	// condition1: the target column must be real column
+	// condition2: the target column has not undergone any transformation
+	// todo: cast expr in target column
+	if targetColumn.IsHidden || targetColumn.OrigName == "" {
```
time-and-fate, line 199: I'm a bit suspecting that if `targetColumn.OrigName == ""` is a reliable check.

planner/core/runtime_filter.go
```diff
@@ -0,0 +1,249 @@
...
+	id             int
+	buildNode      *PhysicalHashJoin
+	srcExprList    []*expression.Column
+	targetExprList []*expression.Column
+	rfType         RuntimeFilterType
+	// The following properties need to be set after assigning a scan node to RF
+	rfMode     RuntimeFilterMode
+	targetNode *PhysicalTableScan
+	// The explain id will be set when runtime filter clone()
+	// It is only used for runtime filter pb
```
time-and-fate, line 70: So, does this mean the `RuntimeFilter` must be cloned before being converted to pb?

planner/core/physical_plans.go
elsa0520: This is a field for tiflash. The compile stage of tiflash, it use this field to recognize special selection

sessionctx/variable/tidb_vars.go
elsa0520: Yes. So what exactly are our rules for adding prefixes? 😂

planner/core/runtime_filter_generator.go
elsa0520: I checked the current code, as long as the column in the original table is currently added to this property. So here is my judgment first. If there is a problem later, it can be repaired in time.

planner/core/runtime_filter_generator.go
elsa0520: It is my fault. The runtime filter could be assign whatever fullscan or not.

planner/core/runtime_filter.go
elsa0520: Currently yes, because as long as the task is sent to tiflash, ```clone``` will be called again.
Secondly, because the attribute explain id is actually used exclusively for the PB structure, it is not used before. 
So I delayed the construction of this property, so that the construction and use of the property should be as close as possible.

planner/core/resolve_indices.go
time-and-fate, line 657:
```suggestion
	err := p.physicalSchemaProducer.ResolveIndices()
```

planner/core/runtime_filter_generator_test.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+func TestRuntimeFilterGenerator(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	defer config.RestoreFunc()()
+	config.UpdateGlobal(func(conf *config.Config) {
+		conf.Log.Level = "debug"
+	})
+	require.NoError(t, logutil.InitLogger(config.GetGlobalConfig().Log.ToLogConfig()))
```
time-and-fate, line 39: Please first `InitLogger` then `CreateMockStore` to avoid data race. You can see https://github.com/pingcap/tidb/issues/39892 for details.

time-and-fate (commented):
Please update the go.mod and resolve the conflicts with the master branch.
Others LGTM.

winoros (approved):

planner/core/runtime_filter_generator.go
```diff
@@ -0,0 +1,242 @@
...
+			// assign rf to current node
+			runtimeFilter.assign(physicalTableScan, scanOutputColumn)
+			currentRFList = append(currentRFList, runtimeFilter)
+		}
+	}
+
+	if len(currentRFList) == 0 {
+		return
+	}
+
```
AilinKid, line 141: what does the usage of `currentRFList`？

planner/core/testdata/runtime_filter_generator_suite_out.json
```diff
@@ -0,0 +1,178 @@
...
+        "Plan": [
+          "TableReader_36 0.80 root  MppVersion: 1, data:ExchangeSender_35",
+          "└─ExchangeSender_35 0.80 mpp[tiflash]  ExchangeType: PassThrough",
+          "  └─HashJoin_29 0.80 mpp[tiflash]  inner join, equal:[eq(test.t2.k1, Column#7)]",
+          "    ├─ExchangeReceiver_33(Build) 1.00 mpp[tiflash]  ",
+          "    │ └─ExchangeSender_32 1.00 mpp[tiflash]  ExchangeType: Broadcast, Compression: FAST",
+          "    │   └─Selection_31 1.00 mpp[tiflash]  not(isnull(test.t2.k1))",
+          "    │     └─TableFullScan_30 1.00 mpp[tiflash] table:t2 pushed down filter:empty, keep order:false",
+          "    └─Projection_34(Probe) 0.80 mpp[tiflash]  test.t1.k1, plus(test.t1.k1, 1)->Column#7",
+          "      └─Selection_26 0.80 mpp[tiflash]  not(isnull(plus(test.t1.k1, 1)))",
```
AilinKid, line 133: Seems we have a todo here.
the simple expression can be deconstructed to `k1 =  values-1`
or
just let Tiflash support expression choose when doing lazy materialization.

sessionctx/variable/session.go
```diff
@@ -3413,3 +3420,118 @@ func (s *SessionVars) GetRelatedTableForMDL() *sync.Map {
 func (s *SessionVars) EnableForceInlineCTE() bool {
 	return s.enableForceInlineCTE
 }
+
+// EnableRuntimeFilter return runtime filter mode whether OFF
+func (s *SessionVars) EnableRuntimeFilter() bool {
```
AilinKid, line 3425: EnableRuntimeFilter sounds like an assignment? while its a judgement here?

sessionctx/variable/session.go
```diff
@@ -3413,3 +3420,118 @@ func (s *SessionVars) GetRelatedTableForMDL() *sync.Map {
...
+}
+
+// GetRuntimeFilterTypes return the session variable runtimeFilterTypes
+func (s *SessionVars) GetRuntimeFilterTypes() []RuntimeFilterType {
+	return s.runtimeFilterTypes
+}
+
+// GetRuntimeFilterMode return the session variable runtimeFilterMode
+func (s *SessionVars) GetRuntimeFilterMode() RuntimeFilterMode {
+	return RFOff
```
AilinKid, line 3436: GET() function should just `return s.runtimeFilterMode` here?

sessionctx/variable/session.go
```diff
@@ -3413,3 +3420,118 @@ func (s *SessionVars) GetRelatedTableForMDL() *sync.Map {
...
+type RuntimeFilterType int64
+
+// In type of runtime filter, like "t.k1 in (?)"
+// MinMax type of runtime filter, like "t.k1 < ? and t.k1 > ?"
+const (
+	In RuntimeFilterType = iota
+	MinMax
+	// todo BloomFilter, bf/in
+)
+
```
AilinKid, line 3449: // function comment is necessary when its name starts with a Cap character.

sessionctx/variable/session.go
```diff
@@ -3413,3 +3420,118 @@ func (s *SessionVars) GetRelatedTableForMDL() *sync.Map {
...
+	case MinMax:
+		return "MIN_MAX"
+	default:
+		return ""
+	}
+}
+
+// RuntimeFilterTypeStringToType convert RuntimeFilterTypeNameString to RuntimeFilterType
+// If name is legal, it will return Runtime Filter Type and true
+// Elsa, it will return -1 and false
```
AilinKid, line 3463:
```suggestion
// Else, it will return -1 and false
```

sessionctx/variable/session.go
```diff
@@ -3413,3 +3420,118 @@ func (s *SessionVars) GetRelatedTableForMDL() *sync.Map {
...
+	case RFGlobal:
+		return "GLOBAL"
+	default:
+		return ""
+	}
+}
+
+// RuntimeFilterModeStringToMode convert RuntimeFilterModeString to RuntimeFilterMode
+// If name is legal, it will return Runtime Filter Mode and true
+// Elsa, it will return -1 and false
```
AilinKid, line 3525: ditto

sessionctx/variable/sysvar.go
```diff
@@ -144,7 +144,8 @@ var defaultSysVars = []*SysVar{
 	{Scope: ScopeSession, Name: TiDBReadStaleness, Value: strconv.Itoa(DefTiDBReadStaleness), Type: TypeInt, MinValue: math.MinInt32, MaxValue: 0, AllowEmpty: true, Hidden: false, SetSession: func(s *SessionVars, val string) error {
 		return setReadStaleness(s, val)
 	}},
-	{Scope: ScopeSession, Name: TiDBEnforceMPPExecution, Type: TypeBool, Value: BoolToOnOff(config.GetGlobalConfig().Performance.EnforceMPP), Validation: func(vars *SessionVars, normalizedValue string, originalValue string, scope ScopeFlag) (string, error) {
+	// todo changed
```
AilinKid, line 147:
```suggestion

```

util/id_generator.go
```diff
@@ -0,0 +1,27 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+// IDGenerator util class used for generate auto-increasing id
+type IDGenerator struct {
```
AilinKid, line 18: seems we can simulate what the planID did to avoid introducing another file that has the same functionality.
```
// PlanID is the unique id of logical and physical plan 
PlanID atomic.Int32
```

planner/core/resolve_indices.go
```diff
@@ -653,6 +653,15 @@ func (p *PhysicalApply) ResolveIndices() (err error) {
 	return
 }
 
+// ResolveIndices implements Plan interface.
+func (p *PhysicalTableScan) ResolveIndices() (err error) {
+	err = p.physicalSchemaProducer.ResolveIndices()
```
AilinKid, line 658:
```suggestion
return p.physicalSchemaProducer.ResolveIndices()
```

planner/core/plan_to_pb.go
```diff
@@ -556,6 +571,13 @@ func (p *PhysicalHashJoin) ToPB(ctx sessionctx.Context, storeType kv.StoreType)
 		probeFiledTypes = append(probeFiledTypes, ty)
 		buildFiledTypes = append(buildFiledTypes, ty)
 	}
+
+	// runtime filter
+	rfListPB, err := RuntimeFilterListToPB(p.runtimeFilterList, sc, client)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	// todo: arenatlx, push down hash join
```
AilinKid, line 580:
```suggestion
```

AilinKid (commented):
REST LGTM

AilinKid (commented):
REST LGTM

util/id_generator.go
elsa0520: We don't have a id util to get auto increasing id. Also the plan id is the unique id of plan but I need a runtimefilter id which is unique id of runtime filter. (There are many runtime filters in one plan)

planner/core/testdata/runtime_filter_generator_suite_out.json
elsa0520: Yes, we need to combine these two optimizations later. This is a todo

planner/core/runtime_filter_generator.go
elsa0520: Oh, this is a future code for physical selection. If the currentRFList is not empty, we need to consider whether adding a selection upper than table scan node

planner/core/runtime_filter_generator.go
```diff
@@ -0,0 +1,242 @@
...
+func (generator *RuntimeFilterGenerator) matchEQPredicate(eqPredicate *expression.ScalarFunction,
+	rightIsBuildSide bool) bool {
+	var targetColumn, srcColumn *expression.Column
+	if rightIsBuildSide {
+		targetColumn = eqPredicate.GetArgs()[0].(*expression.Column)
+		srcColumn = eqPredicate.GetArgs()[1].(*expression.Column)
+	} else {
+		targetColumn = eqPredicate.GetArgs()[1].(*expression.Column)
+		srcColumn = eqPredicate.GetArgs()[0].(*expression.Column)
```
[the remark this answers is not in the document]
elsa0520: Good comment! Fixed

planner/core/runtime_filter.go
```diff
@@ -0,0 +1,257 @@
...
+	rf.buildNode.runtimeFilterList = append(rf.buildNode.runtimeFilterList, rf)
+	rf.targetNode.runtimeFilterList = append(rf.targetNode.runtimeFilterList, rf)
+	logutil.BgLogger().Debug("Assign RF to target node",
+		zap.String("RuntimeFilter", rf.String()))
+}
+
+// ExplainInfo explain info of runtime filter
+func (rf *RuntimeFilter) ExplainInfo(isBuildNode bool) string {
+	var builder strings.Builder
+	fmt.Fprintf(&builder, "%d[%s]", rf.id, rf.rfType)
```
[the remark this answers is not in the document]
elsa0520: But I have actually commented that he is a runtimefilter in general. If a node has multiple runtime filters it will look like:
runtime filter: 0[IN] xxx, 1[IN] xxx

util/id_generator.go
AilinKid: I mean 
```
RuntimeFilterID atomic.Int32
usage like: ctx.GetSessionVars().RuntimeFilterID.Add(1)
```
it's ok, not a strong requirement

AilinKid (approved):
Better `make bazel_prepare` to resolve the conflicts

util/id_generator.go
elsa0520: Oh, This is indeed a method

qw4990 (approved):

PR #43033, opened 2023-04-13
statistics: remove statistics.Column.Count

xuyifangreeneyes:

Issue Number: ref #42160

Problem Summary:

Remove `statistics.Column.Count`. 

Before the PR, in order to maintain `statistics.Column.Count`, we need to read `mysql.stats_top_n` and `mysql.stats_buckets` for each column when init stats, which is time-consuming. On the other hand, the usage of `statistics.Column.Count` is limited. We modify `(*DataSource).getColumnNDV` to get rid of `statistics.Column.Count`.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/planner/core/casetest/integration_test.go
+++ b/planner/core/casetest/integration_test.go
@@ -3205,8 +3205,8 @@ func TestIssue32632(t *testing.T) {
 		"`S_ACCTBAL` decimal(15,2) NOT NULL," +
 		"`S_COMMENT` varchar(101) NOT NULL," +
 		"PRIMARY KEY (`S_SUPPKEY`) /*T![clustered_index] CLUSTERED */)")
-	tk.MustExec("analyze table partsupp;")
-	tk.MustExec("analyze table supplier;")
+	h := dom.StatsHandle()
+	require.NoError(t, h.HandleDDLEvent(<-h.DDLEventCh()))
 	tk.MustExec("set @@tidb_enforce_mpp = 1")
 
 	tbl1, err := dom.InfoSchema().TableByName(model.CIStr{O: "test", L: "test"}, model.CIStr{O: "partsupp", L: "partsupp"})
@@ -3217,7 +3217,6 @@ func TestIssue32632(t *testing.T) {
 	tbl1.Meta().TiFlashReplica = &model.TiFlashReplicaInfo{Count: 1, Available: true}
 	tbl2.Meta().TiFlashReplica = &model.TiFlashReplicaInfo{Count: 1, Available: true}
 
-	h := dom.StatsHandle()
 	statsTbl1 := h.GetTableStats(tbl1.Meta())
 	statsTbl1.RealtimeCount = 800000
 	statsTbl2 := h.GetTableStats(tbl2.Meta())
--- a/planner/core/stats.go
+++ b/planner/core/stats.go
@@ -159,34 +159,35 @@ func (p *baseLogicalPlan) DeriveStats(childStats []*property.StatsInfo, selfSche
 	return profile, nil
 }
 
+// getTotalRowCount returns the total row count, which is obtained when collecting colHist.
+func getTotalRowCount(statsTbl *statistics.Table, colHist *statistics.Column) int64 {
+	if colHist.IsFullLoad() {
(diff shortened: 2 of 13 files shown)


planner/core/stats.go
```diff
@@ -163,9 +163,32 @@ func (p *baseLogicalPlan) DeriveStats(childStats []*property.StatsInfo, selfSche
...
+						analyzeCount = int64(col.TotalRowCount())
+						break
+					}
+				}
+			}
+		}
+		if analyzeCount > 0 {
+			factor := float64(ds.statisticTable.RealtimeCount) / hist.TotalRowCount()
+			ndv *= factor
+		}
```
xuyifangreeneyes, line 191: Here the way to calculate column NDV is changed. We assume that the change is ok for most cases.

planner/core/casetest/integration_test.go
```diff
@@ -3205,8 +3205,8 @@ func TestIssue32632(t *testing.T) {
 		"`S_ACCTBAL` decimal(15,2) NOT NULL," +
 		"`S_COMMENT` varchar(101) NOT NULL," +
 		"PRIMARY KEY (`S_SUPPKEY`) /*T![clustered_index] CLUSTERED */)")
-	tk.MustExec("analyze table partsupp;")
-	tk.MustExec("analyze table supplier;")
```
xuyifangreeneyes, line 3209: If I don't modify the test, I would get the error as follows:
```
Diff:
--- Expected
+++ Actual
@@ -4,4 +4,4 @@
 [    └─HashAgg 1.00 mpp[tiflash]  funcs:sum(test.partsupp.ps_supplycost)->Column#15]
-[      └─Projection 12500.00 mpp[tiflash]  test.partsupp.ps_supplycost]
-[        └─HashJoin 12500.00 mpp[tiflash]  inner join, equal:[eq(test.supplier.s_suppkey, test.partsupp.ps_suppkey)]]
+[      └─Projection 8000000000.00 mpp[tiflash]  test.partsupp.ps_supplycost]
+[        └─HashJoin 8000000000.00 mpp[tiflash]  inner join, equal:[eq(test.supplier.s_suppkey, test.partsupp.ps_suppkey)]]
 [          ├─ExchangeReceiver(Build) 10000.00 mpp[tiflash]  ]
```
Before the PR, in `getColumnNDV`, since `hist.Count` is 0, we use `float64(ds.statisticTable.RealtimeCount) * distinctFactor` to calculate NDV. After the PR, if we still analyze `partsupp` and `supplier` when they don't have any data, since `hist.IsStatsInitialized()` is true, we use `float64(hist.Histogram.NDV)` to calculate NDV. In another word, before the PR we use `hist.Count > 0` to decide whether the table is analyzed while after the PR we use `float64(hist.Histogram.NDV)` to decide that. I think the latter is more reasonable. I remove the two analyze commands to keep cardinality estimation the same as before. Note that it is rare in the reality that `RealtimeCount` reaches 800000/10000 but auto analyze is not triggered.

planner/core/stats.go
```diff
@@ -163,9 +163,32 @@ func (p *baseLogicalPlan) DeriveStats(childStats []*property.StatsInfo, selfSche
...
+				for _, col := range ds.statisticTable.Columns {
+					if col.IsFullLoad() && col.LastUpdateVersion == hist.LastUpdateVersion {
+						analyzeCount = int64(col.TotalRowCount())
+						break
+					}
+				}
+			}
+		}
+		if analyzeCount > 0 {
+			factor := float64(ds.statisticTable.RealtimeCount) / hist.TotalRowCount()
```
time-and-fate, line 189: Is it possible that `hist.TotalRowCount()` is 0 here?

xuyifangreeneyes: I made a mistake. Here we should use `analyzeCount`.

xuyifangreeneyes: Updated.

time-and-fate (approved):

PR #43077, opened 2023-04-14
planner, statistics: introduce lite init stats

xuyifangreeneyes:

Issue Number: ref #42160

Problem Summary:

1. Introduce `InitStatsLite`, which only reads `mysql.stats_meta` and `mysql.stats_histograms`. Index/column stats such as cmsketch/topn/buckets would not be loaded when `InitStatsLite`.
2. Modify the logical of updating stats(`indexStatsFromStorage` and `columnStatsFromStorage`). If we update stats for one index or one primary key, we also need to first whether they are loaded before. If none of their cmsketch/topn/buckets is loaded, we don't need to read the new version of their cmsketch/topn/buckets into memory.
 3. Introduce config `lite-init-stats` and use it to control the logic of init stats and update stats. When `lite-init-stats` is false, the logic of init stats and update stats is the same as before. By default `lite-init-stats` is false.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


config/config.go
```diff
@@ -717,6 +717,8 @@ type Performance struct {
 	MemoryUsageAlarmRatio float64 `toml:"memory-usage-alarm-ratio" json:"memory-usage-alarm-ratio"`
 
 	EnableLoadFMSketch bool `toml:"enable-load-fmsketch" json:"enable-load-fmsketch"`
+
+	LiteInitStats bool `toml:"lite-init-stats" json:"lite-init-stats"`
```
qw4990, line 721: Can we use a SQL variable instead of a config item here? SQL variables are more flexible and east-to-use.

statistics/handle/bootstrap.go
```diff
@@ -362,6 +456,36 @@ func (h *Handle) initStatsBuckets(cache *statsCache) error {
 	return nil
 }
 
+func (h *Handle) InitStatsLite(is infoschema.InfoSchema) (err error) {
```
qw4990, line 459: Please add some comments to explain the differences between `InitStatsLite` and `InitStats`.

qw4990: Or can we merge these 2 code paths into one, e.g. `IndexStats(..., lite bool)`, `initStatsHistograms(..., light bool)`

config/config.go
xuyifangreeneyes: 1. `LiteInitStats` only takes effect when one tidb-server starts. If you change its value but no tidb-server restarts, nothing takes effect. We don't need the flexibility of changing its value online.
2. Making `LiteInitStats` a configuration can provide instance-level config. For example, you can make one tidb-server's `lite-init-stats` true and the other false to observe their different behaviors, which is more elegant than using a global variable to archive that.

statistics/handle/bootstrap.go
xuyifangreeneyes: Updated.

statistics/interact_with_storage.go
```diff
@@ -236,6 +236,24 @@ func indexStatsFromStorage(reader *StatsReader, row chunk.Row, table *Table, tab
 		if histID != idxInfo.ID {
 			continue
 		}
+		notNeedLoad := lease > 0 &&
```
qw4990, line 239: Why add `lease > 0` here? Better to add some comments for this check.

qw4990 (approved):

statistics/interact_with_storage.go
xuyifangreeneyes: Added.

xuyifangreeneyes: `lease > 0` is aligned to `notNeedLoad` conditions of `columnStatsFromStorage`. In UT, `lease` is 0 and all column/index stats are fully loaded.

planner/core/plan_stats.go
```diff
@@ -135,9 +135,13 @@ func collectSyncIndices(ctx sessionctx.Context, histNeededColumns []model.TableI
...
 		}
-		tbl, ok := ctx.GetDomainInfoSchema().(infoschema.InfoSchema).TableByID(column.TableID)
+		is := ctx.GetDomainInfoSchema().(infoschema.InfoSchema)
+		tbl, ok := is.TableByID(column.TableID)
 		if !ok {
-			continue
+			tbl, _, _ = is.FindTableByPartitionID(column.TableID)
+			if tbl == nil {
+				continue
+			}
```
xuyifangreeneyes, line 144: Index stats sync loading has a small bug. I fixed it here.

winoros (approved):

PR #43141, opened 2023-04-18
util: reimplement row level checksum utilities

### LINKED ISSUE #42747: Implement end to end checksum for TiDB and TiCDC
## Background

TiCDC is an important component for TiDB to synchronize data to various downstream systems. When synchronizing data to downstream systems, data integrity is especially important. However, TiCDC does not support end-to-end data integrity verification yet.

## Spec

Provide below cluster level boolean type option in TiDB side. 
```
tidb_enable_row_level_checksum = [true|false]  # the default value is false.
SET GLOBAL tidb_enable_row_level_checksum = true;  
```

After the customer enables this option, every data change for a row in non-system databases will append an invisible field that is used to store a computed checksum value based on the content of the row. This invisible field is just for data correctness checking purposes and is transparent to the customer. 

### Development tracking for the TiDB part

- [ ] Add a global session variable `tidb_enable_row_level_checksum` to enable or disable the checksum calculation when inserting new rows.
- [ ] Add a checksum function `tidb_row_checksum` to return the checksum value of a row.
- [ ] Support writing rows with checksum values 
    - [ ] Calculate the row checksum in the `tablecodec` package when `EncodeRow` function is used. Calculate the CRC XOR result for each column when executing `encodeRowCols`, a checksum result is returned finally.
    - [ ] Make it work with the DDL `add column` schema change, and generate two checksum values if necessary.
    - [ ] Make it work with the DDL `drop column` schema change, and generate two checksum values if necessary.
    - [ ] Make it work with the DDL `modify column` schema change, and generate two checksum values if necessary.
    - [ ] Append the checksum header and checksum result information to the encoded row according to the extended row format protocol.
- [ ] Keep the read request processing compatibility. 
    - [ ] Skip the checksum part processing for `chunckDecoder` in tidb if necessary. 
    - [ ] Skip the checksum part processing in `internal_handle_request` and `PointGetter` for chunk encoding processing read requests if necessary.
    - [ ] Skip the checksum part processing in tiflash if necessary.
- [ ] Compatibility tests, the checksum extended part should not impact the backward compatibility, and downgrade is supported when the checksum row format is used.

zyguan:

Issue Number: ref #42747

Problem Summary:

We have implemented row level checksum utilities in #42859. However, it doesn't match the requirements for e2e checksum verification, because downstream consumers cannot calculate the checksum as tidb does. So, we reimplement checksum algorithm in this PR.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

The overhead of checksum calculation shall be fine.
```
goos: linux
goarch: amd64
pkg: github.com/pingcap/tidb/util/rowcodec
cpu: AMD Ryzen 7 PRO 4750U with Radeon Graphics
BenchmarkChecksum
BenchmarkChecksum-16                    16851555                68.72 ns/op            0 B/op          0 allocs/op
BenchmarkEncode
BenchmarkEncode-16                      15039044                82.57 ns/op            0 B/op          0 allocs/op
```

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/util/rowcodec/common.go
+++ b/util/rowcodec/common.go
@@ -16,12 +16,16 @@ package rowcodec
 
 import (
 	"encoding/binary"
+	"hash/crc32"
+	"math"
 	"reflect"
 	"unsafe"
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/tidb/parser/model"
+	"github.com/pingcap/tidb/parser/mysql"
 	"github.com/pingcap/tidb/parser/types"
+	data "github.com/pingcap/tidb/types"
 )
 
 // CodecVer is the constant number that represent the new row format.
@@ -30,6 +34,7 @@ const CodecVer = 128
 var (
 	errInvalidCodecVer    = errors.New("invalid codec version")
 	errInvalidChecksumVer = errors.New("invalid checksum version")
+	errInvalidChecksumTyp = errors.New("invalid type for checksum")
 )
 
 // First byte in the encoded value which specifies the encoding type.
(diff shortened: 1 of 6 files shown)


util/rowcodec/common.go
```diff
@@ -240,3 +245,109 @@ func IsNewFormat(rowData []byte) bool {
...
+}
+
+// Len implements sort.Interface for RowData.
+func (r RowData) Len() int { return len(r.Cols) }
+
+// Less implements sort.Interface for RowData.
+func (r RowData) Less(i int, j int) bool { return r.Cols[i].ID < r.Cols[j].ID }
+
+// Swap implements sort.Interface for RowData.
+func (r RowData) Swap(i int, j int) { r.Cols[i], r.Cols[j] = r.Cols[j], r.Cols[i] }
```
3AceShowHand, line 273: It looks the `Data` is not sorted

util/rowcodec/common.go
```diff
@@ -240,3 +245,109 @@ func IsNewFormat(rowData []byte) bool {
...
+
+// Encode encodes all columns into bytes (for test purpose).
```
3AceShowHand, line 275: Is this only used for testing purposes?

zyguan: By now, yes.

util/rowcodec/common.go
zyguan: `Data` here is just a buffer for encoding column, so no need to be sorted. It's named `Data` because `RowData.Encode` writes encoded data to it (so user can also access the result from `rd.Data`).

util/rowcodec/common.go
```diff
@@ -240,3 +245,109 @@ func IsNewFormat(rowData []byte) bool {
...
+		r.Data, err = col.Encode(r.Data)
+		if err != nil {
+			return 0, err
+		}
+		checksum = crc32.Update(checksum, crc32.IEEETable, r.Data)
+	}
+	return checksum, nil
+}
+
+func appendDatumForChecksum(buf []byte, dat *data.Datum, typ byte) (out []byte, err error) {
```
3AceShowHand, line 305: It looks for all columns, concatenate the bytes together, does not consider the bytes length.

cfzjywxk (approved):
LGTM

PR #43191, opened 2023-04-19
planner, sessionctx: add `tidb_opt_fix_control` variable

### LINKED ISSUE #43169: Introduce basic compatibility control mechanism to optimizer
### Background

- We already have quite a few variables to control the behavior of optimizers. 
- In some circumstances, the change of optimizer behavior doesn't apply to most scenarios. Appending new variables is painful. 
- In database upgrading, making execution plans stable for a few conservative customers is not possible. 

### Design

Add a new session/global level system variable `tidb_opt_fix_control`.

Its value is a string. The default value is an empty string. A valid value must be in the format of "`<#issue1>:<value1>,<#issue2>:<value2>...`", like `set @@tidb_opt_fix_control = '99999:"on"'`.

This variable provides a list of "fix control", which is for controlling a specific behavior of the optimizer, and can be set to a value. A `<#issueN>:<value>` is for setting a fix control. The `<#issueN>` in it is an issue number that corresponds to a GitHub issue: `https://github.com/pingcap/tidb/issues/<#issueN>`, descriptions related to this fix control should be placed in this issue.

### Test

- Set the system variable to a valid format, the content can be correctly recognized in tidb.
- Set the system variable to a invalid format, tidb will return an error and the internal state of tidb is not updated.
- Set the same "fix control" multiple times, tidb will return a warning and only the last one takes effect.

time-and-fate:

Issue Number: close #43169

- Add a new system variable `tidb_opt_fix_control`
- Add the corresponding field of type `map[uint64]string` to `SessionVars` 

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/sessionctx/variable/session.go
+++ b/sessionctx/variable/session.go
@@ -1423,6 +1423,9 @@ type SessionVars struct {
 	// If there exists an index whose estimated selectivity is smaller than this threshold, the optimizer won't
 	// use the ExpectedCnt to adjust the estimated row count for index scan.
 	OptOrderingIdxSelThresh float64
+
+	// OptimizerFixControl control some details of the optimizer behavior through the tidb_opt_fix_control variable.
+	OptimizerFixControl map[uint64]string
 }
 
 // planReplayerSessionFinishedTaskKeyLen is used to control the max size for the finished plan replayer task key in session
--- a/planner/core/casetest/integration_test.go
+++ b/planner/core/casetest/integration_test.go
@@ -3473,3 +3473,41 @@ func TestIndexJoinRangeFallback(t *testing.T) {
 		}
 	}
 }
+
+func TestFixControl(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test;")
+	s := tk.Session()
+	var input []string
+	var output []struct {
+		SQL        string
+		FixControl map[uint64]string
+		Error      string
+		Warnings   [][]interface{}
+		Variable   []string
+	}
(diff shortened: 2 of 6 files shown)


sessionctx/variable/session.go
```diff
@@ -1423,6 +1423,9 @@ type SessionVars struct {
 	// If there exists an index whose estimated selectivity is smaller than this threshold, the optimizer won't
 	// use the ExpectedCnt to adjust the estimated row count for index scan.
 	OptOrderingIdxSelThresh float64
+
+	// OptimizerFixControl control some details of the optimizer behavior through the tidb_opt_fix_control variable.
```
elsa0520, line 1427: Please give the example of this session variables and explain what is this session variables used for

sessionctx/variable/session.go
```diff
@@ -1423,6 +1423,9 @@ type SessionVars struct {
...
+	OptimizerFixControl map[uint64]string
```
elsa0520, line 1428: How does this variable affect the optimizer? It doesn't seem to be reflected in this PR

sessionctx/variable/session.go
time-and-fate: The expected usage is in the issue. You can have a look.

sessionctx/variable/session.go
time-and-fate: We only introduce this variable now, so we can rely on this in some future bug fixes or behavior changes that might cause regression.

qw4990 (approved):

chrysan (approved):

Issue #43227, opened 2023-04-20
Ignore lossy DDL change for TiCDC


hi-rustin:
### What is lossy DDL and why do we need to special handle it?
Lossy DDL means we execute a DDL which changes the data type. For example:
Create table:
```sql
CREATE TABLE movies(title VARCHAR(50) NOT NULL,
genre VARCHAR(30) NOT NULL,director VARCHAR(60) NOT NULL,
release_year INT NOT NULL,PRIMARY KEY(title));
```
Notice the current length of the genre is `VARCHAR(30)`. If you do an alter:
```sql
ALTER TABLE movies MODIFY COLUMN genre VARCHAR(20)
```
You changed the length from 30 to 20. This is a lossy DDL, it would cause the full amount of data to change.

Also, it causes the TiCDC to spend a lot of resources on handling these data changes from TiCDC. But maybe the users don't care about these changes. They only want to get the right DDL to notice that the data type of this column was changed.

See more at: [Needs Analysis For Optimizing Lossy DDL Synchronization Process ](https://pingcap.feishu.cn/docx/GW4ddnz39oV9dyxqT9Rc2YRanDg?302from=wiki) and [Solution For Optimizing Lossy DDL in TiCDC](https://pingcap.feishu.cn/docx/XeOBdTt7ro8aBhxBP8Ucp8qKnZf?302from=wiki)

PR #43232, opened 2023-04-20
ddl: set lossy DDL change source for column worker

hi-rustin:

Issue Number: close https://github.com/pingcap/tidb/issues/43227

Problem Summary:

set lossy DDL change source for column worker
See more at the issue.

### Check List

Tests 

- [x] Unit test

Side effects

None

Documentation

None

### Release note

```release-note
None
```


kv/option.go
```diff
@@ -96,6 +96,14 @@ const (
...
 	ScanBatchSize
 	// TxnSource set the source of this transaction.
+	// For now, it can represent the transaction is created by a TiCDC or Lossy DDL reorg Backfill job.
+	// 1. Reserve 1 - 15 for TiCDC to implement BDR synchronization.
+	// 2. Reserve 1024 - `To Be Determined` for Lossy DDL Backfill job.
+	//    For now, we only use 1024 for column reorg backfill job.
+	// The reason for having these spans is to keep some room for future expansion of it.
+	// We only can set one source for a transaction, so please make sure the different features which
+	// use this option don't conflict with each other.
+	// Or find a way to set multiple sources for a transaction in an uint64.
```
hi-rustin, line 106: Do you think we need to redesign it?
If we want to add more sources in the future, some sources may conflict with other sources.
For example, if we add a new feature, it depends on a new 2048-2060 range. At the same time, we have a flag for TiKV to indicate what TiKV should do when it gets this kind of txn.
If users also enable the BDR feature, then it conflicts. Because we can only set one `txn_option` for one txn.

So I suggest we use bits to represent different sources.
For example:
```go
package main

import "fmt"

const (
	stateABits  = 8 // Number of bits used for state A
	stateBBits  = 4 // Number of bits used for state B
	stateCBits  = 4 // Number of bits used for state C
	stateAMax   = (1 << stateABits) - 1
	stateBMax   = (1 << stateBBits) - 1
	stateCMax   = (1 << stateCBits) - 1
	stateBShift = stateABits
	stateCShift = stateABits + stateBBits
)

func main() {
	var flags uint64 = 0 // Initial value is 0

	// Set state A to 7
	setStateA(&flags, 7)

	// Set state B to 11
	setStateB(&flags, 11)

	// Set state C to 3
	setStateC(&flags, 3)

	// Check if state A, B, and C are set
	if isStateASet(flags) {
		fmt.Println("State A is set")
	}
	if isStateBSet(flags) {
		fmt.Println("State B is set")
	}
	if isStateCSet(flags) {
		fmt.Println("State C is set")
	}

	// Get the values of state A, B, and C
	a := getStateA(flags)
	b := getStateB(flags)
	c := getStateC(flags)
	fmt.Printf("State A: %d, State B: %d, State C: %d\n", a, b, c)
}

// Set the value of state A
func setStateA(flags *uint64, value uint64) {
	if value > stateAMax {
		value = stateAMax
	}
	*flags |= value
}

// Set the value of state B
func setStateB(flags *uint64, value uint64) {
	if value > stateBMax {
		value = stateBMax
	}
	*flags |= value << stateBShift
}

// Set the value of state C
func setStateC(flags *uint64, value uint64) {
	if value > stateCMax {
		value = stateCMax
	}
	*flags |= value << stateCShift
}

// Check if state A is set
func isStateASet(flags uint64) bool {
	return (flags & stateAMax) != 0
}

// Check if state B is set
func isStateBSet(flags uint64) bool {
	return (flags & (stateBMax << stateBShift)) != 0
}

// Check if state C is set
func isStateCSet(flags uint64) bool {
	return (flags & (stateCMax << stateCShift)) != 0
}

// Get the value of state A
func getStateA(flags uint64) uint64 {
	return flags & stateAMax
}

// Get the value of state B
func getStateB(flags uint64) uint64 {
	return (flags >> stateBShift) & stateBMax
}

// Get the value of state C
func getStateC(flags uint64) uint64 {
	return (flags >> stateCShift) & stateCMax
}
```

Then we can use A for BDR, B for lossy DDL.
Check compatibility:
1. Old TiDB will set txn_option to 1-15.
2. Old TiKV will drop the event if txn_option is not 0.
3. TiUP will update tikv first, so we need to make sure that new TiKV can handle the txn_option set by old TiDB.
4. The new TiKV will get the first 8 bits for the BDR function, because the old TiDB only allows to set 1-15 to txn_option. So it is compatible.

overvenus: LGTM

hi-rustin (commented):
Self check

kv/option.go
```diff
@@ -193,3 +197,59 @@ const (
...
+	// TiCDC uses 1-255 to indicate the change from a lossy DDL reorg Backfill job.
+	// For now, we only use 1 for column reorg backfill job.
+	lossyDDLReorgSourceBits   = 8
+	LossyDDLColumnReorgSource = 1
+	lossyDDLReorgSourceMax    = (1 << lossyDDLReorgSourceBits) - 1
+	lossyDDLReorgSourceShift  = cdcWriteSourceBits
+)
+
+// SetCdcWriteSource sets the TiCDC write source in the txnSource.
+func SetCdcWriteSource(txnSource *uint64, value uint64) error {
```
zimulala, line 220: s/SetCdcWriteSource/SetCDCWriteSource?

kv/option.go
```diff
@@ -193,3 +197,59 @@ const (
...
+	if value > cdcWriteSourceBits {
+		return errors.Errorf("value %d is out of TiCDC write source range, should be in [1, 15]",
+			value)
+	}
+	*txnSource |= value
+
+	return nil
+}
+
+func getCdcWriteSource(txnSource uint64) uint64 {
```
zimulala, line 230: Ditto.

kv/option_test.go
```diff
@@ -0,0 +1,104 @@
...
+		name                        string
+		currentSource               uint64
+		lossyDDLReorgSource         uint64
+		expectedSet                 bool
+		expectedLossyDDLReorgSource uint64
+		expectedError               string
+	}{
+		{
+			name:                        "lossy ddl reorg source is set",
+			currentSource:               12, // SetCdcWriteSource
```
zimulala, line 74: Why here is 12?

ddl/column.go
```diff
@@ -1385,6 +1385,20 @@ func (w *updateColumnWorker) BackfillData(handleRange reorgBackfillTask) (taskCt
...
+		// avoid TiCDC to replicate this kind of change.
+		var txnSource uint64
+		if val := txn.GetOption(kv.TxnSource); val != nil {
+			txnSource, _ = val.(uint64)
+		}
+		err := kv.SetLossyDDLReorgSource(&txnSource, kv.LossyDDLColumnReorgSource)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		txn.SetOption(kv.TxnSource, txnSource)
```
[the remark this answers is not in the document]
hi-rustin: I guess we can keep it. Then we do not need to pass the entire txn to the util function. It makes the util function simpler and easy to test.

session/session.go
```diff
@@ -657,7 +657,21 @@ func (s *session) doCommit(ctx context.Context) error {
...
+	}
+	// If the transaction is started by CDC, we need to set the CDCWriteSource option.
+	if sessVars.CDCWriteSource != 0 {
+		err := kv.SetCdcWriteSource(&txnSource, sessVars.CDCWriteSource)
+		if err != nil {
+			return errors.Trace(err)
+		}
+
+		s.txn.SetOption(kv.TxnSource, txnSource)
+	}
```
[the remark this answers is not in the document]
hi-rustin: See https://github.com/pingcap/tidb/pull/43232#discussion_r1176157023

kv/option.go
hi-rustin: Good catch! Thanks!

kv/option_test.go
```diff
@@ -0,0 +1,104 @@
...
+		name                        string
+		currentSource               uint64
+		lossyDDLReorgSource         uint64
+		expectedSet                 bool
+		expectedLossyDDLReorgSource uint64
+		expectedError               string
+	}{
+		{
+			name:                        "lossy ddl reorg source is set",
+			currentSource:               12, // SetCDCWriteSource
```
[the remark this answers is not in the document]
hi-rustin: 0 means nothing. But the goal of this test is to test whether we can set two sources into the txn_source option.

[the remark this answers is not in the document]
zimulala: I mean could we add this test？And is this allowed?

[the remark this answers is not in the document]
hi-rustin: Oh, added. Thanks!

zimulala (approved):
LGTM

lichunzhu (approved):
LGTM

Issue #42912, opened 2023-04-11
Unexpected result of subquery


DerZc:
### 1. Minimal reproduce step (Required)

Consider the following program:

```
CREATE TABLE t0(c0 TEXT(328) );

CREATE VIEW v0(c0) AS SELECT 'c' FROM t0;

INSERT INTO t0 VALUES (-12);

SELECT v0.c0 AS c0 FROM  v0 NATURAL RIGHT JOIN t0  WHERE (1 !=((v0.c0)REGEXP(-7)));  -- empty

SELECT COUNT(v0.c0) AS c0 FROM v0 WHERE EXISTS(SELECT v0.c0 AS c0 FROM v0 NATURAL RIGHT JOIN t0  WHERE (1 !=((v0.c0)REGEXP(-7))));  -- 1
```
The first query produce empty result, I put the first query in an `EXISTS` expression, which should be equals to `FALSE`, but the second query produce `1`.

### 2. What did you expect to see? (Required)
The second query produce `0`.
I run this query in MySQL, and it produce `0`.  https://www.db-fiddle.com/f/wNt1byF7hj6NXEDKDEsp4B/0

### 3. What did you see instead (Required)
The second query produce `1`.

### 4. What is your TiDB version? (Required)
https://github.com/pingcap/tidb/commit/c233969b2c385b4292a7caaf16517dd8b152d7f1

<img width="720" alt="image" src="https://user-images.githubusercontent.com/34330257/231064543-379ed2db-b2f4-4242-9b69-fb09222bed6d.png">

zanmato1984:
This is an alias resolution issue.

If an alias, e.g. `v1` is assigned to the top-level `v0` to disambiguate the alias resolution, the plan will be:
```
mysql> explain SELECT COUNT(v1.c0) AS c0 FROM v0 v1 WHERE EXISTS(SELECT v0.c0 AS c0 FROM v0 NATURAL RIGHT JOIN t0  WHERE (1 !=((v0.c0)REGEXP(-7))));
+--------------------+---------+------+---------------+----------------------------------+
| id                 | estRows | task | access object | operator info                    |
+--------------------+---------+------+---------------+----------------------------------+
| StreamAgg_28       | 1.00    | root |               | funcs:count(Column#3)->Column#10 |
| └─TableDual_30     | 0.00    | root |               | rows:0                           |
+--------------------+---------+------+---------------+----------------------------------+
2 rows in set (0.01 sec)
```
[...]

PR #43247, opened 2023-04-20
planner: fix the col resolution priority between outer scope and nature join folded col

AilinKid:

Issue Number: close https://github.com/pingcap/tidb/issues/42912

Problem Summary:
a traditional name resolution priority problem as we talked before

image

for the cases, when building a where clause `(1 !=((v0.c0)REGEXP(-7)))`, the resolution flow is: 
1. find the col in current p (built from natural join, only t0.c0 is remained, v0.c0 is folded)
2. find the col in the outer layer scope, (**for case 1, nothing here, find the col in step 3 (non-correlated); for case 2, find the col here (outer scope has exactly the same name v0.c0, Hops!) and return immediately (causing correlated)**)
3. find the col in special cases, for example here, the current layer's nature-join col name folding

the sequence is not that right, we can simply move 3 forward 2 to make this right.

PS. Every outer scope layer should also think about nature-join folding, we didn't think about that as yet.

ref the last comment from issue.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
fix the col resolution priority between outer scope and nature join folded col
```

### DIFF
--- a/planner/core/expression_rewriter.go
+++ b/planner/core/expression_rewriter.go
@@ -2024,6 +2024,18 @@ func (er *expressionRewriter) toColumn(v *ast.ColumnName) {
 		er.ctxStackAppend(column, er.names[idx])
 		return
 	}
+	if _, ok := er.p.(*LogicalUnionAll); ok && v.Table.O != "" {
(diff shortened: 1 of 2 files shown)


fixdb (approved):
+1

winoros (approved):

Issue #43212, opened 2023-04-19
resource_control supports calibrate resource 


nolouch:
## Feature Request

Add new statement `calibrate resource` to estimate the total Request-Units(RU) of the current cluster.
Because the total ru usage is related to workload resource consumption, so the maximum RU can be different with the different workloads. Thus, the maximum RU estimated by this PR is based on a given workload -- TPC-C, and we may support another workload (e.g. sysbench) in the future.

In general, the bottle of a cluster can be one of TiDB CPU, TiKV CPU, and TiKV IO Bandwidth. Currently, we can get the exact IO bandwidth and for most workloads, it is unlikely to be the bottleneck. So here, we only consider TiDB CPU or TiKV CPU as the bottleneck.

And also we consider needing to make the calibrate according to the actual workload to make it more accurate.

### Issuse
-  [x] Static calibrate base on tpcc workload: https://github.com/pingcap/tidb/pull/42165
-  [x] Static calibrate base on sysbench workload: https://github.com/pingcap/tidb/pull/42695
-  [x] Dynamic calibrate base on history workload: https://github.com/pingcap/tidb/pull/43098

PR #43248, opened 2023-04-20
resource_control: make metrics time point match in dynamic calibrate

CabinfeverB:

Issue Number: ref #43212

Problem Summary:
When the value is 0, the corresponding metrics does not return. If `RU` metrics is empty but `CPU` metrics is not (for example, when doing backup or restore), this logic will now cause `RU` and CPU` to be misaligned

Exactly match ru, tidb, and tikv metrics according to metrics timestamps

### Check List

Tests 

- [x] Unit test

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/executor/calibrate_resource.go
+++ b/executor/calibrate_resource.go
@@ -355,6 +355,8 @@ func (t *timeSeriesValues) getValue() float64 {
 
 func (t *timeSeriesValues) advance(target time.Time) bool {
 	for ; t.idx < len(t.vals); t.idx++ {
+		// `target` is maximal time in other timeSeriesValues,
+		// so we should find the time which offset is less than 10s.
 		if t.vals[t.idx].tp.Add(time.Second * 10).After(target) {
 			return t.vals[t.idx].tp.Add(-time.Second * 10).Before(target)
 		}
--- a/executor/calibrate_resource.go
+++ b/executor/calibrate_resource.go
@@ -238,7 +238,7 @@ func (e *calibrateResourceExec) dynamicCalibrate(ctx context.Context, req *chunk
 		if tidbCPUs.getTime().Before(minTime) {
 			minTime = tidbCPUs.getTime()
 		}
-		if !rus.findTime(minTime) || !tikvCPUs.findTime(minTime) || !tidbCPUs.findTime(minTime) {
+		if !rus.advance(minTime) || !tikvCPUs.advance(minTime) || !tidbCPUs.advance(minTime) {
 			break
 		}
 		tikvQuota, tidbQuota := tikvCPUs.getValue()/totalKVCPUQuota, tidbCPUs.getValue()/totalTiDBCPU
@@ -353,7 +353,7 @@ func (t *timeSeriesValues) getValue() float64 {
 	return t.vals[t.idx].val
 }
 
-func (t *timeSeriesValues) findTime(target time.Time) bool {
+func (t *timeSeriesValues) advance(target time.Time) bool {
 	for ; t.idx < len(t.vals); t.idx++ {
 		if t.vals[t.idx].tp.Add(-time.Second * 10).Before(target) {
 			return true
--- a/executor/calibrate_resource.go
+++ b/executor/calibrate_resource.go
@@ -212,15 +212,15 @@ func (e *calibrateResourceExec) dynamicCalibrate(ctx context.Context, req *chunk
 	if err != nil {
 		return err
 	}
-	rus, err := getRUPerSec(e.ctx, ctx, exec, startTime, endTime)
+	rus, err := getRUPerSec(ctx, e.ctx, exec, startTime, endTime)
 	if err != nil {
 		return err
 	}
-	tikvCPUs, err := getComponentCPUUsagePerSec(e.ctx, ctx, exec, "tikv", startTime, endTime)
+	tikvCPUs, err := getComponentCPUUsagePerSec(ctx, e.ctx, exec, "tikv", startTime, endTime)
 	if err != nil {
 		return err
 	}
-	tidbCPUs, err := getComponentCPUUsagePerSec(e.ctx, ctx, exec, "tidb", startTime, endTime)
+	tidbCPUs, err := getComponentCPUUsagePerSec(ctx, e.ctx, exec, "tidb", startTime, endTime)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


executor/calibrate_resource.go
```diff
@@ -334,17 +384,22 @@ func getNumberFromMetrics(ctx context.Context, exec sqlexec.RestrictedSQLExecuto
...
 		return nil, errors.Errorf("metrics '%s' is empty", metrics)
 	}
-	ret := make([]float64, 0, len(rows))
+	ret := make([]*timePointValue, 0, len(rows))
 	for _, row := range rows {
-		ret = append(ret, row.GetFloat64(0))
+		if tp, err := row.GetTime(0).AdjustedGoTime(sctx.GetSessionVars().Location()); err == nil {
+			ret = append(ret, &timePointValue{
+				tp:  tp,
+				val: row.GetFloat64(1),
```
nolouch, line 400: It was wrong before, why is it 1？

executor/calibrate_resource_test.go
```diff
@@ -236,13 +238,147 @@ func TestCalibrateResource(t *testing.T) {
...
+	mockData["process_cpu_usage"] = cpu2
 
 	tk.MustQueryWithContext(ctx, "CALIBRATE RESOURCE START_TIME '2020-02-12 10:35:00' DURATION '10m'").Check(testkit.Rows("5616"))
 	tk.MustQueryWithContext(ctx, "CALIBRATE RESOURCE START_TIME '2020-02-12 10:35:00' END_TIME '2020-02-12 10:45:00'").Check(testkit.Rows("5616"))
 	tk.MustQueryWithContext(ctx, "CALIBRATE RESOURCE START_TIME '2020-02-12 10:35:00' DURATION '10m'").Check(testkit.Rows("5616"))
 	tk.MustQueryWithContext(ctx, "CALIBRATE RESOURCE END_TIME '2020-02-12 10:45:00' START_TIME '2020-02-12 10:35:00'").Check(testkit.Rows("5616"))
 	tk.MustQueryWithContext(ctx, "CALIBRATE RESOURCE END_TIME '2020-02-12 10:45:00' DURATION '5m' START_TIME '2020-02-12 10:35:00' ").Check(testkit.Rows("5616"))
 
+	// Statistical time points do not correspond
+	ruModify1 := [][]types.Datum{
```
nolouch, line 250: maybe add a Offset like +10secs for the datatime to test the case of misalignment.

nolouch (commented):
rest lgtm

executor/calibrate_resource.go
glorv: The expression is changed, the original expr only returns the value, the new expr returns the time and value

executor/calibrate_resource.go
```diff
@@ -312,14 +327,49 @@ func getTiDBTotalCPUQuota(ctx context.Context, exec sqlexec.RestrictedSQLExecuto
...
+
+func (t *timeSeriesValues) getTime() time.Time {
+	return t.vals[t.idx].tp
+}
+
+func (t *timeSeriesValues) getValue() float64 {
+	return t.vals[t.idx].val
+}
+
+func (t *timeSeriesValues) findTime(target time.Time) bool {
```
glorv, line 356:
```suggestion
func (t *timeSeriesValues) advance(target time.Time) bool {
```

executor/calibrate_resource.go
```diff
@@ -312,14 +327,49 @@ func getTiDBTotalCPUQuota(ctx context.Context, exec sqlexec.RestrictedSQLExecuto
...
+	for ; t.idx < len(t.vals); t.idx++ {
+		if t.vals[t.idx].tp.Add(-time.Second * 10).Before(target) {
+			return true
+		}
```
glorv, line 360:
```suggestion
		if t.vals[t.idx].tp.Add(-time.Second * 10).Before(target) {
			return t.vals[t.idx].tp.Add(time.Second * 10).After(target)
		}
```

executor/calibrate_resource.go
```diff
@@ -211,34 +212,48 @@ func (e *calibrateResourceExec) dynamicCalibrate(ctx context.Context, req *chunk
...
-		tikvQuota, tidbQuota := tikvCPUs[idx]/totalKVCPUQuota, tidbCPUs[idx]/totalTiDBCPU
+		// make time point match
+		minTime := rus.getTime()
+		if tikvCPUs.getTime().Before(minTime) {
+			minTime = tikvCPUs.getTime()
+		}
+		if tidbCPUs.getTime().Before(minTime) {
+			minTime = tidbCPUs.getTime()
+		}
+		if !rus.findTime(minTime) || !tikvCPUs.findTime(minTime) || !tidbCPUs.findTime(minTime) {
```
glorv, line 241: I think we should not break here, consider the following case(each number represent a timestamp):
ru ts: 9, 8, 7,  4, 3, 2, 1 ,0
kv cpu ts: 8, 7, 6, 5, 4, 3, 2, 1
db cpu: 5,4,3,2,1,0
After the first step, the minTime should be 5, then rus.findTime will point at 4 and thus return false, the iterator should not break but continue.

BTW: I perfer to let the iterator to use asc order because it's more intuitive.

CabinfeverB: IMO, rus.findTime will point at 4 but return true

executor/calibrate_resource.go
```diff
@@ -312,14 +327,49 @@ func getTiDBTotalCPUQuota(ctx context.Context, exec sqlexec.RestrictedSQLExecuto
 	return getNumberFromMetrics(ctx, exec, query, "tidb_server_maxprocs")
 }
 
-func getRUPerSec(ctx context.Context, exec sqlexec.RestrictedSQLExecutor, startTime, endTime string) ([]float64, error) {
-	query := fmt.Sprintf("SELECT value FROM METRICS_SCHEMA.resource_manager_resource_unit where time >= '%s' and time <= '%s' ORDER BY time desc", startTime, endTime)
-	return getValuesFromMetrics(ctx, exec, query, "resource_manager_resource_unit")
+type timePointValue struct {
+	tp  time.Time
```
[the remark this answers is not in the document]
CabinfeverB: I think it intuitive to use `After` and `Before` in the time library

executor/calibrate_resource.go
glorv: This is also not correct because the other 2 point to 5

CabinfeverB: I mean, https://github.com/pingcap/tidb/pull/43248#discussion_r1174757866 is useful.
THX!

CabinfeverB: `ru3 and cpu3` will test this case.

executor/calibrate_resource_test.go
CabinfeverB: in `ruModify3`

executor/calibrate_resource.go
```diff
@@ -312,14 +327,49 @@ func getTiDBTotalCPUQuota(ctx context.Context, exec sqlexec.RestrictedSQLExecuto
...
+	return t.vals[t.idx].tp
+}
+
+func (t *timeSeriesValues) getValue() float64 {
+	return t.vals[t.idx].val
+}
+
+func (t *timeSeriesValues) advance(target time.Time) bool {
+	for ; t.idx < len(t.vals); t.idx++ {
+		if t.vals[t.idx].tp.Add(time.Second * 10).After(target) {
```
glorv, line 358: Please add a comment for this extra time offset check

glorv (approved):
LGTM

nolouch (approved):
lgtm

PR #43252, opened 2023-04-20
backup: fix retry of fine-grained backup

### LINKED ISSUE #43236: backup: the error message of fine-grained backup is misleading
## Bug Report

Start a backup, and then inject some errors, (for example, make PD always provide stale information.), and begin backup.

The backup should fail at the "fine-grained backup" stage, and reporting some meaningful error message.

It reports "resolve lock timeout".

v6.1.2, however all versions of BR may have this problem.

YuJuncen:

Issue Number: close #43236

Problem Summary:
We are using the `tikv.Backoffer` to implement retry. Which isn't so customizable hence we cannot fully control the printed error messages and back off duration.

This PR added a new structure `RetryWithBackoffer` to adapt the retry logic with the TiDB backoffer logic, at the same time it has enhanced the error message.
This PR will also increase the max retry time of fine-grained backup. The reason is simple: given it begins the fine-grained backup, there must be some problems in the cluster. We need to be more patient.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

### Release note

```release-note
Fixed a bug that may cause BR report false "resolve lock timeout" error.
```


overvenus (approved):

br/pkg/utils/retry.go
```diff
@@ -84,3 +87,76 @@ func FallBack2CreateTable(err error) bool {
...
+// NextSleepInMS returns the time `BackOff` will sleep in ms of the state.
+func (r *RetryWithBackoffer) NextSleepInMS() int {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	return r.nextBackoff
+}
+
+// TotalSleepInMS returns the total sleeped time in ms.
+func (r *RetryWithBackoffer) TotalSleepInMS() int {
+	return r.totalBackoff + r.bo.GetTotalSleep()
```
Leavrth, line 127: It seems that the `r.bo` is always forked, and only the forked `bo`s are updated something. So is `r.bo.GetTotalSleep` always 0?

Leavrth (approved):

br/pkg/utils/retry.go
YuJuncen: Anyway unlocking locks has a bounded backoff time, which is the TTL of the lock. I think omit the retry time it cost won't be really harmful, hopefully...

YuJuncen: The sole difference between the original logic is that resolve lock may fail due to the accumulated sleep time reaches the sleep time limit. I think that would be fine to eliminate that fail case.

Issue #42643, opened 2023-03-28
The etcd client to the registry of Pump/Drainer info does not use auto-sync and fails when the PD cluster address changed


kennytm:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

1. Start a TiDB cluster with 3 PDs ① ② ③ and a Pump connected
2. Scale-out 3 more PDs ④ ⑤ ⑥
3. Wait 31 seconds
4. Scale-in the original PDs ① ② ③
5. Wait 31 seconds
6. Run `SHOW PUMP STATUS;`

### 2. What did you expect to see? (Required)

We see the status of the pump at step 6

### 3. What did you see instead (Required)

Context deadline exceeded in `etcd.(*Client).List`

### 4. What is your TiDB version? (Required)

v4.0.14

kennytm:
In #9575 we added `AutoSyncInterval` so the etcd client can automatically refresh the PD member list. But this change is never carried into the tidb-tools/tidb-binlog package (later merged into this repository):

https://github.com/pingcap/tidb/blob/10f00936b7c88ae7a62d39291b411abb5003c1aa/util/etcd/etcd.go#L81-L85

(This might also manifest in the "MySQL-compatible AUTO_INCREMENT" feature #38809)

https://github.com/pingcap/tidb/blob/00d48f90e6b19676e6d2a18242ede6c0f98f69f9/meta/autoid/autoid.go#L588-L591

While it is easy to just add back the option `AutoSyncInterval: 30 * time.Second` in, I think a bigger issue is why do we need to have 4 different etcd clients rather than sharing the same one with all the correct config :shrug:

jackysp:
similar issue in other repos,
https://github.com/pingcap/tiflow/blob/b20e1cac3e8142f00665e2bdd4e513f4b2514c3c/dm/pkg/etcdutil/etcdutil.go#L62
https://github.com/pingcap/dm/blob/da80bda17dce4d44c2bd830b58d822dac544b08b/dm/worker/server.go#L107
https://github.com/pingcap/tiflow/blob/b20e1cac3e8142f00665e2bdd4e513f4b2514c3c/dm/ctl/common/util.go#L167
https://github.com/pingcap/dm/blob/da80bda17dce4d44c2bd830b58d822dac544b08b/pkg/etcdutil/etcdutil.go#L74

jackysp:
[...]

Issue #33908, opened 2022-04-13
PlacementManager addrs will not update when PD member changes


lcwangchao:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

1. Setup tidb cluster
2. change PD members and remove all old members

### 2. What did you expect to see? (Required)

The PlacementManager's PD addresses should be changed

### 3. What did you see instead (Required)

The PlacementManager's PD addresses should not changed

### 4. What is your TiDB version? (Required)

After 5.4

PR #43529, opened 2023-05-04
meta/autoid: enable etcd client auto sync for autoid service

tiancaiamao:

Issue Number: close #42643 

ref https://github.com/pingcap/tidb/pull/9575, ref #33908

Problem Summary:

When the pd server change, the corresponding etcd address should be auto-refresh on the client side.

> While it is easy to just add back the option AutoSyncInterval: 30 * time.Second in, I think a bigger issue is why do we need to have 4 different etcd clients rather than sharing the same one with all the correct config 🤷

Because 1. tidb repo is too big, and it's not designed to be used as a lib. (this applies to all tidb/util/xxx), so the etcd client utility in the tidb repo is not a good place to be shared by other repos (tools / binlog ...)
2. For autoid service, it's a indenpendent component, the code is put to the tidb repo though. So this case is similiar to tools etc.
It should have shared the same etcd client instance, but it's not available when the server startup.

Maybe we can maintain a `pingcap/etcdutil` lib to solve the " 4 different etcd clients" issue

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/meta/autoid/autoid.go
+++ b/meta/autoid/autoid.go
@@ -586,8 +586,9 @@ func newSinglePointAlloc(store kv.Storage, dbID, tblID int64, isUnsigned bool) *
 	}
 	if len(addrs) > 0 {
 		etcdCli, err := clientv3.New(clientv3.Config{
-			Endpoints: addrs,
-			TLS:       ebd.TLSConfig(),
+			Endpoints:        addrs,
+			AutoSyncInterval: 30 * time.Second,
+			TLS:              ebd.TLSConfig(),
 		})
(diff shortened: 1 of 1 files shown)


zimulala (approved):
LGTM

hawkingrei (approved):

tiancaiamao:
Affect tidb v6.4.0 above

PR #43541, opened 2023-05-05
resource_control: update static calibrate params

HuSharp:

Issue Number: ref #43212

Problem Summary:

### Check List

Tests 

- [x] Unit test

### Release note

```release-note
- The previous test data due to non-exclusive hard disk, io bandwidth is not full, so the tikv is not the bottleneck, the bottleneck is tidb.
But because tidb is easy to expand, tikv is more difficult to expand, so generally tikv is easy to bottleneck
Now the test data are basically playing full tikv, may be slightly more accurate
- cpu formula is ms, the previous calculation forgot to multiply by 1000
```

### DIFF
--- a/executor/calibrate_resource.go
+++ b/executor/calibrate_resource.go
@@ -45,26 +45,26 @@ var (
 	workloadBaseRUCostMap = map[ast.CalibrateResourceType]*baseResourceCost{
 		ast.TPCC: {
 			tidbCPU:       0.6,
-			kvCPU:         0.15,
-			readBytes:     units.MiB / 2,
-			writeBytes:    units.MiB,
-			readReqCount:  300,
-			writeReqCount: 1750,
+			kvCPU:         500,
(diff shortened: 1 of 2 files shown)


HuSharp:
@glorv @CabinfeverB @nolouch PTAL, thx!

nolouch (commented):
Why need change them?

HuSharp:
> Why need change them?

- The previous test data due to non-exclusive hard disk, io bandwidth is not full, so the tikv is not the bottleneck, the bottleneck is tidb.
	But because tidb is easy to expand, tikv is more difficult to expand, so generally tikv is easy to bottleneck.
	Now the test data are basically playing full tikv, may be slightly more accurate.
- cpu formula is ms, the previous calculation forgot to multiply by 1000. @nolouch

HuSharp:
maybe stale data

Issue #43559, opened 2023-05-05
BR: add general commands for operator


YuJuncen:
## Enhancement
During implementing backing up via EBS volumes across k8s clusters, we need to control the gc finer -- because each BR can only access resources in one k8s cluster (Without changing more code in other components -- say, letting PD to synchronize the volume info).
Consequently, we are going to split out some atomic commands for operator to use. For example, the operator can manually pause the GC when backing up -- so it can run BR command in each k8s cluster and the GC won't be advanced.

PR #43562, opened 2023-05-05
br/cmd: implement the `operator pause-gc` command

YuJuncen:

Issue Number: ref https://github.com/pingcap/tidb/issues/43559
Problem Summary:
This PR implements the new (hidden) command "br operator pause-gc", which blocks the shell and then pause the global GC safepoint.
There are two flags for this command:
`--safepoint` or `-t`, the GC safepoint to be kept, default current resolved ts.
`--ttl` or `-i`, how long the service safe point lives, default 5 mins.

This PR just reused the `StartServiceGcSafePointKeeper`.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
(It is not easy to auto test this PR.)
Firstly, we paused the safepoint.
image
The metrics shows that the safepoint has been kept.
image

- [ ] No code

### Release note

```release-note
None
```


br/pkg/task/operator/cmd.go
```diff
@@ -0,0 +1,101 @@
...
+			return err
+		}
+		log.Info("No service safepoint provided, using the minimal resolved TS.", zap.Uint64("min-resolved-ts", rts))
+		sp.BackupTS = rts
+	}
+	err := utils.StartServiceSafePointKeeper(ctx, ctl.GetPDClient(), sp)
+	if err != nil {
+		return err
+	}
+	log.Info("GC is paused.", zap.String("ID", sp.ID), zap.Uint64("at", sp.BackupTS))
```
3pointer, line 76: it's better to log ttl here.

br/pkg/task/operator/cmd.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	mgr, err := pdutil.NewPdController(ctx, pdAddrs, tc, cfg.TLS.ToPDSecurityOption())
+	if err != nil {
+		return nil, err
+	}
+	return mgr, nil
+}
+
+func cleanUpWith(f func(ctx context.Context)) {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
3pointer, line 36: Does 30 seconds enough for operator to start all jobs?

br/pkg/task/operator/cmd.go
```diff
@@ -0,0 +1,101 @@
+package operator
```
3pointer, line 1: license miss

br/pkg/task/operator/cmd.go
```diff
@@ -0,0 +1,101 @@
...
+	eg.Go(func() error { return pauseSchedulerKeeper(ectx, mgr) })
+
+	return eg.Wait()
+}
+
+func pauseGCKeeper(ctx context.Context, cfg *PauseGcConfig, ctl *pdutil.PdController) error {
+	// Note: should we remove the service safepoint as soon as this exits?
+	sp := utils.BRServiceSafePoint{
+		ID:       utils.MakeSafePointID(),
+		TTL:      int64(cfg.TTL.Seconds()),
```
3pointer, line 61: `GC` has a ttl, but `scheduler` resume as soon as function exits. is this work expected?

br/pkg/task/operator/cmd.go
YuJuncen: This time out is used only when the program exits, in fact this program will run until EBS backups in every k8s cluster finished. (At that context, the original context has been canceled.)

br/pkg/task/operator/cmd.go
YuJuncen: I think blocking GC for a little longer won't bring bad things. (But it will probably be a disaster when workloads are applied to a cluster without schedulers...)

br/pkg/task/operator/cmd.go
3pointer: if so, we don't need a timeout here, right?

3pointer: I just wondering how other operators(if exists) use this command

3pointer (approved):

br/cmd/br/main.go
```diff
@@ -49,6 +49,7 @@ func main() {
 		NewBackupCommand(),
 		NewRestoreCommand(),
 		NewStreamCommand(),
+		newOpeartorCommand(),
```
BornChanger, line 52:
```suggestion
		NewOpeartorCommand(),
```

YuJuncen: I think this shouldn't be public function. 🤔️

BornChanger (approved):

PR #43666, opened 2023-05-10
*: Update bootstrap and support pause user DDL when upgrading TiDB

zimulala:

Issue Number: close https://github.com/pingcap/tidb/issues/43665

Problem Summary:
**New version TiDB:**
* set upgrade state and notify all TiDBs 
* wait the owner get the upgrade state
* pause all user DDLs
* _bootstrap(original logic)_
* resume all user DDLs
* set normal state and notify all TiDBs 

**Owner TiDB:**
* If cluster state is upgrading state, pause the user DDL job.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


ddl/syncer/state_syncer.go
```diff
@@ -81,16 +88,17 @@ type serverStateSyncer struct {
 	prompt             string
 	etcdCli            *clientv3.Client
 	session            *concurrency.Session
-	clusterState       *StateInfo
+	clusterState       *atomicutil.UnsafePointer
```
hawkingrei, line 91: atomic.Point[StateInfo] ?

ddl/ddl_worker.go
```diff
@@ -367,14 +370,24 @@ func (d *ddl) addBatchDDLJobs2Table(tasks []*limitJobTask) error {
 		return nil
 	})
 	if err == nil {
-		jobTasks := make([]*model.Job, len(tasks))
+		jobTasks := make([]*model.Job, 0, len(tasks))
```
dhysum, line 373: Need to take care about the memory usage here. There could be a lot of job currently running.

ddl/ddl_worker.go
```diff
@@ -367,14 +370,24 @@ func (d *ddl) addBatchDDLJobs2Table(tasks []*limitJobTask) error {
...
 		for i, task := range tasks {
 			job := task.job
 			job.Version = currentVersion
 			job.StartTS = startTS
 			job.ID = ids[i]
 			setJobStateToQueueing(job)
-			jobTasks[i] = job
+
+			if d.stateSyncer.IsUpgradingState() && !hasSysDB(job) {
+				if err = pauseRunningJob(sess.NewSession(se), job, model.AdminCommandBySystem); err != nil {
```
dhysum, line 382: It would be better to use PauseJobs() instead of pauseRunningJob() here:
- PauseJobs() is an interface to other modules; ddl_worker is different from ddl/ddl.go
- PauseJobs() processes multiple jobs together in one transaction; Normally it would be reduce this loop

ddl/job_table.go
```diff
@@ -137,6 +154,41 @@ func (d *ddl) getJob(se *sess.Session, tp jobType, filter func(*model.Job) (bool
...
+	// TODO: Handle for the name is empty, like ActionCreatePlacementPolicy.
+	for _, name := range sNames {
+		if tidb_util.IsSysDB(name) {
+			return true
+		}
+	}
+	return false
+}
+
+func (d *ddl) filterEndUserDDL(se *sess.Session, job *model.Job) (bool, error) {
```
dhysum, line 168: It would be better to be in ddl/ddl.go

parser/model/ddl.go
```diff
@@ -747,6 +747,12 @@ func (job *Job) IsPaused() bool {
 	return job.State == JobStatePaused
 }
 
+// IsPausedBySystem returns whether the job is paused by system.
+func (job *Job) IsPausedBySystem() bool {
+	// TODO: Replace JobStatePausing with JobStatePaused after merging https://github.com/pingcap/tidb/pull/43297..
+	return job.State == JobStatePausing && job.AdminOperator == AdminCommandBySystem
```
dhysum, line 753: IsPaused() || IsPausing() would be better.

ddl/job_table.go
```diff
@@ -137,6 +154,41 @@ func (d *ddl) getJob(se *sess.Session, tp jobType, filter func(*model.Job) (bool
...
+	if !d.stateSyncer.IsUpgradingState() {
+		if !job.IsPausedBySystem() || hasSysDB(job) {
+			return false, nil
+		}
+		_, err := ResumeJobsBySystem(se.Session(), []int64{job.ID})
```
dhysum, line 173: Good one.

ddl/ddl.go
```diff
@@ -1462,13 +1469,18 @@ func pauseRunningJob(sess *sess.Session, job *model.Job,
 	job.State = model.JobStatePausing
 	job.AdminOperator = byWho
 
+	if job.RawArgs == nil {
+		return nil
+	}
```
dhysum, line 1474: Good one.

Benjamin2037 (approved):
LGTM

ddl/ddl_worker.go
zimulala: This should be the same logic as before, preallocate the same amount of memory, except that the final length of the array is uncertain.

ddl/ddl_worker.go
zimulala: * They are in the the same pkg。
* This logic needn't a new txn, we will do it in `insertDDLJobs2Table`

parser/model/ddl.go
zimulala: paused, it looks like it's okay for now, because I'm just using this one. The other thing is the function name is a little bit simpler, right?

bb7133 (approved):
LGTM

ddl/job_table.go
```diff
@@ -173,6 +183,41 @@ func (d *ddl) getJob(se *sess.Session, tp jobType, filter func(*model.Job) (bool
 	return nil, nil
 }
 
+func hasSysDB(job *model.Job) bool {
+	sNames := job2SchemaNames(job)
```
tangenta, line 187: Just a suggestion: we can avoid the slice allocation by passing a checker function(`func(tableName string) bool`) into `job2SchemaNames`.

session/mock_bootstrap.go
```diff
@@ -39,8 +40,6 @@ var allDDLs = []string{
 	"alter table mock_sys_t add column c4 bigint",
 	"create table test_create_table(a int)",
 	"drop table test_create_table",
-	"create database test_create_db",
-	"drop database test_create_db",
```
tangenta, line 43: Why remove these?

zimulala: Because these aren't the system database, so if we put them in here, they're paused, so they can't execute normally.

ddl/job_table.go
zimulala: Yes, and there's too much code in ddl.go. I'm not going to change it here.

ddl/job_table.go
zimulala: I will use check ID instead of name in the next PR. Also will update it.

PR #43694, opened 2023-05-10
*: use a new algorithm to improve the speed of admin check table/index

### LINKED ISSUE #43693: make admin check table faster
### Background
In the scenario of very large tables (more than 1 billion rows, more than 1000 regions), after a series of optimizations, adding indexes has become very fast. However, verifying the correctness of the index by admin check table (index) is still slow. 

### Idea 
For each row, calculate a value according to the index_column and handle, and finally XOR all rows to obtain a value. Just need to compare the value calculated on the table side and the value calculated on the index side to determine whether the data is consistent.

Example: sbtest1, id is handle, k is the index column.
Calculate the index:
select (bit_xor(concat(id, k))) from sbtest1 use index();
Calculate the table:
select (bit_xor(concat(id, k))) from sbtest1 use index(primary);

wjhuang2016:

Issue Number: close #43693

Problem Summary:
In the scenario of very large tables (more than 1 billion rows, more than 1000 regions), after a series of optimizations, adding indexes has become very fast. However, verifying the correctness of the index by admin check table (index) is still slow.

1. - Use crc32(md5(concat_ws(0x2, indexColumn))) to check if the data are consistent between the table and index.
    - Group by bucket so that we can locate the inconsistent handle.
    - Continue the compute the checksum if a bucket is found inconsistent between the checksums.
    - If the count of rows in a bucket is lower than a threshold, don't group by anymore, compute the checksum for every row in this bucket.
    - report error if the checksum is inconsistent for a row.
3. Add a variable 'tidb_enable_fast_table_check' to control whether to use the new algorithm and set it's default to true.
4. New algorithm doesn't work for multi-value index and prefixed index.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


executor/executor.go
```diff
@@ -2318,3 +2331,470 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+	case w.e.table.Meta().PKIsHandle:
+		pkCols = append(pkCols, w.e.table.Meta().GetPkName().O)
+		primaryIdx = tables.FindPrimaryIndex(w.table.Meta())
+	default: // support decoding _tidb_rowid.
+		pkCols = append(pkCols, model.ExtraHandleName.O)
+	}
+
+	// CheckSum of (handle + index columns).
+	var md5HandleAndIndexCol strings.Builder
+	md5HandleAndIndexCol.WriteString("crc32(md5(concat(")
```
tangenta, line 2451: I don't think `concat()` is appropriate because `concat('ab', 'c')` is identical to `concat('a', 'bc')`. How about adding column IDs to reduce the force-positive cases?

executor/executor.go
```diff
@@ -2318,3 +2331,470 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+			trySaveErr(err)
+			return
+		}
+		defer func() {
+			_, err = se.(sqlexec.SQLExecutor).ExecuteInternal(ctx, "set @@tidb_snapshot = ''")
+			if err != nil {
+				trySaveErr(err)
+				return
+			}
+		}()
```
tangenta, line 2512: Why not set the session variables of `se` directly?

executor/executor.go
```diff
@@ -2318,3 +2331,470 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+	}
+	_, err = se.(sqlexec.SQLExecutor).ExecuteInternal(ctx, "begin")
```
tangenta, line 2514: We should rollback or commit the transaction before returning it to the session pool.

executor/executor.go
```diff
@@ -2318,3 +2331,470 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+			md5HandleAndIndexCol.WriteString(tblCol.GeneratedExprString)
+		} else {
+			md5HandleAndIndexCol.WriteString(col.Name.O)
+		}
+		md5HandleAndIndexCol.WriteString(", 0x1)")
+		if offset != len(idxInfo.Columns)-1 {
+			md5HandleAndIndexCol.WriteString(", ")
+		}
+	}
+	md5HandleAndIndexCol.WriteString(")))")
```
tangenta, line 2469: The practice of splicing strings is vulnerable to string injection attacks. It is recommend to use `%?` and parameters instead.

executor/executor.go
```diff
@@ -2318,3 +2331,470 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+	se, err := w.e.base().getSysSession()
+	se.GetSessionVars().OptimizerUseInvisibleIndexes = true
+	if err != nil {
+		trySaveErr(err)
+		return
+	}
+	defer func() {
+		se.GetSessionVars().OptimizerUseInvisibleIndexes = false
+		w.e.base().releaseSysSession(ctx, se)
+	}()
```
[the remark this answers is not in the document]
wjhuang2016: It should be false for sure.

executor/executor.go
```diff
@@ -2318,3 +2331,470 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+	defer func() { e.done = true }()
+
+	// Here we need check all indexes, includes invisible index
+	e.ctx.GetSessionVars().OptimizerUseInvisibleIndexes = true
+	defer func() {
+		e.ctx.GetSessionVars().OptimizerUseInvisibleIndexes = false
+	}()
+
+	workerPool, err := workerpool.NewWorkerPool[checkIndexTask]("checkIndex",
+		poolutil.CheckTable, 3, e.createWorker, workerpool.OptionSkipRegister[checkIndexTask]{})
```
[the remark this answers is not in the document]
wjhuang2016: For now, I prefer to use the existing concurrency. We can adjust it later if we need to.

executor/executor.go
```diff
@@ -2318,3 +2331,470 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+		se.GetSessionVars().OptimizerUseInvisibleIndexes = false
+		w.e.base().releaseSysSession(ctx, se)
+	}()
+
+	var pkCols []string
+	var pkTypes []*types.FieldType
+	var primaryIdx *model.IndexInfo
+	switch {
+	case w.e.table.Meta().IsCommonHandle:
+		pkColsInfo := w.e.table.Meta().GetPrimaryKey().Columns
```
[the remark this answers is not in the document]
wjhuang2016: GetPrimaryKey is ok when the table is commonhandle, and we need column info.

executor/executor.go
wjhuang2016: No, since we disable multi-statement for internal sql

executor/executor.go
```diff
@@ -2348,3 +2361,463 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+		}
+		whereKey := fmt.Sprintf("((%s - %d) %% %d)", md5Handle.String(), offset, mod)
+		groupByKey := fmt.Sprintf("((%s - %d) div %d %% %d)", md5Handle.String(), offset, mod, bucketSize)
+		if !checkOnce {
+			whereKey = "0"
+		}
+		checkOnce = true
+
+		tblQuery := fmt.Sprintf("select /*+ read_from_storage(tikv[%s.%s]) */ bit_xor(%s), %s, count(*) from %s.%s use index() where %s = 0 group by %s", w.e.dbName, w.e.table.Meta().Name, md5HandleAndIndexCol.String(), groupByKey, w.e.dbName, w.e.table.Meta().Name, whereKey, groupByKey)
+		idxQuery := fmt.Sprintf("select bit_xor(%s), %s, count(*) from %s.%s use index(%s) where %s = 0 group by %s", md5HandleAndIndexCol.String(), groupByKey, w.e.dbName, w.e.table.Meta().Name, idxInfo.Name, whereKey, groupByKey)
```
ywqzzy, line 2555: Why not using stringBuilder?

executor/executor.go
```diff
@@ -2348,3 +2361,463 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+		}()
+	}
+	_, err = se.(sqlexec.SQLExecutor).ExecuteInternal(ctx, "begin")
+	if err != nil {
+		trySaveErr(err)
+		return
+	}
+
+	times := 0
+	const maxTimes = 10
```
ywqzzy, line 2540: Why set it to 10?

executor/executor.go
wjhuang2016: There is no for loop, so it doesn't need to use stringBuilder.

executor/executor.go
wjhuang2016: It's just a limit.

zimulala (commented):
Why do we need a `crc32` outside the `md5` function?

wjhuang2016:
> Why do we need a `crc32` outside the `md5` function?

md5 is string type, we need int type to do offset/mod.

executor/executor.go
```diff
@@ -2348,3 +2361,461 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+	bucketSize := int(CheckTableFastBucketSize.Load())
+
+	ctx := kv.WithInternalSourceType(w.e.contextCtx, kv.InternalTxnAdmin)
+
+	trySaveErr := func(err error) {
+		w.e.err.CompareAndSwap(nil, &err)
+	}
+
+	se, err := w.e.base().getSysSession()
+	se.GetSessionVars().OptimizerUseInvisibleIndexes = true
```
zimulala, line 2448: I think we need to put this line after the check of line2449. Otherwise, it will can't reset `OptimizerUseInvisibleIndexes` when `err != nil` is true.

executor/executor.go
```diff
@@ -2348,3 +2361,461 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+	}
+	checksums := make([]groupByChecksum, 0, len(rows))
+	for _, row := range rows {
+		checksums = append(checksums, groupByChecksum{bucket: row.GetUint64(1), checksum: row.GetUint64(0), count: row.GetInt64(2)})
+	}
+	return checksums, nil
+}
+
+// HandleTask implements the Worker interface.
+func (w *checkIndexWorker) HandleTask(task checkIndexTask) {
```
zimulala, line 2436: This function has nearly 350 lines of code, could you split it?

executor/executor.go
```diff
@@ -2348,3 +2361,461 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+	idxInfo := w.indexInfos[task.indexOffset]
+	bucketSize := int(CheckTableFastBucketSize.Load())
+
+	ctx := kv.WithInternalSourceType(w.e.contextCtx, kv.InternalTxnAdmin)
+
+	trySaveErr := func(err error) {
+		w.e.err.CompareAndSwap(nil, &err)
+	}
+
+	se, err := w.e.base().getSysSession()
```
zimulala, line 2447: Why we need to use a new session here?

executor/executor.go
wjhuang2016: Seems there is a mistake.

executor/executor.go
```diff
@@ -2348,3 +2361,461 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+
+	times := 0
+	const maxTimes = 10
+	for tableRowCntToCheck > lookupCheckThreshold || !checkOnce {
+		times++
+		if times == maxTimes {
+			logutil.BgLogger().Warn("compare checksum by group reaches time limit", zap.Int("times", times))
+			break
+		}
+		whereKey := fmt.Sprintf("((%s - %d) %% %d)", md5Handle.String(), offset, mod)
```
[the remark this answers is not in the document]
wjhuang2016: It's a little complex here, I would add more examples later. But not in this PR.

executor/executor.go
```diff
@@ -2348,3 +2362,457 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+		}
+		tblRow, err := queryToRow(se, tableSQL)
+		if err != nil {
+			trySaveErr(err)
+			return
+		}
+
+		getHandleFromRow := func(row chunk.Row) (kv.Handle, error) {
+			handleDatum := make([]types.Datum, 0)
+			for i, t := range pkTypes {
```
zimulala, line 2653: `pkTypes` is always nil? I only see it is set in https://github.com/pingcap/tidb/pull/43694/files#diff-4c9a4c1e9a402d07a0a78f4db9f766d23f1ff57ceda1e75c4e83b579488ad301R2460.

executor/executor.go
```diff
@@ -2348,3 +2362,457 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+	lookupCheckThreshold := int64(100)
+	checkOnce := false
+
+	if w.e.ctx.GetSessionVars().SnapshotTS != 0 {
+		se.GetSessionVars().SnapshotTS = w.e.ctx.GetSessionVars().SnapshotTS
+		defer func() {
+			se.GetSessionVars().SnapshotTS = 0
+		}()
+	}
+	_, err = se.(sqlexec.SQLExecutor).ExecuteInternal(ctx, "begin")
```
zimulala, line 2531: It seem that I don't see rollback or commit for this txn. Am I wrong?

executor/executor.go
```diff
@@ -2348,3 +2362,457 @@ func isWeakConsistencyRead(ctx sessionctx.Context, node ast.Node) bool {
...
+
+	// CheckSum of (handle + index columns).
+	var md5HandleAndIndexCol strings.Builder
+	md5HandleAndIndexCol.WriteString("crc32(md5(concat_ws(0x2, ")
+	for _, col := range pkCols {
+		md5HandleAndIndexCol.WriteString(dbutil.ColumnName(col))
+		md5HandleAndIndexCol.WriteString(", ")
+	}
+	for offset, col := range idxInfo.Columns {
+		md5HandleAndIndexCol.WriteString("ifnull(")
```
zimulala, line 2482: Why we need to add `ifnull`?

wjhuang2016: concat null would get null, which is unexpected.

executor/executor.go
wjhuang2016: rollback in releaseSysSession

tangenta (approved):

executor/executor.go
zimulala: But we use `concat_ws` which the result isn't nil after executing `select concat_ws('*', null, b) from t;` (b isn't null).

wjhuang2016: Fix

executor/executor.go
wjhuang2016: We can split it later

zimulala (approved):
LGTM

bb7133 (approved):
LGTM

Issue #43818, opened 2023-05-15
Update hasSysDB using check ID instead of name or get all related DBs


zimulala:

zimulala:
handled it in https://github.com/pingcap/tidb/pull/49699

PR #43823, opened 2023-05-15
*: Update hasSysDB using check ID instead of Name, and support all other types.

zimulala:

Issue Number: close https://github.com/pingcap/tidb/issues/43818

Problem Summary:

* Update `hasSysDB` using check ID instead of Name, and support all other types.
* Clean up the log(get global state and global state change) when the TiDB is notOwner.
* Replace `IsPausedBySystem` with `IsPauseBySystem`.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/ddl/ddl_worker.go
+++ b/ddl/ddl_worker.go
@@ -381,21 +381,9 @@ func (d *ddl) addBatchDDLJobs2Table(tasks []*limitJobTask) error {
 		job.ID = ids[i]
 		setJobStateToQueueing(job)
 
-		if d.stateSyncer.IsUpgradingState() {
-			hasSysDB, err := util.HasSysDB(job)
-			if err != nil {
-				task.cacheErr = err
-				logutil.BgLogger().Warn("[ddl] check has system DB failed", zap.Stringer("job", job), zap.Error(err))
-				continue
-			}
-			if !hasSysDB {
-				if err = pauseRunningJob(sess.NewSession(se), job, model.AdminCommandBySystem); err != nil {
-					logutil.BgLogger().Warn("[ddl] pause user DDL by system failed", zap.Stringer("job", job), zap.Error(err))
-					task.cacheErr = err
-					continue
-				}
-				logutil.BgLogger().Info("[ddl] pause user DDL by system successful", zap.Stringer("job", job))
-			}
+		d.setJobPausingForUpgrade(se, task, job)
+		if task.cacheErr != nil {
+			continue
 		}
 
 		jobTasks = append(jobTasks, job)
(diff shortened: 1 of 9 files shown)


session/mock_bootstrap.go
```diff
@@ -107,7 +108,12 @@ func mockUpgradeToVerLatest(s Session, ver int64) {
 		TestHook.OnBootstrap(s)
 		mustExecute(s, sql)
 		logutil.BgLogger().Info("mock upgrade exec", zap.String("sql", sql))
-		time.Sleep(20 * time.Millisecond)
+		// Make test faster.
+		if intest.InTest {
+			time.Sleep(5 * time.Millisecond)
+		} else {
+			time.Sleep(30 * time.Millisecond)
+		}
```
tangenta, line 116: I think the whole function `mockUpgradeToVerLatest` is only used in tests?

ddl/ddl_worker.go
```diff
@@ -369,32 +369,41 @@ func (d *ddl) addBatchDDLJobs2Table(tasks []*limitJobTask) error {
 		startTS = txn.StartTS()
 		return nil
 	})
-	if err == nil {
```
zimulala, line 372: This file only update two place.
One is here, make meet `err != nil` return directly.

ddl/ddl_worker.go
```diff
@@ -369,32 +369,41 @@ func (d *ddl) addBatchDDLJobs2Table(tasks []*limitJobTask) error {
...
+	jobTasks := make([]*model.Job, 0, len(tasks))
+	for i, task := range tasks {
+		job := task.job
+		job.Version = currentVersion
+		job.StartTS = startTS
+		job.ID = ids[i]
+		setJobStateToQueueing(job)
+
+		if d.stateSyncer.IsUpgradingState() {
+			hasSysDB, err := util.HasSysDB(job)
```
zimulala, line 385: Another is check the `HasSysDB`'s error.

session/mock_bootstrap.go
zimulala: No，we will use it in IT

tangenta: Could you please update the comment to indicate that the new one is for unit test?

ddl/ddl_worker.go
tangenta: Maybe wrap this block into a separate function `trySetJobToPausing()` or `setJobPausingForUpgrade()`?

ddl/util/util.go
```diff
@@ -325,3 +326,68 @@ func WrapKey2String(key []byte) string {
...
+const mysqlSchemaID = int64(1)
+
+func isSysDB(dbID int64) bool {
+	return dbID == mysqlSchemaID
+}
+
+// HasSysDB checks if it has a system database.
+func HasSysDB(job *model.Job) (bool, error) {
+	// TODO: Handle for the schema ID is 0, like ActionCreatePlacementPolicy.
+	switch job.Type {
```
tangenta, line 339: I wonder if it's really necessary to enumerate all the possible job types. Can't we determine this by the schema ID or schema name?

Issue #41241, opened 2023-02-09
RFC: reorganize packages layout for this repository


bb7133:
(Comments & suggestions are appreciated!)

## Motivation

As you might notice, the codebase of this repository is organized in a ‘flattened’ way: there are 40+ level-1 directories/packages, which, in my option, has the following cons:

(1) It looks tedious.
(2) Similar to (1), the hierarchical structure is not clear enough. 

So, here I would like to propose the reorganization of the packages.

## The structure

As shown in the figure below, I would like to make the following changes:
* Move the package of Data Platform tools(`br` and `dumpling`) to `tools` directory.
* Move `tidb-server` package to `cmd` directory. 
* Add `pkg` directory, which contains the packages for "TiDB kernel"(most of the level-1 directories today).

```
tidb
├── LICENSES
├── cmd
│   ├── tidb-server
|   └── ... 
├── build
├── docs
├── hooks
├── pkg
|   ├── autoid_service
|   ├── bindinfo
│   ├── config
│   ├── ddl
│   ├── distsql
│   ├── domain
│   ├── errno
│   ├── executor
│   ├── extension
│   ├── errno
│   ├── infoschema
│   ├── keyspace
│   ├── kv
│   ├── lock
│   ├── meta
│   ├── metrics
│   ├── owner
│   ├── meta
│   ├── parser
│   ├── planner
│   ├── plugin
│   ├── resourcemanager
│   ├── server
│   ├── session
│   ├── sessionctx
│   ├── sessiontxn
│   ├── statistics
│   ├── store
│   ├── table
│   ├── tablecodec
│   ├── telemetry
│   ├── testkit
│   ├── tidb-binlog
│   ├── ttl
│   ├── types
│   └── util
│   ├── ttl
|   └── util
|       ├── structure
|       └── ...
├── tests
├── tools
│   ├── br
│   ├── check
│   └── dumpling
└── ...
```

lance6716:
We are really looking forward to it! And I have some considerations:

1. the `tidb-server` in the root is a directory that only contains the main package? can we use a `cmd` directory which will also include tools' main packages under sub-directory?
2. tidb-lightning is put under `br/pkg/lightning`, we want to change its place in PR as well. And for tidb-lightning, part of its functionalities is used by DDL, maybe it's also a good change to move some lightning packages into kernel.
3. I can imagine a lots of git conflict for opening PRs, how will we reduce the impact? For example, provide a reproducing script for moving the package?

[...]

PR #43837, opened 2023-05-15
timer: add user interfaces for timer framework

### LINKED ISSUE #43836: define user models/interfaces of timer framework

lcwangchao:

Issue Number: close #43836

- Add a new package `timer` for timer framework
- Add some new interfaces for user
  - Introduce the model `TimerRecord` to present the meta of the timer
  - Introduce `TimerClient` to provide some operations of timer for user
  - Introduce `Hook` for the users to custom actions when timer event is triggered

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/timer/api/BUILD.bazel
+++ b/timer/api/BUILD.bazel
@@ -1,0 +1,31 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "api",
+    srcs = [
+        "client.go",
+        "hook.go",
+        "store.go",
+        "timer.go",
+    ],
+    importpath = "github.com/pingcap/tidb/timer/api",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//parser/duration",
+        "@com_github_pingcap_errors//:errors",
+    ],
+)
+
+go_test(
+    name = "api_test",
+    timeout = "short",
+    srcs = [
+        "client_test.go",
+        "schedule_policy_test.go",
+        "store_test.go",
+    ],
+    embed = [":api"],
+    flaky = True,
+    shard_count = 7,
+    deps = ["@com_github_stretchr_testify//require"],
+)
--- a/timer/api/timer.go
+++ b/timer/api/timer.go
@@ -1,0 +1,159 @@
+// Copyright 2023 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+import (
+	"time"
+
+	"github.com/pingcap/errors"
+	"github.com/pingcap/tidb/parser/duration"
+)
+
+// SchedPolicyType is the type of the event schedule policy
+type SchedPolicyType string
+
+const (
+	// SchedEventInterval indicates to schedule events every fixed interval
+	SchedEventInterval SchedPolicyType = "INTERVAL"
+)
+
+// SchedEventPolicy is an interface to tell the runtime how to schedule a timer's events
+type SchedEventPolicy interface {
+	// NextEventTime returns the time to schedule the next timer event. If the second return value is true,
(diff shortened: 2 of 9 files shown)


timer/api/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
```
hawkingrei, line 1: according to https://github.com/pingcap/tidb/issues/41241, we should put new package into ```pkg```.

timer/api/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
...
+
+go_library(
```
hawkingrei, line 3: Please add ```timer``` into ```build/nogo_config.json``` to enable some linter such as revive.

timer/api/BUILD.bazel
lcwangchao: I have discussed with @bb7133 that this issue is still in progress and we can move all packages to `pkg` later when it is created.

timer/api/BUILD.bazel
lcwangchao: Done

hawkingrei (approved):

timer/api/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
...
+    srcs = [
+        "client_test.go",
+        "schedule_policy_test.go",
+        "store_test.go",
+    ],
+    embed = [":api"],
+    flaky = True,
+    shard_count = 6,
+    deps = ["@com_github_stretchr_testify//require"],
+)
```
hawkingrei, line 31: Please run ```make bazel_prepare``` again.

timer/api/timer.go
```diff
@@ -0,0 +1,159 @@
...
+const (
+	// SchedEventInterval indicates to schedule events every fixed interval
+	SchedEventInterval SchedPolicyType = "INTERVAL"
+)
+
+// SchedEventPolicy is an interface to tell the runtime how to schedule a timer's events
+type SchedEventPolicy interface {
+	// NextEventTime returns the time to schedule the next timer event. If the second return value is true,
+	// it means we have a next event schedule after `watermark`. Otherwise, it means there is no more event after `watermark`.
+	NextEventTime(watermark time.Time) (time.Time, bool)
```
YangKeao, line 36: Could you give an example about when the second return value will be "false"?

timer/api/timer.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// SchedIntervalPolicy implements SchedEventPolicy, it is the policy of type `SchedEventInterval`
+type SchedIntervalPolicy struct {
```
YangKeao, line 40: I wonder what's the expected behavior if the client found it's too late to start the job (e.g. if the cluster is down for one or two days.). If they are only expected to execute when the time is within a small offset around the "expected" schedule (e.g. `*/5 * * * *` shouldn't happen at 00:02), a function like "NearestSchedTime" (or other better names) which will return the nearest "expected" schedule time would be really helpful.

The only difference between `NearestSchedTime` and for-loop with `NextEventTime` is that the  `NearestSchedTime` can calculate much faster if the timer is regular (e.g. for every 5 seconds). If we don't have this, we'll need to introduce a new range (or limitation) to avoid the "for-loop next" calculates forever, which will also make the expected behavior hard to describe/learn (See the [`.spec.startingDeadlineSeconds`](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#job-creation) in kubernetes cronjob as a counter example :man_facepalming:  )

YangKeao (approved):
Good job. Rest LGTM

timer/api/timer.go
lcwangchao: For example, if one timer only wants to execute once, the second return value can be `false` when the watermark is not zero value. But currently only `INTERVAL` is supported, so it is always true.

timer/api/timer.go
lcwangchao: Good catch! I think there is a principle in the framework that there is only one event triggering at the same time. For example, if the timer is set to trigger at the first minute of each hour but the cluster is down for one or two days. When the cluster recovered, there was still only one event triggered even if we have many events after the watermark. 

Whether to execute all these jobs or ignore some "expired" depends on the user:

1. If the user wants to execute all passed jobs, he should compute `NextEventTime` again and again until the returned event time after now.
2. If the user wants to ignore some jobs after a certain time, he can pass that time instead of the watermark to `NextEventTime` to compute the next event time after it.

When processing one event, a user can submit one or more jobs concurrency, it dependens to him.

PR #43844, opened 2023-05-15
ldap: modify variable name and initialize connection with root_dn and root_pwd

### LINKED ISSUE #43822: LDAP authentication should support the server which doesn't allow anonymous bind
TiDB uses an anonymous bind to detect whether the connection is alive. However, there are some servers which don't support anonymous bind (e.g. the okta LDAP service).

I propose to simple bind with the `root_dn` and `root_pwd` to detect whether the connection is alive. It's also compatible with the original anonymous bind behavior if the `root_dn` and `root_pwd` are set to empty (by default, if the user doesn't need searching).

## Alternatives

I have checked the behavior on two different implementations:

1. Percona server, as shown in its code, will also use `root_dn` and `root_pwd` to initialize the connection.
2. MySQL server doesn't detect the connection status. If a connection is killed (or lost) from the server side, the connection will pollute the connection pool, and all following login will fail. (`tcpdump` on the LDAP server, no request will arrive).

I think the first one is apparently better.

YangKeao:
This is a cherry-pick of https://github.com/pingcap/tidb/pull/43824

Issue Number: close #43822

Problem Summary:

1. TiDB uses variable `root_pw`, which is different from `root_pwd`.
2. TiDB doesn't support the server which doesn't allow anonymous bind.

1. Rename the variables to `root_pwd`.
2. Initialize the connection with binding to the `root_dn` and `root_pwd`.

The only concern is that the the ACL configuration for anonymous user and `root` could be different.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] Manual test
- [ ] No code

I manually tested this patch with okta ldap service, which doesn't support anonymous binding.

Copied from https://github.com/pingcap/tidb/issues/43822, I also attached the "Alternatives" here:

I have checked the behavior on two different implementations:

### Alternatives

1. Percona server, as shown in its code, will also use root_dn and root_pwd to initialize the connection.
2. MySQL server doesn't detect the connection status. If a connection is killed (or lost) from the server side, the connection will pollute the connection pool, and all following login will fail. (tcpdump on the LDAP server, no request will arrive).

I think the first one is apparently better.

### DIFF
--- a/privilege/privileges/ldap/ldap_common.go
+++ b/privilege/privileges/ldap/ldap_common.go
@@ -32,10 +32,10 @@ import (
 type ldapAuthImpl struct {
 	sync.RWMutex
 	// the following attributes are used to search the users
-	bindBaseDN string
-	bindRootDN string
-	bindRootPW string
-	searchAttr string
+	bindBaseDN  string
+	bindRootDN  string
+	bindRootPWD string
+	searchAttr  string
 
 	// the following attributes are used to connect to LDAP server
 	ldapServerHost string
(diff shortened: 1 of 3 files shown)


CbcWestwolf (approved):

bb7133 (approved):
LGTM

PR #43907, opened 2023-05-17
ddl: better approach to pause or resume ddl jobs during upgrade

### LINKED ISSUE #43897: Upgrade pause all jobs failed when pause paused jobs
## Bug Report

1. using tiup deploy a cluster with 3 tidb instance
2. create much table by sysbench `sysbench oltp_common --mysql-host=tiup-peer.testbed-endless-enable-ddl-upgrade-tiup9fghg --mysql-port=3390 --mysql-user=root --mysql-db=sbtest2 --time=600 --threads=300 --report-interval=10 --db-driver=mysql --tables=10000 --table-size=10 --create_secondary=false prepare`
3. tiup cluster upgrade ddl_upgrade v7.1.0 --pre-upgrade-script "sh /home/tidb/prescript.sh", upgrade with -with-mock-upgrade

upgrade cluster success
1. tiup upgrade failed `Error: failed to restart: tidb-1-peer tidb-4000.service, please check the instance's log(/tiup/deploy/tidb-4000/log) for more detail.: timed out waiting for port 4000 to be started after 2m0s`
2. fatal error  for tidb and can not start
![图片](https://github.com/pingcap/tidb/assets/10184437/e94ec377-ebad-4ba8-a02c-8949d657e4bd)


```sql
MySQL [(none)]> select tidb_version();
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| tidb_version()                                                                                                                                                                                                                                                                                               |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Release Version: v7.1.0-upgrade
Edition: Community
Git Commit Hash: 63d719517832dc84cc30b393b6c96a3767c4bead
Git Branch: heads/refs/tags/v7.1.0-upgrade
UTC Build Time: 2023-05-12 11:10:27
GoVersion: go1.20.3
Race Enabled: false
TiKV Min Version: 6.2.0-alpha
Store: tikv |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

dhysum:

Issue Number: close #43897

Problem Summary:
1. TiDB will try to pause jobs again if restarted during upgrade, need to omit ErrPausedDDLJob
2. Once there is error during Pause/Resume a bunch of jobs, it will fail immediately and return. Which should be continue to next job
3. There are some corner cases that jobs could not be paused during upgrade, which should not become runnable

1. Would not run the jobs that could not be paused during TiDB Upgrade
2. Omit the error of `pause` during TiDB Upgrade in DDL Worker
3. Continue to pause/resume next job even error happened
4. Return `ErrPausedDDLJob` if already `paused`
5. Process the errors returned by  PauseAllJobsBySystem and ResumeAllJobsBySystem during upgrade

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
- Support `admin pause ddl jobs 3,5;`
- Support `admin resume ddl jobs 3,5;`
```

### DIFF
--- a/ddl/ddl.go
+++ b/ddl/ddl.go
@@ -1462,9 +1462,11 @@ func cancelRunningJob(sess *sess.Session, job *model.Job,
 // pauseRunningJob check and pause the running Job
 func pauseRunningJob(sess *sess.Session, job *model.Job,
 	byWho model.AdminCommandOperator) (err error) {
-	// It would be much better doing this filter during `getJobsBySQL`, but not now.
+	if job.IsPausing() || job.IsPaused() {
+		return dbterror.ErrPausedDDLJob.GenWithStackByArgs(job.ID)
+	}
 	if !job.IsPausable() {
-		err = dbterror.ErrCannotPauseDDLJob.GenWithStackByArgs(job.ID)
+		err = dbterror.ErrCannotPauseDDLJob.GenWithStackByArgs(job.ID, job.State.String(), job.SchemaState.String())
 		if err != nil {
 			return err
 		}
@@ -1483,10 +1485,14 @@ func pauseRunningJob(sess *sess.Session, job *model.Job,
 // resumePausedJob check and resume the Paused Job
 func resumePausedJob(se *sess.Session, job *model.Job,
 	byWho model.AdminCommandOperator) (err error) {
-	if !job.IsResumable() ||
-		// The Paused job should only be resumed by who paused it
-		job.AdminOperator != byWho {
-		return dbterror.ErrCannotResumeDDLJob.GenWithStackByArgs(job.ID)
+	if !job.IsResumable() {
+		return dbterror.ErrCannotResumeDDLJob.GenWithStackByArgs(job.ID,
+			"job has not been paused")
+	}
+	// The Paused job should only be resumed by who paused it
+	if job.AdminOperator != byWho {
+		return dbterror.ErrCannotResumeDDLJob.GenWithStackByArgs(job.ID,
+			"job is paused by "+job.AdminOperator.String())
 	}
 
 	job.State = model.JobStateQueueing
@@ -1546,18 +1552,14 @@ func processJobs(process func(*sess.Session, *model.Job, model.AdminCommandOpera
 			err = process(ns, job, byWho)
 			if err != nil {
 				errs[i] = err
-				break
+				continue
 			}
 
 			err = updateDDLJob2Table(ns, job, true)
 			if err != nil {
-				break
+				errs[i] = err
+				continue
 			}
-		}
-		// We may meet some error on job update, try it again
-		if err != nil {
-			ns.Rollback()
-			continue
 		}
 
 		// There may be some conflict during the update, try it again
--- a/ddl/ddl.go
+++ b/ddl/ddl.go
@@ -1557,8 +1557,9 @@ func processJobs(process func(*sess.Session, *model.Job, model.AdminCommandOpera
 
 			err = updateDDLJob2Table(ns, job, true)
 			if err != nil {
-				errs[i] = err
-				continue
+				// unexpected error on updating the job table, should roll it back
+				ns.Rollback()
+				return errs, err
 			}
 		}
 
@@ -1624,7 +1625,7 @@ func processAllJobs(process func(*sess.Session, *model.Job, model.AdminCommandOp
 	for {
 		var jobs []*model.Job
 		jobs, err = getJobsBySQL(ns, JobTable,
-			fmt.Sprintf("job_id >= %s order by job_id asc limit %s",
+			fmt.Sprintf("job_id >= %s and processing order by job_id asc limit %s",
 				strconv.FormatInt(jobID, 10),
 				strconv.FormatInt(int64(limit), 10)))
 		if err != nil {
@@ -1636,11 +1637,12 @@ func processAllJobs(process func(*sess.Session, *model.Job, model.AdminCommandOp
 			err = process(ns, job, byWho)
 			if err != nil {
 				jobErrs[job.ID] = err
-				ns.Rollback()
-				return jobErrs, err
+				continue
 			}
+
 			err = updateDDLJob2Table(ns, job, true)
 			if err != nil {
+				// unexpected error on updating the job table, should roll it back
 				ns.Rollback()
 				return jobErrs, err
 			}
--- a/ddl/ddl.go
+++ b/ddl/ddl.go
@@ -1520,7 +1520,7 @@ func processJobs(process func(*sess.Session, *model.Job, model.AdminCommandOpera
 	var errs []error
 	// We should process (and try) all the jobs in one Transaction.
 	for tryN := uint(0); tryN < 10; tryN += 1 {
-		errs = make([]error, 0, len(ids))
+		errs = make([]error, len(ids))
 		// Need to figure out which one could not be paused
 		jobMap := make(map[int64]int, len(ids))
 		idsStr := make([]string, 0, len(ids))
--- a/ddl/ddl.go
+++ b/ddl/ddl.go
@@ -1516,8 +1516,8 @@ func processJobs(process func(*sess.Session, *model.Job, model.AdminCommandOpera
 	}
 
 	ns := sess.NewSession(sessCtx)
-	var errs []error
 
+	var errs []error
 	// We should process (and try) all the jobs in one Transaction.
 	for tryN := uint(0); tryN < 10; tryN += 1 {
 		errs = make([]error, 0, len(ids))
--- a/ddl/ddl.go
+++ b/ddl/ddl.go
@@ -1486,13 +1486,15 @@ func pauseRunningJob(sess *sess.Session, job *model.Job,
 func resumePausedJob(se *sess.Session, job *model.Job,
 	byWho model.AdminCommandOperator) (err error) {
 	if !job.IsResumable() {
-		return dbterror.ErrCannotResumeDDLJob.GenWithStackByArgs(job.ID,
-			"job has not been paused")
+		errMsg := fmt.Sprintf("job has not been paused, job state:%s, schema state:%s",
+			job.State, job.SchemaState)
+		return dbterror.ErrCannotResumeDDLJob.GenWithStackByArgs(job.ID, errMsg)
 	}
 	// The Paused job should only be resumed by who paused it
 	if job.AdminOperator != byWho {
-		return dbterror.ErrCannotResumeDDLJob.GenWithStackByArgs(job.ID,
-			"job is paused by "+job.AdminOperator.String())
+		errMsg := fmt.Sprintf("job has been paused by [%s], should not resumed by [%s]",
+			job.AdminOperator.String(), byWho)
+		return dbterror.ErrCannotResumeDDLJob.GenWithStackByArgs(job.ID, errMsg)
 	}
 
 	job.State = model.JobStateQueueing
--- a/ddl/ddl.go
+++ b/ddl/ddl.go
@@ -1493,7 +1493,7 @@ func resumePausedJob(se *sess.Session, job *model.Job,
 	// The Paused job should only be resumed by who paused it
 	if job.AdminOperator != byWho {
 		errMsg := fmt.Sprintf("job has been paused by [%s], should not resumed by [%s]",
-			job.AdminOperator.String(), byWho)
+			job.AdminOperator.String(), byWho.String())
 		return dbterror.ErrCannotResumeDDLJob.GenWithStackByArgs(job.ID, errMsg)
 	}
 
--- a/ddl/ddl.go
+++ b/ddl/ddl.go
@@ -1557,9 +1557,8 @@ func processJobs(process func(*sess.Session, *model.Job, model.AdminCommandOpera
 
 			err = updateDDLJob2Table(ns, job, true)
 			if err != nil {
-				// unexpected error on updating the job table, should roll it back
-				ns.Rollback()
-				return errs, err
+				errs[i] = err
+				continue
 			}
 		}
 
@@ -1642,9 +1641,8 @@ func processAllJobs(process func(*sess.Session, *model.Job, model.AdminCommandOp
 
 			err = updateDDLJob2Table(ns, job, true)
 			if err != nil {
-				// unexpected error on updating the job table, should roll it back
-				ns.Rollback()
-				return jobErrs, err
+				jobErrs[job.ID] = err
+				continue
 			}
 		}
 
--- a/ddl/ddl.go
+++ b/ddl/ddl.go
@@ -1466,7 +1466,8 @@ func pauseRunningJob(sess *sess.Session, job *model.Job,
 		return dbterror.ErrPausedDDLJob.GenWithStackByArgs(job.ID)
 	}
 	if !job.IsPausable() {
-		err = dbterror.ErrCannotPauseDDLJob.GenWithStackByArgs(job.ID, job.State.String(), job.SchemaState.String())
+		errMsg := fmt.Sprintf("state [%s] or schema state [%s]", job.State.String(), job.SchemaState.String())
+		err = dbterror.ErrCannotPauseDDLJob.GenWithStackByArgs(job.ID, errMsg)
 		if err != nil {
 			return err
 		}
--- a/ddl/ddl.go
+++ b/ddl/ddl.go
@@ -1520,7 +1520,7 @@ func processJobs(process func(*sess.Session, *model.Job, model.AdminCommandOpera
 
 	// We should process (and try) all the jobs in one Transaction.
 	for tryN := uint(0); tryN < 10; tryN += 1 {
-		errs = make([]error, len(ids))
+		errs = make([]error, 0, len(ids))
 		// Need to figure out which one could not be paused
 		jobMap := make(map[int64]int, len(ids))
 		idsStr := make([]string, 0, len(ids))
@@ -1625,7 +1625,7 @@ func processAllJobs(process func(*sess.Session, *model.Job, model.AdminCommandOp
 	for {
 		var jobs []*model.Job
 		jobs, err = getJobsBySQL(ns, JobTable,
-			fmt.Sprintf("job_id >= %s and processing order by job_id asc limit %s",
+			fmt.Sprintf("job_id >= %s order by job_id asc limit %s",
 				strconv.FormatInt(jobID, 10),
 				strconv.FormatInt(int64(limit), 10)))
 		if err != nil {
--- a/ddl/ddl_worker.go
+++ b/ddl/ddl_worker.go
@@ -975,17 +975,49 @@ func (w *worker) runDDLJob(d *ddlCtx, t *meta.Meta, job *model.Job) (ver int64, 
 		return ver, err
 	}
 
-	if job.IsPaused() || job.IsPausing() {
-		logutil.Logger(w.logCtx).Info("[ddl] DDL job paused", zap.String("job", job.String()))
-		return ver, pauseReorgWorkers(w, d, job)
-	}
-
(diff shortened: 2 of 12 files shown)


session/bootstrap.go
```diff
@@ -1179,7 +1179,8 @@ func syncUpgradeState(s Session) {
 	}
 
 	for i := 0; i < retryTimes; i++ {
-		if _, err = ddl.PauseAllJobsBySystem(s); err == nil {
+		_, err = ddl.PauseAllJobsBySystem(s)
+		if err == nil || dbterror.ErrPausedDDLJob.Equal(err) {
```
zimulala, line 1183: If pause 3 jobs， the first one gets `ErrPausedDDLJob`, then other jobs all don't pause.

ddl/ddl_worker.go
```diff
@@ -975,17 +975,49 @@ func (w *worker) runDDLJob(d *ddlCtx, t *meta.Meta, job *model.Job) (ver int64,
...
+		return ver, err
+	} else if job.IsPausedBySystem() && !hasSysDB(job) {
+		_, err = ResumeJobsBySystem(w.sess.Session(), []int64{job.ID})
+		if err != nil {
+			logutil.BgLogger().Warn(
+				"[ddl] failed to resume user DDL by system after upgrade", zap.Stringer("job", job), zap.Error(err))
+			return ver, err
+		}
+		return ver, nil
+	}
```
zimulala, line 1018: Could we abstract this into a function?

dhysum: ok, that would be better.

session/bootstrap.go
```diff
@@ -1193,11 +1204,15 @@ func syncUpgradeState(s Session) {
...
 
 func syncNormalRunning(s Session) {
-	_, err := ddl.ResumeAllJobsBySystem(s)
+	jobErrs, err := ddl.ResumeAllJobsBySystem(s)
 	if err != nil {
 		logutil.BgLogger().Fatal("upgrade pause all jobs failed", zap.Error(err))
 	}
 
+	for _, err = range jobErrs {
+		logutil.BgLogger().Warn("[DDL] unable to resume the job, error: ", zap.Error(err))
```
zimulala, line 1213:
```suggestion
		logutil.BgLogger().Warn("[DDL] upgrade pause all jobs failed, unable to resume the job, error: ", zap.Error(err))
```

ddl/ddl.go
```diff
@@ -1618,7 +1621,7 @@ func processAllJobs(process func(*sess.Session, *model.Job, model.AdminCommandOp
 	for {
 		var jobs []*model.Job
 		jobs, err = getJobsBySQL(ns, JobTable,
-			fmt.Sprintf("job_id >= %s order by job_id asc limit %s",
```
zimulala, line 1621: Why need to add `processing` ? How about the job is queuing?

ddl/ddl.go
```diff
@@ -1542,19 +1548,16 @@ func processJobs(process func(*sess.Session, *model.Job, model.AdminCommandOpera
...
 			if err != nil {
 				errs[i] = err
-				break
+				continue
 			}
 
 			err = updateDDLJob2Table(ns, job, true)
 			if err != nil {
-				break
+				// unexpected error on updating the job table, should roll it back
```
zimulala, line 1556: Maybe we encounter write-conflict error(If the job is running which is likely to happen). I think it also can do it like line1550-1551.

ddl/ddl.go
```diff
@@ -1630,11 +1633,12 @@ func processAllJobs(process func(*sess.Session, *model.Job, model.AdminCommandOp
...
 			if err != nil {
 				jobErrs[job.ID] = err
-				ns.Rollback()
-				return jobErrs, err
+				continue
 			}
+
 			err = updateDDLJob2Table(ns, job, true)
 			if err != nil {
+				// unexpected error on updating the job table, should roll it back
```
zimulala, line 1641: Ditto.

ddl/ddl_worker.go
```diff
@@ -950,6 +950,18 @@ func (w *worker) countForError(err error, job *model.Job) error {
 	return err
 }
 
+func (w *worker) processJobPausingRequest(d *ddlCtx, job *model.Job) (err error, isRunnable bool) {
+	if job.IsPaused() {
+		logutil.Logger(w.logCtx).Debug("[ddl] paused DDL job ", zap.String("job", job.String()))
```
zimulala, line 955: Why change the log  level to debug?

ddl/job_table.go
```diff
@@ -194,28 +165,41 @@ func hasSysDB(job *model.Job) bool {
...
+		// no matter how, we would not run DDL jobs during upgrade even not paused.
+		return err, false
 	}
-	if hasSysDB(job) {
-		return nil
+
+	if job.IsPausedBySystem() && !hasSysDB(job) {
+		var errs []error
+		errs, err = ResumeJobsBySystem(sess.Session(), []int64{job.ID})
+		if len(errs) > 0 {
```
zimulala, line 194: It seems that `len(errs)` always greater than 0
`errs = make([]error, len(ids))`: https://github.com/pingcap/tidb/pull/43907/files#diff-a185df836b8926c4afe351d4def1b56d1e282c65aabe6677c2fa2c54590cc4b2R1519

ddl/job_table.go
```diff
@@ -194,28 +165,41 @@ func hasSysDB(job *model.Job) bool {
...
 }
 
-func (d *ddl) handleUpgradingState(se *sess.Session, job *model.Job) error {
-	if !d.stateSyncer.IsUpgradingState() {
-		if !job.IsPausedBySystem() || hasSysDB(job) {
-			return nil
+func (d *ddl) processJobDuringUpgrade(sess *sess.Session, job *model.Job) (err error, isRunnable bool) {
+	if d.stateSyncer.IsUpgradingState() {
+		if hasSysDB(job) {
+			return err, false
```
zimulala, line 171: I think we need to handle this job which is means we can return true for `isRunnable`.
Besides, I think we'd better return `nil` instead of `err`.

dhysum: good one

ddl/job_table.go
dhysum: good one

ddl/ddl.go
dhysum: Such kind of failures look like more serious. And the behavior  of writing to TiKV should keep the same as line 1533-1536. Also, we could handle the two returned value in different ways.

ddl/ddl.go
dhysum: good one, remove it.

ddl/ddl.go
dhysum: It looks like better to handle two different kind of failure.

ddl/ddl_worker.go
dhysum: It would be much since every time of polling would come to here if paused.

ddl/job_table.go
dhysum: Actually, it also depends how we design the behavior. We could say that all DDL jobs should not be running during upgrade.

session/bootstrap.go
dhysum: Could not get the point. The operation here is "Resume", and why the log here is about "Pause".

zimulala: Sorry. It seems that the log of line1209 is wrong. I want we add "upgrade resume all jobs failed" for these logs.

ddl/ddl.go
zimulala: If all (many) DDL operations are suspended, there is a risk of write-conflict error, and it is possible to try ten times and fail ten times. 
Originally, there was no exit, and it would have been easier for an interface to handle error in the same way as error. 
Or we could do something special here.

ddl/job_table.go
```diff
@@ -194,28 +165,41 @@ func hasSysDB(job *model.Job) bool {
...
+			return errs[0], false
 		}
-		logutil.BgLogger().Info("[ddl] resume user DDL by system successful", zap.Stringer("job", job))
-		return nil
-	}
-	if job.IsPausing() {
-		return nil
+
+		// no matter how, we would not run DDL jobs during upgrade even not paused.
+		return err, false
```
zimulala, line 188: If the job is pausing, we can't do `pauseReorgWorkers`, so I think we need to add the check of `!job.IsPausedBySystem() ` in line170.

dhysum: pausing is kind of request, it will be turned to be 'model.JobStatePaused' in pauseReorgWorkers() in ddl worker.

dhysum: We may let it keep running if the error is ErrPausedDDLJob.

ddl/job_table.go
```diff
@@ -194,28 +165,51 @@ func hasSysDB(job *model.Job) bool {
...
+			if dbterror.ErrPausedDDLJob.Equal(errs[0]) {
+				// In case of 'Pausing', we need to deliver the job to DDL
+				// worker to pause the background reorganization workers.
+				return true, nil
+			}
+			if dbterror.ErrCannotPauseDDLJob.Equal(errs[0]) {
+				// The state could be 'cancelling' or 'rollback', as well as
+				// some corner state
+				return false, nil
+			}
```
zimulala, line 186: I think will only do `job.IsPausing || hasSysDB(job)` like https://github.com/pingcap/tidb/pull/43907/files#diff-91c692cc214077acc788fcaf2f92fb40d0a6c4aef3e34e3b6ed532677dac9f13L210.
Then we needn't distinguish the error msg, and needn't do line197.

dhysum: It is not a good idea to take care of the job's state out of 'PauseJobsBySystem'. 

As the function 'PauseJobsBySystem' will do the check and pause the job accordingly, we should only check the result of it but not moving some of its logical.

As the logic you mentioned:
1. If hasSysDB(), it will continue the running
2. If ErrPausedDDLJob returned from 'PauseJobsBySystem', then the job could be 'paused' or 'pausing'. And no matter what the state is, the ddl worker will handle it accordingly. Then we don't need to do the check here.

It is about cohesion.

session/bootstrap.go
```diff
@@ -1179,7 +1179,18 @@ func syncUpgradeState(s Session) {
 	}
 
 	for i := 0; i < retryTimes; i++ {
```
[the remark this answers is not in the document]
dhysum: Marking it as failed as soon as possible would be a better choice so that the user will know what is going on here.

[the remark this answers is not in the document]
zimulala: Indeed it is.
But the main reason it I  want wait for pausing successful when encountering canceling or rollbacking job(processing).

ddl/job_table.go
zimulala: I means:
1. If `hasSysDB()`, it will continue the running
2. If `job.IsPausing`, it will continue the running(pausing to paused)
3. Do `PauseJobsBySystem`, if it encouters return `false, err`

dhysum: That is the same with current code.

zimulala: But we don't need to determine the error type. That can remove one `if...`

session/bootstrap.go
```diff
@@ -1193,9 +1204,13 @@ func syncUpgradeState(s Session) {
...
 
 func syncNormalRunning(s Session) {
-	_, err := ddl.ResumeAllJobsBySystem(s)
+	jobErrs, err := ddl.ResumeAllJobsBySystem(s)
 	if err != nil {
-		logutil.BgLogger().Fatal("upgrade pause all jobs failed", zap.Error(err))
+		logutil.BgLogger().Warn("[Upgrading] unexpected error to resume all paused jobs: ", zap.Error(err))
+	}
+
+	for _, err := range jobErrs {
```
zimulala, line 1212: Consider about the case of `err == nil`?

ddl/job_table.go
dhysum: Error handling is necessary when we are calling a function.

Once it returned the 'ErrPausedDDLJob', which means the job has already been 'paused' or 'pausing', we could just let it go running; otherwise, we should not run it.

We should implement the logic by calling a function and handle its return value but not moving some logic from it.

session/bootstrap.go
dhysum: It is a map.

zimulala: Got it.

ddl/job_table.go
```diff
@@ -194,28 +165,45 @@ func hasSysDB(job *model.Job) bool {
...
-		logutil.BgLogger().Info("[ddl] resume user DDL by system successful", zap.Stringer("job", job))
-		return nil
-	}
-	if job.IsPausing() {
-		return nil
+
+		// The job has been issued the command 'pause', which has been turned
+		// to be 'pausing' in job table, then we need to deliver it to ddl
+		// worker to turn it to be 'paused'
+		job.State = model.JobStatePausing
```
zimulala, line 191: Could we remove it?

dhysum: Sure, it just is faster to continue the running this time.

zimulala (approved):
LGTM

ddl/job_table.go
```diff
@@ -194,28 +165,41 @@ func hasSysDB(job *model.Job) bool {
...
+			return true, nil
 		}
-		_, err := ResumeJobsBySystem(se.Session(), []int64{job.ID})
-		if err != nil {
-			logutil.BgLogger().Warn("[ddl] resume user DDL by system failed", zap.Stringer("job", job), zap.Error(err))
-			return err
+		var errs []error
+		// During binary upgrade, pause all running DDL jobs
+		errs, err = PauseJobsBySystem(sess.Session(), []int64{job.ID})
+		if errs != nil && errs[0] != nil {
```
zimulala, line 176:
```suggestion
		if len(errs) > 0 && errs[0] != nil {
```

dhysum: Already Changed it back to an array.  If errs not nil, its length will definitely larger than 0.

zimulala (commented):
CI is failed, and do https://github.com/pingcap/tidb/pull/43982/commits/8ad4fcc62f9d35675500152bdebf8ea037d73603 can fix this issue. You can refer to it.

ddl/ddl.go
```diff
@@ -1462,9 +1462,11 @@ func cancelRunningJob(sess *sess.Session, job *model.Job,
 // pauseRunningJob check and pause the running Job
 func pauseRunningJob(sess *sess.Session, job *model.Job,
 	byWho model.AdminCommandOperator) (err error) {
-	// It would be much better doing this filter during `getJobsBySQL`, but not now.
+	if job.IsPausing() || job.IsPaused() {
+		return dbterror.ErrPausedDDLJob.GenWithStackByArgs(job.ID)
+	}
 	if !job.IsPausable() {
-		err = dbterror.ErrCannotPauseDDLJob.GenWithStackByArgs(job.ID)
+		err = dbterror.ErrCannotPauseDDLJob.GenWithStackByArgs(job.ID, job.State.String(), job.SchemaState.String())
```
tangenta, line 1469: The message should be updated when you put more arguments to the error.

Error message:
```
ErrCannotPauseDDLJob:  mysql.Message("Job [%v] can't be paused now: %s", nil),
```

When you call `GenWithStackByArgs`:
```
fmt.Sprintf("state: %s, schema state: %s", job.State.String(), job.SchemaState.String())
```

ddl/ddl_worker.go
```diff
@@ -818,10 +818,13 @@ func (w *worker) HandleDDLJobTable(d *ddlCtx, job *model.Job) (int64, error) {
 	w.registerSync(job)
 
 	if runJobErr != nil {
-		// wait a while to retry again. If we don't wait here, DDL will retry this job immediately,
-		// which may act like a deadlock.
-		logutil.Logger(w.logCtx).Info("[ddl] run DDL job failed, sleeps a while then retries it.",
-			zap.Duration("waitTime", GetWaitTimeWhenErrorOccurred()), zap.Error(runJobErr))
+		// Omit the ErrPausedDDLJob
+		if !dbterror.ErrPausedDDLJob.Equal(runJobErr) {
```
tangenta, line 822: Is it necessary to sleep for `ErrPausedDDLJob`?

ddl/ddl.go
dhysum: Already done that.

session/bootstrap.go
```diff
@@ -1193,9 +1204,13 @@ func syncUpgradeState(s Session) {
...
 func syncNormalRunning(s Session) {
-	_, err := ddl.ResumeAllJobsBySystem(s)
+	jobErrs, err := ddl.ResumeAllJobsBySystem(s)
 	if err != nil {
-		logutil.BgLogger().Fatal("upgrade pause all jobs failed", zap.Error(err))
+		logutil.BgLogger().Warn("[Upgrading] unexpected error to resume all paused jobs: ", zap.Error(err))
+	}
+
+	for _, e := range jobErrs {
+		logutil.BgLogger().Warn("[Upgrading] unable to resume the job, error: ", zap.Error(e))
```
tangenta, line 1213:
```suggestion
		logutil.BgLogger().Warn("[upgrading] unexpected error to resume all paused jobs: ", zap.Error(err))
	}

	for _, e := range jobErrs {
		logutil.BgLogger().Warn("[upgrading] unable to resume the job, error: ", zap.Error(e))
```

These should be in lower case.

session/bootstrap.go
```diff
@@ -1193,9 +1204,13 @@ func syncUpgradeState(s Session) {
...
 func syncNormalRunning(s Session) {
-	_, err := ddl.ResumeAllJobsBySystem(s)
+	jobErrs, err := ddl.ResumeAllJobsBySystem(s)
 	if err != nil {
-		logutil.BgLogger().Fatal("upgrade pause all jobs failed", zap.Error(err))
+		logutil.BgLogger().Warn("[upgrading] unexpected error to resume all paused jobs: ", zap.Error(err))
+	}
+
+	for _, e := range jobErrs {
+		logutil.BgLogger().Warn("[upgrading] unable to resume the job, error: ", zap.Error(e))
```
tangenta, line 1213: What would happen if some of the jobs cannot be resumed?

ddl/multi_schema_change_test.go
```diff
@@ -1298,7 +1298,7 @@ func (c *cancelOnceHook) OnJobUpdated(job *model.Job) {
 	}
 	c.triggered = true
 	errs, err := ddl.CancelJobs(c.s, []int64{job.ID})
-	if errs[0] != nil {
+	if errs != nil && errs[0] != nil {
```
tangenta, line 1301:
```suggestion
	if len(errs) > 0 && errs[0] != nil {
```

Checking the length is less prone to error.

session/bootstrap.go
dhysum: Only 'paused' job can be resumed. If it can't be resumed, than its state could be pausing or others. If 'pausing', it will become 'paused' later on; otherwise, it will go as it is.

tangenta (approved):

ddl/multi_schema_change_test.go
dhysum: It is an array which length is always bigger than 0.

tangenta: I mean if you check the length, such a restriction can be relax when you do the refactoring in the future.

hawkingrei (approved):

ddl/multi_schema_change_test.go
dhysum: good point.

PR #43988, opened 2023-05-18
*: provide a variable to ignore the real-time stats in the planner

### LINKED ISSUE #46080: Provide a variable to disable real-time stats for estimation for the optimizer
## Background

TiDB maintains some real-time statistics, like the real-time table-level row count and the modify count. They are collected in the DML statements and periodically dumped to update the statistics.

This mechanism is for keeping the statistics more up-to-date in addition to the ANALYZE. However, for some users, this mechanism turns out to make some execution plans unstable and make some diagnoses harder. Therefore, we want to provide a mode that the process of plan generation will not be affected by real-time statistics.

## Design

As usual, this mode will be provided through a session/global level system variable. After discussion with the PM, we will use a relatively broad naming here.

Name: `tidb_opt_objective`
Type: string
Possible values: `"moderate"`, `"determinate"`
`Moderate`: the default value. The behavior is not changed.
`determinate`: Optimizer tends to be conservative. The optimizer will ignore real-time statistics.

### Implementation and other details

1. To make the optimizer ignore the real-time stats, we need to ignore the updates in the real-time row count and the modify count.
It might seem straightforward to add a method to access the two fields and check the system variable in it. However, there are too many places accessing them, and some of them should be controlled by the variable, others should not (because they are not in the plan generation process). In some places, we are even unable to access the session variable.
On the other hand, there's only one place where we fetch stats for tables from the stats cache for generating execution plans, which is `getStatsTable()`. We can copy the stats table and reset the two fields when real-time stats need to be ignored. In this approach, we can implement this without too many changes and intrusions.
2. The out-of-range estimation logic strongly relies on the modify count. The modify count acts as an upper bound to prevent the estimation from producing some ridiculous estimation result.
Now that we can't rely on the modify count, we have to come up with some other method to act as the upper bound. Obviously, it's impossible to have a reliable upper bound in all cases. After some tests, 1/NDV (only considering the histogram part) seems a reasonable and good enough upper bound for now. And we will use that as the upper bound when real-time stats need to be ignored.

#### Test the new out-of-range estimation logic when ignoring the modify count
1. Create a table with 3000 rows of data distributed evenly in `[300, 900)` on column `a`.
2. Analyze this table.
3. In `determinate` mode:
    - The estimated row count for `a <= 300` should be `10`.
    - The estimated row count for `a <= 500` should be `1010`.
    - The estimated row count for `a > 900` should be `5`.
    - The estimated row count for `a <= 900` should be `3000`.

#### Test basic estimation behaviors in both modes
1. Create a table without indexes, and insert several rows of data. Wait for the real-time statistics to be updated.
5. Show the execution plan for a query.
    - In `moderate` mode, the `TableFullScan` has the correct row count.
    - In `determinate` mode, the `TableFullScan` has a row count of 10000.
6. Analyze the table.
7. Show the execution plan for the same query.
    - In both modes, the execution plans are the same, and the `TableFullScan` has the correct row count.
8. Insert several rows of data. Wait for the real-time statistics to be updated.
9. Show the execution plan for the same query.
    -  In `moderate` mode, the `TableFullScan` has the correct row count.
    - In `determinate` mode, the execution plan is the same as in 4.

time-and-fate:

Issue Number: close #46080

Problem Summary: please see the issue

- Add the new system variable. (`sessionctx/`)
- Reset the `ModifyCount` and `RealtimeCount` for queries when this feature is enabled. (`planner/core/logical_plan_builder.go`)
- Add another upper bound for out-of-range estimation when the originally used `ModifyCount` should be ignored. (`statistics/histogram.go`)
- Optimize performance for getting the version from the table stats, which is used in plan cache matching. (`planner/core/plan_cache_utils.go`, `planner/core/logical_plan_builder.go`)
- Allow the pseudo table to trigger stats loading in some cases, which is possible in one case introduced in this PR. (`statistics/table.go`, `statistics/index.go`, `statistics/column.go`)
- Other trivial changes.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


planner/core/logical_plan_builder.go
```diff
@@ -4751,6 +4775,44 @@ func getStatsTable(ctx sessionctx.Context, tblInfo *model.TableInfo, pid int64)
 	return statsTbl
 }
 
+// getLatestVersionFromStatsTable gets statistics information for a table specified by "tableID", and get the max
+// LastUpdateVersion among all Columns and Indices in it.
+// Its overall logic is quite similar to getStatsTable(). During plan cache matching, only the latest version is needed.
+// In such case, compared to getStatsTable(), this function can save some copies, memory allocations and unnecessary
```
winoros, line 4781: It's possible to merge this two func in the future?

sessionctx/variable/session.go
```diff
@@ -1534,6 +1534,12 @@ type SessionVars struct {
 	// When set to true, skip missing partition stats and continue to merge other partition stats to global stats.
 	// When set to false, give up merging partition stats to global stats.
 	SkipMissingPartitionStats bool
+
+	// OptObjective indicates whether the optimizer should be more stable, predictable or more aggressive.
+	// For now, the possible values and corresponding behaviors are:
+	// OptObjectiveModerate: The default value. The optimizer considers the real-time stats (real-time row count, modify count).
+	// OptObjectiveDetermined: The optimizer doesn't consider the real-time stats.
+	OptObjective string
```
winoros, line 1542: The var name seems that it will deal with other things besides the statistics?

statistics/column.go
```diff
@@ -45,6 +45,9 @@ type Column struct {
 
 	// StatsLoadedStatus indicates the status of column statistics
 	StatsLoadedStatus
+	// PhysicalID is the physical table id,
+	// or it could possibly be -1, which means "stats not available".
+	// The -1 case could happen in a pseudo stats table, and in this case, this stats should not trigger stats loading.
 	PhysicalID int64
```
winoros, line 51: will you use it heavily? i want to remove this field from the column/index struct to save memory.

statistics/table.go
```diff
@@ -41,6 +41,7 @@ import (
 	"github.com/pingcap/tidb/util/tracing"
 	"go.uber.org/atomic"
 	"go.uber.org/zap"
+	"golang.org/x/exp/maps"
```
winoros, line 44: You could use the non-experimental package after we upgrade to 1.21.

time-and-fate: The standard library `maps` doesn't provide `Keys()`.

statistics/column.go
time-and-fate: No.
I modified this just because this field has already been used for stats loading, and I'm changing the behavior of stats loading for the pseudo table.

sessionctx/variable/session.go
time-and-fate: It's decided by the PM. He's planning broader usage for this variable.

planner/core/logical_plan_builder.go
```diff
@@ -4727,11 +4728,35 @@ func getStatsTable(ctx sessionctx.Context, tblInfo *model.TableInfo, pid int64)
 		statsTbl = statsHandle.GetPartitionStats(tblInfo, pid, cache.WithTableStatsByQuery())
 	}
 
+	allowPseudoTblTriggerLoading := false
+	// In OptObjectiveDetermined mode, we need to ignore the real-time stats.
+	// To achieve this, we copy the statsTbl and reset the real-time stats fields (set ModifyCount to 0 and set
+	// RealtimeCount to the row count from the ANALYZE, which is fetched from loaded stats in GetAnalyzeRowCount()).
+	if ctx.GetSessionVars().GetOptObjective() == variable.OptObjectiveDetermined {
+		analyzeCount := mathutil.Max(int64(statsTbl.GetAnalyzeRowCount()), 0)
```
hawkingrei, line 4736:
```suggestion
		analyzeCount := max(int64(statsTbl.GetAnalyzeRowCount()), 0)
```

planner/core/logical_plan_builder.go
```diff
@@ -4751,6 +4776,44 @@ func getStatsTable(ctx sessionctx.Context, tblInfo *model.TableInfo, pid int64)
...
+	if pid == tblInfo.ID || ctx.GetSessionVars().StmtCtx.UseDynamicPartitionPrune() {
+		statsTbl = statsHandle.GetTableStats(tblInfo, cache.WithTableStatsByQuery())
+	} else {
+		statsTbl = statsHandle.GetPartitionStats(tblInfo, pid, cache.WithTableStatsByQuery())
+	}
+
+	// 2. Table row count from statistics is zero. Pseudo stats table.
+	realtimeRowCount := statsTbl.RealtimeCount
+	if ctx.GetSessionVars().GetOptObjective() == variable.OptObjectiveDetermined {
+		realtimeRowCount = mathutil.Max(int64(statsTbl.GetAnalyzeRowCount()), 0)
```
hawkingrei, line 4801: Ditto

planner/core/logical_plan_builder.go
```diff
@@ -4751,6 +4776,44 @@ func getStatsTable(ctx sessionctx.Context, tblInfo *model.TableInfo, pid int64)
...
+	}
+	if realtimeRowCount == 0 {
+		return 0
+	}
+
+	// 3. Not pseudo stats table. Return the max LastUpdateVersion among all Columns and Indices
+	for _, col := range statsTbl.Columns {
+		version = mathutil.Max(version, col.LastUpdateVersion)
```
hawkingrei, line 4809: Ditto

planner/core/logical_plan_builder.go
```diff
@@ -4727,11 +4728,35 @@ func getStatsTable(ctx sessionctx.Context, tblInfo *model.TableInfo, pid int64)
...
+			// The original stats table from the stats cache is not a pseudo table, but the analyze row count is 0 (probably
+			// because of no col/idx stats are loaded), which will makes it a pseudo table according to the rule 2 below.
+			// Normally, a pseudo table won't trigger stats loading since we assume it means "no stats available", but
+			// in such case, we need it able to trigger stats loading.
+			// That's why we use the special allowPseudoTblTriggerLoading flag here.
+			if !statsTbl.Pseudo && statsTbl.RealtimeCount > 0 && analyzeCount == 0 {
+				allowPseudoTblTriggerLoading = true
+			}
+			// Copy it so we can modify the ModifyCount and the RealtimeCount safely.
+			statsTbl = statsTbl.ShallowCopy()
```
hawkingrei, line 4749: We will have many copy for ```statstics.Table```. sometime we use deep copy. sometime we need to deep copy except some statistics of column or index. 

so we need a name that is easier to understand.

statistics/histogram.go
```diff
@@ -969,14 +969,31 @@ func (hg *Histogram) OutOfRangeRowCount(sctx sessionctx.Context, lDatum, rDatum
 	}
 	rowCount = totalPercent * hg.NotNullCount()
 
+	// Upper bound logic
+
 	// Use the modifyCount as the upper bound. Note that modifyCount contains insert, delete and update. So this is
 	// a rather loose upper bound.
 	// There are some scenarios where we need to handle out-of-range estimation after both insert and delete happen.
 	// But we don't know how many increases are in the modifyCount. So we have to use this loose bound to ensure it
```
hawkingrei, line 977: it can be written.

```
optObjective := sctx.GetSessionVars().GetOptObjective()
if rowCount > float64(modifyCount) && optObjective != variable.OptObjectiveDetermined {
}
if optObjective ==. xxx {
}
```

hawkingrei (approved):

statistics/histogram.go
time-and-fate: Updated.

winoros (approved):

PR #44003, opened 2023-05-19
expression: fix time range check for `UNIX_TIMESTAMP` | tidb-test=pr/2137

### LINKED ISSUE #43987: UNIX_TIMESTAMP() does not work with range >= '2038-01-19 03:14:08' UTC
## Bug Report

In [MySQL 8.0.28](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_unix-timestamp) they increased the range of the argument of UNIX_TIMEZONE() from '2038-01-19 03:14:07.999999' to '3001-01-19 03:14:07.999999' UTC, which TiDB does not support.

(also see #30133)
```sql
set time_zone = UTC;
select unix_timestamp('2038-01-19 03:14:08.000001');
```

```sql
mysql> set time_zone = UTC;
Query OK, 0 rows affected (0,00 sec)

mysql> select unix_timestamp('2038-01-19 03:14:08.000001');
+----------------------------------------------+
| unix_timestamp('2038-01-19 03:14:08.000001') |
+----------------------------------------------+
|                            2147483648.000001 |
+----------------------------------------------+
1 row in set (0,00 sec)
```

```sql
tidb> set time_zone = UTC;

tidb> select unix_timestamp('2038-01-19 03:14:08.000001');
+----------------------------------------------+
| unix_timestamp('2038-01-19 03:14:08.000001') |
+----------------------------------------------+
|                                     0.000000 |
+----------------------------------------------+
1 row in set (0.00 sec)
```

```
tidb_VERSION(): Release Version: v7.2.0-alpha-216-g6dd86d7e02
Edition: Community
Git Commit Hash: 6dd86d7e0254e9f5ee9ba906b553c3087636cd1a
Git Branch: master
UTC Build Time: 2023-05-18 14:20:59
GoVersion: go1.20.4
Race Enabled: false
TiKV Min Version: 6.2.0-alpha
Store: unistore
```

YangKeao:

Issue Number: close #43987

In MySQL 8.0.28, the range of argument of `UNIX_TIMESTAMP` increases to '3001-01-19 03:14:07.999999' UTC. To keep compatible with MySQL, tidb should also extend this limitation.

The only concern is about the precision handling. The original way (of using `UnixNano`) is not able to represent a time in `3001` year now, so I added two decimals to get the final result.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

### Release note

```release-note
The range of argument of `UNIX_TIMESTAMP` is extended from '2038-01-19 03:14:07.999999' to '3001-01-19 03:14:07.999999' UTC
```


expression/builtin_time.go
```diff
@@ -4162,17 +4162,27 @@ func (c *unixTimestampFunctionClass) getFunction(ctx sessionctx.Context, args []
 // MySQL's Unix timestamp ranges in int32. Values out of range should be rewritten to 0.
```
mjonss, line 4162: Maybe update this comment? Since we are not really returning a Timestamp, but a number with same base, but bigger range than the 32-bit timestamp.

mjonss (approved):
LGTM, with some minor comments.

PR #44107, opened 2023-05-23
*: Introduce tidb_tiflash_node_selection_policy to support different policies of TiFlash node selection 

### LINKED ISSUE #44106: Introduce tidb_tiflash_node_selection_policy to support different policies of TiFlash Node Selection
Background
- Big customers always have some IDCs. They want to deploy the cluster in multiple IDCs to improve HA. Network traffic between IDC needs to be avoided, but now TiFlash always tries to use all the nodes to compute to run faster. This will create lots of network traffic in mutli-IDC scenarios.
- If the customer can distribute two TiFlash replicas in two IDCs through the placement rule, the TiFlash can only choose the TiFlash nodes in the same zone as TiDB node, and there will be no network traffic in the cluster.
- There is no work around.

Goals (Must)
Support different policies of TiFlash Node Selection with variable `tidb_tiflash_node_selection_policy`
This feature is only for On-Primise.

tidb_tiflash_node_selection_policy
- Scope: GLOBAL | SESSION
- Type: Enumeration
- Default value: all_nodes
- Value options: all_nodes, priority_local_zone_nodes, only_local_zone_nodes
- This variable is used to set the policy of TiFlash node selection when the query needs the TiFlash engine.
  - all_nodes means using all the available nodes to do analytic computing, regardless of local zone or other zones.
  - priority_local_zone_nodes means using the nodes in the same zone as the entry TiDB. If not all the tiflash data can be accessed, the query will involve the tiflash nodes from other zones.
  - only_local_zone_nodes means using only the nodes in the same zone as the entry TiDB. If not all the tiflash data can be accessed, the query will report an error.

Target user / role (Must)
- The big customers whose cluster is in multi-IDCs.
- Run TiFlash in every IDC and avoid network traffic between IDCs.

Scenarios (Option)
1. Deploy one cluster in multi-IDCs, the TiFlash nodes are set to different zone attributes according to TiUP settings.
2. Load tiflash data and set placement rules, make sure that each zone has a whole TiFlash replica.
Steps above, see https://docs.pingcap.com/tidb/stable/configure-placement-rules
If there's some data disobeying the placement rule, we can move the replica data to target nodes in the zone. See https://docs.pingcap.com/tidb/stable/pd-control#operator-check--show--add--remove
3. Set the TiFlash node selection policy
4. Query the table involves tiflash.
set @@tidb_tiflash_node_selection_policy=priority_local_zone_nodes;
select ... from table_tiflash;

Functional Specs
Add a new variable to specify the policy of TiFlash node selection when the query needs the TiFlash engine.
- **all_nodes** means using all the available nodes to do analytic computing, regardless of local zone or other zones. It is the same as the default policy now.
- **priority_local_zone_nodes** means using the nodes in the same zone as the entry TiDB. If not all the tiflash data can be accessed, the query will involve the tiflash nodes from other zones. And show the data can not be accessed in the local zone in the warning message.
  - RISK: If the local zone does not have all TiFlash replicas needed in the query, the execution will involve the nodes from other zones, and will cause lots of network traffic between zones.
  - Warning message: Total xx region(s) can not be accessed by TiFlash in the zone [zone_name]: region1, region2, ... (list no more than 3 regions)
- **only_local_zone_nodes** means using only the nodes in the same zone as the entry TiDB. If not all the tiflash data can be accessed, the query will report an error, and show an error message. Because of the feature of TiFlash remote read, a small number of regions  in other zones is acceptable, but performance will be affected. And show the data can not be accessed in the local zone in the warning message. 
	Note: The threshold is fixed, 3 regions per tiflash node. 
	If each tiflash node needs to remote read no more than 3 regions, we will plan the execution in the local zone. 

  	- RISK: If the local zone does not have all TiFlash replicas needed in the query, the execution will fail. This may affect business stability.
  	- If the query reports an error because of zones, there will be an error message: No less than xx region(s) can not be accessed by TiFlash in the zone [zone_name]: region1, region2, ...(list no more than 3 regions). 
  	- If the query is executed successfully and activates TiFlash Remote Read, there will be a warning message: Total xx region(s) can not be accessed by TiFlash in the zone [zone_name]: region1, region2, ...(list no more than 3 regions). 

Corner case:
- If TiDB nodes do not set zone attributes and the policy of TiFlash node selection is not all_nodes, the policy of TiFlash node selection will be ignored, all the tiflash nodes will be used in the tiflash query. And there will be a warning message: The variable tidb_tiflash_node_selection_policy is ignored.
- If TiFlash nodes do not set zone attributes, these nodes will be treated as nodes not in any zone.

Other Risk
- Network traffic
Because the regions may be moving, and the tiflash nodes will have to do remote read. The remote read may choose the nodes from the other zone. So if we choose `only_local_zone_nodes`, there may be some network traffic between zones.
	1. Remote read of TiFlash only affects the scan of regions, and will not involve more nodes to do MPP computing.
	2. Remote read is very slow, we will try to avoid this.

This traffic is very small. In normal cases, the cross-zone traffic is less than 1% of all the regions scanned, most of the time it's nearly 0.

XuHuaiyu:
…ch strategy of tiflash tasks

Issue Number: close #44106 

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Introduce tidb_tiflash_node_selection_policy to support different policies of TiFlash node selection.
```


store/copr/batch_coprocessor.go
```diff
@@ -827,8 +867,22 @@ func buildBatchCopTasksCore(bo *backoff.Backoffer, store *kvStore, rangesForEach
 			}
 		}
 
+		if !tiflashReplicaReadPolicy.IsPolicyAllReplicas() && isTiDBLabelZoneSet {
+			allTiflashStores := cache.RegionCache.GetTiFlashStores(tikv.LabelFilterNoTiFlashWriteNode)
+			aliveStores = filterAliveStores(bo.GetCtx(), allTiflashStores, ttl, store)
```
windtalker, line 872: Since `filterAliveStores` is already called here, there is no need to filter alive stores again in `balanceBatchCopTask` when tiflashReplicaReadPolicy is not all replicas?

store/copr/batch_coprocessor.go
```diff
@@ -827,8 +847,22 @@ func buildBatchCopTasksCore(bo *backoff.Backoffer, store *kvStore, rangesForEach
 			}
 		}
 
+		if !tiflashReplicaReadPolicy.IsPolicyAllReplicas() && isTiDBLabelZoneSet {
```
windtalker, line 850: I think we can move this before the `for` loop?

store/copr/batch_coprocessor.go
```diff
@@ -798,15 +784,49 @@ func failpointCheckWhichPolicy(act tiflashcompute.DispatchPolicy) {
...
+		}
+	}
+	// If no available stores in TiDB zone, we need to access TiFlash stores in other zones.
+	if len(storesMatchedPolicy) == 0 {
+		// needsCrossZoneAccess indicates whether we need to access(directly read or remote read) TiFlash stores in other zones.
+		needsCrossZoneAccess = true
+
+		if policy == tiflash.ClosestAdaptive {
+			// If the policy is `ClosestAdaptive`, we can dispatch tasks to the TiFlash stores in other zones.
+			storesMatchedPolicy = allStores
```
windtalker, line 804: Even if there is only one region need cross zone access, it will use all the stores, is it behavior expected?

util/execdetails/execdetails.go
```diff
@@ -762,7 +762,7 @@ func (context *TiFlashScanContext) Merge(other TiFlashScanContext) {
 
 // Empty check whether TiFlashScanContext is Empty, if scan no pack and skip no pack, we regard it as empty
 func (context *TiFlashScanContext) Empty() bool {
-	res := context.totalDmfileScannedPacks == 0 && context.totalDmfileSkippedPacks == 0
+	res := context.totalDmfileScannedPacks == 0 && context.totalDmfileSkippedPacks == 0 && context.totalLocalRegionNum == 0 && context.totalRemoteRegionNum == 0
```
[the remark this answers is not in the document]
XuHuaiyu: It's actually a bug fix of the execution info in explain analyze.

store/copr/batch_coprocessor.go
```diff
@@ -827,8 +845,22 @@ func buildBatchCopTasksCore(bo *backoff.Backoffer, store *kvStore, rangesForEach
...
 		}
 
+		allTiFlashStores := cache.RegionCache.GetTiFlashStores(tikv.LabelFilterNoTiFlashWriteNode)
+		aliveStores = filterAliveStores(bo.GetCtx(), allTiFlashStores, ttl, store)
+		if careTiFlashReplicaRead {
+			aliveStoreIDsInTiDBZone = make(map[uint64]struct{}, len(aliveStores))
+			for _, as := range aliveStores {
+				// If the `zone` label of the TiFlash store is not set, we treat it as a TiFlash store in other zones.
+				if tiflashZone, isSet := as.GetLabelValue(placement.DCLabelKey); isSet && tiflashZone == tidbZone {
+					aliveStoreIDsInTiDBZone[as.StoreID()] = struct{}{}
```
wshwsh12, line 855: aliveStoreIDsInTiDBZone need reset if needRetry is true.

windtalker: Why? I think it is ok to only init `aliveStoreIDsInTiDBZone` once before the for loop

store/copr/batch_coprocessor.go
```diff
@@ -300,54 +302,31 @@ func balanceBatchCopTaskWithContinuity(storeTaskMap map[uint64]*batchCopTask, ca
 //
 // The second balance strategy: Not only consider the region count between TiFlash stores, but also try to make the regions' range continuous(stored in TiFlash closely).
 // If balanceWithContinuity is true, the second balance strategy is enable.
-func balanceBatchCopTask(ctx context.Context, kvStore *kvStore, originalTasks []*batchCopTask, isMPP bool, ttl time.Duration, balanceWithContinuity bool, balanceContinuousRegionCount int64) []*batchCopTask {
+func balanceBatchCopTask(ctx context.Context, aliveStores []*tikv.Store, originalTasks []*batchCopTask, ttl time.Duration, balanceWithContinuity bool, balanceContinuousRegionCount int64) []*batchCopTask {
```
windtalker, line 305: Is `ttl` still needed in this function?

store/copr/batch_coprocessor.go
```diff
@@ -798,15 +781,99 @@ func failpointCheckWhichPolicy(act tiflashcompute.DispatchPolicy) {
 	})
 }
 
+func filterAllStoresAccordingToTiFlashReplicaRead(allStores []uint64, aliveStoreIDs map[uint64]struct{}, aliveStoreIDsInTiDBZone map[uint64]struct{}, isTiDBZoneSet bool, policy tiflash.ReplicaRead) (storesMatchedPolicy []uint64, needsCrossZoneAccess bool) {
+	if policy == tiflash.AllReplicas || !isTiDBZoneSet {
```
windtalker, line 785: return alive store in this case?

store/copr/batch_coprocessor.go
```diff
@@ -798,15 +781,99 @@ func failpointCheckWhichPolicy(act tiflashcompute.DispatchPolicy) {
...
+		}
+	}
+	return aliveStores, aliveStoreIDsInAllZones, aliveStoreIDsInTiDBZone
+}
+
+// filterAccessableStoresAndBuildRegionInfo filters the stores that can be accessed according to:
+// 1. tiflash_replica_read policy
+// 2. whether the store is alive
+// After filtering, it will build the RegionInfo.
+func filterAccessableStoresAndBuildRegionInfo(cache *RegionCache, bo *Backoffer, task *copTask, rpcCtx *tikv.RPCContext, aliveStoreIDsInAllZones, aliveStoreIDsInTiDBZone map[uint64]struct{}, isTiDBLabelZoneSet bool, tiflashReplicaReadPolicy tiflash.ReplicaRead, regionInfoNeedsReloadOnSendFail []RegionInfo, regionsInOtherZones []uint64, maxRemoteReadCountAllowed int, tidbZone string) (regionInfo RegionInfo, _ []RegionInfo, _ []uint64, err error) {
```
windtalker, line 838: Accessable => Accessible

store/copr/batch_coprocessor.go
```diff
@@ -798,15 +781,99 @@ func failpointCheckWhichPolicy(act tiflashcompute.DispatchPolicy) {
...
+			}
+		}
+	}
+	return
+}
+
+// getAliveStoresAndStoreIDs gets all alive TiFlash stores and their IDs.
+// If `careTiFlashReplicaRead` is true, it will also return the IDs of the alive TiFlash stores in TiDB zone.
+func getAliveStoresAndStoreIDs(ctx context.Context, cache *RegionCache, ttl time.Duration, store *kvStore, careTiFlashReplicaRead bool, tidbZone string) (aliveStores []*tikv.Store, aliveStoreIDsInAllZones, aliveStoreIDsInTiDBZone map[uint64]struct{}) {
+	allTiFlashStores := cache.RegionCache.GetTiFlashStores(tikv.LabelFilterNoTiFlashWriteNode)
```
windtalker, line 815: This will trigger a pd rpc, is it expected?

store/copr/batch_coprocessor.go
```diff
@@ -798,15 +781,99 @@ func failpointCheckWhichPolicy(act tiflashcompute.DispatchPolicy) {
...
+				aliveStoreIDsInTiDBZone[as.StoreID()] = struct{}{}
+			}
+		}
+	} else {
+		aliveStoreIDsInAllZones = make(map[uint64]struct{}, len(aliveStores))
+		for _, as := range aliveStores {
+			aliveStoreIDsInAllZones[as.StoreID()] = struct{}{}
+		}
+	}
+	return aliveStores, aliveStoreIDsInAllZones, aliveStoreIDsInTiDBZone
```
windtalker, line 831: Seems it's meaningless the keep both `aliveStoreIDsInAllZones` and `aliveStoreIDsInTiDBZone`, can we just merge the two into one like `aliveStoreIDs`, for `AllReplicas`, it means `aliveStoreIDsInAllZones`, and for `ClosestAdaptive/ClosestReplicas` it means `aliveStoreIDsInTiDBZone`

windtalker: Oh, please ignore this, I think `aliveStoreIDsInAllZones` is still needed.

store/copr/batch_coprocessor.go
```diff
@@ -798,15 +781,99 @@ func failpointCheckWhichPolicy(act tiflashcompute.DispatchPolicy) {
...
+		}
+	}
+	// If no available stores in TiDB zone, we need to access TiFlash stores in other zones.
+	if len(storesMatchedPolicy) == 0 {
+		// needsCrossZoneAccess indicates whether we need to access(directly read or remote read) TiFlash stores in other zones.
+		needsCrossZoneAccess = true
+
+		if policy == tiflash.ClosestAdaptive {
+			// If the policy is `ClosestAdaptive`, we can dispatch tasks to the TiFlash stores in other zones.
+			storesMatchedPolicy = allStores
```
windtalker, line 801: only return the alive stores?

store/copr/batch_coprocessor.go
XuHuaiyu: Yes, it also triggered a pd rpc before.

https://github.com/pingcap/tidb/pull/44107/files#diff-dec6808c7659579faf2c9c6f9d8b7a7e0d7095b61b9a07cfd2cbc9546a3c7c86L332

store/copr/batch_coprocessor.go
```diff
@@ -798,15 +781,108 @@ func failpointCheckWhichPolicy(act tiflashcompute.DispatchPolicy) {
...
 	const cmdType = tikvrpc.CmdBatchCop
 	rangesLen := 0
 
+	tidbZone, isTiDBLabelZoneSet := config.GetGlobalConfig().Labels[placement.DCLabelKey]
+	var (
+		aliveStores               []*tikv.Store
+		aliveStoreIDsInAllZones   map[uint64]struct{}
+		aliveStoreIDsInTiDBZone   map[uint64]struct{}
+		maxRemoteReadCountAllowed int
+		careTiFlashReplicaRead    = !tiflashReplicaReadPolicy.IsAllReplicas() && isTiDBLabelZoneSet
```
windtalker, line 884: How about just overwrite `tiflashReplicaReadPolicy` to `all_replicas` if TiDBLableZone is not set, so we don't need to handle the conflicts of `tiflashReplicaReadPolicy` and `isTiDBLabelZoneSet` in the following code, and `careTiFlashReplicaRead` can also be removed.

store/copr/batch_coprocessor.go
```diff
@@ -798,15 +781,108 @@ func failpointCheckWhichPolicy(act tiflashcompute.DispatchPolicy) {
...
+	}
+	return
+}
+
+// getAliveStoresAndStoreIDs gets all alive TiFlash stores and their IDs.
+// If `careTiFlashReplicaRead` is true, it will also return the IDs of the alive TiFlash stores in TiDB zone.
+func getAliveStoresAndStoreIDs(ctx context.Context, cache *RegionCache, ttl time.Duration, store *kvStore, careTiFlashReplicaRead bool, tidbZone string) (aliveStores []*tikv.Store, aliveStoreIDsInAllZones, aliveStoreIDsInTiDBZone map[uint64]struct{}) {
+	allTiFlashStores := cache.RegionCache.GetTiFlashStores(tikv.LabelFilterNoTiFlashWriteNode)
+	aliveStores = filterAliveStores(ctx, allTiFlashStores, ttl, store)
+	if careTiFlashReplicaRead {
```
windtalker, line 826: I think `aliveStoreIDsInAllZones` is also need to be inited if TiFlashReplicaPolicy is `closet_adaptive`?

store/copr/batch_coprocessor.go
```diff
@@ -882,7 +975,7 @@ func buildBatchCopTasksCore(bo *backoff.Backoffer, store *kvStore, rangesForEach
 			logutil.BgLogger().Debug(msg)
 		}
 		balanceStart := time.Now()
-		batchTasks = balanceBatchCopTask(bo.GetCtx(), store, batchTasks, isMPP, ttl, balanceWithContinuity, balanceContinuousRegionCount)
+		batchTasks = balanceBatchCopTask(bo.GetCtx(), aliveStores, batchTasks, balanceWithContinuity, balanceContinuousRegionCount)
```
windtalker, line 978: There may be some potential problems here by pass `aliveStores` to `balanceBatchCopTask`, I think it should pass  the alive stores that are expected to be used. For example, if `close_replicas` is used, then the expected 
 stores to be used is just a subset of `aliveStores`, in this case, pass `aliveStores` here will make the balance scores less then expected, and fallback to `balanceContinuousRegionCount` strategy, which is actually not expected.

store/copr/batch_coprocessor.go
```diff
@@ -798,15 +781,108 @@ func failpointCheckWhichPolicy(act tiflashcompute.DispatchPolicy) {
 	})
 }
 
+func filterAllStoresAccordingToTiFlashReplicaRead(allStores []uint64, aliveStoreIDs map[uint64]struct{}, aliveStoreIDsInTiDBZone map[uint64]struct{}, isTiDBZoneSet bool, policy tiflash.ReplicaRead) (storesMatchedPolicy []uint64, needsCrossZoneAccess bool) {
```
windtalker, line 784: If we can overwrite `policy` to `allreplicas` when `isTiDBZoneSet` is false, we don't need pass `isTiDBZoneSet` as an argument in this function.

store/copr/batch_coprocessor.go
XuHuaiyu: why?

store/copr/batch_coprocessor.go
```diff
@@ -882,7 +981,11 @@ func buildBatchCopTasksCore(bo *backoff.Backoffer, store *kvStore, rangesForEach
 			logutil.BgLogger().Debug(msg)
 		}
 		balanceStart := time.Now()
-		batchTasks = balanceBatchCopTask(bo.GetCtx(), store, batchTasks, isMPP, ttl, balanceWithContinuity, balanceContinuousRegionCount)
+		expectedAliveStores := aliveStoresInAllZones
+		if !tiflashReplicaReadPolicy.IsAllReplicas() {
+			expectedAliveStores = aliveStoresInTiDBZone
```
windtalker, line 986: For `closest_adaptive`, it could contain some stores that is not in TiDB zone

windtalker (approved):
LGTM

wshwsh12 (approved):

qw4990 (approved):

Issue #41629, opened 2023-02-21
lightning: optimize local-backend duplicate detection, resolution and record policy


gozssky:
## Enhancement

Currently, the workflow of duplicate detection is as follows:
1. During sending kv pairs to TiKV, detect conflicts and record them to local pebble db.
2. After data is imported, read conflicts from local pebble db and write them to table `lightning_task_info`.`conflict_error_v1`.
3. Detect conflicts from TiKV and record them to table `lightning_task_info`.`conflict_error_v1` also.
4. Read conflict records from table `lightning_task_info`.`conflict_error_v1`, and resolve all conflicts by deleting all related kv pairs.

The workflow of duplicate detection is not perfect, and there are some problems:
1. If there are too many conflicts, lightning will become very slow. This is because the conflict records are written to TiDB through the SQL interface.
2. duplicate-resolution only supports removing all duplicate records, but sometimes users want to keep one of them.
3. The option of local-backend duplicate detection is inconsistent with the option of tidb-backend, which is confusing to users.
4. Checksum and analyze are not performed after any conflict is detected, which reduces the reliability of the import.

For the above problems, we plan to improve the duplicate detection function in the following ways:
1. Reuse `on-duplicate` option of tidb-backend, and deprecated the `duplicate-resolution` option of local-backend.
2. Implement `replace` and `ignore` strategies for local-backend duplicate detection, and the default strategy is `replace`.
3. Do not store all conflict records in TiDB, support setting `max-error-records `for local-backend duplicate detection.
4. Support checksum and analyze no matter whether there are conflicts or not.

PR #44185, opened 2023-05-25
util/extsort: parallelize DiskSorter.Sort

gozssky:

Issue Number: ref https://github.com/pingcap/tidb/issues/41629

Problem Summary:

* Refactor `DiskSorter` to improve performance.
* Remove the dependency on `pebble.DB`.

In the test, it takes 18 minutes to sort 400G data on a 32vCPU, 64G RAM machine.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/util/extsort/disk_sorter.go
+++ b/util/extsort/disk_sorter.go
@@ -89,6 +89,9 @@ func newKVStatsCollector(bucketSize int) *kvStatsCollector {
 	}
 }
 
+// Add is called when a key-value pair is added to the sstable.
+// In our case, it's guaranteed that the key-value pairs are added in order,
+// and key.Trailer must be sstable.InternalKeyKindSet.
 func (c *kvStatsCollector) Add(key sstable.InternalKey, value []byte) error {
 	c.curSize += len(key.UserKey) + len(value)
 	c.lastKey = append(c.lastKey[:0], key.UserKey...)
@@ -377,8 +380,13 @@ func (si *sstIter) Close() error {
 }
 
 type mergingIter struct {
-	heap          mergingIterHeap
-	orderedFiles  []*fileMetadata // ordered by start key
+	heap mergingIterHeap
+	// orderedFiles is the files to be merged. It is ordered by start key.
+	orderedFiles []*fileMetadata
+	// nextFileIndex is the index of the next file to be read.
+	// mergingIter only maintains minimum set of opened files.
+	// If current key is less than the start key of next file,
+	// the next file does not need to be opened.
 	nextFileIndex int
 	openIter      openIterFunc
 	curKey        []byte // only used in Next() to avoid alloc
@@ -1020,6 +1028,8 @@ func pickCompactionFiles(
 		return bytes.Compare(a.key, b.key) < 0
 	})
 
+	// Compute the maximum overlap depth of each interval.
+	// See https://en.wikipedia.org/wiki/Sweep_line_algorithm.
 	maxDepth := 0
 	for i := 1; i < len(intervals); i++ {
 		intervals[i].depth += intervals[i-1].depth
--- a/util/extsort/disk_sorter.go
+++ b/util/extsort/disk_sorter.go
@@ -214,6 +214,9 @@ func (sw *sstWriter) Close() (retErr error) {
 		largest := writerMeta.Largest(pebble.DefaultComparer.Compare)
 		meta.startKey = slices.Clone(smallest.UserKey)
 		meta.lastKey = slices.Clone(largest.UserKey)
+		// Make endKey is exclusive. To avoid unnecessary overlap,
+		// we append 0 to make endKey is the smallest key which is
+		// greater than the last key.
 		meta.endKey = append(meta.lastKey, 0)
 		prop, ok := writerMeta.Properties.UserProperties[kvStatsPropKey]
 		if ok {
@@ -880,6 +883,9 @@ func (d *DiskSorter) readFileMetadata(fileNum int) (*fileMetadata, error) {
 	lastKey, _ := iter.Last()
 	if lastKey != nil {
 		meta.lastKey = slices.Clone(lastKey.UserKey)
+		// Make endKey is exclusive. To avoid unnecessary overlap,
+		// we append 0 to make endKey is the smallest key which is
+		// greater than the last key.
 		meta.endKey = append(meta.lastKey, 0)
 	} else if err := iter.Error(); err != nil {
 		return nil, errors.Trace(err)
--- a/util/extsort/disk_sorter.go
+++ b/util/extsort/disk_sorter.go
@@ -508,10 +508,13 @@ func (m *mergingIter) Seek(key []byte) bool {
 	// Although we have opened all files whose range contains the seek key,
 	// it is possible the seeked key is greater than the original seek key.
 	// So we need to check if the next file needs to be opened.
+	//
 	// Consider the following case:
-	//  file 1: a--------f
-	//  file 2:   b------f
-	//  file 3:       d--f
+	//
+	// file 1: a--------f
+	// file 2:   b------f
+	// file 3:       d--f
+	//
 	// If we seek to "c", we will open file 1 and file 2 first. However, the
 	// smallest key which is greater than "c" is "e", so we need to open file 3
 	// to check if it has keys less than "e".
@@ -1175,6 +1178,20 @@ func buildCompactions(files []*fileMetadata, maxCompactionSize int) []*compactio
 		kvSize      int
 		compactions []*compaction
 	)
+	// We assume that data is uniformly distributed in each file as well as
+	// each bucket range. So we can simply merge buckets of different files.
+	// Although the adjacent buckets may not be of the same file, the current
+	// bucket size is approximately the size of kv pairs between the previous
+	// bucket's upper bound and the current bucket's upper bound.
+	//
+	// Consider the following example:
+	//
+	// file 1 buckets: (a, 10),          (c, 10),          (e, 10)
+	// file 2 buckets:          (b, 10),          (d, 10),          (f, 10)
+	// merged buckets: (a, 10), (b, 10), (c, 10), (d, 10), (e, 10), (f, 10).
+	//
+	// For the adjacent buckets (b, 10) and (c, 10), the size of kv pairs
+	// between b and c is approximately 10 (5 from file 1 and 5 from file 2).
 	for i, bucket := range buckets {
 		if i+1 == len(buckets) {
 			compactions = append(compactions, &compaction{
--- a/util/extsort/disk_sorter.go
+++ b/util/extsort/disk_sorter.go
@@ -673,11 +673,11 @@ type DiskSorterOptions struct {
 	// The overlap depth of these files is 3, because file 0, 1, 2 overlap at
 	// the interval [c, d), and file 1, 2, 3 overlap at the interval [d, e).
 	//
-	// If the overlap depth is larger than CompactionThreshold, the sorter will
-	// compact files to reduce the overlap depth during sorting. The larger the
-	// overlap depth, the larger read amplification will be during iteration.
-	// This is a trade-off between read amplification and sorting cost. Setting
-	// this value to math.MaxInt32 will disable the compaction.
+	// If the overlap depth reached CompactionThreshold, the sorter will compact
+	// files to reduce the overlap depth during sorting. The larger the overlap
+	// depth, the larger read amplification will be during iteration. This is a
+	// trade-off between read amplification and sorting cost. Setting this value
+	// to math.MaxInt will disable the compaction.
 	//
 	// The default value is 16.
 	CompactionThreshold int
@@ -915,12 +915,12 @@ func (d *DiskSorter) doSort(ctx context.Context) error {
 	slices.SortFunc(d.orderedFiles, func(a, b *fileMetadata) bool {
 		return bytes.Compare(a.startKey, b.startKey) < 0
 	})
-	files := d.pickCompactionFiles()
+	files := pickCompactionFiles(d.orderedFiles, d.opts.MaxCompactionSize, d.opts.Logger)
 	for len(files) > 0 {
 		if err := d.compactFiles(ctx, files); err != nil {
 			return err
 		}
-		files = d.pickCompactionFiles()
+		files = pickCompactionFiles(d.orderedFiles, d.opts.MaxCompactionSize, d.opts.Logger)
 	}
 	return nil
 }
@@ -993,7 +993,11 @@ func (d *DiskSorter) compactFiles(ctx context.Context, files []*fileMetadata) er
 	return nil
 }
 
-func (d *DiskSorter) pickCompactionFiles() []*fileMetadata {
+func pickCompactionFiles(
+	allFiles []*fileMetadata,
+	compactionThreshold int,
+	logger *zap.Logger,
+) []*fileMetadata {
 	type interval struct {
 		key   []byte
 		depth int
@@ -1001,8 +1005,8 @@ func (d *DiskSorter) pickCompactionFiles() []*fileMetadata {
 	// intervals is a list of intervals that perfectly overlap file boundaries.
 	// For example, if we have two files [a, c] and [b, d], the intervals will be
 	// [a, b), [b, c), [c, d).
-	intervals := make([]interval, 0, len(d.orderedFiles)*2)
-	for _, file := range d.orderedFiles {
+	intervals := make([]interval, 0, len(allFiles)*2)
+	for _, file := range allFiles {
 		intervals = append(intervals, interval{
 			key:   file.startKey,
 			depth: 1,
@@ -1023,12 +1027,12 @@ func (d *DiskSorter) pickCompactionFiles() []*fileMetadata {
 			maxDepth = intervals[i].depth
 		}
 	}
-	if maxDepth <= d.opts.CompactionThreshold {
+	if maxDepth < compactionThreshold {
 		return nil
 	}
 
 	var files []*fileMetadata
-	for _, file := range d.orderedFiles {
+	for _, file := range allFiles {
 		minIntervalIndex := sort.Search(len(intervals), func(i int) bool {
 			return bytes.Compare(intervals[i].key, file.startKey) >= 0
 		})
@@ -1036,16 +1040,16 @@ func (d *DiskSorter) pickCompactionFiles() []*fileMetadata {
 			return bytes.Compare(intervals[i].key, file.endKey) >= 0
 		})
 		for i := minIntervalIndex; i < maxIntervalIndex; i++ {
-			if intervals[i].depth >= d.opts.CompactionThreshold {
+			if intervals[i].depth >= compactionThreshold {
 				files = append(files, file)
 				break
 			}
 		}
 	}
-	d.opts.Logger.Info(
-		"max overlap depth exceeds the compaction threshold, pick files to compact",
+	logger.Info(
+		"max overlap depth reached the compaction threshold, pick files to compact",
 		zap.Int("maxDepth", maxDepth),
-		zap.Int("threshold", d.opts.CompactionThreshold),
+		zap.Int("threshold", compactionThreshold),
 		zap.Int("fileCount", len(files)),
 	)
 	return files
--- a/util/extsort/disk_sorter.go
+++ b/util/extsort/disk_sorter.go
@@ -453,7 +453,7 @@ func (m *mergingIter) Seek(key []byte) bool {
 				toClose = append(toClose, oldHeap[i:]...)
 				break
 			}
-			// The iterator is not in the range of [startKey, endKey), close it.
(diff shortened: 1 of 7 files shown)


gozssky:
/cc @hawkingrei @lance6716 

The implementation is ready for review. The unit tests will be added next week.

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,606 @@
...
+func (p *sstReaderPool) unref(fileNum int) error {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	v, ok := p.mu.readers[fileNum]
+	if !ok {
+		panic(fmt.Sprintf("sstReaderPool: unref a reader that does not exist: %d", fileNum))
+	}
+	if v.refs.Add(-1) == 0 {
+		delete(p.mu.readers, fileNum)
+		return errors.Trace(v.reader.Close())
```
lance6716, line 299: not sure whether `Close` is slow so we need to move it out from lock scope

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,606 @@
...
+		panic("newMergingIter: orderedFiles are not ordered by start key")
+	}
+	return &mergingIter{
+		orderedFiles: orderedFiles,
+		openIter:     openIter,
+	}
+}
+
+func (m *mergingIter) Seek(key []byte) bool {
+	m.err = nil
```
lance6716, line 412: Maybe it's safer that `m.err` needs an explicit method to reset, in case we forget to check m.Error() 😂

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,606 @@
...
+
+	var toClose []mergingIterItem
+	openedIters := make(map[int]struct{})
+	for i, item := range oldHeap {
+		file := m.orderedFiles[item.index]
+		if bytes.Compare(key, file.startKey) >= 0 &&
+			bytes.Compare(key, file.endKey) < 0 &&
+			item.Seek(key) {
+			openedIters[item.index] = struct{}{}
+			heap.Push(&m.heap, item)
```
lance6716, line 424: heap.Init(itemSlice) has less time complexity than N*heap.Push(item), but that's not important 😂

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,606 @@
...
+		} else {
+			m.err = item.Error()
+			if m.err != nil {
+				// Iterators in m.heap should be closed by mergingIter.Close().
```
lance6716, line 428: who should close the iterators that in `oldHeap` but has not been accessed by for-loop due to `break`?

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,606 @@
...
+	}) {
+		panic("newMergingIter: orderedFiles are not ordered by start key")
+	}
+	return &mergingIter{
+		orderedFiles: orderedFiles,
+		openIter:     openIter,
+	}
+}
+
+func (m *mergingIter) Seek(key []byte) bool {
```
lance6716, line 411: just want to learn that what's the benefit that we only keep the needed iterators open? We are afraid of iterators holding some resources like file descriptor?

lance6716 (commented):
will review later

util/extsort/disk_sorter.go
gozssky: The caller is responsible for checking `m.Error` when he received `false` from any method.

If he calls `Seek`, we assume he has checked the error and thought the error is recoverable. 
On `Seek` successfully returns, `m.err` must be `nil`.
On `Seek` encountered any error, `m.err` is impossible to be the old error.
So we clear the error at first.

util/extsort/disk_sorter.go
gozssky: See line 429, all of them are appended to `toClose`.

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,630 @@
...
+	slices.SortFunc(files, func(a, b *fileMetadata) bool {
+		return bytes.Compare(a.lastKey, b.lastKey) > 0
+	})
+
+	// Since we don't need to implement Prev() method,
+	// we can just open the file with the largest last key.
+	for _, file := range files {
+		iter, err := m.openIter(file)
+		if err != nil {
+			m.err = err
```
lance6716, line 589: it will overwrite the m.err in line 573. Maybe we don't need line 573?

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,630 @@
...
+			return false
+		}
+		if iter.Last() {
```
lance6716, line 592: in which case `iter.Last` is false?
In other words, can we directly return the maximun value of lastKey slice?

lance6716 (commented):
will review soon

util/extsort/disk_sorter.go
gozssky: > in which case iter.Last is false?

Empty file.

> can we directly return the maximun value of lastKey slice?

This will affect the implementation of other methods, like `Next`, `Valid`, `UnsafeKey`, `UnsafeValue`.

util/extsort/disk_sorter.go
gozssky: Oh, I should check error after line 576.

util/extsort/disk_sorter.go
```diff
@@ -80,37 +746,123 @@ func OpenDiskSorter(dirname string, opts *DiskSorterOptions) (*DiskSorter, error
...
-	dbOpts := &pebble.Options{
-		MaxConcurrentCompactions: opts.Concurrency,
-		DisableWAL:               true,
-		L0CompactionThreshold:    math.MaxInt,
-		L0StopWritesThreshold:    math.MaxInt,
+	fs := vfs.Default
+	if dirname == "" {
+		fs = vfs.NewMem()
+	}
+	if err := fs.MkdirAll(dirname, 0755); err != nil {
```
lance6716, line 753: not sure what's the effect of an empty dirname 🤔

util/extsort/disk_sorter.go
```diff
@@ -80,37 +746,123 @@ func OpenDiskSorter(dirname string, opts *DiskSorterOptions) (*DiskSorter, error
...
+	if firstKey != nil {
+		meta.startKey = slices.Clone(firstKey.UserKey)
+	} else if err := iter.Error(); err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	lastKey, _ := iter.Last()
+	if lastKey != nil {
+		meta.lastKey = slices.Clone(lastKey.UserKey)
+		meta.endKey = append(meta.lastKey, 0)
```
lance6716, line 861: I'm not sure `slices.Clone` can guarantee to return a slice with no free capacity so `append` will not change the slice in place 🤔

util/extsort/disk_sorter.go
```diff
@@ -131,29 +891,330 @@ func (d *DiskSorter) Sort(ctx context.Context) error {
...
+func splitCompactionFiles(files []*fileMetadata, maxCompactionDepth int) [][]*fileMetadata {
+	// Split files into non-overlapping groups.
+	slices.SortFunc(files, func(a, b *fileMetadata) bool {
+		return bytes.Compare(a.startKey, b.startKey) < 0
+	})
+	var groups [][]*fileMetadata
+	curGroup := []*fileMetadata{files[0]}
+	maxEndKey := files[0].endKey
+	for _, file := range files[1:] {
+		if bytes.Compare(file.startKey, maxEndKey) > 0 {
```
lance6716, line 1062: since maxEndKey is endKey which is exclusive, we can use >= 0?

lance6716 (commented):
will review soon

util/extsort/disk_sorter.go
```diff
@@ -131,29 +891,330 @@ func (d *DiskSorter) Sort(ctx context.Context) error {
...
+			startKey: startKey,
+			endKey:   endKey,
+		})
+	}
+
+	for _, c := range compactions {
+		for _, file := range files {
+			if !(bytes.Compare(file.endKey, c.startKey) <= 0 ||
+				bytes.Compare(file.startKey, c.endKey) >= 0) {
+				c.orderedFiles = append(c.orderedFiles, file)
```
lance6716, line 1140: IMO it's easier to understand to rename `orderedFiles` to overlapping files and change if condition to positively checking overlap. "ordered by start key" can only be mentioned in comments

util/extsort/disk_sorter.go
gozssky: meta is read-only. It does not matter whether they share the same underlying slice.

https://go.dev/play/p/8hpvjZqHC5n

util/extsort/disk_sorter.go
gozssky: Yes, you are right.

lance6716 (commented):
finish the first reading of util/extsort/disk_sorter.go 😭

util/extsort/disk_sorter.go
gozssky: It does nothing when dirname is empty.

util/extsort/disk_sorter.go
gozssky: Renamed in https://github.com/pingcap/tidb/pull/44185/commits/7240482fd168f8fceca7e8c58f54f563535800c9.

But for the if condition, I think the negative version is more easy to understand.

util/extsort/external_sorter_test.go
```diff
@@ -42,19 +36,9 @@ func runCommonTest(t *testing.T, sorter extsort.ExternalSorter) {
 	require.NoError(t, err)
 
 	rng := rand.New(rand.NewSource(0))
```
lance6716, line 38: recommend that we use time.now as seed and t.Log it to help debug the failed case

util/extsort/external_sorter_test.go
```diff
@@ -114,20 +98,10 @@ func runParallelTest(t *testing.T, sorter extsort.ExternalSorter) {
 	}
 
 	rng := rand.New(rand.NewSource(0))
```
lance6716, line 100: ditto

lance6716 (commented):
I'll check disk_sorter.go and its test soon

util/extsort/disk_sorter.go
```diff
@@ -131,29 +900,363 @@ func (d *DiskSorter) Sort(ctx context.Context) error {
...
+			key:   file.startKey,
+			depth: 1,
+		})
+		intervals = append(intervals, interval{
+			key:   file.endKey,
+			depth: -1,
+		})
+	}
+	slices.SortFunc(intervals, func(a, b interval) bool {
+		return bytes.Compare(a.key, b.key) < 0
```
lance6716, line 1028: can we also place depth=-1 before depth=1 when keys are equal? so for [a, b) and [b, c) we will not get maxDepth=2

lance6716 (commented):
will review compaction logic and tests later 😭

util/extsort/disk_sorter.go
```diff
@@ -131,29 +900,363 @@ func (d *DiskSorter) Sort(ctx context.Context) error {
...
+	for _, file := range files {
+		buckets = append(buckets, file.kvStats.Histogram...)
+		if startKey == nil || bytes.Compare(file.startKey, startKey) < 0 {
+			startKey = file.startKey
+		}
+		if endKey == nil || bytes.Compare(file.endKey, endKey) > 0 {
+			endKey = file.endKey
+		}
+	}
+	// If there is kv stats, return a single compaction for all files.
```
lance6716, line 1127:
```suggestion
	// If there is no kv stats, return a single compaction for all files.
```

util/extsort/disk_sorter.go
```diff
@@ -131,29 +900,363 @@ func (d *DiskSorter) Sort(ctx context.Context) error {
...
+			buckets[n] = buckets[i]
+		}
+	}
+	buckets = buckets[:n+1]
+
+	var (
+		kvSize      int
+		compactions []*compaction
+	)
+	for i, bucket := range buckets {
```
lance6716, line 1159: seems we assume that for a bucket, the size are at the point of UpperBound, rather than distributed evenly in (previous UpperBound, UpperBound]? maybe left a comment to explain this estimation

util/extsort/disk_sorter.go
```diff
@@ -131,29 +900,363 @@ func (d *DiskSorter) Sort(ctx context.Context) error {
...
+			return nil, err
+		}
 	}
-	end := slices.Clone(iter.Key())
-	if err := iter.Close(); err != nil {
-		return errors.Trace(err)
+	if err := iter.Error(); err != nil {
+		return nil, err
+	}
+	err = sw.Close()
```
lance6716, line 1236: if the process exit after this line and before 
```
					_ = d.fs.Remove(makeFilename(d.fs, d.dirname, file.fileNum))
```
the data is duplicated, and we must rely on the deduplication when Next()

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,641 @@
...
+func (si *sstIter) Close() error {
+	err := si.iter.Close()
+	if si.onClose != nil {
+		err = goerrors.Join(err, si.onClose())
+	}
+	return errors.Trace(err)
+}
+
+type mergingIter struct {
+	heap mergingIterHeap
```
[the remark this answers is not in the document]
gozssky: >  if a file's range contains heap[0].key

No, it just means the iterator has been opened. It may be closed because of there is no real key after `heap[0].key`.

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,641 @@
...
+	// we can just open the file with the largest last key.
+	for _, file := range files {
+		iter, err := m.openIter(file)
+		if err != nil {
+			m.err = err
+			return false
+		}
+		if iter.Last() {
+			index := -1
+			for i := range m.orderedFiles {
```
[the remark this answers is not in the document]
gozssky: Maybe the `slices` package should provide a `slices.Find` function.

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,641 @@
...
+				if m.orderedFiles[i] == file {
+					index = i
+					break
+				}
+			}
+			heap.Push(&m.heap, mergingIterItem{
+				Iterator: iter,
+				index:    index,
+			})
+			break
```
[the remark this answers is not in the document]
gozssky: It's possible that the file is empty. For this case, `mergingIter.Last` should return false.

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,641 @@
...
+		} else {
+			m.err = iter.Error()
+			closeErr := iter.Close()
+			m.err = goerrors.Join(m.err, closeErr)
+			if m.err != nil {
+				return false
+			}
+		}
+	}
+	return m.maybeOpenNextFiles()
```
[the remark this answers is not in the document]
gozssky: https://github.com/pingcap/tidb/pull/44185/commits/c5b76a8e3371aaefeaaed1f11a428a3e1258eb7e

util/extsort/disk_sorter.go
gozssky: Good catch! Addressed in https://github.com/pingcap/tidb/pull/44185/commits/c5b76a8e3371aaefeaaed1f11a428a3e1258eb7e.

util/extsort/disk_sorter.go
gozssky: That's right.

util/extsort/disk_sorter.go
gozssky: If data is uniformly distributed in each file, size being at the point is equivalent to size being distributed evenly in `(previous UpperBound, UpperBound]`.

lance6716 (approved):

GMHDBJD (approved):
LGTM

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,654 @@
...
+			return errors.Trace(err)
+		}
+		meta := &fileMetadata{
+			fileNum: sw.fileNum,
+		}
+		smallest := writerMeta.Smallest(pebble.DefaultComparer.Compare)
+		largest := writerMeta.Largest(pebble.DefaultComparer.Compare)
+		meta.startKey = slices.Clone(smallest.UserKey)
+		meta.lastKey = slices.Clone(largest.UserKey)
+		meta.endKey = append(meta.lastKey, 0)
```
D3Hunter, line 217: why add 0 to it, can we comment before it?

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,654 @@
...
+
+type sstReaderPool struct {
+	fs      vfs.FS
+	dirname string
+	cache   *pebble.Cache
+	mu      struct {
+		syncutil.RWMutex
+		readers map[int]struct {
+			reader *sstable.Reader
+			refs   *atomic.Int32
```
D3Hunter, line 237: change it to a named struct? it's hard to track fields in GoLand

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,654 @@
...
+			m.err = err
+		}
+	}
+	if m.err != nil {
+		return false
+	}
+
+	m.nextFileIndex = len(m.orderedFiles)
+	for i, file := range m.orderedFiles {
+		if bytes.Compare(file.startKey, key) > 0 {
```
D3Hunter, line 479: `>=` ?

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,654 @@
...
+			return false
+		}
+	}
+	return m.maybeOpenNextFiles()
+}
+
+func (m *mergingIter) maybeOpenNextFiles() bool {
+	for m.nextFileIndex < len(m.orderedFiles) {
+		file := m.orderedFiles[m.nextFileIndex]
+		if len(m.heap) > 0 && bytes.Compare(m.heap[0].UnsafeKey(), file.startKey) < 0 {
```
D3Hunter, line 562: `<=` ?

util/extsort/disk_sorter.go
```diff
@@ -15,37 +15,654 @@
...
+
+func (m *mergingIter) First() bool {
+	return m.Seek(nil)
+}
+
+func (m *mergingIter) Next() bool {
+	m.err = nil
+	m.curKey = append(m.curKey[:0], m.heap[0].UnsafeKey()...)
+	//revive:disable:empty-block
+	for m.next() && bytes.Equal(m.heap[0].UnsafeKey(), m.curKey) {
```
D3Hunter, line 532: the content of the key in those file is `row key + row id`, so we don't need to handle duplicates(should not have those) ?

D3Hunter (commented):
1/2 lines reviews

util/extsort/disk_sorter.go
gozssky: no

util/extsort/disk_sorter.go
gozssky: It does not affect the correctness.

util/extsort/disk_sorter.go
gozssky: It can be any content. `DiskSorter` is a common implementation for `ExternalSorter` which will automatically skip duplicate keys. See the comment in `ExternalSorter` .

util/extsort/disk_sorter.go
gozssky: The implementation is simple enough. Maybe it is ok to have an anonymous struct. Otherwise, it's hard to think a sturct name for it.

util/extsort/disk_sorter.go
gozssky: `endKey` is exclusive. See the comment in `fileMetadata`.

D3Hunter: ok, but we still need a comment here to explain why add 0

util/extsort/disk_sorter.go
D3Hunter: suppose we have 2 files, and we want to seek to `a`, suppose it's the first time we seek, that means heap is empty, so the first file is skpped
```
a...f
 b.....g
```

gozssky: Why? For the first file, `bytes.Compare(file.startKey, key)  > 0` should return false.

util/extsort/disk_sorter.go
gozssky: https://github.com/pingcap/tidb/pull/44185/commits/e843887215eb44107a5d05d30d10fc1f57acb144

util/extsort/disk_sorter.go
D3Hunter: you're right, i misunderstand

D3Hunter (approved):

lance6716 (approved):

GMHDBJD (approved):
LGTM

PR #44188, opened 2023-05-25
*: Fix issue of query in transaction may return rows with same primary key column value

### LINKED ISSUE #44200: Query in pessimistic transaction may return rows with same unique index column value
## Bug Report

This is a sub-issue of https://github.com/pingcap/tidb/issues/24195#issuecomment-825572552

```sql
/* t1 */ drop table if exists t;
/* t1 */ create table t (c1 varchar(10), c2 int, primary key (c1) nonclustered);
/* t1 */ insert into t values ('a', 1);
/* t1 */ begin;
/* t1 */ update t set c1='b' where c1='a';
/* t2 */ begin;
/* t2 */ insert into t values ('a', 2); -- Blocked until t1 commited.
/* t1 */ commit;
/* t2 */ select * from t use index(primary) order by c1, c2; -- [('a', 1), ('a', 2)]
```

```sql
/* t2 */ select * from t use index(primary) order by c1, c2;
+----+----+
| c1 | c2 |
+----+----+
| a  | 2  |
+----+----+
```

```sql
/* t2 */ select * from t use index(primary) order by c1, c2;
+----+----+
| c1 | c2 |
+----+----+
| a  | 1  |
| a  | 2  |
+----+----+
```

```sql
***************************[ 1. row ]***************************
tidb_version() | Release Version: v7.2.0-alpha-277-geca4e9be28
Edition: Community
Git Commit Hash: eca4e9be284e92aca3d5fbb16e44c6a7655fdf82
Git Branch: master
UTC Build Time: 2023-05-26 02:22:59
GoVersion: go1.20.2
Race Enabled: false
TiKV Min Version: 6.2.0-alpha
Store: unistore
```

crazycs520:

Issue Number: close #44200

Problem Summary: Fix issue of query in transaction may return rows with same unique index column value

#### Minimal reproduce step

```sql
/* t1 */ drop table if exists t;
/* t1 */ create table t (c1 varchar(10), c2 int, primary key (c1) nonclustered);
/* t1 */ insert into t values ('a', 1);
/* t1 */ begin;
/* t1 */ update t set c1='b' where c1='a';
/* t2 */ begin;
/* t2 */ insert into t values ('a', 2); -- Blocked until t1 commited.
/* t1 */ commit;
/* t2 */ select * from t use index(primary) order by c1, c2; -- [('a', 1), ('a', 2)]
```

T2 insert will try to write `{ handle: 2, c1: 'a', c2: 2 }`,  but because t1's update gets the lock first, then T2 insert will be blocked. After T1 committed, T2 insert retry writing `{ handle: 3, c1: 'a', c2: 2 } ` due to write conflict. Then T2 select using snapshot read, `UnionScan` reader read ` { handle: 1, c1: 'a', c2: 1 }` from snapshot, and read ` { handle: 3, c1: 'a', c2: 2 }` from txn's memory buffer. Since `UnionScan` currently only use `handle` to resolve duplicate key, without `unique index`, then T2 select will return `(a 1), (a 2)` records, and this violating the c1 uniqueness attribute.

#### How this PR worked?

- `UnionScan` use both `handle` and `unique index` to resolve duplicate key between row in snapshot and row in memory buffer.
- Additional, `UnionScan` should read all `unique index` columns, make sure that they will not be prune column by the optimizer.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

- [x] Performance regression: Consumes more CPU

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix issue of query in transaction may return rows with same unique index column value
```


planner/core/logical_plan_builder.go
```diff
@@ -4872,7 +4872,28 @@ func (b *PlanBuilder) buildDataSource(ctx context.Context, tn *ast.TableName, as
...
+		var uniqueIndexCols []UniqueIndexCols
+		for _, idx := range ds.table.Indices() {
+			if !tables.IsIndexWritable(idx) {
+				continue
+			}
+			idxInfo := idx.Meta()
+			if !idxInfo.Unique {
+				continue
+			}
+			if ds.table.Meta().IsCommonHandle && idxInfo.Primary {
```
ekexium, line 4884: Do we need to check `PKIsHandle` here?

executor/union_scan.go
```diff
@@ -269,6 +262,49 @@ func (us *UnionScanExec) getSnapshotRow(ctx context.Context) ([]types.Datum, err
...
+	}
+
+	// Check conflict by unique index.
+	stmtCtx := us.ctx.GetSessionVars().StmtCtx
+	for _, idxCols := range us.uniqueIndexCols {
+		idxVals := idxCols.FetchIndexValues(row)
+		// Pass handle = 0 to GenIndexKey, due to we only care about distinct key.
+		iter := idxCols.IdxInfo.GenIndexKVIter(stmtCtx, idxVals, kv.IntHandle(0), nil)
+		for iter.Valid() {
+			uniqueKey, _, distinct, err := iter.Next(nil)
```
zyguan, line 291: We also need to take the physical table id into account when building a unqiue key. Considering the following case:
```sql
/* t1 */ drop table if exists t;
/* t1 */ create table t (c1 int, c2 int, primary key (c1) nonclustered) partition by hash(c1) partitions 4;
/* t1 */ insert into t values (1, 1);
/* t1 */ begin;
/* t1 */ update t set c1=2 where c1=1;
/* t2 */ begin;
/* t2 */ insert into t values (1, 2);
/* t1 */ commit;
/* t2 */ select * from t use index(primary) order by c1, c2;  -- [(1, 1), (1, 2)]
```

planner/core/logical_plan_builder.go
crazycs520: Sorry I can't get you, how to check `PKIsHandle` here?

ekexium: Like 
```suggestion
			if (ds.table.Meta().IsCommonHandle || ds.table.Meta().PKIsHandle) && idxInfo.Primary {
```

zyguan: It's ok here, `table.Indices()` dosn't include `PKIsHandle`.

zyguan (approved):
LGTM

This fix may lead to IndexReader falling into IndexLookUp in some scenarios. Do we need to introduce a mechanism for rolling back? cc @cfzjywxk

crazycs520:
> LGTM
> 
> This fix may lead to IndexReader falling into IndexLookUp in some scenarios. Do we need to introduce a mechanism for rolling back? cc @cfzjywxk

Great, should we add a global variable for this to rollback?

you06:
This PR LGTM

 > Great, should we add a global variable for this to rollback?

If so, I prefer something like `tidb_performance_mode` instead of `tidb_check_unique_index_union_scan`, we already have some fixes that'll cause performance degression(like `tidb_lock_untouched_keys`), maybe union them together will be a better choice for users.

cfzjywxk:
Temporarily hold because the product form is not yet determined, as well as the default solution for performance impact.

PR #44289, opened 2023-05-31
disttask: fix a bug that session may not commit before putting into the session pool

### LINKED ISSUE #44286: session may not commit before putting into the session pool.
## Bug Report

``` SQL
func execSQL(ctx context.Context, se sessionctx.Context, sql string, args ...interface{}) ([]chunk.Row, error) {
	rs, err := se.(sqlexec.SQLExecutor).ExecuteInternal(ctx, sql, args...)
	if err != nil {
		return nil, err
	}
	if rs != nil {
		rows, err := sqlexec.DrainRecordSet(ctx, rs, 1)
		if err != nil {
			return nil, err
		}
		err = rs.Close()
		if err != nil {
			return nil, err
		}
		return rows, err
	}
	return nil, nil
}
```

It there is any error in DrainRecordSet, rs is not close.

master

wjhuang2016:

Issue Number: close #44286

Problem Summary:
rs.Close() may not call if there is any error in DrainRecordSet.
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
Start the tiup cluster, kill and restart PD during the test. Add some logs to check if the session is committed. 
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/disttask/framework/storage/task_table.go
+++ b/disttask/framework/storage/task_table.go
@@ -26,6 +26,7 @@ import (
 	"github.com/pingcap/failpoint"
 	"github.com/pingcap/tidb/disttask/framework/proto"
 	"github.com/pingcap/tidb/kv"
+	"github.com/pingcap/tidb/parser/terror"
 	"github.com/pingcap/tidb/sessionctx"
 	"github.com/pingcap/tidb/util/chunk"
 	"github.com/pingcap/tidb/util/logutil"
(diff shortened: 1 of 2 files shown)


hawkingrei (approved):

lance6716 (approved):
I prefer we don't close the issue, just mitigate the severity and add a linter to prevent this problem in future

disttask/framework/storage/task_table.go
```diff
@@ -78,15 +79,8 @@ func execSQL(ctx context.Context, se sessionctx.Context, sql string, args ...int
...
-		if err != nil {
-			return nil, err
-		}
-		err = rs.Close()
-		if err != nil {
-			return nil, err
-		}
-		return rows, err
+		defer terror.Call(rs.Close)
+		return sqlexec.DrainRecordSet(ctx, rs, 1024)
```
tangenta, line 83: Why change `maxChunkSize` to 1024? I think most internal SQLs should not return so many rows.

wjhuang2016: It may get many row when getting all pending subtasks.

Issue #43691, opened 2023-05-10
Runaway query control based on resource group


Connor1996:
## Motivation
Run-away queries are queries that consume more resources beyond user expectation. This could be caused by improper SQL statement, suboptimal plan.
Runaway query can impact overall performance if they are not managed properly. We need to manage run-away queries effectively. Long-running operations should be identified and aborted. 
Currently, we already have the deadline mechanism pushed down to the TiKV layer that one coprocessor request would not execute in TiKV more than 60s by default. But a runaway query may not cost too much time on one single coprocessor request, thus the deadline mechanism can't help avoid run-away queries.  In the meantime, deadlines can't be too small, otherwise, normal requests can be quickly aborted. 

## How to identify run-away queries?
Runaway queries can adversely impact overall performance if they are not managed properly. Resource manager can take action when a query exceeds more than a specified amount of elapsed time. The elasped time indicates the time of being processed, which excludes the waiting time. 
Differentiating run-away queries from queries that really need to perform a full table/index scan is hard. There is no absolute rule. So we just let users define the rule to identify run-away queries. They can twist it on their own needs. The criteria are only the execution time, at least at present. Maybe add more dimension later.
TiKV would send back the scan detail in coprocessor responses. If the total elapsed time of the query exceeds the threshold, then it would be recognized as a run-away query(statement). 

## Task Breakdown

- [x] Extend resource group meta with runaway config 
  - [x] Update kvproto @Connor1996 https://github.com/pingcap/kvproto/pull/1114
  - [x] Extend create/alter resource group statement @CabinfeverB 
     - [x] parser part https://github.com/pingcap/tidb/pull/43843
     - [x] pd part
       - [x] https://github.com/tikv/pd/pull/6475
       - [x] https://github.com/tikv/pd/pull/6510
       - [x] https://github.com/tikv/pd/pull/6515
     - [x] Make alter patch style https://github.com/pingcap/tidb/pull/44322
     - [x] Extend admin table `information_schema.resource_groups`  https://github.com/pingcap/tidb/pull/43877 
- [x] Identify runaway in cop client and perform action @Connor1996 
  - [x] Introduce runaway checker https://github.com/pingcap/tidb/pull/44339
    - [x] Use const default resource group name https://github.com/pingcap/tidb/pull/44526
  - [x] Update kvproto to add `override_priority` https://github.com/pingcap/kvproto/pull/1114
[...]

PR #44322, opened 2023-06-01
resource_group: support patch for altering resource group

CabinfeverB:

Issue Number: ref #43691 
should be merged after https://github.com/pingcap/tidb/pull/43877
Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


CabinfeverB:
@glorv @nolouch

ddl/ddl_api.go
```diff
@@ -3096,7 +3099,9 @@ func SetDirectResourceGroupUnit(resourceGroupSettings *model.ResourceGroupSettin
 // SetDirectResourceGroupRunawayOption tries to set runaway part of the ResourceGroupSettings.
 func SetDirectResourceGroupRunawayOption(resourceGroupSettings *model.ResourceGroupSettings, typ ast.RunawayOptionType, stringVal string, intVal int32) error {
 	if resourceGroupSettings.Runaway == nil {
-		resourceGroupSettings.Runaway = &model.ResourceGroupRunawaySettings{}
+		resourceGroupSettings.Runaway = &model.ResourceGroupRunawaySettings{
+			Action: model.RunawayNoneAction,
```
CabinfeverB, line 3103: Here are new modifications and add test. @glorv @nolouch @Connor1996

parser/parser_test.go
```diff
@@ -3729,23 +3731,25 @@ func TestDDL(t *testing.T) {
 		{"alter resource group x region ='us, 3'", false, ""},
 		{"alter resource group x cpu='8c', io_read_bandwidth='2GB/s', io_write_bandwidth='200MB/s'", false, ""},
 		{"alter resource group x ru_per_sec=1000", true, "ALTER RESOURCE GROUP `x` RU_PER_SEC = 1000"},
-		{"alter resource group x ru_per_sec=2000, BURSTABLE", true, "ALTER RESOURCE GROUP `x` RU_PER_SEC = 2000, BURSTABLE"},
-		{"alter resource group x BURSTABLE, ru_per_sec=3000", true, "ALTER RESOURCE GROUP `x` BURSTABLE, RU_PER_SEC = 3000"},
-		{"alter resource group x BURSTABLE ru_per_sec=4000", true, "ALTER RESOURCE GROUP `x` BURSTABLE, RU_PER_SEC = 4000"},
+		{"alter resource group x ru_per_sec=2000, BURSTABLE", true, "ALTER RESOURCE GROUP `x` RU_PER_SEC = 2000, BURSTABLE = TRUE"},
```
Connor1996, line 3734: should we keep the original way?

CabinfeverB: Only the restore output of the syntax tree is affected here, and the user can use `burstable` or `burstable=true`.

Connor1996 (approved):
LGTM

parser/ast/ddl.go
```diff
@@ -4542,9 +4550,10 @@ func CheckRunawayAppend(ops []*ResourceGroupRunawayOption, newOp *ResourceGroupR
 type AlterResourceGroupStmt struct {
 	ddlNode
 
-	ResourceGroupName       model.CIStr
-	IfExists                bool
-	ResourceGroupOptionList []*ResourceGroupOption
+	ResourceGroupName              model.CIStr
+	IfExists                       bool
+	ResourceGroupOptionList        []*ResourceGroupOption
+	ResourceGroupRunawayOptionList []*ResourceGroupRunawayOption
```
glorv, line 4556: Why need this field? I see there are already ResourceGroupRunawayOptionList in ResourceGroupOptionList

ddl/ddl_api.go
glorv: Seems this default value is useless

parser/model/model.go
```diff
@@ -1866,6 +1866,7 @@ const (
 	RunawayActionDryRun RunawayActionType = iota
 	RunawayActionCooldown
 	RunawayActionKill
+	RunawayNoneAction
```
glorv, line 1869: NIT: better let this default RunawayNoneAction to be the iota

parser/parser.y
```diff
@@ -1818,10 +1818,22 @@ DirectResourceGroupOption:
...
+	{
+		$$ = &ast.ResourceGroupOption{Tp: ast.ResourceBurstableOpiton, BoolValue: $3.(bool)}
+	}
 |	"QUERY_LIMIT" EqOpt '(' ResourceGroupRunawayOptionList ')'
 	{
 		$$ = &ast.ResourceGroupOption{Tp: ast.ResourceGroupRunaway, ResourceGroupRunawayOptionList: $4.([]*ast.ResourceGroupRunawayOption)}
 	}
+|	"QUERY_LIMIT" EqOpt '(' ')'
+	{
+		$$ = &ast.ResourceGroupOption{Tp: ast.ResourceGroupRunaway, ResourceGroupRunawayOptionList: make([]*ast.ResourceGroupRunawayOption, 0)}
```
glorv, line 1831:
```suggestion
		$$ = &ast.ResourceGroupOption{Tp: ast.ResourceGroupRunaway, ResourceGroupRunawayOptionList: nil}
```
I think in this case there is not difference between nil and a empty slice.

parser/parser.y
```diff
@@ -1818,10 +1818,22 @@ DirectResourceGroupOption:
...
+	}
+|	"QUERY_LIMIT" EqOpt "NULL"
+	{
+		$$ = &ast.ResourceGroupOption{Tp: ast.ResourceGroupRunaway, ResourceGroupRunawayOptionList: make([]*ast.ResourceGroupRunawayOption, 0)}
```
glorv, line 1835:
```suggestion
		$$ = &ast.ResourceGroupOption{Tp: ast.ResourceGroupRunaway, ResourceGroupRunawayOptionList: nil}
```

parser/model/model.go
CabinfeverB: > Seems this default value is useless

Because we do not support patch in the runaway setting clause, it means that the user must declare the configuration of action, but action is just an int and cannot judge nil like a pointer, so this default value can judge whether the user has declared action.

> RunawayNoneAction

Because there is no RunawayNoneAction in kvproto, I have made RunawayNoneAction equal to 3 to avoid semantic conversion between kvproto and tidb

parser/ast/ddl.go
CabinfeverB: forget to remove it

parser/model/model.go
glorv: > Because we do not support patch in the runaway setting clause, it means that the user must declare the configuration of action, but action is just an int and cannot judge nil like a pointer, so this default value can judge whether the user has declared action.

When action = RunawayNoneAction, we can just set to ResourceGroupSettings and vice verse, so there is even no need to add this action type here.

glorv (approved):

parser/ast/ddl.go
```diff
@@ -2160,11 +2163,14 @@ func (n *ResourceGroupOption) Restore(ctx *format.RestoreCtx) error {
 			ctx.WritePlain("= ")
 			ctx.WriteString(n.StrValue)
 		case ResourceBurstableOpiton:
-			ctx.WriteKeyWord("BURSTABLE")
+			ctx.WriteKeyWord("BURSTABLE ")
+			ctx.WritePlain("= ")
+			ctx.WritePlain(strings.ToUpper(fmt.Sprintf("%v", n.BoolValue)))
 		case ResourceGroupRunaway:
+			ctx.WritePlain("QUERY_LIMIT")
+			ctx.WritePlain(" = ")
```
Connor1996, line 2171: let's keep it consistent with others
```suggestion
			ctx.WritePlain("QUERY_LIMIT ")
			ctx.WritePlain("= ")
```

nolouch (approved):
lgtm

PR #44360, opened 2023-06-02
*: let handleMap supports partitionHandle

### LINKED ISSUE #44312: After exchange partition, the indexLookUp will return wrong result with keepOrder
## Bug Report

```
create table t (id int not null, store_id int not null )  partition by range (store_id) (partition p0 values less than (6), partition p1 values less than (11), partition p2 values less than (16), partition p3 values less than (21) );
create table t1(id int not null, store_id int not null);
insert into t values (1, 1);
insert into t values (2, 17);
insert into t1 values (0, 18);
alter table t exchange partition p3 with table t1;
alter table t add index idx(id);
analyze table t;
select *,_tidb_rowid from t use index(idx) order by id limit 2;
```

```
mysql> select *,_tidb_rowid from t use index(idx) order by id limit 2;
+----+----------+-------------+
| id | store_id | _tidb_rowid |
+----+----------+-------------+
|  0 |       18 |           1 |
|  1 |        1 |           1 |
+----+----------+-------------+
2 rows in set (0.00 sec)
```

```
mysql> select *,_tidb_rowid from t use index(idx) order by id limit 2;
+----+----------+-------------+
| id | store_id | _tidb_rowid |
+----+----------+-------------+
|  1 |        1 |           1 |
|  0 |       18 |           1 |
+----+----------+-------------+
2 rows in set (0.00 sec)
```

Defined2014:

Issue Number: close #44312

Problem Summary: After exchange partition, a partition table may have same `_tidb_rowid` in different partition tables. So the handleMap should supports partitionHandle.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/executor/partition_table_test.go
+++ b/executor/partition_table_test.go
@@ -821,6 +821,42 @@ func TestOrderByOnUnsignedPk(t *testing.T) {
 	tk.MustQuery("select max(a) from tunsigned_hash").Check(testkit.Rows("9279808998424041135"))
 }
 
+func TestPartitionHandleWithKeepOrder(t *testing.T) {
+	// https://github.com/pingcap/tidb/issues/44312
+	store := testkit.CreateMockStore(t)
+
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("create table t (id int not null, store_id int not null )" +
+		"partition by range (store_id)" +
+		"(partition p0 values less than (6)," +
+		"partition p1 values less than (11)," +
+		"partition p2 values less than (16)," +
+		"partition p3 values less than (21))")
+	tk.MustExec("create table t1(id int not null, store_id int not null)")
+	tk.MustExec("insert into t values (1, 1)")
+	tk.MustExec("insert into t values (2, 17)")
+	tk.MustExec("insert into t1 values (0, 18)")
+	tk.MustExec("alter table t exchange partition p3 with table t1")
+	tk.MustExec("alter table t add index idx(id)")
+	tk.MustExec("analyze table t")
+	tk.MustQuery("select *,_tidb_rowid from t use index(idx) order by id limit 2").Check(testkit.Rows("0 18 1", "1 1 1"))
+
+	tk.MustExec("drop table t, t1")
+	tk.MustExec("create table t (a int, b int, c int, key `idx_ac`(a, c), key `idx_bc`(b, c))" +
+		"partition by range (b)" +
+		"(partition p0 values less than (6)," +
+		"partition p1 values less than (11)," +
+		"partition p2 values less than (16)," +
+		"partition p3 values less than (21))")
+	tk.MustExec("create table t1 (a int, b int, c int, key `idx_ac`(a, c), key `idx_bc`(b, c))")
+	tk.MustExec("insert into t values (1,2,3), (2,3,4), (3,4,5)")
+	tk.MustExec("insert into t1 values (1,18,3)")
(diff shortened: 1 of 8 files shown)


kv/key.go
```diff
@@ -434,16 +465,32 @@ func (m *HandleMap) Set(h Handle, val interface{}) {
...
-		delete(m.ints, h.IntValue())
+		delete(ints, h.IntValue())
 	} else {
-		delete(m.strs, string(h.Encoded()))
+		delete(strs, string(h.Encoded()))
 	}
 }
 
 // Len returns the length of the map.
 func (m *HandleMap) Len() int {
```
guo-shaoge, line 485: Can `ints` and `partitionInts` merge into one? Otherwise `Len()` seems strage, because `ints` and `partitionInts` are exclusive.

planner/core/find_best_task.go
```diff
@@ -1635,6 +1635,12 @@ func (ds *DataSource) convertToIndexScan(prop *property.PhysicalProperty,
 	}
 	if candidate.isMatchProp {
 		cop.keepOrder = true
+		if cop.tablePlan != nil && !ds.tableInfo.IsCommonHandle {
+			col, isNew := cop.tablePlan.(*PhysicalTableScan).appendExtraHandleCol(ds)
+			cop.extraHandleCol = col
+			cop.needExtraProj = cop.needExtraProj || isNew
```
guo-shaoge, line 1641: Does this code related to this issue?

Defined2014: Yes, we should make sure `model.ExtraPhysTblID` in the last column. So move the code to the front.

kv/key.go
Defined2014: I think we could use `partitionInts[0]` to instead of `ints`, because `partitionID` shouldn't equal to 0. But i don't think that's good
1. an intHandle will do one more map lookup
2. implicitly assumes that partitionID will not be equal to 0, a little bit hack here

And I don't get why `Len()` function seems strange here. Because we could store different handle types in one `HandleMap`. Just like `TestHandleMapWithPartialHandle` test.

executor/partition_table_test.go
```diff
@@ -821,6 +821,42 @@ func TestOrderByOnUnsignedPk(t *testing.T) {
...
+	tk.MustExec("insert into t values (1, 1)")
+	tk.MustExec("insert into t values (2, 17)")
+	tk.MustExec("insert into t1 values (0, 18)")
+	tk.MustExec("alter table t exchange partition p3 with table t1")
+	tk.MustExec("alter table t add index idx(id)")
+	tk.MustExec("analyze table t")
+	tk.MustQuery("select *,_tidb_rowid from t use index(idx) order by id limit 2").Check(testkit.Rows("0 18 1", "1 1 1"))
+
+	tk.MustExec("drop table t, t1")
+	tk.MustExec("create table t (a int, b int, c int, key `idx_ac`(a, c), key `idx_bc`(b, c))" +
```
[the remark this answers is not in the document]
Defined2014: This cannot happen. If a table has a primary key, it should not have `_tidb_rowid` . And the primary key should be part of the partition key, so it cannot be duplicated in the partitioned table.

And this PR only changed `index_merge.tableScan`, did not modify `partialTableScan` related codes.

guo-shaoge:
rest lgtm

winoros (approved):

Issue #44314, opened 2023-05-31
Temporary incorrect query results occur after create table.


wd0517:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```
CREATE DATABASE `django_tests` /*!40100 DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci */;
USE `django_tests`;

CREATE TABLE `queries_identifier` (
   `id` int(11) NOT NULL AUTO_INCREMENT,
   `name` varchar(100) COLLATE utf8mb4_general_ci NOT NULL,
   PRIMARY KEY (`id`) /*T![clustered_index] CLUSTERED */
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;


 CREATE TABLE `queries_program` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `identifier_id` int(11) NOT NULL,
  PRIMARY KEY (`id`) /*T![clustered_index] CLUSTERED */,
  UNIQUE KEY `identifier_id` (`identifier_id`),
  CONSTRAINT `queries_program_identifier_id_70ff12a6_fk_queries_identifier_id` FOREIGN KEY (`identifier_id`) REFERENCES `django_tests`.`queries_identifier` (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;

CREATE TABLE `queries_channel` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `identifier_id` int(11) NOT NULL,
  PRIMARY KEY (`id`) /*T![clustered_index] CLUSTERED */,
  UNIQUE KEY `identifier_id` (`identifier_id`),
  CONSTRAINT `queries_channel_identifier_id_06ac3513_fk_queries_identifier_id` FOREIGN KEY (`identifier_id`) REFERENCES `django_tests`.`queries_identifier` (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;

INSERT  INTO queries_identifier(`id`, `name`) values(13, 'i1'), (14, 'i2'), (15, 'i3');
INSERT  INTO queries_program(`id`, `identifier_id`) values(8, 13), (9, 14);
INSERT  INTO queries_channel(`id`, `identifier_id`) values(5, 13);

[...]

PR #44409, opened 2023-06-05
planner: fix join reorder will append remained other condition to an outer join

AilinKid:

Issue Number: close https://github.com/pingcap/tidb/issues/44314

Problem Summary:

As mentioned in the last comment on the issue

Reason: outer join's other condition influence is quite different from inner join's other condition, the latter one functions as a simple filter, just integrated with inner join. The former functions as a simple filter but also decides whether the outer join is matched or not, consequently deciding whether appending an additional null (which means this row itself is not pruned in the final result)

So when enumerating join order when encountering inner join combined with outer join, those original other conditions bonded with an outer join, just let it be; those additional remained other conditions shouldn't be put down into it, adding additional selection above is a resolution.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
planner: fix join reorder will append remained condition to an outer join
```

### DIFF
--- a/planner/core/rule_join_reorder.go
+++ b/planner/core/rule_join_reorder.go
@@ -507,6 +507,13 @@ func (s *baseSingleGroupJoinOrderSolver) makeJoin(leftPlan, rightPlan LogicalPla
 	remainOtherConds, otherConds = expression.FilterOutInPlace(remainOtherConds, func(expr expression.Expression) bool {
 		return expression.ExprFromSchema(expr, mergedSchema)
 	})
+	if (joinType.JoinType == LeftOuterJoin || joinType.JoinType == RightOuterJoin) && len(otherConds) > 0 {
+		// the original outer join's other conditions has been bound to the outer join Edge,
+		// these remained other condition here shouldn't be appended to it because on-mismatch
+		// logic will produce more append-null rows which is banned in original semantic.
+		remainOtherConds = append(remainOtherConds, otherConds...)
+		otherConds = otherConds[:0]
+	}
 	if len(joinType.outerBindCondition) > 0 {
 		remainOBOtherConds := make([]expression.Expression, len(joinType.outerBindCondition))
 		copy(remainOBOtherConds, joinType.outerBindCondition)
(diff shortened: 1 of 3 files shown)


winoros (approved):

planner/core/rule_join_reorder.go
```diff
@@ -507,6 +507,13 @@ func (s *baseSingleGroupJoinOrderSolver) makeJoin(leftPlan, rightPlan LogicalPla
 	remainOtherConds, otherConds = expression.FilterOutInPlace(remainOtherConds, func(expr expression.Expression) bool {
 		return expression.ExprFromSchema(expr, mergedSchema)
 	})
+	if (joinType.JoinType == LeftOuterJoin || joinType.JoinType == RightOuterJoin) && len(otherConds) > 0 {
```
elsa0520, line 510: How about left outer semi join and anti left outer ？

AilinKid: good

planner/core/rule_join_reorder.go
```diff
@@ -507,6 +507,13 @@ func (s *baseSingleGroupJoinOrderSolver) makeJoin(leftPlan, rightPlan LogicalPla
 	remainOtherConds, otherConds = expression.FilterOutInPlace(remainOtherConds, func(expr expression.Expression) bool {
 		return expression.ExprFromSchema(expr, mergedSchema)
 	})
+	if (joinType.JoinType == LeftOuterJoin || joinType.JoinType == RightOuterJoin || joinType.JoinType == LeftOuterSemiJoin || joinType.JoinType == AntiLeftOuterSemiJoin) && len(otherConds) > 0 {
```
chrysan, line 510: What about fullOuter?

planner/core/rule_join_reorder_test.go
```diff
@@ -69,6 +69,34 @@ func TestJoinOrderHintWithBinding(t *testing.T) {
...
+
+	tk.MustExec("INSERT  INTO queries_identifier(`id`, `name`) values(13, 'i1'), (14, 'i2'), (15, 'i3');")
+	tk.MustExec("INSERT  INTO queries_program(`id`, `identifier_id`) values(8, 13), (9, 14);")
+	tk.MustExec("INSERT  INTO queries_channel(`id`, `identifier_id`) values(5, 13);")
+
+	tk.MustQuery("SELECT `queries_identifier`.`id`, `queries_identifier`.`name` FROM `queries_identifier` LEFT OUTER JOIN `queries_channel` ON (`queries_identifier`.`id` = `queries_channel`.`identifier_id`) INNER JOIN `queries_program` ON (`queries_identifier`.`id` = `queries_program`.`identifier_id`) WHERE ((`queries_channel`.`id` = 5 AND `queries_program`.`id` = 9) OR `queries_program`.`id` = 8) ORDER BY `queries_identifier`.`id` ASC;").Check(testkit.Rows("" +
```
chrysan, line 85: Do we need more tests to cover different join types?

planner/core/rule_join_reorder.go
AilinKid: no full outer join type interally

elsa0520 (approved):
LGTM

chrysan (approved):

Issue #41711, opened 2023-02-24
Support `CHECK` constraint


CbcWestwolf:
https://dev.mysql.com/doc/refman/8.0/en/create-table-check-constraints.html

Benjamin2037:
Do we have same behavior as MySQL on check constraint?

fzzf678:
> Do we have same behavior as MySQL on check constraint?

Mostly same, I will supply unsupported behavior in doc later

dveeden:
Looks like this has been implemented in v7.2. Can we close this issue?

```sql
mysql> SELECT TIDB_VERSION()\G
*************************** 1. row ***************************
TIDB_VERSION(): Release Version: v7.2.0
Edition: Community
Git Commit Hash: 9fd5f4a8e4f273a60fbe7d3848f85a1be8f0600b
Git Branch: heads/refs/tags/v7.2.0
UTC Build Time: 2023-06-27 15:04:42
GoVersion: go1.20.5
Race Enabled: false
Check Table Before Drop: false
Store: tikv
1 row in set (0.00 sec)

mysql> SET GLOBAL tidb_enable_check_constraint=ON;
Query OK, 0 rows affected (0.04 sec)

mysql> CREATE TABLE t1 (id INT PRIMARY KEY, CHECK (id > 10), CHECK (id < 20));
Query OK, 0 rows affected (0.20 sec)

mysql> INSERT INTO t1 VALUES (1);
ERROR 3819 (HY000): Check constraint 't1_chk_1' is violated.
mysql> INSERT INTO t1 VALUES (15);
Query OK, 1 row affected (0.02 sec)

mysql> INSERT INTO t1 VALUES (25);
ERROR 3819 (HY000): Check constraint 't1_chk_2' is violated.
```

CbcWestwolf:
@dveeden  Sure!

shawn0915:
@CbcWestwolf could you help confirm is that a bug? #47567

PR #44455, opened 2023-06-06
*: fix check constraintInfo state change

fzzf678:

Issue Number: ref #41711

Problem Summary:

In the process of changing the schemaState of the check constraint, it is stateNone- > stateWriteOnly - > statePublic. The remaining data will be verified during `sateWriteOnly`. At this time, if there is a tidb node still in `stateNone` state, it is possible to insert data that violates the check constraint. 
To avoid this problem,  I added a `stateWriteReorg` state after the `stateWriteOnly `state, where the remaining data is checked. Make sure other tidb nodes are at least in `stateWriteOnly` state
### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


ddl/constraint.go
```diff
@@ -86,15 +87,25 @@ func (w *worker) onAddCheckConstraint(d *ddlCtx, t *meta.Meta, job *model.Job) (
...
-		// write only -> public
-		err = w.verifyRemainRecordsForCheckConstraint(dbInfo, tblInfo, constraintInfoInMeta, job)
-		if err != nil {
-			return ver, errors.Trace(err)
+		job.SchemaState = model.StateWriteReorganization
+		constraintInfoInMeta.State = model.StateWriteReorganization
+		ver, err = updateVersionAndTableInfoWithCheck(d, t, job, tblInfo, originalState != constraintInfoInMeta.State)
+	case model.StateWriteReorganization:
+		skipCheck := false
+		failpoint.Inject("mockVerifyRemainDataSuccess", func(val failpoint.Value) {
```
zimulala, line 99: Maybe we can put it to `verifyRemainRecordsForCheckConstraint`, and if `mockVerifyRemainDataSuccess` is ture, we can return this function. Then we needn't add `skipCheck`.

ddl/constraint.go
```diff
@@ -86,15 +87,25 @@ func (w *worker) onAddCheckConstraint(d *ddlCtx, t *meta.Meta, job *model.Job) (
...
 		ver, err = updateVersionAndTableInfoWithCheck(d, t, job, tblInfo, originalState != constraintInfoInMeta.State)
 	case model.StateWriteOnly:
-		// write only -> public
-		err = w.verifyRemainRecordsForCheckConstraint(dbInfo, tblInfo, constraintInfoInMeta, job)
-		if err != nil {
-			return ver, errors.Trace(err)
+		job.SchemaState = model.StateWriteReorganization
+		constraintInfoInMeta.State = model.StateWriteReorganization
+		ver, err = updateVersionAndTableInfoWithCheck(d, t, job, tblInfo, originalState != constraintInfoInMeta.State)
+	case model.StateWriteReorganization:
```
zimulala, line 97: Why we add a new state for this operation？

fzzf678: Because I want to make sure that the state of all tidb nodes is changed to `StateWriteOnly` before doing remaining data verification, so that no data is written that violates the check constraint.
 If doing stock data verification in `StateWriteOnly` state, the state of some nodes may be still in `StateNone `( am i right?), so even if the remaining data verification passes, there are still `StateNone` nodes that can write data that violates the check constraint. Hence I added a state to ensure that all nodes change at least to `StateWriteOnly` before verifying the data. Is this right or not, please give some advice, thks

zimulala: Ok, there was no PR description before, I don't know why it suddenly added a status, so ask.

ddl/constraint.go
```diff
@@ -205,29 +206,50 @@ func checkDropCheckConstraint(t *meta.Meta, job *model.Job) (*model.TableInfo, *
...
-func (w *worker) onAlterCheckConstraint(d *ddlCtx, t *meta.Meta, job *model.Job) (ver int64, _ error) {
+func (w *worker) onAlterCheckConstraint(d *ddlCtx, t *meta.Meta, job *model.Job) (ver int64, err error) {
 	dbInfo, tblInfo, constraintInfo, enforced, err := checkAlterCheckConstraint(t, job)
 	if err != nil {
 		return ver, errors.Trace(err)
 	}
 
 	// enforced will fetch table data and check the constraint.
-	if constraintInfo.Enforced != enforced && enforced {
-		err = w.verifyRemainRecordsForCheckConstraint(dbInfo, tblInfo, constraintInfo, job)
```
fzzf678, line 216: In the check constraint, when changing from `not enforced` to `enforced`, remaining data verification is required. All nodes in tidb may have two states, `enforced` and `not enforced`. After the remaining data is verified, data that violates the check constraint may also be written. Therefore, state changes are added to ensure that all nodes have reached the `enforced` state before remaining data verification.

TonsnakeLin (approved):
LGTM

ddl/constraint.go
Benjamin2037: Seems here we missed delete-only state, that is before writeonly state.  write reorg means will reorg table data.

Issue #44158, opened 2023-05-25
Upgrade failure Re-upgrade may cause paused system DDL job can't be executed or dirty data of DDL job


zimulala:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

Mock： do upgradeNewVer： Add a new mysql.table(New TiDB upgrade failed, and the DDL is queuing, such as TiKV network problems)
Restart the TiDB. And the TiDB do `ResumeAllJobsBySystem` failed.

### 2. What did you expect to see? (Required)
The upgrade succeeds. 

### 3. What did you see instead (Required)
The upgrade may fail because of paused system DDL.  But as a DDL job of `Add a new mysql.table` may be paused in DDL table

### 4. What is your TiDB version? (Required)

master

Issue #44366, opened 2023-06-02
[upgrading] get owner op failed led tidb can not start


aytrack:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
1. deploy v7.1.0 cluster with tiup cluster
2. use sysbench prepare data
3. using tiup cluster upgrade the cluster to nightly 

### 2. What did you expect to see? (Required)
upgrade success
### 3. What did you see instead (Required)
ugrade failed
```log
[2023/06/02 16:35:24.772 +08:00] [WARN] [bootstrap.go:1188] ["[upgrading] get owner op failed"] [state=none] []
[2023/06/02 16:35:24.973 +08:00] [INFO] [manager.go:344] ["get owner"] ["owner info"="upgrade bootstrap"] ["owner key"=/tidb/ddl/fg/owner/db4887b356a3f6d] [ownerID=c4a4c6ff-e140-4248-9414-c80a1fa74703] [op=none]
[2023/06/02 16:35:24.973 +08:00] [FATAL] [bootstrap.go:1186] ["[upgrading] get owner op failed"] [state=none] [] [stack="github.com/pingcap/tidb/session.syncUpgradeState
	/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/session/bootstrap.go:1186
github.com/pingcap/tidb/session.upgrade
	/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/session/bootstrap.go:1125
github.com/pingcap/tidb/session.runInBootstrapSession
	/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/session/session.go:3540
github.com/pingcap/tidb/session.BootstrapSession
	/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/session/session.go:3317
main.createStoreAndDomain
	/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/tidb-server/main.go:347
main.main
	/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/tidb-server/main.go:241
runtime.main
	/usr/local/go/src/runtime/proc.go:250"]
```
### 4. What is your TiDB version? (Required)
```log
 ["Welcome to TiDB."] ["Release Version"=v7.1.0] [Edition=Community] ["Git Commit Hash"=635a4362235e8a3c0043542e629532e3c7bb2756]
[...]

PR #44469, opened 2023-06-07
ddl, session: fix re-upgrade issues

zimulala:

Issue Number: close https://github.com/pingcap/tidb/issues/44158, and related https://github.com/pingcap/tidb/issues/44366

Problem Summary:
Mock： do upgradeNewVer： Add a new mysql.table(mock system DDL)

**case 1**
1.  The new TiDB upgrade failed(such as TiKV network problems), **and the mock system DDL is queuing**
2.  Re-upgrade the TiDB(This operation will pause all DDL jobs by the system). It will pause the mock system DDL which always goes into `runDDLJob`.

**case 2**
1. The new TiDB upgrade failed(such as TiKV network problems), **and the etcd owner path isn't remove**
2. The etcd owner path TTL is 60s, we retry get owner info is 2s. So we can't get the owner's info normally.

* Make all paused DDL jobs can't be runnable(won't go into `runDDLJob`) when upgrading state.
* Extend the timeout to get owner's info.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
case 2.
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


session/mock_bootstrap.go
```diff
@@ -140,13 +155,20 @@ func modifyBootstrapVersionForTest(store kv.Storage, ver int64) int64 {
 	return ver
 }
 
+// MockUpgradeToVerLatestKind is used to indicate the use of different mock bootstrapVersion.
+var MockUpgradeToVerLatestKind = 1
```
Defined2014, line 159: Split it to two vars like `mockUpgradeToVerLatest` and `mockSimpleUpgradeToVerLatest` is better

Defined2014 (commented):
Could update `job2SchemaNames` function 

```diff
--- a/ddl/job_table.go
+++ b/ddl/job_table.go
@@ -561,12 +561,12 @@ func job2SchemaNames(job *model.Job) []string {
                names = append(names, strings.ToLower(job.SchemaName))
                names = append(names, oldSchemaName.O)
                return names
-       case model.ActionRenameTables:
-               // TODO: Get this action's schema names.
-       case model.ActionExchangeTablePartition:
+       case model.ActionRenameTables, model.ActionExchangeTablePartition:
                // TODO: Get this action's schema names.
+               return nil
+       default:
+               return []string{job.SchemaName}
        }
-       return []string{job.SchemaName}
 }
```

session/bootstraptest/bootstrap_upgrade_test.go
```diff
@@ -363,6 +363,80 @@ func TestUpgradeVersionForPausedJob(t *testing.T) {
 	require.True(t, suc)
 }
 
+func TestUpgradeVersionForSystemPausedJob(t *testing.T) {
```
dhysum, line 366: Shall we add some comment on the case description here?

zimulala: Added.

session/bootstraptest/bootstrap_upgrade_test.go
```diff
@@ -363,6 +363,82 @@ func TestUpgradeVersionForPausedJob(t *testing.T) {
...
+	defer domLatestV.Close()
+	seLatestV := session.CreateSessionAndSetID(t, store)
+	ver, err = session.GetBootstrapVersion(seLatestV)
+	require.NoError(t, err)
+	require.Equal(t, session.CurrentBootstrapVersion+1, ver)
+
+	sql := fmt.Sprintf(" admin show ddl jobs where job_id=%d", jobID)
+	// Make sure the add index operation is successful.
+	suc := false
+	for i := 0; i < 20; i++ {
```
Defined2014, line 426: Could we extract it to a function? I think it also used in other tests.

Defined2014 (approved):

session/bootstrap.go
```diff
@@ -1186,7 +1187,9 @@ func syncUpgradeState(s Session) {
 		if i == retryTimes-1 {
```
dhysum, line 1187: Normally, crash inside a loop is not a good practice, which we may involve more action on the crash later on.

We could do it like :
```
isSuccess = true
for ... {
    ...
    if err == nil ... {
        break
    }

    isSuccess = false
}

if !isSuccess {
    // fatal actions ...   
    ...
    
    logutil.BgLogger().Fatal...
}

```

dhysum: No big deal though.

ddl/job_table.go
```diff
@@ -551,8 +551,7 @@ func job2UniqueIDs(job *model.Job, schema bool) string {
 }
 
 func job2SchemaNames(job *model.Job) []string {
-	switch job.Type {
-	case model.ActionRenameTable:
+	if job.Type == model.ActionRenameTable {
```
dhysum, line 554: Good one, more clear.

ddl/job_table.go
```diff
@@ -562,11 +561,9 @@ func job2SchemaNames(job *model.Job) []string {
 		names = append(names, strings.ToLower(job.SchemaName))
 		names = append(names, oldSchemaName.O)
 		return names
-	case model.ActionRenameTables:
-		// TODO: Get this action's schema names.
-	case model.ActionExchangeTablePartition:
-		// TODO: Get this action's schema names.
 	}
+	// TODO: consider about model.ActionRenameTables and model.ActionExchangeTablePartition, which need to get the schema names.
```
dhysum, line 565: Better one.

If not necessary, we'd better not add the code. "// TODO:" is a option.

It would be much better to talk about such kind of potential possibility in the design procedure, as much as possible.

session/bootstraptest/bootstrap_upgrade_test.go
```diff
@@ -363,6 +367,66 @@ func TestUpgradeVersionForPausedJob(t *testing.T) {
...
+	hook.OnJobRunAfterExported = func(job *model.Job) {
+		if job.SchemaState == model.StateDeleteOnly {
+			se := session.CreateSessionAndSetID(t, store)
+			session.MustExec(t, se, fmt.Sprintf("admin pause ddl jobs %d", job.ID))
+		}
+		if job.State == model.JobStatePaused && jobID == 0 {
+			// Mock pause the ddl job by system.
+			job.AdminOperator = model.AdminCommandBySystem
+			ch <- struct{}{}
+			jobID = job.ID
```
dhysum, line 408: jobID is multi-corouting variable. Should be atomic or controlled by mutex or channel.

session/bootstraptest/bootstrap_upgrade_test.go
```diff
@@ -363,6 +367,66 @@ func TestUpgradeVersionForPausedJob(t *testing.T) {
...
+	store, dom := session.CreateStoreAndBootstrap(t)
+	defer func() { require.NoError(t, store.Close()) }()
+
+	seV := session.CreateSessionAndSetID(t, store)
+	txn, err := store.Begin()
+	require.NoError(t, err)
+	m := meta.NewMeta(txn)
+	err = m.FinishBootstrap(session.CurrentBootstrapVersion - 1)
+	require.NoError(t, err)
+	err = txn.Commit(context.Background())
```
dhysum, line 386: Shall we use the `session` but not `txn` here ?

session/bootstraptest/bootstrap_upgrade_test.go
```diff
@@ -363,6 +367,66 @@ func TestUpgradeVersionForPausedJob(t *testing.T) {
...
+	var jobID int64
+	hook := &callback.TestDDLCallback{}
+	hook.OnJobRunAfterExported = func(job *model.Job) {
+		if job.SchemaState == model.StateDeleteOnly {
+			se := session.CreateSessionAndSetID(t, store)
+			session.MustExec(t, se, fmt.Sprintf("admin pause ddl jobs %d", job.ID))
+		}
+		if job.State == model.JobStatePaused && jobID == 0 {
+			// Mock pause the ddl job by system.
+			job.AdminOperator = model.AdminCommandBySystem
```
dhysum, line 406: It looks like that would be better to do the real upgrade but not mock since it will be more like user's action.

And, the `mock` here normally seems being a litlle bit complex since there are cases that we may not know what going to happen in the future because the behavior of 'job.AdminOperator' may be changed by its module. In another word, the case here involves the inside-logic of 'admin pause' which may introduce unknow error in future.

session/mock_bootstrap.go
```diff
@@ -112,6 +112,21 @@ func mockUpgradeToVerLatest(s Session, ver int64) {
...
 
+// mockSimpleUpgradeToVerLatest mocks a simple bootstrapVersion(make the test faster).
+func mockSimpleUpgradeToVerLatest(s Session, ver int64) {
+	logutil.BgLogger().Info("mock upgrade to ver latest", zap.Int64("old ver", ver), zap.Int64("mock latest ver", mockLatestVer))
+	if ver >= mockLatestVer {
+		return
+	}
+	mustExecute(s, "use mysql")
+	mustExecute(s, `create table if not exists mock_sys_t(
+		c1 int, c2 int, c3 int, c11 tinyint, index fk_c1(c1)
```
dhysum, line 123: It would be better that different column is with different kind of type

session/mock_bootstrap.go
```diff
@@ -140,13 +155,26 @@ func modifyBootstrapVersionForTest(store kv.Storage, ver int64) int64 {
 	return ver
 }
 
+const (
+	defaultMockUpgradeToVerLatest = 0
+	// MockSimpleUpgradeToVerLatest is used to indicate the use of the simple mock bootstrapVersion.
+	MockSimpleUpgradeToVerLatest = 1
```
dhysum, line 161: What does it mean 'Simple' here? The function 'mockSimpleUpgradeToVerLatest'?

And, the function and the enum are sharing the same name seems not a good idea.

session/mock_bootstrap.go
```diff
@@ -140,13 +155,26 @@ func modifyBootstrapVersionForTest(store kv.Storage, ver int64) int64 {
...
 
 	TestHook.OnBootstrapBefore(s)
-	bootstrapVersion = append(bootstrapVersion, mockUpgradeToVerLatest)
+	if MockUpgradeToVerLatestKind == defaultMockUpgradeToVerLatest {
+		bootstrapVersion = append(bootstrapVersion, mockUpgradeToVerLatest)
+	} else {
+		bootstrapVersion = append(bootstrapVersion, mockSimpleUpgradeToVerLatest)
+	}
 	currentBootstrapVersion++
 }
```
dhysum, line 179: LGTM

session/mock_bootstrap.go
zimulala: This is a simple test, hopefully a system DDL job. I don't feel the need to consider these types.

session/bootstraptest/bootstrap_upgrade_test.go
zimulala: This is because this method of `FinishBootstrap` call requires `txn`.

session/bootstraptest/bootstrap_upgrade_test.go
zimulala: Yes, but the test should fail if the internal implementation changes, so just modify the test. And the way I've come up with mocks so far is this. Do you have any other suggestions?

session/bootstrap.go
zimulala: Okay, I won't change it.

ddl/job_table.go
zimulala: Yeah, but here I think it needs to be added.

session/mock_bootstrap.go
zimulala: Comments were added. Another name has a good suggestion can be put forward, I change it

hawkingrei (approved):

XuHuaiyu (approved):

Issue #44487, opened 2023-06-07
support grouping function/col/expression rewriting and physical plan exhaustion for rollup expand OP


AilinKid:
## Feature Request

**Is your feature request related to a problem? Please describe:**
<!-- A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] -->
support grouping function/col/expression rewriting and physical plan exhaustion for rollup expand OP

https://github.com/pingcap/tidb/pull/44216  logical expand support
https://github.com/pingcap/tidb/pull/44436  grouping function refactor

Now, after the above two PR is merged, now we can support transforming a ast rollup syntax to a Expand logical plan.
There is something important that needs to be done before we step into the next physical plan generation --- that's rewriting.

* grouping function needs to be rewritten. what the user typed in the SQL layer is the raw definition of the grouping function which receives multi-raw columns ref as its parameters with a maximum number of up to 64. While in its runtime, what she only considered is the gid(uint64) column, getting the gid uint64 value from every row and comparing it with the meta filled in the rewriting time according to some rule regulated with the grouping mode.

```
case: select grouping(a), grouping(b,a) from t group by a, b with rollup.
                      |             |
                      |             +>.  grouping(gid) with [meta1, meta2]
                      +>    grouping(gid) with meta
```

* except for the grouping function, normal columns or expressions should be considered to be rewritten as grouping set columns or expressions. As we talked about in the previous pull request, the gby expressions will be projected out in one another projection under expand even if it's simple columns already. We do this because gby item will be filled with a null value when we replicate the source rows to distinguish data for different grouping layouts, so it's not a real column ref yet! (column ref doesn't change anything from the source column, including the field type). So is there any chance we should use the source column rather than the copied-changed grouping column? Yes, any column args in an aggregation should use the source column ref, otherwise, use the copied-changed grouping column instead!

```
eg： select   sum(a),  a,  a+1,  grouping(b,a) from t group by a, b with rollup.
                |      |     |             +>   absolutely resolve the grouping set col a and b.
                |      |     +>  resolve to the grouping set col a'#3. (it can show NULL anyway)
                |     +> resolve to the grouping set col a'#3. (it can show NULL anyway)
                +> resolve to the base column a.

logical plan:   
                projection: col#9, a'#3, a'#3+1, grouping(gid#5, metas)
                     aggregation: sum(a) -> col#9
                         expand:   [a, b, a'#3, b'#4, gid#5]
[...]

PR #44436, opened 2023-06-06
expression: refactor grouping function computation and update related tipb

AilinKid:
### What problem does this PR solve?

Issue Number: close https://github.com/pingcap/tidb/issues/44437

related tiflash issue : https://github.com/pingcap/tiflash/issues/7590

related tipb pr: https://github.com/pingcap/tipb/pull/308

Problem Summary:

add one more dimension for grouping marks maintenance.

ideas come from MySQL's support for grouping(a,b,c,...) with maximum of 64 parameters.

```
mysql> select grouping(a,b), grouping(a), grouping(b) from t group by a, b with rollup;
+---------------+-------------+-------------+
| grouping(a,b) | grouping(a) | grouping(b) |
+---------------+-------------+-------------+
|             0 |           0 |           0 |
|             1 |           0 |           1 |
|             0 |           0 |           0 |
|             1 |           0 |           1 |
|             3 |           1 |           1 |
+---------------+-------------+-------------+
5 rows in set (0.00 sec)
mysql> select grouping(b,a), grouping(a), grouping(b) from t group by a, b with rollup;
+---------------+-------------+-------------+
| grouping(b,a) | grouping(a) | grouping(b) |
+---------------+-------------+-------------+
|             0 |           0 |           0 |
|             2 |           0 |           1 |
|             0 |           0 |           0 |
|             2 |           0 |           1 |
|             3 |           1 |           1 |
+---------------+-------------+-------------+
5 rows in set (0.00 sec)
```
Notice the result difference of grouping(a,b) and grouping(b,a)!

#### mysql reference
As seen here, if GROUPING (a,b) returns 3, it means that NULL in column “a” and NULL in column “b” for that row is produce by a ROLLUP operation. If result is 1, NULL in column “b” alone is a result of ROLLUP operation.
https://dev.mysql.com/blog-archive/mysql-8-0-grouping-function/

### when evaluating the grouping function's value, the formula is:
 
grouping(a,b,c) = grouping(a)<<2 + grouping(b)<<1 << grouping(c)

the left shift size is exactly according to their listed position in the grouping function.
so we should maintain all the grouping marks (for every single col) down, and add one more round computation for the left shift plus if the len(groupingMarks) is greater than 1 (means multi col args in grouping function).

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
[...]

PR #44488, opened 2023-06-07
planner: support grouping function/col/expression rewriting and physical plan exhaustion for rollup expand OP

AilinKid:
<!--

Thank you for contributing to TiDB!

PR Title Format:
1. pkg [, pkg2, pkg3]: what's changed
2. *: what's changed

-->

<!--

There MUST be one line starting with "Issue Number:  " and 
linking the relevant issues via the "close" or "ref".

For more info, check https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/contribute-code.html#referring-to-an-issue.

-->

Issue Number: close https://github.com/pingcap/tidb/issues/44487

Problem Summary:

https://github.com/pingcap/tidb/pull/44216  logical expand support
https://github.com/pingcap/tidb/pull/44436  grouping function refactor

Now, after the above two PR is merged, now we can support transforming a ast rollup syntax to a Expand logical plan.
There is something important that needs to be done before we step into the next physical plan generation --- that's rewriting.

* grouping function needs to be rewritten. what the user typed in the SQL layer is the raw definition of the grouping function which receives multi-raw columns ref as its parameters with a maximum number of up to 64. While in its runtime, what she only considered is the gid(uint64) column, getting the gid uint64 value from every row and comparing it with the meta filled in the rewriting time according to some rule regulated with the grouping mode.

```
case: select grouping(a), grouping(b,a) from t group by a, b with rollup.
                      |             |
                      |             +>.  grouping(gid) with [meta1, meta2]
                      +>    grouping(gid) with meta
```

* except for the grouping function, normal columns or expressions should be considered to be rewritten as grouping set columns or expressions. As we talked about in the previous pull request, the gby expressions will be projected out in one another projection under expand even if it's simple columns already. We do this because gby item will be filled with a null value when we replicate the source rows to distinguish data for different grouping layouts, so it's not a real column ref yet! (column ref doesn't change anything from the source column, including the field type). So is there any chance we should use the source column rather than the copied-changed grouping column? Yes, any column args in an aggregation should use the source column ref, otherwise, use the copied-changed grouping column instead!

```
eg： select   sum(a),  a,  a+1,  grouping(b,a) from t group by a, b with rollup.
                |      |     |             +>   absolutely resolve the grouping set col a and b.
                |      |     +>  resolve to the grouping set col a'#3. (it can show NULL anyway)
                |     +> resolve to the grouping set col a'#3. (it can show NULL anyway)
                +> resolve to the base column a.

logical plan:   
                projection: col#9, a'#3, a'#3+1, grouping(gid#5, metas)
                     aggregation: sum(a) -> col#9
                         expand:   [a, b, a'#3, b'#4, gid#5]
                             + projection:  [a, b, a -> a'#3, b -> b'#4]
```
* After comparison between spark and MySQL, the former has a more analyzing ability, cases like below can even be resolved.
```
eg： select   sum(a),  a+1,  grouping(1+a,b) from t group by 1+a, b with rollup.
                |      |                 +>   absolutely resolve the grouping set col a+1 (A#3) and b'#4.
                |      |     
                |      +> resolve to the grouping set ** Expression ** a+1, here we should substitute it with A#3. 
                +> resolve to the base column a.

logical plan:   
                projection: col#9, A#3,  grouping(gid#5, metas)
                     aggregation: sum(a) -> col#9
                         expand:   [a, b, A#3, b'#4, gid#5]
                             + projection:  [a, b, a+1 -> A#3, b -> b'#4]
```
* grouping function rewriting and grouping column/expression-related rewriting can exist in select-list fields/having/order clause. Given the complexity of rewriting them in having/order clause, this PR currently focus on the rewriting of them in the select-list fields.
1. we rewrite the grouping function in place when we first meet/rewrite it in the expressionWriter.
2. we rewrite the grouping expressions/columns after select Expr is built, then traverse down the expression tree to find the shallow-most scope for substitution.  (how to understand shallow-most, let's say we have a field like a+1, then we got two group by item (a(a#3), a+1 (A#4)) with a rollup, the a+1 should be resolved to ** bigger** one a+1 rather than single expression a)
```
          
   +     find gby item using scalar plus function hashcode (shadow layer in tree),  final substitution as  A#4         
  /  \                                                                                                                                 
a     1                                                                                                                    
                                                                                                 
   +       
  /  \                                                                                                                                 
a     1  find gby item using a hashcode (deep layer in tree), final substitution as a'#3 + 1
  

both cases is reasonable, but actually, we need case1: substitute a+1 completely as gby item as A#4
```

* After rewriting is done, then, we put the logical plan tree through the logical optimization rules, at the final rule `resolveExpand` we generate the level projections for this logical Expand. Then, enumerate the possible physical plan for logical Expand, we should keep in mind that currently we only have the MPP Expand Execution mode, so only MPP task type can be generated by now.


### Check List

Tests <!-- At least one of them must be included. -->

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

<!-- compatibility change, improvement, bugfix, and new feature need a release note -->

```release-note
planner: support grouping function/col rewriting and physical plan exhausion for rollup expand OP
```

### DIFF
--- a/expression/expression.go
+++ b/expression/expression.go
@@ -97,6 +97,16 @@ type ReverseExpr interface {
 	ReverseEval(sc *stmtctx.StatementContext, res types.Datum, rType types.RoundingType) (val types.Datum, err error)
 }
 
+// TraverseDown implementing a pre-order traverse for expression tree.
+type TraverseDown interface {
+	Traverse(TraverseAction) Expression
+}
+
+// TraverseAction define the interface for action when traversing down an expression.
+type TraverseAction interface {
+	Transform(Expression) Expression
+}
+
 // Expression represents all scalar expression in SQL.
 type Expression interface {
 	fmt.Stringer
@@ -104,6 +114,7 @@ type Expression interface {
 	VecExpr
 	ReverseExpr
 	CollationInfo
+	TraverseDown
 
 	// Eval evaluates an expression through a row.
 	Eval(row chunk.Row) (types.Datum, error)
@@ -1268,6 +1279,8 @@ func scalarExprSupportedByFlash(function *ScalarFunction) bool {
 	case ast.GetFormat:
 		return true
 	case ast.IsIPv4, ast.IsIPv6:
+		return true
+	case ast.Grouping: // grouping function for grouping sets identification.
 		return true
 	}
 	return false
--- a/expression/expression.go
+++ b/expression/expression.go
@@ -97,11 +97,6 @@ type ReverseExpr interface {
 	ReverseEval(sc *stmtctx.StatementContext, res types.Datum, rType types.RoundingType) (val types.Datum, err error)
 }
 
-// TraverseDown implementing a pre-order traverse for expression tree.
-type TraverseDown interface {
-	Traverse(TraverseAction) Expression
-}
-
 // TraverseAction define the interface for action when traversing down an expression.
 type TraverseAction interface {
 	Transform(Expression) Expression
@@ -114,7 +109,8 @@ type Expression interface {
 	VecExpr
 	ReverseExpr
 	CollationInfo
-	TraverseDown
+
+	Traverse(TraverseAction) Expression
 
 	// Eval evaluates an expression through a row.
 	Eval(row chunk.Row) (types.Datum, error)
--- a/planner/core/casetest/testdata/enforce_mpp_suite_out.json
+++ b/planner/core/casetest/testdata/enforce_mpp_suite_out.json
@@ -1924,5 +1924,121 @@
         ]
       }
     ]
+  },
+  {
+    "Name": "TestRollupMPP",
+    "Cases": [
+      {
+        "SQL": "explain format = 'brief' select count(1) from t group by a, b with rollup; -- 1. simple agg",
+        "Plan": [
+          "TableReader 8000.00 root  MppVersion: 1, data:ExchangeSender",
+          "└─ExchangeSender 8000.00 mpp[tiflash]  ExchangeType: PassThrough",
+          "  └─Projection 8000.00 mpp[tiflash]  Column#8",
+          "    └─HashAgg 8000.00 mpp[tiflash]  group by:Column#5, Column#6, gid, funcs:count(1)->Column#8",
+          "      └─ExchangeReceiver 10000.00 mpp[tiflash]  ",
+          "        └─ExchangeSender 10000.00 mpp[tiflash]  ExchangeType: HashPartition, Compression: FAST, Hash Cols: [name: Column#5, collate: binary], [name: Column#6, collate: binary], [name: gid, collate: binary]",
+          "          └─Expand 10000.00 mpp[tiflash]  level-projection:[<nil>->Column#5, <nil>->Column#6, 0->gid],[Column#5, <nil>->Column#6, 1->gid],[Column#5, Column#6, 3->gid]; schema: [Column: [Column#5,Column#6,gid] Unique key: []]",
+          "            └─TableFullScan 10000.00 mpp[tiflash] table:t keep order:false, stats:pseudo"
+        ],
+        "Warn": null
+      },
+      {
+        "SQL": "explain format = 'brief' select sum(c), count(1) from t group by a, b with rollup; -- 2. non-grouping set col c",
+        "Plan": [
+          "TableReader 8000.00 root  MppVersion: 1, data:ExchangeSender",
+          "└─ExchangeSender 8000.00 mpp[tiflash]  ExchangeType: PassThrough",
+          "  └─Projection 8000.00 mpp[tiflash]  Column#8, Column#9",
+          "    └─HashAgg 8000.00 mpp[tiflash]  group by:Column#17, Column#18, Column#19, funcs:sum(Column#16)->Column#8, funcs:count(1)->Column#9",
+          "      └─Projection 10000.00 mpp[tiflash]  cast(test.t.c, decimal(10,0) BINARY)->Column#16, Column#5, Column#6, gid",
+          "        └─ExchangeReceiver 10000.00 mpp[tiflash]  ",
+          "          └─ExchangeSender 10000.00 mpp[tiflash]  ExchangeType: HashPartition, Compression: FAST, Hash Cols: [name: Column#5, collate: binary], [name: Column#6, collate: binary], [name: gid, collate: binary]",
+          "            └─Expand 10000.00 mpp[tiflash]  level-projection:[test.t.c, <nil>->Column#5, <nil>->Column#6, 0->gid],[test.t.c, Column#5, <nil>->Column#6, 1->gid],[test.t.c, Column#5, Column#6, 3->gid]; schema: [Column: [test.t.c,Column#5,Column#6,gid] Unique key: []]",
+          "              └─Projection 10000.00 mpp[tiflash]  test.t.c, test.t.a, test.t.b",
+          "                └─TableFullScan 10000.00 mpp[tiflash] table:t keep order:false, stats:pseudo"
+        ],
+        "Warn": null
+      },
+      {
+        "SQL": "explain format = 'brief' select count(a) from t group by a, b with rollup; -- 3. should keep the original col a",
+        "Plan": [
+          "TableReader 8000.00 root  MppVersion: 1, data:ExchangeSender",
+          "└─ExchangeSender 8000.00 mpp[tiflash]  ExchangeType: PassThrough",
+          "  └─Projection 8000.00 mpp[tiflash]  Column#8",
+          "    └─HashAgg 8000.00 mpp[tiflash]  group by:Column#5, Column#6, gid, funcs:sum(Column#11)->Column#8",
+          "      └─ExchangeReceiver 8000.00 mpp[tiflash]  ",
+          "        └─ExchangeSender 8000.00 mpp[tiflash]  ExchangeType: HashPartition, Compression: FAST, Hash Cols: [name: Column#5, collate: binary], [name: Column#6, collate: binary], [name: gid, collate: binary]",
+          "          └─HashAgg 8000.00 mpp[tiflash]  group by:Column#5, Column#6, gid, funcs:count(test.t.a)->Column#11",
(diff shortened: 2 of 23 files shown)


expression/expression.go
```diff
@@ -97,13 +97,24 @@ type ReverseExpr interface {
...
+}
+
 // Expression represents all scalar expression in SQL.
 type Expression interface {
 	fmt.Stringer
 	goJSON.Marshaler
 	VecExpr
 	ReverseExpr
 	CollationInfo
+	TraverseDown
```
winoros, line 117: Seems that only `Expression` implements the `TraverseDown`. So we only need to
```suggestion
	Traverse(TraverseAction) Expression
```
?

AilinKid: make sense

planner/core/expression_rewriter.go
```diff
@@ -2001,6 +2007,50 @@ func (er *expressionRewriter) funcCallToExpression(v *ast.FuncCallExpr) {
...
+			// for the second and third grouping set: {a,a,null} and {a,null,null}, a here is the col ref of original
+			// column `a`. So from the static layer, this two grouping set are equivalent, we don't need to copy col
+			// `a double times at the every beginning and resort to gpos to distinguish them.
+			//  {col-a, col-b, gid, gpos}
+			//  {a, b, 0, 1}, {a, null, 1, 2}, {a, null, 1, 3}, {null, null, 2, 4}
+			// grouping function still only need to care about gid is enough, gpos what group and shuffle keys cared.
+			newArg := er.rollExpand.GID.Clone()
+			function, er.err = er.newFunction(v.FnName.L, &v.Type, newArg)
+			if er.err == nil {
+				if scalarFunc, ok := function.(*expression.ScalarFunction); ok {
```
elsa0520, line 2032: Is the indentation normal here?

planner/core/task.go
```diff
@@ -1347,6 +1347,23 @@ func (p *PhysicalTopN) addPartialLimitForSubScans(copSubPlans []PhysicalPlan, fi
...
 
+func (p *PhysicalExpand) attach2Task(tasks ...task) task {
+	t := tasks[0].copy()
+	// current expand can only be run in MPP TiFlash mode.
+	if mpp, ok := t.(*mppTask); ok {
+		p.SetChildren(mpp.p)
+		mpp.p = p
+		return mpp
+	}
+	// future code for TiDB self Expand implementation.
```
guo-shaoge, line 1358: Why not just give error for rootTask? Future code seems useless for now.

planner/core/physical_plans.go
```diff
@@ -1579,11 +1579,18 @@ type PhysicalExpand struct {
 	// GroupingSets is used to define what kind of group layout should the underlying data follow.
 	// For simple case: select count(distinct a), count(distinct b) from t; the grouping expressions are [a] and [b].
 	GroupingSets expression.GroupingSets
+
+	// The level projections is generated from grouping sets，make execution more clearly.
+	LevelExprs [][]expression.Expression
+
+	// The generated column names. Eg: "grouping_id" and so on.
+	GeneratedColNames []string
```
guo-shaoge, line 1587:
```suggestion
	ExtraGroupingColNames []string
```
Because `GeneratedColNames` makes people think it's related to virtual/generated column.

planner/core/logical_plans.go
```diff
@@ -662,23 +668,102 @@ func (p *LogicalExpand) GenLevelProjections() {
 }
 
 // GenerateGroupingMarks generate the groupingMark for the source column specified in grouping function.
-func (p *LogicalExpand) GenerateGroupingMarks(sourceCol *expression.Column) (resMap map[uint64]struct{}) {
+func (p *LogicalExpand) GenerateGroupingMarks(sourceCols []*expression.Column) []map[uint64]struct{} {
+	// since grouping function may have multi args like grouping(a,b), so the source columns may greater than 1.
```
guo-shaoge, line 672:
```suggestion
	// Since grouping function may have multi args like grouping(a,b), so the source columns may greater than 1.
```

planner/core/logical_plans.go
```diff
@@ -662,23 +668,102 @@ func (p *LogicalExpand) GenLevelProjections() {
...
-		resMap = make(map[uint64]struct{}, 1)
-		res := uint64(0)
-		// from high pos to low pos.
-		for i := len(p.distinctGroupByCol) - 1; i >= 0; i-- {
-			if p.distinctGroupByCol[i].UniqueID == sourceCol.UniqueID {
-				// col is needed, fill the corresponding pos as 1.
-				res = res | 1
+		for _, oneCol := range sourceCols {
+			resMap := make(map[uint64]struct{}, 1)
+			res := uint64(0)
```
guo-shaoge, line 687:
```suggestion
			var res uint64
```

planner/core/casetest/testdata/enforce_mpp_suite_out.json
```diff
@@ -1924,5 +1924,121 @@
...
+        "SQL": "explain format = 'brief' select sum(c), count(1) from t group by a, b with rollup; -- 2. non-grouping set col c",
+        "Plan": [
+          "TableReader 8000.00 root  MppVersion: 1, data:ExchangeSender",
+          "└─ExchangeSender 8000.00 mpp[tiflash]  ExchangeType: PassThrough",
+          "  └─Projection 8000.00 mpp[tiflash]  Column#8, Column#9",
+          "    └─HashAgg 8000.00 mpp[tiflash]  group by:Column#17, Column#18, Column#19, funcs:sum(Column#16)->Column#8, funcs:count(1)->Column#9",
+          "      └─Projection 10000.00 mpp[tiflash]  cast(test.t.c, decimal(10,0) BINARY)->Column#16, Column#5, Column#6, gid",
+          "        └─ExchangeReceiver 10000.00 mpp[tiflash]  ",
+          "          └─ExchangeSender 10000.00 mpp[tiflash]  ExchangeType: HashPartition, Compression: FAST, Hash Cols: [name: Column#5, collate: binary], [name: Column#6, collate: binary], [name: gid, collate: binary]",
+          "            └─Expand 10000.00 mpp[tiflash]  level-projection:[test.t.c, <nil>->Column#5, <nil>->Column#6, 0->gid],[test.t.c, Column#5, <nil>->Column#6, 1->gid],[test.t.c, Column#5, Column#6, 3->gid]; schema: [test.t.c,Column#5,Column#6,gid]",
```
guo-shaoge, line 1955: Length of `level-projection` depends on the number of group by column. So the output may be very long. Does that matters?

planner/core/exhaust_physical_plans.go
```diff
@@ -2483,6 +2483,38 @@ func (p *LogicalProjection) TryToGetChildProp(prop *property.PhysicalProperty) (
 	return newProp, true
 }
 
+// exhaustPhysicalPlans enumerate all the possible physical plan for expand operator (currently only mpp case is supported)
+func (p *LogicalExpand) exhaustPhysicalPlans(prop *property.PhysicalProperty) ([]PhysicalPlan, bool, error) {
+	// under the mpp task type, if the sort item is not empty, refuse it, cause expanded data doesn't support any sort items.
+	if !prop.IsSortItemEmpty() {
+		return nil, true, nil
```
guo-shaoge, line 2490: Is the second return value expect?

planner/core/exhaust_physical_plans.go
```diff
@@ -2483,6 +2483,38 @@ func (p *LogicalProjection) TryToGetChildProp(prop *property.PhysicalProperty) (
...
+	}
+	// now Expand mode can only be executed on TiFlash node.
+	// Upper layer shouldn't expect any mpp partition from an Expand operator.
+	// todo: data output from Expand operator should keep the origin data mpp partition.
+	if prop.TaskTp == property.MppTaskType && prop.MPPPartitionTp != property.AnyType {
+		return nil, true, nil
+	}
+	// for property.RootTaskType and property.MppTaskType with no partition option, we can give an MPP Expand.
+	if p.SCtx().GetSessionVars().IsMPPAllowed() || p.SCtx().GetSessionVars().IsMPPEnforced() {
```
guo-shaoge, line 2503:
```suggestion
	if p.SCtx().GetSessionVars().IsMPPAllowed() {
```

planner/core/expression_rewriter.go
AilinKid: seem a normal tab here

planner/core/task.go
AilinKid: make sense

planner/core/casetest/testdata/enforce_mpp_suite_out.json
AilinKid: it's ok, seems spark does the same here

planner/core/exhaust_physical_plans.go
AilinKid: good catch, for sort items not empty, we can return a false to indicate a sort enforcer outside. for others, return true to just avoid enforcing path check.

guo-shaoge:
Rest lgtm

planner/core/exhaust_physical_plans.go
```diff
@@ -2481,6 +2481,39 @@ func (p *LogicalProjection) TryToGetChildProp(prop *property.PhysicalProperty) (
...
+		return nil, true, nil
+	}
+	// now Expand mode can only be executed on TiFlash node.
+	// Upper layer shouldn't expect any mpp partition from an Expand operator.
+	// todo: data output from Expand operator should keep the origin data mpp partition.
+	if prop.TaskTp == property.MppTaskType && prop.MPPPartitionTp != property.AnyType {
+		return nil, true, nil
+	}
+	// for property.RootTaskType and property.MppTaskType with no partition option, we can give an MPP Expand.
+	if p.SCtx().GetSessionVars().IsMPPAllowed() {
```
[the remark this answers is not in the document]
AilinKid: expand only contains simple column ref and literal projection. (always can push down)

wshwsh12 (approved):

PR #43341, opened 2023-04-24
ddl: wait schema change before rename table job is done

tangenta:
### What problem does this PR solve?

Issue Number: close #43338

Problem Summary:

### What is changed and how it works?

We split the renaming table job into two steps:
1. rename table and update the schema version.
2. update the job state to JobStateDone.

This is the requirement from TiCDC because
- it uses the job state to check whether the DDL is finished.
- there is a gap between schema reloading and job state updating:
   when the job state is updated to JobStateDone, before the new schema reloaded,
   there may be DMLs that use the old schema.
- TiCDC cannot handle the DMLs that use the old schema, because
   the commit TS of the DMLs are greater than the job state updating TS.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #44585, opened 2023-06-12
infoschema: update raw args to make it drop correct table in `RENAME TABLE`

### LINKED ISSUE #43714: deleted system table still can be get from information_schema
## Bug Report

create table in mysql db
drop table
check with "select * from information_schema"

can't find this table in information_schema
still exist
master

tangenta:

Issue Number: close #43714

Problem Summary:

Reproducible step:

```sql
create table test.t1 (id int primary key,a text);
insert test.t1 values(1,'334'),(4,'3443435'),(5,'fdf43t536653');
rename table test.t1 to mysql.t1;
SELECT * FROM information_schema.TABLES WHERE (TABLE_SCHEMA = 'mysql') AND (TABLE_NAME = 't1');
```

#43341 changes the behavior of `RENAME TABLE`: it takes two schema versions to complete. However, `applyDiff()` doesn't handle two `SchemaDiff`s properly. The same table is created twice in `infoschema`, causing the phantom described in #43714.

During applying schema diff, we check if the table already exists before creating a new table. Thus, we can skip the second creation.

Other alternative fixes include

- Drop the table first and then create a new table(like what `CREATE TABLE` with foreign key does).
    - Change `job.Args` at second step, so that it contains the new schema ID.
    - There are risks if we change `job.Args`, it is hard to guarantee all the components that relying on this work correctly.
- Check if the old table is dropped. If it is not found, skip creating the second table.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/ddl/table.go
+++ b/ddl/table.go
@@ -1178,6 +1178,9 @@ func finishJobRenameTable(d *ddlCtx, t *meta.Meta, job *model.Job) (int64, error
 		job.State = model.JobStateCancelled
 		return 0, errors.Trace(err)
 	}
+	// Before updating the schema version, we need to reset the old schema ID to new schema ID, so that
+	// the table info can be dropped normally in `ApplyDiff`. This is because renaming table requires two
+	// schema versions to complete.
 	oldRawArgs := job.RawArgs
 	job.Args[0] = job.SchemaID
 	job.RawArgs, err = json.Marshal(job.Args)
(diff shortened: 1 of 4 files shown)


crazycs520 (approved):

zimulala (approved):
LGTM

ddl/table.go
```diff
@@ -1177,10 +1178,17 @@ func finishJobRenameTable(d *ddlCtx, t *meta.Meta, job *model.Job) (int64, error
 		job.State = model.JobStateCancelled
 		return 0, errors.Trace(err)
 	}
+	oldRawArgs := job.RawArgs
```
zimulala, line 1181: Could  we add comments for this logic？

zimulala (approved):

crazycs520 (approved):

PR #44592, opened 2023-06-12
*: support plan replayer for historical stats

### LINKED ISSUE #45038: Support plan replayer for historical stats
## Backgound 

Since we already supported historical stats and make it enabled by default. It would be convenient for the user to be able to download the plan replayer file with historical stats.
To achieve this, we need to enhance the syntax of the `PLAN REPLAYER DUMP` statement to specify the expected time, then make the dump logic able to fetch and dump the expected historical stats, and we can reuse the code of historical stats HTTP API for this.

### Syntax

Add a new clause `WITH STATS AS OF TIMESTAMP [datetime | TSO]` after `PLAN REPLAYER DUMP`. The `AS OF TIMESTAMP` part is consistent with the stale read syntax.

For example:
```sql
plan replayer dump with stats as of timestamp '442012134592479233' explain [analyze] select ...
plan replayer dump with stats as of timestamp '2023-06-14 17:00:00' explain [analyze] select ...
```

### Behavior

- If the user specified a valid time and the corresponding historical stats are available, the stats in the replayer file would be the expected historical stats.
- The TS used to fetch the historical stats would be recorded in `sql_meta.toml` in the plan replayer zip file.
- If the expected historical stats are not available, it would dump the latest stats (the same as when there's no specified time), and the corresponding table names would be recorded in `errors.txt` in the plan replayer zip file.

## Tests

- Create a table (mark as time1), insert some data, then analyze (mark as time2), then insert some data, then analyze again (mark as time3).
  - Try to use the `PLAN REPLAYER DUMP ... AS OF TIMESTAMP time1 ...` to get a plan replayer file. In the zip file:
    - The recorded time in the `sql_meta.toml` should be consistent with time1.
    - The stats should be the same as the stats of time3.
    - It should be recorded in `errors.txt` that it failed to fetch historical stats of the table.
  - Try to get plan replayer file at time2. In the zip file:
    - The recorded time in the `sql_meta.toml` should be consistent with time2.
    - The stats should be the same as the stats of time2.
    - There should be no `errors.txt`.
  - Try to get plan replayer file at time2. In the zip file:
    - The recorded time in the `sql_meta.toml` should be consistent with time3.
    - The stats should be the same as the stats of time3.
    - There should be no `errors.txt`.

time-and-fate:

Issue Number: close #45038

Please see the issue.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


statistics/handle/dump.go
```diff
@@ -154,29 +165,39 @@ func (h *Handle) DumpHistoricalStatsBySnapshot(dbName string, tableInfo *model.T
 	}()
 	pi := tableInfo.GetPartitionInfo()
 	if pi == nil {
-		return h.getTableHistoricalStatsToJSONWithFallback(dbName, tableInfo, tableInfo.ID, snapshot)
+		jt, fallback, err := h.getTableHistoricalStatsToJSONWithFallback(dbName, tableInfo, tableInfo.ID, snapshot)
+		if fallback {
+			fallbackTbls = append(fallbackTbls, tableInfo.Name.O)
```
qw4990, line 170: TableName only seems cannot identify a table, should we add DBName here as well?

server/plan_replayer_test.go
```diff
@@ -242,3 +251,199 @@ func forEachFileInZipBytes(t *testing.T, b []byte, fn func(file *zip.File)) {
...
+		}
+	}
+	return
+}
+
+func TestDumpPlanReplayerAPIWithHistoryStats(t *testing.T) {
+	require.NoError(t, failpoint.Enable("github.com/pingcap/tidb/domain/sendHistoricalStats", "return(true)"))
+	defer func() {
+		require.NoError(t, failpoint.Disable("github.com/pingcap/tidb/domain/sendHistoricalStats"))
+	}()
```
qw4990, line 324: I recommend using the context to inject the failure instead of the `fail-point`.
Here is an example https://github.com/pingcap/tidb/blob/master/planner/core/plan_cache.go#L48.
Using the context seems more go-style and easier to use (don't need to execute 'enable failpoint' to modify all source code...)

qw4990 (approved):
LGTM

statistics/handle/dump.go
time-and-fate: Yes!
Updated.

server/plan_replayer_test.go
time-and-fate: The logic related to the failpoint is irrelevant to the content of this PR. Just that logic is needed to make the test here run as expected.
So I tend not to modify that in this PR.

xuyifangreeneyes (approved):

AilinKid (commented):
Rest LGTM

parser/ast/misc.go
```diff
@@ -281,6 +282,9 @@ type PlanReplayerStmt struct {
 	// 1. plan replayer load 'file';
 	// 2. plan replayer dump explain <analyze> 'file'
 	File string
+
+	// Fields below are currently useless.
```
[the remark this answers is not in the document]
time-and-fate: I noticed that the three fields below this line are introduced a long time ago, but are never used.
So I added a comment to try to make it clear.

parser/parser.y
```diff
@@ -14974,95 +14975,113 @@ RowStmt:
...
+			Stmt:    $6,
 			Analyze: false,
 			Load:    false,
 			File:    "",
 			Where:   nil,
 			OrderBy: nil,
 			Limit:   nil,
 		}
+		if $4 != nil {
+			x.HistoricalStatsInfo = $4.(*ast.AsOfClause)
```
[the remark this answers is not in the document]
time-and-fate: `AsOfClause` is from the stale read syntax, and it's used also in other statements, so we can't modify it at will.
To add the `WITH STATS` part, we need another grammar rule, so I added `PlanReplayerDumpOpt`.

For the `PlanReplayerDumpOpt` and `HistoricalStatsInfo`, you are right, it's a careless mistake. I should make them consistent. I'll update this when modifying `parser.y` the next time.

PR #44601, opened 2023-06-12
SQL digester: Modify sql digester logic for IN statement with single literal `IN (?) => IN (...)`

### LINKED ISSUE #44298: planner: bindings for `in (?)` can not match `in (?, ... ?)`
## Enhancement
See the case below, the binding for `... a in (1, 2, 3)` can not match queries with only 1 element in the in-list `... a in (1)` :
```
mysql> create table t (a int);

mysql> create binding for select * from t where a in(1, 2, 3) using select * from t where a in (1, 2, 3);

mysql> select * from t where a in (1, 2,3);
Empty set (0.01 sec)

mysql> select @@last_plan_from_binding;
+--------------------------+
| @@last_plan_from_binding |
+--------------------------+
|                        1 |
+--------------------------+
1 row in set (0.00 sec)

mysql> select * from t where a in (1, 2);
Empty set (0.01 sec)

mysql> select @@last_plan_from_binding;
+--------------------------+
| @@last_plan_from_binding |
+--------------------------+
|                        1 |
+--------------------------+
1 row in set (0.00 sec)

mysql> select * from t where a in (1);
Empty set (0.00 sec)

mysql> select @@last_plan_from_binding;
+--------------------------+
| @@last_plan_from_binding |
+--------------------------+
|                        0 |
+--------------------------+
1 row in set (0.00 sec)
```

isabella0428:
Issue Number: ref #44298 

Problem Summary:
* Currently, binding rule with in (1,2,3) is not applied to statements with only one literal (1)

The reason why the problem occurs is that the sql digester creates different normalized sql for (1,2,3) and (1).
in (1,2,3) => in (...)
in (1) => in (?)
Thus when the optimizer tries to find a matching binding rule, it fails.

I add an additional rule in sql digester to translate statements like `in (1)` to `in (...)` instead. 

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

1. Unit test
* Run `bindinfo_test`
```
cd bindinfo
go test -run TestBindingInListWithSingleLiteral --tags=intest
```

* Run `parser_test`
```
make parser_unit_test
```

2. Manual test steps (#44298 ):
```
mysql> create table t (a int);

mysql> create binding for select * from t where a in(1, 2, 3) using select * from t where a in (1, 2, 3);

mysql> select * from t where a in (1, 2,3);
Empty set (0.01 sec)

mysql> select @@last_plan_from_binding;
+--------------------------+
| @@last_plan_from_binding |
+--------------------------+
|                        1 |
+--------------------------+
1 row in set (0.00 sec)

mysql> select * from t where a in (1, 2);
Empty set (0.01 sec)

mysql> select @@last_plan_from_binding;
+--------------------------+
| @@last_plan_from_binding |
+--------------------------+
|                        1 |
+--------------------------+
1 row in set (0.00 sec)

mysql> select * from t where a in (1);
Empty set (0.00 sec)

mysql> select @@last_plan_from_binding;
+--------------------------+
| @@last_plan_from_binding |
+--------------------------+
|                        1 |
+--------------------------+
1 row in set (0.00 sec)
```

Side effects

Documentation

- [ ] Affects user behaviors
- [x] Contains syntax changes

### Release note

```release-note
Fixed the bug for binding rule like in (1,2,3) cannot be applied to in (1)
```


isabella0428:
Hi guys, I checked the testcases and I currently think of several ways to get around.

The problem here is that the digest of `in (1,2,3)  => in (...)` and `in (1) => in (?)`. When we matching the binding rule using the hash value of the digest, it leads to a mismatch.

One way to get around this is to change the reduceLit method in `parser/digester.go`. We can add a rule like:
in ( lit ) => in (...) instead of in (?)  => But I am wondering if there would be some backward incompatibility problems.
eg: select ? from `b` where `id` in ( ? , `a` , ? )

Another way to do that is to insert two rules when add binding rule. We can convert all `...` to `?` when adding the second rule. So that we can get a match when we try to look up the rules. But there are still some concerns about inappropriate matching.

Is there any suggestion regarding this PR?
BTW the current commit  wrongly pop the left paren when reducing. I'll fix it after we decide the next steps :D

qw4990:
> Hi guys, I checked the testcases and I currently think of several ways to get around.
> 
> The problem here is that the digest of `in (1,2,3) => in (...)` and `in (1) => in (?)`. When we matching the binding rule using the hash value of the digest, it leads to a mismatch.
> 
> One way to get around this is to change the reduceLit method in `parser/digester.go`. We can add a rule like: in ( lit ) => in (...) instead of in (?) => But I am wondering if there would be some backward incompatibility problems. eg: select ? from `b` where `id` in ( ? , `a` , ? )
> 
> Another way to do that is to insert two rules when add binding rule. We can convert all `...` to `?` when adding the second rule. So that we can get a match when we try to look up the rules. But there are still some concerns about inappropriate matching.
> 
> Is there any suggestion regarding this PR? BTW the current commit wrongly pop the left paren when reducing. I'll fix it after we decide the next steps :D

The first solution seems better. If we apply the second solution on a query with N `in-condition`, e.g. `where a in (1) and b in (1) and c in (1) and d in (1)`, we have to create `2^N` for it, which seems unacceptable.

isabella0428:
> Hi @isabella0428 , sorry for my late reply. Could you let me know if you are still here? One of our customers met this issue recently, and we want to solve this issue in the next few weeks. If you are still here, we'd like to use this PR as our solution, your approach seems good.

Hi Yuanjia, thanks for replying! Glad that this PR may help. In order to apply this change, I may also need to change some existing tests. I will let you know when it is ready for review : )

qw4990:
> > Hi @isabella0428 , sorry for my late reply. Could you let me know if you are still here? One of our customers met this issue recently, and we want to solve this issue in the next few weeks. If you are still here, we'd like to use this PR as our solution, your approach seems good.
> 
> Hi Yuanjia, thanks for replying! Glad that this PR may help. In order to apply this change, I may also need to change some existing tests. I will let you know when it is ready for review : )

OK, one suggestion is that we'd better introduce a new flag on `sqlDigester` to control this behavior to let it only work for bindings to avoid some unnecessary changes.

isabella0428:
> > > Hi @isabella0428 , sorry for my late reply. Could you let me know if you are still here? One of our customers met this issue recently, and we want to solve this issue in the next few weeks. If you are still here, we'd like to use this PR as our solution, your approach seems good.
> > 
> > 
> > Hi Yuanjia, thanks for replying! Glad that this PR may help. In order to apply this change, I may also need to change some existing tests. I will let you know when it is ready for review : )
> 
> OK, one suggestion is that we'd better introduce a new flag on `sqlDigester` to control this behavior to let it only work for bindings to avoid some unnecessary changes and unexpected test failures.
> 
> And you don't need to consider any compatibility issues in your PR. It's a little dirty and tricky to handle them completely. I'll solve compatibility issues after your PR later on. And please let me know if you need any help.

Thanks for the suggestion! For the new flag, I may need a little more help about that. I am currently modifying `reduce rule for sql digest`, which only has the information about the tokens. I am not sure how to get the info `whether the command is used for binding`.

Also there is one error that may be caused by my change. I am not quite sure what has happened. Could u pls take a look at it? Thanks!

```
    result.go:49: 
        	Error Trace:	testkit/result.go:49
        	            				planner/core/casetest/cbotest/cbo_test.go:394
        	Error:      	Not equal: 
        	            	expected: "[TableReader 0.00 root  data:Selection]\n[└─Selection 0.00 cop[tikv]  eq(test.t.b, 1)]\n[  └─TableFullScan 2.00 cop[tikv] table:t keep order:false]\n"
        	            	actual  : "[TableReader 10.00 root  data:Selection]\n[└─Selection 10.00 cop[tikv]  eq(test.t.b, 1)]\n[  └─TableFullScan 10000.00 cop[tikv] table:t keep order:false, stats:pseudo]\n"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,4 +1,4 @@
        	            	-[TableReader 0.00 root  data:Selection]
        	            	-[└─Selection 0.00 cop[tikv]  eq(test.t.b, 1)]
        	            	-[  └─TableFullScan 2.00 cop[tikv] table:t keep order:false]
        	            	+[TableReader 10.00 root  data:Selection]
        	            	+[└─Selection 10.00 cop[tikv]  eq(test.t.b, 1)]
        	            	+[  └─TableFullScan 10000.00 cop[tikv] table:t keep order:false, stats:pseudo]
        	            	 
        	Test:       	TestNullCount
        	Messages:   	sql:explain format = 'brief' select * from t where b = 1, args:[]
```

qw4990:
> > > > Hi @isabella0428 , sorry for my late reply. Could you let me know if you are still here? One of our customers met this issue recently, and we want to solve this issue in the next few weeks. If you are still here, we'd like to use this PR as our solution, your approach seems good.
> > > 
> > > 
> > > Hi Yuanjia, thanks for replying! Glad that this PR may help. In order to apply this change, I may also need to change some existing tests. I will let you know when it is ready for review : )
> > 
> > 
> > OK, one suggestion is that we'd better introduce a new flag on `sqlDigester` to control this behavior to let it only work for bindings to avoid some unnecessary changes and unexpected test failures.
> > And you don't need to consider any compatibility issues in your PR. It's a little dirty and tricky to handle them completely. I'll solve compatibility issues after your PR later on. And please let me know if you need any help.
> 
> @qw4990 Thanks for the suggestion! For the new flag, I may need a little more help about that. I am currently modifying `reduce rule for sql digest`, which only has the information about the tokens. I am not sure how to get the info `whether the command is used for binding`.
> 
> Also there is one error that may be caused by my change. I am not quite sure what has happened. Could u pls take a look at it? Thanks!
> 
> ```
>     result.go:49: 
>         	Error Trace:	testkit/result.go:49
>         	            				planner/core/casetest/cbotest/cbo_test.go:394
>         	Error:      	Not equal: 
>         	            	expected: "[TableReader 0.00 root  data:Selection]\n[└─Selection 0.00 cop[tikv]  eq(test.t.b, 1)]\n[  └─TableFullScan 2.00 cop[tikv] table:t keep order:false]\n"
>         	            	actual  : "[TableReader 10.00 root  data:Selection]\n[└─Selection 10.00 cop[tikv]  eq(test.t.b, 1)]\n[  └─TableFullScan 10000.00 cop[tikv] table:t keep order:false, stats:pseudo]\n"
>         	            	
>         	            	Diff:
>         	            	--- Expected
>         	            	+++ Actual
>         	            	@@ -1,4 +1,4 @@
>         	            	-[TableReader 0.00 root  data:Selection]
>         	            	-[└─Selection 0.00 cop[tikv]  eq(test.t.b, 1)]
>         	            	-[  └─TableFullScan 2.00 cop[tikv] table:t keep order:false]
>         	            	+[TableReader 10.00 root  data:Selection]
>         	            	+[└─Selection 10.00 cop[tikv]  eq(test.t.b, 1)]
>         	            	+[  └─TableFullScan 10000.00 cop[tikv] table:t keep order:false, stats:pseudo]
>         	            	 
>         	Test:       	TestNullCount
>         	Messages:   	sql:explain format = 'brief' select * from t where b = 1, args:[]
> ```

The information about whether this SQL digester is used for binding should be passed by the caller using the digester from outside. In TiDB, the entry point to get a SQL's binding is this function `getBindRecord`, so you can pass a flag from there.

parser/digester.go
```diff
@@ -380,6 +391,17 @@ func (d *sqlDigester) isGenericList(last2 []token) (generic bool) {
 	return
 }
 
+func (d *sqlDigester) isSingleLiteralInList(last3 []token, currToken *token) (generic bool) {
```
qw4990, line 394: Can we merge this function into `reduceInListWithSingleLiteral`?

parser/digester_test.go
```diff
@@ -34,6 +34,9 @@ func TestNormalize(t *testing.T) {
 		{"select \\N", "select ?"},
 		{"SELECT `null`", "select `null`"},
 		{"select * from b where id = 1", "select * from `b` where `id` = ?"},
+		{"select * from t where a in (1)", "select * from `t` where `a` in ( ... )"},
```
qw4990, line 37: Please add 2 more test cases:
`select * from t where (a, b) in ((1, 1))`
`select * from t where (a, b) in ((1, 1), (2, 2))`

qw4990:
As for the CI failures, I debugged it locally but didn't reproduce it.
image

I guess it's not your PR's problem, just ignore them.

bindinfo/tests/bind_test.go
```diff
@@ -429,6 +429,48 @@ func TestBindingSymbolList(t *testing.T) {
...
+	tk.MustQuery(sqlcmd)
+	require.Equal(t, "t:ia", tk.Session().GetSessionVars().StmtCtx.IndexNames[0])
+	require.True(t, tk.MustUseIndex(sqlcmd, "ia(a)"))
+
+	tk.MustExec(binding)
+
+	// after binding
+	tk.MustQuery(sqlcmd)
+	require.Equal(t, "t:ib", tk.Session().GetSessionVars().StmtCtx.IndexNames[0])
+	require.True(t, tk.MustUseIndex(sqlcmd, "ib(b)"))
```
qw4990, line 456: You can run `select @@last_query_from_binding` to check whether it can take effect explicitly.

parser/digester.go
```diff
@@ -334,6 +335,16 @@ func (d *sqlDigester) isGenericLists(last4 []token) bool {
 	return true
 }
 
+func (d *sqlDigester) reduceInListWithSingleLiteral(currTok *token) {
+	// IN (lit) => IN (...)
```
qw4990, line 339: Please add the issue's link here: https://github.com/pingcap/tidb/issues/44298

qw4990:
It can pass all tests
image

isabella0428:
> > > > > Hi @isabella0428 , sorry for my late reply. Could you let me know if you are still here? One of our customers met this issue recently, and we want to solve this issue in the next few weeks. If you are still here, we'd like to use this PR as our solution, your approach seems good.
> > > > 
> > > > 
> > > > Hi Yuanjia, thanks for replying! Glad that this PR may help. In order to apply this change, I may also need to change some existing tests. I will let you know when it is ready for review : )
> > > 
> > > 
> > > OK, one suggestion is that we'd better introduce a new flag on `sqlDigester` to control this behavior to let it only work for bindings to avoid some unnecessary changes and unexpected test failures.
> > > And you don't need to consider any compatibility issues in your PR. It's a little dirty and tricky to handle them completely. I'll solve compatibility issues after your PR later on. And please let me know if you need any help.
> > 
> > 
> > @qw4990 Thanks for the suggestion! For the new flag, I may need a little more help about that. I am currently modifying `reduce rule for sql digest`, which only has the information about the tokens. I am not sure how to get the info `whether the command is used for binding`.
> > Also there is one error that may be caused by my change. I am not quite sure what has happened. Could u pls take a look at it? Thanks!
> > ```
> >     result.go:49: 
> >         	Error Trace:	testkit/result.go:49
> >         	            				planner/core/casetest/cbotest/cbo_test.go:394
> >         	Error:      	Not equal: 
> >         	            	expected: "[TableReader 0.00 root  data:Selection]\n[└─Selection 0.00 cop[tikv]  eq(test.t.b, 1)]\n[  └─TableFullScan 2.00 cop[tikv] table:t keep order:false]\n"
> >         	            	actual  : "[TableReader 10.00 root  data:Selection]\n[└─Selection 10.00 cop[tikv]  eq(test.t.b, 1)]\n[  └─TableFullScan 10000.00 cop[tikv] table:t keep order:false, stats:pseudo]\n"
> >         	            	
> >         	            	Diff:
> >         	            	--- Expected
> >         	            	+++ Actual
> >         	            	@@ -1,4 +1,4 @@
> >         	            	-[TableReader 0.00 root  data:Selection]
> >         	            	-[└─Selection 0.00 cop[tikv]  eq(test.t.b, 1)]
> >         	            	-[  └─TableFullScan 2.00 cop[tikv] table:t keep order:false]
> >         	            	+[TableReader 10.00 root  data:Selection]
> >         	            	+[└─Selection 10.00 cop[tikv]  eq(test.t.b, 1)]
> >         	            	+[  └─TableFullScan 10000.00 cop[tikv] table:t keep order:false, stats:pseudo]
> >         	            	 
> >         	Test:       	TestNullCount
> >         	Messages:   	sql:explain format = 'brief' select * from t where b = 1, args:[]
> > ```
> 
> The information about whether this SQL digester is used for binding should be passed by the caller using the digester from outside. In TiDB, the entry point to get a SQL's binding is this function `getBindRecord`, so you can pass a flag from there.
> 
> I advise you to create a new function like `NormalizeDigestForBinding`, and then just let the optimizer use this new function when matching bindings.

Thanks for the comment! I create a series of APIs to normalize the sql in binding scenarios.

planner/optimize.go
```diff
@@ -621,12 +621,63 @@ func ExtractSelectAndNormalizeDigest(stmtNode ast.StmtNode, specifiledDB string)
 	return nil, "", "", nil
 }
 
+// ExtractSelectAndNormalizeDigestForBinding extract the select statement and normalize it with additional binding specific rules
+func ExtractSelectAndNormalizeDigestForBinding(stmtNode ast.StmtNode, specifiledDB string) (ast.StmtNode, string, string, error) {
```
qw4990, line 625: How about adding a flag in `ExtractSelectAndNormalizeDigest` like `ExtractSelectAndNormalizeDigest(..., forBinding bool)` instead of creating a new similar function?

parser/digester.go
```diff
@@ -231,6 +269,65 @@ func (d *sqlDigester) normalize(sql string, keepHint bool) {
 	d.tokens.reset()
 }
 
+// Normalize the sql command with additional binding specific rules
+func (d *sqlDigester) normalizeForBinding(sql string, keepHint bool) {
```
qw4990, line 273: It's very similar with `sqlDigester.normalize`, how about merging them into one?
For example:
```
func (d *sqlDigester) normalize(sql string, keepHint, forBinding bool) {
    ...
    if forBinding {d.reduceInListWithSingleLiteral(&currTok)}
    ...
}
```

qw4990:
Two small comments, PTAL @isabella0428

bindinfo/tests/bind_test.go
```diff
@@ -429,6 +429,56 @@ func TestBindingSymbolList(t *testing.T) {
...
+
+	rs, err := tk.Exec("select @@last_plan_from_binding")
+	require.NoError(t, err)
+	chk := rs.NewChunk(nil)
+	err = rs.Next(context.TODO(), chk)
+	require.NoError(t, err)
+	useBindingFlag := chk.Column(0).GetInt64(0)
+	require.Equal(t, int64(1), useBindingFlag)
```
qw4990, line 464: You can use `tk.MustQuery("select @@last_plan_from_cache").Check(testkit.Rows("1"))` to simplify this part.

parser/digester_test.go
```diff
@@ -79,6 +79,70 @@ func TestNormalize(t *testing.T) {
...
+		{"select * from `select`", "select * from `select`"},
+		{"select * from 🥳", "select * from `🥳`"},
+		// test syntax error, it will be checked by parser, but it should not make normalize dead loop.
+		{"select * from t ignore index(", "select * from `t` ignore index"},
+		{"select /*+ ", "select "},
+		{"select 1 / 2", "select ? / ?"},
+		{"select * from t where a = 40 limit ?, ?", "select * from `t` where `a` = ? limit ..."},
+		{"select * from t where a > ?", "select * from `t` where `a` > ?"},
+		{"select @a=b from t", "select @a = `b` from `t`"},
+		{"select * from `table", "select * from"},
```
qw4990, line 129: How about removing these cases? They are irrelevant.

parser/digester.go
```diff
@@ -95,17 +95,17 @@ func Normalize(sql string) (result string) {
...
 	d := digesterPool.Get().(*sqlDigester)
-	result = d.doNormalize(sql, true)
+	result = d.doNormalize(sql, true, forBinding)
 	digesterPool.Put(d)
 	return
 }
 
 // NormalizeDigest combines Normalize and DigestNormalized into one method.
-func NormalizeDigest(sql string) (normalized string, digest *Digest) {
+func NormalizeDigest(sql string, forBinding bool) (normalized string, digest *Digest) {
```
qw4990, line 106: How about keeping these exposed functions unchanged and creating a new exposed function? Otherwise this cause too many changes.
For example:
```
func  NormalizeKeepHint(sql string) (result string) {
    ...
    d.doNormalize(sql, true, false)
    ...
}

func NormalizeDigest(sql string) (normalized string, digest *Digest) {
    ...
    d.doNormalize(sql, true, false)
    ...
}

func NormalizeForBinding(sql string) {
    ...
    d.doNormalize(sql, false, true)
    ...
}
```

qw4990: Almost LGTM, the last one comment! @isabella0428

planner/optimize.go
```diff
@@ -584,18 +584,33 @@ func ExtractSelectAndNormalizeDigest(stmtNode ast.StmtNode, specifiledDB string)
...
+			} else {
+				explainSQL = x.Text()
+			}
+
+			if forBinding {
+				// Apply additional binding rules
+				normalizeExplainSQL = parser.NormalizeForBinding(explainSQL)
 			} else {
-				normalizeExplainSQL = parser.Normalize(x.Text())
+				normalizeExplainSQL = parser.Normalize(explainSQL)
```
qw4990, line 611:
```suggestion
				normalizeExplainSQL = parser.Normalize(x.Text())
```

parser/digester.go
```diff
@@ -334,6 +377,20 @@ func (d *sqlDigester) isGenericLists(last4 []token) bool {
 	return true
 }
 
+// IN (lit) => IN (...) Issue: #44298
```
qw4990, line 380:
```suggestion
// IN (?) => IN (...) Issue: #44298
```

parser/digester.go
```diff
@@ -194,6 +231,12 @@ func (d *sqlDigester) normalize(sql string, keepHint bool) {
 
 		d.reduceLit(&currTok)
 
+		// Apply binding matching specific rules
+		if forBinding {
+			// IN (Lit) => IN ( ... ) #44298
```
qw4990, line 236:
```suggestion
			// IN (?) => IN ( ... ) #44298
```

parser/digester.go
```diff
@@ -442,6 +499,26 @@ func (*sqlDigester) isComma(tok token) (isComma bool) {
...
+func (*sqlDigester) isInKeyword(tok token) (isInKeyword bool) {
+	isInKeyword = tok.lit == "in"
+	return
+}
+
+func (*sqlDigester) isQuestionMarkKeyword(tok token) (isQuestionMarkKeyword bool) {
+	isQuestionMarkKeyword = tok.lit == "?"
+	return
+}
+
```
qw4990, line 521:
```suggestion
```

qw4990 (approved):
Well done. Thanks for your contribution @isabella0428 ~

isabella0428:
Hi Yuanjia, thanks for all the comments! Really appreciate that.
Regarding the pr, i still am still meeting with a bug about `select @@last_plan_from_cache`

```
--- FAIL: TestBindingInListWithSingleLiteral (3.19s)
    result.go:49: 
                Error Trace:    /Users/isabella/Desktop/CSLearning/opensource/tidb/testkit/result.go:49
                                                        /Users/isabella/Desktop/CSLearning/opensource/tidb/bindinfo/tests/bind_test.go:458
                Error:          Not equal: 
                                expected: "[1]\n"
                                actual  : "[0]\n"
                                
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,2 +1,2 @@
                                -[1]
                                +[0]
                                 
                Test:           TestBindingInListWithSingleLiteral
                Messages:       sql:select @@last_plan_from_cache, args:[]
FAIL
exit status 1
```

It seems that the binding is working since `index` has been changed to `ib(b)`, but it seems that `select @@last_plan_from_cache` is not working. I've been spending a while on that, but still have no clues. Could u help me look into that? Thanks : D

hawkingrei (approved):

qw4990:
> Hi Yuanjia, thanks for all the comments! Really appreciate that. Regarding the pr, i am still meeting with a bug about `select @@last_plan_from_cache`
> 
> ```
> --- FAIL: TestBindingInListWithSingleLiteral (3.19s)
>     result.go:49: 
>                 Error Trace:    /Users/isabella/Desktop/CSLearning/opensource/tidb/testkit/result.go:49
>                                                         /Users/isabella/Desktop/CSLearning/opensource/tidb/bindinfo/tests/bind_test.go:458
>                 Error:          Not equal: 
>                                 expected: "[1]\n"
>                                 actual  : "[0]\n"
>                                 
>                                 Diff:
>                                 --- Expected
>                                 +++ Actual
>                                 @@ -1,2 +1,2 @@
>                                 -[1]
>                                 +[0]
>                                  
>                 Test:           TestBindingInListWithSingleLiteral
>                 Messages:       sql:select @@last_plan_from_cache, args:[]
> FAIL
> exit status 1
> ```
> 
> It seems that the binding is working since `index` has been changed to `ib(b)`, but it seems that `select @@last_plan_from_cache` is not working. I've been spending a while on that, but still have no clues. Could u help me look into that? Thanks : D

My mistake... it should be `select @@last_plan_from_binding;`

bindinfo/tests/bind_test.go
```diff
@@ -429,6 +429,50 @@ func TestBindingSymbolList(t *testing.T) {
...
+	require.True(t, tk.MustUseIndex(sqlcmd, "ia(a)"))
+
+	tk.MustExec(binding)
+
+	// after binding
+	tk.MustQuery(sqlcmd)
+	require.Equal(t, "t:ib", tk.Session().GetSessionVars().StmtCtx.IndexNames[0])
+	require.True(t, tk.MustUseIndex(sqlcmd, "ib(b)"))
+
+	tk.MustQuery("select @@last_plan_from_cache").Check(testkit.Rows("1"))
```
qw4990, line 458:
```suggestion
	tk.MustQuery("select @@last_plan_from_binding").Check(testkit.Rows("1"))
```

PR #44604, opened 2023-06-13
timer: Support set Tags for timer in timer framework

### LINKED ISSUE #44603: Support set `Tags` for timer in timer framework

lcwangchao:

Issue Number: close #44603

Timer framework is a lib to provide timer functions that serval features can use. We can use `key` indicate different features but it is some what limited. So `Tags` can be used to class different timers in different dimissions.

For example if we have two timers used for GC and backup on time, their keys can be:

`/tidb/back` and `/tidb/gc`

For these above two timers, we can allow users to modify their meta through some other SQL syntax. So we can add tags `user_ctl` to them to mark it.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


timer/tablestore/store.go
```diff
@@ -114,6 +119,11 @@ func (s *tableTimerStoreCore) List(ctx context.Context, cond api.Cond) ([]*api.T
 	}
 	defer back()
 
+	if sessVars := sctx.GetSessionVars(); sessVars.GetEnableIndexMerge() {
```
bb7133, line 122: Is there any plan we could remove this in the future? IMO it's quite confusing to the others.

lcwangchao: It's a limitation of multi-value index that is documented here: https://docs.pingcap.com/tidb/dev/choose-index#use-a-multi-valued-index . I think I can add some comments

bb7133: Got it, please add a comment about this.

bb7133 (approved):
LGTM

YangKeao (approved):
LGTM

timer/api/store.go
```diff
@@ -106,6 +107,8 @@ type TimerCond struct {
 	// If `KeyPrefix is` true, it will check whether the timer's key is prefixed with `TimerCond.Key`.
 	// Otherwise, it will check whether the timer's key equals `TimerCond.Key`.
 	KeyPrefix bool
+	// Tags indicates to filter the timer record with specified tags
+	Tags OptionalVal[[]string]
```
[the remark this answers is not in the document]
lcwangchao: Yes, it does not make a big difference but not all the same. One reason is to make sure that `TimerCond.FieldsSet` can return all set fields correctly...

Issue #44032, opened 2023-05-22
Introduce coordinator for mpp tasks execution


yibin87:
## Enhancement
The main advatages are:

- Better error tracking mechnism: mpp task can report first error directly to coordinator
- Provide infrastucture to support global runtime filter, coordinator can play the role as merging all hash build infos and broadcast to probe mpp tasks.

- [ ] Refact MPPClient and MPPGather to introduce MPPCoordinator interface
- [ ] Modify protocal to prepare for better error tracking
- [ ] Implement better error tracking both in tidb/tiflash side

PR #44697, opened 2023-06-15
executor, distsql: enhance mpp cooridnator to handle execption and execution summaries better

yibin87:

Issue Number: ref #44032

Problem Summary:

Previously only root mpp tasks can "report" errors and execution summaries to tidb. However, when errors were passing among mpp tasks, secondary errors would happen and have chance to be report to tidb first. Besides, due to some limitation of TiFlash, execution summaries may fail to pass to root tasks even the total query completes without errors. So this PR will let every mpp task report directly to mpp coordinator to ensure first error message and execution summaries.

1. When MPPVersion >= 2 && TiDB status server is on, mpp coordinator address will be TiDB status server address, otherwise, an empty string. And it will be included in mpp task meta info.
2. When mpp coordinator address is not empty, and physical plan contains limit plan node, ReportExecutionSummary flag is set true in mpp task meta info.
3. MPP coordinator records the first error message, and will be used when root task report any errors
4. When mpp coordinator closes, if ReportExecutionSummary is true and no error happened during dispatch phase, the main goroutine will wait until all reports are received or after 3 seconds.
5. MPP coordinator manager is responsible for managing all mpp coordinators, and it will starts a background goroutine to timely detect and delete mpp coordinators lives over time due to unknown reasons.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
Run tiflash queries and check the behaviors in compilation failure, run failure, and normal success situations.
Will add new metrics to track mpp coordinator counter and received reports related metrics, and use them in endless tests like random failpoints.
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that mpp tasks with limit operator may lose execution summary.
Fix the issue that confusing error message may be returned to users when mpp tasks failed.
```


go.mod
```diff
@@ -301,3 +302,7 @@ replace (
 	github.com/pingcap/tidb/parser => ./parser
 	go.opencensus.io => go.opencensus.io v0.23.1-0.20220331163232-052120675fac
 )
+
+replace github.com/pingcap/kvproto => github.com/pingcap/kvproto v0.0.0-20230614050532-13569312829f
+
+replace github.com/pingcap/tipb => github.com/pingcap/tipb v0.0.0-20230607071926-bda24015c2d6
```
hawkingrei, line 308: This looks a bit strange?

yibin87:
block merge until 7.2 code freeze

go.mod
yibin87: Yeah, will update the final go mod info when ready to merge

executor/mppcoordmanager/BUILD.bazel
```diff
@@ -14,3 +15,14 @@ go_library(
         "@org_uber_go_zap//:zap",
     ],
 )
+
+go_test(
+    name = "mppcoordmanager_test",
+    srcs = ["mpp_coordinator_manager_test.go"],
+    embed = [":mppcoordmanager"],
+    deps = [
```
hawkingrei, line 23: Can you add ```race ="on",``` here?

yibin87: okk

executor/mppcoordmanager/mpp_coordinator_manager.go
```diff
@@ -0,0 +1,177 @@
...
+			}
+		}
+	}()
+}
+
+func (m *MPPCoordinatorManager) detectAndDelete(nowTs uint64) {
+	var outOfTimeIDs []CoordinatorUniqueID
+	m.mu.Lock()
+	for id := range m.coordinatorMap {
+		if nowTs > id.MPPQueryID.QueryTs+m.maxLifeTime {
```
SeaRise, line 81: Is it possible to delete only the coodinator that has received all the data but has not received the report summary?
I am a little worried that the long-distance mpp query will be deleted by mistake, such as the sql running a large etl

executor/mppcoordmanager/mpp_coordinator_manager.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// Register is to register mpp coordinator
+func (m *MPPCoordinatorManager) Register(coordID CoordinatorUniqueID, mppCoord kv.MppCoordinator) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	_, exists := m.coordinatorMap[coordID]
+	if exists {
+		return errors.Errorf("Mpp coordinator already registered: %d %d %d %d", coordID.MPPQueryID.QueryTs, coordID.MPPQueryID.LocalQueryID, coordID.MPPQueryID.ServerID, coordID.GatherID)
+	}
```
SeaRise, line 122: from https://github.com/pingcap/tiflash/pull/7694#pullrequestreview-1499880922
Multiple different mpp queries may have the same query id, which may cause the regsiter coodinator to fail?

SeaRise: For example, sql with apply

executor/mppcoordmanager/mpp_coordinator_manager.go
SeaRise: Or does it follow the time of sql execution timeout?

distsql/select_result.go
```diff
@@ -556,21 +581,7 @@ func (r *selectResult) updateCopRuntimeStats(ctx context.Context, copStats *copr
...
-			}
-		}
-		num := uint64(0)
-		dummySummary := &tipb.ExecutorExecutionSummary{TimeProcessedNs: &num, NumProducedRows: &num, NumIterations: &num, ExecutorId: nil}
-		for _, planID := range r.copPlanIDs {
-			if _, ok := recorededPlanIDs[planID]; !ok {
-				r.ctx.GetSessionVars().StmtCtx.RuntimeStatsColl.RecordOneCopTask(planID, r.storeType.Name(), callee, dummySummary)
-			}
-		}
+		recordExecutionSummariesForMppTasks(r.ctx.GetSessionVars().StmtCtx, r.selectResp.GetExecutionSummaries(), callee, r.storeType.Name(), r.copPlanIDs)
```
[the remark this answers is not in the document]
yibin87: Make sense, I'll change the name.

executor/mppcoordmanager/mpp_coordinator_manager.go
yibin87: Yeah, that's the longest time one mpp task can take.

executor/mppcoordmanager/mpp_coordinator_manager.go
yibin87: I think that may be caused by unused mpp_gather id, in this PR, mpp_gather id will be assigned meaningful value. I'll double check it.

yibin87: Confirmed that, either LocalQueyId, or GatherID is different, different coordinators won't share the same key.

distsql/select_result.go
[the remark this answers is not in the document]
yibin87: Done.

executor/mppcoordmanager/mpp_coordinator_manager.go
SeaRise: ok

executor/internal/mpp/local_mpp_coordinator.go
```diff
@@ -275,6 +313,22 @@ func (c *localMppCoordinator) fixTaskForCTEStorageAndReader(exec *tipb.Executor,
 	return nil
 }
 
+// DFS to check if plan needs report execution summary through ReportMPPTaskStatus mpp service
+// Currently, return true if plan contains limit operator
+func needReportExecutionSummary(plan plannercore.PhysicalPlan) bool {
+	switch x := plan.(type) {
+	case *plannercore.PhysicalLimit:
```
SeaRise, line 320: Also for TopN

executor/mppcoordmanager/mpp_coordinator_manager.go
yibin87: It's hard coded here: https://github.com/pingcap/tidb/blob/9fc80a922a340d851d69327dc3de189bdb984fb3/store/copr/batch_coprocessor.go#L1133, and passed as timeout parameter to client-go sendRequest

executor/mppcoordmanager/mpp_coordinator_manager.go
```diff
@@ -0,0 +1,177 @@
...
+	}()
+}
+
+func (m *MPPCoordinatorManager) detectAndDelete(nowTs uint64) {
+	var outOfTimeIDs []CoordinatorUniqueID
+	m.mu.Lock()
+	for id := range m.coordinatorMap {
+		if nowTs > id.MPPQueryID.QueryTs+m.maxLifeTime {
+			outOfTimeIDs = append(outOfTimeIDs, id)
+			delete(m.coordinatorMap, id)
```
SeaRise, line 83: Is it necessary to actively trigger the failure of the coodinator here, and terminate the mpp query?

executor/internal/mpp/local_mpp_coordinator.go
yibin87: Does our TopN operator implementation terminate mpp task before receiving all data now?

executor/mppcoordmanager/mpp_coordinator_manager.go
SeaRise: ok

executor/mppcoordmanager/mpp_coordinator_manager.go
yibin87: This background routine is added to ensure mpp coordinator map won't keep terminated coordinators, won't affect coordinator's life cycle. Even if coordinator was deleted from this map before it terminated, no critical error will happen.

executor/internal/mpp/local_mpp_coordinator.go
SeaRise: yes, topn pushed down to tiflash is partial topn, as follows
```
tidb         tiflash
topX <--  topX
                topX
                topX
```
The total number of rows passed from tiflash to tidb will be tiflash_node * X, but tidb closes the executor after receiving X rows.

executor/mppcoordmanager/mpp_coordinator_manager.go
SeaRise: ok

executor/internal/mpp/local_mpp_coordinator.go
SeaRise: My fault, topn won't

yibin87: all right~

executor/internal/mpp/local_mpp_coordinator.go
```diff
@@ -394,9 +451,20 @@ func (c *localMppCoordinator) handleDispatchReq(ctx context.Context, bo *backoff
 // NOTE: We do not retry here, because retry is helpless when errors result from TiFlash or Network. If errors occur, the execution on TiFlash will finally stop after some minutes.
 // This function is exclusively called, and only the first call succeeds sending Tasks and setting all Tasks as cancelled, while others will not work.
 func (c *localMppCoordinator) cancelMppTasks() {
+	usedStoreAddrs := make(map[string]bool)
 	c.mu.Lock()
-	defer c.mu.Unlock()
-	c.sessionCtx.GetMPPClient().CancelMPPTasks(kv.CancelMPPTasksParam{Reqs: c.mppReqs})
+	for _, task := range c.mppReqs {
```
SeaRise, line 456: Maybe add some notes?
- The mpptask of the same addr only needs to send a cancel request, because of xxxxx
- If one mpptask is canceled, other mpp tasks must also be canceled

SeaRise (commented):
Others LGTM

executor/internal/mpp/local_mpp_coordinator.go
yibin87: Make sense

yibin87: Done

executor/mpp_gather.go
```diff
@@ -94,20 +95,24 @@ func (e *MPPGather) Open(ctx context.Context) (err error) {
 		_, e.kvRanges, err = plannercore.GenerateRootMPPTasks(e.ctx, e.startTS, e.mppQueryID, sender, e.is)
 		return err
 	}
-	coord := e.buildCoordinator()
+	planIDs := collectPlanIDS(e.originalPlan, nil)
+	coord := e.buildCoordinator(planIDs)
+	err = mppcoordmanager.InstanceMPPCoordinatorManager.Register(mppcoordmanager.CoordinatorUniqueID{MPPQueryID: e.mppQueryID, GatherID: uint64(e.id)}, coord)
```
windtalker, line 100: If we need support gather level cancel, should not use `e.id` as GatherID

executor/mppcoordmanager/mpp_coordinator_manager.go
windtalker: But this is not the max running time of a mpp query, this timeout means the client does not receive any data for `readTimeoutUltraLong` time, if a mpp query keep return data for every 1 minutes, it never timeout

executor/internal/mpp/local_mpp_coordinator.go
```diff
@@ -448,20 +522,97 @@ func (c *localMppCoordinator) receiveResults(req *kv.MPPDispatchRequest, taskMet
...
-// Close and release the used resources.
+// ReportStatus implements MppCoordinator interface
+func (c *localMppCoordinator) ReportStatus(info kv.ReportStatusRequest) error {
+	taskID := info.Request.Meta.TaskId
+	var errMsg string
+	if info.Request.Error != nil {
+		errMsg = info.Request.Error.Msg
+	}
+	executionInfo := new(tipb.TiFlashExecutionInfo)
+	err := executionInfo.Unmarshal(info.Request.GetData())
```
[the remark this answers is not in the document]
yibin87: As mentioned in the previous comments, it is very corner case to meet the error. And it won't cause hang forever if we don't set flag or add receive counter here.

executor/mppcoordmanager/mpp_coordinator_manager.go
windtalker: Do you double checked sql with `Apply` executor? Apply may use the same gather id, if you use `e.id` as the gather id

executor/mpp_gather.go
yibin87: I see, I'll change it like mpp task id

executor/mppcoordmanager/mpp_coordinator_manager.go
SeaRise: If that's the case, then I suggest only deleting the coordinator that received the data but did not receive all report status messages.

executor/mppcoordmanager/mpp_coordinator_manager.go
yibin87: I checked the "select * from t where id > (select max(a) from s)" query, and get the previous conclusion.

executor/mppcoordmanager/mpp_coordinator_manager.go
yibin87: I see, I'll use the coordinator's close flag as a double check.

executor/mpp_gather.go
yibin87: Done

executor/mppcoordmanager/mpp_coordinator_manager.go
yibin87: Done

executor/mppcoordmanager/mpp_coordinator_manager.go
windtalker: Maybe concurrent apply is not enabled?

SeaRise: https://github.com/pingcap/tidb/pull/44697/files#diff-3fd4cee3c1972379d1b70f86b4cfde79297f1fea3016580771a88dfad9888496R89-R92

SeaRise (approved):
lgtm

executor/mppcoordmanager/mpp_coordinator_manager.go
yibin87: Yes, record one query that uses apply:
select count(*) from t where a > (select avg(b) from t t1 where t.a > t1.a);

Issue #44672, opened 2023-06-14
56 cases of `mysql_client_test` failed


YangKeao:
### 0. Background

`mysql_client_test` is the test cases in [mysql repository](https://github.com/YangKeao/mysql-server/tree/known-bugs/testclients). It's used to test both the server and the client(libmysql) implementation. As a MySQL-compatible database, I think it'd be better to pass all these tests. It can enrich our test cases on protocol layer.

Unfortunately, tons of the test cases have failed. They can be concluded in the following areas (except known unsupported features):

1. Cursor. About the server status, behavior, ... 
2. `ComFieldList`. The type and length of a field is usually different from the expectation of the tests. I'm not sure whether it's expected.
3. The count of warning is unexpected. (Though, I think this problem is not serious, right?).
4. Others...

### 1. Minimal reproduce step (Required)

1. `git clone https://github.com/YangKeao/mysql-server.git -b known-bugs`
5. See the `testclients/mysql_client_test.cc`, and find "TODO" in it. Several lines are commented out with "TODO: need fix", which means this test case cannot pass on TiDB.
6. Uncomment any one of them, and compile the target `mysql_client_test`. Assume the location of `mysql_client_test` binary is at `./cmake-build-mysql/runtime_output_directory/mysql_client_test`
7. Run `MYSQL_TEST_DIR=$(pwd)/mysql-test ./cmake-build-mysql/runtime_output_directory/mysql_client_test -P 4000 -h 127.0.0.1 -u root`, you'll see the error output.

For example, 

### 2. What did you expect to see? (Required)

All tests passed (except those which are known to be unsupported, like procedure, unsupported character set and other features).

### 3. What did you see instead (Required)

A lot of tests failed. 

- [ ] test_func_fields
- [ ] test_select_show
- [ ] test_time_zone
- [ ] test_bug33781442
- [ ] test_simple_temporal
- [ ] test_temporal_param
- [ ] test_date_date
- [ ] test_date_time
- [ ] test_date_ts
- [ ] test_date_dt
- [ ] test_field_misc
- [ ] test_explain_bug
- [ ] test_truncated_rows
- [ ] test_list_fields
- [ ] test_bug1500
- [ ] test_subqueries
- [ ] test_create_drop
- [ ] test_bug4079
- [ ] test_datetime_ranges
- [ ] test_bug4172
- [ ] test_conversion
- [ ] test_basic_cursors
- [ ] test_cursors_with_union
- [ ] test_cursor_for_show
[...]

PR #44730, opened 2023-06-16
server: implement spill disk for cursorFetch result

### LINKED ISSUE #43233: TiDB should track, limit (and maybe spill) the temporary memory used by CursorFetch
## Enhancement

To avoid the bug in concurrent executing statements, TiDB fetches all the results and stored them in the memory. The following `FETCH` command only reads data from memory. It's especially hard for TiDB to support a real "CursorFetch", as the execution model of TiDB assumes that the access to session struct is exclusive, which is not true for "CursorFetch".

This behavior will cause OOM, when the result set is big or there are many cursor fetch statement. It's even more unexpected as the user may only need the first several rows. We could have some ways to relief this problem:

1. Trace the memory through setting a suitable chunk allocator. However, the connection allocator will reset after command execution, so we need a new allocator across different statements.
2. Spill result to disk if needed.

YangKeao:

Issue Number: close #43233

1. Use the `rowContainer` to store the rows after execution.
2. Reset the memory/disk tracker when the statement is closed.

TODO:

- [x] Add tests
- [x] Consider the behavior of `Reset`.
- [x] Consider carefully about the server status. https://github.com/pingcap/tidb/issues/44672 seems to have shown some bugs about server status.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test (TODO)
- [ ] No code

I used the following codes to do some basic tests. The preloaded data is 5M rows `sysbench ./oltp_insert.lua` :

```java
public static void tidbBigSelectTest() throws SQLException, InterruptedException {
    // the following are test codes for MySQL
    ConnectionImpl conn = (ConnectionImpl) DriverManager.getConnection("jdbc:mysql://127.0.0.1:4000/test?useCursorFetch=true&defaultFetchSize=1000", "root", "");
    conn.prepareStatement("set @@tidb_mem_quota_query = 8 << 20\n").execute();
    conn.prepareStatement("begin").execute();

    String sql = "SELECT * FROM test.sbtest1";
    PreparedStatement stmt = conn.prepareStatement(sql, ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
    stmt.setFetchSize(1000);
    Instant beforeExecuteTime = Clock.systemUTC().instant();
    ResultSet rs = stmt.executeQuery();
    Instant afterExecuteTime = Clock.systemUTC().instant();
    System.out.println("It takes " + (Duration.between(beforeExecuteTime, afterExecuteTime).toString()) + " to EXECUTE QUERY");

    ResultSetMetaData metaData = rs.getMetaData();
    int columnCount = metaData.getColumnCount();
    int batchSize = stmt.getFetchSize();
    int rowCount = 0;
    int totalCount = 0;
    while (rs.next()) {
        rowCount++;
        totalCount++;
        Object[] columns = new Object[columnCount];
        for (int i = 1; i <= columnCount; i++) {
            columns[i - 1] = rs.getObject(i);
        }
        Row row = new Row(columns);
        if (rowCount == batchSize) {
            System.out.println("Fetched " + batchSize + " rows");
            System.out.println("allocated memory is " + (Runtime.getRuntime().totalMemory()-Runtime.getRuntime().freeMemory()));
            System.out.println("TotalCount " + totalCount);
            rowCount = 0;
        }
        if (totalCount == 4999999) {
            System.out.println("here");
        }
    }
    if (rowCount > 0) {
        System.out.println("allocated memory is " + (Runtime.getRuntime().totalMemory()-Runtime.getRuntime().freeMemory()));
        System.out.println("Fetched " + rowCount + " rows");
        System.out.println("TotalCount " + totalCount);
    }

    conn.prepareStatement("commit").execute();
}
```

### Release note

```release-note
Automatically spill the memory exceeding `tidb_mem_quota_query` into the disk
```

### DIFF
--- a/server/conn.go
+++ b/server/conn.go
@@ -2410,10 +2410,10 @@ func (cc *clientConn) writeChunks(ctx context.Context, rs ResultSet, binary bool
 // serverStatus, a flag bit represents server information.
 // fetchSize, the desired number of rows to be fetched each time when client uses cursor.
 func (cc *clientConn) writeChunksWithFetchSize(ctx context.Context, rs cursorResultSet, serverStatus uint16, fetchSize int) error {
-	iter := rs.GetRowIterator()
-
 	// construct the rows sent to the client according to fetchSize.
 	var curRows []chunk.Row
+	iter := rs.GetRowContainerReader()
+	// send the rows to the client according to fetchSize.
 	for i := 0; i < fetchSize && iter.Current() != iter.End(); i++ {
 		curRows = append(curRows, iter.Current())
 		iter.Next()
--- a/server/conn.go
+++ b/server/conn.go
@@ -2270,7 +2270,12 @@ func (cc *clientConn) writeResultSet(ctx context.Context, rs ResultSet, binary b
 	cc.initResultEncoder(ctx)
 	defer cc.rsEncoder.Clean()
 	if mysql.HasCursorExistsFlag(serverStatus) {
-		if err := cc.writeChunksWithFetchSize(ctx, rs, serverStatus, fetchSize); err != nil {
+		crs, ok := rs.(cursorResultSet)
+		if !ok {
+			// this branch is actually unreachable
+			return false, errors.New("this cursor is not a resultSet")
+		}
+		if err := cc.writeChunksWithFetchSize(ctx, crs, serverStatus, fetchSize); err != nil {
 			return false, err
 		}
 		return false, cc.flush(ctx)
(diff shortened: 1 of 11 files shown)


server/conn.go
```diff
@@ -2394,61 +2399,62 @@ func (cc *clientConn) writeChunks(ctx context.Context, rs ResultSet, binary bool
...
 // serverStatus, a flag bit represents server information.
 // fetchSize, the desired number of rows to be fetched each time when client uses cursor.
-func (cc *clientConn) writeChunksWithFetchSize(ctx context.Context, rs ResultSet, serverStatus uint16, fetchSize int) error {
-	fetchedRows := rs.GetFetchedRows()
+func (cc *clientConn) writeChunksWithFetchSize(ctx context.Context, rs cursorResultSet, serverStatus uint16, fetchSize int) error {
+	iter := rs.GetRowIterator()
+
+	// construct the rows sent to the client according to fetchSize.
+	var curRows []chunk.Row
+	for i := 0; i < fetchSize && iter.Current() != iter.End(); i++ {
```
YangKeao, line 2407: This performance is toooooooo bad (because of the redundant chunk allocation). I'm optimizing it right now.

util/chunk/row_container_reader.go
```diff
@@ -0,0 +1,169 @@
...
+	close(iter.rowCh)
+}
+
+// NewForwardOnlyIterator4RowContainer creates a forward only iterator for row container
+func NewForwardOnlyIterator4RowContainer(rc *RowContainer) *forwardOnlyIterator4RowContainer {
+	iter := &forwardOnlyIterator4RowContainer{
+		rc: rc,
+	}
+	iter.initializeChannel()
+	go iter.startWorker()
```
YangKeao, line 165: This goroutine leaks. I'll need to add a method to close it. (Or simply observe the status of rowContainer and close it automatically, to keep the interface simpler).

YangKeao:
Github doesn't update this PR after I rebase and solve the conflict. It's really weird. The content of this PR is different from the content in my branch: https://github.com/YangKeao/tidb/tree/cursor-fetch-persist (e.g. the file `/util/chunk/row_container_reader.go`)

YangKeao:
Good. Then it doesn't allow me to reopen it :face_exhaling: . I have to create a new one.

image

PR #44760, opened 2023-06-19
ddl/ingest: fix checkpoint resume when owner changes several times

### LINKED ISSUE #44619: data index inconsistent after DDL owner change several times (at least twice)
## Bug Report

1. deploy a cluter with 3 tidb instance
2. create a partition table and insert some data
3. add index for the table use ingest mode.
4. during add index, upgrade the cluster or resign the ddl owner several times, at least twice
5. after cluster upgraded, admin check index

admin check index pass
check index failed
```sql
error="[admin:8003]table count 10000800 != index(idx1) count 4666669"
```
图片

v7.1.0

tangenta:

Issue Number: close #44619

Problem Summary:

When the owner restarts to check if owner instance address has changed, the partitition ID of the checkpoint record will not be written into the checkpoint manager.

This results in the partition ID not being restored to the global checkpoint when restoring adding index job, and finally missing part of the partition's index records.

Regardless of whether the owner instance address has changed, always record the partition ID of the checkpoint to the checkpoint manager

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


Benjamin2037:
Could you add some brief description about the root cause?

ddl/index.go
```diff
@@ -1850,6 +1850,10 @@ func (w *worker) addTableIndex(t table.Table, reorgInfo *reorgInfo) error {
 			if err != nil {
 				break
 			}
+			w.ddlCtx.mu.RLock()
```
Benjamin2037, line 1853: why here do we need a. read lock?  can we remove lock?

tangenta: This is how we use the hook. There is only one hook instance globally.

ddl/index.go
```diff
@@ -1850,6 +1850,10 @@ func (w *worker) addTableIndex(t table.Table, reorgInfo *reorgInfo) error {
...
+			w.ddlCtx.mu.hook.OnUpdateReorgInfo(reorgInfo.Job, reorgInfo.PhysicalTableID)
+			w.ddlCtx.mu.RUnlock()
```
Benjamin2037, line 1855:
```suggestion
			defer w.ddlCtx.mu.RUnlock()
```
to ensure release lock.

ddl/util/callback/callback.go
```diff
@@ -58,6 +58,8 @@ type TestDDLCallback struct {
 	OnGetJobBeforeExported  func(string)
 	OnGetJobAfterExported   func(string, *model.Job)
 	OnJobSchemaStateChanged func(int64)
+
```
Benjamin2037, line 61:
```suggestion
```

wjhuang2016 (approved):

ddl/index.go
tangenta: No, inside the loop we cannot use "defer". Otherwise it will block `RLock` forever.

ddl/util/callback/callback.go
tangenta: The line break is used to distinguish plain DDL hooks and reorg hooks.

ywqzzy (approved):

Benjamin2037 (approved):
LGTM

PR #44764, opened 2023-06-19
binlog: fix show pump/drainer status

lichunzhu:

Issue Number: ref https://github.com/pingcap/tidb/issues/42643

Problem Summary:

In the past code we directly use tidb's global config path to create etcd registry. After this PR we use binlog client's etcd registry to do the following staff. Binlog client will update etcd urls in time.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

Side effects

- None

Documentation

- None

### Release note

```release-note
None
```

### DIFF
--- a/executor/show.go
+++ b/executor/show.go
@@ -55,6 +55,7 @@ import (
 	"github.com/pingcap/tidb/privilege"
 	"github.com/pingcap/tidb/privilege/privileges"
 	"github.com/pingcap/tidb/sessionctx"
+	"github.com/pingcap/tidb/sessionctx/binloginfo"
 	"github.com/pingcap/tidb/sessionctx/sessionstates"
 	"github.com/pingcap/tidb/sessionctx/stmtctx"
 	"github.com/pingcap/tidb/sessionctx/variable"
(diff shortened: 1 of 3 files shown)


GMHDBJD (approved):
LGTM

executor/show.go
```diff
@@ -1900,18 +1903,21 @@ func (e *ShowExec) fetchShowPumpOrDrainerStatus(kind string) error {
 	return nil
 }
 
-// createRegistry returns an ectd registry
-func createRegistry(urls string) (*node.EtcdRegistry, error) {
+// createRegistry returns an ectd registry, need to close, and error
+func createRegistry(urls string) (*node.EtcdRegistry, bool, error) {
+	if pumpClient := binloginfo.GetPumpsClient(); pumpClient != nil && pumpClient.EtcdRegistry != nil {
```
okJiang, line 1908: Do we need tolerate `pumpClient` is nil?

executor/show.go
```diff
@@ -1876,16 +1877,18 @@ func (e *ShowExec) fetchShowWarnings(errOnly bool) error {
 
 // fetchShowPumpOrDrainerStatus gets status of all pumps or drainers and fill them into e.rows.
 func (e *ShowExec) fetchShowPumpOrDrainerStatus(kind string) error {
-	registry, err := createRegistry(config.GetGlobalConfig().Path)
+	registry, needToClose, err := createRegistry(config.GetGlobalConfig().Path)
```
okJiang, line 1880: Is it better to use a new function? This client, I think, is used for Pump/Drainer specifically.

executor/show.go
okJiang: `createRegistry` is used in `excutor/change.go` too. Should it use the `PumpsClient`?

executor/show.go
lichunzhu: How about changing function's name to avoid abuse?

okJiang (approved):

PR #44801, opened 2023-06-19
lightning: refine fetching table structure error handling

### LINKED ISSUE #44614: tidb-lightning tidb backend import tableA but  report Error 1146: Table 'tableB' doesn't exist
## Bug Report
image

during tidb-lightning import another table dropped by someone else.

image

import successful !

tidb-lighting report another  table not exist.

tidb-lightning v6.5.2

dsdashun:

Issue Number: close #44614 

Problem Summary:

If any errors are encountered while fetching column auto information, log a warning and refrain from adding the table structure to the table structure map. This approach allows the error to be ignored if the table is not relevant, making the importing process proceed without any issues.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


br/pkg/lightning/backend/tidb/tidb.go
```diff
@@ -212,15 +214,24 @@ func (b *targetInfoGetter) FetchRemoteTableModels(ctx context.Context, schemaNam
...
 		// `show table next_row_id` is also not available before tidb v4.0.0
 		if serverInfo.ServerType != version.ServerTypeTiDB || serverInfo.ServerVersion.Major < 4 {
+			results = tables
 			return nil
 		}
 
+		failpoint.Inject(
+			"FetchRemoteTableModels_BeforeFetchTableAutoIDInfos",
+			func() {
+				fmt.Println("failpoint: FetchRemoteTableModels_BeforeFetchTableAutoIDInfos")
```
lance6716, line 224: we can use nil inject function when the inject verb is sleep, like `FailAfterWriteRows`

br/pkg/lightning/importer/get_pre_info.go
```diff
@@ -365,6 +365,19 @@ func (p *PreImportInfoGetterImpl) GetAllTableStructures(ctx context.Context, opt
 
 func (p *PreImportInfoGetterImpl) getTableStructuresByFileMeta(ctx context.Context, dbSrcFileMeta *mydump.MDDatabaseMeta, getPreInfoCfg *ropts.GetPreInfoConfig) ([]*model.TableInfo, error) {
 	dbName := dbSrcFileMeta.Name
+	failpoint.Inject(
+		"getTableStructuresByFileMeta_BeforeFetchRemoteTableModels",
```
lance6716, line 369: Can the caller directly set `FetchRemoteTableModels_BeforeFetchTableAutoIDInfos`?

br/pkg/lightning/backend/tidb/tidb.go
```diff
@@ -141,10 +141,11 @@ func NewTargetInfoGetter(db *sql.DB) backend.TargetInfoGetter {
 // TODO: refactor
 func (b *targetInfoGetter) FetchRemoteTableModels(ctx context.Context, schemaName string) ([]*model.TableInfo, error) {
```
lance6716, line 142: In the lightning logic, is it easy to pass the table-level allow list to this function, so we don't need to care about unrelated tables?

No need to change in this PR

br/pkg/lightning/importer/get_pre_info.go
dsdashun: I have previously attempted this approach where I set a sleep timer of 5 seconds to activate that failpoint. During the integration test, the failpoint was triggered twice, resulting in a total of 10 seconds of delay. However, I was only concerned with the second trigger. In this second trigger, I needed to complete the `drop table` operation within the sleep period. To achieve this, I added an 8-second sleep timer in the integration test script (5 < 8 < 10). Although this approach worked, it lacked flexibility. If there were more or fewer calls to `FetchRemoteTableModels`, I would need to adjust the sleep timer accordingly.

Therefore, I now use a different approach where I introduce a failpoint here that enables the specific failpoint I want to sleep. This way, I only need to adjust the sleep time together with the total sleep time of a single failpoint trigger, making the behavior more controllable.

br/pkg/lightning/backend/tidb/tidb.go
dsdashun: Yes, I understand your point. I print that out to verify in the integration test script that my `drop table` is executed before this failpoint's hook is triggered. This makes it easier to debug. Do you think it's OK to keep the printing message here?

br/pkg/lightning/importer/get_pre_info.go
lance6716: I'm not sure this approach, are we currently equivalently set `sleep(6000)` for `FetchRemoteTableModels_BeforeFetchTableAutoIDInfos`? so we can directly set it.

And another suggestion is maybe we let `FetchRemoteTableModels_BeforeFetchTableAutoIDInfos` use `b.db` to drop the table, to mimic the user action happens at that point 😂  so it's no longer related to sleep duration

dsdashun: > I'm not sure this approach, are we currently equivalently set `sleep(6000)` for `FetchRemoteTableModels_BeforeFetchTableAutoIDInfos`? so we can directly set it.

If directly enable `FetchRemoteTableModels_BeforeFetchTableAutoIDInfos` with `sleep(6000)`, it will actually sleep 12 seconds.  The first 6 seconds is from here: 
https://github.com/pingcap/tidb/blob/e3417913f58cdd5a136259b902bf177eaf3aa637/br/pkg/lightning/importer/import.go#L807 =>
https://github.com/pingcap/tidb/blob/e3417913f58cdd5a136259b902bf177eaf3aa637/br/pkg/lightning/importer/import.go#L601

At this time, the table schemas have not been created, and the errors are totally ignored.

The second 6 seconds is from here: 
https://github.com/pingcap/tidb/blob/e3417913f58cdd5a136259b902bf177eaf3aa637/br/pkg/lightning/importer/import.go#L813 =>
https://github.com/pingcap/tidb/blob/e3417913f58cdd5a136259b902bf177eaf3aa637/br/pkg/lightning/importer/get_pre_info.go#L356-L358 =>
https://github.com/pingcap/tidb/blob/e3417913f58cdd5a136259b902bf177eaf3aa637/br/pkg/lightning/importer/get_pre_info.go#L368

At this time, the table schemas have been created.

For my approach, I initially did not enable the `FetchRemoteTableModels_BeforeFetchTableAutoIDInfos` failpoint. After creating the schemas, I enabled the failpoint here to ensure that the tables have been created during the sleep time of the failpoint. This way, any errors will not be completely ignored.

For the alternative method you mentioned, we still need a precise way to trigger the failpoint only once during the second attempt, ensuring that the error will not be ignored. If the `FetchRemoteTableModels_BeforeFetchTableAutoIDInfos` failpoint is triggered twice and the `drop table` is executed twice, the following will occur:
- The first time, the `drop table` is executed successfully, but the generated error is ignored.
- The second time, the `drop table` cannot be executed successfully because the table has already been dropped. We won't encounter any error because the `drop table` is not executed halfway.

dsdashun: > And another suggestion is maybe we let `FetchRemoteTableModels_BeforeFetchTableAutoIDInfos` use `b.db` to drop the table, to mimic the user action happens at that point 😂 so it's no longer related to sleep duration

This suggestion is technically feasible and presents a good option. However, I am currently struggling with whether to include "DROP TABLE" in the script or in the failpoint hook.

- If we write "DROP TABLE" in the script, it offers flexibility as only the script needs to be modified. However, this approach also has some drawbacks, such as unnecessary additional sleep time that may cause uncontrollable execution sequences. We would then need to tune the sleep time in the script.
- On the other hand, if we write "DROP TABLE" in the failpoint, it eliminates the need for additional sleep time and provides controllable logic. However, it may feel like we are coupling the testing logic into the Lightning code.

Which approach do you prefer?

lance6716: currently I prefer we use failpoint to do it. Let's wait the option from the second reviewer

br/pkg/lightning/backend/tidb/tidb.go
lance6716: Can you check if this idea will work? But we don't need to change this PR

lance6716 (approved):

GMHDBJD (approved):
LGTM

Issue #44795, opened 2023-06-19
got wrong result when enable `tidb_opt_agg_push_down`


aytrack:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
1. deploy a cluster 
2. prepare tpch data with 1sf `tiup bench tpch prepare -T 10`
3. execute q13
```sql
set tidb_opt_agg_push_down=ON;
SELECT c_count, count(*) as custdist
from ( SELECT c_custkey, count(o_orderkey)  as  c_count
       from customer left join orders on c_custkey = o_custkey and o_comment not like '%special%requests%'
       group by c_custkey ) c_orders
group by c_count
order by custdist desc, c_count desc;
```
<!-- a step by step guide for reproducing the bug. -->

### 2. What did you expect to see? (Required)
```sql
[17:41:41]TiDB root:test> SELECT c_count, count(*) as custdist
                       -> from ( SELECT c_custkey, count(o_orderkey)  as  c_count
                       ->        from customer left join orders on c_custkey = o_custkey and o_comment not like '%special%requests%'
                       ->        group by c_custkey ) c_orders
                       -> group by c_count
                       -> order by custdist desc, c_count desc;
+---------+----------+
| c_count | custdist |
+---------+----------+
| 0       | 50005    |
| 9       | 6641     |
| 10      | 6532     |
| 11      | 6014     |
| 8       | 5937     |
| 12      | 5639     |
| 13      | 5024     |
| 19      | 4793     |
| 7       | 4687     |
| 17      | 4587     |
| 18      | 4529     |
| 20      | 4516     |
| 15      | 4505     |
| 14      | 4446     |
| 16      | 4273     |
| 21      | 4190     |
| 22      | 3623     |
| 6       | 3265     |
| 23      | 3225     |
| 24      | 2742     |
| 25      | 2086     |
| 5       | 1948     |
| 26      | 1612     |
| 27      | 1179     |
| 4       | 1007     |
| 28      | 893      |
| 29      | 593      |
| 3       | 415      |
| 30      | 376      |
| 31      | 226      |
| 32      | 148      |
| 2       | 134      |
| 33      | 75       |
| 34      | 50       |
| 35      | 37       |
| 1       | 17       |
| 36      | 14       |
| 38      | 5        |
| 37      | 5        |
| 40      | 4        |
| 41      | 2        |
| 39      | 1        |
+---------+----------+
42 rows in set
```

[...]

PR #44941, opened 2023-06-26
planner: fix agg elimination logic after agg pushed down through a join

AilinKid:

Issue Number: close https://github.com/pingcap/tidb/issues/44795

Problem Summary:

code in `aggregationPushDownSolver` for the switch `p.SCtx().GetSessionVars().AllowAggPushDown` path is too old to use. 

briefly speaking, when we try to push agg down through a join (that's what aggregationPushDownSolver does) and try to combine some aggregation elimination (for example, group item covering unique key, then the aggregation itself can be eliminated):

```
buildKeyInfo(join)
proj := a.tryToEliminateAggregation(agg, opt). // here use an old pointer, whose child is already changed during the logic above
if proj != nil {
	p = proj
}
```
the comment place is actually to eliminate the new pushed-down agg since agg's children have changed, maybe some new unique key can be detected to eliminate itself. Or we should say, the old agg elimination logic is quite different from normal agg eliminations (see comments for more detail)

image

old agg rewriting still uses the `ifnull(col#19, 0, 1)`'s logic, taking every row as a count 1, while after the agg has been pushed down, the col#19 here is already the final aggregation result, rather than in the process of aggregation. Keep it real is the true path, while for now, we banned this kind of old agg elimination logic.

so the handling logic is a mess.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
scripts:

sql = $below
```
SELECT c_count, count(*) as custdist
from ( SELECT c_custkey, count(o_orderkey)  as  c_count
       from customer left join orders on c_custkey = o_custkey and o_comment not like '%special%requests%'
       group by c_custkey ) c_orders
group by c_count
order by custdist desc, c_count desc;
```
execute $sql + "into outfile 'tai1.txt'" under set `tidb_opt_agg_push_down=ON;`
execute $sql + "into outfile 'tai2txt'" under set `tidb_opt_agg_push_down=OFF;`

diff `tai1.txt` `tai2.txt` to see nothing strange

- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
planner: fix agg elimination logic after agg pushed down to join
```

### DIFF
--- a/planner/core/rule_aggregation_elimination.go
+++ b/planner/core/rule_aggregation_elimination.go
@@ -32,13 +32,23 @@ type aggregationEliminator struct {
 }
 
 type aggregationEliminateChecker struct {
+	// in agg pushed down cases, for example:
+	// agg -> join -> datasource1
+	//             -> datasource2
+	// we just make a new agg upon datasource1 or datasource2, while the old agg is still existed and waiting for elimination.
+	// Note when the old agg is like below, and join is an outer join type, rewriting old agg in elimination logic has some problem.
+	// eg:
+	// count(a) -> ifnull(col#x, 0, 1) in rewriteExpr of agg function, since col#x is already the final pushed-down aggregation's
+	// result from new join schema, we don't need to take every row as count 1 when they don't have not-null flag in a.tryToEliminateAggregation(oldAgg, opt),
+	// which is not suitable here.
+	oldAggEliminationCheck bool
 }
 
 // tryToEliminateAggregation will eliminate aggregation grouped by unique key.
 // e.g. select min(b) from t group by a. If a is a unique key, then this sql is equal to `select b from t group by a`.
 // For count(expr), sum(expr), avg(expr), count(distinct expr, [expr...]) we may need to rewrite the expr. Details are shown below.
 // If we can eliminate agg successful, we return a projection. Else we return a nil pointer.
-func (*aggregationEliminateChecker) tryToEliminateAggregation(agg *LogicalAggregation, opt *logicalOptimizeOp) *LogicalProjection {
+func (a *aggregationEliminateChecker) tryToEliminateAggregation(agg *LogicalAggregation, opt *logicalOptimizeOp) *LogicalProjection {
 	for _, af := range agg.AggFuncs {
 		// TODO(issue #9968): Actually, we can rewrite GROUP_CONCAT when all the
 		// arguments it accepts are promised to be NOT-NULL.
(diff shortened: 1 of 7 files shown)


planner/core/rule_aggregation_elimination.go
```diff
@@ -64,6 +74,11 @@ func (*aggregationEliminateChecker) tryToEliminateAggregation(agg *LogicalAggreg
 		}
 	}
 	if coveredByUniqueKey {
+		if a.oldAggEliminationCheck {
+			if !CheckCanConvertAggToProj(agg) {
```
winoros, line 78: And this check can always take effect, instead of only for the pushed agg?

AilinKid: no.
agg1 -> join -> agg2
now the case is that agg1's work has been transferred to agg2, agg1 can be just removed.
while 'we' do the remove-work by aggEliminator, in which it will rewrite some agg func to a new projection, consequently causing problems.

the ideal approach is that to make the agg1's parent shift its args to what join can produce after agg2 is generated, otherwise, the can't-find-column error will occur. (currently too complicated for a quick fix)

we just keep the old agg1 remained, but its work is actually done by agg2, old-agg1 here acts like a role to project what schema the join produced to what the old-agg1 parent wanted as before.

planner/core/rule_aggregation_elimination.go
```diff
@@ -32,13 +32,23 @@ type aggregationEliminator struct {
 }
 
 type aggregationEliminateChecker struct {
+	// in agg pushed down cases, for example:
```
elsa0520, line 35:
```suggestion
	// used for agg pushed down cases, for example:
```

planner/core/rule_aggregation_elimination.go
```diff
@@ -138,6 +151,33 @@ func appendDistinctEliminateTraceStep(agg *LogicalAggregation, uniqueKey express
 	opt.appendStepToCurrent(agg.ID(), agg.TP(), reason, action)
 }
 
+// CheckCanConvertAggToProj check whether a special old aggregation (which has already been pushed down) to projection.
```
elsa0520, line 154:
```suggestion
// CheckCanConvertAggToProj check whether a special old aggregation (which has already been pushed down) to projection.
// link: issue#44795
```

fixdb (approved):
+1

elsa0520 (approved):
LGTM

wshwsh12 (approved):

PR #44979, opened 2023-06-27
executor: add cache for approximate table count

### LINKED ISSUE #44978: Add cache mechanism to line count estimation to reduce backend load
## Enhancement

Currently, every time the line count estimation feature is called, a request is sent to the backend, which may cause a certain burden on the backend. To solve this problem, we can add a cache mechanism to store the results of line count estimation locally, thereby reducing the number of requests sent to the backend.

## Solution

We can add a local cache to store the results of line count estimation. When the line count estimation is called, first check whether there are corresponding results in the local cache. If so, return the results directly from the cache. Otherwise, send a request to the backend and store the results in the local cache.
To avoid cache expiration issues, we can set a cache time, and after this time, the cache will automatically expire and send a new request to the backend to get the latest line count estimation results.

## Implementation Details

Add a cache mechanism in the frontend code to store the results of line count estimation;
Check if there are corresponding results in the cache;
If there are, return the results directly from the cache;
If not, send a request to the backend and store the results in the cache;
Set a cache time, and after this time, the cache will automatically expire and send a new request to the backend to get the latest line count estimation results.

## Expected Outcome

By adding a cache mechanism, we can reduce the number of requests sent to the backend, thereby reducing the backend load, improving system performance and stability. At the same time, we can reduce user waiting time and improve user experience.

hawkingrei:

Issue Number: close #44978

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


executor/internal/pdhelper/pd.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func approximateTableCountKey(tid int64, dbName, tableName, partitionName string) string {
+	return strings.Join([]string{strconv.FormatInt(tid, 10), dbName, tableName, partitionName}, "_")
+}
+
+// GetApproximateTableCountFromStorage gets the approximate count of the table.
+func (p *PDHelper) GetApproximateTableCountFromStorage(sctx sessionctx.Context, tid int64, dbName, tableName, partitionName string) (float64, bool) {
+	key := approximateTableCountKey(tid, dbName, tableName, partitionName)
+	if item := p.cacheForApproximateTableCountFromStorage.Get(key); item != nil {
```
hawkingrei, line 73: ```GetApproximateTableCountFromStorage``` is the same as the old. but we add cache here.

tidb-server/main.go
```diff
@@ -760,6 +761,7 @@ func setGlobalVars() {
 	} else {
 		executor.GlobalMemoryUsageTracker.SetBytesLimit(int64(cfg.Performance.ServerMemoryQuota))
 	}
+	executor.Start()
```
time-and-fate, line 764: Why do we put this in `setGlobalVars`?
Seems a bit confusing.

executor/executor.go
```diff
@@ -183,6 +184,16 @@ func init() {
 	CheckTableFastBucketSize.Store(1024)
 }
 
+// Start the backend components
```
time-and-fate, line 187: I suggest add some comments since we are introducing the "backend components" new concept to this package.

executor/internal/pdhelper/pd.go
```diff
@@ -0,0 +1,123 @@
...
```
time-and-fate, line 77: I think we don't call `Start()` in tests. Is it safe to use it without calling `Start()`?

executor/internal/pdhelper/pd.go
```diff
@@ -0,0 +1,123 @@
...
+	)
+	return &PDHelper{
+		cacheForApproximateTableCountFromStorage: cache,
+	}
+}
+
+// Start starts the PDHelper.
+func (p *PDHelper) Start() {
+	globalPDHelperOnce.Do(func() {
+		p.wg.Run(p.cacheForApproximateTableCountFromStorage.Start)
```
chrysan, line 60: Is it possible to run Start in init()?

hawkingrei: ```init()``` is a bad code style.

refer https://github.com/uber-go/guide/blob/master/style.md#avoid-init

hawkingrei: it can be removed to ```Start()```.

executor/internal/pdhelper/pd.go
```diff
@@ -0,0 +1,123 @@
...
+	})
+	if err != nil {
+		return 0, false
+	}
+	// If this table is not small, we directly use the count from PD,
+	// since for a small table, it's possible that it's data is in the same region with part of another large table.
+	// Thus, we use the number of the regions of the table's table KV to decide whether the table is small.
+	if regionStats.Count > 2 {
+		result := float64(regionStats.StorageKeys)
+		p.cacheForApproximateTableCountFromStorage.Set(key, result, ttlcache.DefaultTTL)
```
chrysan, line 102: What will happen when capacity is full?

tidb-server/main.go
hawkingrei: I hope it is globally unique.

chrysan (commented):
There's no ut in this PR, please do functional testing for this cache.

executor/internal/pdhelper/pd.go
hawkingrei: Because it is globally unique, it started when it startup. ```Start``` is to start the eviction goroutine.it does not affect normal use/

executor/executor.go
chrysan: And we did lots of things in init() of executor. It seems that we should move those into Start()?

executor/internal/pdhelper/pd.go
chrysan: Looks like many "func init()" need to be revisited.

tidb-server/main.go
chrysan: Why not in main()?

time-and-fate: +1. My main points are:
1. Currently, we only set up global states, i.e. setting global variables in `setGlobalVars`. But `executor.Start()` starts a background goroutine, which is quite different.
2. If we call `executor.Start()` here and `executor.Stop()` in `main()`, they won't pair up and will make the code a bit confusing.

executor/executor.go
hawkingrei: Done

tidb-server/main.go
hawkingrei: Done

executor/internal/pdhelper/pd.go
hawkingrei: No, The Least Recently Used one will be deleted.

time-and-fate (approved):

hawkingrei:
@chrysan @qw4990 PTAL

chrysan (approved):

PR #45012, opened 2023-06-28
timer: support manual trigger for timer framework

### LINKED ISSUE #45009: Support trigger event manually for timer framework

lcwangchao:

Issue Number: close #45009

A new field `ManualRequest` is added to `TimerRecord` , it is defined as below

```
// ManualRequest is the request info to trigger timer manually
type ManualRequest struct {
	// ManualRequestID is the id of manual request
	ManualRequestID string
	// ManualRequestTime is the request time
	ManualRequestTime time.Time
	// ManualTimeout is the timeout for the request, if the timer is not triggered after timeout, processed will be set to true
	// with empty event id.
	ManualTimeout time.Duration
	// ManualProcessed indicates the request is processed (triggered or timeout)
	ManualProcessed bool
	// ManualEventID means the triggered event id for the current request
	ManualEventID string
}
```

We can invoke `client.ManualTriggerEvent` to trigger a timer event manually, after it the `ManualRequest` will be set with a request id. After the event is triggered,  `ManualEventID` will be set to the corresponding event id and `ManualProcessed` will be set to `true`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/timer/api/client.go
+++ b/timer/api/client.go
@@ -23,13 +23,14 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/pingcap/errors"
+	"github.com/pingcap/tidb/util"
 	"github.com/pingcap/tidb/util/logutil"
 	"go.uber.org/zap"
 )
 
 const (
-	clientMaxRetry      = 10
-	clientRetryInterval = 2 * time.Second
+	clientMaxRetry     = 5
+	clientRetryBackoff = uint64(1000)
 )
 
 // GetTimerOption is the option to get timers
@@ -131,17 +132,17 @@ const DefaultStoreNamespace = "default"
 
 // defaultTimerClient is the default implement of timer client
 type defaultTimerClient struct {
-	namespace     string
-	store         *TimerStore
-	retryInterval time.Duration
+	namespace    string
+	store        *TimerStore
+	retryBackoff uint64
 }
 
 // NewDefaultTimerClient creates a new defaultTimerClient
 func NewDefaultTimerClient(store *TimerStore) TimerClient {
 	return &defaultTimerClient{
-		namespace:     DefaultStoreNamespace,
-		store:         store,
-		retryInterval: clientRetryInterval,
+		namespace:    DefaultStoreNamespace,
+		store:        store,
+		retryBackoff: clientRetryBackoff,
 	}
 }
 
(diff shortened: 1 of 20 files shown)


timer/tablestore/sql.go
```diff
@@ -104,7 +109,7 @@ func buildInsertTimerSQL(dbName, tableName string, record *api.TimerRecord) (str
 		"EVENT_DATA, "+
 		"SUMMARY_DATA, "+
 		"VERSION) "+
-		"VALUES (%%?, %%?, %%?, 'TIDB', %%?, %%?, %%?, %s, %%?, %%?, %%?, %%?, %s, %%?, %%?, 1)",
+		"VALUES (%%?, %%?, %%?, 'TIDB', %%?, %%?, %%?, %s, %%?, JSON_MERGE_PATCH('{}', %%?), %%?, %%?, %s, %%?, %%?, 1)",
```
YangKeao, line 112: Why is it a `JSON_MERGE_PATCH` but not `cast(? as JSON)`?

YangKeao (commented):
Rest LGTM

timer/api/client.go
```diff
@@ -200,9 +248,27 @@ func (c *defaultTimerClient) CloseTimerEvent(ctx context.Context, timerID string
...
+	update.EventExtra.Set(EventExtra{})
 	return c.store.Update(ctx, timerID, update)
 }
 
 func (c *defaultTimerClient) DeleteTimer(ctx context.Context, timerID string) (bool, error) {
 	return c.store.Delete(ctx, timerID)
 }
+
+func (c *defaultTimerClient) retryWhenVersionNotMatch(ctx context.Context, timerID string, fn func() error) (err error) {
+	for i := 0; i <= clientMaxRetry; i++ {
```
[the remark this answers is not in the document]
lcwangchao: That depends on how to define retry. Here retry count will not include the first action. For example, if `clientMaxRetry` is 0, it means not retry but still have to do it once.

timer/tablestore/sql.go
lcwangchao: That is because `timerExt` object will be used in both insert and update. If we simply cast it to json. We will write some null value to it. For example, when update, the "manual" field should be serialized to `{"request_id": "123", "event_id": null, ...}` to make a full update and to make sure "event_id" is removed. And in insert scene, we want to keep the same behavior

YangKeao (approved):
LGTM

timer/api/client.go
```diff
@@ -200,9 +248,27 @@ func (c *defaultTimerClient) CloseTimerEvent(ctx context.Context, timerID string
...
 	}
+	update.EventExtra.Set(EventExtra{})
 	return c.store.Update(ctx, timerID, update)
 }
 
 func (c *defaultTimerClient) DeleteTimer(ctx context.Context, timerID string) (bool, error) {
 	return c.store.Delete(ctx, timerID)
 }
+
+func (c *defaultTimerClient) retryWhenVersionNotMatch(ctx context.Context, timerID string, fn func() error) (err error) {
```
bb7133, line 259: Could you try to reuse the existing retry functions like [RunWithRetry](https://github.com/pingcap/tidb/blob/e76a3f363ab6335b333b8e89e1c95a7333a18696/util/misc.go#L66)?

timer/api/store_test.go
```diff
@@ -240,7 +240,7 @@ func TestTimerUpdate(t *testing.T) {
 	require.True(t, errors.ErrorEqual(err, ErrEventIDNotMatch))
 	require.Equal(t, tpl, *tm)
 
-	// test apply without check
+	// test apply without check1
```
bb7133, line 243: Is this a typo?

timer/api/timer.go
```diff
@@ -64,6 +64,43 @@ func (p *SchedIntervalPolicy) NextEventTime(watermark time.Time) (time.Time, boo
 	return watermark.Add(p.interval), true
 }
 
+// ManualRequest is the request info to trigger timer manually
```
bb7133, line 67:
```suggestion
// ManualRequest is the request info to trigger timer manually.
```
Ditto for the other comments.

timer/api/store_test.go
lcwangchao: No, there is check2, check3 later. But I'll update the comment message to make it easier to read

bb7133 (approved):
LGTM

Issue #35744, opened 2022-06-27
BR restore failed: error="cannot find rewrite rule: [BR:Restore:ErrRestoreInvalidRewrite]invalid rewrite rule"


Tammyxia:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

- restore sysbench data from minio to TiDB cluster upstream and downstream separately :  

> 2022-06-25T18:08:31.177Z	INFO	host/host.go:30	Execute command	{"command": " /br  restore  full \"-s\" \"s3://nfs/lixia-tools-workload/BR-sysbench-tmp?access-key=minioadmin&secret-access-key=minioadmin&endpoint=http%3a%2f%2fminio.pingcap.net%3a9000&force-path-style=true\" \"-u\" \"http://downstream-pd.cdc-testbed-tps-813939-1-87:2379\" \"--check-requirements=false\"", "timeout": "1h", "resource name": "br"}

> 2022-06-25T18:08:31.181Z	INFO	host/host.go:30	Execute command	{"command": " /br  restore  full \"-s\" \"s3://nfs/lixia-tools-workload/BR-sysbench-tmp?access-key=minioadmin&secret-access-key=minioadmin&endpoint=http%3a%2f%2fminio.pingcap.net%3a9000&force-path-style=true\" \"-u\" \"http://upstream-pd.cdc-testbed-tps-813939-1-87:2379\" \"--check-requirements=false\"", "timeout": "1h", "resource name": "br"}

### 3. What did you see instead (Required)

- restore to upstream cluster successfully, but restore to downstream cluster failed:
   Error: cannot find rewrite rule: [BR:Restore:ErrRestoreInvalidRewrite]invalid rewrite rule\n", "error": "command terminated with exit code 1"

### 4. What is your TiDB version? (Required)

["Welcome to Backup & Restore (BR)"] [release-version=v6.1.0-nightly] [git-hash=ce7b541b2d3efd0120f9bec62db3f5088c723482] [git-branch=heads/refs/tags/v6.1.0-nightly] [go-version=go1.18.2] [utc-build-time="2022-06-24 11:01:21"] [race-enabled=false]

3pointer:
looks like the backup files has some problems.

overvenus:
This issue is caused by restoring to an existing table and the table lacks an index.

Reproduced with following steps

```shell
# Create a table with an index.
[...]

PR #45014, opened 2023-06-28
restore: precheck cluster is empty when first time full restore

3pointer:

Issue Number: close https://github.com/pingcap/tidb/issues/35744

Problem Summary:

1. Add pre-check for the first time FullRestore.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Add a cluster empty precheck for full restore command
```


Leavrth (approved):

br/pkg/task/restore.go
```diff
@@ -769,6 +753,28 @@ func runRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
 		}()
 	}
 
+	if isFullRestore(cmdName) {
+		// we need check cluster is fresh every time. except restore from a checkpoint.
+		if client.IsFull() && len(checkpointSetWithTableID) == 0 {
+			if err = client.CheckTargetClusterFresh(ctx); err != nil {
```
YuJuncen, line 759: This will make BR fail once there are any table in the cluster. (Even there isn't intersection of them and the backup archive.) Will this break some use cases?

YuJuncen (commented):
Generally LGTM.

br/pkg/task/restore.go
3pointer: Yes, It's intended  @benmaoer has confirmed it. so this behaviour will introduced in v7.3. and won't cherry-pick back to any release versions.

benmaoer: Yes,  there are 2 options:

1.  When restoring the entire cluster from a full backup, it is necessary to check whether the target cluster is an empty cluster, which has no user tables. 

2. When restoring specific tables or databases from the backup, it is necessary to detect if there are any conflicts between the tables to be restored and the existing tables in the target cluster. If conflicts are found, an error should be reported in advance to avoid data integrity issues.

YuJuncen (approved):

br/pkg/task/restore.go
dbakit: Sometimes `br restore full` don't mean I realy want to resotre the entire cluster. In certain scenarios, I just want to restore multiple repositories, but they are distributed across different databases, or just want to use regular expressions.
So, can you add a `--force` option to skip this check??

3pointer: I think maybe we should skip the check if user has specified `--filter` argument already. @benmaoer @YuJuncen  WDYT?

Issue #43249, opened 2023-04-20
Optimize performance for `UnionScanExec` and `MemBuffer`


lcwangchao:
## Enhancement

The UnionScan is not very efficient, for example:

- UnionScanExec will filter all rows/indexes in `Open` instead of `Next`. It will take more time when the SQL has a `LIMIT` because `Open` will not considerate it.
- Maybe we can implement`MemBuffer` more efficient.

The union scan's performance will affect the performance of temporary table, cached table and queries in a txn with a lot of uncommitted rows.

tiancaiamao:
```
cd executor;
 go test -run XXX -bench BenchmarkUnionScan -cpuprofile cpu.out -benchtime 45s
```

![image](https://github.com/pingcap/tidb/assets/1420062/3d0883d2-e9f3-4e0b-9a33-344097eb575c)


The inefficency comes from too parts:

1. `Open()` always drain all the data even it might be useless in the `Limit` scenario. It's better to make it a streaming API
2. The decode / encode and row format translation cost. When loading, we decode kv -> row in []Datum representation. and when merging, []Datum -> []Datum, and finally output is translated from row representation to chunk, i.e. []Datum -> chunk

tiancaiamao:
```
cd pkg/executor;
go test -tags intest -run XXX -bench BenchmarkUnionScanIndexReadDescRead -benchmem -benchtime 30s

Before:
   79682            447983 ns/op          135108 B/op       1920 allocs/op
After:
   81997            432501 ns/op          124309 B/op       1814 allocs/op
```

The bench test only buffer 100 rows, so the result differ is not significant enough https://github.com/pingcap/tidb/blob/1c8d383487172801d600c7e96d84794d87edd72c/pkg/executor/union_scan_test.go#L330-L332

tiancaiamao:
```
make ut X='run pkg/executor --coverprofile cover.out
make failpoint-disable
go tool cover -html cover.out
```

The coverage of mem_reader.go is 58.3%, but the code is this PR are almost included.
What's missing is the IndexMerge code branch, there are no UT cover that.

ekexium:
Hey @lcwangchao do you have more elaboration in "Maybe we can implementMemBuffer more efficient."? The flamegraph shows that time spent in memdb is quite little. Did you mean codec-related work?

tiancaiamao:
[...]

PR #45051, opened 2023-06-29
executor: union scan refactor, optimize the membuffer decoding process

tiancaiamao:

Issue Number: ref #43249

Problem Summary:

The decoding of kv data from membuffer is very inefficient, this commit try to optimize it.

Use `ChunkDecoder` to replace the `ByteDecoder` implementaion.

The old `ByteDecoder` way is quite inefficient.
It decodes the key-value data to []bytes (in the new row encoding format), then decode []bytes to `[]types.Datum`
To evaluate the filter condition, it then translate `[]types.Datum` to `chunk.MutRow` and then to  `chunk.Row`  for expression evaluation.
As you can see, the whole process is quite inefficient and involves a lot of allocation.

The **new implementation is 10x better than before** considering the object allocation.

```
Before vs After:
cd executor
go test -tags intest -run XXX -bench BenchmarkUnionScanRead  -benchmem -cpuprofile cpu1.out -benchtime 45s
    2091          25964141 ns/op        17071121 B/op     334832 allocs/op
    5428           9718020 ns/op         1665827 B/op      34998 allocs/op
```

FlameGrap before vs after:

image
image

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


executor/mem_reader.go
```diff
@@ -231,6 +248,7 @@ func buildMemTableReader(ctx context.Context, us *UnionScanExec, kvRanges []kv.K
 		buffer: allocBuf{
 			handleBytes: make([]byte, 0, 16),
 			rd:          rd,
+			cd:          cd,
```
Defined2014, line 251: how about use `NewRowDecoder` here, maybe a litter bit slow, but it's simpler and less error-prone

Defined2014 (approved):

executor/mem_reader.go
tiancaiamao: Done

executor/mem_reader.go
```diff
@@ -196,17 +197,24 @@ func buildMemTableReader(ctx context.Context, us *UnionScanExec, kvRanges []kv.K
 		colIDs[col.ID] = i
 	}
 
+	var pkColIDs []int64
```
Defined2014, line 200: After f7ded1dc44a2d9d91258c76ad294001132e1e198, I think `pkColIDs` related columns needn't anymore.

executor/builder.go
```diff
@@ -5115,6 +5115,13 @@ func NewRowDecoder(ctx sessionctx.Context, schema *expression.Schema, tbl *model
 		}
 	}
 	defVal := func(i int, chk *chunk.Chunk) error {
+		if reqCols[i].ID < 0 {
+			// model.ExtraHandleID, ExtraPidColID, ExtraPhysTblID... etc
+			// Don't set the default value for that column.
+			chk.AppendNull(i)
```
lcwangchao, line 5121: Do these codes only affect the rows in membuffer? If so, could you have a future explanation because I still not know why not to set the default value for it

lcwangchao (approved):

Issue #44771, opened 2023-06-19
Configurable Replica Read Timeout with Retry


Tema:
## Configurable Replica Read Timeout with Retry Feature Request

**Is your feature request related to a problem? Please describe:**
One of common problems running TiDB in the cloud on network attached disks (Amazon EBS, Google PD or Azure managed disks) is temporary elevated disk IO latency. This can happen if  a cloud provider storage node fails and goes through a repair procedure. During the repair phase a network attached disk exhibits 100ms or even single digit second latency vs single digit millisecond latency under the normal conditions.

**Describe the feature you'd like:**
If TiDB customer uses Follower Read or Stale Read feature, it is possible to retry a request initially landed on the TiKV node with network disk exhibiting elevated latency on the other TiKV replica. While retry policy already exists in [tikv go-client](https://github.com/tikv/client-go), the default network timeout is [10s of seconds](https://github.com/tikv/client-go/blob/ce9203ef66e99dcc8b14f68777c520830ba99099/internal/client/client.go#L82-L83).   

OLTP workload on TiDB could leverage an introduction of system variable `tidb_tikv_read_timeout` which then could be passed as a context timeout for TiKV requests made by TiDB layer and rely on existing selector logic to retry requests on other replicas. The implementation of this feature needs also to take care of the following:   
* The intermediate timeouts should still bubble up in metrics
* Back-off should be skipped on this timeout
* Right now Stale Read retry always goes to the leader (https://github.com/tikv/client-go/pull/765). This needs to be changed so that it retries on leader only for `DataIsNot` ready error, but after timeout it should go to another replica.

**Describe alternatives you've considered:**
TiDB already has a `max_execution_timeout` system variable, but it is not used as a context deadline in go-client to network calls from TiDB to TiKV. Moreover, if TiKV request takes longer than `max_execution_timeout`, then the [session is marked as killed](https://github.com/tikv/client-go/blob/a78518eb37dafb3c40f1d4e9d97aa1a6ecf62978/internal/locate/region_request.go#L1051-L1055) and retry won’t happen.

**Teachability, Documentation, Adoption, Migration Strategy:**
The feature would be fully controlled by session variable `tidb_tikv_read_timeout`.

cfzjywxk:
The RFC: https://github.com/pingcap/tidb/blob/7105505a78fc886c33258caa5813baf197b15247/docs/design/2023-06-30-configurable-kv-timeout.md
[...]

PR #45093, opened 2023-06-30
doc: rfc for configurable kv timeout

cfzjywxk:

Issue Number: ref #44771 

Problem Summary:
Add the detailed design document for the configurable kv timeout proposal.

### Check List

Tests 

Side effects

Documentation

### Release note

```release-note
None
```

### DIFF
--- a/docs/design/2023-06-30-configurable-kv-timeout.md
+++ b/docs/design/2023-06-30-configurable-kv-timeout.md
@@ -3,9 +3,9 @@
 * Authors: [cfzjywxk](https://github.com/cfzjywxk)
 * Tracking issue: TODO
 
-## Motivation
+## Motivation & Background
 
-The associated TiKV requests are typically processed quickly, taking only a few milliseconds. However, if 
+TiKV requests are typically processed quickly, taking only a few milliseconds. However, if 
 there is disk IO or network latency jitter on a specified TiKV node, the duration of these requests may spike 
 to several seconds or more. This usually happens when there is an EBS issue like the EBS is repairing its 
 replica.
@@ -19,6 +19,16 @@ it could not retry or redirect the requests to other nodes quickly as the timeou
 tens of seconds. As a result, the application may experience a query latency spike when TiKV node jitter is 
 encountered.
 
+For latency sensitive applications, providing predictable sub-second read latency by set fast retrying
+is valuable. queries that are usually very fast (such as point-select query), setting the value
+of `tidb_kv_read_timeout` to short value like  `500ms`, the TiDB cluster would be more tolerable
+for network latency or io latency jitter for a single storage node, because the retry are more quickly.
+
+This would be helpful if the requests could be processed by more than 1 candidate targets, for example
+follower or stale read requests that could be handled by multiple available peers.
+
+## Improvement Proposal
+
 A possible improvement suggested by [#44771](https://github.com/pingcap/tidb/issues/44771) is to make the
 timeout values of specific KV requests configurable. For example:
 - Adding a session variable `tidb_kv_read_timeout`, which is used to control the timeout for a single 
@@ -28,7 +38,7 @@ value of `ReadTimeoutShort` and `ReadTimeoutMedium`.
 - Adding statement level hint like `SELECT /*+ tidb_kv_read_timeout(500ms) */ * FROM t where id = ?;` to
 set the timeout value of the KV requests of this single query to the certain value.
 
-## Example Usage
+### Example Usage
 
 Consider the stale read usage case:
 
@@ -41,19 +51,12 @@ select * from t where id = 1;
 select /*+ tidb_kv_read_timeout(500ms) */ * FROM t where id = 1;
 ```
 
-## Benefits
-
-For latency sensitive applications, providing predictable sub-second read latency by set fast retrying 
-is valuable. queries that are usually very fast (such as point-select query), setting the value 
-of `tidb_kv_read_timeout` to short value like  `500ms`, the TiDB cluster would be more tolerable 
-for network latency or io latency jitter for a single storage node, because the retry are more quickly.
-
-## Problems
+### Problems
 
 - Setting the variable `tidb_kv_read_timeout ` may not be easy if it affects the timeout for all 
 TiKV read requests, such as Get, BatchGet, Cop in this session.A timeout of 1 second may be sufficient for GET requests, 
 but may be small for COP requests. Some large COP requests may keep timing out and could not be processed properly.
-- If the value of the variable `tidb_kv_read_timeout` is set too small, more retries will occur, 
+- If the value of the variable or query hint `tidb_kv_read_timeout` is set too small, more retries will occur, 
 increasing the load pressure on the TiDB cluster. In the worst case the query would not return until the 
 max backoff timeout is reached if the `tidb_kv_read_timeout` is set to a value which none of the replicas
 could finish processing within that time.
@@ -72,10 +75,10 @@ type StmtHints struct {
 	KVReadTimeout: Duration
 }
 ```
-- Support `tidb_kv_read_timeout` processing in `ExtractTableHintsFromStmtNode` and `handleStmtHints`, convert
-the user input hint value into `StmtContext` so it could be used later.
+- Support `tidb_kv_read_timeout` processing in `ExtractTableHintsFromStmtNode` and `handleStmtHints` functions, 
+convert the user input hint value into correspond field of `StmtContext` so it could be used later.
 
-## Support Timeout Configuration For Get And Batch-Get
+### Support Timeout Configuration For Get And Batch-Get
 
 - Add related filed in the `KVSnapshot` struct like
 
@@ -88,7 +91,7 @@ type KVSnapshot struct {
 
 This change is to be done in the `client-go` repository.
 
-- Add set interface `SetKVReadTimeout` and get interface `GetKVReadTimeout`
+- Add set interface `SetKVReadTimeout` and get interface `GetKVReadTimeout` for `KVSnapshot`
 ```go
 func (s *KVSnapshot) SetKVReadTimeout(KVTimeout Duration) {
 	s.KVReadTimeout = KVTimeout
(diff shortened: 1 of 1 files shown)


docs/design/2023-06-30-configurable-kv-timeout.md
```diff
@@ -0,0 +1,230 @@
...
+```
+- Change the `kvproto` to pass the read timeout value ~ the [`BatchGetRequest`](https://github.com/pingcap/kvproto/blob/master/proto/kvrpcpb.proto#L414)
+```protobuf
+message BatchGetRequest {
+    Context context = 1;
+    repeated bytes keys = 2;
+    uint64 version = 3;
+    uint32 read_timeout = 4; // Add this new field.
+}
+```
```
crazycs520, line 119: Currently `kvproto` already has a field `max_execution_duration_ms` in  `Context` https://github.com/pingcap/kvproto/blob/master/proto/kvrpcpb.proto#L820, so no need to add new timeout field?

docs/design/2023-06-30-configurable-kv-timeout.md
```diff
@@ -0,0 +1,230 @@
...
+        self.max_handle_duration, // Here use the specified timeout value.
+        peer,
+        Some(is_desc_scan),
+        start_ts.into(),
+        cache_match_version,
+        self.perf_level,
+    );
+}
+```
+This change needs to be done in the `tikv` repository.
```
crazycs520, line 197: ditto?

docs/design/2023-06-30-configurable-kv-timeout.md
```diff
@@ -0,0 +1,230 @@
...
+set the timeout value of the KV requests of this single query to the certain value.
+
+## Example Usage
+
+Consider the stale read usage case:
+
+```SQL
+set @@tidb_read_staleness=-5;
+# The unit is miliseconds. The session variable usage.
+set @@tidb_tikv_tidb_timeout=500;
```
crazycs520, line 38:
```suggestion
set @@ tidb_kv_read_timeout=500;
```

docs/design/2023-06-30-configurable-kv-timeout.md
cfzjywxk: Yes we could re-use the field in the `Context` which is used by write requests now, thanks for the reminding.

docs/design/2023-06-30-configurable-kv-timeout.md
```diff
@@ -0,0 +1,238 @@
...
+
+The query hint usage would be more flexible and safer as the impact is limited to a single query.
+
+### Add Hint Support For `tidb_kv_read_timeout`
+
+- Add related field in the `StatementContext` struct like
+```go
+type StmtHints struct {
+	...
+	kv_read_timeout: Duration
```
crazycs520, line 72:
```suggestion
	KVReadTimeout: Duration
```

docs/design/2023-06-30-configurable-kv-timeout.md
```diff
@@ -0,0 +1,238 @@
...
+is valuable. queries that are usually very fast (such as point-select query), setting the value 
+of `tidb_kv_read_timeout` to short value like  `500ms`, the TiDB cluster would be more tolerable 
+for network latency or io latency jitter for a single storage node, because the retry are more quickly.
+
+## Problems
+
+- Setting the variable `tidb_kv_read_timeout ` may not be easy if it affects the timeout for all 
+TiKV read requests, such as Get, BatchGet, Cop in this session.A timeout of 1 second may be sufficient for GET requests, 
+but may be small for COP requests. Some large COP requests may keep timing out and could not be processed properly.
+- If the value of the variable `tidb_kv_read_timeout` is set too small, more retries will occur,
```
ekexium, line 56: We can solve this by letting this timeout setting take effect only once, because it is less likely to have multiple nodes affected by jitter at the same time. Unlimited retry can also lead to congestion, increasing latency of all requests and make it worse.

docs/design/2023-06-30-configurable-kv-timeout.md
```diff
@@ -0,0 +1,238 @@
...
+
+A possible improvement suggested by [#44771](https://github.com/pingcap/tidb/issues/44771) is to make the
+timeout values of specific KV requests configurable. For example:
+- Adding a session variable `tidb_kv_read_timeout`, which is used to control the timeout for a single 
+TiKV read RPC request. When the user sets the value of this variable, all read RPC request timeouts will use this value. 
+The default value of this variable is 0, and the timeout of TiKV read RPC requests is still the original 
+value of `ReadTimeoutShort` and `ReadTimeoutMedium`.
+- Adding statement level hint like `SELECT /*+ tidb_kv_read_timeout(500ms) */ * FROM t where id = ?;` to
+set the timeout value of the KV requests of this single query to the certain value.
+
```
ekexium, line 30: We can briefly describe the benefit of configurable timeout at the end of the Motivation section.

docs/design/2023-06-30-configurable-kv-timeout.md
```diff
@@ -0,0 +1,238 @@
...
+* Authors: [cfzjywxk](https://github.com/cfzjywxk)
+* Tracking issue: TODO
+
+## Motivation
+
+The associated TiKV requests are typically processed quickly, taking only a few milliseconds. However, if 
+there is disk IO or network latency jitter on a specified TiKV node, the duration of these requests may spike 
+to several seconds or more. This usually happens when there is an EBS issue like the EBS is repairing its 
+replica.
```
ekexium, line 11: Shall we make it clear that it should only affect requests that have more than 1 candidate target?

docs/design/2023-06-30-configurable-kv-timeout.md
```diff
@@ -0,0 +1,238 @@
...
+  ...
+  uint64 max_execution_duration_ms = 14;
+  ...
+}
+```
+
+- Support timeout check during the request handling in TiKV. When there's new point get and batch point get 
+requests are created, the `kv_read_timeout` value should be read from the `GetReuqest` and `BatchGetRequest`
+and passed to the `pointGetter`. But by now there's no canceling mechanism when the task is scheduled to the
+read pool in TiKV, a simpler way is to check the deadline duration before next processing and try to return
```
ekexium, line 114: Shall we distinguish different DeadlineExceeded Errors? If it's caused by `kv_read_timeout` it's supposed to retry. What happens if it is caused by other deadlines setting?

docs/design/2023-06-30-configurable-kv-timeout.md
```diff
@@ -0,0 +1,238 @@
...
+- The timeout retry details should be recorded and handled properly in the request tracker, so are the metrics.
+
+## Timeout Task Canceling
+
+- If the timeout value is configured to a small value like hundreds of milliseconds, there could be more
+timeout tasks. When task is spawned to the read pool in TiKV, it could not be canceled immediately for 
+serveral reasons:
+  - When the read task is polled and executed, there's no timeout checking mechanism in the task scheduler
+  by now.
+  - The read operations are synchronous, so if the read task is blocked by slow io the task could not be
```
ekexium, line 214: So how about letting client-go simply sends the second request when timeout exceeded in the client side? It can cover more cases like network delay

docs/design/2023-06-30-configurable-kv-timeout.md
```diff
@@ -0,0 +1,238 @@
...
+- By now some simple timeout checks could be added in certain places like "after getting snapshot". A complete
+  timeout check or scheduling design is needed in the future.
+
+
+## Compatibility
+
+The requests with configurable timeout values would take effect on newer versions. These fields are expected not to take effect when down-grading the
+cluster theoretically.
+
+## More comprehensive solution
```
ekexium, line 225: When we support cancelling a request from outside commands, we can implement a real hedge policies. For example
- For requests taking more than a threshold of time, send a hedge request to another replica. Either response can be taken as the final result.
- When either response is returned, cancel the other request.
- TiKV prioritizes non-hedge requests so the hedge policy won't increase latencies.

docs/design/2023-06-30-configurable-kv-timeout.md
crazycs520: There don't seem to be any other deadlines setting in TiDB.

crazycs520 (approved):
REST LGTM

docs/design/2023-06-30-configurable-kv-timeout.md
ekexium: TiKV coprocessor requests set their deadlines from the config item `end_point_request_max_handle_duration`.

docs/design/2023-06-30-configurable-kv-timeout.md
cfzjywxk: Good idea. Or the max retry times could be set to the same value as the `max-replica`  configuration which is the available replicas that can handle stale or follower read requests.

docs/design/2023-06-30-configurable-kv-timeout.md
cfzjywxk: It would be more clear to distinguish them as the `kv_read_timeout` is supposed to only affect read requests.

docs/design/2023-06-30-configurable-kv-timeout.md
cfzjywxk: This could be realized if the real `in-time canceling` is supported as described above.

zhangjinpeng1987 (commented):
Please organize the RFC as:
1) Motivation/Background
2) The usage of proposed solution, detailed design
3) Other alternatives we have considered, list their pros and cons

docs/design/2023-06-30-configurable-kv-timeout.md
cfzjywxk: Added at the end of this section.

docs/design/2023-06-30-configurable-kv-timeout.md
cfzjywxk: Added in the `Motivation & Background` section.

docs/design/2023-06-30-configurable-kv-timeout.md
cfzjywxk: Described in the `Timeout Retry` part, current strategy is to let the timeout retry take effect at most `available-replicas - 1` times.

docs/design/2023-06-30-configurable-kv-timeout.md
```diff
@@ -0,0 +1,253 @@
+# Proposal: Configurable KV Timeout
+
+* Authors: [cfzjywxk](https://github.com/cfzjywxk)
+* Tracking issue: TODO
```
ekexium, line 4: We can file a tracking issue now

ekexium (approved):

PR #45130, opened 2023-07-03
util/chunk: optimize `(*ListInDisk).GetChunk` and add a fast row container reader

### LINKED ISSUE #45125: Implement a fast row container reader to dump rows from disk
## Enhancement

The current implementation of reading rows from `RowContainer` is not fast enough. The following function is used to dump a chunk:

```
// GetRowAndAppendToChunk gets a Row from the ListInDisk by RowPtr. Return the Row and the Ref Chunk.
func (l *ListInDisk) GetRowAndAppendToChunk(ptr RowPtr, chk *Chunk) (row Row, _ *Chunk, err error) {
	off, err := l.getOffset(ptr.ChkIdx, ptr.RowIdx)
	if err != nil {
		return
	}
	r := l.dataFile.getSectionReader(off)
	format := rowInDisk{numCol: len(l.fieldTypes)}
	_, err = format.ReadFrom(r)
	if err != nil {
		return row, nil, err
	}
	row, chk = format.toRow(l.fieldTypes, chk)
	return row, chk, err
}
```

It would be better to pipeline the `format.ReadFrom` (which is blocked by disk) and the `format.toRow` (which is blocked by CPU). 

Also, with the assumption that a whole chunk is stored together tightly, we can get the first row offset and use it for the whole chunk.

YangKeao:

Issue Number: close #45125

Problem Summary:

The existing reading method of `RowContainer` (`GetChunk(...)`) is not fast enough for dumping a lot of rows from disk (for the `cursorFetch` use case). 

The existing `Iterator4RowContainer` is even slower, as it allocates a new chunk for each row :facepalm:.

This PR is extracted from https://github.com/pingcap/tidb/pull/44730 (with a some refractor).

This PR pipelines the IO and CPU calculation, to make full use of the IO bandwidth. It should also help other features using `rowContainer`, as `GetChunk` is now much faster.

The performance of existing benchmark `BenchmarkListInDisk_GetChunk` increases from `2877471ns/op` to `462864ns/op`

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

### DIFF
--- a/util/chunk/row_container_reader.go
+++ b/util/chunk/row_container_reader.go
@@ -1,0 +1,166 @@
+// Copyright 2023 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package chunk
+
+import (
+	"context"
+	"runtime"
+	"sync"
+
+	"github.com/pingcap/tidb/util/logutil"
+)
+
+// RowContainerReader is a forward-only iterator for the row container. It provides an interface similar to other
+// iterators, but it doesn't provide `ReachEnd` function and requires manually closing to release goroutine.
+//
+// It's recommended to use the following pattern to use it:
+//
+//	for iter := NewRowContainerReader(rc); iter.Current() != iter.End(); iter.Next() {
+//		...
+//	}
+//	iter.Close()
+//	if iter.Error() != nil {
+//	}
+type RowContainerReader interface {
+	// Next returns the next Row.
+	Next() Row
+
+	// Current returns the current Row.
+	Current() Row
+
+	// End returns the invalid end Row.
+	End() Row
+
+	// Error returns none-nil error if anything wrong happens during the iteration.
+	Error() error
+
+	// Close closes the dumper
+	Close()
+}
+
+var _ RowContainerReader = &rowContainerReader{}
+
+// rowContainerReader is a forward-only iterator for the row container
+// It will spawn two goroutines for reading chunks from disk, and converting the chunk to rows. The row will only be sent
+// to `rowCh` inside only after when the full chunk has been read, to avoid concurrently read/write to the chunk.
+//
+// TODO: record the memory allocated for the channel and chunks.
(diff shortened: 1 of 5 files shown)


YangKeao:
As tested, spawning a new goroutine to read is faster for long row, but slower for short rows. 

```
4096                     1024                     512                       64
8562687 ns/op   2573457 ns/op   1363930 ns/op   370070 ns/op    (serial)
6035161 ns/op   2086350 ns/op   1249451 ns/op   439846 ns/op    (parallel)
```

util/chunk/row_container_test.go
```diff
@@ -303,3 +307,243 @@ func TestActionBlocked(t *testing.T) {
...
+			length := rand2.Uint32()
+			randomBytes := make([]byte, length%4096)
+			_, err := rand.Read(randomBytes)
+			require.NoError(t, err)
+
+			chk.AppendBytes(0, randomBytes)
+			allRows = append(allRows, randomBytes)
+		}
+		rc.Add(chk)
+	}
```
wshwsh12, line 340: extract a function to generate data?

util/chunk/row_container_test.go
```diff
@@ -303,3 +307,243 @@ func TestActionBlocked(t *testing.T) {
...
+	for i := 0; i < 16; i++ {
+		for j := 0; j < 16; j++ {
+			row := reader.Current()
+			require.Equal(t, allRows[i*16+j], row.GetBytes(0))
+			reader.Next()
+		}
+	}
+}
+
+func TestCloseRowContainerReader(t *testing.T) {
```
wshwsh12, line 353: What the test means?   seems the same as the last one?

util/chunk/row_container_reader.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// Error implements RowContainerReader
+func (reader *rowContainerReader) Error() error {
+	return reader.err
+}
+
+func (reader *rowContainerReader) initializeChannel() {
+	reader.rc.m.RLock()
+	defer reader.rc.m.RUnlock()
```
wshwsh12, line 102: Why need add rc.m.Rlock()？ rc.NumChunks() and rc.NumRowsOfChunk() have acquire the lock...

util/chunk/row_container_test.go
YangKeao: Good.

Done.

util/chunk/row_container_test.go
YangKeao: It only reads 8.5 chunks, but doesn't drain out the whole row container. It tests whether the reader can be closed successfully (and no goroutine leaks) when it doesn't reach the end.

util/chunk/row_container_reader.go
YangKeao: You are right. It's not needed now. They should be removed.

wshwsh12 (approved):

xhebox (approved):

PR #45140, opened 2023-07-04
executor: support partition table KeepOrder in UnionScan

### LINKED ISSUE #45007: UnionScan with partition table + keepOrder can't get correct result
## Bug Report

```
create table t(a int, b int, primary key(b)) PARTITION BY HASH(b) partitions 4;
analyze table t;
begin;
insert into t(a, b) values (6,6),(3,3),(9,9),(4,4),(5,5),(7,7),(8,8);
select * from t where b > 1 order by b limit 5;
```

```
mysql> select * from t where b > 1 order by b limit 5;
+------+---+
| a    | b |
+------+---+
|    3 | 3 |
|    4 | 4 |
|    5 | 5 |
|    6 | 6 |
|    7 | 7 |
+------+---+
5 rows in set (0.00 sec)
```

```
mysql> select * from t where b > 1 order by b limit 5;
+------+---+
| a    | b |
+------+---+
|    4 | 4 |
|    8 | 8 |
|    5 | 5 |
|    9 | 9 |
|    6 | 6 |
+------+---+
5 rows in set (0.00 sec)
```

Defined2014:

Issue Number: close #45007

Problem Summary: Let UnionScan supports keepOrder with Partition Table in dynamic pruning. The IndexMergeExec still not support keepOrder for both partition table and normal table, we could do it later.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Support partition table KeepOrder in UnionScan
```


tiancaiamao (approved):
Sort the []types.Datum requires loading all the rows in memory, break the iterator interface.
This may be a temporary solution to fix the bug quickly, but not the best way archive good performance.

tiancaiamao:
In addition, I notice two things during the code review.
1. keep order for partition table requires merge sort inside the TableReader executor. 
The distsql request could guarantee the keep order property, but **the order is defined by the order of `key-range`**.
While **the required order is actually `(partition id, b)`**
... so we need to merge sort the results from each key range

```
CREATE TABLE `t` (
  `a` int(11) DEFAULT NULL,
  `b` int(11) NOT NULL,
  PRIMARY KEY (`b`)  CLUSTERED
) PARTITION BY HASH (`b`) PARTITIONS 4;

mysql> explain select * from t where b > 1 order by b limit 5;
+-----------------------------+---------+-----------+---------------+-----------------------------------------------+
| id                          | estRows | task      | access object | operator info                                 |
+-----------------------------+---------+-----------+---------------+-----------------------------------------------+
| Limit_11                    | 5.00    | root      |               | offset:0, count:5                             |
| └─TableReader_18            | 5.00    | root      | partition:all | data:Limit_17                                 |
|   └─Limit_17                | 5.00    | cop[tikv] |               | offset:0, count:5                             |
|     └─TableRangeScan_16     | 5.00    | cop[tikv] | table:t       | range:(1,+inf], keep order:true, stats:pseudo |
+-----------------------------+---------+-----------+---------------+-----------------------------------------------+
4 rows in set (0.00 sec)
```

The distsql query result from each key range is sorted by `b`, the clustered primary key.
To get the correct `keep order` result, merge sort is used internally.

https://github.com/pingcap/tidb/blob/fdf8c2cd0d63585e8ce445c3f79bc42d7421013f/distsql/select_result.go#L126

Note, **union scan do not have such utilities**

2. The rows from membuffer does `keep order`, it means https://github.com/pingcap/tidb/issues/45007 bug may not limted to partitioned table!

winoros (approved):

PR #45156, opened 2023-07-04
session: Add mysql.user columns used by Navicat | tidb-test=pr/2249

### LINKED ISSUE #45154: Navicat user management compatibility
## Bug Report

Try to manage users in TiDB with Navicat

With Navicat 16.2.2 click the "User" button.

User management

image

```
Release Version: v7.2.0
Edition: Community
Git Commit Hash: 9fd5f4a8e4f273a60fbe7d3848f85a1be8f0600b
Git Branch: heads/refs/tags/v7.2.0
UTC Build Time: 2023-06-27 15:04:42
GoVersion: go1.20.5
Race Enabled: false
Store: tikv
```

dveeden:

Issue Number: close #45154 

Problem Summary:

Add columns that Navicat expects

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Compatibility with Navicat User Management was improved
```


dveeden:
The vim modeline is to set the tabwidth to 4.

With `set ts=8`:
image

With `set ts=4`:
image

Looks like GitHub has 8 as default and this can cause the layout to break. Looks like VS Code has 4 as default.

bb7133:
Should we update the data of the column?

dveeden:
> Should we update the data of the column?

The parser at least supports some of the resource options (`MAX_USER_CONNECTIONS`, etc). I can see if we can actually update these columns on `CREATE USER` and `ALTER USER`.  If that works adding this to `SHOW CREATE USER` etc. should be easy as well.

However then the next step would be to actually enforce these. Would the limits be in server scope or in cluster scope (etcd?)

I haven't looked at the `ssl_type` etc.

dveeden:
The `ssl_type` is stored in a different table. I think that Navicat is creating and altering user with `CREATE USER` and `ALTER USER` and only queries the `mysql.user` table to get a list of users as there is no good interface for this. So, updating this column when we update `mysql.global_priv` should be ok.
```
mysql> CREATE USER 'u1'@'%';

mysql> CREATE USER 'u2'@'%' REQUIRE SSL;

mysql> CREATE USER 'u3'@'%' REQUIRE X509;

mysql> TABLE mysql.global_priv;
+------+------+----------------+
| Host | User | Priv           |
+------+------+----------------+
| %    | u1   | {}             |
| %    | u2   | {"ssl_type":1} |
| %    | u3   | {"ssl_type":2} |
+------+------+----------------+
3 rows in set (0.00 sec)
```

dveeden:
This now:

1. Adds SSL/X509 and Resource Limits columns to `mysql.user`
2. When upgrading it fills the SSL/X509 fields based on `mysql.global_priv`
3. `SHOW CREATE USER` now shows resource limits
4. `CREATE USER` now updates the SSL/X509 and Resource Limits columns

Note that:
- The SSL/X509 columns in `mysql.user` are not used as `mysql.global_priv` is used instead. This is only for third-party tool compatibility.
- The Resource limits are not used at all
- `ALTER USER` etc. hasn't been updated.

dveeden:
> Should we update the data of the column?

What do you think about the latest updates? Now the columns get populated when running `CREATE USER`..

executor/show.go
```diff
@@ -1677,6 +1678,28 @@ func (e *ShowExec) fetchShowCreateUser(ctx context.Context) error {
 		tokenIssuer = " token_issuer " + tokenIssuer
 	}
 
+	maxQuestions := rows[0].GetUint64(10)
+	maxUpdates := rows[0].GetUint64(11)
+	maxConnections := rows[0].GetUint64(12)
+	maxUserConnections := rows[0].GetUint64(13)
+	var resourceStr string
+	if maxQuestions > 0 || maxUpdates > 0 || maxConnections > 0 || maxUserConnections > 0 {
```
CbcWestwolf, line 1686: We need to add some test cases to cover this code snippet.

executor/simple.go
```diff
@@ -1180,7 +1181,36 @@ func (e *SimpleExec) executeCreateUser(ctx context.Context, s *ast.CreateUserStm
 		}
 
 		hostName := strings.ToLower(spec.User.Hostname)
-		sqlexec.MustFormatSQL(sql, valueTemplate, hostName, spec.User.Username, pwd, authPlugin, userAttributesStr, plOptions.lockAccount, recordTokenIssuer, plOptions.passwordExpired, plOptions.passwordLifetime)
+
+		var resQryHour, resUpdHour, resConHour, resUserCon int64 = 0, 0, 0, 0
+		for _, resOpt := range s.ResourceOptions {
```
CbcWestwolf, line 1186: If currently these options are implemented only in the parser level, we should give warnings to users to mention it.

CbcWestwolf (approved):
Almost LGTM from my side. Please fix the CI failure :-)

CbcWestwolf:
> This now:
> 
> 1. Adds SSL/X509 and Resource Limits columns to `mysql.user`
> 2. When upgrading it fills the SSL/X509 fields based on `mysql.global_priv`
> 3. `SHOW CREATE USER` now shows resource limits
> 4. `CREATE USER` now updates the SSL/X509 and Resource Limits columns
> 
> Note that:
> 
> * The SSL/X509 columns in `mysql.user` are not used as `mysql.global_priv` is used instead. This is only for third-party tool compatibility.
> * The Resource limits are not used at all
> * `ALTER USER` etc. hasn't been updated.

After these changes are merged, please modify the relative description in pingcap/docs

executor/simple.go
dveeden: We already do this since a3182eb84347a264dfa6872642ab5d7cac950fdf:
```
sql> CREATE USER 'foo'@'bar.example.com' IDENTIFIED BY 'foobar' WITH MAX_CONNECTIONS_PER_HOUR 1;
Query OK, 0 rows affected, 1 warning (0.0162 sec)
Warning (code 1064): You have an error in your SQL syntax; check the manual that corresponds to your TiDB version for the right syntax to use line 1 column 90 near ""TiDB does not support WITH ConnectionOptions now, they would be parsed but ignored. 

sql> 
```

https://github.com/pingcap/tidb/blob/67e8f902f32fa793f45808bd7468fb6fe4bc5249/parser/parser.y#L13320

executor/show.go
dveeden: Done in 8b8ba8b648f24fabbd4c0e2fcf9d6e5bdde97a4b

session/bootstrap.go
```diff
@@ -2789,6 +2802,36 @@ func upgradeToVer172(s Session, ver int64) {
 	mustExecute(s, CreateDoneRunawayWatchTable)
 }
 
+func upgradeToVer173(s Session, ver int64) {
+	if ver >= version173 {
+		return
+	}
+	doReentrantDDL(s, `ALTER TABLE mysql.user
+	ADD COLUMN ssl_type enum('','ANY','X509','SPECIFIED') NOT NULL DEFAULT '',
```
TonsnakeLin, line 2810: The column order of mysq.user in TiDB is different from MySQL,
Will we keep it consistent?

dveeden: Tools should avoid using any `mysql.*` tables directly. Instead they should prefer using things like `CREATE USER`, `ALTER USER`, `GRANT`, `REVOKE`, etc.

When that's not possible (e.g. to list users) it might be needed to directly query `mysql.user`. When this is done the query should specify the columns that the tool needs. Using `SELECT * FROM ...` should not be used by applications unless the application is willing to handle different order of columns and more or less columns than expected etc.

We could try to keep this the same, but there are already a lot of differences. If we do this we should do this outside of this PR.

```
sql> SELECT VERSION(); SHOW COLUMNS FROM mysql.user;
+-----------+
| VERSION() |
+-----------+
| 8.1.0     |
+-----------+
1 row in set (0.0007 sec)
+--------------------------+-----------------------------------+------+-----+-----------------------+-------+
| Field                    | Type                              | Null | Key | Default               | Extra |
+--------------------------+-----------------------------------+------+-----+-----------------------+-------+
| Host                     | char(255)                         | NO   | PRI |                       |       |
| User                     | char(32)                          | NO   | PRI |                       |       |
| Select_priv              | enum('N','Y')                     | NO   |     | N                     |       |
| Insert_priv              | enum('N','Y')                     | NO   |     | N                     |       |
| Update_priv              | enum('N','Y')                     | NO   |     | N                     |       |
| Delete_priv              | enum('N','Y')                     | NO   |     | N                     |       |
| Create_priv              | enum('N','Y')                     | NO   |     | N                     |       |
| Drop_priv                | enum('N','Y')                     | NO   |     | N                     |       |
...
```

```
sql> SELECT VERSION(); SHOW COLUMNS FROM mysql.user;
+------------------------------------------+
| VERSION()                                |
+------------------------------------------+
| 5.7.25-TiDB-v7.3.0-alpha-309-g8b8ba8b648 |
+------------------------------------------+
1 row in set (0.0011 sec)
+------------------------+-----------------------------------+------+-----+-------------------+-------+
| Field                  | Type                              | Null | Key | Default           | Extra |
+------------------------+-----------------------------------+------+-----+-------------------+-------+
| Host                   | char(255)                         | NO   | PRI | NULL              |       |
| User                   | char(32)                          | NO   | PRI | NULL              |       |
| authentication_string  | text                              | YES  |     | NULL              |       |
| plugin                 | char(64)                          | YES  |     | NULL              |       |
| Select_priv            | enum('N','Y')                     | NO   |     | N                 |       |
| Insert_priv            | enum('N','Y')                     | NO   |     | N                 |       |
| Update_priv            | enum('N','Y')                     | NO   |     | N                 |       |
| Delete_priv            | enum('N','Y')                     | NO   |     | N                 |       |
| Create_priv            | enum('N','Y')                     | NO   |     | N                 |       |
| Drop_priv              | enum('N','Y')                     | NO   |     | N                 |       |
| Process_priv           | enum('N','Y')                     | NO   |     | N                 |       |
...
```

Note that TiDB also has at least one extra column: `Token_issuer`

session/bootstrap.go
```diff
@@ -2789,6 +2802,36 @@ func upgradeToVer172(s Session, ver int64) {
...
+
+	// Update mysql.user table with data from mysql.global_priv
+	mustExecute(s, `UPDATE mysql.user u JOIN mysql.global_priv p ON u.Host=p.Host AND u.User=p.User
+	SET u.ssl_type='ANY' WHERE p.Priv->'$.ssl_type'=1`)
+	mustExecute(s, `UPDATE mysql.user u JOIN mysql.global_priv p ON u.Host=p.Host AND u.User=p.User
+	SET u.ssl_type='X509' WHERE p.Priv->'$.ssl_type'=2`)
+	mustExecute(s, `UPDATE mysql.user u JOIN mysql.global_priv p ON u.Host=p.Host AND u.User=p.User
+	SET u.ssl_type='SPECIFIED' WHERE p.Priv->'$.ssl_type'=3`)
+
+	mustExecute(s, `UPDATE mysql.user u JOIN mysql.global_priv p ON u.Host=p.Host AND u.User=p.User
```
TonsnakeLin, line 2827: Do we have any test cases for this ?

TonsnakeLin:
The rest is LGTM.

lastincisor (approved):

dveeden:
PTAL @TonsnakeLin

PR #45230, opened 2023-07-07
planner: simplify StatsCacheWrapper

qw4990:

Issue Number: ref #44461

Problem Summary: planner: simplify `StatsCacheWrapper`

planner: simplify `StatsCacheWrapper`

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/statistics/handle/cache/internal/inner.go
+++ b/statistics/handle/cache/internal/inner.go
@@ -21,10 +21,12 @@ import (
 // StatsCacheInner is the interface to manage the statsCache, it can be implemented by map, lru cache or other structures.
 type StatsCacheInner interface {
 	// GetByQuery retrieves cache triggered by a query. Usually used in LRU.
+	// TODO: merge this method with Get.
 	GetByQuery(int64) (*statistics.Table, bool)
 	// Get gets the cache.
 	Get(int64) (*statistics.Table, bool)
 	// PutByQuery puts a cache triggered by a query. Usually used in LRU.
+	// TODO: merge this method with Put.
 	PutByQuery(int64, *statistics.Table)
 	// Put puts a cache.
 	Put(int64, *statistics.Table)
@@ -32,11 +34,8 @@ type StatsCacheInner interface {
 	Del(int64)
 	// Cost returns the memory usage of the cache.
 	Cost() int64
-	// Keys returns the keys of the cache.
-	Keys() []int64
 	// Values returns the values of the cache.
 	Values() []*statistics.Table
-	Map() map[int64]*statistics.Table
 	// Len returns the length of the cache.
 	Len() int
 	FreshMemUsage()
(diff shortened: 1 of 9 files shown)


statistics/handle/cache/internal/inner.go
```diff
@@ -21,29 +21,33 @@ import (
...
 	Put(int64, *statistics.Table)
 	// Del deletes a cache.
 	Del(int64)
 	// Cost returns the memory usage of the cache.
 	Cost() int64
-	// Keys returns the keys of the cache.
-	Keys() []int64
 	// Values returns the values of the cache.
 	Values() []*statistics.Table
-	Map() map[int64]*statistics.Table
```
qw4990, line 39: Remove 2 unnecessary methods to make this interface easier to maintain and understand.

hawkingrei (approved):

xuyifangreeneyes (approved):

statistics/handle/handle.go
```diff
@@ -618,7 +617,7 @@ func (h *Handle) Update(is infoschema.InfoSchema, opts ...cache.TableStatsOpt) e
 		tbl.TblInfoUpdateTS = tableInfo.UpdateTS
 		tables = append(tables, tbl)
 	}
-	h.updateStatsCache(oldCache.Update(tables, deletedTableIDs, lastVersion, opts...))
+	h.updateStatsCache(oldCache.Update(tables, deletedTableIDs, opts...))
```
chrysan, line 620: What about adding logic here? If maxTableVersion loaded here is not larger than oldCache.Version, we do not do update.

qw4990: I've tried this but found some bad cases. For example, in the beginning, we have 2 tables, and their stats versions are `t1:100` and `t2:200` respectively, then the cache version should be `200`.
And then we drop the table `t2` and then try to update the cache, since we'll find we cannot update the cache, since cache version is `100` which is less than `200`.

chrysan: If we do not update cache version when Delete, why does cache version become 100 from 200?

statistics/handle/cache/statscache.go
```diff
@@ -67,11 +67,9 @@ func (s *StatsCache) UpdateCache(newCache StatsCacheWrapper) (updated bool, newC
 	s.mu.Lock()
 	oldCache := s.cache.Load()
 	newCost = newCache.Cost()
-	if oldCache.version < newCache.version || (oldCache.version == newCache.version && oldCache.minorVersion < newCache.minorVersion) {
-		s.memTracker.Consume(newCost - oldCache.Cost())
-		s.cache.Store(&newCache)
-		updated = true
-	}
+	s.memTracker.Consume(newCost - oldCache.Cost())
```
xuyifangreeneyes, line 70: Maybe we can still keep the comparison of version since `(s *StatsCache).UpdateCache` can be called concurrently.

qw4990: 1) even if we keep the comparison, there are still some bad cases that can cause unexpected overwriting; 2) the side-effect of overwriting here seems trivial and acceptable; So to make it easier to understand and maintain, maybe removing the version comparison here is better.

PR #45232, opened 2023-07-07
server: add more integration/unit tests for cursor fetch

YangKeao:
Add more tests to cover the following situations:

1. Execute returns error in the early check
2. Execute / Fetch can return the expected data
3. Execute multiple times will not return an error, and the original cursor will be closed automatically
4. Reset will close the opened cursor automatically
5. Command send long data works well
6. Reset will removes the stored "long data"

### DIFF
--- a/server/conn_stmt_test.go
+++ b/server/conn_stmt_test.go
@@ -15,6 +15,7 @@
 package server
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"crypto/rand"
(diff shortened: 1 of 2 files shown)


hawkingrei (approved):

server/conn_stmt_test.go
```diff
@@ -352,3 +357,269 @@ func TestCursorFetchErrorInFetch(t *testing.T) {
...
+	require.Equal(t, expected, out.Bytes())
+}
+
+func TestCursorFetchSendLongDataReset(t *testing.T) {
+	store, dom := testkit.CreateMockStoreAndDomain(t)
+	srv := CreateMockServer(t, store)
+	srv.SetDomain(dom)
+	defer srv.Close()
+
+	appendUint32 := binary.LittleEndian.AppendUint32
```
Defined2014, line 581: How about change `appendUint16`, `appendUint32`, `appendUint64` funcs to global variables?

PR #45357, opened 2023-07-14
statcache: faster LFU

hawkingrei:

Issue Number: ref #45367

Problem Summary:

```
BenchmarkCachePut
BenchmarkCachePut/LRU
BenchmarkCachePut/LRU-8         	   35224	     65941 ns/op
BenchmarkCachePut/LFU
BenchmarkCachePut/LFU-8         	  143685	     10771 ns/op

BenchmarkCachePutGet/LRU
BenchmarkCachePutGet/LRU-8         	   30358	     40205 ns/op
BenchmarkCachePutGet/LFU
BenchmarkCachePutGet/LFU-8         	   85431	     14172 ns/op
```
### Check List

Tests 

- [x] Unit test
- [ ] Integration test

add benchmark

- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/statistics/handle/cache/internal/lfu/key_set.go
+++ b/statistics/handle/cache/internal/lfu/key_set.go
@@ -1,0 +1,52 @@
+// Copyright 2023 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package lfu
+
+import (
+	"sync"
+
+	"golang.org/x/exp/maps"
+)
+
+type keySet struct {
+	set map[int64]struct{}
+	mu  sync.RWMutex
+}
+
+func (ks *keySet) Add(key int64) {
+	ks.mu.Lock()
+	ks.set[key] = struct{}{}
+	ks.mu.Unlock()
+}
+
+func (ks *keySet) Remove(key int64) {
+	ks.mu.Lock()
+	delete(ks.set, key)
+	ks.mu.Unlock()
+}
+
+func (ks *keySet) Keys() []int64 {
+	ks.mu.RLock()
+	result := maps.Keys(ks.set)
+	ks.mu.RUnlock()
+	return result
+}
+
+func (ks *keySet) Len() int {
+	ks.mu.RLock()
+	result := len(ks.set)
+	ks.mu.RUnlock()
+	return result
+}
--- a/statistics/handle/cache/internal/lfu/lfu_cache.go
+++ b/statistics/handle/cache/internal/lfu/lfu_cache.go
@@ -1,0 +1,132 @@
+// Copyright 2023 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package lfu
+
+import (
+	"sync/atomic"
+
+	"github.com/dgraph-io/ristretto"
+	"github.com/pingcap/tidb/statistics"
+	"github.com/pingcap/tidb/statistics/handle/cache/internal"
+	"github.com/pingcap/tidb/util/intest"
+)
+
+// LFU is a LFU based on the ristretto.Cache
+type LFU struct {
+	cache        *ristretto.Cache
+	resultKeySet *keySetShard
+	cost         atomic.Int64
+}
+
+// NewLFU creates a new LFU cache.
+func NewLFU(totalMemCost int64) (*LFU, error) {
+	result := &LFU{}
+	bufferItems := int64(64)
+	if intest.InTest {
+		bufferItems = 1
+	}
+	cache, err := ristretto.NewCache(&ristretto.Config{
+		NumCounters:        1024 * 1024 * 1024, // TODO(hawkingrei): make it configurable for NumCounters and MaxCost
+		MaxCost:            totalMemCost,
+		BufferItems:        bufferItems,
+		OnEvict:            result.onEvict,
+		OnExit:             result.onExit,
+		IgnoreInternalCost: intest.InTest,
+		Metrics:            intest.InTest,
+	})
+	result.cache = cache
+	result.resultKeySet = newKeySetShard()
+	return result, err
+}
+
+// Get implements statsCacheInner
+func (s *LFU) Get(tid int64, _ bool) (*statistics.Table, bool) {
+	result, ok := s.cache.Get(tid)
+	if !ok {
+		return nil, ok
+	}
+	return result.(*statistics.Table), ok
+}
+
+// Put implements statsCacheInner
+func (s *LFU) Put(tblID int64, tbl *statistics.Table, _ bool) {
+	ok := s.cache.Set(tblID, tbl, tbl.MemoryUsage().TotalTrackingMemUsage())
+	if ok { // NOTE: `s.cache` and `s.resultKeySet` may be inconsistent since the update operation is not atomic, but it's acceptable for our scenario
+		s.resultKeySet.Add(tblID)
+		s.cost.Add(tbl.MemoryUsage().TotalTrackingMemUsage())
+	}
+}
+
+// Del implements statsCacheInner
+func (s *LFU) Del(tblID int64) {
+	s.cache.Del(tblID)
+	s.resultKeySet.Remove(tblID)
+}
+
+// Cost implements statsCacheInner
+func (s *LFU) Cost() int64 {
+	return s.cost.Load()
+}
+
+// Values implements statsCacheInner
+func (s *LFU) Values() []*statistics.Table {
+	result := make([]*statistics.Table, 0, 512)
+	for _, k := range s.resultKeySet.Keys() {
+		if value, ok := s.cache.Get(k); ok {
+			result = append(result, value.(*statistics.Table))
+		}
+	}
+	return nil
+}
+
+func (s *LFU) onEvict(item *ristretto.Item) {
+	// We do not need to calculate the cost during onEvict, because the onexit function
+	// is also called when the evict event occurs.
+	s.resultKeySet.Remove(int64(item.Key))
+}
+
+func (s *LFU) onExit(val interface{}) {
+	s.cost.Add(-1 * val.(*statistics.Table).MemoryUsage().TotalTrackingMemUsage())
+}
+
+// Len implements statsCacheInner
+func (s *LFU) Len() int {
+	return s.resultKeySet.Len()
+}
+
+// Front implements statsCacheInner
+func (*LFU) Front() int64 {
+	return 0
+}
+
+// Copy implements statsCacheInner
+func (s *LFU) Copy() internal.StatsCacheInner {
+	return s
+}
+
+// SetCapacity implements statsCacheInner
+func (s *LFU) SetCapacity(maxCost int64) {
+	s.cache.UpdateMaxCost(maxCost)
+}
+
+// wait blocks until all buffered writes have been applied. This ensures a call to Set()
+// will be visible to future calls to Get(). it is only used for test.
+func (s *LFU) wait() {
+	s.cache.Wait()
+}
+
+func (s *LFU) metrics() *ristretto.Metrics {
+	return s.cache.Metrics
+}
--- a/statistics/handle/cache/internal/lfu/lfu_cache.go
+++ b/statistics/handle/cache/internal/lfu/lfu_cache.go
@@ -88,7 +88,7 @@ func (s *LFU) Values() []*statistics.Table {
 			result = append(result, value.(*statistics.Table))
 		}
 	}
-	return nil
+	return result
 }
 
 func (s *LFU) onEvict(item *ristretto.Item) {
--- a/statistics/handle/cache/internal/lfu/lfu_cache_test.go
+++ b/statistics/handle/cache/internal/lfu/lfu_cache_test.go
@@ -109,18 +109,17 @@ func TestCacheLen(t *testing.T) {
 
 func TestLFUCachePutGetWithManyConcurrency(t *testing.T) {
 	// to test DATA RACE
-	capacity := int64(12)
+	capacity := int64(100000000000)
 	lfu, err := NewLFU(capacity)
 	require.NoError(t, err)
 	var wg sync.WaitGroup
+	wg.Add(2000)
 	for i := 0; i < 1000; i++ {
-		wg.Add(1)
 		go func(i int) {
 			defer wg.Done()
 			t1 := testutil.NewMockStatisticsTable(1, 1, true, false, false)
 			lfu.Put(int64(i), t1, true)
 		}(i)
-		wg.Add(1)
 		go func(i int) {
 			defer wg.Done()
 			lfu.Get(int64(i), true)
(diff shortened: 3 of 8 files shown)


statistics/handle/cache/internal/mapcache/map_cache.go
```diff
@@ -49,12 +52,16 @@ func NewMapCache() *MapCache {
...
 func (m *MapCache) Get(k int64, _ bool) (*statistics.Table, bool) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
 	v, ok := m.tables[k]
 	return v.value, ok
 }
 
 // Put implements StatsCacheInner
 func (m *MapCache) Put(k int64, v *statistics.Table, _ bool) {
+	m.mu.Lock()
```
hawkingrei, line 63: If not this mutex, it cannot benchmark. I think it will be removed next PR.

bb7133:
> Tests 
>
> - [ ] Unit test
> - [ ] Integration test
> - [ ] Manual test (add detailed scripts or steps below)
> - [ ] No code

Please do remember to update the description.

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return nil
+}
+
+func (s *LFU) onEvict(item *ristretto.Item) {
+	idx := shard(int64(item.Key))
+	s.resultKeySet[idx].Remove(int64(item.Key))
+}
+
+func (s *LFU) onExit(val interface{}) {
```
qw4990, line 106: When this function will be called?

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -0,0 +1,132 @@
...
+const keySetCnt = 128
+
+func shard(key int64) int {
+	return int(key) % keySetCnt
+}
+
+// LFU is a LFU based on the ristretto.Cache
+type LFU struct {
+	cache        *ristretto.Cache
+	resultKeySet [keySetCnt]keySet
```
qw4990, line 35: Can we implement the shard key set as a new structure like: 
```
type shardKeySet structure {
    []sets keySet
    numShards int
}
```

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -0,0 +1,132 @@
...
+	return s.cost.Load()
+}
+
+// Values implements statsCacheInner
+func (s *LFU) Values() []*statistics.Table {
+	result := make([]*statistics.Table, 0, len(s.resultKeySet))
+	for key := range s.resultKeySet {
+		if value, ok := s.cache.Get(key); ok {
+			result = append(result, value.(*statistics.Table))
+		}
```
qw4990, line 96: This implementation seems not correct and not thread-safe.

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// Front implements statsCacheInner
+func (*LFU) Front() int64 {
+	return 0
+}
+
+// Copy implements statsCacheInner
+func (s *LFU) Copy() internal.StatsCacheInner {
+	return s
```
qw4990, line 126: Why not implement this method?

statistics/handle/cache/internal/lfu/lfu_cache_test.go
```diff
@@ -0,0 +1,106 @@
...
+	lru.Put(int64(2), t2, false)
+	time.Sleep(5 * time.Second)
+	require.Equal(t, lru.Len(), 1)
+
+	// put t3, t1/t2 should be evicted all items and disappeared from the list
+	t3 := testutil.NewMockStatisticsTable(2, 1, true, false, false)
+	lru.Put(int64(3), t3, false)
+	time.Sleep(1 * time.Second)
+	require.Equal(t, lru.Len(), 1)
+}
```
qw4990, line 106: Better to add some concurrent test cases, e.g. start some goroutines to read/write this cache and see whether there is any DATA RACE and whether the status is still consistent after some operations.

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: When ristretto's cache deletes or replaces the cache item, it will be called.

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: LFU can not be copied. it can not copy its cm sketch. Copy interface will be deleted later。

statistics/handle/cache/internal/lfu/lfu_cache_test.go
hawkingrei: Done

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: The previous code was inaccurate, so I rewrote it.

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: Do you want to implement dynamically configurable sharding numbers? This will be a bit tricky.

qw4990: Now, just encapsulate a new structure for this to make it easier to use.

statistics/handle/cache/internal/lfu/lfu_cache.go
qw4990: Can we implement a simple version of `copy` above the `LFU`? e.g. read all keys and values from the old cache and add them into a new cache.

statistics/handle/cache/internal/lfu/lfu_cache_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+func TestLFUCachePutGet(t *testing.T) {
```
qw4990, line 109: It seems this is not enough. We'd better:
```
for i := 0; i < 5; i++ {
    go func() { // reader
        for i := 0; i < 1000; i++ {
            // randomly read some keys
        }
    }
}
for i := 0; i < 5; i++ {
    go func() { // writer
        for i := 0; i < 1000; i++ {
            // randomly write some keys
        }
    }
}
wg.Done() 
// check whether the status is consistent, e.g. cost and `keySet` are consistent to the underlying cache.
```

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: It can be implemented. But this copy is to COW. so if we do not use the COW, it is useless.

statistics/handle/cache/internal/lfu/lfu_cache_test.go
hawkingrei: Done

statistics/handle/cache/internal/lfu/key_set.go
```diff
@@ -0,0 +1,59 @@
...
+	delete(ks.set, key)
+	ks.mu.Unlock()
+}
+
+func (ks *keySet) Contains(key int64) bool {
+	ks.mu.RLock()
+	_, ok := ks.set[key]
+	ks.mu.RUnlock()
+	return ok
+}
```
qw4990, line 45: No place uses this method, can we remove it? Better to keep our implementation as simple as possible.

hawkingrei: Removed.

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -0,0 +1,115 @@
...
+type LFU struct {
+	cache        *ristretto.Cache
+	resultKeySet *keySetShard
+	cost         atomic.Int64
+}
+
+// NewLFU creates a new LFU cache.
+func NewLFU(cost int64) (*LFU, error) {
+	result := &LFU{}
+	cache, err := ristretto.NewCache(&ristretto.Config{
```
qw4990, line 36: Should we close this cache when exit?

hawkingrei: It seems that there won't be too many issues if it doesn't close when exiting, and the handle also does not consider the cleaning issue when exiting.

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -0,0 +1,119 @@
...
+	if !ok {
+		return nil, ok
+	}
+	return result.(*statistics.Table), ok
+}
+
+// Put implements statsCacheInner
+func (s *LFU) Put(tblID int64, tbl *statistics.Table, _ bool) {
+	ok := s.cache.Set(tblID, tbl, tbl.MemoryUsage().TotalTrackingMemUsage())
+	if ok {
```
qw4990, line 65:
```suggestion
	if ok {  // NOTE: `s.cache` and `s.resultKeySet` may be inconsistent since the update operation is not atomic, but it's acceptable for our scenario
```

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -0,0 +1,119 @@
...
+	for _, k := range s.resultKeySet.Keys() {
+		if value, ok := s.cache.Get(k); ok {
+			result = append(result, value.(*statistics.Table))
+		}
+	}
+	return nil
+}
+
+func (s *LFU) onEvict(item *ristretto.Item) {
+	s.resultKeySet.Remove(int64(item.Key))
```
qw4990, line 94: Should we update the cost here?

statistics/handle/cache/internal/lfu/lfu_cache_test.go
```diff
@@ -0,0 +1,161 @@
...
+	lru, err := NewLFU(1000)
+	require.NoError(t, err)
+	t1 := testutil.NewMockStatisticsTable(1, 1, true, false, false)
+	t2 := testutil.NewMockStatisticsTable(2, 2, true, false, false)
+	t3 := testutil.NewMockStatisticsTable(3, 3, true, false, false)
+	lru.Put(int64(1), t1, false)
+	lru.Put(int64(2), t2, false)
+	lru.Put(int64(3), t3, false)
+	time.Sleep(1 * time.Second)
+	require.Equal(t, lru.Cost(), 6*mockCMSMemoryUsage+6*mockCMSMemoryUsage)
```
qw4990, line 61: Why not use `t1.MemoryUsage().TotalMemUsage + t2.MemoryUsage().TotalMemUsage + t3.MemoryUsage().TotalMemUsage`?

statistics/handle/cache/internal/lfu/lfu_cache_test.go
```diff
@@ -0,0 +1,161 @@
...
+	t4 := testutil.NewMockStatisticsTable(2, 1, true, false, false)
+	lru.Put(int64(1), t4, false)
+	time.Sleep(4 * time.Second)
```
qw4990, line 64: It seems like these `sleep` can be removed.

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: Done

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: Done

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -0,0 +1,119 @@
...
+
+// LFU is a LFU based on the ristretto.Cache
+type LFU struct {
+	cache        *ristretto.Cache
+	resultKeySet *keySetShard
+	cost         atomic.Int64
+}
+
+// NewLFU creates a new LFU cache.
+func NewLFU(cost int64) (*LFU, error) {
```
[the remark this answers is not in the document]
hawkingrei: Done

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// SetCapacity implements statsCacheInner
+func (s *LFU) SetCapacity(maxCost int64) {
+	s.cache.UpdateMaxCost(maxCost)
+}
+
+// Wait blocks until all buffered writes have been applied. This ensures a call to Set()
+// will be visible to future calls to Get(). it is only used for test.
+func (s *LFU) Wait() {
```
qw4990, line 126: We can let this method unexposed since it is only used in test.

statistics/handle/cache/internal/lfu/lfu_cache_test.go
```diff
@@ -0,0 +1,165 @@
...
+	require.Equal(t, uint64(lfu.Cost()), lfu.metrics().CostAdded()-lfu.metrics().CostEvicted())
+}
+
+func TestLFUCachePutGetWithManyConcurrency2(t *testing.T) {
+	// to test DATA RACE
+	capacity := int64(12)
+	lfu, err := NewLFU(capacity)
+	require.NoError(t, err)
+	var wg sync.WaitGroup
+	var cnt atomic.Int64
```
qw4990, line 140: This variable seems useless... This variable is used to test for what?

qw4990 (approved):

statistics/handle/cache/internal/lfu/lfu_cache_test.go
hawkingrei: removed.

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: Done

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// Cost implements statsCacheInner
+func (s *LFU) Cost() int64 {
+	return s.cost.Load()
+}
+
+// Values implements statsCacheInner
+func (s *LFU) Values() []*statistics.Table {
+	result := make([]*statistics.Table, 0, 512)
```
chrysan, line 85: Why 512?

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -0,0 +1,132 @@
...
+		if value, ok := s.cache.Get(k); ok {
+			result = append(result, value.(*statistics.Table))
+		}
+	}
+	return result
+}
+
+func (s *LFU) onEvict(item *ristretto.Item) {
+	// We do not need to calculate the cost during onEvict, because the onexit function
+	// is also called when the evict event occurs.
```
chrysan, line 96: Why does it design 2 interfaces, onEvict and onExit?

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -0,0 +1,132 @@
...
+		Metrics:            intest.InTest,
+	})
+	result.cache = cache
+	result.resultKeySet = newKeySetShard()
+	return result, err
+}
+
+// Get implements statsCacheInner
+func (s *LFU) Get(tid int64, _ bool) (*statistics.Table, bool) {
+	result, ok := s.cache.Get(tid)
```
chrysan, line 56: So the Get from internal logic will also add one to the frequency, right?

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return result.(*statistics.Table), ok
+}
+
+// Put implements statsCacheInner
+func (s *LFU) Put(tblID int64, tbl *statistics.Table, _ bool) {
+	ok := s.cache.Set(tblID, tbl, tbl.MemoryUsage().TotalTrackingMemUsage())
+	if ok { // NOTE: `s.cache` and `s.resultKeySet` may be inconsistent since the update operation is not atomic, but it's acceptable for our scenario
+		s.resultKeySet.Add(tblID)
+		s.cost.Add(tbl.MemoryUsage().TotalTrackingMemUsage())
```
chrysan, line 68: Will the cost be accumulated repeatedly?

chrysan (commented):
And we need a benchmark like, with 500,000 tables, and the stats update to cache happens frequently, how's the Get QPS?

hawkingrei:
@chrysan It is a sad story. I think I wrongly use the LGTM3.

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: It is not 2 public interface. ```onEvist``` is a function called when ristretto evicts s cache item. ```onExit``` is a function called when ristretto deletes a cache item.

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: it has a buffer to deal with the counter.

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: No, because if the operator is ```update```, although this will add the new object's memory, OnExit will be called in the background and delete the old object's memory.

statistics/handle/cache/internal/lfu/lfu_cache.go
chrysan: Looks like the cache itself does not provide the parameter to ignore frequency counting. Tha't fine.

statistics/handle/cache/internal/lfu/lfu_cache.go
chrysan: Get. onExit will be called at both eviction and manual deletion.

chrysan (commented):
Please handle put failure when buffer is full. Make sure that table put failure will not impact the update to table stats later.

PR #45394, opened 2023-07-17
lightning: tidb backend will check conflict threshold

lance6716:

Issue Number: ref https://github.com/pingcap/tidb/issues/41629

Problem Summary:

use affected rows to check if batch insert meets duplications

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/lightning/backend/tidb/tidb.go
+++ b/br/pkg/lightning/backend/tidb/tidb.go
@@ -276,12 +276,29 @@ var _ backend.Backend = (*tidbBackend)(nil)
 //
 // The backend does not take ownership of `db`. Caller should close `db`
 // manually after the backend expired.
-func NewTiDBBackend(ctx context.Context, db *sql.DB, onDuplicate string, errorMgr *errormanager.ErrorManager) backend.Backend {
-	switch onDuplicate {
-	case config.ReplaceOnDup, config.IgnoreOnDup, config.ErrorOnDup:
-	default:
-		log.FromContext(ctx).Warn("unsupported action on duplicate, overwrite with `replace`")
+func NewTiDBBackend(
+	ctx context.Context,
+	db *sql.DB,
+	conflict config.Conflict,
+	errorMgr *errormanager.ErrorManager,
+) backend.Backend {
+	var onDuplicate string
+	switch conflict.Strategy {
+	case config.ErrorOnDup:
+		onDuplicate = config.ErrorOnDup
+	case config.ReplaceOnDup:
 		onDuplicate = config.ReplaceOnDup
+	case config.IgnoreOnDup:
+		if conflict.MaxRecordRows == 0 {
+			onDuplicate = config.IgnoreOnDup
+		} else {
+			// need to stop batch insert on error and fall back to row by row insert
+			// to record the row
+			onDuplicate = config.ErrorOnDup
+		}
+	default:
+		log.FromContext(ctx).Warn("unsupported conflict strategy, overwrite with `error`")
+		onDuplicate = config.ErrorOnDup
 	}
 	return &tidbBackend{
 		db:          db,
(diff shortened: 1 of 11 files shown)


dsdashun (approved):

br/pkg/lightning/backend/tidb/tidb.go
```diff
@@ -276,12 +276,29 @@ var _ backend.Backend = (*tidbBackend)(nil)
...
+		onDuplicate = config.ErrorOnDup
+	case config.ReplaceOnDup:
 		onDuplicate = config.ReplaceOnDup
+	case config.IgnoreOnDup:
+		if conflict.MaxRecordRows == 0 {
+			onDuplicate = config.IgnoreOnDup
+		} else {
+			// need to stop batch insert on error and fall back to row by row insert
+			// to record the row
+			onDuplicate = config.ErrorOnDup
```
D3Hunter, line 297:
```
mysql> insert into t values(1);
ERROR 1062 (23000): Duplicate entry '1' for key 't.PRIMARY'
mysql> insert ignore into t values(1);

mysql>
mysql> replace into t values(1);
```
`insert ingore into` does have different RowAffected as len(rows), but `replace into` not, shouldn't we use this checking on `replace`?

lance6716: > The REPLACE statement returns a count to indicate the number of rows affected. This is the sum of the rows deleted and inserted. If the count is 1 for a single-row REPLACE, a row was inserted and no rows were deleted. If the count is greater than 1, one or more old rows were deleted before the new row was inserted. It is possible for a single row to replace more than one old row if the table contains multiple unique indexes and the new row duplicates values for different old rows in different unique indexes.

The affected-rows count makes it easy to determine whether REPLACE only added a row or whether it also replaced any rows: Check whether the count is 1 (added) or greater (replaced).

https://dev.mysql.com/doc/refman/8.4/en/replace.html

seems the behavior is not the same as documents

D3Hunter: I tried a real replaced case, it works. and it seems we can also tell how many rows ignored using RowAffected
```
mysql> create table t(id int primary key, v int);
Query OK, 0 rows affected (0.15 sec)

mysql> insert into t values(1,1);

mysql> replace into t values(1,1);

mysql> replace into t values(2,2);

mysql> replace into t values(1,2);
Query OK, 2 rows affected (0.00 sec)

mysql> insert ignore into t values(3,3);

mysql> insert ignore into t values(1,1);
```

lance6716: I should have implemented it. When the user doesn't require conflict row details, lightning will use REPLACE and check affected rows to calculate if conflict threshold is exceeded

D3Hunter: yes, but for `ignore` we will set it to `error` if `conflict.threshold > 0`, it seems we can implement the same as `replace`

lance6716: we will use `IgnoreOnDup` when `conflict.MaxRecordRows == 0` (line 292). But when `conflict.MaxRecordRows` is not 0, we need to know which row is conflicting to record it, so we can't use batched INSERT IGNORE INTO.

D3Hunter: why not just record a count like `replace`, for compatibility?

lance6716: This is two features. One is set an early fail threshold for large amount of duplicate data. The other one is display the duplicate data.

lance6716: Displaying the duplicate data has more details than only limit the count, so it causes worse performance (fallback to row-by-row INSERT) and feature compatibility (can't use REPLACE)

D3Hunter: get, but i think we can make `ignore` works like `replace` to record count only later. I met a case that user uses `ignore` but it still print `ERROR` logs on duplicate and the SQL is `INSERT INTO` unlike the doc says as `IGNORE INTO`(it's fixed here https://github.com/pingcap/docs/pull/18911/files), it's confusing though i get the point after checking the code

lance6716: We can achieve it if it's 0 MaxRecordRows. Maybe change the default value to 0? Currently it's 10000

```golang
	if c.Threshold < 0 {
		switch c.Strategy {
		case ErrorOnDup, NoneOnDup:
			c.Threshold = 0
		case IgnoreOnDup, ReplaceOnDup:
			c.Threshold = DefaultRecordDuplicateThreshold
		}
	}
```

D3Hunter: lgtm

PR #45465, opened 2023-07-19
*: global runaway watch by system table and impl exector for `query watch`

CabinfeverB:

Issue Number: ref https://github.com/pingcap/tidb/issues/43691

Problem Summary:

1. global runaway watch by system table 
> create table `mysql.tidb_runaway_watch` and table `mysql.tidb_runaway_watch_done` to persist runaway watch and deletion of runaway watch. And all TiDb servers read both tables periodically to sync runaway watch among TiDBs.
2. impl exector for query watch

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
About runaway watch sync:
Execute query on TiDB1, and then execute query on TiDB2
image
We can see TiDB1 returns error 8253, and TiDB2 returns error 8254. And mysql.tidb_runaway_watch has the record.
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


CabinfeverB:
cc @nolouch @glorv @Connor1996

session/bootstrap.go
```diff
@@ -618,18 +618,34 @@ const (
...
 	) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;`
 
+	// CreateDoneRunawayWatchTable stores the condition which is used to check whether query should be quarantined.
+	CreateDoneRunawayWatchTable = `CREATE TABLE IF NOT EXISTS mysql.tidb_runaway_watch_done (
+		id BIGINT(20) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+		record_id BIGINT(20) not null,
+		resource_group_name varchar(32) not null,
+		start_time datetime(6) NOT NULL,
+		end_time datetime(6),
+		watch  bigint(10) NOT NULL,
```
Connor1996, line 642: extra blank

session/bootstrap.go
```diff
@@ -947,11 +963,12 @@ const (
 	version168 = 168
 	version169 = 169
 	version170 = 170
+	version171 = 171
```
Connor1996, line 966: no comment?

ddl/util/util.go
```diff
@@ -31,7 +31,7 @@ import (
 	"github.com/pingcap/tidb/util/logutil"
 	"github.com/pingcap/tidb/util/sqlexec"
 	"github.com/tikv/client-go/v2/tikvrpc"
-	"go.etcd.io/etcd/client/v3"
+	clientv3 "go.etcd.io/etcd/client/v3"
```
Connor1996, line 34: necessary?

domain/resourcegroup/runaway.go
```diff
@@ -227,30 +372,40 @@ func (r *RunawayChecker) BeforeExecutor() error {
...
 }
 
 // BeforeCopRequest checks runaway and modifies the request if necessary before sending coprocessor request.
 func (r *RunawayChecker) BeforeCopRequest(req *tikvrpc.Request) error {
+	if r.setting == nil {
+		return nil
+	}
 	marked := r.marked.Load()
 	if !marked {
+		// if err := r.BeforeExecutor(); err != nil {
```
Connor1996, line 406: clean it

domain/resourcegroup/runaway.go
```diff
@@ -121,38 +205,90 @@ func (rm *RunawayManager) DeriveChecker(resourceGroupName string, originalSQL st
...
+	key := record.GetRecordKey()
+	if force {
+		rm.queryLock.Lock()
+		if rm.watchList.Get(key) != nil {
+			rm.watchList.Delete(key)
+		}
+		rm.watchList.Set(key, record, ttl)
+		rm.queryLock.Unlock()
+	} else {
+		if rm.watchList.Get(key) == nil {
```
Connor1996, line 245: Should it be in lock?

domain/runaway.go
```diff
@@ -0,0 +1,525 @@
...
+			err := updateFn()
+			do.runawaySyncer.doneChan <- err
+		case <-runawayWatchSyncTicker.C:
+			updateFn()
+		}
+	}
+}
+
+func (do *Domain) AddRunawayWatch(record *resourcegroup.QuarantineRecord) (int64, error) {
+	if err := do.handleRunawayWatch(record); err != nil {
```
glorv, line 176: I think addwatch should return success after the record is inseted into the watch table. No need to wait for the background task refresh.  BTW, if a statement returns error, you should ensure there is no side effect in the system.

domain/runaway.go
```diff
@@ -0,0 +1,525 @@
...
+	se, err := do.sysSessionPool.Get()
+	defer func() {
+		do.sysSessionPool.Put(se)
+	}()
+	if err != nil {
+		return errors.Annotate(err, "get session failed")
+	}
+	exec, _ := se.(sqlexec.SQLExecutor)
+	ctx := kv.WithInternalSourceType(context.Background(), kv.InternalTxnStats)
+	_, err = exec.ExecuteInternal(ctx, "BEGIN")
```
glorv, line 307: why ignore checking this error

domain/runaway.go
```diff
@@ -0,0 +1,525 @@
...
+	defer func() {
+		if err != nil {
+			_, err1 := exec.ExecuteInternal(ctx, "ROLLBACK")
+			terror.Log(err1)
+			return
+		}
+		_, err = exec.ExecuteInternal(ctx, "COMMIT")
```
glorv, line 314: What if this commit failed? should not do a commit in defer

domain/runaway.go
```diff
@@ -0,0 +1,525 @@
...
+
+func (do *Domain) RemoveRunawayWatch(recordID int64) error {
+	records, err := do.runawaySyncer.getWatchRecordByID(recordID)
+	if err != nil {
+		return err
+	}
+	if len(records) != 1 {
+		return errors.Errorf("no runaway watch with the specific ID")
+	}
+	err = do.handleRunawayWatchDone(records[0])
```
glorv, line 214: The sql in `handleRunawayWatchDone` and `handleRemoveStaleRunawayWatch` should be executed in the same txn to ensure atomic

domain/resourcegroup/runaway.go
```diff
@@ -121,38 +205,90 @@ func (rm *RunawayManager) DeriveChecker(resourceGroupName string, originalSQL st
...
+	if item == nil {
+		return nil
+	}
+	return item.Value()
+}
+
+func (rm *RunawayManager) AddWatch(record *QuarantineRecord) {
+	ttl := time.Until(record.EndTime)
+	if ttl <= 0 {
+		rm.staleQuarantineRecord <- record
```
glorv, line 270: Why need this explicit stale. Outdated record should be auto cleaned up by the clean up loop

session/bootstrap.go
```diff
@@ -618,18 +618,34 @@ const (
...
-	// CreateRunawayQuarantineWatchTable stores the condition which is used to check whether query should be quarantined.
-	CreateRunawayQuarantineWatchTable = `CREATE TABLE IF NOT EXISTS mysql.tidb_runaway_quarantined_watch (
+	// CreateRunawayWatchTable stores the condition which is used to check whether query should be quarantined.
+	CreateRunawayWatchTable = `CREATE TABLE IF NOT EXISTS mysql.tidb_runaway_watch (
+		id BIGINT(20) NOT NULL AUTO_INCREMENT PRIMARY KEY,
 		resource_group_name varchar(32) not null,
-		start_time TIMESTAMP NOT NULL,
-		end_time TIMESTAMP NOT NULL,
-		watch varchar(12) NOT NULL,
+		start_time datetime(6) NOT NULL,
```
glorv, line 625: How to deal with the timezone difference if the cluster is cross region deployed?

ddl/util/util.go
CabinfeverB: It is changed by gofmt automatically

domain/resourcegroup/runaway.go
CabinfeverB: This is first check, because we generally believe that in most cases, we will not add a watch list to a key repeatedly

domain/resourcegroup/runaway.go
```diff
@@ -75,84 +78,259 @@ type RunawayRecord struct {
...
 	default:
 		// TODO: add warning for discard flush records
 	}
 }
+
+func (rm *RunawayManager) addWatchList(record *QuarantineRecord, ttl time.Duration, force bool) {
+	key := record.GetRecordKey()
+	item := rm.watchList.Get(key)
+	if force {
+		rm.queryLock.Lock()
```
Connor1996, line 269: so getting from watchList is lock-free and delete and set isn't?

domain/resourcegroup/runaway.go
```diff
@@ -75,84 +78,259 @@ type RunawayRecord struct {
...
 
 // RunawayManager is used to detect and record runaway queries.
 type RunawayManager struct {
-	queryLock          sync.Mutex
-	resourceGroupCtl   *rmclient.ResourceGroupsController
-	watchList          *ttlcache.Cache[string, struct{}]
-	serverID           string
-	runawayQueriesChan chan *RunawayRecord
-	quarantineChan     chan *QuarantineRecord
+	queryLock        sync.Mutex
```
Connor1996, line 179: add a comment for what's the lock used for

session/bootstrap.go
CabinfeverB: `types.Time` always saves UTC time, so I think no problem.

domain/resourcegroup/runaway.go
CabinfeverB: Yes.

domain/resourcegroup/runaway.go
CabinfeverB: I remove the the clean up loop. The reason for this is that the information_schema table was not involved before, and I wanted to gc it as quickly as possible

domain/runaway.go
CabinfeverB: fix

domain/runaway.go
CabinfeverB: In the design, we want to return the ID of the added watch record to the user.

domain/runaway.go
CabinfeverB: In the design, we want to return the ID of the added watch record to the user.

domain/resourcegroup/runaway.go
```diff
@@ -121,38 +205,90 @@ func (rm *RunawayManager) DeriveChecker(resourceGroupName string, originalSQL st
...
+	}
+	force := false
+	if record.Source == "manual" {
+		force = true
+	}
+	rm.addWatchList(record, ttl, force)
+}
+
+func (rm *RunawayManager) RemoveWatch(record *QuarantineRecord) {
+	rm.watchList.Delete(record.GetRecordKey())
```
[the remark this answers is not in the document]
CabinfeverB: fixed

domain/runaway.go
glorv: Then how do you handle the case if the record did insert successfully but the wait watch timeout, in your current impl it will return an error but because the record is inserted it will still take effect later and on other instances. Please ensure if you return an error to the user(which means the txn is not successful) then there should be no unexpected side-effect in the system.

CabinfeverB:
/test unit-test
--

session/bootstrap.go
glorv: I don't see how the `Time` type will help you handle the timezone offset

domain/runaway.go
```diff
@@ -0,0 +1,615 @@
...
+func (do *Domain) handleRunawayWatch(record *resourcegroup.QuarantineRecord) error {
+	se, err := do.sysSessionPool.Get()
+	defer func() {
+		do.sysSessionPool.Put(se)
+	}()
+	if err != nil {
+		return errors.Annotate(err, "get session failed")
+	}
+	exec, _ := se.(sqlexec.SQLExecutor)
+	ctx := kv.WithInternalSourceType(context.Background(), kv.InternalTxnStats)
```
nolouch, line 328: I think the source is not stats, maybe  `InternalTxnOthers`.

domain/runaway.go
```diff
@@ -0,0 +1,615 @@
...
+func (do *Domain) handleRunawayWatchDone(record *resourcegroup.QuarantineRecord) error {
+	se, err := do.sysSessionPool.Get()
+	defer func() {
+		do.sysSessionPool.Put(se)
+	}()
+	if err != nil {
+		return errors.Annotate(err, "get session failed")
+	}
+	exec, _ := se.(sqlexec.SQLExecutor)
+	ctx := kv.WithInternalSourceType(context.Background(), kv.InternalTxnStats)
```
nolouch, line 358: same

domain/runaway.go
```diff
@@ -0,0 +1,615 @@
...
+func (do *Domain) handleRemoveStaleRunawayWatch(record *resourcegroup.QuarantineRecord) error {
+	se, err := do.sysSessionPool.Get()
+	defer func() {
+		do.sysSessionPool.Put(se)
+	}()
+	if err != nil {
+		return errors.Annotate(err, "get session failed")
+	}
+	exec, _ := se.(sqlexec.SQLExecutor)
+	ctx := kv.WithInternalSourceType(context.Background(), kv.InternalTxnStats)
```
nolouch, line 393: same

domain/runaway.go
```diff
@@ -0,0 +1,615 @@
...
+func (do *Domain) execRestrictedSQL(sql string, params []interface{}) ([]chunk.Row, error) {
+	se, err := do.sysSessionPool.Get()
+	defer func() {
+		do.sysSessionPool.Put(se)
+	}()
+	if err != nil {
+		return nil, errors.Annotate(err, "get session failed")
+	}
+	exec := se.(sqlexec.RestrictedSQLExecutor)
+	ctx := kv.WithInternalSourceType(context.Background(), kv.InternalTxnStats)
```
nolouch, line 423: same

domain/runaway.go
```diff
@@ -0,0 +1,615 @@
...
+	builder.WriteString(" where ")
+	builder.WriteString(r.KeyCol)
+	builder.WriteString(" > %? order by ")
+	builder.WriteString(r.KeyCol)
+	params = append(params, r.CheckPoint)
+	return builder.String(), params
+}
+
+func (r *SystemTableReader) Read(exec sqlexec.RestrictedSQLExecutor, genFn func() (string, []interface{})) ([]chunk.Row, error) {
+	ctx := kv.WithInternalSourceType(context.Background(), kv.InternalTxnStats)
```
nolouch, line 619: Ditto

domain/runaway.go
CabinfeverB: It should be a very rare situation. I think there are no side effects, but we should return a clearer error message that the user added successfully, but the TiDB server load timed out, you can slightly query information_schame.runaway_watches to confirm.

domain/runaway.go
```diff
@@ -0,0 +1,625 @@
...
+					err := do.handleRemoveStaleRunawayWatch(r)
+					if err == nil {
+						break
+					}
+					logutil.BgLogger().Error("remove stale runaway watch", zap.Error(err))
+					time.Sleep(1000 * time.Millisecond)
+				}
+			}()
+		}
+	}
```
nolouch, line 316: time.Second

domain/runaway.go
CabinfeverB: fix

session/bootstrap.go
CabinfeverB: Now I convert all time to UTC and save, read and parse them as UTC

Connor1996 (approved):
LGTM

domain/resourcegroup/runaway.go
```diff
@@ -75,84 +79,291 @@ type RunawayRecord struct {
...
+		return
+	}
+
+	force := false
+	// The manual record replaces the old record.
+	force = record.Source == ManualSource || record.Source == rm.serverID
+	rm.addWatchList(record, ttl, force)
+}
+
+// RemoveWatch is used to remove watch items from system table.
```
glorv, line 341: misleading comment, this function only remove records from in-memory cache

domain/resourcegroup/runaway.go
```diff
@@ -75,84 +79,291 @@ type RunawayRecord struct {
...
+	item := rm.watchList.Get(key)
+	if item != nil {
+		return item.Value()
+	}
+	return nil
+}
+
 func (rm *RunawayManager) markRunaway(resourceGroupName, originalSQL, planDigest string, action string, matchType RunawayMatchType, now *time.Time) {
+	source := rm.serverID
+	if len(source) > 128 {
```
glorv, line 364: Why need truncate here?

domain/runaway.go
```diff
@@ -0,0 +1,625 @@
...
+	for {
+		select {
+		case <-timer.C:
+			return 0, errors.Errorf("the query watch is added successfully, but the TiDB server load timed out, you can query information_schame.runaway_watches later to confirm")
+		case <-ticker.C:
+			r := do.runawayManager.GetWatchByKey(record.GetRecordKey())
+			if r.ID > 0 {
+				return r.ID, nil
+			}
+		case err := <-do.runawaySyncer.doneChan:
```
glorv, line 198: What if multiple add watches happen at the same time, how can you ensure you get the right message from this chan?

domain/runaway.go
glorv: Don't get what you want to explain, but it's still clear that the current logic cannot ensure the procedure to be atomic, so it can deliver wrong result

executor/internal/querywatch/query_watch.go
```diff
@@ -16,13 +16,143 @@ package querywatch
...
+	}
+	rg, err := client.GetResourceGroup(record.ResourceGroupName)
+	if err != nil {
+		return err
+	}
+	if rg == nil {
+		return infoschema.ErrResourceGroupNotExists.GenWithStackByArgs(record.ResourceGroupName)
+	}
+	if record.Action == rmpb.RunawayAction_NoneAction {
+		if rg.RunawaySettings == nil {
```
glorv, line 146: What happens if when the RunawaySettings is not nil when adding the record and then changed to none, then what is the action of the record

session/bootstrap.go
glorv: So what is the benefit for this change, I didn't see any other table to use datetime to represent time

domain/resourcegroup/runaway.go
CabinfeverB: I think the server IP in k8s will be domain name, so the length of it will be very long. And the length of feild is 128, do we need to change it to blob?

domain/resourcegroup/runaway.go
CabinfeverB: Well, it does have ambiguity

domain/resourcegroup/runaway.go
```diff
@@ -75,84 +79,291 @@ type RunawayRecord struct {
...
+	if record.EndTime.Equal(NullTime) {
+		ttl = 0
+	} else if ttl <= 0 {
+		rm.staleQuarantineRecord <- record
+		return
+	}
+
+	force := false
+	// The manual record replaces the old record.
+	force = record.Source == ManualSource || record.Source == rm.serverID
```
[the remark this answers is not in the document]
CabinfeverB: manually added records has higher priority. I will change it.

executor/internal/querywatch/query_watch.go
CabinfeverB: This watch will be uselss but not be removed from watch list. You can see `BeforeExecutor`, If no match action, it will do nothing.

session/bootstrap.go
CabinfeverB: Because timestamp only store time between 1970-01-01 00:00:01 UTC and 2038-01-19 03:14:07 UTC. You can find datetime in stats_history.

domain/runaway.go
CabinfeverB: Well, I know what you mean. The main point of disagreement is whether the function needs to keep atomic. I didn't think we needed it, so I didn't implement it strictly.

domain/resourcegroup/runaway.go
glorv: You should set a field length that is long enough to store the data

domain/resourcegroup/runaway.go
```diff
@@ -183,9 +402,18 @@ func (rm *RunawayManager) QuarantineRecordChan() <-chan *QuarantineRecord {
...
+
 // examineWatchList check whether the query is in watch list.
-func (rm *RunawayManager) examineWatchList(resourceGroupName string, convict string) bool {
-	return rm.watchList.Get(resourceGroupName+"/"+convict) != nil
+func (rm *RunawayManager) examineWatchList(resourceGroupName string, convict string) (bool, rmpb.RunawayAction) {
+	item := rm.getWatchFromWatchList(resourceGroupName + "/" + convict)
+	if item == nil {
+		return false, 0
+	}
+	return true, item.Action
```
glorv, line 416: For a manual add item without action, the action is none, in this case you should use the setting's action instead. BTW, if we support let manual record reflect the change of query limit's action change, I would expect watch record reflect this change too.

domain/resourcegroup/runaway.go
CabinfeverB: OK, let's set 512 same as analyze job

domain/resourcegroup/runaway.go
CabinfeverB: You can see L461-463. 
make sense, I will update it.

CabinfeverB: done

qw4990 (approved):

session/bootstrap.go
```diff
@@ -2773,10 +2777,16 @@ func upgradeToVer171(s Session, ver int64) {
 	if ver >= version171 {
 		return
 	}
+	mustExecute(s, "ALTER TABLE mysql.tidb_runaway_queries CHANGE COLUMN `tidb_server` `tidb_server` varchar(512)")
```
CabinfeverB, line 2780: Due to TestUpgradeVersionForResumeJob, I try to pass it by add two upgrade functions. Maybe it's not a good method, I just want to pass test first.

domain/runaway.go
```diff
@@ -0,0 +1,626 @@
...
+var systemSchemaCIStr = model.NewCIStr("mysql")
+
+func (do *Domain) deleteExpiredRows(tableName, colName string, expiredDuration time.Duration) {
+	if !do.DDL().OwnerManager().IsOwner() {
+		return
+	}
+	failpoint.Inject("FastRunawayGC", func() {
+		expiredDuration = time.Second * 1
+	})
+	expiredTime := time.Now().Add(-expiredDuration)
```
glorv, line 65: Does the timezone offset also impact this?

domain/runaway.go
CabinfeverB: Can't ensure. If keys of records are different and it gets wrong message , It can't get watch and will try wo get watch later. If keys are same, refer to https://github.com/pingcap/tidb/pull/45465/files#r1277100060

domain/runaway.go
CabinfeverB: Now gc clean up loop only affects `tidb_runaway_queries`, it does not affect correctness.

domain/runaway.go
glorv: As you only send a message to notifyChan only once, one routine receives another routine's return message and drops it, then the other routine is blocked, this is a more severe situation.

glorv (approved):

Issue #45367, opened 2023-07-14
Provide a Higher Performance LFU Cache Implementation for Statscache


hawkingrei:
## Enhancement

Currently, our statscache uses a single lock to protect thread safety, but due to the nature of locks, performance is poor. Therefore, we would like to provide a higher performance LRU cache implementation for statscache.
Our solution is to use sharding map and LFU to implement LRU cache. This way, we can avoid using a single global lock and instead shard the cache and use individual locks for each shard. This can improve concurrency performance and cache efficiency. At the same time, the LFU algorithm can better adapt to the data access pattern and improve cache hit rate.
We plan to implement this new LRU cache implementation in the next release and would appreciate your feedback and suggestions.

PR #45484, opened 2023-07-20
cache: add statscache metrics and enable LFU

hawkingrei:

Issue Number: ref #45367

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -130,3 +148,9 @@ func (s *LFU) wait() {
 func (s *LFU) metrics() *ristretto.Metrics {
 	return s.cache.Metrics
 }
+
+// Stop implements statsCacheInner
+func (s *LFU) Stop() {
```
hawkingrei, line 153: Add ```stop``` to avoid leak.

statistics/handle/cache/statscacheinner.go
```diff
@@ -105,10 +121,29 @@ func (sc *StatsCache) PutFromInternal(id int64, t *statistics.Table) {
...
+	return ok
+}
+
 // Put puts the table statistics to the cache.
 func (sc *StatsCache) put(id int64, t *statistics.Table, moveLRUFront bool) {
-	sc.c.Put(id, t, moveLRUFront)
-
+	ok := sc.putCache(id, t, moveLRUFront)
+	if !ok {
+		logutil.BgLogger().Warn("fail to put the stats cache")
```
time-and-fate, line 144: I suggest we make the log content more detailed here, like adding the table name, stack trace, or anything useful.

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -32,6 +34,14 @@ type LFU struct {
 
 // NewLFU creates a new LFU cache.
 func NewLFU(totalMemCost int64) (*LFU, error) {
+	if totalMemCost == 0 {
+		memTotal, err := memory.MemTotal()
+		if err != nil {
+			return nil, err
+		}
+		totalMemCost = int64(memTotal / 2)
```
time-and-fate, line 42: Looks like this is a behavior change, is there any rationale behind this change?

statistics/handle/cache/statscacheinner.go
```diff
@@ -129,9 +164,15 @@ func (sc *StatsCache) Cost() int64 {
 
 // SetCapacity sets the memory capacity of the cache.
 func (sc *StatsCache) SetCapacity(c int64) {
+	metrics.CostGauge.Set(float64(c))
```
time-and-fate, line 167: We updated this both here and in `(*LFU).SetCapacity()`. I think this is unintended.

statistics/handle/handle.go
```diff
@@ -1012,7 +1022,6 @@ func (h *Handle) statsCacheLen() int {
 // Callers should add retry loop if necessary.
 func (h *Handle) updateStatsCache(newCache *cache.StatsCache) (updated bool) {
 	h.statsCache.Replace(newCache)
-	handle_metrics.CostGauge.Set(float64(newCache.Cost()))
```
time-and-fate, line 1015: Does this mean now we no longer update the metrics when the stats cache updates?

statistics/handle/cache/internal/lfu/lfu_cache.go
```diff
@@ -130,3 +148,9 @@ func (s *LFU) wait() {
 func (s *LFU) metrics() *ristretto.Metrics {
 	return s.cache.Metrics
 }
+
+// Stop implements statsCacheInner
+func (s *LFU) Stop() {
+	s.cache.Close()
+	s.cache.Wait()
```
time-and-fate, line 155: Would this possibly wait for a long time?

statistics/handle/handle.go
hawkingrei: we do not use to update the stats cache to replace the old with the old one. we just set the cache by mutex.

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: No, the goroutine in the cache will receive the close signal and then exit.

statistics/handle/cache/internal/inner.go
```diff
@@ -26,7 +26,7 @@ type StatsCacheInner interface {
 	// Get gets the cache.
 	Get(tid int64, moveFront bool) (*statistics.Table, bool)
 	// Put puts a cache.
-	Put(tid int64, tbl *statistics.Table, moveLRUFront bool)
+	Put(tid int64, tbl *statistics.Table, moveLRUFront bool) bool
```
qw4990, line 29: In our scenarios, it seems that we don't need to expose the `bool`.

statistics/handle/cache/internal/lfu/lfu_cache.go
qw4990, line 153: How about renaming it from `Stop()` to `Close()` to keep aligned with the underlying `s.cache.Close()`?

statistics/handle/cache/internal/lfu/lfu_cache.go
qw4990: Will it cause some dead-lock or dead-goroutine if we call `cache.Close()` before `cache.Wait()`?

domain/domain.go
```diff
@@ -1022,6 +1022,9 @@ func (do *Domain) Close() {
 	if intest.InTest {
 		infosync.MockGlobalServerInfoManagerEntry.Close()
 	}
+	if handle := do.statsHandle.Load(); handle != nil {
```
qw4990, line 1025: Is this change necessary? If it is not necessary for this PR, I prefer to keep it unchanged to avoid any risk 🤦

statistics/handle/cache/internal/lru/lru_cache.go
```diff
@@ -1,443 +0,0 @@
-// Copyright 2023 PingCAP, Inc.
```
qw4990, line 1: Better to keep it unchanged, we can remove it in the next few PRs after finishing the new LFU completely.

statistics/handle/cache/statscacheinner.go
hawkingrei: If we read the code from the coprocessor cache, we will know the capacity is from the cost and the user can config the max memory usage of the coprocessor. this LFU is the same approach.

hawkingrei: the function can be changed as ```SetCost```.

statistics/handle/cache/statscacheinner.go
```diff
@@ -129,9 +164,15 @@ func (sc *StatsCache) Cost() int64 {
...
 	sc.c.SetCapacity(c)
 }
 
+// Stop stops the cache.
+func (sc *StatsCache) Stop() {
```
qw4990, line 172: Better to use `Close` to aligned with the underlying cache's method name.
And in addition, `Close` seems a better name in Golang than `Stop`, many standard libs use `Close` instead of `Stop` like `io.Closer`.

statistics/handle/cache/statscacheinner.go
```diff
@@ -57,17 +59,21 @@ func WithTableStatsByQuery() TableStatsOpt {
...
 
 // NewStatsCache creates a new StatsCacheWrapper.
-func NewStatsCache() *StatsCache {
+func NewStatsCache() (*StatsCache, error) {
 	enableQuota := config.GetGlobalConfig().Performance.EnableStatsCacheMemQuota
 	if enableQuota {
 		capacity := variable.StatsCacheMemQuota.Load()
-		return &StatsCache{
-			c: lru.NewStatsLruCache(capacity),
+		stats, err := lfu.NewLFU(capacity)
```
qw4990, line 66: In method `CopyAndUpdate`, the `StatsCache` copies the underlying `cache`, will that cause any risk since the LFU cache doesn't implement `copy` method?

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: Done

domain/domain.go
hawkingrei: It is to avoid the ```goleak```. so that we do not add ```ristretto``` into goleak config in the every ```main_test```.

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: Let's start by looking at the implementation of ```wait``` and ```close```. 

- ```Close``` is at the https://github.com/dgraph-io/ristretto/blob/main/cache.go#L354.

Close is to close some channels to send a close signal. so that the background task will be stopped like evict policy. Finally, it will be marked as closed.

- ```Wait``` is at the https://github.com/dgraph-io/ristretto/blob/main/cache.go#L213

 It seems that "wait" doesn't have any effect because once you close it, it returns immediately.

hawkingrei: it has been marked as closed. All the ```get```/```set``` will return nothing immediately.

statistics/handle/cache/internal/inner.go
hawkingrei: Actually, in the ```StatsCache```, the retry logic is already included, so the internal cache needs to expose its status for ```StatsCache``` to handle retries and logging.

statistics/handle/cache/internal/lru/lru_cache.go
hawkingrei: Done, it has been recovered.

statistics/handle/cache/statscacheinner.go
hawkingrei: Done

statistics/handle/cache/statscacheinner.go
hawkingrei:
```
h.updateStatsCache(oldCache.CopyAndUpdate(tables, deletedTableIDs, opts...))
```

I want to merge ```updateStatsCache``` and ```CopyAndUpdate``` into a function. if users use ```LFU```. we do not use COW. But this change will be big. so I will do it in the next PR.

hawkingrei: It is at https://github.com/pingcap/tidb/pull/45518.

statistics/handle/cache/statscacheinner.go
hawkingrei: Done

statistics/handle/cache/internal/lfu/lfu_cache.go
hawkingrei: In the old LRU, If you set capacity 0, you will have unlimited memory in the cache. It is dangerous.

planner/core/plan_stats_test.go
```diff
@@ -313,7 +313,6 @@ func TestPlanStatsStatusRecord(t *testing.T) {
 	// drop stats in order to change status
 	domain.GetDomain(tk.Session()).StatsHandle().SetStatsCacheCapacity(1)
 	tk.MustQuery("select * from t where b >= 1")
-	require.Equal(t, tk.Session().GetSessionVars().StmtCtx.RecordedStatsLoadStatusCnt(), 2)
```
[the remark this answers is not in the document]
hawkingrei: LFU cannot delete LRU after decreasing the capacity as soon as possible.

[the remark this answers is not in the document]
hawkingrei: the implementation of ristretto is to evict the cache when the user writes cache.

statistics/handle/cache/statscacheinner.go
hawkingrei: Done

statistics/handle/handle.go
hawkingrei: It mean current actual size.

time-and-fate (approved):

hawkingrei:
@qw4990 PTAL

statistics/handle/cache/statscacheinner.go
```diff
@@ -105,10 +122,30 @@ func (sc *StatsCache) PutFromInternal(id int64, t *statistics.Table) {
...
+		return ok
+	}
+	// retry three times and sleep exponentially. 20ms, 40ms and 80ms.
+	for i := 0; i < 3; i++ {
+		time.Sleep(time.Duration(1<<uint(i)) * 10 * time.Millisecond)
+		ok = sc.c.Put(id, t, moveLRUFront)
+		if ok {
+			return ok
+		}
+	}
```
qw4990, line 137: Please don't do that... this retry seems very dangerous, if the `StatsHandler` call this, in the worst case, `StatsHandler` will be blocked for `140ms`.
I suggest that we just log a warning here and do nothing about retry.

qw4990 (approved):

PR #45689, opened 2023-07-31
[DNM] 6.5.2 debug patch: print select result

ekexium:

Issue Number: close #xxx

Problem Summary:

For a select statement like
```sql
create table print_result (c int primary key, v varchar(20))
insert into print_result values (1, 'a')
insert into print_result values (2, 'b')
select * from print_result
```

Print a log like
```log
[2023/08/01 12:58:31.255 +08:00] [INFO] [util.go:607] ["statement result"] [conn=4014666597707809173] [txnStartTS=443250353436884993] [forUpdateTS=443250353436884993] [sql="select * from print_result"] [rows="(\u00011\u0001a);\n (\u00012\u0001b)"]
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/server/util.go
+++ b/server/util.go
@@ -37,20 +37,24 @@ package server
 
 import (
 	"bytes"
+	"context"
 	"encoding/binary"
 	"io"
 	"math"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/pingcap/tidb/config"
+	"github.com/pingcap/tidb/parser/ast"
 	"github.com/pingcap/tidb/parser/charset"
 	"github.com/pingcap/tidb/parser/mysql"
 	"github.com/pingcap/tidb/types"
 	"github.com/pingcap/tidb/util/chunk"
 	"github.com/pingcap/tidb/util/hack"
 	"github.com/pingcap/tidb/util/logutil"
+	"github.com/pingcap/tidb/util/sqlexec"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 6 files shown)


session/session.go
```diff
@@ -2236,6 +2237,49 @@ func (s *session) ExecuteStmt(ctx context.Context, stmtNode ast.StmtNode) (sqlex
...
+		zap.Uint64("start ts", vars.TxnCtx.StartTS),
+		zap.Uint64("for update ts", vars.TxnCtx.GetForUpdateTS()),
+		zap.String("rows", "("+strings.Join(resultRows, ");\n (")+")"),
+	)
+
+	copiedRs := &sqlexec.CopiedRecordSet{
+		Records:      records,
+		ResultFields: rs.Fields(),
+	}
+	rs.Close()
```
cfzjywxk, line 2279: The lifetime of `rs` is changed compared with the original version, would it introduce extra risks I'm not sure if there are implicit dependencies of it.

Or is it possible to print the results of the `Next` call so it's unnecessary to copy or clone here?

session/session.go
```diff
@@ -2236,6 +2237,49 @@ func (s *session) ExecuteStmt(ctx context.Context, stmtNode ast.StmtNode) (sqlex
...
+		for i := 0; i < chk.NumRows(); i++ {
+			row := chk.GetRow(i)
+			records = append(records, row)
+			resultRows = append(resultRows, row.ToString(fieldTypes))
+		}
+		chk = chunk.Renew(chk, 32)
+	}
+
+	// Print all rows
+	logutil.Logger(ctx).Info("statement returned result",
```
cfzjywxk, line 2267: If there are quite a lot of rows, there could be memory risk.

session/session.go
```diff
@@ -2359,8 +2403,18 @@ func runStmt(ctx context.Context, se *session, s sqlexec.Statement) (rs sqlexec.
 				}
 			}
 		}
+
+		copiedRs := rs
```
cfzjywxk, line 2407: May need a switch to turn it on or off the hook.

MyonKeminta (commented):
LGTM. The only problem of me is that, as the timing of performing physical read is changed (if I understood it correctly), I'm not sure whether it will affect reproduction of the issue.

session/session.go
MyonKeminta: Can we assume that the transaction sizes are not too big in the environment that we'are going to use this?
Another way to print them is to collect them when the `resultSet` is being consumed by calling `Next`

session/session.go
ekexium: I've set the maximum number of rows to store and log to 5. Do you think it still necessary to add a switch?

cfzjywxk: Fine with me, the limit `5` is a little bit small, maybe `50`. If it's enough to cover the questionable query that's OK.

ekexium:
I've changed it to a wrapper style. PTAL @MyonKeminta

server/util.go
```diff
@@ -543,3 +549,71 @@ func (h CorsHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+	zapFields     []zap.Field
+	maxCachedRows int
+}
+
+func (r *recordSetLoggerWrapper) Fields() []*ast.ResultField {
+	return r.rs.Fields()
+}
+
+func (r *recordSetLoggerWrapper) Next(ctx context.Context, req *chunk.Chunk) error {
+	chk := r.rs.NewChunk(nil)
```
MyonKeminta, line 576: It looks to me that the better way is to directly read to `req` and then get some rows from it, but it seems kind of difficult to know when `req` is reset? 🤔

MyonKeminta (commented):
LGTM

server/util.go
ekexium: Yeah that's not something we can control.

Issue #45719, opened 2023-08-01
Support global sort on S3


wjhuang2016:
## Enhancement
This PR collects some PRs to support global sort. 
TiDB servers can sort a large amount of data on shared storage(S3 for now), then ingect into TiKV without overlap.

PR #45724, opened 2023-08-01
br/lightning: add byte reader implementation

tangenta:
Issue Number: ref https://github.com/pingcap/tidb/issues/45719

Problem Summary: N/A

Add the byte reader implementation.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/lightning/backend/external/byte_reader.go
+++ b/br/pkg/lightning/backend/external/byte_reader.go
@@ -24,6 +24,7 @@ import (
 	"go.uber.org/zap"
 )
 
+// byteReader provides structured reading on a byte stream of external storage.
 type byteReader struct {
 	ctx           context.Context
 	storageReader storage.ReadSeekCloser
@@ -31,9 +32,10 @@ type byteReader struct {
 	buf       []byte
 	bufOffset int
 
-	auxBuf   []byte
-	cowSlice [][]byte // copy-on-write slices
-	isEOF    bool
+	auxBuf []byte
+	isEOF  bool
+
+	retPointers []*[]byte
 }
 
 func openStoreReaderAndSeek(
@@ -62,31 +64,17 @@ func newByteReader(ctx context.Context, storageReader storage.ReadSeekCloser, bu
 	}
 }
 
-type slice interface {
-	get() []byte
-}
-
-type commonSlice []byte
-
-func (c commonSlice) get() []byte {
-	return c
-}
-
-type lazySlice struct {
-	cowIdx     int
-	byteReader *byteReader
-}
-
-func (s lazySlice) get() []byte {
-	return s.byteReader.cowSlice[s.cowIdx]
-}
-
-// sliceNext reads the next n bytes from the reader and returns a buffer slice containing those bytes.
-func (r *byteReader) sliceNext(n int) (slice, error) {
+// readNBytes reads the next n bytes from the reader and returns a buffer slice containing those bytes.
+// The returned slice (pointer) can not be used after r.reset. In the same interval of r.reset,
+// byteReader guarantees that the returned slice (pointer) will point to the same content
+// though the slice may be changed.
+func (r *byteReader) readNBytes(n int) (*[]byte, error) {
 	b := r.next(n)
 	readLen := len(b)
 	if readLen == n {
-		return mkLazySlice(r, b), nil
+		ret := &b
+		r.retPointers = append(r.retPointers, ret)
+		return ret, nil
 	}
 	// If the reader has fewer than n bytes remaining in current buffer,
 	// `auxBuf` is used as a container instead.
@@ -105,24 +93,19 @@ func (r *byteReader) sliceNext(n int) (slice, error) {
 		copy(r.auxBuf[readLen:], b)
 		readLen += len(b)
 	}
-	return commonSlice(r.auxBuf), nil
+	return &r.auxBuf, nil
 }
 
-func mkLazySlice(r *byteReader, s []byte) lazySlice {
-	r.cowSlice = append(r.cowSlice, s)
-	return lazySlice{
-		cowIdx:     len(r.cowSlice) - 1,
-		byteReader: r,
-	}
-}
-
+// reset releases the tracked slice (pointer) returned by r.readNBytes.
 func (r *byteReader) reset() {
-	r.cowSlice = r.cowSlice[:0]
+	r.retPointers = r.retPointers[:0]
 }
 
 func (r *byteReader) cloneSlices() {
-	for i := range r.cowSlice {
-		r.cowSlice[i] = append([]byte(nil), r.cowSlice[i]...)
+	for i := range r.retPointers {
+		copied := make([]byte, len(*r.retPointers[i]))
+		copy(copied, *r.retPointers[i])
+		*r.retPointers[i] = copied
 	}
 }
 
--- a/br/pkg/lightning/backend/external/byte_reader.go
+++ b/br/pkg/lightning/backend/external/byte_reader.go
@@ -36,8 +36,12 @@ type byteReader struct {
 	isEOF    bool
 }
 
-func openStoreReaderAndSeek(ctx context.Context, store storage.ExternalStorage,
-	name string, initFileOffset uint64) (storage.ReadSeekCloser, error) {
+func openStoreReaderAndSeek(
+	ctx context.Context,
+	store storage.ExternalStorage,
+	name string,
+	initFileOffset uint64,
+) (storage.ReadSeekCloser, error) {
 	storageReader, err := store.Open(ctx, name)
 	if err != nil {
 		return nil, err
--- a/br/pkg/lightning/backend/external/byte_reader.go
+++ b/br/pkg/lightning/backend/external/byte_reader.go
@@ -82,13 +82,14 @@ func (s lazySlice) get() []byte {
 }
 
 // sliceNext reads the next n bytes from the reader and returns a buffer slice containing those bytes.
-// If the reader has fewer than n bytes remaining in current buffer, `auxBuf` is used as a container instead.
 func (r *byteReader) sliceNext(n int) (slice, error) {
 	b := r.next(n)
 	readLen := len(b)
 	if readLen == n {
 		return mkLazySlice(r, b), nil
 	}
+	// If the reader has fewer than n bytes remaining in current buffer,
+	// `auxBuf` is used as a container instead.
 	if cap(r.auxBuf) < n {
 		r.auxBuf = make([]byte, n)
 	}
--- a/br/pkg/lightning/backend/external/byte_reader.go
+++ b/br/pkg/lightning/backend/external/byte_reader.go
@@ -1,0 +1,157 @@
+// Copyright 2023 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package external
+
+import (
+	"context"
+	"io"
+
+	"github.com/pingcap/tidb/br/pkg/storage"
+	"github.com/pingcap/tidb/util/logutil"
+	"github.com/pingcap/tidb/util/mathutil"
+	"go.uber.org/zap"
+)
+
+type byteReader struct {
+	ctx           context.Context
+	storageReader storage.ReadSeekCloser
+
+	buf       []byte
+	bufOffset int
+
+	auxBuf   []byte
+	cowSlice [][]byte // copy-on-write slices
+	isEOF    bool
+}
+
+func openStoreReaderAndSeek(ctx context.Context, store storage.ExternalStorage,
+	name string, initFileOffset uint64) (storage.ReadSeekCloser, error) {
+	storageReader, err := store.Open(ctx, name)
+	if err != nil {
+		return nil, err
+	}
+	_, err = storageReader.Seek(int64(initFileOffset), io.SeekStart)
+	if err != nil {
+		return nil, err
+	}
+	return storageReader, nil
+}
+
+func newByteReader(ctx context.Context, storageReader storage.ReadSeekCloser, bufSize int) *byteReader {
+	return &byteReader{
+		ctx:           ctx,
+		storageReader: storageReader,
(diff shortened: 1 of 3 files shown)


br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -0,0 +1,161 @@
...
+	cowIdx     int
+	byteReader *byteReader
+}
+
+func (s lazySlice) get() []byte {
+	return s.byteReader.cowSlice[s.cowIdx]
+}
+
+// sliceNext reads the next n bytes from the reader and returns a buffer slice containing those bytes.
+// If the reader has fewer than n bytes remaining in current buffer, `auxBuf` is used as a container instead.
```
lance6716, line 85: I'm not sure what should the caller do with `auxBuf`. My suggestion is, the function comment only need to mention the needed concepts to caller. Comments inside the function can have more detailed explanation of code.

br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -0,0 +1,161 @@
...
+		r.cowSlice[i] = append([]byte(nil), r.cowSlice[i]...)
+	}
+}
+
+func (r *byteReader) eof() bool {
+	return r.isEOF && len(r.buf) == r.bufOffset
+}
+
+func (r *byteReader) next(n int) []byte {
+	end := mathutil.Min(r.bufOffset+n, len(r.buf))
```
lance6716, line 133: can we use a bytes.Reader?

br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -0,0 +1,161 @@
...
+	byteReader *byteReader
+}
+
+func (s lazySlice) get() []byte {
+	return s.byteReader.cowSlice[s.cowIdx]
+}
+
+// sliceNext reads the next n bytes from the reader and returns a buffer slice containing those bytes.
+// If the reader has fewer than n bytes remaining in current buffer, `auxBuf` is used as a container instead.
+func (r *byteReader) sliceNext(n int) (slice, error) {
```
lance6716, line 86: Is this a pattern that 
- The buffer may not have enough data (`n` bytes) so we need to read from source, and
- The data returned to caller maybe consumed asynchronously so COW is introduced ?

Do we have other consideration? I think there may exist some problems 🤔

lance6716: bufio.NewReader should be able to handle the first specification.

br/pkg/lightning/backend/external/byte_reader.go
tangenta: OK. I will move this inside the function.

br/pkg/lightning/backend/external/byte_reader.go
tangenta: There is copy overhead in `bytes.Reader`.

br/pkg/lightning/backend/external/byte_reader.go
tangenta: COW is not used for asynchronous handling. It is simply because `reload()` may change the slices referenced by previous `sliceNext()`.

br/pkg/lightning/backend/external/byte_reader.go
lance6716: Oh I see. We need to return the slice to caller so underlying byte array can be reused.

br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -0,0 +1,162 @@
...
+
+func mkLazySlice(r *byteReader, s []byte) lazySlice {
+	r.cowSlice = append(r.cowSlice, s)
+	return lazySlice{
+		cowIdx:     len(r.cowSlice) - 1,
+		byteReader: r,
+	}
+}
+
+func (r *byteReader) reset() {
```
lance6716, line 119: How often will the caller invoke `reset`? Because I see this is the only method that reduce the `cowSlice`, if it's not very often called, I'm a bot worried that `sliceNext` will copy all the `cowSlice` at last so we can't avoid the copy overhead.

lance6716: And if the caller always call `reset` after every `sliceNext`, we don't need to implement `lazySlice`. The usage pattern is important I think

lance6716: OK I found the usage in #45356

lance6716: In https://github.com/pingcap/tidb/blob/1b723a219b1fbb5888b68f2ab53e010655a2534c/br/pkg/lightning/backend/sharedisk/reader.go#L52
we reset the cowSlice, but the slice is still used in the kvHeap
https://github.com/pingcap/tidb/blob/1b723a219b1fbb5888b68f2ab53e010655a2534c/br/pkg/lightning/backend/sharedisk/iter.go#L110-L118
So following sliceNext will overwrite the underlying array, and thus change the slice in kvHeap

```
func TestNextKV(t *testing.T) {
	ms := &mockExtStore{src: []byte("0123456789")}
	br := newByteReader(context.Background(), ms, 1)
	err := br.reload()
	require.NoError(t, err)

	nextKV := func() ([]byte, []byte) {
		br.reset()
		key, err := br.sliceNext(1)
		require.NoError(t, err)
		val, err := br.sliceNext(1)
		require.NoError(t, err)
		return key.get(), val.get()
	}
	k1, v1 := nextKV()
	require.Equal(t, []byte("0"), k1)
	require.Equal(t, []byte("1"), v1)
	k2, v2 := nextKV()
	require.Equal(t, []byte("0"), k1)
	require.Equal(t, []byte("1"), v1) // fail because underlying array is changed to "3"
	require.Equal(t, []byte("2"), k2)
	require.Equal(t, []byte("3"), v2)
}
```

tangenta: > we reset the cowSlice, but the slice is still used in the kvHeap.

Yes, you are correct. We need to guarantee that the slice is not used anymore before calling `nextKV()/reset()`.

br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -0,0 +1,162 @@
...
+	"context"
+	"io"
+
+	"github.com/pingcap/tidb/br/pkg/storage"
+	"github.com/pingcap/tidb/util/logutil"
+	"github.com/pingcap/tidb/util/mathutil"
+	"go.uber.org/zap"
+)
+
+type byteReader struct {
```
lance6716, line 27: I want to add some comments for this structure, like:

byteReader provides structured reading on a byte stream of external storage.

br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -0,0 +1,162 @@
...
+func newByteReader(ctx context.Context, storageReader storage.ReadSeekCloser, bufSize int) *byteReader {
+	return &byteReader{
+		ctx:           ctx,
+		storageReader: storageReader,
+		buf:           make([]byte, bufSize),
+		bufOffset:     0,
+	}
+}
+
+type slice interface {
```
lance6716, line 65: can we use `*[]byte` rather than an interface?

br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -0,0 +1,162 @@
...
+	cowIdx     int
+	byteReader *byteReader
+}
+
+func (s lazySlice) get() []byte {
+	return s.byteReader.cowSlice[s.cowIdx]
+}
+
+// sliceNext reads the next n bytes from the reader and returns a buffer slice containing those bytes.
+func (r *byteReader) sliceNext(n int) (slice, error) {
```
lance6716, line 85: I slightly prefer we name it `readNBytes`

and these properties can be added to comment:
The returned slice (pointer) can not be used after r.reset. In the same interval of r.reset, byteReader guarantees that the returned slice (pointer) will point to the same content though the slice may be changed.

br/pkg/lightning/backend/external/byte_reader.go
lance6716, line 119:
```suggestion
// reset releases the tracked slice (pointer) returned by r.readNBytes.
func (r *byteReader) reset() {
```

br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -0,0 +1,162 @@
...
+		byteReader: r,
+	}
+}
+
+func (r *byteReader) reset() {
+	r.cowSlice = r.cowSlice[:0]
+}
+
+func (r *byteReader) cloneSlices() {
+	for i := range r.cowSlice {
```
lance6716, line 124: Now say slices A and B are referring to the `buf`. And "copy-on-write" means when we are going to write `buf`, we should copy the `buf` rather than copy A and B? Maybe change the name or change the code

br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	// If the reader has fewer than n bytes remaining in current buffer,
+	// `auxBuf` is used as a container instead.
+	if cap(r.auxBuf) < n {
+		r.auxBuf = make([]byte, n)
+	}
+	r.auxBuf = r.auxBuf[:n]
+	copy(r.auxBuf, b)
+	for readLen < n {
+		r.cloneSlices()
```
lance6716, line 87: we only need to cloneSlices once in one round of r.reset(). Maybe after the slice is cloned we don't need to put them back to r.retPointers?

br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -0,0 +1,145 @@
...
+	return ret
+}
+
+func (r *byteReader) reload() error {
+	nBytes, err := io.ReadFull(r.storageReader, r.buf[0:])
+	if err == io.EOF {
+		logutil.Logger(r.ctx).Error("unexpected EOF")
+		r.isEOF = true
+		return err
+	} else if err != nil && err == io.ErrUnexpectedEOF {
```
lance6716, line 129:
```suggestion
	} else if err == io.ErrUnexpectedEOF {
```

br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -0,0 +1,145 @@
...
+		r.cloneSlices()
+		err := r.reload()
+		if err != nil {
+			return nil, err
+		}
+		b = r.next(n - readLen)
+		copy(r.auxBuf[readLen:], b)
+		readLen += len(b)
+	}
+	return &r.auxBuf, nil
```
lance6716, line 96: the returned auxBuf is not cloned so its content will be changed. I'll check it by a test soon

br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// reset releases the tracked slice (pointer) returned by r.readNBytes.
+func (r *byteReader) reset() {
+	r.retPointers = r.retPointers[:0]
```
lance6716, line 101: and assign them to nil can avoid some `buf` of many rounds ago still being referenced by the underlying array of `retPointers`. For example, we have 4 call to `readNBytes` so 4 `retPointers` in one round of r.reset. But in following rounds we have 3 `retPointers`. The 4th element will always hold a reference to its `buf`

br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (r *byteReader) next(n int) []byte {
+	end := mathutil.Min(r.bufOffset+n, len(r.buf))
+	ret := r.buf[r.bufOffset:end]
+	r.bufOffset += len(ret)
+	return ret
+}
+
+func (r *byteReader) reload() error {
```
lance6716, line 123: need to mention that reload must be called before use firstly. Or we let readNBytes call reload for the first time?

br/pkg/lightning/backend/external/byte_reader.go
lance6716:
```golang
func TestAuxBuf(t *testing.T) {
	ms := &mockExtStore{src: []byte("0123456789")}
	br := newByteReader(context.Background(), ms, 1)
	require.NoError(t, br.reload())
	y1, err := br.readNBytes(1)
	require.NoError(t, err)
	y2, err := br.readNBytes(2)
	require.NoError(t, err)
	require.Equal(t, []byte("0"), *y1)
	require.Equal(t, []byte("12"), *y2)

	y3, err := br.readNBytes(1)
	require.NoError(t, err)
	y4, err := br.readNBytes(2)
	require.NoError(t, err)
	require.Equal(t, []byte("3"), *y3)
	require.Equal(t, []byte("45"), *y4)
	require.Equal(t, []byte("0"), *y1)
	require.Equal(t, []byte("12"), *y2)
}
```

Seems an irrelevant fail

lance6716 (commented):
we still need r.reset? otherwise data will always be cloned once. With reset `retPointers` can be released earlier.

rest lgtm

br/pkg/lightning/backend/external/byte_reader.go
tangenta: Done. I set the content of slice to `nil` explicitly.

br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -93,6 +93,10 @@ func (r *byteReader) readNBytes(n int) (*[]byte, error) {
 	return &auxBuf, nil
 }
 
+func (r *byteReader) reset() {
```
lance6716, line 96: not important: we can also set elements to nil before truncate the `retPointers`

lance6716 (approved):

br/pkg/lightning/backend/external/byte_reader.go
lance6716: maybe also add more reset() in unit tests

br/pkg/lightning/backend/external/byte_reader.go
```diff
@@ -0,0 +1,146 @@
...
+	end := mathutil.Min(r.bufOffset+n, len(r.buf))
+	ret := r.buf[r.bufOffset:end]
+	r.bufOffset += len(ret)
+	return ret
+}
+
+func (r *byteReader) reload() error {
+	nBytes, err := io.ReadFull(r.storageReader, r.buf[0:])
+	if err == io.EOF {
+		logutil.Logger(r.ctx).Error("unexpected EOF")
```
wjhuang2016, line 127: The EOF may be expected, so it shouldn't be ERROR

ywqzzy (approved):

wjhuang2016 (approved):

PR #45786, opened 2023-08-03
executor: union scan refactor for index read scenario

tiancaiamao:

Issue Number: ref #43249

Problem Summary:

- Implement the iterator for union scan index reader code path.
- refactor `txnMemBufferIter` 

The problem of the old `iterTxnMemBuffer` is that it drain all the rows all in once, so even the query is 'select * ... limit 1' it read all the data.
So I change the API to a iterator interface with `Next()` method (do it for TableReader first).

When I try to implement `Next()` for IndexReader, I realized that the code is not composable!

For composable code, the interface should be:

```
type Iter interface {
    Next() (key []byte, value []byte)
}
type Decoder {
    Process(key []byte, value []byte) []Datum
}
```

Combine iter and decoder into one make the code less re-useable.

```
type Iter interface {
    Next() []Datum
}
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


tiancaiamao:
For  `SELECT `id` FROM `AccessoryMaster` limit 1` using index reader case,
this PR can **improve the performance by 120X**，from 35.2 QPS --> 4.2K QPS, where the sql file of the table size is 40M.

executor/mem_reader.go
```diff
@@ -90,15 +90,36 @@ func buildMemIndexReader(ctx context.Context, us *UnionScanExec, idxReader *Inde
...
+	}
+
+	kvIter, err := newTxnMemBufferIter(m.ctx, m.cacheTable, m.kvRanges, m.desc)
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-	return &defaultRowsIter{data: data}, nil
+	tps := m.getTypes()
+	hdStatus := tablecodec.HandleDefault
+	if mysql.HasUnsignedFlag(tps[len(tps)-1].GetFlag()) {
```
mjonss, line 107: How does this work with PK(i int, u int unsigned) vs PK(u int unsigned, i int) ? I.e. can the handle be said to be either HandleIsUnsigned or HandleDefault in either of those cases?

executor/mem_reader.go
```diff
@@ -258,108 +284,95 @@ func buildMemTableReader(ctx context.Context, us *UnionScanExec, kvRanges []kv.K
...
+type kvProcessor interface {
+	processKV(key, value []byte) []types.Datum
+}
 
-	reverse  bool
-	cd       *rowcodec.ChunkDecoder
-	chk      *chunk.Chunk
-	datumRow []types.Datum
+// txnMemBufferIter implements a kv.Iterator, it is an iterator that combines the membuffer data and snapshot data.
+type txnMemBufferIter struct {
```
hawkingrei, line 292: Too much code is put into ```executor```. can we put it into ```executor/internal``` to save the cache ?

executor/mem_reader.go
tiancaiamao: I don't change the code logic, it's here originally:
https://github.com/pingcap/tidb/pull/45786/files#diff-f7f3339231e4104abed7ddd0b9a3e12d2284761165df20d2141406104051767cL159-L164

mjonss: Cool, do you see any possible bugs due to this? (not related to this pr)

executor/mem_reader.go
```diff
@@ -150,18 +174,15 @@ func (m *memIndexReader) getMemRows(ctx context.Context) ([][]types.Datum, error
 			}
 			return ret
 		})
-		return m.addedRows, err
```
mjonss, line 153: Should this be kept as `return nil, err`? In the case of the `m.compare` function gives an error in `slices.SortFunc()`?

tiancaiamao:
> > For `SELECT `id`FROM`AccessoryMaster` limit 1` using index reader case, this PR can **improve the performance by 120X**，from 35.2 QPS --> 4.2K QPS, where the sql file of the table size is 40M.
> 
> It should be that this pr plus other changes in planner?

No, just this commit. As is stated:

> The problem of the old `iterTxnMemBuffer` is that it drain all the rows all in once, so even the query is 'select * ... limit 1' it read all the data.

So that's where the 120X comes

@winoros

winoros:
> > > For `SELECT `id`FROM`AccessoryMaster` limit 1` using index reader case, this PR can **improve the performance by 120X**，from 35.2 QPS --> 4.2K QPS, where the sql file of the table size is 40M.
> > 
> > 
> > It should be that this pr plus other changes in planner?
> 
> No, just this commit. As is stated:
> 
> > The problem of the old `iterTxnMemBuffer` is that it drain all the rows all in once, so even the query is 'select * ... limit 1' it read all the data.
> 
> So that's where the 120X comes
> 
> @winoros

ohhhh, I got your idea. You mean that the performance is bad even if we just want to read data with a random limit. This pull can improve that case.

winoros (approved):

pkg/executor/mem_reader.go
```diff
@@ -859,6 +861,115 @@ func (iter *defaultRowsIter) Next() ([]types.Datum, error) {
...
+
+func (iter *memRowsIterForTable) Next() ([]types.Datum, error) {
+	curr := iter.kvIter
+	var ret []types.Datum
+	for curr.Valid() {
+		key := curr.Key()
+		value := curr.Value()
+		if err := curr.Next(); err != nil {
+			return nil, errors.Trace(err)
+		}
```
Defined2014, line 881: Will this cause some problems? For example, the current key and value meet the requirements, but `curr.Next()` causing an error to be returned.

Defined2014 (commented):
Please fill the performance `Check List`

pkg/executor/mem_reader.go
```diff
@@ -151,18 +175,15 @@ func (m *memIndexReader) getMemRows(ctx context.Context) ([][]types.Datum, error
 			}
 			return ret
 		})
-		return m.addedRows, err
 	}
 	return m.addedRows, nil
```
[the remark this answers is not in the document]
tiancaiamao: Good catch!  It must be a mistake introduced when I resolve the confict.

pkg/executor/mem_reader.go
tiancaiamao: OK, let me explain why the code looks like this:

The signature for Valid() method of kv.Iterator is Valid() bool
It's not expected to return an error.

txnMemBufferIter implements kv.Iterator, but in its Valid() implementation, error might happen.
So it record error in its iter.err field.

And when Next() is called, the error would be throw out.

So here it is basically equvalent to

```
if err := iter.Valid(); err != nil {
    return err
}
```

And once error happens, the error is throw to the top most of the caller, and I guess finally to the   mysql client.

Defined2014: So when `Valid()` meets error, it also return true which feel strange to me.

hawkingrei (approved):

pkg/executor/mem_reader.go
tiancaiamao: I think the current kv.Iterator API is somehow better than https://pkg.go.dev/database/sql#Rows.Next

It has a `Err()` method, and `Next()` return false for two reasons:
1. finish
2. error

While our `Valid()` just serve one purpose: 
1. finish

Our kv.Iterator do not provide the `Err()`, (because it's really common to forget check that after the for-loop). 
We return the error in the `Next()` which is much easier for the user to notice it.

tiancaiamao:
```
cd pkg/executor;
go test -tags intest -run XXX -bench BenchmarkUnionScanIndexReadDescRead -benchmem -benchtime 30s

Before:
   79682            447983 ns/op          135108 B/op       1920 allocs/op
After:
   81997            432501 ns/op          124309 B/op       1814 allocs/op
```

The bench test only buffer 100 rows, so the result differ is not significant enough https://github.com/pingcap/tidb/blob/1c8d383487172801d600c7e96d84794d87edd72c/pkg/executor/union_scan_test.go#L330-L332



=============

```
make ut X='run pkg/executor --coverprofile cover.out
make failpoint-disable
go tool cover -html cover.out
```

The coverage of mem_reader.go is 58.3%, but the code is this PR are almost included.
What's missing is the IndexMerge code branch, there are no UT cover that.

PR #45794, opened 2023-08-03
br: adjust default tidb global config for br

### LINKED ISSUE #45793: failed to restore DDL when upstream has some different limit config from default value
## Bug Report

1. upstream, set some limit config, suche as table-column-count, to its max value.
2. create a table with too many column or index.
3. backup
4. restore to downstream

1. success
failed to restore for DDL check error.
master

Leavrth:

Issue Number: close #45793

Problem Summary:
When the upstream set the column count limit to any value larger than the default one, the downstream would be not able to restore the data for DDL check error.
use the max value as the default value for br.
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/task/restore.go
+++ b/br/pkg/task/restore.go
@@ -1117,6 +1117,10 @@ func enableTiDBConfig() func() {
 		// when upstream and downstream both set this value greater than default(3072)
 		conf.MaxIndexLength = config.DefMaxOfMaxIndexLength
 		log.Warn("set max-index-length to max(3072*4) to skip check index length in DDL")
+		conf.IndexLimit = config.DefMaxOfIndexLimit
+		log.Warn("set index-limit to max(64*8) to skip check index count in DDL")
+		conf.TableColumnCountLimit = config.DefMaxOfTableColumnCountLimit
+		log.Warn("set table-column-count to max(4096) to skip check column count in DDL")
 	})
 	return restoreConfig
 }


br/pkg/task/restore.go
```diff
@@ -1117,6 +1117,10 @@ func enableTiDBConfig() func() {
 		// when upstream and downstream both set this value greater than default(3072)
 		conf.MaxIndexLength = config.DefMaxOfMaxIndexLength
```
lance6716, line 1118: seems fail on same type of mistake again? Do you consider to skip checks when creating table? like https://github.com/pingcap/tidb/blob/522a00840385dfa43509a3af2bc79e6ff50252a8/ddl/schematracker/dm_tracker.go#L178

lance6716: https://github.com/pingcap/tidb/blob/522a00840385dfa43509a3af2bc79e6ff50252a8/ddl/schematracker/dm_tracker_test.go#L50

DM has met a similar problem before, which is upstream MySQL has more capability than TiDB, so we use a special DDL implementation to apply DDL SQL.

Leavrth: There is a check before create table with DDL.
https://github.com/pingcap/tidb/blob/57d778f0b1b6daebe0f330af2baa54e9e5310ea4/ddl/ddl_api.go#L2337-L2352

When use br, every check config is default value. If upstream has large value, for example set `TableColumnCountLimit` to 4000, the restore might be failed when there is a table with 4000 columns.

Leavrth: BR import the DDL package, but doesn't expose these configs.

3pointer (approved):
please leave to TODO here, we need customize check for `createTableWithInfo`

PR #17149, opened 2020-05-12
ddl: Support exchange partition

zhaox1n:
UCP #15005

### What is changed and how it works?

Support exchange partition

### Check List 

Tests 

- Unit test

### Release note 
- No release note

PR #45877, opened 2023-08-07
ddl: Exchange partition rollback

### LINKED ISSUE #45791: exchange partition will fail when table has placement policy and then insert will fail forever
## Bug Report

```sql
CREATE PLACEMENT POLICY rule1 FOLLOWERS=4;
CREATE TABLE `t1` (
  `date` date NOT NULL ,
  `name` varchar(10)  NOT NULL,
  UNIQUE KEY `uk_1` (`date`,`name`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin PLACEMENT POLICY=`rule1`;


CREATE TABLE `t1_partition` (
  `date` date NOT NULL ,
  `name` varchar(10)  NOT NULL,
  UNIQUE KEY `uk_1` (`date`,`name`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin  PLACEMENT POLICY=`rule1`
PARTITION BY RANGE COLUMNS(`date`)
(PARTITION `p202307` VALUES LESS THAN ('2023-08-01'),
 PARTITION `p202308` VALUES LESS THAN ('2023-09-01'),
 PARTITION `p202309` VALUES LESS THAN ('2023-10-01'),
 PARTITION `p202310` VALUES LESS THAN ('2023-11-01'),
 PARTITION `p202311` VALUES LESS THAN ('2023-12-01'),
 PARTITION `p202312` VALUES LESS THAN ('2024-01-01'),
 PARTITION `pfuture` VALUES LESS THAN (MAXVALUE));

alter table t1_partition exchange partition p202307 with table t1; -- this query will fail
insert into t1(date, name) values ("2023-08-06","0000"); -- then this insert will fail too
```

no error

```
TiDB root@127.0.0.1:test> alter table t1_partition exchange partition p202307 with table t1;
(1736, 'Tables have different definitions')

TiDB root@127.0.0.1:test> insert into t1(date, name) values ("2023-08-06","0000");
(1748, 'Found a row not matching the given partition set')
```

master, but I think it affects TiDB after v5.4.0

### LINKED ISSUE #45920: Table gone after concurrent EXCHANGE PARTITION
## Bug Report

```sql
tidb1> create table t (a int);
tidb1> create table tp (a int) partition by hash (a) partitions 3;
tidb1> create table t2 (a int);

tidb2> begin;
tidb2> insert into tp values (1);

tidb1> alter table tp exchange partition p0 with table t;

tidb3> alter table tp exchange partition p0 with table t2; -- fails error 8038, information schema changed

tidb2> rollback;

tidb1> select * from t2; -- Fails: ERROR 1146 (42S02): Table 'test.t2' doesn't exist

```
Failure with out the table removed

Failure and table t2 no longer found.

```
tidb_version(): Release Version: v7.4.0-alpha
Edition: Community
Git Commit Hash: d590ccb14e2e36610f7a1313ed75904088b7d84e
Git Branch: heads/refs/tags/v7.4.0-alpha
UTC Build Time: 2023-08-07 14:24:53
GoVersion: go1.20.7
Race Enabled: false
Store: tikv
```

mjonss:

Issue Number: close #45791, close #45920

Problem Summary:
The first schema change state adds a restriction to the non-partitioned table, so that it can only write (insert/update) rows that would fit into the partition to be exchanged with. But if something fails afterwards, this is never cleared.

Adding rollback, in case of user cancellation or failure during the DDL execution, after the first state change.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Manual test for upgrade scenario, revert the changes made to `ddl/ddl_worker.go` (which sets the `diff` struct) and run the tests manually.

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


ddl/partition.go
```diff
@@ -2388,35 +2391,58 @@ func (w *worker) onExchangeTablePartition(d *ddlCtx, t *meta.Meta, job *model.Jo
...
-		return ver, dbterror.ErrInvalidDDLState.GenWithStack("table %s is not in public, but %s", pt.Name, pt.State)
-	}
-
-	err = checkExchangePartition(pt, nt)
+	index, partDef, err := getPartitionDef(pt, partName)
 	if err != nil {
-		job.State = model.JobStateCancelled
 		return ver, errors.Trace(err)
 	}
+	if job.SchemaState == model.StateNone {
```
mjonss, line 2398: If there is a running Exchange Partition from previous version, there is no harm in redoing this step and set job.SchemaState and reset ExchangePartitionInfo.

ddl/partition.go
```diff
@@ -2425,27 +2451,19 @@ func (w *worker) onExchangeTablePartition(d *ddlCtx, t *meta.Meta, job *model.Jo
...
-			job.State = model.JobStateCancelled
+			job.State = model.JobStateRollingback
 			return ver, errors.Trace(err)
 		}
 	}
 
 	// partition table auto IDs.
 	ptAutoIDs, err := t.GetAutoIDAccessors(ptSchemaID, ptID).Get()
 	if err != nil {
-		job.State = model.JobStateCancelled
```
mjonss, line 2436: Let the runDDLJob etc handle the error, by retrying instead of cancelling or rolling back.

infoschema/builder.go
```diff
@@ -309,6 +311,47 @@ func (b *Builder) applyReorganizePartition(m *meta.Meta, diff *model.SchemaDiff)
...
+func (b *Builder) applyExchangeTablePartition(m *meta.Meta, diff *model.SchemaDiff) ([]int64, error) {
+	// The partitioned table is not affected until the last stage
+	if diff.OldTableID == diff.TableID && diff.OldSchemaID == diff.SchemaID {
+		return b.applyTableUpdate(m, diff)
+	}
+	ntSchemaID := diff.OldSchemaID
+	ntID := diff.OldTableID
+	ptSchemaID := diff.SchemaID
+	ptID := diff.TableID
+	if len(diff.AffectedOpts) > 0 {
```
bb7133, line 323: When we have `len(diff.AffectedOpts) > 0` for exchange partitions?

I found that `AffectedOpts` was introduced in https://github.com/pingcap/parser/pull/883 with https://github.com/pingcap/tidb/pull/17149. However, I cannot think of any case of `len(diff.AffectedOpts) > 0`.

/cc @crazycs520 @tiancaiamao

ddl/db_partition_test.go
```diff
@@ -3282,6 +3282,71 @@ func TestExchangePartitionTableCompatiable(t *testing.T) {
...
+	 PARTITION p202312 VALUES LESS THAN ('2024-01-01'),
+	 PARTITION pfuture VALUES LESS THAN (MAXVALUE))`)
+
+	tk.MustExec(`insert into t1 values ("2023-08-06","0000")`)
+	tk.MustContainErrMsg(`alter table t1p exchange partition p202307 with table t1 with validation`,
+		"[ddl:1737]Found a row that does not match the partition")
+	tk.MustExec(`insert into t1 values ("2023-08-06","0001")`)
+}
+
+func TestExchangePartitionPlacementPolicy(t *testing.T) {
```
lcwangchao, line 3317: I think you can add some cases in `TestExchangePartitionWithPlacement` in the file `placement_policy_test.go` so that you can check some placement bundle settings are also set correctly.

infoschema/builder.go
mjonss: This is in the case of upgrades, if there is an existing diff that is not yet processed, from an earlier version. Not sure the risk of that to happen though...
I manually tested it by reverting the `ddl/ddl_worker.go` file, which is where I changed/cleaned how to set the diff.

ddl/db_partition_test.go
mjonss: Thanks for the suggestion, I just added more complete tests there.

ddl/ddl_api.go
```diff
@@ -4746,7 +4746,6 @@ func checkExchangePartition(pt *model.TableInfo, nt *model.TableInfo) error {
 		return errors.Trace(dbterror.ErrPartitionExchangeForeignKey.GenWithStackByArgs(nt.Name))
 	}
 
-	// NOTE: if nt is temporary table, it should be checked
```
mjonss, line 4749: this was already checked, in `checkTableDefCompatible()`

bb7133 (approved):
LGTM

Issue #38361, opened 2022-10-10
Wrong results and nonoptimal plan when datetime column comparing number constant


time-and-fate:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```sql
CREATE TABLE `tt` (
  `CREATED_DATE` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
  KEY `ii` (`CREATED_DATE`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;

-- 5 values on the day before yesterday
insert into tt value
(date_sub(curdate(), interval 2 day)),
(date_add(date_sub(curdate(), interval 2 day), interval 1 minute)),
(date_sub(date_sub(now(), interval 2 day), interval 1 hour)),
(date_add(date_sub(now(), interval 2 day), interval 1 hour)),
(date_sub(date_sub(curdate(), interval 1 day), interval 1 minute));
-- 5 values on yesterday
insert into tt value
(date_sub(curdate(), interval 1 day)),
(date_add(date_sub(curdate(), interval 1 day), interval 1 minute)),
(date_sub(date_sub(now(), interval 1 day), interval 1 hour)),
(date_add(date_sub(now(), interval 1 day), interval 1 hour)),
(date_sub(curdate(), interval 1 minute));
-- 5 values on today
insert into tt value
(curdate()),
(date_add(curdate(), interval 1 minute)),
(date_sub(now(), interval 1 hour)),
(date_add(now(), interval 1 hour)),
(date_sub(date_add(curdate(), interval 1 day), interval 1 minute));
-- 5 values on tomorrow
insert into tt value
(date_add(curdate(), interval 1 day)),
(date_add(date_add(curdate(), interval 1 day), interval 1 minute)),
(date_sub(date_add(now(), interval 1 day), interval 1 hour)),
(date_add(date_add(now(), interval 1 day), interval 1 hour)),
(date_sub(date_add(curdate(), interval 2 day), interval 1 minute));

insert into tt value ('2000-1-1'); -- an extra value
select * from tt a where a.CREATED_DATE > curdate()-1;
select * from tt a where a.CREATED_DATE > subdate(curdate(),1);
```

### 2. What did you expect to see? (Required)

Result of MySQL 8.0.30:
```
127.0.0.1:3306[test]> select * from tt a where a.CREATED_DATE > curdate()-1;
+---------------------+
[...]

PR #45945, opened 2023-08-09
expression: Fix different behaviors with MySQL when comparing datetime column with numeric constant | tidb-test=pr/2196

yibin87:

Issue Number: close #38361

Problem Summary: When compare datetime/timestamp column with numeric constant, try convert numeric constant to timestamp datetime, do nothing if conversion failed.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/executor/executor_test.go
+++ b/executor/executor_test.go
@@ -125,3 +125,37 @@ func TestSetOperation(t *testing.T) {
 	rows := result.Rows()
 	require.Len(t, rows, 0)
 }
+
+func TestCompareIssue38361(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("drop database if exists TEST1")
+	tk.MustExec("create database TEST1")
+	tk.MustExec("use TEST1")
+	tk.MustExec("create table t(a datetime, b bigint)")
+	tk.MustExec("insert into t values(cast('2023-08-09 00:00:00' as datetime), 20230809)")
+
+	tk.MustQuery("select a > 20230809 from t").Check(testkit.Rows("0"))
+	tk.MustQuery("select a = 20230809 from t").Check(testkit.Rows("1"))
+	tk.MustQuery("select a < 20230810 from t").Check(testkit.Rows("1"))
+	tk.MustQuery("select 20230809 < a from t").Check(testkit.Rows("0"))
+	tk.MustQuery("select 20230809 = a from t").Check(testkit.Rows("1"))
+	tk.MustQuery("select 20230810 > a from t").Check(testkit.Rows("1"))
+
+	// constant datetime cmp numeric constant should be compared as real data type
+	tk.MustQuery("select cast('2023-08-09 00:00:00' as datetime) > 20230809 from t").Check(testkit.Rows("1"))
+	tk.MustQuery("select cast('2023-08-09 00:00:00' as datetime) = 20230809 from t").Check(testkit.Rows("0"))
+	tk.MustQuery("select cast('2023-08-09 00:00:00' as datetime) < 20230810 from t").Check(testkit.Rows("0"))
+	tk.MustQuery("select 20230809 < cast('2023-08-09 00:00:00' as datetime) from t").Check(testkit.Rows("1"))
+	tk.MustQuery("select 20230809 = cast('2023-08-09 00:00:00' as datetime) from t").Check(testkit.Rows("0"))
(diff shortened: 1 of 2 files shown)


executor/executor_test.go
```diff
@@ -125,3 +125,43 @@ func TestSetOperation(t *testing.T) {
...
+	tk.MustExec("drop database if exists TEST1")
+	tk.MustExec("create database TEST1")
+	tk.MustExec("use TEST1")
+	tk.MustExec("create table t(a datetime, b bigint, c bigint)")
+	tk.MustExec("insert into t values(cast('2023-08-09 00:00:00' as datetime), 20230809, 20231310)")
+
+	tk.MustQuery("select a > 20230809 from t").Check(testkit.Rows("0"))
+	tk.MustQuery("select a = 20230809 from t").Check(testkit.Rows("1"))
+	tk.MustQuery("select a < 20230810 from t").Check(testkit.Rows("1"))
+	tk.MustQuery("select a < 20231310 from t").Check(testkit.Rows("0"))
```
windtalker, line 142: This is the case that numeric can't convert to datetime, so use real data type to compare?

expression/builtin_compare.go
```diff
@@ -1699,6 +1727,31 @@ func (c *compareFunctionClass) refineArgs(ctx sessionctx.Context, args []Express
...
 
+// see https://github.com/pingcap/tidb/issues/38361 for more details
+func (c *compareFunctionClass) refineNumericConstantCmpDatetime(ctx sessionctx.Context, args []Expression, constArg *Constant, constArgIdx int) []Expression {
+	dt, err := constArg.Eval(chunk.Row{})
+	if err != nil {
+		return args
+	}
+	sc := ctx.GetSessionVars().StmtCtx
+	var timestampDatum types.Datum
+	targetFieldType := types.NewFieldType(mysql.TypeTimestamp)
```
windtalker, line 1738: Why the target type is timestamp instead of datetime?

expression/builtin_compare.go
```diff
@@ -1699,6 +1727,31 @@ func (c *compareFunctionClass) refineArgs(ctx sessionctx.Context, args []Express
...
+	timestampDatum, err = dt.ConvertTo(sc, targetFieldType)
+	if err != nil {
+		return args
+	}
```
windtalker, line 1742: maybe also need to return original args if `timestampDatum` is null but `dt` is not null

executor/executor_test.go
yibin87: Yes

expression/builtin_compare.go
yibin87: According to MySQL document, "If one of the arguments is a [TIMESTAMP](https://dev.mysql.com/doc/refman/8.0/en/datetime.html) or [DATETIME](https://dev.mysql.com/doc/refman/8.0/en/datetime.html) column and the other argument is a constant, the constant is converted to a timestamp before the comparison is performed"

expression/builtin_compare.go
yibin87: Make sense, add it.

expression/builtin_compare.go
yibin87: Check that the MySQL seems convert to datetime instead, because it can handle numeric constant that is beyond timestamp's valid value range.

executor/executor_test.go
windtalker: Add some comments here?

expression/builtin_compare.go
```diff
@@ -1699,6 +1727,31 @@ func (c *compareFunctionClass) refineArgs(ctx sessionctx.Context, args []Express
...
+func (c *compareFunctionClass) refineNumericConstantCmpDatetime(ctx sessionctx.Context, args []Expression, constArg *Constant, constArgIdx int) []Expression {
+	dt, err := constArg.Eval(chunk.Row{})
+	if err != nil {
+		return args
+	}
+	sc := ctx.GetSessionVars().StmtCtx
+	var datetimeDatum types.Datum
+	targetFieldType := types.NewFieldType(mysql.TypeDatetime)
+	datetimeDatum, err = dt.ConvertTo(sc, targetFieldType)
+	if err != nil || datetimeDatum.IsNull() {
```
windtalker, line 1740: Don't need to check `dt.IsNull()` here?

windtalker (approved):
Others LGTM

expression/builtin_compare.go
yibin87: ok, added the comments and check.

wshwsh12 (approved):

PR #45952, opened 2023-08-09
executor: change log level to INFO when fetching next chunk failed

### LINKED ISSUE #45951: The log level when fetching next chunk failed is a bit high
## Enhancement
When fetching next chunk failed, it may be caused by the server canceled the SQL execution, and it's  message will finally send to the client, so it could not be ignored. But the current log level of it is ERROR, it is a bit too high, which makes some users concern about it.

jackysp:

Issue Number: close #45951

Problem Summary:

Change the log level to INFO.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/executor/distsql.go
+++ b/executor/distsql.go
@@ -1492,7 +1492,7 @@ func (w *tableWorker) executeTask(ctx context.Context, task *lookupTableTask) er
 		chk := tryNewCacheChunk(tableReader)
 		err = Next(ctx, tableReader, chk)
 		if err != nil {
-			logutil.Logger(ctx).Error("table reader fetch next chunk failed", zap.Error(err))
+			logutil.Logger(ctx).Info("table reader fetch next chunk failed", zap.Error(err))
 			return err
 		}
 		if chk.NumRows() == 0 {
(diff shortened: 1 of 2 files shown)


executor/distsql.go
```diff
@@ -1492,7 +1492,7 @@ func (w *tableWorker) executeTask(ctx context.Context, task *lookupTableTask) er
 		chk := tryNewCacheChunk(tableReader)
 		err = Next(ctx, tableReader, chk)
 		if err != nil {
-			logutil.Logger(ctx).Error("table reader fetch next chunk failed", zap.Error(err))
+			logutil.Logger(ctx).Info("table reader fetch next chunk failed", zap.Error(err))
```
jackysp, line 1495: It seems it's hard to reach this line in the unit tests.

hawkingrei: Now tidb is in go1.21. in Go 1.21, The new [AfterFunc](https://tip.golang.org/pkg/context/#AfterFunc) function registers a function to run after a context has been cancelled. so maybe we check it to set log level.

hawkingrei: WithCancelCause will be better?

jackysp: I found that sometimes this log still has unique value. It feels better to keep it as it is.

PR #45975, opened 2023-08-10
parser: update golang.org/x/exp/slices

### LINKED ISSUE #45629: slices.SortFunc API change breaks build
## Bug Report

Update golang.org/x/exp dependency to latest.

Successful build.

Build failures due to API change for golang.org/x/exp/slices.SortFunc. SortFunc now takes a compare function (returns <= -1 for less, 0 for equal, >= 1 for greater) instead of a less function (returns true for less, false for greater than or equal).

See new API here: https://pkg.go.dev/golang.org/x/exp@v0.0.0-20230725093048-515e97ebf090/slices#SortFunc

Commit that introduced breaking change here: https://cs.opensource.google/go/x/exp/+/302865e7556b4ae5de27248ce625d443ef4ad3ed

master

### LINKED ISSUE #45976: parser:  golang.org/x/exp/slices usage is outdated
```
# github.com/pingcap/tidb/parser/charset
/go/pkg/mod/github.com/pingcap/tidb/parser@v0.0.0-20230810102603-1005894c17ff/charset/charset.go:94:28: type func(i *Charset, j *Charset) bool of func(i, j *Charset) bool {…} does not match inferred type func(a *Charset, b *Charset) int for func(a E, b E) int
```

The `slices.SortFunc` changed it's signature.

ernado:

Issue Number: close #45976 

close #45629

Update golang.org/x/exp/slices and it's usage.

### Check List

Tests 

- [x] Unit test

Side effects

Update for `golang.org/x/exp/slices` is required.

### Release note

```release-note
None
```


parser/charset/charset.go
```diff
@@ -91,8 +91,8 @@ func GetSupportedCharsets() []*Charset {
 	}
 
 	// sort charset by name.
-	slices.SortFunc(charsets, func(i, j *Charset) bool {
```
hawkingrei, line 94: Now tidb use go1.21. we can use std package ```slices``` to replace the ```exp/slices```.

hawkingrei:
@ernado It seems that there will be a considerable amount of work involved.

parser/charset/charset.go
ernado: Should we also require `go1.21` for `github.com/pingcap/tidb/parser`?
It can be used outside `tidb` and is currently has `go 1.18` in [go.mod](https://github.com/pingcap/tidb/blob/master/parser/go.mod)

ernado:
Oh, I see, the `slices` is used in much more places.

I'll close this PR for now, thanks 🙃

I will stick to my workaround (`replace` directive with manually fixed error in vendored `parser package`.

hawkingrei:
> Oh, I see, the `slices` is used in much more places.
> 
> I'll close this PR for now, thanks 🙃
> 
> I will stick to my workaround (`replace` directive with manually fixed error in vendored `parser` package.

Don‘t worry about it. We will definitely solve this problem.

parser/charset/charset.go
lance6716: considering parser may be publicly used by other applications, I prefer we don't upgrade golang versions too aggressively. my opinion is parser module should use the second latest major version of golang. @hawkingrei

PR #46021, opened 2023-08-12
partition: make ExchangePartition follow check constraints(part1)

### LINKED ISSUE #45922: EXCHANGE PARTITION does not follow CHECK Constraints
## Bug Report

```
set @@global.tidb_enable_check_constraint = 1;
create table t (a int check (a > 50) ENFORCED);
create table tp (a int check (a < 50) ENFORCED) partition by range (a) (partition p1 values less than (100));
insert into t values (75);
insert into tp values (25);
alter table tp exchange partition p1 with table t;
select * from t; -- shows 25, which breaks the CHECK constraint
insert into t values (25); -- fails as it should
select * from tp; -- shows 75, which breaks the CHECK constraint
insert into tp values (75); -- fail as it should
```

That the alter would keep the CHECK Constraint enforced

The CHECK constraint is not enforced.

```
tidb_version(): Release Version: v7.4.0-alpha-43-gcdeac4c66f
Edition: Community
Git Commit Hash: cdeac4c66f728ad60a5166ef7700efa52cbd1384
Git Branch: master
UTC Build Time: 2023-08-08 14:27:32
GoVersion: go1.20.7
Race Enabled: false
Store: unistore
```

jiyfhust:

Issue Number: close #45922

Problem Summary:

Part1: 
only verify records by check constraints.
prohibiting insert or update records which not satisfy check constraints is not included. 

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


jiyfhust:
It is ready for review now. @mjonss

ddl/partition.go
```diff
@@ -2448,6 +2455,14 @@ func (w *worker) onExchangeTablePartition(d *ddlCtx, t *meta.Meta, job *model.Jo
 		}
 	}
 
+	if variable.EnableCheckConstraint.Load() {
```
mjonss, line 2458: I think this should also be under the `withValidation` check, i.e. if `WITHOUT VALIDATION` is given, we should trust the user and not start a big scan to validate the CHECK, just like we would skip checking that all rows match the partition.

ddl/partition.go
```diff
@@ -3313,6 +3328,76 @@ func checkExchangePartitionRecordValidation(w *worker, pt *model.TableInfo, inde
...
+			if !con.Enforced {
+				continue
+			}
+			if con.State == model.StateDeleteOnly || con.State == model.StateDeleteReorganization {
+				continue
+			}
+			writeableConstraintExpr = append(writeableConstraintExpr, con.ExprString)
+		}
+		return writeableConstraintExpr
+	}
```
mjonss, line 3344: Would it be possible to reuse `TableCommon.WritableConstraints()` instead?

ddl/partition.go
jiyfhust: Yes, It is a good suggestion.  
But there is also a problem,  when withValidation is on, duraing write only state, there seems to no good method to check withValidation is on.   
And current Validation check also has the same problem.

ddl/partition.go
jiyfhust: If add a interface for "TableCommon.WritableConstraints" it will ok. 
I did not find a way to directly change table to TableCommon and then call WritableConstraints.
I add a interface wrapper in the other pr.

ddl/partition.go
mjonss: The `WITH[OUT] VALIDATION` is only for existing rows in the non-partitioned table (and with CHECK also for the partition). It does not need to control the insert/update that may concurrently happen, it is OK if those concurrent checks is always on. Ref [Exchange partition in MySQL Manual](https://dev.mysql.com/doc/refman/8.0/en/partitioning-management-exchange.html)

jiyfhust:
PTAL @mjonss

ddl/partition.go
```diff
@@ -2446,6 +2453,21 @@ func (w *worker) onExchangeTablePartition(d *ddlCtx, t *meta.Meta, job *model.Jo
 			job.State = model.JobStateRollingback
 			return ver, errors.Trace(err)
 		}
+		if variable.EnableCheckConstraint.Load() {
```
mjonss, line 2456: I would prefer if the check is combined with the `checkExchangePartitionRecordValidation` above, to only do a single scan of the non-partitioned table.

tiancaiamao (approved):

mjonss (approved):
Please create a followup issue, to test null handling in validation, since I assume there may be hidded issues in non-list cases?

jiyfhust:
> Please create a followup issue, to test null handling in validation, since I assume there may be hidded issues in non-list cases?

Issue created #47167.

PR #46030, opened 2023-08-13
partition: make ExchangePartition follow check constraints during writeOnly state(Part2)

jiyfhust:

Issue Number: ref #45922

Problem Summary:

Part2:
prohibiting insert or update records which not satisfy check constraints during ddl write only state.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


executor/write.go
```diff
@@ -45,6 +46,76 @@ var (
...
+				return err
+			}
+		}
+	} else {
+		if variable.EnableCheckConstraint.Load() {
+			p, ok := t.(table.PartitionedTable)
+			if !ok {
+				return errors.Errorf("exchange partition process assert table partition failed")
+			}
+			physicalTable, err := p.GetPartitionByRow(sctx, row)
```
mjonss, line 89: If this was called from `AddRecord()`/`UpdateRecord()` instead of the `fillRow()`/`updateRecord()`, then the partition pruning would already been done. Could that be a follow-up PR?

parser/model/model.go
```diff
@@ -518,7 +518,11 @@ type TableInfo struct {
 	// StatsOptions is used when do analyze/auto-analyze for each table
 	StatsOptions *StatsOptions `json:"stats_options"`
 
+	// Current table is non-partition table.
 	ExchangePartitionInfo *ExchangePartitionInfo `json:"exchange_partition_info"`
+	// Current table is partition table.
+	// Key: partID, value: non-partition tableID.
+	ExchangePartitionPartIDs map[int64]int64 `json:"exchange_partition_partids"`
```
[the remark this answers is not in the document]
mjonss: Oh, sorry, please ignore this suggestion, it was part of a draft review.

executor/write.go
jiyfhust: I'm not very sure about what you said. Could you explain more about it?

ddl/ddl_worker.go
```diff
@@ -1404,6 +1404,8 @@ func updateSchemaVersion(d *ddlCtx, t *meta.Meta, job *model.Job, multiInfos ...
 			// Keep this as Schema ID of non-partitioned table
 			// to avoid trigger early rename in TiFlash
 			diff.AffectedOpts[0].SchemaID = job.SchemaID
+			// Mark it is not StatePublic.
+			diff.AffectedOpts[0].OldSchemaID = ptSchemaID
```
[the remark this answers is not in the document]
mjonss: I think you can add the partitioned table's TableInfo + SchemaID into a new schemaIDAndTableInfo in the `updateVersionAndTableInfoWithCheck()` call in `onExchangeTablePartition()` function, so it becomes a single `multiInfos` argument, then if it is set, you can use it in the `updateSchemaVersion()` function.

jiyfhust:
PTAL @mjonss

executor/write.go
mjonss: Hi and sorry for the delay. Here is the stack from this line:
Screenshot 2023-09-08 at 16 45 16
And later in AddRecord locatePartition is called again:
Screenshot 2023-09-08 at 16 45 37
Which then results in the constraint check and insert (similar with update/deletes) will be applied to the physical partition here:
Screenshot 2023-09-08 at 16 49 44

So I'm just curious if we should change the place in the code where these checks for Exchange Partition is executed, to somewhere where we already have the located partition?
But we can fix that in a followup PR, in that case.

jiyfhust: Good advice, you really showed strong experience in this. I'm agree with you, and fix it in this pr together.

ddl/partition.go
```diff
@@ -2457,16 +2457,27 @@ func (w *worker) onExchangeTablePartition(d *ddlCtx, t *meta.Meta, job *model.Jo
...
+			ptInfo = append(ptInfo, schemaIDAndTableInfo{
+				schemaID: ptSchemaID,
+				tblInfo:  pt,
+			})
+		}
 		nt.ExchangePartitionInfo = &model.ExchangePartitionInfo{
-			ExchangePartitionID:    ptID,
-			ExchangePartitionDefID: defID,
+			ExchangePartitionTableID:     ptID,
+			ExchangePartitionPartitionID: defID,
```
mjonss, line 2473: Not sure `ExchangePartitionPartitionID` is better than `ExchangePartitionPartitionID`?
```suggestion
			ExchangePartitionDefID: defID,
```

executor/insert_common.go
```diff
@@ -690,28 +689,15 @@ func (e *InsertValues) fillRow(ctx context.Context, row []types.Datum, hasValue
...
-			tbl.ExchangePartitionInfo.ExchangePartitionDefID,
-		)
-		if err != nil {
+	tbl := e.Table.Meta()
+	if tbl.ExchangePartitionInfo != nil && tbl.GetPartitionInfo() == nil {
+		if err := checkRowForExchangePartition(e.Ctx(), row, tbl); err != nil {
 			return nil, err
 		}
 	}
+
```
mjonss, line 700: Does this check need to be here, or could it be moved to `AddRecord` [just before checking the constraints](https://github.com/pingcap/tidb/blob/1159bc8d88c4b35a957087813d1087fbcd519725/table/tables/tables.go#L970)?

parser/model/model.go
```diff
@@ -1164,8 +1164,9 @@ func (p PartitionType) String() string {
 
 // ExchangePartitionInfo provides exchange partition info.
 type ExchangePartitionInfo struct {
-	ExchangePartitionID    int64 `json:"exchange_partition_id"`
-	ExchangePartitionDefID int64 `json:"exchange_partition_def_id"`
+	// It is nt tableID when table which has the info is a partition table, else pt tableID.
+	ExchangePartitionTableID     int64 `json:"exchange_partition_table_id"`
+	ExchangePartitionPartitionID int64 `json:"exchange_partition_partition_id"`
```
mjonss, line 1169: Same, I think you can keep `ExchangePartitionDefID` here
```suggestion
	ExchangePartitionDefID int64 `json:"exchange_partition_def_id"`
```

executor/insert_common.go
jiyfhust: I think it it not a very bad way to put it here because `AddRecord ` is  a function.of physical table.  In ExchangePartition case, both non-partition table and physical table of partition table will go here right? There seems no good method to distingrish it.

mjonss: OK, let's keep it there. I do think the TableCommon.AddRecord could be used for both partitioned and non-partitioned tables, but that can be a follow up issue/PR, which is not necessary.

mjonss (approved):
LGTM

tiancaiamao (approved):

PR #46040, opened 2023-08-14
br: configure the httpclient for external storage

### LINKED ISSUE #46011: cannot reuse tcp connection for the external storage
## Bug Report

1. run br truncate

``` 
netstat | grep "TIME_WAIT"
0
```
```
netstat | grep "TIME_WAIT"
11111
```
master

Leavrth:

Issue Number: close #46011 

Problem Summary:
br use the default http client to connect the external storage, which doesn't have enough reuse-connection-pool for each host(the default value is 2). This leads to slow batch files processing and TCP connections exceeding the system configuration limit.

set the `MaxIdleConnsPerHost` as the same as `MaxIdleConns`.
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

before configuration
img_v2_6f0edf8a-a936-4ed6-aa18-77f5e663fa7g

after configuration
fd011f9c-050a-4c66-9a2a-57bfab0ab882

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/storage/storage.go
+++ b/br/pkg/storage/storage.go
@@ -171,9 +171,6 @@ func New(ctx context.Context, backend *backuppb.StorageBackend, opts *ExternalSt
 	if opts == nil {
 		opts = &ExternalStorageOptions{}
 	}
-	if opts.HTTPClient == nil {
-		opts.HTTPClient = GetDefaultHttpClient(DefaultRequestConcurrency)
-	}
 	switch backend := backend.Backend.(type) {
 	case *backuppb.StorageBackend_Local:
 		if backend.Local == nil {
--- a/br/pkg/storage/storage.go
+++ b/br/pkg/storage/storage.go
@@ -172,7 +172,7 @@ func New(ctx context.Context, backend *backuppb.StorageBackend, opts *ExternalSt
 		opts = &ExternalStorageOptions{}
 	}
 	if opts.HTTPClient == nil {
-		opts.HTTPClient = GetDefaultHttpClient(DefaultRequestConcurrency)
+		//opts.HTTPClient = GetDefaultHttpClient(DefaultRequestConcurrency)
 	}
 	switch backend := backend.Backend.(type) {
 	case *backuppb.StorageBackend_Local:
--- a/br/pkg/storage/storage.go
+++ b/br/pkg/storage/storage.go
@@ -172,7 +172,7 @@ func New(ctx context.Context, backend *backuppb.StorageBackend, opts *ExternalSt
 		opts = &ExternalStorageOptions{}
 	}
 	if opts.HTTPClient == nil {
-		//opts.HTTPClient = GetDefaultHttpClient(DefaultRequestConcurrency)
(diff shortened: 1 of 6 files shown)


br/pkg/storage/storage.go
```diff
@@ -198,3 +205,29 @@ func New(ctx context.Context, backend *backuppb.StorageBackend, opts *ExternalSt
...
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+			DialContext: defaultTransportDialContext(&net.Dialer{
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+			}),
+			ForceAttemptHTTP2:     true,
+			MaxIdleConns:          int(concurrency),
+			MaxIdleConnsPerHost:   int(concurrency),
+			IdleConnTimeout:       90 * time.Second,
```
BornChanger, line 228: Can we avoid to hard code them but copy attributes from http.DefaultTransport

Leavrth: the `http.DefaultTransport` is a interface type( interface `RoundTripper` ), so we only can force convert back to Transport and ignore the possible error.
And it has some fields, such as sync.Mutex and map, I'm worried that the copy leaves behind the risk of golang updating.

YuJuncen (approved):

br/pkg/storage/storage.go
```diff
@@ -198,3 +205,29 @@ func New(ctx context.Context, backend *backuppb.StorageBackend, opts *ExternalSt
 		return nil, errors.Annotatef(berrors.ErrStorageInvalidConfig, "storage %T is not supported yet", backend)
 	}
 }
+
+// copy from `http.defaultTransportDialContext`
```
lance6716, line 209: another way is use go:linkname like https://github.com/pingcap/tidb/blob/5a305400a94464d226f95503c6a3068ef881e69d/br/pkg/lightning/backend/local/local_test.go#L752-L753

br/pkg/storage/storage.go
lance6716: I also like @BornChanger 's idea. We can add an unit test to detect the future implementation change of interface `RoundTripper` , and use `Transport.Clone` to correctly handle its private fields.

br/pkg/storage/storage_test.go
```diff
@@ -0,0 +1,26 @@
...
+	require.True(t, ok)
+	require.True(t, transport.MaxConnsPerHost == 0)
+	require.True(t, transport.MaxIdleConns > 0)
+}
+
+func TestDefaultHttpClient(t *testing.T) {
+	var concurrency uint = 128
+	transport, ok := storage.GetDefaultHttpClient(concurrency).Transport.(*http.Transport)
+	require.True(t, ok)
+	require.Equal(t, transport.MaxIdleConnsPerHost, int(concurrency))
```
lance6716, line 24: nit: except for `t`, the first argument is "expected" and the second one is "got", so we should swap them

lance6716 (commented):
lgtm

Issue #46100, opened 2023-08-15
AUTO_INCREMENT increase too large after importing NONCLUSTERED AUTO_ID_CACHE=1 table by lightning


mayjiang0203:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

<!-- a step by step guide for reproducing the bug. -->
Following is schema that will import by lighting
```
cat smalldb.t1-schema.sql
CREATE TABLE `t1` (
    `cacheid` int(11) NOT NULL AUTO_INCREMENT,
    `rk` varchar(16) NOT NULL,
    `cfq` varchar(10) NOT NULL,
    `ts` bigint(20) NOT NULL,
    `value` varchar(2048) DEFAULT NULL,
    PRIMARY KEY(`cacheid`) NONCLUSTERED)ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin AUTO_ID_CACHE 1;
```
After import finished, hit following error when try to insert new rows.
```
mysql> show table t1 next_row_id;
+---------+------------+-------------+--------------------+----------------+
| DB_NAME | TABLE_NAME | COLUMN_NAME | NEXT_GLOBAL_ROW_ID | ID_TYPE        |
+---------+------------+-------------+--------------------+----------------+
| smalldb | t1         | _tidb_rowid |                  1 | _TIDB_ROWID    |
| smalldb | t1         | _tidb_rowid |          577898371 | AUTO_INCREMENT |
+---------+------------+-------------+--------------------+----------------+
2 rows in set (0.21 sec)

mysql> select count(*) from t1;
+----------+
| count(*) |
+----------+
|  3960000 |
+----------+
1 row in set (0.78 sec)

mysql> show create table t1;
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                                                                                                                                                                                                              |
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| t1    | CREATE TABLE `t1` (
  `cacheid` int(11) NOT NULL AUTO_INCREMENT,
  `rk` varchar(16) NOT NULL,
  `cfq` varchar(10) NOT NULL,
[...]

PR #46127, opened 2023-08-16
*: fix the duplicate entry error when using BR to restore a NONCLUSTERED AUTO_ID_CACHE=1 table

### LINKED ISSUE #46093: Duplicate entry when using BR to restore a NONCLUSTERED AUTO_ID_CACHE=1 table
## Bug Report

1. prepare

```sql
use test;
create table t (a int primary key nonclustered auto_increment, b int) auto_id_cache = 1;
insert into t (b) values (1), (2), (3);
show table t next_row_id;
-- _TIDB_ROWID = 30001, AUTO_INCREMENT = 4
```

2. backup & restore

```sql
backup table t to 'local:///tmp/br';
drop table t;
restore table t from 'local:///tmp/br';
```

(explained below)

3. the next_row_id are wrong
```sql
show table t next_row_id;
-- _TIDB_ROWID = 1, AUTO_INCREMENT = 4001
```

4. try insert more rows, result in duplicate keys of `_tidb_rowid`
```sql
insert into t (b) values (4), (5), (6);
-- ERROR 1062 (23000): Duplicate entry '1' for key 't.PRIMARY'
```

5. after `_tidb_rowid` increases to the available zone, we see the primary key's auto-inc values have a gap
```sql
insert into t (b) values (7), (8), (9);
select * from t;
```
```
+------+------+
| a    | b    |
+------+------+
|    1 |    1 |
|    2 |    2 |
|    3 |    3 |
| 4004 |    7 |
| 4005 |    8 |
| 4006 |    9 |
+------+------+
```

```
Release Version: v7.1.1
Edition: Community
Git Commit Hash: cf441574864be63938524e7dfcf7cc659edc3dd8
Git Branch: heads/refs/tags/v7.1.1
UTC Build Time: 2023-07-19 10:20:53
GoVersion: go1.20.6
Race Enabled: false
TiKV Min Version: 6.2.0-alpha
Store: tikv
```

tiancaiamao:

Issue Number: close #46093

Problem Summary:

For a NONCLUSTERED table with AUTO_ID_CACHE=1, like the case in #46093,
there are both `auto_increment_id` and `_tidb_rowid`.

The `model.TableInfo` only get one field `AutoIncID`, the information is not enough to recover the table correctly.
So a `AutoIncIDExtra` field is added for that case. We can record  `_tidb_rowid` in `AutoIncIDExtra` and recover  the `_tidb_rowid` from it.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

Test the case follow the steps in  #46093 manually.

----

I'm trying to add some test but find it difficult to  do that in a unit test.
For example, I tried to add it in the `tests/realtikvtest/brietest` but found that it does not start the server layer so the auto id service does not run and I can't test the AUTO_ID_CACHE=1
I tried to add it in the `server/tests` where a server is started, but it use unistore and `br recover table ...` can not run with unistore.

```
 func TestIssue46093(t *testing.T) {
        ts := createTidbTestSuite(t)

        ts.RunTestsOnNewDB(t, nil, "br", func(dbt *testkit.DBTestKit) {
                dbt.MustExec("drop table if exists t")
                dbt.MustExec("create table t (a int primary key nonclustered auto_increment, b int) auto_id_cache = 1;")
                dbt.MustExec("insert into t (b) values (1), (2), (3);")
                rows := dbt.MustQuery("show table t next_row_id;")
                ts.CheckRows(t, rows, "br t _tidb_rowid 30001 _TIDB_ROWID",
                        "br t _tidb_rowid 4 AUTO_INCREMENT")
                dbt.MustExec("backup table t to 'local:///tmp/br';")
                dbt.MustExec("drop table t;")
                dbt.MustExec("restore table t from 'local:///tmp/br';")
                rows = dbt.MustQuery("show table t next_row_id;")
                ts.CheckRows(t, rows, "1 4001")
                dbt.MustExec("insert into t (b) values (4), (5), (6);")
                dbt.MustExec("insert into t (b) values (7), (8), (9);")
                rows = dbt.MustQuery("select * from t;")
                ts.CheckRows(t, rows, "1 1", "2 2", "3 3", "4004 7", "4005 8", "4006 9")
        })
}
```

- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


bb7133 (commented):
Please add some tests.

br/pkg/backup/client.go
```diff
@@ -633,6 +633,12 @@ func BuildBackupSchemas(
 			default:
 				if tableInfo.SepAutoInc() {
 					globalAutoID, err = autoIDAccess.IncrementID(tableInfo.Version).Get()
+					// For a nonclustered table with auto_increment column, both auto_increment_id and _tidb_rowid are required.
+					// See also https://github.com/pingcap/tidb/issues/46093
+					if rowID, err1 := autoIDAccess.RowID().Get(); err1 == nil {
+						rowID = rowID + 1
```
bb7133, line 639: Can you explain why we need `rowID +=1` here and `*tbInfo.AutoIncIDExtra-1` in `ddl_api.go`(line 2633)?

tiancaiamao: I follow the convention here https://github.com/pingcap/tidb/pull/46127/files/a6403bc4931d7ffa386cb06ea56f23c7dedb09fd#diff-e611f02c279107dbfa433418c809e99569c0b24c2ea92cc7aef9f7f41fbcc3b9L643

br/pkg/lightning/common/common.go
```diff
@@ -89,17 +89,26 @@ func getGlobalAutoIDAlloc(store kv.Storage, dbID int64, tblInfo *model.TableInfo
...
+	// If AUTO_ID_CACHE=1 is set (i.e. MySQL compatible AUTO_INCREMENT), the limitations do not hold anymore.
 	switch {
-	case hasRowID || hasAutoIncID:
+	case hasRowID:
 		return autoid.NewAllocator(store, dbID, tblInfo.ID, tblInfo.IsAutoIncColUnsigned(),
 			autoid.RowIDAllocType, noCache, tblVer), nil
+	case hasAutoIncID:
+		allocType := autoid.RowIDAllocType
+		if tblInfo.SepAutoInc() {
+			allocType = autoid.AutoIncrementType
```
lance6716, line 108: seems for lightning only the type is changed, and the allocated ID of import data is also changed. What's the new behaviour of this new type? and I'd ask if the data has non-continuous AUTO INCREMENT ID after import is it acceptable?

lance6716 (commented):
You can add integration test like this

https://github.com/pingcap/tidb/blob/0fb21c505805b7745c5be23e0608775c8eb3ab48/br/tests/lightning_tikv_multi_rocksdb/run.sh#L21-L22

restart_service is defined in https://github.com/pingcap/tidb/blob/master/br/tests/_utils/run_services

ddl/ddl_api.go
```diff
@@ -2628,6 +2628,12 @@ func (d *ddl) createTableWithInfoPost(
 			return errors.Trace(err)
 		}
 	}
+	// For issue https://github.com/pingcap/tidb/issues/46093
+	if tbInfo.AutoIncIDExtra != nil && *tbInfo.AutoIncIDExtra != 0 {
```
kennytm, line 2632: given the `*tbInfo.AutoIncIDExtra != 0` condition i think making `AutoIncIDExtra` an `int64` rather than `*int64` is good enough.

we could add a `omitempty` to prevent it from showing up in JSON on zero

br/pkg/lightning/common/common.go
tiancaiamao: > What's the new behaviour of this new type?

For lightning, this change is not related to the fix of #46093 
I make this change just because this is the correct behaviour, when `SepAutoInc() == true`, the allocator should use the autoid's meta key rather than rowid's meta key.


> and I'd ask if the data has non-continuous AUTO INCREMENT ID after import is it acceptable?

This is because the cached ids in the memory of auto id service is dropped, BR get the current id by reading the auto id's meta key.
If the auto id service panic unexpecitally, we also lose some cached ids, and there would be holes in the id sequence. I think that's acceptable.

ddl/ddl_api.go
tiancaiamao: Good suggestion, I'll take a try.

br/pkg/lightning/common/common.go
```diff
@@ -89,17 +89,26 @@ func getGlobalAutoIDAlloc(store kv.Storage, dbID int64, tblInfo *model.TableInfo
...
 	//    See https://github.com/pingcap/tidb/issues/982.
 	// 2. Auto random column must be a clustered primary key. That is to say,
 	//    there is no implicit row ID for tables with auto random column.
 	// 3. There is at most one auto column in a table.
 	// Therefore, we assume there is only one auto column in a table and use RowID allocator if possible.
+	//
+	// If AUTO_ID_CACHE=1 is set (i.e. MySQL compatible AUTO_INCREMENT), the limitations do not hold anymore.
 	switch {
-	case hasRowID || hasAutoIncID:
+	case hasRowID:
```
D3Hunter, line 102: i have a unfinished pr to fix this https://github.com/pingcap/tidb/issues/46100 on lightning part

this won't work as lightning don't rebase this alloctor when running in single instance

tiancaiamao: Get it, I'll reset the change here.

br/pkg/backup/client.go
```diff
@@ -633,6 +633,11 @@ func BuildBackupSchemas(
 			default:
 				if tableInfo.SepAutoInc() {
 					globalAutoID, err = autoIDAccess.IncrementID(tableInfo.Version).Get()
+					// For a nonclustered table with auto_increment column, both auto_increment_id and _tidb_rowid are required.
+					// See also https://github.com/pingcap/tidb/issues/46093
+					if rowID, err1 := autoIDAccess.RowID().Get(); err1 == nil {
```
3pointer, line 638: need handle error here

tiancaiamao: print a log?
return the error to the caller seems no suitable, because I'm not 100% sure the rowid meta key exist.
If the rowid does not exist, then the error is expected.

tiancaiamao: Handling of err1 != nil updated. @3pointer

kennytm:
could we at least add an integration test (into `br/tests/br_*`)?

kennytm (approved):

br/tests/br_autoid/run.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/sh
```
D3Hunter, line 1: this file need add to run_group.sh

tiancaiamao: Done

br/tests/br_autoid/run.sh
```diff
@@ -0,0 +1,51 @@
...
+run_sql "create database if not exists ${DB}"
+run_sql "create table $DB.issue46093 (a int primary key nonclustered auto_increment, b int) auto_id_cache = 1;"
+run_sql "insert into $DB.issue46093 (b) values (1), (2), (3);"
+run_sql "show table $DB.issue46093 next_row_id;"
+check_contains "NEXT_GLOBAL_ROW_ID: 30001"
+check_contains "NEXT_GLOBAL_ROW_ID: 4"
+
+run_sql "backup table $DB.issue46093 to 'local:///tmp/br'";
+run_sql "drop table $DB.issue46093;"
+run_sql "restore table $DB.issue46093 from 'local:///tmp/br';"
```
Leavrth, line 29: please backup to `local://$TEST_DIR/$DB` to avoid duplicate backup paths in different tests.

tiancaiamao: Done

PR #46232, opened 2023-08-18
CI: disable codecov's reports in commit status and github cheks

wuhuizuo:
Reverts pingcap/tidb#46180

changes:
- disable report git commit status.
- disable report github checks.
- only keep comments.
- recover `carryforward` for flags.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -8,14 +8,22 @@ coverage:
   round: down
   range: "65...90"
 
+  # status:
+  #   project:
+  #     default:
+  #       threshold: 0.2 #Allow the coverage to drop by threshold%, and posting a success status.
+  #   patch:
+  #     default:
+  #       target: 0% # trial operation
+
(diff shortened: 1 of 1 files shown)


.codecov.yml
```diff
@@ -8,15 +8,23 @@ coverage:
...
   range: "65...90"
 
+  # status:
+  #   project:
+  #     default:
+  #       threshold: 0.2 #Allow the coverage to drop by threshold%, and posting a success status.
+  #   patch:
+  #     default:
+  #       target: 0% # trial operation
+
```
hi-rustin, line 18:
```suggestion
```

hi-rustin (approved):

.codecov.yml
wuhuizuo: I will try to persuade BR members to run the integration job before merge, it will be recovered when they agreed it.

PR #46235, opened 2023-08-18
planner: refactor `Next` function of `AnalyzeExec`

hi-rustin:

Issue Number: None

Problem Summary:

- Split the `Next` function into small functions
- Used minimal channel size: `concurrency = min(len(tasks), concurrency)`
- Called `e.Ctx().GetSessionVars()` only once
- Used `strings.Join` to generate the warning message

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/executor/analyze.go
+++ b/executor/analyze.go
@@ -179,7 +179,7 @@ func filterAndCollectTasks(tasks []*analyzeTask, statsHandle *handle.Handle, inf
 	for _, task := range tasks {
 		tableID := getTableIDFromTask(task)
 		isLocked := statsHandle.IsTableLocked(tableID)
-		if isLocked {
+		if !isLocked {
 			filteredTasks = append(filteredTasks, task)
 		}
 		if _, ok := tids[tableID]; !ok {
--- a/executor/analyze.go
+++ b/executor/analyze.go
@@ -99,7 +99,7 @@ func (e *AnalyzeExec) Next(ctx context.Context, _ *chunk.Chunk) error {
 		return nil
 	}
 
-	// Get the max number of goroutines for parallel execution.
+	// Get the min number of goroutines for parallel execution.
 	concurrency, err := getBuildStatsConcurrency(e.Ctx())
 	if err != nil {
 		return err
--- a/executor/analyze.go
+++ b/executor/analyze.go
@@ -183,6 +183,7 @@ func filterAndCollectTasks(tasks []*analyzeTask, statsHandle *handle.Handle, inf
 			if statsHandle.IsTableLocked(tableID) {
 				tbl, ok := infoSchema.TableByID(tableID)
 				if !ok {
+					logutil.BgLogger().Warn("Unknown table ID in analyze task", zap.Int64("tid", tableID))
 					continue
 				}
 				skippedTables = append(skippedTables, tbl.Meta().Name.L)
--- a/executor/analyze.go
+++ b/executor/analyze.go
@@ -171,10 +171,12 @@ func (e *AnalyzeExec) Next(ctx context.Context, _ *chunk.Chunk) error {
 
 // filterAndCollectTasks filters the tasks that are not locked and collects the table IDs.
 func filterAndCollectTasks(tasks []*analyzeTask, statsHandle *handle.Handle, infoSchema infoschema.InfoSchema) ([]*analyzeTask, uint, []string) {
-	var filteredTasks []*analyzeTask
-	tids := make(map[int64]struct{})
-	var skippedTables []string
-	needAnalyzeTableCnt := uint(0)
+	var (
+		filteredTasks       []*analyzeTask
+		skippedTables       []string
+		needAnalyzeTableCnt uint
+		tids                = make(map[int64]struct{})
+	)
 
 	for _, task := range tasks {
 		tableID := getTableIDFromTask(task)
--- a/executor/analyze.go
+++ b/executor/analyze.go
@@ -198,13 +198,13 @@ func filterAndCollectTasks(tasks []*analyzeTask, statsHandle *handle.Handle, inf
 func warnLockedTableMsg(sessionVars *variable.SessionVars, tids map[int64]struct{}, skippedTables []string) {
 	if len(skippedTables) > 0 {
 		tables := strings.Join(skippedTables, ", ")
-		msg := "skip analyze locked table"
+		msg := "skip analyze locked table: %s"
 		if len(tids) > 1 {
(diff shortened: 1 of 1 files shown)


executor/analyze.go
```diff
@@ -198,14 +154,77 @@ func (e *AnalyzeExec) Next(ctx context.Context, _ *chunk.Chunk) error {
...
+	for _, task := range tasks {
+		tableID := getTableIDFromTask(task)
+		if !statsHandle.IsTableLocked(tableID) {
+			filteredTasks = append(filteredTasks, task)
+		}
+		if _, ok := tids[tableID]; !ok {
+			if statsHandle.IsTableLocked(tableID) {
+				tbl, ok := infoSchema.TableByID(tableID)
+				if !ok {
+					continue
```
XuHuaiyu, line 186: Why continue here, it's `return nil` before

hi-rustin: Good catch. This was a mistake.

qw4990 (approved):

hi-rustin:
This PR is ready for review. I want to add more tests for these small functions. But I need to move the analyze code into the analyze package first. Otherwise I have to expose a lot of private fields from a bunch of structures to use it in the `executor_test` package. I tried to move all tests from the `executor_test` package to the `executor` package, but I got an `import cycle` error.
So let's move this pull request forward. I will move the code and add tests in my next PR.

executor/analyze.go
```diff
@@ -198,14 +154,78 @@ func (e *AnalyzeExec) Next(ctx context.Context, _ *chunk.Chunk) error {
...
+	for _, task := range tasks {
+		tableID := getTableIDFromTask(task)
+		if !statsHandle.IsTableLocked(tableID) {
+			filteredTasks = append(filteredTasks, task)
+		}
+		if _, ok := tids[tableID]; !ok {
+			if statsHandle.IsTableLocked(tableID) {
+				tbl, ok := infoSchema.TableByID(tableID)
+				if !ok {
+					logutil.BgLogger().Warn("Unknown table ID in analyze task", zap.Int64("tid", tableID))
```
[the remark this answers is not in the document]
hi-rustin: This is already a warning. Do you mean just note it with a comment?

executor/analyze.go
```diff
@@ -88,79 +88,31 @@ const (
...
-		if len(tids) > 1 {
-			if len(tids) > len(skippedTables) {
-				msg = "skip analyze locked tables: " + tables + ", other tables will be analyzed"
-			} else {
-				msg = "skip analyze locked tables: " + tables
-			}
-		} else {
-			msg = "skip analyze locked table: " + tables
-		}
+	infoSchema := e.Ctx().GetInfoSchema().(infoschema.InfoSchema)
```
hawkingrei, line 91: use ```sessiontxn.GetTxnManager(ctx).GetTxnInfoSchema``` to replace it. because this method is deprecated.

hi-rustin: Addressed. Thanks!

executor/analyze.go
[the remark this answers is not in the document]
hi-rustin: I updated it as a debug log.

hawkingrei (approved):

hawkingrei:
@XuHuaiyu PTAL

executor/analyze.go
time-and-fate: It's almost impossible to meet such cases in normal scenarios. Even if the table is dropped during the execution of this statement, we should still be able to get the table from the `infoschema` here.

I agree that the previous `return nil` is wrong.
Personally, I think DEBUG level logging is basically useless since the default is INFO level, and I would suggest that we directly record it like `table id: 123` to the `skippedTables`.

executor/analyze.go
```diff
@@ -198,14 +155,79 @@ func (e *AnalyzeExec) Next(ctx context.Context, _ *chunk.Chunk) error {
...
+// warnLockedTableMsg warns the locked table IDs.
+func warnLockedTableMsg(sessionVars *variable.SessionVars, tids map[int64]struct{}, skippedTables []string) {
+	if len(skippedTables) > 0 {
+		tables := strings.Join(skippedTables, ", ")
+		msg := "skip analyze locked table: %s"
+		if len(tids) > 1 {
+			if len(tids) > len(skippedTables) {
+				msg = "skip analyze locked tables: %s, other tables will be analyzed"
+			}
+		}
```
time-and-fate, line 209: They can be merged into one `if` block.

executor/analyze.go
hi-rustin: The reason I didn't generate a statement warning here is that we can't get the table, so if we use TableID here then it would be too confusing for users.

hi-rustin: > It's almost impossible to meet such cases in normal scenarios. Even if the table is dropped during the execution of this statement, we should still be able to get the table from the `infoschema` here.

So let's change it to a warning. That would be helpful.

time-and-fate: LGTM.

time-and-fate (approved):

executor/analyze.go
```diff
@@ -198,14 +155,88 @@ func (e *AnalyzeExec) Next(ctx context.Context, _ *chunk.Chunk) error {
...
+		if _, ok := tids[tableID]; !ok {
+			if isLocked {
+				tbl, ok := infoSchema.TableByID(tableID)
+				if !ok {
+					logutil.BgLogger().Warn("Unknown table ID in analyze task", zap.Int64("tid", tableID))
+					// Ignore this table because it may have been dropped.
+					continue
+				}
+				skippedTables = append(skippedTables, tbl.Meta().Name.L)
```
time-and-fate, line 195:
```suggestion
				if !ok {
					logutil.BgLogger().Warn("Unknown table ID in analyze task", zap.Int64("tid", tableID))
				} else {
					skippedTables = append(skippedTables, tbl.Meta().Name.L)
				}
```
Looks like we shouldn't `continue` here because we still need to run `tids[tableID] = struct{}{}` below.

And the comment can be removed because dropping the table actually won't cause this.

hi-rustin: Addressed in [`4b587e8` (#46235)](https://github.com/pingcap/tidb/pull/46235/commits/4b587e8d07cdb82787c883aacaf0407fbb2ac54b)

Issue #46203, opened 2023-08-17
refine lightning retry on region scatter


mittalrishabh:
## Bug Report

Lightning split a region into few regions and call scatter regions on all the regions once before it back off. In my view it should back off and retry scattering the same region if it is failed. This logic is similar to tikv client scatter logic. 

https://github.com/tikv/client-go/blob/a0ac170698524cbf20a869cb3471e969608d599a/tikv/split_region.go#L198

Advantage of this approach is
1. We don't send scatter unnecessarily to PD if it is failing. 
2. Each scatter operation creation increase the scatter count on a store even though it is cancelled later. It might cause some skewness.
 
Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
1. Set store limit to 16
4. Run lightning in physical import mode for large data set (100G)
5. You will see lot of scatter operators cancelled. And you might see skewed regions.

### 2. What did you expect to see? (Required)
I expect to see balanced regions and less number of operators cancelled

### 3. What did you see instead (Required)

I found regions skewed and scatter operations cancelled

### 4. What is your TiDB version? (Required)
6.5.1

lance6716:
> retry scattering the same region if it is failed

lightning should already retry scattering. in `waitForScatterRegions` -> `checkRegionScatteredOrReScatter`, lightning will retry scattering when status is canceled

https://github.com/pingcap/tidb/blob/4084b077d615f9dc0a41cf2e30bc6e1a02332df2/br/pkg/lightning/backend/local/localhelper.go#L504-L513

> It might cause some skewness.

we should have fixed it in master and v7.1.1. This is the issue for skewness https://github.com/pingcap/tidb/issues/43510

mittalrishabh:
I have created this PR. https://github.com/pingcap/tidb/pull/46245

It is not a final PR. But it will give you the clear idea what i am trying to do

D3Hunter:
region skew part is duplicate with https://github.com/pingcap/tidb/issues/43510

D3Hunter:
change to `enhancement` as region skew part has already fixed in https://github.com/pingcap/tidb/issues/43510

this issue is only for refine retry logic of region scatter

D3Hunter:
> I have created this PR. #46245
> 
> It is not a final PR. But it will give you the clear idea what i am trying to do

#46245 will cause long tail issue when parallel importing using multiple instances, some instance might be very slow at split&scatter
[...]

Issue #43510, opened 2023-05-04
for large import task, lightning needs to improve region split/scatter/check


lance6716:
## Enhancement

we meet a problem that for a 10 lightning parallel import task, region actions are slow so lightning exits with error.

there are two cases at least

- scattering region timed out and lightning continues to import. And due to region are imbalanced some TiKV nodes run out of disk
- checking region continuous timed out with message "region endKey not equal to next region startKey" and lightning exited

PR #46245, opened 2023-08-18
pkg/lightning: retry individual scatter region

mittalrishabh:

Issue Number: ref #46203

Problem Summary:
Lightning try to scatter all the regions and at the end it backoff and retry it again. It causes few issues
1. If we are hiting a store limit, than further scatter regions cause unnecessarily load on PD. 
2. Regions can be skewed as stores are allocated based on number of scatter operation scheduled. It doesn't take into number of scatter operations cancelled because of store limit.
 
I am proposing the change similar to tikv go client which doesn't retry and back off after scattering all the regions. 

https://github.com/tikv/client-go/blob/a0ac170698524cbf20a869cb3471e969608d599a/tikv/split_region.go#L198

### Check List

Tests 

I tested the similar logic in 6.5 and 6.6 by using the below steps
1. Set store limit to 16
2. Parallel import and verify that regions are not skewed

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/lightning/backend/local/localhelper.go
+++ b/br/pkg/lightning/backend/local/localhelper.go
@@ -399,6 +399,7 @@ func (local *Backend) SplitAndScatterRegionByRanges(
 	return nil
 }
 
+// it scatter region and retry if it fails. It retuns error if can not scatter after max_retry
 func (local *Backend) ScatterRegion(ctx context.Context, regionInfo *split.RegionInfo) error {
 	backoffer := split.NewWaitRegionOnlineBackoffer().(*split.WaitRegionOnlineBackoffer)
 	_ = utils.WithRetry(ctx, func() error {
@@ -435,9 +436,10 @@ func (local *Backend) BatchSplitRegions(
 			// Wait for a while until the regions successfully splits.
 			ok, err2 := local.hasRegion(ctx, region.Region.Id)
 			if !ok || err2 != nil {
-				failedErr = err2
-				if failedErr == nil {
+				if err2 == nil {
 					failedErr = errors.Errorf("region %d not found", region.Region.Id)
+				} else {
+					failedErr = err2
 				}
 				retryRegions = append(retryRegions, region)
 				continue
--- a/br/pkg/lightning/backend/local/localhelper.go
+++ b/br/pkg/lightning/backend/local/localhelper.go
@@ -404,6 +404,9 @@ func (local *Backend) ScatterRegion(ctx context.Context, regionInfo *split.Regio
 	_ = utils.WithRetry(ctx, func() error {
 		return (local.splitCli.ScatterRegion(ctx, regionInfo))
 	}, backoffer)
+	if ctx.Err() != nil {
+		log.FromContext(ctx).Warn("scatter region failed", zap.Error(ctx.Err()))
+	}
 	return ctx.Err()
 }
 
@@ -423,12 +426,12 @@ func (local *Backend) BatchSplitRegions(
 		return nil, nil, errors.Annotatef(err, "batch split regions failed")
 	}
 	var failedErr error
-	scatterRegions := newRegions
+	splitRegions := newRegions
 	// wait for regions to be split
 	backoffer := split.NewWaitRegionOnlineBackoffer().(*split.WaitRegionOnlineBackoffer)
 	_ = utils.WithRetry(ctx, func() error {
 		retryRegions := make([]*split.RegionInfo, 0)
-		for _, region := range scatterRegions {
+		for _, region := range splitRegions {
 			// Wait for a while until the regions successfully splits.
 			ok, err2 := local.hasRegion(ctx, region.Region.Id)
 			if !ok || err2 != nil {
@@ -445,16 +448,16 @@ func (local *Backend) BatchSplitRegions(
 		}
 		// if the number of becomes smaller, we can infer TiKV side really
 		// made some progress so don't increase the retry times.
-		if len(retryRegions) < len(scatterRegions) {
+		if len(retryRegions) < len(splitRegions) {
 			backoffer.Stat.ReduceRetry()
 		}
 		log.FromContext(ctx).Warn("split region failed", zap.Int("regionCount", len(newRegions)),
 			zap.Int("failedCount", len(retryRegions)), zap.Error(failedErr))
-		scatterRegions = retryRegions
+		splitRegions = retryRegions
 		// although it's not PDBatchScanRegion, WaitRegionOnlineBackoffer will only
 		// check this error class so we simply reuse it. Will refine WaitRegionOnlineBackoffer
 		// later
-		failedErr = errors.Annotatef(berrors.ErrPDBatchScanRegion, "scatter region failed")
+		failedErr = errors.Annotatef(berrors.ErrPDBatchScanRegion, "split region failed")
 		return failedErr
 	}, backoffer)
 
@@ -463,10 +466,11 @@ func (local *Backend) BatchSplitRegions(
 	}
 
 	// scatter regions
+	scatterRegions := newRegions
 	for _, region := range scatterRegions {
 		err = local.ScatterRegion(ctx, region)
 		if failedErr == nil {
-			failedErr = err
+			failedErr = errors.Annotatef(berrors.ErrPDBatchScanRegion, "scatter region failed")
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


mittalrishabh:
It is not a final PR. I will clean it up more. But would like to know your opinion before that.

br/pkg/lightning/backend/local/localhelper.go
```diff
@@ -399,6 +399,14 @@ func (local *Backend) SplitAndScatterRegionByRanges(
 	return nil
 }
 
+func (local *Backend) ScatterRegion(ctx context.Context, regionInfo *split.RegionInfo) error {
+	backoffer := split.NewWaitRegionOnlineBackoffer().(*split.WaitRegionOnlineBackoffer)
+	_ = utils.WithRetry(ctx, func() error {
+		return (local.splitCli.ScatterRegion(ctx, regionInfo))
```
lance6716, line 405: Oh I understand now. We should retry for multiple times when `ScatterRegion` fails. Will take a look later

mittalrishabh: did you get chance to look into it

br/pkg/lightning/backend/local/localhelper.go
```diff
@@ -430,10 +439,6 @@ func (local *Backend) BatchSplitRegions(
 				retryRegions = append(retryRegions, region)
 				continue
 			}
-			if err = local.splitCli.ScatterRegion(ctx, region); err != nil {
```
D3Hunter, line 433: we have retry scatter after checking `hasRegion` previously, now it's done after `hasRegion` check of all regions. how this affect region skewness?

and i think change from `try scatter next region` to `try scatter same region` after some scatter failed due to store limit, does NOT reduce unnecessarily load on PD

mittalrishabh: Why do we want to scatter other regions in the list If we hit a store limit on one scatter operation. In my view it should back off if one scatter operation fails

mittalrishabh: If scatter operation fails It is adding the regions into retryRegions currently and keep doing scatter of other regions in the list.  It can cause lot of cancel scatter operation in PD.

D3Hunter: > Why do we want to scatter other regions in the list If we hit a store limit on one scatter operation.

i get it. but still not get why this affect region skewness.

as in `It doesn't take into number of scatter operations cancelled because of store limit.`, those cancelled operation will be retried and finally became be scheduled.

mittalrishabh: There are two problems

In 6.5 , with parallel import and one region being split into 1000 regions , it hit the retry limit and failing to the scatter operations. It increase the retry limit only after scattering all the regions in the list and the retry limit is set to only 5.

PD scatter the regions on a node which has the least scatter operations scheduled. It doesn't take cancelled scatter regions into account. Lets say you there are 5 tikv node and you send 6 scatter operations than 1 and 5 can get scheduled to same node even though 2 to 4 scatter operations are cancelled.

I understand that master has some fixes like region-check-backoff-limit which is set to 1800 and it also ncreases the retry limit more conservatively which might fix some cases.

D3Hunter: > In 6.5

yes, it's a [known issue](https://github.com/pingcap/tidb/issues/43510), which might cause data skew on large data import, it's fixed on master/7.1.x, but not on to 6.5.x(as the pr introduces new parameters)

> However i think that for a complete solution , retry limit should be applied to individual regions rather than the whole list.

lgtm. now i know this pr only enhance this part, it's not related to region skewness

mittalrishabh: can you review this PR

br/pkg/lightning/backend/local/localhelper.go
D3Hunter: plz add a warn log on err, it helps diagnose, and the err need to be wrapped for now, see

https://github.com/pingcap/tidb/blob/64a2e1c20a0666f8f8c4ae46330dedcc406a5d47/br/pkg/lightning/backend/local/localhelper.go#L421-L425

mittalrishabh: fixed.

hawkingrei:
@mittalrishabh Please fix the linter.

```
br/pkg/lightning/backend/local/localhelper.go:443:3: ineffectual assignment to err (ineffassign)

br/pkg/lightning/backend/local/localhelper.go:372:1: exported: comment on exported method Backend.ScatterRegion should be of the form "ScatterRegion ..." (revive)

```

br/pkg/lightning/backend/local/localhelper.go
```diff
@@ -415,48 +427,58 @@ func (local *Backend) BatchSplitRegions(
...
+		return region, newRegions, ctx.Err()
+	}
+
+	// scatter regions
+	failedErr = nil
+	scatterRegions := newRegions
+	for _, region := range scatterRegions {
+		err = local.ScatterRegion(ctx, region)
+		if err != nil && failedErr == nil {
+			failedErr = errors.Annotatef(berrors.ErrPDBatchScanRegion, "scatter region failed")
```
D3Hunter, line 476: this should be done inside scope of `split.NewWaitRegionOnlineBackoffer().(*split.WaitRegionOnlineBackoffer)`

br/pkg/lightning/backend/local/localhelper.go
```diff
@@ -399,6 +399,18 @@ func (local *Backend) SplitAndScatterRegionByRanges(
 	return nil
 }
 
+// ScatterRegion scatter the regions and retry if it fails. It returns error if can not scatter after max_retry.
+func (local *Backend) ScatterRegion(ctx context.Context, regionInfo *split.RegionInfo) error {
+	backoffer := split.NewWaitRegionOnlineBackoffer().(*split.WaitRegionOnlineBackoffer)
+	_ = utils.WithRetry(ctx, func() error {
+		return (local.splitCli.ScatterRegion(ctx, regionInfo))
+	}, backoffer)
+	if ctx.Err() != nil {
```
D3Hunter, line 408: i think we need check err of `utils.WithRetry` here

br/pkg/lightning/backend/local/localhelper.go
mittalrishabh: i intentionally made it out of scope. Why we want to have two retry loops for scatter regions.

br/pkg/lightning/backend/local/localhelper.go
mittalrishabh: I don't understand. Won't ctx.err() be set if utils.WithRetry fails. Or are you asking to print error inside utils.withretry.

br/pkg/lightning/backend/local/localhelper.go
D3Hunter: see this, the backoffer only retry on `ErrPDBatchScanRegion` error, too bad😅

https://github.com/pingcap/tidb/blob/abec0c68e368d6dd5c076f725c42813a426f8ccc/br/pkg/restore/split/split.go#L221-L235

mittalrishabh: oh, should i change the Nextbackoffer. NextBackoff should retry on any error ideally. It should be the caller responsibility.

br/pkg/lightning/backend/local/localhelper.go
mittalrishabh: got it

br/pkg/lightning/backend/local/localhelper.go
D3Hunter: as commented in here, we plan to do it later, but if you have time, you can refactor this part too
https://github.com/pingcap/tidb/blob/64a2e1c20a0666f8f8c4ae46330dedcc406a5d47/br/pkg/lightning/backend/local/localhelper.go#L421-L425

br/pkg/lightning/backend/local/localhelper.go
```diff
@@ -415,48 +432,62 @@ func (local *Backend) BatchSplitRegions(
...
+		return region, newRegions, ctx.Err()
+	}
+
+	// scatter regions
+	failedErr = nil
+	scatterRegions := newRegions
+	for _, region := range scatterRegions {
+		err = local.ScatterRegion(ctx, region)
+		if err != nil && failedErr == nil {
+			failedErr = errors.Annotatef(berrors.ErrPDBatchScanRegion, "scatter region failed")
```
D3Hunter, line 485: not necessary not annotate the err here

D3Hunter (commented):
rest lgtm

lance6716 (approved):

Benjamin2037:
already in 7.5 @mittalrishabh

PR #46249, opened 2023-08-20
planner: remove outer join on unique keys when there is no column referenced

### LINKED ISSUE #46248: Planner fails to eliminate outer join on unique keys
```sql
CREATE TABLE `customer` (
  `C_CUSTKEY` bigint(20) NOT NULL,
  `C_NAME` varchar(25) NOT NULL,
  `C_ADDRESS` varchar(40) NOT NULL,
  `C_NATIONKEY` bigint(20) NOT NULL,
  `C_PHONE` char(15) NOT NULL,
  `C_ACCTBAL` decimal(15,2) NOT NULL,
  `C_MKTSEGMENT` char(10) NOT NULL,
  `C_COMMENT` varchar(117) NOT NULL,
  PRIMARY KEY (`C_CUSTKEY`) /*T![clustered_index] CLUSTERED */
);
CREATE TABLE `nation` (
  `N_NATIONKEY` bigint(20) NOT NULL,
  `N_NAME` char(25) NOT NULL,
  `N_REGIONKEY` bigint(20) NOT NULL,
  `N_COMMENT` varchar(152) DEFAULT NULL,
  PRIMARY KEY (`N_NATIONKEY`) /*T![clustered_index] CLUSTERED */
)

```
Join should be eliminated

```sql
explain select count(*) from customer a left join nation b on a.C_NATIONKEY=b.N_NATIONKEY;
+-------------------------------+----------+-----------+---------------+---------------------------------------------------------------------------------+
| id                            | estRows  | task      | access object | operator info                                                                   |
+-------------------------------+----------+-----------+---------------+---------------------------------------------------------------------------------+
| HashAgg_8                     | 1.00     | root      |               | funcs:count(1)->Column#13                                                       |
| └─HashJoin_19                 | 12500.00 | root      |               | left outer join, equal:[eq(test.customer.c_nationkey, test.nation.n_nationkey)] |
|   ├─TableReader_24(Build)     | 10000.00 | root      |               | data:TableFullScan_23                                                           |
|   │ └─TableFullScan_23        | 10000.00 | cop[tikv] | table:b       | keep order:false, stats:pseudo                                                  |
|   └─TableReader_22(Probe)     | 10000.00 | root      |               | data:TableFullScan_21                                                           |
|     └─TableFullScan_21        | 10000.00 | cop[tikv] | table:a       | keep order:false, stats:pseudo                                                  |
+-------------------------------+----------+-----------+---------------+---------------------------------------------------------------------------------+
6 rows in set (0.00 sec)
```

master

fixdb:

Issue Number: close #46248 

Problem Summary:
Previously we check all the referenced columns come from outer table or not, if yes, then we proceed to check whether we can eliminate outer join or not. However, when there is no column referenced from outer table and inner table, e.g. SELECT 1 FROM R LOJ S ON x=y, there is no column matching with outer table columns, we decided that we can't eliminate outer join, which is incorrect.

This patch fixes above issue by checking as long as there is no column referenced from inner table, we should continue to next step to see whether we can remove outer join or not, instead of bailing out quickly.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/planner/core/rule_join_elimination.go
+++ b/planner/core/rule_join_elimination.go
@@ -28,8 +28,8 @@ type outerJoinEliminator struct {
 }
 
 // tryToEliminateOuterJoin will eliminate outer join plan base on the following rules
-//  1. outer join elimination: For example left outer join, if the parent only use the
-//     columns from left table and the join key of right table(the inner table) is a unique
+//  1. outer join elimination: For example left outer join, if the parent doesn't use the
+//     columns from right table and the join key of right table(the inner table) is a unique
 //     key of the right table. the left outer join can be eliminated.
 //  2. outer join elimination with duplicate agnostic aggregate functions: For example left outer join.
 //     If the parent only use the columns from left table with 'distinct' label. The left outer join can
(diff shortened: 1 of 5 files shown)


planner/core/rule_join_elimination.go
```diff
@@ -51,12 +51,20 @@ func (o *outerJoinEliminator) tryToEliminateOuterJoin(p *LogicalJoin, aggCols []
 	for _, outerCol := range outerPlan.Schema().Columns {
 		outerUniqueIDs.Insert(outerCol.UniqueID)
 	}
-	matched := IsColsAllFromOuterTable(parentCols, outerUniqueIDs)
-	if !matched {
-		return p, false, nil
+
```
elsa0520, line 54: Please update function comment together.
"For example left outer join, if the parent don't use any right table columns and xxxx"

elsa0520 (approved):
LGTM

AilinKid (approved):

planner/core/rule_join_elimination.go
fixdb: done

chrysan (approved):

PR #46331, opened 2023-08-23
timer: support cron syntax for timer framework

### LINKED ISSUE #46330: Support cron syntax for timer framework

lcwangchao:

Issue Number: close #46330

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


timer/api/timer.go
```diff
@@ -211,3 +272,13 @@ func (r *TimerRecord) Clone() *TimerRecord {
 	cloned.TimerSpec = *r.TimerSpec.Clone()
 	return &cloned
 }
+
+// ValidateTimeZone validates the TimeZone field.
+func ValidateTimeZone(tz string) error {
+	if tz != "" && !strings.EqualFold(tz, TimeZoneTiDB) {
+		if _, ok := timeutil.ParseTimeZone(tz); !ok {
+			return errors.Errorf("Invalid timezone '%s'", tz)
```
bb7133, line 280: IMO it's better to use `ErrUnknownTimeZone`? `errors.Errorf` gives 'unknown error code' which we should always avoid when possible.

And, is there any specific reason we need to change the return value of `ParseTimeZone` from `error` to `bool`?

timer/api/timer.go
```diff
@@ -15,18 +15,26 @@
...
+	"strings"
 	"time"
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/tidb/parser/duration"
+	"github.com/pingcap/tidb/util/timeutil"
+	"github.com/robfig/cron/v3"
 )
 
+// TimeZoneTiDB indicates to use the tidb timezone.
```
bb7133, line 27: It seems that `TimeZoneTiDB` is the same with an empty string for 'timezone' fields in timer. IMO its semantic is not better than an empty string to me -_-.

How about just removing it and explicitly explain that in the client api? For example we can declare that 'if timezone is empty string, we use the (global) `time_zone` system variable'

lcwangchao: In the previous implement of timer, it uses `TiDB` as the default value and has already persisted into `mysql.tidb_timers`. 

But I think we can just support parsing `TiDB` value without  supporting write `TiDB` for compatitability

timer/api/timer.go
lcwangchao: `ParseTimeZone` returns bool is easy to write without introducing another `errors.go` file... But anyway it not very hard to do it, I'll update the PR.

bb7133 (approved):
LGTM

YangKeao (approved):
LGTM

easonn7:
/approve

there is no sysvar behavior change.

PR #46370, opened 2023-08-23
*: support the variable of `default_collation_for_utf8mb4` and corresponding behaviors

zimulala:

Issue Number: close https://github.com/pingcap/tidb/issues/46371

Problem Summary:
https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_default_collation_for_utf8mb4

> The default utf8mb4 collation is used in the following statements:
> 
> 1. [SHOW COLLATION](https://dev.mysql.com/doc/refman/8.0/en/show-collation.html) and [SHOW CHARACTER SET](https://dev.mysql.com/doc/refman/8.0/en/show-character-set.html).
> 
> 2. [CREATE TABLE](https://dev.mysql.com/doc/refman/8.0/en/create-table.html) and [ALTER TABLE](https://dev.mysql.com/doc/refman/8.0/en/alter-table.html) having a CHARACTER SET utf8mb4 clause without a COLLATION clause, either for the table character set or for a column character set.
> 
> 3. [CREATE DATABASE](https://dev.mysql.com/doc/refman/8.0/en/create-database.html) and [ALTER DATABASE](https://dev.mysql.com/doc/refman/8.0/en/alter-database.html) having a CHARACTER SET utf8mb4 clause without a COLLATION clause.
> 
> 4. Any statement containing a string literal of the form _utf8mb4'some text' without a COLLATE clause.

* Support the variable of `default_collation_for_utf8mb4 `
* Support show collation and show character set
* Support `create/alter table` behaviors
* Support `create/alter database` behaviors
* This PR doesn't support item 4.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


ddl/ddl_api.go
```diff
@@ -775,10 +784,38 @@ func buildColumnsAndConstraints(
...
+	}
+	defaultCollation, err := sessVars.GetSessionOrGlobalSystemVar(context.Background(), variable.DefaultCollationForUTF8MB4)
+	if err != nil {
+		return "", err
+	}
+	return defaultCollation, nil
+}
+
+// GetDefaultCollation returns the default collation for charset and handle the default collation for UTF8MB4.
+func GetDefaultCollation(sessVars *variable.SessionVars, cs string) (string, error) {
```
YangKeao, line 799: Why is this function in `ddl` package, but not in `charset` package?

I suggest to modify the current `charset.GetDefaultCollation`, add an argument `sessVars`. We have only 10+ reference to the `charset.GetDefaultCollation`, and most of them have `sctx` in the environment, so I think it won't be a big change to do this refractor. But I'm not sure whether it's the expected behavior... I see this PR doesn't touch the `expression` package, but `(*convertFunctionClass).getFunction` will also call `GetDefaultCollation`. Is it expected?

Besides the `GetDefaultCollation`, I also found `GetDefaultCollationLegacy` used in parser. I'm not sure whether it should also be modified :thinking: .

zimulala: I put it here because it is currently only used for DDL operations, and the PR description explains the scope of this variable's influence, that is, it should not affect the expression.

YangKeao (approved):
LGTM

sessionctx/variable/sysvar.go
```diff
@@ -1383,6 +1383,9 @@ var defaultSysVars = []*SysVar{
 		}
 		return nil
 	}},
+	{Scope: ScopeGlobal | ScopeSession, Name: DefaultCollationForUTF8MB4, Value: mysql.DefaultCollationName, Validation: func(vars *SessionVars, normalizedValue string, originalValue string, scope ScopeFlag) (string, error) {
```
bb7133, line 1386: Please add a warning for `SET default_collation_for_utf8mb4`:

```
mysql> set @@default_collation_for_utf8mb4='utf8mb4_general_ci';

mysql> show warnings;
+---------+------+--------------------------------------------------------------------------------------------------------+
| Level   | Code | Message                                                                                                |
+---------+------+--------------------------------------------------------------------------------------------------------+
| Warning | 1681 | Updating 'default_collation_for_utf8mb4' is deprecated. It will be made read-only in a future release. |
+---------+------+--------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

ddl/ddl_api.go
bb7133: > it should not affect the expression

Well this actually affects the expression, we should expect the following result:

```
tidb> set @@default_collation_for_utf8mb4='utf8mb4_0900_ai_ci';

tidb> select _utf8mb4'a' = _utf8mb4'A', _utf8mb4'a' = _utf8mb4'a  ';
+---------------------------+-----------------------------+
| _utf8mb4'a' = _utf8mb4'A' | _utf8mb4'a' = _utf8mb4'a  ' |
+---------------------------+-----------------------------+
|                         1 |                           0 |
+---------------------------+-----------------------------+
1 row in set (0.01 sec)
```

As mentioned in the PR desc:

> Any statement containing a string literal of the form _utf8mb4'some text' without a COLLATE clause.

So I think @YangKeao 's comment is reasonable.

ddl/ddl_api.go
```diff
@@ -775,10 +784,38 @@ func buildColumnsAndConstraints(
 	return cols, constraints, nil
 }
 
+func getDefaultCollationForUTF8MB4(sessVars *variable.SessionVars, cs string) (string, error) {
```
bb7133, line 787: IMHO it's weird to allow a `cs` parameter for `getDefaultCollationForUTF8MB4` and it returns nothing when `cs != 'utf8mb4'`.

For example, the [code above](https://github.com/zimulala/tidb/blob/2de36755d11812c0a99ce893871370895c92e5e8/ddl/ddl_api.go#L150-L151) confused me at the first glance:

```
	if !explicitCollation && explicitCharset {
		coll, err := getDefaultCollationForUTF8MB4(ctx.GetSessionVars(), charsetOpt.Chs)
```

Could you consider removing this new function and just merge the logic to `charset.GetDefaultCollation`?

ddl/ddl_api.go
bb7133: > it should not affect the expression

Well this actually affects the expression, we should expect the following result:

```
tidb> set @@default_collation_for_utf8mb4='utf8mb4_0900_ai_ci';

tidb> select _utf8mb4'a' = _utf8mb4'A', _utf8mb4'a' = _utf8mb4'a  ';
+---------------------------+-----------------------------+
| _utf8mb4'a' = _utf8mb4'A' | _utf8mb4'a' = _utf8mb4'a  ' |
+---------------------------+-----------------------------+
|                         1 |                           0 |
+---------------------------+-----------------------------+
1 row in set (0.01 sec)
```

As mentioned in the PR desc:

> Any statement containing a string literal of the form _utf8mb4'some text' without a COLLATE clause.

So I think @YangKeao 's comment is reasonable.

zimulala: Yes, as I said above, currently only DDL is affected, and PR also describes that "This PR doesn't support item 4."

ddl/ddl_api.go
zimulala: `charset. GetDefaultCollation` places many, this function(`getDefaultCollationForUTF8MB4`) is used and the function affects only DDL operations. So you can't put this function in

sessionctx/variable/sysvar.go
zimulala: Done.

ddl/ddl_api.go
bb7133: Sorry I missed the description!

bb7133 (approved):
LGTM

Issue #42788, opened 2023-04-03
Deep nested OR expressions caused compile oom


chrysan:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
Run a sql with deep nested OR expressions. 
![img_v2_68600fb7-25ba-4b4f-aa55-d9252b69f4fg](https://user-images.githubusercontent.com/5151781/229529568-c7bc3149-24e3-4fa1-b519-0581bbf86594.jpg)

We need:
1. memory track and threshold 

### 2. What did you expect to see? (Required)
SQL runs normally or is killed due to memory quota. 

### 3. What did you see instead (Required)
TiDB oom.

### 4. What is your TiDB version? (Required)
v6.5

chrysan:
We need:
1. track the memory of the exprs' hashCode and make the sql killed when exceeds memory threshold
2. need a new implementation of ReHashCode which is more memory efficient

AilinKid:
Before Fix
```
➜  core git:(fix-42788) ✗ go tool pprof m5.file
___5BenchmarkSubstituteExpression_in_github_com_pingcap_tidb_planner_core.test
Build ID: ea7c603fe0cf5e18deac5bf65d36f115467fce80
Type: alloc_space
Time: Aug 28, 2023 at 3:58pm (CST)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) list BenchmarkSubstituteExpression
Total: 1.40GB
ROUTINE ======================== github.com/pingcap/tidb/planner/core_test.BenchmarkSubstituteExpression in /home/arenatlx/go/src/github.com/pingcap/tidb/planner/core/rule_generate_column_substitute_test.go
         0   173.44MB (flat, cum) 12.12% of Total
         .          .     29:func BenchmarkSubstituteExpression(b *testing.B) {
         .   169.91MB     30:   store := testkit.CreateMockStore(b)
         .          .     31:   tk := testkit.NewTestKit(b, store)
         .          .     32:   tk.MustExec("use test")
         .          .     33:   tk.MustExec("drop table if exists tai")
         .   512.19kB     34:   tk.MustExec("create table tai(a varchar(256), b varchar(256), c int as (a+1), d int as (b+1))")
         .          .     35:   is := domain.GetDomain(tk.Session()).InfoSchema()
         .          .     36:   _, err := is.TableByName(model.NewCIStr("test"), model.NewCIStr("tai"))
[...]

PR #46450, opened 2023-08-28
planner: don't recompute the hashcode when generated column substitution doesn't happen

AilinKid:

Issue Number: close https://github.com/pingcap/tidb/issues/42788

Problem Summary:

currently Expression.Hashcode() functionality is only used in some limited comparison cases, so lazy computation is necessary.
Say memory formular for every DNF item is m(?)
```
OR                 root node                 hashcode mem consumption: 
(a  OR)                                       m(a) + m(child) = m(a) + m(b) + m(c) + ... m(y) + m(z)
  (b  OR)
     (c  OR)
        (d  OR)
           (e  OR)
              (f  OR)                         m(f) + m(child) = m(f) + m(x) + m(y) + m(y)  
                 (x  OR)                      m(x) + m(child) = m(x) + m(y) + m(y)  
                   (y   z)                    m(y);  m(z);      
                                              total: m(a)+m(b)*2+m(c)*3+ ....+m(x)*(n-1)+m(y)*n+m(z)*n, assume the tree depth is n.
```
from the case above, when computing the hashcode from the raw tree format, additional mem consumption should be allocated since the hashcode cache in every `OR` Expression except for basic column hashcode comsumption: (m(a) + m(b) + m(c) + ... m(y) + m(z))

additionally, actually we rarely need to compute the hashcode from root node, from the most usage from TiDB expression deduplication cases after flatten DNF expressions, what we only need to do is to output every hashcode for every single flattenDNFItem above, which consuming as just few as `(m(a) + m(b) + m(c) + ... m(y) + m(z))` as we said above.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
planner: don't recompute the hashcode when generated column substitution 't happened
```

### DIFF
--- a/planner/core/rule_generate_column_substitute.go
+++ b/planner/core/rule_generate_column_substitute.go
@@ -76,12 +76,15 @@ func collectGenerateColumn(lp LogicalPlan, exprToColumn ExprColumnMap) {
 	}
 }
 
-func tryToSubstituteExpr(expr *expression.Expression, lp LogicalPlan, candidateExpr expression.Expression, tp types.EvalType, schema *expression.Schema, col *expression.Column, opt *logicalOptimizeOp) {
+func tryToSubstituteExpr(expr *expression.Expression, lp LogicalPlan, candidateExpr expression.Expression, tp types.EvalType, schema *expression.Schema, col *expression.Column, opt *logicalOptimizeOp) bool {
+	changed := false
 	if (*expr).Equal(lp.SCtx(), candidateExpr) && candidateExpr.GetType().EvalType() == tp &&
 		schema.ColumnIndex(col) != -1 {
 		*expr = col
 		appendSubstituteColumnStep(lp, candidateExpr, col, opt)
+		changed = true
 	}
+	return changed
 }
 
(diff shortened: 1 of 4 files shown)


planner/core/rule_generate_column_substitute.go
```diff
@@ -97,26 +100,38 @@ func appendSubstituteColumnStep(lp LogicalPlan, candidateExpr expression.Express
...
+		if partial && !changed {
+			changed = true
+		}
+	}
 	defer func() {
 		// If the argument is not changed, hash code doesn't need to recount again.
-		// But we always do it to keep the code simple and stupid.
-		expression.ReHashCode(sf, sctx)
+		if changed {
+			expression.ReHashCode(sf, sctx)
```
AilinKid, line 123: for better we could even just clean the hashcode as [:0] here because the computed hashcode or top  OR expression is never being used.

qw4990 (approved):

winoros (approved):

fixdb (approved):
+1

Issue #46258, opened 2023-08-21
Enhance distributed task execution framework


ywqzzy:
## Enhancement
We need to make Distributed task framework implementation more modular, reduce complexity of implement, decouple the implemented and future function. Moreover, the framework must be more robust and easy to use.
1. By implementing a goroutine pool to control the concurrency of running tasks in the distributed framework, we can effectively prevent background tasks from consuming excessive compute resources and improve management scalability.
2. Use one Dispatcher take care one task, simplify the complexity of Dispatcher management.
3. The task status should follow the state machine, the whole procedure should be simple and easy to reasoning.
4. When the Dispatcher is in error status and does not work,  the impact scope can only be limited only to the task it takes care of. That means the dispatcher and scheduler's execution are decoupled. It will make the task run in distributed more robust.
5. Consider error handling of the task, make the task can be more robust when encountering many abnormal scenes.
6. Support cleanUp routines to do garbage collection for tasks.
7. Support subtask history table to provide more management of the framework.
8. Support metrics to maintain observability.
9. Support role to make the node selection policy more flexible.
10. Provide plan interface to make the whole execution process follow the plan, the business logic will be much clearer to trace and implement.
11. Introduce operator abstraction to make it easier to compose functionalities and provide more sophisticated task runtime statistics collect logic. Further more, the cpu resource should be easy to manage.
#### HA

- [x] 1. #45935 
- [x] 2. #47300 
- [x] 3. #47313
- [x] 4. #47334
- [x] 5. #47886 
- [x] 6. #48033 
- [x] 7. #48095 
- [x] 8. #48660 

#### Pause And Resume

- [x] Support pause and resume. #47037
#### CleanUp routine
- [x] 1. #47084
- [x] 2. #47179 
- [x] 3. #47218

#### Refinement

- [x] 1. #45460
- [x] 2. #45828 
- [x] 3. #45932 
- [x] 4. #45566
- [x] 5. #46497 
- [x] 6. #46357 
- [x] 7. #46472 
- [x] 8. #46603 
- [x] 9. #46669 
- [x] 10. #46686 
- [x] 11. #46722
- [x] 12. #46922
- [x] 13. #46952 
- [x] 14. #46913 
- [x] 15. #46957
- [x] 16. Refine onNextSubtasksBatch #46957 
- [x] 17. #47257 
- [x] 18. #47406 
- [x] 19. #47334 
- [x] 20. #47611 
- [x] 21. #47575 

#### role

[...]

PR #46464, opened 2023-08-28
disttask, ddl: dispatch step-two tasks to proper instances

tangenta:

Issue Number: ref #46258

Problem Summary:

Create a cluster with two TiDB instances.

```sql
mysql> use test
Database changed
mysql> set global tidb_enable_dist_task=1;

mysql> create table t(a int);
Query OK, 0 rows affected (0.13 sec)

mysql> insert into t values(1);

mysql> alter table t add index idx(a);
ERROR 8247 (HY000): Ingest failed: ingest engine not found
```

If a TiDB has never run read-index stage(step one), no ingest engine is created. When it tries to run ingest-stage(step two), "ingest engine not found" error is reported.

Add index dispatcher should only dispatch step-two tasks to those instances that have run step-one.

- Consider previous instance IDs in `GetEligibleInstances()` of add index dispatcher.
- Other tiny code refactoring.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
  
  It is not easy to mock multi-instance environment in unit tests for functional tests.
  ```sql
  mysql> set global tidb_enable_dist_task=1;

  mysql> create table t(a int);
  Query OK, 0 rows affected (0.11 sec)

  mysql> insert into t values(1);
  Query OK, 1 row affected (0.02 sec)

  mysql> alter table t add index idx(a);
  Query OK, 0 rows affected (4.80 sec)

  mysql> admin check table t;
  ```
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


ywqzzy:
Maybe we can refine the `processNormalFlow` interface to make it return ([][]byte, flag bool, error)
If in next stage we want to dispatch the same subtask to nodes that run previous stage.
Just return one subtask and set the bool return value to true.
If the flag == true Then when we dispatch subtasks, just send the subtask to nodes that run previous stage.
Else we use the previous round robin way to dispatch subtasks.

disttask/framework/storage/task_table.go
```diff
@@ -510,6 +510,26 @@ func (stm *TaskManager) GetSchedulerIDsByTaskID(taskID int64) ([]string, error)
 	return instanceIDs, nil
 }
 
+// GetSchedulerIDsByTaskIDAndStep gets the scheduler IDs of the given global task ID and step.
+func (stm *TaskManager) GetSchedulerIDsByTaskIDAndStep(taskID int64, step int64) ([]string, error) {
```
ywqzzy, line 514: Add ut?

ywqzzy (approved):
Rest LGTM

GMHDBJD (approved):
LGTM

tangenta:
Wait for https://github.com/pingcap/tidb/pull/46671

util/disttask/idservice.go
```diff
@@ -32,13 +32,18 @@ func GenerateExecID(ip string, port uint) string {
...
+func FindServerInfo(serverInfos []*infosync.ServerInfo, schedulerID string) int {
+	for i, serverInfo := range serverInfos {
 		serverID := GenerateExecID(serverInfo.IP, serverInfo.Port)
 		if serverID == schedulerID {
-			return true
+			return i
 		}
 	}
-	return false
+	return -1
```
Benjamin2037, line 46: Should we use a InvalidServerID const to mean -1?

Issue #46351, opened 2023-08-23
Tracking issue of `lock/unlock stats`


hi-rustin:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Original pull request: https://github.com/pingcap/tidb/pull/38387
[
The design doc.
](https://pingcap.feishu.cn/docx/FaSgdt775oKkpdxlUrocfjn9nph)
**Describe the feature you'd like:**
Sometimes performance drop due to the explained change caused by statistic info change after data changes in the table.
So we want to introduce table statistic lock/unlock function to avoid unexpected performance drop by statistic changes

PR #46478, opened 2023-08-29
statistics: rewrite unlock stats handler

hi-rustin:

Problem Summary:

- Rewrote unlock stats handler
- Used `exec.ExecuteInternal` instead of `h.execRestrictedSQL`
- Added `updateStatsAndUnlockTable` to reuse code.
- Added tests for `RemoveIfTableLocked` and `LockTableIndexOf`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/statistics/handle/handletest/statslock/stats_lcok_test.go
+++ b/statistics/handle/handletest/statslock/stats_lcok_test.go
@@ -61,7 +61,7 @@ func TestStatsLockAndUnlockTable(t *testing.T) {
 	require.Equal(t, tblStats, tblStats1)
 
 	tableLocked1 := handle.GetTableLockedAndClearForTest()
-	err = handle.LoadLockedTables()
+	err = handle.LoadLockedTables(8)
 	require.Nil(t, err)
 	tableLocked2 := handle.GetTableLockedAndClearForTest()
 	require.Equal(t, tableLocked1, tableLocked2)
@@ -117,7 +117,7 @@ func TestStatsLockTableRepeatedly(t *testing.T) {
 		"Warning 1105 skip locking locked table: test.t",
 	))
 
-	err = handle.LoadLockedTables()
+	err = handle.LoadLockedTables(8)
 	require.Nil(t, err)
 	tableLocked2 := handle.GetTableLockedAndClearForTest()
 	require.Equal(t, tableLocked1, tableLocked2)
@@ -176,7 +176,7 @@ func TestStatsLockAndUnlockTables(t *testing.T) {
 	require.Equal(t, tbl2Stats, tbl2Stats1)
 
 	tableLocked1 := handle.GetTableLockedAndClearForTest()
-	err = handle.LoadLockedTables()
+	err = handle.LoadLockedTables(8)
 	require.Nil(t, err)
 	tableLocked2 := handle.GetTableLockedAndClearForTest()
 	require.Equal(t, tableLocked1, tableLocked2)
--- a/statistics/handle/handletest/statslock/stats_lcok_test.go
+++ b/statistics/handle/handletest/statslock/stats_lcok_test.go
@@ -76,7 +76,7 @@ func TestStatsLockAndUnlockTable(t *testing.T) {
 	require.Equal(t, int64(2), tblStats2.RealtimeCount)
 }
 
-func TestStatsLockTableRepeatedly(t *testing.T) {
+func TestStatsLockTableAndUnlockTableRepeatedly(t *testing.T) {
 	restore := config.RestoreFunc()
 	defer restore()
 	config.UpdateGlobal(func(conf *config.Config) {
@@ -121,6 +121,22 @@ func TestStatsLockTableRepeatedly(t *testing.T) {
 	require.Nil(t, err)
 	tableLocked2 := handle.GetTableLockedAndClearForTest()
 	require.Equal(t, tableLocked1, tableLocked2)
+
+	// Unlock the table.
+	tk.MustExec("unlock stats t")
+	rows = tk.MustQuery("select count(*) from mysql.stats_table_locked").Rows()
+	num, _ = strconv.Atoi(rows[0][0].(string))
+	require.Equal(t, num, 0)
+
+	tk.MustExec("analyze table test.t")
+	tblStats2 := handle.GetTableStats(tbl.Meta())
+	require.Equal(t, int64(2), tblStats2.RealtimeCount)
+
+	// Unlock the table again and check the warning.
+	tk.MustExec("unlock stats t")
+	tk.MustQuery("show warnings").Sort().Check(testkit.Rows(
+		"Warning 1105 skip unlocking unlocked table: test.t",
+	))
 }
 
 func TestStatsLockAndUnlockTables(t *testing.T) {
--- a/statistics/handle/handletest/statslock/stats_lcok_test.go
+++ b/statistics/handle/handletest/statslock/stats_lcok_test.go
@@ -113,7 +113,7 @@ func TestStatsLockTableAndUnlockTableRepeatedly(t *testing.T) {
 	// Lock the table again and check the warning.
 	tableLocked1 := handle.GetTableLockedAndClearForTest()
 	tk.MustExec("lock stats t")
-	tk.MustQuery("show warnings").Sort().Check(testkit.Rows(
+	tk.MustQuery("show warnings").Check(testkit.Rows(
 		"Warning 1105 skip locking locked table: test.t",
 	))
 
@@ -134,7 +134,7 @@ func TestStatsLockTableAndUnlockTableRepeatedly(t *testing.T) {
 
 	// Unlock the table again and check the warning.
 	tk.MustExec("unlock stats t")
-	tk.MustQuery("show warnings").Sort().Check(testkit.Rows(
+	tk.MustQuery("show warnings").Check(testkit.Rows(
 		"Warning 1105 skip unlocking unlocked table: test.t",
 	))
 }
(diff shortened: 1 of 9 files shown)


hi-rustin:
1. Start the TiDB cluster: `tiup playground nightly --tiflash 0 --db.binpath /Users/hi-rustin/GolandProjects/tidb/bin/tidb-server`
2. Create table and analyze it:
```sql
create table t(a int, b int);
insert into t values (1,2), (3,4), (5,6), (7,8);
analyze table t;
show warnings;
```
3. Lock table: `lock stats t;`
4. Analyze it again: `analyze table t;`
5. Check the warnings: `show warnings;`
```sql
mysql> show warnings;
+---------+------+-----------------------------------------------------------------------------------------------------------------------------------------+
| Level   | Code | Message                                                                                                                                 |
+---------+------+-----------------------------------------------------------------------------------------------------------------------------------------+
| Note    | 1105 | Analyze use auto adjusted sample rate 1.000000 for table test.t, reason to use this rate is "use min(1, 110000/4) as the sample-rate=1" |
| Warning | 1105 | skip analyze locked table: t                                                                                                            |
+---------+------+-----------------------------------------------------------------------------------------------------------------------------------------+
```
6. Check the `mysql.stats_table_locked` table:
```sql
mysql> use mysql;

Database changed
mysql> select * from stats_table_locked;
+----------+--------------+-------+--------------------+
| table_id | modify_count | count | version            |
+----------+--------------+-------+--------------------+
|       98 |            4 |     4 | 443863948666601483 |
+----------+--------------+-------+--------------------+
1 row in set (0.00 sec)
```
7. Try to lock it again: `lock stats t;`
```sql
mysql> use test;

Database changed
mysql> lock stats t;
Query OK, 0 rows affected, 1 warning (0.01 sec)

mysql> show warnings;
+---------+------+-----------------------------------+
| Level   | Code | Message                           |
+---------+------+-----------------------------------+
| Warning | 1105 | skip locking locked table: test.t |
+---------+------+-----------------------------------+
1 row in set (0.00 sec)
```

statistics/handle/unlock_stats_handler.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/pingcap/tidb/kv"
+	"github.com/pingcap/tidb/parser/ast"
+	"github.com/pingcap/tidb/statistics/handle/cache"
+	"github.com/pingcap/tidb/util/sqlexec"
+	"go.uber.org/zap"
+)
+
+// RemoveLockedTables remove tables from table locked array
+func (h *Handle) RemoveLockedTables(tids []int64, pids []int64, tables []*ast.TableName, maxChunkSize int) (string, error) {
```
qw4990, line 28: Please add some comments for these arguments, are `tids`, `pids` and `tables` the target table IDs to remove here?

qw4990, line 28: I recommend removing `maxChunkSize` here. This operation of removing locked tables is very lightweight, so this argument has almost no impact on this operation. I recommend removing it to simplify our code, and we can just use an arbitrary number here.

hi-rustin: Sounds good to me.

hi-rustin: Addressed in [`e83859a` (#46478)](https://github.com/pingcap/tidb/pull/46478/commits/e83859aeb36a1bc0ed05e21adc55dd77eb41d79c)

hi-rustin: Good suggestion!

hi-rustin: Added. Thanks!

statistics/handle/unlock_stats_handler.go
```diff
@@ -0,0 +1,126 @@
...
+
+	_, err := exec.ExecuteInternal(ctx, "BEGIN PESSIMISTIC")
+	if err != nil {
+		return "", err
+	}
+
+	// Load tables to check locked before delete.
+	tableLocked, err := loadLockedTables(ctx, exec, maxChunkSize)
+	if err != nil {
+		return "", err
```
qw4990, line 47: It seems like the transaction is not finished. We'd better use `defer finishTransaction(...)` above.

statistics/handle/unlock_stats_handler.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return err
+	}
+
+	if _, err := exec.ExecuteInternal(ctx,
+		"UPDATE mysql.stats_meta SET version = %?, count = count + %?, modify_count = modify_count + %? WHERE table_id = %?",
+		version, count, modifyCount, tid); err != nil {
+		return err
+	}
+	cache.TableRowStatsCache.Invalidate(tid)
```
qw4990, line 99: I recommend removing this operation of refreshing Delta to make the whole unlocking process simpler since there is another background task goroutine updating all Delta data periodically and automatically.
In addition, a TiDB cluster has multiple TiDB instances, so just refreshing this instance's Delta seems useless, the Delta of this table is still inaccurate.
So I recommend removing this to make it simpler.

statistics/handle/handletest/statslock/stats_lcok_test.go
```diff
@@ -121,6 +121,22 @@ func TestStatsLockTableRepeatedly(t *testing.T) {
...
+
+	tk.MustExec("analyze table test.t")
+	tblStats2 := handle.GetTableStats(tbl.Meta())
+	require.Equal(t, int64(2), tblStats2.RealtimeCount)
+
+	// Unlock the table again and check the warning.
+	tk.MustExec("unlock stats t")
+	tk.MustQuery("show warnings").Sort().Check(testkit.Rows(
+		"Warning 1105 skip unlocking unlocked table: test.t",
+	))
```
hawkingrei, line 139: If result is one, it is unnecessary to sort it.

statistics/handle/lock_stats_handler.go
```diff
@@ -19,20 +19,36 @@ import (
...
+	lockedStatus   = "locked"
+	unlockedStatus = "unlocked"
+)
+
+var (
+	// maxChunkSize is the max chunk size for load locked tables.
+	// We use 1024 as the default value, which is the same as the default value of session.maxChunkSize.
+	// The reason why we don't use session.maxChunkSize is that we don't want to introduce a new dependency.
+	// See: https://github.com/pingcap/tidb/pull/46478#discussion_r1308786474
+	maxChunkSize = 1024
```
hawkingrei, line 41: I think we can create a new package in the ```util``` to put all const magic number.

hi-rustin: This is only for the lock and unlock stats feature.

statistics/handle/unlock_stats_handler.go
hi-rustin: Good catch! I totally forgot that.

hi-rustin: Addressed in [`13cc3fd` (#46478)](https://github.com/pingcap/tidb/pull/46478/commits/13cc3fdf175c81efb78b44dfcc6dd5ed8a1ed5f5)

statistics/handle/handletest/statslock/stats_lcok_test.go
hi-rustin: Good catch! Addressed in [`32beb78` (#46478)](https://github.com/pingcap/tidb/pull/46478/commits/32beb785df4911cb8a8ac8515d58ce4091ce3e4a)

statistics/handle/unlock_stats_handler.go
hi-rustin: > I recommend removing this operation of refreshing Delta to make the whole unlocking process simpler since there is another background task goroutine updating all Delta data periodically and automatically.

Good point!

hi-rustin: But in `dumpTableStatCountToKV` we also skip locked tables. This is the only chance to update the delta into the `mysql.stats_meta`.

hi-rustin: > In addition, a TiDB cluster has multiple TiDB instances, so just refreshing this instance's Delta seems useless, the Delta of this table is still inaccurate.

I didn't get this. Could you please explain it a bit?

qw4990 (approved):

statistics/handle/lock_stats_handler.go
```diff
@@ -84,31 +95,26 @@ func (h *Handle) AddLockedTables(tids []int64, pids []int64, tables []*ast.Table
...
-	}
-
 	// Update handle.tableLocked after transaction success, if txn failed, tableLocked won't be updated.
 	h.tableLocked = tableLocked
 
-	mag := generateDuplicateTablesMessage(tids, dupTables)
-	return mag, nil
+	mag := generateSkippedTablesMessage(tids, skippedTables, lockAction, lockedStatus)
+	// Note: defer commit transaction, so we can't use `return nil` here.
+	return mag, err
```
hawkingrei, line 103:
```suggestion
	msg := generateSkippedTablesMessage(tids, skippedTables, lockAction, lockedStatus)
	// Note: defer commit transaction, so we can't use `return nil` here.
	return msg, err
```

statistics/handle/unlock_stats_handler.go
qw4990: Each TiDB instance has its own `Delta Data`, and they use `Delta Data` to update some system tables periodically and asynchronously, so the effect of `Delta Data` on these system tables is "eventually consistent".
So triggering an update operation here doesn't change this attribute, it's useless and redundant.

statistics/handle/unlock_stats_handler.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if _, err := exec.ExecuteInternal(ctx, "DELETE FROM mysql.stats_table_locked WHERE table_id = %?", tid); err != nil {
+		return err
+	}
+
+	return nil
```
hawkingrei, line 105:
```suggestion
	_, err := exec.ExecuteInternal(ctx, "DELETE FROM mysql.stats_table_locked WHERE table_id = %?", tid)
	return err
```

hawkingrei (approved):

Issue #45395, opened 2023-07-17
reduce rule numbers if not set the role in the rule SQL


nolouch:
## Problem
```
 create placement policy retail_pos constraints="[+retail_region=fast]";
 CREATE TABLE t1 (a INT);
 ALTER TABLE t1  PLACEMENT POLICY=retail_pos;
```
The rules in PD side:
```
[
  {
    "group_id": "TiDB_DDL_90",
    "id": "table_rule_90_0",
    "index": 40,
    "start_key": "7480000000000000ff5a00000000000000f8",
    "end_key": "7480000000000000ff5b00000000000000f8",
    "role": "leader",
    "is_witness": false,
    "count": 1,
    "label_constraints": [
      {
        "key": "retail_region",
        "op": "in",
        "values": [
          "fast"
        ]
      },
      ....
    ],
    "create_timestamp": 1689589620
  },
  {
    "group_id": "TiDB_DDL_90",
    "id": "table_rule_90_1",
    "index": 40,
    "start_key": "7480000000000000ff5a00000000000000f8",
    "end_key": "7480000000000000ff5b00000000000000f8",
    "role": "voter",
    "is_witness": false,
    "count": 2,
    "label_constraints": [
      {
        "key": "retail_region",
        "op": "in",
        "values": [
          "fast"
        ]
      },
      ...
    ],
    "create_timestamp": 1689589620
  },
  {
    "group_id": "pd",
    "id": "default",
    "start_key": "",
    "end_key": "",
    "role": "voter",
    "is_witness": false,
    "count": 3
  }
]

```

There is no need to split into two rules because we do not need to distinguish roles, two rules perhaps exist some problems. such as `location-labels` may not work in https://github.com/tikv/pd/issues/6637

PR #46529, opened 2023-08-30
placement: merge the rules if the constraints are the same

nolouch:

Issue Number: close https://github.com/pingcap/tidb/issues/45395

Problem Summary:

```
 create placement policy retail_pos constraints="[+retail_region=fast]";
 CREATE TABLE t1 (a INT);
 ALTER TABLE t1  PLACEMENT POLICY=retail_pos;
```
The rules in PD side:
```
[
  {
    "group_id": "TiDB_DDL_90",
    "id": "table_rule_90_0",
    "index": 40,
    "start_key": "7480000000000000ff5a00000000000000f8",
    "end_key": "7480000000000000ff5b00000000000000f8",
    "role": "leader",
    "is_witness": false,
    "count": 1,
    "label_constraints": [
      {
        "key": "retail_region",
        "op": "in",
        "values": [
          "fast"
        ]
      },
      ....
    ],
    "create_timestamp": 1689589620
  },
  {
    "group_id": "TiDB_DDL_90",
    "id": "table_rule_90_1",
    "index": 40,
    "start_key": "7480000000000000ff5a00000000000000f8",
    "end_key": "7480000000000000ff5b00000000000000f8",
    "role": "voter",
    "is_witness": false,
    "count": 2,
    "label_constraints": [
      {
        "key": "retail_region",
        "op": "in",
        "values": [
          "fast"
        ]
      },
      ...
    ],
    "create_timestamp": 1689589620
  },
  {
    "group_id": "pd",
    "id": "default",
    "start_key": "",
    "end_key": "",
    "role": "voter",
    "is_witness": false,
    "count": 3
  }
]

```

There is no need to split into two rules because we do not need to distinguish roles, two rules perhaps exist some problems. such as `location-labels` may not work in https://github.com/tikv/pd/issues/6637

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

### Release note

```release-note
None
```


ddl/placement/constraints.go
```diff
@@ -110,3 +113,34 @@ func (constraints *Constraints) Add(label Constraint) error {
 	}
 	return nil
 }
+
+// FingerPrint returns a unique string for the constraints.
+func (constraints Constraints) FingerPrint() string {
+	copied := make(Constraints, len(constraints))
+	copy(copied, constraints)
+	sort.SliceStable(copied, func(i, j int) bool {
```
hawkingrei, line 121: slices.SortStableFunction is better. it can remove BCE.

nolouch:
PTAL @CabinfeverB @JmPotato

ddl/placement/bundle.go
```diff
@@ -320,36 +306,140 @@ func (b *Bundle) Tidy() error {
...
-		})
+		existing.rules = append(existing.rules, rule)
+	}
+	for _, group := range groups {
+		group.MergeRulesByRole()
+	}
+	if len(groups) == 1 {
+		for _, group := range groups {
+			group.MergeTransformableRoles()
+		}
```
lcwangchao, line 331: I think we can use `groups[0].MergeTransformableRoles()` instead ?

ddl/placement/constraints.go
```diff
@@ -110,3 +115,34 @@ func (constraints *Constraints) Add(label Constraint) error {
 	}
 	return nil
 }
+
+// FingerPrint returns a unique string for the constraints.
+func (constraints Constraints) FingerPrint() string {
```
lcwangchao, line 120: `Constraints` => `*Constraints`? I got a warning in IDE: "Struct Constraints has methods on both value and pointer receivers. Such usage is not recommended by the Go Documentation. "

ddl/placement/bundle.go
```diff
@@ -320,36 +306,140 @@ func (b *Bundle) Tidy() error {
...
+	// sort by id
+	sort.SliceStable(finalRules, func(i, j int) bool {
+		return finalRules[i].ID < finalRules[j].ID
+	})
+	b.Rules = finalRules
+	return nil
+}
+
+// ConstraintsGroup is a group of rules with the same constraints.
+type ConstraintsGroup struct {
```
lcwangchao, line 349: Is `ConstraintsGroup` only used in the current package and as an intermediate object? If so, I think it's better to declare it as private

ddl/placement/bundle.go
```diff
@@ -320,36 +306,140 @@ func (b *Bundle) Tidy() error {
...
+				return ErrInvalidPlacementOptions
+			}
+			leaderGroup = group
+		}
+		if group.canBecameLeader {
+			canBecameLeaderNum++
+		}
+	}
+	if leaderGroup != nil && canBecameLeaderNum == 1 {
+		leaderGroup.MergeTransformableRoles()
```
lcwangchao, line 370: Could you add some comments here or somewhere to explain why when two or more groups have a leader or voter role, we should skip `MergeTransformableRoles`?

ddl/placement/bundle.go
```diff
@@ -288,27 +288,13 @@ func (b *Bundle) String() string {
...
 // Tidy will post optimize Rules, trying to generate rules that suits PD.
 func (b *Bundle) Tidy() error {
-	extraCnt := map[PeerRoleType]int{}
-	newRules := b.Rules[:0]
-
-	// One Bundle is from one PlacementSettings, rule share same location labels, so we can use the first rule's location labels.
-	var locationLabels []string
+	tempRules := b.Rules[:0]
+	id := 0
 	for _, rule := range b.Rules {
```
lcwangchao, line 293: Could you add some extra checks for rules in Tidy? I think `Tidy` can only be called when some fields are empty (`StartKeyHex/EndKeyHex` .etc).

ddl/placement/constraints.go
```diff
@@ -110,3 +115,34 @@ func (constraints *Constraints) Add(label Constraint) error {
...
+	// Calculate the SHA256 hash of the concatenated constraints
+	hash := sha256.Sum256([]byte(combinedConstraints))
+
+	// Encode the hash as a base64 string
+	hashStr := base64.StdEncoding.EncodeToString(hash[:])
+
+	return hashStr
+}
+
+func constraintToString(c Constraint) string {
```
xhebox, line 141: `func (*Constraint) String()`?

ddl/placement/bundle.go
```diff
@@ -320,36 +306,140 @@ func (b *Bundle) Tidy() error {
...
+			return err
+		}
+	}
+	for _, group := range groups {
+		finalRules = append(finalRules, group.rules...)
+	}
+	// sort by id
+	sort.SliceStable(finalRules, func(i, j int) bool {
+		return finalRules[i].ID < finalRules[j].ID
+	})
```
xhebox, line 343: Do we need to?

ddl/placement/bundle.go
```diff
@@ -320,36 +306,140 @@ func (b *Bundle) Tidy() error {
...
+	canBecameLeader bool
+	isLeaderGroup   bool
+}
+
+func transformableLeaderConstraint(groups map[string]*ConstraintsGroup) error {
+	var leaderGroup *ConstraintsGroup
+	canBecameLeaderNum := 0
+	for _, group := range groups {
+		if group.isLeaderGroup {
+			if !(leaderGroup == nil) {
```
xhebox, line 360:
```suggestion
			if leaderGroup != nil {
```

ddl/placement/bundle.go
xhebox: I think we can just use `if err := transformableLeaderConstraint(groups); err != nil ` instead. It is same.

Or probably `groups.transformableLeaderConstraint()`.

ddl/placement/bundle.go
```diff
@@ -320,36 +306,140 @@ func (b *Bundle) Tidy() error {
...
+		c.rules = append(c.rules, mergedRule)
+	}
+}
+
+// MergeTransformableRoles merges all the rules to one that can be transformed to other roles.
+func (c *ConstraintsGroup) MergeTransformableRoles() {
+	if len(c.rules) == 0 {
+		return
+	}
+	newRules := make([]*Rule, 0, len(c.rules))
```
CabinfeverB, line 416: The maximum length is 2?

ddl/placement/bundle.go
CabinfeverB: I agree. But if 1 leader and 2 followers in group1 and 2 voters in group2, can we merge group1 to voters?

xhebox: > I agree. But if 1 leader and 2 followers in group1 and 2 voters in group2, can we merge group1 to voters?

No. They don't share same constraints.

ddl/placement/bundle.go
nolouch: As the comments saied, it's optimized rules, try to simplify the rules.

ddl/placement/bundle.go
nolouch: > I agree. But if 1 leader and 2 followers in group1 and 2 voters in group2, can we merge group1 to voters?

that means the `group 1` has leader priority than group 2, so it's not transformable.

ddl/placement/bundle.go
nolouch: yes, because merge may change the orders, and it also make the test stable.

nolouch:
> And I have a question: as PR description, the cause of multiple rules is multiple tables instead of role?

it is caused by multi-role setting. see https://github.com/pingcap/tidb/blob/2472154f70948a3cd5695c7cd17d9abbd3ac95a4/ddl/placement/bundle.go#L57-L135

ddl/placement/bundle.go
nolouch: maybe more.

nolouch:
PTAL @lcwangchao @xhebox @CabinfeverB  thanks.

xhebox (approved):

CabinfeverB (approved):

PR #46533, opened 2023-08-30
*: Exchange partition, fix LIST COLUMNs validation as well as NULL validation

### LINKED ISSUE #46492: EXCHANGE PARTITION issue: Non-partitioned tables that nonmatch the partition definition can be swapped successfully
## Bug Report

<img width="751" alt="image" src="https://github.com/pingcap/tidb/assets/66671526/35757598-aec0-45e1-8032-155398769fc7">
https://docs.pingcap.com/tidb/v6.5/partitioned-table#partition-management

**partition table:**
mysql> create table test_p(id int,create_ts datetime,name varchar(10)) partition by list columns(create_ts)(partition p20230829 values in ('2023-08-29'),partition p20230830 values in ('2023-08-30'));
Query OK, 0 rows affected (0.13 sec)
mysql> insert into test_p values (1,'2023-08-29','a');
mysql> insert into test_p values (2,'2023-08-30','b');
mysql> insert into test_p values (3,'2023-08-31','c');
ERROR 1526 (HY000): Table has no partition for value from column_list

**non-partitioned table:**
mysql> create table test_p_swap(id int,create_ts datetime,name varchar(10));
Query OK, 0 rows affected (0.12 sec)
mysql> insert into test_p_swap values (3,'2023-08-31','c');

**exchange partition:**
mysql> alter table test_p EXCHANGE PARTITION p20230829 WITH TABLE test_p_swap;
Query OK, 0 rows affected, 1 warning (0.25 sec)

mysql> show warnings;
+---------+------+---------------------------------------------------------------------------------------+
| Level   | Code | Message                                                                               |
+---------+------+---------------------------------------------------------------------------------------+
| Warning | 1105 | after the exchange, please analyze related table of the exchange to update statistics |
+---------+------+---------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> select * from test_p;
+------+---------------------+------+
| id   | create_ts           | name |
+------+---------------------+------+
|    2 | 2023-08-30 00:00:00 | b    |
|    3 | 2023-08-31 00:00:00 | c    |
+------+---------------------+------+
2 rows in set, 1 warning (0.00 sec)

mysql> show create table test_p;
+--------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table  | Create Table                                                                                                                                                                                                                                                                                                                     |
+--------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| test_p | CREATE TABLE `test_p` (
  `id` int(11) DEFAULT NULL,
  `create_ts` datetime DEFAULT NULL,
  `name` varchar(10) DEFAULT NULL
PARTITION BY LIST COLUMNS(`create_ts`)
(PARTITION `p20230829` VALUES IN ('2023-08-29'),
 PARTITION `p20230830` VALUES IN ('2023-08-30')) |
+--------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

The exchange partition statement will fail.

Non-partitioned tables that nonmatch the partition definition can be swapped successfully.

v6.5.3

mjonss:

Issue Number: close #46492

Problem Summary:
The main issue was that the values are store in single quoted strings, and when adding them to the parser as parameters, they are quoted again, which issues (ignored) internal warnings, and no matching rows will be found.

Also the check did not handle NULL, so I added proper null-safe comparison.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


mjonss:
I wonder if it would be better to just run the partition pruning on each row, and see if they match or not? (but then we would not be able to use any indexes etc.)

YangKeao:
> I wonder if it would be better to just run the partition pruning on each row, and see if they match or not? (but then we would not be able to use any indexes etc.)

It sounds not scalable? Even without index, the SQL can distribute the seletions to multiple TiKVs and make it faster.

ddl/partition.go
```diff
@@ -3461,73 +3463,94 @@ func buildCheckSQLForRangeExprPartition(pi *model.PartitionInfo, index int, sche
...
-		inValues = append(inValues, inValue...)
+	// How to find a match?
+	// (row <=> vals1) OR (row <=> vals2)
+	// How to find a non-matching row:
+	// NOT ( (row <=> vals1) OR (row <=> vals2) ... )
+	buf.WriteString("select 1 from %n.%n where not (")
+	colNames := make([]string, 0, len(pi.Columns))
+	for i := range pi.Columns {
+		// TODO: check if there are no proper quoting function for this?
+		n := "`" + strings.ReplaceAll(pi.Columns[i].O, "`", "``") + "`"
```
YangKeao, line 3534: It's a little complex, but maybe helpful. See `/parser/format/format.go: (*RestoreContext).WriteName`.

It's also possible to "Restore" the whole SQL from ast, and maybe building an AST is more intuitive than concating strings for a complex statement, (but I'm not sure).

YangKeao (commented):
Rest LGTM

mjonss:
> > I wonder if it would be better to just run the partition pruning on each row, and see if they match or not? (but then we would not be able to use any indexes etc.)
> 
> It sounds not scalable? Even without index, the SQL can distribute the seletions to multiple TiKVs and make it faster.

You are right, I realised that as well afterwards, since this would not allow to push down the filtering to TiKV/TiFlash, so all rows would need to go to TiDB, which would scale poorly.
So let us just keep it as is.
The only technical issue is that we need have this code complexity of generating the WHERE clause as in this PR, instead of reusing existing code for partitioning.

ddl/partition.go
mjonss: OK, that is the function I saw as well, but the overhead of using it was bigger than the benefit, so I propose to use this direct quoting. Also generating the ast tree instead of sql string is probably more complex than the benefits would be.

xhebox (approved):

ddl/partition.go
tiancaiamao: This should work for most of the cases, I would suggest double check the internal sql not using ANSI/ANSI_QUOTE sql_mode

tiancaiamao (approved):

PR #46555, opened 2023-08-31
statistics: remove lockedTable cache

hi-rustin:

Problem Summary:

### What changed and how it works?

- Removed the `lockedTable` cache. The reason we're removing it is because it doesn't help. It's locked and unlocked so infrequently that it doesn't matter if we query it every time. This way we don't need to load it as often.

- Removed `LoadLockedTables` background task
- Use the read lock to protect the SQLExecutor
- Move all code to `lockstats` pkg

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


hi-rustin:
1. Start the TiDB cluster: `tiup playground nightly --tiflash 0 --db.binpath /Users/hi-rustin/GolandProjects/tidb/bin/tidb-server`
2. Create table and analyze it:
```sql
create table t(a int, b int);
insert into t values (1,2), (3,4), (5,6), (7,8);
analyze table t;
show warnings;
```
3. Lock table: `lock stats t;`
4. Analyze it again: `analyze table t;`
5. Check the warnings: `show warnings;`
```sql
mysql> show warnings;
+---------+------+-----------------------------------------------------------------------------------------------------------------------------------------+
| Level   | Code | Message                                                                                                                                 |
+---------+------+-----------------------------------------------------------------------------------------------------------------------------------------+
| Note    | 1105 | Analyze use auto adjusted sample rate 1.000000 for table test.t, reason to use this rate is "use min(1, 110000/4) as the sample-rate=1" |
| Warning | 1105 | skip analyze locked table: t                                                                                                            |
+---------+------+-----------------------------------------------------------------------------------------------------------------------------------------+
```
6. Check the `mysql.stats_table_locked` table:
```sql
mysql> use mysql;

Database changed
mysql> select * from stats_table_locked;
+----------+--------------+-------+--------------------+
| table_id | modify_count | count | version            |
+----------+--------------+-------+--------------------+
|       98 |            4 |     4 | 443863948666601483 |
+----------+--------------+-------+--------------------+
1 row in set (0.00 sec)
```
7. Try to lock it again: `lock stats t;`
```sql
mysql> use test;

Database changed
mysql> lock stats t;
Query OK, 0 rows affected, 1 warning (0.01 sec)

mysql> show warnings;
+---------+------+-----------------------------------+
| Level   | Code | Message                           |
+---------+------+-----------------------------------+
| Warning | 1105 | skip locking locked table: test.t |
+---------+------+-----------------------------------+
1 row in set (0.00 sec)
```
8. Show locked stats: `show stats_locked;`.
```sql
mysql> show stats_locked;
+---------+------------+----------------+--------+
| Db_name | Table_name | Partition_name | Status |
+---------+------------+----------------+--------+
| test    | t          |                | locked |
+---------+------------+----------------+--------+
1 row in set (0.00 sec)
```
9. Unlock the table: `unlock stats t;`
10. Check the `mysql.stats_table_locked` table again:
```sql
mysql> select * from stats_table_locked;
Empty set (0.00 sec)
```
11. Show locked stats: `show stats_locked;`
```sql
mysql> show stats_locked;
Empty set (0.00 sec)
```
12. Unlock the table again and check warnings: `unlock stats t;`
```sql
mysql> unlock stats t;

mysql> show warnings;
+---------+------+---------------------------------------+
| Level   | Code | Message                               |
+---------+------+---------------------------------------+
| Warning | 1105 | skip unlocking unlocked table: test.t |
+---------+------+---------------------------------------+
1 row in set (0.00 sec)
```

statistics/handle/lock_stats_handler.go
```diff
@@ -49,161 +29,50 @@ var (
...
-	// Update all locked tables.
-	h.tableLocked = tableLocked
-
-	return nil
+	return lockstats.RemoveLockedTables(h.mu.ctx.(sqlexec.SQLExecutor), tids, pids, tables)
 }
 
 // IsTableLocked check whether table is locked in handle with Handle.Mutex
-func (h *Handle) IsTableLocked(tableID int64) bool {
+func (h *Handle) IsTableLocked(tableID int64) (bool, error) {
```
qw4990, line 50: Can we allow this method to accept multiple table IDs like `IsTableLocked(tableIDs ...int64) ([]bool, error)` to get a better performance?

statistics/handle/lock_stats_handler.go
```diff
@@ -49,161 +29,50 @@ var (
...
 }
 
-func loadLockedTables(ctx context.Context, exec sqlexec.SQLExecutor, maxChunkSize int) ([]int64, error) {
-	recordSet, err := exec.ExecuteInternal(ctx, "SELECT table_id FROM mysql.stats_table_locked")
-	if err != nil {
-		return nil, err
-	}
-	rows, err := sqlexec.DrainRecordSet(ctx, recordSet, maxChunkSize)
+// isTableLockedWithoutLock check whether table is locked in handle without Handle.Mutex
+func (h *Handle) isTableLockedWithoutLock(tableID int64) (bool, error) {
```
qw4990, line 65: `h.mu` lock's granularity is too coarse and not reentrant, which brings lots of risks to us.
For example, someone may don't know whether he should use `h.IsTableLocked` or `h.IsTableLockedWithoutLock`, and once they use the wrong method, we'll get a terrible DATA RACE or DEADLOCK problem.
How about adding a TODO here to notify us to replace this lock with a reentrant lock in the future?

statistics/handle/lock_stats_handler.go
hi-rustin: Good point. I tried it this morning. It makes the code complicated. I will try it again.

hi-rustin: Addressed in [`a6a0bf3` (#46555)](https://github.com/pingcap/tidb/pull/46555/commits/a6a0bf31b0559451da544c5ace98819dce977a3f)

statistics/handle/update.go
```diff
@@ -716,9 +719,24 @@ func (h *Handle) HandleAutoAnalyze(is infoschema.InfoSchema) (analyzed bool) {
...
+			continue
+		}
+
 		for _, tbl := range tbls {
-			//if table locked, skip analyze
-			if h.IsTableLocked(tbl.Meta().ID) {
+			// If table locked, skip analyze.
+			if checkedTableIDs[tbl.Meta().ID] {
+				logutil.BgLogger().Info("skip analyze locked table", zap.String("category", "stats"),
+					zap.String("db", db), zap.String("table", tbl.Meta().Name.O))
```
qw4990, line 739: Please add some comments here to notice this check is not complete. For example, some other TiDB node may lock or unlock this table between this check and the analyze command.

qw4990: Better to enhance or remove this logic in the future.

qw4990 (approved):

hawkingrei (approved):

time-and-fate (commented):
Why are we switching from `Lock` to `RLock`? Would this possibly cause problems when there are concurrent operations?

hi-rustin:
> Why are we switching from `Lock` to `RLock`? Would this possibly cause problems when there are concurrent operations?

It was a mistake. I reverted it.

executor/analyze.go
```diff
@@ -170,37 +173,45 @@ func (e *AnalyzeExec) Next(ctx context.Context, _ *chunk.Chunk) error {
...
-					logutil.BgLogger().Warn("Unknown table ID in analyze task", zap.Int64("tid", tableID))
-				} else {
-					skippedTables = append(skippedTables, tbl.Meta().Name.L)
-				}
+			filteredTasks = append(filteredTasks, taskMap[tid]...)
+			needAnalyzeTableCnt++
+		} else {
+			tbl, ok := infoSchema.TableByID(tid)
+			if !ok {
+				// Ignore this table because it may have been dropped.
```
time-and-fate, line 206: This line can be removed because dropping this table also wouldn't make this case happen.

time-and-fate (approved):
1. A typo in filename `statistics/handle/handletest/statslock/stats_lcok_test.go`. Seems it has been there for a long time.
2. Personally, I think it'll be better if we extract a `query_lock.go` from `lock_stats.go`. Feel free to update it or keep it unchanged.

hi-rustin:
> 1. A typo in filename `statistics/handle/handletest/statslock/stats_lcok_test.go`. Seems it has been there for a long time.
> 2. Personally, I think it'll be better if we extract something like `query_lock.go` from `lock_stats.go`. Feel free to update it or keep it unchanged.

I will deal with them in my next PR.
Thanks for your review!  💚 💙 💜 💛

PR #46559, opened 2023-08-31
planner: introduce a new empirical rule into the Skyline pruning

### LINKED ISSUE #45132: planner: introduce a configurable rule to adjust the optimizer's tendency between `large IndexLookup` and `large FullScan` to make index selection more stable
## Enhancement

The current optimizer's cost model tends to avoid using `large IndexLookup`, because it may trigger massive requests, which is resource-consuming and may slow down the whole system.

But this tendency is not always correct, it can lead to some unreasonable `large FullScan` plans:

```
create table t (a int, b int, key(a));
insert into t values (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1);
insert into t values (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1);
insert into t values (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1);
insert into t values (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1);
insert into t values (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1);
insert into t values (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1);
insert into t values (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1);
insert into t values (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1);
insert into t values (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1);
insert into t values (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1);
insert into t values (2, 2);  -- 100 : 1

insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
analyze table t;

mysql> explain select * from t where a=2;
+-------------------------+------------+-----------+---------------+------------------+
| id                      | estRows    | task      | access object | operator info    |
+-------------------------+------------+-----------+---------------+------------------+
| TableReader_7           | 16305.14   | root      |               | data:Selection_6 |
| └─Selection_6           | 16305.14   | cop[tikv] |               | eq(test.t.a, 2)  |
|   └─TableFullScan_5     | 1654784.00 | cop[tikv] | table:t       | keep order:false |
+-------------------------+------------+-----------+---------------+------------------+
3 rows in set (0.00 sec)

mysql> explain select /*+ use_index(t, a) */ * from t where a=2;
+-------------------------------+----------+-----------+---------------------+-------------------------------+
| id                            | estRows  | task      | access object       | operator info                 |
+-------------------------------+----------+-----------+---------------------+-------------------------------+
| IndexLookUp_7                 | 16305.14 | root      |                     |                               |
| ├─IndexRangeScan_5(Build)     | 16305.14 | cop[tikv] | table:t, index:a(a) | range:[2,2], keep order:false |
| └─TableRowIDScan_6(Probe)     | 16305.14 | cop[tikv] | table:t             | keep order:false              |
+-------------------------------+----------+-----------+---------------------+-------------------------------+
3 rows in set (0.01 sec)
```

In the case above, the optimizer selects a `1654784-row FullScan` instead of a `16305-row IndexLookup`.
We believe in most cases `16305-row IndexLookup` should be better than `1654784-row FullScan`, but the optimizer tends to use `FullScan` here.

To make index selection more stable, we decided to introduce a rule to adjust this tendency.
This rule is based on the `estRows` of `FullScan` and `IndexLookup`, if their ratio is larger than a threshold (e.g. `FullScan-Rows / IndexLookup-Rows > 100`, we bypass the cost model and use `IndexLookup` directly.

qw4990:

Issue Number: close #45132

Problem Summary: planner: introduce a new empirical rule into the Skyline pruning

Add a new empirical rule to the Skyline pruning: If x's range row count is significantly lower than y's, for example, 1000 times, we think x is better than y and prune y from the candidate.

Use fixcontrol to control this behavior.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/planner/core/casetest/integration_test.go
+++ b/planner/core/casetest/integration_test.go
@@ -1578,6 +1578,38 @@ func TestIndexJoinRangeFallback(t *testing.T) {
 	}
 }
 
+func TestFixControl45132(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec(`use test`)
+	tk.MustExec(`create table t (a int, b int, key(a))`)
+	for i := 0; i < 10; i++ {
+		tk.MustExec(`insert into t values (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1)`)
+	}
+	tk.MustExec(`insert into t values (2, 2)`) // count(1) : count(2) == 100 : 1
+	for i := 0; i < 7; i++ {
+		tk.MustExec(`insert into t select * from t`)
+	}
+	tk.MustExec(`analyze table t`)
+	// the cost model prefers to use TableScan instead of IndexLookup to avoid double requests.
+	tk.MustQuery(`explain select * from t where a=2`).Check(testkit.Rows(
+		`TableReader_7 128.00 root  data:Selection_6`,
+		`└─Selection_6 128.00 cop[tikv]  eq(test.t.a, 2)`,
+		`  └─TableFullScan_5 12928.00 cop[tikv] table:t keep order:false`))
+
+	tk.MustExec(`set @@tidb_opt_fix_control = "45132:99"`)
+	tk.MustQuery(`explain select * from t where a=2`).Check(testkit.Rows(
+		`IndexLookUp_7 128.00 root  `,
+		`├─IndexRangeScan_5(Build) 128.00 cop[tikv] table:t, index:a(a) range:[2,2], keep order:false`,
+		`└─TableRowIDScan_6(Probe) 128.00 cop[tikv] table:t keep order:false`))
+
+	tk.MustExec(`set @@tidb_opt_fix_control = "45132:500"`)
+	tk.MustQuery(`explain select * from t where a=2`).Check(testkit.Rows(
+		`TableReader_7 128.00 root  data:Selection_6`,
+		`└─Selection_6 128.00 cop[tikv]  eq(test.t.a, 2)`,
+		`  └─TableFullScan_5 12928.00 cop[tikv] table:t keep order:false`))
+}
+
 func TestFixControl44262(t *testing.T) {
 	store := testkit.CreateMockStore(t)
 	tk := testkit.NewTestKit(t, store)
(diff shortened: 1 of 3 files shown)


planner/core/casetest/integration_test.go
```diff
@@ -1578,6 +1578,44 @@ func TestIndexJoinRangeFallback(t *testing.T) {
 	}
 }
 
+func TestFixControl45132(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec(`use test`)
+	tk.MustExec(`create table t (a int, b int, key(a))`)
+	for i := 0; i < 10; i++ {
+		tk.MustExec(`insert into t values (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1), (1, 1)`)
```
hawkingrei, line 1587: It can merge one SQL. you can reference the code from executor/test/indexmergereadtest/index_merge_reader_test.go:1201

hawkingrei (approved):

planner/core/casetest/integration_test.go
qw4990: updated

time-and-fate (approved):

PR #45893, opened 2023-08-08
planner, variable: add test and ga the SET_VAR hint

winoros:
### What problem does this PR solve?

Issue Number: close https://github.com/pingcap/tidb/issues/45892

Problem Summary:

### What is changed and how it works?

Adding tests to test that the hint can take effect. And the vars can be restored after the SQL is executed.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #45538, opened 2023-07-24
planner: support `no_hash_join` hint on optimizer

qw4990:
### What problem does this PR solve?

Issue Number: ref #45520

Problem Summary: planner: support `no_hash_join` hint on optimizer

### What is changed and how it works?

planner: support `no_hash_join` hint on optimizer

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #46575, opened 2023-08-31
planner:  variable `tidb_opt_enable_hash_join` to skip hash join

### LINKED ISSUE #46695: Proposal: introduce system variable to ignore hash join in TiDB
### Background 
* For most OLTP queries, hash join is not selected as it normally takes longer to fetch very few rows. 
* MySQL doesn’t support hash join before MySQL 8.0.
* In 8.0 or higher, MySQL provides an optimizer switch () to disable in session/global level. 
* There is undocumented parameter (_hash_join_enabled) in Oracle to disable hash join. 
 
It makes sense we introduce the mechanism to control this optimizer behavior at session or cluster level. 

# User case
* As an application developer, I’m very confident the queries don’t benefit from hash join. However, at times, bad plans were generated in which a hash join was selected. I will config the connection pool, and initialize the connections with  “set tidb_opt_enable_hash_join=no”. 
* As a DBA, I’m aware the application is OLTP workload, which works properly in MySQL ~5.7.  I’m very sure the SQLs don’t need hash join.  I’ve seen cases in the past. bad plans were generated in which a hash join was selected. After confirming with application developers, I will disable hash join in cluster level with “set global tidb_opt_enable_hash_join=no” 

#### Introduce variable to disable hash join in session/global 
* Name: tidb_opt_enable_hash_join 
* Scope:  session / global 
* Applies to set_var(): YES 
* Default value: YES 
* Possible values: YES /NO 

#### Behavior 

* With tidb_opt_enable_hash_join=NO, Optimizer ignores hash join when generating execution plan. 
* Optimizer hints overrides the behavior of the variable.  If tidb_opt_enable_hash_join is set to “NO”, hash join is still enforced for the queries with hint “HASH_JOIN()”, “HASH_JOIN_PROBE()” or “HASH_JOIN_BUILD()”.  In this case,  lower level control makes exceptions possible. (The same behavior as MySQL) 
* It’s possible queries come with hints which could lead to no join type available.  For example
```
set tidb_opt_enable_hash_join=NO 
select /*+ no_merge_join(), no_index_join(), no_index_hash_join()/ … 
```

It’s not acceptable SQL fails caused by optimizer variable changes, as the scope of variables is wide.   **We have to avoid no available plan because of this variable.** 

In this case, we shall keep the hints working and  ignore the behavior change caused by this variable.  Query still generates a plan with hash_join, and appends messages to warnings. 
```
variable tidb_opt_enable_hash_join=NO is ignored due to no join type available.
```

coderplay:
This PR adds a global/session knob to disable hash join, in order to prevent query plan regression. 
Use cases:

* As an application developer, I’m very confident the queries don’t benefit from hash join. However, at times, bad plans were generated in which a hash join was selected. I will config the connection pool, and initialize the connections with  `set tidb_opt_enable_hash_join=no`.
* As a DBA, I’m aware the application is OLTP workload, which works properly in MySQL ~5.7.  I’m very sure the SQLs don’t need hash join.  I’ve seen cases in the past. bad plans were generated in which a hash join was selected. After confirming with application developers, I will disable hash join in cluster level with `set global tidb_opt_enable_hash_join=off`

Issue Number: close #46695

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Add a new global/session variable `tidb_opt_enable_hash_join` to control whether hash join is enabled or not.
```


planner/core/casetest/rule/testdata/join_reorder_suite_out.json
```diff
@@ -782,6 +782,167 @@
...
+          "HashJoin 100000000.00 root  CARTESIAN inner join",
+          "├─TableReader(Build) 10000.00 root  data:TableFullScan",
+          "│ └─TableFullScan 10000.00 cop[tikv] table:t2 keep order:false, stats:pseudo",
+          "└─TableReader(Probe) 10000.00 root  data:TableFullScan",
+          "  └─TableFullScan 10000.00 cop[tikv] table:t1 keep order:false, stats:pseudo"
+        ],
+        "Warning": null
+      },
+      {
+        "SQL": "select /*+ leading(t4, t3, t2, t1) */ * from t1, t2, t3, t4",
```
coderplay, line 909: Why hashjoin can't be disabled in this case?

qw4990: I'll take a look at this.

qw4990:
@coderplay Great, thanks for your contribution, I'll review it soon.

planner/core/casetest/rule/testdata/join_reorder_suite_out.json
qw4990: `no` is not a valid value for a bool variable in TiDB, only `on`, `off`, `1`, `0`.

planner/core/exhaust_physical_plans.go
```diff
@@ -450,11 +454,10 @@ func (p *LogicalJoin) getHashJoins(prop *property.PhysicalProperty) (joins []Phy
...
-	forced = (p.preferJoinType&preferHashJoin > 0) || forceLeftToBuild || forceRightToBuild
-	noHashJoin := (p.preferJoinType & preferNoHashJoin) > 0
-	if !forced && noHashJoin {
+	forced = forceLeftToBuild || forceRightToBuild
+	if !forced && p.shouldSkipHashJoin() {
 		return nil, false
-	} else if forced && noHashJoin {
+	} else if forced && p.shouldSkipHashJoin() {
 		p.SCtx().GetSessionVars().StmtCtx.AppendWarning(ErrInternal.GenWithStack(
 			"Some HASH_JOIN and NO_HASH_JOIN hints conflict, NO_HASH_JOIN is ignored"))
```
qw4990, line 462: The warning message should be updated, in the case below, the warning message is not accurate:
```
mysql> set tidb_opt_enable_hash_join=off;

mysql> explain select /*+ hash_join_build(t1) */ * from t1, t2;
+-----------------------------+--------------+-----------+---------------+--------------------------------+
| id                          | estRows      | task      | access object | operator info                  |
+-----------------------------+--------------+-----------+---------------+--------------------------------+
| HashJoin_8                  | 100000000.00 | root      |               | CARTESIAN inner join           |
| ├─TableReader_10(Build)     | 10000.00     | root      |               | data:TableFullScan_9           |
| │ └─TableFullScan_9         | 10000.00     | cop[tikv] | table:t1      | keep order:false, stats:pseudo |
| └─TableReader_12(Probe)     | 10000.00     | root      |               | data:TableFullScan_11          |
|   └─TableFullScan_11        | 10000.00     | cop[tikv] | table:t2      | keep order:false, stats:pseudo |
+-----------------------------+--------------+-----------+---------------+--------------------------------+
5 rows in set, 1 warning (0.00 sec)

mysql> show warnings;
+---------+------+-------------------------------------------------------------------------+
| Level   | Code | Message                                                                 |
+---------+------+-------------------------------------------------------------------------+
| Warning | 1815 | Some HASH_JOIN and NO_HASH_JOIN hints conflict, NO_HASH_JOIN is ignored |
+---------+------+-------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

planner/core/casetest/rule/testdata/join_reorder_suite_out.json
coderplay: @qw4990  My original meaning was that even in your PR, the hash join is not disabled when applying a `leading` hint with a `no_hash_join_hint`.  

see my comment on https://github.com/pingcap/tidb/pull/45538

chrysan:
Do you want this variable work in set_var hint? If yes, what about adding one test for it? https://github.com/pingcap/tidb/pull/45893/files#diff-c543ebb4dd3e432d5b6f6d276d6f7f4d740d366cd199f3ce955d627b54035449R29-R101

planner/core/casetest/rule/testdata/join_reorder_suite_out.json
qw4990: In my case, the `no_hash_join` only works for `t2` and `t3`. The top Hash Join is applied on `t1` and `result of (t2, t3, t4)`, so `no_hash_join(t2)` cannot take effect on it.
image

planner/core/exhaust_physical_plans.go
```diff
@@ -450,11 +454,10 @@ func (p *LogicalJoin) getHashJoins(prop *property.PhysicalProperty) (joins []Phy
...
 	}
 
-	forced = (p.preferJoinType&preferHashJoin > 0) || forceLeftToBuild || forceRightToBuild
-	noHashJoin := (p.preferJoinType & preferNoHashJoin) > 0
-	if !forced && noHashJoin {
+	forced = forceLeftToBuild || forceRightToBuild
+	if !forced && p.shouldSkipHashJoin() {
 		return nil, false
-	} else if forced && noHashJoin {
+	} else if forced && p.shouldSkipHashJoin() {
```
qw4990, line 460: Another problem that needs to solve is to avoid `no-plan` error. 
In the case below, after disabling `Hash Join`, the optimizer can't find a plan for this query.
```
mysql> explain select /*+ no_merge_join(t1) */ * from t1, t2 where t1.a=t2.a;
+------------------------------+----------+-----------+---------------+----------------------------------------------+
| id                           | estRows  | task      | access object | operator info                                |
+------------------------------+----------+-----------+---------------+----------------------------------------------+
| HashJoin_8                   | 12487.50 | root      |               | inner join, equal:[eq(test.t1.a, test.t2.a)] |
| ├─TableReader_15(Build)      | 9990.00  | root      |               | data:Selection_14                            |
| │ └─Selection_14             | 9990.00  | cop[tikv] |               | not(isnull(test.t2.a))                       |
| │   └─TableFullScan_13       | 10000.00 | cop[tikv] | table:t2      | keep order:false, stats:pseudo               |
| └─TableReader_12(Probe)      | 9990.00  | root      |               | data:Selection_11                            |
|   └─Selection_11             | 9990.00  | cop[tikv] |               | not(isnull(test.t1.a))                       |
|     └─TableFullScan_10       | 10000.00 | cop[tikv] | table:t1      | keep order:false, stats:pseudo               |
+------------------------------+----------+-----------+---------------+----------------------------------------------+
7 rows in set (0.00 sec)

mysql> set tidb_opt_enable_hash_join=off;

mysql> explain select /*+ no_merge_join(t1) */ * from t1, t2 where t1.a=t2.a;
ERROR 1815 (HY000): Internal : Can't find a proper physical plan for this query
```

Can we implement this in `LogicalJoin.exhaustPhysicalPlans` where we can know whether there is any other possible join plans?
Then to implement it in this way:
```
func (p *LogicalJoin) exhaustPhysicalPlans(...) {
    ...
	if noHashJoin && len(otherJoins) == 0 {
        hashJoins := p.getHashJoins(prop)
        warning("no hash join can not be applied")
        return hashJoins
    } else {
        ...
    }
}
```

coderplay:
> Do you want this variable work in set_var hint? If yes, what about adding one test for it? https://github.com/pingcap/tidb/pull/45893/files#diff-c543ebb4dd3e432d5b6f6d276d6f7f4d740d366cd199f3ce955d627b54035449R29-R101

As we discussed offline, we don't want because it's supposed to be effective global wise.

planner/core/exhaust_physical_plans.go
coderplay: How did you manage to load t1, t2 data into above tidb cluster?

coderplay:
BTW, how I can run the unit test? I follow the instructions here: https://pingcap.github.io/tidb-dev-guide/get-started/setup-an-ide.html

But got an error : 

```
/Users/min.zhou/.gvm/gos/go1.21/bin/go test -v -race --tags=intest,deadlock ./planner/core/... -check.f TestEnforceMPP -check.vv
go: downloading github.com/pingcap/kvproto v0.0.0-20230904082117-ecdbf1f8c130
go: downloading github.com/tikv/client-go/v2 v2.0.8-0.20230919031511-be2b4c78a910
go: downloading github.com/tikv/pd/client v0.0.0-20230912103610-2f57a9f050eb
go: downloading golang.org/x/text v0.13.0
go: downloading github.com/klauspost/compress v1.17.0
go: downloading github.com/dolthub/swiss v0.2.1
go: downloading golang.org/x/sys v0.12.0
go: downloading golang.org/x/net v0.15.0
go: downloading golang.org/x/term v0.12.0
go: downloading golang.org/x/crypto v0.13.0
go: downloading github.com/dolthub/maphash v0.1.0
flag provided but not defined: -check.f
Usage of /var/folders/0m/552hpx955lg2p0hx71wsz0_h0000gp/T/go-build968870319/b001/core.test:
  -alsologtostderr
```

planner/core/exhaust_physical_plans.go
qw4990: OK, then we'd better add this into the user doc later on just like what we've done for hints (https://docs.pingcap.com/tidb/dev/optimizer-hints#using-hints-causes-the-cant-find-a-proper-physical-plan-for-this-query-error).

planner/core/exhaust_physical_plans.go
qw4990: Emm no data in them. I just create 2 empty tables and then use `stats:pseudo`.

coderplay:
--- PASS: TestNoHashJoinHint (0.63s)
PASS
ok      github.com/pingcap/tidb/planner/core/casetest/rule      0.682s

--- PASS: TestOptEnableHashJoin (0.65s)
PASS
ok      github.com/pingcap/tidb/planner/core/casetest/rule      0.701s

planner/core/exhaust_physical_plans.go
coderplay: I am creating another PR for the doc.

coderplay:
@qw4990 Just addressed all the feedbacks, please let me know if you have any suggestions. Thx

planner/core/exhaust_physical_plans.go
coderplay: see https://github.com/pingcap/docs/pull/14905

qw4990 (approved):
LGTM!

qw4990:
@time-and-fate Please take a look at this PR at your convenience

time-and-fate (approved):

planner/core/exhaust_physical_plans.go
```diff
@@ -391,6 +391,10 @@ var ForceUseOuterBuild4Test = atomic.NewBool(false)
 // TODO: use hint and remove this variable
 var ForcedHashLeftJoin4Test = atomic.NewBool(false)
 
+func (p *LogicalJoin) shouldSkipHashJoin() bool {
+	return (p.preferJoinType&preferNoHashJoin) > 0 || (p.SCtx().GetSessionVars().EnableHashJoin == false)
```
qw4990, line 395:
```suggestion
	return (p.preferJoinType&preferNoHashJoin) > 0 || (!p.SCtx().GetSessionVars().EnableHashJoin)
```

qw4990:
And do we need to pick this PR to v7.1? @coderplay @songrijie

planner/core/optimizer.go
```diff
@@ -162,6 +162,7 @@ type logicalOptRule interface {
 func BuildLogicalPlanForTest(ctx context.Context, sctx sessionctx.Context, node ast.Node, infoSchema infoschema.InfoSchema) (Plan, types.NameSlice, error) {
 	sctx.GetSessionVars().PlanID.Store(0)
 	sctx.GetSessionVars().PlanColumnID.Store(0)
+	sctx.GetSessionVars().EnableHashJoin = true
```
qw4990, line 165: Too many CI failures due to the default value of this variable:
image

Can we change the name of the variable from `EnableHashJoin` to `DisableHashJoin` and set it as `false` by default?

1) for most users, `HashJoin` should be allowed, so using `false` as the default value seems safer.
2) for some tests, the test framework won't use `DefEnableHashJoin` to initialize `sctx.EnableHashJoin`, instead it uses the default value of `Bool` in `Golang`, which is false; If we use `EnableHashJoin`, there will be plenty of test failures, changing it to `DisableHashJoin` can fix this.

qw4990: After discussing with PM, we decided to keep it unchanged as `EnableHashJoin`.

qw4990: > Too many CI failures... For safety, how about using another solution, which is using `DisableHashJoin` internally, and exposing `enable_hash_join` to users.

@coderplay I've pushed some code to your PR directly (tomorrow is the code freeze deadline of v7.4 ...) by using this approach to solve CI failures.

qw4990 (approved):

qw4990:
@easonn7 PTAL

planner/core/optimizer.go
coderplay: Thanks for fixing those @qw4990 .  How can I detect those issues locally before submitting this PR?

PR #46626, opened 2023-09-04
executor: return query watch id

CabinfeverB:

Issue Number: ref #43691

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


CabinfeverB:
cc @nolouch @glorv

domain/runaway.go
```diff
@@ -176,8 +176,18 @@ func (do *Domain) runawayWatchSyncLoop() {
...
-	return do.handleRunawayWatch(record)
+func (do *Domain) AddRunawayWatch(record *resourcegroup.QuarantineRecord) (uint64, error) {
+	if err := do.handleRunawayWatch(record); err != nil {
+		return 0, err
+	}
+	if err := do.TryToUpdateRunawayWatch(); err != nil {
+		return 0, err
+	}
+	r := do.runawayManager.GetWatchByKey(record.GetRecordKey())
+	if r != nil {
```
glorv, line 187: Here goes the same issue as before:
- If `handleRunawayWatch` return success, it means the record is inserted successfully, so it's like runaway manager can see this record within one update interval. But It's possible `TryToUpdateRunawayWatch` failed due to arbitrary errors(e.g. network error), thus the function returns an error but the record is actually takes effect.
- In another case, since table `tidb_runaway_watch`'s primary key is the auto_increment id, it's possible that two concurrent insert with the same recordkey can overwrite each other when put into the runaway manager, so user `select LAST_INSERT_ID()` here to get the id may be the better choice.

CabinfeverB: Get it. I think we can retry when `select LAST_INSERT_ID()` fails and tell the user that the insertion is successful but getting id fails.

CabinfeverB:
PTAL @nolouch @glorv

pkg/domain/runaway.go
```diff
@@ -308,7 +306,22 @@ func (do *Domain) handleRunawayWatch(record *resourcegroup.QuarantineRecord) err
...
+			rows, err1 := sqlexec.DrainRecordSet(ctx, rs, 1)
+			err2 := rs.Close()
+			if err1 == nil && err2 == nil && len(rows) == 1 {
+				return rows[0].GetUint64(0), nil
+			}
+		}
+		logutil.BgLogger().Warn("failed to get last insert id when adding runaway watch", zap.Error(err))
+		time.Sleep(time.Millisecond * time.Duration(retry*100))
+	}
+	return 0, errors.Errorf("An error occurred while getting the ID of the newly added watch record. Try querying information_schema.runaway_watches later")
```
glorv, line 324: better wrap the `SELECT LAST_INSERT_ID` error into this return error

glorv (approved):

nolouch (approved):

CabinfeverB:
Can u help approve this PR @zimulala @lance6716 @tangenta @wjhuang2016. Thx~

CabinfeverB:
Hi, @zimulala , could u take a look this PR?

pkg/domain/runaway.go
```diff
@@ -308,7 +306,28 @@ func (do *Domain) handleRunawayWatch(record *resourcegroup.QuarantineRecord) err
...
+			rs.Close()
+			if err == nil {
+				if len(rows) == 1 {
+					return rows[0].GetUint64(0), nil
+				}
+				err = errors.Errorf("unexpected result length: %d", len(rows))
+			}
+		}
+		logutil.BgLogger().Warn("failed to get last insert id when adding runaway watch", zap.Error(err))
+		time.Sleep(time.Millisecond * time.Duration(retry*100))
```
lance6716, line 328: Better to use `select` to see `do.exit` is closed while sleeping.

pkg/domain/runaway.go
```diff
@@ -308,7 +306,28 @@ func (do *Domain) handleRunawayWatch(record *resourcegroup.QuarantineRecord) err
...
 	sql, params := record.GenInsertionStmt()
 	_, err = exec.ExecuteInternal(ctx, sql, params...)
-	return err
+	if err != nil {
+		return 0, err
+	}
+	for retry := 1; retry <= maxIDRetries; retry++ {
+		var rs sqlexec.RecordSet
+		rs, err = exec.ExecuteInternal(ctx, `SELECT LAST_INSERT_ID();`)
+		if err == nil {
```
lance6716, line 315: just a coding style suggestion, we can use 
```
if err != nil {
  continue
}
```

to reduce the indentation. Also for below if-check

CabinfeverB: If we use `continue `, how do I gracefully write the logic to wait for interval between retries?

pkg/domain/runaway.go
CabinfeverB: Got it, thx~

pkg/domain/runaway.go
lance6716: maybe move the sleep to the beginning of loop (skip the sleep for the first try), or use goto 😂

pkg/domain/runaway.go
```diff
@@ -325,7 +325,12 @@ func (do *Domain) AddRunawayWatch(record *resourcegroup.QuarantineRecord) (uint6
 			}
 		}
 		logutil.BgLogger().Warn("failed to get last insert id when adding runaway watch", zap.Error(err))
-		time.Sleep(time.Millisecond * time.Duration(retry*100))
+		select {
```
lance6716, line 328: sorry I mean 
select {
case <-do.exit
case <-time.After
}

lance6716 (approved):

PR #46676, opened 2023-09-05
executor: record history stats when estimation is not accurate 

yibin87:

Issue Number: ref #46773 

Problem Summary: 
Recognize "Selection + TableFullScan" and "TableFullScan" pattern in physical plan after statement executed successfully. Compare actual row count with estimate row count, and record the history stats when max(est, actual) > min(est, actual) * Threshold. Since ddl changes for stats_meta is not ready, just record in log.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
Create a new table and insert some data. Set the system var and execute "explain analyze select * from t where xxxxxx", check the log info.
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


executor/adapter.go
```diff
@@ -1492,6 +1497,52 @@ func resetCTEStorageMap(se sessionctx.Context) error {
...
+			if _, ok := x.Children()[0].(*plannercore.PhysicalTableScan); ok {
+				checkAndRecordHistoryStats(stmtCtx, sql, x)
+			}
+		}
+	}
+}
+
+func checkAndRecordHistoryStats(stmtCtx *stmtctx.StatementContext, sql string, x *plannercore.PhysicalSelection) {
+	estRows := plannercore.GetPhysicalPlanEstRowCount(x)
+	actualRows := GetResultRowsCountForCopStats(stmtCtx, x.ID())
```
ghazalfamilyusa, line 1530: This function can fail (-1 return). Can you also bail out in that case?

executor/adapter.go
```diff
@@ -1700,6 +1751,21 @@ func GetResultRowsCount(stmtCtx *stmtctx.StatementContext, p plannercore.Plan) i
 	return rootStats.GetActRows()
 }
 
+// GetResultRowsCountForCopStats gets the count of the statement result rows of cop stats
+// Return -1 if no cop stats exists
+func GetResultRowsCountForCopStats(stmtCtx *stmtctx.StatementContext, planID int) int64 {
```
ghazalfamilyusa, line 1756: It would be cleaner if you return pass/fail from the function.

executor/adapter.go
```diff
@@ -1492,6 +1497,52 @@ func resetCTEStorageMap(se sessionctx.Context) error {
...
+		}
+		if x, isSelect := pp.(*plannercore.PhysicalSelection); isSelect {
+			if _, ok := x.Children()[0].(*plannercore.PhysicalTableScan); ok {
+				checkAndRecordHistoryStats(stmtCtx, sql, x)
+			}
+		}
+	}
+}
+
+func checkAndRecordHistoryStats(stmtCtx *stmtctx.StatementContext, sql string, x *plannercore.PhysicalSelection) {
```
ghazalfamilyusa, line 1528: Can also cover full table scans as well? May be in next PR.

expression/util.go
```diff
@@ -1794,3 +1798,51 @@ func ExprsToStringsForDisplay(exprs []Expression) []string {
 	}
 	return strs
 }
+
+// SortAndGenMD5HashForCNFExprs sorts cnf exprs by their hash codes
```
ghazalfamilyusa, line 1802: Does this sort along CNF and DNF recursively?
Initially, sorting along CNF at the outermost level is good enough.

ghazalfamilyusa (changes requested):

ghazalfamilyusa:
Is there and end to end test?

executor/adapter.go
```diff
@@ -1492,6 +1497,52 @@ func resetCTEStorageMap(se sessionctx.Context) error {
 	return nil
 }
 
+// RecordHistoryStats is used to record history stats that is more accurate than estimation
+// Currently, only support "Selection + TableScan" pattern
+func (a *ExecStmt) RecordHistoryStats() {
+	sessVars := a.Ctx.GetSessionVars()
+	if sessVars.InRestrictedSQL || !sessVars.StmtCtx.InSelectStmt {
```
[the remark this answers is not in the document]
yibin87: Sure, will add it.

executor/adapter.go
```diff
@@ -1492,6 +1497,52 @@ func resetCTEStorageMap(se sessionctx.Context) error {
...
+				checkAndRecordHistoryStats(stmtCtx, sql, x)
+			}
+		}
+	}
+}
+
+func checkAndRecordHistoryStats(stmtCtx *stmtctx.StatementContext, sql string, x *plannercore.PhysicalSelection) {
+	estRows := plannercore.GetPhysicalPlanEstRowCount(x)
+	actualRows := GetResultRowsCountForCopStats(stmtCtx, x.ID())
+	const Threshold float64 = 1.0
```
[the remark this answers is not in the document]
yibin87: ok, will change it.

executor/adapter.go
```diff
@@ -1492,6 +1497,52 @@ func resetCTEStorageMap(se sessionctx.Context) error {
...
+	if actualRows >= 0 && math.Abs(float64(actualRows)-estRows) > 2*Threshold {
```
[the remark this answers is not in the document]
yibin87: Yes, I think it is hard to figure out a reasonable check for all situations: when abs(actualRows) is small, differences seems resonable; when abs(actualRows) is big, the ratio one makes more sense. However, I'm not sure how to define "small" or "big" here.

executor/adapter.go
yibin87: Get it~

executor/adapter.go
yibin87: Do you mean table scan operator without any filters?

expression/util.go
yibin87: Now it only sorts standard CNF expression:   (A or B or C) and (D or E or F) and (G or H or I), **and** expression inside DNF will not be sorted recursively.

yibin87:
> Is there and end to end test?

Not yet, only do some manually end to end test to check the log message, no automatic tests.

executor/adapter.go
yibin87: ok.

expression/util.go
yibin87: It is more general to sort the expression along CNF and DNF recursively, I'll change it.

executor/adapter.go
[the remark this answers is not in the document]
ghazalfamilyusa: How about tidb_optimizer_history_stats?

executor/adapter.go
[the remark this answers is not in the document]
ghazalfamilyusa: I think ratio is important for all cases in the optimizer for small or big actual. I would suggest to just use ratio >= threshold

executor/adapter.go
ghazalfamilyusa: Yes. In this case, we may have no or stalled analyze.

executor/adapter.go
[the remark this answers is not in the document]
yibin87: OK，I'll change it.

expression/util.go
```diff
@@ -1794,3 +1798,62 @@ func ExprsToStringsForDisplay(exprs []Expression) []string {
...
+				logutil.BgLogger().Warn("Expression is too complex to record history stat.",
+					zap.Int("DNFExprLen", len(dnfExprs)))
+				return false, 0
+			}
+			dnfHashCodes := make([][]byte, 0, len(dnfExprs))
+			for _, dnf := range dnfExprs {
+				dnfHashCode := dnf.HashCode(sc)
+				dnfHashCodes = checkHashLengthAndAppend(dnfHashCode, dnfHashCodes)
+			}
+			sort.Slice(dnfHashCodes, func(i, j int) bool { return bytes.Compare(dnfHashCodes[i], dnfHashCodes[j]) < 0 })
```
XuHuaiyu, line 1832: What will be compared, the value or pointer of the byte slice?

expression/constant.go
```diff
@@ -112,8 +112,9 @@ type Constant struct {
 	DeferredExpr Expression
 	// ParamMarker holds param index inside sessionVars.PreparedParams.
 	// It's only used to reference a user variable provided in the `EXECUTE` statement or `COM_EXECUTE` binary protocol.
-	ParamMarker *ParamMarker
-	hashcode    []byte
+	ParamMarker               *ParamMarker
+	hashcode                  []byte
+	ignoreParamMarkerHashCode []byte // use ConstantFlag+EncodedValue even if ParamMarker is not nil
```
XuHuaiyu, line 117: Why do we need a new one, can we reuse the `hashcode`?

executor/adapter.go
```diff
@@ -1492,6 +1502,92 @@ func resetCTEStorageMap(se sessionctx.Context) error {
...
+func needRecordHistoryStats(stmtCtx *stmtctx.StatementContext, x plannercore.PhysicalPlan) (bool, float64, int64) {
+	estRows := plannercore.GetPhysicalPlanEstRowCount(x)
+	hasCopStats, actualRows := GetResultRowsCountForCopStats(stmtCtx, x.ID())
+	if !hasCopStats {
+		return false, estRows, actualRows
+	}
+	actualRowsFloat := float64(actualRows)
+	max := math.Max(actualRowsFloat, estRows)
+	min := math.Min(actualRowsFloat, estRows)
+	return max > min*HistoryStatsUpdateThreshold, estRows, actualRows
```
XuHuaiyu, line 1588: What if the actRows is 2 and estRows is 1, do we still need to update the history stats?

executor/adapter.go
```diff
@@ -1370,6 +1375,11 @@ func (a *ExecStmt) FinishExecuteStmt(txnTS uint64, err error, hasMoreResults boo
 		// but we set it again in case we missed some code paths.
 		sessVars.StmtCtx.SetPlan(a.Plan)
 	}
+
+	/// Triage for historical stats
+	if succ {
+		a.RecordHistoryStats()
```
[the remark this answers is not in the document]
yibin87: I think it won't. Because the record work will be done only when estimation is not accurate. For benchmarks, when "warm up", these estimation should be adjusted if "optimizer_history_stats" is on.

expression/constant.go
yibin87: Revert the change now. It is somewhat like the "canonicalhashcode" in scalar function. Different goals lead to different hash code. However, also need to consider add it in scalar function level to avoid the cache mechanism, so revert it currently.

executor/adapter.go
yibin87: Good question, according to Ahmad, even if row count is very small, he believes the ratio value is meaningful for optimization.

expression/util.go
yibin87: For [][]byte, I'd like to compare byte[i] slice with byte[j] slice, I'll check the grammar here to ensure it.

expression/util.go
```diff
@@ -1794,3 +1798,57 @@ func ExprsToStringsForDisplay(exprs []Expression) []string {
...
+
+			flatDnfHashCode := bytes.Join(dnfHashCodes, nil)
+			cnfHashCodes = append(cnfHashCodes, flatDnfHashCode)
```
windtalker, line 1830: Why don't need to call `checkHashLengthAndAppend`?

executor/adapter.go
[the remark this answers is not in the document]
yibin87: I agree, when the write path is completed, I'll do some benchmark testing.

executor/adapter.go
```diff
@@ -1492,6 +1502,92 @@ func resetCTEStorageMap(se sessionctx.Context) error {
...
 
+// RecordHistoryStats is used to record history stats that is more accurate than estimation
+// Currently, only support "Selection + TableScan" pattern
+func (a *ExecStmt) RecordHistoryStats() {
+	sessVars := a.Ctx.GetSessionVars()
+	if !sessVars.EnableOptimizerHistoryStats || sessVars.InRestrictedSQL || !sessVars.StmtCtx.InSelectStmt {
+		return
+	}
+
+	sql := FormatSQL(a.GetTextToLog(true))()
```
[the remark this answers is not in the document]
yibin87: Yes, will remove one.

executor/adapter.go
```diff
@@ -1492,6 +1502,92 @@ func resetCTEStorageMap(se sessionctx.Context) error {
...
+				if !needRecord {
+					break
+				}
+				succ, hashCode := genHistoryStatsHashCodeForSelection(stmtCtx, x)
+				if !succ {
+					break
+				}
+				recordHistoryStats(sql, tableScan.Table.ID, x.ID(), false, hashCode, estRows, actualRows)
+			}
+		case *plannercore.PhysicalTableScan:
```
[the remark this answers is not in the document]
yibin87: Good points, I believe we can simply check and disable case 1 and case 3. However, case 2 is challenging, I'll do some research and discuss it with Ahmad.

expression/util.go
yibin87: Yes, it reminds me that I need to restrict the memory usage more carefully here, I'll update it.

ghazalfamilyusa (approved):

PR #46718, opened 2023-09-06
util/misc: pass the array field in column info to TiKV

### LINKED ISSUE #46717: Array field should be passed to the `columnInfo`
The following sql will fail:

```
create table t (pk varchar(4) primary key clustered, j json, str varchar(255), value int, key idx((cast(j as char(100) array)), str));
insert into t values ("1", '["a"]', 'b', 1);
select * from t use index(idx) where "a" member of (j);
```

Because tikv thinks it's a normal column and try to restore the data. However, the restore data doesn't exist in the value because tidb think it's a json and don't need to restore.

A better way to fix this is to set the collation to `binary`. See https://github.com/pingcap/tidb/issues/46691. This issue should also be  fixed for backward compatibility (for existing data / indexes).

YangKeao:

Issue Number: close #46717

This PR copies the information of whether the type is an array from the `FieldType` to the `tipb.ColumnInfo` and passed it to tikv.

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code


hawkingrei:
/test all 

I have uploaded the new depedency.

bb7133 (approved):
LGTM

However, the title & description is a little misleading: it adds array type to `tipb.ColumnInfo` and pass it to TiKV, the [`ColumnInfo`](https://github.com/pingcap/tidb/blob/d3d30f5f556e826ae5ec2d25c8d8fd6b9412e72a/parser/model/model.go#L105) struct in TiDB contains array type already, please update the title & desc.

go.mod
```diff
@@ -305,4 +305,6 @@ replace (
 	github.com/dgrijalva/jwt-go => github.com/form3tech-oss/jwt-go v3.2.6-0.20210809144907-32ab6a8243d7+incompatible
 	github.com/go-ldap/ldap/v3 => github.com/YangKeao/ldap/v3 v3.4.5-0.20230421065457-369a3bab1117
 	github.com/pingcap/tidb/parser => ./parser
+
+	github.com/pingcap/tipb => github.com/YangKeao/tipb v0.0.0-20230906072756-8d43143257bf
```
YangKeao, line 309: I'll hold this PR until https://github.com/pingcap/tipb/pull/317 is merged.

YangKeao:
/hold

I realized that this PR may be wrong if the user is accessing a visible array column not for index. I'll test and try to find a better way to solve it 🤔 .

YangKeao:
/hold

I proposed another solution with less modification and is more suitable with the original multi-valued index implementation. https://github.com/pingcap/tidb/pull/46993 PTAL @bb7133 

I'll also revert the tipb's PR if we agree https://github.com/pingcap/tidb/pull/46993 is better.

YangKeao:
https://github.com/pingcap/tidb/pull/46993 has been merged. Close this one.

Issue #46656, opened 2023-09-05
ddl: attach the right resource group name in kv request for ddl/import tasks


glorv:
## Development Task

The resource control feature implements quota limiter and scheduling logic in the client-go and tikv side, it uses the resource group name in the kv request to identify the task attached resource group. 

For most foreground SQLs, we set the resource-group-name parameter from the sessionctx but for DDL/Import like task, it uses a unified worker pool so the sessionctx's resource group is not the same as the user's session. We should attach the correct resource-group-name to let resource control feature work correctly.

okJiang:
Can you tell me where `set the resource-group-name parameter from the sessionctx`? I will have a look. @glorv

PR #46756, opened 2023-09-07
rc: add resource group name to ddl

okJiang:

Issue Number: ref https://github.com/pingcap/tidb/issues/46656

Problem Summary:

as title

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

add index
image

- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


ddl/ddl_worker.go
```diff
@@ -762,10 +764,14 @@ func (w *worker) HandleDDLJobTable(d *ddlCtx, job *model.Job) (int64, error) {
 	}
 	w.setDDLLabelForTopSQL(job.ID, job.Query)
 	w.setDDLSourceForDiagnosis(job.ID, job.Type)
-	jobContext := w.jobContext(job.ID)
+	jobContext := w.jobContext(job.ID, job.ReorgMeta)
 	if tagger := w.getResourceGroupTaggerForTopSQL(job.ID); tagger != nil {
 		txn.SetOption(kv.ResourceGroupTagger, tagger)
 	}
+	if job.ReorgMeta != nil {
```
zimulala, line 771: Could we remove this logical? It seems we do it in the funcion of 
 `jobContext` on the line767(https://github.com/pingcap/tidb/pull/46756/files#diff-a185df836b8926c4afe351d4def1b56d1e282c65aabe6677c2fa2c54590cc4b2R477)

ddl/ddl_worker.go
```diff
@@ -762,10 +764,14 @@ func (w *worker) HandleDDLJobTable(d *ddlCtx, job *model.Job) (int64, error) {
...
+		txn.SetOption(kv.ResourceGroupName, job.ReorgMeta.ResourceGroupName)
+	}
```
okJiang, line 773:
```suggestion
	txn.SetOption(kv.ResourceGroupName, jobContext.resourceGroupName)
```

How about using this? line767 does not `SetOption` @zimulala

zimulala: OK.

okJiang: fixed in https://github.com/pingcap/tidb/pull/46756/commits/37d96b3217f3a30e589c4d0060978487342e1cee

ddl/backfilling.go
```diff
@@ -684,7 +684,7 @@ func (dc *ddlCtx) writePhysicalTableRecord(sessPool *sess.Pool, t table.Physical
 		}
 	})
 
-	jc := dc.jobContext(job.ID)
+	jc := dc.jobContext(job.ID, job.ReorgMeta)
```
zimulala, line 687: Why not use `reorgInfo.NewJobContext()` instead of this? In addition, how to determine whether to use 'reorgInfo.NewJobContext()' or 'dc.jobContext(job.ID, job.ReorgMeta)'

okJiang: `reorgInfo` is a concept at the higher level of DDL job, in this function, we can only get DDL job. If we can get `reorgInfo`, we prefer to call `NewJobContext` first, because this function does not need to pass additional parameters. Otherwise, call `jobContext()`.

zimulala: But this function has the argument of  `reorgInfo *reorgInfo`, so why not just use it?

ddl/ddl_api.go
```diff
@@ -4373,10 +4373,11 @@ func (d *ddl) AlterTablePartitioning(ctx sessionctx.Context, ident ast.Ident, sp
...
 		ReorgMeta: &model.DDLReorgMeta{
-			SQLMode:       ctx.GetSessionVars().SQLMode,
-			Warnings:      make(map[errors.ErrorID]*terror.Error),
-			WarningsCount: make(map[errors.ErrorID]int64),
-			Location:      &model.TimeZoneLocation{Name: tzName, Offset: tzOffset},
+			SQLMode:           ctx.GetSessionVars().SQLMode,
+			Warnings:          make(map[errors.ErrorID]*terror.Error),
+			WarningsCount:     make(map[errors.ErrorID]int64),
+			Location:          &model.TimeZoneLocation{Name: tzName, Offset: tzOffset},
+			ResourceGroupName: ctx.GetSessionVars().ResourceGroupName,
```
zimulala, line 4380: Do we need to add it to `MultiSchemaChange`'s `ReorgMeta`?

zimulala: Besides, could we add a `NewReorgMeta` to unify the assignment? This helps reduce the possibility of missing parameter assignments later on

ddl/backfilling.go
okJiang: fixed in https://github.com/pingcap/tidb/pull/46756/commits/390b696357ddf7881dceca61ccfe70c6d52ffdbd

ddl/ddl_api.go
okJiang: > Besides, could we add a NewReorgMeta to unify the assignment? This helps reduce the possibility of missing parameter assignments later on

lgtm

okJiang: fixed all in https://github.com/pingcap/tidb/pull/46756/commits/d3ad2bf7bb2f8ebb47be4fc5bea197dfa2307695

zimulala (approved):
LGTM

ddl/ddl.go
```diff
@@ -465,13 +465,19 @@ func (dc *ddlCtx) removeJobCtx(job *model.Job) {
...
 	dc.jobCtx.RLock()
 	defer dc.jobCtx.RUnlock()
+	var ctx *JobContext
 	if jobContext, exists := dc.jobCtx.jobCtxMap[jobID]; exists {
-		return jobContext
+		ctx = jobContext
+	} else {
+		ctx = NewJobContext()
+	}
+	if reorgMeta != nil && len(ctx.resourceGroupName) == 0 {
```
glorv, line 477:
```suggestion
	if reorgMeta != nil {
```
No need to check ctx's resource-group, it may be the default value `default`

glorv (approved):
LGTM

ddl/ddl.go
okJiang: Is `default` necessary? If it is, we need add a lock for it. Otherwise, it will occur data race.

glorv: Is there any case that depend on the ctx.resourceGroupName here? otherwise, should always override with reorgMeta's

PR #46762, opened 2023-09-07
util: memory: fix segfaults in meminfo when VirtualMemory errors

### LINKED ISSUE #46761: Segfault in util/memory/meminfo.go when mem.VirtualMemory() returns an error
## Bug Report

Get mem.VirtualMemory() to return an error. 

The error should have been returned

SIGSEGV because the function does
```
if err != nil {
    return v.Total, err
}
```
but in this case v is nil.

This bug is on master, I will push a pull request to fix it.

paveyry:
Issue number: close #46761

Problem Summary:

When VirtualMemory() returns an error, the actual value of `v` is nil, so returning v.Total when err != nil was causing segfaults

Simply return 0 when there is an error instead of trying to dereference nil

### Check List

Tests 

- [x] Unit test

### Release note

```release-note
util: memory: fix a bad error handling case which could cause a segmentation fault in case of error
```

### DIFF
--- a/util/memory/meminfo.go
+++ b/util/memory/meminfo.go
@@ -53,7 +53,7 @@ func MemTotalNormal() (uint64, error) {
 	}
 	v, err := mem.VirtualMemory()
 	if err != nil {
-		return v.Total, err
+		return 0, err
 	}
 	memLimit.set(v.Total, time.Now())
 	return v.Total, nil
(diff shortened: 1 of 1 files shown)


hawkingrei:
@paveyry Please sign the CLA.

hawkingrei:
If you can add some test cases for it. It will be better.

hawkingrei (approved):

Defined2014 (approved):

hawkingrei:
@paveyry Hello, Please sign the CLA, then we can merge this PR.

paveyry:
Your CLA form requires access to some private data, this is really annoying. It's really a deterrent from contributing to your codebase. 
Anyway, I signed it.

PR #46781, opened 2023-09-08
types, collation: utf8mb4_0900_bin shouldn't have RestoreData

### LINKED ISSUE #46690: `utf8mb4_0900_bin` shouldn't write `RestoredData`
The `NeedRestoredData` defines whether we need to write `RestoredData` for a type. See:

```
// NeedRestoredData returns if a type needs restored data.
// If the type is char and the collation is _bin, NeedRestoredData() returns false.
func NeedRestoredData(ft *FieldType) bool {
	if collate.NewCollationEnabled() &&
		IsNonBinaryStr(ft) &&
		!(collate.IsBinCollation(ft.GetCollate()) && !IsTypeVarchar(ft.GetType())) {
		return true
	}
	return false
}
```

However, `utf8mb4_0900_bin` is `NO PAD`, so it shouldn't contain restore data in the index value. It may give incorrect result for this collation when scanning through the index (but I'm not sure, haven't constructed an example).

YangKeao:

Issue Number: close #46690

Problem Summary:

This PR fixed three problems:

1. Add `utf8mb4_0900_bin` to the collation to id map in the `parser/mysql`. Without this field, it'll be converted to `utf8mb4_bin` (the default collation) when it's serialized into protobuf.
2. Fix the `IsBinCollation` for `utf8mb4_0900_bin`.
3. Make sure the `utf8mb4_0900_bin` doesn't need restore.

The composition of these three bugs make the behavior really weird.

1. When writing the index, because `IsBinCollation` is false while `NeedRestoreData` is true, it will write the original data into the index value.
2. When decoding the restore data (in unistore or tikv), the collation is `utf8mb4_bin` (because of the first problem), so it tries to decode an integer from the `RestoreData`, and it'll definitely fail. See the behavior tracked in issue https://github.com/pingcap/tidb/issues/46690

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

### Release note

```release-note
None
```

### DIFF
--- a/parser/mysql/charset.go
+++ b/parser/mysql/charset.go
@@ -79,7 +79,7 @@ var CharsetIDs = map[string]uint8{
 }
 
 // Collations maps MySQL collation ID to its name.
-var Collations = map[uint8]string{
+var Collations = map[uint16]string{
 	1:   "big5_chinese_ci",
 	2:   "latin2_czech_cs",
 	3:   "dec8_swedish_ci",
@@ -300,10 +300,11 @@ var Collations = map[uint8]string{
 	246: "utf8mb4_unicode_520_ci",
 	247: "utf8mb4_vietnamese_ci",
 	255: "utf8mb4_0900_ai_ci",
+	309: "utf8mb4_0900_bin",
 }
 
 // CollationNames maps MySQL collation name to its ID
-var CollationNames = map[string]uint8{
+var CollationNames = map[string]uint16{
 	"big5_chinese_ci":          1,
 	"latin2_czech_cs":          2,
 	"dec8_swedish_ci":          3,
@@ -524,6 +525,7 @@ var CollationNames = map[string]uint8{
 	"utf8mb4_unicode_520_ci":   246,
 	"utf8mb4_vietnamese_ci":    247,
 	"utf8mb4_0900_ai_ci":       255,
+	"utf8mb4_0900_bin":         309,
 }
 
 // MySQL collation information.
--- a/types/etc.go
+++ b/types/etc.go
@@ -121,7 +121,8 @@ func IsNonBinaryStr(ft *FieldType) bool {
 func NeedRestoredData(ft *FieldType) bool {
 	if collate.NewCollationEnabled() &&
 		IsNonBinaryStr(ft) &&
-		!(collate.IsBinCollation(ft.GetCollate()) && !IsTypeVarchar(ft.GetType())) {
+		(!collate.IsBinCollation(ft.GetCollate()) || IsTypeVarchar(ft.GetType())) &&
+		ft.GetCollate() != "utf8mb4_0900_bin" {
 		return true
 	}
 	return false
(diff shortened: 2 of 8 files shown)


util/collate/collate.go
```diff
@@ -360,7 +360,8 @@ func ConvertAndGetBinCollation(collate string) Collator {
 func IsBinCollation(collate string) bool {
 	return collate == charset.CollationASCII || collate == charset.CollationLatin1 ||
 		collate == charset.CollationUTF8 || collate == charset.CollationUTF8MB4 ||
-		collate == charset.CollationBin
+		collate == charset.CollationBin || collate == "utf8mb4_0900_bin"
```
zimulala, line 363: Could we put "utf8mb4_0900_bin" to charset.go as a constant? Then we can call it like `charset.CollationUTF8MB4`.

types/etc.go
```diff
@@ -121,7 +121,8 @@ func IsNonBinaryStr(ft *FieldType) bool {
 func NeedRestoredData(ft *FieldType) bool {
 	if collate.NewCollationEnabled() &&
 		IsNonBinaryStr(ft) &&
-		!(collate.IsBinCollation(ft.GetCollate()) && !IsTypeVarchar(ft.GetType())) {
+		(!collate.IsBinCollation(ft.GetCollate()) || IsTypeVarchar(ft.GetType())) &&
+		ft.GetCollate() != "utf8mb4_0900_bin" {
```
zimulala, line 125: We have added "utf8mb4_0900_bin" to "IsBinCollation" in this PR, so could we remove this line?

util/collate/collate.go
YangKeao: I think `CollationUTF8MB4` somehow represents the default collation for `utf8mb4` (it's `utf8mb4_bin` now). 

Maybe in the future PR I'll use `CollationUTF8MB40900Bin` to represent `utf8mb4_0900_bin` and replace every collation string literals, but not in this PR. I don't hope that big refractor hide the main modification made by this PR.

parser/mysql/charset.go
```diff
@@ -300,10 +300,11 @@ var Collations = map[uint8]string{
 	246: "utf8mb4_unicode_520_ci",
 	247: "utf8mb4_vietnamese_ci",
 	255: "utf8mb4_0900_ai_ci",
+	305: "utf8mb4_0900_bin",
```
zimulala, line 303: Why this is "305" not "309"?

util/collate/collate.go
zimulala: Got it.

parser/mysql/charset.go
YangKeao: It's a mistake. Fixed.

types/etc.go
YangKeao: On the other hand: If `NeedRestoredData` is true, then:

1. If `IsBinCollation` is true, the count of tailing spaces will be written into the value as an integer.
2. If `IsBinCollation` is false, the original string will be written into the value as a string.

zimulala (approved):

CbcWestwolf (approved):

PR #46927, opened 2023-09-12
executor: fix the error that is raised when executing the window function with range type frame

### LINKED ISSUE #46298: Window function raise error when executing with range frame
## Bug Report

```
drop table if exists test.first_range;
create table test.first_range(p int not null, o int not null, v int not null);
insert into test.first_range (p, o, v) values (0, 0, 0), (1, 1, 1), (1, 2, 2), (1, 4, 4), (1, 8, 8), (2, 0, 0), (2, 3, 3), (2, 10, 10), (2, 13, 13), (2, 15, 15), (3, 1, 1), (3, 3, 3), (3, 5, 5), (3, 9, 9), (3, 15, 15), (3, 20, 20), (3, 31, 31);
select *, first_value(v) over (partition by p order by o range between 3 preceding and 2.9 following) as a from test.first_range;
```

Return results

It raises error: ERROR 1105 (HY000): runtime error: index out of range [40] with length 40

nightly c0c54f23b350e0d7c56215

xzhangxian1008:

When executing window function with range type frame, tidb needs to compare the result of `CalFunc` and `order by column`. Sometimes, CalFunc returns decimal type while `order by column` is an int column. In this circumstance, tidb will get the value of `order by column` with decimal type which it's an int column, which result the raised error.

Issue Number: close #46298

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


planner/core/logical_plan_builder.go
```diff
@@ -6826,6 +6827,11 @@ func (b *PlanBuilder) buildWindowFunctionFrameBound(_ context.Context, spec *ast
 	if err != nil {
 		return nil, err
 	}
+	funcName = ast.Cast
+	bound.CastFuncs[0], err = expression.NewFunctionBase(b.ctx, funcName, bound.CalcFuncs[0].GetType(), col)
```
windtalker, line 6831: How about when there is no need to cast?

executor/pipelined_window.go
```diff
@@ -264,7 +264,7 @@ func (e *PipelinedWindowExec) getStart(ctx sessionctx.Context) (uint64, error) {
 			var res int64
 			var err error
 			for i := range e.orderByCols {
-				res, _, err = e.start.CmpFuncs[i](ctx, e.orderByCols[i], e.start.CalcFuncs[i], e.getRow(start), e.getRow(e.curRowIdx))
+				res, _, err = e.start.CmpFuncs[i](ctx, e.start.CastFuncs[i], e.start.CalcFuncs[i], e.getRow(start), e.getRow(e.curRowIdx))
```
windtalker, line 267: From the name `PipelinedWindowExec`, I assume there is a normal `WindowExec`, will this exec meet the same issue?

xzhangxian1008: > From the name `PipelinedWindowExec`, I assume there is a normal `WindowExec`, will this exec meet the same issue?

I find `window.go` will encounter the same problem. I will fix it

planner/core/logical_plan_builder.go
xzhangxian1008: > How about when there is no need to cast?

okk, I put the col expression to CastFuncs when there is no need to cast.

planner/core/logical_plan_builder.go
```diff
@@ -6759,6 +6759,7 @@ func (b *PlanBuilder) buildWindowFunctionFrameBound(_ context.Context, spec *ast
 	}
 
 	bound.CalcFuncs = make([]expression.Expression, len(orderByItems))
+	bound.CastFuncs = make([]expression.Expression, len(orderByItems))
```
windtalker, line 6762: Maybe rename it to `compareCols` since it is not always a cast function?

planner/core/logical_plan_builder.go
```diff
@@ -6826,6 +6827,15 @@ func (b *PlanBuilder) buildWindowFunctionFrameBound(_ context.Context, spec *ast
 	if err != nil {
 		return nil, err
 	}
+	if bound.CalcFuncs[0].GetType() != col.GetType() {
```
windtalker, line 6830: I think you can use EvalType to compare here because in my understanding, if `EvalType` is the same, TiDB can compare directly.

xzhangxian1008: > I think you can use EvalType to compare here because in my understanding, if `EvalType` is the same, TiDB can compare directly.

okk

planner/core/logical_plan_builder.go
xzhangxian1008: > Maybe rename it to `compareCols` since it is not always a cast function?

okk

planner/core/logical_plan_builder.go
xzhangxian1008: > I think you can use EvalType to compare here because in my understanding, if `EvalType` is the same, TiDB can compare directly.

okk

planner/core/logical_plan_builder.go
```diff
@@ -6826,6 +6827,15 @@ func (b *PlanBuilder) buildWindowFunctionFrameBound(_ context.Context, spec *ast
 	if err != nil {
 		return nil, err
 	}
+	if bound.CalcFuncs[0].GetType().EvalType() != col.GetType().EvalType() {
```
windtalker, line 6830: Add some tests to test different fieldType but the same evalType, for example, test the case that the order column is tinyint

xzhangxian1008: > Add some tests to test different fieldType but the same evalType, for example, test the case that the order column is tinyint

Done

`tk.MustQuery("select *, first_value(v) over (partition by p order by o range between 3 preceding and 2.9 following) as a from test.first_range;")`, for preceding, it will be tinyint vs bigint, for following it will be tinyint vs decimal.

planner/core/integration_test.go
```diff
@@ -5279,3 +5279,13 @@ func TestIssue45033(t *testing.T) {
...
 				                where  alias4.c2 = alias2.alias_col1);`).Check(testkit.Rows("0"))
 }
+
+func TestIssue46298(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("drop table if exists test.first_range;")
+	tk.MustExec("create table test.first_range(p int not null, o tinyint not null, v int not null);")
+	tk.MustExec("insert into test.first_range (p, o, v) values (0, 0, 0), (1, 1, 1), (1, 2, 2), (1, 4, 4), (1, 8, 8), (2, 0, 0), (2, 3, 3), (2, 10, 10), (2, 13, 13), (2, 15, 15), (3, 1, 1), (3, 3, 3), (3, 5, 5), (3, 9, 9), (3, 15, 15), (3, 20, 20), (3, 31, 31);")
```
windtalker, line 5289: Should add test for both `WindowExec` and `PipelinedWindowExec`

xzhangxian1008: > Should add test for both `WindowExec` and `PipelinedWindowExec`

Done

windtalker (approved):
LGTM

planner/core/logical_plan_builder.go
```diff
@@ -6826,6 +6830,15 @@ func (b *PlanBuilder) buildWindowFunctionFrameBound(_ context.Context, spec *ast
 	if err != nil {
 		return nil, err
 	}
+
+	if bound.CalcFuncs[0].GetType().EvalType() != col.GetType().EvalType() {
+		funcName = ast.Cast
+		bound.CompareCols[0], err = expression.NewFunctionBase(b.ctx, funcName, bound.CalcFuncs[0].GetType(), col)
```
guo-shaoge, line 6836:
```suggestion
		bound.CompareCols[0], err = expression.NewFunctionBase(b.ctx, ast.Cast, bound.CalcFuncs[0].GetType(), col)
```

guo-shaoge (approved):

PR #46957, opened 2023-09-13
disttask: refactor task step and batch subtask dispatching

D3Hunter:
### What problem does this PR solve?

Issue Number: ref #46258

Problem Summary:

- revert StepInit to -1 and StepImport to 1 to avoid potential compatibility issue
- remove StageFinished and Finished, add GetNextStep
- successfully task MUST go from StepInit to business steps, then StepDone.

since subtask metas might be too large to save in one transaction,
we plan to save subtasks in batch later, so add this:
invariant: task.Step always means the most recent step that all
corresponding subtasks have been saved to system table.

when all subtasks of task.Step is finished, we call OnNextSubtasksBatch
to generate subtasks of next step. after all subtasks of next step are
saved to system table, we will update task.Step to next step, so the
invariant hold.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #46993, opened 2023-09-14
ddl,util: modify the collation of the column of multi-valued index to "binary"

### LINKED ISSUE #46717: Array field should be passed to the `columnInfo`

YangKeao:

Issue Number: close #46717

This PR changes the collation of the column of multi-valued index to "binary". I think it's a better way to solve #46717. It doesn't need to modify tipb/tikv and the influence is much more limited.

#### Background

1. The design doc of multi-valued index tells that the collation of multi-valued index string should be "binary" (MySQL uses "utf8mb4_0900_as_cs", which is not supported by us for now).
2. The TiDB actually uses the "binary" collation to query the index (see https://github.com/pingcap/tidb/pull/40644), and the tailing whitespaces in multi-valued index string is not trimmed, so we didn't find any correctness issue about this.
3. The "restored data" is a part of the value of an index K-V. Because the generated key of a string datum can be different from the original value according to the collation, so we need the help of the "restored data" to reconstruct the original value. For example, as the `utf8mb4_bin` will trim the tailing whitespace, the "restored data" of the `utf8mb4_bin` string is an integer, which represents the count of tailing whitespaces.
4. The column corresponding to multi-valued index is written as `utf8mb4_bin` collation, but the restored data is not written and the tailing whitespaces are not trimmed. When the `TiKV` is about to restore the original value from the index, it'll fail to read the "restored data" and report an error.

This issue will only happen in compound index which contains at least one string column (which will actually write restored data), because when the index value is too short, both TiDB and TiKV will just regard it as an "old format" and skip the restore logic, then nothing is reported.

#### Q&A

1. Is it fine to modify the collation directly? Yes, the multi-valued index is written without restored data, and the whitespace in the key is also kept. For the index lookup, the planner always uses the "binary" collation (see https://github.com/pingcap/tidb/pull/40644). It's fine to read the value as binary collations.
6. How does it solve the problem in #46717? By setting the collation to "binary", TiKV will not try to read the restored value.

Another question is, why it's a better solution than  https://github.com/pingcap/tipb/pull/317 https://github.com/tikv/tikv/pull/15537  https://github.com/pingcap/tidb/pull/46718 ?

1. https://github.com/tikv/tikv/pull/15537 gives the wrong type for array columns. Though for now array columns are only used in multi-valued index, the original design of TiDB marks it as a JSON type. If it's read for index, it'll use the original type. Theoretically, TiKV doesn't need to know whether this K-V is an array or not, as it'll only handle the original value. https://github.com/tikv/tikv/pull/15537 lets the TiKV regard it as a JSON, but it's actually a string (or integer, or something else), which can cause bugs in the future.
2. The influence and logic of this PR is much simpler: use "binary" collation for multi-valued index. We don't need to consider the interaction between tidb/tikv, and how different query engines handle the types.

When I was drafting  https://github.com/pingcap/tipb/pull/317 https://github.com/tikv/tikv/pull/15537  https://github.com/pingcap/tidb/pull/46718, I didn't fully understand the design of original multi-valued index type. Now I'm pretty confident about this PR :+1: . 

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

You can test whether #46717 is solved by using the following SQL:

```sql
create table t (pk varchar(4) primary key clustered, j json, str varchar(255), value int, key idx((cast(j as char(100) array)), str));
insert into t values ("1", '["a"]', 'b', 1);
select * from t use index(idx) where "a" member of (j);
```

I'll add them to `tidb-test` later, as I want to test across tidb and tikv :thinking: . (I'm not sure whether `tests/integrationtest` will run with TiKV after merged. I'll confirm later. If so, `tests/integrationtest` looks better).

### Release note

```release-note
Change the collation of the column of multi-valued index to "binary"
```


planner/core/plan_to_pb_test.go
```diff
@@ -79,4 +79,19 @@ func TestColumnToProto(t *testing.T) {
 	}
 	pc = util.ColumnToProto(col2, false)
 	require.Len(t, pc.Elems, 2)
+
+	tp = types.NewFieldTypeBuilder().
+		SetType(mysql.TypeString).
+		SetCharset("utf8mb4").
```
bb7133, line 85: If we'll modify the collation to 'binary', why setting 'utf8mb4'/'utf8mb4_bin' in this case?

YangKeao: The `tp` here is the input of `util.ColumnToProto`. This test shows the function `util.ColumnToProto` can converts any collation (if `.array` is true) into `binary` collation. The expected output is defined in L94:

```
expect = &tipb.ColumnInfo{ColumnId: 0, Tp: 0xfe, Collation: 63, ColumnLen: 100, Decimal: 0, Flag: 10, Elems: []string(nil), DefaultVal: []uint8(nil), PkHandle: false, XXX_unrecognized: []uint8(nil)}
```

The `Collation: 63` means binary collation.

util/misc.go
```diff
@@ -423,6 +423,12 @@ func ColumnToProto(c *model.ColumnInfo, forIndex bool) *tipb.ColumnInfo {
 	if forIndex {
 		// Use array type for read the multi-valued index.
 		pc.Tp = int32(c.FieldType.ArrayType().GetType())
+		if c.FieldType.IsArray() {
```
zimulala, line 426: If the data stored in TiKV is already utf8mb4 (data stored before this PR), is there a compatibility problem with binary when reading here?

zimulala (approved):
LGTM

YangKeao:
/hold

This PR will make `br` fail. Still investigating.

bb7133 (approved):
LGTM

YangKeao:
/unhold

I think it's a bug of dist task framework. It occurs less frequently in master, but I think it's still possible. I'll confirm it with the developer of dist task framework.

YangKeao:
> /unhold
> 
> I think it's a bug of dist task framework. It occurs less frequently in master, but I think it's still possible. I'll confirm it with the developer of dist task framework.

The reported bug (importing succeed with 0 row) is fixed by https://github.com/pingcap/tidb/pull/46957 (though, maybe not intentional).

Issue #46704, opened 2023-09-06
Enhanced Features in 'Import Into' SQL Statement


D3Hunter:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

The 'import into' SQL was introduced in version 7.2, integrating the Lightning local backend (physical import mode). In version 7.2, it was released as an experimental feature and Only support 1 TiB source data and can’t resume from breakpoint. So we have to deliver the following new features to make "import into " more powerful.

This newly added feature in version 7.4 ：

1. Support "S3 global sort" to avoid problems caused by data range overlap, to make import more stable.
2. support parameters like "SPLIT_FILE", and  support import compressed source files like .gzip, .gz, .zstd, .zst and .snappy.
3. Expected to support data imports of up to 100 TiB.（The final size depends on the results of testing.）
4. The import performance can reach up to 1 TiB per hour.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

D3Hunter:
this part is release in v7.4.0, so close it

PR #46998, opened 2023-09-15
importinto: integrate global sort(without merge-sort) part 1

D3Hunter:

Issue Number: ref #46704

Problem Summary:

- add cloud_storage_uri option
- when use global sort: job step goes from none -> global-sorting -> importing -> validating -> none.
    - otherwise, it's none -> importing -> validating -> none.
- integrate global sort(without merge-sort part), but since this pr is large, will add test for it later.
- add Init to dispatcher, if Init returns error, dispatcher manager will fail the task directly, so the returned error should be a fatal error.
- some refactor

when using global sort, we write index kv using IndexRouteWriter
we route kvs of different index to different writer in order to make
merge sort easier, else kv data of all subtasks will all be overlapped.

drawback of doing this is that the number of writers need to open will be
index-count * encode-concurrency, when the table has many indexes, and each
writer will take 256MiB buffer on default,
this will take a lot of memory, or even OOM.
we can adjust memory buf size later, but might write too many small files.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


ywqzzy (commented):
Framework PART LGTM

br/pkg/lightning/backend/external/util.go
```diff
@@ -241,3 +241,79 @@ func GetMaxOverlapping(points []Endpoint) int {
...
+	m.MaxKey = NotNilMax(m.MaxKey, other.MaxKey)
+	m.TotalKVSize += other.TotalKVSize
+
+	m.DataFiles = append(m.DataFiles, other.DataFiles...)
+	m.StatFiles = append(m.StatFiles, other.StatFiles...)
+}
+
+// MergeSummary merges the WriterSummary into this SortedKVMeta.
+func (m *SortedKVMeta) MergeSummary(summary *WriterSummary) {
+	m.MinKey = NotNilMin(m.MinKey, summary.Min)
```
lance6716, line 282: not a big problem. Seems we can use NewSortedKVMeta and Merge to implement this

br/pkg/lightning/backend/external/writer.go
```diff
@@ -427,3 +430,39 @@ func (w *Writer) createStorageWriter(ctx context.Context) (
...
+}
+
+// AppendRows implements backend.EngineWriter interface.
+func (e *EngineWriter) AppendRows(ctx context.Context, _ []string, rows encode.Rows) error {
+	kvs := kv.Rows2KvPairs(rows)
+	if len(kvs) == 0 {
+		return nil
+	}
+	for _, item := range kvs {
+		err := e.w.WriteRow(ctx, item.Key, item.Val, nil)
```
lance6716, line 451:
```suggestion
		err := e.w.WriteRow(ctx, item.Key, item.Val, item.RowID)
```

br/pkg/lightning/backend/local/local.go
```diff
@@ -1689,6 +1689,18 @@ func (local *Backend) GetImportedKVCount(engineUUID uuid.UUID) int64 {
 	return e.importedKVCount.Load()
 }
 
+// GetExternalEngineKVStatistics returns kv statistics of some engine.
+func (local *Backend) GetExternalEngineKVStatistics(engineUUID uuid.UUID) (
+	totalKVSize int64, totalKVCount int64) {
+	v, ok := local.externalEngine[engineUUID]
+	if !ok {
+		// we get it after import, but before clean up, so this should not happen
```
lance6716, line 1697: I don't understand it. Can you explain more details about the process of local engine import?

lance6716 (commented):
15 / 52 files viewed

will review later

disttask/importinto/dispatcher.go
```diff
@@ -194,10 +195,12 @@ func (dsp *ImportDispatcherExt) unregisterTask(ctx context.Context, task *proto.
 // OnNextSubtasksBatch generate batch of next stage's plan.
 func (dsp *ImportDispatcherExt) OnNextSubtasksBatch(ctx context.Context, taskHandle dispatcher.TaskHandle, gTask *proto.Task) (
 	resSubtaskMeta [][]byte, err error) {
+	nextStep := dsp.GetNextStep(gTask)
 	logger := logutil.BgLogger().With(
 		zap.String("type", gTask.Type),
 		zap.Int64("task-id", gTask.ID),
-		zap.String("step", stepStr(gTask.Step)),
+		zap.String("curr-step", stepStr(gTask.Step)),
```
lance6716, line 202: In master code, I have added `nextStep` as a parameter of `OnNextSubtasksBatch`. 😂 My consideration is, `GetNextStep` can be called once while `OnNextSubtasksBatch` maybe called for multiple time for a subtask

lance6716 (commented):
21 / 52 files viewed

tablecodec/tablecodec.go
```diff
@@ -276,6 +276,19 @@ func DecodeKeyHead(key kv.Key) (tableID int64, indexID int64, isRecordKey bool,
 	return
 }
 
+// DecodeIndexID decodes indexID from the key.
+// this method simply extract index id part, and no other checking.
+// Caller should make sure the key is an index key.
+func DecodeIndexID(key kv.Key) (int64, error) {
```
lance6716, line 282: seems in below `DecodeTableID` we should also consider keyspace of APIv2

executor/importer/table_import.go
```diff
@@ -147,6 +147,29 @@ func GetCachedKVStoreFrom(pdAddr string, tls *common.TLS) (tidbkv.Storage, error
 	return kvStore, nil
 }
 
+// GetRegionSplitSizeKeys gets the region split size and keys from PD.
+func GetRegionSplitSizeKeys(ctx context.Context) (regionSplitSize int64, regionSplitKeys int64, err error) {
+	tidbCfg := tidb.GetGlobalConfig()
```
lance6716, line 152: maybe in future we can get the singleton PD client of this TiDB instance, and use the latest PD address, in case of PD node is changed.

executor/importer/table_import.go
```diff
@@ -147,6 +147,29 @@ func GetCachedKVStoreFrom(pdAddr string, tls *common.TLS) (tidbkv.Storage, error
...
+	hostPort := net.JoinHostPort("127.0.0.1", strconv.Itoa(int(tidbCfg.Status.StatusPort)))
```
lance6716, line 153: I see that the `hostPort` is used by `tls`, and in fact `tls` will overwrite the host by `WithHost` in `getSplitConfFromStore`. So maybe don't need to set the actual value.

lance6716 (commented):
33 / 52 files viewed

tablecodec/tablecodec.go
D3Hunter: only that method checks it, seems we don't need to check it here for index kv
```
	// `V2` use new encoding for RawKV & TxnKV to support more features.
	//
	// Key Encoding:
	//  TiDB: start with `m` or `t`, the same as `V1`.
	//  TxnKV: prefix with `x`, encoded as `MCE( x{keyspace id} + {user key} ) + timestamp`.
	//  RawKV: prefix with `r`, encoded as `MCE( r{keyspace id} + {user key} ) + timestamp`.
```

br/pkg/lightning/backend/external/writer.go
D3Hunter: `e.w` requires a `handle`, import into don't have conflict detect now, we can do it later.

br/pkg/lightning/backend/local/local.go
D3Hunter: copied from prev `GetImportedKVCount`, will remove it

disttask/importinto/dispatcher.go
D3Hunter: i say that `GetNextStep` now has a `handle` param, we can remove it, and move those logic into `OnNextSubtasksBatch`, if there's no need to do merge-sort, just return 0 subtask metas, and let dispatcher call it again on next step.

lance6716: If `GetNextStep` is heavy to compute, it's better we let framework call it once and send the result to the argument of `OnNextSubtasksBatch`

> if there's no need to do merge-sort, just return 0 subtask metas

Currently I think it's also acceptable. In future we may have a complex step transition and must use non-linear steps.

tablecodec/tablecodec.go
lance6716: Oh, keyspace may be automatically adjusted in client-go, we don't need to care about it here.

disttask/importinto/dispatcher.go
```diff
@@ -223,18 +226,28 @@ func (dsp *ImportDispatcherExt) OnNextSubtasksBatch(ctx context.Context, taskHan
...
 	}()
 
+	var previousSubtaskMetas [][]byte
 	switch gTask.Step {
 	case proto.StepInit:
 		if metrics, ok := metric.GetCommonMetric(ctx); ok {
 			metrics.BytesCounter.WithLabelValues(metric.StateTotalRestore).Add(float64(taskMeta.Plan.TotalFileSize))
 		}
-		if err := preProcess(ctx, taskHandle, gTask, taskMeta, logger); err != nil {
+		jobStep := importer.JobStepImporting
```
lance6716, line 235: and these logic seems more natural as a part of `GetNextStep`

lance6716 (commented):
35 / 52 files viewed

disttask/importinto/encode_and_sort_operator.go
```diff
@@ -116,6 +126,40 @@ func (op *encodeAndSortOperator) Done() <-chan struct{} {
...
+func newChunkWorker(ctx context.Context, op *encodeAndSortOperator, workerID int64) *chunkWorker {
+	w := &chunkWorker{
+		ctx: ctx,
+		op:  op,
+	}
+	if op.tableImporter.IsGlobalSort() {
+		// sorted index kv storage path: /{taskID}/{subtaskID}/index/{indexID}/{workerID}
+		indexWriterFn := func(indexID int64) *external.Writer {
+			builder := external.NewWriterBuilder().
+				SetOnCloseFunc(func(summary *external.WriterSummary) {
```
lance6716, line 143: I plan to let the "WriterSummary" result be returned when Writer.Close, not passing it around using OnClose. What do you think?

disttask/importinto/encode_and_sort_operator.go
```diff
@@ -116,6 +126,40 @@ func (op *encodeAndSortOperator) Done() <-chan struct{} {
...
+					op.sharedVars.mergeIndexSummary(indexID, summary)
+				})
+			prefix := path.Join(strconv.Itoa(int(op.taskID)), strconv.Itoa(int(op.subtaskID)))
+			writerID := path.Join("index", strconv.Itoa(int(indexID)), strconv.Itoa(int(workerID)))
```
lance6716, line 147: Considering two nodes are running same subtask (maybe the old one suffers network partition and framework dispatch a new node), we should use random writerID to avoid overwritten. Or maybe WriterBuilder should assign UUID internally 🤔 

The result file path will be returned by OnClose, so caller seems don't need to care about the writerID

disttask/importinto/encode_and_sort_operator.go
```diff
@@ -125,10 +169,27 @@ func (w *chunkWorker) HandleTask(task *importStepMinimalTask, _ func(workerpool.
...
 		w.op.onError(err)
 	}
 }
 
-func (*chunkWorker) Close() {
+func (w *chunkWorker) Close() {
+	closeCtx := w.ctx
+	if closeCtx.Err() != nil {
+		// in case of context canceled, we need to create a new context to close writers.
+		closeCtx = context.Background()
```
lance6716, line 181: maybe some timeout to avoid blocking main logic. resource leak is not that harmful

lance6716 (commented):
37 / 52 files viewed

disttask/importinto/encode_and_sort_operator.go
lance6716: cc @wjhuang2016 @tangenta

executor/importer/import.go
```diff
@@ -654,6 +666,25 @@ func (p *Plan) initOptions(seCtx sessionctx.Context, options []*plannercore.Load
...
+		}
+		// set cloud storage uri to empty string to force uses local sort when
+		// the global variable is set.
+		if v != "" {
+			b, err := storage.ParseBackend(v, nil)
+			if err != nil {
+				return exeerrors.ErrInvalidOptionVal.FastGenByArgs(opt.Name)
+			}
+			// only support s3 and gcs now.
+			if b.GetS3() == nil && b.GetGcs() == nil {
```
lance6716, line 682: @wjhuang2016 we will return error when it's google cloud storage? I remember we have some S3 only implementation. Which release version do we plan to remove the restriction?

lance6716 (commented):
41 / 52 files viewed

executor/importer/import.go
D3Hunter: in doc, import-into only support s3 too, see https://github.com/pingcap/docs-cn/pull/15021/files#diff-218ca9ffe9f1f361d40ff3e8ff7ea091223220e4193574191cf763a64a5de299R129

but for integration test, we need gcs

disttask/importinto/dispatcher.go
D3Hunter: no, that's `import-into` job's step, exposed to user and has less steps

disttask/importinto/encode_and_sort_operator.go
D3Hunter: lgtm

disttask/importinto/encode_and_sort_operator.go
D3Hunter: > Considering two nodes are running same subtask

makes sense, will change it to a uuid here

> so caller seems don't need to care about the writerID

that would be better

executor/importer/table_import.go
D3Hunter: https://github.com/pingcap/tidb/pull/46998/commits/e25413d647f9ae7b6ac9a8991cd292a828523bab

br/pkg/lightning/backend/local/local.go
D3Hunter: https://github.com/pingcap/tidb/pull/46998/commits/e25413d647f9ae7b6ac9a8991cd292a828523bab

br/pkg/lightning/backend/external/util.go
D3Hunter: https://github.com/pingcap/tidb/pull/46998/commits/e25413d647f9ae7b6ac9a8991cd292a828523bab

disttask/importinto/encode_and_sort_operator.go
D3Hunter: https://github.com/pingcap/tidb/pull/46998/commits/a3e9817aaf3f5987c3ca20965bf63682096ed59c

executor/importer/import.go
```diff
@@ -497,13 +503,19 @@ func (p *Plan) initDefaultOptions() {
...
 	p.Charset = &v
 }
 
 func (p *Plan) initOptions(seCtx sessionctx.Context, options []*plannercore.LoadDataOpt) error {
-	p.initDefaultOptions()
+	cloudStorageURI, err1 := seCtx.GetSessionVars().GlobalVarsAccessor.
+		GetGlobalSysVar(variable.TiDBCloudStorageURI)
+	if err1 != nil {
+		return errors.Trace(err1)
+	}
```
tangenta, line 517: Use `variable.CloudStorageURI.Load()` instead?

br/pkg/lightning/backend/external/util.go
```diff
@@ -239,3 +239,71 @@ func GetMaxOverlapping(points []Endpoint) int64 {
...
+}
+
+// MergeSummary merges the WriterSummary into this SortedKVMeta.
+func (m *SortedKVMeta) MergeSummary(summary *WriterSummary) {
+	m.Merge(NewSortedKVMeta(summary))
+}
+
+// NotNilMin returns the smallest of a and b, ignoring nil values.
+func NotNilMin(a, b []byte) []byte {
+	if len(a) == 0 {
```
Benjamin2037, line 285: what if both a and b nil？it will return b（nil）？

D3Hunter: that's moved from add-index, didn't change that

br/pkg/lightning/backend/external/util.go
```diff
@@ -239,3 +239,71 @@ func GetMaxOverlapping(points []Endpoint) int64 {
...
+	m.StatFiles = append(m.StatFiles, other.StatFiles...)
+}
+
+// MergeSummary merges the WriterSummary into this SortedKVMeta.
+func (m *SortedKVMeta) MergeSummary(summary *WriterSummary) {
+	m.Merge(NewSortedKVMeta(summary))
+}
+
+// NotNilMin returns the smallest of a and b, ignoring nil values.
+func NotNilMin(a, b []byte) []byte {
```
D3Hunter, line 284: moved from ddl/backfilling_dispatcher.go

disttask/importinto/subtask_executor.go
```diff
@@ -77,8 +78,14 @@ func (e *importMinimalTaskExecutor) Run(ctx context.Context) error {
...
 	chunkCheckpoint := toChunkCheckpoint(e.mTtask.Chunk)
 	sharedVars := e.mTtask.SharedVars
-	if err := importer.ProcessChunk(ctx, &chunkCheckpoint, sharedVars.TableImporter, sharedVars.DataEngine, sharedVars.IndexEngine, sharedVars.Progress, logger); err != nil {
-		return err
+	if sharedVars.TableImporter.IsLocalSort() {
+		if err := importer.ProcessChunk(ctx, &chunkCheckpoint, sharedVars.TableImporter, sharedVars.DataEngine, sharedVars.IndexEngine, sharedVars.Progress, logger); err != nil {
+			return err
+		}
+	} else {
+		if err := importer.ProcessChunkWith(ctx, &chunkCheckpoint, sharedVars.TableImporter, dataWriter, indexWriter, sharedVars.Progress, logger); err != nil {
```
Benjamin2037, line 86: global sort use ProcessChunkWith, but local use ProcessChunk

D3Hunter: that's expected

br/pkg/lightning/backend/external/util.go
tangenta: @Benjamin2037 Yes, returning a `nil` value is expected.

disttask/importinto/scheduler.go
```diff
@@ -103,31 +111,37 @@ func (s *importStepExecutor) RunSubtask(ctx context.Context, subtask *proto.Subt
...
+		if err != nil {
+			return err
+		}
+		// Unlike in Lightning, we start an index engine for each subtask,
+		// whereas previously there was only a single index engine globally.
+		// This is because the scheduler currently does not have a post-processing mechanism.
+		// If we import the index in `cleanupSubtaskEnv`, the dispatcher will not wait for the import to complete.
+		// Multiple index engines may suffer performance degradation due to range overlap.
+		// These issues will be alleviated after we integrate s3 sorter.
+		// engineID = -1, -2, -3, ...
```
Benjamin2037, line 126: what is this comment used for “// engineID = -1, -2, -3, ...”？

disttask/importinto/subtask_executor.go
Benjamin2037: I mean the name of function is not consistency.

br/pkg/lightning/backend/external/util.go
```diff
@@ -239,3 +239,71 @@ func GetMaxOverlapping(points []Endpoint) int64 {
...
+	m.DataFiles = append(m.DataFiles, other.DataFiles...)
+	m.StatFiles = append(m.StatFiles, other.StatFiles...)
+}
+
+// MergeSummary merges the WriterSummary into this SortedKVMeta.
+func (m *SortedKVMeta) MergeSummary(summary *WriterSummary) {
+	m.Merge(NewSortedKVMeta(summary))
+}
+
+// NotNilMin returns the smallest of a and b, ignoring nil values.
```
Benjamin2037, line 283:
```suggestion
// NotNilMin returns the smallest of a and b.
```

tangenta (approved):

disttask/importinto/scheduler.go
D3Hunter: each subtask will have a index engine, that's how we alloc it's id

PR #47070, opened 2023-09-18
statstics: reuse fmsketch

### LINKED ISSUE #47071: Reuse FMsketch to Avoid High Memory Allocation
## Enhancement

The fmsketch module often has to create and discard objects, leading to high memory allocation and frequent garbage collection. This process can be resource-intensive and can negatively impact the performance of the application, especially in high-load scenarios.

hawkingrei:

Issue Number: close #47071

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test

before:

image

after:

image

- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/statistics/fmsketch.go
+++ b/statistics/fmsketch.go
@@ -214,6 +214,9 @@ func (s *FMSketch) reset() {
 
 // Destroy resets the FMSketch and puts it back to the pool.
 func (s *FMSketch) Destroy() {
+	if s == nil {
+		return
+	}
 	s.reset()
 	fmSketchPool.Put(s)
 }
--- a/statistics/fmsketch.go
+++ b/statistics/fmsketch.go
@@ -165,10 +165,8 @@ func FMSketchFromProto(protoSketch *tipb.FMSketch) *FMSketch {
 	if protoSketch == nil {
 		return nil
 	}
-	sketch := &FMSketch{
-		hashset: make(map[uint64]bool, len(protoSketch.Hashset)),
-		mask:    protoSketch.Mask,
-	}
+	sketch := fmSketchPool.Get().(*FMSketch)
+	sketch.mask = protoSketch.Mask
 	for _, val := range protoSketch.Hashset {
 		sketch.hashset[val] = true
 	}
--- a/statistics/fmsketch.go
+++ b/statistics/fmsketch.go
@@ -212,8 +212,8 @@ func (s *FMSketch) reset() {
 	s.maxSize = 0
 }
 
-// Destroy resets the FMSketch and puts it back to the pool.
-func (s *FMSketch) Destroy() {
+// DestroyAndPutToPool resets the FMSketch and puts it to the pool.
+func (s *FMSketch) DestroyAndPutToPool() {
 	if s == nil {
 		return
 	}
--- a/statistics/fmsketch.go
+++ b/statistics/fmsketch.go
@@ -61,15 +61,12 @@ func (s *FMSketch) Copy() *FMSketch {
 	if s == nil {
 		return nil
 	}
-	hashset := make(map[uint64]bool)
+	result := NewFMSketch(s.maxSize)
 	for key, value := range s.hashset {
-		hashset[key] = value
+		result.hashset[key] = value
 	}
-	return &FMSketch{
-		hashset: hashset,
-		mask:    s.mask,
-		maxSize: s.maxSize,
-	}
+	result.mask = s.mask
+	return result
 }
 
 // NDV returns the ndv of the sketch.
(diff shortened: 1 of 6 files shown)


qw4990 (approved):

statistics/fmsketch.go
```diff
@@ -201,3 +205,18 @@ func (s *FMSketch) MemoryUsage() (sum int64) {
...
 }
+
+func (s *FMSketch) reset() {
+	maps.Clear(s.hashset)
+	s.mask = 0
+	s.maxSize = 0
+}
+
+// Destroy resets the FMSketch and puts it back to the pool.
+func (s *FMSketch) Destroy() {
```
time-and-fate, line 216: I suggest we name it something like `DestroyAndPutToPool`.
Because it's meaningless to explicitly "destroy", the main point is that we want to put it back in the pool. Also, I think it's important that other people are aware that we'll reuse this struct when calling this method.

time-and-fate (commented):
I think the logic in this PR is very fragile.
In this PR:
1. We put back the `allFms` in the pool in `MergePartitionStats2GlobalStats`
2. We put back the `GlobalStats.Fms` in the pool in `handleGlobalStats` and `updateGlobalStats`, which is returned from `MergePartitionStats2GlobalStats`
3. We put back the `tableAllPartitionStats.Column.FMSketch` in the pool in `handleGlobalStats`, which is also created in `MergePartitionStats2GlobalStats`.
The assumptions are very fragile, if we share structs in any two of the above three places, there will be bugs. And it's hard to realize this, because we put them back in the pool in totally different places.

time-and-fate:
And we are handling the logic inconsistently.
When we call `MergePartitionStats2GlobalStats` from `updateGlobalStats`, we don't reuse the FM sketch in the `allPartitionStats`, but when we call `MergePartitionStats2GlobalStats` from `handleGlobalStats`, we reuse them.
Even in `handleGlobalStats`, we only reuse FM sketches from `Column`, but not reuse those from `Index`.

time-and-fate:
My suggestion is to do all the `Put()` in `MergePartitionStats2GlobalStats`. And it looks like we can even remove `Fms` from the `struct GlobalStats`.

statistics/fmsketch.go
hawkingrei: I have updated.

hawkingrei:
> I think the logic in this PR is very fragile. In this PR:
> 
>     1. We put back the `allFms` in the pool in `MergePartitionStats2GlobalStats`
> 
>     2. We put back the `GlobalStats.Fms` in the pool in `handleGlobalStats` and `updateGlobalStats`, which is returned from `MergePartitionStats2GlobalStats`
> 
>     3. We put back the `tableAllPartitionStats.Column.FMSketch` in the pool in `handleGlobalStats`, which is also created in `MergePartitionStats2GlobalStats`.
> 
> 
> The assumptions are very fragile, if we share structs in any two of the above three places, there will be bugs. And it's hard to realize this, because we put them back in the pool in totally different places.

In fact, we optimize the two places.

1、create FM sketch from the NewFmSketch 
2、Create from the proto


all the implements create Fmsketch from the pool and release Fmsketch at the end of the lift.

the root cause is at the ```tableAllPartitionStats``` which the life of FMsketch is out of ```MergePartitionStats2GlobalStats```.

hawkingrei:
> My suggestion is to do all the `Put()` in `MergePartitionStats2GlobalStats`. And it looks like we can even remove `Fms` from the `struct GlobalStats`.

I have refactor this code. Now we have an external cache in the mergeGlobalStatus. if it is not from the external, we will release it in the ```mergeGlobalStatus```.

statistics/handle/globalstats/global_stats.go
```diff
@@ -93,6 +93,14 @@ func (g *GlobalStatusHandler) MergePartitionStats2GlobalStats(
...
 ) (globalStats *GlobalStats, err error) {
 	partitionNum := len(globalTableInfo.Partition.Definitions)
+	externalCache := false
+	if allPartitionStats == nil {
+		allPartitionStats = make(map[int64]*statistics.Table)
+	} else {
+		externalCache = true
+	}
+	// initialized the globalStats
+	globalStats = new(GlobalStats)
```
time-and-fate, line 103: Looks like we don't need this. We initiate `globalStats` below.

statistics/table.go
```diff
@@ -505,6 +505,16 @@ func (t *Table) IsOutdated() bool {
 	return false
 }
 
+// ReleaseAndPutToPool releases data structures of Table and put itself back to pool.
+func (t *Table) ReleaseAndPutToPool() {
+	for _, col := range t.Columns {
+		col.FMSketch.DestroyAndPutToPool()
```
time-and-fate, line 511: I think it's safer to set `FMSketch` to `nil` after putting to the pool.

time-and-fate: Ok to me to improve them in another PR or in another way at your convenience.

hawkingrei: Done，use ```map.clear``` to clean it.

time-and-fate (approved):

Issue #46304, opened 2023-08-22
Placement rule supports set follower strict


nolouch:
## Enhancement
Currently, all `FOLLOWER_CONSTRAINTS` will set the role in the rule to be `voter`. It can cover major scenarios and reduce user's burden. but in some scenarios cannot cover, like:

there are three regions, `region1`, `region2`, `region3`. the user wants the replicas in `region1` and `region2` can be the leader, and in normal situations, the leader should be placed in `region1`, if `region1` is gone, the leader is better in `region2.` and `region3` reject be the leader,  `region3` can be seen as a backup zone, it's not used to serve.

maybe we can enhance the `FOLLOWER_CONSTRAINTS`, like
##### option 1
```
Dictionary:  e.g. {"+region=us-east": 2, "+region=us-west,#evict-leader": 1}
```
`#evict-leader` explains voters should be followers, and reject becoming leaders.

full SQL will like this:
```
CREATE PLACEMENT POLICY `mydeploy`  LEADER_CONSTRAINTS='{"+region=us-east-1": 1}' FOLLOWER_CONSTRAINTS='{"+region=us-east-1": 1,  "+region=use-east-2":2, "+region=us-west,#evict-leader": 1};
```
##### option 2
or we can call `FOLLOWER_STRICT_CONSTRAINTS` to set the `follower` role

#### option 3
add `VOTER_CONSTRAINTS` and change the `FOLLOWER_CONSTRAINTS` role be `follower`,  it has compatibility issues.

PR #47307, opened 2023-09-26
placement: supports follower role with evict-leader hint

nolouch:

Issue Number: close https://github.com/pingcap/tidb/issues/46304

Problem Summary:

Currently, all `FOLLOWER_CONSTRAINTS` will set the role in the rule to be `voter`. It can cover major scenarios and reduce the user's burden. but in some scenarios cannot cover, like:

there are three az, zone1, zone2, zone3. the user wants the replicas in zone 1 and zone 2 can be the leader, and normal situations, the leader should be placed in zone 1, if zone 1 is gone, the leader is better in zone 2. and zone3 reject be the leader,  zone 3 can be seen as a backup zone, it's not used to serve.

maybe we can enhance the `FOLLOWER_CONSTRAINTS`, like
#####  add attribute
```
Dictionary:  e.g. {"+region=us-east": 2, "+region=us-west,#evict-leader": 1}
```
`#evict-leader` explains voters should be followers, and reject becoming leaders.

full SQL will like this:
```
CREATE PLACEMENT POLICY `mydeploy`  LEADER_CONSTRAINTS='{"+region=us-east-1": 1}' FOLLOWER_CONSTRAINTS='{"+region=us-east-1": 1,  "+region=use-east-2":2, "+region=us-west,#evict-leader": 1};
```

#### alternative plan

**warn**: this is the current easy way to achieve this issue scenario. we would be supported `leader_preferences` in the future to define the leader choice priority.

add new keywords.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test

### Release note

```release-note
None
```

### DIFF
--- a/ddl/placement/rule.go
+++ b/ddl/placement/rule.go
@@ -39,6 +39,12 @@ const (
 	Learner PeerRoleType = "learner"
 )
 
+const (
+	attributePrefix = "#"
+	// AttributeEvictLeader is used to evict leader from a store.
+	attributeEvictLeader = "evict-leader"
+)
+
 // RuleGroupConfig defines basic config of rule group
 type RuleGroupConfig struct {
 	ID       string `json:"id"`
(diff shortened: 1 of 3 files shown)


ddl/placement/rule.go
```diff
@@ -200,12 +206,29 @@ func NewRules(role PeerRoleType, replicas uint64, cnstr string) ([]*Rule, error)
...
+							overrideRole = Follower
+						}
+					default:
+						return rules, fmt.Errorf("%w: unsupported attribute '%s'", ErrUnsupportedConstraint, str)
+					}
+					continue
+				}
+				newLabels = append(newLabels, str)
+			}
+			labelConstraints, err := NewConstraints(newLabels)
```
xhebox, line 226: LGTM, but what about adding this into `NewConstraints`?

nolouch: currently, the constraint has no `role` semantics, so I still put it outer the `NewConstraints`.

xhebox (approved):

nolouch:
PTAL @CabinfeverB

rleungx (approved):
LGTM, but I was wondering if the user uses # as part of the label, it may have a problem. It's better to add the notice in the doc.

PR #47319, opened 2023-09-27
statistics: Update global count and modify_count only if partition is not locked

hi-rustin:

close https://github.com/pingcap/tidb/issues/47335

Problem Summary:

-  Update global count and modify_count only if the partition is not locked.
-  Do not allow to use of `unlock stats t` to unlock the partition.
-  Added more tests.
-  Changed it to use zap.Any() to print log.
-  Made sure we can have negative deltas in lock info
```log
[2023/09/27 16:28:31.742 +08:00] [INFO] [lock_stats.go:86] ["lock table"] [category=stats] [tables="{\"102\":{\"PartitionInfo\":null,\"FullName\":\"test.t\"},\"104\":{\"PartitionInfo\":null,\"FullName\":\"test.t1\"}}"]
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [x] [Manual test ](https://pingcap.feishu.cn/docx/KY2edTqS1ob5fAxo6mgcBYcDnrd#part-VJU4dN9rfoggKrxa7GLcvw5RnIh)
- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


qw4990 (approved):

statistics/handle/update.go
```diff
@@ -332,9 +333,9 @@ func (h *Handle) dumpTableStatCountToKV(is infoschema.InfoSchema, physicalTableI
 			return
 		}
 		affectedRows += sctx.GetSessionVars().StmtCtx.AffectedRows()
-		// If only the partition is locked, we don't need to update the global-stats.
+		// If the partition is locked, we don't need to update the global-stats.
 		// We will update its global-stats when the partition is unlocked.
-		if isTableLocked || !isPartitionLocked {
+		if !isPartitionLocked {
```
time-and-fate, line 338: Looks like it should be `!isTableLocked && !isPartitionLocked`?

time-and-fate: I think there are only two cases for the global stats now:
1. The table id and the partition id are both not locked: just directly dump the delta.
2. Otherwise, we neither dump nor "stash" the delta for the table id, and we just "stash" the delta for the partition id. And when unlocking, delta will be dumped for both partition id and table id.

statistics/handle/lockstats/lock_stats.go
```diff
@@ -42,15 +42,20 @@ var (
...
+	PartitionInfo map[int64]string
+	// schema name + table name.
+	FullName string
+}
+
 // AddLockedTables add locked tables id to store.
 // - exec: sql executor.
-// - tidAndNames: table ids and names of which will be locked.
-// - pidAndNames: partition ids and names of which will be locked.
+// - tables: tables of which will be locked.
```
time-and-fate, line 54:
```suggestion
// - tables: tables that will be locked.
```
I think the meanings are the same but simpler? :laughing:

statistics/handle/update.go
hi-rustin: I don't think we need this condition.

```go
		// If the partition is locked, we don't need to update the global-stats.
		// We will update its global-stats when the partition is unlocked.
		// 1. If table is locked and partition is locked, we only stash the delta in the partition's lock info.
		//    we will update its global-stats when the partition is unlocked.
		// 2. If table is locked and partition is not locked(new partition after lock), we only stash the delta in the table's lock info.
		//    we will update its global-stats when the table is unlocked. We don't need to specially handle this case.
		// 3. If table is not locked and partition is locked, we only stash the delta in the partition's lock info.
		//    we will update its global-stats when the partition is unlocked.
		// 4. If table is not locked and partition is not locked, we update the global-stats.
		// To sum up, we only need to update the global-stats when the partition is unlocked.
```

hi-rustin: Let me test it again.

hi-rustin: Updated! Thanks for your review! I am so buggy 😢

statistics/handle/lockstats/lock_stats.go
hi-rustin: Updated.

statistics/handle/lockstats/unlock_stats.go
```diff
@@ -26,20 +26,18 @@ import (
 
 const (
 	selectDeltaSQL = "SELECT count, modify_count, version FROM mysql.stats_table_locked WHERE table_id = %?"
-	updateDeltaSQL = "UPDATE mysql.stats_meta SET version = %?, count = count + %?, modify_count = modify_count + %? WHERE table_id = %?"
+	updateDeltaSQL = "UPDATE mysql.stats_meta SET version = %?, count = CASE WHEN %? < 0 THEN IF(count > ABS(%?), count - ABS(%?), 0) ELSE count + %? END, modify_count = modify_count + %? WHERE table_id = %?"
```
hi-rustin, line 29: @time-and-fate Do you have any ideas about how to improve this?

time-and-fate: Looks like these two expressions are equivalent.
```sql
CASE WHEN %? < 0 THEN
    IF(count > ABS(%?), count - ABS(%?), 0)
ELSE
    count + %?
END
```
```sql
IF(count + %? > 0, count + %?, 0)
```

hi-rustin: Good point!

hi-rustin: Updated in [`e852581` (#47319)](https://github.com/pingcap/tidb/pull/47319/commits/e8525818f7531a95fd3d7774345131148612cff8)

time-and-fate (approved):

zanmato1984 (approved):

PR #47376, opened 2023-09-29
snapshot_backup: deny executing tidb-lightning import while running s…

YuJuncen:
#47341


YuJuncen:
it has been replaced by the advanced prepare

Issue #46867, opened 2023-09-11
persist stats object consumes too much memory compared with the table meta


winoros:
## Enhancement

![image](https://github.com/pingcap/tidb/assets/7846227/151af028-d25c-4924-9975-44f34c24fe4c)

Here you can see that the newly created tables only costs 200+MB for its meta. But the persist stats object has already gone to 1+GB. Huge difference between them.

We need to first reduce the memory of the persist stats object to the same level as the table meta.

PR #47496, opened 2023-10-09
planner, stats: reduce the persist memory usage of stats after startup

winoros:

Issue Number: close https://github.com/pingcap/tidb/issues/46867

Problem Summary:

We've introduced lite init before. Only load the metadata(NDV, NULL count etc) of the column/index's statistics.
But the Column/Index's meta object still consumes too much memory.

So, in this pull, we use a tiny map to represent the existence of statistics of Column/Index. And load the stats only when it's used.

The changes in the pull are all around this idea.

And there's still some remaining problems to be addressed later, like that Pseudo table still costs much memory than we expected.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


statistics/table.go
```diff
@@ -68,6 +70,113 @@ type Table struct {
...
+
+// IsEmpty checks whether the map is empty.
+func (m *ColAndIndexExistenceMap) IsEmpty() bool {
+	return len(m.m1)+len(m.m2) == 0
+}
+
+// Clone deeply copies the map.
+func (m *ColAndIndexExistenceMap) Clone() *ColAndIndexExistenceMap {
+	mm := NewColAndIndexExistenceMap(len(m.m1), len(m.m2))
+	for k, v := range m.m1 {
```
qw4990, line 157: The name `m1` and `m2` seem a little unclear, can we rename them like `colInfo` and `indexInfo`?

statistics/table.go
```diff
@@ -610,45 +733,46 @@ func (coll *HistColl) GenerateHistCollFromColumnInfo(tblInfo *model.TableInfo, c
 // Usually, we don't want to trigger stats loading for pseudo table.
 // But there are exceptional cases. In such cases, we should pass allowTriggerLoading as true.
 // Such case could possibly happen in getStatsTable().
-func PseudoTable(tblInfo *model.TableInfo, allowTriggerLoading bool) *Table {
-	const fakePhysicalID int64 = -1
+func PseudoTable(tblInfo *model.TableInfo, allowTriggerLoading bool, allowFillHistMeta bool) *Table {
```
qw4990, line 736: Please add some comments to `allowFillHistMeta`.

statistics/table.go
```diff
@@ -68,6 +70,113 @@ type Table struct {
...
+	if isIndex {
+		analyzed, ok := m.indexAnalyzed[id]
+		return ok && analyzed
+	}
+	analyzed, ok := m.colAnalyzed[id]
+	return ok && analyzed
+}
+
+// InsertCol inserts a column with its meta into the map.
+func (m *ColAndIndexExistenceMap) InsertCol(id int64, info *model.ColumnInfo, analyzed bool) {
```
qw4990, line 128: Do we need to add a lock to make this structure thread-safe?

statistics/table.go
```diff
@@ -68,6 +70,113 @@ type Table struct {
 	// and the schema of the table does not change, we don't need to load the stats for this
 	// table again.
 	TblInfoUpdateTS uint64
+
+	IsPkIsHandle bool
+}
+
+// ColAndIndexExistenceMap is the meta map for statistics.Table.
+// It can tell whether a column/index really has its statistics. So we won't send useless kv request when we do online stats loading.
+type ColAndIndexExistenceMap struct {
```
qw4990, line 79: Great, this structure is helpful for us to identify the loading status of a table's statistics.

tests/integrationtest/r/explain_shard_index.result
```diff
@@ -19,8 +19,8 @@ Projection	10.00	root		explain_shard_index.test3.id, explain_shard_index.test3.a
   └─TableRowIDScan(Probe)	10.00	cop[tikv]	table:test3	keep order:false, stats:pseudo
 explain format=brief select * from test3 where a=100 or b = 200;
 id	estRows	task	access object	operator info
-Projection	8000.00	root		explain_shard_index.test3.id, explain_shard_index.test3.a, explain_shard_index.test3.b
-└─Selection	8000.00	root		or(and(eq(tidb_shard(explain_shard_index.test3.a), 8), eq(explain_shard_index.test3.a, 100)), eq(explain_shard_index.test3.b, 200))
+Projection	11.00	root		explain_shard_index.test3.id, explain_shard_index.test3.a, explain_shard_index.test3.b
+└─Selection	10.01	root		or(and(eq(tidb_shard(explain_shard_index.test3.a), 8), eq(explain_shard_index.test3.a, 100)), eq(explain_shard_index.test3.b, 200))
```
qw4990, line 23: Are these estimation changes OK?

statistics/handle/bootstrap.go
```diff
@@ -117,19 +114,7 @@ func (h *Handle) initStatsHistograms4ChunkLite(is infoschema.InfoSchema, cache *
...
-				Info:       idxInfo,
-				StatsVer:   statsVer,
-				Flag:       flag,
-				PhysicalID: tblID,
-			}
-			lastAnalyzePos.Copy(&index.LastAnalyzePos)
-			if index.IsAnalyzed() {
-				index.StatsLoadedStatus = statistics.NewStatsAllEvictedStatus()
-			}
-			table.Indices[hist.ID] = index
```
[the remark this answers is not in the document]
winoros: yes, the api can be refactored for better use later.

statistics/table.go
winoros: We don't need it. A simple proof is that the `map[int64]*Column` and `map[int64]*Index` also don't consider the concurrent updates.

statistics/table.go
winoros: Yes, it's a temp solution to add this new param. We meet the problem with the pseudo table's memory usage.
But fully refactoring needs to also change some estimation logic. After some hard trying, I removed those changes in the pull since this pull is already large enough.:joy:

statistics/table.go
winoros: yes, I used those names at the beginning since there're just two maps. But later I found that we need to extend to four maps to make this pull as small as possible.

tests/integrationtest/r/explain_shard_index.result
winoros: The estimation change should be looked into. I missed this change.

statistics/table.go
winoros: This pull reduces some part of the pseudo table's memory usage, but not fully resolve it🤣

tests/integrationtest/r/explain_shard_index.result
winoros: The most challenging problem of this pr is that the pseudo estimation logic is heavily influenced by this pr.
Many changes are to keep the pseudo-estimation behavior the same as before.

statistics/table.go
qw4990: OOOK, we can remove `StatsLoadedStatus` in the future, this structure is more accurate and specific than `StatsLoadedStatus` for identifying stats loading status.

statistics/index.go
```diff
@@ -127,23 +127,30 @@ func (idx *Index) TotalRowCount() float64 {
 	return idx.Histogram.TotalRowCount()
 }
 
-// IsInvalid checks if this index is invalid.
-func (idx *Index) IsInvalid(sctx sessionctx.Context, collPseudo bool) (res bool) {
-	idx.CheckStats()
+// IndexStatsIsInvalid checks whether the index has valid stats or not.
+func IndexStatsIsInvalid(idxStats *Index, sctx sessionctx.Context, coll *HistColl, cid int64) (res bool) {
```
qw4990, line 131: The first argument `idxStats` seems can be removed? Since `idxStats = coll.Index[cid]`.

statistics/column.go
```diff
@@ -140,13 +138,10 @@ func (c *Column) MemoryUsage() CacheItemMemoryUsage {
...
 
-// IsInvalid checks if this column is invalid.
+// ColumnStatsIsInvalid checks if this column is invalid.
 // If this column has histogram but not loaded yet,
 // then we mark it as need histogram.
-func (c *Column) IsInvalid(
-	sctx sessionctx.Context,
-	collPseudo bool,
-) (res bool) {
+func ColumnStatsIsInvalid(colStats *Column, sctx sessionctx.Context, histColl *HistColl, cid int64) (res bool) {
```
qw4990, line 144: `colsStats` seems redundant since it's equal to `histColl.Columns[cid]`, can we remove it?

time-and-fate (commented):
Looks like it's hard to make sure everything works well when sync load is disabled and we need the metadata like NDV.

In the case below, lazy load can't be triggered for `t.a`.
```sql
create table t(a int, b int);
--- insert some data and analyze it
explain select * from t t1 left join t t2 on t1.a = t2.b;
```

statistics/handle/storage/read.go
```diff
@@ -562,12 +593,32 @@ func loadNeededIndexHistograms(sctx sessionctx.Context, statsCache *cache.StatsC
...
+	if ok {
+		hgMeta = &index.Histogram
+		idxInfo = index.Info
+		flag = index.Flag
+		lastAnalyzePos = new(types.Datum)
+		lastAnalyzePos.SetUint64(index.LastUpdateVersion)
+	} else {
+		hgMeta, lastAnalyzePos, statsVer, flag, err = HistMetaFromStorage(sctx, &idx, nil)
+		if err != nil {
+			return err
```
AilinKid, line 617: should add `statistics.HistogramNeededItems.Delete` here like above？

statistics/handle/handle_hist.go
```diff
@@ -154,7 +155,7 @@ func (h *Handle) removeHistLoadedColumns(neededItems []model.TableItemID) []mode
 			continue
 		}
 		colHist, ok := tbl.Columns[item.ID]
-		if ok && colHist.IsStatsInitialized() && !colHist.IsFullLoad() {
+		if (!ok && tbl.ColAndIndexExistenceMap.HasAnalyzed(item.ID, false)) || (ok && colHist.IsStatsInitialized() && !colHist.IsFullLoad()) {
```
time-and-fate, line 158: Why do we need `HasAnalyzed` here? We are using `Has()` everywhere else.

statistics/handle/handle_hist.go
```diff
@@ -222,30 +223,39 @@ func (h *Handle) HandleOneTask(sctx sessionctx.Context, lastTask *NeededItemTask
...
-	result := stmtctx.StatsLoadResult{Item: task.TableItemID}
+	result := stmtctx.StatsLoadResult{Item: task.Item.TableItemID}
 	item := result.Item
 	oldCache := h.statsCache.Load()
 	tbl, ok := oldCache.Get(item.TableID)
 	if !ok {
 		h.writeToResultChan(task.ResultCh, result)
 		return nil, nil
 	}
+	logutil.BgLogger().Warn("?")
```
time-and-fate, line 234: Please remove debug logs.

statistics/table.go
```diff
@@ -68,6 +70,113 @@ type Table struct {
...
+		_, ok := m.m2[id]
+		return ok
+	}
+	_, ok := m.m1[id]
+	return ok
+}
+
+// HasAnalyzed checks whether a column/index stats exists and it has stats.
+// TODO: the map should only keep the analyzed cols.
+func (m *ColAndIndexExistenceMap) HasAnalyzed(id int64, isIndex bool) bool {
```
time-and-fate, line 118: I think this description is quite unclear. What does it mean by "a stats exists and it has stats"? There are also not any explanations for `struct ColAndIndexExistenceMap`.

statistics/table.go
time-and-fate, line 79: btw, shouldn't it be named `ColAndIdxExistenceMap` or `ColumnAndIndexExistenceMap`?

statistics/table.go
winoros: Currently, the map needs to tell
- whether this table has this index/column.
- whether this index/column has statistics.
I'll update the comment.

statistics/handle/handle_hist.go
winoros: Other places just need to know whether this column exists or not.
Here should tell whether this column really has stats or not.

planner/core/plan_stats.go
```diff
@@ -114,11 +114,9 @@ func RequestLoadStats(ctx sessionctx.Context, neededHistItems []model.TableItemI
 	if err != nil {
 		stmtCtx.IsSyncStatsFailed = true
 		if variable.StatsLoadPseudoTimeout.Load() {
-			logutil.BgLogger().Warn("RequestLoadStats failed", zap.Error(err))
 			stmtCtx.AppendWarning(err)
 			return nil
 		}
-		logutil.BgLogger().Error("RequestLoadStats failed", zap.Error(err))
```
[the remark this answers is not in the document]
winoros: It should be mistake.

planner/cardinality/selectivity.go
```diff
@@ -117,12 +117,16 @@ func Selectivity(
...
-		colStats := coll.Columns[id]
-		col := expression.ColInfo2Col(extractedCols, colStats.Info)
-		if col != nil {
+	slices.SortFunc(extractedCols, func(a *expression.Column, b *expression.Column) int {
+		return cmp.Compare(a.ID, b.ID)
+	})
+	for i, col := range extractedCols {
+		if i != 0 && extractedCols[i].ID == extractedCols[i-1].ID {
+			continue
+		}
```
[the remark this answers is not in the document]
winoros: The same column should be evaluated only once.

statistics/table.go
```diff
@@ -68,6 +70,113 @@ type Table struct {
...
+}
+
+// ColAndIndexExistenceMap is the meta map for statistics.Table.
+// It can tell whether a column/index really has its statistics. So we won't send useless kv request when we do online stats loading.
+type ColAndIndexExistenceMap struct {
+	m1            map[int64]*model.ColumnInfo
+	colAnalyzed   map[int64]bool
+	m2            map[int64]*model.IndexInfo
+	indexAnalyzed map[int64]bool
+	someAnalyzed  bool
```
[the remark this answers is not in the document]
winoros: Oh this field can be deleted.

statistics/handle/storage/read.go
winoros: Other places that meet errors also do not delete it from the map. I think this is the original behavior.

statistics/handle/handle_hist.go
time-and-fate: I think the question is: if the col/idx stats are not in the `tbl.Columns`/`tbl.Indices`, what needs to be satisfied to trigger a loading? `Has()` or `HasAnalyzed()`?

statistics/index.go
winoros: Here just want to save once of the accessing the map(

planner/cardinality/selectivity.go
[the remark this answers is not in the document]
time-and-fate: Please add a comment to make it explicit then. It's not obvious at first glance.
Or I think just replacing it with a dedup is also good.

statistics/index.go
qw4990: `Once of accessing the map` seems OK? 
If we keep both `*Index` and `*HistColl` here, it seems unsafe since there is an implicit assumption that `the *Index must come from the *HistColl`. 
Other developers may not know this implicit assumption, and what if the `*Index` is not from `*HistColl`...

statistics/handle/handle_hist.go
```diff
@@ -276,85 +285,92 @@ func (h *Handle) handleOneItemTask(sctx sessionctx.Context, task *NeededItemTask
...
-	if item.IsIndex {
-		isIndexFlag = 1
+	hg, lastAnalyzePos, statsVer, flag, err := storage.HistMetaFromStorage(sctx, &item, w.colInfo)
+	if err != nil {
+		return nil, err
+	}
+	if hg == nil {
+		logutil.BgLogger().Error("fail to get stats version for this histogram", zap.Int64("table_id", item.TableID),
+			zap.Int64("hist_id", item.ID), zap.Bool("is_index", item.IsIndex))
+		return nil, errors.Trace(fmt.Errorf("fail to get stats version for this histogram, table_id:%v, hist_id:%v, is_index:%v", item.TableID, item.ID, item.IsIndex))
```
time-and-fate, line 306: I think this log is misleading and insufficient:
1. We are not only fetching the stats version here.
2. It doesn't say that it's a failure during sync loading.
3. It doesn't make it clear that the reason is there is no corresponding row in `stats_histograms`.

statistics/column.go
```diff
@@ -165,29 +158,29 @@ func (c *Column) IsInvalid(
...
-			// In this case, we don't trigger stats loading.
-			if c.Info != nil && c.PhysicalID > 0 {
-				HistogramNeededItems.insert(model.TableItemID{TableID: c.PhysicalID, ID: c.Info.ID, IsIndex: false})
-			}
+		if (colStats == nil || !colStats.IsStatsInitialized() || colStats.IsLoadNeeded()) &&
+			stmtctx != nil &&
+			histColl.PhysicalID > 0 &&
+			!histColl.CanNotTriggerLoad {
+			HistogramNeededItems.insert(model.TableItemID{TableID: histColl.PhysicalID, ID: cid, IsIndex: false})
 		}
```
time-and-fate, line 166: Why are the logs and comments removed?

planner/cardinality/row_count_index.go
```diff
@@ -60,9 +60,7 @@ func GetRowCountByIndexRanges(sctx sessionctx.Context, coll *statistics.HistColl
 		}
 	}
 	recordUsedItemStatsStatus(sctx, idx, coll.PhysicalID, idxID)
-	// For the mv index case, now we have supported collecting stats and async loading stats, but sync loading and
-	// estimation is not well-supported, so we keep mv index using pseudo estimation for this period of time.
-	if !ok || idx.IsInvalid(sctx, coll.Pseudo) || isMVIndex {
+	if statistics.IndexStatsIsInvalid(idx, sctx, coll, idxID) || isMVIndex {
 		colsLen := -1
```
time-and-fate, line 64: Why is the comment removed?

statistics/index.go
```diff
@@ -127,23 +127,30 @@ func (idx *Index) TotalRowCount() float64 {
 	return idx.Histogram.TotalRowCount()
 }
 
-// IsInvalid checks if this index is invalid.
-func (idx *Index) IsInvalid(sctx sessionctx.Context, collPseudo bool) (res bool) {
-	idx.CheckStats()
```
time-and-fate, line 132: `(*Index).CheckStats()` can be removed now. This is the only place using it.

statistics/index.go
```diff
@@ -127,23 +127,30 @@ func (idx *Index) TotalRowCount() float64 {
...
+// IndexStatsIsInvalid checks whether the index has valid stats or not.
+func IndexStatsIsInvalid(idxStats *Index, sctx sessionctx.Context, coll *HistColl, cid int64) (res bool) {
+	// When we are using stats from PseudoTable(), all column/index ID will be -1.
+	if ((idxStats != nil && idxStats.IsFullLoad()) || coll.PhysicalID <= 0) && !coll.CanNotTriggerLoad {
```
time-and-fate, line 133: I think this check is wrong.

time-and-fate: And can we have a unified checking in this method instead of two different points of trigger loading?
Comments are also needed.

time-and-fate (commented):
As commented yesterday, I think we need to make one thing clear: when should we trigger a loading?
The sync/lazy and index/column all have different logic. There are many assumptions and considerations not clarified.

statistics/handle/storage/read.go
```diff
@@ -592,19 +643,21 @@ func loadNeededIndexHistograms(sctx sessionctx.Context, statsCache util.StatsCac
...
+	if statsVer == -1 {
+		rows, _, err := util.ExecRows(sctx, "select stats_ver from mysql.stats_histograms where is_index = 1 and table_id = %? and hist_id = %?", idx.TableID, idx.ID)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		if len(rows) == 0 {
+			logutil.BgLogger().Error("fail to get stats version for this histogram", zap.Int64("table_id", idx.TableID), zap.Int64("hist_id", idx.ID))
+			return errors.Trace(fmt.Errorf("fail to get stats version for this histogram, table_id:%v, hist_id:%v", idx.TableID, idx.ID))
+		}
 	}
```
time-and-fate, line 655: Forget to use the query result?

planner/cardinality/row_count_index.go
winoros: Should be mistake when resolving merging conflicts

statistics/column.go
winoros: Should be mistake when resolving merging conflicts

winoros: That check became the condition `histColl.CanNotTriggerLoad`.

statistics/index.go
winoros:
```
	// When we are using stats from PseudoTable(), all column/index ID will be -1.
	if idx.IsFullLoad() || idx.PhysicalID <= 0 {
		return
	}
	HistogramNeededItems.insert(model.TableItemID{TableID: idx.PhysicalID, ID: idx.Info.ID, IsIndex: true})
```
It's copied from the `CheckStats`. But yes, we can redesign it. I just want to make the review more easier by copying it instead of fully refactoring it(

statistics/handle/storage/read.go
winoros: Which result?

time-and-fate: The query result. You need to use it to update the `statsVer`, right?

pkg/sessionctx/stmtctx/stmtctx.go
```diff
@@ -1458,6 +1458,8 @@ type UsedStatsInfoForTable struct {
 	ModifyCount           int64
 	ColumnStatsLoadStatus map[int64]string
 	IndexStatsLoadStatus  map[int64]string
+	StatsTbl              interface{}
+	ColAndIdxStatus       interface{}
```
qw4990, line 1462: Could you add more comments about these 2 fields? Add can we use a specific type instead of interface{} here?

pkg/planner/core/collect_column_stats_usage.go
```diff
@@ -39,8 +41,8 @@ type columnStatsUsageCollector struct {
 	// we don't know `ndv(t.a, t.b)`(see (*LogicalAggregation).DeriveStats and getColsNDV for details). So when calculating the statistics
 	// of column `e`, we may use the statistics of column `t.a` and `t.b`.
 	colMap map[int64]map[model.TableItemID]struct{}
-	// histNeededCols records histogram-needed columns
-	histNeededCols map[model.TableItemID]struct{}
+	// histNeededCols records histogram-needed columns. The value field of the map indicates that whether we need to load the full stats of the time or not.
```
qw4990, line 44: `time` should be `item` here?

winoros: yes

pkg/sessionctx/stmtctx/stmtctx.go
winoros: It will cause cyclic importing

pkg/statistics/handle/storage/read.go
```diff
@@ -246,18 +275,8 @@ func indexStatsFromStorage(sctx sessionctx.Context, row chunk.Row, table *statis
...
-				Flag:       flag,
-				PhysicalID: table.PhysicalID,
-			}
-			if idx.IsAnalyzed() {
-				idx.StatsLoadedStatus = statistics.NewStatsAllEvictedStatus()
-			}
-			lastAnalyzePos.Copy(&idx.LastAnalyzePos)
-			break
+			delete(table.Indices, histID)
+			return nil
```
time-and-fate, line 279: I think this is a dangerous change.
Before, we never proactively kick stats out of the stats cache. The only possibility that would happen is eviction caused by memory insufficiency.
My suggestion is that, if this column already has metadata, we just keep it with metadata. And only let the LRU handle the eviction.

pkg/statistics/handle/storage/read.go
```diff
@@ -343,20 +363,8 @@ func columnStatsFromStorage(sctx sessionctx.Context, row chunk.Row, table *stati
...
-				StatsVer:   statsVer,
-			}
-			if col.StatsAvailable() {
-				col.StatsLoadedStatus = statistics.NewStatsAllEvictedStatus()
-			}
-			lastAnalyzePos.Copy(&col.LastAnalyzePos)
-			col.Histogram.Correlation = correlation
-			break
+			delete(table.Columns, histID)
+			return nil
```
time-and-fate, line 367: ditto.

pkg/statistics/handle/storage/read.go
```diff
@@ -562,9 +642,15 @@ func loadNeededColumnHistograms(sctx sessionctx.Context, statsCache statstypes.S
...
-	tbl.Columns[c.ID] = colHist
+	tbl.Columns[col.ID] = colHist
 	statsCache.UpdateStatsCache([]*statistics.Table{tbl}, nil)
 	statistics.HistogramNeededItems.Delete(col)
+	if sctx.GetSessionVars().StmtCtx.StatsLoad.Timeout > 0 {
+		logutil.BgLogger().Warn("Hist for column should already be loaded as sync but not found.",
+			zap.Int64("table_id", c.PhysicalID),
+			zap.Int64("column_id", c.Info.ID),
+			zap.String("column_name", c.Info.Name.O))
+	}
```
time-and-fate, line 653: I think the log here is incorrect.

pkg/statistics/handle/storage/read.go
```diff
@@ -533,23 +611,25 @@ func loadNeededColumnHistograms(sctx sessionctx.Context, statsCache statstypes.S
...
+	if statsVer == -1 {
+		rows, _, err := util.ExecRows(sctx, "select stats_ver from mysql.stats_histograms where is_index = 0 and table_id = %? and hist_id = %?", col.TableID, col.ID)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		if len(rows) == 0 {
+			logutil.BgLogger().Error("fail to get stats version for this histogram", zap.Int64("table_id", col.TableID), zap.Int64("hist_id", col.ID))
+			return errors.Trace(fmt.Errorf("fail to get stats version for this histogram, table_id:%v, hist_id:%v", col.TableID, col.ID))
+		}
+		statsVer = rows[0].GetInt64(0)
```
time-and-fate, line 623: I think this logic can be merged into `HistMetaFromStorage` above. Now we always need to access the table once.

pkg/statistics/handle/storage/read.go
```diff
@@ -593,19 +700,21 @@ func loadNeededIndexHistograms(sctx sessionctx.Context, statsCache statstypes.St
...
-		return errors.Trace(fmt.Errorf("fail to get stats version for this histogram, table_id:%v, hist_id:%v", idx.TableID, idx.ID))
+	if statsVer == -1 {
+		rows, _, err := util.ExecRows(sctx, "select stats_ver from mysql.stats_histograms where is_index = 1 and table_id = %? and hist_id = %?", idx.TableID, idx.ID)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		if len(rows) == 0 {
+			logutil.BgLogger().Error("fail to get stats version for this histogram", zap.Int64("table_id", idx.TableID), zap.Int64("hist_id", idx.ID))
+			return errors.Trace(fmt.Errorf("fail to get stats version for this histogram, table_id:%v, hist_id:%v", idx.TableID, idx.ID))
+		}
```
time-and-fate, line 711: ditto.

pkg/statistics/handle/storage/read.go
time-and-fate: And looks like the `if c != nil {` branch above can be removed.

pkg/sessionctx/stmtctx/stmtctx.go
```diff
@@ -1400,6 +1400,8 @@ type UsedStatsInfoForTable struct {
 	ModifyCount           int64
 	ColumnStatsLoadStatus map[int64]string
 	IndexStatsLoadStatus  map[int64]string
+	StatsTbl              any
```
time-and-fate, line 1403: Not used.

pkg/statistics/index.go
```diff
@@ -127,23 +127,31 @@ func (idx *Index) TotalRowCount() float64 {
...
 				"TotalCount", totalCount,
 			)
 			debugtrace.LeaveContextCommon(sctx)
 		}()
 	}
-	totalCount = idx.TotalRowCount()
-	return (collPseudo) || totalCount == 0
+	// If the given index statistics is nil or we found that the index's statistics hasn't been fully loaded, we add this index to NeededItems.
+	// Also, we need to check that this HistColl has its physical ID and it is permitted to trigger the stats loading.
+	if (idxStats == nil || !idxStats.IsFullLoad()) && coll.PhysicalID > 0 && !coll.CanNotTriggerLoad {
```
time-and-fate, line 146: Looks like we can reuse `IndexIsLoadNeeded` here.

pkg/statistics/column.go
```diff
@@ -163,31 +158,29 @@ func (c *Column) IsInvalid(
...
-					zap.Int64("column_id", c.Info.ID),
-					zap.String("column_name", c.Info.Name.O))
-			}
-			// In some tests, the c.Info is not set, so we add this check here.
-			// When we are using stats from PseudoTable(), the table ID will possibly be -1.
-			// In this case, we don't trigger stats loading.
-			if c.Info != nil && c.PhysicalID > 0 {
-				HistogramNeededItems.insert(model.TableItemID{TableID: c.PhysicalID, ID: c.Info.ID, IsIndex: false})
-			}
+		if (colStats == nil || !colStats.IsStatsInitialized() || colStats.IsLoadNeeded()) &&
```
time-and-fate, line 161: Ditto, but `ColumnIsLoadNeeded`.

pkg/statistics/column.go
```diff
@@ -163,31 +158,29 @@ func (c *Column) IsInvalid(
...
+			stmtctx != nil &&
+			histColl.PhysicalID > 0 &&
```
time-and-fate, line 163: I think this won't be equal to or less than 0 and can be removed.

pkg/statistics/index.go
time-and-fate: I think `coll.PhysicalID` won't be equal to or less than 0 and can be removed.

winoros: The IndexIsLoadNeeded is attached on the `statistics.Table`, not the `statistics.HistColl`

pkg/statistics/handle/syncload/stats_syncload.go
```diff
@@ -456,18 +472,27 @@ func (s *statsSyncLoad) updateCachedItem(item model.TableItemID, colHist *statis
 	}
 	if !item.IsIndex && colHist != nil {
 		c, ok := tbl.Columns[item.ID]
-		if !ok || c.IsFullLoad() {
+		// If we don't have this column or its stats is fully loaded, we skip it.
+		if (!ok && !tbl.ColAndIdxExistenceMap.Has(item.ID, false)) || (ok && c.IsFullLoad()) {
```
time-and-fate, line 476: Why do need this `Has()` check?

pkg/statistics/handle/syncload/stats_syncload.go
```diff
@@ -456,18 +472,27 @@ func (s *statsSyncLoad) updateCachedItem(item model.TableItemID, colHist *statis
...
 			return true
 		}
 		tbl = tbl.Copy()
-		tbl.Columns[c.ID] = colHist
+		tbl.Columns[item.ID] = colHist
+		// If the column is analyzed we refresh the map for the possible change.
+		if colHist.StatsAvailable() {
+			tbl.ColAndIdxExistenceMap.InsertCol(item.ID, colHist.Info, true)
```
time-and-fate, line 483: I think this is not necessary, right?

time-and-fate (commented):
Should `tryAutoAnalyzeTable()` also be updated?
Looks like it should use `HasAnalyzed` to check if a new index hasn't been analyzed yet.

time-and-fate:
> Should `tryAutoAnalyzeTable()` also be updated? Looks like it should use `HasAnalyzed` to check if a new index hasn't been analyzed yet.

And also the newly introduced `checkIndexesNeedAnalyze()`.

pkg/planner/core/collect_column_stats_usage.go
```diff
@@ -163,10 +166,33 @@ func (c *columnStatsUsageCollector) addHistNeededColumns(ds *DataSource) {
...
+	stats := domain.GetDomain(ds.SCtx()).StatsHandle()
+	tblStats := stats.GetPartitionStats(ds.tableInfo, ds.physicalTableID)
+	skipPseudoCheckForTest := false
+	failpoint.Inject("disablePseudoCheck", func() {
+		skipPseudoCheckForTest = true
+	})
+	// Since we can not get the stats tbl, this table is not analyzed. So we don't need to consider load stats.
+	if tblStats.Pseudo && !skipPseudoCheckForTest {
+		return
+	}
```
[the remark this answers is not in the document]
winoros: To reduce the possible background SQL

pkg/statistics/handle/syncload/stats_syncload.go
winoros: From the newly added test, i think it's yes.

pkg/planner/core/plan_stats.go
```diff
@@ -114,11 +114,9 @@ func RequestLoadStats(ctx PlanContext, neededHistItems []model.TableItemID, sync
 	if err != nil {
 		stmtCtx.IsSyncStatsFailed = true
 		if variable.StatsLoadPseudoTimeout.Load() {
-			logutil.BgLogger().Warn("RequestLoadStats failed", zap.Error(err))
 			stmtCtx.AppendWarning(err)
 			return nil
 		}
-		logutil.BgLogger().Error("RequestLoadStats failed", zap.Error(err))
```
hawkingrei, line 121: This log should be retained to facilitate problem discovery from the log.

pkg/planner/cardinality/selectivity.go
```diff
@@ -119,12 +119,16 @@ func Selectivity(
...
-		colStats := coll.Columns[id]
-		col := expression.ColInfo2Col(extractedCols, colStats.Info)
-		if col != nil {
+	slices.SortFunc(extractedCols, func(a *expression.Column, b *expression.Column) int {
+		return cmp.Compare(a.ID, b.ID)
+	})
+	for i, col := range extractedCols {
+		if i != 0 && extractedCols[i].ID == extractedCols[i-1].ID {
+			continue
+		}
```
time-and-fate, line 128: We can use `slices.CompactFunc` from `golang.org/x/exp/slices` to dedup `extractedCols`. I think that will be easier to understand.

hawkingrei (approved):
LGTM

tangenta (approved):

PR #47526, opened 2023-10-10
planner: Refactor framework of rule engine.(Support interaction rule of new rule engine )

### LINKED ISSUE #43360: A new rule engine for Optimizer
## Feature Request

Current optimizer rewrite rules are executed sequentially once and some specific rules (column prunning) are done multiple times hard coded. 
- Applying rules multiple times is not possible if a rule triggers another rule (for example: eliminating projections triggers join elimination). Missing such cases are lost optimization opportunities. 

The New rule engine for pre-processing phase changes logical plan to another logical plan.
It also should satisfy following requirements:

- Applying rules multiple times by rule engine instead of hard code manually. 
- Clean fast path for fast queries
- Easier testing for enabling/disabling rules
- Adding new rules is easier with this new architecture 

**Design**

Design

- Code factoring:
  - Functional components (FD, derived statistics and hint) out of logical plan building. #43361
  - Make sure each optimization rewrite has no side effects. If any, remove the side and apply it as a functional rewrite. 
- Rule engine design is based on the following abstraction
  - List of API's for rule1, rule2, ... RuleN. (ordered rule set as set of interfaces) 
  - Order of rules "RuleOrder" = {RuleX, RuleY, ....}
  - 2 dimensional array of rule interactions where M(i,j) = true means that If Rulei is applied then we should also do Rulej

The final product is depicted in the diagram below where we do pre-processing (FD, derived stats and hint annotations) of rewrites first and then we execute the rewrite engine.

image

elsa0520:

This PR mainly add the interaction rule framework in new rule engine. It doesn't be enabled by default.
The interaction rule will be trigger when it satisfies following conditions:
1. The related rule has been trigger and changed the plan
2. The interaction rule is enabled

The `optimizer` interface of logical rule has been changed. This PR add a new bool return value which used to flag whether plan changed or not during this logical rule.
If the logical rule changed the plan, it will return true. 

In this PR, I just add a new framework and not enable the interaction progress. So all of rules plan changed flag is false.

Issue Number: close #43360

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > The basic function of all logical rule has been covered by existing tests. Because this is an architectural adjustment, its correctness can be verified as long as the existing test can pass.

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/planner/core/optimizer.go
+++ b/planner/core/optimizer.go
@@ -117,6 +117,13 @@ var optRuleList = []logicalOptRule{
 	&resolveExpand{},
 }
 
+// Interaction Rule List
+/* The interaction rule will be trigger when it satisfies following conditions:
+1. The related rule has been trigger and changed the plan
+2. The interaction rule is enabled
+*/
+var optInteractionRuleList = map[logicalOptRule]logicalOptRule{}
+
 type logicalOptimizeOp struct {
 	// tracer is goring to track optimize steps during rule optimizing
 	tracer *tracing.LogicalOptimizeTracer
@@ -1125,6 +1132,7 @@ func logicalOptimize(ctx context.Context, flag uint64, logic LogicalPlan) (Logic
 		}()
 	}
 	var err error
+	var againRuleList []logicalOptRule
 	for i, rule := range optRuleList {
 		// The order of flags is same as the order of optRule in the list.
 		// We use a bitmask to record which opt rules should be used. If the i-th bit is 1, it means we should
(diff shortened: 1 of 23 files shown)


planner/core/optimizer.go
```diff
@@ -1133,11 +1148,27 @@ func logicalOptimize(ctx context.Context, flag uint64, logic LogicalPlan) (Logic
...
+		interactionRule, ok := optInteractionRuleList[rule]
+		if planChanged && ok && isLogicalRuleDisabled(interactionRule) {
+			againRuleList = append(againRuleList, interactionRule)
+		}
+	}
+
+	// Trigger the interaction rule
+	for i, rule := range againRuleList {
+		opt.appendBeforeRuleOptimize(i, rule.name(), logic)
+		logic, _, err = rule.optimize(ctx, logic, opt)
```
qw4990, line 1166: Should we check the `planChanged` again? And then trigger these interaction rules again until there is no change?

planner/core/plan_stats.go
```diff
@@ -33,9 +33,10 @@ import (
 
 type collectPredicateColumnsPoint struct{}
 
-func (collectPredicateColumnsPoint) optimize(_ context.Context, plan LogicalPlan, _ *logicalOptimizeOp) (LogicalPlan, error) {
+func (collectPredicateColumnsPoint) optimize(_ context.Context, plan LogicalPlan, _ *logicalOptimizeOp) (LogicalPlan, bool, error) {
```
winoros, line 36: I think we need to add a TODO here. The sync load/col usage analysis should be moved out of from the rule iteration.

qw4990 (approved):

planner/core/plan_stats.go
winoros: It relies on the predicate push-down currently. But i think it will be better if we detach it and move it out as a pre-opt operation.

elsa0520: There is two stats rule in Rule List : collectPredicateColumnsPoint,  syncWaitStatsLoadPoint.
My origin plan about New Rule Engine is separate those stats rule from Rule List.

We can come down and talk about the risks. Because it seems that some rules rely on statistical information?

winoros (approved):

PR #47552, opened 2023-10-11
*: Add `Assert` to do assertions in test

### LINKED ISSUE #47551: add a new function `Assert` to do some assertions in test
## Enhancement
 
Usually, we have some constraints in code, but it is not applicable to check it every time in the production environment. So we propose a new function `Assert` to do some assertions. The `Assert` function only works in test environment, and in a release build, it will return directly. It's implement is like this:

```go
func Assert(condition any, msgAndArgs ...any) {
    if intest.Intest { // intest.Intest is true only in test env
        doAssert(condition, msgAndArgs...)
    }
}
```

We can use this function like below:

```go
var foo *Foo
foo = ...

 // assert a bool is true
assert.Assert(foo != nil);

// or directly assert a pointer is not nil
assert.Assert(foo)

// assert the function's return value is true
assert.FuncAssert(func() bool) {
    return foo != nil
}

// with some custom messages
assert.Assert(foo != nil, "foo should not be nil")
assert.Assert(a == 0, "a should be zeror, but current is %d", a)
```

if you have some code that is sensitive in performance and don't want extra cost in stack jump, you can check if `intest.Intest` is true before assert.

```go
if intest.Test {
    assert.Assert(foo != nil)
}
```

However, I wonder if it is necessary because go will auto inline simple functions in most cases.

lcwangchao:

Issue Number: close #47551

See description in issue

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/util/intest/assert.go
+++ b/util/intest/assert.go
@@ -1,0 +1,83 @@
+// Copyright 2023 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build intest
+
+package intest
+
+import (
+	"fmt"
+	"reflect"
+)
+
+// Assert asserts a condition. It only works in test (intest.InTest == true).
+// You can assert a condition like this to assert a variable `foo` is not nil: `assert.Assert(foo != nil)`.
+// Or you can pass foo as a parameter directly for simple: `assert.Assert(foo)`
+// You can also assert a function that returns a bool: `intest.Assert(func() bool { return foo != nil })`
+// If you pass a function without a signature `func() bool`, the function will always panic.
+func Assert(cond any, msgAndArgs ...any) {
+	if InTest {
+		assert(cond, msgAndArgs...)
+	}
+}
+
+// AssertFunc asserts a function condition
+func AssertFunc(fn func() bool, msgAndArgs ...any) {
+	if InTest {
+		assert(fn(), msgAndArgs...)
+	}
+}
+
+func assert(cond any, msgAndArgs ...any) {
+	if !checkAssertObject(cond) {
+		doPanic(msgAndArgs...)
+	}
+}
+
+func doPanic(msgAndArgs ...any) {
+	panic(assertionFailedMsg(msgAndArgs...))
+}
+
+func assertionFailedMsg(msgAndArgs ...any) string {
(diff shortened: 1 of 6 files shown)


tiancaiamao (approved):

lance6716:
can you check that go's compiler really eliminated the Assert function? And maybe we can define an empty Assert with `//go:build !intest`

lcwangchao:
> can you check that go's compiler really eliminated the Assert function? And maybe we can define an empty Assert with `//go:build !intest`

That's a good idea. I have updated the PR.

util/intest/assert.go
```diff
@@ -0,0 +1,83 @@
...
+	"reflect"
+)
+
+// Assert asserts a condition. It only works in test (intest.InTest == true).
+// You can assert a condition like this to assert a variable `foo` is not nil: `assert.Assert(foo != nil)`.
+// Or you can pass foo as a parameter directly for simple: `assert.Assert(foo)`
+// You can also assert a function that returns a bool: `intest.Assert(func() bool { return foo != nil })`
+// If you pass a function without a signature `func() bool`, the function will always panic.
+func Assert(cond any, msgAndArgs ...any) {
+	if InTest {
```
lance6716, line 30: we can rely on `InTest` is always true when `//go:build intest`, or to make it more obvious, merge assert.go and intest.go into one file.

util/intest/assert_stub.go
lance6716, line 1: also maybe merge this file with common.go

lance6716 (commented):
rest lgtm

util/intest/assert_stub.go
lcwangchao: done

util/intest/assert.go
lcwangchao: I prefer to leave it here. Some PR are trying to use `testing.Testing()` to replace `inTest` flag. Though this PR cannot merge because of some bazel bug now, but one day it should merge to master, it's a good way to make sure this place can be modified in right way.

lcwangchao: We can still eliminated the Assert function in build phase at that time if we want. just to add a new flag `inRelease` when building a product version...

YangKeao (approved):

util/intest/assert.go
lance6716: > Some PR are trying to use testing.Testing() to replace inTest flag

Do you expect Assert should be enabled for testing.Testing()? With the "intest" build tag Assert is no-op.

> just to add a new flag inRelease when building a product version...

makes me think intest and indebug/inrelease are two separate things.

lcwangchao:
```suggestion
Do you expect Assert should be enabled for testing.Testing()? With the "intest" build tag Assert is no-op.
```

Yes，at that time intest flag may be removed.

```
makes me think intest and indebug/inrelease are two separate things.
```

At that time, `inrelease` tag will be used to optimize some performance like `-O` in C, but it may not be very necessary because I don't think it has a very big performance affect.

lance6716 (approved):

hawkingrei (approved):

PR #47620, opened 2023-10-13
table/tables: fix cached table size limitation violation during inserting

### LINKED ISSUE #47619: Insert to a cached table larger than 64M does not report error
## Bug Report

There is a size limitation for cached table, and when a cached table is too large, insert to it should report error.

```
# TestInsertTableTooLargeErr
drop table if exists t,t2;
# -- create a table first to provide the data
create table t (pad varchar(200));
insert into t values (repeat('x', 200));
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
# -- create a cached table
create table t2 like t;
alter table t2 cache;
# -- fill data and check the error
insert into t2 select * from t;
insert into t2 select * from t;
insert into t2 select * from t;
insert into t2 select * from t;
--error table too large
insert into t2 select * from t;
```

It's expected to get an error when the cached table t2 is big enough:

```
--error table too large
insert into t2 select * from t;
```

No error ...

master  f4b287db7f0c6890e381764151a07ecf08c362d7

tiancaiamao:

Issue Number: close #47619

Problem Summary:

The old code check the size limitation indeed, 

```
	if atomic.LoadInt64(&c.totalSize) > cachedTableSizeLimit {
		return nil, table.ErrOptOnCacheTable.GenWithStackByArgs("table too large")
	}
```

But the value of table size `c.totalSize` is loaded just when the cached table been loaded, and **never change**.

So if a user just 'insert ... insert ... insert ...', and never 'select ...', the table cache is not loaded, and the table size value is stale, and result in wrong behaviour.

This commit will try to update date the `c.totalSize` on insert and update, thus make the size limitation check work in that case.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


Defined2014 (approved):

pkg/table/tables/cache.go
```diff
@@ -261,6 +278,12 @@ func txnCtxAddCachedTable(sctx sessionctx.Context, tid int64, handle *cachedTabl
 
 // UpdateRecord implements table.Table
 func (c *cachedTable) UpdateRecord(ctx context.Context, sctx sessionctx.Context, h kv.Handle, oldData, newData []types.Datum, touched []bool) error {
+	txn, err := sctx.Txn(true)
+	if err != nil {
+		return err
+	}
+	defer handleTableSizeDelta(c, txn.Size(), txn)
```
djshow832, line 285: Why change the size when updating the record instead of removing the record?

tests/integrationtest/t/table/cache.test
```diff
@@ -118,3 +118,34 @@ select * from cache_add_column use index(k) where f1 = 1;
...
+# -- create a cached table
+create table t2 like t;
+alter table t2 cache;
+# -- fill data and check the error
+insert into t2 select * from t;
+insert into t2 select * from t;
+insert into t2 select * from t;
+insert into t2 select * from t;
+--error table too large
+insert into t2 select * from t;
```
djshow832, line 151: Please add an ending line here.

pkg/table/tables/cache.go
tiancaiamao: Because we just care the size limit of the table, we just need to check the growth of the table size, **remove will not make the table growth**.

One possible fail is that the user remove ... remove ... remove ... and insert/update, the table size is not accurate enough because we don't take the remove into account.
But the real table size is get from loading the cached, that's more accurate than calculate the size by transaction's size delta.

PR #47661, opened 2023-10-16
linter: setup a linter to validate constructor

### LINKED ISSUE #47660: Add a linter to ensure the `StmtCtx` is constructed correctly
## Enhancement

In https://github.com/pingcap/tidb/pull/47592, we added a new constructor for `StmtCtx`. However, it's hard to ensure all contexts are created through these constructors, and we cannot make sure that the `TypeContext` of the statement context is created correctly (even without considering the time zone info).

Maybe the suggested way (in golang) is to make it a private struct, and only provide the constructors to create it. However, it's impossible for now considering the existing big code base. I'd like to develop a linter to validate whether there are some manually created `StmtCtx`, and report errors.

YangKeao:

Issue Number: close #47660

Add a linter to ensure the `stmtCtx` is constructed correctly, at least the `AppendWarning` of `TypeCtx` is set.

Problem Summary:

A simple linter, as documented in the comment, it will detect the following cases:

1. Create struct directly, like `SomeStruct{}` or `&SomeStruct{}`.
2. Create struct with `new`: `new(SomeStruct)`
3. Struct literal in slice or other struct: `[]SomeStruct{{}}, []*SomeStruct{{}}`
4. Define variables through `var`: `var a SomeStruct`.
5. Implicit zero value in other struct literal: `type OtherStruct struct{SomeStruct}; other := OtherStruct{}`

After this PR is merged, the `StmtCtx` can only be constructed through:

```go
stmtctx.NewStmtCtx()
stmtctx.NewStmtCtxWithTimeZone()
```

Or `Reset` an existing `StmtCtx`.

Though we still exposed the `TypeCtx` and the code can still modify the `TimeZone` and `TypeCtx` manually, but it's easier to modify and protect them in the future (e.g. make the `TypeCtx` private).

To add the constructor validator to a struct, you can add the `_ constructor.Constructor` field like the following example:

```go
type someStruct struct {
    _ constructor.Constructor `ctor:"NewSomeStruct"`
}
```

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  >   

### Release note

```release-note
None
```

### DIFF
--- a/build/linter/constructor/analyzer.go
+++ b/build/linter/constructor/analyzer.go
@@ -1,0 +1,176 @@
+// Copyright 2023 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package constructor
+
+import (
+	"go/ast"
+	"go/types"
+	"slices"
+	"strings"
+
+	"github.com/fatih/structtag"
+	"golang.org/x/tools/go/analysis"
+	"golang.org/x/tools/go/ast/inspector"
+)
+
+// Analyzer is the analyzer struct of constructor.
+// constructor only allows constructing a struct manually in some specific functions, which is specified with tags for
+// `constructor.Constructor` field.
+//
+// It can detect the following pattern and give error (if not in constructor functions):
+//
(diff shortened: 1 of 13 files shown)


build/linter/constructor/analyzer.go
```diff
@@ -0,0 +1,176 @@
...
+//
+// TODO: verify whether this linter can work well with generics
+var Analyzer = &analysis.Analyzer{
+	Name:     "constructor",
+	Doc:      "Check developers don't create structs manually without using constructors",
+	Requires: []*analysis.Analyzer{},
+	Run:      run,
+}
+
+func getConstructorList(t types.Type) []string {
```
YangKeao, line 48: I'm not sure whether (and how) I need to add a cache to this function. I don't know whether this linter will make the compilation significantly slower.

But I think it's not worse than the `noCopy` linter (see https://github.com/golang/tools/blob/12b5dadef96835a9e447f2079a6cb949220c3856/go/analysis/passes/copylock/copylock.go#L330)

xhebox (approved):

YangKeao:
```
analysistest.go:283: go tool not available: exec: "go": executable file not found in $PATH
```
🤔 
/retest

PR #47671, opened 2023-10-16
planner: implementing multi index merge from CNF/DNF condition | tidb-test=pr/2229

AilinKid:

Issue Number: close https://github.com/pingcap/tidb/issues/47759

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
planner: implementing index merge from CNF condition
```


pkg/planner/core/indexmerge_path.go
```diff
@@ -596,7 +867,144 @@ func (ds *DataSource) generateIndexMergeOnDNF4MVIndex(normalPathCnt int, filters
 	return
 }
 
-// generateIndexMergeJSONMVIndexPath generates paths for (json_member_of / json_overlaps / json_contains) on multi-valued index.
+// generateIndexMerge4ComposedIndex generates index path composed of multi indexes including multivalued index from
+// (json_member_of / json_overlaps / json_contains) and single-valued index from normal indexes.
+/*
+	1. select * from t where ((1 member of (a) and c=1) and (2 member of (b) and d=2) and (other index predicates))
```
qw4990, line 873: Could we add some comments about DNF?

pkg/planner/core/indexmerge_path.go
```diff
@@ -596,7 +867,144 @@ func (ds *DataSource) generateIndexMergeOnDNF4MVIndex(normalPathCnt int, filters
...
+func (ds *DataSource) generateIndexMerge4ComposedIndex(normalPathCnt int,
+	indexMergeConds []expression.Expression) error {
+	isPossibleIdxMerge := len(indexMergeConds) > 0 && // have corresponding access conditions, and
+		len(ds.possibleAccessPaths) > 1 // have multiple index paths
+	if !isPossibleIdxMerge {
+		return nil
+	}
+	if strings.HasPrefix(ds.SCtx().GetSessionVars().StmtCtx.OriginalSQL, "explain select /*+ use_index_merge(t2, "+
+		"idx2, idx) */ * from t2 where ( json_overlaps") {
+		fmt.Println(1)
```
qw4990, line 914: `fmt.Println(1)` should be removed?

pkg/planner/core/indexmerge_path.go
```diff
@@ -846,6 +1254,66 @@ func (ds *DataSource) collectFilters4MVIndex(filters []expression.Expression, id
...
+// about them should be built as self-independent index path, deriving the final intersection/union handles,
+// which means a mv index path may be reused for multi related conditions. Here means whether (2 member of a)
+// And (1 member of a) is valid composed range or empty range can't be told until runtime intersection/union.
+//
+// therefore, for multi condition about a single mv index virtual json col here: (2 member of a) and (1 member of a)
+// we should build indexMerge above them, and each of them can access to the same mv index. That's why
+// we should derive the mutations of virtual json col's access condition, output the accessFilter combination
+// for each mutation of it.
+
+// todo: case like idx(x, cast(a as array), z), condition like: x=1 and x=2 and (
```
qw4990, line 1275: If 'x' is a normal column, should 'x=1 and x=2' always be false?

pkg/planner/core/indexmerge_path.go
```diff
@@ -846,6 +1254,66 @@ func (ds *DataSource) collectFilters4MVIndex(filters []expression.Expression, id
...
+// 2 member of a)? we can derive the x is invalid range?
+func (ds *DataSource) collectFilters4MVIndexMutations(filters []expression.Expression,
+	idxCols []*expression.Column) (accessFilters, remainingFilters []expression.Expression, mvColOffset int, mvFilterMutations []expression.Expression) {
```
qw4990, line 1278: Could you add a concrete example int the comment? For example, `for input 'filters' and 'cols', the output is ...`

pkg/planner/core/indexmerge_path.go
```diff
@@ -846,6 +1254,66 @@ func (ds *DataSource) collectFilters4MVIndex(filters []expression.Expression, id
...
+			if usedAsAccess[i] {
+				continue
+			}
+			if ds.checkFilter4MVIndexColumn(f, col) {
+				// todo: ban the virtual col in mv index except the json col it self.
+				if col.VirtualExpr != nil {
+					// assert jsonColOffset should always be the same.
+					// if the filter is from virtual expression, it means it is about the mv json col.
+					mvFilterMutations = append(mvFilterMutations, f)
+					mvColOffset = z
```
qw4990, line 1297: This line contains an implicit assumption: there should be only 1 virtual column in a MVIndex? Can we add this into the comment?

pkg/planner/core/indexmerge_path.go
```diff
@@ -478,6 +595,159 @@ func (ds *DataSource) generateIndexMergeAndPaths(normalPathCnt int) *util.Access
...
+			if !needSelection && bestNeedSelection {
+				// collect one path's need selection flag.
+				needSelection = bestNeedSelection
+			}
+		}
+	}
+	return mvAndPartialPaths, usedMap, needSelection, nil
+}
+
+func (ds *DataSource) generateIndexMergePartialPaths4And(normalPathCnt int,
```
qw4990, line 698: Please add some comments for this function, especially a concrete example like "the input is ..., the output is ..."

pkg/planner/core/indexmerge_path.go
```diff
@@ -478,6 +595,159 @@ func (ds *DataSource) generateIndexMergeAndPaths(normalPathCnt int) *util.Access
...
+		if len(accessFilters) == 0 { // cannot use any filter on this MVIndex
+			continue
+		}
+		// in traveling of these mv index conditions, we can only use one of them to build index merge path, just fetch it out first.
+		// build index merge partial path for every mutation combination access filters.
+		for _, mvFilterMu := range mvFilterMutations {
+			// derive each mutation access filters
+			accessFilters[mvColOffset] = mvFilterMu
+
+			partialPaths, isIntersection, ok, err := ds.buildPartialPaths4MVIndex(accessFilters, idxCols, possibleMVIndexPaths[idx].Index)
```
qw4990, line 726: Should we copy the whole `accessFilters` here? I'm not sure whether it will be overwritten if we don't copy it.

pkg/planner/core/indexmerge_path.go
```diff
@@ -391,19 +479,48 @@ func (ds *DataSource) generateIndexMergeAndPaths(normalPathCnt int) *util.Access
...
 		}
+		if composedWithMvIndex {
+			// case:
+			// idx1: mv(c, cast(`a` as signed array))
+			// idx2: idx(c)
+			// idx3: idx(c, d)
+			// for condition: (1 member of a) AND (c = 1) AND (d = 2), we should pick idx1 and idx3,
+			// since idx2's access cond has already been covered by idx1.
+			containRelation := true
+			for _, access := range originalPath.AccessConds {
```
qw4990, line 501: Should we order all paths first? Otherwise if the input order is `idx2`, `idx1` `idx3`, the filter logic seems not working.

qw4990: But maybe it's a little complex to order them...

pkg/planner/core/indexmerge_path.go
```diff
@@ -478,6 +595,159 @@ func (ds *DataSource) generateIndexMergeAndPaths(normalPathCnt int) *util.Access
...
+select * from t where ((1 member of (a) and b=1) or (2 member of (a) and b=2)) and (c > 10)
+
+	IndexMerge(OR)
+		IndexRangeScan(a, b, [1 1, 1 1])
+		IndexRangeScan(a, b, [2 2, 2 2])
+		Selection(c > 10)
+			TableRowIdScan(t)
+
+Two limitations now:
+1). Not support the embedded index merge case, which (DNF_Item1 OR DNF_Item2), for every DNF item,
```
qw4990, line 608: Where is the limitation 2 :D

pkg/planner/core/indexmerge_path.go
```diff
@@ -478,6 +595,159 @@ func (ds *DataSource) generateIndexMergeAndPaths(normalPathCnt int) *util.Access
...
+				maxCountAfterAccess = math.Max(maxCountAfterAccess, p.CountAfterAccess)
+			}
+			// Note that: here every path is about mv index path.
+			// find the most valuable mv index path, which means it has the minimum countAfterAccess.
+			if len(bestPaths) == 0 {
+				bestPaths = paths
+				bestCountAfterAccess = maxCountAfterAccess
+				bestNeedSelection = len(remainingFilters) != 0
+			} else {
+				if bestCountAfterAccess > maxCountAfterAccess {
```
qw4990, line 678: Can just use `else if bestCountAfterAccess > maxCountAfterAccess` here?

pkg/planner/core/casetest/index/index_test.go
```diff
@@ -56,6 +56,57 @@ func TestIndexJoinUniqueCompositeIndex(t *testing.T) {
 	}
 }
 
+func TestIndexMergeFromComposedDNFCondition(t *testing.T) {
```
qw4990, line 59: Better to add one more test to check the result is correct not just the plan, for example "TestMVIndexRandom".

pkg/planner/core/indexmerge_path.go
```diff
@@ -596,7 +867,144 @@ func (ds *DataSource) generateIndexMergeOnDNF4MVIndex(normalPathCnt int, filters
...
+		mvIndexPartialPaths, usedAccessMap, needSelection4MVIndex, err := ds.generateIndexMergePartialPath4Or(normalPathCnt, dnfFilters)
+		if err != nil {
+			return err
+		}
+		remainedDNFItems := make([]expression.Expression, 0, len(dnfFilters))
+		for i, b := range usedAccessMap {
+			if !b {
+				remainedDNFItems = append(remainedDNFItems, dnfFilters[i])
+			}
+		}
```
qw4990, line 934: Just suggest that put all filters in the table-side Selection if `needSelection4MVIndex == True` for safety.

pkg/planner/core/indexmerge_path.go
AilinKid: yes, it a different portal, when expressions {x=1 and x=2} being filled in `fillIndexPath`, we can derive it's index range is invalid.

for mv index （json predicate）here, fillIndexPath can not work, we should extract filters and map them with the mv index columns ourselves. In this process, json predicate is converted as EQ condition combination, finally deriving the partial linear index range according `detachCondAndBuildRangeForPath`. But in the outer side when it's still a json predicate, converging several json predicate linked with AND into a linear range is impossible。

AilinKid: 这个函数主要服务于这个 CNF
（A_json=1）AND（A_json=2） AND（A_json=2）AND（C=1）
我们想走 index merge, 不能独立每个 cnf item 来看，因为有些 index 可能就是复合索引，所以是尽量从一堆中去找当前index 自己想要的。对于 mv index1 （A_json, c）来说，现在有这么一个 path 了，可以从一堆中可以找到，（A_json=1）和 （c=1）还有一个选择就是（A_json=2）和 （c=1）; 对于 mv index1 （B_json, c）来说，可以从一堆中找到 （B_json=1）和 （c=1）;

AilinKid: 研究了一下，目前 （c=1）和 （c=2） 这种情况，在 table path 就会被启发式规则（也是通过 detach ranger）算出来这是 empty range，然后直接走 table path 进而 dual 计划了，无法走到上述 2 的假如情况，但是 ranger 可能探测不出来 json predicate 相悖的情况。所以我们不用对一个常规的 index column 收集多个 access 条件，而仅对 json col 收集下多个 mutation 的条件来组合。

pkg/planner/core/indexmerge_path.go
AilinKid: I changed the if condition to: `if col.VirtualExpr != nil && col.VirtualExpr.GetType().IsArray() `

pkg/planner/core/indexmerge_path.go
AilinKid: you are right, make senseeeee.

pkg/planner/core/indexmerge_path.go
AilinKid: you mean when c is not used, which index below should be used to build normal index partial path？yes, it a question, for now it's random, first coming first usage, map it as used. so idx3 is ignored, unless there is a new condition d=1 remained.

idx2: idx(c)
idx3: idx(c, d)

pkg/planner/core/indexmerge_path.go
AilinKid: yes, changed, for now, usedAccessMap is only used to identify whether one dnf item is completely not used, if so, composed-index-merge should be stopped soon.

pkg/planner/core/casetest/index/index_test.go
AilinKid: sure, it will be in my next pull request

pkg/planner/core/indexmerge_path.go
```diff
@@ -596,7 +930,191 @@ func (ds *DataSource) generateIndexMergeOnDNF4MVIndex(normalPathCnt int, filters
...
+		// todo:  make estimation more correct rather than pruning other index merge path.
+		var indexMergeTableFilters []expression.Expression
+		if needSelection4MVIndex || needSelection4NormalIndex {
+			indexMergeTableFilters = indexMergeConds
+		}
+		mvp := ds.buildPartialPathUp4MVIndex(combinedPartialPaths, false, indexMergeTableFilters)
+		ds.possibleAccessPaths = append(ds.possibleAccessPaths, mvp)
+		return nil
+	}
+	// CNF path.
```
time-and-fate, line 1061: What if we just extend the previous `generateIndexMergeAndPaths()` to treat the intersection type index merge path as a candidate partial path?
I think the new logic has exactly the same rationale as the previous one, but we are re-implementing many things.

time-and-fate (commented):
Though I haven't finished reviewing it yet, this PR gives me a feeling that it's much more complicated than it needs to be.

pkg/planner/core/indexmerge_path.go
```diff
@@ -596,7 +930,191 @@ func (ds *DataSource) generateIndexMergeOnDNF4MVIndex(normalPathCnt int, filters
...
+		// 1: some mv index partial path, no normal index path, it means multi mv index merge.
+		// 2: some mv index partial path, some normal index path, it means hybrid index merge.
+		// 3: no mv index partial path, several normal index path, it means multi normal index merge.
+		var combinedPartialPaths []*util.AccessPath
+		if len(normalIndexPartialPaths) != 0 {
+			combinedPartialPaths = normalIndexPartialPaths
+		}
+		if len(mvIndexPartialPaths) != 0 {
+			combinedPartialPaths = append(combinedPartialPaths, mvIndexPartialPaths...)
+		}
```
time-and-fate, line 1047:
```suggestion
		combinedPartialPaths = append(combinedPartialPaths, normalIndexPartialPaths...)
		combinedPartialPaths = append(combinedPartialPaths, mvIndexPartialPaths...)
```

pkg/planner/core/indexmerge_path.go
```diff
@@ -596,7 +930,191 @@ func (ds *DataSource) generateIndexMergeOnDNF4MVIndex(normalPathCnt int, filters
...
+		flatten as: select * from t where 1 member of (a) and json_contains(b, '[1, 2, 3]') and c=1 and c=2 and other index predicates
+		analyze: find and utilize index access filter items as much as possible:
+		IndexMerge(AND-INTERSECTION)                  --- ROOT
+			IndexRangeScan(mv-index-a)(1)             --- COP
+			IndexMerge(mv-index-b AND-INTERSECTION)   --- ROOT (embedded index merge) ---> simplify: we can defer the TableRowIdScan(t) to the outer index merge.
+				IndexRangeScan(a, [1,1])              --- COP
+				IndexRangeScan(a, [2,2])              --- COP
+				IndexRangeScan(a, [3,3])              --- COP
+			IndexRangeScan(non-mv-index-if-any)(?)    --- COP
+			TableRowIdScan(t)                         --- COP
```
time-and-fate, line 956: Got a bit confused by those nested IndexMerge and "simplify" descriptions. For me, I think they make the comments harder to understand.

pkg/planner/core/indexmerge_path.go
```diff
@@ -94,7 +98,12 @@ func (ds *DataSource) generateIndexMergePath() error {
 	}
 
 	// If len(indexMergeHints) > 0 and some index-merge paths were added, then prune all other non-index-merge paths.
-	ds.possibleAccessPaths = ds.possibleAccessPaths[regularPathCount:]
+	// if len(ds.possibleAccessPaths) > oldIndexMergeCount, it means composed index merge path is generated, prune others.
```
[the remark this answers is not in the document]
AilinKid: borrow from the old logic here, since composed cnf/dnf path is hard to be choosed as we said in the lark group
```
@Ryan Zhang 对，在没有 prune 之前，我想用这个 cost model 自己来判来着，结果发现 index plan 上累加 cost 的速度 >> table plan 上来减少的 row count 带来的代价减小
```

pkg/planner/core/indexmerge_path.go
AilinKid: image
From the pic, for normal index path in CNF case, I have tried to reuse this function, but for MV index conds, there some limitations, and that's why I specially split it out.

the way you say is the first version I have been tried, take this as a simple example
```
(1 member of a) and (2 member of a)
```
the old index merge case will only gen ONE candidate path for ONE MV index path for this, picking the first encountered/satisfied conds as it access conds, so for this case, there is exactly one index candidate partial path under extending generateIndexMergeAndPaths function as you said.

AilinKid: image
there is a additional todo already added

pkg/planner/core/indexmerge_path.go
```diff
@@ -596,7 +930,183 @@ func (ds *DataSource) generateIndexMergeOnDNF4MVIndex(normalPathCnt int, filters
...
+		composed := (len(mvIndexPartialPaths) > 1) || (len(mvIndexPartialPaths) == 1 && len(normalIndexPartialPaths) >= 1)
+		if !composed {
+			return nil
+		}
+		// if any cnf item is not used as index partial path, index merge is not valid anymore.
+		for _, one := range usedAccessMap {
+			if !one {
+				return nil
+			}
+		}
```
time-and-fate, line 1035: `slices.Contains` can make it simpler.

pkg/planner/core/indexmerge_path.go
```diff
@@ -391,19 +510,54 @@ func (ds *DataSource) generateIndexMergeAndPaths(normalPathCnt int) *util.Access
...
-	if len(partialPaths) < 2 {
+	if len(partialPaths) < 1 {
 		return nil
 	}
+	if len(partialPaths) == 1 {
+		// even single normal index path here, it can be composed with other mv index partial paths.
+		if !composedWithMvIndex {
+			return nil
+		}
+	}
```
time-and-fate, line 560: I think they can be in just one `if` block. 😂

pkg/planner/core/indexmerge_path.go
```diff
@@ -478,6 +632,185 @@ func (ds *DataSource) generateIndexMergeAndPaths(normalPathCnt int) *util.Access
...
+
+	for offset, dnfCond := range indexMergeDNFConds {
+		var cnfConds []expression.Expression
+		if sf, ok := dnfCond.(*expression.ScalarFunction); !ok {
+			continue
+		} else if sf.FuncName.L == ast.LogicAnd {
+			cnfConds = expression.FlattenCNFConditions(sf)
+		} else {
+			cnfConds = []expression.Expression{sf}
+		}
```
time-and-fate, line 674:
```suggestion
		if sf, ok := dnfCond.(*expression.ScalarFunction); !ok {
			continue
		} else {
			cnfConds = expression.FlattenCNFConditions(sf)
		}
```

pkg/planner/core/indexmerge_path.go
```diff
@@ -107,6 +116,70 @@ func (ds *DataSource) generateIndexMergePath() error {
...
+			// the entire index merge is not valid anymore.
+			return nil, false, usedMap, nil
+		}
+		// prune out global indexes.
+		for i := len(itemPaths) - 1; i >= 0; i-- {
+			path := itemPaths[i]
+			if path.Index != nil && path.Index.Global {
+				itemPaths = append(itemPaths[:i], itemPaths[i+1:]...)
+			}
+		}
```
time-and-fate, line 149: Now we can use `slices.DeleteFunc` to make it simpler.

pkg/planner/core/indexmerge_path.go
```diff
@@ -376,12 +449,58 @@ func (ds *DataSource) buildIndexMergeOrPath(
...
+		return nil, usedAccessMap, false, err
+	}
+	// If all the partialPaths use the same index, we will not use the indexMerge.
+	singlePath := true
+	for i := len(paths) - 1; i >= 1; i-- {
+		if paths[i].Index != paths[i-1].Index {
+			singlePath = false
+			break
+		}
+	}
```
time-and-fate, line 475: I think most of the logic here (including the logic of `generateNormalIndexPartialPaths4DNF`) is the same as `generateIndexMergeOrPaths`, can we try to reuse them?

pkg/planner/core/indexmerge_path.go
```diff
@@ -596,7 +930,183 @@ func (ds *DataSource) generateIndexMergeOnDNF4MVIndex(normalPathCnt int, filters
...
+			}
+		}
+		// todo: make this code as a portal of all index merge path.
+		// if we derive:
+		// 1: some mv index partial path, no normal index path, it means multi mv index merge.
+		// 2: some mv index partial path, some normal index path, it means hybrid index merge.
+		// 3: no mv index partial path, several normal index path, it means multi normal index merge.
+		var combinedPartialPaths []*util.AccessPath
+		combinedPartialPaths = normalIndexPartialPaths
+		combinedPartialPaths = append(combinedPartialPaths, mvIndexPartialPaths...)
```
time-and-fate, line 1043: If you want to reuse the slice, this can be more simplified.
```suggestion
		combinedPartialPaths := append(normalIndexPartialPaths, mvIndexPartialPaths...)
```

pkg/planner/core/indexmerge_path.go
time-and-fate:
```
(1 member of a) and (2 member of a)
```
For this case, I think that means we should enhance the old mv index implementation.
Besides, I think this leads to another question: Are we relying on the new logic in this PR to handle this case?
If yes, what's the scope of this PR? I thought this PR is for generating index merge paths that contain different indexes (and including mv index).

pkg/planner/core/indexmerge_path.go
qw4990: got it

qw4990 (approved):

pkg/planner/core/indexmerge_path.go
AilinKid: not exactly here, if a function is not ast.AND, let's say `plus(a,b)`, I want to keep it real and return as `[]expression.Expression{plus}`

while after the trails and errors, I find it will returns `[]expression.Expression{a,b}` to me, the below is the core code inside them

image

pkg/planner/core/indexmerge_path.go
AilinKid: cool

pkg/planner/core/indexmerge_path.go
time-and-fate: You are right, my mistake. It should be `expression.SplitCNFItems()`.

pkg/planner/core/indexmerge_path.go
AilinKid: basically and partially can be reused, I can make it in next pr, otherwise, review will be a burden

pkg/planner/core/indexmerge_path.go
AilinKid: The answer is yes, it targeting for composition of multi different or same mv/normal index partial paths. and `(1 member of a) and (2 member of a)` is not mentioned in the user case, but it's done by the way. 😂

pkg/planner/core/indexmerge_path.go
```diff
@@ -596,7 +927,176 @@ func (ds *DataSource) generateIndexMergeOnDNF4MVIndex(normalPathCnt int, filters
...
+
+	if len(indexMergeConds) == 1 {
+		// DNF path.
+		sf, ok := indexMergeConds[0].(*expression.ScalarFunction)
+		if !ok || sf.FuncName.L != ast.LogicOr {
+			// targeting: cond1 or cond2 or cond3
+			return nil
+		}
+		dnfFilters := expression.FlattenDNFConditions(sf)
+		mvIndexPartialPaths, usedAccessMap, needSelection4MVIndex, err := ds.generateIndexMergePartialPath4Or(normalPathCnt, dnfFilters)
```
time-and-fate, line 1010: Why does this function have a different naming style? Shouldn't it be `generateMVIndexMergePartialPath4Or`?

pkg/planner/core/indexmerge_path.go
```diff
@@ -478,6 +631,183 @@ func (ds *DataSource) generateIndexMergeAndPaths(normalPathCnt int) *util.Access
...
+//			index built on column ---> idx_c(c), since condition c is already used in previous mv index partial path,
+//			index idx_c is ignored. If there is a additional cnf item as d=1 and there is a idx_d(d) on it. Since
+//	     d=1 is not used in previous mv index partial path, we can build normal index partial path idx_d out from
+//			d=1.
+//
+// step3: after all mv index partial paths and normal index partial paths is all built. Build them up as a whole
+//
+//	index merge path together.
+//
+// this function mainly focus on step1.
```
time-and-fate, line 755: Then why do we put irrelevant comments here? I think comments should aim to make the code logic clear and should be easy to understand.
If some background information is needed, we can put that in outer functions like `generateIndexMerge4ComposedIndex`, and then reference those comments here.

time-and-fate (approved):

pkg/planner/core/indexmerge_path.go
```diff
@@ -391,17 +511,50 @@ func (ds *DataSource) generateIndexMergeAndPaths(normalPathCnt int) *util.Access
...
 			continue
 		}
+		if composedWithMvIndex {
+			// case:
+			// idx1: mv(c, cast(`a` as signed array))
+			// idx2: idx(c)
+			// idx3: idx(c, d)
+			// for condition: (1 member of a) AND (c = 1) AND (d = 2), we should pick idx1 and idx3,
+			// since idx2's access cond has already been covered by idx1.
+			containRelation := true
```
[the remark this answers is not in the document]
AilinKid: in Chinese means: "包含关系"
if a normal index's condition are all contained in previous mv's access condition (a kind of 包含), then this normal index is redundant

pkg/planner/core/indexmerge_path.go
AilinKid: let me see, Em, you are right, changed

Issue #47424, opened 2023-10-07
execution: semi join takes too long to run


qw4990:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

<!-- a step by step guide for reproducing the bug. -->

From https://asktug.com/t/topic/1014053/4

Prepare TPCH dataset with `sf=1`:
```
go-tpc tpch --sf=1 prepare
```

Then the 2-way semi-join with `150000` rows takes 13s to run.
```
mysql> EXPLAIN ANALYZE select count(*) from customer a where exists (select 1 from customer b where a.C_NATIONKEY=b.C_NATIONKEY);
+-------------------------------+-----------+---------+-----------+---------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+-----------------------------------------------------------------------------+----------+---------+
| id                            | estRows   | actRows | task      | access object | execution info                                                                                                                                                                                                                                                                                                                                                                                     | operator info                                                               | memory   | disk    |
+-------------------------------+-----------+---------+-----------+---------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+-----------------------------------------------------------------------------+----------+---------+
| HashAgg_10                    | 1.00      | 1       | root      |               | time:13.1s, loops:2, RU:34.346561, partial_worker:{wall_time:13.1241415s, concurrency:5, task_num:148, tot_wait:1m5.616187036s, tot_exec:4.228498ms, tot_time:1m5.620451542s, max:13.124107333s, p95:13.124107333s}, final_worker:{wall_time:13.124157041s, concurrency:5, task_num:5, tot_wait:1m5.620598336s, tot_exec:19.292µs, tot_time:1m5.620619751s, max:13.124129334s, p95:13.124129334s}  | funcs:count(1)->Column#18                                                   | 165.3 KB | N/A     |
[...]

PR #47764, opened 2023-10-18
executor: optimize (left outer) (anti) semi join which has no other condition

gengliqi:

Issue Number: close https://github.com/pingcap/tidb/issues/47424

Problem Summary:
The sql will be very slow if there are a lot of matched row for each row from build side in (left) (anti) semi join.

Recognize this case then optimize it.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

Before
```
mysql> select count(*) from customer a where exists (select 1 from customer b where a.C_NATIONKEY=b.C_NATIONKEY);
+----------+
| count(*) |
+----------+
|   150000 |
+----------+
1 row in set (36.32 sec)
```
After
```
mysql> select count(*) from customer a where exists (select 1 from customer b where a.C_NATIONKEY=b.C_NATIONKEY);
+----------+
| count(*) |
+----------+
|   150000 |
+----------+
1 row in set (0.16 sec)
```

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
optimize (left outer) (anti) semi join which has no other condition
```

### DIFF
--- a/pkg/executor/hash_table.go
+++ b/pkg/executor/hash_table.go
@@ -611,6 +611,11 @@ func (es *entryStore) GetStore() (e *entry, memDelta int64) {
 
 type baseHashTable interface {
 	Put(hashKey uint64, rowPtr chunk.RowPtr)
+	// e := Get(hashKey)
+	// for ; e != nil; e = e.next {
+	//    rowPtr := e.ptr
+	//    ...
+	// }
 	Get(hashKey uint64) *entry
 	Len() uint64
 	// GetAndCleanMemoryDelta gets and cleans the memDelta of the baseHashTable. Memory delta will be cleared after each fetch.
--- a/pkg/executor/hash_table.go
+++ b/pkg/executor/hash_table.go
@@ -162,7 +162,7 @@ func (c *hashRowContainer) GetOneMatchedRow(probeKey uint64, probeRow chunk.Row,
 	var matchedRow chunk.Row
 
 	c.chkBuf = nil
-	cap := 0
+	capacity := 0
 
 	for i := 0; innerEntry != nil; i, innerEntry = i+1, innerEntry.next {
 		ptr := innerEntry.ptr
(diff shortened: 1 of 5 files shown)


pkg/executor/hash_table.go
```diff
@@ -152,6 +152,42 @@ func (c *hashRowContainer) GetMatchedRows(probeKey uint64, probeRow chunk.Row, h
...
+// GetOneMatchedRow get one matched rows from probeRow.
+func (c *hashRowContainer) GetOneMatchedRow(probeKey uint64, probeRow chunk.Row, hCtx *hashContext) (*chunk.Row, error) {
+	var err error
+	innerEntry := c.hashTable.Get(probeKey)
+	if innerEntry == nil {
+		return nil, err
+	}
+	var matchedRow chunk.Row
+
+	c.chkBuf = nil
```
windtalker, line 164: Maybe don't need to reset it to nil for every call of `GetOneMatchedRow`?

pkg/executor/joiner.go
```diff
@@ -490,6 +499,10 @@ func (naaj *nullAwareAntiSemiJoiner) onMissMatch(_ bool, outer chunk.Row, chk *c
 	chk.AppendRowByColIdxs(outer, naaj.lUsed)
 }
 
+func (naaj *nullAwareAntiSemiJoiner) isSemiJoinWithoutCondition() bool {
+	return len(naaj.conditions) == 0
```
windtalker, line 503: Is this optimazation works for nullaware join? And since only `joinMatchedProbeSideRow2Chunk` is check `isSemiJoinWithoutCondition`, nullaware join is not optimized even if it return `true` here.

gengliqi: No. This function is useless for null-aware semi join.

pkg/executor/hash_table.go
gengliqi: Indeed.

gengliqi: Indeed.

windtalker (approved):
LGTM

pkg/executor/hash_table.go
```diff
@@ -565,7 +611,7 @@ func (es *entryStore) GetStore() (e *entry, memDelta int64) {
 
 type baseHashTable interface {
 	Put(hashKey uint64, rowPtr chunk.RowPtr)
-	Get(hashKey uint64) (rowPtrs []chunk.RowPtr)
+	Get(hashKey uint64) *entry
```
XuHuaiyu, line 614: Will it be better to add a comment example here for how to iterator the stored ptrs?

gengliqi: Added. Nice suggestion.

XuHuaiyu (approved):

Issue #47572, opened 2023-10-12
import with global sort is slow, both on encode&sort and write&ingest step


D3Hunter:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
we can see the encode speed is about 90m/s, but without global sort, it's around 210m/s.
write tikv speed is slow too. the total import time is 495m(global sort) vs 23m(local sort)
![64035cef-82c0-4d87-a56b-fb6820a903a1](https://github.com/pingcap/tidb/assets/3312245/972a6433-22a6-4ca3-b4de-9a33521ac1f7)

### 4. What is your TiDB version? (Required)
master/7.4

D3Hunter:
in version 7.5, we release global sort as a experionmental feature

PR #47793, opened 2023-10-19
external: pre-alloc mem for writer, use offset to avoid golang slice

D3Hunter:

Issue Number: ref #47572

Problem Summary:

pre-alloc memory for writer, we align to block-size, default is 16MiB, for import-into data kv we use 32MiB
store each KV inside buf directly and we use a custom location to pointer to it, to avoid the overhead of golang slice, reduce from 3*24(golang slice size) to 16.
this way we can have less smaller object, and can use memory more efficiently, less GC

by using this way, we put limition on KV size, the max kv size is 16MiB for index, and 32MiB for data.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
in below test, we set those 2 params to avoid tidb mem tunner causes too many GC as it's known global sort will uses  a lot of mem
```sql
mysql> show variables like 'tidb_server_memory_limit_gc_trigger';
+-------------------------------------+-------+
| Variable_name                       | Value |
+-------------------------------------+-------+
| tidb_server_memory_limit_gc_trigger | 0.99  |
+-------------------------------------+-------+
1 row in set (0.04 sec)

mysql> show variables like 'tidb_server_memory_limit';
+--------------------------+-------+
| Variable_name            | Value |
+--------------------------+-------+
| tidb_server_memory_limit | 88%   |
+--------------------------+-------+
```

after this pr, on a case of 158.8GB, import takes 46m(after) vs 52m(before). the time takes on `encode&sort` is 12m2s(after) vs 18m45(before). after this pr, the time takes on `encode&sort` step is close to `local sort`, and the encode speed is more stable

before:
image

after
image

- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


wjhuang2016 (approved):

br/pkg/lightning/backend/external/file.go
```diff
@@ -46,51 +49,26 @@ func NewKeyValueStore(
...
-	n, err = s.dataWriter.Write(
-		s.ctx,
-		binary.BigEndian.AppendUint64(b[:0], uint64(len(value))),
-	)
-	if err != nil {
-		return err
-	}
-	kvLen += n
-	n, err = s.dataWriter.Write(s.ctx, value)
+func (s *KeyValueStore) AddData(val []byte) error {
```
lance6716, line 57: Just noticed no other package is using this function so we can make it private. Another reason is now the function need caller to build data in specific layout, which is not friendly as a public function.

br/pkg/lightning/backend/external/kv_buf.go
```diff
@@ -0,0 +1,72 @@
...
+// allocation, all space in this buffer will be reused when reset.
+type preAllocKVBuf struct {
+	blocks      [][]byte
+	blockSize   int
+	curBlock    []byte
+	curBlockIdx int
+	curIdx      int
+}
+
+func newPreAllocKVBuf(memLimit uint64, blockSize int) *preAllocKVBuf {
```
lance6716, line 32: it looks similar to membuf.Buffer 🤔 not sure we can reuse it

br/pkg/lightning/backend/external/writer.go
```diff
@@ -304,22 +319,36 @@ type Writer struct {
...
-	if w.batchSize >= w.memSizeLimit {
+	encodedKeyLen := keyAdapter.EncodedLen(idxKey, rowID)
+	length := encodedKeyLen + len(idxVal) + lengthBytes*2
+	blockIdx, dataBuf, off, allocated := w.kvBuffer.Alloc(length)
+	if !allocated {
 		if err := w.flushKVs(ctx, false); err != nil {
 			return err
 		}
+		blockIdx, dataBuf, off, allocated = w.kvBuffer.Alloc(length)
+		// we now don't support KV larger than blockSize
```
lance6716, line 335: seems we will panic? And why do we need to split `memLimit` into multiple blocks.

D3Hunter: why panic? the mem-limit might be several GB on add-index, it's too large, and hasn't test, but the performance should be similar.

lance6716: I changed an unit test of memLimit = 16, blockSize = 8, Alloc(9). Alloc found the current block is not enough to hold the data and moved to next block, but does not check the pre-allocated blocks have enough capacity

br/pkg/lightning/backend/external/kv_buf.go
D3Hunter: fields are similar, but not the code, will leave it for now

br/pkg/lightning/backend/external/file.go
D3Hunter: https://github.com/pingcap/tidb/pull/47793/commits/c8e33104edb36bdca262e713451c77a20dca0953

br/pkg/lightning/backend/external/writer.go
D3Hunter: https://github.com/pingcap/tidb/pull/47793/commits/0d842cb3a6d1d2e573e0bfc7e4d9eee32d0fa50c

pkg/disttask/importinto/encode_and_sort_operator.go
```diff
@@ -250,3 +257,10 @@ func getNumOfIndexGenKV(tblInfo *model.TableInfo) int {
 	}
 	return count
 }
+
+func getKVGroupBlockSize(group string) int {
```
lance6716, line 261: I still worried about the block size is not enough. In lightning's CSV parser, we can tolerate ~120M KV size (however the only case user get such a big KV pair is that line terminator is not correct / source file is corrupted so the multiple lines are parsed as one line)

https://github.com/pingcap/tidb/blob/b0c01fa9a44e0abbbed9abca306aefd29e16bda2/br/pkg/lightning/mydump/csv_parser.go#L44

And if we don't split `memLimit` into multiple blocks, we will not meet this problem. I don't have strong oponion about this suggestion

lance6716 (commented):
rest lgtm

pkg/disttask/importinto/encode_and_sort_operator.go
D3Hunter: discussed with @Benjamin2037 last week, plan to add a restriction in doc about kv size when import using global sort

> a big KV pair is that line terminator is not correct / source file is corrupted

for this case, if multiple line is concated and parsed successfully, we can identified it easily

PR #47794, opened 2023-10-19
types, stmtctx, *: use ZeroDate, ZeroInDate and InvalidDate related flags to replace variable in statement context

### LINKED ISSUE #47507: use ZeroDate related flags to handle some in functions  in `types` package

### LINKED ISSUE #47508: use ZeroInDate and InvalidDate related flags to handle cases in functions  in `types` package

YangKeao:

Issue Number: close #47508, close #47507

Problem Summary:

This PR replaces the field `NoZeroDate`, `IgnoreZeroInDate` and `AllowInvalidDate` to the flag in types context.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  >   

They are all covered by existing tests


pkg/types/time.go
```diff
@@ -2041,11 +2034,11 @@ func ParseTimeFromYear(_ *stmtctx.StatementContext, year int64) (Time, error) {
 
 // ParseTimeFromNum parses a formatted int64,
 // returns the value which type is tp.
-func ParseTimeFromNum(sc *stmtctx.StatementContext, num int64, tp byte, fsp int) (Time, error) {
+func ParseTimeFromNum(ctx Context, num int64, tp byte, fsp int) (Time, error) {
 	// MySQL compatibility: 0 should not be converted to null, see #11203
 	if num == 0 {
 		zt := NewTime(ZeroCoreTime, tp, DefaultFsp)
-		if sc != nil && sc.InCreateOrAlterStmt && !sc.TypeFlags().TruncateAsWarning() && sc.NoZeroDate {
+		if !ctx.Flags().TruncateAsWarning() && !ctx.Flags().IgnoreZeroDateErr() {
```
YangKeao, line 2041: All modifications except this line of this PR are all simple.

This PR removes the `NoZeroDate` flag and replace it with `IgnoreZeroDateErr`, which roughly means `!NoZeroDate`. And `IgnoreZeroDateErr` is set as the default statement type context flag.

What about `InCreateOrAlterStmt`? As this line is the only logic which will read the `NoZeroDate` (or `IgnoreZeroDateErr`) flag, I turn to only set `IgnoreZeroDateErr` when it's about to set `InCreateOrAlterStmt`. See https://github.com/pingcap/tidb/pull/47794/files#diff-cedde52f04eef8a5117a506a088e41a878c00e147d529fcda309d834c966a010R222 and https://github.com/pingcap/tidb/pull/47794/files#diff-b41713213596a17bc034a9cc9d33e2f26ee345da66c23ca56154be85c5b44c44R2112 (in `pkg/ddl/backfilling_scheduler.go` and `pkg/executor/executor.go`).

lcwangchao: Or could we also remove out `!ctx.Flags().TruncateAsWarning()` here because we only set `IgnoreZeroDateErr` to true when `reorgInfo.ReorgMeta.SQLMode.HasStrictMode()` is false, no need to check `TruncateAsWarning` again.

BTW, according to the document of MySQL: https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html

xhebox (approved):

pkg/types/time.go
YangKeao: > Or could we also remove out !ctx.Flags().TruncateAsWarning() here because we only set IgnoreZeroDateErr to true when reorgInfo.ReorgMeta.SQLMode.HasStrictMode() is false.

Actually `IgnoreZeroDateErr` is by default. For most cases, it will be `true`. (Just like for most cases, the `NoZeroMode` is not set and will be false).

YangKeao: See https://github.com/pingcap/tidb/pull/47794/files#diff-0a77759e2e60ac7cd97742f830a29b9691aea4aa2fa883d2087f0d41191d0915R259

lcwangchao (approved):

3pointer (approved):

pkg/types/time.go
YangKeao: Good! I understand. I could rewrite `IgnoreZeroDateErr = !NoZeroDate` to `IgnoreZeroDateErr = !NoZeroDate || !HasStrictMode`. Nice suggestion. Thanks 👍 .

D3Hunter (approved):
lightning/domain part lgtm

zanmato1984 (approved):
LGTM

Issue #35948, opened 2022-07-05
[stats] auto analyze failed


zhuzhenwen-github:
## Bug Report

[analyze.go:1435] ["analyze worker panicked"] [recover="runtime error: slice bounds out of range [-1:]"] [stack="github.com/pingcap/tidb/executor.(*AnalyzeColumnsExec).subBuildWorker.func1\n\t/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/executor/analyze.go:1435\nruntime.gopanic\n\t/usr/local/go/src/runtime/panic.go:838\nruntime.goPanicSliceB\n\t/usr/local/go/src/runtime/panic.go:117\ngithub.com/pingcap/tidb/statistics.BuildHistAndTopN\n\t/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/statistics/builder.go:346\ngithub.com/pingcap/tidb/executor.(*AnalyzeColumnsExec).subBuildWorker\n\t/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/executor/analyze.go:1559\ngithub.com/pingcap/tidb/executor.(*AnalyzeColumnsExec).buildSamplingStats.func3\n\t/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/executor/analyze.go:1136\ngithub.com/pingcap/tidb/executor.(*notifyErrorWaitGroupWrapper).Run.func1\n\t/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/executor/analyze.go:2653"]

### 1. Minimal reproduce step (Required)
Execute commands manually "analyze table  ${tableName}"

### 2. What did you expect to see? (Required)
auto analyze success

### 3. What did you see instead (Required)
SHOW ANALYZE STATUS;
State:failed
Fail_reason:runtime error: slice bounds out of range [-1:]

### 4. What is your TiDB version? (Required)
Release Version: v6.1.0
Edition: Community
Git Commit Hash: 1a89decdb192cbdce6a7b0020d71128bc964d30f
Git Branch: heads/refs/tags/v6.1.0
UTC Build Time: 2022-06-05 05:15:11
GoVersion: go1.18.2
Race Enabled: false
TiKV Min Version: v3.0.0-60965b006877ca7234adaced7890d7b029ed1306
Check Table Before Drop: false

chrysan:
[...]

PR #47928, opened 2023-10-23
statistic: fix panic when building topN

hi-rustin:
Signed-off-by: hi-rustin <rustin.liu@gmail.com>

Issue Number: close https://github.com/pingcap/tidb/issues/35948

Problem Summary:

The issue arises when a value appears twice in the topN list, triggering a panic.  The solution involves breaking out of the loop as soon as we find the duplicate value. This bug seems to occur only when the column is of JSON type. Since we haven't been able to reproduce the issue, we've added a warning log to print all topN values, helping us identify the duplicated value.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] We will test it on users env.
- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fixed a bug that caused a panic while generating the topN statistics
```

### DIFF
--- a/pkg/statistics/builder.go
+++ b/pkg/statistics/builder.go
@@ -382,12 +382,9 @@ func BuildHistAndTopN(
 		)
 		for j := 0; j < len(topNList); j++ {
 			if bytes.Equal(sampleBytes, topNList[j].Encoded) {
-				// First time to find the same value in topN: need to record the sample data for debugging.
-				if !foundTwice {
-					firstTimeSample = samples[i].Value
-				} else {
-					// This should never happen, but we met this panic before, so we add this check here.
-					// See: https://github.com/pingcap/tidb/issues/35948
+				// This should never happen, but we met this panic before, so we add this check here.
+				// See: https://github.com/pingcap/tidb/issues/35948
+				if foundTwice {
 					datumString, err := samples[i].Value.ToString()
 					if err != nil {
 						logutil.BgLogger().With(
@@ -419,6 +416,8 @@ func BuildHistAndTopN(
 					// how to remove it from the samples.
 					break
 				}
+				// First time to find the same value in topN: need to record the sample data for debugging.
+				firstTimeSample = samples[i].Value
 				// Found the same value in topn: need to skip over this value in samples.
 				copy(samples[i:], samples[uint64(i)+topNList[j].Count:])
 				samples = samples[:uint64(len(samples))-topNList[j].Count]
--- a/pkg/statistics/builder.go
+++ b/pkg/statistics/builder.go
@@ -400,6 +400,7 @@ func BuildHistAndTopN(
 					break
 				} else {
 					i--
+					foundTwice = true
 					continue
 				}
 			}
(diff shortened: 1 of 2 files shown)


hi-rustin (commented):
🔢

winoros (approved):

hawkingrei (approved):

hi-rustin:
/hold

@time-and-fate Could you please take another look? Thanks!

time-and-fate (commented):
1. Here we only logged the encoded bytes. But essentially, we need to know why two `types.Datum`s are encoded into the same bytes. If we only log the encoded bytes, we'll need some extra work to know that after we obtain the log.
I think it's best to log two `types.Datum`s with the same encoded bytes, with the full `Datum` struct printed. But that might not be easy to implement, so it's just a suggestion.
2. Have we verified that the bytes can be correctly escaped and printed by `zap`? Here the encoded bytes are probably not valid UTF-8 strings, and according to `zap` doc we should use `zap.Binary()` in this case.

pkg/statistics/builder.go
```diff
@@ -373,12 +375,53 @@ func BuildHistAndTopN(
...
+							zap.String("category", "stats"),
+						).Error("try to convert datum to string failed", zap.Error(err))
+					}
+
+					logutil.BgLogger().With(
+						zap.String("category", "stats"),
+					).Warn(
+						"invalid sample data",
+						zap.Int64("i", i),
+						zap.Int64("j", int64(j)),
```
winoros, line 406: we cannot use the `i` and `j` to do debugging?

pkg/statistics/builder.go
```diff
@@ -373,12 +375,53 @@ func BuildHistAndTopN(
...
+						zap.String("datum", datumString),
+						zap.Binary("sampleBytes", sampleBytes),
+						zap.Binary("topNBytes", topNList[j].Encoded),
+						zap.String("firstTimeSample", firstTimeSampleString),
+					)
```
winoros, line 411: We missed the `hist_id`(col/index's id) and `is_index`.

pkg/statistics/builder.go
```diff
@@ -373,12 +375,53 @@ func BuildHistAndTopN(
...
+			foundTwice      bool
+			firstTimeSample types.Datum
+		)
 		for j := 0; j < len(topNList); j++ {
 			if bytes.Equal(sampleBytes, topNList[j].Encoded) {
-				// find the same value in topn: need to skip over this value in samples
+				// This should never happen, but we met this panic before, so we add this check here.
+				// See: https://github.com/pingcap/tidb/issues/35948
+				if foundTwice {
+					datumString, err := samples[i].Value.ToString()
```
time-and-fate, line 388: If we meet this bug again, the `i` is actually `-1`, and I think we will panic again here.

pkg/statistics/builder.go
hi-rustin: Added.

winoros (approved):

pkg/statistics/builder.go
hi-rustin: Fixed.

time-and-fate (approved):

hi-rustin:
Thanks for your review!  💚 💙 💜 💛 ❤️

PR #47997, opened 2023-10-26
types: always handle overflow error outside the types package

### LINKED ISSUE #47517: always return overflow error for the functions in `types` package

lcwangchao:

Issue Number: close #47517

All functions in `types` package should respect settings `types.Context`. However, for most functions, it will ignore `overflowAsWarning` settings and return this error directly. To make sure all methods follows the same behavior, we unified this behavior to always return overflow error and the overflow handle should be outside the `types` package

1. Remove `HandleOverflow` in `types.ConvertJSONToInt` and `types.ProduceDecWithSpecifiedTp`
2. Add `HandleOverflow` in `builtinCastXXXAsDecimalSig.evalDecimal` and `builtinCastJSONAsIntSig.evalInt`.
3. Remove useless `FlagIgnoreOverflowError`, `FlagZeroDateAsWarning`, `FlagZeroInDateAsWarning` and `FlagInvalidDateAsWarning` flags.  

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


YangKeao (approved):
LGTM

Do we agree that the error handling in `types` package should follow similar pattern? Which means the function in `types` package should do its best-effort, and return valid values and all errors, the caller of these functions should take the responsibility to handle these errors.

Also, should we do the same refractoring for `HandleTruncate` in the future?

The problem is that, some functions will return both `TruncateError` and `OverflowError` (in different cases), then we may need to provide a function to handle all possible errors :thinking:, but I wonder whether it's too hard/complex to implement this function right considering all these situations. As MySQL itself is not clean, the attitude to warnings in different situations is different. For example, casting a json string to integer can have different warnings compared with casting a normal string to integer:

```
select cast('-1' as unsigned); -- has warning
select cast(cast('"-1"' as json) as unsigned); -- no warning
```

(Though, TiDB doesn't do well in warning compatibility now. Maybe it's not a big problem).

lcwangchao:
@YangKeao 

> Do we agree that the error handling in types package should follow similar pattern? Which means the function in types package should do its best-effort, and return valid values and all errors, the caller of these functions should take the responsibility to handle these errors.

Yes, I agree with you. The best way is to remove all flags and force return the error. However, I'm not sure it is practice now because too many places rely on the error processing inside the `types` package. Maybe we should reconsider it after finishing decoupling `types` and `stmtctx`.

> The problem is that, some functions will return both TruncateError and OverflowError (in different cases), then we may need to provide a function to handle all possible errors 🤔, but I wonder whether it's too hard/complex to implement this function right considering all these situations. 

Maybe need serval functions to translate `SQLMode` and other information such as statement type to `ErrorHandleContext`. At this time, we can still have one function to handle these error and just make the `ErrorHandleContext` as the input parameter which has different settings in different scene.

XuHuaiyu (approved):

PR #48013, opened 2023-10-26
dumpling: support output dialect

### LINKED ISSUE #47964: Allow Dumpling CSV being ingested by Redshift
## Enhancement

Usually, a BINARY column is represented in VARBYTE type in Redshift. Currently Dumpling outputs CSV as follows:

```sql
create table t (i int, b binary);
insert into t values (1, 'a'), (2, 'b'), (3, 'c');
```

Generated CSV:

```
"i","b"
1,"a"
2,"b"
3,"c"
```

However, Redshift requires the CSV content represented in Hex format when ingesting CSV files: https://docs.aws.amazon.com/redshift/latest/dg/copy-usage-varbyte.html

We propose to provide a new parameter called `--dialect` to control the output behavior.

When `--dialect=redshift`, the BINARY column in the CSV will be output as what Redshift requires.

Also, in future, when we discover other incompatible problems we could enhance this parameter, for example, we could add more transformations to the `--dialect=redshift` when necessary, and even introduce other dialects.

Lloyd-Pottiger:

Issue Number: close #47964

Problem Summary:

Now dumpling use UTF8 encoding for binary by default, but not all other databases can read UTF8 encoding directly.

Add `--csv-output-dialect` flag to support more encoding methods and even not standard methods.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Dumpling add `--csv-output-dialect` flag to support more encoding methods
```

### DIFF
--- a/dumpling/export/config.go
+++ b/dumpling/export/config.go
@@ -75,10 +75,43 @@ const (
 	flagReadTimeout              = "read-timeout"
 	flagTransactionalConsistency = "transactional-consistency"
 	flagCompress                 = "compress"
+	flagCsvOutputDialect         = "csv-output-dialect"
 
 	// FlagHelp represents the help flag
 	FlagHelp = "help"
 )
+
+// CSVDialect is the dialect of the CSV output for compatible with different import target
+type CSVDialect int
+
+const (
+	// CSVDialectDefault is the default dialect, which is MySQL/MariaDB/TiDB etc.
+	CSVDialectDefault CSVDialect = iota
+	// CSVDialectSnowflake is the dialect of Snowflake
+	CSVDialectSnowflake
+	// CSVDialectRedshift is the dialect of Redshift
+	CSVDialectRedshift
+)
+
+// BinaryFormat is the format of binary data
+// Three standard formats are supported: UTF8, HEX and Base64 now.
+type BinaryFormat int
+
+const (
+	// BinaryFormatUTF8 is the default format, format binary data as UTF8 string
+	BinaryFormatUTF8 BinaryFormat = iota
+	// BinaryFormatHEX format binary data as HEX string, e.g. 12ABCD
+	BinaryFormatHEX
+	// BinaryFormatBase64 format binary data as Base64 string, e.g. 123qwer==
+	BinaryFormatBase64
+)
+
+// DialectBinaryFormatMap is the map of dialect and binary format
+var DialectBinaryFormatMap = map[CSVDialect]BinaryFormat{
+	CSVDialectDefault:   BinaryFormatUTF8,
+	CSVDialectSnowflake: BinaryFormatHEX,
+	CSVDialectRedshift:  BinaryFormatHEX,
+}
 
 // Config is the dump config for dumpling
 type Config struct {
@@ -142,6 +175,7 @@ type Config struct {
 	SessionParams       map[string]interface{}
 	Tables              DatabaseTables
 	CollationCompatible string
+	CsvOutputDialect    CSVDialect
 
 	Labels       prometheus.Labels       `json:"-"`
 	PromFactory  promutil.Factory        `json:"-"`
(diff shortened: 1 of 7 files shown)


dumpling/export/config.go
```diff
@@ -313,6 +342,7 @@ func (*Config) DefineFlags(flags *pflag.FlagSet) {
 	flags.Bool(flagTransactionalConsistency, true, "Only support transactional consistency")
 	_ = flags.MarkHidden(flagTransactionalConsistency)
 	flags.StringP(flagCompress, "c", "", "Compress output file type, support 'gzip', 'snappy', 'zstd', 'no-compression' now")
+	flags.String(flagOutputDialect, "default", "The dialect of output file, support 'default', 'snowflake', 'redshift' now")
```
lance6716, line 345: Maybe it better to add a "binary-encoding" option with available value of "quote-escaped", "base64", "hex", "hex-with-leading-0x".

If we use "snowflake", "redshift" as an user I think this software will promise to keep all feature capable with "snowflake", "redshift"

breezewish: Yes, you are right. The purpose of `output-dialect` is exactly what you mentioned: We would like the output of our toolchain to be compatible with Redshift / Snowflake.

https://github.com/pingcap/tidb/issues/47964

If we only introduce binary-encoding without compatibility promising, then this PR may not resolve the above issue, after all it will only introduce a new feature called "transform binary encoding" instead of "allow the CSV being ingested by Redshift".

dumpling/export/config.go
```diff
@@ -75,11 +75,44 @@ const (
 	flagReadTimeout              = "read-timeout"
 	flagTransactionalConsistency = "transactional-consistency"
 	flagCompress                 = "compress"
+	flagOutputDialect            = "output-dialect"
 
 	// FlagHelp represents the help flag
 	FlagHelp = "help"
 )
 
+// Dialect is the dialect of different database
```
breezewish, line 84:
```suggestion
// Dialect is the dialect of the CSV output for compatible with different import target
```

dumpling/export/config.go
```diff
@@ -546,6 +582,21 @@ func (conf *Config) ParseFromFlags(flags *pflag.FlagSet) error {
 		return errors.Trace(err)
 	}
 
+	dialect, err := flags.GetString(flagOutputDialect)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	if dialect != "" && conf.SQL != "" {
+		return errors.New("--output-dialect is only supported when dumping whole table to csv, not compatible with --sql")
```
breezewish, line 590:
```suggestion
		return errors.New("--output-dialect is currently not compatible with --sql")
```

breezewish (commented):
Mostly looks good. A better impl might be modifying `WriteToBufferInCsv`, so that it becomes a true "output dialect" that can be fully compatible with `--sql`.

breezewish (approved):

dumpling/export/sql_type.go
```diff
@@ -64,6 +66,10 @@ func initColTypeRowReceiverMap() {
 	}
 }
 
+func initUsedBinaryFormat(bf BinaryFormat) {
+	usedBinaryFormat = bf
```
lichunzhu, line 70: How about putting this configuration in `*csvOption` like other csv options?
DM will may start several dumpling tasks, I'm afraid there will be a conflict.

lichunzhu (commented):
Rest LGTM. Great job!

dumpling/export/sql_type.go
Lloyd-Pottiger: Good idea, fixed.

dumpling/export/config.go
```diff
@@ -313,6 +351,7 @@ func (*Config) DefineFlags(flags *pflag.FlagSet) {
 	flags.Bool(flagTransactionalConsistency, true, "Only support transactional consistency")
 	_ = flags.MarkHidden(flagTransactionalConsistency)
 	flags.StringP(flagCompress, "c", "", "Compress output file type, support 'gzip', 'snappy', 'zstd', 'no-compression' now")
+	flags.String(flagCsvOutputDialect, "", "The dialect of output file, support 'snowflake', 'redshift' now, only take effect when dump whole table to csv")
```
breezewish, line 354:
```suggestion
	flags.String(flagCsvOutputDialect, "", "The dialect of output CSV file, support 'snowflake', 'redshift' now")
```

breezewish (approved):

lichunzhu (approved):
LGTM :)

PR #48017, opened 2023-10-26
domain: use with kv timeout feature schema load kv timeout

cfzjywxk:

Issue Number: close #xxx

Problem Summary:
Use kv timeout feature for schema reload.

When the leader peer of meta region is slow, use kv timeout could help alleviated by enabling the KV read timeout, avoid errors like "schema lease expire". For example when injecting slowness into the meta region TiKV node,

img_v2_bdffa267-615e-43ef-8ff3-2afc80a499fg

### Check List

Tests 

- [ ] Unit test

Side effects

Documentation

### Release note

```release-note
None
```

### DIFF
--- a/domain/domain.go
+++ b/domain/domain.go
@@ -795,7 +795,7 @@ func (do *Domain) loadSchemaInLoop(ctx context.Context, lease time.Duration) {
 	defer util.Recover(metrics.LabelDomain, "loadSchemaInLoop", nil, true)
 	// Lease renewal can run at any frequency.
 	// Use lease/2 here as recommend by paper.
-	ticker := time.NewTicker(lease / 2)
+	ticker := time.NewTicker(lease / 10)
 	defer func() {
 		ticker.Stop()
 		do.wg.Done()
--- a/domain/domain.go
+++ b/domain/domain.go
@@ -184,6 +184,7 @@ func (do *Domain) loadInfoSchema(startTS uint64) (infoschema.InfoSchema, bool, i
 		loadSchemaDurationTotal.Observe(time.Since(beginTime).Seconds())
 	}()
 	snapshot := do.store.GetSnapshot(kv.NewVersion(startTS))
+	snapshot.SetOption(kv.TiKVClientReadTimeout, uint64(3000)) // 3000ms.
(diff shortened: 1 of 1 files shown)


domain/domain.go
```diff
@@ -184,6 +184,7 @@ func (do *Domain) loadInfoSchema(startTS uint64) (infoschema.InfoSchema, bool, i
 		loadSchemaDurationTotal.Observe(time.Since(beginTime).Seconds())
 	}()
 	snapshot := do.store.GetSnapshot(kv.NewVersion(startTS))
+	snapshot.SetOption(kv.TiKVClientReadTimeout, uint64(3000)) // 3000ms.
```
zimulala, line 187: What is the default value if not set?

cfzjywxk: @zimulala 
The default value of `Get` is 30s and `BatchGet` is 60s.

cfzjywxk: If `snapshot.SetOption(kv.ReplicaRead, kv.ReplicaReadMixed)` is not set, the leader peer would be tried first. If timeout happens the [error is cached](https://github.com/tikv/client-go/blob/tidb-6.5-with-kv-timeout-feature/internal/locate/region_request.go#L1569), the replica selector [would be transferfred to](https://github.com/tikv/client-go/blob/tidb-6.5-with-kv-timeout-feature/internal/locate/region_request.go#L913) `TryFollower` automatically , then follower read [would be used](https://github.com/tikv/client-go/blob/tidb-6.5-with-kv-timeout-feature/internal/locate/region_request.go#L439) on the other peers.

So it's not necessary to involve follower read in the first try I think.

cfzjywxk: @Tema 

The fast automatic fallback happens only when the timeout is configured to a value smallter than the default value, [here](https://github.com/tikv/client-go/blob/tidb-6.5-with-kv-timeout-feature/internal/locate/region_request.go#L901) is the check logic.
In the default path, the retry may happen at most `maxReplicaAttempt` times.

crazycs520 (approved):

you06 (approved):

PR #48025, opened 2023-10-27
executor/brie: use the default value from flags

### LINKED ISSUE #48000: br: disable the stats information backup by default for sql command
## Enhancement
Stats information backup needs to call `dumpStatsToJson` function which costs a lot of memory especially when there are a lot of database objects.

For br clp, flag `ignore-stats` is turned on by default, while it's not for backup sql command. So, backup could bring OOM issues.

We need to sync the behavior to as the clp, and also expose the parameter to sql command.

Via this issue, we also needs to sync up the default flags including `use-checkpoint`.

YuJuncen:

Issue Number: close #48000

Problem Summary:
The current default config parameter is different from the BR CLI's. For example, `SkipStatistics` is always `false` (in BR it is defaultly `true`), `UseCheckpoint` is always `false` too (in BR it is defaultly `true`).

This PR exports the default config via the command line parameter of BR. And we will use it at the BRIE via SQL.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

### Release note

```release-note
Now, BRIE via SQL will use the same default config with BR CLI.
```

### DIFF
--- a/pkg/executor/brie.go
+++ b/pkg/executor/brie.go
@@ -277,21 +277,15 @@ func (b *executorBuilder) buildBRIE(s *ast.BRIEStmt, schema *expression.Schema) 
 	}
 
 	tidbCfg := config.GetGlobalConfig()
-	cfg := task.Config{
-		TLS: task.TLSConfig{
-			CA:   tidbCfg.Security.ClusterSSLCA,
-			Cert: tidbCfg.Security.ClusterSSLCert,
-			Key:  tidbCfg.Security.ClusterSSLKey,
-		},
-		PD:          strings.Split(tidbCfg.Path, ","),
-		Concurrency: 4,
-		Checksum:    true,
-		SendCreds:   true,
-		LogProgress: true,
-		CipherInfo: backuppb.CipherInfo{
-			CipherType: encryptionpb.EncryptionMethod_PLAINTEXT,
-		},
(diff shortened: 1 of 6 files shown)


BornChanger:
please associate it with issue #48000 @YuJuncen

BornChanger:
please also provide those 2 parameters to BR sql command option parser. @YuJuncen

pkg/executor/brie.go
```diff
@@ -392,7 +388,9 @@ func (b *executorBuilder) buildBRIE(s *ast.BRIEStmt, schema *expression.Schema)
 		}
 
 	case ast.BRIEKindRestore:
-		e.restoreCfg = &task.RestoreConfig{Config: cfg}
+		rcfg := task.DefaultRestoreConfig()
+		rcfg.Config = cfg
```
Leavrth, line 392: It seems that `rcfg.Config` has been parsed in the `task.DefaultRestoreConfig()`.

Leavrth (approved):

pkg/executor/brie.go
YuJuncen: We will override the default config with the new config. Will this make things wrong?

BornChanger (approved):

YuJuncen:
√

PR #48032, opened 2023-10-27
expression: fix the return type of `coalesce` when arg type is `DATE`

### LINKED ISSUE #46475: The result of comparison function on date value should return DATE type
## Bug Report

```
bash$ mysql --column-type-info

mysql> select coalesce(current_date(),null) from dual;
```

```
Field   1:  `coalesce(current_date(),null)`
Catalog:    `def`
Database:   ``
Table:      ``
Org_table:  ``
Type:       DATE
...
```

```
Field   1:  `coalesce(current_date(),null)`
Catalog:    `def`
Database:   ``
Table:      ``
Org_table:  ``
Type:       DATETIME
...
```

v6.6.0

xzhangxian1008:

Issue Number: close #46475

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the return type of coalesce
```

### DIFF
--- a/pkg/expression/builtin_compare.go
+++ b/pkg/expression/builtin_compare.go
@@ -141,7 +141,9 @@ func (c *coalesceFunctionClass) getFunction(ctx sessionctx.Context, args []Expre
 		return nil, err
 	}
 
-	bf.tp = resultFieldType
+	if resultFieldType.GetType() == mysql.TypeDate {
+		bf.setDecimalAndFlenForDate()
+	}
 
 	bf.tp.AddFlag(resultFieldType.GetFlag())
 	resultFieldType.SetFlen(0)
--- a/pkg/expression/builtin_compare.go
+++ b/pkg/expression/builtin_compare.go
@@ -1105,9 +1105,11 @@ func (b *builtinIntervalIntSig) evalInt(row chunk.Row) (int64, bool, error) {
 	}
 	isUint1 := mysql.HasUnsignedFlag(b.args[0].GetType().GetFlag())
 	var idx int
-
-	idx, err = b.linearSearch(arg0, isUint1, b.args[1:], row)
-
+	if b.hasNullable {
+		idx, err = b.linearSearch(arg0, isUint1, b.args[1:], row)
+	} else {
+		idx, err = b.binSearch(arg0, isUint1, b.args[1:], row)
+	}
 	return int64(idx), err != nil, err
 }
 
@@ -1140,6 +1142,42 @@ func (b *builtinIntervalIntSig) linearSearch(target int64, isUint1 bool, args []
 	return i, nil
 }
 
+// binSearch is a binary search method.
+// All arguments are treated as integers.
+// It is required that arg[0] < args[1] < args[2] < ... < args[n] for this function to work correctly.
+// This is because a binary search is used (very fast).
+func (b *builtinIntervalIntSig) binSearch(target int64, isUint1 bool, args []Expression, row chunk.Row) (_ int, err error) {
+	i, j, cmp := 0, len(args), false
+	for i < j {
+		mid := i + (j-i)/2
+		v, isNull, err1 := args[mid].EvalInt(b.ctx, row)
+		if err1 != nil {
+			err = err1
+			break
+		}
+		if isNull {
+			v = target
+		}
+		isUint2 := mysql.HasUnsignedFlag(args[mid].GetType().GetFlag())
+		switch {
+		case !isUint1 && !isUint2:
+			cmp = target < v
+		case isUint1 && isUint2:
+			cmp = uint64(target) < uint64(v)
+		case !isUint1 && isUint2:
+			cmp = target < 0 || uint64(target) < uint64(v)
+		case isUint1 && !isUint2:
+			cmp = v > 0 && uint64(target) < uint64(v)
+		}
+		if !cmp {
+			i = mid + 1
+		} else {
+			j = mid
+		}
+	}
+	return i, err
+}
+
 type builtinIntervalRealSig struct {
 	baseBuiltinFunc
 	hasNullable bool
@@ -1164,7 +1202,11 @@ func (b *builtinIntervalRealSig) evalInt(row chunk.Row) (int64, bool, error) {
 	}
 
 	var idx int
-	idx, err = b.linearSearch(arg0, b.args[1:], row)
+	if b.hasNullable {
+		idx, err = b.linearSearch(arg0, b.args[1:], row)
+	} else {
+		idx, err = b.binSearch(arg0, b.args[1:], row)
+	}
 	return int64(idx), err != nil, err
 }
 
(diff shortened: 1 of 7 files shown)


pkg/expression/builtin_compare.go
```diff
@@ -141,6 +141,10 @@ func (c *coalesceFunctionClass) getFunction(ctx sessionctx.Context, args []Expre
 		return nil, err
 	}
 
+	if resultFieldType.GetType() == mysql.TypeDate {
+		bf.setDecimalAndFlenForDate()
+	}
```
guo-shaoge, line 146: I'm ok with this fix for this specific issue.
But I wonder why there is no `ETDate`, which makes this type deduction wrong, which cause we have to add this kind of code for every similar situation.

xzhangxian1008: > Why not just set resultFieldType to bf.tp? For control functions(case when/ifnull), it use this way to avoid this problem.

The type of resultFieldType is `EvalType` and `EvalType` doesn't have `Date` type.

windtalker: I think `resultEvalType` is the `EvalType`, `resultFieldType` is field type.

xzhangxian1008: > I think `resultEvalType` is the `EvalType`, `resultFieldType` is field type.

okk, I will change it.

pkg/expression/builtin_compare.go
```diff
@@ -141,6 +141,8 @@ func (c *coalesceFunctionClass) getFunction(ctx sessionctx.Context, args []Expre
 		return nil, err
 	}
 
+	bf.tp = resultFieldType
+
 	bf.tp.AddFlag(resultFieldType.GetFlag())
```
windtalker, line 146: This line can be removed

windtalker: And I found there is lots of code(L145-L197) to adjust `resultFieldType` according to input fieldtype, can you double check if we can use `resultFieldType, err := InferType4ControlFuncs(xxx)`, and remove these extra adjustments? Because theoretically, the type infer for `coalesce` should be the same as `case when/ifnull`, which all use `InferType4ControlFuncs`

pkg/expression/builtin_compare.go
xzhangxian1008: > I think `resultEvalType` is the `EvalType`, `resultFieldType` is field type.

From the failure of ut, we can see that when sql is `select coalesce(c_int_d, c_datetime) from xxx;` the test will fail. So, we can't directly set resultFieldType to bf.tp.

pkg/expression/builtin_compare.go
xzhangxian1008: > And I found there is lots of code(L145-L197) to adjust `resultFieldType` according to input fieldtype, can you double check if we can use `resultFieldType, err := InferType4ControlFuncs(xxx)`, and remove these extra adjustments? Because theoretically, the type infer for `coalesce` should be the same as `case when/ifnull`, which all use `InferType4ControlFuncs`

After check, I think we can not remove these extra adjustments.

pkg/expression/builtin_compare.go
```diff
@@ -119,18 +119,19 @@ func (c *coalesceFunctionClass) getFunction(ctx sessionctx.Context, args []Expre
 		return nil, err
 	}
 
-	fieldTps := make([]*types.FieldType, 0, len(args))
+	flag := uint(0)
```
windtalker, line 122: Why need to handle this NotNullFlag here? I think `InferType4ControlFuncs` already handles this flag?

pkg/expression/builtin_compare.go
```diff
@@ -1157,11 +1105,9 @@ func (b *builtinIntervalIntSig) evalInt(row chunk.Row) (int64, bool, error) {
 	}
 	isUint1 := mysql.HasUnsignedFlag(b.args[0].GetType().GetFlag())
 	var idx int
-	if b.hasNullable {
```
windtalker, line 1160: why change this?

pkg/expression/builtin_control.go
```diff
@@ -117,37 +117,60 @@ func setDecimalFromArgs(evalType types.EvalType, resultFieldType *types.FieldTyp
...
+	}
+	return false
+}
+
 func addCollateAndCharsetAndFlagFromArgs(ctx sessionctx.Context, funcName string, evalType types.EvalType, resultFieldType *types.FieldType, args ...Expression) error {
 	switch funcName {
-	case ast.If, ast.Ifnull, ast.WindowFuncLead, ast.WindowFuncLag:
-		if len(args) != 2 {
-			panic("unexpected length of args for if/ifnull/lead/lag")
+	case ast.If, ast.Ifnull, ast.WindowFuncLead, ast.WindowFuncLag, ast.Coalesce:
```
windtalker, line 142: maybe `coalesce` should be treated the same as `case`, so you don't need to change these codes

pkg/expression/builtin_control.go
```diff
@@ -198,8 +221,8 @@ func InferType4ControlFuncs(ctx sessionctx.Context, funcName string, args ...Exp
 		tempFlag := resultFieldType.GetFlag()
 		types.SetTypeFlag(&tempFlag, mysql.NotNullFlag, false)
 		resultFieldType.SetFlag(tempFlag)
-		// If both arguments are NULL, make resulting type BINARY(0).
-		resultFieldType.SetType(mysql.TypeString)
+
+		resultFieldType.SetType(mysql.TypeNull)
```
windtalker, line 225: Is this behavior the same as MySQL?

xzhangxian1008: > Is this behavior the same as MySQL?

Yes, when we execute `select coalesce(null,null);` the return type is:
```
Field   1:  `coalesce(null,null)`
Catalog:    `def`
Database:   ``
Table:      ``
Org_table:  ``
Type:       NULL
Collation:  binary (63)
Length:     0
Max_length: 0
Decimals:   0
Flags:      BINARY NUM
```

pkg/expression/builtin_compare.go
xzhangxian1008: > Why need to handle this NotNullFlag here? I think `InferType4ControlFuncs` already handles this flag?

After check, I think `InferType4ControlFuncs` does not handle this flag.

windtalker: `InferType4ControlFuncs` is a common type infer function that aims at aggregating some input types into one "common return type", it should be used by all the functions whose return type has this pattern, for example: `ifnull/case when/coalesce`, if you need add code to handle `NotNullFlag` in `coalesce`, you should double check the following question:
- Why you need to add special code to handle `NotNullFlag` in `coalesce` only? Will the problem you meet also happens in other functions that use `InferType4ControlFuncs` as its type infer function?

windtalker (approved):
LGTM

wshwsh12 (approved):

PR #48058, opened 2023-10-28
ddl: Ensure correct errors for duplicate constraint names | tidb-test=pr/2231

### LINKED ISSUE #48040: Constraint check name uses wrong error code
## Bug Report

In MySQL error 3822 is used for duplicate check constraint names, but TiDB uses 1061.

```
CREATE TABLE t3 (id INT PRIMARY KEY, CONSTRAINT c3 CHECK (id<50), CONSTRAINT c3 CHECK (id<40));
```

MySQL 8.2.0
```
sql> CREATE TABLE t3 (id INT PRIMARY KEY, CONSTRAINT c3 CHECK (id<50), CONSTRAINT c3 CHECK (id<40));
ERROR: 3822 (HY000): Duplicate check constraint name 'c3'.
```

```
sql> SET GLOBAL tidb_enable_check_constraint=ON;
Query OK, 0 rows affected (0.0029 sec)

sql> CREATE TABLE t3 (id INT PRIMARY KEY, CONSTRAINT c3 CHECK (id<50), CONSTRAINT c3 CHECK (id<40));
ERROR: 1061 (42000): duplicate key name c3
```

```
Release Version: v7.6.0-alpha-23-gdfd9080bcf-dirty
Edition: Community
Git Commit Hash: dfd9080bcf49083e52d6cb714668da9b881dc54d
Git Branch: master
UTC Build Time: 2023-10-27 06:35:10
GoVersion: go1.21.0
Race Enabled: false
Store: unistore
```

dveeden:

Issue Number: close #48040 

Problem Summary:

Ensure correct error codes and error text for duplicate constraint names.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  >   

Side effects

Documentation

- [x] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Duplicate check constraints now use the correct error code and text
```

### DIFF
--- a/pkg/ddl/ddl_api.go
+++ b/pkg/ddl/ddl_api.go
@@ -1828,16 +1828,26 @@ func setEmptyConstraintName(namesMap map[string]bool, constr *ast.Constraint) {
 
 func checkConstraintNames(tableName model.CIStr, constraints []*ast.Constraint) error {
 	constrNames := map[string]bool{}
-	checkConstraints := make([]*ast.Constraint, 0, len(constraints))
+	fkNames := map[string]bool{}
 
 	// Check not empty constraint name whether is duplicated.
 	for _, constr := range constraints {
-		err := checkDuplicateConstraint(constrNames, constr.Name, constr.Tp)
-		if err != nil {
-			return errors.Trace(err)
+		if constr.Tp == ast.ConstraintForeignKey {
+			err := checkDuplicateConstraint(fkNames, constr.Name, constr.Tp)
+			if err != nil {
+				return errors.Trace(err)
+			}
+		} else {
+			err := checkDuplicateConstraint(constrNames, constr.Name, constr.Tp)
+			if err != nil {
+				return errors.Trace(err)
+			}
 		}
+	}
 
-		// Set empty constraint names.
+	// Set empty constraint names.
+	checkConstraints := make([]*ast.Constraint, 0, len(constraints))
+	for _, constr := range constraints {
 		if constr.Tp != ast.ConstraintForeignKey {
 			setEmptyConstraintName(constrNames, constr)
 		}
--- a/pkg/ddl/ddl_api.go
+++ b/pkg/ddl/ddl_api.go
@@ -1760,16 +1760,20 @@ func checkColumnAttributes(colName string, tp *types.FieldType) error {
 	return nil
 }
 
-func checkDuplicateConstraint(namesMap map[string]bool, name string, foreign bool) error {
+func checkDuplicateConstraint(namesMap map[string]bool, name string, constraintType ast.ConstraintType) error {
 	if name == "" {
 		return nil
 	}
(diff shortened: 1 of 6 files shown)


CbcWestwolf (approved):
LGTM

pkg/ddl/ddl_api.go
```diff
@@ -7252,7 +7256,7 @@ func (d *ddl) createIndex(ctx sessionctx.Context, ti ast.Ident, keyType ast.Inde
 				"a background job is trying to add the same index, "+
 				"please check by `ADMIN SHOW DDL JOBS`", indexName)
 		} else {
-			err = dbterror.ErrDupKeyName.GenWithStack("index already exist %s", indexName)
+			err = dbterror.ErrDupKeyName.GenWithStackByArgs(indexName)
```
zimulala, line 7259: Do we need to update line 7255 like this(using `Duplicate key name` instead of `index already exist`)?
Besides, do we need to unify this error msg in https://github.com/pingcap/tidb/blob/3682bd82cac05a75241ed9fc9f29a1d320d776a7/pkg/ddl/index.go#L567

dveeden: Yes this is needed to address https://github.com/pingcap/tidb/issues/48040#issuecomment-1783773545

Looks like this is used in three locations:
```
$ git grep '"index already exist' | grep -v '_test.go'
pkg/ddl/ddl_api.go:			err = dbterror.ErrDupKeyName.GenWithStack("index already exist %s; "+
pkg/ddl/index.go:			err = dbterror.ErrDupKeyName.GenWithStack("index already exist %s", indexName)
pkg/ddl/schematracker/dm_tracker.go:		return dbterror.ErrDupKeyName.GenWithStack("index already exist %s", indexName)
```

If might be useful to see what situations trigger those and if this is compatible with MySQL.

dveeden: For compatibility with MySQL we should make sure that:
- The error code is the same unless this is really impossible.
- The error text is the same unless there is a strong reason for deviating. If we deviate we should try to add to the message instead of replacing it.

dveeden: For the error in `pkg/ddl/ddl_api.go`:

This seems to happen when a background job is adding the same index. I think this is a TiDB specific situation, so this is probably fine.

zimulala: I think it should be changed to what I said about `pkg/ddl/ddl_api.go`, leaving the supplementary information intact, but replacing `index already exist` with `Duplicate key name`. 
Because I think line7255 and line7259 are the same error, but line7255 adds some extra information.

dveeden: done in 2e73d713dc900caf11a8fe4be30155feb6b33ae2

zimulala (approved):
LGTM

PR #48237, opened 2023-11-02
parse: fix the type of date/time parameters

### LINKED ISSUE #45190: Date parameters are converted to string automatically
TiDB handles the date parameters as string, which is not expected. TiDB should construct a data, but not string for these parameters. See:

```go
		case mysql.TypeDate, mysql.TypeTimestamp, mysql.TypeDatetime:
			if len(paramValues) < (pos + 1) {
				err = mysql.ErrMalformPacket
				return
			}
			// See https://dev.mysql.com/doc/internals/en/binary-protocol-value.html
			// for more details.
			length := paramValues[pos]
			pos++
			switch length {
			case 0:
				tmp = types.ZeroDatetimeStr
			case 4:
				pos, tmp = binaryDate(pos, paramValues)
			case 7:
				pos, tmp = binaryDateTime(pos, paramValues)
			case 11:
				pos, tmp = binaryTimestamp(pos, paramValues)
			case 13:
				pos, tmp = binaryTimestampWithTZ(pos, paramValues)
			default:
				err = mysql.ErrMalformPacket
				return
			}
			args[i] = types.NewDatum(tmp) // FIXME: After check works!!!!!!
			continue
```

YangKeao:

Issue Number: close #45190

Problem Summary:

1. Refractor the `stmtctx` in `parse` to be `types.Context`.
2. Construct the date and time from string.

   I didn't construct the `Datetime` and `Duration` directly from the binary, because developing a binary version of `ParseXXX` can be a little complicated (especially for handling warnings/errors). To be simple and compatible with old versions, I re-use the string from old version and parse from it.

3. Refractor the function of parsing parameters into two parts: one for splitting binary protocols into multiple parameters, one for parsing the binary parameters into expressions. The second one needs to run after when the `stmtctx` is reset, so they need to be split. It also solves some historical problem, as I also saw `HandleTruncate`s besides the `date/time` part.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

### Release note

```release-note
None
```


pkg/param/binary_params.go
```diff
@@ -0,0 +1,275 @@
+// Copyright 2023 PingCAP, Inc.
```
YangKeao, line 1: Most of this file is copied from `/pkg/internal/parse`. Because it's not expected to be "internal" inside the `server`.

pkg/server/conn_stmt_params_test.go
```diff
@@ -0,0 +1,298 @@
...
+		return err
+	}
+
+	for i := 0; i < len(args); i++ {
+		args[i] = parsedArgs[i]
+	}
+	return
+}
+
+func TestParseExecArgs(t *testing.T) {
```
YangKeao, line 52: These tests are also moved from `internal/parse/parse_test.go`.

lcwangchao (approved):

lcwangchao:
LGTM, but is it possible to add some intergartion tests?

xhebox (approved):

YangKeao:
> LGTM, but is it possible to add some intergartion tests?

Good. Add some in the latest commit :smile:

PR #48406, opened 2023-11-08
planner: don't inline CTE when it's consumed by a recursive CTE

### LINKED ISSUE #47603: CTE query execute failed:Recursive Common Table Expression '' can contain neither aggregation nor window functions in recursive query block
## Bug Report

The results of cte query execution are different in v6.1.6 and v6.5.2 .

create table a (id int);
create table b (id int);
create table temp (lvl int);
with a as (select 8 as id from dual),
maxa as (select max(id) as max_id from a),
b as (
with recursive temp as (
select 1 as lvl from dual
union all
select lvl+1 from temp, maxa where lvl < max_id
)
select * from temp
)
select * from b;

**v6.1.6**
mysql> with a as (select 8 as id from dual),
    -> maxa as (select max(id) as max_id from a),
    -> b as (
    -> with recursive temp as (
    -> select 1 as lvl from dual
    -> union all
    -> select lvl+1 from temp, maxa where lvl < max_id
    -> )
    -> select * from temp
    -> )
    -> select * from b;
+------+
| lvl  |
+------+
|    1 |
|    2 |
|    3 |
|    4 |
|    5 |
|    6 |
|    7 |
|    8 |
+------+
8 rows in set (0.01 sec)

**v6.5.2**
mysql> with a as (select 8 as id from dual),
    -> maxa as (select max(id) as max_id from a),
    -> b as (
    -> with recursive temp as (
    -> select 1 as lvl from dual
    -> union all
    -> select lvl+1 from temp, maxa where lvl < max_id
    -> )
    -> select * from temp
    -> )
    -> select * from b;
**ERROR 3575 (HY000): Recursive Common Table Expression '' can contain neither aggregation nor window functions in recursive query block**

v6.5.2

YangKeao:

Issue Number: close #47603 

Problem Summary:

When a CTE containing aggregation is inlined into a recursive CTE, it may return error.

I add some logic in preprocess to detect whether a CTE is consumed by a recursive CTE, if so (and it contains an aggregation), don't inline it.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  >   

Side effects

### Release note

```release-note
Fix the issue that a CTE with aggregation can be automatically inlined into a recursive CTE and cause error.
```

### DIFF
--- a/pkg/parser/ast/dml.go
+++ b/pkg/parser/ast/dml.go
@@ -1067,6 +1067,8 @@ type CommonTableExpression struct {
 
 	// Record how many consumers the current cte has
 	ConsumerCount int
+	// ConsumedByRecursiveCTE records whether this CTE is used by a recursive CTE
+	ConsumedByRecursiveCTE bool
 }
 
 // Restore implements Node interface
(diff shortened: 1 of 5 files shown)


YangKeao:
Duplicated with https://github.com/pingcap/tidb/pull/48188. Close.

Issue #48411, opened 2023-11-08
select for update statement can't got stmt-count-limit error


crazycs520:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

For minimum reproduce, the TiDB's configuration as follow:

```sql
[performance]
stmt-count-limit=3
```

Reproduce SQL:

```sql
create table t (id int key);
set @@tidb_disable_txn_auto_retry=0;
set autocommit=0;
begin optimistic;
insert into t values (1);
insert into t values (2);
select * from t for update;
insert into t values (3);
insert into t values (4);
commit;
select * from t;
```

### 2. What did you expect to see? (Required)

When execute `select * from t for update;`, should got error:

```sql
mysql> set @@tidb_disable_txn_auto_retry=0;
Query OK, 0 rows affected (0.00 sec)

mysql> set autocommit=0;
Query OK, 0 rows affected (0.00 sec)

mysql> begin optimistic;
Query OK, 0 rows affected (0.00 sec)

mysql> insert into t values (1);
Query OK, 1 row affected (0.00 sec)

mysql> insert into t values (2);
Query OK, 1 row affected (0.00 sec)

mysql> select * from t for update;
ERROR 1105 (HY000): statement count 4 exceeds the transaction limitation, transaction has been rollback, autocommit = false
mysql> insert into t values (3);
Query OK, 1 row affected (0.00 sec)

mysql> insert into t values (4);
Query OK, 1 row affected (0.00 sec)

mysql> commit;
Query OK, 0 rows affected (0.00 sec)

mysql> select * from t;
+----+
| id |
+----+
|  3 |
|  4 |
+----+
2 rows in set (0.00 sec)
```

### 3. What did you see instead (Required)

All SQL are executed successfully, but the finally select result are following, **the user may think that the transaction has been partially committed.**

```
mysql> set @@tidb_disable_txn_auto_retry=0;
Query OK, 0 rows affected (0.00 sec)

mysql> set autocommit=0;
Query OK, 0 rows affected (0.00 sec)

mysql> begin optimistic;
Query OK, 0 rows affected (0.00 sec)

mysql> insert into t values (1);
Query OK, 1 row affected (0.00 sec)

mysql> insert into t values (2);
Query OK, 1 row affected (0.00 sec)

mysql> select * from t for update;
+----+
| id |
+----+
|  1 |
|  2 |
+----+
2 rows in set (0.00 sec)

[...]

PR #48412, opened 2023-11-08
session: fix select for update statement can't get stmt-count-limit error

crazycs520:

Issue Number: close #48411 

Problem Summary: fix select for update statement can't got stmt-count-limit error

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix issue of select for update statement can't get stmt-count-limit error
```


pkg/session/session.go
```diff
@@ -2430,6 +2430,11 @@ func runStmt(ctx context.Context, se *session, s sqlexec.Statement) (rs sqlexec.
 				}
 			}
 		}
+		if err == nil && sessVars.TxnCtx.CouldRetry && !s.IsReadOnly(sessVars) {
```
cfzjywxk, line 2433: This checking condition is confusing, could you explain it and why put it here?
IMO the `checkStmtLimit` should be put to a unified place as a hook which all SQL statement processing would use.

pkg/session/tidb.go
```diff
@@ -305,17 +305,21 @@ func autoCommitAfterStmt(ctx context.Context, se *session, meetsErr error, sql s
 	return nil
 }
 
-func checkStmtLimit(ctx context.Context, se *session) error {
+func checkStmtLimit(ctx context.Context, se *session, beforeExec bool) error {
```
ekexium, line 308: Should it be named like `beforeFinish`?

tiancaiamao (dismissed):

pkg/session/tidb.go
crazycs520: already updated, PTAL

pkg/session/session.go
crazycs520: Nice catch , already updated, PTAL

pkg/session/session.go
```diff
@@ -2417,6 +2417,12 @@ func runStmt(ctx context.Context, se *session, s sqlexec.Statement) (rs sqlexec.
 	if err != nil {
 		return nil, err
 	}
+	if sessVars.TxnCtx.CouldRetry && !s.IsReadOnly(sessVars) {
+		// Only when the txn is could retry and the statement is not read only, need to do stmt-count-limit check.
```
ekexium: Maybe a link to https://docs.pingcap.com/tidb/stable/tidb-configuration-file#stmt-count-limit is more clear to explain `CouldRetry`.

ekexium (approved):

pkg/session/session.go
crazycs520: done.

jackysp (commented):
I think if there is a `select for update` statement in an optimistic transaction, it can bypass the stmt-count-limit, and abort its history.

tiancaiamao (approved):

PR #48447, opened 2023-11-08
server,session: make sure ResultSet.Close() errors return to the client

### LINKED ISSUE #48446: error happened but not returned to the client
## Enhancement

By reproduce https://github.com/pingcap/tidb/issues/48411 and after reading the code, I found it might not be the only case that 
**error happened but not returned to the client**.

```
defer terror.Call(rs.Close())
for {
    data = rs.Next()
    writePacket(data)
}
writeEOF()
```

`terror.Call` ignores error, so the error is not returned to the client.

But **NOTE**, even we to write it as this, the code is still wrong:

```
defer func() {
    err = rs.Close()
}()
for {
    data = rs.Next()
    writePacket(data)
}
writeEOF()
```

The problem is that after we `writeEOF()`, we finish this mysql response telling the client that all the data is done.
So even we find an error later (for example, in defer rs.Close()), there is no way to tell the mysql client any more!

The correct order should be:

```
for {
    data = rs.Next()
    writePacket(data)
}
if err = rs.Close(); err != nil {
    writeErr()
} else {
    writeEOF()
}
```

tiancaiamao:

Issue Number: close #48446, ref #48411

Problem Summary:

This also fix #48411 in another way, and can prevent more bugs.

Close the result set after all rs.Next(), and before cc.writeEOF()
Avoid the case that error happen and can't be sent to the client after writeEOF.

### Check List

Tests 

The test code in https://github.com/pingcap/tidb/pull/48412 should also apply for this one, as the both can fix #48411, so I don't bother writer another duplicated test.

- [x] Unit test
- [ ] Integration test

See https://github.com/pingcap/tidb/issues/48411

- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/server/conn.go
+++ b/pkg/server/conn.go
@@ -2049,11 +2049,7 @@ func (cc *clientConn) handleStmt(ctx context.Context, stmt ast.StmtNode, warns [
 			execStmt.(*executor.ExecStmt).FinishExecuteStmt(0, err, false)
 		}
 	}
-	if err != nil {
-		return false, err
-	}
-
-	return false, nil
+	return false, err
 }
 
 func (cc *clientConn) handleFileTransInConn(ctx context.Context, status uint16) (bool, error) {
@@ -2284,6 +2280,9 @@ func (cc *clientConn) writeChunks(ctx context.Context, rs resultset.ResultSet, b
 		if stmtDetail != nil {
 			stmtDetail.WriteSQLRespDuration += time.Since(start)
 		}
+	}
+	if err := rs.Close(); err != nil {
+		return false, err
 	}
 
 	if stmtDetail != nil {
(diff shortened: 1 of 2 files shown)


crazycs520 (approved):

YangKeao (approved):
LGTM

tiancaiamao:
Reuse the unit test of the other PR https://github.com/pingcap/tidb/pull/48412/files#diff-d7a5d1557d888bb61179ecad9eb02d6275358a3954f74a8b5ab600e332843075R2461-R2470

Where the check need to be changed:

```
rows := dbt.MustQuery("select * from t for update;")
for rows.Next() {}
err := rows.Err() 
require.Error(t, err)
```

Because https://github.com/pingcap/tidb/pull/48412 return error to the client by the first packet, while this one return error in the last packet.

Issue #23363, opened 2021-03-17
Risk of breaking consistency when changing column from null to not null


sticnarf:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

Initial table:

```sql
CREATE TABLE t (id INT PRIMARY KEY, v INT);
INSERT INTO t VALUES (0, 0);
```
Change the value to `NULL` and change the value to not null concurrently.

```sql
-- Session 1
BEGIN;

-- Session 2 (Concurrently)
ALTER TABLE t MODIFY COLUMN v INT NOT NULL;

-- Session 1 (Concurrently)
UPDATE t SET v = NULL WHERE ID = 0;
COMMIT;
```

### 2. What did you expect to see? (Required)

Either the DDL or the transaction fails.

### 3. What did you see instead (Required)

I think it is possible that both the DDL and the transaction succeed. Because under async commit, there is no check after the not null flag is added.

Then, the table contains NULL, but the column schema is NOT NULL.

I'm still trying to create a test case that can reproduce it.

### 4. What is your TiDB version? (Required)

5.0/nightly

wjhuang2016:
Dose this can only be reproduced when using the async-commit?

sticnarf:
> Dose this can only be reproduced when using the async-commit?

Yes. If it's a normal transaction, the transaction will be aborted by the schema check between prewrite and commit.

ti-srebot:
## Please edit this comment or add a new comment to complete the following information

### Not a bug

1. Remove the 'type/bug' label
2. Add notes to indicate why it is not a bug

### Duplicate bug

1. Add the 'type/duplicate' label
2. Add the link to the original bug

### Bug

Note: Make Sure that 'component', and 'severity' labels are added
Example for how to fill out the template: https://github.com/pingcap/tidb/issues/20100

#### 1. Root Cause Analysis (RCA) (optional) 


#### 2. Symptom (optional)



#### 3. All Trigger Conditions (optional)



#### 4. Workaround (optional)

#### 5. Affected versions



#### 6. Fixed versions

PR #48524, opened 2023-11-11
ddl: fix foreign key and check constraint verify records under async commit

### LINKED ISSUE #48297: Is delayForAsyncCommit needed before verify records by foreign_key and constraints check?
## Bug Report

I see there is delayForAsyncCommit before verifying records, for example:
https://github.com/pingcap/tidb/blob/4f00ece106b1f9b13086d3e683939527d91718de/pkg/ddl/partition.go#L2496-L2498

But there is no delay used by foreign_key and constraints check.

https://github.com/pingcap/tidb/blob/4f00ece106b1f9b13086d3e683939527d91718de/pkg/ddl/foreign_key.go#L672
https://github.com/pingcap/tidb/blob/4f00ece106b1f9b13086d3e683939527d91718de/pkg/ddl/constraint.go#L339

Is it a potential bug under async commit?

I don't known how to reproduce it to verify this question. 

master

jiyfhust:

Issue Number: close #48297

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  >  
Code review instead of tests. 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/column.go
+++ b/pkg/ddl/column.go
@@ -731,6 +731,9 @@ func (w *worker) doModifyColumnTypeWithData(
 	case model.StateDeleteOnly:
 		// Column from null to not null.
 		if !mysql.HasNotNullFlag(oldCol.GetFlag()) && mysql.HasNotNullFlag(changingCol.GetFlag()) {
+			if d.lease > 0 {
+				delayForAsyncCommit()
+			}
 			// Introduce the `mysql.PreventNullInsertFlag` flag to prevent users from inserting or updating null values.
 			err := modifyColsFromNull2NotNull(w, dbInfo, tblInfo, []*model.ColumnInfo{oldCol}, targetCol, oldCol.GetType() != changingCol.GetType())
 			if err != nil {
--- a/pkg/ddl/constraint.go
+++ b/pkg/ddl/constraint.go
@@ -101,6 +101,9 @@ func (w *worker) onAddCheckConstraint(d *ddlCtx, t *meta.Meta, job *model.Job) (
 		constraintInfoInMeta.State = model.StateWriteReorganization
 		ver, err = updateVersionAndTableInfoWithCheck(d, t, job, tblInfo, true)
 	case model.StateWriteReorganization:
+		if d.lease > 0 {
+			delayForAsyncCommit()
+		}
 		err = w.verifyRemainRecordsForCheckConstraint(dbInfo, tblInfo, constraintInfoInMeta)
 		if err != nil {
 			if dbterror.ErrCheckConstraintIsViolated.Equal(err) {
@@ -238,6 +241,9 @@ func (w *worker) onAlterCheckConstraint(d *ddlCtx, t *meta.Meta, job *model.Job)
 			constraintInfo.State = model.StateWriteOnly
 			ver, err = updateVersionAndTableInfoWithCheck(d, t, job, tblInfo, true)
 		case model.StateWriteOnly:
+			if d.lease > 0 {
+				delayForAsyncCommit()
+			}
 			err = w.verifyRemainRecordsForCheckConstraint(dbInfo, tblInfo, constraintInfo)
 			if err != nil {
 				if dbterror.ErrCheckConstraintIsViolated.Equal(err) {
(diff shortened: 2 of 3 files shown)


jiyfhust:
/cc @CbcWestwolf @mjonss @Benjamin2037 
Hi，brothers, this pr is easy to review.

mjonss (commented):
Is it possible to add a test case showing the effect of this PR?

jiyfhust:
em... Who could help for this pr?

jiyfhust:
em... Who could help for this pr?

jiyfhust:
For example, modify column null to not null, also use (delayAsyncCommit)[https://github.com/pingcap/tidb/blob/0fd779c623a77ce13c7b6d9e42da615aed16f8e1/pkg/ddl/column.go#L1544]

jiyfhust:
The case:
1. set `tidb_enable_metadata_lock` off
2. set `tidb_enable_async_commit` on
3. execute a `add check foreign key` ddl
4. when ddl enter (model.StateNone)[https://github.com/pingcap/tidb/blob/4f00ece106b1f9b13086d3e683939527d91718de/pkg/ddl/foreign_key.go#L51C1-L52C1], a dml execute with the schemaVersion and commit use async commit.
5. while the commit is slow for some reason, and the ddl enter (model.StateWriteOnly)[https://github.com/pingcap/tidb/blob/4f00ece106b1f9b13086d3e683939527d91718de/pkg/ddl/foreign_key.go#L66]
6. the dml finish the prepare use a maxTx which is greate than the `checkForeignKeyConstrain` readTs
7. then the dml will be a potential bug for the check.

Is there some thing wrong above?
It is not eazy to reproduce the case, but logically it can occurrs.

right?  not right?

jiyfhust:
Related #23363.

pkg/ddl/column.go
```diff
@@ -731,6 +731,9 @@ func (w *worker) doModifyColumnTypeWithData(
 	case model.StateDeleteOnly:
 		// Column from null to not null.
 		if !mysql.HasNotNullFlag(oldCol.GetFlag()) && mysql.HasNotNullFlag(changingCol.GetFlag()) {
+			if d.lease > 0 {
+				delayForAsyncCommit()
```
YangKeao, line 735: I think it's not needed. This branch will only happen when the modifying column is implemented through creating a new column and removing the old one. When the new column is in the `DeleteOnly` state, it'll have no chance to insert `NULL` values into it :thinking:.

YangKeao (commented):
Let's merge this PR! Would you mind adding a test for it?

Ref https://github.com/pingcap/tidb/pull/56349/files, as I've tested, the test and failpoint in that PR should work.

It's also preferred to add more similar tests related to other situations (like `check constraints`, `modify column`...).

jiyfhust:
> Let's merge this issue! Would you mind adding a test for it?

I'm pleased for it，we can also work together to make this pr more complete.

mjonss:
For tests with different DDL States, I wrote tests for partitioning DDL, that have multiple domains, so one can even tests new transactions started in previous schema version, see [here](https://github.com/pingcap/tidb/pull/56082/files#diff-5be7c002803760ae62e04c0c1e92a8d7651ac6948fd691a93a62a735cb5d5137). @YangKeao, @jiyfhust That may work as inspiration for writing tests for this PR as well?

jiyfhust:
> Let's merge this PR! Would you mind adding a test for it?
> 
> Ref https://github.com/pingcap/tidb/pull/56349/files, as I've tested, the test and failpoint in that PR should work.
> 
> It's also preferred to add more similar tests related to other situations (like `check constraints`, `modify column`...).

My poor virtual host with only 4G memory has diffcuity to reproduce the case, expecially to run txn tests you added.
I have checked the code again, except for the "foreign key", we can't verify there really need `delayForAsyncCommit`(they add a schema version to handle it, but not sure can  handle all cases) , I think you can fix the 'foreign key' case as you  pull request does, and we now can left the others aside.

For others, maybe case like this(this is only for note):
```
normal 2pc and mdl off:
1. prepare
2. get commit_ts
3. if commit_ts < curruent_schema_version 
        return error

however:
async 2pc:
1. tidb node1 prepare with schema v1, not return from tikv yet, it may table a long time but before `SafeWindow` expire
2. tidb ddl owner which is node2 run ddl and schema version enter v3
3. node1 prepare success alther it is not normal now(v1->v3, the schema version is too old)
4. node1 don't need check commit_ts with schema version becasuse it is async commit
5. finally transaction committed. 
```

Issue #31648, opened 2022-01-13
`SET @@timestamp=UNIX_TIMESTAMP` does not take effect in TiDB


espresso98:
## Bug Report

The date portion from MySQL is based on TIMESTAMP but in TiDB it is based on current time.

### 1. Minimal reproduce step 

```sql
SET @@timestamp=UNIX_TIMESTAMP('2001-01-01 00:00:00');
SELECT CAST(CAST('00:00:00.123456' AS TIME(6)) + INTERVAL 1 DAY AS DATETIME(6));

DROP TABLE IF EXISTS t1;
SET @@timestamp=UNIX_TIMESTAMP('2008-01-03 10:20:30.1');
CREATE TABLE t1 (a DATETIME(6));
INSERT INTO t1 VALUES (CURRENT_TIME), (CURRENT_TIME(6));
INSERT INTO t1 VALUES (TIME'08:00:00.123456'), (TIME'240:00:00.000001');
INSERT INTO t1 VALUES (TIME'-10:00:00.000001'), (TIME'-240:00:00.000001');
SELECT * FROM t1;
```
### 2. What did you expect to see? 
```sql
mysql> SELECT CAST(CAST('00:00:00.123456' AS TIME(6)) + INTERVAL 1 DAY AS DATETIME(6));
+--------------------------------------------------------------------------+
| CAST(CAST('00:00:00.123456' AS TIME(6)) + INTERVAL 1 DAY AS DATETIME(6)) |
+--------------------------------------------------------------------------+
| 2001-01-02 00:00:00.123456                                               |
+--------------------------------------------------------------------------+

mysql> SELECT * FROM t1;
+----------------------------+
| a                          |
+----------------------------+
| 2008-01-03 10:20:30.000000 |
| 2008-01-03 10:20:30.100000 |
| 2008-01-03 08:00:00.123456 |
| 2008-01-13 00:00:00.000001 |
| 2008-01-02 13:59:59.999999 |
| 2007-12-23 23:59:59.999999 |
+----------------------------+
```

### 3. What did you see instead 

```sql
tidb> SELECT CAST(CAST('00:00:00.123456' AS TIME(6)) + INTERVAL 1 DAY AS DATETIME(6));
+--------------------------------------------------------------------------+
| CAST(CAST('00:00:00.123456' AS TIME(6)) + INTERVAL 1 DAY AS DATETIME(6)) |
+--------------------------------------------------------------------------+
| 2022-01-13 00:00:00.123456                                               |
[...]

PR #48565, opened 2023-11-13
expression, types: fix the logic of converting duration to year

### LINKED ISSUE #48557: casting `Time` to `Year` gives the wrong result

```sql
select cast(cast('22:03' as time) as year);
```

```sql
mysql> select cast(cast('22:03' as time) as year);
+-------------------------------------+
| cast(cast('22:03' as time) as year) |
+-------------------------------------+
|                                2023 |
+-------------------------------------+
1 row in set (0.00 sec)
```

```sql
mysql> select cast(cast('22:03' as time) as year);
+-------------------------------------+
| cast(cast('22:03' as time) as year) |
+-------------------------------------+
|                              220300 |
+-------------------------------------+
1 row in set (0.00 sec)
```

```
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| tidb_version()                                                                                                                                                                              |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Release Version: None
Edition: Community
Git Commit Hash: None
Git Branch: None
UTC Build Time: None
GoVersion: go1.21.3
Race Enabled: false
Store: unistore |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

YangKeao:

Issue Number: close #48557

Problem Summary:

The correct behavior of converting time to year should be mixing the `Now` and duration, and then return the year.

Actually, it has two different behaviors when converting the `TIME` to `YEAR`.

1. In DDL, e.g. modify the column type from `TIME` to `YEAR`, it follows the "concatenation" rule. The `00:19:98` will be converted to `1998`. TiDB follows this behavior in all scenario before this PR. It's more like `TIME -> CHAR() -> YEAR`. This behavior has been covered by existing tests in `ddl/column_type_change` and `ddl/modify_column`.
2. In `SELECT` (cast implicitly or explicitly) and `INSERT` (inserting a `TIME` to a `YEAR` column), it's more like `TIME -> DATETIME -> YEAR`. It will mix the duration with the current time, and return the year.

This PR makes the second one the default behavior, and use a flag to follow the first behavior in `DDL` backfilling.

Modify the logic of both `CastDurationAsInt` in expression and the `ConvertTo` in `types`.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

### Release note

```release-note
Fix the issue that converting the `TIME` to `YEAR` gives an integer which represents the string of the time, rather then the year part of mixing the `TIME` and current time.
```

### DIFF
--- a/pkg/expression/builtin_cast_vec.go
+++ b/pkg/expression/builtin_cast_vec.go
@@ -338,6 +338,7 @@ func (b *builtinCastDurationAsIntSig) vecEvalInt(ctx sessionctx.Context, input *
 	var duration types.Duration
 	ds := buf.GoDurations()
 	fsp := b.args[0].GetType().GetDecimal()
+	isYear := b.tp.GetType() == mysql.TypeYear
 	for i := 0; i < n; i++ {
 		if result.IsNull(i) {
 			continue
@@ -345,11 +346,18 @@ func (b *builtinCastDurationAsIntSig) vecEvalInt(ctx sessionctx.Context, input *
 
 		duration.Duration = ds[i]
 		duration.Fsp = fsp
-		dur, err := duration.RoundFrac(types.DefaultFsp, ctx.GetSessionVars().Location())
-		if err != nil {
-			return err
+
+		if isYear {
+			i64s[i], err = duration.ConvertToYear(ctx.GetSessionVars().StmtCtx.TypeCtx())
+		} else {
+			var dur types.Duration
+			dur, err = duration.RoundFrac(types.DefaultFsp, ctx.GetSessionVars().Location())
+			if err != nil {
+				return err
+			}
+			i64s[i], err = dur.ToNumber().ToInt()
 		}
-		i64s[i], err = dur.ToNumber().ToInt()
+
 		if err != nil {
 			return err
 		}
--- a/pkg/types/convert_test.go
+++ b/pkg/types/convert_test.go
@@ -271,7 +271,7 @@ func TestConvertType(t *testing.T) {
 	require.Equal(t, int64(2015), v)
 	v, err = Convert(ZeroDuration, ft)
 	require.NoError(t, err)
-	require.Equal(t, int64(0), v)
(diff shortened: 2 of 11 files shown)


xhebox (approved):

tangenta (approved):

pkg/expression/builtin_cast_vec.go
```diff
@@ -345,11 +345,18 @@ func (b *builtinCastDurationAsIntSig) vecEvalInt(ctx sessionctx.Context, input *
 
 		duration.Duration = ds[i]
 		duration.Fsp = fsp
-		dur, err := duration.RoundFrac(types.DefaultFsp, ctx.GetSessionVars().Location())
-		if err != nil {
-			return err
+
+		if b.tp.GetType() == mysql.TypeYear {
```
zanmato1984, line 349: I would recommend to use a local bool variable outside the for loop to do this check, so that the loop body can be tighter.

pkg/types/convert_test.go
```diff
@@ -271,7 +271,7 @@ func TestConvertType(t *testing.T) {
 	require.Equal(t, int64(2015), v)
 	v, err = Convert(ZeroDuration, ft)
 	require.NoError(t, err)
-	require.Equal(t, int64(0), v)
+	require.Equal(t, int64(time.Now().Year()), v)
```
zanmato1984, line 274: This test may fail when line 272 executes in the past year than this line. But it's ok :)

zanmato1984 (approved):
LGTM

pkg/types/convert_test.go
YangKeao: Yes. Once TiDB fixes https://github.com/pingcap/tidb/issues/31648, we can make these tests stabler.

YangKeao:
/retest

Yellow timeout :facepalm:

PR #48583, opened 2023-11-14
*: reuse chunk in the histogram

### LINKED ISSUE #48584: reuse memory in the histogram
## Enhancement

In our current system, memory utilization needs to be optimized for higher efficiency. Specifically, we have observed potential room for improvement in the memory usage within the histogram component of our system.

The challenge is to reuse memory in the histogram, which could significantly enhance our system performance and reduce unnecessary memory allocation and deallocation, leading to less memory fragmentation.

hawkingrei:

Issue Number: close #48584

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [x] Integration test

analyze a table with 292 columns.

before:

spend 2517.959s

image

Allocate Speed: 500MB/s

after:

spend 2422.801s

image

Allocate Speed: ~400MB/s

- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/util/chunk/pool.go
+++ b/pkg/util/chunk/pool.go
@@ -34,7 +34,6 @@ func getChunkFromPool(initCap int, fields []*types.FieldType) *Chunk {
 	pool, ok := globalChunkPool[initCap]
 	globalChunkPoolMutex.RUnlock()
 	if ok {
-
 		return pool.GetChunk(fields)
 	}
 	globalChunkPoolMutex.Lock()
--- a/pkg/util/chunk/pool.go
+++ b/pkg/util/chunk/pool.go
@@ -30,11 +30,11 @@ var (
 func getChunkFromPool(initCap int, fields []*types.FieldType) *Chunk {
 	globalChunkPoolMutex.RLock()
 	pool, ok := globalChunkPool[initCap]
+	globalChunkPoolMutex.RUnlock()
 	if ok {
-		globalChunkPoolMutex.RUnlock()
+
 		return pool.GetChunk(fields)
 	}
-	globalChunkPoolMutex.RUnlock()
 	globalChunkPoolMutex.Lock()
 	defer globalChunkPoolMutex.Unlock()
 	globalChunkPool[initCap] = NewPool(initCap)
@@ -44,12 +44,11 @@ func getChunkFromPool(initCap int, fields []*types.FieldType) *Chunk {
 func putChunkFromPool(initCap int, fields []*types.FieldType, chk *Chunk) {
 	globalChunkPoolMutex.RLock()
 	pool, ok := globalChunkPool[initCap]
+	globalChunkPoolMutex.RUnlock()
 	if ok {
-		globalChunkPoolMutex.RUnlock()
 		pool.PutChunk(fields, chk)
 		return
 	}
-	globalChunkPoolMutex.RUnlock()
 	globalChunkPoolMutex.Lock()
 	defer globalChunkPoolMutex.Unlock()
 	globalChunkPool[initCap] = NewPool(initCap)
--- a/pkg/util/chunk/pool.go
+++ b/pkg/util/chunk/pool.go
@@ -18,10 +18,11 @@ import (
 	"sync"
 
 	"github.com/pingcap/tidb/pkg/types"
+	"github.com/pingcap/tidb/pkg/util/syncutil"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


pkg/util/chunk/pool.go
```diff
@@ -18,8 +18,44 @@ import (
...
+var (
+	globalChunkPoolMutex syncutil.RWMutex
+	// globalChunkPool is a chunk pool, the key is the init capacity.
+	globalChunkPool = make(map[int]*Pool)
+)
+
+func getChunkFromPool(initCap int, fields []*types.FieldType) *Chunk {
+	globalChunkPoolMutex.RLock()
+	pool, ok := globalChunkPool[initCap]
+	if ok {
```
elsa0520, line 33:
```suggestion
    globalChunkPoolMutex.RUnlock()
	if ok {
			return pool.GetChunk(fields)
	}
```

pkg/util/chunk/pool.go
```diff
@@ -18,8 +18,44 @@ import (
...
+	defer globalChunkPoolMutex.Unlock()
+	globalChunkPool[initCap] = NewPool(initCap)
+	return globalChunkPool[initCap].GetChunk(fields)
+}
+
+func putChunkFromPool(initCap int, fields []*types.FieldType, chk *Chunk) {
+	globalChunkPoolMutex.RLock()
+	pool, ok := globalChunkPool[initCap]
+	if ok {
+		globalChunkPoolMutex.RUnlock()
```
elsa0520, line 48: Same as above

pkg/util/chunk/pool.go
```diff
@@ -18,8 +18,44 @@ import (
...
 )
 
+var (
+	globalChunkPoolMutex syncutil.RWMutex
+	// globalChunkPool is a chunk pool, the key is the init capacity.
+	globalChunkPool = make(map[int]*Pool)
+)
+
+func getChunkFromPool(initCap int, fields []*types.FieldType) *Chunk {
+	globalChunkPoolMutex.RLock()
```
elsa0520, line 31: Compare with before, is there any performance issue for the destroy and get ?

pkg/util/chunk/pool.go
```diff
@@ -18,8 +18,44 @@ import (
...
+	pool, ok := globalChunkPool[initCap]
+	if ok {
+		globalChunkPoolMutex.RUnlock()
+		return pool.GetChunk(fields)
+	}
+	globalChunkPoolMutex.RUnlock()
+	globalChunkPoolMutex.Lock()
+	defer globalChunkPoolMutex.Unlock()
+	globalChunkPool[initCap] = NewPool(initCap)
+	return globalChunkPool[initCap].GetChunk(fields)
```
AilinKid, line 41: is that we need exactly a matched capacity-size pool, a little bit less-than is adequate?

AilinKid (commented):
Rest LGTM

pkg/util/chunk/pool.go
hawkingrei: Fixed.

pkg/util/chunk/pool.go
hawkingrei: In fact, as long as we don't encounter a new size, we are dealing with a read mutex. But since there are only so many sizes, it would only impact at startup. After that, it should all be read mutex.

hawkingrei: Additionally, if we really encounter performance issues,  this part can be turned into a shardmap. However, this would increase the code complexity here and affect the review process. We can consider this for subsequent optimizations.

hawkingrei: BTW， this initCap is the size of bucket in the histogram. I think it will not have too many difference values.

pkg/util/chunk/pool.go
hawkingrei: this initCap is the size of bucket in the histogram. I think it can meet the requirement.

hawkingrei: I add some comments for it.

elsa0520 (approved):
LGTM

pkg/util/chunk/pool.go
hawkingrei: Done

AilinKid (approved):

PR #48587, opened 2023-11-14
br/storage: enable async prefetch data

lance6716:

Issue Number: close https://github.com/pingcap/tidb/issues/48781

Problem Summary:

offload the network reading from the main goroutine of merge iterator

image

Still need to improve the performance in future.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

use this command to run the specific "benchmark" unit test

```
go test ./br/pkg/lightning/backend/external -v --tags=intest -test.run TestCompareReader --testing-storage-uri "xxx"
```

On master with ks3

```
    bench_test.go:414: merge iter read speed for 1258291200 bytes: 109.49 MB/s
```

This PR with ks3

```
    bench_test.go:412: merge iter read speed for 1258291200 bytes: 121.49 MB/s
```

This PR with memstore

```
    bench_test.go:412: merge iter read speed for 1258291200 bytes: 163.92 MB/s
```

And I checked the heap usage, I have set 64MB in the unit test, the heap does not exceed it.

image

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/util/prefetch/reader.go
+++ b/pkg/util/prefetch/reader.go
@@ -22,8 +22,8 @@ func NewPrefetchReader(r io.ReadCloser, prefetchSize int) io.ReadCloser {
 		bufCh: make(chan []byte, 0),
 		err:   nil,
 	}
-	ret.buf[0] = make([]byte, prefetchSize)
-	ret.buf[1] = make([]byte, prefetchSize)
+	ret.buf[0] = make([]byte, prefetchSize/2)
+	ret.buf[1] = make([]byte, prefetchSize/2)
 	go ret.run()
 	return ret
 }
--- a/pkg/util/prefetch/reader.go
+++ b/pkg/util/prefetch/reader.go
@@ -1,4 +1,16 @@
-// Copyright 2023 PingCAP, Inc. Licensed under Apache-2.0.
+// Copyright 2023 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 
 package prefetch
 
--- a/pkg/util/prefetch/reader.go
+++ b/pkg/util/prefetch/reader.go
@@ -17,6 +17,7 @@ package prefetch
 import (
 	"bytes"
 	"io"
+	"sync"
 )
 
 // Reader is a reader that prefetches data from the underlying reader.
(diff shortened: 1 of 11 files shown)


pkg/util/prefetch/reader.go
```diff
@@ -0,0 +1,94 @@
...
+	return ret
+}
+
+func (r *Reader) run() {
+	for {
+		r.bufIdx = (r.bufIdx + 1) % 2
+		buf := r.buf[r.bufIdx]
+		n, err := r.r.Read(buf)
+		buf = buf[:n]
+		r.bufCh <- buf
```
D3Hunter, line 51: might block forever if we call close before read

and we should wait this routine exist on close

D3Hunter (commented):
rest lgtm

pkg/util/prefetch/reader.go
lance6716: https://github.com/pingcap/tidb/pull/48587/commits/b407fbf785e7b35fbfce042c8251176f34f42dc1

pkg/util/prefetch/reader.go
```diff
@@ -88,7 +100,12 @@ func (r *Reader) Read(data []byte) (int, error) {
 	}
 }
 
-// Close implements io.Closer.
+// Close implements io.Closer. Close should not be called concurrently with Read.
 func (r *Reader) Close() error {
-	return r.r.Close()
+	ret := r.r.Close()
+	r.closeOnce.Do(func() {
+		close(r.closed)
```
D3Hunter, line 107: we don't close reader twice, seems no need this closeOnce

pkg/util/prefetch/reader.go
```diff
@@ -88,7 +100,12 @@ func (r *Reader) Read(data []byte) (int, error) {
 	}
 }
 
-// Close implements io.Closer.
+// Close implements io.Closer. Close should not be called concurrently with Read.
```
D3Hunter, line 103: seems most io.Reader is thread unsafe

pkg/util/prefetch/reader.go
```diff
@@ -88,7 +100,12 @@ func (r *Reader) Read(data []byte) (int, error) {
...
 func (r *Reader) Close() error {
-	return r.r.Close()
+	ret := r.r.Close()
```
D3Hunter, line 105: not sure if the undering reader is safe to call `close/read` concurrently

pkg/util/prefetch/reader.go
lance6716: I simply checked s3.go. In some cases s3ObjectReader will be called by line 893 and line 911

https://github.com/pingcap/tidb/blob/3543275dcf4b6454eb874c1362c87d31a963da6d/br/pkg/storage/s3.go#L872-L905

So I think we should tolerate the caller call Close() for multiple times. But all reader methods are not thread safe, I will change it to a bool rather than sync.Once

pkg/util/prefetch/reader.go
lance6716: io.Reader interface does not limit the invocation, for example, the reader of io.Pipe is thread safe. So I want to notice the caller about safety.

pkg/util/prefetch/reader.go
lance6716: I think we can't assume it so I added the comment

pkg/util/prefetch/reader_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	source = bytes.NewReader([]byte("01234567890"))
+	r = NewReader(io.NopCloser(source), 3)
+	buf = make([]byte, 11)
+	n, err = r.Read(buf)
+	require.NoError(t, err)
+	require.EqualValues(t, 11, n)
+	_, err = r.Read(buf)
+	require.ErrorIs(t, err, io.EOF)
+}
```
ywqzzy, line 66: Add a case that len(read buf) > len(source []byte)

lance6716: https://github.com/pingcap/tidb/pull/48587/commits/d731f28df8f68847813f7d2c1bd8ed9ef76048e4

D3Hunter (approved):

3pointer (approved):

PR #48626, opened 2023-11-15
build: don't run linter on excluded files

### LINKED ISSUE #48624: skip linter for generated files and external libraries
Running `make bazel_build` on a local machine requires more than 40GB ram (I have tested it through `systemd-run --scope -p MemoryMax=40G ...`, and it gets OOM killed). We could try to eliminate a lot of RAM and CPU usage by skipping some big files.

Though we add `_generated.go` to the `exclude_files` in `nogo_config.json`, the linters will still actually run on these files. See https://github.com/bazelbuild/rules_go/blob/master/go/tools/builders/nogo_main.go#L438 .

Still nogo doesn't provide such function, we can modify the linters to achieve our goal, for example in `gofmt`:

```go
for _, f := range fileNames {
  diff, err := gofmt.Run(f, needSimplify)
  if err != nil {
	  return nil, fmt.Errorf("could not run gofmt: %w (%s)", err, f)
  }
...
```

We can add a line to check the filename and skip:

```go
for _, f := range fileNames {
  if strings.Contains(f, "_generated.go") {
    continue;
  }

  diff, err := gofmt.Run(f, needSimplify)
  if err != nil {
	  return nil, fmt.Errorf("could not run gofmt: %w (%s)", err, f)
  }
...
```

(It's just a PoC. We can actually re-use the `nogo_config.json` and use the regex in it to check.)

YangKeao:

Issue Number: close #48624 

Problem Summary:

As I have tested, this PR makes it possible to run `make bazel_build` within 16GB ram (without cache). 

Previously, it'll need more than 40GB ram. As I only have 48GB ram for my development environment and the `bazel` processes are always OOM-killed, I'm not sure how much it'll actually take.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

### Release note

```release-note
None
```

### DIFF
--- a/build/BUILD.bazel
+++ b/build/BUILD.bazel
@@ -1,6 +1,6 @@
 package(default_visibility = ["//visibility:public"])
 
-load("@io_bazel_rules_go//go:def.bzl", "nogo")
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "nogo")
 load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
 load("//build/linter/staticcheck:def.bzl", "staticcheck_analyzers")
 
@@ -180,4 +180,14 @@ nogo(
                ],
                "//conditions:default": [],
            }),
+)
+
+go_library(
+    name = "build",
+    srcs = ["config.go"],
+    embedsrcs = [
+        "nogo_config.json",
+    ],
+    importpath = "github.com/pingcap/tidb/build",
+    visibility = ["//visibility:public"],
 )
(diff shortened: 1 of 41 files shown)


hawkingrei (approved):

hawkingrei:
@YangKeao I think you need to add this code to ```build/linter/util/util.go:104```

hawkingrei:
> @YangKeao I think you need to add this code to `build/linter/util/util.go:104`

You directly edit the analysis.Pass.Files to delete skip files.

lance6716:
lgtm, but I want to ask do you know where the linter framework generates *analysis.Pass? maybe it can pre-filter files based on every linter names.

xhebox (approved):

PR #48653, opened 2023-11-17
pkg: stop using math/rand.Seed

szpnygo:

Issue Number: https://github.com/golang/go/issues/54880

Problem Summary:

```golang


// Seed uses the provided seed value to initialize the default Source to a
// deterministic state. Seed values that have the same remainder when
// divided by 2³¹-1 generate the same pseudo-random sequence.
// Seed, unlike the Rand.Seed method, is safe for concurrent use.
//
// If Seed is not called, the generator is seeded randomly at program startup.
//
// Prior to Go 1.20, the generator was seeded like Seed(1) at program startup.
// To force the old behavior, call Seed(1) at program startup.
// Alternately, set GODEBUG=randautoseed=0 in the environment
// before making any calls to functions in this package.
//
// Deprecated: As of Go 1.20 there is no reason to call Seed with
// a random value. Programs that call Seed with a known value to get
// a specific sequence of results should use New(NewSource(seed)) to
// obtain a local random generator.
func Seed(seed int64) {

```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/lightning/mydump/loader_test.go
+++ b/br/pkg/lightning/mydump/loader_test.go
@@ -23,7 +23,6 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
-	"time"
 
 	"github.com/pingcap/tidb/br/pkg/lightning/common"
 	"github.com/pingcap/tidb/br/pkg/lightning/config"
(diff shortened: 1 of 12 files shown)


br/pkg/lightning/mydump/loader_test.go
```diff
@@ -1129,9 +1128,6 @@ func TestSampleParquetDataSize(t *testing.T) {
 	pwriter.RowGroupSize = 128 * 1024 * 1024 //128M
 	pwriter.PageSize = 8 * 1024              //8K
 	pwriter.CompressionType = parquet.CompressionCodec_SNAPPY
-	seed := time.Now().Unix()
-	t.Logf("seed: %d", seed)
```
lance6716, line 1133: We need to print the current seed, so when the unit tests fails we can set to that specific seed and reproduce it. Can you keep the printing here? like

```
t.Logf("seed: %d. To reproduce the random behaviour, manually set `rand.Seed(seed)`", seed)
```

lance6716 (approved):

hawkingrei (approved):

Issue #35319, opened 2022-06-13
Unified the PD API request interface


JmPotato:
As https://github.com/pingcap/tidb/issues/35268 and https://github.com/pingcap/tidb/issues/35284 show, the PD API request lacks a unified interface, which causes the failover, error handling, and other necessary mechanisms to be missed. So I proposed that we should provide a unified PD API request interface inside TiDB rather than maintaining different implementations everywhere.

lhy1024:
etcd too?

PR #48738, opened 2023-11-21
infosync: integrate PD HTTP client into the label manager

JmPotato:

Issue Number: ref #35319, https://github.com/tikv/pd/issues/7300

Problem Summary:

Integrate PD HTTP client into the label manager.

- Replace the etcd client inside `LabelRuleManager` with PD HTTP client.
- Reuse the type definitions within the PD SDK as much as possible.

### Check List

Tests 

- [x] Unit test
- [x] Integration test

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/label/rule.go
+++ b/pkg/ddl/label/rule.go
@@ -23,6 +23,7 @@ import (
 	"github.com/pingcap/tidb/pkg/parser/ast"
 	"github.com/pingcap/tidb/pkg/tablecodec"
 	"github.com/pingcap/tidb/pkg/util/codec"
+	pd "github.com/tikv/pd/client/http"
 	"gopkg.in/yaml.v2"
 )
 
@@ -53,13 +54,7 @@ var (
 )
 
 // Rule is used to establish the relationship between labels and a key range.
-type Rule struct {
-	ID       string        `json:"id"`
-	Index    int           `json:"index"`
-	Labels   Labels        `json:"labels"`
-	RuleType string        `json:"rule_type"`
-	Data     []interface{} `json:"data"`
-}
+type Rule pd.LabelRule
 
 // NewRule creates a rule.
 func NewRule() *Rule {
@@ -69,7 +64,7 @@ func NewRule() *Rule {
 // ApplyAttributesSpec will transfer attributes defined in AttributesSpec to the labels.
 func (r *Rule) ApplyAttributesSpec(spec *ast.AttributesSpec) error {
 	if spec.Default {
-		r.Labels = []Label{}
+		r.Labels = []pd.RegionLabel{}
 		return nil
 	}
 	// construct a string list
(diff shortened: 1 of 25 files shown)


pkg/server/handler/tests/http_handler_test.go
```diff
@@ -449,26 +450,22 @@ func TestBinlogRecover(t *testing.T) {
...
 func (ts *basicHTTPHandlerTestSuite) startServer(t *testing.T) {
 	var err error
-	ts.Port = uint(rand.Int31n(50000)) + 10000
-	ts.StatusPort = ts.Port + 1
-	ts.store, err = mockstore.NewMockStore(
-		mockstore.WithTiKVOptions(
-			tikv.WithPDHTTPClient([]string{ts.Addr()}),
-		),
-	)
+	ts.store, err = mockstore.NewMockStore()
```
JmPotato, line 453: The previous commit made changes to this code section, but it was later determined to be unnecessary during testing. This PR reverts those changes.

hawkingrei (approved):

JmPotato:
PTAL

/cc @tangenta @zimulala

pkg/domain/infosync/info.go
```diff
@@ -209,7 +209,10 @@ func GlobalInfoSyncerInit(
 	if err != nil {
 		return nil, err
 	}
-	is.labelRuleManager = initLabelRuleManager(etcdCli)
+	if pdHTTPCli != nil {
+		pdHTTPCli = pdHTTPCli.WithRespHandler(pdResponseHandler)
```
zimulala, line 213: Could we put it to `initLabelRuleManager`? It seems we use `pdHTTPCli` only in `initLabelRuleManager`.

JmPotato: Later we will integrate the PD HTTP client into the Placement Rule manager below also, so I keep it initialized here.

pkg/ddl/label/rule.go
```diff
@@ -157,16 +153,14 @@ func (r *Rule) Reset(dbName, tableName, partName string, ids ...int64) *Rule {
...
-}
-
 // NewRulePatch returns a patch of rules which need to be set or deleted.
-func NewRulePatch(setRules []*Rule, deleteRules []string) *RulePatch {
-	return &RulePatch{
-		SetRules:    setRules,
+func NewRulePatch(setRules []*Rule, deleteRules []string) *pd.LabelRulePatch {
+	labelRules := make([]*pd.LabelRule, 0, len(setRules))
+	for _, rule := range setRules {
+		labelRules = append(labelRules, (*pd.LabelRule)(rule))
```
zimulala, line 160: Does changing the type of `setRules`(line157) to `pd.LabelRule` also eliminate the need for conversions?

JmPotato: The main consideration here is that TiDB has made additional extensions to the SQL layer concept for the `Rule` type internally. Therefore, I decided to keep the relevant implementation of the `Rule` type in TiDB and hide the type conversion in the lower layer from the caller. Otherwise, we would have needed to completely replace `Rule` with `pd.LabelRule`, or constantly introduce type conversions throughout. However, prioritizing code readability over a slight performance sacrifice seems like a better choice.

zimulala (approved):
LGTM

PR #48775, opened 2023-11-22
planner: fix possible inconsistent output cols among union's children

### LINKED ISSUE #48755: UNION ALL might return error when the first child is table dual
## Bug Report

```
create table t(a int, b int);
set @@tidb_max_chunk_size=32;
select a from (select 100 as a, 100 as b union all select * from t) t where b != 0;
```
make sure that the number of satisfied rows from t is more than 32.

Return the correct result.

Error thrown `Internal error: UnionExec chunk column count mismatch, req: 2, result: 1`

winoros:

Issue Number: close #48755 

Problem Summary:

When UNIONALL's first child, it's possible that the selection is pushed at a position different with other children. This will cause inconsistent output columns among its children.

You can check the changed codes together with https://github.com/pingcap/tidb/blob/master/pkg/planner/core/rule_predicate_push_down.go#L479-L483 for more details.

And the failed case is something like `Proj{c1}->UnionAll{c1, c2, c3}->(Select{c2 > 0, c3 > 0}->Proj{c1, c2, c3}, Proj{c1, c2, c3}->Select{c2>0, c3>0})`.
UnionALL first only keeps the col `c1`. But its first child's selection also requires `c2` and `c3`, so the projection of its first children will keep `c2` and `c3`. The second child's projection doesn't need the `c2` and `c3` since the selection is pushed down.

We use the UNIONALL's self output columns to do the alignment.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  >   

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the potential execution failure when UNION ALL's first child is TableDual
```


elsa0520 (approved):
LGTM

pkg/planner/core/rule_column_pruning.go
```diff
@@ -284,16 +287,14 @@ func (p *LogicalUnionAll) PruneColumns(parentUsedCols []*expression.Column, opt
...
-				prunedColumns = append(prunedColumns, p.schema.Columns[i])
-				p.schema.Columns = append(p.schema.Columns[:i], p.schema.Columns[i+1:]...)
-			}
+	for i := len(used) - 1; i >= 0; i-- {
+		if !used[i] {
+			prunedColumns = append(prunedColumns, p.schema.Columns[i])
+			p.schema.Columns = append(p.schema.Columns[:i], p.schema.Columns[i+1:]...)
 		}
-		appendColumnPruneTraceStep(p, prunedColumns, opt)
+	}
```
time-and-fate, line 295: Now we can use `slices.Delete` or `slices.DeleteFunc` to make it cleaner.

time-and-fate (approved):

AilinKid (approved):

pkg/planner/core/rule_column_pruning.go
```diff
@@ -284,16 +288,14 @@ func (p *LogicalUnionAll) PruneColumns(parentUsedCols []*expression.Column, opt
...
-		used := expression.GetUsedList(p.SCtx(), p.children[0].Schema().Columns, p.schema)
-		for i := len(used) - 1; i >= 0; i-- {
-			if !used[i] {
-				prunedColumns = append(prunedColumns, p.schema.Columns[i])
-				p.schema.Columns = append(p.schema.Columns[:i], p.schema.Columns[i+1:]...)
-			}
+	for i := len(used) - 1; i >= 0; i-- {
+		if !used[i] {
+			prunedColumns = append(prunedColumns, p.schema.Columns[i])
+			slices.Delete(p.schema.Columns, i, i)
```
time-and-fate, line 294: You need to write
`p.schema.Columns = slices.Delete(p.schema.Columns, i, i)`.

time-and-fate: This is Golang :rofl:

PR #48902, opened 2023-11-26
*: improve proformance for BuildHistAndTopN

hawkingrei:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/statistics/builder.go
```diff
@@ -287,97 +294,97 @@ func BuildHistAndTopN(
...
 
 	topNList = pruneTopNItem(topNList, ndv, nullCount, sampleNum, count)
-
+	samples = make([]*SampleItem, 0, sampleTree.Len())
+	for sampleTree.Len() != 0 {
+		item, _ := sampleTree.DeleteMin()
+		samples = append(samples, item.Sample)
+	}
 	// Step2: exclude topn from samples
 	for i := int64(0); i < int64(len(samples)); i++ {
```
ngaut, line 389: Key Changes:
Introduced a Map for Efficient Lookup: A map topNMap is created to store the topNList items. This allows for O(1) lookup time to check if a sample is in the topN list.

Eliminated the Inner Loop: The inner loop that iterated over topNList for each sample is removed. Instead, the map is used for checking the presence of a sample in the topN list.

Preserving Samples Not in TopN: A new slice newSamples is created to hold samples that are not in the topN list. This avoids modifying the original samples slice while iterating over it.

This optimization should significantly reduce the time complexity, particularly when dealing with a large number of samples and a sizable topN list. However, it's important to note that optimizations should always be validated with real-world data and use cases to ensure they provide the expected benefits.


// Part of BuildHistAndTopN function with optimization
topNMap := make(map[string]uint64) // Create a map to hold topN items for quick lookup
for _, topN := range topNList {
    topNMap[string(topN.Encoded)] = topN.Count
}

newSamples := make([]*SampleItem, 0, len(samples))
for _, sample := range samples {
    sampleBytes, err := getComparedBytes(sample.Value)
    if err != nil {
        return nil, nil, errors.Trace(err)
    }

    if count, exists := topNMap[string(sampleBytes)]; exists {
        // If found in topN, skip the next 'count' samples
        continue
    }
    newSamples = append(newSamples, sample) // Keep the sample if not in topN
}

samples = newSamples // Replace old samples with the filtered ones

hawkingrei: I think If NDV is high, we can merge sampling by hashmap. so it can avoid too much compare when to sort. I need to add more benchmark to verify it. but if NDV is low, it is unavailable.

time-and-fate (commented):
1. Please at least keep the comments valid and up-to-date after you copied them from elsewhere.
2. If you need to copy ~80% of an existing function, I suggest we try to find some methods to avoid the large duplication.
3. From your benchmark result, I think we are just trading space for time instead of a pure improvement. So what's the reason behind this change, and have we checked the risk?

hawkingrei:
> 1. Please at least keep the comments valid and up-to-date after you copied them from elsewhere.
> 
>     2. If you need to copy ~80% of an existing function, I suggest we try to find some methods to avoid the large duplication.
> 
>     3. From your benchmark result, I think we are trading space for time instead of a pure improvement. So what's the reason behind this change, and have we checked the risk?

I have done the refactor to merge the same code and reduce memory allocate

Now, the new one is better than the old one.

time-and-fate:
I think there is little performance difference in the new benchmark result.

pkg/statistics/builder.go
```diff
@@ -242,6 +354,7 @@ func BuildHistAndTopN(
 	isColumn bool,
 	memTracker *memory.Tracker,
 	needExtStats bool,
+	highNDVMode bool,
```
xhebox, line 357:
```suggestion
	lowNDVMode bool,
```

Why it is high NDV?

Issue #31880, opened 2022-01-21
`make ut` build the test binary too slow


tiancaiamao:
## Enhancement

It use this pattern:

```
for _, pkg := range pkgs {
    cd pkg;
    go test -c -o pkg.test
}
```

The test binary is built one by one...
If we can 

```
go test -c ./...
```

It will be much faster

tiancaiamao:
A enhancement for #30822

PR #48985, opened 2023-11-28
tools/check: increase the parallelism for building test binaries

bb7133:

Issue Number: ref #31880

Problem Summary:

Increase the parallelism when building the test binaries, by default it is the number of cores(https://pkg.go.dev/cmd/go#hdr-Compile_packages_and_dependencies): 

> 
>-p n
>	the number of programs, such as build commands or
>	test binaries, that can be run in parallel.
>	The default is GOMAXPROCS, normally the number of CPUs available.
> 

I found that when running the 'building binaries stage' of `make ut`, the CPU usage is not high with the default value. So I think it's a better idea to increase the parallelism. In this PR, I doubled it.

The result of `make ut` time in my Macbook M2 Pro:

(Before this PR)
```
bash> make ut
...
building task finish, maxproc=10, count=4822, takes=12m32.547299958s 
...
```
(After this PR)
```
bash> make ut
...
building task finish, parallelism=20, count=4822, takes=6m45.4092545s
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > This is a change of the test framework, it does not change any behavior of the release, and it is trivial enough.

Side effects

- None

Documentation

- None

### Release note

```release-note
None
```

### DIFF
--- a/tools/check/ut.go
+++ b/tools/check/ut.go
@@ -29,6 +29,7 @@ import (
 	"regexp"
 	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -97,6 +98,7 @@ func (t *task) String() string {
 }
 
 var p int
+var buildParallel int
 var workDir string
 
 func cmdList(args ...string) bool {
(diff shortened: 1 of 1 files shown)


tiancaiamao (approved):

tools/check/ut.go
```diff
@@ -448,6 +450,8 @@ func main() {
 
 	// Get the correct count of CPU if it's in docker.
 	p = runtime.GOMAXPROCS(0)
```
wuhuizuo, line 452: Does it work correctly in container environments?

bb7133: Actually, I'm not sure, but as the old code it should be fine.

Defined2014 (approved):

hawkingrei (approved):

PR #49018, opened 2023-11-29
design: Priority Queue for Auto Analyze

hi-rustin:

Issue Number: None

Problem Summary:

Added design doc: `Priority Queue for Auto Analyze`.

[Rendered](https://github.com/hi-rustin/tidb/blob/rustin-patch-design/docs/design/2023-11-29-priority-queue-for-auto-analyze.md)

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > - [x] I checked and no code files have been changed.
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


docs/design/2023-11-29-priority-queue-for-auto-analyze.md
```diff
@@ -0,0 +1,270 @@
...
+|----------------------|-------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Percentage of Change | The percentage of change since the last analysis.                       | 50% <= x < 70%: 1 <br/> 70% <= x < 80%: 1.3 <br/> x >= 80%: 1.5  <br/>                                                                                                                                                                                                                                                                                            |
```
hawkingrei, line 90: Besides, some other things can also be taken into consideration.

1. Is it partition table?

partition table will analyze every partition and do ```merge```. so it will take more time.

3. NDV

after this [PR](https://github.com/pingcap/tidb/pull/48902), the low NDV table will analyze faster than the high NDV table.

hi-rustin: > 1. Is it partition table?

During the meeting, we discussed this indicator that is related to the size of a table. However, we don't want to focus solely on the size of the table because even small tables can lead to some really bad plans for certain join queries. Instead, we want to focus on the `Percentage of Change` of the table. Even if we give a partition table a higher priority, it doesn't solve the issue of resource starvation. To address this problem, we need to support the execution of analysis tasks on different nodes. If we continue to execute tasks on a single node, even with complicated priority rules, we will still face limitations due to the lack of resources to execute them one by one.

hi-rustin: > 3\. NDV

I don't understand. Could you please explain why it's important?

docs/design/2023-11-29-priority-queue-for-auto-analyze.md
```diff
@@ -0,0 +1,270 @@
...
+| Last Failed Interval | The interval between now and the last failed analysis ends time.        | Here we cannot easily determine what interval is long and what interval is short. Therefore, we can decide by querying the interval between each automatic analysis in history. <br/> interval < average automatic analysis interval: 0 <br/> interval >=  average automatic analysis interval: 0.5 <br/> interval >=  2 * average automatic analysis interval: 1 |
+| Special Event        | For example, the table has a new index but it hasn't been analyzed yet. | HasNewIndexWithoutStats: 1.5 <br/> TableNeverBeenAnalyzed: 1.5                                                                                                                                                                                                                                                                                                    |
+
+The calculation formula is: `priority_score = (change_percentage[size] * last_failed_time_weight[interval] * special_event[event])`
```
qw4990, line 94: So we calculate the score of each table by using this formula each time when triggering auto-analyze, and pick the table with the highest score to analyze?

qw4990: Just want to remind that don't forget to log this formula and scores each time when triggering auto-analyze.

qw4990 (approved):
LGTM

docs/design/2023-11-29-priority-queue-for-auto-analyze.md
```diff
@@ -0,0 +1,270 @@
...
+
+The above process is the current auto analyze process. We can see that the current auto analyze process is a random selection algorithm. We need to design a priority queue to solve the above problems.
+
+Since we perform analysis tasks synchronously on a single point for each table, we need to carry out weighted sorting on the tables that need analysis.
+
+   1. We still use tidb_auto_analyze_ratio to determine if the table needs to be analyzed. The default value is 0.5. It means that auto-analyze is triggered when greater than 50% of the rows in a table have been modified.
+   2. For special events, it is necessary to put the table into the priority queue. For example, the table has a new index but it hasn't been analyzed yet.
+   3. If a table is added to the auto-analysis queue, a weight must be assigned to it to determine the execution order.
+
+Weight table:
```
time-and-fate, line 87: For the weight table and the formula below, they are what the actual code would be like, instead of just for an example, right?

docs/design/2023-11-29-priority-queue-for-auto-analyze.md
hi-rustin: > So we calculate the score of each table by using this formula each time when triggering auto-analyze, and pick the table with the highest score to analyze?

Yes. 



> Just want to remind that don't forget to log this formula and scores each time when triggering auto-analyze.

I will add it.

docs/design/2023-11-29-priority-queue-for-auto-analyze.md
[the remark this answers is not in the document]
hi-rustin: > * Should these numbers be adjusted accordingly when `tidb_auto_analyze_ratio` changes?
> * Just curious, why do we choose to define 3 fixed weights instead of something like a linear function?

I used a fixed number because I wanted to start with a simple solution. We can always improve it later after testing. Perhaps using a linear function would be a better solution, but we can change it based on the results of our tests.

> 3\. For tables that are never analyzed, the percentage is 100%, right?

I want to consider this as a special event. However, it seems reasonable to assume that its percentage of change is 100% because all the data is new.

docs/design/2023-11-29-priority-queue-for-auto-analyze.md
hi-rustin: Yes.

docs/design/2023-11-29-priority-queue-for-auto-analyze.md
```diff
@@ -0,0 +1,270 @@
...
+| Percentage of Change | The percentage of change since the last analysis.                       | 50% <= x < 70%: 1 <br/> 70% <= x < 80%: 1.3 <br/> x >= 80%: 1.5  <br/>                                                                                                                                                                                                                                                                                            |
+| Last Failed Interval | The interval between now and the last failed analysis ends time.        | Here we cannot easily determine what interval is long and what interval is short. Therefore, we can decide by querying the interval between each automatic analysis in history. <br/> interval < average automatic analysis interval: 0 <br/> interval >=  average automatic analysis interval: 0.5 <br/> interval >=  2 * average automatic analysis interval: 1 |
```
time-and-fate, line 91: If we fail to get any of the two values, should we use `1`?

docs/design/2023-11-29-priority-queue-for-auto-analyze.md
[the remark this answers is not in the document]
time-and-fate: > I used a fixed number because I wanted to start with a simple solution. We can always improve it later after testing. Perhaps using a linear function would be a better solution, but we can change it based on the results of our tests.

Got it. I have no problem with this.
So, do we need a number for `x < 50%` when the user changes `tidb_auto_analyze_ratio` to a smaller values?

> I want to consider this as a special event.

I think, according to the formula, even if it's a "Special Event", it still needs a `change_percentage` score to calculate the total score.

PR #49022, opened 2023-11-29
planner: Adjust index scan estimate for order with limit

### LINKED ISSUE #50237: Improve estimated rows for LIMIT query with filtering separate from ordering
## Enhancement
The current row estimation is too optimistic for queries with WHERE clause filtering is against columns not in the ordering index.

Given the following example of a 1 million row table. The query estimate is 9.85 rows, and the actual is 915,000.

```
tidb> show create table 
+-------+---------------------------------------------------+
| Table | Create Table                                      |
+-------+---------------------------------------------------+
| t1    | CREATE TABLE `t1` (
  `a` int(11) DEFAULT NULL,
  `b` int(11) DEFAULT NULL,
  KEY `ia` (`a`),
  KEY `ib` (`b`)
+-----------------------------------------------------------+

tidb> explain analyze format='brief' select * from t1 where b > 900000 order by a limit 1;
+--------------------------------+---------+---------+-----------+-----------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+-----------------------+---------+------+
| id                             | estRows | actRows | task      | access object         | execution info                                                                                                                                                                                                                                                                                                                                                        | operator info         | memory  | disk |
+--------------------------------+---------+---------+-----------+-----------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+-----------------------+---------+------+
| Limit                          | 1.00    | 1       | root      |                       | time:3.45s, loops:2, RU:1349.521569                                                                                                                                                                                                                                                                                                                                   | offset:0, count:1     | N/A     | N/A  |
| └─Projection                   | 1.00    | 1       | root      |                       | time:3.45s, loops:1, Concurrency:OFF                                                                                                                                                                                                                                                                                                                                  | test.t1.a, test.t1.b  | 1.11 KB | N/A  |
|   └─IndexLookUp                | 1.00    | 1       | root      |                       | time:3.45s, loops:1, index_task: {total_time: 3.44s, fetch_handle: 3.39s, build: 52.3ms, wait: 203µs}, table_task: {total_time: 267.3ms, num: 59, concurrency: 5}, next: {wait_index: 3.29s, wait_table_lookup_build: 40.8ms, wait_table_lookup_resp: 115.8ms}                                                                                                        |                       | 1.72 MB | N/A  |
|     ├─IndexFullScan(Build)     | 9.85    | 915968  | cop[tikv] | table:t1, index:ia(a) | time:3.38s, loops:925, cop_task: {num: 26, max: 225.2ms, min: 25.2ms, avg: 132.1ms, p95: 192.5ms, tot_proc: 3.42s, rpc_num: 26, rpc_time: 3.43s, copr_cache_hit_ratio: 0.00, build_task_duration: 34.4µs, max_distsql_concurrency: 1}, tikv_task:{proc max:225.1ms, min:25.2ms, avg: 132ms, p80:190.2ms, p95:192.5ms, iters:0, tasks:26}                              | keep order:true       | N/A     | N/A  |
|     └─Selection(Probe)         | 1.00    | 12896   | cop[tikv] |                       | time:161.7ms, loops:72, cop_task: {num: 60, max: 4.22ms, min: 140.9µs, avg: 2.7ms, p95: 3.77ms, tot_proc: 136ms, rpc_num: 60, rpc_time: 161.4ms, copr_cache_hit_ratio: 0.00, build_task_duration: 448.5µs, max_distsql_concurrency: 2, max_extra_concurrency: 1}, tikv_task:{proc max:4.18ms, min:130.3µs, avg: 2.68ms, p80:3.49ms, p95:3.75ms, iters:0, tasks:60}    | gt(test.t1.b, 900000) | N/A     | N/A  |
|       └─TableRowIDScan         | 9.85    | 912896  | cop[tikv] | table:t1              | tikv_task:{proc max:4.18ms, min:130.3µs, avg: 2.68ms, p80:3.49ms, p95:3.75ms, iters:0, tasks:60}                                                                                                                                                                                                                                                                      | keep order:false      | N/A     | N/A  |
+--------------------------------+---------+---------+-----------+-----------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+-----------------------+---------+------+
```

terry1purcell:
Issue Number: close #50237

Add penalty to index scan for limit with order by when filtering exists outside of the ordering index. Add session variable to adjust ratio.

For a query that has filtering (WHERE clause), ordering (ORDER BY) and LIMIT - if there is an index that provides order, but filtering resides outside of that index, the optimizer currently has a very optimistic estimate as to when the qualified row(s) will be found.

For example in issue #50237, the query estimate is that the LIMIT 1 row will be found within 9.85 rows, when the actual is that 915,000 rows are scanned.

### Check List

Tests

- [x] Unit test
- [x] Integration test
- [ ] No need to test
 
Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

This enhancement will improve the bias towards a query plan that provides filtering vs a plan that provides ordering, but the filtering is in not in the chosen index. This can improve the stability of plans chosen by the optimizer. Given that the optimizer cannot estimate when a qualified row will be found (this is not a statistic that is available to the optimizer) - then being too optimistic creates risk.

The default is that this enhancement is disabled, but users can control the optimism/pessimism associated with this estimation with variable tidb_opt_ordering_index_selectivity_ratio.
1. Default = -1 : new estimation is disabled.
2. Value 0 : rows are estimated to be found immediately within the scan range.
3. Value >=0 and < 1 : a percentage of the scan range when a row will be found. 0.1 means 10% of the scan range. 0.5 means 50%. 1 means 100%.

### DIFF
--- a/pkg/sessionctx/variable/session.go
+++ b/pkg/sessionctx/variable/session.go
@@ -1514,6 +1514,11 @@ type SessionVars struct {
 	// use the ExpectedCnt to adjust the estimated row count for index scan.
 	OptOrderingIdxSelThresh float64
 
+	// OptOrderingIdxSelRatio is the ratio for optimizer to determine when qualified rows from filtering outside
+	// of the index will be found during the scan of an ordering index.
+	// If all filtering is applied as matching on the ordering index, this ratio will have no impact.
+	OptOrderingIdxSelRatio float64
+
 	// EnableMPPSharedCTEExecution indicates whether we enable the shared CTE execution strategy on MPP side.
 	EnableMPPSharedCTEExecution bool
 
--- a/pkg/planner/cardinality/cross_estimation.go
+++ b/pkg/planner/cardinality/cross_estimation.go
@@ -92,6 +92,10 @@ func AdjustRowCountForIndexScanByLimit(sctx sessionctx.Context,
 		correlationFactor := math.Pow(1-abs, float64(sctx.GetSessionVars().CorrelationExpFactor))
 		selectivity := dsStatsInfo.RowCount / rowCount
 		rowCount = min(expectedCnt/selectivity/correlationFactor, rowCount)
+		orderRatio := sctx.GetSessionVars().OptOrderingIdxSelRatio
+		if dsStatsInfo.RowCount < path.CountAfterAccess && orderRatio > 0 {
+			rowCount = max(rowCount, (path.CountAfterAccess-dsStatsInfo.RowCount)*orderRatio)
+		}
 	}
 	return rowCount
 }
--- a/pkg/planner/cardinality/cross_estimation.go
+++ b/pkg/planner/cardinality/cross_estimation.go
@@ -89,12 +89,21 @@ func AdjustRowCountForIndexScanByLimit(sctx sessionctx.Context,
 	if ok {
 		rowCount = count
 	} else if abs := math.Abs(corr); abs < 1 {
-		correlationFactor := math.Pow(1-abs, float64(sctx.GetSessionVars().CorrelationExpFactor))
-		selectivity := dsStatsInfo.RowCount / rowCount
-		rowCount = min(expectedCnt/selectivity/correlationFactor, rowCount)
+		// If OptOrderingIdxSelRatio is enabled - estimate the difference between index and table filtering, as this represents
+		// the possible scan range when LIMIT rows will be found. orderRatio is the estimated percentage of that range when the first
+		// row isexpected to be found. Index filtering applies orderRatio twice. Once found - rows are estimated to be clustered (expectedCnt).
+		// This formula is to bias away from non-filtering (or poorly filtering) indexes that provide order due, where filtering exists
+		// outside of that index. Such plans have high risk since we cannot estimate when rows will be found.
 		orderRatio := sctx.GetSessionVars().OptOrderingIdxSelRatio
-		if dsStatsInfo.RowCount < path.CountAfterAccess && orderRatio > 0 {
-			rowCount = max(rowCount, (path.CountAfterAccess-dsStatsInfo.RowCount)*orderRatio)
+		if dsStatsInfo.RowCount < path.CountAfterAccess && orderRatio > -1 {
+			rowsToMeetFirst := (((path.CountAfterAccess - path.CountAfterIndex) * orderRatio) + (path.CountAfterIndex - dsStatsInfo.RowCount)) * orderRatio
+			rowCount = rowsToMeetFirst + expectedCnt
+		} else {
+			// Assume rows are linearly distributed throughout the range - for example: selectivity 0.1 assumes that a
+			// qualified row is found every 10th row.
+			correlationFactor := math.Pow(1-abs, float64(sctx.GetSessionVars().CorrelationExpFactor))
+			selectivity := dsStatsInfo.RowCount / rowCount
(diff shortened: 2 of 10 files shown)


pkg/sessionctx/variable/session.go
```diff
@@ -1514,6 +1514,11 @@ type SessionVars struct {
 	// use the ExpectedCnt to adjust the estimated row count for index scan.
 	OptOrderingIdxSelThresh float64
 
+	// OptOrderingIdxSelRatio is the ratio for optimizer to determine when qualified rows from filtering outside
+	// of the index will be found during the scan of an ordering index.
+	// If all filtering is applied as matching on the ordering index, this ratio will have no impact.
+	OptOrderingIdxSelRatio float64
```
elsa0520, line 1520: It seems that every query need different OptOrderingIdxSelRatio, even th values of different "where column=value" in the same query template should use different configurations.
So I feel it should be more like a hint rather than a session variable

terry1purcell: This was modeled after the prior solution for OptOrderingIdxSelThreshold - which has the same challenge. However, a hint is not the preferred approach here - since we already have a way to hint an individual query with hinting the individual index. We need a solution that applies more generally - which is why this has a global variable.

winoros (commented):
Seems that we need to also modify the func `AdjustRowCountForTableScanByLimit`?

winoros:
And then we can prepare to remove the variable `correlationFactor`?

terry1purcell:
> Seems that we need to also modify the func `AdjustRowCountForTableScanByLimit`?

This was modeled off a prior enhancement that introduced tidb_opt_ordering_index_selectivity_threshold - that also avoided changing the table scan estimation.

Similar to that prior enhancement - func 'AdjustRowCountForTableScanByLimit' wasn't considered because the problem that was exposed relates to LIMIT query with WHERE clause filtering that is separate from the ORDER by column(s) - such that there is a need to penalize an ordering plan that doesn't also provide (adequate) filtering.

This costing does however relate to both when the sort property matches the ordering, but also when the sort property is empty. In unit testing - the row estimate adjustment does then carry over to the table scan plan (if there was an index that resulted in AdjustRowCountForIndexScanByLimit) - such that it is then a cost based decision between a table scan and other available indexes that may or may not have this penalty.

It was my assessment therefore that the need to also update AdjustRowCountForTableScanByLimit was for scenarios where there aren't other viable indexes. Upon such analysis - it does seem that the estimate for finding a row via a table scan (when another index hasn't adjusted the estimate) is similarly optimistic and thus will assume a LIMIT row is found early in a scan.

My initial conclusion therefore was that this should be investigated separately if this issue is exposed by a customer or in other testing.

terry1purcell:
> And then we can prepare to remove the variable `correlationFactor`?

When the new tidb_opt_ordering_index_selectivity_ratio is set to it's default of -1, then the correlationFactor is still used for the original row estimation. It would make sense to remove correlationFactor at the time when we can deprecate the original estimation logic.

PR #49205, opened 2023-12-06
parser: support `OPTIMIZE` statement

### LINKED ISSUE #49204: The parser can't parse the `OPTIMIZE` DDL operation.
## Bug Report

```go
import (
	"fmt"
	"log"
	"reflect"

	"github.com/pingcap/tidb/parser"
	_ "github.com/pingcap/tidb/types/parser_driver"
)

func main() {
	p := parser.New()
	stmt, err := p.ParseOneStmt("OPTIMIZE TABLE foo", "", "")
	if err != nil {
		log.Fatalln(err)
	}
}
```

`ast.OptimizeTableStmt{}`

```go
2023/12/05 22:13:42 line 1 column 8 near "OPTIMIZE TABLE foo;" 
```

```
github.com/pingcap/tidb v1.1.0-beta.0.20230719095317-cf441574864b
github.com/pingcap/tidb/parser v0.0.0-20230719095317-cf441574864b
```

ub-3:
Now it is possible to parse SQL with OPTIMIZE statement.
```sql
OPTIMIZE [NO_WRITE_TO_BINLOG | LOCAL]
    TABLE tbl_name [, tbl_name] ...
```

Issue Number: close #49204

Problem Summary: There is no support for the `OPTIMIZE` statement

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Added support for the "OPTIMIZE" statement.
```


pkg/parser/parser.y
```diff
@@ -14721,6 +14734,27 @@ OptionalShardColumn:
...
 	}
 
+/********************************************************************
+ * OptimizeTableStmt
+ *
+ * OPTIMIZE [NO_WRITE_TO_BINLOG | LOCAL]
+ *     TABLE tbl_name [, tbl_name] ...
+ *******************************************************************/
+OptimizeTableStmt:
+	"OPTIMIZE" NoWriteToBinLogOpt LocalOpt "TABLE" TableNameList
```
Defined2014, line 14744: It will support stmt like `OPTIMIZE NO_WRITE_TO_BINLOG LOCAL TABLE t;` which is not allowed in MySQL.

bb7133: @ub-3 you could just follow the [yacc definition in MySQL](https://github.com/mysql/mysql-server/blob/1bfe02bdad6604d54913c62614bde57a055c8332/sql/sql_yacc.yy#L9658-L9669):

> opt_no_write_to_binlog:
>          /* empty */ { $$= 0; }
>        | NO_WRITE_TO_BINLOG { $$= 1; }
>        | LOCAL_SYM { $$= 1; }

bb7133:
It would be confusing to TiDB users when 'OPTIMIZE TABLE' is accepted but does nothing. Do you have any specific requirement to have this supported in syntax only?

lance6716:
> It would be confusing to TiDB users when 'OPTIMIZE TABLE' is accepted but does nothing. Do you have any specific requirement to have this supported in syntax only?

I agree that since this PR changes the behaviour for user, we need to discuss what's the expected behaviour for TiDB of `OPTIMIZE` statement. For example, should we invoke TiKV's functions? Don't want to change it back and forth in future.

Maybe If you only care about the parser's golang API, you can still let TiDB report error for OPTIMIZE.

ub-3:
@bb7133 To be honest, I'm only interested in parser support. I should read MySQL slow-logs and analyzes queries, and your package is just what I need. I'm not so familiar with the project, but it seems to me, given that the parser used to be in a separate project, adding this support is not such a problem.

ub-3:
@lance6716 
> Maybe If you only care about the parser's golang API, you can still let TiDB report error for OPTIMIZE.

Can you explain what it means? Maybe there is an example?

mjonss (commented):
LGTM, but probably something is needed in TiDB code to either implement it or to block/fail it. With current patch it will just say OK, without any checks for table etc.
```
tidb> drop table if exists t;

tidb> show warnings;
+-------+------+------------------------+
| Level | Code | Message                |
+-------+------+------------------------+
| Note  | 1051 | Unknown table 'test.t' |
+-------+------+------------------------+
1 row in set (0.00 sec)

tidb> optimize table t;
```

ub-3:
@mjonss What about this kind of message ?
```
mysql> optimize table t;
ERROR 1235 (42000): This version of TiDB doesn't yet support 'OPTIMIZE'
```
This can be done by handling the statement as follows.
```go
pkg/planner/core/planbuilder.go:819
...
	case *ast.OptimizeTableStmt:
		return nil, ErrNotSupportedYet.GenWithStackByArgs("OPTIMIZE")
...
```

ub-3:
@mjonss Thanks for the good explanation!

Defined2014 (approved):
LGTM
Need update a private repo to pass mysql-tester, which I have done. https://github.com/PingCAP-QE/tidb-test/pull/2261

mjonss (approved):

mjonss:
I holding this PR, so we can merge it with the matching PR for passing the mysql-tester CI test.

dveeden:
This looks good. However it should have set some of the items in the description as this affects mysql compatibility, user behavior and syntax (all of them in a positive way).

tangenta (approved):

PR #49229, opened 2023-12-06
ddl: fix partition definition for literal with non-utf8 charsets + binary column | tidb-test=pr/2262

### LINKED ISSUE #36433: string literals in partition definition doesn't respect charset settings
## Bug Report

```
my:root@127.0.0.1:4000=> use test;
USE
my:root@127.0.0.1:4000=> set character_set_connection=gbk;
SET
my:root@127.0.0.1:4000=> create table `test`.`a` (id bigint, name varbinary(16) unique key) partition by range columns(name) (
my:root@127.0.0.1:4000(> partition p0 values less than ('你好'),
my:root@127.0.0.1:4000(> partition p1 values less than ('我好'),
my:root@127.0.0.1:4000(> partition p2 values less than MAXVALUE);
CREATE TABLE
my:root@127.0.0.1:4000=> insert into a values(1, '你好');
INSERT 1
my:root@127.0.0.1:4000=> select * from a;
 id |       name
----+------------------
  1 | \xc4\xe3\xba\xc3
(1 row)

my:root@127.0.0.1:4000=> select * from a partition(p0);
 id |       name
----+------------------
  1 | \xc4\xe3\xba\xc3
(1 row)
my:root@127.0.0.1:4000=> set character_set_connection=utf8mb4;
SET
my:root@127.0.0.1:4000=> insert into a values(1, '你好');
INSERT 1
my:root@127.0.0.1:4000=> select * from a partition(p0);
 id |       name
----+------------------
  1 | \xc4\xe3\xba\xc3
(1 row)

my:root@127.0.0.1:4000=> select * from a partition(p1);
 id | name
----+------
  1 | 你好
(1 row)
```
Seems TiDB always uses utf8mb4 to store string literals in partition DDL. So when I use gbk charset to insert '你好', it falls into partition p0.

In MySQL 8, the same procedure:

```
mysql> set character_set_connection=gbk;
mysql> create table `test`.`a` (id bigint, name varbinary(16) unique key) partition by range columns(name) (
    -> partition p0 values less than ('你好'),
    -> partition p1 values less than ('我好'),
    -> partition p2 values less than MAXVALUE);
mysql> insert into a values(1, '你好');

mysql> select * from a;
+------+------------+
| id   | name       |
+------+------------+
|    1 | 0xC4E3BAC3 |
+------+------------+
1 row in set (0.00 sec)
mysql> select * from a partition(p0);
Empty set (0.00 sec)
mysql> select * from a partition(p1);
+------+------------+
| id   | name       |
+------+------------+
|    1 | 0xC4E3BAC3 |
+------+------------+
1 row in set (0.00 sec)
mysql> set character_set_connection=utf8mb4;
mysql> insert into a values(1, '你好');
mysql> select * from a partition(p0);
Empty set (0.01 sec)
mysql> select * from a partition(p1);
+------+------------+
| id   | name       |
+------+------------+
|    1 | 0xC4E3BAC3 |
+------+------------+
1 row in set (0.00 sec)
mysql> select * from a partition(p2);
+------+----------------+
| id   | name           |
+------+----------------+
|    1 | 0xE4BDA0E5A5BD |
+------+----------------+
1 row in set (0.00 sec)
```

Seems MySQL respects the charset settings when executing partition DDL.

nightly

### LINKED ISSUE #49251: TiDB gives hard-to-understand error message for definition with `NULL` in range partition
## Bug Report

```
mysql> create table a (col1 int, col2 int, unique key (col1, col2)) partition by range  columns (col1, col2) (partition p0 values less than (NULL, 1 ));
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your TiDB version for the right syntax to use line 1 column 28 near ",1))"
```

Expected:

```
mysql> create table a (col1 int, col2 int, unique key (col1, col2)) partition by range  columns (col1, col2) (partition p0 values less than (NULL, 1));
ERROR 1566 (HY000): Not allowed to use NULL value in VALUES LESS THAN
```

YangKeao:

Issue Number: close #36433, close #49251

Problem Summary:

The string literal in the partition clause is stored as a formated string literal. For example, the `'你好'` will be kept as `'你好'` in the schema, and the collation information is lost. However, when running the `getRangeLocateExprs`, the expression is constructed with the DDL session, which always use `utf8mb4` collation.

When inserting the row, the string (after converted to the collation corresponding to the column) will be compared with the `utf8mb4` literal string, which may cause unexpected behavior. In the issue #36433, the insert value `你好` is converted to binary encoding GBK `\xc4\xe3\xba\xc3`, and compare it with the utf8 string `'你好'` through binary collation directly, then former is smaller than the later one.

MySQL records the binary representation in table schema. If we run `show create table a` in MySQL, it'll give a hex literal  with `_binary` prefix in the partition clause.

If the target is binary charset, use the hex literal to represent the string. Actually this PR does two change:

1. Always use the evaluated string to represent the partition clause.
2. If the target column collation is binary, use the `_binary` + hex literal to represent the definition, as it may be an invalid utf8 string.

As TiDB always use utf8 to represent a string internally (except binary), it's fine to not consider other charset like GBK.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
Fix the issue that the partition clause is compared through the collation of `utf8` charset, but not the column charset.
```

### DIFF
--- a/pkg/ddl/partition.go
+++ b/pkg/ddl/partition.go
@@ -38,6 +38,7 @@ import (
 	"github.com/pingcap/tidb/pkg/metrics"
 	"github.com/pingcap/tidb/pkg/parser"
 	"github.com/pingcap/tidb/pkg/parser/ast"
+	"github.com/pingcap/tidb/pkg/parser/charset"
 	"github.com/pingcap/tidb/pkg/parser/format"
 	"github.com/pingcap/tidb/pkg/parser/model"
 	"github.com/pingcap/tidb/pkg/parser/mysql"
@@ -1309,13 +1310,14 @@ func buildListPartitionDefinitions(ctx expression.BuildContext, defs []*ast.Part
 			return nil, err
 		}
 		clause := def.Clause.(*ast.PartitionDefinitionClauseIn)
+		partVals := make([][]types.Datum, 0, len(clause.Values))
 		if len(tbInfo.Partition.Columns) > 0 {
 			for _, vs := range clause.Values {
-				// TODO: use the generated strings / normalized partition values
-				_, err := checkAndGetColumnsTypeAndValuesMatch(ctx, colTypes, vs)
+				vals, err := checkAndGetColumnsTypeAndValuesMatch(ctx, colTypes, vs)
 				if err != nil {
 					return nil, err
 				}
+				partVals = append(partVals, vals)
 			}
 		} else {
 			for _, vs := range clause.Values {
@@ -1339,16 +1341,32 @@ func buildListPartitionDefinitions(ctx expression.BuildContext, defs []*ast.Part
 		}
 
 		buf := new(bytes.Buffer)
-		for _, vs := range clause.Values {
+		for valIdx, vs := range clause.Values {
 			inValue := make([]string, 0, len(vs))
-			for i := range vs {
-				vs[i].Accept(exprChecker)
-				if exprChecker.err != nil {
-					return nil, exprChecker.err
+			isDefault := false
+			if len(vs) == 1 {
+				if _, ok := vs[0].(*ast.DefaultExpr); ok {
+					isDefault = true
 				}
-				buf.Reset()
-				vs[i].Format(buf)
-				inValue = append(inValue, buf.String())
+			}
+			if len(partVals) > valIdx && !isDefault {
+				for colIdx := range partVals[valIdx] {
+					partVal, err := generatePartValuesWithTp(partVals[valIdx][colIdx], colTypes[colIdx])
+					if err != nil {
+						return nil, err
+					}
+					inValue = append(inValue, partVal)
+				}
+			} else {
+				for i := range vs {
+					vs[i].Accept(exprChecker)
+					if exprChecker.err != nil {
+						return nil, exprChecker.err
+					}
+					buf.Reset()
+					vs[i].Format(buf)
+					inValue = append(inValue, buf.String())
+				}
 			}
 			piDef.InValues = append(piDef.InValues, inValue)
 			buf.Reset()
@@ -1387,10 +1405,10 @@ func buildRangePartitionDefinitions(ctx expression.BuildContext, defs []*ast.Par
 			return nil, err
 		}
 		clause := def.Clause.(*ast.PartitionDefinitionClauseLessThan)
-		var partValStrings []string
+		var partValDatums []types.Datum
 		if len(tbInfo.Partition.Columns) > 0 {
 			var err error
-			if partValStrings, err = checkAndGetColumnsTypeAndValuesMatch(ctx, colTypes, clause.Exprs); err != nil {
+			if partValDatums, err = checkAndGetColumnsTypeAndValuesMatch(ctx, colTypes, clause.Exprs); err != nil {
 				return nil, err
 			}
 		} else {
@@ -1424,19 +1442,23 @@ func buildRangePartitionDefinitions(ctx expression.BuildContext, defs []*ast.Par
 				return nil, exprChecker.err
 			}
 			// If multi-column use new evaluated+normalized output, instead of just formatted expression
-			if len(partValStrings) > i && len(colTypes) > 1 {
-				partVal := partValStrings[i]
-				switch colTypes[i].EvalType() {
-				case types.ETInt:
-					// no wrapping
-				case types.ETDatetime, types.ETString, types.ETDuration:
-					if _, ok := clause.Exprs[i].(*ast.MaxValueExpr); !ok {
-						// Don't wrap MAXVALUE
-						partVal = driver.WrapInSingleQuotes(partVal)
-					}
-				default:
-					return nil, dbterror.ErrWrongTypeColumnValue.GenWithStackByArgs()
+			if len(partValDatums) > i {
+				var partVal string
+				if partValDatums[i].Kind() == types.KindNull {
+					return nil, dbterror.ErrNullInValuesLessThan
 				}
+				if _, ok := clause.Exprs[i].(*ast.MaxValueExpr); ok {
+					partVal, err = partValDatums[i].ToString()
+					if err != nil {
+						return nil, err
+					}
+				} else {
+					partVal, err = generatePartValuesWithTp(partValDatums[i], colTypes[i])
+					if err != nil {
+						return nil, err
+					}
+				}
+
 				piDef.LessThan = append(piDef.LessThan, partVal)
 			} else {
 				expr.Format(buf)
@@ -4383,4 +4405,31 @@ func AppendPartitionDefs(partitionInfo *model.PartitionInfo, buf *bytes.Buffer, 
 			fmt.Fprintf(buf, " /*T![placement] PLACEMENT POLICY=%s */", stringutil.Escape(def.PlacementPolicyRef.Name.O, sqlMode))
 		}
 	}
+}
+
+func generatePartValuesWithTp(partVal types.Datum, tp types.FieldType) (string, error) {
+	if partVal.Kind() == types.KindNull {
+		return "NULL", nil
+	}
+
+	s, err := partVal.ToString()
+	if err != nil {
+		return "", err
+	}
+
+	switch tp.EvalType() {
+	case types.ETInt:
+		return s, nil
+	case types.ETString:
+		// The `partVal` can be an invalid utf8 string if it's converted to BINARY, then the content will be lost after
+		// marshaling and storing in the schema. In this case, we use a hex literal to work around this issue.
+		if tp.GetCharset() == charset.CharsetBin {
+			return fmt.Sprintf("_binary 0x%x", s), nil
+		}
+		return driver.WrapInSingleQuotes(s), nil
+	case types.ETDatetime, types.ETDuration:
+		return driver.WrapInSingleQuotes(s), nil
+	}
+
+	return "", dbterror.ErrWrongTypeColumnValue.GenWithStackByArgs()
 }
--- a/tests/integrationtest/r/ddl/partition.result
+++ b/tests/integrationtest/r/ddl/partition.result
@@ -280,3 +280,180 @@ UNIQUE KEY (id, id1))
 PARTITION BY KEY()
 PARTITIONS 2;
 Error 1105 (HY000): Table partition metadata not correct, neither partition expression or list of partition columns
+set character_set_connection=gbk;
+drop table if exists t;
+create table t (col1 varbinary(16) unique key) partition by list columns(col1)
+(partition p0 values in ('你好', '我好'), partition p1 values in ('大家好'), partition p2 DEFAULT);
+show create table t;
+Table	Create Table
+t	CREATE TABLE `t` (
+  `col1` varbinary(16) DEFAULT NULL,
+  UNIQUE KEY `col1` (`col1`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
+PARTITION BY LIST COLUMNS(`col1`)
+(PARTITION `p0` VALUES IN (_binary 0xc4e3bac3,_binary 0xced2bac3),
+ PARTITION `p1` VALUES IN (_binary 0xb4f3bcd2bac3),
+ PARTITION `p2` DEFAULT)
+insert into t values ("你好");
+select hex(col1) from t partition(p0);
+hex(col1)
+C4E3BAC3
+select hex(col1) from t partition(p1);
+hex(col1)
+select hex(col1) from t partition(p2);
+hex(col1)
+drop table t;
+create table t (col1 varbinary(16) unique key) partition by range columns(col1)
+(partition p0 values less than ('你好'), partition p1 values less than ('我好'), partition p2 values less than MAXVALUE);
+show create table t;
+Table	Create Table
+t	CREATE TABLE `t` (
+  `col1` varbinary(16) DEFAULT NULL,
+  UNIQUE KEY `col1` (`col1`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
+PARTITION BY RANGE COLUMNS(`col1`)
+(PARTITION `p0` VALUES LESS THAN (_binary 0xc4e3bac3),
+ PARTITION `p1` VALUES LESS THAN (_binary 0xced2bac3),
+ PARTITION `p2` VALUES LESS THAN (MAXVALUE))
+insert into t values ("你好");
+select hex(col1) from t partition(p0);
+hex(col1)
+select hex(col1) from t partition(p1);
+hex(col1)
+C4E3BAC3
+select hex(col1) from t partition(p2);
+hex(col1)
+drop table t;
+create table t (col1 varchar(16) collate gbk_chinese_ci unique key) partition by list columns(col1)
+(partition p0 values in ('你好', '我好'), partition p1 values in ('大家好'), partition p2 DEFAULT);
+show create table t;
+Table	Create Table
+t	CREATE TABLE `t` (
+  `col1` varchar(16) CHARACTER SET gbk COLLATE gbk_chinese_ci DEFAULT NULL,
+  UNIQUE KEY `col1` (`col1`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
+PARTITION BY LIST COLUMNS(`col1`)
+(PARTITION `p0` VALUES IN ('你好','我好'),
+ PARTITION `p1` VALUES IN ('大家好'),
+ PARTITION `p2` DEFAULT)
+insert into t values ("你好");
+select hex(col1) from t partition(p0);
+hex(col1)
+C4E3BAC3
+select hex(col1) from t partition(p1);
+hex(col1)
+select hex(col1) from t partition(p2);
+hex(col1)
+drop table t;
+create table t (col1 varchar(16) collate gbk_chinese_ci unique key) partition by range columns(col1)
+(partition p0 values less than ('你好'), partition p1 values less than ('我好'), partition p2 values less than MAXVALUE);
+show create table t;
+Table	Create Table
+t	CREATE TABLE `t` (
+  `col1` varchar(16) CHARACTER SET gbk COLLATE gbk_chinese_ci DEFAULT NULL,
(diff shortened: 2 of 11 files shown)


pkg/ddl/partition.go
```diff
@@ -1399,21 +1417,22 @@ func buildRangePartitionDefinitions(ctx sessionctx.Context, defs []*ast.Partitio
 				return nil, exprChecker.err
 			}
 			// If multi-column use new evaluated+normalized output, instead of just formatted expression
-			if len(partValStrings) > i && len(colTypes) > 1 {
+			if len(partValStrings) > i && len(tbInfo.Partition.Columns) > 0 {
```
YangKeao, line 1420: Not sure whether it's correct to replace `len(colTypes) > 1` with `len(tbInfo.Partition.Columns) > 0`. Is the number `1` a special number or just a typo?

Is it also correct to only set `len(partValStrings) > i` without other conditions here? @mjonss PTAL

pkg/ddl/partition.go
```diff
@@ -1314,16 +1316,32 @@ func buildListPartitionDefinitions(ctx sessionctx.Context, defs []*ast.Partition
...
-				vs[i].Accept(exprChecker)
-				if exprChecker.err != nil {
-					return nil, exprChecker.err
+			isDefault := false
+			if len(vs) == 1 {
+				if _, ok := vs[0].(*ast.DefaultExpr); ok {
+					isDefault = true
+				}
+			}
+			if len(partVals) > valIdx && len(tbInfo.Partition.Columns) > 0 && !isDefault {
```
mjonss, line 1327: When is `len(partVals) > valIdx` not true when `len(tbInfo.Partition.Columns)` is?

pkg/ddl/partition.go
mjonss: Above there is a check for `if len(colTypes) != len(tbInfo.Partition.Columns)` so they should be the same here.
I think there was a reason for keeping the output for single column RANGE COLUMNS partitioning (IIRC the evaluation+normalization of the values was done when adding multi-column RANGE COLUMNS, so I just wanted to avoid changing the single column at the same time).

So if we also want the evaluated+normalized value string here also for single column RANGE COLUMNS, then it is OK to also remove the `&& len(tbInfo.Partition.Columns)` part.
I.e. OK to change now so also single column RANGE COLUMNS uses the better formatting.

pkg/ddl/partition.go
```diff
@@ -1399,21 +1417,22 @@ func buildRangePartitionDefinitions(ctx sessionctx.Context, defs []*ast.Partitio
...
-					return nil, dbterror.ErrWrongTypeColumnValue.GenWithStackByArgs()
 				}
+
 				piDef.LessThan = append(piDef.LessThan, partVal)
 			} else {
+				// TODO: return the integer value from `checkPartitionValuesIsInt` and use it to construct `piDef` like
+				// the above condition.
+				// It's fine to not consider the collation/charset problem as it only supports integer now, but once it
+				// supports more functions and other types, we'll need to reconsider how to format the expression in this
+				// situation.
```
mjonss, line 1435: I don't think RANGE/LIST/HASH will support other types than INT based ones, that is the reason for RANGE/LIST COLUMNS. IIRC adding COLUMNS partitioning was to separate that implementation from the simpler INT based ones when it was added in MySQL.

tests/integrationtest/t/ddl/partition.test
```diff
@@ -135,3 +135,62 @@ show create table t;
 alter table t partition by key(a) partitions 5;
 show create table t;
 
+# TestPartitionInNonUTF8Charset
```
mjonss, line 138: Could you also add some `SHOW CREATE TABLE` output?

mjonss (commented):
Will this affect existing partitioned tables? I mean if someone already have a partitioned table with data that is affected by the issue/PR, will such tables need either the old logic, or be upgraded/repaired?

pkg/ddl/partition.go
YangKeao: OK. I'll remove this condition. Then only partition with simple expression (and without column) will use `expr.Format`, all other situations will use the evaluated value.

pkg/ddl/partition.go
YangKeao: `len(tbInfo.Partition.Columns) > 0` is redundant. I'll only keep `len(partVals) > valIdx && !isDefault`. :+1: .

pkg/ddl/partition.go
YangKeao: Got it. Thanks for the clarifications.

tests/integrationtest/t/ddl/partition.test
YangKeao: Done :+1:

YangKeao:
> Will this affect existing partitioned tables? I mean if someone already have a partitioned table with data that is affected by the issue/PR, will such tables need either the old logic, or be upgraded/repaired?

@mjonss I think we have no way to fix the existing partitioned tables 100% correctly, as we have lost the information about the charsets of session when the user is creating the table. Consider the following situation:

```
create table t (col1 varbinary(16) unique key) partition by list columns(col1)     (partition p0 values in ('你好', '我好'), partition p1 values in ('大家好'), partition p2 DEFAULT);
insert into t values('你好');
select * from t partition(p2);
```

If the sessions `set character_set_connection=gbk;` before the first line, `p2` will be empty and the row locates in `p0`, but if `set character_set_connection=gbk;` after `create` and before `insert`, the row will locate in p2. However, the schema (and behavior) is the same without this PR.


As a reference, here is the current bug behavior. With or without this PR, existing table will still follow these behaviors:

1. For multi-column range partition + non-utf8 session charset + binary column charset, this problem is quite scary. It'll produce invalid utf8 string. All invalid characters will be truncated as unicode 0xFFFD. It may cause multiple partition with the same condition. For example: 
```
set character_set_connection=gbk;
create table t (col1 varbinary(16), col2 varbinary(16), unique key (col1, col2)) partition by range columns(col1, col2)     (partition p0 values less than ('你好','你好'), partition p1 values less than ('你在','你在'), partition p2 values less than (MAXVALUE, MAXVALUE));
show create table t;
```
It will show that the condition of `p0` and `p1` are the same.
```
+-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                                                                                                                                                                                                                                                     |
+-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `col1` varbinary(16) DEFAULT NULL,
  `col2` varbinary(16) DEFAULT NULL,
  UNIQUE KEY `col1` (`col1`,`col2`)
PARTITION BY RANGE COLUMNS(`col1`,`col2`)
(PARTITION `p0` VALUES LESS THAN ('����','����'),
 PARTITION `p1` VALUES LESS THAN ('����','����'),
 PARTITION `p2` VALUES LESS THAN (MAXVALUE,MAXVALUE))                                 |
+-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

Then, `0xFFFD` (actually `EF BF BD` in utf8 encoding) is greater than most strings, so many string will locate in `p0`.

(Maybe we can add a check to avoid producing invalid utf8 schema. I record an issue for it https://github.com/pingcap/tidb/issues/49259)

2. For list partition / single-column range partition + non-utf8 session charset + binary column charset, as described above, the location of rows may be unexpected.

mjonss:
> > Will this affect existing partitioned tables? I mean if someone already have a partitioned table with data that is affected by the issue/PR, will such tables need either the old logic, or be upgraded/repaired?
> 
> @mjonss I think we have no way to fix the existing partitioned tables 100% correctly, as we have lost the information about the charsets of session when the user is creating the table. ...
> 1. For multi-column range partition + non-utf8 session charset + binary column charset, this problem is quite scary. It'll produce invalid utf8 string. All invalid characters will be truncated as unicode 0xFFFD. It may cause multiple partition with the same condition. For example:
> ...
> 2. For list partition / single-column range partition + non-utf8 session charset + binary column charset, as described above, the location of rows may be unexpected.

@YangKeao Is it only the `SHOW CREATE TABLE` that is affected, or also how the partition definitions are interpreted and used internally?
I.e. Does the table partitioning still works as expected (and stable and does not change with this PR) when it comes to placing rows it the correct partitions? Like if the partition was created with a session charset of 'gbk' it will place the rows accordingly (even if there are gbk character encodings that are non-valid utf8 encodings)?
If there is only the output from `SHOW CREATE TABLE` that is different, then I think we have much more freedom in fixing this.
If we change how the partition definition values are interpreted when evaluating which partition a row belongs to, so there is a possibility that rows belongs to other partitions after this PR, we need to take care and probably also introduce some kind of table/partition setting to distinguish between the behavior before this PR and after.

YangKeao:
> > > Will this affect existing partitioned tables? I mean if someone already have a partitioned table with data that is affected by the issue/PR, will such tables need either the old logic, or be upgraded/repaired?
> > 
> > 
> > @mjonss I think we have no way to fix the existing partitioned tables 100% correctly, as we have lost the information about the charsets of session when the user is creating the table. ...
> > 
> > 1. For multi-column range partition + non-utf8 session charset + binary column charset, this problem is quite scary. It'll produce invalid utf8 string. All invalid characters will be truncated as unicode 0xFFFD. It may cause multiple partition with the same condition. For example:
> >    ...
> > 2. For list partition / single-column range partition + non-utf8 session charset + binary column charset, as described above, the location of rows may be unexpected.
> 
> @YangKeao Is it only the `SHOW CREATE TABLE` that is affected, or also how the partition definitions are interpreted and used internally? I.e. Does the table partitioning still works as expected (and stable and does not change with this PR) when it comes to placing rows it the correct partitions? Like if the partition was created with a session charset of 'gbk' it will place the rows accordingly (even if there are gbk character encodings that are non-valid utf8 encodings)? If there is only the output from `SHOW CREATE TABLE` that is different, then I think we have much more freedom in fixing this. If we change how the partition definition values are interpreted when evaluating which partition a row belongs to, so there is a possibility that rows belongs to other partitions after this PR, we need to take care and probably also introduce some kind of table/partition setting to distinguish between the behavior before this PR and after.

@mjonss 
> Is it only the `SHOW CREATE TABLE` that is affected, or also how the partition definitions are interpreted and used internally?

Not only `SHOW CREATE TABLE`. It modifies how DDL statements (related with partition) are interpreted and converted to schema. It doesn't change how the existing partition definitions (in schema) are interpreted into expressions, and therefore, doesn't change the behavior of existing table.

This PR only modifies `buildListPartitionDefinitions` and `buildRangePartitionDefinitions`. IIUC, these functions are only used in DDL (`CREATE TABLE`, `ALTER ...`).

> Like if the partition was created with a session charset of 'gbk' it will place the rows accordingly (even if there are gbk character encodings that are non-valid utf8 encodings)?

The behavior of existing table is stable (but with bug) before/after this PR. I didn't try (and it's impossible) to fix this issue for existing tables in this PR.

The behavior of newly created tables can be different from existing table created with the same `CREATE TABLE` statement. I'm not sure whether it's a big problem :thinking:.

pkg/ddl/partition.go
```diff
@@ -1399,19 +1417,21 @@ func buildRangePartitionDefinitions(ctx sessionctx.Context, defs []*ast.Partitio
...
+				if err != nil {
+					return nil, err
+				}
+				if partValDatums[i].Kind() == types.KindNull {
+					return nil, dbterror.ErrNullInValuesLessThan
+				}
+				if _, ok := clause.Exprs[i].(*ast.MaxValueExpr); !ok {
+					partVal, err = generatePartValuesWithTp(partValDatums[i], colTypes[i])
+					if err != nil {
+						return nil, err
```
mjonss, line 1431: [minor] Maybe rewrite this to something like:
```
				if partValDatums[i].Kind() == types.KindNull {
					return nil, dbterror.ErrNullInValuesLessThan
				}
				if _, ok := clause.Exprs[i].(*ast.MaxValueExpr); ok {
					partVal, err := partValDatums[i].ToString()
					if err != nil {
						return nil, err
					}
				} else {
					partVal, err = generatePartValuesWithTp(partValDatums[i], colTypes[i])
					if err != nil {
						return nil, err
					}
				}
```

mjonss (approved):
Approved, with a minor suggestion.

mjonss:
> ...
> The behavior of newly created tables can be different from existing table created with the same `CREATE TABLE` statement. I'm not sure whether it's a big problem 🤔.

@YangKeao The only issue I can think of is that if an old table is affected by this, and that table is then ALTERED, is that it will change the VALUES IN/LESS for existing partitions, in a way that the partitioning would place rows differently.
Not sure if it is limited to only ADD/REMOVE/REORGANIZE PARTITION, since any ALTER would regenerate the partitioning expression I assume?

So this is my main concern.

YangKeao:
> > ...
> > The behavior of newly created tables can be different from existing table created with the same `CREATE TABLE` statement. I'm not sure whether it's a big problem 🤔.
> 
> @YangKeao The only issue I can think of is that if an old table is affected by this, and that table is then ALTERED, is that it will change the VALUES IN/LESS for existing partitions, in a way that the partitioning would place rows differently. Not sure if it is limited to only ADD/REMOVE/REORGANIZE PARTITION, since any ALTER would regenerate the partitioning expression I assume?
> 
> So this is my main concern.

Got it. Let me do some tests / read more codes about the `ALTER ...` process :thinking:. 

/hold

YangKeao:
These two modified functions have following callstacks:

1. `BuildAddedPartitionInfo` -> `buildPartitionDefinitionsInfo`. It's used in `ALTER ... ADD PARTITION ...`. In this case, the existing partition will not be affected by this PR, so the existing data will not be located to another partition. However, the range maybe confusing as the behavior changed within the same table.
2. `GetModifiableColumnJob` -> `buildPartitionDefinitionsInfo`. The return value of `buildPartitionDefinitionsInfo` is not used. It's only used to check whether the column change will break the existing partition configuration.
3. `buildTablePartitionInfo` -> `buildPartitionDefinitionsInfo`
   1. `AlterTablePartitioning` -> `buildTablePartitionInfo` -> `buildPartitionDefinitionsInfo`. This happends in `ALTER TABLE ... PARTITION ...` (`AlterTablePartition`). In this case, all existing partition will be removed and re-organized into new partition definitions. BTW, `REORGANIZE PARTITION` will not affect unrelated partitions (just like `ADD PARTITION`).
   2. `BuildTableInfoWithStmt` -> `buildTablePartitionInfo` -> `buildPartitionDefinitionsInfo`. This routine is called in `CREATE TABLE ...` and `ADMIN REPAIR TABLE`.

Another thing comes into my mind is that I'm not sure whether dm, cdc, br... can work well for this situation across the upgradation :facepalm:. But as long as the schema itself is synced, it should be fine.

Below is the conclusion of the behavior:

1. For `ALTER TABLE ... ADD/REMOVE/REORGANIZE PARTITION ...` will not change the schema of unrelated partitions. If a partition is not touched, it'll keep the original behavior. The newly added partitions will follow the new behavior.
2. For `ALTER TABLE ... PARTITION ...`, the full partition definition will be rebuilt. Therefore, after running this kind of DDL, all partitions will follow the new behavior.
3. For tools, not sure. I also don't know whether they will set session variables like `character_set_connection`. The statement given by `SHOW CREATE TABLE` of an existing table can have different behaviors when the `character_set_connection` is set to different charsets. The `SHOW CREATE TABLE` of a new table created after this PR should represent the creation accurately (and stably).
4. In other situations, existing tables/partitions will not be changed.

@bb7133 and @songrijie PTAL

tests/integrationtest/t/ddl/partition.test
```diff
@@ -135,3 +135,74 @@ show create table t;
...
+set character_set_connection=gbk;
+drop table if exists t;
+create table t (col1 varbinary(16) unique key) partition by list columns(col1)
```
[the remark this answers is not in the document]
YangKeao: > I'm a bit curious here, this file is in utf8 and even the server will receive the partition values in utf8 encoding. 

The session variable `character_set_connection` is set as an attribute of the parser. With the charset specified as `gbk`, the parser will "mark" the string literal as the `gbk` charset. You can find related logic in `parser.y` by searching `parser.charset`. 

> Where are those values converted to gbk encoding?

TiDB always represents string with utf8 encoding in memory. For example, in this issue, the string "你好" is still represented in utf8 encoding in memory, but the `ValueExpr` records the charset of it is `gbk`. When the string `ValueExpr` is converted to `binary`, it'll be converted to gbk encoding and store in binary representation. If the `gbk` `ValueExpr` is converted to `utf8`, it'll actually do nothing (as the string is already represented in `utf8`) but only change the charset/collation field in `ValueExpr` and `Datum`.

For this issue, the conversion happens in `checkAndGetColumnsTypeAndValuesMatch` -> `ConvertTo`.

tests/integrationtest/r/ddl/partition.result
```diff
@@ -177,3 +177,180 @@ t	CREATE TABLE `t` (
...
+  UNIQUE KEY `col1` (`col1`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
+PARTITION BY LIST COLUMNS(`col1`)
+(PARTITION `p0` VALUES IN (_binary 0xc4e3bac3,_binary 0xced2bac3),
+ PARTITION `p1` VALUES IN (_binary 0xb4f3bcd2bac3),
+ PARTITION `p2` DEFAULT)
+insert into t values ("你好");
+select * from t partition(p0);
+col1
+���
```
tiancaiamao, line 197: It seems @Defined2014 have modify mysql-test to display the binary human-readable somehow

tiancaiamao (approved):

tiancaiamao:
Need a ddl owner's review
PTAL @wjhuang2016

tests/integrationtest/r/ddl/partition.result
Defined2014: @YangKeao Please change it to `select hex(col1) from t partition(p0);`.

bb7133 (approved):
LGTM

zimulala (approved):

PR #26098, opened 2021-07-10
executor, planner: add support for SQL_CALC_FOUND_ROWS

morgo:
### What problem does this PR solve?

Issue Number: close https://github.com/pingcap/tidb/issues/21786

Problem Summary:

The feature `SQL_CALC_FOUND_ROWS` is deprecated in MySQL, but it is unfortunately used in a number of common apps like wordpress and drupal. This causes compatibility issues with TiDB, and can leave a bad impression because such 'simple' applications fail.

The implementation disables all push-down to the coprocessor and counts rows inside TiDB. The performance requirements of this feature are not important, since it is deprecated and only provided for compatibility purposes.

### What is changed and how it works?

What's Changed:

Support for SELECT SQL_CALC_FOUND_ROWS has been added for MySQL compatibility. However, where possible it is recommended to send 2 queries to TiDB instead.

### Related changes

- PR to update `pingcap/docs`/`pingcap/docs-cn`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Increased code complexity.

### Release note 

- Support for SELECT SQL_CALC_FOUND_ROWS has been added for MySQL compatibility. However, where possible it is recommended to send 2 queries to TiDB instead.

Issue #21786, opened 2020-12-15
Support sql_calc_found_rows + select found_rows()


morgo:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently TiDB does not support `sql_calc_found_rows` correctly. To protect against incorrect use, https://github.com/pingcap/tidb/issues/19383 was added.

**Describe the feature you'd like:**

This feature request is to add correct support for `sql_calc_found_rows` + `found_rows()`. It is not a bug, since the current behavior is as intended.

**Describe alternatives you've considered:**

This feature is actually deprecated in MySQL 8.0. I suspect it might take some time until it is removed because it is still quite common in usage.

An alternative that may be considered is 'do not implement this feature'. I think that is okay, since at least now TiDB will not return incorrect results.

**Teachability, Documentation, Adoption, Migration Strategy:**

Improves MySQL compatibility, teachability and docs is easy to explain.

PR #49283, opened 2023-12-08
planner: Support SQL_CALC_FOUND_ROWS | tidb-test=pr/2263

breezewish:

Issue Number: ref https://github.com/pingcap/tidb/issues/21786

Problem Summary:

The feature SQL_CALC_FOUND_ROWS is deprecated in MySQL, but it is unfortunately used in a number of common apps like wordpress and drupal. This causes compatibility issues with TiDB, and can leave a bad impression because such 'simple' applications fail.

**The behavior of SQL_CALC_FOUND_ROWS is described as follows (according to [doc](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_found-rows)):**

A [SELECT](https://dev.mysql.com/doc/refman/8.0/en/select.html) statement may include a LIMIT clause to restrict the number of rows the server returns to the client. **In some cases, it is desirable to know how many rows the statement would have returned without the LIMIT, but without running the statement again. To obtain this row count, include an SQL_CALC_FOUND_ROWS option in the [SELECT](https://dev.mysql.com/doc/refman/8.0/en/select.html) statement, and then invoke [FOUND_ROWS()](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_found-rows) afterward**:

```
mysql> SELECT SQL_CALC_FOUND_ROWS * FROM tbl_name
    -> WHERE id > 100 LIMIT 10;
mysql> SELECT FOUND_ROWS();
```

The second [SELECT](https://dev.mysql.com/doc/refman/8.0/en/select.html) returns a number indicating how many rows the first [SELECT](https://dev.mysql.com/doc/refman/8.0/en/select.html) would have returned had it been written without the LIMIT clause.

...

SQL_CALC_FOUND_ROWS and [FOUND_ROWS()](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_found-rows) can be useful in situations when you want to restrict the number of rows that a query returns, but also determine the number of rows in the full result set without running the query again. An example is a Web script that presents a paged display containing links to the pages that show other sections of a search result. Using [FOUND_ROWS()](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_found-rows) enables you to determine how many other pages are needed for the rest of the result.

The use of SQL_CALC_FOUND_ROWS and [FOUND_ROWS()](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_found-rows) is more complex for [UNION](https://dev.mysql.com/doc/refman/8.0/en/union.html) statements than for simple [SELECT](https://dev.mysql.com/doc/refman/8.0/en/select.html) statements, because LIMIT may occur at multiple places in a [UNION](https://dev.mysql.com/doc/refman/8.0/en/union.html). It may be applied to individual [SELECT](https://dev.mysql.com/doc/refman/8.0/en/select.html) statements in the [UNION](https://dev.mysql.com/doc/refman/8.0/en/union.html), or global to the [UNION](https://dev.mysql.com/doc/refman/8.0/en/union.html) result as a whole.

**The intent of SQL_CALC_FOUND_ROWS for [UNION](https://dev.mysql.com/doc/refman/8.0/en/union.html) is that it should return the row count that would be returned without a global LIMIT**. The conditions for use of SQL_CALC_FOUND_ROWS with [UNION](https://dev.mysql.com/doc/refman/8.0/en/union.html) are:

- The SQL_CALC_FOUND_ROWS keyword must appear in the first [SELECT](https://dev.mysql.com/doc/refman/8.0/en/select.html) of the [UNION](https://dev.mysql.com/doc/refman/8.0/en/union.html).

- The value of [FOUND_ROWS()](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_found-rows) is exact only if [UNION ALL](https://dev.mysql.com/doc/refman/8.0/en/union.html) is used. If [UNION](https://dev.mysql.com/doc/refman/8.0/en/union.html) without ALL is used, duplicate removal occurs and the value of [FOUND_ROWS()](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_found-rows) is only approximate.

- If no LIMIT is present in the [UNION](https://dev.mysql.com/doc/refman/8.0/en/union.html), SQL_CALC_FOUND_ROWS is ignored and returns the number of rows in the temporary table that is created to process the [UNION](https://dev.mysql.com/doc/refman/8.0/en/union.html)

A previous impl is https://github.com/pingcap/tidb/pull/26098.

This PR uses a different approach by simply rewriting the expression.

For example, for SQL like:

```sql
SELECT SQL_CALC_FOUND_ROWS *, a FROM x WHERE ... LIMIT ...
```

We will do the followings:

1. Execute the SQL but without the LIMIT part, for example:

   ```sql
   SELECT *, a FROM x WHERE ...
   ```

2. Retrieve the rows of the SQL in step 1. The rows will be recorded in the session, which will be used for the output of `found_rows()`.

3. Finally, execute and return the result of SQL without SQL_CALC_FOUND_ROWS:

   ```sql
   SELECT *, a FROM x WHERE ... LIMIT ...
   ```

Step 2 is implemented by adding a COUNT(*) aggregator over the plan, so that some optimizations may be applied, without involving a full scan.

This approach supports all edge cases not covered by https://github.com/pingcap/tidb/pull/26098, including subqueries, unions and expressions without a limit.

- [x] TODO: Support prepared statement with `EXECUTE stmt` syntax
- [x] TODO: Disable plan cache (as we calculate rows in the planning stage)
- [x] TODO: Test with JOIN
- [x] TODO: Test with UNION
- [x] TODO: Test with SubQuery
- [x] TODO: Throw errors when SQL_CALC_FOUND_ROWS is not occurred in the first SELECT
- [x] TODO: Add unit tests

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Support `SELECT SQL_CALC_FOUND_ROWS`. However, where possible it is recommended to send 2 queries to TiDB instead.
```


pkg/expression/integration_test/integration_test.go
```diff
@@ -250,6 +233,245 @@ func TestInfoBuiltin(t *testing.T) {
...
+	tk.MustQuery("select SQL_CALC_FOUND_ROWS b,count(*) as c from t2 group by b order by c limit 1")
+	tk.MustQuery("select found_rows()").Check(testkit.Rows("2"))
+	tk.MustQuery("select b,count(*) as c from t2 group by b order by c limit 1")
+	tk.MustQuery("select found_rows()").Check(testkit.Rows("1"))
+
+	// test expressions without a FROM
+	tk.MustQuery("select SQL_CALC_FOUND_ROWS 1, 1").Check(testkit.Rows("1 1"))
+	tk.MustQuery("select found_rows()").Check(testkit.Rows("1"))
+
+	// test prepared statement
```
bb7133, line 293: I'm not sure if we need some e2e integration test, @qw4990 PTAL.

PR #25485, opened 2021-06-16
executor,planner: Add support for the HELP statement

dveeden:
### What problem does this PR solve?

Issue Number: close #33609 

This adds support for server side help: e.g. `HELP 'select'`. This make it easier to lookup documentation for statements, etc. This depends on the content of the `mysql.help_topic` table so we can later extend this with documentation about functions and variables as well if we want. 

- https://dev.mysql.com/doc/refman/8.0/en/mysql-server-side-help.html
- https://dev.mysql.com/doc/refman/8.0/en/help.html

A ChatGPT based implementation is available in #49313

### What is changed and how it works?

How it Works:

This reads the data from `mysql.help_topic`

### Related changes

- Script for generating `mysql.help_topic` data: https://github.com/pingcap/docs/pull/5796

### Check List 

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

```
mysql> INSERT INTO mysql.help_topic VALUES(1,'demo','1',"This is a demonstration of server side help\n","This is an example\n",'https://example.com/something');
Query OK, 1 row affected (0.00 sec)

mysql> help 'demo';
Name: 'demo'
Description:
This is a demonstration of server side help
Examples:
This is an example

mysql> 
```

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [x] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [x] Changes MySQL compatibility

### Release note 

```release-note
 Support for server side help was added.
```

PR #49313, opened 2023-12-09
executor, planner: Server side help, powered by OpenAI ChatGPT

### LINKED ISSUE #33609: CLI help ---  ERROR 8108 (HY000): Unsupported type *ast.HelpStmt

dveeden:
This requires the user to set a `OPENAI_API_KEY` environment variable.

Issue Number: close #33609

Problem Summary:

This adds support for server side help: e.g. `HELP 'select'`. This make it easier to lookup documentation for statements, etc. This uses ChatGPT.

- https://dev.mysql.com/doc/refman/8.0/en/mysql-server-side-help.html
- https://dev.mysql.com/doc/refman/8.0/en/help.html

A more classical implementation is in #25485

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [x] Affects user behaviors
- [ ] Contains syntax changes
- [x] Contains experimental features

### Release note

```release-note
Server side powered by ChatGPT was added.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -163,6 +163,7 @@ require (
 	go.opentelemetry.io/otel/metric v1.21.0 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20231030173426-d783a09b4405 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20231120223509-83a465c0220f // indirect
+	gopkg.in/resty.v1 v1.12.0 // indirect
 	k8s.io/utils v0.0.0-20230711102312-30195339c3c7 // indirect
 )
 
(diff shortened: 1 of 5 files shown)


dveeden:
```
mysql> help 'AUTO_RANDOM';
Name: 'AUTO_RANDOM'
Description:
AUTO_RANDOM is a feature in TiDB SQL that automatically generates a random number as the value for a specified column in a table. This feature is typically used when you need to assign a random value to a column during the insertion of a new row.

For example, suppose we have a table named `users` with columns `id` and `name`. And we want to assign a random number for the `id` column when inserting a new row. We can define the `id` column as an AUTO_RANDOM column.

```sql
CREATE TABLE users (
  id BIGINT AUTO_RANDOM PRIMARY KEY,
  name VARCHAR(50)
);
```

Now, every time we insert a new row into the `users` table without specifying a value for the `id` column, TiDB will automatically generate a random number for it.

```sql
INSERT INTO users (name) VALUES ('John Doe');
```

The resulting row in the `users` table might look like this:

```
+----+----------+
| id | name     |
+----+----------+
| 5  | John Doe |
+----+----------+
```

Note that the specific value of the `id` column will differ each time due to the random generation
```

**Comment by dveeden** [comment]
Another example:
````
mysql> help 'query hints for tiflash';
Name: 'query hints for tiflash'
Description:
Query hints in TiDB SQL allow you to provide suggestions to the query optimizer for how it should execute a specific SQL query. One use case for query hints is to specify that a particular table should be queried using TiFlash, which is a columnar storage engine in TiDB for faster analytical queries.

To use query hints for TiFlash, you can add the `tiflash` hint as a comment in your SQL query. Here is an example:

```sql
SELECT /*+ read_from_storage(tiflash[t1]) */ * FROM t1;
```

In this example, the `read_from_storage(tiflash[t1])` hint is added as a comment in the query. It instructs the optimizer to read the table `t1` from the TiFlash storage engine. This hint is particularly useful when there are multiple replicas of a table, and you want to explicitly specify which one to use.

Note that query hints are optional, and the TiDB optimizer will automatically choose the best execution plan based on statistics and cost estimations. Query hints should only be used when you have specific knowledge or requirements about the execution plan.
````

This is how this looks:
image

okJiang:
very cool~

dveeden:
While this is interesting, I don't think this is the way forward. Instead we should have reviewed documentation from the docs repo for this

Issue #48126, opened 2023-10-31
Tracking Issue for Partition Statistics


hi-rustin:
## Enhancement

In the past, there were issues with statistics information related to partition operations. Here are some examples:

1. https://github.com/pingcap/tidb/issues/39681 When we truncate a partition table, the global statistics remain unchanged.
2. https://github.com/pingcap/tidb/issues/47354 After we exchange a partition table, the global statistics remain unchanged.

This will impact how we generate the correct query plan. Therefore, we need to test all partition operations and fix any potential issues.

You can see more at: https://pingcap.feishu.cn/docx/VIW7d4I7KoXpx5xOsqRczZ1hnUc

Task list:

- [x] https://github.com/pingcap/tidb/issues/39681
- [x] https://github.com/pingcap/tidb/issues/47354
- [x] https://github.com/pingcap/tidb/issues/48182
- [x] https://github.com/pingcap/tidb/issues/48226
- [x] https://github.com/pingcap/tidb/issues/48235
- [x] https://github.com/pingcap/tidb/issues/48233
- [x] https://github.com/pingcap/tidb/issues/48231
- [ ] https://github.com/pingcap/tidb/issues/49547

hi-rustin:
Test report: https://pingcap.feishu.cn/docx/XSnOdZ51IoxtjZxP3DwcOjwinoh?from=from_copylink

hi-rustin:
Design doc: https://pingcap.feishu.cn/docx/OLCrdxHj2ojb9KxlUvUch60LnBe

hi-rustin:
We are going to use the solution 2.

hi-rustin:
The test report after fixing: https://pingcap.feishu.cn/docx/XPYIdC4Ihoc7jNxuPficBbOgn8g

PR #49331, opened 2023-12-11
statistics: do not get stats meta from cache when handling drop partition DDL event

hi-rustin:

Issue Number: ref https://github.com/pingcap/tidb/issues/48126

Problem Summary:

1. Moved the code to a separate file to have better readability.
2. Do not use the stats meta from the cache, maybe it is outdated.
3. ~~Added a TODO item. There are some bugs in ResetTableStats2KVForDrop which prevent it from storing history correctly. I will fix it in my next PR, and then I will complete this TODO.~~ After offline discussion with @time-and-fate, it's OK to put them in the different txns.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/handle/ddl/drop_partition.go
+++ b/pkg/statistics/handle/ddl/drop_partition.go
@@ -1,0 +1,72 @@
+// Copyright 2023 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ddl
+
+import (
+	"github.com/pingcap/errors"
+	"github.com/pingcap/tidb/pkg/sessionctx"
+	"github.com/pingcap/tidb/pkg/sessionctx/variable"
+	"github.com/pingcap/tidb/pkg/statistics/handle/lockstats"
+	"github.com/pingcap/tidb/pkg/statistics/handle/storage"
+	"github.com/pingcap/tidb/pkg/statistics/handle/util"
+)
+
+func (h *ddlHandlerImpl) onDropPartitions(t *util.DDLEvent) error {
+	globalTableInfo, droppedPartitionInfo := t.GetDropPartitionInfo()
+	// Note: Put all the operations in a transaction.
+	return util.CallWithSCtx(h.statsHandler.SPool(), func(sctx sessionctx.Context) error {
+		count := int64(0)
+		for _, def := range droppedPartitionInfo.Definitions {
+			// Get the count and modify count of the partition.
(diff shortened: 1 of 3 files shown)


hi-rustin (commented):
🔢 Self-check

pkg/statistics/handle/ddl/drop_partition.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+func (h *ddlHandlerImpl) onDropPartitions(t *util.DDLEvent) error {
+	globalTableInfo, droppedPartitionInfo := t.GetDropPartitionInfo()
+	// Note: Put all the operations in a transaction.
+	return util.CallWithSCtx(h.statsHandler.SPool(), func(sctx sessionctx.Context) error {
+		count := int64(0)
+		for _, def := range droppedPartitionInfo.Definitions {
+			// Get the count and modify count of the partition.
+			tableCount, _, _, err := storage.StatsMetaCountAndModifyCount(sctx, def.ID)
```
time-and-fate, line 33: Do we need to subtract the modify count from partition stats when we update the modify count as we did in the exchange partition?

hi-rustin: Considering the exchange partition logic, it seems we should it as `update + delete`. So maybe we should keep it?

time-and-fate: In my understanding, in the exchange partition logic, the modify count update consists of two parts when we remove a partition from the partitioned table:
1. We consider that we deleted this table, so we should add the row count of the partition to the global modify count.
2. Some of the original global modify count is contributed by this partition. Now that this partition is removed, we need to subtract this modify count.

hi-rustin: > 2\. Some of the original global modify count is contributed by this partition. Now that this partition is removed, we need to subtract this modify count.

But for dropped partitions, it happened before it got dropped. So I think it should count as changes to the global table. For instance, this situation more like: 
1. The partition has 1000 rows.
2. I updated 500 rows.
3. I deleted 999 rows.

"Should we consider the modify count as 500 plus 999, or just 999?"

Regarding the exchange partition logic, we include the modify count of the exchange table in the global stats. 
https://github.com/pingcap/tidb/blob/cf23666766d03ec9c2551465fbb92e0f9bba25a2/pkg/statistics/handle/ddl/exchange_partition.go#L52
Here we not only subtract the old partition's modify count, but we also add the new table's modify count.

time-and-fate: I didn't fully understand your meaning.
For the exchange partition formula, I think the removing partition contributes to the `partCount - partModifyCount` part, and the adding partition contributes to the `tableCount + tableModifyCount` part. So I think we missed the ` - partModifyCount` part here.

hi-rustin: After we discussed offline: We shouldn't drop those updates, because it happens before the drop operation.

For example: 
1. The partition has 1000 rows.
2. I deleted 1000 rows.
3. I dropped the partition.

We shouldn't update the global table's modify count to 0.

time-and-fate (approved):

hawkingrei (approved):

PR #49352, opened 2023-12-12
gcworker: resolve locks earlier than safepoint

### LINKED ISSUE #49349: Resolve locks could not based on GC safepoint
## Enhancement

In some scenarios, the user wants to set the GC lifetime longer to keep they can flashback to any point when some mistakes happen. But they may not want to resolve lock too late.

When TiDB stopped, there may lots of running transactions, many locks need to be resolved by GC or some other transactions after TiDB restarted. This will cause some latency spike for the new transaction. It will be OK after the first round GC, but if the GC life time is too large, the spike may happened in a very long time.

It may be better to set the resolve lock time earlier in GC, which will not make the latency spike take too long time.

jackysp:

Issue Number: close #49349

Problem Summary:

Make GC resolve lock at now - gcDefRunInterval

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/store/gcworker/gc_worker.go
+++ b/pkg/store/gcworker/gc_worker.go
@@ -1152,12 +1152,25 @@ func (w *GCWorker) checkUsePhysicalScanLock() (bool, error) {
 }
 
 func (w *GCWorker) resolveLocks(ctx context.Context, safePoint uint64, concurrency int, usePhysical bool) (bool, error) {
+	now, err := w.getOracleTime()
+	if err != nil {
+		logutil.Logger(ctx).Error("get Oracle time failed", zap.String("category", "gc worker"),
+			zap.String("uuid", w.uuid),
+			zap.Uint64("safePoint", safePoint),
+			zap.Error(err))
+	}
+	w.checkGCInterval()
+	resolveLockPoint := oracle.ComposeTS(now.Add(-gcDefaultRunInterval).Unix()*1000, 0)
+	if safePoint > resolveLockPoint {
+		resolveLockPoint = safePoint
+	}
+
 	if !usePhysical {
-		return false, w.legacyResolveLocks(ctx, safePoint, concurrency)
+		return false, w.legacyResolveLocks(ctx, resolveLockPoint, concurrency)
 	}
 
 	// First try resolve locks with physical scan
-	err := w.resolveLocksPhysical(ctx, safePoint)
+	err = w.resolveLocksPhysical(ctx, resolveLockPoint)
 	if err == nil {
 		return true, nil
 	}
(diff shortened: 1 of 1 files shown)


jackysp:
What are your thoughts on the matter? @MyonKeminta @AndreMouche

pkg/store/gcworker/gc_worker.go
```diff
@@ -1152,22 +1152,36 @@ func (w *GCWorker) checkUsePhysicalScanLock() (bool, error) {
...
 
 func (w *GCWorker) resolveLocks(ctx context.Context, safePoint uint64, concurrency int, usePhysical bool) (bool, error) {
+	now, err := w.getOracleTime()
+	if err != nil {
+		logutil.Logger(ctx).Error("get Oracle time failed", zap.String("category", "gc worker"),
+			zap.String("uuid", w.uuid),
+			zap.Uint64("safePoint", safePoint),
+			zap.Error(err))
+	}
+	w.checkGCInterval()
```
MyonKeminta, line 1162: What's the purposo of this call?

pkg/store/gcworker/gc_worker.go
```diff
@@ -1152,22 +1152,36 @@ func (w *GCWorker) checkUsePhysicalScanLock() (bool, error) {
...
+	resolveLockPoint := oracle.ComposeTS(now.Add(-gcDefaultRunInterval).Unix()*1000, 0)
+	if safePoint > resolveLockPoint {
+		resolveLockPoint = safePoint
+	}
```
MyonKeminta, line 1166: I'm sorry, 1) you seem to be mixing the concept of GC run interval and GC lifetime; 2) you are not considering user's configuration at all, and a user don't have choice to disable this new behavior; 3) you are breakig the constraints of alive transactions and service safepoints so that long-running transactions is possible to be killed. A previously added but later reverted change to this part is to do *soft* resolving lock at now-5min (do not resolve locks no transactions whose tts is not expired yet), and do *hard*(force) resolving lock at safepoint (which is how the current GC resolving lock is being done). It looks more proper and you may consider adopt that way.
Also consider making the soft resolve lock interval configurable.

MyonKeminta (changes requested):

pkg/store/gcworker/gc_worker.go
jackysp: It was added by a mistake.

pkg/store/gcworker/gc_worker.go
jackysp: I have to say that it is not the final version, so it doesn't have a configurable variable. I just want to know why the "soft resolve lock" reverted.

jackysp: Actually, there's another way to handle the original issue. A new SQL like `resolve locks` could be introduced for users to resolve locks on demand. However, it will increase the learning cost for users.

PR #33084, opened 2022-03-15
mpp: throw an err when mpp generate empty table scan task list

SeaRise:
### What problem does this PR solve?

Issue Number: close #33083

Problem Summary:

### What is changed and how it works?
Now mpp cannot handle empty table scan task list.
We temporarily throw an error for this case.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #49401, opened 2023-12-13
executor: support recovery mem limit error in disagg tiflash mode

### LINKED ISSUE #49574: support mpp err recovery
## Enhancement
In disaggregted tiflash situation, AutoScaler assigns `init number` of CN to one tenant(normally 1) to handle MPP query.
But when user run a big query which one node cannot handle(e.g. MemLimit exceeds), the query will give error.

But actually we make AutoScaler scale out immedicately and retry this query internally, so the query will run successfully.

guo-shaoge:

Issue Number: close #49574

Problem Summary: Support mpp error recovery. This pr only support recovery MemLimit error.

1. TiDB hold mpp result for n*MaxChunkSize rows
2. During holding result, if got mem limit error, will send recovery http request to AutoScaler(with ori CN cnt, and AutoScaler will check ori CN count to scale out new CN)
3. For now, I didn't add sys var, instead use hard code value. Because for master, mpp recovery will not take effccts. It only works for Serverless Tier(or with AutoScaler). So no need to ask PM to add new sys var, and I will add sys var when cherry pick to Serverless Tier repo
4. AutoScaler changes: https://github.com/tidbcloud/tiflash-autoscale/pull/33

### Check List

Tests 

- [x] Unit test
- [ ] Integration test

1. Create EKS cluster on AWS. Internal doc describe this.
    1. Two c5.xlarge EC2 node(4c8g for each one)
    5. Limit CN config 3G by setting `profiles.default.max_memory_usage and max_memory_usage_for_all_queries`
    6. AutoScaler config: min-pod: 1, max-pod: 4
2. Load tpch10 dataset
7. Run workload: `tiup bench tpch run -U root -D tpch10 --queries q1 -T 5 --conn-params tidb_isolation_read_engines='tiflash'`

Before this pr: `tiup bench` will report error
img_v3_0269_9d921829-ac27-4d9f-896a-384a00b5ea5g

After this pr: `tiup bench` run successfully
img_v3_0269_146fa439-9273-4243-addf-71cc5a030c8g

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/mpperr/mpp_err_recovery.go
```diff
@@ -0,0 +1,181 @@
...
+		enable:   enable,
+		handlers: []handlerImpl{newMemLimitHandlerImpl(useAutoScaler)},
+		holder:   newMPPResultHolder(holderCap, parent),
+		// Default recovery 3 time.
+		maxRecoveryCnt: 3,
+	}
+}
+
+// CanHoldResult tells whether we can insert intermediate results.
+func (m *RecoveryHandler) CanHoldResult(chk *chunk.Chunk) bool {
```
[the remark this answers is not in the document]
guo-shaoge: d

pkg/executor/mpp_gather.go
```diff
@@ -132,11 +186,48 @@ func (e *MPPGather) Next(ctx context.Context, chk *chunk.Chunk) error {
...
+			e.mppErrRecovery.ResetHolder()
+
+			continue
+		}
+
+		if tmpChk.NumRows() == 0 {
+			break
+		}
+
+		e.mppErrRecovery.HoldResult(tmpChk)
```
guo-shaoge, line 218: no ok. 
need refactor.

pkg/executor/mpp_gather.go
```diff
@@ -104,19 +160,31 @@ func (e *MPPGather) Open(ctx context.Context) (err error) {
...
-	if err != nil {
+	if err = e.setupRespIter(ctx, false); err != nil {
 		return err
 	}
-	var resp kv.Response
-	resp, e.kvRanges, err = coord.Execute(ctx)
-	if err != nil {
-		return errors.Trace(err)
+
+	enableMPPRecovery := true
```
SeaRise, line 167:
```suggestion
    // For now, mpp err recovery only support MemLimit, which is only useful when AutoScaler is used.
	enableMPPRecovery := config.GetGlobalConfig().UseAutoScaler && config.GetGlobalConfig().DisaggregatedTiFlash
```

pkg/executor/mpp_gather.go
```diff
@@ -126,17 +194,104 @@ func (e *MPPGather) buildCoordinator(planIDs []int) kv.MppCoordinator {
...
 // Next fills data into the chunk passed by its caller.
 func (e *MPPGather) Next(ctx context.Context, chk *chunk.Chunk) error {
 	chk.Reset()
 	if e.dummy {
 		return nil
 	}
-	err := e.respIter.Next(ctx, chk)
-	if err != nil {
+
+	if err := e.nextWithRecovery(ctx); err != nil {
```
SeaRise, line 273: Will MPPGather cancel prev query before retry?

guo-shaoge: If got error, cancel happens in lower layer. Check https://github.com/pingcap/tidb/blob/master/pkg/executor/internal/mpp/local_mpp_coordinator.go#L536 and https://github.com/pingcap/tidb/blob/master/pkg/executor/internal/mpp/local_mpp_coordinator.go#L536

SeaRise: ok, seems that `needTriggerFallback` should be checked too.
https://github.com/pingcap/tidb/blob/aeefec893d4b2b9550560a0011c778a953f54e16/pkg/executor/internal/mpp/local_mpp_coordinator.go#L533

SeaRise: Cannot recover if `needTriggerFallback` is true

pkg/executor/internal/mpp/local_mpp_coordinator.go
```diff
@@ -145,6 +145,9 @@ type localMppCoordinator struct {
 	enableCollectExecutionInfo bool
 	reportExecutionInfo        bool // if each mpp task needs to report execution info directly to coordinator through ReportMPPTaskStatus
 
+	// Record max task cnt among all the fragments.
+	// This is the number of nodes that involved in MPP computation.
+	maxTaskCnt int
```
SeaRise, line 150: For sql with union, there may be `sender1(node1, node2) and sender2(node3, node4)`, so the node count will be task count of sender1 + task count of sender.

SeaRise: from https://github.com/pingcap/tidb/pull/33084/files, the mpptask count must not be zero, seems `GetNodeCnt` can be removed.

pkg/executor/mpp_gather.go
```diff
@@ -104,19 +160,25 @@ func (e *MPPGather) Open(ctx context.Context) (err error) {
...
 	}
-	var resp kv.Response
-	resp, e.kvRanges, err = coord.Execute(ctx)
-	if err != nil {
-		return errors.Trace(err)
+
+	// For now, mpp err recovery only support MemLimit, which is only useful when AutoScaler is used.
+	disaggTiFlashWithAutoScaler := config.GetGlobalConfig().DisaggregatedTiFlash && config.GetGlobalConfig().UseAutoScaler
+	enableMPPRecovery := disaggTiFlashWithAutoScaler
+	holdCap := mppErrRecoveryHoldChkCap * e.Ctx().GetSessionVars().MaxChunkSize
```
SeaRise, line 170:
```suggestion
	holdCap := max(1, mppErrRecoveryHoldChkCap * e.Ctx().GetSessionVars().MaxChunkSize)
```

pkg/executor/mpp_gather.go
guo-shaoge: done

pkg/executor/internal/mpp/local_mpp_coordinator.go
guo-shaoge: For union, `sender1(node1, node2) and sender2(node3, node4)`, because `sender1` and `sender2` will be in different fragment, so `coord.maxTaskCnt` is still `max(sender1_node_num, sender2_node_num)`, because we compute `c.maxTaskCnt` for each fragment.

SeaRise: > This is the number of nodes that involved in MPP computation.
But `GetNodeCnt` means all nodes participating in mpp compute? So the node cnt here should be 4 instead of 2.

SeaRise: It seems that `getNodeCnt` is only used to check whether `node_cnt` is equal to 0. However, an error has already been thrown for task_count being equal to 0 in https://github.com/pingcap/tidb/pull/33084/files.

guo-shaoge: `GetNodeCnt()` is for update `MPPGather.nodeCnt`, so we can send http request with node info to AutoScaler.
And node cnt will be the hint of AutoScaler to scale out new CN

guo-shaoge: > > This is the number of nodes that involved in MPP computation.
> 
> But `GetNodeCnt` means all nodes participating in mpp compute? So the node cnt here should be 4 instead of 2.

Yeah, you're right. nodeCnt should be 4

SeaRise: ok, maybe the right `node_cnt = count(distinct node of scan executor)`

guo-shaoge: I changed the implementation, using `map` to record all different nodes. ptal

pkg/executor/mpp_gather.go
```diff
@@ -104,19 +161,32 @@ func (e *MPPGather) Open(ctx context.Context) (err error) {
...
+	enableMPPRecovery := disaggTiFlashWithAutoScaler
+
+	failpoint.Inject("mpp_recovery_test_force_enable", func() {
+		enableMPPRecovery = true
+	})
+
+	// No need to recovery when can fallback to tikv.
+	_, allowTiFlashFallback := e.Ctx().GetSessionVars().AllowFallbackToTiKV[kv.TiFlash]
+	if allowTiFlashFallback {
+		enableMPPRecovery = false
```
SeaRise, line 189: Seems it needs to be moved before L174, otherwise it will conflict with `mpp_recovery_test_force_enable`.
why not just
```
// For now, mpp err recovery only support MemLimit, which is only useful when AutoScaler is used.
// No need to recovery when can fallback to tikv.
enableMPPRecovery := disaggTiFlashWithAutoScaler && !allowTiFlashFallback
```

guo-shaoge: ok. I use different if branch to make the logic cleaner.

guo-shaoge: `mpp_recovery_test_force_enable` is only for case, and `allowTiFlashFallback` is default to be off.

guo-shaoge: and we can add test to ensure when `allowTiFlashFallback` is true, mpp err recovery should not enable.
and maybe rename `mpp_recovery_test_force_enable` as `mpp_recovery_test_mock_enable` is better.

guo-shaoge: done

pkg/executor/mpperr/mpp_err_recovery.go
```diff
@@ -0,0 +1,201 @@
...
+	if strings.Contains(mppErr.Error(), memLimitErrPattern) && h.useAutoScaler {
+		return true
+	}
+	return false
+}
+
+func (*memLimitHandlerImpl) doRecovery(info *RecoveryInfo) error {
+	// Ignore fetched topo, because AutoScaler will keep the topo for a while.
+	// And the new topo will be fetched when dispatch mpp task again.
+	if _, err := tiflashcompute.GetGlobalTopoFetcher().RecoveryAndGetTopo(tiflashcompute.RecoveryTypeMemLimit, info.NodeCnt); err != nil {
```
windtalker, line 164: The new topo is ready after this rpc, or do we need to do some backoff to make sure the new topo is ready during retry

pkg/executor/mpp_gather.go
```diff
@@ -126,17 +196,104 @@ func (e *MPPGather) buildCoordinator(planIDs []int) kv.MppCoordinator {
...
+			if err := e.setupRespIter(ctx, true); err != nil {
+				logutil.BgLogger().Error("setup resp iter when recovery mpp err failed", zap.Any("err", err))
+				return mppErr
+			}
+			e.mppErrRecovery.ResetHolder()
+
+			continue
+		}
+
+		if tmpChk.NumRows() == 0 {
```
windtalker, line 250: If `tmpChk.NumRows() == 0` is true but the hold is not full, the next time when call `Next()`, it will still call `nextWithRecovery` first?

pkg/planner/core/fragment.go
```diff
@@ -606,6 +610,10 @@ func (e *mppTaskGenerator) constructMPPTasksImpl(ctx context.Context, ts *Physic
 			SessionAlias:       sessionAlias,
 		}
 		tasks = append(tasks, task)
+		addr := meta.GetAddress()
+		if _, ok := e.nodeInfo[addr]; !ok {
+			e.nodeInfo[addr] = true
```
windtalker, line 615: Looks like just insert the addr to nodeInfo is ok? Why need to check it first?

pkg/executor/mpperr/mpp_err_recovery.go
guo-shaoge: RecoveryAndGetTopo() will block until new topo is ready. And if timeout, will report error

pkg/executor/mpp_gather.go
guo-shaoge: Yes, but it's, because `respIter.Next()` is ok to call even when all data is consumed.

pkg/planner/core/fragment.go
guo-shaoge: changed

pkg/executor/mpp_gather.go
guo-shaoge: Or we can add a flag like `childDone` , so `nextWithRecovery()` can return directly.

guo-shaoge: done. set `mppErrRecovery.cannotHold` in `PopFrontChk()`

pkg/executor/internal/mpp/local_mpp_coordinator.go
```diff
@@ -522,13 +523,6 @@ func (c *localMppCoordinator) receiveResults(req *kv.MPPDispatchRequest, taskMet
 				return
 			}
 
-			if err1 := bo.Backoff(tikv.BoTiKVRPC(), errors.Errorf("recv stream response error: %v", err)); err1 != nil {
-				if errors.Cause(err) == context.Canceled {
```
windtalker, line 526: Keep the logs?

windtalker (approved):
Others LGTM

pkg/executor/internal/mpp/local_mpp_coordinator.go
guo-shaoge: done

pkg/executor/mpp_gather.go
```diff
@@ -101,22 +165,28 @@ func (e *MPPGather) Open(ctx context.Context) (err error) {
...
-		return errors.Trace(err)
-	}
-	e.respIter = distsql.GenSelectResultFromResponse(e.Ctx(), e.RetFieldTypes(), planIDs, e.ID(), resp)
+
+	holdCap := mathutil.Max(32, mppErrRecoveryHoldChkCap*e.Ctx().GetSessionVars().MaxChunkSize)
+
+	disaggTiFlashWithAutoScaler := config.GetGlobalConfig().DisaggregatedTiFlash && config.GetGlobalConfig().UseAutoScaler
+	_, allowTiFlashFallback := e.Ctx().GetSessionVars().AllowFallbackToTiKV[kv.TiFlash]
+	// For now, mpp err recovery only support MemLimit, which is only useful when AutoScaler is used.
+	// For cached table, will not dispatch tasks to TiFlash, so no need to recovery.
```
SeaRise, line 180: And keep the comment about `allowTiFlashFallback`?

SeaRise (commented):
Others LGTM

pkg/executor/internal/mpp/local_mpp_coordinator.go
```diff
@@ -522,13 +523,6 @@ func (c *localMppCoordinator) receiveResults(req *kv.MPPDispatchRequest, taskMet
...
-					logutil.BgLogger().Info("stream recv timeout", zap.Error(err), zap.Uint64("timestamp", taskMeta.StartTs), zap.Int64("task", taskMeta.TaskId), zap.Int64("mpp-version", taskMeta.MppVersion))
-				} else {
-					logutil.BgLogger().Info("stream unknown error", zap.Error(err), zap.Uint64("timestamp", taskMeta.StartTs), zap.Int64("task", taskMeta.TaskId), zap.Int64("mpp-version", taskMeta.MppVersion))
-				}
-			}
```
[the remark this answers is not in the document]
guo-shaoge: i think the `backoff` is useless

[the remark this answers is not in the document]
SeaRise: ok

pkg/executor/mpp_gather.go
guo-shaoge: done

SeaRise (approved):

Issue #36503, opened 2022-07-25
tidb oom when insert 4m big value


Lily2025:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
tidb：8core 32GB
insert 4m big value with 256 Concurrency
code：
https://github.com/pingcap/endless/blob/master/pkg/workload/append.go

### 2. What did you expect to see? (Required)
tidb shoud not oom

### 3. What did you see instead (Required)
After running for a while, one tidb oom
![SgMUMT1bv3](https://user-images.githubusercontent.com/84712107/180680119-2115e8fc-a98c-413e-8d44-6cbffbc68035.jpg)
w := workload.AppendWorkload{
		DB:          db,
		Concurrency: 256,
		Tables:      256,
		PadLength:   4000000,
	}
    	Try(w.Prepare())
		w.Run(gctx)

### 4. What is your TiDB version? (Required)
./tidb-server -V
 Release Version: v6.2.0-alpha
Edition: Community
Git Commit Hash: 28a8ffcf67f9b1979920d03c29a54842314aa921
Git Branch: heads/refs/tags/v6.2.0-alpha
UTC Build Time: 2022-07-23 11:07:09
GoVersion: go1.18.2
Race Enabled: false
TiKV Min Version: 6.2.0-alpha
Check Table Before Drop: false
Store: unistore

XuHuaiyu:
reproducible script:
[main.txt](https://github.com/pingcap/tidb/files/9179435/main.txt)

XuHuaiyu:
1. The dashboard shows that inuse space is about 7GB~13GB.
 
<img width="875" alt="截屏2022-07-25 下午4 39 23" src="https://user-images.githubusercontent.com/9039012/180735201-ab388f2f-d4bf-40b4-87b5-83354a3e6df1.png">

2. We can only see ~5.6GB inuse memory space from the profile
<img width="1702" alt="截屏2022-07-25 下午4 37 59" src="https://user-images.githubusercontent.com/9039012/180735226-6b8dc2f0-a3a5-4d49-a8b1-9a0e5389029c.png">

3. We can see from the alloc_space that `GetTextToLog` allocates too many small objects, which are GCed thus we can not see from the profile inuse space.
[...]

PR #49420, opened 2023-12-13
sessionctx: change @@tidb_enable_gogc_tuner default value to 0

tiancaiamao:

Issue Number: ref #36503

Problem Summary:

We prefer stability (jitters, OOM risk, unpredictable latency etc) over performance.
In essential, gogc_tuner is trading stability for performance, it works by reducing Go GC frequency to get better throughput and improve the performance. 
Maybe in POC or when the memory usage of the workload is low, we can choose performance first, otherwise it's better to work with the default Go GC settings.

This commit change the default value of @@tidb_enable_gogc_tuner from true to false.
It only affect the new bootstraped cluster, the stored global session variable is not changed, 
so upgradation of the old clusters do not affect by this change (they can update this session variable depends on their need). 

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [X] Contains variable changes

### Release note

```release-note
Change @@tidb_enable_gogc_tuner default value to 0, tidb memory usage may reduce by this change. Only affect the newly bootstrapped cluster.
```

### DIFF
--- a/pkg/sessionctx/variable/tidb_vars.go
+++ b/pkg/sessionctx/variable/tidb_vars.go
@@ -1375,7 +1375,7 @@ const (
 	DefTiDBServerMemoryLimitSessMinSize          = 128 << 20
 	DefTiDBMergePartitionStatsConcurrency        = 1
 	DefTiDBServerMemoryLimitGCTrigger            = 0.7
-	DefTiDBEnableGOGCTuner                       = true
+	DefTiDBEnableGOGCTuner                       = false
 	// DefTiDBGOGCTunerThreshold is to limit TiDBGOGCTunerThreshold.
 	DefTiDBGOGCTunerThreshold                 float64 = 0.6
 	DefTiDBGOGCMaxValue                               = 500
--- a/pkg/sessionctx/variable/varsutil_test.go
+++ b/pkg/sessionctx/variable/varsutil_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pingcap/tidb/pkg/kv"
 	"github.com/pingcap/tidb/pkg/parser/mysql"
 	"github.com/pingcap/tidb/pkg/parser/terror"
+	"github.com/pingcap/tidb/pkg/util/gctuner"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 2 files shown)


jackysp (approved):
LGTM

bb7133:
> We prefer stability (jitters, OOM risk, unpredictable latency etc) over performance.

I agree with you about this, but could you please provide some real-scenario case that GOGC Tuner leads to stableness issue? IMO it is just designed to provide a 'smart' timeing for GC.

tiancaiamao:
See what's the attitude from the kernel developers of Go team:

> This out-of-memory avoidance led to the Go community developing its own homegrown garbage collector tuning.
> 
> The first example of such tuning is the heap ballast. In order to increase their productivity metrics while also avoiding out-of-memory errors, users sometimes pick a low GOGC value, and fool the GC into thinking that there's a fixed amount of memory live. This solution elegantly scales with GOGC: as the real live heap increases, the impact of that fixed set decreases, and GOGC's contribution to the heap size dominates. In effect, GOGC is larger when the heap is smaller, and smaller (down to its actual value) when the heap is larger. Unfortunately, this solution is not portable across platforms, and is not guaranteed to continue to work as the Go runtime changes. Furthermore, users are forced to do complex calculations and estimate runtime overheads in order to pick a heap ballast size that aligns with their memory limits.
> 
> The second example of such tuning is calling runtime/debug.FreeOSMemory at some regular interval, forcing a garbage collection to trigger sooner, usually to respect some memory limit. This case is much more dangerous, because calling it too frequently can lead a process to entirely freeze up, spending all its time on garbage collection. Working with it takes careful consideration and experimentation to be both effective and avoid serious repercussions.
> 


Ref https://github.com/golang/go/issues/48409
Ref https://github.com/golang/proposal/blob/master/design/48409-soft-memory-limit.md

We're doing all anti-best practice things

kennytm:
Is this still being worked on? I think we've encountered a real-world customer issue that the gctuner mistakenly set `$GOGC` to 500 similar to pingcap/tiflow#9762 (which eventually causes us to disable gctuner on TiCDC by default).

PR #49428, opened 2023-12-13
expression: remove useless ctx input in `newBaseBuiltinFuncWithFieldType`

### LINKED ISSUE #47958: proposal: remove `ctx` inner `ScalarFunction` and use `sessionctx.Context` as a input for all `EvalXXX` method
## Enhancement

In the current implementation, `ScalarFunction` is stateful and keeps a `ctx` field:

https://github.com/pingcap/tidb/blob/f3dfd896b5e6b6c09ebfc44294fbf6346def2189/pkg/expression/builtin.go#L49-L52

I think it's a bad practice because `ScalarFunction` also implements `Expression` which uses `sessionctx.Context` as the parameters in its `EvalXXX` methods:

https://github.com/pingcap/tidb/blob/f3dfd896b5e6b6c09ebfc44294fbf6346def2189/pkg/expression/expression.go#L117-L132

So, in the implementation of `ScalarFunction`, it will skip the parameter `ctx` and use its inner `ctx` instead, for example:

https://github.com/pingcap/tidb/blob/ecaa1c518cc9367844ebb5206f2e970461c8bf28/pkg/expression/scalar_function.go#L59-L61

It is ugly, what's worse, it may introduce some potential bugs such as #14370. So we decided to remove `ctx` field from `ScalaFunction` and always use the parameter `sessionctx.Context` to provide the context. It has some benefits:

- It is more clear which context we are using and it will reduce some bugs.
- All expressions will be immutable, it is thread-safe and easier to share.

For this issue, we should also add a `ctx` parameter for `Expression.Eval`. The old implementation forces to use the inner ctx when calling `Expression.Eval`.

### Remove inner context step-by-step

We can achieve the above goal with several steps.

**Step1: Add a new parameters `ctx` to methods in `builtinFunc`**

We can just add a new parameter `ctx` without using it. For example, we will change the signature of method `builtinFunc.evalInt`:

```
evalInt(row chunk.Row) (val int64, isNull bool, err error)
```

to 

```
evalInt(ctx sessionctx.Context, row chunk.Row) (val int64, isNull bool, err error)
```

We do not need to change the methods' implementations in this step, we'll do it in step2

**Step2: use the parameter `ctx` instead of inner context in `evalXXX` methods of `builtinFunc`***

for example:

```
func (b *builtinASCIISig) evalInt(ctx sessionctx.Context, row chunk.Row) (int64, bool, error) {
	val, isNull, err := b.args[0].EvalString(b.ctx, row)
        ....
```

will be changed to

```
func (b *builtinASCIISig) evalInt(ctx sessionctx.Context, row chunk.Row) (int64, bool, error) {
	val, isNull, err := b.args[0].EvalString(ctx, row)
        ....
```

**Step3: After all implements of `builtinFunc` uses the parameter ctx as the evaluate context, remove `baseBuiltinFunc.ctx` and move it to `ScalarFunction`**

```
// ScalarFunction is the function that returns a value.
type ScalarFunction struct {
        ...
        // add ctx here
        ctx sessionctx.Context
       ...
}
```

This is because the method `Expression.Eval` does not have `ctx` parameter, we should still cache it now.

**Step4: Add `ctx` parameter to `Expression.Eval`** 

The new `Eval` method is 

```
	// Eval evaluates an expression through a row.
	Eval(ctx sessionctx.Context, row chunk.Row) (types.Datum, error)
```

`Expression.Eval` is used everywhere, and not all have a `ctx` to pass to the `Eval` method. So we have to pass nil to `Expression.Eval` temporarily. `ScalarFunction.Eval` should handle this case:

```
// Eval implements Expression interface.
func (sf *ScalarFunction) Eval(ctx sessionctx.Context, row chunk.Row) (d types.Datum, err error) {
	if ctx == nil {
		ctx = sf.ctx
	}
        ....
```

**Step5: Refactor all places that call `Eval` to make sure the input ctx is not nil**

**Step6: Remove `ctx` field in `ScalarFunction`**

### sub tasks

- [x] #48338
- [x] #48408
- [x] #48595
- [x] #48608
- [x] #48665
- [x] #48770
- [x] #48783
- [x] #48787
- [x] #48790
- [x] #48793
- [x] #48914
- [x] #48916

lcwangchao:

Issue Number: ref #47958

Problem Summary:

After we removed ctx cache inner `ScalarFunction`. The `ctx` parameter in `newBaseBuiltinFuncWithFieldType` is useless, just remove it.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > Simple code, no need to test

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/expression/builtin.go
+++ b/pkg/expression/builtin.go
@@ -286,10 +286,7 @@ func newBaseBuiltinFuncWithFieldTypes(ctx sessionctx.Context, funcName string, a
 
 // newBaseBuiltinFuncWithFieldType create BaseBuiltinFunc with FieldType charset and collation.
 // do not check and compute collation.
-func newBaseBuiltinFuncWithFieldType(ctx sessionctx.Context, tp *types.FieldType, args []Expression) (baseBuiltinFunc, error) {
-	if ctx == nil {
-		return baseBuiltinFunc{}, errors.New("unexpected nil session ctx")
-	}
+func newBaseBuiltinFuncWithFieldType(tp *types.FieldType, args []Expression) (baseBuiltinFunc, error) {
 	bf := baseBuiltinFunc{
 		bufAllocator:           newLocalColumnPool(),
 		childrenVectorizedOnce: new(sync.Once),
--- a/pkg/expression/builtin_grouping_test.go
+++ b/pkg/expression/builtin_grouping_test.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/pingcap/tidb/pkg/sessionctx"
 	"github.com/pingcap/tidb/pkg/testkit/testutil"
 	"github.com/pingcap/tidb/pkg/types"
 	"github.com/pingcap/tidb/pkg/util/chunk"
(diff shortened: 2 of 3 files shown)


XuHuaiyu (approved):

okJiang (approved):

YangKeao (approved):
LGTM

PR #49514, opened 2023-12-16
Lightning : make pebbleDB block size configurable

### LINKED ISSUE #45037: Ligthtning : Increase io size(BlockSize) of pebble DB
## Enhancement

Well, recently, I got a question that's if we have any option to increase io size of pebble DB.
1. I found [here](https://github.com/pingcap/tidb/blob/eb35c773b512e4e00c42caf7f04ea7397d00c127/br/pkg/lightning/backend/local/local.go#L764-L769 ) is the only area changing something about LevelOptions. So, untill now, Tidb-lightning hasn't had a way to control it. 

2. What the user want is when they are hiting the IOPS limit of the EBS drives in read data from pebble into SSTs of TiKV.  In this way, "Pebble could relatively read more data at once to reduce iops"

```go
type LevelOptions struct {
// BlockRestartInterval is the number of keys between restart points
// for delta encoding of keys.
//
// The default value is 16.
BlockRestartInterval int

// BlockSize is the target uncompressed size in bytes of each table block.
//
// The default value is 4096.
BlockSize int
```

more info --> https://pingcap.zendesk.com/agent/tickets/1896

mittalrishabh:

Issue Number: close #45037

Problem Summary:
We have seen that lightning is bounded by EBS IOPS which is 16k and is taking more than 3 hours to transfer 7 TB of data. 

I have made the block size configurable through lightning config. We are setting it 128k internally. 
### Check List

Tests 
- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Its running in production with 128k size. However, i haven't tested the master code. I am relying on pingcap CI to test it. 
Side effects

I haven't seen much increase in memory and CPU after increasing the block size. 
Documentation

- [X] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/lightning/backend/local/engine.go
+++ b/br/pkg/lightning/backend/local/engine.go
@@ -1349,7 +1349,7 @@ func (w *Writer) addSST(ctx context.Context, meta *sstMeta) error {
 
 func (w *Writer) createSSTWriter() (*sstWriter, error) {
 	path := filepath.Join(w.engine.sstDir, uuid.New().String()+".sst")
-	writer, err := newSSTWriter(path)
+	writer, err := newSSTWriter(path, w.engine.config.BlockSize)
 	if err != nil {
 		return nil, err
 	}
@@ -1365,7 +1365,7 @@ type sstWriter struct {
 	logger log.Logger
 }
 
-func newSSTWriter(path string) (*sstable.Writer, error) {
+func newSSTWriter(path string, blockSize int) (*sstable.Writer, error) {
 	f, err := os.Create(path)
 	if err != nil {
 		return nil, errors.Trace(err)
@@ -1374,7 +1374,7 @@ func newSSTWriter(path string) (*sstable.Writer, error) {
 		TablePropertyCollectors: []func() pebble.TablePropertyCollector{
 			newRangePropertiesCollector,
 		},
-		BlockSize: 16 * 1024,
+		BlockSize: blockSize,
 	})
 	return writer, nil
 }
@@ -1561,7 +1561,7 @@ func (i dbSSTIngester) mergeSSTs(metas []*sstMeta, dir string) (*sstMeta, error)
 	heap.Init(mergeIter)
 
 	name := filepath.Join(dir, fmt.Sprintf("%s.sst", uuid.New()))
-	writer, err := newSSTWriter(name)
+	writer, err := newSSTWriter(name, 16*1024)
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
--- a/br/pkg/lightning/backend/local/engine.go
+++ b/br/pkg/lightning/backend/local/engine.go
@@ -642,7 +642,7 @@ func (e *Engine) ingestSSTLoop() {
 					}
 					ingestMetas := metas.metas
 					if e.config.Compact {
-						newMeta, err := e.sstIngester.mergeSSTs(metas.metas, e.sstDir)
+						newMeta, err := e.sstIngester.mergeSSTs(metas.metas, e.sstDir, e.config.BlockSize)
 						if err != nil {
 							e.setError(err)
 							return
@@ -1504,7 +1504,7 @@ func (h *sstIterHeap) Next() ([]byte, []byte, error) {
 // sstIngester is a interface used to merge and ingest SST files.
 // it's a interface mainly used for test convenience
 type sstIngester interface {
-	mergeSSTs(metas []*sstMeta, dir string) (*sstMeta, error)
+	mergeSSTs(metas []*sstMeta, dir string, blockSize int) (*sstMeta, error)
 	ingest([]*sstMeta) error
 }
 
@@ -1512,7 +1512,7 @@ type dbSSTIngester struct {
 	e *Engine
 }
 
-func (i dbSSTIngester) mergeSSTs(metas []*sstMeta, dir string) (*sstMeta, error) {
+func (i dbSSTIngester) mergeSSTs(metas []*sstMeta, dir string, blockSize int) (*sstMeta, error) {
 	if len(metas) == 0 {
 		return nil, errors.New("sst metas is empty")
 	} else if len(metas) == 1 {
@@ -1561,7 +1561,7 @@ func (i dbSSTIngester) mergeSSTs(metas []*sstMeta, dir string) (*sstMeta, error)
 	heap.Init(mergeIter)
 
 	name := filepath.Join(dir, fmt.Sprintf("%s.sst", uuid.New()))
-	writer, err := newSSTWriter(name, 16*1024)
+	writer, err := newSSTWriter(name, blockSize)
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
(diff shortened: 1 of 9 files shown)


lance6716:
Hi this is the error from CI

```
br/pkg/lightning/backend/local/local_test.go:593:31: not enough arguments in call to newSSTWriter
	have (string)
	want (string, int)
```

I will review soon

br/pkg/lightning/backend/local/engine.go
```diff
@@ -1561,7 +1561,7 @@ func (i dbSSTIngester) mergeSSTs(metas []*sstMeta, dir string) (*sstMeta, error)
 	heap.Init(mergeIter)
 
 	name := filepath.Join(dir, fmt.Sprintf("%s.sst", uuid.New()))
-	writer, err := newSSTWriter(name)
+	writer, err := newSSTWriter(name, 16*1024)
```
lyzx2001, line 1564: we can also use the block size passed in by user instead of hard-code value

br/pkg/lightning/config/config.go
```diff
@@ -1070,6 +1070,7 @@ type TikvImporter struct {
 	StoreWriteBWLimit       ByteSize `toml:"store-write-bwlimit" json:"store-write-bwlimit"`
 	// default is PausePDSchedulerScopeTable to compatible with previous version(>= 6.1)
 	PausePDSchedulerScope PausePDSchedulerScope `toml:"pause-pd-scheduler-scope" json:"pause-pd-scheduler-scope"`
+	BlockSize             int                   `toml:"block-size" json:"block-size"`
```
lyzx2001, line 1073: add a comment that the unit of `BlockSize` is KiB

lyzx2001 (commented):
also need to add the code to check whether `BlockSize` is valid before using it

lyzx2001:
> Hi this is the error from CI
> 
> ```
> br/pkg/lightning/backend/local/local_test.go:593:31: not enough arguments in call to newSSTWriter
> 	have (string)
> 	want (string, int)
> ```
> 
> I will review soon

can be solved by changing br/pkg/lightning/backend/local/local_test.go:593 to
```
writer, err := newSSTWriter(path, 16*1024)
```

br/pkg/lightning/backend/local/engine.go
mittalrishabh: this is the test code. And i couldn't access to config here

lyzx2001: This is not the test code, this is lightning merging pebble SST, and it should use the block size that user passes in, otherwise this PR does not have actual effect since `dbSSTIngester` is what really has effect. 
You can try to pass `BlockSize` through functions and struct to `dbSSTIngester` by adding arguments or fields, then here you can acess `BlockSize` using `dbSSTIngester`.

mittalrishabh: oh ok. I didn't realize it. I will make these changes.

mittalrishabh: As per the comments, 
``sstIngester is a interface used to merge and ingest SST files.
1505 // it's a interface mainly used for test convenience```

It is used only for testing. And the callers in 6.5 are only from test code. However, i see that there are additional callers in 7.5 . What is the reason of that ?

mittalrishabh: I understand it now. In our case, rows are always ordered so compact is always set to false. That is why it worked for us. I will make the proposed change. 
```657     if !tr.tableMeta.IsRowOrdered {
 658         dataEngineCfg.Local.Compact = true
 659         dataEngineCfg.Local.CompactConcurrency = 4
 660         dataEngineCfg.Local.CompactThreshold = local.CompactionUpperThreshold
 661     }

File: br/pkg/lightning/config/config.go [hunk 3]
```diff
@@ -1457,6 +1459,7 @@ func NewConfig() *Config {
 			DiskQuota:               ByteSize(math.MaxInt64),
 			DuplicateResolution:     DupeResAlgNone,
 			PausePDSchedulerScope:   PausePDSchedulerScopeTable,
+			BlockSize:               8192,
```
lance6716, line 1462: In order to keep compatibility, we should use the old effective value 16 * 1024, as in `newSSTWriter` from backend/local/engine.go line 1377

lance6716 (commented):
rest lgtm

pkg/ddl/ingest/config.go
```diff
@@ -91,6 +91,7 @@ func generateLocalEngineConfig(id int64, dbName, tbName string) *backend.EngineC
 			Compact:            true,
 			CompactThreshold:   int64(compactMemory),
 			CompactConcurrency: compactConcurrency,
+			BlockSize:          8192, // using default for DDL
```
lance6716, line 94:
```suggestion
			BlockSize:          16 * 1024, // using default for DDL
```

pkg/executor/importer/table_import.go
```diff
@@ -446,6 +446,7 @@ func (ti *TableImporter) OpenIndexEngine(ctx context.Context, engineID int32) (*
 		Compact:            threshold > 0,
 		CompactConcurrency: 4,
 		CompactThreshold:   threshold,
+		BlockSize:          8192,
```
lance6716, line 449:
```suggestion
		BlockSize:          16 * 1024,
```

lance6716 (approved):

tangenta (approved):

ywqzzy (approved):

br/pkg/lightning/config/config.go
```diff
@@ -1070,6 +1070,8 @@ type TikvImporter struct {
 	StoreWriteBWLimit       ByteSize `toml:"store-write-bwlimit" json:"store-write-bwlimit"`
 	// default is PausePDSchedulerScopeTable to compatible with previous version(>= 6.1)
 	PausePDSchedulerScope PausePDSchedulerScope `toml:"pause-pd-scheduler-scope" json:"pause-pd-scheduler-scope"`
+	// block size is in KiB
+	BlockSize int `toml:"block-size" json:"block-size"`
```
lance6716, line 1074: I just checked that we want the configuration syntax like `RegionSplitSize`

```
block-size = '16KiB'
```

https://github.com/pingcap/tidb/issues/45037#issuecomment-1685512854

We should change the type to `ByteSize`, remove the "block size is in KiB" comment and remove `* units.KiB` in local.go line 876

br/pkg/lightning/config/config.go
```diff
@@ -1070,6 +1070,8 @@ type TikvImporter struct {
 	StoreWriteBWLimit       ByteSize `toml:"store-write-bwlimit" json:"store-write-bwlimit"`
 	// default is PausePDSchedulerScopeTable to compatible with previous version(>= 6.1)
 	PausePDSchedulerScope PausePDSchedulerScope `toml:"pause-pd-scheduler-scope" json:"pause-pd-scheduler-scope"`
+	// block size is in KiB
```
lance6716, line 1073:
```suggestion
```

lance6716 (approved):

PR #49518, opened 2023-12-16
Lightning: increase backoff if split fails

### LINKED ISSUE #49517: Lightning: split backoff is not long enough, and fail when EBS BR disabled region split
## Bug Report

1. Run EBS BR while lightning is running. It might fail lightning if BR suspends region split for more than 1 - 2 min. 

Lightning to not fail if region split are suspended because of EBS BR
Lightning failed
6.5

mittalrishabh:

Issue Number: close #49517

Problem Summary:
Lightning job is failing intermittently after enable BR on prod clusters because of batch split failure. During the backup process, BR temporarily disables the schedulers for a period of 1-2 minutes. However, the Lightning job, which does not have a backoff mechanism, continues to retry the process for up to 5 times without pausing or spacing out the retries.
i am adding a exponential back off before each retry and now it will retry upto 930 seconds before if fails the job
### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects
NO

Documentation
NO
- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/lightning/backend/local/local.go
+++ b/br/pkg/lightning/backend/local/local.go
@@ -811,7 +811,7 @@ func (local *Backend) prepareAndSendJob(
 		needSplit = true
 	})
 	logger := log.FromContext(ctx).With(zap.String("uuid", engine.ID())).Begin(zap.InfoLevel, "split and scatter ranges")
-	backOffTime := 30 * time.Second
+	backOffTime := 10 * time.Second
 	maxbackoffTime := 120 * time.Second
 	for i := 0; i < maxRetryTimes; i++ {
 		failpoint.Inject("skipSplitAndScatter", func() {
--- a/br/pkg/lightning/backend/local/local.go
+++ b/br/pkg/lightning/backend/local/local.go
@@ -811,8 +811,8 @@ func (local *Backend) prepareAndSendJob(
 		needSplit = true
 	})
 	logger := log.FromContext(ctx).With(zap.String("uuid", engine.ID())).Begin(zap.InfoLevel, "split and scatter ranges")
-	// we are going to do max backoff of 15 min 30 second
 	backOffTime := 30 * time.Second
+	maxbackoffTime := 120 * time.Second
 	for i := 0; i < maxRetryTimes; i++ {
 		failpoint.Inject("skipSplitAndScatter", func() {
 			failpoint.Break()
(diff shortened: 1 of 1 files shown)


br/pkg/lightning/backend/local/local.go
```diff
@@ -1129,6 +1129,8 @@ func (local *Backend) prepareAndSendJob(
 		needSplit = true
 	})
 	logger := log.FromContext(ctx).With(zap.String("uuid", engine.ID())).Begin(zap.InfoLevel, "split and scatter ranges")
+	// we are going to do max backoff of 15 min 30 second
+	backOffTime := 30 * time.Second
```
D3Hunter, line 1133: init backoff is too large for transient error, and there's no max-backoff

D3Hunter:
some background about this pr: https://github.com/pingcap/tidb/issues/49517#issuecomment-1862018888

D3Hunter:
there are already backoff in region split
https://github.com/pingcap/tidb/blob/0719329ee71f6aef85fdaaf1193c370af936a4db/br/pkg/lightning/backend/local/localhelper.go#L171-L174

mittalrishabh:
> there are already backoff in region split， just not long enough
> 
> https://github.com/pingcap/tidb/blob/0719329ee71f6aef85fdaaf1193c370af936a4db/br/pkg/lightning/backend/local/localhelper.go#L171-L174

It is only for 4 seconds. BR can suspend split from 60 seconds to 10 min. And i don't want to increase this backoff time as it is going to add backoff in each region split which will be very expensive.

br/pkg/lightning/backend/local/local.go
mittalrishabh: i added the max backoff. BR can suspend split for 60 to 300 seconds

mittalrishabh:
can you please review this PR

br/pkg/lightning/backend/local/local.go
lance6716: I also suggest to lower the initial backoff time. If there's some transient error like network problems, sleeping 30 seconds is too long. Maybe starting from 1 second.

lance6716 (commented):
rest lgtm

br/pkg/lightning/backend/local/local.go
mittalrishabh: Lets start with 10 seconds. There is already a 1 second back off during region split and scatter.

mittalrishabh:
Let me know if you agree with 10 second initial back off

lance6716 (approved):
just noticed that `SplitAndScatterRegionByRanges` has a maximum 4 seconds sleep for 8 times, so I think here we sleep 10 seconds is good.

mittalrishabh:
I need lgtm label to checkin

lance6716:
PTAL @D3Hunter

D3Hunter (approved):

Issue #49464, opened 2023-12-14
remove input argument for `Expression.ConstItem`


lcwangchao:
## Enhancement

The method `Expression.ConstItem` returns a bool to indicate whether the expression returns a constant ignoring the input row. However, it has an input argument with type `*stmtctx.StatementContext`.

When we take a closer look at its implementation. We found the input stmtctx is required by `Constant`:

https://github.com/pingcap/tidb/blob/d076bc142886e795cdf0ec9ccb57df7033095161/pkg/expression/constant.go#L426-L428

So this function has two meanings:

1. When `UseCache` is true. This returns whether the expression is constant ignoring the `EvalContext` which means it will return a constant value at any time.
2. When `UseCache` is false. This returns whether the expression is constant only with the same `EvalContext`.

The concern of `UseCache` field is to make sure some intermediate states in `ScalarExpression` can be cached correctly. See implement of regexp: https://github.com/pingcap/tidb/blob/master/pkg/expression/builtin_regexp.go . When an expression will be cached in plan cache, it should not cache the intermediate state because some other statements may use it later.

However, it will has some problems:

1.  The semantic of `ConstItem` is confusing, and the expression is coupled with plan cache heavily.
2. We cannot cache intermediate states when using plan cache sometimes.
3. Sometimes we do not care `UseCache` and just want to know whether the expression returns a constant with same context. But the current implement forces us to check it: #49343  

### Remove input argument in `ConstItem` to decouple with plan cache

Make it clear whether the expression is a pure constant or context-level constant.

First option, introduce a new enum `ConstLevel` to tell the expression's level of const

```go
type ConstLevel unit

const (
    ConstLevelNone ConstLevel = iota
    ConstLevelConext
    ConstLevelStrict
)

type Expression interface {
    // ...
    ConstLevel() ConstLevel
    // ...
}

```

Another option, introduce two methods to indicate the const property seperately

```go
type Expression interface {
    // ...
    ConstItemStrictly() boll 
    ConstItemContextful() bool
    // ...
}
```

### Make `ScalarFunction` stateless and move cache to context

To make sure we can also cache some states when plan cache enabled, we can move the expression cache from scalar function to context:

```go
func (s *xxxFuncSig) evalInt(ctx EvalContext, row chunk.Row) (int64, bool, error) {
    var state Foo
    if s.args[0].ConstLevel > ConstLevelNone {
        state, ok := ctx.GetCache(s.cacheKey()).(Foo)
        if !ok {
[...]

PR #49584, opened 2023-12-19
expression: refactor cache logic for `builtinRegexpSubstrFuncSig`

### LINKED ISSUE #49583: refactor cache implement for function `regexp_substr`
## Enhancement

see #49464

`regexp_substr` caches some intermediate `regexp.Regexp` inside the function when some args are constant. Some issues for the current implement:

1. The cache logic is coupled with plan cahe, and it is hard to say the function is `stateless`.
2. When plan cahe is enabled, we cannot cache the object `regexp.Regexp` some times. For example, when the regex pattern is from the variable.

lcwangchao:

Issue Number: close #49583

Problem Summary:

See #49583 and #49464

1. Add `CtxID` to `StatementContext` to return a unique ID for statement context. 
2. Introduce `builtinFuncCache` to provide a cache mechanism for built function. It uses context ID to separate cache between different statements so that we can still cache some immediate states even if plan cache is enabled now.
3. Introduce `regexpNewBaseFuncSig` as a new base for regex related functions. We'll use `regexpNewBaseFuncSig` to replace the old `regexpBaseFuncSig` in the next several PRs.
4. refactor `builtinRegexpSubstrFuncSig` to use `regexpNewBaseFuncSig` to manage the cache, so that:
  - It's implement can decouple with plan cache
  - It can still cache the parsed `*regexp.Regexp` when plan cache is on and regex pattern is read from a variable

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

We did a benchmark test and the performance of `regexp_substr` increased:

```
func BenchmarkRegexpSubStr(b *testing.B) {
	store := testkit.CreateMockStore(b)
	tk := testkit.NewTestKit(b, store)
	tk.MustExec("use test")
	tk.MustExec("create table t(a varchar(256))")
	for i := 0; i < b.N; i++ {
		tk.MustExec(fmt.Sprintf("insert into t values('pingcaper%d@pingcap.com')", i))
	}
	tk.MustExec(`set @regexp='^\\w+([-+.]\\w+)*@\\w+([-.]\\w+)*\\.\\w+([-.]\\w+)*$'`)
	tk.MustExec("prepare s from 'select regexp_substr(a, ?) from t'")

	b.ResetTimer()
	tk.MustQuery("execute s using @regexp")
	b.StopTimer()
}
```

Master:

```
go test -bench='^BenchmarkRegexpSubStr$' -run='^$' -benchtime=1000000x --tags=intest
...
 1000000	      1804 ns/op
PASS
ok  	github.com/pingcap/tidb/pkg/expression/integration_test	64.611s
```

This PR:
```
go test -bench='^BenchmarkRegexpSubStr$' -run='^$' -benchtime=1000000x --tags=intest
...
 1000000	       333.3 ns/op
PASS
ok  	github.com/pingcap/tidb/pkg/expression/integration_test	64.772s
```

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
improve the performance of function `regexp_substr` in some cases when plan cache enabled
```


pkg/expression/builtin.go
```diff
@@ -1017,3 +1019,46 @@ func (b *baseBuiltinFunc) MemoryUsage() (sum int64) {
...
+	if p := c.cached.Load(); p != nil && p.ctxID == ctxID {
+		return p.item, true
+	}
+	return v, false
+}
+
+func (c *builtinFuncCache[T]) getOrInitCache(ctx EvalContext, constructCache func() (T, error)) (T, error) {
+	intest.Assert(constructCache != nil)
+	ctxID := ctx.GetSessionVars().StmtCtx.CtxID()
+	if item, ok := c.getCache(ctxID); ok {
```
xzhangxian1008, line 1043: Why we double check the cache outside of the lock protection?

pkg/sessionctx/stmtctx/stmtctx.go
```diff
@@ -425,26 +429,36 @@ type StatementContext struct {
 
 // NewStmtCtx creates a new statement context
 func NewStmtCtx() *StatementContext {
```
YangKeao, line 431: Not related with this PR. Is it possible to use `NewStmtCtxWithTimeZone(time.UTC)` to implement `NewStmtCtx`?

YangKeao (approved):
LGTM

pkg/expression/builtin_test.go
```diff
@@ -161,6 +163,71 @@ func TestDisplayName(t *testing.T) {
 	require.Equal(t, "other_unknown_func", GetDisplayName("other_unknown_func"))
 }
 
+func TestBuiltinFuncCache(t *testing.T) {
```
xzhangxian1008, line 166: I think it's better to add some concurrent tests.

pkg/expression/builtin.go
lcwangchao: To achieve a better performance, ensure a concurrent call of `getOrInitCache` will not block each other.

pkg/sessionctx/stmtctx/stmtctx.go
lcwangchao: Yes, I think it is a better way when we have more fields to init

pkg/expression/builtin.go
```diff
@@ -1017,3 +1019,46 @@ func (b *baseBuiltinFunc) MemoryUsage() (sum int64) {
...
+		return item, nil
+	}
+
+	c.Lock()
```
YangKeao, line 1047: Maybe we can implement it in CAS pattern as it's already an atomic pointer.

```go
	old := c.cached.Load()
	if old.ctxID == ctxID {
		return old.item, nil
	}

	newItem, err := constructCache()
	if err != nil {
		var def T
		return def, err
	}
	newCache := &builtinFuncCacheItem[T]{
		item:  newItem,
		ctxID: ctxID,
	}
	for !c.cached.CompareAndSwap(old, newCache) {
		old := c.cached.Load()
		if old.ctxID == ctxID {
			return old.item, nil
		}
	}

	return newItem, nil
```

But actually, I have no opinion which one is better :thinking:.

lcwangchao: The `constructCache` seems to have a probability of being called more than once even if ctxID not change (but I think it is also acceptable).

pkg/expression/builtin_test.go
lcwangchao: A new test case `TestBuiltinFuncCacheConcurrency` added

xzhangxian1008 (approved):
LGTM

XuHuaiyu (approved):

Issue #49291, opened 2023-12-08
planner: care to use errors.new and errors.errorf as your warnings and notes


AilinKid:
## Enhancement
github/pingcap/errors.git has their own encapsulation of errors.new and errors.errorf, in which their will fetch stack trace of currently places. If your generation of this errors is used like warning and note, inside which the stack trace is never necessary, so try to avoid usage of this or use another stack-less errors

AilinKid:
![image](https://github.com/pingcap/tidb/assets/15825830/f05ce04a-ec10-4f6f-8cfd-0227a659348f)
```
红色圈圈部分，我们在 v6.5.0 上是看不到的，为什么会在 v7.1.0 上成为性能问题呢（虽然生活可能整体来说占比不是很大，导致 planner 部分增长了约 （18.5-15.3）/18.5  = 17 个点）。研究发现，这部分基本上都是系统调用 caller的部分，主要是服务与 error.New() 和 errors.Errorf() 里面的填充的 stack 字段部分（stack: callers()）。这部分代码是 7.1新增用来 log warning 或者 note 用的。
```


we did find a performance problem from internal incall concluded by a lark document in planner component, you can search the keyword by `Care to use errors.New and errors.Errorf` in lark app, which is not suitable to put the link here.

stack-less error is called like `errors.NewNoStackError("msg")`

jackysp:
<img width="1823" alt="image" src="https://github.com/pingcap/tidb/assets/4352397/2d10afc9-02ee-437b-88ec-f2520c8d4fff">
In version 7.1, I saw that in the batch write scenario, the CPU usage can reach up to 25%.

PR #49600, opened 2023-12-20
executor: refactor executor pkg's warning and note generation

AilinKid:

Issue Number: close https://github.com/pingcap/tidb/issues/49291

Problem Summary:

* if a error is only generated and used like warning and note, don't bind unnecessary stack info with it. Because showExec (show warning logic) only fetch errCode and errMsg to show.
* if a error is generated in deep call chain, by now, always generating it with stack, because you couldn't know whether it will be used like error/warning/note later and outer.
* if a error is treated several ways according sqlMode/var/hint/ticker and some thing, eg: like the code below, the error can be possibly treated like an error, we couldn't control generating logic with it. 
* if this error is generated in current level, like the second case below, generating it without stack info, and wrapping stack with it when warn is returned like error, because of stack info is quite same.
```go
if err != nil {
    // deep call chain for generation of ERROR here.
    stmtCtx.IsSyncStatsFailed = true
    if variable.StatsLoadPseudoTimeout.Load() {
	logutil.BgLogger().Warn("SyncWaitStatsLoad failed", zap.Error(err))
        // treated like warning
	stmtCtx.AppendWarning(err)
	return nil
    }
    logutil.BgLogger().Error("SyncWaitStatsLoad failed", zap.Error(err))
    // treated like user error
    return err
}


if exists {
    err := infoschema.ErrTableExists.FastGenByArgs(ast.Ident{Schema: s.Table.Schema, Name: s.Table.Name})
    if s.IfNotExists {
  	e.Ctx().GetSessionVars().StmtCtx.AppendNote(err)
	return nil
    }
    return errors.Trace(err)
}
```
* error's stack info will be useful, because it can be used to quickly locate where the error comes from.
```go
func (cc *clientConn) Run(ctx context.Context) {
    // ...
    logutil.Logger(ctx).Info("command dispatched failed",
	zap.String("connInfo", cc.String()),
	zap.String("command", mysql.Command2Str[data[0]]),
	zap.String("status", cc.SessionStatusToString()),
	zap.Stringer("sql", getLastStmtInConn{cc}),
	zap.String("txn_mode", txnMode),
	zap.Uint64("timestamp", startTS),
	zap.String("err", errStrForLog(err, cc.ctx.GetSessionVars().EnableRedactLog)),  // fetch stack info here
}
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/compact_table.go
+++ b/pkg/executor/compact_table.go
@@ -87,7 +87,7 @@ func (e *CompactTableTiFlashExec) Next(ctx context.Context, chk *chunk.Chunk) er
 func (e *CompactTableTiFlashExec) doCompact(execCtx context.Context) error {
 	vars := e.Ctx().GetSessionVars()
 	if e.tableInfo.TiFlashReplica == nil || e.tableInfo.TiFlashReplica.Count == 0 {
-		vars.StmtCtx.AppendWarning(errors.Errorf("compact skipped: no tiflash replica in the table"))
+		vars.StmtCtx.AppendWarning(errors.NewNoStackErrorf("compact skipped: no tiflash replica in the table"))
 		return nil
 	}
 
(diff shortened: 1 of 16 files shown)


qw4990 (approved):

pkg/executor/compact_table.go
```diff
@@ -257,33 +257,38 @@ func (task *storeCompactTask) compactOnePhysicalTable(physicalTableID int64) (bo
 		if err != nil {
 			// Even after backoff, the request is still failed.., or the request is cancelled or timed out
 			// For example, the store is down. Let's simply don't compact other partitions.
-			warn := errors.Errorf("compact on store %s failed: %v", task.targetStore.Address, err)
+			warn := errors.NewNoStackErrorf("compact on store %s failed: %v", task.targetStore.Address, err)
 			task.parentExec.Ctx().GetSessionVars().StmtCtx.AppendWarning(warn)
+			newErr := errors.Trace(warn)
```
XuHuaiyu, line 262: what's the difference

hawkingrei (approved):

pkg/executor/compact_table.go
AilinKid: AppendWarning(warn) will be stack free

AilinKid: from the stack info itself, the only difference is the come-out line number of deepest frame

XuHuaiyu (approved):

Benjamin2037 (approved):
LGTM

PR #49625, opened 2023-12-20
planner: fix the issue that the optimizer cannot convert OUTER JOIN to INNER JOIN with nested AND/OR in some cases

### LINKED ISSUE #49616: planner: the optimizer cannot convert OUTER JOIN to INNER JOIN with nested AND/OR predicates in some cases
## Enhancement
```
create table t1 (k int, a int);
create table t2 (k int, b int, key(k));

explain select /*+ tidb_inlj(t2, t1) */ *
  from t2 left join t1 
    on t1.k=t2.k
  where 
    a>0 or (a=0 and b>0);
+--------------------------------+----------+-----------+---------------+---------------------------------------------------------------+
| id                             | estRows  | task      | access object | operator info                                                 |
+--------------------------------+----------+-----------+---------------+---------------------------------------------------------------+
| Selection_7                    | 9990.00  | root      |               | or(gt(test.t1.a, 0), and(eq(test.t1.a, 0), gt(test.t2.b, 0))) |
| └─HashJoin_8                   | 12487.50 | root      |               | left outer join, equal:[eq(test.t2.k, test.t1.k)]             |
|   ├─TableReader_14(Build)      | 9990.00  | root      |               | data:Selection_13                                             |
|   │ └─Selection_13             | 9990.00  | cop[tikv] |               | not(isnull(test.t1.k))                                        |
|   │   └─TableFullScan_12       | 10000.00 | cop[tikv] | table:t1      | keep order:false, stats:pseudo                                |
|   └─TableReader_11(Probe)      | 10000.00 | root      |               | data:TableFullScan_10                                         |
|     └─TableFullScan_10         | 10000.00 | cop[tikv] | table:t2      | keep order:false, stats:pseudo                                |
+--------------------------------+----------+-----------+---------------+---------------------------------------------------------------+
```

In the case above, the `tidb_inlj` hint doesn't take effect, because the optimizer cannot convert this OUTER JOIN into an INNER JOIN, since the predicate `a>0 or (a=0 and b>0)` doesn't pass the null-rejective check (see `isNullRejected`, `evaluateExprWithNullInNullRejectCheck`). 
But actually, the predicate is null-rejective since no matter the value of `b`, if `a` is `null`, then the eventual value of this predicate is `null`, and so all unmatched OUTER JOIN rows (whose `a` is `null`) will be eliminated by this predicate. So it's safe to convert this OUTER JOIN to an INNER JOIN, but the optimizer didn't recognize this.

This issue is caused by https://github.com/pingcap/tidb/pull/38430, our current null-rejective check is too strict.

qw4990:

Issue Number: close #49616

Problem Summary: planner: fix the issue that the optimizer cannot convert OUTER JOIN to INNER JOIN with nested AND/OR in some cases

planner: fix the issue that the optimizer cannot convert OUTER JOIN to INNER JOIN with nested AND/OR in some cases

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/integration_test.go
+++ b/pkg/planner/core/integration_test.go
@@ -1626,6 +1626,17 @@ func TestLeastGretestStringPushDownToTiFlash(t *testing.T) {
 	tk.MustQuery("explain select greatest(a, b) from t;").CheckAt([]int{0, 2, 4}, rows)
 }
 
+func TestIssue49616(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec(`use test`)
+	tk.MustExec(`create table t1 (k int, a int)`)
+	tk.MustExec(`create table t2 (k int, b int, key(k))`)
+	tk.MustHavePlan(`select /*+ tidb_inlj(t2, t1) */ *
+  from t2 left join t1 on t1.k=t2.k
+  where a>0 or (a=0 and b>0)`, `IndexJoin`)
+}
+
 func TestTiFlashReadForWriteStmt(t *testing.T) {
 	store, dom := testkit.CreateMockStoreAndDomain(t)
 	tk := testkit.NewTestKit(t, store)
--- a/pkg/planner/core/integration_test.go
+++ b/pkg/planner/core/integration_test.go
@@ -107,7 +107,7 @@ func TestAggPushDownEngine(t *testing.T) {
 		"  └─TableFullScan 10000.00 cop[tikv] table:t keep order:false, stats:pseudo"))
 }
 
-func TestIssue15110(t *testing.T) {
+func TestIssue15110And49616(t *testing.T) {
 	store := testkit.CreateMockStore(t)
 	tk := testkit.NewTestKit(t, store)
 	tk.MustExec("use test")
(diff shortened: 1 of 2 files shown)


pkg/planner/core/integration_test.go
```diff
@@ -1626,6 +1626,17 @@ func TestLeastGretestStringPushDownToTiFlash(t *testing.T) {
...
 
+func TestIssue49616(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec(`use test`)
+	tk.MustExec(`create table t1 (k int, a int)`)
+	tk.MustExec(`create table t2 (k int, b int, key(k))`)
+	tk.MustHavePlan(`select /*+ tidb_inlj(t2, t1) */ *
+  from t2 left join t1 on t1.k=t2.k
+  where a>0 or (a=0 and b>0)`, `IndexJoin`)
```
hawkingrei, line 1637: I suggest to move this test into ```pkg/planner/core/issuetest```.

pkg/planner/core/rule_predicate_push_down.go
```diff
@@ -454,6 +458,47 @@ func isNullRejected(ctx sessionctx.Context, schema *expression.Schema, expr expr
...
+	}
+	for i := 0; i < 2; i++ {
+		andFunc := isFunc(orFunc.GetArgs()[i], ast.LogicAnd)
+		if andFunc == nil {
+			continue
+		}
+		if !isNullRejected(ctx, schema, orFunc.GetArgs()[1-i]) {
+			continue // the other side should be null-rejected: null-rejected OR (... AND ...)
+		}
+		for _, andItem := range expression.FlattenCNFConditions(andFunc) {
```
time-and-fate, line 493: I think it should be `SplitCNFItems` here.
```suggestion
		for _, andItem := range expression.SplitCNFItems(andFunc) {
```

pkg/planner/core/integration_test.go
qw4990: fixed

time-and-fate (approved):

winoros (approved):

AilinKid (approved):

ghazalfamilyusa (commented):
The general solution to null filtering condition NFC is simple and wondering why we need to limit to this special case. 
NFC (A and B) = NFC(A) OR NFC(B)
NFC (A OR B) = NFC(A) AND NFC(B) 
We do have NFC logic for primitive conditions and should be easy to do this general fix.

Issue #48359, opened 2023-11-07
planner: the optimizer cannot choose the best IndexMerge plan


qw4990:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```
mysql> create table t (a int, b int, c int, key a(a), key b(b), key ac(a, c), key bc(b, c));
Query OK, 0 rows affected (0.02 sec)

mysql> explain format='verbose' select * from t where a=1 or b=1 order by c;
+----------------------------------+---------+---------+-----------+---------------------+---------------------------------------------+
| id                               | estRows | estCost | task      | access object       | operator info                               |
+----------------------------------+---------+---------+-----------+---------------------+---------------------------------------------+
| Sort_5                           | 19.99   | 5431.39 | root      |                     | test.t.c                                    |
| └─IndexMerge_14                  | 19.99   | 1019.04 | root      |                     | type: union                                 |
|   ├─IndexRangeScan_11(Build)     | 10.00   | 2035.00 | cop[tikv] | table:t, index:a(a) | range:[1,1], keep order:false, stats:pseudo |
|   ├─IndexRangeScan_12(Build)     | 10.00   | 2035.00 | cop[tikv] | table:t, index:b(b) | range:[1,1], keep order:false, stats:pseudo |
|   └─TableRowIDScan_13(Probe)     | 19.99   | 4881.56 | cop[tikv] | table:t             | keep order:false, stats:pseudo              |
+----------------------------------+---------+---------+-----------+---------------------+---------------------------------------------+
5 rows in set, 1 warning (0.00 sec)

mysql> explain format='verbose' select /*+ use_index_merge(t, ac, bc) */ * from t where a=1 or b=1 order by c;
+----------------------------------+---------+---------+-----------+-------------------------+--------------------------------------------+
| id                               | estRows | estCost | task      | access object           | operator info                              |
+----------------------------------+---------+---------+-----------+-------------------------+--------------------------------------------+
| Projection_17                    | 19.99   | 1225.69 | root      |                         | test.t.a, test.t.b, test.t.c               |
| └─IndexMerge_16                  | 19.99   | 1219.71 | root      |                         | type: union                                |
|   ├─IndexRangeScan_13(Build)     | 10.00   | 2273.08 | cop[tikv] | table:t, index:ac(a, c) | range:[1,1], keep order:true, stats:pseudo |
[...]

PR #49632, opened 2023-12-20
planner: make index merge union case aware of order property

AilinKid:

Issue Number: close https://github.com/pingcap/tidb/issues/48359

Problem Summary:

As shown in the issue, currently index merge path generation can not feel its parent's property, and this kind of generation contains pruning logic inside ahead, so for this case, since ac and bc index pathes here has the same countAfterAccess as index a and b did, so it will only keep the later one remained.

generation should aware of order property and produce the most suitable index partial pathes combination, which means we should move this generation logic a little bit later when physical enumeration take place, in which parent can ask for what kind of property they want (order by c for this case here)

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
planner: make index merge union case aware of order property
```

### DIFF
--- a/pkg/planner/core/casetest/dag/testdata/plan_suite_out.json
+++ b/pkg/planner/core/casetest/dag/testdata/plan_suite_out.json
@@ -16,7 +16,7 @@
       },
       {
         "SQL": "select * from t where (t.c > 0 and t.c < 2) or (t.c > 4 and t.c < 6) or (t.c > 8 and t.c < 10) or (t.c > 12 and t.c < 14) or (t.c > 16 and t.c < 18)",
-        "Best": "TableReader(Table(t)->Sel([or(or(and(gt(test.t.c, 0), lt(test.t.c, 2)), and(gt(test.t.c, 4), lt(test.t.c, 6))), or(and(gt(test.t.c, 8), lt(test.t.c, 10)), or(and(gt(test.t.c, 12), lt(test.t.c, 14)), and(gt(test.t.c, 16), lt(test.t.c, 18)))))]))"
+        "Best": "IndexLookUp(Index(t.c_d_e)[[1,1] [5,5] [9,9] [13,13] [17,17]], Table(t))"
       },
       {
         "SQL": "select * from t where (t.c > 0 and t.c < 1) or (t.c > 2 and t.c < 3) or (t.c > 4 and t.c < 5) or (t.c > 6 and t.c < 7) or (t.c > 9 and t.c < 10)",
--- a/pkg/planner/core/find_best_task.go
+++ b/pkg/planner/core/find_best_task.go
@@ -860,9 +860,19 @@ func (ds *DataSource) matchPropForIndexMergeAlternatives(path *util.AccessPath, 
 	// step1: match the property from all the index partial alternative paths.
 	determinedIndexPartialPaths := make([]*util.AccessPath, 0, len(path.PartialAlternativeIndexPaths))
 	usedIndexMap := make(map[int64]struct{}, 1)
-	matchIdxes := make([]int, 0, 1)
-	for _, oneItemAlternatives := range path.PartialAlternativeIndexPaths {
-		matchIdxes = matchIdxes[:0]
+	type idxWrapper struct {
+		matchIdx []int
+		pathIdx  int
+	}
+	allMatchIdxes := make([]idxWrapper, 0, len(path.PartialAlternativeIndexPaths))
+	// special logic for alternative paths:
+	// index merge:
+	//  path1: {pk, index1}
+	//  path2: {pk}
+	// if we choose pk in the first path, then path2 has no choice but pk, this will result in all single index failure.
+	// so we should collect all match prop paths down, stored as matchIdxes here.
+	for pathIdx, oneItemAlternatives := range path.PartialAlternativeIndexPaths {
+		matchIdxes := make([]int, 0, 1)
 		for i, oneIndexAlternativePath := range oneItemAlternatives {
 			// if there is some sort items and this path doesn't match this prop, continue.
 			if !noSortItem && !ds.isMatchProp(oneIndexAlternativePath, prop) {
@@ -893,20 +903,35 @@ func (ds *DataSource) matchPropForIndexMergeAlternatives(path *util.AccessPath, 
 				return cmp.Compare(lhsCountAfter, rhsCountAfter)
 			})
 		}
-
+		allMatchIdxes = append(allMatchIdxes, idxWrapper{matchIdxes, pathIdx})
+	}
+	// sort allMatchIdxes by its element length.
+	// index merge:                index merge:
+	//  path1: {pk, index1}  ==>    path2: {pk}
+	//  path2: {pk}                 path1: {pk, index1}
+	// here for the fixed choice pk of path2, let it be the first one to choose, left choice of index1 to path1.
+	slices.SortFunc(allMatchIdxes, func(a, b idxWrapper) int {
+		lhsLen := len(a.matchIdx)
+		rhsLen := len(b.matchIdx)
+		return cmp.Compare(lhsLen, rhsLen)
+	})
+	for _, matchIdxes := range allMatchIdxes {
+		// since matchIdxes are ordered by matchIdxes's length,
+		// we should use matchIdxes.pathIdx to locate where it comes from.
+		alternatives := path.PartialAlternativeIndexPaths[matchIdxes.pathIdx]
 		found := false
 		// pick a most suitable index partial alternative from all matched alternative paths according to asc CountAfterAccess,
 		// By this way, a distinguished one is better.
-		for _, oneIdx := range matchIdxes {
+		for _, oneIdx := range matchIdxes.matchIdx {
 			var indexID int64
-			if oneItemAlternatives[oneIdx].IsTablePath() {
+			if alternatives[oneIdx].IsTablePath() {
 				indexID = -1
 			} else {
-				indexID = oneItemAlternatives[oneIdx].Index.ID
+				indexID = alternatives[oneIdx].Index.ID
 			}
 			if _, ok := usedIndexMap[indexID]; !ok {
 				// try to avoid all index partial paths are all about a single index.
-				determinedIndexPartialPaths = append(determinedIndexPartialPaths, oneItemAlternatives[oneIdx].Clone())
+				determinedIndexPartialPaths = append(determinedIndexPartialPaths, alternatives[oneIdx].Clone())
 				usedIndexMap[indexID] = struct{}{}
 				found = true
 				break
@@ -915,7 +940,7 @@ func (ds *DataSource) matchPropForIndexMergeAlternatives(path *util.AccessPath, 
 		if !found {
 			// just pick the same name index (just using the first one is ok), in case that there may be some other
 			// picked distinctive index path for other partial paths latter.
-			determinedIndexPartialPaths = append(determinedIndexPartialPaths, oneItemAlternatives[0].Clone())
+			determinedIndexPartialPaths = append(determinedIndexPartialPaths, alternatives[0].Clone())
 			// uedIndexMap[oneItemAlternatives[oneIdx].Index.ID] = struct{}{} must already be colored.
 		}
 	}
--- a/pkg/planner/core/find_best_task.go
+++ b/pkg/planner/core/find_best_task.go
@@ -1022,9 +1022,12 @@ func (ds *DataSource) matchPropForIndexMergeAlternatives(path *util.AccessPath, 
 	shouldKeepCurrentFilter := path.KeepIndexMergeORSourceFilter
 	for _, path := range determinedIndexPartialPaths {
 		// If any partial path contains table filters, we need to keep the whole DNF filter in the Selection.
-		if len(path.TableFilters) > 0 || (len(path.TableFilters) != 0 && !expression.CanExprsPushDown(ds.SCtx().GetExprCtx(), path.TableFilters, ds.SCtx().GetClient(), kv.TiKV)) {
+		if len(path.TableFilters) > 0 {
+			if !expression.CanExprsPushDown(ds.SCtx().GetExprCtx(), path.TableFilters, ds.SCtx().GetClient(), kv.TiKV) {
+				// if this table filters can't be pushed down, all of them should be kept in the table side, cleaning the lookup side here.
+				path.TableFilters = nil
+			}
 			shouldKeepCurrentFilter = true
-			path.TableFilters = nil
 		}
 		// If any partial path's index filter cannot be pushed to TiKV, we should keep the whole DNF filter.
 		if len(path.IndexFilters) != 0 && !expression.CanExprsPushDown(ds.SCtx().GetExprCtx(), path.IndexFilters, ds.SCtx().GetClient(), kv.TiKV) {
--- a/pkg/planner/core/find_best_task.go
+++ b/pkg/planner/core/find_best_task.go
@@ -897,14 +897,14 @@ func (ds *DataSource) matchPropForIndexMergeAlternatives(path *util.AccessPath, 
 		// pick a most suitable index partial alternative from all matched alternative paths according to asc CountAfterAccess,
 		// By this way, a distinguished one is better.
 		for _, oneIdx := range matchIdxes {
-			indexId := oneItemAlternatives[oneIdx].Index.ID
+			indexID := oneItemAlternatives[oneIdx].Index.ID
 			if oneItemAlternatives[oneIdx].IsTablePath() {
-				indexId = -1
+				indexID = -1
 			}
-			if _, ok := usedIndexMap[indexId]; !ok {
+			if _, ok := usedIndexMap[indexID]; !ok {
 				// try to avoid all index partial paths are all about a single index.
 				determinedIndexPartialPaths = append(determinedIndexPartialPaths, oneItemAlternatives[oneIdx].Clone())
-				usedIndexMap[indexId] = struct{}{}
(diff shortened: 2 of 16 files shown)


pkg/planner/core/indexmerge_path.go
```diff
@@ -180,6 +180,10 @@ func (ds *DataSource) generateNormalIndexPartialPaths4DNF(dnfItems []expression.
 }
 
 // getIndexMergeOrPath generates all possible IndexMergeOrPaths.
+// For index merge union case，the order property from its partial
+// path can be kept and multi-way merged and output. So we don't
+// generate a concrete index merge path out, but an un-determined
+// alternatives set index merge path instead.
```
qw4990, line 186: Could you add an example here? For example, for `select ... order by ...`, the `partialAlternativePaths` can be `...`.

pkg/planner/core/find_best_task.go
```diff
@@ -835,6 +836,172 @@ func (ds *DataSource) isMatchProp(path *util.AccessPath, prop *property.Physical
 	return isMatchProp
 }
 
+func (ds *DataSource) matchPropForIndexMergeAlternatives(path *util.AccessPath, prop *property.PhysicalProperty) (*util.AccessPath, bool) {
```
qw4990, line 839: Could you add an example for this method?

pkg/planner/core/find_best_task.go
```diff
@@ -835,6 +836,172 @@ func (ds *DataSource) isMatchProp(path *util.AccessPath, prop *property.Physical
...
+				return cmp.Compare(lhsCountAfter, rhsCountAfter)
+			})
+		}
+		allMatchIdxes = append(allMatchIdxes, idxWrapper{matchIdxes, pathIdx})
+	}
+	// sort allMatchIdxes by its element length.
+	// index merge:                index merge:
+	//  path1: {pk, index1}  ==>    path2: {pk}
+	//  path2: {pk}                 path1: {pk, index1}
+	// here for the fixed choice pk of path2, let it be the first one to choose, left choice of index1 to path1.
```
time-and-fate, line 908: I'm a bit skeptical about whether we really need to deliberately avoid choosing the same index.

pkg/planner/util/path.go
```diff
@@ -51,6 +51,12 @@ type AccessPath struct {
 	// PartialIndexPaths store all index access paths.
 	// If there are extra filters, store them in TableFilters.
 	PartialIndexPaths []*AccessPath
+	// For every dnf/cnf item, there maybe several matched partial index paths to be determined later in property detecting and cost model.
+	PartialAlternativeIndexPaths [][]*AccessPath
+	ShouldBeKeptCurrentFilter    bool
+	// ShouldBeKeptCurrentFilter indicates that there are some expression inside this dnf that couldn't be pushed down, and we should keep the entire dnf above.
+	ShouldBeKeptCurrentFilterExpression expression.Expression
```
time-and-fate, line 58: Some suggestions about the comments and naming:
1. Make it clear that an access path with `PartialAlternativeIndexPaths` is a different one, and it can't have `PartialIndexPaths`, and a single index or table path also doesn't use this field.
2. Specify that `ShouldBeKeptCurrentFilter` and `ShouldBeKeptCurrentFilterExpression` are only used with `PartialAlternativeIndexPaths`, which means for the new type of access path.
3. "should be kept" sounds a bit strange to me, and so does the `CurrentFilter`. How about changing the names, like `IndexMergeSourceFilter` and `KeepIndexMergeSourceFilter`?

pkg/planner/util/path.go
```diff
@@ -118,6 +124,11 @@ func (path *AccessPath) IsTablePath() bool {
 	return path.IsIntHandlePath || path.IsCommonHandlePath
 }
 
+// IsIndexMergePath returns true if it's Index Merge Path.
+func (path *AccessPath) IsIndexMergePath() bool {
+	return len(path.PartialIndexPaths) > 0 || len(path.PartialAlternativeIndexPaths) > 0
+}
```
time-and-fate, line 130: Looks like no places are using this.

time-and-fate (commented):
I suggest we replace all `SortFunc` with `SortStableFunc` to make it deterministic.

pkg/planner/core/indexmerge_path.go
```diff
@@ -210,48 +214,55 @@ func (ds *DataSource) generateIndexMergeOrPaths(filters []expression.Expression)
...
-			partialPaths = append(partialPaths, partialPath)
-		}
-		// If all of the partialPaths use the same index, we will not use the indexMerge.
-		singlePath := true
-		for i := len(partialPaths) - 1; i >= 1; i-- {
-			if partialPaths[i].Index != partialPaths[i-1].Index {
-				singlePath = false
-				break
+			// 2.1: trade off on countAfterAccess.
+			theoreticalMinCountAfterAccessPath := ds.buildIndexMergePartialPath(partialAlternativePaths[i])
```
time-and-fate, line 241:
```suggestion
			minCountAfterAccessPath := ds.buildIndexMergePartialPath(partialAlternativePaths[i])
```
I think we don't need to specially say it's "theoretical".

pkg/planner/util/path.go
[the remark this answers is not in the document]
AilinKid: impressive, thx

pkg/planner/core/indexmerge_path.go
```diff
@@ -210,48 +214,55 @@ func (ds *DataSource) generateIndexMergeOrPaths(filters []expression.Expression)
...
-			possiblePath := ds.buildIndexMergeOrPath(filters, partialPaths, i, shouldKeepCurrentFilter)
+		// 2.2 get the theoretical whole count after access for index merge.
+		accessDNF := expression.ComposeDNFCondition(ds.SCtx(), accessConds...)
+		sel, _, err := cardinality.Selectivity(ds.SCtx(), ds.tableStats.HistColl, []expression.Expression{accessDNF}, nil)
+		if err != nil {
+			logutil.BgLogger().Debug("something wrong happened, use the default selectivity", zap.Error(err))
+			sel = SelectionFactor
+		}
+
+		if len(partialAlternativePaths) > 1 {
```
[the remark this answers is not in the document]
AilinKid: not really, `if len(indexMap) == 1 {` will fail it as they are all about single same index

pkg/planner/core/find_best_task.go
AilinKid: i think if we have a>2 or a<1, we can directly derive a $(-\infty, 1) \cup (2, +\infty)$ on a simple index range scan of a.
rather than index merge { $a1:  (-\infty, 1),  a2: (2, +\infty)$ }

old logic, just keep them identical here.

pkg/planner/core/testdata/index_merge_suite_in.json
```diff
@@ -39,5 +39,14 @@
       "select /*+ use_index_merge(t3, ia, ibc, id, ie) */ * from t3 where a > 10 and b = 20 and c < 35 and d < 45 and e = 100",
       "select /*+ use_index_merge(t4, ia, ibc, id, ie) */ * from t4 where a > 10 and b = 20 and c < 35 and d in (1,3,8,9) and e = 100"
     ]
+  },
+  {
+    "name": "TestIndexMergeWithOrderProperty",
+    "cases": [
```
qw4990, line 45: Could you add more cases:
```
-- duplicated column in DNF
where a=1 or a=2 order by c
where a=1 or a=2 or b=3 order by c
-- more than 2 columns
where a=1 or b=1 or c=1 order by d
where a=1 or b=1 or c=1 or d=1 order by e -- idx(a, e), (b, e), ...
-- composite predicates
where (a=1 and b=1) or c=1 order by d  -- idx(a, b, d), idx(c, d)
where (a=1 and b=1) or (c=1 and d=1) order by e
```

qw4990 (approved):

pkg/planner/core/testdata/index_merge_suite_in.json
AilinKid: great test

pkg/planner/core/find_best_task.go
time-and-fate: That makes sense. However, if so, do we really need this sorting logic? I can't come up with any real cases that need this logic.

pkg/planner/core/indexmerge_path.go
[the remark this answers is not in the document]
time-and-fate: I mean we write `if len(partialAlternativePaths) == 1 { continue}` there.

tests/integrationtest/r/planner/core/issuetest/planner_issue.result
```diff
@@ -113,10 +113,13 @@ explain select  /*+ use_index_merge( tbl_39) */ col_239  from tbl_39 where not(
...
+    └─IndexMerge_29	382.00	root		type: union
+      ├─Selection_25(Build)	0.05	cop[tikv]		not(in(planner__core__issuetest__planner_issue.tbl_39.col_239, 2004, 2010, 2010))
+      │ └─IndexRangeScan_24	0.14	cop[tikv]	table:tbl_39, index:PRIMARY(col_239)	range:[1994,1994], keep order:true, stats:pseudo
+      ├─Selection_27(Build)	458.26	cop[tikv]		or(lt(cast(planner__core__issuetest__planner_issue.tbl_39.col_239, double UNSIGNED BINARY), 2026), gt(cast(planner__core__issuetest__planner_issue.tbl_39.col_239, double UNSIGNED BINARY), 2011))
+      │ └─IndexRangeScan_26	477.36	cop[tikv]	table:tbl_39, index:idx_223(col_239)	range:(1996,+inf], keep order:true, stats:pseudo
+      └─TableRowIDScan_28(Probe)	382.00	cop[tikv]	table:tbl_39	keep order:false, stats:pseudo
```
time-and-fate, line 122: Looks like we are choosing a worse execution plan. I guess this should be a cost model issue?

pkg/planner/core/casetest/dag/testdata/plan_suite_out.json
```diff
@@ -16,7 +16,7 @@
       },
       {
         "SQL": "select * from t where (t.c > 0 and t.c < 2) or (t.c > 4 and t.c < 6) or (t.c > 8 and t.c < 10) or (t.c > 12 and t.c < 14) or (t.c > 16 and t.c < 18)",
-        "Best": "TableReader(Table(t)->Sel([or(or(and(gt(test.t.c, 0), lt(test.t.c, 2)), and(gt(test.t.c, 4), lt(test.t.c, 6))), or(and(gt(test.t.c, 8), lt(test.t.c, 10)), or(and(gt(test.t.c, 12), lt(test.t.c, 14)), and(gt(test.t.c, 16), lt(test.t.c, 18)))))]))"
+        "Best": "IndexLookUp(Index(t.c_d_e)[[1,1] [5,5] [9,9] [13,13] [17,17]], Table(t))"
```
time-and-fate, line 19: Why is this changed? The old and the new execution plans both don't contain `IndexMerge`.

pkg/planner/core/indexmerge_intersection_test.go
```diff
@@ -43,6 +43,40 @@ func TestPlanCacheForIntersectionIndexMerge(t *testing.T) {
...
+func TestIndexMergeWithOrderProperty(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("drop table if exists t")
+	tk.MustExec("create table t (a int, b int, c int, d int, e int, key a(a), key b(b), key c(c), key ac(a, c), key bc(b, c), key ae(a, e), key be(b, e)," +
+		" key abd(a, b, d), key cd(c, d))")
+	tk.MustExec("create table t2 (a int, b int, c int, key a(a), key b(b), key ac(a, c))")
+
+	tk.MustExec("")
```
time-and-fate, line 55: ?

pkg/planner/core/indexmerge_intersection_test.go
```diff
@@ -43,6 +43,40 @@ func TestPlanCacheForIntersectionIndexMerge(t *testing.T) {
...
+
+	var (
+		input  []string
+		output []struct {
+			SQL    string
+			Plan   []string
+			Result []string
```
time-and-fate, line 62: We didn't insert any data. So I think we can remove the `Result` field to make the test result cleaner.

pkg/planner/core/find_best_task.go
AilinKid: sorting according to valid candidate length here can avoid the miss-pick across paths.
 
in the deriveStats, we already checked the simple index rule before, when generating a concrete path here, we should successfully generate it out, rather than “can't find a proper physical plan” caused by miss-pick here, saying 	
```
// index merge:                index merge:
//  path1: {pk, index1}  ==>    path2: {pk}
//  path2: {pk}                 path1: {pk, index1}
```
this index merge was validated before, when generating concrete paths, we should choose a path for path2 first, otherwise when picking pk for path1 is done, path2 has no choices then. (error)

pkg/planner/core/indexmerge_test.go
```diff
@@ -39,11 +39,19 @@ func getIndexMergePathDigest(paths []*util.AccessPath, startIndex int) string {
 		}
 		path := paths[i]
 		idxMergeDisgest += "{Idxs:["
-		for j := 0; j < len(path.PartialIndexPaths); j++ {
+		for j := 0; j < len(path.PartialAlternativeIndexPaths); j++ {
```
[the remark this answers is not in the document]
AilinKid: this functionality is used in test: TestIndexMergePathGeneration, which has made some assertions with the logical plan after deriveStats, where PartialIndexPaths will be nil according to the design in this pull request

tests/integrationtest/r/planner/core/issuetest/planner_issue.result
AilinKid: maybe this just show respect to the hint /*+ use_index_merge( tbl_39) */

pkg/planner/core/indexmerge_path.go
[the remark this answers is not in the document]
AilinKid: make sense

pkg/planner/core/find_best_task.go
```diff
@@ -835,6 +836,235 @@ func (ds *DataSource) isMatchProp(path *util.AccessPath, prop *property.Physical
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			// just pick the same name index (just using the first one is ok), in case that there may be some other
+			// picked distinctive index path for other partial paths latter.
+			determinedIndexPartialPaths = append(determinedIndexPartialPaths, alternatives[matchIdxes.matchIdx[0]].Clone())
+			// uedIndexMap[oneItemAlternatives[oneIdx].Index.ID] = struct{}{} must already be colored.
+		}
```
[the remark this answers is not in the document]
AilinKid: //  path1: {pk, index1}  ==>    path1: {index1}
//  path2: {pk}              ==>    path2: {pk}  
//  path3: {pk}              ==>    path3: {pk}  

this case is allowable, when picking path for path3, when pk is already colored, we can also pick pk as well.

pkg/planner/core/casetest/dag/testdata/plan_suite_out.json
AilinKid: i'm quite strange about this, after debug inside i find some difference in the estimation of selectivity
image
image
and finally leading the ds.statsCount difference about 1250 vs 50

pkg/planner/util/path.go
```diff
@@ -51,6 +51,19 @@ type AccessPath struct {
...
 	PartialIndexPaths []*AccessPath
+
+	// ************************************************** special field below *********************************************************
+	// For every dnf/cnf item, there maybe several matched partial index paths to be determined later in property detecting and cost model.
+	// when PartialAlternativeIndexPaths is not empty, it means a special state for index merge path, and it can't have PartialIndexPaths
+	// at same time. Normal single index or table path also doesn't use this field.
+	PartialAlternativeIndexPaths [][]*AccessPath
+	// ShouldBeKeptCurrentFilter and ShouldBeKeptCurrentFilterExpression are only used with PartialAlternativeIndexPaths, which means for
+	// the new state/type of access path. (undetermined index merge path)
+	KeepIndexMergeORSourceFilter bool
```
time-and-fate, line 62: Please update the comments to be consistent with the new variable name.

pkg/planner/core/find_best_task.go
time-and-fate: I mean if there are real examples like SQLs?
In such a case, I think, can we just choose a single-index plan on `pk`, or just allow using two partial paths using the same index?

tests/integrationtest/r/planner/core/issuetest/planner_issue.result
time-and-fate: Yes. I think you are right.

time-and-fate (commented):
Others LGTM.

pkg/planner/core/find_best_task.go
AilinKid: > I mean if there are real examples like SQLs? 
the modification here comes from an old unit test case failure, so I changed the logic like this.

the main reason is that: the user specified an index merge hint, then we pruned other paths in deriveStats there, the only left path is validated for index merge choosing (not all a simple index, otherwise, index merge hint is inapplicable right?), so the concrete path should successfully be chosen by a kind of order that we want before.

time-and-fate (approved):

PR #49667, opened 2023-12-21
session, chunk: refactor re-used chunk allocator in executor

### LINKED ISSUE #49661: re-organize the reuse chunk pool implementation
## Enhancement

Sometimes, we re-used the chunk pool to avoid re-allocating memories across multiple statements. This chunk pool located in the "connection", and is set into session context when it's possible to re-use the allocator. This part of code is quite confusing:

1. `chunk.Allocator` is not safe to call concurrently. Therefore, we always use `GetNewChunkWithCapacity` to allocate chunks from the allocator, and it'll use `(*SessionVars).ChunkPool.mu` to protect the allocator inside the `BaseExecutor`. It's reasonable because usually all `BaseExecutor` just used the `ChunkPool` in session context. However, just from the code, the relationship between the `(*SessionVars).ChunkPool.mu` and the `BaseExecutor.Alloc` is not significant.
2. We expect to remove session context from the `BaseExecutor`, so we have to remove the `GetNewChunkWithCapacity` function call in the implementation of `BaseExecutor`.

And also, we'll need to use a session variable to tell whether the re-use chunk pool optimization is enabled. Also, we have a read-only session variable to tell the users whether last statement used the re-used chunk :facepalm: . It's a little complicated.

YangKeao:

Issue Number: close #49661

Problem Summary:

I considered three features:

1. The allocator in the `BaseExecutor` should be able to allocate concurrently.
2. The session variable `tidb_enable_reuse_chunk` should be able to set whether to reuse chunks between statements.
3. The session variable `last_sql_use_alloc` should tell the user whether the last query used the cached chunk.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/sessionctx/variable/session.go
```diff
@@ -1633,35 +1621,38 @@ func (s *SessionVars) GetUseChunkAlloc() bool {
...
 		return
 	}
 
-	// If an error is reported, re-apply for alloc
-	// Prevent the goroutine left before, affecting the execution of the next sql
-	// issuse 38918
-	s.ChunkPool.mu.Lock()
-	s.ChunkPool.Alloc = nil
-	s.ChunkPool.mu.Unlock()
+	s.chunkPool = nil
```
lcwangchao, line 1655: Is it safe to remove lock here？

YangKeao: The usage of `.chunkPool` field is quite limited. After analyzing the codes, I think it's fine.

I've also discussed with @keeplearning20221 (who is the original author of the re-used chunk). He also agrees it's safe to do so.

lcwangchao (approved):

pkg/executor/join.go
```diff
@@ -332,7 +332,7 @@ func (w *buildWorker) fetchBuildSideRows(ctx context.Context, chkCh chan<- *chun
 		if w.hashJoinCtx.finished.Load() {
 			return
 		}
-		chk := sessVars.GetNewChunkWithCapacity(w.buildSideExec.RetFieldTypes(), sessVars.MaxChunkSize, sessVars.MaxChunkSize, w.hashJoinCtx.allocPool)
+		chk := w.hashJoinCtx.allocPool.Alloc(w.buildSideExec.RetFieldTypes(), sessVars.MaxChunkSize, sessVars.MaxChunkSize)
```
tiancaiamao, line 335: By the way, the use of `MaxChunkSize` is not reasonable, we met Go GC issues when the column length is varchar(1500)/varchar(3000)... in that case the chunk object size amplify too much and cause high pressue on Go GC.

tiancaiamao: We can take a close look of all `sessVars.MaxChunkSize` chunk allocating later.

pkg/sessionctx/variable/session.go
```diff
@@ -1606,19 +1600,13 @@ func (s *SessionVars) IsPlanReplayerCaptureEnabled() bool {
...
 
-// GetNewChunkWithCapacity Attempt to request memory from the chunk pool
-// thread safety
-func (s *SessionVars) GetNewChunkWithCapacity(fields []*types.FieldType, capacity int, maxCachesize int, pool chunk.Allocator) *chunk.Chunk {
-	if pool == nil {
-		return chunk.New(fields, capacity, maxCachesize)
+// GetChunkAllocator returns a vaid chunk allocator.
+func (s *SessionVars) GetChunkAllocator() chunk.Allocator {
+	if s.chunkPool == nil {
+		return chunk.NewEmptyAllocator()
```
tiancaiamao, line 1606: Can we retun something like a `DefaultEmptyAllocator` instead?  (see also `context.Background()`)
Creating a `&emptyAllocator{}` is unnecessary and it increase object allocating

pkg/util/chunk/alloc_test.go
```diff
@@ -289,3 +291,68 @@ func TestColumnAllocatorCheck(t *testing.T) {
...
+	alloc := NewSyncAllocator(NewAllocator())
+
+	wg := &sync.WaitGroup{}
+	for i := 0; i < 1000; i++ {
+		wg.Add(1)
+		go func() {
+			for j := 0; j < 1000; j++ {
+				chk := alloc.Alloc(fieldTypes, 5, 100)
+				require.NotNil(t, chk)
+			}
```
tiancaiamao, line 352: `alloc.Reset()`?

tiancaiamao (commented):
Could you please verify this change does not break  `make bench-test TO=/tmp/2023-12-25.json` for me?

pkg/executor/join.go
YangKeao: Do you think it would be better to also use `BaseExecutor.InitCap` and `BaseExecutor.MaxChunkSize` here? Though the current context doesn't have this information, we can pass them (to the `hashJoinCtx`).

pkg/util/chunk/alloc_test.go
YangKeao: Done

YangKeao:
> Could you please verify this change does not break `make bench-test TO=/tmp/2023-12-25.json` for me?

Do you mean `make bench-daily`? `make bench-daily` doesn't have significant difference between `6bbc6b` and `821993`(the based master commit).

pkg/executor/join.go
tiancaiamao: I haven't check the code for how `MaxChunkSize` is used yet, so no suggestions.

tiancaiamao (approved):

PR #49705, opened 2023-12-22
executor: Support Parallel Sort

### LINKED ISSUE #49217: Support parallel sort
## Feature Request

xzhangxian1008:

Issue Number: close #49217

Problem Summary:

#### Golang Builtin Benchmark

```
Unparallel Sort:
BenchmarkSortExec/(rows:300000,_orderBy:[0_1],_ndvs:_[0_0])-72       254143338 ns/op
BenchmarkSortExec/(rows:300000,_orderBy:[0_1],_ndvs:_[1_0])-72       355555629 ns/op
BenchmarkSortExec/(rows:300000,_orderBy:[0],_ndvs:_[1_0])-72          45852456 ns/op
BenchmarkSortExec/(rows:300000,_orderBy:[1],_ndvs:_[1_0])-72         254146326 ns/op
BenchmarkSortExec/(rows:300000,_orderBy:[0_1],_ndvs:_[10000_0])-72   296893378 ns/op
BenchmarkSortExec/(rows:300000,_orderBy:[0],_ndvs:_[10000_0])-72     219969174 ns/op
BenchmarkSortExec/(rows:300000,_orderBy:[1],_ndvs:_[10000_0])-72     252317562 ns/op
average                                                              239839694 ns/op

Parallel Sort:
BenchmarkSortExec/(rows:300000,_orderBy:[0_1],_ndvs:_[0_0])-72       139268537 ns/op
BenchmarkSortExec/(rows:300000,_orderBy:[0_1],_ndvs:_[1_0])-72       158654422 ns/op
BenchmarkSortExec/(rows:300000,_orderBy:[0],_ndvs:_[1_0])-72          35870240 ns/op
BenchmarkSortExec/(rows:300000,_orderBy:[1],_ndvs:_[1_0])-72         128380614 ns/op
BenchmarkSortExec/(rows:300000,_orderBy:[0_1],_ndvs:_[10000_0])-72   144901004 ns/op
BenchmarkSortExec/(rows:300000,_orderBy:[0],_ndvs:_[10000_0])-72     118420609 ns/op
BenchmarkSortExec/(rows:300000,_orderBy:[1],_ndvs:_[10000_0])-72     119021069 ns/op
average                                                              120645213 ns/op

Increase                                                             98.80%
```

#### TPCH10 Benchmark

Dataset: tpch10
Nodes: 1\*tidb, 3\*tikv
sql1: explain analyze select L_COMMENT, L_EXTENDEDPRICE from lineitem where L_SUPPKEY > 95000 order by L_COMMENT desc, L_EXTENDEDPRICE asc;
sql2: explain analyze select * from lineitem where L_SUPPKEY > 95000 order by L_COMMENT desc, L_EXTENDEDPRICE asc;

  | Sql1 | Sql2
-- | -- | --
Unparallel | 18.19s | 26.69s
Parallel | 5.39s | 11.43s
Increase | 237.48% | 133.51%

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/sortexec/parallel_sort_worker.go
```diff
@@ -0,0 +1,224 @@
...
+	return p.lessRowFunc(p.rowBuffer[i], p.rowBuffer[j])
+}
+
+func (p *parallelSortWorker) sortChunkAndGetSortedRows(chk *chunk.Chunk) sortedRows {
+	rowNum := chk.NumRows()
+	p.rowBuffer = make(sortedRows, rowNum)
+	for i := 0; i < rowNum; i++ {
+		p.rowBuffer[i] = chk.GetRow(i)
+	}
+	sort.Slice(p.rowBuffer, p.keyColumnsLess)
```
hawkingrei, line 123: use ```slices.SortFunc```. It is faster.

xzhangxian1008: > use `slices.SortFunc`. It is faster.

get

SeaRise:
Please add a performance comparison report

pkg/util/chunk/mpmc_queue.go
```diff
@@ -0,0 +1,152 @@
...
+// ChunkWithMemoryUsage contains chunk and memory usage.
+// However, some of memory usage may also come from other place,
+// not only the chunk's memory usage.
+type ChunkWithMemoryUsage struct {
+	Chk         *Chunk
+	MemoryUsage int64
+}
+
+// MPMCQueue means multi producer and multi consumer.
+type MPMCQueue struct {
```
SeaRise, line 52: why not using `channel` directly

pkg/executor/sortexec/sort.go
```diff
@@ -97,30 +138,59 @@ func (e *SortExec) Open(ctx context.Context) error {
...
-// Sort constructs the result following these step:
+// Sort constructs the result following these step in unparallel mode:
 //  1. Read as mush as rows into memory.
 //  2. If memory quota is triggered, sort these rows in memory and put them into disk as partition 1, then reset
 //     the memory quota trigger and return to step 1
 //  3. If memory quota is not triggered and child is consumed, sort these rows in memory as partition N.
 //  4. Merge sort if the count of partitions is larger than 1. If there is only one partition in step 4, it works
 //     just like in-memory sort before.
+//
+// TODO add introduction of parallel mode
```
SeaRise, line 166: it would be better to add the introduction in this PR like https://github.com/pingcap/tidb/blob/4ed493dcc1e3d55dd5af46db49d0b6cd34a24c07/pkg/executor/projection.go#L125-L181

pkg/executor/sortexec/parallel_sort_worker.go
```diff
@@ -0,0 +1,189 @@
...
+
+		// Memory usage of the chunk has been tracked at the producer side.
+		chk, res := p.mpmcQueue.Pop()
+		if res != chunk.OK {
+			// Check if the queue is closed by an error.
+			err := p.checkError()
+			return err == nil
+		}
+
+		p.totalMemoryUsage += chk.MemoryUsage
```
SeaRise, line 90: Where to subtract memory used?
why need `totalMemoryUsage`

pkg/executor/sortexec/parallel_sort_worker.go
```diff
@@ -0,0 +1,189 @@
...
+
+		sortedRows := p.sortChunkAndGetSortedRows(chk.Chk)
+		p.globalSortedRowsQueue.add(sortedRows)
+	}
+}
+
+func (p *parallelSortWorker) keyColumnsLess(i, j chunk.Row) int {
+	if p.timesOfRowCompare >= SignalCheckpointForSort {
+		// Trigger Consume for checking the NeedKill signal
+		p.memTracker.Consume(1)
```
SeaRise, line 100: some hacks...
Wouldn't `memory_tracker.consume` be triggered when merging two rows?

pkg/util/chunk/mpmc_queue.go
xzhangxian1008: > why not using `channel` directly

okk, I think `channel` is better.

pkg/executor/sortexec/parallel_sort_worker.go
xzhangxian1008: > Where to subtract memory used? why need `totalMemoryUsage`

Spill action needs this usage, so that it can unconsume some memory after spill.

pkg/executor/sortexec/sort.go
xzhangxian1008: > it would be better to add the introduction in this PR like
> 
> https://github.com/pingcap/tidb/blob/4ed493dcc1e3d55dd5af46db49d0b6cd34a24c07/pkg/executor/projection.go#L125-L181

okk

xzhangxian1008:
> Please add a performance comparison report

okk

pkg/executor/sortexec/sort.go
```diff
@@ -336,11 +474,112 @@ func (e *SortExec) fetchRowChunks(ctx context.Context) error {
...
+func (e *SortExec) fetchChunksParallel(ctx context.Context) error {
+	workerNum := len(e.Parallel.workers)
+
+	// Wait for the finish of all workers
+	workersWaiter := sync.WaitGroup{}
+
+	// Add before the start of goroutine to avoid that the counter is minus to negative.
+	workersWaiter.Add(workerNum + 1)
+
+	// Fetch chunks from child and put chunks into MPMCQueue
```
SeaRise, line 491:
```suggestion
	// Fetch chunks from child and put chunks into chunkChannel
```

pkg/executor/sortexec/sort.go
```diff
@@ -336,11 +474,112 @@ func (e *SortExec) fetchRowChunks(ctx context.Context) error {
...
+	workersWaiter.Wait()
+	err := e.checkErrorForParallel()
+	if err != nil {
+		return err
+	}
+	e.fetchResultFromQueue()
+	return nil
+}
+
+// Fetch chunks from child and put chunks into MPMCQueue
```
SeaRise, line 510:
```suggestion
// Fetch chunks from child and put chunks into chunkChannel
```

pkg/executor/sortexec/sort.go
```diff
@@ -336,11 +474,112 @@ func (e *SortExec) fetchRowChunks(ctx context.Context) error {
...
+		}
+
+		rowCount := chk.NumRows()
+		if rowCount == 0 {
+			break
+		}
+
+		chkWithMemoryUsage := &chunkWithMemoryUsage{
+			Chk:         chk,
+			MemoryUsage: chk.MemoryUsage() + chunk.RowSize*int64(rowCount),
```
SeaRise, line 535: why `chunk.RowSize*int64(rowCount)` is needed, doesn't `chk.MemoryUsage()` contains it?

pkg/executor/sortexec/sort.go
```diff
@@ -336,11 +474,112 @@ func (e *SortExec) fetchRowChunks(ctx context.Context) error {
...
+		e.Parallel.err = err
+	}
+}
+
+func (e *SortExec) fetchResultFromQueue() {
+	sortedRowsNum := e.Parallel.globalSortedRowsQueue.getSortedRowsNumNoLock()
+	if sortedRowsNum > 1 {
+		panic("Sort is not completed.")
+	}
+
```
SeaRise, line 572:
```suggestion

	idx = 0
```

pkg/executor/sortexec/sort.go
```diff
@@ -357,20 +596,18 @@ func (e *SortExec) buildKeyColumns() {
 	}
 }
 
-func (e *SortExec) lessRow(rowI, rowJ chunk.Row) bool {
+func (e *SortExec) lessRow(rowI, rowJ chunk.Row) int {
```
wshwsh12, line 599: Why modify this function?

pkg/executor/sortexec/parallel_sort_worker.go
```diff
@@ -0,0 +1,217 @@
...
+	p.timesOfRowCompare++
+
+	return p.lessRowFunc(i, j)
+}
+
+func (p *parallelSortWorker) run() {
+	defer func() {
+		if r := recover(); r != nil {
+			processPanicAndLog(p.processError, r)
+			p.tryToCloseChunkChannel()
```
wshwsh12, line 206: WHY sortWorker need to call this?

pkg/executor/sortexec/sort.go
xzhangxian1008: > why `chunk.RowSize*int64(rowCount)` is needed, doesn't `chk.MemoryUsage()` contains it?

Because we will extract rows from chunks and save them in a field with `[]chunk.Row` type, memory consumed by `chunk.Row` is also needed to be tracked.

pkg/executor/sortexec/sort.go
xzhangxian1008: > 

Why we need to add `idx = 0`?

pkg/executor/sortexec/sort.go
```diff
@@ -52,36 +58,67 @@ type SortExec struct {
...
 
-	// sortPartitions is the chunks to store row values for partitions. Every partition is a sorted list.
-	sortPartitions []*sortPartition
+	IsUnparallel bool
+
+	Unparallel struct {
+		Idx int
+
+		// rowChunks is the chunks to store row values.
+		rowChunks *chunk.SortedRowContainer
```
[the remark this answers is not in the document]
xzhangxian1008: > Remove this.

done

pkg/executor/sortexec/sort.go
```diff
@@ -52,36 +58,67 @@ type SortExec struct {
...
+
+		// sortPartitions is the chunks to store row values for partitions. Every partition is a sorted list.
+		sortPartitions []*sortPartition
+
+		// PartitionList is the chunks to store row values for partitions. Every partition is a sorted list.
+		PartitionList []*chunk.SortedRowContainer
```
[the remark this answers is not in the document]
xzhangxian1008: > Remove this.

done

pkg/executor/sortexec/sort.go
xzhangxian1008: > Why modify this function?

Because `slices.SortFunc` is faster and it needs a compare function returning int type.

pkg/executor/sortexec/parallel_sort_worker.go
xzhangxian1008: > WHY sortWorker need to call this?

In case all workers panic, no one will consume this chunkChannel and data fetcher will be blocked in the channel when the channel is full. So, we need to close this channel when worker panic.

pkg/executor/sortexec/sort.go
SeaRise: `getRows` is called in `sortChunkAndGetSortedRows`, so `memory_tracker.consume(chunk.RowSize*int64(rowCount))` should be called in `sortChunkAndGetSortedRows` instead of here?

pkg/executor/sortexec/sort.go
SeaRise: In golang, is the default value of int 0? If so, there is no need to call `idx = 0` here.

pkg/executor/sortexec/parallel_sort_worker.go
wshwsh12: Closing this channel will make data fetcher panic

pkg/executor/sortexec/parallel_sort_worker.go
```diff
@@ -0,0 +1,213 @@
...
+				randNum := rand.Int31n(10000)
+				if randNum < 100 {
+					time.Sleep(5 * time.Millisecond)
+				}
+			}
+		}
+	})
+}
+
+// Fetching chunks from MPMCQueue and sort them.
```
SeaRise, line 93:
```suggestion
// Fetching chunks from chunkChannel and sort them.
```

SeaRise:
> > Please add a performance comparison report
> 
> okk

any updated?

pkg/executor/sortexec/sort_util.go
```diff
@@ -48,3 +91,39 @@ func (h *multiWayMerge) Pop() interface{} {
...
+	return res
+}
+
+// chunkWithMemoryUsage contains chunk and memory usage.
+// However, some of memory usage may also come from other place,
+// not only the chunk's memory usage.
+type chunkWithMemoryUsage struct {
+	Chk         *chunk.Chunk
+	MemoryUsage int64
+}
```
SeaRise, line 118: why need this?
Seems `chunkWithMemoryUsage::MemoryUsage` is not used

xzhangxian1008:
> > > Please add a performance comparison report
> > 
> > 
> > okk
> 
> any updated?

updated, and here is the [doc](https://pingcap.feishu.cn/wiki/ALqkwLFnui9CXCkrTjacyR5rn3b)

pkg/executor/sortexec/sort.go
xzhangxian1008: > In golang, is the default value of int 0? If so, there is no need to call `idx = 0` here.

Yes, the default value of int in golang is 0.

pkg/executor/sortexec/sort.go
xzhangxian1008: > `getRows` is called in `sortChunkAndGetSortedRows`, so `memory_tracker.consume(chunk.RowSize*int64(rowCount))` should be called in `sortChunkAndGetSortedRows` instead of here?

I think it's better to call at here, as memory consumed by `chunk.Row` must be allocated in the future and we can call `Consume()` only once.

pkg/executor/sortexec/parallel_sort_worker.go
```diff
@@ -0,0 +1,213 @@
...
+
+import (
+	"math/rand"
+	"sync"
+	"time"
+
+	"github.com/pingcap/failpoint"
+	"github.com/pingcap/tidb/pkg/util/chunk"
+	"github.com/pingcap/tidb/pkg/util/memory"
+	"golang.org/x/exp/slices"
```
wshwsh12, line 25: Use "slices"

pkg/executor/sortexec/sort.go
```diff
@@ -52,36 +56,61 @@ type SortExec struct {
...
+	if e.Unparallel.spillAction != nil {
+		e.Unparallel.spillAction.SetFinished()
+	}
+
+	if e.IsUnparallel {
+		for _, partition := range e.Unparallel.sortPartitions {
+			partition.close()
+		}
+	} else {
+		if e.Parallel.err != nil {
```
wshwsh12, line 105: Need to ensure that all goroutines have exited.
And always close children(0).

pkg/executor/sortexec/sort.go
```diff
@@ -52,36 +56,61 @@ type SortExec struct {
...
-	e.diskTracker = nil
-	e.multiWayMerge = nil
-	if e.spillAction != nil {
-		e.spillAction.SetFinished()
+	if e.memTracker != nil {
+		e.memTracker.ReplaceBytesUsed(0)
 	}
-	e.spillAction = nil
-	return exec.Close(e.Children(0))
+	return e.Children(0).Close()
```
wshwsh12, line 113: Please use exec.Close(e.Children(0))

pkg/executor/sortexec/sort.go
```diff
@@ -336,11 +465,112 @@ func (e *SortExec) fetchRowChunks(ctx context.Context) error {
...
+	e.Unparallel.sortPartitions = append(e.Unparallel.sortPartitions, e.curPartition)
 	e.curPartition = nil
 	return nil
 }
 
+func (e *SortExec) fetchChunksParallel(ctx context.Context) error {
+	workerNum := len(e.Parallel.workers)
+
+	// Wait for the finish of all workers
+	workersWaiter := sync.WaitGroup{}
```
wshwsh12, line 477: Use util.WaitGroupWrapper

SeaRise:
> > > > Please add a performance comparison report
> > > 
> > > 
> > > okk
> > 
> > 
> > any updated?
> 
> updated, and here is the [doc](https://pingcap.feishu.cn/wiki/ALqkwLFnui9CXCkrTjacyR5rn3b)

It would be better to put it in the PR description.
Feishu documents are private.

pkg/executor/sortexec/sort.go
SeaRise: Tracker accuracy might be more important?
The overhead of tracker.consume seems small.
Does chunk's memory_used need to be subtracted in `sortChunkAndGetSortedRows`?

pkg/executor/sortexec/sort.go
```diff
@@ -52,42 +56,76 @@ type SortExec struct {
...
 	}
-	e.sortPartitions = e.sortPartitions[:0]
 
-	e.memTracker = nil
-	e.diskTracker = nil
-	e.multiWayMerge = nil
-	if e.spillAction != nil {
-		e.spillAction.SetFinished()
+	if e.memTracker != nil {
+		e.memTracker.ReplaceBytesUsed(0)
```
SeaRise, line 115: Should cleanup chunkChannel and global_queue before `memTracker.ReplaceBytesUsed(0)`
https://github.com/pingcap/tidb/blob/245951f07d5767047b1b67cf2f9d1ef17003a84d/pkg/executor/projection.go#L323-L324

pkg/executor/sortexec/sort.go
```diff
@@ -52,42 +56,76 @@ type SortExec struct {
...
+		e.Unparallel.spillAction.SetFinished()
+	}
+
+	if e.IsUnparallel {
+		for _, partition := range e.Unparallel.sortPartitions {
+			partition.close()
+		}
+	} else {
+		if e.Parallel.err != nil {
+			return e.Parallel.err
```
SeaRise, line 109: If so, other steps in `close()` will be skipped?

pkg/executor/sortexec/sort.go
```diff
@@ -336,11 +480,115 @@ func (e *SortExec) fetchRowChunks(ctx context.Context) error {
...
+	return e.Parallel.err
+}
+
+func (e *SortExec) processErrorForParallel(err error) {
+	e.Parallel.errRWLock.Lock()
+	defer e.Parallel.errRWLock.Unlock()
+	if e.Parallel.err == nil {
+		e.Parallel.err = err
+	}
+	e.tryToCloseFinishChannel()
```
SeaRise, line 573: To reduce the scope of lock
```suggestion
        e.tryToCloseFinishChannel()
	e.Parallel.errRWLock.Lock()
	defer e.Parallel.errRWLock.Unlock()
	if e.Parallel.err == nil {
		e.Parallel.err = err
	}
```

pkg/executor/sortexec/sort.go
```diff
@@ -52,42 +56,76 @@ type SortExec struct {
...
+		e.tryToCloseChunkChannel()
 	}
-	e.sortPartitions = e.sortPartitions[:0]
 
-	e.memTracker = nil
-	e.diskTracker = nil
-	e.multiWayMerge = nil
-	if e.spillAction != nil {
-		e.spillAction.SetFinished()
+	if e.memTracker != nil {
```
SeaRise, line 114: `e.Parallel.result` also need to be cleanup

pkg/executor/sortexec/parallel_sort_worker.go
xzhangxian1008: > Closing this channel will make data fetcher panic

Yes, and it's ok, data fetcher will recover this panic.

pkg/executor/sortexec/sort.go
xzhangxian1008: > exec.Close(e.Children(0))

done

pkg/executor/sortexec/sort.go
xzhangxian1008: > util.WaitGroupWrapper

done

pkg/executor/sortexec/sort.go
xzhangxian1008: > If so, other steps in `close()` will be skipped?

Yes, I think it's ok.

pkg/executor/sortexec/sort.go
xzhangxian1008: > `e.Parallel.result` also need to be cleanup

done

pkg/executor/sortexec/sort.go
xzhangxian1008: > Need to ensure that all goroutines have exited. And always close children(0).

done

pkg/executor/sortexec/sort.go
SeaRise: Sure? `child.close` be skipped is ok?

xzhangxian1008: > Sure? `child.close` be skipped is ok?

After check, Parallel.err shouldn't be returned, and I will delete it.

pkg/executor/sortexec/sort.go
```diff
@@ -97,31 +131,122 @@ func (e *SortExec) Open(ctx context.Context) error {
...
+           │   Put       │   Put            Put   │
+           │    │        │    │              │    │
+           │    │       Pop   │              │    │
+           │    │        │    ▼              │    │
+           │    │   ┌────┴─────────┐         │    │
+           │    └──►│              │◄────────┘    │
+           │        │ Global Queue │              │
+          Pop───────┤              ├─────────────Pop
+                    └──────────────┘
+*/
```
SeaRise, line 218: update the comments?

pkg/executor/sortexec/sort.go
```diff
@@ -336,11 +518,96 @@ func (e *SortExec) fetchRowChunks(ctx context.Context) error {
...
+	e.Unparallel.sortPartitions = append(e.Unparallel.sortPartitions, e.curPartition)
 	e.curPartition = nil
 	return nil
 }
 
+func (e *SortExec) fetchChunksParallel(ctx context.Context) error {
+	// Wait for the finish of all workers
+	workersWaiter := util.WaitGroupWrapper{}
+	// Wait for the finish of chunk fetcher
+	fetcherWaiter := util.WaitGroupWrapper{}
```
SeaRise, line 530: Seems they can be combined into one.

xzhangxian1008: > Seems they can be combined into one.

We can't combine them into one, as when one worker panics all workers may exit without consuming chunks sent by fetcher, this will lead to hang in sort spill.

pkg/executor/sortexec/sort.go
xzhangxian1008: > update the comments?

done

pkg/executor/sortexec/sort.go
```diff
@@ -97,31 +134,124 @@ func (e *SortExec) Open(ctx context.Context) error {
...
+                     Multi-way Merge
+                             │
+                      ┌──────┴──────┐
+                      │  Generator  │
+                      └──────┬──────┘
+                             │
+							Push
+                             │
+                             ▼
+					   resultChannel
```
SeaRise, line 222:
```suggestion
                            Push
                             │
                             ▼
                    resultChannel
```

SeaRise (approved):

pkg/executor/sortexec/parallel_sort_worker.go
```diff
@@ -0,0 +1,184 @@
...
+const SignalCheckpointForSort uint = 20000
+
+type parallelSortWorker struct {
+	workerIDForTest int
+
+	lessRowFunc func(chunk.Row, chunk.Row) int
+
+	chunkChannel  chan *chunk.Chunk
+	errOutputChan chan rowWithError
+	finishCh      chan struct{}
```
wshwsh12, line 37:
```suggestion
	chunkChannel  <-chan *chunk.Chunk
	errOutputChan chan<- rowWithError
	finishCh      <-chan struct{}
```

pkg/executor/sortexec/sort.go
```diff
@@ -52,42 +54,74 @@ type SortExec struct {
...
+	} else {
+		for {
+			_, ok := <-e.Parallel.chunkChannel
+			if !ok {
+				break
+			}
+		}
+		e.Parallel.mergerLock.Lock()
+		e.Parallel.merger = nil
+		e.Parallel.mergerLock.Unlock()
```
wshwsh12, line 110: Why need reset merger to nil here? In generateResult is better?
And I think we don't need to add a lock.

pkg/executor/sortexec/sort.go
```diff
@@ -52,42 +54,74 @@ type SortExec struct {
...
+		e.Unparallel.spillAction.SetFinished()
+	}
+
+	if e.IsUnparallel {
+		for _, partition := range e.Unparallel.sortPartitions {
+			partition.close()
 		}
+	} else {
+		for {
+			_, ok := <-e.Parallel.chunkChannel
```
wshwsh12, line 103: What happend when calling `Close` after `Open` without calling `Next`?

pkg/executor/sortexec/sort.go
```diff
@@ -138,29 +266,81 @@ func (e *SortExec) Next(ctx context.Context, req *chunk.Chunk) error {
...
+		}
+	}()
+
+	e.Parallel.mergerLock.Lock()
+	defer e.Parallel.mergerLock.Unlock()
+	if e.Parallel.merger == nil {
+		// Sort has been closed
+		return
+	}
+	e.Parallel.merger.init()
```
wshwsh12, line 291: Check e.finishch is ok?

pkg/executor/sortexec/parallel_sort_worker.go
xzhangxian1008: > <-

done

pkg/executor/sortexec/sort.go
xzhangxian1008: > Why need reset merger to nil here? In generateResult is better? And I think we don't need to add a lock.

okk, I think it's better to put it in generateResult.

pkg/executor/sortexec/sort.go
xzhangxian1008: > What happend when calling `Close` after `Open` without calling `Next`?

I will fix it.

pkg/executor/sortexec/sort.go
xzhangxian1008: > Check e.finishch is ok?

okk

pkg/sessionctx/variable/tidb_vars.go
```diff
@@ -1465,6 +1467,7 @@ const (
 	DefTiDBOptObjective                               = OptObjectiveModerate
 	DefTiDBSchemaVersionCacheLimit                    = 16
 	DefTiDBIdleTransactionTimeout                     = 0
+	DefEnableParallelSort                             = false
```
winoros, line 1470: Just a small question, why we don't enable it by default?

xzhangxian1008: > Just a small question, why we don't enable it by default?

Because we want to simultaneously enable the parallel sort and parallel sort spill in v8.1 .

wshwsh12 (approved):

PR #49756, opened 2023-12-25
expression: pass const bool to all calls for expression.ConstItem

### LINKED ISSUE #49755: pass const bool to all calls for expression.ConstItem
see description in #49756

lcwangchao:

Issue Number: close #49755

Problem Summary:

The method `expression.ConstItem` receives a bool argument to indicate whether the expression is a constant consider/not consider different contexts. Many calls use `StatementContext.UseCache` as an input but most of them are necessary. We'd better to use a const bool value as the input argument to make the code easier to understand.

- We use `ConstItem(false)` to check that `APPROX_PERCENTILE` should be a constant. It is to make sure its behaviors are the same when in or not in the plan cache.
- We use `ConstItem(true)` in `WrapWithCastAsDecimal` to do some find-grained precision optimization in the expression's build phase. 
- We use `ConstItem(false)` in the vector evaluation implement of function `aes_encrypt`/`aes_decrypt` to do optimize  when key constant.
-  the pattern of `ilike` should be a constant so we use `ConstItem(true)` to check it. 

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/expression/aggregation/base_func.go
```diff
@@ -145,7 +145,7 @@ func (a *baseFuncDesc) typeInfer4ApproxPercentile(ctx sessionctx.Context) error
 		return errors.New("APPROX_PERCENTILE should take 2 arguments")
 	}
 
-	if !a.Args[1].ConstItem(ctx.GetSessionVars().StmtCtx.UseCache) {
+	if !a.Args[1].ConstItem(false) {
```
lcwangchao, line 148: This check should not be affected by plan cache, PTAL @qw4990

pkg/expression/builtin_ilike_vec.go
```diff
@@ -52,28 +53,30 @@ func (b *builtinIlikeSig) canMemorize(param *funcParam) bool {
...
+	return pattern, true
 }
 
 func (b *builtinIlikeSig) getEscape(ctx EvalContext, input *chunk.Chunk, result *chunk.Column) (int64, bool, error) {
 	rowNum := input.NumRows()
 	escape := int64('\\')
 
-	if !b.args[2].ConstItem(ctx.GetSessionVars().StmtCtx.UseCache) {
+	if !b.args[2].ConstItem(true) {
 		return escape, true, errors.Errorf("escape should be const")
```
lcwangchao, line 80: This line cannot never reached because the parser blocked all other formats except for a literal format. I just use `ConstItem(true)` to make it a some behavior with parser

YangKeao (approved):
LGTM

qw4990 (approved):

XuHuaiyu (approved):

Issue #49717, opened 2023-12-22
Import into using global sort failed with error "failed to get gcs file attribute" on gcp


aytrack:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
1. Import into data from gcs and use gcs as cloud storage

### 2. What did you expect to see? (Required)
Import success
### 3. What did you see instead (Required)
```sql
[14:52:33]TiDB root:test> show import jobs \G
Job_ID           | 2
Data_Source      | gs://xxxx.*csv
Target_Table     | `test`.`xxx`
Table_ID         | 102
Phase            | importing
Status           | failed
Source_File_Size | 9.535TiB
Imported_Rows    | <null>
Result_Message   | failed to get gcs file attribute, file info: input.bucket='xxx', input.key='2/32/data/aa7061f5-4230-4697-bebe-d6b2f64916b6_stat/40': Get "https://storage.googleapis.com/storage/v1/b/xxx%2Faa7061f5-4230-4697-bebe-d6b2f64916b6_stat%2F40?alt=json&prettyPrint=false&projection=full": write tcp 10.65.88.3:40373->172.253.117.207:443: write: broken pipe
Create_Time      | 2023-12-22 03:31:50.555484
Start_Time       | 2023-12-22 03:31:53.923179
End_Time         | 2023-12-22 06:51:03.490446
Created_By       | root@%
```
### 4. What is your TiDB version? (Required)
```sql
[15:09:25]TiDB root:test> select tidb_version();
+-----------------------------------------------------------+
| tidb_version()                                            |
+-----------------------------------------------------------+
| Release Version: v7.6.0-alpha                             |
| Edition: Community                                        |
| Git Commit Hash: 5c279d800774ae7ce5ff9934e1334740f1c7207b |
| Git Branch: heads/refs/tags/v7.6.0-alpha                  |
| UTC Build Time: 2023-12-21 07:56:10                       |
| GoVersion: go1.21.5                                       |
| Race Enabled: false                                       |
| Check Table Before Drop: false                            |
| Store: tikv                                               |
+-----------------------------------------------------------+
```

PR #49768, opened 2023-12-25
global sort: add 2-level iterator for merge property and range splitter

lance6716:

Issue Number: close https://github.com/pingcap/tidb/issues/49717

Problem Summary:

- For `MultipleFilesStat`, it's maintained every 500 files, and `MultipleFilesStat.MaxOverlappingNum` is the number of the key range maximum overlaping of the 500 files. In this PR, `MultipleFilesStat.Filenames` is also sorted by start key of each file, so that we only need to keep number of `MultipleFilesStat.MaxOverlappingNum` files open from left to right, and get the minimum key among the opened files.
- We introduced a `limitSizeMergeIter`, which can have reader openers, their weights, and maximum weight limit as arguments. `limitSizeMergeIter` will automatically open readers in given order and keep readers' weight sum under the limit.
- For `RangeSplitter`, now it takes `[]MultipleFilesStat` as arguments. It has two levels of `limitSizeMergeIter`. The upper level is a whole `MultipleFilesStat` with weight `MultipleFilesStat.MaxOverlappingNum`, and the limit for this level is `MergeSortOverlapThreshold`. The second level is intra-`MultipleFilesStat` which means the 500 files, with weight `1`, and the limit is `MultipleFilesStat.MaxOverlappingNum`. In this way `RangeSplitter` can still output ordered key but won't open all files.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/lightning/backend/external/iter.go
+++ b/br/pkg/lightning/backend/external/iter.go
@@ -250,9 +250,9 @@ func (i *mergeIter[T, R]) close() error {
 // exhausted, i.err will be nil instead of io.EOF. For other errors, i.err will
 // be set.
 //
-// readerClosed >=0 means that reader is closed after last
-// invocation, -1 means no reader is closed.
-func (i *mergeIter[T, R]) next() (ok bool, closeReaderIdx int) {
+// closeReaderIdx >= 0 means that reader is closed after last invocation, -1
+// means no reader is closed.
+func (i *mergeIter[T, R]) next() (closeReaderIdx int, ok bool) {
 	closeReaderIdx = -1
 	if i.lastReaderIdx >= 0 {
 		if i.checkHotspot {
@@ -287,7 +287,7 @@ func (i *mergeIter[T, R]) next() (ok bool, closeReaderIdx int) {
 					err := (*rp).switchConcurrentMode(isHotspot)
 					if err != nil {
 						i.err = err
-						return false, closeReaderIdx
+						return closeReaderIdx, false
 					}
 				}
 				i.checkHotspotCnt = 0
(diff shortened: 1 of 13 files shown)


D3Hunter (commented):
12/13

br/pkg/lightning/backend/external/iter.go
```diff
@@ -292,23 +313,121 @@ func (i *mergeIter[T, R]) next() bool {
...
+	i := &limitSizeMergeIter[T, R]{
+		mergeIter:     iter,
+		readerOpeners: readerOpeners,
+		weights:       weights,
+		nextReaderIdx: end,
+		limit:         limit,
+	}
+	for j := 0; j < end; j++ {
+		i.weightSum += weights[j]
+	}
```
D3Hunter, line 385:
```suggestion
	i.weightSum = cur
```

br/pkg/lightning/backend/external/iter.go
```diff
@@ -292,23 +313,121 @@ func (i *mergeIter[T, R]) next() bool {
...
+}
+
+// limitSizeMergeIter acts like a mergeIter, except that each reader has a weight
+// and it will try to open more readers with the total weight doesn't exceed the
+// limit.
+//
+// Because it's like a mergeIter it's expected to iterate in ascending order,
+// caller should set a proper limit to do not block opening new readers
+// containing the next minimum element.
+type limitSizeMergeIter[T heapElem, R sortedReader[T]] struct {
```
D3Hunter, line 343: if the files are ordered by firstKey, we can open those files when current iterated key >= firstKey, so no need to calculate weight.

it's error-prone to let caller make sure the limit is enough and might cause the iterated key might not in ascending order.

lance6716: We need to open the file to know the first key (or record first key in another place). In current design, we don't record first key for intra-`MultipleFilesStat`. Another point is it's more cheap to saving a number representing the maximum overlapping level than saving every start key.

> make sure the limit is enough and might cause the iterated key might not in ascending order.

the caller can use `mergeIter` directly, like `MergeKVIter`. Currently for this problem I can only figure out `limitSizeMergeIter` with this prerequisite.

lance6716: And in fact in my first design half a year ago I suggest to record `MinKey`, `MaxKey` for every file and save it in a `MetadataStorage` interface (a table in downstream database should be OK), so these actions are easy to implement 😂 Current global sort system component design is complicated and I think they are over-optimized.

D3Hunter (approved):

Benjamin2037 (approved):
LGTM

Issue #49318, opened 2023-12-11
Improve the observability of resource control


nolouch:
### System table  
- [x] show RU consumption in statement level in the `statement_summary`, `statements_summary_history` table @glorv 
        https://github.com/tidbcloud/tidb-cse/pull/678
- [x] (option) show RU in `mysql.request_unit_by_group` https://github.com/pingcap/tidb/pull/49873 @glorv
- [ ] (option) show RU in the new table `information_schema.resource_group_stats` ??

### Log: 
- [x] Slow Log https://github.com/pingcap/tidb/issues/42892 @glorv 
- [x] Trace Log https://github.com/tikv/pd/issues/7509 @nolouch 

### Metrics @bufferflies 
To check the service quality:   https://github.com/pingcap/tidb/pull/50086

- [x] QPS & TPS by resource groups
- [x] P99/P95 by resource groups
- [x] DML latency by resource groups
- [x] command by resource groups
- [x] Failed command by resource groups
- [x] Connection Count by resource groups 

### Dashboard @nolouch 
- [x] add ru supports in the dashboard `statement` page  https://github.com/pingcap/tidb-dashboard/pull/1639
- [x] https://github.com/pingcap/tidb-dashboard/pull/1642
- [x] https://github.com/pingcap/tidb-dashboard/pull/1646
- [x] https://github.com/pingcap/tidb-dashboard/pull/1647

### TiFlash
- [ ] https://github.com/pingcap/tidb/issues/48168

### Runtime 
- [x] Introduce a session `variable ru_by_last_statement` https://github.com/pingcap/tidb/pull/49769 @nolouch 

### Docs

- [x] https://github.com/pingcap/docs-cn/pull/15935

PR #49769, opened 2023-12-25
*: add last ru consumption for tidb_last_query_info

nolouch:

Issue Number: ref https://github.com/pingcap/tidb/issues/49318

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
with debug version (last query)
image

### Release note

```release-note
None
```


pkg/session/session.go
```diff
@@ -2351,23 +2351,6 @@ func runStmt(ctx context.Context, se *session, s sqlexec.Statement) (rs sqlexec.
 
 	sessVars := se.sessionVars
 
-	// Record diagnostic information for DML statements
```
glorv, line 2354: This seems to be a behavior change. Before this PR, `LastQueryInfo` is record even if this function returns error, but after this change, if `runStmt` returns error and the returned RecordSet is nil, `FinishExecuteStmt` won't be called so LastQueryInfo won't be recorded either.

nolouch: Most errors occur during the execution stage, I think it's reasonable move to here.  and you can see the manual test on the issue, I kill the pd, it can record `pd timeout` error during execution.

pkg/sessionctx/variable/session.go
```diff
@@ -792,6 +792,9 @@ type SessionVars struct {
 	// LastFoundRows is the number of found rows of last query statement
 	LastFoundRows uint64
 
+	// LastRUConsumption is the total RU consumption of last query statement
+	LastRUConsumption float64
```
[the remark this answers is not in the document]
nolouch: Here is an independent variable like `sessVars.TxnCtx.StartTS`, but I reuse the user-aware variable `last_query_info` to collect them with some condition checks. I think it can keep the consist with those conditions.

pkg/session/session.go
nolouch: cc @djshow832 is it ok to move to FinishExecuteStmt?

pkg/executor/adapter.go
```diff
@@ -1449,6 +1450,38 @@ func (a *ExecStmt) FinishExecuteStmt(txnTS uint64, err error, hasMoreResults boo
...
+		var lastRUConsumption float64
+		if ruDetailRaw := a.GoCtx.Value(util.RUDetailsCtxKey); ruDetailRaw != nil {
+			ruDetail := ruDetailRaw.(*util.RUDetails)
+			lastRUConsumption = ruDetail.RRU() + ruDetail.WRU()
+		}
+		failpoint.Inject("mockRUConsumption", func(_ failpoint.Value) {
+			lastRUConsumption = float64(len(sessVars.StmtCtx.OriginalSQL))
+		})
+		// Keep the previous queryInfo for `show session_states` because the statement needs to encode it.
+		sessVars.LastQueryInfo = sessionstates.QueryInfo{
```
glorv, line 1473: Another question: is it better that when if the statement type is not exec or dml, the lastQueryInfo/LastRUConsumption should be reset instead of just keep unchanged?

nolouch: I think we can keep unchanged.

glorv (approved):

CabinfeverB (approved):

Issue #46539, opened 2023-08-31
Support cost-based access path selection for the multi-valued index


time-and-fate:
## Enhancement
Now the statistics collection on the multi-valued index is skipped in tidb. Because of user requirements, we need to support this to make estimation for multi-valued index access path more accurate.
This issue is for tracking the work on the series of functionalities and also acts as a simple design doc.

### Background
Currently, statistics collection on the multi-valued index is directly disabled. So cost calculation for the multi-valued index access paths is done without actual statistics, which means it's based on "pseudo estimation" in tidb.
To achieve cost-based access path selection for the multi-valued index based on real statistics, there are generally several parts of work remaining needs to be done:
1. Support collecting (and storing) statistics for multi-valued index
2. Support loading and maintaining the statistics in the memory
3. Support using these statistics to estimate row count

Among these things, many infrastructures are already there and we don't need to implement them again. Like:
1. Build statistics on columns/indexes using table row samples or full index data.
2. Maintain the statistics in the system table.
3. Load the statistics metadata (like row count, NDV, etc.) into tidb memory (stats cache).
4. Load the concrete statistics (like histograms, TopN, etc.) into tidb memory according to the query content synchronously or asynchronously.
5. Use the statistics to estimate the row count of a range on an index.
6. Use independence assumption to handle estimation for the composite expressions.

We just need to enhance them, make them work correctly for the multi-valued index, and implement some new logic based on them.

#### Resource consumption of statistics collection
In the standard method to collect statistics in v2 analyze, we should use the same samples to build stats for all columns/indexes. If we implement collecting statistics for multi-valued index in this way, it will need tidb to load the samples of JSON columns to tidb. However, the JSON values may cost too much memory, and we can't handle such case very well now.
Therefore, we choose to reuse another infrastructure in tidb/tikv here, which is collect statistics on an index by a sequential full scan on tikv. Statistics are built during this process in tikv and then merged in tidb. In this way, we only need to collect and merge statistics in tidb, which would cost much less resources in tidb than collecting JSON value samples and building full statistics in tidb.

#### Specialness of the multi-valued index
As the definition of multi-valued index, the row count and NDV of this index may be higher than the table row count.
This needs us to treat it differently and carefully in many places. For example, when collecting (and storing) statistics, we need to avoid using it as the table-level row count. And when maintaining and using the statistics, we need to allow the row count and NDV of the multi-valued index to exceed the table-level row count and NDV.

### Development
- Support collecting statistics for multi-valued index
[...]

PR #49852, opened 2023-12-27
planner, statistics: support estimation for mv index access path | tidb-test=pr/2270

time-and-fate:

Issue Number: close https://github.com/pingcap/tidb/issues/46539

- Enable estimation for ranges on the mv index, and adjust the realtime row count for it. (`pkg/planner/cardinality/row_count_index.go`, `pkg/statistics/table.go`)
- Use the independence assumption to estimate the total selectivity from the estimation of the partial paths. (`CalcTotalSelectivityForMVIdxPath()` in `pkg/planner/cardinality/selectivity.go`)
- Reuse previous logic to make `Selectivity()` support estimating filters that can be access condition of the mv index. (`pkg/planner/cardinality/selectivity.go`, `pkg/planner/core/indexmerge_path.go`, `pkg/statistics/table.go`)
- Fix a bug introduced by #46292, which will cause wrong stats when running tidb without real tikv. (`pkg/statistics/cmsketch.go`)

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/cardinality/selectivity.go
```diff
@@ -621,6 +671,34 @@ func getMaskAndRanges(ctx sessionctx.Context, exprs []expression.Expression, ran
...
+	ctx sessionctx.Context,
+	coll *statistics.HistColl,
+	id int64,
+	remainedExprs []expression.Expression,
+) (float64, int64, bool) {
+	cols := coll.MVIdx2Columns[id]
+	if len(cols) == 0 {
+		return 1.0, 0, false
+	}
+	accessConds, _ := CollectFilters4MVIndex(ctx, remainedExprs, cols)
```
qw4990, line 684: Should we multiply the selectivity with 0.8 if there is any remaining filter?

pkg/planner/cardinality/selectivity.go
```diff
@@ -621,6 +671,34 @@ func getMaskAndRanges(ctx sessionctx.Context, exprs []expression.Expression, ran
 	return mask, ranges, false, nil
 }
 
+func getMaskAndSelectivityForMVIndex(
```
qw4990, line 674: Could you please add some example comments with the code, for example:
```
func getMaskAndSelectivityForMVIndex(
    ...
    remainedExprs []Expr // [(x member of j1), (y member of j2), (others)]
) {
    ....
    accessConds, _ := CollectFilters4MVIndex(ctx, remainedExprs, cols) // [(x member of j1), (y member of j2)]
    ....
}
```
Like this:
image

pkg/planner/cardinality/selectivity.go
```diff
@@ -417,6 +434,39 @@ OUTER:
...
+	coll *statistics.HistColl,
+	partialPaths []*planutil.AccessPath,
+	isIntersection bool,
+) float64 {
+	selectivities := make([]float64, 0, len(partialPaths))
+	for _, path := range partialPaths {
+		// Note that although the total row count of the mv index is usually larger than the table, the CountAfterAccess
+		// of a partial path will never be larger than the table total row count.
+		// And for an index merge path with only one partial path, the CountAfterAccess will be exactly the same as the
+		// CountAfterAccess of the partial path (currently there's no index filter for partial path of mv index merge path).
```
AilinKid, line 450: ”currently there's no index filter for partial path of mv index merge path“
image
we did?

pkg/planner/cardinality/row_count_index.go
```diff
@@ -432,13 +430,15 @@ func expBackoffEstimation(sctx sessionctx.Context, idx *statistics.Index, coll *
...
-			foundStats bool
+			count       float64
+			selectivity float64
+			err         error
+			foundStats  bool
 		)
 		if col, ok := coll.Columns[colID]; ok && !col.IsInvalid(sctx, coll.Pseudo) {
 			foundStats = true
 			count, err = GetRowCountByColumnRanges(sctx, coll, colID, tmpRan)
+			selectivity = count / float64(coll.RealtimeCount)
```
AilinKid, line 441: one case here
if we have a index(normal-col1, json)
when we have normal-col1 > xxx
and we use prefix column stats to est the rows that this mv index can access, that seems also unfair

AilinKid (commented):
Rest LGTM

pkg/planner/cardinality/selectivity.go
time-and-fate: Which is this case? Is this a pure MV index path?

AilinKid: o sorry, its not mv case

qw4990 (approved):

pkg/planner/cardinality/selectivity.go
time-and-fate: No. This is only for finding the access conditions. If some conditions don't become access conditions, they are recognized as not covered, and processed by the existing logic in `Selectivity()` by * 0.8.

pkg/planner/cardinality/selectivity.go
time-and-fate: I think this logic is not complicated, and the existing comments on `collectFilters4MVIndex()` and `buildPartialPaths4MVIndex()` are enough, so I added a reference to those comments.

pkg/statistics/cmsketch.go
```diff
@@ -861,8 +861,8 @@ func SortTopnMeta(topnMetas []TopNMeta) {
 
 // TopnMetaCompare compare topnMeta
 func TopnMetaCompare(i, j TopNMeta) int {
-	c := cmp.Compare(i.Count, j.Count)
-	if c == 0 {
+	c := cmp.Compare(j.Count, i.Count)
```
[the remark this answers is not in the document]
time-and-fate: You can take a look at the changes in the PR mentioned in the description:
https://github.com/pingcap/tidb/pull/46292/files#diff-53469762b11ee4c62b1e4993444e9cb60a39091867f8f6a8a28fee2c617a9e0d

This change replaces the original `SortFunc` with this `TopnMetaCompare`, but it's wrong.
Previously, the sorting logic was: first by `Count`, from large to small, then by `Encoded`, from small to large, which is different from `TopnMetaCompare` here.

pkg/planner/cardinality/row_count_index.go
time-and-fate: After offline communication, we agree that this logic is reasonable. Because the row count from `GetRowCountByColumnRanges` should correspond to the table total row count instead of the row count of this index.

pkg/planner/cardinality/selectivity.go
```diff
@@ -820,10 +898,11 @@ func getEqualCondSelectivity(sctx sessionctx.Context, coll *statistics.HistColl,
 	}
 	val := types.NewBytesDatum(bytes)
 	if outOfRangeOnIndex(idx, val) {
+		realtimeCnt, _ := coll.GetScaledRealtimeAndModifyCnt(idx)
```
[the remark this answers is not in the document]
time-and-fate: I tend to leave this unchanged for now.
Because this will only happen for the mv index, and not all places will need this scaling for the mv index. Indeed, it needs some consideration to choose the correct row count to use. But I believe there won't be more new functionality that needs this logic for some time. So I tend not to make this logic more generalized.

pkg/statistics/cmsketch.go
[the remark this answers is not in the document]
time-and-fate: Actually, before this fix, the test result will be different between real tikv and the unistore.

AilinKid (approved):

pkg/planner/cardinality/selectivity.go
```diff
@@ -444,11 +444,55 @@ func CalcTotalSelectivityForMVIdxPath(
...
+				break
+			}
+		}
+		cols := expression.ExtractColumnsFromExpressions(nil, path.AccessConds, func(column *expression.Column) bool {
+			return virtualCol != nil && column.UniqueID == virtualCol.UniqueID
+		})
+		realtimeCount := coll.RealtimeCount
+		// If we can't find the virtual column from the access conditions, it's the case 2.
+		if len(cols) == 0 {
+			realtimeCount, _ = coll.GetScaledRealtimeAndModifyCnt(coll.Indices[path.Index.ID])
```
AilinKid, line 493: make sense

PR #50035, opened 2024-01-03
feat: add interval util functions for priority queue

hi-rustin:

Issue Number: ref https://github.com/pingcap/tidb/pull/49018

Problem Summary:

Added some functions to get the analysis intervals.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/handle/autoanalyze/priorityqueue/interval.go
+++ b/pkg/statistics/handle/autoanalyze/priorityqueue/interval.go
@@ -66,6 +66,7 @@ const lastFailedDurationQueryForPartition = `
 `
 
 // getAverageAnalysisDuration returns the average duration of the last 5 successful analyses for each specified partition.
+// If there are no successful analyses, it returns 0.
 func getAverageAnalysisDuration(
 	sctx sessionctx.Context,
 	schema, tableName string,
@@ -86,10 +87,12 @@ func getAverageAnalysisDuration(
 		return 0, err
 	}
 
+	// NOTE: if there are no successful analyses, we return 0.
 	var avgDuration *types.MyDecimal
-	if len(rows) > 0 {
-		avgDuration = rows[0].GetMyDecimal(0)
+	if len(rows) == 0 {
+		return 0, nil
 	}
+	avgDuration = rows[0].GetMyDecimal(0)
 	duration, err := avgDuration.ToFloat64()
 	if err != nil {
 		return 0, err
@@ -99,6 +102,7 @@ func getAverageAnalysisDuration(
 }
 
 // getLastFailedAnalysisDuration returns the duration since the last failed analysis.
+// If there is no failed analysis, it returns 0.
 func getLastFailedAnalysisDuration(
 	sctx sessionctx.Context,
 	schema, tableName string,
@@ -119,10 +123,12 @@ func getLastFailedAnalysisDuration(
 		return 0, err
 	}
 
+	// NOTE: if there are no failed analyses, we return 0.
 	var lastFailedDuration uint64
-	if len(rows) > 0 {
-		lastFailedDuration = rows[0].GetUint64(0)
+	if len(rows) == 0 {
+		return 0, nil
 	}
+	lastFailedDuration = rows[0].GetUint64(0)
 
 	return time.Duration(lastFailedDuration) * time.Second, nil
 }
--- a/pkg/statistics/handle/autoanalyze/priorityqueue/interval.go
+++ b/pkg/statistics/handle/autoanalyze/priorityqueue/interval.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package analyzequeue
+package priorityqueue
 
 import (
 	"time"
--- a/pkg/statistics/handle/autoanalyze/priorityqueue/interval.go
+++ b/pkg/statistics/handle/autoanalyze/priorityqueue/interval.go
@@ -19,7 +19,6 @@ import (
 
 	"github.com/pingcap/tidb/pkg/sessionctx"
 	"github.com/pingcap/tidb/pkg/statistics/handle/util"
-	"github.com/pingcap/tidb/pkg/types"
 )
 
 const avgDurationQueryForTable = `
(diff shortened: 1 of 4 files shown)


hi-rustin (commented):
🔢 Self-check

pkg/statistics/handle/autoanalyze/priorityqueue/interval.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// getLastFailedAnalysisDuration returns the duration since the last failed analysis.
+func getLastFailedAnalysisDuration(
+	sctx sessionctx.Context,
+	schema, tableName string,
+	partitionNames ...string,
+) (time.Duration, error) {
+	query := ""
+	params := make([]interface{}, 0, len(partitionNames)+3)
```
hawkingrei, line 108:
```suggestion
	params := make([]any, 0, len(partitionNames)+3)
```

pkg/statistics/handle/autoanalyze/priorityqueue/interval_test.go
```diff
@@ -0,0 +1,274 @@
...
+	"time"
+
+	"github.com/pingcap/tidb/pkg/session"
+	"github.com/pingcap/tidb/pkg/sessionctx"
+	"github.com/pingcap/tidb/pkg/testkit"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetAverageAnalysisDuration(t *testing.T) {
+	store, _ := testkit.CreateMockStoreAndDomain(t)
```
hawkingrei, line 28:
```suggestion
	store := testkit.CreateMockStore(t)
```

pkg/statistics/handle/autoanalyze/priorityqueue/interval_test.go
```diff
@@ -0,0 +1,274 @@
...
+	require.Equal(t, time.Duration(3600)*time.Second, avgDuration)
+	// Non-partitioned table.
+	insertFinishedJob(tk, "example_schema1", "example_table1", "")
+	avgDuration, err = getAverageAnalysisDuration(sctx, "example_schema1", "example_table1", "")
+	require.NoError(t, err)
+	require.Equal(t, time.Duration(3600)*time.Second, avgDuration)
+}
+
+func TestGetLastFailedAnalysisDuration(t *testing.T) {
+	store, _ := testkit.CreateMockStoreAndDomain(t)
```
hawkingrei, line 61:
```suggestion
	store := testkit.CreateMockStore(t)
```

hawkingrei (approved):

pkg/statistics/handle/autoanalyze/priorityqueue/interval.go
```diff
@@ -0,0 +1,128 @@
...
+`
+
+// getAverageAnalysisDuration returns the average duration of the last 5 successful analyses for each specified partition.
+func getAverageAnalysisDuration(
+	sctx sessionctx.Context,
+	schema, tableName string,
+	partitionNames ...string,
+) (time.Duration, error) {
+	query := ""
+	params := make([]interface{}, 0, len(partitionNames)+3)
```
hi-rustin, line 75:
```suggestion
	params := make([]any, 0, len(partitionNames)+3)
```

hi-rustin (commented):
🔢 Self-check

time-and-fate (commented):
Looks like we didn't consider what would happen if there are no such rows in the analyze history.

pkg/statistics/handle/autoanalyze/priorityqueue/interval.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/pingcap/tidb/pkg/statistics/handle/util"
+	"github.com/pingcap/tidb/pkg/types"
+)
+
+const avgDurationQueryForTable = `
+	SELECT AVG(TIMESTAMPDIFF(SECOND, start_time, end_time)) AS avg_duration
+	FROM mysql.analyze_jobs
+	WHERE table_schema = %? AND table_name = %? AND state = 'finished' AND fail_reason IS NULL AND partition_name = ''
+	ORDER BY id DESC
+	LIMIT 5;
```
time-and-fate, line 30: This SQL is a bit questionable with both an `avg` and `ORDER BY + LIMIT`.

hi-rustin:
> Looks like we didn't consider what would happen if there are no such rows in the analyze history.

Then we will return 0 and compare it with 0 by `0 < 0`, then we can analyze it because it doesn't satisfy the condition.

time-and-fate:
> > Looks like we didn't consider what would happen if there are no such rows in the analyze history.
> 
> Then we will return 0 and compare it with 0 by `0 < 0`, then we can analyze it because it doesn't satisfy the condition.

Have you tested this case to make sure?
`AVG` and `MIN` on empty result set would be `NULL` instead of `0`.

And I think we'd better add some comments. For example, it's not obvious that `len(rows) == 0` in `getLastFailedAnalysisDuration()` is an expected case.

pkg/statistics/handle/autoanalyze/priorityqueue/interval.go
hi-rustin: I want to calculate the average analysis time for the most recent 5 records. Could you please provide more information about the issue that is causing you worry?

time-and-fate: The semantic of a SELECT statement is generally `JOIN -> WHERE -> AGG -> HAVING -> ORDER BY -> LIMIT`. You would need to write it like `avgDurationQueryForPartition` below to get what you want.
You can see this example: https://www.db-fiddle.com/f/3moV1YrnrZchZTaK41fk2h/1

hi-rustin:
> > > Looks like we didn't consider what would happen if there are no such rows in the analyze history.
> > 
> > 
> > Then we will return 0 and compare it with 0 by `0 < 0`, then we can analyze it because it doesn't satisfy the condition.
> 
> Have we tested this case to make sure? `AVG` and `MIN` on empty result set would be `NULL` instead of `0`.
> 
> And I think we'd better add some comments. For example, it's not obvious that `len(rows) == 0` in `getLastFailedAnalysisDuration()` is an expected case.

I added some tests to cover it and also updated the implementation to make it more obvious.

time-and-fate (approved):

hi-rustin:
@ti-chi-bot Hi man

PR #50050, opened 2024-01-03
statistics/handle/usage: modify the `SessionIndexUsageCollector` to implement new index usage design

### LINKED ISSUE #49835: Modify the existing `SessionIndexUsageCollector` for new index usage design

YangKeao:

Issue Number: close #49835

Problem Summary:

There is already an implementation of index usage session collector. However, it doesn't fit in our need perfectly and need some modification.

1. Copy the `SessionIndexUsageCollector` from `pkg/statistics/handle/usage/index_usage.go` to `pkg/statistics/handle/usage/indexusage/collector.go` to solve circle dependency.
2. Add more fields to the stats data.
3. Add more tests.
4. Deprecate the configuration `"index-usage-sync-lease"` and use a hard-coded duration to setup GC and Sweep ticker.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```


pkg/statistics/handle/usage/index_usage_integration_test.go
```diff
@@ -0,0 +1,107 @@
+// Copyright 2023 PingCAP, Inc.
```
hawkingrei, line 1:
```suggestion
// Copyright 2024 PingCAP, Inc.
```

YangKeao: Done.

pkg/statistics/handle/usage/indexusage/collector.go
```diff
@@ -0,0 +1,218 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/pingcap/tidb/pkg/parser/model"
+)
+
+// GlobalIndexID is the key type for indexUsageMap.
+type GlobalIndexID struct {
+	TableID int64
```
qw4990, line 26: If this table is a partitioning table, it seems like we'll record index usage for each partition individually, should we merge the same index data from different partitions together?

pkg/statistics/handle/usage/indexusage/collector.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	if percentage == 1.0 {
+		bucket = len(bucketBound)
+	}
+
+	return bucket
+}
+
+// NewSample creates a new data point for index usage.
+func NewSample(queryTotal uint64, kvReqTotal uint64, rowAccess uint64, tableTotalRows uint64) *Sample {
```
qw4990, line 62: Should we maintain an object pool for `*Sample`? It seems like this allocation operation is very high-frequent.

pkg/statistics/handle/usage/indexusage/collector.go
```diff
@@ -0,0 +1,218 @@
...
+// SessionIndexUsageCollector is a list item that holds the index usage mapper. If you want to write or read mapper, you must lock it.
+// TODO: use a third-party thread-safe list implementation instead of maintaining the list manually.
+// TODO: merge this list into SessionStatsList.
+/*
+                            [session1]                [session2]                        [sessionN]
+                                |                         |                                 |
+                            update into              update into                       update into
+                                |                         |                                 |
+                                v                         v                                 v
+[StatsHandle.Head] --> [session1.IndexUsage] --> [session2.IndexUsage] --> ... --> [sessionN.IndexUsage]
```
qw4990, line 93: Can we rewrite this mechanism to use goroutines to collect information, which seems more safe and go-style?

pkg/statistics/handle/usage/indexusage/collector.go
YangKeao: It's modified to use `Sample`.

pkg/statistics/handle/usage/indexusage/collector.go
```diff
@@ -0,0 +1,225 @@
...
+	collector collector.SessionCollector[indexUsage]
+}
+
+// Update updates the mapper in SessionIndexUsageCollector
+func (s *SessionIndexUsageCollector) Update(tableID int64, indexID int64, sample Sample) {
+	s.mapper.update(tableID, indexID, sample)
+}
+
+// Report reports the mapper in `SessionIndexUsageCollector` to the global collector
+func (s *SessionIndexUsageCollector) Report() {
```
YangKeao, line 208: This function is not called in this PR. I prepare to use it after (or before) each statement. When the collector for each statements (which is required to record `query count` for indexes) is reset, it'll also try to call the `Report`.

pkg/statistics/handle/usage/indexusage/collector.go
YangKeao: Done :heart:

pkg/statistics/handle/usage/indexusage/collector.go
YangKeao: I agree that the final display on `index_usage` table will only give one row for each table/index. 

However, I prefer to handle the aggregation of partition local indexes when we are retrieving the table, so that this package doesn't need to handle `infoschema` related things. Also, when we are retrieving the table, we'll need to filter out the table/index according to the session infoschema.

pkg/statistics/handle/types/interfaces.go
```diff
@@ -68,17 +69,23 @@ type StatsUsage interface {
...
 
-	// NewSessionIndexUsageCollector creates a new IndexUsageCollector on the list.
+	// NewSessionIndexUsageCollector creates a new Collector on the list.
 	// The returned value's type should be *usage.SessionIndexUsageCollector, use interface{} to avoid cycle import now.
 	// TODO: use *usage.SessionIndexUsageCollector instead of interface{}.
-	NewSessionIndexUsageCollector() interface{}
-
-	// DumpIndexUsageToKV dumps all collected index usage info to storage.
-	DumpIndexUsageToKV() error
+	NewSessionIndexUsageCollector() *indexusage.SessionIndexUsageCollector
```
qw4990, line 75: Better to add a new separate interface `IndexUsage` for it, we can do it later. But could you add a comment here, like "TODO: a new separate interface for it".

pkg/statistics/handle/usage/indexusage/collector.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+// Sample is the data structure to store index usage information.
+type Sample struct {
+	LastUsedAt     time.Time
+	QueryTotal     uint64
+	KvReqTotal     uint64
+	RowAccessTotal uint64
+	// 0, 0-1, 1-10, 10-20, 20-50, 50-100, 100
+	PercentageAccess [7]uint64
```
qw4990, line 38: Please add more comments for those fields.

pkg/statistics/handle/usage/indexusage/collector.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+		if !foundIdx {
+			delete(c.mapper, k)
+		}
+	}
+}
+
+// SessionIndexUsageCollector collects index usage per-session
+type SessionIndexUsageCollector struct {
+	mapper indexUsage
```
qw4990, line 197: `mapper` seems too general and not a good name, could we rename it?

pkg/statistics/handle/usage/indexusage/collector.go
```diff
@@ -0,0 +1,225 @@
...
+	for id := range destMap {
+		item := destMap[id]
+		m.updateByKey(id, item)
+	}
+}
+
+// Collector represents a data structure to record the index usage for the whole node
+type Collector struct {
+	collector collector.GlobalCollector[indexUsage]
+	mapper    indexUsage
```
qw4990, line 115: Could we rename `mapper` to another name like `indexUsage`?

pkg/statistics/handle/usage/indexusage/collector.go
```diff
@@ -0,0 +1,225 @@
...
+	iuc := &Collector{
+		mapper: make(indexUsage),
+	}
+	iuc.collector = collector.NewGlobalCollector[indexUsage](iuc.merge)
+
+	return iuc
+}
+
+// GetIndexUsage returns the index usage information
+func (c *Collector) GetIndexUsage(tableID int64, indexID int64) *Sample {
```
qw4990, line 132: Is it safe to return a pointer here? What if the caller modify it?

qw4990: Maybe it's safer to assume `Sample` is immutable after creation, and never update it in-place like `Sample.QueryTotal += XX`.

pkg/statistics/handle/types/interfaces.go
YangKeao: Good! I've moved these functions to an `IndexUsage` interface. Also, I added a TODO for the following `table delta and stats usage`. :beers:

pkg/statistics/handle/usage/indexusage/collector.go
YangKeao: Done

pkg/statistics/handle/usage/indexusage/collector.go
YangKeao: Done

pkg/statistics/handle/usage/indexusage/collector.go
YangKeao: Yes! It has been renamed to `indexUsage`.

pkg/statistics/handle/usage/indexusage/collector.go
YangKeao: Nice catch. The safety seems fine in the collector, but I agree it's more safe / performant to use a `Sample` here. It avoids the potential risk in the future, and also avoids a heap allocation in this function. I've modified this function to use `Sample`.

qw4990 (approved):

tangenta (approved):

PR #50130, opened 2024-01-05
membuf: count memory cost for returned []byte or SliceLocation

### LINKED ISSUE #50131: count memory cost for returned []byte or SliceLocation
## Enhancement

We have limited the total memory block acquired from membuf.Pool to 16G, but the keys and values type which has their underlying array pointing to membuf.Pool also costs a lot memory so the node may OOM.

image
⬇️
image

lance6716:

Issue Number: close #50131

Problem Summary:

in ADD INDEX test cases KV are small, so for 16G data the number of KV is larger. So the overhead of storing []byte are not negligible. I think we should take them into consideration.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


br/pkg/membuf/buffer.go
```diff
@@ -172,17 +175,51 @@ func (p *Pool) NewBuffer(opts ...BufferOption) *Buffer {
...
+// smallObjOverheadBatch is the batch size to acquire memory from limiter. 256KB
+// can store 256KB/24B = 10922 []byte objects, or 256KB/12B = 21845 SliceLocation
+// objects.
+const smallObjOverheadBatch = 256 * 1024
+
+// recordSmallObjOverhead records the memory cost of []byte or SliceLocation into
+// pool's limiter. The caller will ensure the pool's limiter is not nil.
+func (b *Buffer) recordSmallObjOverhead(n int) {
+	if n > b.smallObjOverheadCache {
+		b.pool.limiter.Acquire(smallObjOverheadBatch)
```
tangenta, line 187: I think for each goroutine, `p.limiter.Acquire()` should be invoked only once? Otherwise, there is a risk of deadlock.

```
Total memory: 10

goroutine 1:
func AllocBytesWithSliceLocation() {
    b.allocBytesWithSliceLocation() // limiter.Acquire(9), ok
->  b.recordSmallObjOverhead()      // limiter.Acquire(2), but 9 + 2 > 10, block
}

goroutine 2:
func AllocBytesWithSliceLocation() {
->  b.allocBytesWithSliceLocation() // limiter.Acquire(9), but 9 + 9 > 10, block 
    b.recordSmallObjOverhead()
}
```

lance6716: In this lock acquisition pattern we only rely the caller don't have too many concurrency and don't acquire too many memory to avoid deadlock 😂 

We don't know how many []byte will be allocated from the `Buffer` so it can't call `Acquire` once.

tangenta (approved):

wjhuang2016 (approved):

br/pkg/membuf/buffer.go
```diff
@@ -172,17 +175,51 @@ func (p *Pool) NewBuffer(opts ...BufferOption) *Buffer {
...
-// Reset resets the buffer, the memory is still retained in this buffer.
+// smallObjOverheadBatch is the batch size to acquire memory from limiter. 256KB
+// can store 256KB/24B = 10922 []byte objects, or 256KB/12B = 21845 SliceLocation
+// objects.
+const smallObjOverheadBatch = 256 * 1024
+
+// recordSmallObjOverhead records the memory cost of []byte or SliceLocation into
+// pool's limiter. The caller will ensure the pool's limiter is not nil.
+func (b *Buffer) recordSmallObjOverhead(n int) {
+	if n > b.smallObjOverheadCache {
```
D3Hunter, line 186: seems `n` never bigger than it, and b.smallObjOverheadCache always <= 0?

lance6716: the initial value of `smallObjOverheadCache` is `0`. For the first invocation `n` is bigger than it and will go through line 187~189. Then the next few invocations `n` is less than `smallObjOverheadCache` so directly goes to line 191.

D3Hunter: my fault, i read it as `if n > smallObjOverheadBatch {` 😓

Leavrth (approved):

PR #50183, opened 2024-01-08
planner: avoid scanning multi-valued index when there is no condition related | tidb-test=pr/2276

### LINKED ISSUE #50125: Current MV index maintenance logic will cause wrong query result on empty array values
## Bug Report

```sql
create table t(a int, d json, index iad(a, (cast(d->'$.b' as signed array))));
insert into t value(1,'{"b":[]}'), (2,'{"b":[]}');
select * from t use index (iad) where a = 1;
select * from t ignore index (iad) where a = 1;
```

```sql
> select * from t use index (iad) where a = 1;
+------+-----------+
| a    | d         |
+------+-----------+
|    1 | {"b": []} |
+------+-----------+

> select * from t ignore index (iad) where a = 1;
+------+-----------+
| a    | d         |
+------+-----------+
|    1 | {"b": []} |
+------+-----------+
```

```sql
> select * from t use index (iad) where a = 1;
Empty set (0.002 sec)

> select * from t ignore index (iad) where a = 1;
+------+-----------+
| a    | d         |
+------+-----------+
|    1 | {"b": []} |
+------+-----------+
1 row in set (0.001 sec)
```

```sql
> explain analyze select * from t use index (iad) where a = 1;
+-------------------------------+---------+---------+-----------+----------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------------+-----------+------+
| id                            | estRows | actRows | task      | access object                                                              | execution info                                                                                                                                                                                                                                                                                            | operator info                               | memory    | disk |
+-------------------------------+---------+---------+-----------+----------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------------+-----------+------+
| IndexMerge_7                  | 10.00   | 0       | root      |                                                                            | time:606.4µs, loops:1                                                                                                                                                                                                                                                                                     | type: union                                 | 270 Bytes | N/A  |
| ├─IndexRangeScan_5(Build)     | 10.00   | 0       | cop[tikv] | table:t, index:iad(a, cast(json_extract(`d`, _utf8'$.b') as signed array)) | time:480.1µs, loops:1, cop_task: {num: 1, max: 409.8µs, proc_keys: 0, rpc_num: 1, rpc_time: 384.5µs, copr_cache_hit_ratio: 0.00, build_task_duration: 33.9µs, max_distsql_concurrency: 1}, tikv_task:{time:0s, loops:1}, scan_detail: {total_keys: 1, get_snapshot_time: 20µs, rocksdb: {block: {}}}      | range:[1,1], keep order:false, stats:pseudo | N/A       | N/A  |
| └─TableRowIDScan_6(Probe)     | 10.00   | 0       | cop[tikv] | table:t                                                                    |                                                                                                                                                                                                                                                                                                           | keep order:false, stats:pseudo              | N/A       | N/A  |
+-------------------------------+---------+---------+-----------+----------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------------+-----------+------+
```

From v6.6.0 to the latest master (v7.6.0)

YangKeao:

Issue Number: close #50125

We cannot blindly scan on the multi-valued index, because scanning on it assumes that the array is not empty. I added a lot comments on related codes to describe why it's safe now.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
Fix the issue that multi-valued index can give incorrect result when the array is empty.
```

### DIFF
--- a/pkg/planner/core/indexmerge_path.go
+++ b/pkg/planner/core/indexmerge_path.go
@@ -109,24 +109,27 @@ func (ds *DataSource) generateIndexMergePath() error {
 	}
 
 	// If without hints, it means that `enableIndexMerge` is true
-	if len(ds.indexMergeHints) == 0 {
-		return nil
-	}
-	// If len(indexMergeHints) > 0, then add warnings if index-merge hints cannot work.
-	if regularPathCount == len(ds.possibleAccessPaths) {
-		if warningMsg == "" {
-			warningMsg = "IndexMerge is inapplicable"
+	if len(ds.indexMergeHints) != 0 {
+		// If len(indexMergeHints) > 0, then add warnings if index-merge hints cannot work.
+		if regularPathCount == len(ds.possibleAccessPaths) {
+			if warningMsg == "" {
+				warningMsg = "IndexMerge is inapplicable"
+			}
+			return nil
 		}
-		return nil
+
+		// If len(indexMergeHints) > 0 and some index-merge paths were added, then prune all other non-index-merge paths.
+		// if len(ds.possibleAccessPaths) > oldIndexMergeCount, it means composed index merge path is generated, prune others.
+		if len(ds.possibleAccessPaths) > oldIndexMergeCount {
+			ds.possibleAccessPaths = ds.possibleAccessPaths[oldIndexMergeCount:]
+		} else {
+			ds.possibleAccessPaths = ds.possibleAccessPaths[regularPathCount:]
+		}
 	}
 
-	// If len(indexMergeHints) > 0 and some index-merge paths were added, then prune all other non-index-merge paths.
-	// if len(ds.possibleAccessPaths) > oldIndexMergeCount, it means composed index merge path is generated, prune others.
-	if len(ds.possibleAccessPaths) > oldIndexMergeCount {
-		ds.possibleAccessPaths = ds.possibleAccessPaths[oldIndexMergeCount:]
-	} else {
-		ds.possibleAccessPaths = ds.possibleAccessPaths[regularPathCount:]
-	}
+	// If there is a multi-valued index hint, remove all paths which don't use the specified index.
+	ds.cleanAccessPathForMVIndexHint()
+
 	return nil
 }
 
@@ -1193,10 +1196,17 @@ func buildPartialPaths4MVIndex(
 	if virColID == -1 { // unexpected, no vir-col on this MVIndex
 		return nil, false, false, nil
 	}
-	if len(accessFilters) <= virColID { // no filter related to the vir-col, build a partial path directly.
-		partialPath, ok, err := buildPartialPath4MVIndex(sctx, accessFilters, idxCols, mvIndex, histColl)
-		return []*util.AccessPath{partialPath}, false, ok, err
+	if len(accessFilters) <= virColID {
+		// No filter related to the vir-col, cannot build a path for multi-valued index. Scanning on a multi-valued
+		// index will only produce the rows whose corresponding array is not empty.
+		return nil, false, false, nil
 	}
+	// If the condition is related with the array column, all following condition assumes that the array is not empty:
+	// `member of`, `json_contains`, `json_overlaps` all return false when the array is empty, except that
+	// `json_contains('[]', '[]')` is true. Therefore, using `json_contains(array, '[]')` is also not allowed here.
+	//
+	// Only when the condition implies that the array is not empty, it'd be safe to scan on multi-valued index without
+	// worrying whether the row with empty array will be lost in the result.
 
 	virCol := idxCols[virColID]
 	jsonType := virCol.GetType().ArrayType()
@@ -1221,7 +1231,9 @@ func buildPartialPaths4MVIndex(
 	case ast.JSONContains: // (json_contains(a->'$.zip', '[1, 2, 3]')
 		isIntersection = true
 		virColVals, ok = jsonArrayExpr2Exprs(sctx, sf.GetArgs()[1], jsonType)
-		if !ok || len(virColVals) == 0 { // json_contains(JSON, '[]') is TRUE
+		if !ok || len(virColVals) == 0 {
+			// json_contains(JSON, '[]') is TRUE. If the row has an empty array, it'll not exist on multi-valued index,
+			// but the `json_contains(array, '[]')` is still true, so also don't try to scan on the index.
 			return nil, false, false, nil
 		}
 	case ast.JSONOverlaps: // (json_overlaps(a->'$.zip', '[1, 2, 3]')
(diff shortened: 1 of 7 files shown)


pkg/planner/core/indexmerge_path.go
```diff
@@ -1193,10 +1193,17 @@ func buildPartialPaths4MVIndex(
 	if virColID == -1 { // unexpected, no vir-col on this MVIndex
 		return nil, false, false, nil
 	}
-	if len(accessFilters) <= virColID { // no filter related to the vir-col, build a partial path directly.
-		partialPath, ok, err := buildPartialPath4MVIndex(sctx, accessFilters, idxCols, mvIndex, histColl)
-		return []*util.AccessPath{partialPath}, false, ok, err
+	if len(accessFilters) <= virColID {
```
AilinKid, line 1196:
```
mysql> show create table t;
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                                                              |
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `a` int DEFAULT NULL,
  `d` json DEFAULT NULL,
  UNIQUE KEY `iad` (`a`,(cast(json_extract(`d`,_utf8mb3'$.b') as signed array))),
  KEY `ua` (`a`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci |
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.01 sec)

mysql> explain  analyze select a from t use index(ua);
+-----------------------------------------------------------------------------------------------------+
| EXPLAIN                                                                                             |
+-----------------------------------------------------------------------------------------------------+
| -> Covering index scan on t using ua  (cost=0.45 rows=2) (actual time=0.072..0.084 rows=2 loops=1)
 |
+-----------------------------------------------------------------------------------------------------+
1 row in set (0.01 sec)

mysql> explain  analyze select a from t use index(iad);
+-----------------------------------------------------------------------------------+
| EXPLAIN                                                                           |
+-----------------------------------------------------------------------------------+
| -> Table scan on t  (cost=0.45 rows=2) (actual time=0.086..0.100 rows=2 loops=1)
 |
+-----------------------------------------------------------------------------------+
1 row in set (0.01 sec)

mysql> explain  analyze select d->'$.b' from t use index(iad);
+-----------------------------------------------------------------------------------+
| EXPLAIN                                                                           |
+-----------------------------------------------------------------------------------+
| -> Table scan on t  (cost=0.45 rows=2) (actual time=0.068..0.082 rows=2 loops=1)
 |
+-----------------------------------------------------------------------------------+
1 row in set (0.01 sec)
```

seems MySQL does know itself have some rows missing for mv index value, and avoids using mv index to just fetch data for those only mv index columns.

so here is clear, avoid using mv index to fetch data when json column's is not covered by conditions.

tests/integrationtest/r/planner/core/indexmerge_path.result
```diff
@@ -663,12 +654,27 @@ insert into t value(2,2,2, '{"b":[3,4,5,6]}');
...
 explain select * from t use index (iad) where a = 1 and (2 member of (d->'$.b'));
 id	estRows	task	access object	operator info
 IndexMerge_7	1.00	root		type: union
 ├─IndexRangeScan_5(Build)	1.00	cop[tikv]	table:t, index:iad(a, cast(json_extract(`d`, _utf8mb4'$.b') as signed array))	range:[1 2,1 2], keep order:false, stats:partial[d:unInitialized]
 └─TableRowIDScan_6(Probe)	1.00	cop[tikv]	table:t	keep order:false, stats:partial[d:unInitialized]
+drop table if exists t;
+create table t(a int, d json, index iad(a, (cast(d->'$.b' as signed array))));
+insert into t value(1,'{"b":[]}'), (2,'{"b":[]}');
+select * from t use index (iad) where a = 1;
+Error 1815 (HY000): Internal : Can't find a proper physical plan for this query
```
AilinKid, line 667: this error is undesirable
a basic table scan should be conductable

AilinKid (commented):
/test check_dev

pkg/planner/core/indexmerge_path.go
YangKeao: > avoid using mv index to fetch data when json column's is not covered by conditions

Yes. That's what I was trying to do in this PR :beers: .

pkg/planner/core/planbuilder.go
```diff
@@ -1216,6 +1216,19 @@ func getPossibleAccessPaths(ctx sessionctx.Context, tableHints *hint.TableHintIn
...
+	// If all available paths are Multi-Valued Index, it's possible that the only multi-valued index is inapplicable,
+	// so that the table paths are still added here to avoid failing to find any physical plan.
+	allMVIIndexPath := true
+	for _, availablePath := range available {
+		if !isMVIndexPath(availablePath) {
+			allMVIIndexPath = false
+		}
+	}
+	if allMVIIndexPath {
+		available = append(available, tablePath)
```
YangKeao, line 1229: Here is how I fixed the "Can't find a proper physical plan for this query". 

Is the former `if len(available) == 0` only an optimization to explore less paths, or it's necessary to ensure the hint works properly? I saw some paths are added to `preferredPath` according to the `.Force` field later, but I'm not sure whether it's really a hard restriction, or just some soft preference :thinking: .

@AilinKid PTAL

YangKeao: Oops, I realized that the index merge access path doesn't have `.Force` :facepalm: . Let me think about it.

YangKeao: Oops. I'll need to remove these plan when after finding a valid index merge based in multi- valued index.

YangKeao: Done

tests/integrationtest/r/planner/core/indexmerge_path.result
```diff
@@ -499,11 +499,18 @@ TableReader_7	8000.00	root		data:Selection_6
 drop table if exists t;
 create table t(a int, j json, index kj((cast(j as signed array))));
 explain format = 'brief' select /*+ use_index(t, kj) */ * from t;
-Error 1815 (HY000): Internal : Can't find a proper physical plan for this query
+id	estRows	task	access object	operator info
+TableReader	10000.00	root		data:TableFullScan
+└─TableFullScan	10000.00	cop[tikv]	table:t	keep order:false, stats:pseudo
```
YangKeao, line 504: This is a behavior change. Now it's more compatible with MySQL.

pkg/planner/core/planbuilder.go
AilinKid: LGTM here

AilinKid (approved):

tests/integrationtest/r/planner/core/indexmerge_path.result
AilinKid: make sense

pkg/planner/core/indexmerge_path.go
```diff
@@ -1440,6 +1451,56 @@ func (ds *DataSource) collectFilters4MVIndexMutations(filters []expression.Expre
...
+}
+
+// indexMergeContainSpecificIndex checks whether the index merge path contains at least one index in the `indexSet`
+func indexMergeContainSpecificIndex(path *util.AccessPath, indexSet map[int64]struct{}) bool {
+	if path.PartialIndexPaths == nil {
+		return false
+	}
+	for _, p := range path.PartialIndexPaths {
+		// each partial index path can be an index scan, or still an index merge
+		if len(p.PartialIndexPaths) > 0 {
```
time-and-fate, line 1487: Currently it's impossible, an index merge access path can only be "a single layer".
It's better to update the comments, while I think it's ok to leave this code.

pkg/planner/core/indexmerge_path.go
```diff
@@ -108,6 +108,8 @@ func (ds *DataSource) generateIndexMergePath() error {
 		}
 	}
 
+	ds.cleanAccessPathForMVIndexHint()
```
time-and-fate, line 111: Looks like this function will interfere with the logic below and we may need to move it to the end of this function.

YangKeao: Done.

pkg/planner/core/indexmerge_path.go
YangKeao: Thanks for the information :heart:  . I added a comment here.

PR #50185, opened 2024-01-08
planner: choose the better unique index

### LINKED ISSUE #50184: Choose the better unique index
## Enhancement

If both of unique index need to double scan, we should better choose the unique index which can match all of where predicates.
The reason is that if the unique index can match the all of where predicates, it will skip the double scan when the qualify row is zero.

## For example
1. create table
```
CREATE TABLE t ( a int NOT NULL ,  b int NOT NULL,
			c varchar(64) NOT NULL ,  d varchar(64) NOT NULL  ,
			UNIQUE KEY ua (a),
			UNIQUE KEY uab (a, b));
```

2. query
```
select * from t where a=1 and b=1;
```

3. choose the uab index

uab index is better than ua. When there is no row can match the where predicates "a=1 and b=1", the uab index can skip the double scan while the ua index can not.

```
"Point_Get 1.00 root table:t, index:uab(a, b) "
```

elsa0520:
Fix #50184

Issue Number: close #50184

Problem Summary:

If both of unique index need to double scan, we should better choose the unique index which can match all of where predicates.
The reason is that if the unique index can match the all of where predicates, it will skip the double scan when the qualify row is zero.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/stats.go
```diff
@@ -359,7 +359,12 @@ func (ds *DataSource) derivePathStatsAndTryHeuristics() error {
 		for _, uniqueIdx := range uniqueIdxsWithDoubleScan {
 			uniqueIdxAccessCols = append(uniqueIdxAccessCols, uniqueIdx.GetCol2LenFromAccessConds(ds.SCtx()))
 			// Find the unique index with the minimal number of ranges as `uniqueBest`.
-			if uniqueBest == nil || len(uniqueIdx.Ranges) < len(uniqueBest.Ranges) {
+			/*
+				If the range is same but one of unique index can match all of where predicates, it will better than current unique index.
+				Example in the test "TestPointgetIndexChoosen".
+			*/
+			if uniqueBest == nil || len(uniqueIdx.Ranges) < len(uniqueBest.Ranges) ||
+				(len(uniqueIdx.Ranges) == len(uniqueBest.Ranges) && len(uniqueIdx.TableFilters) == 0) {
```
time-and-fate, line 367: Looks like this would be better.
```suggestion
				(len(uniqueIdx.Ranges) == len(uniqueBest.Ranges) && len(uniqueIdx.TableFilters) == 0 && len(uniqueBest.TableFilters) > 0) {
```

pkg/planner/core/casetest/physicalplantest/testdata/plan_suite_out.json
```diff
@@ -3719,5 +3719,17 @@
...
       }
     ]
+  },
+  {
+    "Name": "TestPointgetIndexChoosen",
+    "Cases": [
+      {
+        "SQL": "select * from t where a=1 and b=1;",
+        "Plan": [
+          "Point_Get 1.00 root table:t, index:uab(a, b) "
```
AilinKid, line 3729: seems currently tidb can do the right thing for this query, any other meaningful cases?
image

pkg/planner/core/stats.go
elsa0520: If the both unique index table filter > 0 , there are same efficient  plan. 
So we just set the uniqueBest when current unique index has no table filter.
Otherwise, we just pick the first one (same as original logic)

pkg/planner/core/casetest/physicalplantest/testdata/plan_suite_out.json
elsa0520: I already change the test to more accurate test this case

AilinKid (approved):

go.sum
```diff
@@ -15,43 +17,162 @@ cloud.google.com/go v0.62.0/go.mod h1:jmCYTdRCQuc1PHIIJ/maLInMho30T/Y0M4hTdTShOY
...
+cloud.google.com/go/apigateway v1.6.4/go.mod h1:0EpJlVGH5HwAN4VF4Iec8TAzGN1aQgbxAWGJsnPCGGY=
+cloud.google.com/go/apigeeconnect v1.6.4/go.mod h1:CapQCWZ8TCjnU0d7PobxhpOdVz/OVJ2Hr/Zcuu1xFx0=
+cloud.google.com/go/apigeeregistry v0.8.2/go.mod h1:h4v11TDGdeXJDJvImtgK2AFVvMIgGWjSb0HRnBSjcX8=
+cloud.google.com/go/appengine v1.8.4/go.mod h1:TZ24v+wXBujtkK77CXCpjZbnuTvsFNT41MUaZ28D6vg=
+cloud.google.com/go/area120 v0.8.4/go.mod h1:jfawXjxf29wyBXr48+W+GyX/f8fflxp642D/bb9v68M=
+cloud.google.com/go/artifactregistry v1.14.6/go.mod h1:np9LSFotNWHcjnOgh8UVK0RFPCTUGbO0ve3384xyHfE=
+cloud.google.com/go/asset v1.15.3/go.mod h1:yYLfUD4wL4X589A9tYrv4rFrba0QlDeag0CMcM5ggXU=
+cloud.google.com/go/assuredworkloads v1.11.4/go.mod h1:4pwwGNwy1RP0m+y12ef3Q/8PaiWrIDQ6nD2E8kvWI9U=
+cloud.google.com/go/automl v1.13.4/go.mod h1:ULqwX/OLZ4hBVfKQaMtxMSTlPx0GqGbWN8uA/1EqCP8=
+cloud.google.com/go/baremetalsolution v1.2.3/go.mod h1:/UAQ5xG3faDdy180rCUv47e0jvpp3BFxT+Cl0PFjw5g=
```
hawkingrei, line 33: I think you should remove unnecessary update.

go.sum
```diff
@@ -1,3 +1,5 @@
+4d63.com/gocheckcompilerdirectives v1.2.1/go.mod h1:yjDJSxmDTtIHHCqX0ufRYZDL6vQtMG7tJdKVeWwsqvs=
+4d63.com/gochecknoglobals v0.2.1/go.mod h1:KRE8wtJB3CXCsb1xy421JfTHIIbmT3U5ruxw2Qu8fSU=
```
AilinKid, line 2: go sum shouldn't be changed

hawkingrei (approved):

PR #50357, opened 2024-01-12
disttask/ddl: fix hang when reverting

ywqzzy:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/disttask/framework/taskexecutor/manager.go
+++ b/pkg/disttask/framework/taskexecutor/manager.go
@@ -237,17 +237,18 @@ func (m *Manager) handleExecutableTasks(tasks []*proto.Task) {
 			continue
 		}
 		logutil.Logger(m.logCtx).Info("detect new subtask", zap.Int64("task-id", task.ID))
+		if task.State == proto.TaskStateRunning {
+			canAlloc, tasksNeedFree := m.slotManager.canAlloc(task)
+			if len(tasksNeedFree) > 0 {
+				m.cancelTaskExecutors(tasksNeedFree)
+				// do not handle the tasks with lower priority if current task is waiting tasks free.
+				break
+			}
 
-		canAlloc, tasksNeedFree := m.slotManager.canAlloc(task)
-		if len(tasksNeedFree) > 0 {
-			m.cancelTaskExecutors(tasksNeedFree)
-			// do not handle the tasks with lower priority if current task is waiting tasks free.
-			break
-		}
-
-		if !canAlloc {
-			logutil.Logger(m.logCtx).Debug("no enough slots to run task", zap.Int64("task-id", task.ID))
-			continue
+			if !canAlloc {
+				logutil.Logger(m.logCtx).Debug("no enough slots to run task", zap.Int64("task-id", task.ID))
+				continue
+			}
 		}
 		m.addHandlingTask(task.ID)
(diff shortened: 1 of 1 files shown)


pkg/disttask/framework/taskexecutor/manager.go
```diff
@@ -237,17 +237,18 @@ func (m *Manager) handleExecutableTasks(tasks []*proto.Task) {
 			continue
 		}
 		logutil.Logger(m.logCtx).Info("detect new subtask", zap.Int64("task-id", task.ID))
+		if task.State == proto.TaskStateRunning {
```
tangenta, line 240: Maybe put this condition to `canAlloc()` is better.

pkg/disttask/framework/taskexecutor/slot.go
```diff
@@ -37,6 +37,9 @@ type slotManager struct {
 
 // subtasks inside a task will be run in serial, so they takes task.Concurrency slots.
 func (sm *slotManager) alloc(task *proto.Task) {
+	if task.State != proto.TaskStateRunning {
+		return
+	}
```
D3Hunter, line 42: shouldn't we fix the case that disttask is still `reverting`, but ddl mark it as `rollback done`

it's ok for reverting task to take slots, as it's transient normally

ywqzzy: I am working on it

pkg/disttask/framework/scheduler/state_transform.go
```diff
@@ -38,11 +38,8 @@ func VerifyTaskStateTransform(from, to proto.TaskState) bool {
 		proto.TaskStateSucceed: {},
 		proto.TaskStateReverting: {
 			proto.TaskStateReverted,
-			// no revert_failed now
-			// proto.TaskStateRevertFailed,
 		},
-		proto.TaskStateFailed:       {},
-		proto.TaskStateRevertFailed: {},
```
ywqzzy, line 45: Remove it since we will remove all reverting logic

D3Hunter:
> Revert process run forever since we didn't mark failed revert subtask as revert_failed.

if we have `revert_failed` subtask, task need to be this state too. but i don't think we should maintain it

maybe we can fix the `Init` fail in , so current framework logic can continuehttps://github.com/pingcap/tidb/blob/095227d4ef77f15ef4e29f0e4b9782175c884f54/pkg/ddl/backfilling_dist_executor.go#L150

ywqzzy:
> > Revert process run forever since we didn't mark failed revert subtask as revert_failed.
> 
> if we have `revert_failed` subtask, task need to be this state too. but i don't think we should maintain it
> 
> maybe we can fix the `Init` fail in below part, so current framework logic can keep working
> 
> https://github.com/pingcap/tidb/blob/095227d4ef77f15ef4e29f0e4b9782175c884f54/pkg/ddl/backfilling_dist_executor.go#L147

Just a tmp fix for previous version of tidb, we can fire another pr to remove reverting process completely.

Issue #50215, opened 2024-01-09
Tracking issue: pipelined DML


ekexium:
## Enhancement

- Main changes are tracked in TiKV: https://github.com/tikv/tikv/issues/16291

PR #50389, opened 2024-01-12
txn: introduce pipelined DML

you06:

Issue Number: ref #50215
Problem Summary:

Currently TiDB stores all the mutations in its memory before committing, which is the main bottleneck of large DML in many usecases.

Support pipelined DML, which runs large DML within less memory.

* Add a variable `tidb_dml_type` to control the feature, can be `standard` and `bulk`. When `tidb_dml_type` is set to `bulk`, auto committed DML will be executed in pipelined DML.
* Written the mutations into pipelined memdb and check flush every row.
* The main implementation is in the client side, see tikv/client-go#1114.
image

#### Before:
image

#### After:
image

The added metrics of pipelined flush:
image
image
image

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Implement pipelined DML to enable TiDB to execute unlimited size DMLs.
```


you06:
Mark it "ready for review" for running CI. Actually it's not really ready.

pkg/session/txn.go
```diff
@@ -683,29 +683,36 @@ func (txnFailFuture) Wait() (uint64, error) {
...
+		return tf.store.Begin(tikv.WithTxnScope(tf.txnScope), tikv.WithPipelinedMemDB())
+	}
 	return tf.store.Begin(tikv.WithTxnScope(tf.txnScope))
 }
 
 // HasDirtyContent checks whether there's dirty update on the given table.
 // Put this function here is to avoid cycle import.
 func (s *session) HasDirtyContent(tid int64) bool {
-	if s.txn.Transaction == nil {
+	if s.txn.Transaction == nil || s.txn.Transaction.IsPipelined() {
```
ekexium, line 715: Why put this? If it has to, can we add comments explaining it?

pkg/sessionctx/variable/sysvar.go
```diff
@@ -3039,6 +3039,11 @@ var defaultSysVars = []*SysVar{
 			vars.EnableParallelSort = TiDBOptOn(s)
 			return nil
 		}},
+	{Scope: ScopeGlobal | ScopeSession, Name: TiDBEnablePipelinedDML, Value: BoolToOnOff(DefTiDBEnablePipelinedTxn), Type: TypeBool,
```
ekexium, line 3042: The name needs change.

pkg/store/driver/txn/unionstore_driver.go
```diff
@@ -67,20 +68,30 @@ func (m *memBuffer) GetFlags(key kv.Key) (kv.KeyFlags, error) {
 }
 
 func (m *memBuffer) Staging() kv.StagingHandle {
+	if m.isPipelinedDML {
+		// 0 stands for staging not supported.
+		return 0
```
ekexium, line 73: We should panic in these methods to avoid incorrect operations.

ekexium: And add checks before every usage of these methods.

ekexium (commented):
We also need to appropriately ignore `ErrNotFound`s when using `GetFlags` if it is on a pipelined memdb.

pkg/executor/insert.go
```diff
@@ -92,6 +92,7 @@ func (e *InsertExec) exec(ctx context.Context, rows [][]types.Datum) error {
 			return err
 		}
 	} else {
+		ctx = kv.WithPipelinedMemDBSkipRemoteBuffer(ctx)
```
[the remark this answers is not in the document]
you06: It's ok to reomve this optimization, but there will be a significant performance loss without this.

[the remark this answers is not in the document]
ekexium: We must guarantee correctness before performance. As long as we are confident in this optimization we can keep it.

pkg/store/driver/txn/unionstore_driver.go
you06: I did this follow BR's implemenration and the comment.

https://github.com/pingcap/tidb/blob/ac057ebbc37c7dfee2df1d31a43d18db1cbf672a/br/pkg/lightning/backend/kv/session.go#L174-L177

https://github.com/pingcap/tidb/blob/ac057ebbc37c7dfee2df1d31a43d18db1cbf672a/pkg/table/tables/mutation_checker.go#L86-L89

If we want to panic in those case, better to add a method `SupportStaging` and avoid calling `Staging` and `Release` and `Cleanup` for BR also.

pkg/session/txn.go
you06: Comment added.

pkg/executor/insert.go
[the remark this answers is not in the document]
you06: Removed.

pkg/store/driver/txn/unionstore_driver.go
ekexium: Okay it's not a big problem

pkg/sessionctx/variable/session.go
```diff
@@ -2801,6 +2801,9 @@ type Concurrency struct {
 
 	// IdleTransactionTimeout indicates the maximum time duration a transaction could be idle, unit is second.
 	IdleTransactionTimeout int
+
+	// BulkDMLParallelism indicates whether to enable pipelined transaction.
+	BulkDMLParallelism int
```
ekexium, line 2806: There is no parallelism in current implementation. And renaming it won't break anything. So I suppose we don't to name it so aggressively. Can be a simple boolean

pkg/sessionctx/variable/sysvar.go
```diff
@@ -3065,6 +3065,19 @@ var defaultSysVars = []*SysVar{
 			vars.EnableParallelSort = TiDBOptOn(s)
 			return nil
 		}},
+	{Scope: ScopeGlobal | ScopeSession, Name: TiDBDMLType, Value: DefTiDBDMLType, Type: TypeStr,
```
ekexium, line 3068: Should be ScopeSession only.

pkg/sessionctx/variable/tidb_vars.go
```diff
@@ -1154,6 +1154,10 @@ const (
 	TiDBEnableParallelSort = "enable_parallel_sort"
 	// TiDBLowResolutionTSOUpdateInterval defines how often to refresh low resolution timestamps.
 	TiDBLowResolutionTSOUpdateInterval = "tidb_low_resolution_tso_update_interval"
+	// TiDBDMLType indicates whether the execution type of DML in TiDB.
```
ekexium, line 1157:
```suggestion
	// TiDBDMLType indicates the execution type of DML in TiDB.
```

pkg/store/driver/txn/unionstore_driver.go
ekexium: Is it problematic that we do not check usages of these methods in tables.go?

you06: Yes, a little obscure because it's actually no effects and no error returned.

pkg/kv/option.go
```diff
@@ -273,3 +274,6 @@ func getLossyDDLReorgSource(txnSource uint64) uint64 {
 func isLossyDDLReorgSourceSet(txnSource uint64) bool {
 	return (txnSource >> lossyDDLReorgSourceShift) != 0
 }
+
+// WithPipelinedMemDBSkipRemoteBuffer create context will skip remote buffer.
+var WithPipelinedMemDBSkipRemoteBuffer = tikv.WithPipelinedMemDBSkipRemoteBuffer
```
crazycs520, line 279: This is never used?

pkg/executor/insert.go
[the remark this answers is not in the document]
you06: @crazycs520 it's a risky optimization, so we don't use it in this PR.

pkg/kv/option.go
you06: Yes, it's an risky optimization, you can check the earlier discussion in this PR about this.

pkg/session/pipelineddmltest/pipelineddml_test.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	tk.MustQuery("select * from t order by a asc").Check(testkit.Rows(results...))
+}
+
+func TestPipelinedDMLInsert(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	prepareData(tk)
+	tk.MustExec("set session tidb_dml_type = bulk")
```
crazycs520, line 63: Could you add a test that without pipelined DML, the following bulk insert will fail cause by exceeded memory limit?

pkg/session/pipelineddmltest/pipelineddml_test.go
```diff
@@ -0,0 +1,156 @@
...
+	tk.MustExec("create table _t like t")
+	results := make([]string, 0, 10000)
+	for i := 0; i < 10000; i++ {
+		tk.MustExec("insert into t values (?, ?)", i, i)
+		results = append(results, fmt.Sprintf("%d %d", i, i))
+	}
+	tk.MustQuery("select * from t order by a asc").Check(testkit.Rows(results...))
+}
+
+func TestPipelinedDMLInsert(t *testing.T) {
```
crazycs520, line 58: Could you add a test case that pipelined DML failed cause by duplicated key error? and the duplicate key error is detected during flush memory table?

Another similar test case with upper, but use `insert ignore`, then TiDB should ignore the duplicate key error that detected during flush memory table.

pkg/session/session.go
```diff
@@ -3815,10 +3815,13 @@ func (s *session) PrepareTSFuture(ctx context.Context, future oracle.Future, sco
 		future = txnFailFuture{}
 	})
 
+	pipelined := s.sessionVars.BulkDMLEnabled && s.sessionVars.IsAutocommit() && !s.sessionVars.InTxn() &&
```
cfzjywxk, line 3818: Is this the only entry that auto-commit transaction would be activiated? Is the prepared protocol execution path also covered here?

cfzjywxk (commented):
It seems there is not much executor changes for pipelined DML, is it by design because all the details are already wrapped in the client-go implementations?

pkg/session/session.go
```diff
@@ -3815,10 +3815,13 @@ func (s *session) PrepareTSFuture(ctx context.Context, future oracle.Future, sco
...
+		!config.GetGlobalConfig().PessimisticTxn.PessimisticAutoCommit.Load()
```
ekexium, line 3819: Can we add a check for internal stmts?

pkg/kv/option.go
crazycs520: Since it is never used in TiDB, can we remove it, I'm worried if someone misuses this for performance optimization in the future.

ekexium (changes requested):
Let's forbid statement retrying (using pessimistic mode) for p-dml.

pkg/session/session.go
ekexium: And for `BinlogClient == nil`

pkg/session/session.go
```diff
@@ -4296,6 +4297,26 @@ func (s *session) NewStmtIndexUsageCollector() *indexusage.StmtIndexUsageCollect
...
+// isPipelinedDML returns the current statement can be executed as a pipelined DML.
+func (s *session) isPipelinedDML() bool {
+	if !s.sessionVars.BulkDMLEnabled {
+		return false
+	}
+	stmtCtx := s.sessionVars.StmtCtx
+	if stmtCtx == nil {
+		return false
+	}
+	if !stmtCtx.InInsertStmt && !stmtCtx.InDeleteStmt && !stmtCtx.InUpdateStmt {
```
ekexium, line 4309: Just a question, how about `LOAD DATA`? We may not support it but it's good to clarify. Because in release notes we mentioned "any DML"

you06: Just claim we donot support `LOAD DATA` yet(maybe in the future)?

cfzjywxk: `LOAD DATA` is a main entry for the usages of large tranctions, we need to clarify with PM if we decide not to support it by now.

pkg/session/pipelineddmltest/pipelineddml_test.go
you06: I added it in the end of this file, note the mem quota is carefully choosed because when mutations are not such large(reach GB level), memory used by read executors can be large also.

pkg/session/session.go
you06: Added test for positive & negative cases.

ekexium (approved):

cfzjywxk (approved):

pkg/session/pipelineddmltest/pipelineddml_test.go
```diff
@@ -0,0 +1,392 @@
...
+	tk1.MustExec("set session tidb_dml_type = bulk")
+	errCh := make(chan error)
+	go func() {
+		errCh <- tk1.ExecToErr("update t1 set b = b + 20")
+	}()
+	time.Sleep(500 * time.Millisecond)
+	tk2.MustExec("update t1 set b = b + 10")
+	require.Nil(t, failpoint.Disable("tikvclient/beforePipelinedFlush"))
+	err := <-errCh
+	require.Error(t, err)
```
crazycs520, line 391: How about check the error msg here, I want to known what's kind of error here.

you06: Added, it's a write conflict error.

you06:
@benjamin2037  @easonn7 this PR need your approve

I slightly change the membuffer interface and make lightning backend to compatible with it.

Besides, new sys variable is introduced.

tests/realtikvtest/txntest/pipelineddml_test.go
```diff
@@ -0,0 +1,433 @@
...
+			buf.WriteString(fmt.Sprintf("(%d, %d, 'abcdefghijklmnopqrstuvwxyz1234567890,.?+-=_!@#$&*()_+')", i, i))
+		}
+	}
+	buf.WriteString(" on duplicate key update b = values(b) + 2000")
+	tk.MustExec("set session tidb_dml_type = bulk")
+	fmt.Println(buf.String())
+	tk.MustExec(buf.String())
+	tk.MustQuery("select count(*) from t1").Check(testkit.Rows(fmt.Sprintf("%d", 1999)))
+	tk.MustQuery("select a, b from t1 where a = 250").Check(testkit.Rows("250 2250"))
+}
```
crazycs520, line 265:
```suggestion
func TestPipelinedDMLInsertOnDuplicateKeyUpdateInTxn(t *testing.T) {
	require.Nil(t, failpoint.Enable("tikvclient/pipelinedMemDBMinFlushKeys", `return(10)`))
	require.Nil(t, failpoint.Enable("tikvclient/pipelinedMemDBMinFlushSize", `return(100)`))
	require.Nil(t, failpoint.Enable("tikvclient/pipelinedMemDBForceFlushSizeThreshold", `return(10240)`))
	defer func() {
		require.Nil(t, failpoint.Disable("tikvclient/pipelinedMemDBMinFlushKeys"))
		require.Nil(t, failpoint.Disable("tikvclient/pipelinedMemDBMinFlushSize"))
		require.Nil(t, failpoint.Disable("tikvclient/pipelinedMemDBForceFlushSizeThreshold"))
	}()
	store := realtikvtest.CreateMockStoreAndSetup(t)
	tk := testkit.NewTestKit(t, store)
	tk.MustExec("use test")
	tk.MustExec("drop table if exists t1")
	tk.MustExec("create table t1 (a int, b int, c varchar(128), unique index idx(b))")
	cnt := 2000
	values := bytes.NewBuffer(make([]byte, 0, 10240))
	for i := 0; i < cnt; i++ {
		if i > 0 {
			values.WriteString(", ")
		}
		if i == 1500 {
			values.WriteString(fmt.Sprintf("(%d, %d, 'abcdefghijklmnopqrstuvwxyz1234567890,.?+-=_!@#$&*()_+')", i, 250))
		} else {
			values.WriteString(fmt.Sprintf("(%d, %d, 'abcdefghijklmnopqrstuvwxyz1234567890,.?+-=_!@#$&*()_+')", i, i))
		}
	}
	tk.MustExec("set session tidb_dml_type = bulk")
	// Test insert meet duplicate key error.
	tk.MustGetErrMsg("insert into t1 values "+values.String(), "[kv:1062]Duplicate entry '250' for key 't1.idx'")
	tk.MustQuery("select count(*) from t1").Check(testkit.Rows("0"))
	// Test insert ignore
	tk.MustExec("insert ignore into t1 values " + values.String())
	tk.MustQuery("show warnings").Check(testkit.Rows("Warning 1062 Duplicate entry '250' for key 't1.idx'"))
	tk.MustQuery("select count(*) from t1").Check(testkit.Rows("1999"))
	tk.MustQuery("select a, b from t1 where a in (250, 1500)").Check(testkit.Rows("250 250"))
	// Test insert on duplicate key update.
	if !*realtikvtest.WithRealTiKV {
		// TODO: fix me. skip for real TiKV because now we have assertion issue.
		tk.MustExec("delete from t1")
		tk.MustExec("insert into t1 values " + values.String() + " on duplicate key update b = values(b) + 2000")
		tk.MustQuery("select count(*) from t1").Check(testkit.Rows("1999"))
		tk.MustQuery("select a, b from t1 where a in (250, 1500)").Check(testkit.Rows("250 2250"))
	}
}
```

Add more test case.

tests/realtikvtest/txntest/pipelineddml_test.go
```diff
@@ -0,0 +1,440 @@
...
+	// Test insert meet duplicate key error.
+	tk.MustGetErrMsg("insert into t1 values "+values.String(), "[kv:1062]Duplicate entry '250' for key 't1.idx'")
+	tk.MustQuery("select count(*) from t1").Check(testkit.Rows("0"))
+	// Test insert ignore
+	tk.MustExec("insert ignore into t1 values " + values.String())
+	tk.MustQuery("show warnings").Check(testkit.Rows("Warning 1062 Duplicate entry '250' for key 't1.idx'"))
+	tk.MustQuery("select count(*) from t1").Check(testkit.Rows("1999"))
+	tk.MustQuery("select a, b from t1 where a in (250, 1500)").Check(testkit.Rows("250 250"))
+	// Test insert on duplicate key update.
+	if !*realtikvtest.WithRealTiKV {
```
crazycs520, line 265:
```suggestion
	if !*realtikvtest.WithRealTiKV {
		// TODO: fix me. skip for real TiKV because now we have assertion issue.
		// Test replace into.
		tk.MustExec("delete from t1")
		tk.MustExec("replace into t1 values " + values.String())
		tk.MustQuery("select count(*) from t1").Check(testkit.Rows("1999"))
		tk.MustQuery("select a, b from t1 where a in (250, 1500)").Check(testkit.Rows("1500 250"))
		// Test insert on duplicate key update.
```

crazycs520: add test for replace into

you06:
`//tests/realtikvtest/txntest:txntest_test                               TIMEOUT in 3 out of 3 in 300.7s`

Reduced the test data size.

crazycs520:
Please add following test with some change, otherwise, some execution statistics will be missed.

```go
func TestPipelinedDMLInsertRPC(t *testing.T) {
	tk.MustExec("use test")
	tk.MustExec("drop table if exists t1")
	tk.MustExec("create table t1 (a int, b int, unique index idx(b))")
	res := tk.MustQuery("explain analyze insert ignore into t1 values (1,1), (2,2), (3,3), (4,4), (5,5)")
	explain := getExplainResult(res)
	require.Regexp(t, "Insert.* check_insert: {total_time: .* rpc:{BatchGet:{num_rpc:1, total_time:.*}}}.*", explain)

	// Test with bulk dml.
	tk.MustExec("set session tidb_dml_type = bulk")
	// Test normal insert.
	tk.MustExec("truncate table t1")
	res = tk.MustQuery("explain analyze insert into t1 values (1,1), (2,2), (3,3), (4,4), (5,5)")
	explain = getExplainResult(res)
	// TODO: try to optimize the rpc count, when use bulk dml, insert will send many BufferBatchGet rpc.
	require.Regexp(t, "Insert.* insert:.*, rpc:{BufferBatchGet:{num_rpc:10, total_time:.*}}.*", explain)
	// Test insert ignore.
	tk.MustExec("truncate table t1")
	res = tk.MustQuery("explain analyze insert ignore into t1 values (1,1), (2,2), (3,3), (4,4), (5,5)")
	explain = getExplainResult(res)
	// TODO: try to optimize the rpc count, when use bulk dml, insert ignore will send 5 BufferBatchGet and  1 BatchGet rpc.
	// but without bulk dml, it will only use 1 BatchGet rpcs.
	require.Regexp(t, "Insert.* check_insert: {total_time: .* rpc:{.*BufferBatchGet:{num_rpc:5, total_time:.*}}}.*", explain)
	require.Regexp(t, "Insert.* check_insert: {total_time: .* rpc:{.*BatchGet:{num_rpc:1, total_time:.*}}}.*", explain)
}

func getExplainResult(res *testkit.Result) string {
	resBuff := bytes.NewBufferString("")
	for _, row := range res.Rows() {
		_, _ = fmt.Fprintf(resBuff, "%s\t", row)
	}
	return resBuff.String()
}
```

some code change diff:

```diff
diff --git a/pkg/executor/insert.go b/pkg/executor/insert.go
index 99bf773f2a..21eb2b0fbb 100644
--- a/pkg/executor/insert.go
+++ b/pkg/executor/insert.go
@@ -92,7 +92,12 @@ func (e *InsertExec) exec(ctx context.Context, rows [][]types.Datum) error {
                        return err
                }
        } else {
-               e.collectRuntimeStatsEnabled()
+               if e.collectRuntimeStatsEnabled() {
+                       if snapshot := txn.GetSnapshot(); snapshot != nil {
+                               snapshot.SetOption(kv.CollectRuntimeStats, e.stats.SnapshotRuntimeStats)
+                               defer snapshot.SetOption(kv.CollectRuntimeStats, nil)
+                       }
+               }
                start := time.Now()
                for i, row := range rows {
                        var err error
diff --git a/pkg/executor/insert_common.go b/pkg/executor/insert_common.go
index 77d25e7a26..762e0a9b4c 100644
--- a/pkg/executor/insert_common.go
+++ b/pkg/executor/insert_common.go
@@ -1493,6 +1493,11 @@ func (e *InsertRuntimeStat) String() string {
                buf.WriteString("}")
        } else {
                fmt.Fprintf(buf, "insert:%v", execdetails.FormatDuration(e.CheckInsertTime))
+               if e.SnapshotRuntimeStats != nil {
+                       if rpc := e.SnapshotRuntimeStats.String(); len(rpc) > 0 {
+                               fmt.Fprintf(buf, ", rpc:{%s}", rpc)
+                       }
+               }
        }
        return buf.String()
 }
```

crazycs520 (approved):

Issue #50358, opened 2024-01-12
Incorrect cross join results for mpp queries occasionally


yibin87:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
create table t0 ( a int , b int);
create table t1( c int);
insert into t0 values(10, 20);
insert into t1 values(10);
alter table t0 set tiflash replica 1;
alter table t1 set tiflash replica 1;
set session tidb_isolation_read_engines='tiflash';
select 5 from t0 join t1;

### 2. What did you expect to see? (Required)
+---+
| 5 |
+---+
| 5 |
+---+

### 3. What did you see instead (Required)
Empty result sets.
Note, it is hard to reproduce the issue manually. The reproduce possiblity is rather low.

yibin87:
The issue is introduced by several steps during optimization:
Part of the initial logicalPlan looks like this:  Join(t0, t1) => Projection(schema: constant_column)
1. rule_join_reorder
For cross join node, this rule reorders t0 join t1 to t1 join t0 due to its cost estimation algorithm. And it detects schema changes, and try to add a new projection above join node:
https://github.com/pingcap/tidb/blob/17d65b9317e054c80f781680d2412081de6556c6/pkg/planner/core/rule_join_reorder.go#L316C21-L324C21
Now, the LogicalPlan will have two projections as parent and child:  Join(t0, t1) => Projection1(schema: old_join_schema) => Projection(schema: constant_column)
2. rule_column_pruning
This rule detects that none of the Projection1's output columns are needed by Projection, thus it prunes all the schema, and the LogicalPlan looks like this:
 Join(t0, t1) => Projection1(schema: empty_schema) => Projection(schema: constant_column)
3. Generate physical plan:
When init PhysicalTableReader node, its schema will be set to the Projection1's schema, and won't be updated again
https://github.com/pingcap/tidb/blob/17d65b9317e054c80f781680d2412081de6556c6/pkg/planner/core/initialize.go#L484C1-L484C33
4. Build MPPGather:
MppGather node reuses TableReader's schema
https://github.com/pingcap/tidb/blob/17d65b9317e054c80f781680d2412081de6556c6/pkg/executor/builder.go#L3345
5. Decode mpp response:
Decode uses MppGather's schema to decode tiflash response, and produce empty result
[...]

PR #50425, opened 2024-01-15
executor: check schema when build table reader executor

yibin87:

Issue Number: ref #50358

Problem Summary:
Described in #50358
Quick fix for the issue in executor. Had tried to fix it in planner phase, but failed with so many test failures: https://github.com/pingcap/tidb/pull/50410
May try to fix it in planner phase later.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
Try to run cross join queries with high concurrency, and check no empty result sets are found
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/builder.go
+++ b/pkg/executor/builder.go
@@ -3416,6 +3416,10 @@ func (b *executorBuilder) buildTableReader(v *plannercore.PhysicalTableReader) e
 			failpoint.Return(nil)
 		}
 	})
+	// https://github.com/pingcap/tidb/issues/50358
+	if len(v.Schema().Columns) == 0 && len(v.GetTablePlan().Schema().Columns) > 0 {
+		v.SetSchema(v.GetTablePlan().Schema())
+	}
 	useMPP := useMPPExecution(b.ctx, v)
 	useTiFlashBatchCop := v.ReadReqType == plannercore.BatchCop
 	useTiFlash := useMPP || useTiFlashBatchCop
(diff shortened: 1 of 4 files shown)


pkg/executor/builder.go
```diff
@@ -3416,6 +3416,10 @@ func (b *executorBuilder) buildTableReader(v *plannercore.PhysicalTableReader) e
 			failpoint.Return(nil)
 		}
 	})
+	// https://github.com/pingcap/tidb/issues/50358
```
elsa0520, line 3419: Please add the test ~

yibin87: ok

elsa0520 (changes requested):
This is not a good changed for this issue.
We should keep the tablereader schema correctly in planner.

yibin87:
> This is not a good changed for this issue. We should keep the tablereader schema correctly in planner.

Yeah, that's the reason why this is not a PR that closes the issue but ref it. This PR aims to pass some regression tests. I tried  to fix the issue in planner phase in #50410 and recorded some other tries, while introducing many test failures. Thus I submit this PR to pass regression tests.

elsa0520:
> > This is not a good changed for this issue. We should keep the tablereader schema correctly in planner.
> 
> Yeah, that's the reason why this is not a PR that closes the issue but ref it. This PR aims to pass some regression tests. I tried to fix the issue in planner phase in #50410 and recorded some other tries, while introducing many test failures. Thus I submit this PR to pass regression tests.

OK, for quickly pass the test I can approve this PR. But we should directly fix this issue by following planner PR later.

elsa0520 (approved):
Later to fix it correctly.

windtalker (approved):
LGTM

PR #50550, opened 2024-01-18
disttask: remove revert subtask

### LINKED ISSUE #50549: disttask: remove reverting subtask
## Enhancement
Revert subtask is used for rollback task. It's created when the task failed or cancelled on each executor node of the task, and they have the same step as the current task step. It has the following issues:
- There's no rollback in any business type now, they are only used to notify us that we need cancel other pending/running subtasks which we can do without Revert subtask.
- Even though we DO need to rollback later, there's no need to create one subtask on each node.
- Node scale in/out, and those revert subtask are not balanced, so they will block the task. And even if they are balanced to active nodes, it's meaningless to run them there.

If rollback is required later, we can:
- Have a special step to do rollback.
- Rollback shouldn't be bind to some node. We can split the rollback into multiple subtasks though.

If we want to do some cleanup(such as cleanup local sorted kv) on each node, we should do it on subtask cleanup. Or start a background routine to do it if cleanup fails, or tidb exit abnormally.

D3Hunter:

Issue Number: close #50549

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/disttask/framework/storage/task_state.go
```diff
@@ -59,6 +59,19 @@ func (mgr *TaskManager) FailTask(ctx context.Context, taskID int64, currentState
...
 }
 
+// RevertTask implements the scheduler.TaskManager interface.
+func (mgr *TaskManager) RevertTask(ctx context.Context, taskID int64, taskState proto.TaskState, taskErr error) error {
+	_, err := mgr.ExecuteSQLWithNewSession(ctx, `
+		update mysql.tidb_global_task
+		set state = %?,
+			error = %?,
+			state_update_time = CURRENT_TIMESTAMP()
+		 where id = %? and state = %?`,
```
ywqzzy, line 69:
```suggestion
		where id = %? and state = %?`,
```

pkg/disttask/framework/testutil/task_util.go
```diff
@@ -28,12 +28,8 @@ import (
 
 // CreateSubTask adds a new task to subtask table.
 // used for testing.
-func CreateSubTask(t *testing.T, gm *storage.TaskManager, taskID int64, step proto.Step, execID string, meta []byte, tp proto.TaskType, concurrency int, isRevert bool) {
-	state := proto.SubtaskStatePending
-	if isRevert {
-		state = proto.SubtaskStateRevertPending
-	}
-	InsertSubtask(t, gm, taskID, step, execID, meta, state, tp, concurrency)
+func CreateSubTask(t *testing.T, gm *storage.TaskManager, taskID int64, step proto.Step, execID string, meta []byte, tp proto.TaskType, concurrency int) {
```
ywqzzy, line 31: This function is useless now

pkg/disttask/framework/proto/task.go
```diff
@@ -20,40 +20,37 @@ import (
...
-//		   │             │└────────►│reverting├────►│reverted│
-//		   │             ▼          └────┬────┘     └────────┘
-//		   │          ┌──────────┐    ▲  │          ┌─────────────┐
-//		   └─────────►│cancelling├────┘  └─────────►│revert_failed│
-//		              └──────────┘                  └─────────────┘
-//	 1. succeed:		pending -> running -> succeed
-//	 2. failed:			pending -> running -> reverting -> reverted/revert_failed, pending -> failed
-//	 3. canceled:		pending -> running -> cancelling -> reverting -> reverted/revert_failed
-//	 4. pause/resume:	pending -> running -> pausing -> paused -> running
+// Note: if task fail during running, it will end with `reverted` state.
```
ywqzzy, line 23:
```suggestion
// Note: if a task fails during running, it will end with `reverted` state.
```

ywqzzy (approved):

pkg/disttask/framework/testutil/task_util.go
D3Hunter: a wrapper to create pending subtask

pkg/disttask/framework/storage/table_test.go
```diff
@@ -93,12 +84,10 @@ func TestTaskTable(t *testing.T) {
...
 	require.GreaterOrEqual(t, task4[0].StateUpdateTime, task.StateUpdateTime)
 
-	prevState := task.State
-	task.State = proto.TaskStateRunning
-	retryable, err := gm.UpdateTaskAndAddSubTasks(ctx, task, nil, prevState)
+	err = gm.SwitchTaskStep(ctx, task, proto.TaskStateRunning, proto.StepOne, nil)
 	require.NoError(t, err)
-	require.True(t, retryable)
 
+	task.State = proto.TaskStateRunning
```
okJiang, line 90: how about updating state and step in `SwitchTaskStep`, and returning the new task?

okJiang (commented):
rest lgtm

pkg/disttask/framework/storage/table_test.go
D3Hunter: that's side effects we should avoid

okJiang (approved):

PR #50574, opened 2024-01-18
planner: Correct cardinality estimate for subset index match | tidb-test=pr/2286

### LINKED ISSUE #50573: Cardinality estimate can be lower for partial index match compared to full match
## Enhancement
Given a multi-column index of 3 or more columns, an index match on a subset of the full index can result in a lower cardinality (number of rows) estimate as compared to a full match on the index. It should not be possible that fewer predicates results in fewer rows qualifying (assuming the predicates in the subset are an exact match for the same columns in the superset predicates).

The below example is a simple 3 column table, with an index on all 3 columns. Rows are inserted into the table such that there are 1 million rows, but the number of distinct values of the 3 columns = 10,000.

```
tidb> select count(distinct a, b, c), count(*) from t2;
+-------------------------+----------+
| count(distinct a, b, c) | count(*) |
+-------------------------+----------+
|                   10000 |  1000000 |
+-------------------------+----------+
```

The following estimates from 2 explain statements demonstrate the problem. With 3 equal predicates on all index columns, the estimate uses 1/(index cardinality) * table cardinality = 1/10000 * 1000000 = 100 (estimate is 99.52 due to sampling from ANALYZE).

```
tidb> explain select * from t2 where a = 1 and b = 1 and c = 1;
+------------------------+---------+-----------+-------------------------------+---------------------------------------+
| id                     | estRows | task      | access object                 | operator info                         |
+------------------------+---------+-----------+-------------------------------+---------------------------------------+
| IndexReader_6          | 99.52   | root      |                               | index:IndexRangeScan_5                |
| └─IndexRangeScan_5     | 99.52   | cop[tikv] | table:t2, index:iabc(a, b, c) | range:[1 1 1,1 1 1], keep order:false |
+------------------------+---------+-----------+-------------------------------+---------------------------------------+
```

With predicates on only 2 columns, the estimate is 3.14.

```
tidb> explain select * from t2 where a = 1 and b = 1;
+------------------------+---------+-----------+-------------------------------+-----------------------------------+
| id                     | estRows | task      | access object                 | operator info                     |
+------------------------+---------+-----------+-------------------------------+-----------------------------------+
| IndexReader_6          | 3.14    | root      |                               | index:IndexRangeScan_5            |
| └─IndexRangeScan_5     | 3.14    | cop[tikv] | table:t2, index:iabc(a, b, c) | range:[1 1,1 1], keep order:false |
+------------------------+---------+-----------+-------------------------------+-----------------------------------+
```

The formula for a subset index match sorts the column cardinalities in desc sequence, and the multiplies the 1st by the sqrt of the 2nd (and continues with the 3rd root for the 3rd column, 4th root of the 4th etc).

```
tidb> select count(distinct a), count(distinct b) from t2;
+-------------------+-------------------+
| count(distinct a) | count(distinct b) |
+-------------------+-------------------+
|             10000 |              1000 |
+-------------------+-------------------+
```

1/10000 * sqrt(1/1000) * 1000000 = 3.16 (close enough to 3.14).

Table DDL follows.

```
tidb> show create table t2;
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                                     |
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| t2    | CREATE TABLE `t2` (
  `a` int(11) DEFAULT NULL,
  `b` int(11) DEFAULT NULL,
  `c` int(11) DEFAULT NULL,
  KEY `iabc` (`a`,`b`,`c`)
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

terry1purcell:

Issue Number: close #50573 

Problem Summary:

Corrects an issue where the estimated number of rows can be less if predicates exist on a subset of the index columns compared to full index match.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Improves the estimated number of rows when predicates exist on a subset of the full index, and the combined cardinalities of the individual columns exceed the full index cardinality. This can result in a larger estimated number of rows when predicates are applied to a subset of the full index columns.
```


winoros (approved):

pkg/planner/cardinality/row_count_index.go
```diff
@@ -478,14 +478,29 @@ func expBackoffEstimation(sctx sessionctx.Context, idx *statistics.Index, coll *
...
-	} else if l == 3 {
-		return singleColumnEstResults[0] * math.Sqrt(singleColumnEstResults[1]) * math.Sqrt(math.Sqrt(singleColumnEstResults[2])), true, nil
 	} else if l == 0 {
 		return 0, false, nil
 	}
-	return singleColumnEstResults[0] * math.Sqrt(singleColumnEstResults[1]) * math.Sqrt(math.Sqrt(singleColumnEstResults[2])) * math.Sqrt(math.Sqrt(math.Sqrt(singleColumnEstResults[3]))), true, nil
+	// Do not allow the exponential backoff to go below the available index bound. If the number of predicates
+	// is less than the number of index columns - use 90% of the bound to differentiate a subset from full index match.
+	// If there is an individual column selectivity that goes below this bound, use that selectivity only.
+	idxLowBound := max(1/float64(idx.NDV), 1/float64(coll.RealtimeCount))
```
hawkingrei, line 487:
```suggestion
	idxLowBound := 1/float64(min(idx.NDV, coll.RealtimeCount))
```

More simple.

terry1purcell: Are there situations where idx.NDV is unavailable? And will this result in us choosing an incorrect lower bound?

terry1purcell: Changed the code as recommended. But also added a check to make sure that we don't run into a divide by zero error for idx.NDV.

pkg/planner/cardinality/row_count_index.go
```diff
@@ -478,14 +478,33 @@ func expBackoffEstimation(sctx sessionctx.Context, idx *statistics.Index, coll *
...
+	// If there is an individual column selectivity that goes below this bound, use that selectivity only.
+	histNDV := coll.RealtimeCount
+	if idx.NDV > 0 {
+		histNDV = idx.NDV
+	}
+	idxLowBound := 1 / float64(min(histNDV, coll.RealtimeCount))
+	if l < len(idx.Info.Columns) {
+		idxLowBound /= 0.9
+	}
+	minTwoCol := min(singleColumnEstResults[0], min(singleColumnEstResults[1], idxLowBound))
```
time-and-fate, line 495:
```suggestion
	minTwoCol := min(singleColumnEstResults[0], singleColumnEstResults[1], idxLowBound)
```

time-and-fate (commented):
Just a golang comment. Others LGTM.
Though there should be some bad cases in theory, I think their influence and risk should be low.

pkg/planner/cardinality/row_count_index.go
terry1purcell: Applied recommended code change

hawkingrei (approved):

hawkingrei:
/hold

@elsa0520 PTAL

PR #50624, opened 2024-01-22
variable: remove `StrictSQLMode` in `variable.SessionVars`

### LINKED ISSUE #50589: remove `StrictSQLMode` in `SessionVars`
## Enhancement

`StrictSQLMode` is introduced by an earlier PR: #1263 which does not support full SQL mode. And then in PR #2739, full SQL mode is supported. Now we should guarantee that `StrictSQLMode == SQLMode.HasStrictMode()` see:

https://github.com/pingcap/tidb/blob/188726aec62936b2cf805de053cb28d7d290cd4f/pkg/sessionctx/variable/sysvar.go#L1449-L1450

However, this is not always true in the current implementation. We always forget to set `StrictSQLMode` after setting `SQLMode`, see here: https://github.com/pingcap/tidb/blob/188726aec62936b2cf805de053cb28d7d290cd4f/pkg/ddl/backfilling_scheduler.go#L163 and here: https://github.com/pingcap/tidb/blob/188726aec62936b2cf805de053cb28d7d290cd4f/br/pkg/lightning/backend/kv/session.go#L289

I'm unsure whether we have some bugs if `StrictSQLMode` is missing for these cases, but at least the above codes have some potential risks.

Since keeping two truths is so annoying and not easy to management, I think it's better to remove `StrictSQLMode` and we should always use `SQLMode` as the single truth.

lcwangchao:

Issue Number: close #50589

Problem Summary:

see description in issue

remove `StrictSQLMode` in `variable.SessionVars`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


hawkingrei (approved):

pkg/expression/builtin_time_test.go
```diff
@@ -41,6 +41,7 @@ import (
 
 func TestDate(t *testing.T) {
 	ctx := createContext(t)
+	ctx.GetSessionVars().SQLMode = mysql.DelSQLMode(ctx.GetSessionVars().SQLMode, mysql.ModeNoZeroDate)
```
XuHuaiyu, line 44: Why modify this

lcwangchao: The `mock.NewContext` returns a context with an empty sql mode but set `StrictSQLMode` to true in the previous implement. So in the current implementation we set SQL mode to the default `DefaultSQLMode` to make it a real strict mode to make some UT pass. 

`DefaultSQLMode` contains `NO_ZERO_DATE` date, it will cause some cases fail in this test (this test suppose `NO_ZERO_DATE` not set).

lcwangchao: But I think this line can be moved before: https://github.com/pingcap/tidb/pull/50624/files/8dc99b08aa6d569da36fc734aea6c87ff01aa88c#diff-210d7ae9ffb6cec2ad0af174f954bfcfa98ed15aa3a5efe2b831dfcefb0e20d2R231

XuHuaiyu (approved):

tangenta (approved):

PR #50643, opened 2024-01-22
executors: record the index usage in each executors

### LINKED ISSUE #50261: Record the index usages at each related executors.

YangKeao:

Issue Number: close #50261

Problem Summary:

1. Add a `IndexUsageCollector` in the `executor` package, to help dump stats info from the executor and report them to the collector in the statement. You can find the collector in `pkg/executor/internal/exec/indexusage.go`
2. Use this collector to collect the usage for each executors, include `IndexReaderExecutor`, `IndexLookUpExecutor`, `IndexMergeReaderExecutor`, `PointGetExecutor`, `BatchPointGetExecutor`.
   1. It will pre-load stats data before building the executor. See function `LoadTableStats` for each physical plan.
   2. The builder has a helper function to build the index usage collector: see `b.buildIndexUsageReporter(v)`.
   3. In the `Close` of each executor, the index usage information is uploaded.

Note: for partition table and local index, most of the time, the index usage is recorded in a partition and the percentage is calculated according to the partition size. However, for `BatchPointGet` which may access multiple partition through a single request, this PR records the stats on the table itself.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

### Release note

```release-note
None
```


pkg/executor/internal/exec/indexusage.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	accessRows := basic.GetActRows()
+
+	sample := indexusage.NewSample(0, uint64(kvRequestTotal), uint64(accessRows), uint64(tableRowCount))
+	e.reporter.Update(tableID, indexID, sample)
+}
+
+// getTableRowCount returns the `RealtimeCount` of a table
+func (e *IndexUsageReporter) getTableRowCount(tableID int64) (tableRowCount int64) {
+	if e.statsMap != nil {
```
qw4990, line 74: If the stats is empty or pseudo, should we skip this report operation, since it can pollute our metrics?

YangKeao: I agree with you :+1: . I'll check that the version is not `PseudoVersion`.

I think the stats shouldn't be empty. I pre-load the table statistics if it doesn't appear in the `statsMap` before building the executors. Do you mean it should be skipped when the table is empty? The current behavior is that it'll be recorded in the `100%` bucket :thinking: I think it's also fine to skip, as the definition is not so good.

YangKeao: Done

qw4990 (approved):
LGTM

pkg/statistics/handle/usage/indexusage/collector.go
```diff
@@ -232,3 +232,38 @@ func (s *SessionIndexUsageCollector) Flush() {
...
+	// here.
+	s.Lock()
+	defer s.Unlock()
+
+	idxID := GlobalIndexID{IndexID: indexID, TableID: tableID}
+	if _, ok := s.recordedIndex[idxID]; !ok {
+		sample.QueryTotal = 1
+		s.recordedIndex[idxID] = struct{}{}
+	} else {
+		sample.QueryTotal = 0
```
bb7133, line 265: IMO the logic here is a bit confusing(and tricky?), please add some comments here.

bb7133:
LGTM

bb7133 (approved):
LGTM

PR #50787, opened 2024-01-29
lightning: fix duplicate extend columns when resume from checkpoint

### LINKED ISSUE #50786: lightning: Column count doesn't match value count at row 1 when use extract-schema
## Bug Report

create task with extract-schema in dm
pause task in load stage
resume task

task succeed
task failed with [error="Error 1136 (21S01): Column count doesn't match value count at row 1"]

master

GMHDBJD:

Issue Number: close #50786

Problem Summary:

- copy ColumnPermutation for extend columns so it will not be saved in checkpoint
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
```
1. manually create dm task with extend column
2. pause task in load stage
3. resume task and task succeed
```
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/lightning/importer/chunk_process.go
+++ b/br/pkg/lightning/importer/chunk_process.go
@@ -301,9 +301,13 @@ func (cr *chunkProcessor) encodeLoop(
 	// but since ColumnPermutation also depends on the hypothesis that the columns in one source file is the same
 	// so this should be ok.
 	var (
-		filteredColumns []string
-		extendVals      []types.Datum
+		filteredColumns   []string
+		extendVals        []types.Datum
+		// some columns should not record in checkpoint, like extra columns
+		// so we copy a new one from chunk.ColumnPermutation
+		columnPermutation = make([]int, 0, len(cr.chunk.ColumnPermutation))
 	)
+	columnPermutation = append(columnPermutation, cr.chunk.ColumnPermutation...)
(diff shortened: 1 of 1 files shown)


br/pkg/lightning/importer/chunk_process.go
```diff
@@ -362,6 +366,7 @@ func (cr *chunkProcessor) encodeLoop(
 						if err = t.initializeColumns(columnNames, cr.chunk); err != nil {
```
lance6716, line 366: I prefer we move the "extend column" related logic into `initializeColumns`

D3Hunter, line 366: `cr.chunk.ColumnPermutation` is still filled and stored into checkpoint here

D3Hunter: maybe we can fix it here https://github.com/pingcap/tidb/blob/aab2da047904901725e5a3bf745b47b1488eb8c7/br/pkg/lightning/importer/chunk_process.go#L102-L104

when there's extend column, no need to set columns

lance6716: It's OK to store them into checkpoint, but it must not to add extended columns again after loading from checkpoint

D3Hunter: the ColumnPermutation is correct, the problem here is that we have 1 more column generated after restart, see code link above

PR #50816, opened 2024-01-30
executor: record the index usage on a the whole table

### LINKED ISSUE #50814: Record the index usage on the whole table rather than each partition
## Enhancement

I proposed to record the index usage on each partition, and then aggregate them together when the user is querying the `tidb_index_usage` table. However, it turns out that I was wrong 🤦 . The `QueryTotal` will not be recorded correctly, because a single query on different partition will not be deduplicated by the `StmtIndexUsageCollector`. For example:

```sql
create table t (id_1 int, unique key idx_1(id_1))
  partition by range (id_1) (
  partition p0 values less than (10),
  partition p1 values less than (20),
  partition p2 values less than (50),
  partition p3 values less than MAXVALUE);

select id_1 from t where id_1 >= 30;
```

A single query can make the `QueryTotal` greater than 1, because the usage on each partition will add `1` to the `QueryTotal`.

The simplest way to fix it is using the id of the whole table to record the index usage, but be careful that the we'll still use the physical table's row count to calculate the percentage, so the physical table id is still needed.

YangKeao:

Issue Number: close #50814

Problem Summary:

Previously, the usage is recorded on each partitions, so that the queries on different partitions will not be deduplicated 🤦  and will be recorded multiple times.

This PR adds one more argument on the `ReportPointGetIndexUsage` and `ReportCopIndexUsage` function to get both the `physicalTableID` and `tableID`.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/batch_point_get.go
+++ b/pkg/executor/batch_point_get.go
@@ -168,10 +168,9 @@ func (e *BatchPointGetExec) Close() error {
 	}
 	if e.indexUsageReporter != nil && e.idxInfo != nil {
 		kvReqTotal := e.stats.GetCmdRPCCount(tikvrpc.CmdBatchGet)
-		// We cannot distinguish how many rows are coming from each partition. Here, we record all index usages on the
-		// table itself (but not for the partition physical table), which is different from how an index usage is
-		// treated for a local index.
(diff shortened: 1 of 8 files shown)


bb7133 (approved):
LGTM

qw4990 (approved):

Issue #49883, opened 2023-12-28
import into from select


D3Hunter:
## Feature Request

**Is your feature request related to a problem? Please describe:**

“Insert into from select” statement do normal insert in one txn, the performance can not satisfy user expection(see https://github.com/pingcap/tidb/issues/48637 too), so we add a similar  `IMPORT into from select` which uses physical import same as lightning to speed up.
**Describe the feature you'd like:**

IMPORT into doesn't support transaction, and the target table must be a empty one. Selected rows will be encoded into kv, sorted and write to local disk, data will be ingest into tikv directly, just as what lightning physical mode does.

during prototype test, it will speed up 2~8 times compared to INSERT INTO depends on scenario and configured concurrency.

It also support import from snapshot, but only using `as of timestamp 'xxx'` inside the `select-statement` is supported, set `tidb_snapshot` is not.

### the grammar:
```sql
IMPORT INTO table[column, column...] FROM select-statement [WITH thread=xx]

EXPLAIN import-into-statement
```

### tasks:
- [x] https://github.com/pingcap/tidb/pull/49884
- [x] https://github.com/pingcap/tidb/pull/49936
- [x] https://github.com/pingcap/tidb/pull/49976
- [x] https://github.com/pingcap/tidb/pull/50341
- [x] https://github.com/pingcap/tidb/pull/50558
- [x] https://github.com/pingcap/tidb/pull/50852
- [x] https://github.com/pingcap/tidb/pull/51077

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

AndrewDi:
Could `SELECT Statement` support `tidb_snapshot`?

D3Hunter:
> Could `SELECT Statement` support `tidb_snapshot`?

not now, can you describe your usage scenario?

AndrewDi:
> > Could `SELECT Statement` support `tidb_snapshot`?
> 
> not now, can you describe your usage scenario?

We need to replicate a `table with specific timestamp ` to a new table

D3Hunter:
> > > Could `SELECT Statement` support `tidb_snapshot`?
> > 
> > 
> > not now, can you describe your usage scenario?
> 
> We need to replicate a `table with specific timestamp ` to a new table

how frequent you do it?

what's the new table used for?
- as a backup? why not use backup&restore tool?
- as a persisted snapshot (avoid gc) that can be queried for analysis later?

AndrewDi:
[...]

PR #50852, opened 2024-01-31
importinto: support import from stale read

D3Hunter:

Issue Number: ref #49883

Problem Summary:

support import from stale read, but only using `as of timestamp 'xxx'` inside the `select-statement` is supported, set `tidb_snapshot` (history snapshot read) is not.
### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/planbuilder.go
+++ b/pkg/planner/core/planbuilder.go
@@ -47,7 +47,6 @@ import (
 	"github.com/pingcap/tidb/pkg/sessionctx"
 	"github.com/pingcap/tidb/pkg/sessionctx/stmtctx"
 	"github.com/pingcap/tidb/pkg/sessionctx/variable"
-	"github.com/pingcap/tidb/pkg/sessiontxn"
 	"github.com/pingcap/tidb/pkg/sessiontxn/staleread"
 	"github.com/pingcap/tidb/pkg/statistics"
 	"github.com/pingcap/tidb/pkg/table"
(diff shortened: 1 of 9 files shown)


pkg/planner/core/planbuilder.go
```diff
@@ -4217,8 +4217,22 @@ func (b *PlanBuilder) buildImportInto(ctx context.Context, ld *ast.ImportIntoStm
 		b.visitInfo = appendVisitInfo(b.visitInfo, mysql.FilePriv, "", "", "", plannererrors.ErrSpecificAccessDenied.GenWithStackByArgs("FILE"))
 	}
 	tableInfo := p.Table.TableInfo
-	tableInPlan, ok := b.is.TableByID(tableInfo.ID)
+	// we use the latest IS to support IMPORT INTO dst FROM SELECT * FROM src AS OF TIMESTAMP '2020-01-01 00:00:00'
+	// Note: we need to get p.Table when preprocessing, at that time, IS of session
```
lance6716, line 4221: My understanding is, `p.Table.TableInfo` may be a latest or stale TableInfo. And in expectation the target table of IMPORT INTO should always be latest TableInfo.

And my suggestions are:
1. Can we modify the preprocessing to add a special branch for IMPORT INTO? I see there's such a special branch for `InsertStmt` https://github.com/pingcap/tidb/blob/134d2bfbe0002b24177155873038188d4544e538/pkg/planner/core/preprocess.go#L249-L253
2. This PR uses a "may be stale" table ID with a "latest" IS. This is a little weird. Can we use `TableByName` to get the latest TableInfo again and check the latest table ID is the same as `tableInfo.ID`?

D3Hunter: i don't want to support the case that user uses import into when the session is already in stale-read state by set `tidb_snapshot`

> Can we modify the preprocessing to add a special branch for IMPORT INTO?

that makes no difference, both requires special handling

> Can we use TableByName to get the latest TableInfo again and check the latest table ID is the same as tableInfo.ID?

checking existence of `maybe stale id`  VS  get latest id and compare, both lgtm, as tidb doesn't reuse id.

lance6716: >> Can we modify the preprocessing to add a special branch for IMPORT INTO?
>
>that makes no difference, both requires special handling

The difference is we get a latest TableInfo vs we get an uncertain TableInfo and check it has the same table ID with latest one. Will there be problems that the TableInfo of two different versions have same table ID? For example, some DDL has modified the columns or indices and table ID is not changed.

D3Hunter: when will this happen without in a stale-read session?

lance6716: Seems only happen in the session that already set `tidb_snapshot`. In this function we report error only when latest TableInfo is not found, so I'm afraid an old version TableInfo with the same table ID will cause some problems, when it's accessed after this function and before `handleNoDelayExecutor`. I didn't see problems now, so it's OK.

And in this page https://docs.pingcap.com/tidb/stable/basic-features#advanced-sql-features I found 2 terminology
"Stale read" and "Read historical data", I prefer these comment use "history snapshot read", "tidb_snapshot read" or some thing rather than "stale read" when refer to  `tidb_snapshot`

lance6716 (commented):
rest lgtm

pkg/planner/core/planbuilder.go
D3Hunter: we're mixing them in here https://docs.pingcap.com/tidb/stable/as-of-timestamp

pkg/planner/core/planbuilder.go
```diff
@@ -4217,8 +4217,22 @@ func (b *PlanBuilder) buildImportInto(ctx context.Context, ld *ast.ImportIntoStm
...
+	// transaction is used, if the session ctx is already in stale read, we might
```
D3Hunter, line 4222:
```suggestion
	// transaction is used, if the session ctx is already in snapshot read using tidb_snapshot, we might
```

lance6716 (approved):

PR #9933, opened 2019-03-28
types: fix incorrect time fraction parsing method

WangXiangUSTC:
### What problem does this PR solve? 
fix issue https://github.com/pingcap/tidb/issues/9734
tidb don't support date time format like "2019.01.01"

### What is changed and how it works?
some minor update, and add unit test

### Check List 

Tests 

 - Unit test

Related changes

 - Need to cherry-pick to the release branch

PR #50958, opened 2024-02-04
expression: Fix the issue that cast str as datetime with time_zone results in loss of precision

### LINKED ISSUE #49555: Insert a special datetime with time_zone results in loss of precision
## Bug Report

```
create table t(a datetime(3));
insert into t values ('2020-01-01 12:00:00.123456 +0600 PST'); # The format is Golang’s default format
insert into t values ('2020-01-01 12:00:00.123456 -0600 PST');
select * from t;
```

Got error during insert or keep result like `2020-01-01 12:00:00.123`

MySQL will report error during insert.

```
mysql> select * from t;
+-------------------------+
| a                       |
+-------------------------+
| 2020-01-01 12:00:00.123 |
| 2020-01-01 12:00:00.000 |
+-------------------------+
2 rows in set (0.01 sec)
```

https://github.com/pingcap/tidb/commit/23953a7cbe82a18390f2a213777a57f48e203d18

SeaRise:

Issue Number: close #49555

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that cast str as datetime with time_zone results in loss of precision
```


pkg/types/time_test.go
```diff
@@ -114,6 +114,8 @@ func TestDateTime(t *testing.T) {
 
 		// For issue 35291
 		{"2020-01-01 12:00:00.123456+05:00", "2020-01-01 07:00:00.123456"},
+		// For issue 49555
+		{"2020-01-01 12:00:00.123456-05:00", "2020-01-01 07:00:00.123456"},
```
Defined2014, line 118:
```suggestion
		{"2020-01-01 12:00:00.123456-05:00", "2020-01-01 17:00:00.123456"},
```

pkg/types/time.go
```diff
@@ -587,7 +590,7 @@ func GetFracIndex(s string) (index int) {
 	}
 	index = -1
 	for i := end; i >= 0; i-- {
-		if unicode.IsPunct(rune(s[i])) {
+		if s[i] != '+' && s[i] != '-' && unicode.IsPunct(rune(s[i])) {
```
Defined2014, line 593: We have some functions in `types/time.go` and `types/helper.go`, like `isValidSeparator` and `isPunctuation`. Not sure could we directly use them.

SeaRise: It looks like there is no function available.
BTW, I replace `unicode.IsPunct` by `isPunctuation`.

Defined2014: Ok, just a question more than suggestion.

Seems `punctuation` in different language has its own meaning. I think `isPunctuation` function is followed `std::ispunct` which may different with `unicode.IsPunct`. Not sure which one is correct.

SeaRise: As `isPunctuation` was used consistently in time.go, I have replaced it here. It seems that in time.go, either all function should use `isPunctuation`, or all instances should use `unicode.IsPunct`.

SeaRise: I believe replacing it with `isPunctuation` is safe here, as this code snippet was introduced by https://github.com/pingcap/tidb/pull/9933 to handle the case like `2019.01.01 00:00:00`.

Defined2014 (approved):

windtalker (approved):
LGTM

PR #50970, opened 2024-02-04
planner: adjust estimated rows to account for modified rows

### LINKED ISSUE #47523: Repeatedly updating the same row will cause the stats_healthy to become low and generate wrong query plan
## Report
Repeatedly updating the same row will cause the stats_healthy to become low, which affects the correctness of the execution plan

```
mysql> create table t(a int);
mysql> insert into t values(1),(1),(1),(1),(1),(1),(1),(1);
Query OK, 8 row affected (0.01 sec)
mysql> show stats_meta where table_name='t';
| test    | t          |                | 2023-10-10 16:36:50 |            0 |         8 |
1 row in set (0.00 sec)
mysql> update t set a=2 where a=1 limit 1;
Rows matched: 1  Changed: 1  Warnings: 
mysql> update t set a=3 where a=2 limit 1;
Rows matched: 1  Changed: 1  Warnings: 0
mysql> select * from t;
+------+
| a    |
+------+
|    3 |
|    1 |
|    1 |
|    1 |
|    1 |
|    1 |
|    1 |
|    1 |
+------+
8 rows in set (0.00 sec)

mysql> update t set a=4 where a=3 limit 1;
Rows matched: 1  Changed: 1  Warnings: 0
```

```
mysql> show stats_meta where table_name='t';
| test    | t          |                | 2023-10-10 16:40:47 |            1 |         8 |
1 row in set (0.00 sec)

mysql> show stats_healthy where table_name='t';
+---------+------------+----------------+---------+
| Db_name | Table_name | Partition_name | Healthy |
+---------+------------+----------------+---------+
| test    | t          |                |      88 |
+---------+------------+----------------+---------+
1 row in set (0.00 sec)

```

```
mysql> show stats_meta where table_name='t';
| test    | t          |                | 2023-10-10 16:40:47 |            3 |         8 |
1 row in set (0.00 sec)

mysql> show stats_healthy where table_name='t';
+---------+------------+----------------+---------+
| Db_name | Table_name | Partition_name | Healthy |
+---------+------------+----------------+---------+
| test    | t          |                |      12 |
+---------+------------+----------------+---------+
1 row in set (0.00 sec)
```

v6.5.3

terry1purcell:

Issue Number: close #47523 

Problem Summary: Out of range estimation has two significant deficiencies that impact the reported issue:
1. The NDV used is for the full index, which may be too high if predicates match a subset of the columns.
2. The out-of-range estimation does adequately account for the fact that modifications may include inserts (or updates) that mean that the WHERE clause predicate is now in range.

Estimated row calculation is changed to account for modify row count and usage of the subset NDV if one exists.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


terry1purcell:
reopen after rebase

pkg/statistics/histogram.go
```diff
@@ -1052,38 +1052,41 @@ func (hg *Histogram) OutOfRangeRowCount(
...
+	// If the modifyCount is large (as a percentage of the base table rows), then any out of range estimate is
+	// unreliable - because this indicates that our statistics are stale. These bounds are therefore approximations.
+	// ModifyCount counts all inserts/updates/deletes - this logic has a bias towards modifications being inserts.
+	// Assume a minimum of 1 value is found if we have a high modifyCount. This targest the situation where we search
+	// for a statisically out of range value, but inserts result in the value actually being in that modified range.
+	modifyPercent := min(float64(modifyCount)/hg.NotNullCount(), 1.0)
+	if modifyPercent == 1 || modifyPercent > totalPercent {
+		rowCount = max(rowCount, upperBound)
+	} else if rowCount < upperBound {
+		rowCount *= hg.GetIncreaseFactor(realtimeRowCount)
```
time-and-fate, line 1086: Here is a small pitfall.
`hg.GetIncreaseFactor` can only consider the row count in the histogram. The row count in TopN and NULL is not considered. But `realtimeRowCount` counts everything.
So probably it's better to calculate the "increase factor" using `(*Index).GetIncreaseFactor()` or `(*Column).GetIncreaseFactor()` in the outside and pass it into this function.

pkg/statistics/histogram.go
```diff
@@ -1052,38 +1052,41 @@ func (hg *Histogram) OutOfRangeRowCount(
...
+		// We use 1/NDV here (only the Histogram part is considered) and it seems reasonable and good enough for now.
+		return min(rowCount, upperBound)
+	}
+	// If the modifyCount is large (as a percentage of the base table rows), then any out of range estimate is
+	// unreliable - because this indicates that our statistics are stale. These bounds are therefore approximations.
+	// ModifyCount counts all inserts/updates/deletes - this logic has a bias towards modifications being inserts.
+	// Assume a minimum of 1 value is found if we have a high modifyCount. This targest the situation where we search
+	// for a statisically out of range value, but inserts result in the value actually being in that modified range.
+	modifyPercent := min(float64(modifyCount)/hg.NotNullCount(), 1.0)
+	if modifyPercent == 1 || modifyPercent > totalPercent {
```
time-and-fate, line 1083: The `modifyPercent > totalPercent` is a bit hard to understand for me.
`totalPercent` represents how large the out-of-range part is.
It looks like our criteria for stale stats are changing according to the size of the out-of-range part.

terry1purcell: As you say - totalPercent represents the out-of-range portion. But this is calculated based upon histograms that were collected at last ANALYZE. NotNullCount is the number of table rows minus NULLs - also collected from last ANALYZE.
realtimeRowCount and modifyCount are a count of the changes since last ANALYZE. It is my understanding that realtime/modifyCount do not update the histograms. Therefore, the larger the number of changes since the histograms were created - the greater the risk that any "totalPercent" (out of range estimate).
Therefore - what the above code is attempting to do - if the percentage of modifications to the original table (modifyCount/NotNullCount) is greater than the percentage we estimate out of range - it is possible that all of the rows we estimate to be out of range are actually within the modifyCount.

time-and-fate: Yes. Realtime row count and modify count don't update the histograms.


So, to my core question, if I understand it correctly, we are indeed using the relative size of modifications and out-of-range portion as criteria, which means that, if the modification is large, or if the out-of-range portion is small, we think the lower bound is needed.

pkg/statistics/histogram.go
terry1purcell: Corrected in latest revision

pkg/statistics/histogram.go
terry1purcell: Correct - but I have attempted to address your concerns about the risk of using modifyCount in the latest revision.

time-and-fate (approved):

qw4990 (approved):

qw4990:
This PR is great but is the related issue number correct? #47523 seems about the display and update of modify_cnt and stats_healthy, not about cardinality estimation.

PR #50997, opened 2024-02-06
expression: refactor functions to build expressions

### LINKED ISSUE #50996: refactor methods to provide more clear interfaces to to build expressions
see descriptions in #50997

lcwangchao:

Issue Number: close #50996

We have many functions to build expressions, but it is unclear which we should use for a certain case and some codes in these functions are duplicated. It's better to refactor them to make it clear.

In this PR, we provide two groups of functions to build expressions.

The first function group is exposed in the package `expression` and used to build simple expressions. The "simple" means they do not include some special expressions such as subquery. Here we simplified to 3 functions:

- `expression.BuildSimpleExpr`: build a simple expression from an AST node.
- `expression.ParseSimpleExpr`: parse a string and then build a simple expression
- `expression.EvalSimpleAst`: build a simple expression from an AST node and evaluate it directly.

The first group requires limited context and can be used in some places where you do not want to introduce too many dependencies. However, if you need to build some complex expressions like subquery, we should use the second group. The second group is exposed in the package `planner/util` and can build any expression. However, it requires the full context of a planner. These functions includes:

- `util.RewriteAstExprWithPlanCtx`: build an expression with the plan context
- `util.EvalAstExprWithPlanCtx`: evaluates an expression directly with the plan context
- `util.ParseExprWithPlanCtx`: parse an expression from the string with the plan context

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


lance6716 (approved):

YangKeao (approved):
LGTM

pkg/planner/core/expression_test.go
```diff
@@ -40,22 +40,40 @@ func parseExpr(t *testing.T, expr string) ast.ExprNode {
 	return stmt.Fields.Fields[0].Expr
 }
 
-func buildExpr(t *testing.T, ctx sessionctx.Context, exprNode any, opts ...expression.BuildOption) (expression.Expression, error) {
+func buildExpr(t *testing.T, ctx sessionctx.Context, exprNode any, opts ...expression.BuildOption) (expression.Expression, expression.Expression, error) {
```
mjonss, line 43: (Minor comment, since it is in a test...)
Why having a function that depending on input returns an expression to either of the two resulting values, depending on the input type?
Would it not be more clear to have two separate functions? Or to simply return only one result expression, independently of the input?
Also where is the second return value used?

pkg/planner/core/expression_rewriter.go
```diff
@@ -112,6 +122,20 @@ func buildExprWithAst(ctx expression.BuildContext, node ast.ExprNode, opts ...ex
 		return nil, errors.New("InputSchema and InputNames should be the same length")
 	}
 
+	if len(options.InputNames) > 0 {
+		intest.AssertFunc(func() bool {
```
mjonss, line 126: Is it possible to have the `if len(options.InputNames) > 0` inside the `intest.AssertFunc`?

mjonss (approved):

pkg/planner/core/expression_test.go
lcwangchao: It is to check the `ParseSimpleExpr` will build the same expression with parse and rewrite. However, yes, it is a little difficult to understand, I'll change it.

zimulala (approved):

XuHuaiyu (approved):

PR #20610, opened 2020-10-23
*: refactor `CompilePattern` and `DoMatch` used by `like`

xiongjiwei:
### What problem does this PR solve?

- Issue Number: close #20608 
- refactor code

Problem Summary:

there are three `CompilePattern` and `DoMatch` function in different places, they are almost the same but in parameters. This PR aims to unify them as one and provide some adapters. And I believe these two functions should take account of "real string" but "bytes of string".

### Check List 

Tests 

- Unit test

### Release note 

- fix a bug that makes the `like` function get the wrong result if pattern string is a Unicode string

PR #51113, opened 2024-02-18
util: Adjust binary collate behavior to do pattern matching in bytes

### LINKED ISSUE #50393: Join between blob type with matching returns incorrect result.
## Bug Report

```sql
create table t2(a blob);
create table t3(a blob);
insert into t2 values(0xC2A0);
insert into t3 values(0xC2);
select * from t2,t3 where t2.a like concat("%",t3.a,"%");
```

Returns one row, as MySQL(8.0.33) does.

```sql
mysql> select * from t2,t3 where t2.a like concat("%",t3.a,"%");
+------------+------------+
| a          | a          |
+------------+------------+
| 0xC2A0     | 0xC2       |
+------------+------------+
1 row in set (0.01 sec)
```

Returns empty set.

```sql
> select * from t2,t3 where t2.a like concat("%",t3.a,"%");
Empty set (0.00 sec)
```

It looks weird since in the case below, TiDB works well.
```sql
create table t2(a blob);
create table t3(a blob);
insert into t2 values(0xC2A020);
insert into t3 values(0xC2A0);
select * from t2,t3 where t2.a like concat("%",t3.a,"%");
```

In this case, TiDB returns one row:
```sql
> select * from t2,t3 where t2.a like concat("%",t3.a,"%");
+------+------+
| a    | a    |
+------+------+
|      |      |
+------+------+
1 row in set (0.00 sec)
```

which is similar with MySQL.(The difference in display is reported in #50392).

```sql
mysql> select * from t2,t3 where t2.a like concat("%",t3.a,"%");
+------------+------------+
| a          | a          |
+------------+------------+
| 0xC2A020   | 0xC2A0     |
+------------+------------+
1 row in set (0.02 sec)
```

The charsets of TiDB and MySQL are the same(utf8mb4).

```sql
>  SELECT tidb_version();
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| tidb_version()

       |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Release Version: v7.4.0
Edition: Community
Git Commit Hash: 38cb4f3312be9199a983c0ef282d2ea2e28a7824
Git Branch: heads/refs/tags/v7.4.0
UTC Build Time: 2023-10-10 14:18:50
GoVersion: go1.21.1
Race Enabled: false
Store: tikv |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

yibin87:

Issue Number: close #50393

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/util/collate/bin.go
```diff
@@ -65,6 +65,16 @@ func (*binPaddingCollator) Pattern() WildcardPattern {
 	return &binPattern{}
 }
 
+// binPureCollator match pattern in bytes
+type binPureCollator struct {
```
YangKeao, line 69: How about modify the behavior of `binPattern` directly? I don't think there is anything depending on the original wrong behavior 🤔. Is there any special reason to keep the original `binPattern`?

pkg/util/stringutil/string_util.go
```diff
@@ -228,53 +281,76 @@ func CompileLike2Regexp(str string) string {
 	return string(result)
 }
 
-// DoMatchBytes is a adapter for `DoMatchInner`, `str` can only be an ascii string.
+// DoMatchBytes is an adapter for `DoMatchInner`, `str` can only be an ascii string.
 func DoMatchBytes(str string, patChars, patTypes []byte) bool {
```
YangKeao, line 285: Also, I think we don't need to distinguish `DoMatchBytes` and `DoMatchPureBytes`. We can directly modify the behavior of `DoMatchBytes` (under proper tests).

pkg/util/stringutil/string_util.go
```diff
@@ -228,53 +281,76 @@ func CompileLike2Regexp(str string) string {
...
+	bytes := []byte(str)
+	lenBytes := len(bytes)
+	lenPatWeights := len(patChars)
+	return DoMatchInner(lenPatWeights, lenBytes, patTypes, func(a, b int) bool { return bytes[a] == patChars[b] })
 }
 
-// DoMatch is a adapter for `DoMatchInner`, `str` can be any unicode string.
+// DoMatch is an adapter for `DoMatchInner`, `str` can be any unicode string.
 func DoMatch(str string, patChars []rune, patTypes []byte) bool {
-	return DoMatchInner(str, patChars, patTypes, matchRune)
```
YangKeao, line 238: How about `return DoMatchCustomized(str, patChars, patTypes, matchRune)`.

YangKeao (commented):
Rest LGTM.

BTW, I have checked that the behavior of TiKV is correct 👍 . It decodes the character according to the charset.

pkg/util/collate/bin.go
yibin87: According to https://github.com/pingcap/tidb/pull/20610, some utf_bin collation needs to use Rune characters.

pkg/util/stringutil/string_util.go
yibin87: Make sense, I'll try.

pkg/util/collate/bin.go
YangKeao: You are right. I realized that `utf8_bin` and `utf8mb4_bin` are using `binPaddingCollator`, and `utf8mb4_0900_bin` is using `binCollator` 🤦 (and also `gbkBinCollator` uses the `binPattern` directly).

windtalker:
And please double check if TiFlash has the same issue

xiongjiwei (approved):

xzhangxian1008 (approved):
LGTM

pkg/util/stringutil/string_util.go
```diff
@@ -228,53 +281,62 @@ func CompileLike2Regexp(str string) string {
...
-	for pIdx < len(patWeights) || rIdx < lenRunes {
-		if pIdx < len(patWeights) {
+	lenPatWeights := len(patWeights)
+	return DoMatchInner(lenPatWeights, lenRunes, patTypes, func(a, b int) bool { return matcher(runes[a], patWeights[b]) })
+}
+
+// DoMatchInner matches the string with patChars and patTypes.
+// The algorithm has linear time complexity.
+// https://research.swtch.com/glob
+func DoMatchInner(lenPatWeights int, lenChars int, patTypes []byte, matcher func(a, b int) bool) bool {
```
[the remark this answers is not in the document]
yibin87: ok

pkg/util/stringutil/string_util.go
```diff
@@ -138,21 +138,27 @@ const (
...
 func CompilePatternBytes(pattern string, escape byte) (patChars, patTypes []byte) {
 	patWeights, patTypes := CompilePatternInner(pattern, escape)
 	patChars = []byte(string(patWeights))
 
 	return patChars, patTypes
 }
 
-// CompilePattern is a adapter for `CompilePatternInner`, `pattern` can be any unicode string.
+// CompilePatternPureBytes is used for binary strings.
+func CompilePatternPureBytes(pattern string, escape byte) (patChars, patTypes []byte) {
```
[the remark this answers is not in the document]
yibin87: ok

pkg/util/collate/collate.go
```diff
@@ -400,8 +400,8 @@ func init() {
 	newCollatorMap = make(map[string]Collator)
 	newCollatorIDMap = make(map[int]Collator)
 
-	newCollatorMap["binary"] = &binCollator{}
-	newCollatorIDMap[CollationName2ID("binary")] = &binCollator{}
+	newCollatorMap["binary"] = &binPureCollator{}
```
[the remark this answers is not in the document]
yibin87: ok

yibin87:
> And please double check if TiFlash has the same issue

The same issue detected in TiFlash, filed a issue to track it: https://github.com/pingcap/tiflash/issues/8776

windtalker (approved):
LGTM

PR #51128, opened 2024-02-18
planner, executor: enhance the index join's choice

### LINKED ISSUE #47233: planner: support using partial EQ predicates to construct IndexJoins
## Enhancement

See the example below, the third query cannot use `IndexJoin`:
```
create table a(id int,name varchar(100));
create table b(id int,name varchar(100), key(id));

select /*+ INL_JOIN(a,b) */ *  from a,b where a.id=b.id;                                                                -- IndexJoin
select /*+ INL_JOIN(a,b) */ *  from a,b where a.id=b.id and a.name!=substr(b.name,1,10);      -- IndexJoin
select /*+ INL_JOIN(a,b) */ *  from a,b where a.id=b.id and a.name=substr(b.name,1,10);       -- HashJoin, cannot use IndexJoin
```

The root cause is that both `a.id=b.id` and `a.name=substr(b.name)` are `EQ predicates`, so the optimizer **has to** use them both as the `Join Keys` when constructing Joins. 
But `IndexJoin` cannot support `substr`, then the optimizer has to use `HashJoin` for the third query if the two predicates have to be used as `Join Keys`.

A better solution is to only use `a.id=b.id` as the `Join Key` and put `a.name=substr(b.name)` into joins' other conditions to construct IndexJoin:
image

winoros:

Issue Number: close https://github.com/pingcap/tidb/issues/47233

Problem Summary:

Modify the codes to support:
- The projection has expression ->column.
- Not only one of UnionScan/Selection/Projection Exists.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/builder.go
+++ b/pkg/executor/builder.go
@@ -4136,14 +4136,8 @@ func (builder *dataReaderBuilder) buildExecutorForIndexJoinInternal(ctx context.
 		return builder.buildIndexLookUpReaderForIndexJoin(ctx, v, lookUpContents, indexRanges, keyOff2IdxOff, cwc, memTracker, interruptSignal)
 	case *plannercore.PhysicalUnionScan:
 		return builder.buildUnionScanForIndexJoin(ctx, v, lookUpContents, indexRanges, keyOff2IdxOff, cwc, canReorderHandles, memTracker, interruptSignal)
-	// The inner child of IndexJoin might be Projection when a combination of the following conditions is true:
-	// 	1. The inner child fetch data using indexLookupReader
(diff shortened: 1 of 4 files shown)


winoros:
The first round is to test that existing tests are passed.

winoros:
I've added the new case.

elsa0520 (approved):
LGTM

hawkingrei (approved):

PR #51137, opened 2024-02-19
log_backup: added more info for slow regions in log backup advancer

### LINKED ISSUE #51046: log_backup: add some hint about the slow region
## Enhancement
For now, the slow region hint of log backup only provides the range info. It is pretty hard to find the region contains that range in a huge cluster (We need run pd-ctl or something, also note that PD ctl doesn't support querying history data.)
It will be hard to debug checkpoint stuck issues: usually we need to fetch all TiKV's logs and query the start key of some region in *ALL* logs(!). It is painful.

We can fetch the region metadata once there are some slow regions. That will be pretty lightweight.

YuJuncen:

Issue Number: close #51046

Problem Summary:
It is painful for debugging slow regions in log backup for now. Exactly when there are many TiKV nodes in the cluster: we only have a range but know nothing about which region this key belongs to.
Also we must pick the TiDB log (which isn't always easily accessible.) for such information.

This PR added two new series to the log backup advancer, which presents the current slowest region's leader location and its region id.
Also we will append more detailed hints about the slowest range.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
A simple log:
CleanShot 2024-02-19 at 14 20 11@2x
- [ ] No need to test
  > 

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/streamhelper/advancer.go
+++ b/br/pkg/streamhelper/advancer.go
@@ -5,6 +5,7 @@ package streamhelper
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"math"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


br/pkg/streamhelper/advancer.go
```diff
@@ -272,20 +273,49 @@ func (c *CheckpointAdvancer) NewCheckpoints(cps *spans.ValueSortedFull) {
...
 			return true
 		})
 		minValue = vsf.Min()
 	})
-	log.Info("current last region", zap.String("category", "log backup advancer hint"),
+	sctx, cancel := context.WithTimeout(ctx, time.Second)
+	// Always fetch the hint and update the metrics.
+	hint := c.fetchRegionHint(sctx, minValue.Key.StartKey)
+	logger := log.Debug
+	if minValue.Value < thresholdTso {
```
Leavrth, line 310: how about using `<=`?

Leavrth (approved):

BornChanger (approved):

br/pkg/streamhelper/advancer.go
YuJuncen: Given the scale of TSO is millisecond, I guess the difference is minor between `<=` and `<`? 🤔

Issue #42588, opened 2023-03-25
Panic triggered at chunk.appendCellByCell (util/chunk/chunk.go:421) 


JZuming:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

**Setup the environment:**

```
~/tidb/bin/tidb-server &
mysql -h "127.0.0.1" -u root -P 4000 -D testdb < mysql_bk.sql
```
**mysql_bk.sql:** [mysql_bk.txt](https://github.com/pingcap/tidb/files/11069677/mysql_bk.txt)

**Testcase**
```
select  
  (select  
      subq_1.c0 as c0
    from 
      t273 as ref_84
    where exists (
        select  
            (select  
                  ref_86.c1147 as c0
                from 
                  t272 as ref_86
                where (subq_1.c0) > (subq_1.c0)
                window w0 as (partition by ref_86.c1147 order by ref_86.c1143 desc)
                order by c0 limit 1
            ) as c3, 
            (select  
                  subq_1.c0 as c0
                from 
                  t273 as ref_89
                order by c0 limit 1) as c4 
          from 
            t271 as ref_85
          )
    order by c0 desc limit 1) as c1 
from 
  (select 1 as c0) as subq_1;
```


### 2. What did you expect to see? (Required)

No panic.

### 3. What did you see instead (Required)
```
ERROR 1105 (HY000): probeWorker[0] meets error: runtime error: index out of range [1] with length 1
```
Log of TiDB server:
```
err="probeWorker[0] meets error: runtime error: index out of range [1] with length 1
github.com/pingcap/tidb/executor.(*probeWorker).handleProbeWorkerPanic
    /root/tidb/executor/join.go:402
github.com/pingcap/tidb/util.(*WaitGroupWrapper).RunWithRecover.func1.1
    /root/tidb/util/wait_group_wrapper.go:168
runtime.gopanic
    /usr/local/go/src/runtime/panic.go:884
runtime.goPanicIndex
    /usr/local/go/src/runtime/panic.go:113
github.com/pingcap/tidb/util/chunk.appendCellByCell
    /root/tidb/util/chunk/chunk.go:421
github.com/pingcap/tidb/util/chunk.(*Chunk).AppendPartialRowByColIdxs
    /root/tidb/util/chunk/chunk.go:408
github.com/pingcap/tidb/util/chunk.(*Chunk).AppendRowByColIdxs
[...]

PR #45831, opened 2023-08-04
planner: refactor Join and Limit's ResolveIndices

winoros:
### What problem does this PR solve?

Issue Number: close #45758 , close #45805 

Problem Summary:

### What is changed and how it works?

In https://github.com/pingcap/tidb/pull/45158, we use `ResolveIndices` for the inline projection's column matching.

It's possible that the join/limit's child(ren)'s output columns are `[col0, col0, col1]`.
And if we call ResolveIndices, the two `col0` from join/limit will resolve their indexes to the first col0. But actually, the first col0 should resolve to the first col0 of its child(ren). The second should resolve to the second of its child(ren).

This pr aims to solve this issue.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #51203, opened 2024-02-20
planner,executor: fix join resolveIndex won't find its column from children schema & amend join's lused and rused logic for reversed column ref from join schema to its children

AilinKid:

Issue Number: close https://github.com/pingcap/tidb/issues/42588

Problem Summary:

issue https://github.com/pingcap/tidb/issues/42588 is caused by many factors and this pr is try to resolve them all.
* first: resolve parent pointer modification error caused by the apply elimination, as mentioned here https://github.com/pingcap/tidb/issues/42588#issuecomment-1929026146
* second: since TiDB internal has many non-rigorous logical optimizing rules, occasionally two or more of them will leading a critical assumption which will lead a resolveIndex error as mentioned here: https://github.com/pingcap/tidb/pull/45831#discussion_r1481031471
* third: in the execution of join, the inline projection's `lused` and `rused` are order unawareness, which means even if we  allow reversed column ref in join schema itself to its children, it will break its assumption here too. So we need to fix here as well.
example: 
```
join schema[col11, col22, col21], child1 [col1], child2 [col21, col22]
the pull/45831 make an assumption here that join schema won't refer a column-order revered from the order of its children. Like here [col22, col21] in join schema is reversed with  [col21, col22] from its child2. Step2 try to fix this.

in the execution layer.
join schema[col11, col22, col21], child1 [col1], child2 [col21, col22]
the lused will be [true]
the rused will be [true, true]
when constructing the join row result, the order couldn't be restored from a simple bool array. Consequently one join result will be shown like: <col1, col21, col22> while the field type is totally different from what the datum in the same position. So step3 try to fix this.
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/benchmark_test.go
+++ b/pkg/executor/benchmark_test.go
@@ -15,9 +15,11 @@
 package executor
 
 import (
+	"cmp"
 	"context"
 	"fmt"
 	"os"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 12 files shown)


pkg/executor/benchmark_test.go
```diff
@@ -1350,6 +1372,20 @@ func prepareMergeJoinExec(tc *mergeJoinTestCase, joinSchema *expression.Schema,
...
+		for _, childSchema := range tc.childrenUsedSchema {
+			used := make([]int, 0, len(childSchema))
+			for idx, one := range childSchema {
+				if one {
+					used = append(used, idx)
+				}
+			}
+			usedIdx = append(usedIdx, used)
+		}
+	}
```
winoros, line 1387: Why do we need a special one for the merge join case?

pkg/executor/benchmark_test.go
```diff
@@ -1595,6 +1631,7 @@ func benchmarkMergeJoinExecWithCase(b *testing.B, tc *mergeJoinTestCase, innerDS
 		if err := executor.Close(); err != nil {
 			b.Fatal(err)
 		}
+		fmt.Println("current round: ", i, "/", b.N)
```
winoros, line 1634: debugging log.

pkg/executor/benchmark_test.go
AilinKid: because merge join's usedArray(bool type) is passed in the test case itself, we need an adaptor here to change it as int array type to feed the below joiner initialization

winoros (approved):

PR #51058, opened 2024-02-07
infoschema,domain: introduce InfoSchemaV2, core data struct etc

tiancaiamao:
### What problem does this PR solve?

Issue Number: ref #50959 

Problem Summary:

### What changed and how does it work?

This commit introduce the `infoschema_v2.go` file, include the core data struct, implement some of the APIs.

It's not a completed work, the code can barely run some basic operations.
There is still a long way to make the code smooth, but I fire this PR early and it is ready for review now. Why? 
Because I hope we can provide infoschema v2 in tidb 8.0, GA or at least RC, time is limited.
This commit provide a solid foundation that **we can work on together**. 

With the core data struct introduced, we can fullfill the missing things and split the tasks parallelly. 
Later, I, @GMHDBJD and other team members will continue the development.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [X] Manual test (add detailed scripts or steps below)

Enable infoschema v2 here:

https://github.com/pingcap/tidb/compare/master...tiancaiamao:infoschema-v2-pr?expand=1#diff-d82e5bb6499a8d8c00f3d55ec95bccda3fab6d4cc45529cecf06552bfa973b1bR21

```
func init() {
    enableV2.Store(true)
}
```

Start tidb, create some tabe, insert data, drop table, verify the basic operations etc...

This commit doesn't break anyting because the new code is not used now.
We need prepare a test plan document and add tests later.

- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

Issue #50959, opened 2024-02-04
InfoSchema V2


tiancaiamao:
## Enhancement

When we have millions of tables, the memory footprint for meta data is not negligible.
Some users use TiDB to provide services to their tenant, so there could be a huge amount of tables.
We need to develop a kind of caching mechanism to ensure that even in the extreme cases, the meta data do not cause TiDB OOM.

In the new design, the table.Table will not be persisted in memory, insted, they will be cached.

## Development tasks

- [ ] add design document
- [X] internal test plan document [(link)](https://docs.google.com/document/d/1GSFFmJhsye6jwpJAHIGHfV3kBL-ULlOIs42HrukhV_w/edit?usp=sharing)
- [x] [introduce the core data struct for InfoSchemaV2](https://github.com/pingcap/tidb/pull/51058)
- [x] #51256
- [x] #51680
- [x] #51679
- [x] #52150
- [x] #52237
- [x] #52234
- [x] https://github.com/pingcap/tidb/issues/54281
- [ ] observability: about the jitter when cache miss
    - [ ] display refill cache latency to slow query log
    - [x] display refill cache latency to trace statement
    - [ ] display refill cache latency to explain analyze etc

PR #51257, opened 2024-02-22
*: support building infoschema V2, and introduce @@tidb_schema_cache_size

### LINKED ISSUE #51256: introduce @@tidb_infoschema_cache_size switch to control the behaviour

tiancaiamao:

Issue Number: close #51256 
ref #50959

~~!!! *Need to merge https://github.com/pingcap/tidb/pull/51058 first before reviewing this PR.*~~ (DONE)

Problem Summary:

1. The infoschema.Builder now support building both infoschema v1 and v2

Before this change, when infoschema Builder builds the infoschema v2, it depends on building the infoschema v1 object
and rely on an `infoschemaProxy` object to pretender to be an InfoSchema instance.
Now it support build and use infoschemaV2 object directly.

2. Introduce the `@@tidb_schema_cache_size` global session variable

The variable work as a switch for us to enable v2 if it's value is not equal to 0.
We can now change to infoschema v2 and change back to v1, it make the testing easier.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [X] Contains variable changes

**Caution: if we expose this global variable, but infoschema v2 implementation has not finish yet, modify the default value would be buggy.** There are two options:
- 1. hide it totally from the document, so nobody would know and modifty it. 
- 2. add this change to the document, but mark as expiremental and notify the consequence explicitly.

I would prefer 1 because it's still at an early stage for infoschema v2.

### Release note

```release-note
Introduce `@@tidb_schema_cache_size` global session variable, infoschema V2 implementation would be used when its value is not zero
```


pkg/sessionctx/variable/tidb_vars.go
```diff
@@ -946,6 +946,9 @@ const (
 
 	// TiDBTxnEntrySizeLimit indicates the max size of a entry in membuf.
 	TiDBTxnEntrySizeLimit = "tidb_txn_entry_size_limit"
+
+	// TiDBInfoSchemaCacheSize indicates the size of infoschema meta data which are cached in V2 implementation.
+	TiDBInfoSchemaCacheSize = "tidb_infoschema_cache_size"
```
ywqzzy, line 951: When modify the variable, should we force trigger reload infoschema?

tiancaiamao: Yes, we can, but it seems we need to register callback function in `SetGlobal` to call the function in domain.
How to do that is a bit complex, considering infoschema reload is called eventually anyway, I skip the force trigger part.

ywqzzy (approved):

pkg/domain/domain.go
```diff
@@ -448,7 +453,10 @@ func (do *Domain) tryLoadSchemaDiffs(m *meta.Meta, usedVersion, newVersion int64
 		}
 		diffs = append(diffs, diff)
 	}
-	builder := infoschema.NewBuilder(do, do.sysFacHack, do.infoCache.Data).InitWithOldInfoSchema(do.infoCache.GetLatest())
+	builder, err := infoschema.NewBuilder(do, do.sysFacHack, do.infoCache.Data).InitWithOldInfoSchema(do.infoCache.GetLatest())
```
ywqzzy, line 456: In line 307, we may choose different ways of init builder.
Ok for now

pkg/infoschema/builder.go
```diff
@@ -177,7 +178,8 @@ func (b *bundleInfoBuilder) updateTableBundles(is *infoSchema, tableID int64) {
 
 // Builder builds a new InfoSchema.
 type Builder struct {
-	is *infoSchema
+	enableV2 bool
```
ywqzzy, line 181: IMHO, we can have a brand new builder for V2. Since many schema update work messed up with v1 and v2 now. Some of the structures should not updated for v2 may be updated by accident during developing.

Ok for now.

tiancaiamao: Yes, I agree with the brand new builder for V2, but it involves much more code changes.
At the current phase we can keep the minimal code changes and go ahead.

GMHDBJD (approved):
LGTM

Issue #51036, opened 2024-02-07
Merge lightning preprocess duplicate detection and post-import conflict detection 


lyzx2001:
## Enhancement
Merge lightning preprocess duplicate detection and post-import conflict detection 

```
[conflict]
strategy = "none"
```

1. none: Do not conduct conflict detection. If the imported data has conflict, checksum phase will fail and return error.
2. replace: Remove all the rows that have conflicts except the latest ones.
3. error: Fail and return error if conflicts are detected during import phase.

We remove the old configuration options, including record and remove, which are less useful and confusing for users.
tikv-importer.duplicate-resolutionwill not be used after preprocess duplicate detection and post-import conflict detection are merged.
Add a parameter `conflict.precheck-conflict-before-import` to control whether to turn on preprocess duplicate detection. The valid value is `true` and `false`, and the default is `false`.

Users should be able to check table `lightning_task_info.conflict_error_v2` for all the conflicts, including conflicts from both preprocess duplicate detection and post-import conflict detection. Previously, conflicts from preprocess duplicate detection were stored in table `lightning_task_info.conflict_records`, so here we need to integrate them into `lightning_task_info.conflict_error_v2`.

Pre-work: Support `replace` mode for lightning post-import conflict detection https://github.com/pingcap/tidb/issues/45774

lyzx2001:
Further optimization will be tracked by https://github.com/pingcap/tidb/issues/52306

PR #51277, opened 2024-02-23
lightning: improve post-import conflict detection 'error' semantic

lyzx2001:

Issue Number: ref #51036 

Problem Summary:
Merge preprocess duplicate detection and post-import conflict detection.
Improve post-import conflict detection 'error' semantic.
### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -459,7 +467,7 @@ func (m *DupeDetector) HasDuplicate() bool {
 }
 
 // RecordDataConflictError records data conflicts to errorMgr. The key received from stream must be a row key.
-func (m *DupeDetector) RecordDataConflictError(ctx context.Context, stream DupKVStream) error {
+func (m *DupeDetector) RecordDataConflictError(ctx context.Context, stream DupKVStream, tbl table.Table, algorithm config.DuplicateResolutionAlgorithm) error {
```
lance6716, line 470: seems we can use m.tbl, no need to add a tbl parameter

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -481,6 +489,12 @@ func (m *DupeDetector) RecordDataConflictError(ctx context.Context, stream DupKV
 		if err != nil {
 			return errors.Trace(err)
 		}
+
+		if algorithm == config.DupeResAlgErr {
+			err = common.ErrFoundDuplicateKeys.FastGenByArgs(key, val)
+			return ConvertToErrFoundConflictRecords(err, tbl)
```
lance6716, line 495: here we can directly use key and val. No need to create a temporary err variable. You can add another function like NewErrFoundConflictRecords(key, value, tbl) and let ConvertToErrFoundConflictRecords use it internally.

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -550,6 +564,12 @@ func (m *DupeDetector) RecordIndexConflictError(ctx context.Context, stream DupK
 		if err != nil {
 			return errors.Trace(err)
 		}
+
+		if algorithm == config.DupeResAlgErr {
+			err = common.ErrFoundDuplicateKeys.FastGenByArgs(key, val)
```
lance6716, line 569: ditto

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -573,6 +593,111 @@ func (m *DupeDetector) RecordIndexConflictError(ctx context.Context, stream DupK
 	return nil
 }
 
+// RetrieveKeyAndValueFromErrFoundDuplicateKeys retrieves the key and value
+// from ErrFoundDuplicateKeys error.
+func RetrieveKeyAndValueFromErrFoundDuplicateKeys(err error) ([]byte, []byte, error) {
+	errString := err.Error()
```
lance6716, line 599: instead of using string matching which is less efficient and error-prone, we can use a structure to save the fields. Like https://github.com/pingcap/tidb/blob/332086ebc863b0effb459c1c0eb5af6defced047/pkg/parser/mysql/error.go#L28-L38

pkg/ddl/ingest/backend.go
```diff
@@ -99,7 +99,7 @@ func (bc *litBackendCtx) CollectRemoteDuplicateRows(indexID int64, tbl table.Tab
 		SQLMode: mysql.ModeStrictAllTables,
 		SysVars: bc.sysVars,
 		IndexID: indexID,
-	})
+	}, lightning.DupeResAlgNone)
```
lance6716, line 102: use "error"?

lyzx2001: modified

pkg/ddl/ingest/backend.go
```diff
@@ -140,7 +140,7 @@ func (bc *litBackendCtx) FinishImport(indexID int64, unique bool, tbl table.Tabl
 			SQLMode: mysql.ModeStrictAllTables,
 			SysVars: bc.sysVars,
 			IndexID: ei.indexID,
-		})
```
lance6716, line 143: ditto

br/pkg/lightning/importer/import.go
```diff
@@ -2026,6 +2027,11 @@ func (rc *Controller) enforceDiskQuota(ctx context.Context) {
 			for _, engine := range largeEngines {
 				// Use a larger split region size to avoid split the same region by many times.
 				if err := localBackend.UnsafeImportAndReset(ctx, engine, int64(config.SplitRegionSize)*int64(config.MaxSplitRegionSizeRatio), int64(config.SplitRegionKeys)*int64(config.MaxSplitRegionSizeRatio)); err != nil {
+					engineIns := localBackend.EngineMgr.RLockEngine(engine)
```
lance6716, line 2030: Sorry I just noticed that this error will not be reported to user, so we don't need to convert it. Please revert https://github.com/pingcap/tidb/pull/51277/commits/3fda42b6a98d6d475402b0b468294ecf59b9c118 except for the DupeResAlgErr change in pkg/ddl/ingest/backend.go

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -481,6 +488,11 @@ func (m *DupeDetector) RecordDataConflictError(ctx context.Context, stream DupKV
 		if err != nil {
 			return errors.Trace(err)
 		}
+
+		if algorithm == config.DupeResAlgErr {
+			return NewErrFoundConflictRecords(ctx, key, val, m.tbl)
```
lance6716, line 493: need to check with PM if we still need to save a record in downstream table.

And seems we can use `h.String()` and `m.decoder.DecodeRawRowDataAsStr(h, val)`? See line 499~500

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -573,6 +591,117 @@ func (m *DupeDetector) RecordIndexConflictError(ctx context.Context, stream DupK
...
+		if err != nil {
+			str = string(val)
+		}
+		if types.IsBinaryStr(colInfos[i].Ft) || types.IsTypeBit(colInfos[i].Ft) {
+			str = tidbutil.FmtNonASCIIPrintableCharToHex(str)
+		}
+		valueStr = append(valueStr, str)
+	}
+
+	return errors.Trace(common.ErrFoundIndexConflictRecords.FastGenByArgs(tbl.Meta().Name, idxValues[0], valueStr[0]))
```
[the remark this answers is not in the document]
lyzx2001: I think here we only need to show any one value of the conflict key, to conform to the returned error of data KV, which also only shows one value of the conflict key. It may not be easy to change the code for data KV to show all the values of the conflict key.

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -550,6 +562,11 @@ func (m *DupeDetector) RecordIndexConflictError(ctx context.Context, stream DupK
 		if err != nil {
 			return errors.Trace(err)
 		}
+
+		if algorithm == config.DupeResAlgErr {
```
lance6716, line 566: we also have variable `indexInfo`, no need to find the index info again inside `NewErrFoundConflictRecords`

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -573,6 +590,97 @@ func (m *DupeDetector) RecordIndexConflictError(ctx context.Context, stream DupK
 	return nil
 }
 
+// RetrieveKeyAndValueFromErrFoundDuplicateKeys retrieves the key and value
+// from ErrFoundDuplicateKeys error.
+func RetrieveKeyAndValueFromErrFoundDuplicateKeys(err error) ([]byte, []byte, error) {
+	tErr, ok := errors.Cause(err).(*terror.Error)
+	if !ok {
+		return nil, nil, err
+	}
```
lance6716, line 599: check tErr.ID is expected, which is a ErrFoundDuplicateKeys

br/pkg/lightning/backend/local/duplicate.go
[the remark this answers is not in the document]
lance6716: You can check what's the meaning of idxValues[0] or valueStr[0]. Maybe they are from a multiple column index, so all columns are needed to identify the row.

br/pkg/lightning/backend/local/duplicate.go
lyzx2001: In `RecordIndexConflictError` we need to use `NewErrFoundConflictRecords` to decode the value since there is no value for `Row`
https://github.com/pingcap/tidb/blob/99390e2a0e3912df57e8ea2ed05cb24efb114d0e/br/pkg/lightning/backend/local/duplicate.go#L553-L557
So here I also use `NewErrFoundConflictRecords` to make them consistent

lyzx2001: I suppose for the ease of code, I can change to use `h.String()` and `m.decoder.DecodeRawRowDataAsStr(h, val)`

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -494,6 +502,10 @@ func (m *DupeDetector) RecordDataConflictError(ctx context.Context, stream DupKV
 			}
 			dataConflictInfos = dataConflictInfos[:0]
 		}
+
+		if algorithm == config.DupeResAlgErr {
+			return errors.Trace(common.ErrFoundDataConflictRecords.FastGenByArgs(m.tbl.Meta().Name, h.String(), m.decoder.DecodeRawRowDataAsStr(h, val)))
```
lance6716, line 507: This is the usage of DDL

https://github.com/pingcap/tidb/blob/b2c3e67fdca6bf641d92ed63b41d8aee4ccf374d/pkg/ddl/ingest/backend.go#L139-L152

Seems when it wants to fail fast for duplicate error, CollectRemoteDuplicateRows should not return error. Instead, it should return a `true` `hasDupe`, and ddl logic will then convert it to tikv.ErrKeyExists so TestAddIndexRemoteDuplicateCheck will pass.

This fail fast behaviour is different with import's fail fast behaviour, where it expects there's a non-nil error.

Do you think it's better that we change DupeController.CollectRemoteDuplicateRows to return 3 variables, `hasDupe bool, errorFromDupeResAlgErr error, runningErr error`?

lance6716: /cc @tangenta 

And in old ddl logic, DDL will not report the duplicate entry in error message, can you guide us how to return a correct format to build the error message?

br/pkg/lightning/backend/local/duplicate_test.go
```diff
@@ -70,3 +73,182 @@ func TestBuildDupTask(t *testing.T) {
...
+	data2RowValue := kvPairs.Pairs[2].Val
+	data3IndexKey := kvPairs.Pairs[5].Key
+	data3IndexValue := kvPairs.Pairs[5].Val
+
+	ctx := context.Background()
+
+	originalErr := common.ErrFoundDuplicateKeys.FastGenByArgs(data2RowKey, data2RowValue)
+
+	newErr := local.ConvertToErrFoundConflictRecords(ctx, originalErr, tbl)
+	require.EqualError(t, newErr, "[Lightning:Restore:ErrFoundDataConflictRecords]found data conflict records in table a, primary key is '2', row data is '[KindInt64 2 KindInt64 6 KindString 2.csv KindInt64 102]'")
```
lance6716, line 248: `[KindInt64 2 KindInt64 6 KindString 2.csv KindInt64 102]` is not very good in reading, you can use `decoder.DecodeRawRowDataAsStr` to get a more friendly string

br/pkg/lightning/backend/local/duplicate.go
lyzx2001: DDL behavior has been maintained to conform to the original behavior.

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -573,6 +590,115 @@ func (m *DupeDetector) RecordIndexConflictError(ctx context.Context, stream DupK
...
+
+// NewErrFoundConflictRecords generate an error ErrFoundDuplicateKeys
+// according to key and value.
+func NewErrFoundConflictRecords(ctx context.Context, key []byte, value []byte, tbl table.Table, idxInfo *model.IndexInfo) error {
+	logger := log.FromContext(ctx).With(zap.String("table", tbl.Meta().Name.L)).Begin(zap.InfoLevel, "[new-ErrFoundConflictRecords] new ErrFoundConflictRecords")
+	sessionOpts := encode.SessionOptions{
+		SQLMode: mysql.ModeStrictAllTables,
+	}
+
+	logger2 := logger.With(zap.String("tableName", tbl.Meta().Name.L))
```
lance6716, line 622: tableName is the same field of `logger`'s `table`

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -573,6 +590,115 @@ func (m *DupeDetector) RecordIndexConflictError(ctx context.Context, stream DupK
...
+	if !keyIsByte || !valIsByte {
+		return nil, nil, err
+	}
+	return key, value, nil
+}
+
+// NewErrFoundConflictRecords generate an error ErrFoundDuplicateKeys
+// according to key and value.
+func NewErrFoundConflictRecords(ctx context.Context, key []byte, value []byte, tbl table.Table, idxInfo *model.IndexInfo) error {
+	logger := log.FromContext(ctx).With(zap.String("table", tbl.Meta().Name.L)).Begin(zap.InfoLevel, "[new-ErrFoundConflictRecords] new ErrFoundConflictRecords")
```
lance6716, line 617: This is a simple function, we can directly log needed information, no need to use `.Begin` to create a Task struct.

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -564,6 +577,10 @@ func (m *DupeDetector) RecordIndexConflictError(ctx context.Context, stream DupK
 			}
 			indexHandles.truncate()
 		}
+
+		if algorithm == config.DupeResAlgErr {
+			return NewErrFoundConflictRecords(ctx, key, val, m.tbl, indexInfo)
```
lance6716, line 582: We can extract the part of NewErrFoundConflictRecords into ErrFoundIndexConflictRecords. 
NewErrFoundConflictRecords will check if the key is a record key and construct a ErrFoundDataConflictRecords, otherwise find the indexInfo from tableInfo and call ErrFoundIndexConflictRecords.
And here we can directly call ErrFoundIndexConflictRecords.

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -573,6 +590,115 @@ func (m *DupeDetector) RecordIndexConflictError(ctx context.Context, stream DupK
...
+		idxInfo = model.FindIndexInfoByID(tbl.Meta().Indices, idxID)
+	}
+
+	h, err := decoder.DecodeHandleFromIndex(idxInfo, key, value)
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	rowKey := tablecodec.EncodeRowKeyWithHandle(tbl.Meta().ID, h)
+	handle, err := tablecodec.DecodeRowKey(rowKey)
```
lance6716, line 656: `handle` is the same as `h`? Revise the basic concepts of TiDB's key encoding rule and be clear about what's the purpose of the code.

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -573,6 +590,115 @@ func (m *DupeDetector) RecordIndexConflictError(ctx context.Context, stream DupK
...
+	colInfos := tables.BuildRowcodecColInfoForIndexColumns(idxInfo, tbl.Meta())
+	values, err := tablecodec.DecodeIndexKV(key, value, idxColLen, tablecodec.HandleNotNeeded, colInfos)
+	if err != nil {
+		logger.Warn("decode index key value failed", zap.String("index", indexName),
+			zap.String("key", hex.EncodeToString(key)), zap.String("value", hex.EncodeToString(value)), zap.Error(err))
+		return errors.Trace(common.ErrFoundIndexConflictRecords.FastGenByArgs(tbl.Meta().Name, key, value))
+	}
+	valueStr := make([]string, 0, idxColLen)
+	for i, val := range values[:idxColLen] {
+		d, err := tablecodec.DecodeColumnValue(val, colInfos[i].Ft, time.Local)
```
lance6716, line 672: I see these codes are the same as DDL's `genKeyExistsErr`, maybe ask @tangenta extract a function to avoid duplication. Also when DDL convert ErrFoundIndexConflictRecords to ErrKeyExists it can directly use `valueStr`

br/pkg/lightning/backend/local/duplicate.go
lyzx2001: Yes, `handle` is the same as `h`, they are both the handle decoded from the index.

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -573,6 +590,122 @@ func (m *DupeDetector) RecordIndexConflictError(ctx context.Context, stream DupK
...
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	_, idxID, _, err := tablecodec.DecodeIndexKey(key)
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	if idxInfo == nil {
```
lance6716, line 659: can we restrict that idxInfo must not be nil? I see a possible nil path from ConvertToErrFoundConflictRecords -> NewErrFoundConflictRecords -> NewErrFoundIndexConflictRecords. We can let NewErrFoundConflictRecords rather than NewErrFoundIndexConflictRecords find the idxInfo

pkg/ddl/ingest/backend.go
```diff
@@ -99,8 +100,8 @@ func (bc *litBackendCtx) CollectRemoteDuplicateRows(indexID int64, tbl table.Tab
 		SQLMode: mysql.ModeStrictAllTables,
 		SysVars: bc.sysVars,
 		IndexID: indexID,
-	})
-	if err != nil {
+	}, lightning.DupeResAlgErr)
+	if err != nil && !common.ErrFoundDataConflictRecords.Equal(err) && !common.ErrFoundIndexConflictRecords.Equal(err) {
```
lance6716, line 104: for DDL, only ErrFoundIndexConflictRecords is possible

pkg/ddl/ingest/backend.go
```diff
@@ -140,8 +141,8 @@ func (bc *litBackendCtx) FinishImport(indexID int64, unique bool, tbl table.Tabl
 			SQLMode: mysql.ModeStrictAllTables,
 			SysVars: bc.sysVars,
 			IndexID: ei.indexID,
-		})
-		if err != nil {
+		}, lightning.DupeResAlgErr)
+		if err != nil && !common.ErrFoundDataConflictRecords.Equal(err) && !common.ErrFoundIndexConflictRecords.Equal(err) {
```
lance6716, line 145: ditto

lance6716 (commented):
rest lgtm

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -573,6 +590,127 @@ func (m *DupeDetector) RecordIndexConflictError(ctx context.Context, stream DupK
...
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	valueStr, err := GenIndexValueFromIndex(key, value, tbl.Meta(), idxInfo)
+	indexName := fmt.Sprintf("%s.%s", tbl.Meta().Name.String(), idxInfo.Name.String())
+	if err != nil {
+		log.L().Warn("decode index key value / column value failed", zap.String("index", indexName),
+			zap.String("key", hex.EncodeToString(key)), zap.String("value", hex.EncodeToString(value)), zap.Error(err))
+		return errors.Trace(common.ErrFoundIndexConflictRecords.FastGenByArgs(tbl.Meta().Name, indexName, key, value))
```
lance6716, line 697: in this case, `key` and `value` does noy satisfy the meaning of message template of ErrFoundIndexConflictRecords. Should we use a new error?

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -984,8 +1124,12 @@ func (local *DupeController) CollectRemoteDuplicateRows(ctx context.Context, tbl
 	if err != nil {
 		return false, errors.Trace(err)
 	}
-	if err := duplicateManager.CollectDuplicateRowsFromTiKV(ctx, local.importClientFactory); err != nil {
-		return false, errors.Trace(err)
+	err = duplicateManager.CollectDuplicateRowsFromTiKV(ctx, local.importClientFactory, algorithm)
+	if err != nil {
+		if !common.ErrFoundDataConflictRecords.Equal(err) && !common.ErrFoundIndexConflictRecords.Equal(err) {
```
lance6716, line 1129: `return common.ErrFoundDataConflictRecords.Equal(err) || common.ErrFoundIndexConflictRecords.Equal(err), errors.Trace(err)`

pkg/ddl/index.go
```diff
@@ -1594,31 +1595,12 @@ func (w *addIndexTxnWorker) checkHandleExists(idxInfo *model.IndexInfo, key kv.K
 }
 
 func genKeyExistsErr(key, value []byte, idxInfo *model.IndexInfo, tblInfo *model.TableInfo) error {
-	idxColLen := len(idxInfo.Columns)
+	valueStr, err := local.GenIndexValueFromIndex(key, value, tblInfo, idxInfo)
```
lance6716, line 1598: Maybe we can move `GenIndexValueFromIndex` into another package? It is not the responsibility of `local` package to build the message of duplicate entry. (In fact the old logic is copied from `ddl` package)

pkg/ddl/ingest/backend.go
```diff
@@ -99,15 +101,29 @@ func (bc *litBackendCtx) CollectRemoteDuplicateRows(indexID int64, tbl table.Tab
...
 		return err
 	} else if hasDupe {
 		logutil.Logger(bc.ctx).Error(LitErrRemoteDupExistErr,
 			zap.String("table", tbl.Meta().Name.O), zap.Int64("index ID", indexID))
-		return tikv.ErrKeyExists
+
+		if common.ErrFoundIndexConflictRecords.Equal(err) {
+			tErr, ok := errors.Cause(err).(*terror.Error)
+			if !ok {
+				return err
```
lance6716, line 116: here we should return tikv.ErrKeyExists. Also for next lines.

pkg/ddl/ingest/backend.go
```diff
@@ -140,15 +156,29 @@ func (bc *litBackendCtx) FinishImport(indexID int64, unique bool, tbl table.Tabl
...
 			return err
 		} else if hasDupe {
 			logutil.Logger(bc.ctx).Error(LitErrRemoteDupExistErr,
 				zap.String("table", tbl.Meta().Name.O), zap.Int64("index ID", indexID))
-			return tikv.ErrKeyExists
+
+			if common.ErrFoundIndexConflictRecords.Equal(err) {
+				tErr, ok := errors.Cause(err).(*terror.Error)
+				if !ok {
+					return err
```
lance6716, line 171: ditto

br/pkg/lightning/backend/local/duplicate.go
lyzx2001: Here I conform to the original `genKeyExistsErr` function, where it also returns the expected `ErrKeyExists` error with the raw key
https://github.com/pingcap/tidb/blob/897d93db8c9b0959803cc7ebae419bf72da0ded3/pkg/ddl/index.go#L1604
I think here we can also use the raw key and value, they satisfy the meaning of ErrFoundIndexConflictRecords, although they are in the form of raw but not human readable.

Leavrth (approved):

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -573,6 +587,100 @@ func (m *DupeDetector) RecordIndexConflictError(ctx context.Context, stream DupK
...
+	value, valIsByte := tErr.Args()[1].([]byte)
+	if !keyIsByte || !valIsByte {
+		return nil, nil, err
+	}
+	return key, value, nil
+}
+
+// NewErrFoundConflictRecords generate an error ErrFoundDataConflictRecords / ErrFoundIndexConflictRecords
+// according to key and value.
+func NewErrFoundConflictRecords(key []byte, value []byte, tbl table.Table) error {
```
D3Hunter, line 613: why not create the error in where `ErrFoundDuplicateKeys` is created, so no need to convert them back and forth

br/tests/lightning_duplicate_resolution_error_uk_multiple_files_multicol_index/run.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+#
+# Copyright 2021 PingCAP, Inc.
```
D3Hunter, line 3:
```suggestion
# Copyright 2024 PingCAP, Inc.
```

D3Hunter (commented):
21/27

br/pkg/lightning/backend/local/duplicate.go
lyzx2001: Because `ErrFoundDuplicateKeys` is created here
https://github.com/pingcap/tidb/blob/69ab2a4de69f458db967026c7da5da0bbfba1c79/br/pkg/lightning/common/dupdetect.go#L96-L102
The error will be returned to lightning/ddl, and we need to maintain the behavior of ddl, which will convert `ErrFoundDuplicateKeys` to `ErrKeyExists`. But in lightning, we will need to convert `ErrFoundDuplicateKeys` to `ErrFoundDataConflictRecords` or `ErrFoundIndexConflictRecords` for better readability according to PM, so we do separate conversion in lightning.
So actually there are no converting back and forth, there are just two paths for converting depending on which package the error is returned to.

br/tests/lightning_duplicate_resolution_error_uk_multiple_files_multicol_index/run.sh
lyzx2001: modified

pkg/table/tables/tables.go
```diff
@@ -2386,3 +2386,30 @@ func (t *TemporaryTable) SetSize(v int64) {
 func (t *TemporaryTable) GetMeta() *model.TableInfo {
 	return t.meta
 }
+
+// GenIndexValueFromIndex generate index value from index.
+func GenIndexValueFromIndex(key []byte, value []byte, tblInfo *model.TableInfo, idxInfo *model.IndexInfo) ([]string, error) {
```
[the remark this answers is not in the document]
lyzx2001: UT added

MODULE.bazel.lock
Benjamin2037, line 1: whether this doc needs to be included in this PR?

pkg/table/tables/index_test.go
```diff
@@ -168,3 +171,49 @@ func buildTableInfo(t *testing.T, sql string) *model.TableInfo {
...
 }
+
+func TestGenIndexValueFromIndex(t *testing.T) {
+	p := parser.New()
+	node, _, err := p.ParseSQL("create table a (a int primary key, b int not null, c text, unique key key_b(b));")
+	require.NoError(t, err)
+	mockSctx := mock.NewContext()
+	info, err := ddl.MockTableInfo(mockSctx, node[0].(*ast.CreateTableStmt), 108)
+	require.NoError(t, err)
+	info.State = model.StatePublic
```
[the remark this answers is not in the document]
lyzx2001: modified

br/pkg/lightning/backend/local/duplicate.go
D3Hunter: can we return `ErrFoundDataConflictRecords` and `ErrFoundIndexConflictRecords ` directly, and change ddl part to convert `ErrFoundIndexConflictRecords` to `ErrKeyExists`

pkg/table/tables/index_test.go
```diff
@@ -168,3 +171,44 @@ func buildTableInfo(t *testing.T, sql string) *model.TableInfo {
...
+	indexValue := kvPairs.Pairs[1].Val
+
+	_, idxID, _, err := tablecodec.DecodeIndexKey(indexKey)
+	require.NoError(t, err)
+
+	idxInfo := model.FindIndexInfoByID(tbl.Meta().Indices, idxID)
+
+	valueStr, err := tables.GenIndexValueFromIndex(indexKey, indexValue, tbl.Meta(), idxInfo)
+	require.NoError(t, err)
+	require.Equal(t, []string{"23"}, valueStr)
```
D3Hunter, line 213: what's the output for binary or bits?

D3Hunter (commented):
25/30

br/pkg/lightning/backend/local/duplicate.go
lyzx2001: The original `ErrFoundDuplicateKeys` error contains the most basic information of the conflict. When creating `ErrFoundDataConflictRecords` or `ErrFoundIndexConflictRecords`, we need to conduct extra decoding and encoding for information like row key, row value, etc. 
`ErrKeyExists` does not need those extra information, so it seems unnecessary to convert `ErrKeyExists` from `ErrFoundIndexConflictRecords`.

pkg/table/tables/index_test.go
lyzx2001: `tables.GenIndexValueFromIndex`'s content is retrieved from the original function `genKeyExistsErr` in `ddl`
https://github.com/pingcap/tidb/blob/5745d3df9de4afa98909eb972c8456f8135d65f5/pkg/ddl/index.go#L1601-L1629

br/pkg/lightning/backend/local/duplicate_test.go
```diff
@@ -70,3 +73,178 @@ func TestBuildDupTask(t *testing.T) {
...
+	newErr := local.ConvertToErrFoundConflictRecords(originalErr, tbl)
+	require.EqualError(t, newErr, "[Lightning:Restore:ErrFoundDataConflictRecords]found data conflict records in table a, primary key is '2', row data is '(2, 6, \"2.csv\")'")
+
+	originalErr = common.ErrFoundDuplicateKeys.FastGenByArgs(data3IndexKey, data3IndexValue)
+
+	newErr = local.ConvertToErrFoundConflictRecords(originalErr, tbl)
+	require.EqualError(t, newErr, "[Lightning:Restore:ErrFoundIndexConflictRecords]found index conflict records in table a, index name is 'a.key_b', unique key is '[7]', primary key is '3'")
+}
+
+func TestConvertToErrFoundConflictRecordsMultipleColumnsIndex(t *testing.T) {
```
zimulala, line 185: `TestConvertToErrFoundConflictRecords` and `TestConvertToErrFoundConflictRecordsMultipleColumnsIndex` only difference seems to be the number of columns in the table and the index. Could we unify it?

br/pkg/lightning/backend/local/duplicate.go
lance6716: The error message template of `ErrKeyExists` is "Duplicate entry '%-.64s' for key '%-.192s'". We have a different template. Arguments order and meaning is wrong, not whether it's raw or not.

br/pkg/lightning/backend/local/duplicate_test.go
lyzx2001: The three unit tests have been unified by extracting `buildTableForTestConvertToErrFoundConflictRecords` function.

D3Hunter (commented):
rest lgtm

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -984,8 +1093,9 @@ func (local *DupeController) CollectRemoteDuplicateRows(ctx context.Context, tbl
 	if err != nil {
 		return false, errors.Trace(err)
 	}
-	if err := duplicateManager.CollectDuplicateRowsFromTiKV(ctx, local.importClientFactory); err != nil {
-		return false, errors.Trace(err)
+	err = duplicateManager.CollectDuplicateRowsFromTiKV(ctx, local.importClientFactory, algorithm)
+	if err != nil {
+		return common.ErrFoundDataConflictRecords.Equal(err) || common.ErrFoundIndexConflictRecords.Equal(err), errors.Trace(err)
```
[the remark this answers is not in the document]
D3Hunter: see below comment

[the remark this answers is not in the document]
lyzx2001: I don' think `hasDup` is useless. If error happened, it may be other error like decoding failed or other failure, not necessarily means we encounter duplicate. We still need `common.ErrFoundDataConflictRecords.Equal(err) || common.ErrFoundIndexConflictRecords.Equal(err)` to determine whether we have duplicates.

`CollectDuplicateRowsFromTiKV` will not return `ErrFoundDuplicateKeys`. Here it can only return `ErrFoundDataConflictRecords` or `ErrFoundIndexConflictRecords`.

pkg/ddl/ingest/backend.go
```diff
@@ -90,6 +92,33 @@ type litBackendCtx struct {
 	etcdClient      *clientv3.Client
 }
 
+func (bc *litBackendCtx) handleErrorAfterCollectRemoteDuplicateRows(err error, indexID int64, tbl table.Table, hasDupe bool) error {
+	if err != nil && !common.ErrFoundIndexConflictRecords.Equal(err) {
```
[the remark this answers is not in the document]
lyzx2001: I think here the error is correctly handled. We first determine
`if err != nil && !common.ErrFoundIndexConflictRecords.Equal(err) `.
In this case, the error means something that we cannot handle any further.

Then we use `hasDupe`, in this case either we have no error, or the error is `ErrFoundIndexConflictRecords`. In both cases, we can directly use `hasDupe` for further operation.

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -494,6 +499,10 @@ func (m *DupeDetector) RecordDataConflictError(ctx context.Context, stream DupKV
 			}
 			dataConflictInfos = dataConflictInfos[:0]
 		}
+
+		if algorithm == config.DupeResAlgErr {
+			return errors.Trace(common.ErrFoundDataConflictRecords.FastGenByArgs(m.tbl.Meta().Name, h.String(), m.decoder.DecodeRawRowDataAsStr(h, val)))
```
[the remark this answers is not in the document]
lyzx2001: PM said it'd be better to also show the row data in the error as well

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -573,6 +587,99 @@ func (m *DupeDetector) RecordIndexConflictError(ctx context.Context, stream DupK
...
+		return nil, nil, err
+	}
+	return key, value, nil
+}
+
+// newErrFoundConflictRecords generate an error ErrFoundDataConflictRecords / ErrFoundIndexConflictRecords
+// according to key and value.
+func newErrFoundConflictRecords(key []byte, value []byte, tbl table.Table) error {
+	sessionOpts := encode.SessionOptions{
+		SQLMode: mysql.ModeStrictAllTables,
```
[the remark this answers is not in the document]
lyzx2001: Conflict detection related code like the following all use `mysql.ModeStrictAllTables`, I think it would be ok if here we also use `mysql.ModeStrictAllTables`?
https://github.com/pingcap/tidb/blob/0d742d35f80eb9178105b1a09ee41eb693e6dbb4/br/pkg/lightning/importer/table_import.go#L1023-L1027
https://github.com/pingcap/tidb/blob/0d742d35f80eb9178105b1a09ee41eb693e6dbb4/pkg/ddl/ingest/backend.go#L98-L102

zimulala (approved):

pkg/ddl/ingest/backend.go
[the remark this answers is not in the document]
D3Hunter: it's very uncommon to used other return values when there's error returned

and this error is actually works as a struct to pass context info when duplicate is detected. so i prefer write in the way in previous comment

if there's no time to change, i think we can leave a todo to do it later

br/pkg/lightning/backend/local/duplicate.go
```diff
@@ -965,15 +1074,16 @@ func (local *DupeController) CollectLocalDuplicateRows(ctx context.Context, tbl
...
+	if err := duplicateManager.CollectDuplicateRowsFromDupDB(ctx, local.duplicateDB, local.keyAdapter, algorithm); err != nil {
 		return false, errors.Trace(err)
 	}
 	return duplicateManager.HasDuplicate(), nil
 }
 
 // CollectRemoteDuplicateRows collect duplicate keys from remote TiKV storage. This keys may be duplicate with
 // the data import by other lightning.
-func (local *DupeController) CollectRemoteDuplicateRows(ctx context.Context, tbl table.Table, tableName string, opts *encode.SessionOptions) (hasDupe bool, err error) {
+// TODO: revise the returned arguments to (hasDupe bool, dupInfo *DupInfo, err error) to distinguish the conflict error and the common error
```
D3Hunter, line 1085:
```suggestion
// TODO: revise the returned arguments to (dupInfo *DupInfo, err error) to distinguish the conflict error and the common error
```
hasDup can be infered from `dupInfo != nil`

D3Hunter (approved):

br/pkg/lightning/backend/local/duplicate.go
lyzx2001: Maybe not necessarily? Only when conflict detection mode is set to `error`, `CollectDuplicateRowsFromTiKV` would return `ErrFoundDataConflictRecords` or `ErrFoundIndexConflictRecords`. For other modes, we still need `duplicateManager.HasDuplicate()` to determine whether we have duplicate. And this information is needed for lightning
https://github.com/pingcap/tidb/blob/7548df70b1a4e84bd2ad2fa6cb52dfa819f4e5d3/br/pkg/lightning/importer/table_import.go#L1055-L1059

PR #51354, opened 2024-02-27
planner, executor: support aggregation as inner side of index join

### LINKED ISSUE #37068: Let Agg->DataSource be inner child of IndexJoin
## Enhancement

Currently, Agg->DataSource cannot be inner child of IndexJoin. Here is an example.

```
mysql> create table ta(a int, b int, index idx_a(a));
Query OK, 0 rows affected (0.03 sec)

mysql> create table tb(a int, b int, index idx_a(a));

mysql> explain select /*+ inl_join(t2@sel_2) */ t1.b,
    ->     (select count(t2.b) from tb t2 where t2.a = t1.a)
    -> from ta t1;
+------------------------------------+----------+-----------+---------------+---------------------------------------------------------------------------------------------+
| id                                 | estRows  | task      | access object | operator info                                                                               |
+------------------------------------+----------+-----------+---------------+---------------------------------------------------------------------------------------------+
| Projection_11                      | 10000.00 | root      |               | test.ta.b, ifnull(Column#10, 0)->Column#10                                                  |
| └─HashJoin_13                      | 10000.00 | root      |               | left outer join, equal:[eq(test.ta.a, test.tb.a)]                                           |
|   ├─HashAgg_34(Build)              | 7992.00  | root      |               | group by:test.tb.a, funcs:count(Column#15)->Column#10, funcs:firstrow(test.tb.a)->test.tb.a |
|   │ └─TableReader_35               | 7992.00  | root      |               | data:HashAgg_27                                                                             |
|   │   └─HashAgg_27                 | 7992.00  | cop[tikv] |               | group by:test.tb.a, funcs:count(test.tb.b)->Column#15                                       |
|   │     └─Selection_33             | 9990.00  | cop[tikv] |               | not(isnull(test.tb.a))                                                                      |
|   │       └─TableFullScan_32       | 10000.00 | cop[tikv] | table:t2      | keep order:false, stats:pseudo                                                              |
|   └─TableReader_26(Probe)          | 10000.00 | root      |               | data:TableFullScan_25                                                                       |
|     └─TableFullScan_25             | 10000.00 | cop[tikv] | table:t1      | keep order:false, stats:pseudo                                                              |
+------------------------------------+----------+-----------+---------------+---------------------------------------------------------------------------------------------+
9 rows in set, 1 warning (0.00 sec)

mysql> show warnings;
+---------+------+--------------------------------------------------------------------------------------------------------------------------------------------------+
| Level   | Code | Message                                                                                                                                          |
+---------+------+--------------------------------------------------------------------------------------------------------------------------------------------------+
| Warning | 1815 | There are no matching table names for (t2) in optimizer hint /*+ INL_JOIN(t2) */ or /*+ TIDB_INLJ(t2) */. Maybe you can use the table alias name |
+---------+------+--------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```
If both `ta` and `tb` are very large, HashJoin may consume too much memory and cause OOM. So we hope that IndexJoin can be used.

Here is the current logic rejecting Agg->DataSource as inner child of IndexJoin.
https://github.com/pingcap/tidb/blob/430e75e04f726d11015100b0a751e3c51d27dc6c/planner/core/exhaust_physical_plans.go#L669-L693

elsa0520:

Issue Number: close #37068

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/aggregate/agg_hash_executor.go
+++ b/pkg/executor/aggregate/agg_hash_executor.go
@@ -236,7 +236,11 @@ func (e *HashAggExec) Open(ctx context.Context) error {
 	if err := e.BaseExecutor.Open(ctx); err != nil {
 		return err
 	}
+	return e.OpenSelf()
+}
 
+// OpenSelf just opens the hash aggregation executor.
+func (e *HashAggExec) OpenSelf() error {
 	e.prepared.Store(false)
 
 	if e.memTracker != nil {
--- a/pkg/executor/aggregate/agg_stream_executor.go
+++ b/pkg/executor/aggregate/agg_stream_executor.go
@@ -64,7 +64,11 @@ func (e *StreamAggExec) Open(ctx context.Context) error {
 	}
 	// If panic in Open, the children executor should be closed because they are open.
 	defer closeBaseExecutor(&e.BaseExecutor)
+	return e.OpenSelf()
+}
 
+// OpenSelf just opens the StreamAggExec.
+func (e *StreamAggExec) OpenSelf() error {
 	e.childResult = exec.TryNewCacheChunk(e.Children(0))
 	e.executed = false
 	e.IsChildReturnEmpty = true
--- a/pkg/planner/core/exhaust_physical_plans.go
+++ b/pkg/planner/core/exhaust_physical_plans.go
@@ -744,6 +744,20 @@ func (p *LogicalJoin) getIndexJoinByOuterIdx(prop *property.PhysicalProperty, ou
 	return p.buildIndexJoinInner2IndexScan(prop, innerChildWrapper, innerJoinKeys, outerJoinKeys, outerIdx, avgInnerRowCnt)
 }
 
+// indexJoinInnerChildWrapper is a wrapper for the inner child of an index join.
+// It contains the lowest DataSource operator and other inner child operator
+// which is flattened into a list structure from tree structure .
+// For example, the inner child of an index join is a tree structure like:
+//
+//	Projection
+//	       Aggregation
+//				Selection
+//					DataSource
+//
+// The inner child wrapper will be:
+// DataSource: the lowest DataSource operator.
+// hasDitryWrite: whether the inner child contains dirty data.
+// zippedChildren: [Projection, Aggregation, Selection]
 type indexJoinInnerChildWrapper struct {
 	ds             *DataSource
 	hasDitryWrite  bool
@@ -764,7 +778,7 @@ childLoop:
 		case *DataSource:
 			wrapper.ds = child
 			break childLoop
-		case *LogicalProjection, *LogicalSelection:
+		case *LogicalProjection, *LogicalSelection, *LogicalAggregation:
 			if !p.SCtx().GetSessionVars().EnableINLJoinInnerMultiPattern {
 				return nil
 			}
@@ -989,7 +1003,8 @@ type indexJoinBuildHelper struct {
 	curPossibleUsedKeys []*expression.Column
 	curNotUsedIndexCols []*expression.Column
 	curNotUsedColLens   []int
-	curIdxOff2KeyOff    []int
+	// Store the corresponding innerKeys offset of each column in the current path, reset by "resetContextForIndex()"
+	curIdxOff2KeyOff []int
 }
 
 func (ijHelper *indexJoinBuildHelper) buildRangeDecidedByInformation(idxCols []*expression.Column, outerJoinKeys []*expression.Column) string {
@@ -1092,24 +1107,33 @@ func (p *LogicalJoin) constructInnerTableScanTask(
 	ts.PlanPartInfo = copTask.physPlanPartInfo
 	selStats := ts.StatsInfo().Scale(selectivity)
 	ts.addPushedDownSelection(copTask, selStats)
-	t := copTask.ConvertToRootTask(ds.SCtx()).(*RootTask)
-	reader := t.GetPlan()
-	t.SetPlan(p.constructInnerByWrapper(wrapper, reader))
-	return t
+	return p.constructIndexJoinInnerSideTask(copTask, ds, nil, wrapper)
 }
 
-func (p *LogicalJoin) constructInnerByWrapper(wrapper *indexJoinInnerChildWrapper, child base.PhysicalPlan) base.PhysicalPlan {
-	for i := len(wrapper.zippedChildren) - 1; i >= 0; i-- {
-		switch x := wrapper.zippedChildren[i].(type) {
+func (p *LogicalJoin) constructInnerByZippedChildren(zippedChildren []LogicalPlan, child base.PhysicalPlan) base.PhysicalPlan {
+	for i := len(zippedChildren) - 1; i >= 0; i-- {
+		switch x := zippedChildren[i].(type) {
 		case *LogicalUnionScan:
 			child = p.constructInnerUnionScan(x, child)
 		case *LogicalProjection:
 			child = p.constructInnerProj(x, child)
 		case *LogicalSelection:
 			child = p.constructInnerSel(x, child)
+		case *LogicalAggregation:
+			child = p.constructInnerAgg(x, child)
 		}
 	}
 	return child
+}
+
+func (*LogicalJoin) constructInnerAgg(logicalAgg *LogicalAggregation, child base.PhysicalPlan) base.PhysicalPlan {
+	if logicalAgg == nil {
+		return child
+	}
+	physicalHashAgg := NewPhysicalHashAgg(logicalAgg, logicalAgg.StatsInfo(), nil)
+	physicalHashAgg.SetSchema(logicalAgg.schema.Clone())
+	physicalHashAgg.SetChildren(child)
+	return physicalHashAgg
 }
 
 func (*LogicalJoin) constructInnerSel(sel *LogicalSelection, child base.PhysicalPlan) base.PhysicalPlan {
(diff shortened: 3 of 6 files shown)


elsa0520:
TODO, add stream agg, two-phase-hash agg , one-phase -hash agg case.
test the correct of result

pkg/executor/aggregate/agg_hash_executor.go
```diff
@@ -236,7 +236,11 @@ func (e *HashAggExec) Open(ctx context.Context) error {
 	if err := e.BaseExecutor.Open(ctx); err != nil {
 		return err
 	}
+	return e.OpenInternal()
+}
 
+// OpenInternal just opens the hash aggregation executor.
+func (e *HashAggExec) OpenInternal() error {
```
qw4990, line 243: Should we rename it to 'openInternal' to not expose this function since it's 'internal'?

elsa0520: Renamed to "OpenExecutor" ? This is a part of logic from original Open function. The outside need to only open Agg executor so I separate it.

pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -1377,10 +1387,105 @@ func (p *LogicalJoin) constructInnerIndexScanTask(
...
+
+// construct the inner join task by inner child plan tree
+// The Logical include two parts: logicalplan->physicalplan, phsicaplan->task
+// Step1: whether agg can be pushed down to coprocessor
+//
+//	Step1.1: If the agg can be pushded down to coprocessor, we will build a copTask and attach the agg to the copTask
+//	There are two kinds of agg: stream agg and hash agg. Stream agg depends on some conditions, such as the group by cols
+//
+// Step2: build other inner plan node to task
+func (p *LogicalJoin) constructIndexJoinInnerSideTask(dsCopTask *copTask, ds *DataSource, path *util.AccessPath, wrapper *indexJoinInnerChildWrapper) task {
```
qw4990, line 1401: Could you add some comments for `indexJoinInnerChildWrapper`, I just found there is no any comment for this tricky structure...

pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -1377,10 +1387,105 @@ func (p *LogicalJoin) constructInnerIndexScanTask(
...
+	if len(wrapper.zippedChildren) == 0 {
+		t := dsCopTask.convertToRootTask(ds.SCtx())
+		return t
+	}
+	la, canPushAggToCop := wrapper.zippedChildren[len(wrapper.zippedChildren)-1].(*LogicalAggregation)
```
qw4990, line 1406: Why can we assume there is only one Agg in `zippedChildren` and its index must be `len(zippedChildren)-1`?

pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -1377,10 +1387,105 @@ func (p *LogicalJoin) constructInnerIndexScanTask(
...
+			}
+			sctx := p.SCtx()
+			for i, groupbyCol := range groupByCols {
+				if path.IdxColLens[i] != types.UnspecifiedLength ||
+					!groupbyCol.EqualByExprAndID(sctx.GetExprCtx().GetEvalCtx(), path.IdxCols[i]) {
+					preferStream = false
+				}
+			}
+		} else {
+			preferStream = false
```
qw4990, line 1445: Can we remove those checks of `preferStream` and just enforce the optimizer to always use `HashAgg`, which can reduce our risks and let the code simpler?

pkg/planner/core/exhaust_physical_plans.go
elsa0520: If there is an agg node at the bottom closest to the datasource, we need to consider whether it should be pushed down to tikv. If not, skip this part of the logic directly.

pkg/planner/core/exhaust_physical_plans.go
elsa0520: This is not better for performance if we directly use ```HashAgg```.

pkg/executor/aggregate/agg_hash_executor.go
elsa0520: Changed

pkg/planner/core/exhaust_physical_plans.go
elsa0520: Added on the structure

qw4990 (approved):

pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -1378,10 +1402,105 @@ func (p *LogicalJoin) constructInnerIndexScanTask(
...
+		groupByCols := la.GetGroupByCols()
+		if len(groupByCols) != len(la.GroupByItems) {
+			preferStream = false
+		}
+		if la.HasDistinct() && !la.distinctArgsMeetsProperty() {
+			preferStream = false
+		}
+		// sort items must be the super set of group by items
+		if path != nil && path.Index != nil && ds.tableInfo.GetPartitionInfo() == nil {
+			if len(path.IdxCols) < len(groupByCols) {
```
AilinKid, line 1449: skip mv index?

AilinKid: remember change this

AilinKid (approved):

pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -1378,10 +1402,105 @@ func (p *LogicalJoin) constructInnerIndexScanTask(
...
+		if !ok {
+			return nil
+		}
+		result.p = p.constructInnerByZippedChildren(wrapper.zippedChildren[0:len(wrapper.zippedChildren)-1], result.p)
+		return result
+	}
+
+	result := dsCopTask.ConvertToRootTask(ds.SCtx())
+	result.SetPlan(p.constructInnerByZippedChildren(wrapper.zippedChildren, result.GetPlan()))
+	return result
```
time-and-fate, line 1503: Looks like this case can be combined into the `if len(wrapper.zippedChildren) == 0` case above, just like the previous code, and can make the code structure slightly simpler.

pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -742,6 +742,20 @@ func (p *LogicalJoin) getIndexJoinByOuterIdx(prop *property.PhysicalProperty, ou
...
+//
+//	Projection
+//	       Aggregation
+//				Selection
+//					DataSource
+//
+// The inner child wrapper will be:
+// DataSource: the lowest DataSource operator.
+// hasDitryWrite: whether the inner child contains dirty data.
+// zippedChildren: [Selection, Aggregation, Projection]
```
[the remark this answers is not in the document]
time-and-fate: And we don't have test cases for `Aggregation` and other operators appear together in `zippedChildren`.

tests/integrationtest/t/planner/core/indexjoin.test
```diff
@@ -17,4 +17,31 @@ select /*+ INL_JOIN(t1) */ * from t, t1 where t.a=t1.a and t.b = t1.a+t1.b;
...
-set @@tidb_enable_inl_join_inner_multi_pattern=default;
\ No newline at end of file
+# Test agg as inner side of index join
+explain format='brief' select /*+ INL_JOIN(tmp) */ * from (select a, count(b) from t group by a) tmp, t1 where tmp.a=t1.a;
+select /*+ INL_JOIN(tmp) */ t1.a, tmp.count_b from (select a, count(b) count_b from t group by a) tmp, t1 where tmp.a=t1.a order by t1.a, tmp.count_b;
+select /*+ HASH_JOIN(tmp) */ t1.a, tmp.count_b from (select a, count(b) count_b from t group by a) tmp, t1 where tmp.a=t1.a order by t1.a, tmp.count_b;
+explain format='brief' select /*+ INL_JOIN(tmp) */ * from (select a, count(b) count_b from t group by a) tmp, t1 where tmp.a=t1.a and tmp.count_b = t1.b;
+select /*+ INL_JOIN(tmp) */  t1.a, tmp.count_b from (select a, count(b) count_b from t group by a) tmp, t1 where tmp.a=t1.a and tmp.count_b = t1.b order by  t1.a, tmp.count_b;
+select /*+ HASH_JOIN(tmp) */  t1.a, tmp.count_b from (select a, count(b) count_b from t group by a) tmp, t1 where tmp.a=t1.a and tmp.count_b = t1.b order by  t1.a, tmp.count_b;
+# hint doesn't work
```
time-and-fate, line 27: So this "hint doesn't work" is expected, right? And only the test case at L28 can't generate IndexJoin. I see the test case at L29 and L32 can generate IndexJoin.
I think we need to clarify this.

pkg/planner/core/exhaust_physical_plans.go
elsa0520: Add the skip logic of mv index

pkg/planner/core/exhaust_physical_plans.go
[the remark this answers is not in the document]
elsa0520: Changed comment ~
Also add some test for other operators in ```zippedChildren```

elsa0520:
> We don't have any limitations on the group by columns? I think that will cause wrong query results.

We don't need to additional limitation of the group by columns because we already removed all of not source column in the front logic

AilinKid (approved):

pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -1379,10 +1403,115 @@ func (p *LogicalJoin) constructInnerIndexScanTask(
...
+
+// construct the inner join task by inner child plan tree
+// The Logical include two parts: logicalplan->physicalplan, physicalplan->task
+// Step1: whether agg can be pushed down to coprocessor
+//
+//	Step1.1: If the agg can be pushded down to coprocessor, we will build a copTask and attach the agg to the copTask
+//	There are two kinds of agg: stream agg and hash agg. Stream agg depends on some conditions, such as the group by cols
+//
+// Step2: build other inner plan node to task
+func (p *LogicalJoin) constructIndexJoinInnerSideTask(dsCopTask *CopTask, ds *DataSource, path *util.AccessPath, wrapper *indexJoinInnerChildWrapper) base.Task {
```
time-and-fate, line 1417: I think this method name can be improved.
1. It's too similar to the `constructInnerIndexScanTask` above.
2. This method is mostly about `Aggregate` but the name doesn't reflect this.

tests/integrationtest/t/planner/core/indexjoin.test
```diff
@@ -17,4 +17,44 @@ select /*+ INL_JOIN(t1) */ * from t, t1 where t.a=t1.a and t.b = t1.a+t1.b;
...
+select /*+ INL_JOIN(tmp) */ tmp.a, t1.b from (select a, b from t group by a, b) tmp, t1 where tmp.a=t1.a order by tmp.a, t1.b;
+select /*+ HASH_JOIN(tmp) */ tmp.a, t1.b from (select a, b from t group by a, b) tmp, t1 where tmp.a=t1.a order by tmp.a, t1.b;
+explain format='brief' select /*+ INL_JOIN(tmp) */ * from (select a, b from t where a=1 group by a, b having a>0 ) tmp, t1 where tmp.a=t1.a;
+select /*+ INL_JOIN(tmp) */ tmp.a, t1.b from (select a, b from t where a=1 group by a, b having a>0 ) tmp, t1 where tmp.a=t1.a order by tmp.a, t1.b;
+select /*+ HASH_JOIN(tmp) */ tmp.a, t1.b from (select a, b from t where a=1 group by a, b having a>0 ) tmp, t1 where tmp.a=t1.a order by tmp.a, t1.b;
+# hint doesn't work
+explain format='brief' select /*+ INL_JOIN(tmp) */ * from (select a +1 as a1, b from t group by a, b ) tmp, t1 where tmp.a1=t1.a;
+
+# Test the diff type with same column
+# The firstrow() function will changed the column type from float to double but it will not changed the unique id of column. So we need to test the result of index join.
```
time-and-fate, line 40: But looks like this query can't generate index join.

time-and-fate (commented):
Don't forget to check the estimation logic~

pkg/executor/aggregate/agg_stream_executor.go
```diff
@@ -64,7 +64,11 @@ func (e *StreamAggExec) Open(ctx context.Context) error {
 	}
 	// If panic in Open, the children executor should be closed because they are open.
 	defer closeBaseExecutor(&e.BaseExecutor)
```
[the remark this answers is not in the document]
elsa0520: Panic can be recovered by the outer layer, but if a problem occurs during the open process, it is not sure whether the children will be closed and whether to close the children.
But looking at the previous logic, it seems that close children are not handled

Issue #43527, opened 2023-05-04
Got interface conversion error when there are get_sys_var expr


guo-shaoge:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
```
create table test (a datetime , b bigint,c bigint);
insert into test values('2010-10-10 10:10:10', 100, 100);
select a,b as 'asd',@total:=@total+b from (select date(a) as a,sum(c) as b from test group by date(a)) as temp,(select @total:=0) as T1 order by a;
```
<!-- a step by step guide for reproducing the bug. -->

### 2. What did you expect to see? (Required)
runs ok
### 3. What did you see instead (Required)
```
ERROR 1105 (HY000): interface conversion: interface {} is nil, not *types.MyDecimal
```
### 4. What is your TiDB version? (Required)
```
MySQL [test]> select tidb_version();
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| tidb_version()                                                                                                                                                                                                                                                                                           |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Release Version: v7.2.0-alpha
Edition: Community
Git Commit Hash: 7eff8ce629d00cf4ab1eeda82a1ecd5a4e394b02
Git Branch: heads/refs/tags/v7.2.0-alpha
UTC Build Time: 2023-05-03 14:33:44
GoVersion: go1.20.3
Race Enabled: false
TiKV Min Version: 6.2.0-alpha
Check Table Before Drop: false
Store: tikv |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.009 sec)
```
<!-- Paste the output of SELECT tidb_version() -->

guo-shaoge:
1. `(select @total:=0)` makes `@total` as int type
2. `@total:=@total+b`: `b` is decimal type, so `+` will call `getVar.evalDecimal`, but `@total` is actually a int value, so we got error [here](https://github.com/pingcap/tidb/blob/master/types/datum.go#L347)

A `cast` expr is missing for `getVar`, check [how getVar is built](https://github.com/pingcap/tidb/blob/master/expression/scalar_function.go#L185), we should add cast when build `getVar` expr

zanmato1984:
[...]

PR #51395, opened 2024-02-28
expression: use the correct type when eval decimal system var

hi-rustin:
Signed-off-by: hi-rustin <rustin.liu@gmail.com>

Issue Number: close https://github.com/pingcap/tidb/issues/43527

Problem Summary:

```sql
create table test (a datetime, b bigint, c decimal(10, 2), d float);
insert into test values('2010-10-10 10:10:10', 100, 100.01, 100);
SELECT @total := @total + c FROM (SELECT c FROM test) AS temp, (SELECT @total := 200) AS T1;
```

In the SQL, `@total` type is int64, but when we do eval decimal we will directly convert it as decimal then we will get a panic because there is no `c` for int value.
GetMysqlDecimal: 
```go
func (d *Datum) GetMysqlDecimal() *MyDecimal {
	return d.x.(*MyDecimal)
}
```
ToDecimal: 
```go
func (d *Datum) ToDecimal(ctx Context) (*MyDecimal, error) {
	switch d.Kind() {
	case KindMysqlTime:
		return d.GetMysqlTime().ToNumber(), nil
	case KindMysqlDuration:
		return d.GetMysqlDuration().ToNumber(), nil
	default:
		return ConvertDatumToDecimal(ctx, *d)
	}
}
```

So we have to try to convert it explicitly.

Same for the float value eval process. We also need to convert it.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fixed an issue that TiDB could not correctly convert the type of a session variable
```

### DIFF
--- a/pkg/expression/builtin_other.go
+++ b/pkg/expression/builtin_other.go
@@ -1052,7 +1052,11 @@ func (b *builtinGetRealVarSig) evalReal(ctx EvalContext, row chunk.Row) (float64
 	}
 	varName = strings.ToLower(varName)
 	if v, ok := sessionVars.GetUserVarVal(varName); ok {
-		return v.GetFloat64(), false, nil
+		d, err := v.ToFloat64(typeCtx(ctx))
+		if err != nil {
+			return 0, false, err
+		}
+		return d, false, nil
 	}
 	return 0, true, nil
 }
--- a/pkg/expression/builtin_other.go
+++ b/pkg/expression/builtin_other.go
@@ -1092,7 +1092,11 @@ func (b *builtinGetDecimalVarSig) evalDecimal(ctx EvalContext, row chunk.Row) (*
 	}
 	varName = strings.ToLower(varName)
 	if v, ok := sessionVars.GetUserVarVal(varName); ok {
-		return v.GetMysqlDecimal(), false, nil
+		d, err := v.ToDecimal(typeCtx(ctx))
+		if err != nil {
+			return nil, false, err
+		}
+		return d, false, nil
 	}
 	return nil, true, nil
 }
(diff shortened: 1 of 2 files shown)


hi-rustin (commented):
🔢 Self-check

AilinKid (approved):

elsa0520 (approved):
LGTM

pkg/expression/builtin_other.go
```diff
@@ -1052,7 +1052,11 @@ func (b *builtinGetRealVarSig) evalReal(ctx EvalContext, row chunk.Row) (float64
 	}
 	varName = strings.ToLower(varName)
 	if v, ok := sessionVars.GetUserVarVal(varName); ok {
-		return v.GetFloat64(), false, nil
```
XuHuaiyu, line 1055: What about the implementation`(b *builtinGetRealVarSig) vecEvalReal`

zanmato1984: +1

zanmato1984: +1

zanmato1984 (changes requested):
Is there any case we can add into this UT: https://github.com/pingcap/tidb/blob/27ce02afd2e3d312b2bb92e20489cefce000e320/pkg/expression/builtin_other_test.go#L118.
As well as https://github.com/pingcap/tidb/blob/27ce02afd2e3d312b2bb92e20489cefce000e320/pkg/expression/builtin_other_vec_test.go , assuming you are going to add vectorized version of `builtinGetRealVarSig`.

pkg/expression/builtin_other.go
zanmato1984: Yes, please. Much appreciated.

Let's talk about enabling vectorization for get/set var function in the future.

hi-rustin (commented):
🔢 Self-check

zanmato1984 (approved):
LGTM

PR #51435, opened 2024-02-29
planner: trigger stats load on a column in determinate mode to make sure analyze row count is available

### LINKED ISSUE #48257: TableFullScan uses pseudo stats in determinate mode
## Bug Report

Create a table without indexes, and insert some rows of data. Wait a minute and analyze the table.

```
mysql> show stats_meta;
| test    | test       |                | 2023-11-03 10:07:27 |            0 |      2048 |
1 row in set (0.00 sec)

mysql> explain select * from test;
+-----------------------+---------+-----------+---------------+----------------------+
| id                    | estRows | task      | access object | operator info        |
+-----------------------+---------+-----------+---------------+----------------------+
| TableReader_5         | 2048.00 | root      |               | data:TableFullScan_4 |
| └─TableFullScan_4     | 2048.00 | cop[tikv] | table:test    | keep order:false     |
+-----------------------+---------+-----------+---------------+----------------------+
2 rows in set (0.00 sec)

mysql> set tidb_opt_objective='determinate';

mysql> explain select * from test;
+-----------------------+----------+-----------+---------------+--------------------------------+
| id                    | estRows  | task      | access object | operator info                  |
+-----------------------+----------+-----------+---------------+--------------------------------+
| TableReader_5         | 10000.00 | root      |               | data:TableFullScan_4           |
| └─TableFullScan_4     | 10000.00 | cop[tikv] | table:test    | keep order:false, stats:pseudo |
+-----------------------+----------+-----------+---------------+--------------------------------+
2 rows in set (0.00 sec)
```

In both modes, the TableFullScan has the correct row count.

In determinate mode, the TableFullScan has a row count of 10000.

v7.6.0-alpha-nightly-20231101

time-and-fate:

Issue Number: close  #48257

Problem Summary:

Currently, if there are no predicates on one table, no stats loading will be triggered.
But in the determinate mode, if there are no full-loaded stats on this table, we are unable to get the "analyze row count".
This won't be a problem in most times, but it may cause a problem if it's joined with other tables.

In the determinate mode, if:
1. no sync stats loading is triggered on a table, and
2. there are no full-loaded stats on this table

then we arbitrarily choose one column to trigger stats loading.

This is implemented in `collectPredicateColumnsPoint` while effective in both sync/async load modes.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/plan_stats.go
```diff
@@ -39,24 +40,83 @@ func (collectPredicateColumnsPoint) optimize(_ context.Context, plan LogicalPlan
...
+			if colStats := tblStats.Columns[col.ID]; colStats != nil {
+				// Choose the first column we meet to trigger stats loading.
+				if colToTriggerLoad == nil && colStats.IsStatsInitialized() {
+					colToTriggerLoad = &model.TableItemID{TableID: int64(physicalTblID), ID: col.ID, IsIndex: false}
+				}
+				// If any stats are already full loaded, we don't need to trigger stats loading on this table.
+				if colStats.IsFullLoad() {
+					colToTriggerLoad = nil
+					break
+				}
```
winoros, line 95: Do we need to check it here since we have the check `physTblIDsWithNeededCols.Has(physicalTblID)`?

pkg/planner/core/collect_column_stats_usage.go
```diff
@@ -159,6 +167,10 @@ func (c *columnStatsUsageCollector) collectPredicateColumnsForUnionAll(p *Logica
 }
 
 func (c *columnStatsUsageCollector) addHistNeededColumns(ds *DataSource) {
+	c.visitedPhysTblIDs.Insert(int(ds.physicalTableID))
```
winoros, line 170: If we insert the id into the map after the line178 and check `if len(columns) == 0 then add it to the map`. We don't need the struct `physTblIDsWithNeededCols` in the `plan_stats.go`'s `optimize()`?

time-and-fate: Yes.
But personally, I tend to output information for possibly more general usage and keep the "determinate mode" specific logic out of this collector.

pkg/planner/core/plan_stats.go
time-and-fate: They are different.
`physTblIDsWithNeededCols.Has(physicalTblID)` means we will trigger sync loading on some columns for this table now, no matter if there are already some usable stats.
`colStats.IsFullLoad()` means we already have some column stats that can be used and also implies that these stats are not needed by the predicates.

winoros (approved):

PR #51441, opened 2024-03-01
br: Enable checkpoint advancer to pause tasks lagged too large

### LINKED ISSUE #50803: Circuit breaker log backup task when checkpoint ts does not advance for a long time
## Bug Report

log backup stop advance by whatever reason. (usually it caused by a bug like https://github.com/tikv/tikv/issues/16056). eventually we need a protection mechanism. 

log backup won't block GC for a long time.

log backup block GC for a month.

below 6.5.6

RidRisR:

Issue Number: close #50803

Problem Summary:

Now, a new option called `CheckPointLagLimit` is added to the advancer config. When it is set, advancer will check if the checkpoint is lagged too large on every tick. If so, advancer would send a warning to PD and unstall the task.
PS: Theoretically, we need a resume signal when the task could be resumed. However, there is no channel to send this signal now. Maybe we will add it in the future.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
A new option called `CheckPointLagLimit` is added to the advancer config. When it is set, advancer will check if the checkpoint is lagged too large on every tick. If so, advancer would send a warning to PD and unstall the task.
```


br/pkg/streamhelper/advancer.go
```diff
@@ -544,13 +553,43 @@ func (c *CheckpointAdvancer) subscribeTick(ctx context.Context) error {
...
 }
 
+func (c *CheckpointAdvancer) isCheckpointLagged(ctx context.Context) (bool, error) {
+	if c.cfg.CheckPointLagLimit <= 0 {
+		return false, nil
+	}
+
+	now, err := c.env.FetchCurrentTS(ctx)
+	if err != nil {
+		return true, err
```
YuJuncen, line 563: We may encounter some temporary errors when fetching TSO from a remote node. Maybe always return `false` when `err` isn't `nil`?

br/pkg/streamhelper/advancer_cliext.go
```diff
@@ -127,13 +154,22 @@ func (t AdvancerExt) startListen(ctx context.Context, rev int64, ch chan<- TaskE
...
 			select {
-			case resp, ok := <-c:
+			case resp, ok := <-taskCh:
+				if !ok {
+					return
+				}
+				if !handleResponse(resp) {
+					return
+				}
+			case resp, ok := <-pauseCh:
```
YuJuncen, line 172: We may want to "merge" all remaining events from both `taskCh` and `pauseCh`, so I think we'd better return once both `taskCh` and `pauseCh` are closed.
Tips: a `nil` channel in go will block forever, so you may consume all data in the two channels via replacing the `!ok` branch with:
```go
// ...
if !ok {
  taskCh = nil
}
// ...
```

br/pkg/streamhelper/regioniter_test.go
```diff
@@ -81,6 +82,10 @@ func (c constantRegions) BlockGCUntil(ctx context.Context, at uint64) (uint64, e
 	return 0, status.Error(codes.Unimplemented, "Unsupported operation")
 }
 
+func (c constantRegions) FetchCurrentTS(ctx context.Context) (uint64, error) {
+	return uint64(time.Now().Unix()), nil
```
YuJuncen, line 86: For making it an TSO, you may also need to append it with a "logical part" and shift the physical unix millisecond left. Read [this](https://docs.pingcap.com/tidb/stable/tso#timestamp-oracle-tso-in-tidb) for more details.

TL;DR: You may directly use `oracle.ComposeTS(time.Now().UnixMilli(), 0)` here.

br/pkg/streamhelper/regioniter.go
```diff
@@ -42,6 +42,9 @@ type TiKVClusterMeta interface {
 	// NOTE: once we support multi tasks, perhaps we need to allow the caller to provide a namespace.
 	// For now, all tasks (exactly one task in fact) use the same checkpoint.
 	BlockGCUntil(ctx context.Context, at uint64) (uint64, error)
+
+	//TODO: It should be able to synchoronize the current TS with the PD.
```
YuJuncen, line 46: I guess we'd better add this TODO at the implementation side, instead of the interface side.

br/pkg/streamhelper/advancer_test.go
```diff
@@ -463,3 +464,80 @@ func TestRemoveTaskAndFlush(t *testing.T) {
...
+	adv := streamhelper.NewCheckpointAdvancer(env)
+	adv.UpdateConfigWith(func(c *config.Config) {
+		c.CheckPointLagLimit = 1 * time.Minute
+	})
+	adv.StartTaskListener(ctx)
+	c.advanceClusterTimeBy(1 * time.Minute)
+	require.NoError(t, adv.OnTick(ctx))
+	c.advanceClusterTimeBy(1 * time.Minute)
+	require.ErrorContains(t, adv.OnTick(ctx), "lagged too large")
+	// after some times, the isPaused will be set and ticks are skipped
```
YuJuncen, line 507: Would you also check whether the task has been paused?

br/pkg/streamhelper/advancer.go
```diff
@@ -544,13 +553,43 @@ func (c *CheckpointAdvancer) subscribeTick(ctx context.Context) error {
...
+	if err != nil {
+		return errors.Annotate(err, "failed to check timestamp")
+	}
+	if isLagged {
+		err := c.env.PauseTask(ctx, c.task.Name)
+		if err != nil {
+			return errors.Annotate(err, "failed to pause task")
+		}
+		return errors.Annotate(errors.Errorf("check point lagged too large"), "check point lagged too large")
+	}
```
YuJuncen, line 592: Note: we may unblock the GC immediately here. But this will make the task cannot be resumed any more. Perhaps we need a discussion about this.

br/pkg/streamhelper/advancer.go
```diff
@@ -446,6 +447,14 @@ func (c *CheckpointAdvancer) onTaskEvent(ctx context.Context, e TaskEvent) error
 			log.Warn("failed to remove service GC safepoint", logutil.ShortError(err))
 		}
 		metrics.LastCheckpoint.DeleteLabelValues(e.Name)
+	case EventPause:
+		if c.task.GetName() == e.Name {
+			c.isPaused.CompareAndSwap(false, true)
+		}
+	case EventResume:
+		if c.task.GetName() == e.Name {
+			c.isPaused.CompareAndSwap(true, false)
```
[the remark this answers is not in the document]
RidRisR: Yes, they are equivalent. I just want to avoid to modify the value if redundant pause or resume signal are captured

br/pkg/streamhelper/advancer_test.go
RidRisR: It has been checked. When the task has been paused, `isPaused` field will be set and `tick` will run normally without error (for it is skipped).

br/pkg/streamhelper/advancer.go
RidRisR: That may lead to some unexpected problem when the errors are not just "temporary", but I that make sense.

YuJuncen (approved):
LGTM. 
Note: for now it seems this function is disabled by default. For now, the configuration of TiDB cannot be passed to advancer, which means this function won't be usable in a TiDB binary. Should we enable this with a value safe enough, say, 72h or even longer? cc @3pointer @BornChanger

Leavrth (approved):

br/pkg/streamhelper/advancer_cliext.go
```diff
@@ -127,21 +154,26 @@ func (t AdvancerExt) startListen(ctx context.Context, rev int64, ch chan<- TaskE
...
-					return
+			case resp, ok := <-taskCh:
+				if !ok || !handleResponse(resp) {
+					taskCh = nil
 				}
-				if !handleResponse(resp) {
-					return
+			case resp, ok := <-pauseCh:
+				if !ok || !handleResponse(resp) {
+					pauseCh = nil
```
YuJuncen, line 175: Nits: I think we'd better bring the default branch back and remove the `if`. Even in this context the `ctx` should be canceled and the channel must be closed. In a cleanup context, unbounded blocking is somehow dangerous -- it may lead to unexpected leakage or make the program fail to exit.

3pointer:
> LGTM. Note: for now it seems this function is disabled by default. For now, the configuration of TiDB cannot be passed to advancer, which means this function won't be usable in a TiDB binary. Should we enable this with a value safe enough, say, 72h or even longer? cc @3pointer @BornChanger

I think we should enable it with a default value,  3 days is safe enough and we could not let log backup block GC to much.

PR #51484, opened 2024-03-04
br: skip the later pool tasks right away if context is canceled

### LINKED ISSUE #50800: backup/restore print much WARN log with context canceled error after failed
## Bug Report

1. backup/restore
2. failed by some error

1. the log has no unnecessary warning logs with error `context canceled`.
1. many warning logs with error `context canceled`.
master

Leavrth:

Issue Number: close #50800

Problem Summary:
if context is canceled, the workerpool still load the new tasks.
check the status of context before load a task
### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


br/pkg/utils/worker.go
```diff
@@ -61,41 +63,54 @@ func (pool *WorkerPool) Apply(fn taskFunc) {
...
 	})
+	return nil
 }
 
 // ApplyWorker apply a worker.
-func (pool *WorkerPool) ApplyWorker() *Worker {
+func (pool *WorkerPool) ApplyWorker(ctx context.Context) (*Worker, error) {
 	var worker *Worker
+	if ctx.Err() != nil {
+		return nil, context.Cause(ctx)
```
lance6716, line 103: we still return "context cancel" error and how does it solve origin issue?

lance6716 (commented):
And I think we do not need to log or return "context cancel" error because they are derived from other errors. As for using egCtx, I think the callback function should be aware of it, rather than WorkerPool, unless the quick `ApplyWorker` and `RecycleWorker` causes CPU spike and further problems

br/pkg/utils/worker.go
Leavrth: If the function in the workerpool return an error, then the errgroup `eg` would cancel the context `ectx` with the error.
Therefore the context `ectx` has these values:
```
ectx: context.Context {
  Error: "context cancel",
  Cause: "{error}",
}
```

PR #51618, opened 2024-03-08
distsql: use a more accurate type of the context of distsql

### LINKED ISSUE #51617: Use a more accurate `DistSQLContext` to avoid depending on the whole session variables
## Enhancement

As the `DistSQLContext` only uses a small subset of all variables and session tools, we can use a more accurate `DistSQLContext` to avoid depending on a whole session variables. And it'll help us to understand how to clone it.

There are also problems about the `Reset` and dependency loop. Obviously the `DistSQLContext` should be re-generated after each statement, but some fields can only be defined well in session (because of circular dependency).

YangKeao:

Issue Number: close #51617

Problem Summary:

The `distsql` only needs a subset of the whole session variables and context, so we can use a more accurate type to represent it. This refactor will help us to clone it in the future.

1. Give a struct `DistSQLContext`, which contains all information needed by the session.
2. Add a cache on the `StmtCtx`, because the cache should be invalidated after each execution.
3. Remove the dependency of `stmtctx` (and session, maybe) from `distsql`.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > It should be covered by existing tests. Some tests also need to be modified.

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/adapter.go
```diff
@@ -504,21 +505,25 @@ func (a *ExecStmt) Exec(ctx context.Context) (_ sqlexec.RecordSet, err error) {
...
 		_, planDigest := GetPlanDigest(stmtCtx)
-		_, digest := stmtCtx.SQLDigest()
-		stmtCtx.RunawayChecker = domain.GetDomain(sctx).RunawayManager().DeriveChecker(sctx.GetSessionVars().StmtCtx.ResourceGroupName, stmtCtx.OriginalSQL, digest.String(), planDigest.String())
+		_, sqlDigest := stmtCtx.SQLDigest()
+		stmtCtx.RunawayChecker = domain.GetDomain(sctx).RunawayManager().DeriveChecker(sctx.GetSessionVars().StmtCtx.ResourceGroupName, stmtCtx.OriginalSQL, sqlDigest.String(), planDigest.String())
 		if err := stmtCtx.RunawayChecker.BeforeExecutor(); err != nil {
 			return nil, err
 		}
 	}
+	a.initKvExecCounter()
```
YangKeao, line 519: The most risky part of this PR. I adjusted the order of initialization of multiple fields.
1. The `RunawayChecker`
2. The `KvExecCounter`
3. The plan from `Execute` plan

The `plan` should be initialized before `KvExecCounter` and `RunawayChecker`, to use the correct digest. However, previously the `plan` for `Execute` statements are extracted inside `buildExecutor` (from `ExecuteExec`). I modifed to build the `plan` from `*plannercore.Execute`, and initialize them before `buildExecutor`.

I hope everything needed by executors can be initialized before `buildExecutor`. However, it's not true now, a lot of fields are initialized after `buildExecutor` (by setting fields of `StmtCtx`), and the dependency between them are not clear.

lcwangchao (approved):

pkg/sessionctx/stmtctx/stmtctx.go
```diff
@@ -167,6 +168,15 @@ type StatementContext struct {
 	// errCtx is used to indicate how to handle the errors
 	errCtx errctx.Context
 
+	DistSQLContext distsqlctx.DistSQLContext
```
zimulala, line 171: Do we use this variable？

zimulala (approved):

pkg/sessionctx/stmtctx/stmtctx.go
YangKeao: No. I'll delete it :beers: .

XuHuaiyu (approved):

PR #51652, opened 2024-03-11
txn: save RPCs for Pipelined-DML

### LINKED ISSUE #51625: reduce pipelined DML RPC count in some case to optimize performance
## Enhancement

#50389 introduce pipelined DML, we can enhancement the performance in some situation. See following test:

```go
func TestPipelinedDMLInsertRPC(t *testing.T) {
	tk.MustExec("use test")
	tk.MustExec("drop table if exists t1")
	tk.MustExec("create table t1 (a int, b int, unique index idx(b))")
	res := tk.MustQuery("explain analyze insert ignore into t1 values (1,1), (2,2), (3,3), (4,4), (5,5)")
	explain := getExplainResult(res)
	require.Regexp(t, "Insert.* check_insert: {total_time: .* rpc:{BatchGet:{num_rpc:1, total_time:.*}}}.*", explain)
	tk.MustExec("set session tidb_dml_type = bulk")
	res = tk.MustQuery("explain analyze insert ignore into t1 values (1,1), (2,2), (3,3), (4,4), (5,5)")
	explain = getExplainResult(res)
	// TODO: try to optimize the rpc count, when use bulk dml, insert ignore will use 5 BufferBatchGet and  1 BatchGet rpcs.
	// but without bulk dml, it will only use 1 BatchGet rpcs.
	require.Regexp(t, "Insert.* check_insert: {total_time: .* rpc:{.*BufferBatchGet:{num_rpc:5, total_time:.*}}}.*", explain)
	require.Regexp(t, "Insert.* check_insert: {total_time: .* rpc:{.*BatchGet:{num_rpc:1, total_time:.*}}}.*", explain)
	tk.MustExec("delete from t1")
}

func getExplainResult(res *testkit.Result) string {
	resBuff := bytes.NewBufferString("")
	for _, row := range res.Rows() {
		_, _ = fmt.Fprintf(resBuff, "%s\t", row)
	}
	return resBuff.String()
}
```

ekexium:

Issue Number: close #51625

Problem Summary:

For pipelined DML, `membuffer.Get()` may involve a RPC which is too expensive.

I searched all places in TiDB where this method is used:
- `batch_point_get.go`: this is used for cached table, which is incompatible by design, ignore it.
- `point_get.go`: a pipelined-dml is not designed for point-get, do not optimize it now.
- `unionstore_driver.go`: pass-through methods, ignore it.
- `index.go`
  - `index.Create`: **for untouched flag**. We disable the untouched optimization temporarily.
  - `index.Create`: **Read value**. We enforce that P-DML must lazily do existence check, and set `PresumeNotExist` flag if local memory doesn't contain the value.
  - `index.Delete` for partition id. This is not a common case. We do not optimize it in this PR.
- `tables.go`: **AddRecord**. We enforce that P-DML must lazily do existence check, and set `PresumeNotExist` flag if local memory doesn't contain the value.

This PR handles 
1. The **bold** items listed above. It solves the performance problem for the basic `INSERT` statement. 
2. For `REPLACE`, `INSERT IGNORE` and `INSERT ON DUPLICATE KEY UPDATE` statements, we do BatchGet to prefetch the keys just like the snapshot cache, so that later `membuffer.Get()` can use the cache to reduce RPCs.

Depends on https://github.com/tikv/client-go/pull/1212

### Check List

Tests 

- [x] Unit test: we set `tidb_dml_type=bulk` for all TiDB UT and checked all failures.
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/session/session.go
```diff
@@ -4307,6 +4307,45 @@ func (s *session) isPipelinedDML() bool {
 	if s.isInternal() {
 		return false
 	}
+	if !s.GetSessionVars().LazyCheckKeyNotExists() {
+		// we enforce that pipelined DML must lazily check key.
+		return false
```
you06, line 4312: The implementation seems to be always lazy check in p-dml.

pkg/session/session.go
```diff
@@ -4307,6 +4307,45 @@ func (s *session) isPipelinedDML() bool {
...
+	}
+	
+	{
```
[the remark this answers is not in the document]
ekexium: The default value is ON so it would simply reject nearly all cases.

pkg/executor/batch_checker.go
```diff
@@ -266,7 +266,15 @@ func formatDataForDupError(data []types.Datum) (string, error) {
 // t could be a normal table or a partition, but it must not be a PartitionedTable.
 func getOldRow(ctx context.Context, sctx sessionctx.Context, txn kv.Transaction, t table.Table, handle kv.Handle,
 	genExprs []expression.Expression) ([]types.Datum, error) {
-	oldValue, err := txn.Get(ctx, tablecodec.EncodeRecordKey(t.RecordPrefix(), handle))
+	var (
+		oldValue []byte
+		err      error
+	)
+	if txn.IsPipelined() {
+		oldValue, err = txn.GetFromPrefetchCache(ctx, tablecodec.EncodeRecordKey(t.RecordPrefix(), handle))
```
ekexium, line 274: ~~Should it be like this?~~
~~1. `GetLocal`~~
~~2. If not found, `GetFromPrefetchCache`~~

Okay I see this is implemented inside `GetFromPrefetchCache`. It's just a ambiguous name.

pkg/executor/insert.go
```diff
@@ -149,6 +149,9 @@ func prefetchUniqueIndices(ctx context.Context, txn kv.Transaction, rows []toBeC
 			batchKeys = append(batchKeys, k.newKey)
 		}
 	}
+	if txn.IsPipelined() {
+		return txn.Prefetch(ctx, batchKeys)
+	}
 	return txn.BatchGet(ctx, batchKeys)
```
ekexium, line 155: Is it necessary to do this `txn.BatchGet` for a pipelined txn?

pkg/table/tables/index.go
```diff
@@ -546,7 +552,12 @@ func FetchDuplicatedHandle(ctx context.Context, key kv.Key, distinct bool,
 		return fetchDuplicatedHandleForTempIndexKey(ctx, key, distinct, txn, tableID, isCommon)
 	}
 	// The index key is not from temp index.
-	val, err := getKeyInTxn(ctx, txn, key)
+	var val []byte
+	if txn.IsPipelined() {
+		val, err = getKeyInPrefetchCache(ctx, txn, key)
```
ekexium, line 557: Are we sure this key is prefetched? If so we can add an assertion that it can be found from the cache. If not we should fallback to a txn.Get?

pkg/store/driver/txn/unionstore_driver.go
```diff
@@ -157,6 +157,16 @@ func (m *memBuffer) MayFlush() error {
...
 }
 
+// GetLocal implements kv.MemBuffer interface
+func (m *memBuffer) GetLocal(ctx context.Context, key []byte) ([]byte, error) {
+	data, err := m.MemBuffer.GetLocal(ctx, key)
+	return data, derr.ToTiDBErr(err)
+}
+
+func (m *memBuffer) BatchGet(ctx context.Context, keys [][]byte) (map[string][]byte, error) {
+	return m.MemBuffer.BatchGet(ctx, keys)
```
ekexium, line 167: Does it need `derr.ToTiDBErr(err)`? Everything is an error. It's so obscure.

you06: Yes, added.

pkg/executor/insert.go
you06: Union the `Prefetch` with `txn.BatchGet` now.

pkg/executor/insert.go
```diff
@@ -88,6 +88,9 @@ func (e *InsertExec) exec(ctx context.Context, rows [][]types.Datum) error {
 	// If `ON DUPLICATE KEY UPDATE` is specified, and no `IGNORE` keyword,
 	// the to-be-insert rows will be check on duplicate keys and update to the new rows.
 	if len(e.OnDuplicate) > 0 {
+		if ignoreErr && txn.IsPipelined() {
+			return errors.New("cannot use pipelined mode with insert ignore and on duplicate update")
```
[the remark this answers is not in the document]
crazycs520: Oh, I see the test `TestInsertIgnoreOnDuplicateKeyUpdate`.

pkg/executor/delete.go
```diff
@@ -55,10 +55,19 @@ type DeleteExec struct {
...
+		err = e.deleteMultiTablesByChunk(ctx)
+	} else {
+		err = e.deleteSingleTableByChunk(ctx)
 	}
-	return e.deleteSingleTableByChunk(ctx)
+	if err != nil {
+		return err
+	}
+	if txn, _ := e.Ctx().Txn(false); txn != nil {
+		return txn.GetMemBuffer().MayFlush()
```
crazycs520, line 68: Too late to call `MayFlush` here? I think you may need move `MayFlush` into `deleteSingleTableByChunk` and `deleteMultiTablesByChunk`.

you06: Good catch.

cfzjywxk (approved):

GMHDBJD (approved):
LGTM

PR #51657, opened 2024-03-11
expression: force scalar expression to implement `RequireOptionalEvalProps` explicitly

### LINKED ISSUE #51656: force scalar expression to implement `RequireOptionalEvalProps` explicitly

lcwangchao:

Issue Number: close #51656

This PR removes default implement for `RequireOptionalEvalProps` in `baseBuiltinFunc` that forcing all scalar functions to have their own `RequireOptionalEvalProps` defined to tell the framework which optional properties they need. This is better than the previous one because it reduces the chances that the coder forget to declar optional property requirements when adding a new function.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/expression/builtin.go
+++ b/pkg/expression/builtin.go
@@ -46,6 +46,16 @@ import (
 	"github.com/pingcap/tipb/go-tipb"
 )
 
+var _ contextopt.RequireOptionalEvalProps = notRequireOptionalEvalProps{}
+
+// notRequireOptionalEvalProps indicates the function does not require any optional properties.
+type notRequireOptionalEvalProps struct{}
+
+// RequiredOptionalEvalProps implements the RequireOptionalEvalProps interface.
+func (notRequireOptionalEvalProps) RequiredOptionalEvalProps() (set OptionalEvalPropKeySet) {
+	return
+}
+
 // baseBuiltinFunc will be contained in every struct that implement builtinFunc interface.
 type baseBuiltinFunc struct {
 	bufAllocator columnBufferAllocator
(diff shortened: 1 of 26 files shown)


pkg/expression/builtin.go
```diff
@@ -46,6 +46,16 @@ import (
 	"github.com/pingcap/tipb/go-tipb"
 )
 
+var _ contextopt.RequireOptionalEvalProps = notRequireOptionalEvalProps{}
+
+// notRequireOptionalEvalProps indicates the function does not require any optional properties.
+type notRequireOptionalEvalProps struct{}
+
+// RequiredOptionalEvalProps implements the RequireOptionalEvalProps interface.
+func (notRequireOptionalEvalProps) RequiredOptionalEvalProps() (set OptionalEvalPropKeySet) {
```
YangKeao, line 55: What about a type cast like this

```go
type XXX interface {
    RequiredOptionalEvalProps() (set OptionalEvalPropKeySet
}
if s, ok := sig.(); ok {
    ...
}
```

If a sig didn't implement `RequiredOptionalEvalProps()`, we can assume that it does not need any `OptionalEvalPropKeySet` :thinking: . It leaves the complexity from the implementation to the caller. In my imagination, the count of callers will be much smaller than the implementation, so it'll lead to less modification. 

(But actually, this opinion is not very strong).

PR #51667, opened 2024-03-11
txn: Add test case to cover the case that check_txn_status on primary lock of primary txn may affect transaction correctness

### LINKED ISSUE #51666: Add test to cover the issue that check_txn_status being called on optimistic transaction may lead to transaction consistency issue
The bug itself is tracked in: https://github.com/tikv/tikv/issues/16620

Consider the test case added in #51667 . Without the fix to tikv, consider the checks at last:

https://github.com/pingcap/tidb/blob/163bbb08cddc5e72edabcc8c6de8e08d0f937d74/tests/realtikvtest/txntest/txn_test.go#L529-L533

* No error will be received at line 529, where there is supposed to be.
* Check fail at line 531:
    ```
                Error:          Not equal:
                                expected: "[1 11]\n[2 21]\n"
                                actual  : "[1 111]\n[2 21]\n"
    ```
    Which means that atomicity of transaction `tk1` (adds 100 to both row 1 and row 2) is broken.
* Data inconsistency will be found at line 532:
    ```
                Error:          Received unexpected error:
                                [admin:8223]data inconsistency in table: t2, index: v, handle: 1, index-values:"" != record-values:"handle: 1, values: [KindInt64 11]"
    ```
    Which is because `tk1`'s write to table t2 (which should be a pair of row and index) is partially committed.

MyonKeminta:

Issue Number: close #51666

Problem Summary:

This PR adds a test to ensure that #51666 (https://github.com/tikv/tikv/issues/16620) is fixed.

Requires:
* https://github.com/tikv/client-go/pull/1211
* https://github.com/tikv/tikv/pull/16621

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


tests/realtikvtest/txntest/txn_test.go
```diff
@@ -342,3 +350,185 @@ func TestTxnEntrySizeLimit(t *testing.T) {
...
+
+		assert.NoError(t, failpoint.Disable("tikvclient/beforePrewrite"))
+		select {
+		case <-tkPrepapre1Ch:
+		case <-time.After(time.Second):
+			assert.Fail(t, "tkPrepare1 not resumed after unsetting failpoint")
+		}
+
+		var err error
+		//tk1.MustQuery("select json_extract(@@tidb_last_txn_info, '$.txn_commit_mode')").Check(testkit.Rows("async_commit"))
```
cfzjywxk, line 428: Need to be deleted?

tests/realtikvtest/txntest/txn_test.go
```diff
@@ -342,3 +350,185 @@ func TestTxnEntrySizeLimit(t *testing.T) {
...
+		lastCommitTS, err = strconv.ParseUint(tkPrepare1.MustQuery("select json_extract(@@tidb_last_txn_info, '$.commit_ts')").Rows()[0][0].(string), 10, 64)
+		assert.NoError(t, err)
+		currentStartTS, err := strconv.ParseUint(tk1.MustQuery("select @@tidb_current_ts").Rows()[0][0].(string), 10, 64)
+		assert.NoError(t, err)
+		if currentStartTS == lastCommitTS {
```
cfzjywxk, line 433: Is it possible to use a failpoint to inject specific ts instead of using a `for` loop?

MyonKeminta: Maybe, but I'm afraid it's hard to claim that our simulation is correct and our simulated transactions can exist in real cases. And IIRC it's not the first time I use this way to simulate ts collision in tests.

cfzjywxk: The case is quite complex, a little bit concern its stability. 
Perhaps it is indeed difficult to encounter in real scenes, which is why it is so difficult to simulate..

MyonKeminta: Yes. Actually I'm kind of doubting about how well the ts-collision cases of async commit/1PC are covered by our random tests... but making these cases right is crucial to the correctness

cfzjywxk (approved):

crazycs520 (approved):

MyonKeminta:
This is a test for covering the bugfix of https://github.com/tikv/tikv/issues/16620 , and we decided not to cherry-pick this PR to release branches. Why is the bot triggered by the PR which is not marked as a bugfix and the corresponding issue of which is not marked as a bug?

PR #43479, opened 2023-04-27
*: add tidb_row_checksum() as a builtin function

zyguan:
### What problem does this PR solve?

Issue Number: ref #42747

Problem Summary: as decribed in #42747, we need to implement a builtin function `tidb_row_checksum` to export row checksums to users.

### What is changed and how it works?

This is a quick & dirty implementation of `tidb_row_checksum`, which can only be used as a select field in a fast point plan. `tidb_row_checksum()` will be rewritten as a dummy column when `buildSchemaFromFields` in `TryFastPlan` and `ChunkDecoder` will fill this dummy column with row checksum info.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #51678, opened 2024-03-12
expression: support tidb encode key function

### LINKED ISSUE #51683: Support querying MVCC information through builtin function
## Enhancement

MVCC information is useful for debugging data inconsistenct issue. For now, we have HTTP API `mvcc/key/...` to query MVCC. However, it is not always convenient to construct a HTTP query. For example, when there are complex data types in index columns or clustered primary key columns like `timestamp`, `datetime` or even `binary`, query parameters cannot represent these values.

Maybe we can provide a builtin function like `tidb_mvcc_info()` achieve this purpose.

Note: the priviege should be considered to make sure there are no security issues.

tangenta:

Issue Number: close #51683

Problem Summary: See #51683.

This PR supports three builtin functions, namely `tidb_mvcc_info()`, `tidb_encode_record_key()` and `tidb_encode_index_key()`:

- `tidb_mvcc_info(key)` returns MVCC information of a given key.
- `tidb_encode_record_key(dbName, tableName, handle/pk columns...)` returns a row record key. The table must exist.
- `tidb_encode_index_key(dbName, tableName, indexName, idx columns..., handle/pk columns...)` returns a index record key. The index must exist.

Please also note that *tableName* can be one of formats as below:
- `t`: refer to non-partitioned table `t`.
- `t(p1)`: refer to partition table `t` and the partition `p1`.

```sql
mysql> create table t (a int, b int, c timestamp, primary key (a) clustered, index idx_c(c));
Query OK, 0 rows affected (0.14 sec)

mysql> insert into t values (1, 1, now());

mysql> table t;
+---+------+---------------------+
| a | b    | c                   |
+---+------+---------------------+
| 1 |    1 | 2024-03-12 11:14:35 |
+---+------+---------------------+
1 row in set (0.00 sec)

mysql> select tidb_encode_record_key('test', 't', 1);
+----------------------------------------+
| tidb_encode_record_key('test', 't', 1) |
+----------------------------------------+
| 7480000000000000705f728000000000000001 |
+----------------------------------------+
1 row in set (0.00 sec)

mysql> select tidb_mvcc_info('7480000000000000705f728000000000000001');
+-----------------------------------------------------------------------------------------------------------------------------------+
| tidb_mvcc_info('7480000000000000705f728000000000000001')                                                                          |
+-----------------------------------------------------------------------------------------------------------------------------------+
| {"info":{"writes":[{"start_ts":448322148825038849,"commit_ts":448322148825038850,"short_value":"gAACAAAAAgMBAAkAAQAAAKMz2LIZ"}]}} |
+-----------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> select tidb_encode_index_key('test', 't', 'idx_c', '2024-03-12 11:14:35', 1);
+----------------------------------------------------------------------------+
| tidb_encode_index_key('test', 't', 'idx_c', '2024-03-12 11:14:35', 1)      |
+----------------------------------------------------------------------------+
| 7480000000000000705f6980000000000000010419b2d833a3000000038000000000000001 |
+----------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> select tidb_decode_key(tidb_encode_index_key('test', 't', 'idx_c', '2024-03-12 11:14:35', 1));
+----------------------------------------------------------------------------------------+
| tidb_decode_key(tidb_encode_index_key('test', 't', 'idx_c', '2024-03-12 11:14:35', 1)) |
+----------------------------------------------------------------------------------------+
| {"index_id":1,"index_vals":{"c":"2024-03-12 11:14:35"},"table_id":112}                 |
+----------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> select tidb_mvcc_info(tidb_encode_index_key('test', 't', 'idx_c', '2024-03-12 11:14:35', 1));
+-----------------------------------------------------------------------------------------------------------+
| tidb_mvcc_info(tidb_encode_index_key('test', 't', 'idx_c', '2024-03-12 11:14:35', 1))                     |
+-----------------------------------------------------------------------------------------------------------+
| {"info":{"writes":[{"start_ts":448322148825038849,"commit_ts":448322148825038850,"short_value":"MA=="}]}} |
+-----------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


ekexium:
I love this
We need this
Please do continue this

D3Hunter:
i hope those functions would works in normal query like :
```
select tidb_handle(), tidb_idx_key(index-name) from test.t where id = 1
select tidb_mvcc(tidb_handle()) from test.t where id = 1
```

tangenta:
@D3Hunter If we put this in normal query, I think we cannot get the MVCC of deleted rows/indexes?

kennytm:
@D3Hunter I don't think a function should need or be able to know the current plan is a SELECT operating on exactly one physical table.

kennytm (changes requested):
This must not be merged before the privilege check bypass is fixed. 

See https://github.com/pingcap/tidb/issues/51683#issuecomment-2168137763 for a PoC.

D3Hunter:
> i hope those functions would works in normal query like :
> 
> ```
> select tidb_handle(), tidb_idx_key(index-name) from test.t where id = 1
> select tidb_mvcc(tidb_handle()) from test.t where id = 1
> ```

ok, for downvoters, get from query is much simplier, i don't like to input so many parameters to just get the key of it, and it's error-prone to do that as we have to remember the definition of the pk/index

> If we put this in normal query, I think we cannot get the MVCC of deleted rows/indexes?

can use `as of timestamp`?

> I don't think a function should need or be able to know the current plan is a SELECT operating on exactly one physical table.

don't get your point, where is `a function should need or be able to know the current plan is a SELECT operating on exactly one physical table.` rule came from?

kennytm:
@D3Hunter what is the result of `tidb_handle()` in the following?

```sql
select tidb_handle() from a, b;
select tidb_handle() from dual;
select tidb_handle() from (select * from t) _;

create view v as select 1;
select tidb_handle() from v;
```

D3Hunter:
> @D3Hunter what is the result of `tidb_handle()` in the following?
> 
> ```sql
> select tidb_handle() from a, b;
> select tidb_handle() from dual;
> select tidb_handle() from (select * from t) _;
> 
> create view v as select 1;
> select tidb_handle() from v;
> ```

if we use the pattern, just check and report error in this case

kennytm:
> if we use the pattern, just check and report error in this case

No SQL flavor (MySQL or Postgres or T-SQL etc) would determine the validity of a function by the chosen query plan.

On TiDB there is a single exception of `tidb_row_checksum()` (#43479) which the PR description claimed this **is a quick & dirty implementation**, and I absolutely oppose making this the future convention. A pseudo-column that attaches to the table object i.e. `t._tidb_rowid` makes much more sense than the plan-dependent `tidb_row_checksum()` model (even if both are hacks).

D3Hunter:
> No SQL flavor (MySQL or Postgres or T-SQL etc) would determine the validity of a function by the chosen query plan.

unfair to compare with them, they don't have the requirements to access underlying encoded DATA(KV) too

> absolutely oppose

please give insights, now just your standpoint

kennytm:
@D3Hunter

> unfair to compare with them, they don't have the requirements to access underlying encoded DATA(KV) too

T-SQL has informational functions like [OBJECT_ID](https://learn.microsoft.com/en-us/sql/t-sql/functions/object-id-transact-sql?view=sql-server-ver16) to obtain the low-level information and using similar semantic as the current proposal --- identify the table via a string

```sql
select object_id('db.schema.tbl');
```

Similar for [PostgreSQL](https://www.postgresql.org/docs/current/functions-info.html) which the functions either return something that can be uniquely identified (`current_schema()`, `pg_current_xact_id()`, ...) or requires user to supply the identity (` col_description(table_oid, col_number), ...`). 

I can't find any function that has implicit assumption about which table the SELECT is working on.

> please give insights, now just your standpoint

Keep the current syntax unchanged `tidb_encode_record_key('schema', 'tbl', pk_val)` and `tidb_encode_index_key('schema', 'tbl', 'idx', idx_vals, pk_val)`.

(One could introduce a pseudo-column `t._tidb_record_key` but it doesn't work for index keys nicely.)

kennytm:
I think OracleDB's [Pseudocolumns](https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/Pseudocolumns.html) are pretty close in spirit. But I don't think there's much mainstream DB products that exposes low-level row encoding via SQL.

Another alternative is a table function like

```sql
select tbl.a, tbl.c, mvcc.record, mvcc.index_idx_c 
from tbl 
join tidb_mvcc_info(tbl) mvcc on true 
where tbl.a = 1;
```

but probably don't bother when we don't even support the MySQL-compatible `JSON_TABLE()` table function :shrug: 

Also both pseudo-column and table function can't give us MVCC info for ***deleted*** rows that are not yet GC'ed which sometimes we may want to access.

BTW shouldn't these syntax questions be better commented on #51683 rather than this PR?

tangenta:
@kennytm I added the privilege check in [this commit](https://github.com/pingcap/tidb/pull/51678/commits/03a9dc0df6b82269b1d1d5766a632cafccb8c271), PTAL

pkg/planner/core/expression_rewriter.go
```diff
@@ -2640,6 +2641,185 @@ func hasCurrentDatetimeDefault(col *model.ColumnInfo) bool {
...
+	if err != nil || isNull {
+		return nil, isNull, err
+	}
+	tblName, isNull, err := args[1].EvalString(ctx, row)
+	if err != nil || isNull {
+		return nil, isNull, err
+	}
+	is := isVer.(infoschema.InfoSchema)
+	if is == nil {
+		return nil, false, errors.New("missing information schema")
```
kennytm, line 2660:
```suggestion
	is, ok := isVer.(infoschema.InfoSchema)
	if !ok {
		return nil, false, errors.New("missing information schema")
```

the cast will crash if `isVer` is actually nil 

```
panic: interface conversion: interface is nil, not infoschema.InfoSchema
```

pkg/planner/core/expression_rewriter.go
```diff
@@ -2640,6 +2641,185 @@ func hasCurrentDatetimeDefault(col *model.ColumnInfo) bool {
...
+	}
+	tbl, err := is.TableByName(context.Background(), model.NewCIStr(dbName), model.NewCIStr(tblName))
+	if err != nil {
+		return nil, false, err
+	}
+	tblInfo := tbl.Meta()
```
kennytm, line 2779: is `tbl` actually needed here, or is `TableInfoByName` sufficient (TableInfoByName requires only MetaOnlyInfoSchema).

kennytm (approved):

pkg/planner/core/expression_rewriter.go
tangenta: Yes, we need `table.Table` to choose the correct partition. Previous implementation does not consider partition. This bug is fixed in https://github.com/pingcap/tidb/pull/51678/commits/4e07b96d212ad91e4ca8e1d272c9147a72c8b18b.

pkg/planner/core/expression_rewriter.go
tangenta: It cannot be nil, I removed the nil-check.

D3Hunter:
please ask PM for approval for new feature

D3Hunter:
/hold

unhold it after PM approve

Frank945946 (approved):

wjhuang2016 (approved):

D3Hunter (commented):
~add some test for user privileges?~

XuHuaiyu (approved):

time-and-fate (approved):

PR #51701, opened 2024-03-12
ttl: fix expire time drift for daylighting-save time case

### LINKED ISSUE #51675: ttl: 1 hour's deviation for 'expire time' in TTL
## Bug Report

1. Set time zone as 'PDT': `set @@time_zone='America/Los_Angeles'` on PDT time.
2. Run test 'TestSubmitJob':
```
go test --tags=intest -timeout 30s -run ^TestSubmitJob$ "github.com/pingcap/tidb/pkg/ttl/ttlworker"
```

Test succeeded.

Test failed:
```
--- FAIL: TestSubmitJob (0.96s)
    job_manager_integration_test.go:477:
        	Error Trace:	/Users/bb7133/Projects/gopath/src/github.com/pingcap/tidb/pkg/ttl/ttlworker/job_manager_integration_test.go:477
        	Error:      	Max difference between 1710172125 and 1710175725 allowed is 300, but difference was -3600
        	Test:       	TestSubmitJob
FAIL
```

Latest master branch

### 5. Root cause
The [calculation of 'expire time'](https://github.com/pingcap/tidb/blob/e2718640b58e3cef5eb2be1275ad59fdab36ea61/pkg/ttl/cache/table.go#L199-L201) is problematic when facing daylight saving time:
```
fmt.Sprintf("SELECT FROM_UNIXTIME(0) + INTERVAL %d SECOND - INTERVAL %s %s", now.Unix(), expireExpr, unit.String()),
```
For areas like Los Angeles, the time of `FROM_UNIXTIME(0)` is in PST when `now()` is in PDT, there will be 1-hour deviation.

As a result, users will also notice this bug checking job status in PDT time:

```
tidb> create table ttlp1(a int, t timestamp) TTL=`t`+interval 1 HOUR PARTITION BY RANGE (a) (
    -> PARTITION p0 VALUES LESS THAN (10),
    -> PARTITION p1 VALUES LESS THAN (100));

tidb> select * from mysql.tidb_ttl_table_status\G
...
*************************** 3. row ***************************
                      table_id: 110
               parent_table_id: 109
              table_statistics: NULL
                   last_job_id: bfbe0a4394ad4a50852ef6da315a9640
           last_job_start_time: 2024-03-11 14:20:39
          last_job_finish_time: 2024-03-11 14:20:54
           **last_job_ttl_expire**: 2024-03-11 12:20:39
...
```

lcwangchao:

Issue Number: close #51675

To make sure we have a right behavior with different timezones, this PR has some modifications:

- `EvalExpireTime` uses go time lib to compute the expired time instead of executing SQL to avoid some problems like: https://bugs.mysql.com/bug.php?id=95219
- `EvalExpireTime` must compute the expired time according to the TiDB global time zone.
- Considering some SQLs are using string format to update timestamp columns. We must ensure that the time argument with type `time.Time` has a location that is the same as the session's location. So I added some asserts.
- We force the session's time zone to 'UTC' for TTL meta operations to make sure we can insert a right time even in daylight saving time.
- We should also considerate the timezone for TTL job window. In this PR, we also use the global time zone in TiDB instead of the local time zone.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
fix expire time drift for daylighting-save time case
```

### DIFF
--- a/pkg/ttl/cache/table.go
+++ b/pkg/ttl/cache/table.go
@@ -208,7 +208,11 @@ func (t *PhysicalTable) EvalExpireTime(ctx context.Context, se session.Session,
 		return time.Time{}, err
 	}
 
-	expire = now.In(globalTz).
+	expire = now.
+		In(globalTz).
+		// Truncate to second to make sure the precision is always the same with the one stored in a table to avoid some
+		// comparing problems in testing.
+		Truncate(time.Second).
 		AddDate(-int(y), -int(m), -int(d)).
 		Add(-time.Duration(nanosecond)).
 		In(now.Location())
--- a/pkg/ttl/cache/table.go
+++ b/pkg/ttl/cache/table.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"encoding/binary"
 	"encoding/hex"
-	"fmt"
 	"math"
 	"time"
 
@@ -32,7 +31,6 @@ import (
 	"github.com/pingcap/tidb/pkg/tablecodec"
 	"github.com/pingcap/tidb/pkg/ttl/session"
 	"github.com/pingcap/tidb/pkg/types"
-	"github.com/pingcap/tidb/pkg/util/chunk"
 	"github.com/pingcap/tidb/pkg/util/codec"
 	"github.com/pingcap/tidb/pkg/util/collate"
 	"github.com/pingcap/tidb/pkg/util/intest"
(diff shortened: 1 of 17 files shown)


YangKeao (approved):
LGTM

pkg/ttl/cache/table.go
```diff
@@ -189,24 +189,61 @@ func (t *PhysicalTable) ValidateKeyPrefix(key []types.Datum) error {
 func (t *PhysicalTable) EvalExpireTime(ctx context.Context, se session.Session,
 	now time.Time) (expire time.Time, err error) {
 	tz := se.GetSessionVars().Location()
+	now = now.In(tz)
 
 	expireExpr := t.TTLInfo.IntervalExprStr
 	unit := ast.TimeUnitType(t.TTLInfo.IntervalTimeUnit)
 
+	if now.Unix() < 0 {
+		return time.Time{}, errors.Errorf("current time must after UTC time 1970-01-01 00:00:00, but %s", now)
```
bb7133, line 198:
```suggestion
		return time.Time{}, errors.Errorf("Current time must be after the UTC time January 1, 1970, 00:00:00. Received time: %s.", now)
```

pkg/ttl/cache/table.go
```diff
@@ -189,24 +189,61 @@ func (t *PhysicalTable) ValidateKeyPrefix(key []types.Datum) error {
...
+		// this time does not "exist" in timezone `America/Los_Angeles` because after "2024-03-10 02:00:00",
+		// the time will leap to "2024-03-10 03:00:00" directly.
+		// If then we parse "2024-03-10 02:30:00" with `time.ParseInLocation` we will get a Go time "2024-03-10 01:30:00".
+		// So, return the parsed time directly.
+		return expire, nil
+	}
+
+	_, offsetNow := now.Zone()
+	_, offsetExpire := expire.Zone()
+	if interval := offsetExpire - offsetNow; interval < 0 {
```
bb7133, line 234: What if we have interval > 0(turn 1 hour backward when daylight saving time ends)? Could you also add a test case for it?

lcwangchao: When interval > 0, its safe for daylight saving case because the computed expire time is eariler. I've added a test case

pkg/ttl/cache/table.go
```diff
@@ -189,24 +189,64 @@ func (t *PhysicalTable) ValidateKeyPrefix(key []types.Datum) error {
 func (t *PhysicalTable) EvalExpireTime(ctx context.Context, se session.Session,
 	now time.Time) (expire time.Time, err error) {
 	tz := se.GetSessionVars().Location()
+	now = now.In(tz)
```
mjonss, line 192: Would it be possible to only work with either time.Time or Unix epochs? as soon as one goes to strings or time representations with time zones it will add a lot of complexity and mental overhead.

pkg/ttl/cache/table.go
```diff
@@ -189,24 +189,64 @@ func (t *PhysicalTable) ValidateKeyPrefix(key []types.Datum) error {
...
+
+	if expire.Hour() != tm.Hour() {
+		// For the daylight-saving time, the hour may be different, and it means the tm we get is not a "valid" date.
+		// For example, if we computed an expired `types.Time` with value "2024-03-10 02:30:00",
+		// this time does not "exist" in timezone `America/Los_Angeles` because after "2024-03-10 02:00:00",
+		// the time will leap to "2024-03-10 03:00:00" directly.
+		// If then we parse "2024-03-10 02:30:00" with `time.ParseInLocation` we will get a Go time "2024-03-10 01:30:00".
+		// So, return the parsed time directly.
+		return expire, nil
+	}
```
mjonss, line 233: There is also a weird time zone with 30 min offset for Day Light Saving Time: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#LORD_HOWE

For all the different time zones:
https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

mjonss (commented):
I think there should be a function to just get the interval in seconds from the `t.TTLInfo.IntervalExprStr` and `t.TTLInfo.IntervalTimeUnit`. Then just adjust a time.Time internally, instead of going over SQL, including various internal time zone adjustments.

lcwangchao:
> I think there should be a function to just get the interval in seconds from the `t.TTLInfo.IntervalExprStr` and `t.TTLInfo.IntervalTimeUnit`. Then just adjust a time.Time internally, instead of going over SQL, including various internal time zone adjustments.

I've make a new implement of `EvalExpireTime` without executing SQL.

pkg/ttl/cache/table.go
```diff
@@ -185,28 +183,40 @@ func (t *PhysicalTable) ValidateKeyPrefix(key []types.Datum) error {
...
+		return time.Time{}, err
 	}
 
-	tm := rows[0].GetTime(0)
-	return tm.CoreTime().GoTime(tz)
+	// Use the global time zone to compute expire time.
+	// Different timezones may have different results event with the same "now" time and TTL expression.
+	// Consider a TTL setting with the expiration `INTERVAL 1 MONTH`.
+	// If the current timezone is `Asia/Shanghai` and now is `2021-03-01 00:00:00 +0800`
+	// the expired time should be `2022-02-01 00:00:00 +0800`, corresponding to UTC time `2022-01-31 16:00:00 UTC`.
```
mjonss, line 203:
```suggestion
	// the expired time should be `2021-02-01 00:00:00 +0800`, corresponding to UTC time `2021-01-31 16:00:00 UTC`.
```

pkg/ttl/cache/table.go
```diff
@@ -185,28 +183,40 @@ func (t *PhysicalTable) ValidateKeyPrefix(key []types.Datum) error {
...
+	// But if we use the `UTC` time zone, the current time is `2021-02-28 16:00:00 UTC`,
+	// and the expired time should be `2021-01-28 16:00:00 UTC` that is not the same the previous one.
```
mjonss, line 205: Good catch and well written comment!
Should we mention this in the manual as well, that the TTL will be executed in the global time zone?
I think the other way around with a global time zone west of GMT is more confusing, where '2021-02-28 16:00:00 -08:00' will delete up to '2021-02-01 00:00:00 UTC', if the global time zone is UTC?
Meaning that in the users local time zone, it will not have access to one month old data, i.e. the user is still in Februrary, but cannot access the last data in January.
Probably should be mentioned for any unit that includes 'Month' and 'Quarter'?

pkg/ttl/cache/table.go
```diff
@@ -185,28 +183,40 @@ func (t *PhysicalTable) ValidateKeyPrefix(key []types.Datum) error {
 	return nil
 }
 
-// EvalExpireTime returns the expired time
+// EvalExpireTime returns the expired time.
+// It uses the global timezone to compute the expired time.
+// Then we'll reset the returned expired time to the same timezone as the input `now`.
```
mjonss, line 188: In which case can this be executed in a different time zone than the global time zone?

mjonss (approved):
LGTM, just some curious questions left :)

mjonss:
I just added a 'hold', so you can merge it when you feel like you completed it, since it is now approved :)

pkg/ttl/cache/table.go
lcwangchao: For this PR, the session time zone will be always different with global time zone if the global time zone is not `UTC`. That is because we set the session time zone to `UTC` to avoid some problems. For example, the DST problem that may case a wrong inserted timestamp value.

pkg/ttl/cache/table.go
```diff
@@ -185,28 +183,40 @@ func (t *PhysicalTable) ValidateKeyPrefix(key []types.Datum) error {
...
+	globalTz, err := se.GlobalTimeZone(ctx)
+	if err != nil {
+		return time.Time{}, err
+	}
+
+	expire = now.
+		In(globalTz).
+		// Truncate to second to make sure the precision is always the same with the one stored in a table to avoid some
+		// comparing problems in testing.
+		Truncate(time.Second).
```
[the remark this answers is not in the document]
lcwangchao: I think it is also fine to make it truncated in the production environment. If no other reasons, I prefer to keep the same behavior between test and production

pkg/ttl/cache/table.go
lcwangchao: Thanks for your comment. It's better to mention it in the manual.

bb7133 (approved):
LGTM

Issue #51477, opened 2024-03-04
"optional" properties for `expression.EvalContext`


lcwangchao:
The `EvalContext` has many fields to provide but not all of them are required in every scene. So it's better to classify these fields into two categories:

1. **Basic properties** such as `type.Context`, `errctx.Context`, `SQLMode`, etc... These properties are used by functions widely. For example, `errctx.Context` used by most functions to handle some errors. The basic properties are also simple, they should be easy to clone and mock. The `EvalContext` MUST provide basic properties. 
2. **Optional properties**, as its name, is "optional". In some scenes such as lightning, it only requires some of the properties, so these properties are provided on demand. Optional properties are not widely used, for example, `tidb_is_ddl_owner` requires `EvalContext` to provide the information on whether the current tidb is DDL owner, but this function is forbidden in most other scenes except SQL query, like partition expressions or generated column expressions. Some optional properties are complex, so not all of them can be cloned or migrated.

PR #51725, opened 2024-03-13
expression: add more optional properties for `EvalContext`

lcwangchao:

Issue Number: ref #51477

Add more optional properties for `EvalContext`

- OptPropSessionVars: to provide `variable.SessionVars` .
- OptPropAdvisoryLock: to provide advisory lock operations.
- OptPropDDLOwnerInfo: to provide DDL owner information.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/expression/contextopt/sessionvars.go
```diff
@@ -0,0 +1,57 @@
...
+type SessionVarsPropReader struct{}
+
+// RequiredOptionalEvalProps implements the RequireOptionalEvalProps interface.
+func (SessionVarsPropReader) RequiredOptionalEvalProps() context.OptionalEvalPropKeySet {
+	return context.OptPropSessionVars.AsPropKeySet()
+}
+
+// GetSessionVars returns the session vars from the context
+func (SessionVarsPropReader) GetSessionVars(ctx context.EvalContext) (*variable.SessionVars, error) {
+	p, err := getPropProvider[*SessionVarsPropProvider](ctx, context.OptPropSessionVars)
```
zimulala, line 52: The key itself must exist, what is the point of adding this judgment? The main thing is that an extra error judgment feels that the code is not very elegant.

Defined2014: If this is an impossible situation, is it better to just panic instead of return an error?

Specifically, when will a return error occur?

lcwangchao: In the future, we may have multiple implementations for `EvalContext` and not all of them are required to contain all properties. The expression should fail when a property is not provided but required by an expression.

Usually, the error occurs when a coder constructs a `EvalContext` but missing to provide some required properties or using some unsupported expression in a specified scene. For example, only some simple expressions are allowed in a partition expression, but if someone uses function `GetVar`, it will fail.

Though in the above partition expression case, we should do some prior checks but we should also have a check in `GetSessionVars` to make the code strong. IMO, panic is not a good option because it will make tidb (at leat a session) crash.

zimulala: Is it to be changed in the future, and panic is handled at present?

YangKeao (approved):
LGTM

windtalker (approved):
LGTM

Issue #51778, opened 2024-03-14
Enable optimizer to generate index merge path for mv index + OR list nested in AND list


time-and-fate:
## Enhancement
Currently, tidb can't generate index merge access path for such case:
```sql
CREATE TABLE `t` (
  `a` int(11) DEFAULT NULL,
  `b` varchar(30) DEFAULT NULL,
  `c` float DEFAULT NULL,
  `j` json DEFAULT NULL,
  `pk` int(11) NOT NULL,
  PRIMARY KEY (`pk`) /*T![clustered_index] CLUSTERED */,
  KEY `mvi1` (`c`,(cast(json_extract(`j`, _utf8'$.a') as unsigned array)),`b`),
  KEY `mvi2` (`a`,(cast(json_extract(`j`, _utf8'$.c') as unsigned array))),
  KEY `mvi3` ((cast(json_extract(`j`, _utf8'$.d') as unsigned array)),`c`),
  KEY `idx` (`b`,`c`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;

explain format=brief select /*+ use_index_merge(t, mvi1) */ * from t where ( 1 member of (j->'$.a') or  2 member of (j->'$.a') ) and a = 20 and c = 10;
```
```
> explain format=brief select /*+ use_index_merge(t, mvi1) */ * from t where ( 1 member of (j->'$.a') or  2 member of (j->'$.a') ) and a = 20 and c = 10;
+-----------------------+----------+-----------+---------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| id                    | estRows  | task      | access object | operator info                                                                                                                                                                  |
+-----------------------+----------+-----------+---------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| TableReader           | 0.01     | root      |               | data:Selection                                                                                                                                                                 |
| └─Selection           | 0.01     | cop[tikv] |               | eq(test.t.a, 20), eq(test.t.c, 10), or(json_memberof(cast(1, json BINARY), json_extract(test.t.j, "$.a")), json_memberof(cast(2, json BINARY), json_extract(test.t.j, "$.a"))) |
|   └─TableFullScan     | 10000.00 | cop[tikv] | table:t       | keep order:false, stats:pseudo                                                                                                                                                 |
[...]

PR #51716, opened 2024-03-12
planner: support OR list nested in AND list for mv index | tidb-test=pr/2309

time-and-fate:
### What problem does this PR solve?

Issue Number: close  #51778

### What changed and how does it work?

Use new methods to replace the original implementation for DNF case in `generateIndexMergeOnDNF4MVIndex()` and `generateIndexMerge4ComposedIndex()`.
Please see comments for details.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #51780, opened 2024-03-14
planner: code prepare for support OR list nested in AND list for mv index

time-and-fate:

Issue Number: ref #51778

Problem Summary:

Some code preparation to make  #51716 smaller.

- Add test cases for comparison in the next PR. (`tests/integrationtest/r/planner/core/indexmerge_path.result`, `tests/integrationtest/t/planner/core/indexmerge_path.test`)
- Fix two test cases where the values in the SQL text use the wrong type. (`pkg/planner/core/indexmerge_path_test.go`)
- Change the passed parameters in several functions. The effect of the functions are not affected. (`pkg/planner/core/indexmerge_path.go`)
- Enhance several existing methods. (`pkg/planner/core/indexmerge_path.go`)
  - `PrepareCols4MVIndex() -> PrepareIdxColsAndUncoverArrayType()`: Add a parameter to allow this function not to check there is exactly 1 virtual column. In the next PR, we'll use this function for the non-MV index.
  - `jsonArrayExpr2Exprs()`: Add a parameter to allow skipping the check for skipping plan cache. This will be used if we are not building an access path and are only doing checks and collecting information. This will be used in the next function.
  - `checkFilter4MVIndexColumn()`
    - Before, we only do simple checks on the expression to see if it can be an access filter. Now, we'll also extract values from the expression and do similar checks as in `buildPartialPaths4MVIndex()`. Using these values, we also return the "type" of the access filter.
  - `collectFilters4MVIndex()`: Based on changes in `checkFilter4MVIndexColumn()`, provide a "type" for the access filters in the return value.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/indexmerge_path.go
```diff
@@ -1590,57 +1637,130 @@ func indexMergeContainSpecificIndex(path *util.AccessPath, indexSet map[int64]st
...
-			if con, isCon := args[0].(*expression.Constant); isCon {
-				argCol, argConst = c, con
+		for _, v := range virColVals {
+			if !isSafeTypeConversion4MVIndexRange(v.GetType(), idxCol.GetType()) {
+				return false, unspecifiedFilterTp
 			}
 		}
-		if argCol == nil || argConst == nil {
-			return false
+		if (tp == multiValuesOROnMVColTp || tp == multiValuesANDOnMVColTp) && len(virColVals) == 1 {
```
qw4990, line 1724: Please add some comments here.

qw4990 (approved):

pkg/planner/core/indexmerge_path.go
```diff
@@ -1590,57 +1643,132 @@ func indexMergeContainSpecificIndex(path *util.AccessPath, indexSet map[int64]st
 	return false
 }
 
+const (
+	unspecifiedFilterTp int = iota
+	eqOnNonMVColTp
+	multiValuesOROnMVColTp
+	multiValuesANDOnMVColTp
+	singleValueOnMVColTp
```
AilinKid, line 1651: you can add more comments on the situation they stands

pkg/planner/core/indexmerge_path.go
AilinKid: +1

pkg/planner/core/indexmerge_path.go
```diff
@@ -1428,18 +1473,26 @@ func PrepareCols4MVIndex(
...
 			if usedAsAccess[i] {
 				continue
 			}
-			if checkFilter4MVIndexColumn(sctx, f, col) {
+			if ok, tp := checkFilter4MVIndexColumn(sctx, f, col); ok {
 				accessFilters = append(accessFilters, f)
 				usedAsAccess[i] = true
 				found = true
+				// access filter type on mv col overrides normal col for the return value of this function
+				if accessTp == unspecifiedFilterTp || accessTp == eqOnNonMVColTp {
```
AilinKid, line 1493: why use accessTp to do the comparson？
it isn't be assigned with any value before？

AilinKid (approved):

PR #51795, opened 2024-03-14
topsql: set RPC interceptor immediately after a txn is activated

### LINKED ISSUE #52845: Flush requests are not tracked by TopSQL
## Bug Report

ref #50215 

Run pipelined DMLs

TopSQL can track the resource usage of Flush commands.

TopSQL cannot track the resource usage of Flush commands.

master(09313090225382b3756f54438ea7bc53ae6bd937)

ekexium:

Issue Number: close #52845

Problem Summary:

This is to fix the resource group tag problem found in testing of pipelined-dml. The tagger was nil for Flush requests.

Set resource group tagger immediately after a txn is activated so that the following Flush requests can use it.

### Check List

Tests 

- [x] Unit test: existing topsql tests are not broken.
- [ ] Integration test
   Enable P-DML for all UTs, those that failed because of lack of resource group tag now pass.
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/sessiontxn/isolation/base.go
```diff
@@ -284,6 +284,8 @@ func (p *baseTxnContextProvider) ActivateTxn() (kv.Transaction, error) {
 		return nil, err
 	}
 
+	txn.SetOption(kv.ResourceGroupTagger, p.sctx.GetSessionVars().StmtCtx.GetResourceGroupTagger())
```
MyonKeminta, line 287: I'm actually not sure whether this is the proper place to do that. And should `RPCInterceptor` be also set? Or should `setOptionForTopSQL` be used instead?

ekexium: I put it here to let it work as a backup plan when some auto-commit pipelined dml stmts do not have such option set. It can be override later in other places. Or should we put it in `ExecuteStmt()` after `Compile()`?  @crazycs520 @mornyx What do you think?

you06: If the committer is already inited here, `SetOption(kv.ResourceGroupTagger, ...)` won't affect the committer.

https://github.com/tikv/client-go/blob/0606e74e8e373d224ad6e1392729b9780b76a8e8/txnkv/transaction/txn.go#L309-L312

ekexium: The pipelined committer is created when txn is created. It seems we need to modify SetOption

you06: LGTM

you06 (approved):
client-go size is updated in https://github.com/tikv/client-go/pull/1212

pkg/sessiontxn/isolation/base.go
ekexium: > I'm actually not sure whether this is the proper place to do that. And should RPCInterceptor be also set? Or should setOptionForTopSQL be used instead?

I copied the code to set these options before `commit`, and only apply them for pipelined txns, in favor of caution.

MyonKeminta (approved):

crazycs520 (approved):

PR #51801, opened 2024-03-15
lightning: skip CREATE DATABASE when downstream exists, and return error when parse failed

### LINKED ISSUE #51800: When lightning is less capabile than downstream, `USE <database>` may miss so failed with "no database selected"
The source schema files may contain some unsupported feature, like unsupported chatset. When lightning failed to rewrite the statement, it directly uses the source file content. However, `USE <database>` is missing

https://github.com/pingcap/tidb/blob/c1e3daebf183f93e3ed61e3ef56b8da430621d21/br/pkg/lightning/importer/import.go#L625-L638

lance6716:

Issue Number: close #51800

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/lightning/importer/import.go
+++ b/br/pkg/lightning/importer/import.go
@@ -630,7 +630,8 @@ func (worker *restoreSchemaWorker) addJob(sqlStr string, job *schemaJob) error {
 			zap.String("table", job.tblName),
 			zap.String("statement", sqlStr),
 			zap.Error(err))
-		job.stmts = []string{sqlStr}
+		useStmt := common.SprintfWithIdentifiers("USE %s", job.dbName)
+		job.stmts = []string{useStmt, sqlStr}
 	} else {
 		job.stmts = stmts
 	}
--- a/br/pkg/lightning/importer/import.go
+++ b/br/pkg/lightning/importer/import.go
@@ -633,15 +633,35 @@ func (worker *restoreSchemaWorker) addJob(sqlStr string, job *schemaJob) error {
 
 func (worker *restoreSchemaWorker) makeJobs(
 	dbMetas []*mydump.MDDatabaseMeta,
+	getDBs func(context.Context) ([]*model.DBInfo, error),
 	getTables func(context.Context, string) ([]*model.TableInfo, error),
 ) error {
 	defer func() {
 		close(worker.jobCh)
 		worker.quit()
 	}()
-	var err error
+
 	// 1. restore databases, execute statements concurrency
+
+	dbs, err := getDBs(worker.ctx)
+	if err != nil {
+		worker.logger.Warn("get databases from downstream failed", zap.Error(err))
+	}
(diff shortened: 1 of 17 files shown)


br/pkg/lightning/importer/import.go
```diff
@@ -630,7 +630,8 @@ func (worker *restoreSchemaWorker) addJob(sqlStr string, job *schemaJob) error {
 			zap.String("table", job.tblName),
 			zap.String("statement", sqlStr),
 			zap.Error(err))
-		job.stmts = []string{sqlStr}
+		useStmt := common.SprintfWithIdentifiers("USE %s", job.dbName)
+		job.stmts = []string{useStmt, sqlStr}
```
D3Hunter, line 634: we should return error if we cannot parse the sql, so user can see the real err. lightning version >= tidb-version in most case, so there shouldn't be cases that lightning cannot parse, but tidb can execute it.

lance6716: This is a cloud oncall, tiflow's version is lower than downstream TiDB

D3Hunter: then we should update tiflow with latest tidb dependency

also the logic to use original sql is error prone when the unparsable sql already contains dbname that might be the one we want

lance6716: Our rewrite logic will not change the error message if downstream can not execute them, the user still can see a meaningful error message which is reported by downstream TiDB's parser.

D3Hunter: i mean when the sql is unparsable by tiflow, but ok for tidb, then we might created it in the wrong db

okJiang: I agree `update tiflow with latest tidb dependency`

But since we already support the function of executing SQL even if it cannot be parsed, I think this PR can be fixed as a bug. Unless we firmly decide to remove it

lance6716: I should remove the fallback logic if tiflow can be newer than downstream TiDB

br/tests/lightning_character_sets/config.toml
Benjamin2037, line 1: why do we need a empty file in this pr？

lance6716: The `run_lightning` script needs a configuration file, otherwise it will report error. Empty file means default configuration.

br/pkg/lightning/importer/import.go
Frank945946: I also agree with "update tiflow with latest tidb dependency"

okJiang (approved):

3pointer (approved):

PR #51863, opened 2024-03-18
ttl: fix time zone issue in TTL

### LINKED ISSUE #51675: ttl: 1 hour's deviation for 'expire time' in TTL

lcwangchao:

Issue Number: close #51675

fix the time zone issue again after we reverted #51701. In this PR, we still use SQL to compute the expired time to make sure it has the same behavior as the expression. However, to make the result right we:

1. Set the session's time zone to UTC to make the time zone stable and has no daylight-saving time.
2. Make a new start time with the same string value but it is in time zone 'UTC'. For example, if the input now is `2020-01-01 08:00:00 +08:00`, the new start time is `2020-01-01 08:00:00 UTC`.
3. Compute the expired time with the new started time, for example, if the expire interval is `1 hour`, the expire time is `2020-01-01 07:00:00 UTC`
4. The interval between the above two times is the expired duration.
5. Compute the real expired time by `now.Add(-interval)`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
 fix time zone issue in TTL
```

### DIFF
--- a/pkg/ttl/cache/table.go
+++ b/pkg/ttl/cache/table.go
@@ -185,28 +185,101 @@ func (t *PhysicalTable) ValidateKeyPrefix(key []types.Datum) error {
 	return nil
 }
 
-// EvalExpireTime returns the expired time
+var mockExpireTimeKey struct{}
+
+// SetMockExpireTime can only used in test
+func SetMockExpireTime(ctx context.Context, tm time.Time) context.Context {
+	return context.WithValue(ctx, mockExpireTimeKey, tm)
+}
+
+// EvalExpireTime returns the expired time.
+// It uses the global timezone to compute the expired time.
+// Then we'll reset the returned expired time to the same timezone as the input `now`.
 func (t *PhysicalTable) EvalExpireTime(ctx context.Context, se session.Session,
 	now time.Time) (expire time.Time, err error) {
-	tz := se.GetSessionVars().Location()
+	if intest.InTest {
+		if tm, ok := ctx.Value(mockExpireTimeKey).(time.Time); ok {
+			return tm, err
+		}
+	}
 
 	expireExpr := t.TTLInfo.IntervalExprStr
 	unit := ast.TimeUnitType(t.TTLInfo.IntervalTimeUnit)
 
+	// Use the global time zone to compute expire time.
+	// Different timezones may have different results event with the same "now" time and TTL expression.
+	// Consider a TTL setting with the expiration `INTERVAL 1 MONTH`.
+	// If the current timezone is `Asia/Shanghai` and now is `2021-03-01 00:00:00 +0800`
+	// the expired time should be `2021-02-01 00:00:00 +0800`, corresponding to UTC time `2021-01-31 16:00:00 UTC`.
+	// But if we use the `UTC` time zone, the current time is `2021-02-28 16:00:00 UTC`,
+	// and the expired time should be `2021-01-28 16:00:00 UTC` that is not the same the previous one.
+	globalTz, err := se.GlobalTimeZone(ctx)
+	if err != nil {
+		return time.Time{}, err
+	}
+
 	var rows []chunk.Row
+
+	// We should set the session time zone to UTC because the next SQLs should be executed in the UTC timezone.
+	// The session time zone should be reverted to the original one after the SQLs are executed.
+	rows, err = se.ExecuteSQL(ctx, "SELECT @@time_zone")
+	if err != nil {
+		return
+	}
+
+	originalTZ := rows[0].GetString(0)
+	if _, err = se.ExecuteSQL(ctx, "SET @@time_zone='UTC'"); err != nil {
+		return
+	}
+
+	defer func() {
+		_, restoreErr := se.ExecuteSQL(ctx, "SET @@time_zone=%?", originalTZ)
+		if err == nil {
+			err = restoreErr
+		}
+	}()
+
+	// Firstly, we should use the UTC time zone to compute the expired time to avoid time shift caused by DST.
+	// The start time should be a time with the same datetime string as `now` but it is in the UTC timezone.
+	// For example, if global timezone is `Asia/Shanghai` with a string format `2020-01-01 08:00:00 +0800`.
+	// The startTime should be in timezone `UTC` and have a string format `2020-01-01 08:00:00 +0000` which is not the
+	// same as the original one (`2020-01-01 00:00:00 +0000` in UTC actually).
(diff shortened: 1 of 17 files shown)


YangKeao (approved):
LGTM

pkg/ttl/cache/table.go
```diff
@@ -185,28 +185,98 @@ func (t *PhysicalTable) ValidateKeyPrefix(key []types.Datum) error {
...
+		_, restoreErr := se.ExecuteSQL(ctx, "SET @@time_zone=%?", originalTZ)
+		if err == nil {
+			err = restoreErr
+		}
+	}()
+
+	// Firstly, we should use the UTC time zone to compute the expired time to avoid time shift caused by DST.
+	// The start time should be a time with the same datetime string as `now` but it is in the UTC timezone.
+	nowInGlobalTZ := now.In(globalTz)
+	startTime := time.Date(
```
bb7133, line 245: I don't really get the point here: `nowInGlobalTZ` and `startTime` are `time.Time`, what's the difference between this and `startTime := time.In(UTC)`?

pkg/ttl/ttlworker/job_manager_integration_test.go
```diff
@@ -63,6 +64,45 @@ func sessionFactory(t *testing.T, store kv.Storage) func() session.Session {
...
+
+	// global time zone should not change
+	tk.MustQuery("select @@global.time_zone").Check(testkit.Rows("Europe/Berlin"))
+	tz, err := se.GlobalTimeZone(context.TODO())
+	require.NoError(t, err)
+	require.Equal(t, "Europe/Berlin", tz.String())
+	// session variables should be set
+	tk.MustQuery("select @@time_zone, @@tidb_retry_limit, @@tidb_enable_1pc, @@tidb_enable_async_commit").
+		Check(testkit.Rows("UTC 0 1 1"))
+
```
bb7133, line 98: Please remove this blank line.

pkg/ttl/cache/table.go
```diff
@@ -185,28 +185,98 @@ func (t *PhysicalTable) ValidateKeyPrefix(key []types.Datum) error {
...
+		nowInGlobalTZ.Hour(), nowInGlobalTZ.Minute(), nowInGlobalTZ.Second(),
+		nowInGlobalTZ.Nanosecond(), time.UTC,
+	)
+
 	rows, err = se.ExecuteSQL(
 		ctx,
 		// FROM_UNIXTIME does not support negative value, so we use `FROM_UNIXTIME(0) + INTERVAL <current_ts>`
 		// to present current time
-		fmt.Sprintf("SELECT FROM_UNIXTIME(0) + INTERVAL %d SECOND - INTERVAL %s %s", now.Unix(), expireExpr, unit.String()),
+		fmt.Sprintf("SELECT FROM_UNIXTIME(0) + INTERVAL %d MICROSECOND - INTERVAL %s %s",
```
bb7133, line 255: It would be much easier(and more robust) way to calculate the expire time without SQL executions here. However, there's no straightforward way to do it because of `expireExpr`.

I can't imagine a better way but this can be treated as a backlog.

pkg/ttl/cache/table.go
lcwangchao: For example, if global time zone is `Asia/Shanghai` and nowInGlobalTZ is `2020-01-01 08:00:00 +08:00`.

`nowInGlobalTZ.In(time.UTC)` is: `2020-01-01 00:00:00 +00:00`
`startTime` is: `2020-01-01 08:00:00 +00:00`

`startTime` should have a same formated time string with `nowInGlobalTZ` but it they are not the same point

pkg/ttl/cache/table.go
lcwangchao: I think we can use `expression.ParseSimpleExpr` to parse it to an expression and evaluate it. But an `expression.BuildContext` is not very easy to build now. Maybe we can make it after expression refactor finished.

pkg/ttl/cache/table.go
bb7133: This is a bit confusing, could you please add your example to the comment?

lcwangchao: Done

bb7133 (approved):
LGTM

Issue #51664, opened 2024-03-11
TopDown optimizing approach exploring and implementation based on cascades/columbia


AilinKid:
## Enhancement

For modern advanced HTAP databases, the cascades-based optimizing approach will make the entire optimizing framework more flexible and extensible (modular, easy for adding/updating rules, making catalog and cost model more adaptive), eliminating the redundant searching space by group-pruning/low-bound admission branch check/equiv class classification, without completely do enumeration of each logical plan node from the bottom up.

### Infrastructure
* implement a stack mechanism  https://github.com/pingcap/tidb/pull/51663 
* Implement a serializing scheduling https://github.com/pingcap/tidb/pull/51866
* implement mutable and immutable fieldType https://github.com/pingcap/tidb/pull/51916
* make the task dir clean out of memo https://github.com/pingcap/tidb/pull/52083
* separate the pattern logic of the memo directory https://github.com/pingcap/tidb/pull/52117
#### HashEqual infra
  * add base hasher and equaler for taking in primitive type https://github.com/pingcap/tidb/pull/55570
  * regular the normal way to integrate HashEqual into base.LogicalPlan and expression.Expression https://github.com/pingcap/tidb/pull/55652
  * introduce hashEquals to column/collationinfo/fieldType https://github.com/pingcap/tidb/pull/55691
  * introduce hashEqual interface for datum https://github.com/pingcap/tidb/pull/55727
  * introduce hashEquals interface for expression.Expression https://github.com/pingcap/tidb/pull/55793

### Refactoring
The current `plan/core` pkg is quite huge not as slim as we expected. The hybrid placement of logicalOp, physicalOp, property, task, logical-rewrite, build-phase, binder, cost, exhaustion, etc makes the hierarchy complicated. The boundary of them is also not as clear as we desired. I concluded that there is some reason for this phenomenon.

* One is if Golang's structure wants to implement member functions with itself as a receiver, including the interface implementation, which can only be defined in the same pkg where the structure is defined. (here clearly the pkg is `core`) [stackoverflow ref](https://stackoverflow.com/questions/41109708/cant-define-receiver-from-another-package-in-go) 
[...]

PR #51866, opened 2024-03-18
planner: add simple serializing scheduler

AilinKid:

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/memo/task_scheduler.go
```diff
@@ -0,0 +1,53 @@
...
+		if err := task.execute(); err != nil {
+			s.Err = err
+			return
+		}
+	}
+}
+
+// TaskSchedulerContext is defined for scheduling logic calling, also facilitate interface-oriented coding and testing.
+type TaskSchedulerContext interface {
+	getStack() *TaskStack
```
winoros, line 51: Is this method for test only?

winoros:
You should use `ref #xxx` in the pr description instead of `close #xxx`(

pkg/planner/memo/task_scheduler.go
```diff
@@ -0,0 +1,53 @@
...
+	pushTask(task Task)
```
winoros, line 52: We can change the array from `[]task` to `[]*task` to avoid the overhead of pass by value.
Maybe we can make a benchmark later for the two kinds of usage.

If the task object is not small, it would be better to allocate it once on the heap instead of passing by value and allocating it at stack when we enter the `getTask` and `pushTask`.

winoros:
I think we don't need to export the `TaskStack` to other packages now that there's a `TaskSchedulerContext` to be exported?

pkg/planner/memo/task_scheduler.go
AilinKid: Task is a interface definition
```
type Task interface {
	// task self executing logic
	execute() error
	// task self description string.
	desc() string
}
type TaskStack struct {
	tasks []Task
}
```

AilinKid: https://github.com/pingcap/tidb/pull/51866/commits/8b62caef209f740c505fb02025444d6763299b96

AilinKid: // BenchmarkTestStack2Pointer-8   	   43802	     27569 ns/op	   24000 B/op	    2000 allocs/op
// BenchmarkTestStackInterface-8   	  108644	     10736 ns/op	    8000 B/op	    1000 allocs/op

AilinKid:
> I think we don't need to export the `TaskStack` to other packages now that there's a `TaskSchedulerContext` to be exported?

make sense, changed, currently, we only exported the Stack interface.

pkg/planner/memo/task_scheduler.go
AilinKid: no, TaskSchedulerContext is just a wrapper of Stack, scheduler should be responsible for how tasks are taken out and how it runs (operate on stack elements)

winoros (approved):

hawkingrei (approved):

PR #51869, opened 2024-03-18
ddl: fix some unit tests in ddl package when using infoschema v2

tiancaiamao:

Issue Number: ref #50959 

Problem Summary:

- set the ts field in infoschema when calling `do.tryLoadSchemaDiffs`
- make table ID unique in test TestCreateTableWithInfo
- fix nil pointer panic for test TestPlacementPolicyInUse

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

```
diff --git a/pkg/sessionctx/variable/tidb_vars.go b/pkg/sessionctx/variable/tidb_vars.go
index d5e9b34324..73e6f48387 100644
--- a/pkg/sessionctx/variable/tidb_vars.go
+++ b/pkg/sessionctx/variable/tidb_vars.go
@@ -1483,7 +1483,7 @@ const (
        DefTiDBSchemaVersionCacheLimit                    = 16
        DefTiDBIdleTransactionTimeout                     = 0
        DefTiDBTxnEntrySizeLimit                          = 0
-       DefTiDBSchemaCacheSize                            = 0
+       DefTiDBSchemaCacheSize                            = 1024
        DefTiDBLowResolutionTSOUpdateInterval             = 2000
        DefDivPrecisionIncrement                          = 4
        DefTiDBDMLType                                    = "STANDARD"
```

```
make ut X='run pkg/ddl' 1>a.log 2>b.log
grep FAIL a.log |wc -l
```

This commit fix more than 70+ cases in ddl packages.

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/domain/domain.go
```diff
@@ -271,7 +271,11 @@ func (do *Domain) loadInfoSchema(startTS uint64) (infoschema.InfoSchema, bool, i
 	// 4. No regenrated schema diff.
 	startTime := time.Now()
 	if currentSchemaVersion != 0 && neededSchemaVersion > currentSchemaVersion && neededSchemaVersion-currentSchemaVersion < LoadSchemaDiffVersionGapThreshold {
-		is, relatedChanges, diffTypes, err := do.tryLoadSchemaDiffs(m, currentSchemaVersion, neededSchemaVersion, schemaTs)
+		schemaTsOrStartTs := schemaTs
+		if schemaTsOrStartTs == 0 {
+			schemaTsOrStartTs = startTS
```
zimulala, line 276: Could set it before line242？ Then we needn't do this check.

tiancaiamao: `getTimestampForSchemaVersionWithNonEmptyDiff` always return schemaTs=0 ... (something is wrong but all the test cases still pass)
**I'm not sure using startTs as schemaTs for infoschema v1 is correct or not, so I put the code here.
This way it only affect infoscheme v2.**

zimulala: We also can use `schemaTsOrStartTs=startTs` in line242. So the `schemaTsOrStartTs` only affect v2?

zimulala: What I'm saying is that you can move this line over here, so that it doesn't affect the original logic and it does what you want

GMHDBJD (approved):
LGTM

ywqzzy (approved):

PR #51912, opened 2024-03-19
ldap: add timeout mechanism and optimize lock for LDAP dial and requests

### LINKED ISSUE #51883: A dead LDAP upstream can block the authentication and `show global variables`.
## Enhancement

Some functions about `LDAP` don't have timeout mechanism (e.g. `StartTLS`), therefore, the `RLock` of it cannot be released until the upstream returns.

I'd like to address the following enhancement:

1. Don't include any network / IO in the `RLock`. An `RLock` in golang can block the `Lock`, and a pending write lock will blocks all other `RLock`, which will make the things much worse.
2. Add timeout mechanism for `LDAP` functions.
3. Add metrics for them.

YangKeao:

Issue Number: close #51883

Problem Summary:

There are two problems:

1. A pending LDAP authentication process will block `rebuildSysVarCache`.
2. If the LDAP connection lost after the first handshake, the LDAP goroutine and function call will hang forever.

I have done two modification to fix this problem:

1. Refactor the original `ldapAuthImpl` to `ldapAuthImplBuilder`, which will copy the configurations (and take a reference to the connection pool, which include a closure with a copy of connection configurations). Therefore, during the scope of any lock, there didn't exist any IO operation, so that the `rebuildSysVarCache` will not be blocked by a network connection.
2. Add timeout to the LDAP requests.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

I used the `yangkeao/ldap-sasl-example` container to run the manual test, steps:

TODO

### Release note

```release-note
None
```

### DIFF
--- a/pkg/privilege/privileges/ldap/ldap_common.go
+++ b/pkg/privilege/privileges/ldap/ldap_common.go
@@ -66,32 +66,32 @@ type ldapAuthImplBuilder struct {
 	ldapConnectionPool *pools.ResourcePool
 }
 
-func (impl *ldapAuthImplBuilder) build() *ldapAuthImpl {
-	impl.RLock()
-	defer impl.RUnlock()
+func (builder *ldapAuthImplBuilder) build() *ldapAuthImpl {
+	builder.RLock()
+	defer builder.RUnlock()
 
 	return &ldapAuthImpl{
-		bindBaseDN:         impl.bindBaseDN,
-		bindRootDN:         impl.bindRootDN,
-		bindRootPWD:        impl.bindRootPWD,
-		searchAttr:         impl.searchAttr,
-		ldapConnectionPool: impl.ldapConnectionPool,
+		bindBaseDN:         builder.bindBaseDN,
+		bindRootDN:         builder.bindRootDN,
+		bindRootPWD:        builder.bindRootPWD,
+		searchAttr:         builder.searchAttr,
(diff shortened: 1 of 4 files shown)


pkg/privilege/privileges/ldap/ldap_common.go
```diff
@@ -52,49 +65,20 @@ type ldapAuthImpl struct {
...
-	}
-
-	if len(result.Entries) == 0 {
-		return "", errors.New("LDAP user not found")
-	}
-
-	dn = result.Entries[0].DN
-	return
-}
+func (impl *ldapAuthImplBuilder) build() *ldapAuthImpl {
```
bb7133, line 68:
```suggestion
func (builder *ldapAuthImplBuilder) build() *ldapAuthImpl {
```
Renaming it to `builder`?

YangKeao: Fixed. :+1:

YangKeao:
Now, I think this PR is ready to merge. cc @bb7133 and @CbcWestwolf

YangKeao:
/hold

I found a bug. If `ldapConnectionPool` is re-initialized, and an already built `ldapAuthImpl` is getting a connection from the pool, there'll be a data race :facepalm: . Let me fix it.

PR #51926, opened 2024-03-20
extension: disable some optimizations for extension function

### LINKED ISSUE #51925: some optimizations may cause some wrong result for extension functions.
## Bug Report

We found serval problems when a extension function is optimized by planner:

- extension function fold by mistake even if it has some side effect.
- extension function result may be cached in some expression such as `regexp`
- extension function maybe cached in plan cache that will cause privilege check pass even if some privileges are revoked.

You can see the test: https://github.com/pingcap/tidb/pull/51926/files#diff-b06ca4f8065e5716bc7b3e5d5cfe389156ad9125bddb40ff1d95a59d889d45c8R461 for detail

lcwangchao:

Issue Number: close #51925

Problem Summary:

If an extension function has some side-effect, it may be optimized by mistake. For example, if a function `myfunc` returns a random int, `select id, myfunc() from t` may produce same values for `myfunc` because it have been fold to a constant.

Another problem is that if an extension function requires some privilege, it will only be checked when building it. That causes some problems when the plan cache is enabled because the privilege will not be check again when evaluating a function and a user's privileges may be revoked.

We should avoid these unnecessary optimizations.

Disable some optimizations for extension functions.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


YangKeao (approved):
LGTM

qw4990 (approved):

windtalker (approved):
LGTM

PR #51927, opened 2024-03-20
ldap: add timeout and retry-backoff for ldap

### LINKED ISSUE #51883: A dead LDAP upstream can block the authentication and `show global variables`.

YangKeao:

Issue Number: close #51883

This PR is a smaller version of #51912. We'll finally get #51912 merged, but we need a smaller one to focus on the timeout mechanism (which is much simpler than refactor the locks).

If the LDAP connection lost after the first handshake, the LDAP goroutine and function call will hang forever.

I have done two modifications in this PR:

1. Add a timeout for LDAP dialing and requests. Therefore, the lock will be held for at most several seconds.
2. Add an interval to the retry mechanism, to avoid using all quotas provided by the LDAP service.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

`docker run --network host -it yangkeao/ldap-sasl-example:d2b324 /bin/bash` to get an environment with LDAP server. Then running a TiDB server at port 4000.

Create user, setup variables, and prepare CA:

```sql
set global authentication_ldap_sasl_server_host='127.0.0.1';
set global authentication_ldap_sasl_bind_root_dn='cn=admin,dc=example,dc=org';
set global authentication_ldap_sasl_bind_root_pwd='123456';
set global authentication_ldap_sasl_ca_path='/tmp/ca.crt';
set global authentication_ldap_simple_server_host='127.0.0.1';
set global authentication_ldap_simple_bind_root_dn='cn=admin,dc=example,dc=org';
set global authentication_ldap_simple_bind_root_pwd='123456';
set global authentication_ldap_simple_ca_path='/tmp/ca.crt';
create user yangkeao IDENTIFIED WITH authentication_ldap_simple as 'cn=yangkeao+uid=yangkeao,dc=example,dc=org';
```

```bash
sudo cp /proc/$(pidof mysqld)/root/etc/ssl/certs/example.crt /tmp/ca.crt
```

Then you can login to TiDB with `yangkeao` user:

```
LIBMYSQL_ENABLE_CLEARTEXT_PLUGIN=1 mysql -h 127.0.0.1 -u yangkeao -P 4000 -p123456
```

Enable or disable `authentication_ldap_simple_tls` are both fine.

Use the following iptables command to drop all packets to LDAP server:

```
sudo iptables -A INPUT -p tcp --dport 389 -j DROP
```

Then login without TLS will timeout after 10 seconds:

```
root@home:/# time LIBMYSQL_ENABLE_CLEARTEXT_PLUGIN=1 mysql -h 127.0.0.1 -u yangkeao -P 4000 -p123456
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'yangkeao'@'127.0.0.1' (using password: YES)

real    0m10.023s
user    0m0.000s
sys     0m0.009s
```

Then login with TLS will timeout after 20 seconds:

```
root@home:/# time LIBMYSQL_ENABLE_CLEARTEXT_PLUGIN=1 mysql -h 127.0.0.1 -u yangkeao -P 4000 -p123456
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'yangkeao'@'127.0.0.1' (using password: YES)

real    0m20.023s
user    0m0.005s
sys     0m0.004s
```

This PR also fixed some tiny issues: like rebuilding the connection pool after resetting the connection related variables, to avoid having wrong connection in the pool.

### Release note

```release-note
None
```

### DIFF
--- a/pkg/privilege/privileges/ldap/ldap_common.go
+++ b/pkg/privilege/privileges/ldap/ldap_common.go
@@ -22,11 +22,24 @@ import (
 	"os"
 	"strconv"
 	"sync"
+	"time"
 
 	"github.com/go-ldap/ldap/v3"
 	"github.com/ngaut/pools"
 	"github.com/pingcap/errors"
+	"github.com/pingcap/tidb/pkg/util/intest"
+	"github.com/pingcap/tidb/pkg/util/logutil"
+	"go.uber.org/zap"
 )
+
+// ldapTimeout is set to 10s. It works on both the TCP/TLS dialing timeout, and the LDAP request timeout. For connection with TLS, the
+// user may find that it fails after 2*ldapTimeout, because TiDB will try to connect through both `StartTLS` (from a normal TCP connection)
+// and `TLS`, therefore the total time is 2*ldapTimeout.
+var ldapTimeout = 10 * time.Second
+
+// skipTLSForTest is used to skip trying to connect with TLS directly in tests. If it's set to false, connection will only try to
+// use `StartTLS`
+var skipTLSForTest = false
 
 // ldapAuthImpl gives the internal utilities of authentication with LDAP.
 // The getter and setter methods will lock the mutex inside, while all other methods don't, so all other method call
@@ -115,10 +128,13 @@ func (impl *ldapAuthImpl) initializeCAPool() error {
 }
 
 func (impl *ldapAuthImpl) tryConnectLDAPThroughStartTLS(address string) (*ldap.Conn, error) {
-	ldapConnection, err := ldap.Dial("tcp", address)
+	ldapConnection, err := ldap.DialURL("ldap://"+address, ldap.DialWithDialer(&net.Dialer{
+		Timeout: ldapTimeout,
+	}))
 	if err != nil {
 		return nil, err
 	}
+	ldapConnection.SetTimeout(ldapTimeout)
 
 	err = ldapConnection.StartTLS(&tls.Config{
 		RootCAs:    impl.caPool,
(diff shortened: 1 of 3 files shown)


pkg/privilege/privileges/ldap/ldap_common.go
```diff
@@ -22,12 +22,25 @@ import (
...
 
 	"github.com/go-ldap/ldap/v3"
 	"github.com/ngaut/pools"
 	"github.com/pingcap/errors"
+	"github.com/pingcap/tidb/pkg/util/intest"
+	"github.com/pingcap/tidb/pkg/util/logutil"
+	"go.uber.org/zap"
 )
 
+// ldapTimeout is set to 15s. It works on both the TCP/TLS dialing timeout, and the LDAP request timeout. For connection with TLS, the
```
bb7133, line 35:
```suggestion
// ldapTimeout is set to 10s. It works on both the TCP/TLS dialing timeout, and the LDAP request timeout. For connection with TLS, the
```

bb7133 (approved):
LGTM

pkg/privilege/privileges/ldap/ldap_common.go
```diff
@@ -191,13 +220,19 @@ func (impl *ldapAuthImpl) getConnection() (*ldap.Conn, error) {
...
+
 			// fail to bind to anonymous user, just release this connection and try to get a new one
 			impl.ldapConnectionPool.Put(nil)
 
 			retryCount++
 			if retryCount >= getConnectionMaxRetry {
 				return nil, errors.Wrap(err, "fail to bind to anonymous user")
 			}
+			// Be careful that it's still holding the lock of the system variables, so it's not good to sleep here.
+			// TODO: refactor the `RWLock` to avoid the problem of holding the lock.
```
hawkingrei, line 234: You can use ```syncutil.RWMutex``` to replace ```sync.RWMutex```. which can find deadlock in the test.

YangKeao: The current issue is not about deadlock (there's only one lock, and is not recursive, so it's impossible to deadlock here).

The problem is that a pending write lock (by rebuildSysVarCache) will block all other read lock, which I didn't realize before 🤦 😢 .

CbcWestwolf (approved):

PR #51996, opened 2024-03-21
server, metrics: remove the connection count on server, only use the metrics

### LINKED ISSUE #51889: Connection count metric can be less than the real value
This issue has similar reason with #42761, because the `conn.Close()` will be called multiple times :facepalm:.

```go
func (s *Server) onConn(conn *clientConn) {
...
	defer func() {
		terror.Log(conn.Close())
		logutil.Logger(ctx).Debug("connection closed")
	}()
...
}
```

And 

```go
func (cc *clientConn) Run(ctx context.Context) {
	defer func() {
		...
		if cc.getStatus() != connStatusShutdown {
			err := cc.Close()
			terror.Log(err)
		}

		close(cc.quit)
	}()
```

We have two ways to solve this problem:

1. Don't include any logic in `conn.Close()`. Only have logic in the `sync.Once` inside the `closeConn`.
2. Enhance the code, **don't close the connection for multiple times!** (it's more intuitive for the developers).

YangKeao:

Issue Number: close #51889

Problem Summary:

1. The connection count is not correct, if the session resets the resource group manually.
2. Maintaining a separated map on `server` to count the connection for each resource group seems to be not necessary. We can use the `GaugeVec` itself. The connection count is not used anywhere else.
3. The grafana panel of connection count is confusing. I added a legend and now it looks like:

image

Maybe someone prefers to sum up them together. Actually I think both are fine, feel free to comment your opinion.

Personally, I don't like the idea to use the uncontrollable string as a label :facepalm:, because the prometheus will have much greater pressure when these labels multiply together.

As for now, it seems not a big issue, but it'll become much worse if anyone adds more similar labels.

1. Use the `GaugeVec` as the only counter of the connection.
2. Also increase and decrease the counter when the user `set resource group` manually.
3. Modify the grafana definition. I found that some modification on the json file is not represented in the jsonnet, so I also fix them.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
Fix the issue that the connection count is not calculated and shown correctly.
```

### DIFF
--- a/pkg/executor/simple.go
+++ b/pkg/executor/simple.go
@@ -42,6 +42,7 @@ import (
 	"github.com/pingcap/tidb/pkg/infoschema"
 	"github.com/pingcap/tidb/pkg/kv"
 	"github.com/pingcap/tidb/pkg/meta"
+	"github.com/pingcap/tidb/pkg/metrics"
 	"github.com/pingcap/tidb/pkg/parser/ast"
 	"github.com/pingcap/tidb/pkg/parser/auth"
 	"github.com/pingcap/tidb/pkg/parser/model"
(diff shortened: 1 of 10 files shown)


tiancaiamao (approved):

nolouch (approved):
lgtm

PR #52000, opened 2024-03-21
ddl: determine whether to use merge sort by each index separately

### LINKED ISSUE #52004: Multi-schema adding index should not trigger merge step
## Bug Report

Add multiple indexes for 1TB sysbench table, for example

```sql
alter table sbtest1 
  add index idx_p(pad),
  add index idx_k(k),
  add index idx_c(c);
```

TiDB should not run merge sort because the dataset is not large enough.

TiDB run merge sort step.

img_v3_0296_5456d25a-ebc2-42e4-855b-5d3f8e82b21g

It may take more than 100% time to complete.

69c26ae8a

tangenta:

Issue Number: close #52004

Problem Summary:

Before this PR, when TiDB adds multiple indexes with global sort, merge sort step is determined by external file count of ALL indexes. For example, the threshold is 4000, there are 3 indexes with 3000 files respectively, then there are 9000(>4000) files. Merge sort step will be executed. However, merge sort is not necessary in this case, because the records from different indexes are not overlap.

- Use file count from each index to determine whether use merge sort.
- Refine a error message and add a test case.
- Use average row size to determine reader count for ingest mode.
- Increase system session pool from 128 to 512, because readers could increase a lot and block session allocation.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/ddl/backfilling_scheduler.go
```diff
@@ -329,18 +331,37 @@ type ingestBackfillScheduler struct {
...
 
+func estimateAvgRowSize(tbl table.PhysicalTable) int {
+	size := 0
+	for _, col := range tbl.Meta().Columns {
+		ft := col.FieldType
+		storeLen := ft.StorageLength()
+		if storeLen != types.VarStorageLen {
+			size += storeLen
+		} else {
+			if types.IsString(ft.GetType()) {
```
Benjamin2037, line 356: will the Json and blob be treated as char type？What is the return length of them.

pkg/ddl/backfilling_scheduler.go
```diff
@@ -501,15 +524,21 @@ func (b *ingestBackfillScheduler) createCopReqSenderPool() (*copReqSenderPool, e
...
-func expectedIngestWorkerCnt() (readerCnt, writerCnt int) {
+func expectedIngestWorkerCnt(avgRowSize int) (readerCnt, writerCnt int) {
 	workerCnt := int(variable.GetDDLReorgWorkerCounter())
-	readerCnt = min(workerCnt/2, maxBackfillWorkerSize)
-	readerCnt = max(readerCnt, 1)
-	writerCnt = min(workerCnt/2+2, maxBackfillWorkerSize)
+	readerRatio := []float64{0.5, 1, 2, 4, 8}
+	rowSize := []int{200, 500, 1000, 3000, 5000}
+	for i, s := range rowSize {
+		if avgRowSize <= s {
```
Benjamin2037, line 536: how about when avgRowSize > 5000, reader and writer cnt?

pkg/ddl/backfilling_scheduler.go
```diff
@@ -501,15 +524,21 @@ func (b *ingestBackfillScheduler) createCopReqSenderPool() (*copReqSenderPool, e
...
+			readerCnt = int(float64(workerCnt) * readerRatio[i])
+			writerCnt = workerCnt
```
Benjamin2037, line 538: also. < 200, the writerCnt need +2 otherwise, I remeber the performance will degrade bigger than 10%

tangenta: I prefer to make it simple and easy to predict.

pkg/domain/domain.go
```diff
@@ -1175,7 +1175,7 @@ func (do *Domain) Init(
 	sysFac := func() (pools.Resource, error) {
 		return sysExecutorFactory(do)
 	}
-	sysCtxPool := pools.NewResourcePool(sysFac, 128, 128, resourceIdleTimeout)
+	sysCtxPool := pools.NewResourcePool(sysFac, 512, 512, resourceIdleTimeout)
```
Benjamin2037, line 1178: What 128 to 512 mean?

pkg/ddl/backfilling_scheduler.go
Benjamin2037: OK, let's see the perf testing result.

Benjamin2037 (approved):
rest LGTM

pkg/ddl/backfilling_dist_scheduler.go
```diff
@@ -466,32 +466,42 @@ func generateMergePlan(
...
+	if allSkip {
 		logger.Info("skip merge sort")
 		return nil, nil
 	}
 
-	// generate merge sort plan.
-	dataFiles := make([]string, 0, 1000)
+	metaArr := make([][]byte, 0, 16)
 	for i, g := range kvMetaGroups {
+		dataFiles := make([]string, 0, 1000)
```
ywqzzy, line 504: Why not generate subtasks for each index that really needs merge sort？

tangenta: If we do this, in step 3, we need to scan subtasks meta from both step 2 and step 1.

wjhuang2016 (approved):

PR #52070, opened 2024-03-25
infoschema: introduce `MetaOnlyInfoSchema` to provide meta only information schema

### LINKED ISSUE #52072: introduce MetaOnlyInfoSchema to provide meta only information schema
## Enhancement

see descriptions in #52070

lcwangchao:

Issue Number: close #52072

Problem Summary:

The API of `infoschema.InfoSchema` introduces too many dependencies. The main reason is that it not only provides "meta" of tables but also returns `table.Table` for table mutations. It brings two problems.

The first problem is the dependency cycle. `infoschema.InfoSchema` is a basic lib and is required by many other packages. This may cause a dependency cycle easily.

Another problem is that, if we want to break the dependency cycle. We have to use some "magic ways". For example, return `infoschema.InfoSchemaVersion` instead of `infoschema.InfoSchema`  in function signature but cast it to `infoschema.InfoSchema` when using it. However, the casting `is.(infoschema.InfoSchema)` is ugly and may lost the benefit of static checks in compile phase.

In this PR, we extended `infoschema.InfoschemaVersion` to `infoschema.MetaOnlyInfoSchema` to provide more common use methods in API and avoid importing the `table` package there. In most scenes, we only need the table meta and do not want to modify it, so `infoschema.MetaOnlyInfoSchema` is enough.

We also added some methods to provide table meta only:

```
type MetaOnlyInfoSchema interface {
	// ...
	TableInfoByName(schema, table model.CIStr) (*model.TableInfo, error)
	TableInfoByID(id int64) (*model.TableInfo, bool)
	FindTableInfoByPartitionID(partitionID int64) (*model.TableInfo, *model.DBInfo, *model.PartitionDefinition)
	SchemaTableInfos(schema model.CIStr) []*model.TableInfo
	// ...
}
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/ddl/schematracker/info_store.go
```diff
@@ -181,3 +181,9 @@ func (i InfoStoreAdaptor) TableByName(schema, table model.CIStr) (t table.Table,
 	}
 	return tables.MockTableFromMeta(tableInfo), nil
 }
+
+// TableInfoByName implements the InfoSchema interface.
+// nolint:unused
```
lance6716, line 186: public function will not trigger "unused" linter?

lance6716 (approved):

pkg/ddl/schematracker/info_store.go
lcwangchao: I'm not sure, just copied from exists method like `TableByName` ...

tangenta (approved):

pkg/ddl/schematracker/info_store.go
tangenta: How about trying to remove it?

lcwangchao: ok, I'll try to remove it

windtalker (approved):
LGTM

Issue #51807, opened 2024-03-15
Exchange partition cannot be used after creating/dropping a foreign key


YangKeao:
### 1. Minimal reproduce step (Required)

```
create table parent (id int unique);
create table child (id int, parent_id int, foreign key (parent_id) references parent(id));
create table child_with_partition(id int, parent_id int) partition by range(id) (partition p1 values less than (100));
alter table child drop foreign key fk_1;
alter table child drop key fk_1;
alter table child_with_partition exchange partition p1 with table child;
```

### 2. What did you expect to see? (Required)

Partition exchanged successfully.

### 3. What did you see instead (Required)

```
mysql> alter table child_with_partition exchange partition p1 with table child;
ERROR 1740 (HY000): Table to exchange with partition has foreign key references: 'child'
```

YangKeao:
Duplicated with https://github.com/pingcap/tidb/issues/51762. Close

PR #52073, opened 2024-03-25
lint: add a linter to avoid using `==nil` to assert empty slice or map

### LINKED ISSUE #52074: add a linter to avoid asserting whether slice is empty (or not) by `arr == nil`.
## Enhancement

Ref #51807. It's easy to make mistake that `[]T{} == nil` returns false, so that we'd better to always use `len(arr) > 0` to assert whether a slice is empty. We should also provide an utility to assert whether the slice is really nil.

YangKeao:
It's such a huge change!!! I'm not sure whether merging this PR is good. Need discussion.

They are fixed automatically by the linter.

Issue Number: close #52074, ref #51807

Problem Summary:

It's easy to make mistake to use `arr == nil` to tell whether a slice is empty, but if the slice is `[]T{}`, this condition will not meet. #51807 is one of the bugs made by it. Therefore I tried to add a linter to avoid this situation.

This PR also fixed similar issues for `map`.

1. Don't allow using `arr == nil`.
2. If you really want to use `arr == nil` to check whether is not initialized (or empty return), you can use `emptyslice.IsNil(arr)`, which is ignored by the linter.

It gives a lot of error, and I didn't fix them all yet. I'm not sure whether it's good to add this linter.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```


build/linter/emptynil/analyzer.go
```diff
@@ -0,0 +1,181 @@
...
+
+var Analyzer = &analysis.Analyzer{
+	Name:     "emptynil",
+	Doc:      "avoid using `x == nil` to check whether a slice/map is empty",
+	Requires: []*analysis.Analyzer{inspect.Analyzer},
+	Run:      run,
+}
+
+const msgForEQ = "should use len(x) == 0 to check whether a slice or map is empty. If you indeed need to check whether it is nil, use emptynil.IsNilSlice(x) or emptynil.IsNilMap(x)."
+const msgForNEQ = "should use len(x) > 0 to check whether a slice or map is not empty. If you indeed need to check whether it is nil, use !emptynil.IsNilSlice(x) or emptynil.IsNilMap(x)."
```
bb7133, line 38:
```suggestion
const msgForNEQ = "should use len(x) > 0 to check whether a slice or map is not empty. If you indeed need to check whether it is nil, use !emptynil.IsNilSlice(x) or !emptynil.IsNilMap(x)."
```

bb7133 (commented):
LGTM, but...

How to trigger this lint manually? I found a lot of checks related to lint like `make lint`, `make static-check`, `make bazel_lint`... :(

build/linter/emptynil/analyzer.go
YangKeao: Nice catch :+1: . Fixed.

YangKeao:
> LGTM, but...
> 
> How to trigger this lint manually? I found a lot of checks related to lint like `make lint`, `make static-check`, `make bazel_lint`... :(

I think `make bazel_lint` can trigger this lint.

This PR also provides a main package in `/build/linter/emptynil/cmd`. Using `-fix` argument will fix the found issues automatically.

D3Hunter:
i think it's a little over-reacting to the bug in description. misusing between `len(xx) == 0` and `xx == nil` are human errors, that cannot be avoided by replacing `xx == nil` with `isNil(xx)`, we still might make the same mistake by misusing `len(xx) == 0` and `isNil(xx)`, and it sacrifices readability.

Issue #52133, opened 2024-03-27
Column prune related improvements


yibin87:
## Enhancement
Column prune will improve performance in several aspects:

- Reduce memory usage, since fewer column data may be loaded.
- Reduce network overhead, because we have dependant strorage systems: tikv, tiflash, fewer columns, fewer data will be transfered
- Reduce CPU usage, some operators' workload are related with the size of their input columns.  For example, PhysicalExchange node will partition all its input columns, therefore, fewer columns, fewer workload.

However, there are still several column prune related issues
1. Useless columns in union operator may not be pruned
Currently, when a LogicalUnionAll operator's parent operator doesn't use the LogicalUnionAll operator's output columns(e.g. select count(*) from (xx union xx)),  these useless columns won't be pruned:
https://github.com/pingcap/tidb/blob/b96f08134d66196ca760a6caf4d0e9cd044fff97/pkg/planner/core/rule_column_pruning.go#L303C1-L309C3
2. Useless columns used in filter expressions won't be pruned
When column prune optimization is executed, filter is still included in DataSource operator. Currently, the DataSource operator's output schema should contains columns that are not only needed by its parent, but also used by the filter operator. These columns which are used by the filter operator, while not needed by its parent, can be pruned. 
For MPP mode, add a new projection above the DataSource operator is enough.
For Tikv mode, we need to [push down the projection](https://github.com/pingcap/tidb/issues/51876) to tikv to reduce network overhead.
3. MPP physical join operator abandons the column prune achievements
During the process of building MPP tasks, physical join operaotr's schema is reset to its full semantic output:
https://github.com/pingcap/tidb/blob/b96f08134d66196ca760a6caf4d0e9cd044fff97/pkg/planner/core/task.go#L521

PR #52143, opened 2024-03-27
Column prune improvement

yibin87:

Issue Number: ref #52133 

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/rule_column_pruning.go
```diff
@@ -414,6 +416,15 @@ func (ds *DataSource) PruneColumns(parentUsedCols []*expression.Column, opt *log
...
 	}
+	if !addOneHandle && ds.schema.Len() > len(parentUsedCols) && ds.SCtx().GetSessionVars().IsMPPEnforced() {
+		proj := LogicalProjection{
+			Exprs: expression.Column2Exprs(parentUsedCols),
+		}.Init(ds.SCtx(), ds.QueryBlockOffset())
+		proj.SetStats(ds.StatsInfo())
+		proj.SetSchema(expression.NewSchema(parentUsedCols...))
+		proj.SetChildren(ds)
+		return proj, nil
+	}
```
winoros, line 427: It's not a very good idea to add it here. Let me take a deeper look.

yibin87: Please let me know if any suggestions.

elsa0520: > It's not a very good idea to add it here. Let me take a deeper look.

1. I think this is not a only mpp task enhancement but also suite for tikv task
2. Should we change the output schema of datasource rather than add a projection operator in here ?

pkg/planner/core/task.go
```diff
@@ -544,6 +542,31 @@ func (p *PhysicalHashJoin) attach2TaskForMpp(tasks ...task) task {
 		partTp:   outerTask.partTp,
 		hashCols: outerTask.hashCols,
 	}
+	defaultSchema := BuildPhysicalJoinSchema(p.JoinType, p)
+	if p.schema.Len() < defaultSchema.Len() {
+		if p.schema.Len() > 0 {
```
elsa0520, line 547: We'd better column pruning in logical rule  phase rather than attachToTask .

pkg/planner/core/rule_column_pruning.go
yibin87: Yeah, for point 1, tikv task will benefit from this if projection push down work, and this will be the next work I'll focus on(will see more detail in [issue](https://github.com/pingcap/tidb/issues/52133)), currently, limit it to mpp task only.
For point 2, as I know, we can't change the output schema here. Please correct me if I'm wrong @winoros

pkg/planner/core/task.go
yibin87: I think, since this work takes effect only for mpp task, it seems reasonable to add it here.

winoros (approved):
LGTM for now

elsa0520 (approved):
LGTM

Issue #45532, opened 2023-07-24
Support `BatchPointGet` for Range and List Partition


Defined2014:
## Enhancement
We could support  `BatchPointGet` plan for Range and List Partition and remove the code below. https://github.com/pingcap/tidb/blob/e68c8ff2a59817920b5ab7c1db4eb9a2ed7fff4c/planner/core/find_best_task.go#L1126-L1134

Some things need to consider,
1. https://github.com/pingcap/tidb/issues/45702
2. Explain info, like https://github.com/pingcap/tidb/issues/45888
3. https://github.com/pingcap/tidb/issues/45720
4. https://github.com/pingcap/tidb/issues/45889
5. https://github.com/pingcap/tidb/blob/6b536ae1cb5cf39e15008621df7907bce7a7a1fa/server/conn.go#L1875, `prefetchPointPlanKeys` function, not sure if this will affect.

Most test cases are added in `TestBatchPointGetTablePartition`, but please take care about all of them.

Defined2014:
Some interesting things
```
mysql> show create table t;
+-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                                                                                                                     |
+-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `a` int(11) NOT NULL,
  `b` int(11) DEFAULT NULL,
  PRIMARY KEY (`a`) /*T![clustered_index] CLUSTERED */
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
PARTITION BY RANGE (`a`)
(PARTITION `p0` VALUES LESS THAN (10),
 PARTITION `p1` VALUES LESS THAN (20)) |
+-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> explain select * from t where a = 1 or a = 11;
+------------------------+---------+-----------+---------------+------------------------------------------------------+
| id                     | estRows | task      | access object | operator info                                        |
+------------------------+---------+-----------+---------------+------------------------------------------------------+
| TableReader_6          | 2.00    | root      | partition:all | data:TableRangeScan_5                                |
[...]

Issue #33031, opened 2022-03-14
return wrong results when using plan-cache and dynamic-mode


ChenPeng2013:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

prepared-plan-cache.enabled = true
```
use test;
drop table if exists UK_RP17396;
CREATE TABLE `UK_RP17396` (
  `COL1` int(16) DEFAULT '29' COMMENT 'NUMERIC UNIQUE INDEX',
  `COL2` varchar(20) DEFAULT NULL,
  `COL4` datetime DEFAULT NULL,
  `COL3` bigint(20) DEFAULT NULL,
  `COL5` float DEFAULT NULL,
  UNIQUE KEY `UK_COL1` (`COL1`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
PARTITION BY RANGE (`COL1`)
(PARTITION `P0` VALUES LESS THAN (-1964648596),
 PARTITION `P1` VALUES LESS THAN (-1798171365),
 PARTITION `P2` VALUES LESS THAN (-224353317),
 PARTITION `P3` VALUES LESS THAN (74739562),
 PARTITION `P4` VALUES LESS THAN (330180811),
 PARTITION `P5` VALUES LESS THAN (1267470870));

insert into UK_RP17396 values(-556906299, "竂揿蜓锌璸籾珵岄浑祿恺鮘茷拦没盬饎醱僛橉", "9269-08-13 18:36:02", 363208857940760420, -1.28355e38);


set tidb_partition_prune_mode=static;
prepare stmt from 'select * from UK_RP17396 where col3 < ? and col1 in (?, ?, ?) and col2 != ?;';
set @a=4020185825752602279, @b=-4331, @c=-29749, @d=-13306, @e="AgHVfFsnfEO";
execute stmt using @a,@b,@c,@d,@e;
set @a=2972768264503339938, @b=5150, @c=-556906299, @d=1232847360, @e="=姵犵駘鏤ⱴȴ褋韨窷晢镸辈夁沶蚩皭兩隧撬";
execute stmt using @a,@b,@c,@d,@e;


set tidb_partition_prune_mode=dynamic;
prepare stmt from 'select * from UK_RP17396 where col3 < ? and col1 in (?, ?, ?) and col2 != ?;';
[...]

PR #52170, opened 2024-03-27
planner: plan cache+partitioned tables benchmarks

mjonss:

Issue Number: ref #33031 , ref #45532

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


Makefile
```diff
@@ -298,15 +298,18 @@ endif
 # Usage:
 #	make bench-daily TO=/path/to/file.json
 bench-daily:
-	go test -tags intest github.com/pingcap/tidb/pkg/session -run TestBenchDaily -bench Ignore --outfile bench_daily.json
+	go test -tags intest github.com/pingcap/tidb/pkg/distsql -run TestBenchDaily -bench Ignore --outfile bench_daily.json
```
mjonss, line 301: I just sorted according to the packages.

Makefile
```diff
@@ -298,15 +298,18 @@ endif
...
 	go test -tags intest github.com/pingcap/tidb/pkg/executor -run TestBenchDaily -bench Ignore --outfile bench_daily.json
 	go test -tags intest github.com/pingcap/tidb/pkg/executor/test/splittest -run TestBenchDaily -bench Ignore --outfile bench_daily.json
-	go test -tags intest github.com/pingcap/tidb/pkg/tablecodec -run TestBenchDaily -bench Ignore --outfile bench_daily.json
 	go test -tags intest github.com/pingcap/tidb/pkg/expression -run TestBenchDaily -bench Ignore --outfile bench_daily.json
+	go test -tags intest github.com/pingcap/tidb/pkg/planner/core/tests/partition -run '^[^T]' -bench . -benchmem \
+		| awk 'BEGIN{printf("[")} /^Benchmark.*ns\/op/ {if (found>0) {printf(",")}; sub(/-[0-9]*/,"",$$1); printf("{\"Name\":\"%s\",\"NsPerOps\":%d,\"AllocsPerOp\":%d,\"BytesPerOp\":%d}",$$1,$$3,$$7,$$5); found++ } END{printf("]")}' \
+		> pkg/planner/core/tests/partition/bench_daily.json
```
mjonss, line 307: I could not find a way to run SubBenchmarks to be reported as separate tests, so I am just running it as command line bench marks and transforming the output to json through old fashioned texts processing...

mjonss: Maybe we should remove the awk script with https://pkg.go.dev/go.bobheadxi.dev/gobenchdata#readme-command-line-interface ?

Makefile
```diff
@@ -298,15 +298,18 @@ endif
...
+	go test -tags intest github.com/pingcap/tidb/pkg/session -run TestBenchDaily -bench Ignore --outfile bench_daily.json
+	go test -tags intest github.com/pingcap/tidb/pkg/statistics -run TestBenchDaily -bench Ignore --outfile bench_daily.json
+	go test -tags intest github.com/pingcap/tidb/pkg/tablecodec -run TestBenchDaily -bench Ignore --outfile bench_daily.json
 	go test -tags intest github.com/pingcap/tidb/pkg/util/rowcodec -run TestBenchDaily -bench Ignore --outfile bench_daily.json
 	go test -tags intest github.com/pingcap/tidb/pkg/util/codec -run TestBenchDaily -bench Ignore --outfile bench_daily.json
-	go test -tags intest github.com/pingcap/tidb/pkg/distsql -run TestBenchDaily -bench Ignore --outfile bench_daily.json
-	go test -tags intest github.com/pingcap/tidb/pkg/statistics -run TestBenchDaily -bench Ignore --outfile bench_daily.json
 	go test -tags intest github.com/pingcap/tidb/pkg/util/benchdaily -run TestBenchDaily -bench Ignore \
```
mjonss, line 313: Would it be possible to use https://pkg.go.dev/go.bobheadxi.dev/gobenchdata#readme-command-line-interface for merging the files? Or extend benchdaily to extract the benchmark data and output it to json, instead of the awk script? (not that it is anything wrong with awk...)

hawkingrei (approved):

Makefile
```diff
@@ -363,15 +363,18 @@ endif
...
 	go test github.com/pingcap/tidb/pkg/executor -run TestBenchDaily -bench Ignore --outfile bench_daily.json
 	go test github.com/pingcap/tidb/pkg/executor/test/splittest -run TestBenchDaily -bench Ignore --outfile bench_daily.json
-	go test github.com/pingcap/tidb/pkg/tablecodec -run TestBenchDaily -bench Ignore --outfile bench_daily.json
 	go test github.com/pingcap/tidb/pkg/expression -run TestBenchDaily -bench Ignore --outfile bench_daily.json
-	go test github.com/pingcap/tidb/pkg/util/rowcodec -run TestBenchDaily -bench Ignore --outfile bench_daily.json
-	go test github.com/pingcap/tidb/pkg/util/codec -run TestBenchDaily -bench Ignore --outfile bench_daily.json
-	go test github.com/pingcap/tidb/pkg/distsql -run TestBenchDaily -bench Ignore --outfile bench_daily.json
+	go test github.com/pingcap/tidb/pkg/planner/core/tests/partition -run '^[^T]' -bench . -benchmem \
+		| awk 'BEGIN{printf("[")} /^Benchmark.*ns\/op/ {if (found>0) {printf(",")}; sub(/-[0-9]*/,"",$$1); printf("{\"Name\":\"%s\",\"NsPerOps\":%d,\"AllocsPerOp\":%d,\"BytesPerOp\":%d}",$$1,$$3,$$7,$$5); found++ } END{printf("]")}' \
```
tiancaiamao, line 371: I found the generate json file can not pass json pretty format, so this maybe not quite stable.
If you do not call `benchdaily.Run()` directly, maybe you can wrap around this? The utilities are all there and simple enough.

https://github.com/pingcap/tidb/blob/81da4f8811f2fc150ed176d7ef3bcde3ce112d9f/pkg/util/benchdaily/bench_daily.go#L77-L78

tiancaiamao: You can expose `benchmarkResultToJSON`

tiancaiamao (commented):
```
time go test github.com/pingcap/tidb/pkg/planner/core/tests/partition -run '^[^T]' -bench . -benchmem 

ok      github.com/pingcap/tidb/pkg/planner/core/tests/partition        492.708s

real    8m18.244s
user    31m26.115s
sys     0m46.970s
```

It takes 8min to run this new added tests...  do we really need to cover so many combinated case?

mjonss:
> @tiancaiamao It takes 8min to run this new added tests... do we really need to cover so many combinated case?

I'm not sure, but I wanted to have as much coverage as possible for various methods of partitioning (and include non-partitioned as well for comparison).

@hawkingrei What is your take on adding ~10 min of benchmarks for the daily-bench suite?

Makefile
mjonss: I updated the Makefile with gobenchdata + gojq to extract and format the daily_bench.json file instead.
The issue with `testing.Benchmark(t)` is that it runs a full 'main' benchmark, and sums all the 'sub-benchmarks', so no handy way of getting each sub-benchmark numbers.

mjonss:
I tested to create a single Benchmark function for each benchmark, i.e. instead of having one main benchmark per partitionBy+isPrepared, and one sub benchmark per accessMethod+usePlanCache, only having main benchmarks, one for each combination of partitionBy, isPrepared, accessMethod, usePlanCache.

That way it is possible to run the `benchdaily.Run(...)` and collect the benchmark results the same way as for all our other benchmarks.

However it seems like it will take almost double the time, due to not able to reuse the created tables and their data between different runs, especially for the IndexLookup and TableScan cases, where I chose to use 1k rows instead of just 5 rows.

The reason for having more data for IndexLookup and TableScan benchmarks are so they are more real and also to avoid using optimizer hints for forcing IndexLookup.

After some thinking I propose that I complement the main/sub benchmarks with only main and only listing a few partitioning variants in `benchdaily.Run()`, and that way keeping the additional time a bit more reasonable. While we can still run all variants on demand, both the previous main+sub, new main-only and a subset of the main-only in `benchdaily.Run()` in the standard way.

@tiancaiamao @hawkingrei  What is your take on this?

mjonss:
@tiancaiamao I have limited the daily benchmarks to 12, and it now takes ~1 minute. I have kept all the variations so one can run them manually when needed for finding improvements and regressions. Please take a look :)

tiancaiamao (approved):

PR #52218, opened 2024-03-28
statistics: fix the potential error when merging global stats

### LINKED ISSUE #52251: Possible wrong global histogram after merging the partition statistics
img_v3_029e_2a4ddc4e-edd6-44f4-946c-9936eabe1f2g

After the `slices.SortFunc`, The order like `[2, 2], [1, 3], [2, 3]` will be a valid sorted result. And the following merging process just checks the bucket size. So it's possible that we treat the `[2, 2]` as the first global bucket and merge `[1, 3]` and `[2, 3]` as the second.

The merge result is wrong.

winoros:

Issue Number: close #52251

Problem Summary:

It's another way compared with https://github.com/pingcap/tidb/pull/52122.
This one should be better to solve problems.

See the changed tests to know the error of the previous codes.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/test/analyzetest/analyze_test.go
```diff
@@ -1429,9 +1429,9 @@ func TestAnalyzeColumnsWithDynamicPartitionTable(t *testing.T) {
 			tk.MustQuery("show stats_buckets where db_name = 'test' and table_name = 't' and is_index = 0").Sort().Check(
 				// db, tbl, part, col, is_index, bucket_id, count, repeats, lower, upper, ndv
 				testkit.Rows("test t global a 0 0 5 2 1 4 0",
-					"test t global a 0 1 12 2 17 17 0",
+					"test t global a 0 1 12 2 11 17 0",
```
winoros, line 1432: image
Expected change.

pkg/executor/test/analyzetest/analyze_test.go
```diff
@@ -1429,9 +1429,9 @@ func TestAnalyzeColumnsWithDynamicPartitionTable(t *testing.T) {
...
 					"test t global c 0 0 6 1 2 6 0",
-					"test t global c 0 1 14 2 13 13 0",
+					"test t global c 0 1 14 2 7 13 0",
```
winoros, line 1434: image

expected change

pkg/statistics/histogram_test.go
```diff
@@ -153,27 +154,20 @@ func TestMergePartitionLevelHist(t *testing.T) {
...
-					repeat: 3,
-					ndv:    5,
-				},
-				{
-					lower:  7,
-					upper:  11,
-					count:  13,
-					repeat: 3,
-					ndv:    3,
+					upper:  9,
```
winoros, line 157: The result becomes more balanced compared with previous.

pkg/statistics/handle/globalstats/global_stats_internal_test.go
```diff
@@ -351,10 +351,15 @@ func testIssues24349(testKit *testkit.TestKit) {
...
 	testKit.MustExec("analyze table t with 1 topn, 3 buckets")
+	testKit.MustQuery("show stats_topn where partition_name = 'global'").Sort().Check(testkit.Rows(
+		"test t global a 0 1 6",
+		"test t global b 0 2 4",
+	))
 	testKit.MustQuery("show stats_buckets where partition_name='global'").Check(testkit.Rows(
-		"test t global a 0 0 2 2 0 2 0",
-		"test t global b 0 0 3 1 1 2 0",
-		"test t global b 0 1 10 1 4 4 0",
+		"test t global a 0 0 4 4 0 0 0",
```
winoros, line 359: As you can see the TopN of column a is only the value 1. But column a has three values 0, 1, and 2.
So the added bucket is an expected one.

hawkingrei (approved):

pkg/statistics/histogram.go
```diff
@@ -1489,35 +1478,40 @@ func MergePartitionHist2GlobalHist(sc *stmtctx.StatementContext, hists []*Histog
...
+		return nil, err
 	}
 
 	var sum, prevSum int64
 	r, prevR := len(buckets), 0
 	bucketCount := int64(1)
 	gBucketCountThreshold := (totCount / expBucketNumber) * 80 / 100 // expectedBucketSize * 0.8
 	var bucketNDV int64
+	mergeBuffer := make([]*bucket4Merging, 0, (len(buckets)+int(expBucketNumber)-1)/int(expBucketNumber))
+	fullyInsideBuffer := make([]*bucket4Merging, 0, (len(buckets)+int(expBucketNumber))/int(expBucketNumber))
```
time-and-fate, line 1492: Looks like this variable is not used.

pkg/statistics/histogram.go
```diff
@@ -1489,35 +1478,40 @@ func MergePartitionHist2GlobalHist(sc *stmtctx.StatementContext, hists []*Histog
...
+			if err != nil {
+				return nil, err
+			}
+			if res > 0 {
+				currentLeftMost = buckets[i].lower
+			}
+		}
 		sum += buckets[i].Count
 		bucketNDV += buckets[i].NDV
 		if sum >= totCount*bucketCount/expBucketNumber && sum-prevSum >= gBucketCountThreshold {
```
time-and-fate, line 1509: Though it's not changed in this PR. Can we change it into `if !(original_check) { continue }` to make it simpler?

pkg/statistics/histogram.go
```diff
@@ -1528,10 +1522,85 @@ func MergePartitionHist2GlobalHist(sc *stmtctx.StatementContext, hists []*Histog
...
+				cutBkt := newBucket4Meging()
+				buckets[i-1].upper.Copy(cutBkt.upper)
+				currentLeftMost.Copy(cutBkt.lower)
+				cutBkt.Count = overlappedCount
+				cutBkt.NDV = overlappedNDV
+				mergeBuffer = append(mergeBuffer, cutBkt)
+				cutAndFixBuffer = append(cutAndFixBuffer, cutBkt)
+
+				currentLeftMost.Copy(buckets[i-1].upper)
+				buckets[i-1].Repeat = 0
```
time-and-fate, line 1580: Don't we need a `break` for the last case here?

pkg/statistics/histogram.go
winoros: Yes, i think this would be a better way.
My concern is that it may introduce more pressure on reviewing.

pkg/statistics/histogram.go
winoros: No, there'll be multiple buckets that meet the same requirements. `[small, currentLeftMost+1]`, `[small2, currentLeftMost+2]`, `[small3, currentLeftMost+x]`...

pkg/statistics/histogram.go
```diff
@@ -1528,10 +1521,83 @@ func MergePartitionHist2GlobalHist(sc *stmtctx.StatementContext, hists []*Histog
...
+				if err != nil {
+					return nil, err
+				}
+				// The content in the merge buffer don't need a re-sort since we just fix some lower bound for them.
+				mergeBuffer = append(mergeBuffer, buckets[origI:r]...)
+				merged, err = mergePartitionBuckets(sc, mergeBuffer)
+				for _, bkt := range cutAndFixBuffer {
+					releasebucket4MergingForRecycle(bkt)
+				}
+				i = origI
```
time-and-fate, line 1594: I think this `origI` will be wrong in some cases.
The "totally inside" and "overlap and cut" cases can happen together.
We need this `origI` to restore the `i` to its position before "overlap and cut", but currently `origI` is saved before handling "totally inside".
I think we need to save `origI` before the `continue`.

pkg/statistics/histogram.go
```diff
@@ -1542,40 +1608,48 @@ func MergePartitionHist2GlobalHist(sc *stmtctx.StatementContext, hists []*Histog
...
+			res, err := leftMost.Compare(sc.TypeCtx(), b.lower, collate.GetBinaryCollator())
+			if err != nil {
+				return nil, err
+			}
+			if res > 0 {
+				leftMost = b.lower
+			}
 		}
 
 		if len(globalBuckets) > 0 && bucketSum < gBucketCountThreshold { // merge them into the previous global bucket
```
time-and-fate, line 1626: I think this case will also be incorrect if "overlap and cut" happens. Because in this case, the `buckets` has already been modified and is incomplete.

pkg/statistics/histogram.go
```diff
@@ -1602,6 +1676,27 @@ func MergePartitionHist2GlobalHist(sc *stmtctx.StatementContext, hists []*Histog
 	return globalHist, nil
 }
 
+// sortBucketsByUpperBound the bucket by upper bound first, then by lower bound.
+// If bkt[i].upper = bkt[i+1].upper, then we'll get bkt[i].lower < bukt[i+1].lower.
```
time-and-fate, line 1680:
```suggestion
// If bkt[i].upper = bkt[i+1].upper, then we'll get bkt[i].lower < bkt[i+1].lower.
```

pkg/statistics/histogram.go
winoros: It's set at line#1510, which should be what you mean.

pkg/statistics/histogram.go
winoros: The fix is too long to introduce. I just removed this if block.

pkg/statistics/histogram.go
time-and-fate: Let me add an example:
| bucket idx   |      [lower,upper]      |  comments |
|:----------|:-------------|:------|
| ... |  ... |  |
|x+1|[10,35]| partial overlap (similar to x+3)|
|x+2|[35,40]| totally inside (this buckets should be directly added to `mergeBuffer`)|
|x+3|[20,45]| partial overlap <br/> (it should be cut, this buckets becomes [20,30], and a new bucket [30,45] added to `mergeBuffer`|
|x+4|[30,50]| threshold met when this bucket is added to `sum`|
| ... |  ... |  |

In this case:
- How should the "re-sort" happen?
- What should `origI`, which is also `r` for the next iteration be?

pkg/statistics/histogram.go
```diff
@@ -1507,56 +1498,139 @@ func MergePartitionHist2GlobalHist(sc *stmtctx.StatementContext, hists []*Histog
...
+				//    After cutting, the remained part is [lower_bound_i, currentLeftMost]
+				// To do the next round of merging, we need to kick out the 1st kind of buckets.
+				// And after the re-sort, the 2nd kind of buckets will be in the front.
+				leftMostInvalidPosForNextRound := leftMostValidPosForNonOverlapping
+				for ; leftMostInvalidPosForNextRound > i; leftMostInvalidPosForNextRound-- {
+					res, err := buckets[leftMostInvalidPosForNextRound-1].lower.Compare(sc.TypeCtx(), currentLeftMost, collate.GetBinaryCollator())
+					if err != nil {
+						return nil, err
+					}
+					// Once the lower bound < currentLeftMost, we'll skip all the 1st kind of bucket.
```
time-and-fate, line 1589:
```suggestion
					// Once the lower bound < currentLeftMost, we've skipped all the 1st kind of bucket.
```

pkg/statistics/histogram.go
```diff
@@ -1507,56 +1498,139 @@ func MergePartitionHist2GlobalHist(sc *stmtctx.StatementContext, hists []*Histog
...
+					// We can break here.
+					if res < 0 {
+						break
```
time-and-fate, line 1592: Since this part of logic is very complicated, I suggest we add an assert here to make sure `upper == currentLeftMost` is true for all buckets in `[i,leftMostInvalidPosForNextRound-1]`.

winoros: good idea, added.

time-and-fate (approved):

PR #52225, opened 2024-03-29
util/ranger: Handle boundary value correctly in ranger to avoid incorrect tableDual plan 

### LINKED ISSUE #50051: incorrect tableDual plan which would cause no result
## Bug Report

``` SQL
CREATE TABLE `t5cb5d4a7` (   `col_12` binary(172) NOT NULL,   `col_13` bigint(20) unsigned NOT NULL,   KEY `idx_2` (`col_12`,`col_13`),   PRIMARY KEY (`col_13`,`col_12`)  ,   KEY `idx_5` (`col_13`,`col_12`) ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci;
desc SELECT 1 FROM `t5cb5d4a7` WHERE `t5cb5d4a7`.`col_13`>9223372036854775807 AND `t5cb5d4a7`.`col_13`>8900084226548014260;
```

TableRangeScan

``` SQL
mysql> desc SELECT 1 FROM `t5cb5d4a7` WHERE `t5cb5d4a7`.`col_13`>9223372036854775807 AND `t5cb5d4a7`.`col_13`>8900084226548014260;
+-------------------+---------+------+---------------+---------------+
| id                | estRows | task | access object | operator info |
+-------------------+---------+------+---------------+---------------+
| Projection_4      | 3333.33 | root |               | 1->Column#3   |
| └─TableDual_5     | 3333.33 | root |               | rows:0        |
+-------------------+---------+------+---------------+---------------+
2 rows in set (0.00 sec)

mysql> desc SELECT 1 FROM `t5cb5d4a7` WHERE `t5cb5d4a7`.`col_13`>9223372036854775807;
+--------------------------+---------+-----------+-----------------+------------------------------------------------------------------+
| id                       | estRows | task      | access object   | operator info                                                    |
+--------------------------+---------+-----------+-----------------+------------------------------------------------------------------+
| Projection_4             | 3333.33 | root      |                 | 1->Column#3                                                      |
| └─TableReader_6          | 3333.33 | root      |                 | data:TableRangeScan_5                                            |
|   └─TableRangeScan_5     | 3333.33 | cop[tikv] | table:t5cb5d4a7 | range:(9223372036854775807,+inf], keep order:false, stats:pseudo |
+--------------------------+---------+-----------+-----------------+------------------------------------------------------------------+
3 rows in set (0.00 sec)

mysql> desc SELECT 1 FROM `t5cb5d4a7` WHERE `t5cb5d4a7`.`col_13`>8900084226548014260;
+--------------------------+---------+-----------+-----------------+------------------------------------------------------------------+
| id                       | estRows | task      | access object   | operator info                                                    |
+--------------------------+---------+-----------+-----------------+------------------------------------------------------------------+
| Projection_4             | 3333.33 | root      |                 | 1->Column#3                                                      |
| └─TableReader_6          | 3333.33 | root      |                 | data:TableRangeScan_5                                            |
|   └─TableRangeScan_5     | 3333.33 | cop[tikv] | table:t5cb5d4a7 | range:(8900084226548014260,+inf], keep order:false, stats:pseudo |
+--------------------------+---------+-----------+-----------------+------------------------------------------------------------------+
3 rows in set (0.00 sec)
```

33480e8c8d8874811a5362767c81d40c9a057cb6

onlyacat:

Issue Number: close #50051 

Problem Summary:

The function `excludeToIncludeForIntPoint` should deal with the boundary value correctly.

I guess it is not a bad idea to change the type of the point in this case.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


onlyacat:
PTAL @wjhuang2016 @fixdb

pkg/util/ranger/detacher.go
```diff
@@ -487,7 +487,9 @@ func (d *rangeDetacher) detachCNFCondAndBuildRangeForIndex(conditions []expressi
 // excludeToIncludeForIntPoint converts `(i` to `[i+1` and `i)` to `i-1]` if `i` is integer.
 // For example, if p is `(3`, i.e., point { value: int(3), excl: true, start: true }, it is equal to `[4`, i.e., point { value: int(4), excl: false, start: true }.
 // Similarly, if p is `8)`, i.e., point { value: int(8), excl: true, start: false}, it is equal to `7]`, i.e., point { value: int(7), excl: false, start: false }.
-// If return value is nil, it means p is unsatisfiable. For example, `(MaxInt64` is unsatisfiable.
+// If return value is nil, it means p is unsatisfiable. For example, `(MaxUint64` is unsatisfiable.
+// The boundary value will be treated as the bigger type: For example, `(MaxInt64` of type KindInt64 will become `[MaxInt64+1` of type KindUint64,
+// and vice versa for `0)` of type KindUint64 will become `-1]` of type KindInt64.
 func excludeToIncludeForIntPoint(p *point) *point {
```
winoros, line 493: Seems that this function is only used to check whether the range is a point or not.
Why it will affect table dual or not?

onlyacat: > Seems that this function is only used to check whether the range is a point or not. Why it will affect table dual or not?

This function will convert the point to include mode. For example, input `c>9223372036854775807` will been changed to `c>=9223372036854775808`.
If an nil is returned, it means the condition is unsatisfiable. 

Hence the condition `c>9223372036854775807` with `c` (unsigned bigint) will return nil because of the bug.

This func is invoked in `allSinglePoints`. If nil is returned, it will simply skip this range(ignore unsatisfiable condition), the pos does not increase, which returns an empty slice.

https://github.com/pingcap/tidb/blob/d2eb9026a476dd76a3fbb9449834867c4ad2a87b/pkg/util/ranger/detacher.go#L533-L539

Function `allSinglePoints` has only been called inside `ExtractEqAndInCondition`, here an empty silce returned means the range is invalid and finally cause the TableDual Outside.

https://github.com/pingcap/tidb/blob/d2eb9026a476dd76a3fbb9449834867c4ad2a87b/pkg/util/ranger/detacher.go#L661-L670

wjhuang2016: If there exists an interval whose length is large than 0, return nil.
I think it's ok to return nil here according to the comment.

wjhuang2016 (approved):

pkg/util/ranger/detacher.go
winoros: So this change will make the func `excludeToIncludeForIntPoint` impossible to return `nil`. I think the the codes in `allSinglePoints` should also be changed.

And the actual issue might be that: The range point may be Signed Int64 while the column is Unsigned Int64 or vice versa?

pkg/util/ranger/detacher.go
```diff
@@ -487,7 +487,9 @@ func (d *rangeDetacher) detachCNFCondAndBuildRangeForIndex(conditions []expressi
 // excludeToIncludeForIntPoint converts `(i` to `[i+1` and `i)` to `i-1]` if `i` is integer.
 // For example, if p is `(3`, i.e., point { value: int(3), excl: true, start: true }, it is equal to `[4`, i.e., point { value: int(4), excl: false, start: true }.
 // Similarly, if p is `8)`, i.e., point { value: int(8), excl: true, start: false}, it is equal to `7]`, i.e., point { value: int(7), excl: false, start: false }.
-// If return value is nil, it means p is unsatisfiable. For example, `(MaxInt64` is unsatisfiable.
+// If return value is nil, it means p is unsatisfiable. For example, `(MaxUint64` is unsatisfiable.
+// The boundary value will be treated as the bigger type: For example, `(MaxInt64` of type KindInt64 will become `[MaxInt64+1` of type KindUint64,
```
winoros, line 491: If so, the comment should also be updated. Or we pass the column's field type to this function, to do a more accurate judgment.

wjhuang2016: @onlyacat Could you check this review？

pkg/util/ranger/detacher.go
onlyacat: > So actual issue might be that: The range point may be Signed Int64 while the column is Unsigned Int64 or vice versa?

Yes.

pkg/util/ranger/detacher.go
onlyacat: > If so, the comment should also be updated. Or we pass the column's field type to this function, to do a more accurate judgment.

> @onlyacat Could you check this review？

Sorry for the late reply. 
Since this is just an early termination step, I guess it is not a bad idea to keep it simple and the unsatisfied value for specific field type will be eliminated in the later stage.
As for the comment, it described what I edited as actual. Not sure which should be updated.

winoros (approved):

Issue #51533, opened 2024-03-06
Unify the "split client" package for BR and lightning


lance6716:
## Enhancement

now BR and lightning have separate functions to do one thing. Better to unify them

okJiang:
Can you tell me which modules still need to be unified for this issue? @lance6716

lance6716:
> Can you tell me which modules still need to be unified for this issue? @lance6716

Do you mean why this issue is not finished? Reasons:

1. Currently we don't have enough test clsuter to verify it
2. I'm considering if the "split client" should have built-in coarse-grained + finer-grained two level split, don't let caller bother it.

PR #52235, opened 2024-03-29
br,lightning: refine GetSplitKeysOfRegions

lance6716:

Issue Number: ref #51533

Problem Summary:

- GetSplitKeysOfRegions returns region as map key, so no need to maintaion another map from region ID to region

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
   lightning CI https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fpull_lightning_integration_test/detail/pull_lightning_integration_test/144/pipeline
   br CI https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fpull_br_integration_test/detail/pull_br_integration_test/89/pipeline
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


br/pkg/restore/split/split.go
```diff
@@ -288,34 +288,88 @@ func (b *BackoffMayNotCountBackoffer) Attempt() int {
...
+			//
+			// key2 don't need to be split on. Considering the original region, we can have
+			// [key1, a), [a, b), [b, key3) rather than [key1, a), [a, key2), [key2, b), [b,
+			// key3)
+
+			if moveRegionCnt > 0 {
+				clear(pendingSplitKeyMap)
+				// save key2 in pendingSplitKeyMap. When we move to key3 we can check
+				// moveRegionCnt again to know if key2 can be skipped.
+				pendingSplitKeyMap[region] = append(pendingSplitKeyMap[region], sortedKeys[curKeyIndex])
```
D3Hunter, line 343: seems no need to be a map, its length is <= 1 always and key is always current region

br/pkg/restore/split/split.go
```diff
@@ -288,34 +288,88 @@ func (b *BackoffMayNotCountBackoffer) Attempt() int {
...
 			}
-			regionID := region.Region.GetId()
-			splitKeyMap[regionID] = append(splitKeyMap[regionID], sortedKeys[curKeyIndex])
+
+			//   key1      key2    key3
+			//    |    a    |    b  |
+			// ---+----+----+----+--+-----------
+			// region1 | region2 | region3 | region4
+			//
+			// key2 don't need to be split on. Considering the original region, we can have
```
D3Hunter, line 335: if data cluster around key2, we would have a large region

D3Hunter: based on this, seems we don't need split on `15` for this:
regions: (, 10), [10, 100), [100, 200), [200,)
keys:  11, 15, 111, 121

br/pkg/restore/split_test.go
```diff
@@ -202,7 +202,14 @@ func TestSplitAndScatter(t *testing.T) {
 	err := regionSplitter.ExecuteSplit(ctx, ranges, 1, false, func(key [][]byte) {})
 	require.NoError(t, err)
 	regions := client.GetAllRegions()
-	if !validateRegions(regions) {
+	// old regions: [, aay), [aay, bba), [bba, bbh), [bbh, cca), [cca, )
+	// ranges: [bbd, bbf), [bbf, bbj), [xxa, xxe), [xxe, xxz)
+	// should split on: bbf, xxa, xxe, xxz,
```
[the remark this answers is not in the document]
lance6716: Yes I think we should not split on the first/last key. However caller should use `keys[1:len(keys)-1]`,`keys[:]` or other form as the argument of this function. And this function does not care about it. 

In BR's usage, I see the first key is not included in "split keys", but the last key is included. BR always has non-overlapping split keys, so for the two clients with key range [a, b), [b, c) and [c, d), [d, e), the first client will split on b, c and the second d, e. The result regions are even. Do I get it correctly? @Leavrth 

For lightning global sort usage, we should also mimic the above rule and passing keys[1:]

For lighnting local sort usage, we don't know how data is overlapped, I plan to just port the old logic.

br/pkg/restore/split/split.go
lance6716: > if data cluster around key2, we would have a large region

yes, but I'm just porting the old logic. I'll check it and provide the reference to the old logic soon

> we don't need split on 15 for this

Current rule is we skip 15 only if [11, 15) and [15, 111) are both split. But [11, 15) is not. Even if data is clustered around 15 like they are mostly located in [14.9, 15.1), but because the upper layer choose 15 it means [11, 15) and [15, 111) and ~96MB data so we should also split 15

D3Hunter: yes, but the regions already splitted on `100`, it met the case described in the comment

lance6716: [11, 15) is not split, nearest region is 10 and 100 which are not inside [11, 15)

lance6716: >  I'll check it and provide the reference to the old logic soon

I found I understand the old logic wrong 😂 In old logic, we only skip the ranges which are the exactly same as the region.
https://github.com/pingcap/tidb/blob/8ba1fa452b1ccdbfb85879ea94b9254aabba2916/br/pkg/lightning/backend/local/localhelper.go#L150-L165

I'll update this PR

br/pkg/lightning/backend/local/localhelper.go
```diff
@@ -143,37 +141,12 @@ func (local *Backend) SplitAndScatterRegionByRanges(
...
-			idx := sort.Search(len(regions), func(i int) bool {
-				return beforeEnd(startKey, regions[i].Region.EndKey)
-			})
-			if idx < 0 || idx >= len(regions) {
-				log.FromContext(ctx).Error("target region not found", logutil.Key("start_key", startKey),
-					logutil.RegionBy("first_region", regions[0].Region),
-					logutil.RegionBy("last_region", regions[len(regions)-1].Region))
-				return errors.New("target region not found")
-			}
-			if bytes.Compare(startKey, regions[idx].Region.StartKey) > 0 || bytes.Compare(endKey, regions[idx].Region.EndKey) < 0 {
```
lance6716, line 161: deleted because it only removes the range which has the same start key and end key with the region. `GetSplitKeysOfRegions` can also do it.

br/pkg/lightning/backend/local/localhelper.go
```diff
@@ -138,42 +134,12 @@ func (local *Backend) SplitAndScatterRegionByRanges(
 		log.FromContext(ctx).Info("paginate scan region finished", logutil.Key("minKey", minKey), logutil.Key("maxKey", maxKey),
 			zap.Int("regions", len(regions)))
 
-		if !needSplit {
```
lance6716, line 141: if not needSplit, we don't need to scatter it.

D3Hunter (approved):

br/pkg/restore/split/split.go
```diff
@@ -288,34 +288,58 @@ func (b *BackoffMayNotCountBackoffer) Attempt() int {
...
+			curKeyIndex++
+			if curKeyIndex >= len(sortedKeys) {
+				return splitKeyMap
+			}
+			splitKey = codec.EncodeBytesExt(nil, sortedKeys[curKeyIndex], isRawKV)
 		}
 	}
+	lastKey := sortedKeys[len(sortedKeys)-1]
+	endOfLastRegion := sortedRegions[len(sortedRegions)-1].Region.GetEndKey()
+	if !bytes.Equal(lastKey, endOfLastRegion) {
```
Leavrth, line 336: I think the `lastKey` can be less than the `endOfLastRegion`.

Leavrth (approved):

Issue #52306, opened 2024-04-02
Optimization for lightning conflict detection


lyzx2001:
## Enhancement
1. Merge conflict record tables for preprocess duplicate detection and post-import conflict detection.
2. Performance optimization for post-import conflict detection.

Merge lightning preprocess duplicate detection and post-import conflict detection:
https://github.com/pingcap/tidb/issues/51036

Optimize lightning conflict recording tables:
https://github.com/pingcap/tidb/issues/52393

PR #52307, opened 2024-04-02
lightning: merge conflict record tables for preprocess duplicate detection and post-import conflict detection

lyzx2001:

Issue Number: ref #52306
ref #52393 

Problem Summary:

Merge conflict record tables for preprocess duplicate detection and post-import conflict detection.
### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/lightning/errormanager/errormanager.go
```diff
@@ -83,37 +85,51 @@ const (
...
+			row_id 	             bigint NOT NULL COMMENT 'the row id of the conflicted row',
+			row_data             text NOT NULL COMMENT 'the row data of the conflicted row',
+			INDEX (task_id, table_name),
+			INDEX (is_precheck_conflict)
 		);
 	`
 
+	createConflictView = `
+    	CREATE OR REPLACE VIEW  %s.` + ConflictViewName + `
+			AS SELECT is_precheck_conflict, task_id, create_time, table_name, index_name, key_data, row_data,
```
D3Hunter, line 125: can we use `select 0 as is_precheck_conflict, ..... from xxx`? so we don't have maintain it in each table

D3Hunter (commented):
14/18

pkg/lightning/errormanager/errormanager.go
lyzx2001: modified

pkg/lightning/config/config.go
```diff
@@ -82,7 +82,7 @@ const (
 	// defaultMetaSchemaName is the default database name used to store lightning metadata
 	defaultMetaSchemaName     = "lightning_metadata"
 	defaultTaskInfoSchemaName = "lightning_task_info"
-	defaultMaxRecordRows      = 100
+	DefaultThreshold          = 10000
```
lance6716, line 85: name it DefaultRecordDuplicateThreshold or something, to make its usage more obvious

lightning/tidb-lightning.toml
```diff
@@ -97,17 +97,19 @@ driver = "file"
...
 # - "ignore": When encountering conflicting primary or unique key records, TiDB Lightning retains the old data and ignores the new data. This option can only be used in the logical import mode.
 strategy = ""
 # Controls whether to enable preprocess conflict detection, which check conflicts in the data before importing it to TiDB. In scenarios where the ratio of conflict records is greater than or equal to 1%, it is recommended to enable preprocess conflict detection for better performance in conflict detection.
 # In other scenarios, it is recommended to disable it. The default value is false, indicating that TiDB Lightning only checks conflicts after the import. If you set it to true, TiDB Lightning checks conflicts both before and after the import. This parameter is experimental, and it can be used only in the physical import mode.
 # precheck-conflict-before-import = false
-# Controls the maximum number of conflict errors that can be handled when strategy is "replace" or "ignore". You can set it only when strategy is "replace" or "ignore". The default value is 9223372036854775807, which means that almost all errors are tolerant.
-# threshold = 9223372036854775807
```
lance6716, line 107: I remember the reason of separate `threshold` and `max-record-rows` is the user wants the import task success for the first time, but the user did not know how many duplicate records in it. So `threshold` should be set to a large value.

But if we record too many rows in the conflict table, the performance is not affordable, so there's another configuration `max-record-rows`.

How to resolve this case now?

lyzx2001: This change was determined by PM. The default `threshold`/`max-record-rows` is now set to 10000. In most scenarios, if the user wants to import data with more than 10000 duplicate records, there is probably something wrong with the data source, and the previous experience is that maybe the configuration is not set correctly, or the downstream table structure pk or uk is not well defined. So there is no meaning to keep importing although not further recording conflict data.
The change is the first step that we are gradually deprecating `max-record-rows`, which is to directly set the value of `max-record-rows` to be that of `threshold`.

pkg/lightning/config/config.go
```diff
@@ -1381,45 +1381,30 @@ func (c *Conflict) adjust(i *TikvImporter, l *Lightning) error {
 		case ErrorOnDup:
 			c.Threshold = 0
 		case IgnoreOnDup, ReplaceOnDup:
-			c.Threshold = math.MaxInt64
+			c.Threshold = DefaultRecordDuplicateThreshold
 		case NoneOnDup:
```
lance6716, line 1385: can merge the case with ErrorOnDup

```
case ErrorOnDup, NoneOnDup:
```

pkg/lightning/config/config.go
```diff
@@ -1381,45 +1381,30 @@ func (c *Conflict) adjust(i *TikvImporter, l *Lightning) error {
...
-			maxAccepted = c.Threshold
-		}
-		if c.Strategy == ReplaceOnDup && i.Backend == BackendTiDB {
-			// due to we use batch insert, we can't know which row is duplicated.
-			maxAccepted = 0
+	if c.Strategy == ReplaceOnDup && i.Backend == BackendTiDB {
+		// due to we use batch insert, we can't know which row is duplicated.
+		if c.MaxRecordRows >= 0 {
+			// only warn when it is set by user.
+			log.L().Warn(`Cannot record duplication (conflict.max-record-rows > 0) when use tikv-importer.backend = \"tidb\" and conflict.strategy = \"replace\".
```
lance6716, line 1398: this behaviour also changed. In the old behaviour tidb backend + replace will see error if set max-record-rows, the error will tell the user can't find the error details in conflict table in advance. Now the user will not see the error. Please double check with PM.

pkg/lightning/errormanager/errormanager.go
```diff
@@ -114,6 +116,16 @@ const (
 		);
 	`
 
+	createConflictView = `
+    	CREATE OR REPLACE VIEW  %s.` + ConflictViewName + `
```
lance6716, line 120:
```suggestion
    	CREATE OR REPLACE VIEW %s.` + ConflictViewName + `
```

pkg/lightning/errormanager/errormanager.go
```diff
@@ -269,6 +281,14 @@ func (em *ErrorManager) Init(ctx context.Context) error {
 		}
 	}
 
+	if em.conflictV1Enabled && em.conflictV2Enabled {
```
lance6716, line 284: should we always tell the user one VIEW to find the duplicate error, no matter how lightning is configurated?

pkg/lightning/errormanager/errormanager.go
```diff
@@ -269,6 +281,14 @@ func (em *ErrorManager) Init(ctx context.Context) error {
...
+		sql := [2]string{"create conflict view", createConflictView}
```
lance6716, line 285: no need to use this variable? Directly `exec.Exec(ctx, "create conflict view", common.SprintfWithIdentifiers(createConflictView, em.schema, em.schema, em.schema))`

pkg/lightning/errormanager/errormanager.go
lyzx2001: I think it is better that we only use VIEW when there are data in both tables. For the cases where the table contain a huge amount of data, using VIEW may slow down the performance when the result is exactly the same with not using VIEW.
It will be clearly stated in the log which table or view to check the conflict data.

br/tidb-lightning.toml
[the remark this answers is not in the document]
lyzx2001: Do we also need to update the approve privilege for this file?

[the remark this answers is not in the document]
D3Hunter: the dir already have one

pkg/lightning/errormanager/errormanager.go
lance6716: this will increase the complexity to user, also check it with PM. And for the performance problem I think 10000 records will not change response time a lot, you can test it afterwards.

br/tidb-lightning.toml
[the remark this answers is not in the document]
lyzx2001: https://github.com/pingcap/tidb/blob/69d7770335a7611d05bb2f3a0888c79afb03e92a/br/OWNERS#L4-L10
in the old directory we have different privilege for `tidb-lightning.toml`?

pkg/lightning/config/config.go
lyzx2001: PM confirmed that the user will no longer see this error. Instead there will be a warning in the log

pkg/lightning/errormanager/errormanager.go
lyzx2001: PM said the current implementation is acceptable, but it is better that we use VIEW to return to users regardless of the configuration in the later optimization. Will revise this in the next PR

br/tidb-lightning.toml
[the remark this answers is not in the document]
D3Hunter: o, we can remove it here

lance6716 (approved):

D3Hunter (approved):

BornChanger (approved):

tangenta (approved):

br/OWNERS
```diff
@@ -1,10 +1,5 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 options:
   no_parent_owners: true
-filters:
-  "(tidb-lightning\\.toml)$":
-    approvers:
-      - sig-critical-approvers-tidb-lightning
```
wuhuizuo, line 7: @lyzx2001 you forgot to move this part into `lightning/OWNERS`.

lyzx2001: done
https://github.com/pingcap/tidb/pull/52745

PR #52363, opened 2024-04-07
executor, planner: minimize the context used by the `ranger` package

### LINKED ISSUE #52362: Minimize the context used in `ranger`
## Enhancement

We use a whole `PlanContext` to access the `ranger`. However, the executors will also use ranger functions at any time (after already building a plan). Therefore, it'd be better to split a standalone context used by ranger.

YangKeao:

Issue Number: close #52362

Problem Summary:

The `tableReaderExecutorCtx` depends on a whole `pctx` and its session variables, which is not easy to copy. However, it only uses ranger related functions and `ToPB` related functions.

This PR adds a standalone context for `ranger` package, to make it easier to copy it.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > 
  > Should be guarded by existing tests.

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/context/context.go
```diff
@@ -72,3 +73,20 @@ type EmptyPlanContextExtended struct{}
 
 // AdviseTxnWarmup advises the txn to warm up.
 func (EmptyPlanContextExtended) AdviseTxnWarmup() error { return nil }
+
+// GetRangerCtx creates a RangerContext from PlanContext
+func GetRangerCtx(sctx PlanContext) *rangerctx.RangerContext {
```
lcwangchao, line 78: We use `GetRangerCtx` instead of giving a method `PlanContext.GetRangerCtx` because dependency cycle?

pkg/util/ranger/context/context.go
```diff
@@ -0,0 +1,43 @@
...
+// RangeFallbackHandler is used to handle range fallback.
+// If there are too many ranges, it'll fallback and add a warning.
+type RangeFallbackHandler interface {
+	RecordRangeFallback(rangeMaxSize int64)
+	SetSkipPlanCache(err error)
+	contextutil.WarnHandler
+}
+
+// RangerContext is the context used to build range.
+type RangerContext struct {
```
lcwangchao, line 33: Could make the inner fields private to avoid some unexpected modifications?

lcwangchao (commented):
LGTM, with only some tiny comments. But I think it's fine to merge

lcwangchao (approved):

pkg/planner/context/context.go
YangKeao: No. Maybe it'd be better to add a method `PlanContext.GetRangerCtx`. I used `GetRangerCtx` because it only depends on the public interface defined in `PlanContext`, but now I think it'd be better to add a method because maybe `GetSessionVars()` (etc) will be removed finally 🤔 .

I'll do it right now.

YangKeao: I think I can add this method in `PlanCtxExtendedImpl`. It looks cool 🤔

YangKeao: Done. PTAL

YangKeao: Done. PTAL

hawkingrei (approved):

PR #52372, opened 2024-04-07
distsql, bench: fix the benchmark of distsql to represent the performance more accurately

### LINKED ISSUE #52371: Using `(*mock.Context).GetDistSQLCtx()` will make the benchmark slower
## Enhancement

This issue only affects the benchmark, and it doesn't affect the performance of real `tidb-server`. I'd like to fix it to make the performance of benchmark aligned with the real tidb server.

The real session cache the `distsqlctx`. We should also cache the `distsqlctx` in the `mockResponse`.

YangKeao:

Issue Number: close #52371

Problem Summary:

The original benchmark didn't cache the `distsqlctx`, which is not aligned with the behavior of real session.

```
Before f7d66f:
BenchmarkSelectResponseChunk_SmallResponse-16    	    2487	    457227 ns/op	  624643 B/op	    5112 allocs/op

Before:
BenchmarkSelectResponseChunk_SmallResponse-16    	    2366	    495870 ns/op	  704644 B/op	    5512 allocs/op

After:
BenchmarkSelectResponseChunk_SmallResponse-16    	    2430	    447747 ns/op	  624644 B/op	    5112 allocs/op
```

This PR caches the `distsqlctx` in the `mockResponse`, so that it'll not be re-created multiple times.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```

### DIFF
--- a/pkg/distsql/distsql_test.go
+++ b/pkg/distsql/distsql_test.go
@@ -20,6 +20,7 @@ import (
 	"testing"
 	"time"
 
+	distsqlctx "github.com/pingcap/tidb/pkg/distsql/context"
 	"github.com/pingcap/tidb/pkg/kv"
 	"github.com/pingcap/tidb/pkg/parser/charset"
 	"github.com/pingcap/tidb/pkg/parser/mysql"
@@ -205,7 +206,7 @@ type mockResponse struct {
 	count int
 	total int
 	batch int
-	ctx   sessionctx.Context
+	ctx   *distsqlctx.DistSQLContext
 	sync.Mutex
 }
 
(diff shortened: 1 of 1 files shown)


hawkingrei (approved):

hawkingrei:
@YangKeao Please fix the unit test.

zimulala (approved):

YangKeao:
Close this PR, as it needs too much modification on tests, and it only affects the performance of the test. If we use the benchmark result of new codes as a baseline, I think it's fine to keep it.

PR #52427, opened 2024-04-08
planner, statistics: fix the inconsistent est when table has no stats

### LINKED ISSUE #52426: Inconsistent row count estimation between lite stats init and non-lite init when table has no stats
## Bug Report

create a small table and insert some rows
```
create table t(a int);
insert into t values(1), (2), (3), (4), (5), (6), (7), (8), (9), (10);
```

Wait for a while after the stats delta is updated.
image

Then run `explain select * from t where a in (1, 2, 3, 4, 5, 6,7,8)` with and without setting `lite-init-stats` in config to false(default is true).

Estimation is the same, both is `0.08` while the full scan is `10`. (0.08 = 10 * 1/1000 * 8)

The lite init got 
image
8 = 10 * 0.8(defaultSelectivityFactor)

current master.

winoros:

Issue Number: close #52426 

Problem Summary:

The `Selectivity` needs a memory obj to trigger the pseudo estimation.

We add one.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/collect_column_stats_usage.go
+++ b/pkg/planner/core/collect_column_stats_usage.go
@@ -195,11 +195,21 @@ func (c *columnStatsUsageCollector) addHistNeededColumns(ds *DataSource) {
 	colIDSet := intset.NewFastIntSet()
 
 	for _, col := range columns {
+		// If the column is plan-generated one, Skip it.
+		// TODO: we may need to consider the ExtraHandle.
+		if col.ID < 0 {
+			continue
+		}
 		tblColID := model.TableItemID{TableID: ds.physicalTableID, ID: col.ID, IsIndex: false}
 		colIDSet.Insert(int(col.ID))
 		c.histNeededCols[tblColID] = true
 	}
 	for _, col := range ds.Columns {
+		// If the column is plan-generated one, Skip it.
+		// TODO: we may need to consider the ExtraHandle.
+		if col.ID < 0 {
+			continue
+		}
 		if !colIDSet.Has(int(col.ID)) && !col.Hidden {
 			tblColID := model.TableItemID{TableID: ds.physicalTableID, ID: col.ID, IsIndex: false}
 			if _, ok := c.histNeededCols[tblColID]; ok {
--- a/pkg/statistics/handle/storage/read.go
+++ b/pkg/statistics/handle/storage/read.go
@@ -572,7 +572,9 @@ func CleanFakeItemsForShowHistInFlights(statsCache statstypes.StatsCache) int {
 		if item.IsIndex {
 			_, loadNeeded = tbl.IndexIsLoadNeeded(item.ID)
 		} else {
-			_, loadNeeded = tbl.ColumnIsLoadNeeded(item.ID, true)
+			var analyzed bool
+			_, loadNeeded, analyzed = tbl.ColumnIsLoadNeeded(item.ID, true)
+			loadNeeded = loadNeeded && analyzed
 		}
 		if !loadNeeded {
 			statistics.HistogramNeededItems.Delete(item)
@@ -589,8 +591,8 @@ func loadNeededColumnHistograms(sctx sessionctx.Context, statsCache statstypes.S
 		return nil
 	}
 	var colInfo *model.ColumnInfo
-	_, loadNeeded := tbl.ColumnIsLoadNeeded(col.ID, true)
-	if !loadNeeded {
+	_, loadNeeded, analyzed := tbl.ColumnIsLoadNeeded(col.ID, true)
+	if !loadNeeded || !analyzed {
 		statistics.HistogramNeededItems.Delete(col)
 		return nil
 	}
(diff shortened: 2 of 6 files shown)


pkg/statistics/handle/storage/read.go
```diff
@@ -572,7 +572,9 @@ func CleanFakeItemsForShowHistInFlights(statsCache statstypes.StatsCache) int {
 		if item.IsIndex {
 			_, loadNeeded = tbl.IndexIsLoadNeeded(item.ID)
 		} else {
-			_, loadNeeded = tbl.ColumnIsLoadNeeded(item.ID, true)
```
winoros, line 575: We keep the async load unchanged. Wait for later patch to handle it.

hawkingrei (approved):

pkg/planner/core/collect_column_stats_usage.go
```diff
@@ -195,11 +195,21 @@ func (c *columnStatsUsageCollector) addHistNeededColumns(ds *DataSource) {
 	colIDSet := intset.NewFastIntSet()
 
 	for _, col := range columns {
+		// If the column is plan-generated one, Skip it.
```
time-and-fate, line 198: What does "plan-generated" mean?

pkg/planner/core/collect_column_stats_usage.go
```diff
@@ -195,11 +195,21 @@ func (c *columnStatsUsageCollector) addHistNeededColumns(ds *DataSource) {
...
+		// TODO: we may need to consider the ExtraHandle.
+		if col.ID < 0 {
+			continue
+		}
 		tblColID := model.TableItemID{TableID: ds.physicalTableID, ID: col.ID, IsIndex: false}
 		colIDSet.Insert(int(col.ID))
 		c.histNeededCols[tblColID] = true
 	}
 	for _, col := range ds.Columns {
+		// If the column is plan-generated one, Skip it.
```
time-and-fate, line 208: ditto.

pkg/statistics/table.go
```diff
@@ -629,25 +629,37 @@ func (t *Table) GetStatsHealthy() (int64, bool) {
 // The Column should be visible in the table and really has analyzed statistics in the stroage.
 // Also, if the stats has been loaded into the memory, we also don't need to load it.
 // We return the Column together with the checking result, to avoid accessing the map multiple times.
-func (t *Table) ColumnIsLoadNeeded(id int64, fullLoad bool) (*Column, bool) {
+// The first bool is whether we have it in memory. The second bool is whether this column has stats or not.
```
time-and-fate, line 632: I think "has stats" is a bit ambiguous.
```suggestion
// The first bool is whether we have it in memory. The second bool is whether this column has stats in the system table or not.
```

pkg/statistics/table.go
```diff
@@ -629,25 +629,37 @@ func (t *Table) GetStatsHealthy() (int64, bool) {
...
 	hasAnalyzed := t.ColAndIdxExistenceMap.HasAnalyzed(id, false)
 
-	// If it's not analyzed yet. Don't need to load it.
+	// If it's not analyzed yet.
 	if !hasAnalyzed {
-		return nil, false
+		// If we don't have it in memory, we create a fake hist for pseudo estimation.
+		if !ok {
+			// If we don't have this column. We skip it.
+			// It's something ridiculous. But it's possible that the stats don't have some ColumnInfo.
```
time-and-fate, line 645: So this means this column has stats but doesn't have `ColumnInfo`.
In what case would this happen?

pkg/statistics/table.go
```diff
@@ -629,25 +629,37 @@ func (t *Table) GetStatsHealthy() (int64, bool) {
...
+		return nil, false, false
+	}
 	col, ok := t.Columns[id]
 	hasAnalyzed := t.ColAndIdxExistenceMap.HasAnalyzed(id, false)
 
-	// If it's not analyzed yet. Don't need to load it.
+	// If it's not analyzed yet.
 	if !hasAnalyzed {
-		return nil, false
+		// If we don't have it in memory, we create a fake hist for pseudo estimation.
```
time-and-fate, line 642:
```suggestion
		// If we don't have it in memory, we create a fake hist for pseudo estimation (see handleOneItemTask()).
```

pkg/planner/core/collect_column_stats_usage.go
winoros: Such that ID=-1 means ExtraHandleCol

pkg/statistics/table.go
winoros: Such that `alter table add column` ddl event is not handled by the StatsHandle. Currently, the statistis.Table is initialized by the stats. not by the model.TableInfo

time-and-fate (approved):

PR #52433, opened 2024-04-09
br: use common scan-split-scatter logic

lance6716:

Issue Number: ref #51533

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
  BR CI https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fpull_br_integration_test/detail/pull_br_integration_test/102/pipeline
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


br/pkg/restore/split.go
```diff
@@ -125,14 +106,14 @@ func (rs *RegionSplitter) executeSplitByRanges(
 		roughSortedSplitKeys = append(roughSortedSplitKeys, sortedKeys[curRegionIndex])
 	}
 	if len(roughSortedSplitKeys) > 0 {
-		if err := rs.executeSplitByKeys(ctx, splitContext, scanStartKey, roughSortedSplitKeys); err != nil {
```
3pointer, line 128: `scanStartKey` is the start key. but `sortedKeys` are all end keys. if didn't scan with start key will cause error.

lance6716: we don't need to split the `scanStartKey` so theoretically we don't need to care about it. For the error I should have fixed in https://github.com/pingcap/tidb/blob/da775f61845ac93544019538bb6f48fb6482d346/br/pkg/restore/split/client.go#L541-L548

lance6716: update: I fixed a bug and add tests in https://github.com/pingcap/tidb/pull/52433/commits/59e6f0f9f3fe14494e18a48c2569f3f76fccba26

3pointer: Okay, We need make sure not use same key as start/end key to scan regions. or we will get an error from PD. I assume there is no change to get a   slice with only one nil/"" value in it.

br/pkg/restore/split.go
```diff
@@ -409,7 +338,7 @@ func (helper *LogSplitHelper) splitRegionByPoints(
 	}
 
 	helper.pool.ApplyOnErrorGroup(helper.eg, func() error {
-		newRegions, errSplit := regionSplitter.splitAndScatterRegions(ctx, region, splitPoints, false)
+		newRegions, errSplit := regionSplitter.splitAndScatterRegions(ctx, region, splitPoints)
```
Leavrth, line 341: Now it can directly call `regionSplitter.client.SplitWaitAndScatter` and the function `splitAndScatterRegions` can be removed.

Leavrth (approved):

3pointer (approved):

PR #52461, opened 2024-04-09
cop: ignore locks with the same start_ts

### LINKED ISSUE #52460: Pipelined DML cannot ignore its own locks when reading
## Bug Report

ref #50215 

Keep repeating this SQL (to make the dataset large enough and trigger region splits)
`insert into t select * from t`

No errors.

`Resolve lock timeout` error, which is reported by coprocessor.

master(ab90c771a21590a14c842e6dd1db03c998a9e3a2)

ekexium:

Issue Number: close #52460 

Problem Summary:

If this cop is not a stale read, which means its start_ts is fetched from PD, which means the start_ts cannot conflict with other transactions, add this ts to resolved_locks so that in a pipelined txn the locks belonging to the same txn will not block coprocessor.

For get and batch_get, they already put start_ts in resolved_locks.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below): repeating `insert into t select * from t` no longer reports `resolve lock timeout` error. I found it hard to be tested in unit test.
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/store/copr/coprocessor.go
+++ b/pkg/store/copr/coprocessor.go
@@ -190,6 +190,17 @@ func (c *CopClient) BuildCopIterator(ctx context.Context, req *kv.Request, vars 
 		buildTaskElapsed: *buildOpt.elapsed,
 		runawayChecker:   req.RunawayChecker,
 	}
+	// Pipelined-dml can flush locks when it is still reading.
+	// The coprocessor of the txn should not be blocked by itself.
+	// It should be the only case where a coprocessor can read locks of the same ts.
+	//
+	// But when start_ts is not obtained from PD,
+	// the start_ts could conflict with another pipelined-txn's start_ts.
+	// in which case the locks of same ts cannot be ignored.
+	// We rely on the assumption: start_ts is not from PD => this is a stale read.
+	if !req.IsStaleness {
+		it.resolvedLocks.Put(req.StartTs)
+	}
 	it.tasks = tasks
 	if it.concurrency > len(tasks) {
 		it.concurrency = len(tasks)
--- a/pkg/store/mockstore/unistore/cophandler/closure_exec.go
+++ b/pkg/store/mockstore/unistore/cophandler/closure_exec.go
@@ -1185,7 +1185,7 @@ func safeCopy(b []byte) []byte {
 }
 
 func checkLock(lock mvcc.Lock, key []byte, startTS uint64, resolved []uint64) error {
-	if isResolved(startTS, resolved) {
+	if isResolved(lock.StartTS, resolved) {
 		return nil
 	}
 	lockVisible := lock.StartTS < startTS


you06 (approved):

pkg/store/copr/coprocessor.go
```diff
@@ -190,6 +190,17 @@ func (c *CopClient) BuildCopIterator(ctx context.Context, req *kv.Request, vars
...
 	}
+	// Pipelined-dml can flush locks when it is still reading.
+	// The coprocessor of the txn should not be blocked by itself.
+	// It should be the only case where a coprocessor can read locks of the same ts.
+	//
+	// But when start_ts is not obtained from PD,
+	// the start_ts could conflict with another pipelined-txn's start_ts.
+	// in which case the locks of same ts cannot be ignored.
+	// We rely on the assumption: start_ts is not from PD => this is a stale read.
+	if !req.IsStaleness {
```
cfzjywxk, line 201: For `tidb_snapshot` read path, the `staleness` may not be set.

ekexium: This is a bit confusing. Is it intentional or an oversight? I suppose that `tidb_snapshot` is solely used for stale reads.

cfzjywxk: The `stale` flag is not set in the kv request if I do not remember wrong, there are two stale read implementaions in tidb and `tidb_snasphot` is the earlier one.

ekexium: Yes I've verified that. I mean if it's unintentional we can fix it by setting `IsStaleness` for this case.

cfzjywxk: Do you mean setting `IsStaleness` for `tidb_snaoshot` read?

ekexium: Yes

cfzjywxk: In theory, this is the way to proceed with the change. 
However, it's a breaking change because `tidb_snapshot` wasn't set before (setting it will trigger a check on `safe_ts`). Currently, it's unclear if other components rely on this behavior. 
We can proceed by merging this PR first.

cfzjywxk (approved):

pkg/store/copr/coprocessor.go
ekexium: Then it's possible that when `tidb_snapshot` is set to a ts conflicting with a pipelined txn, coprocessor ignores the locks while point_get doens't?

cfzjywxk: I think so though `tidb_snapshot` seems not to be used widely currenly.

PR #52511, opened 2024-04-11
point-get: calculate the row level checksum on the real time to avoid incorrect checksum caused by schema change

### LINKED ISSUE #52590: tidb_row_checksum() calculate the returned value on the real time instead of previous stored value
## Enhancement

At the moment, when use `tidb_row_checksum()` function return the checksum value which was calculated when the data inserted, this value may be stable since the table schema change.

```
create table t (a int primary key, b int);
insert into t values (1, 2);                         # assume the checksum is `123`

alter table t add column c int default 1; 

select tidb_row_checksum() from t where a = 1;  ## `123` is returned, but the real checksum may be `234`
```

3AceShowHand:

Issue Number: close #52590 

Problem Summary:

* `tidb_row_checksum` calculate the column-level checksum in the real time, instead of return the checksum store in the TiKV when the row inserted

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/point_get.go
```diff
@@ -400,6 +407,93 @@ func (e *PointGetExecutor) Next(ctx context.Context, req *chunk.Chunk) error {
...
+) error {
+	if !rowcodec.IsNewFormat(val) {
+		return nil
+	}
+	targetIndex, ok := shouldFillRowChecksum(schema)
+	if !ok {
+		return nil
+	}
+
+	columns := tblInfo.Cols()
```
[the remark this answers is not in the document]
3AceShowHand: I think the checksum should be calculated on all columns shown to the user on the client query they query the row.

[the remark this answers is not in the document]
3AceShowHand: the checksum should be calculated on all columns, decoded from the underline raw bytes.

zyguan (approved):

pkg/executor/point_get.go
```diff
@@ -387,19 +389,122 @@ func (e *PointGetExecutor) Next(ctx context.Context, req *chunk.Chunk) error {
...
+	}
+	tz := sctx.GetSessionVars().TimeZone
+	ft := []*types.FieldType{schema.Columns[checksumColumnIndex].GetType()}
+	checksumCols := chunk.NewChunkWithCapacity(ft, req.Capacity())
+	for i := start; i < end; i++ {
+		handle, val := handles[i], values[i]
+		if !rowcodec.IsNewFormat(val) {
+			checksumCols.AppendNull(0)
+			continue
+		}
```
tangenta, line 459: Will this be explained in user documentation?

3AceShowHand: no, there is no changes compare to the previous implementation, just fill the null as placeholder, and reset by the `fillChecksum` method.

3AceShowHand: I got your point, I think we can update the documentation to let user's know tat checksum only calculated on the new format.

tangenta (approved):

XuHuaiyu (approved):

Issue #49752, opened 2023-12-25
Create table accelerate


GMHDBJD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

create table in non owner
**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

Issue #27197, opened 2021-08-13
when run sysbench prepare，report “Write conflict” error


Lily2025:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
1、run sysbench prepare 
threads=512
tableNum = 512
tableSize = 10000000000

### 2. What did you expect to see? (Required)
sysbench report “Write conflict” error
<img width="919" alt="writeconflict " src="https://user-images.githubusercontent.com/84712107/129317463-0b4a4269-eb17-418d-a0eb-46d34e95e51b.png">

### 3. What did you see instead (Required)
no “Write conflict” error

### 4. What is your TiDB version? (Required)
------------+
| Release Version: v5.2.0
Edition: Community
Git Commit Hash: 30cf15a59db11c34ffe05fc926152a43327eaa61
Git Branch: heads/refs/tags/v5.2.0
UTC Build Time: 2021-08-09 12:01:42
GoVersion: go1.16.4
Race Enabled: false
TiKV Min Version: v3.0.0-60965b006877ca7234adaced7890d7b029ed1306
Check Table Before Drop: false |

jingshanglu:
@Lily2025 Can you add information about executing sql?

tiancaiamao:
The conflicts on the meta key `NextGlobalID`

The create table statement looks like this:

```
CREATE TABLE sbtest181(
  id BIGINT NOT NULL AUTO_INCREMENT,
  k INTEGER DEFAULT '0' NOT NULL,
  c CHAR(120) DEFAULT '' NOT NULL,
  pad CHAR(60) DEFAULT '' NOT NULL,
  PRIMARY KEY (id)
) /*! ENGINE = innodb */
```

One create table statement will use a transaction to increase the `NextGlobalID` meta key twice, here is the calling stack:

```
github.com/pingcap/tidb/structure.(*TxStructure).Inc(0xc011c7adc0, 0x5268a90, 0xc, 0xc, 0x1, 0xa, 0x5ecfba739940001, 0xc011c7adc0)
	/home/genius/project/src/github.com/pingcap/tidb/structure/string.go:64 +0x99
github.com/pingcap/tidb/meta.(*Meta).GenGlobalIDs(0xc012001a70, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/genius/project/src/github.com/pingcap/tidb/meta/meta.go:130 +0xcc
[...]

PR #52543, opened 2024-04-12
fast_create_table: add retry for local worker

GMHDBJD:

Issue Number: ref https://github.com/pingcap/tidb/issues/49752

Problem Summary:

- add retry for retryable error
- add id allocator to reduce write conflict
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/job_table.go
+++ b/pkg/ddl/job_table.go
@@ -494,7 +494,17 @@ func (d *ddl) delivery2LocalWorker(pool *workerPool, task *limitJobTask) {
 			metrics.DDLRunningJobCount.WithLabelValues(pool.tp().String()).Dec()
 		}()
 
-		err := wk.HandleLocalDDLJob(d.ddlCtx, job)
+		maxRetryTime := 100
+		var err error
+		for i := 0; i < maxRetryTime; i++ {
+			err = wk.HandleLocalDDLJob(d.ddlCtx, job)
+			// since local the job is not inserted into the ddl job queue, we need to add retry logic here.
+			if err == nil || !isRetryableError(err) {
+				break
+			}
+			logutil.DDLLogger().Info("handle local ddl job", zap.Int("retry times", i), zap.Error(err))
+			time.Sleep(time.Second)
+		}
 		pool.put(wk)
 		if err != nil {
 			logutil.DDLLogger().Info("handle ddl job failed", zap.Error(err), zap.Stringer("job", job))
--- a/pkg/ddl/job_table.go
+++ b/pkg/ddl/job_table.go
@@ -494,10 +494,8 @@ func (d *ddl) delivery2LocalWorker(pool *workerPool, task *limitJobTask) {
 			metrics.DDLRunningJobCount.WithLabelValues(pool.tp().String()).Dec()
 		}()
 
-		maxRetryTime := 100
-		var err error
-		for i := 0; i < maxRetryTime; i++ {
-			err = wk.HandleLocalDDLJob(d.ddlCtx, job)
+		for i := int64(0); i < variable.GetDDLErrorCountLimit(); i++ {
+			err := wk.HandleLocalDDLJob(d.ddlCtx, job)
 			// since local the job is not inserted into the ddl job queue, we need to add retry logic here.
 			if err == nil || !isRetryableError(err) {
 				break
--- a/pkg/ddl/job_table.go
+++ b/pkg/ddl/job_table.go
@@ -500,7 +500,7 @@ func (d *ddl) delivery2LocalWorker(pool *workerPool, task *limitJobTask) {
 			if err == nil || !isRetryableError(err) {
 				break
 			}
-			logutil.DDLLogger().Info("handle local ddl job", zap.Int("retry times", i), zap.Error(err))
+			logutil.DDLLogger().Info("handle local ddl job", zap.Int64("retry times", i), zap.Error(err))
 			time.Sleep(time.Second)
 		}
 		pool.put(wk)


pkg/ddl/id_allocator.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"context"
+	"sync"
+
+	"github.com/pingcap/tidb/pkg/kv"
+	"github.com/pingcap/tidb/pkg/meta"
+)
+
+const (
```
Benjamin2037, line 29: Can you add some comments to let user easy to know the func and parameters usage?

pkg/executor/importer/importer_testkit_test.go
```diff
@@ -348,7 +348,7 @@ func TestProcessChunkWith(t *testing.T) {
 		require.Len(t, progress.GetColSize(), 3)
 		checksumMap := checksum.GetInnerChecksums()
 		require.Len(t, checksumMap, 1)
-		require.Equal(t, verify.MakeKVChecksum(111, 3, 14231358899564314836), *checksumMap[verify.DataKVGroupID])
+		require.Equal(t, verify.MakeKVChecksum(111, 3, 4697906519963456140), *checksumMap[verify.DataKVGroupID])
```
GMHDBJD, line 351: checksum changed due to different table id

lance6716 (commented):
/approve

pkg/ddl/job_table.go
```diff
@@ -404,7 +404,16 @@ func (d *ddl) delivery2LocalWorker(pool *workerPool, task *limitJobTask) {
 			metrics.DDLRunningJobCount.WithLabelValues(pool.tp().String()).Dec()
 		}()
 
-		err := wk.HandleLocalDDLJob(d.ddlCtx, job)
+		maxRetryTime := 10
+		var err error
+		for i := 0; i < maxRetryTime; i++ {
+			err = wk.HandleLocalDDLJob(d.ddlCtx, job)
+			if err == nil || !isRetryableError(err) {
```
zimulala, line 411: I remember that if the job itself is not finished (`job.IsFinished` is false), it will retry itself. Why is there a special emphasis here?

pkg/ddl/id_allocator.go
```diff
@@ -0,0 +1,84 @@
...
+	IDAllocatorStep = 100
+)
+
+// IDAllocator is used to allocate global ID.
+type IDAllocator interface {
+	AllocIDs(n int) ([]int64, error)
+}
+
+type allocator struct {
+	cache []int64
```
zimulala, line 40: If the cache exists, if there are multiple tidbs in a cluster, is the following possible?
TiDB1: Execute DDL1, DDL2...DDL201 on table tblx. ,
TiDB2: Executes DDL302 (job.ID: 302) on table tbl;
After 5 minutes, DDL201 is not finished
TiDB1: Execute DDLn (job.ID: 202) on table tbl. ,

Then DDLn will start executing 5 minutes later than DDL302, but the actual execution will start and finish first.
There's no error problem, but it's weird. In addition, this ID will also be used for the ID of the table, which may also have similar problems.

pkg/ddl/ddl_worker.go
```diff
@@ -406,16 +402,16 @@ func (d *ddl) addBatchDDLJobs(tasks []*limitJobTask) error {
 		bdrRole = string(ast.BDRRoleNone)
 	)
 
-	if newTasks, err := d.combineBatchCreateTableJobs(tasks); err == nil {
-		tasks = newTasks
-	}
+	// TODO: since ticdc doesn't support batch create tables now, we disable until ticdc support it.
+	// if newTasks, err := d.combineBatchCreateTableJobs(tasks); err == nil {
```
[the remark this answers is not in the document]
GMHDBJD: They said it has not been scheduled yet, and we hope to support CDC in 8.1, so disable it first.

pkg/ddl/job_table.go
GMHDBJD: for fast create table, the job does not exist in the ddl job table, so the original retry mechanism (taken out from the ddl job table and re-executed) will not take effect.

pkg/ddl/id_allocator.go
GMHDBJD: Yes, that's possible. See id_allocator_test.go. If multiple tidbs execute ddl at the same time, the order of job ids is no longer guaranteed. And there may be id gaps between two jobs

GMHDBJD: However, the previous implementation also had similar problems. For example, the job that requested the ID later was executed for a long time before being committed. The order of the IDs was not guaranteed and had no impact.

pkg/ddl/job_table.go
zimulala: Got it, could we add a comment here?

pkg/ddl/id_allocator.go
zimulala: The original ID order is not guaranteed, but the interval is relatively small. It doesn't seem to be a problem at the moment, but this feels kind of weird. table ID also has this problem. Does this give an optimized performance comparison? See if the optimization effect is obvious.

GMHDBJD: Performance is one aspect, but more is robustness. When multiple tidbs allocate IDs at the same time, it will cause a lot of write conflicts, and changing to pessimistic transactions will also affect performance.  see https://github.com/pingcap/tidb/issues/27197

zimulala: 1. If you change step to 10, maybe the cases I mentioned above are less.
2. The occurrence probability of this scenario(#27197) you mentioned is relatively small. In addition, theoretically we have limitDDLJobs, so this kind of problem should not occur for a TiDB. But we use the same key for tableID and jobID, so a create table operation may result in multiple ID assignments. 
Suggestion: If we put tableID and jobID assignments together, this conflict will be reduced a lot. And, it does not affect the old logic (which considers the ID assignment to be relatively incremental).

GMHDBJD: In fact, we already have global id lock in the original logic, which means that there will no longer be a write conflict in the same tidb, but in multiple tidbs, this problem cannot be avoided.

pkg/ddl/ddl_worker.go
```diff
@@ -425,11 +421,10 @@ func (d *ddl) addBatchDDLJobs(tasks []*limitJobTask) error {
 	}
 
 	ctx := kv.WithInternalSourceType(context.Background(), kv.InternalTxnDDL)
-	// lock to reduce conflict
-	d.globalIDLock.Lock()
+
+	ids, err = d.genGlobalIDs(len(tasks))
```
D3Hunter, line 425: seems this optimize only works when we connects to multiple Nodes to do `create-table`

have you any data on how much this can improve performance?

GMHDBJD: img_v3_02cg_769e1e9e-dd6e-40a5-9ea6-b5ac4596e8dg
to
img_v3_02cc_cb333fa6-74f9-41c4-b4c5-ff62c7249f9g

pkg/ddl/id_allocator.go
D3Hunter: `txn.SetOption(kv.Pessimistic, true)` doesn't mean pessimistic transactions, we have to lock the required key explicitly, i tested `lock + incr(global id)` using 1000 routines, qps is about 1k

> Suggestion: If we put tableID and jobID assignments together, this conflict will be reduced a lot

agree with this, we can consider caching id after this optimization, if this part is still worth optimize.

FYI i have a on-going pr separating local ddl jobs from normal ones in `addBatchDDLJobs`, and `change job id allocation + insert into table` run in one txn, to make sure job submitted in id order

pkg/ddl/job_table.go
```diff
@@ -494,7 +494,17 @@ func (d *ddl) delivery2LocalWorker(pool *workerPool, task *limitJobTask) {
 			metrics.DDLRunningJobCount.WithLabelValues(pool.tp().String()).Dec()
 		}()
 
-		err := wk.HandleLocalDDLJob(d.ddlCtx, job)
+		maxRetryTime := 100
```
tangenta, line 497: Maybe use [tidb_ddl_error_count_limit](https://docs.pingcap.com/tidb/stable/system-variables#tidb_ddl_error_count_limit) instead?

tangenta (approved):

D3Hunter (approved):

Issue #52343, opened 2024-04-03
PreCalculateScalar consumes a lot of CPU when non-lite init stats


winoros:
## Enhancement

![image](https://github.com/pingcap/tidb/assets/7846227/88d75de8-d179-44d8-aa15-e07e5daf7ca6)

After https://github.com/pingcap/tidb/issues/52342 is fixed, we found another CPU consumer.
Too many small obj are allocated and causing too much GC pressure.

winoros:
And it's also a possible issue when we load needed histograms.

PR #52623, opened 2024-04-15
statistics: reduce small obj alloc for PreCalculateScalar

winoros:

Issue Number: close https://github.com/pingcap/tidb/issues/52343

Problem Summary:

We don't need to re-alloc anything since we only read from the chunk, no modification after reading it.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/scalar.go
+++ b/pkg/statistics/scalar.go
@@ -101,24 +101,26 @@ func (hg *Histogram) PreCalculateScalar() {
 	}
 	switch hg.GetLower(0).Kind() {
 	case types.KindMysqlDecimal, types.KindMysqlTime:
-		lower, upper := new(types.Datum), new(types.Datum)
+		var lower, upper types.Datum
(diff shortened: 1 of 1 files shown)


hi-rustin (commented):
I'm not very familiar with these structures.
Is it safe to reuse? Are there other places that use it in this way?
If it is, please add some comments to note that we trying to rescue it to save memory here.

winoros:
> I'm not very familiar with these structures. Is it safe to reuse? Are there other places that use it in this way? If it is, please add some comments to note that we trying to rescue it to save memory here.

You can jump into the func `convertDatumToScalar`, which calculates a float64 score for the datum. So the datum is read-only before it's destroyed.

hi-rustin (approved):
/hold

Please add some comments to it.

PR #52720, opened 2024-04-18
server: handle the case the type of param is set for the param sent by `SEND_LONG_DATA`

### LINKED ISSUE #52711: The type of the `param` sent through `SEND_LONG_DATA` is not set according to the request

Create the table

```
CREATE TABLE t (j JSON);
```

Prepare the following statement

```
INSERT INTO t (j) VALUES (cast(? as json))
```

Execute this statement with `SEND_LONG_DATA` and pass a string parameter . Usually, it can be triggered by using a too long parameter.

The json value is a normal string.

The json value is a opaque blob value, which means the parameter is used as a blob value.

Nightly.

The wrong code:

```
	for i := 0; i < len(params); i++ {
		// if params had received via ComStmtSendLongData, use them directly.
		// ref https://dev.mysql.com/doc/internals/en/com-stmt-send-long-data.html
		// see clientConn#handleStmtSendLongData
		if boundParams[i] != nil {
			params[i] = param.BinaryParam{
				Tp:  mysql.TypeBlob,
				Val: enc.DecodeInput(boundParams[i]),
			}
			continue
		}
```

TiDB should get the type from `paramTypes`

YangKeao:

Issue Number: close #52711

After sending the param through `SEND_LONG_DATA`, it's still possible that the type is given in the `EXECUTE` statement. But actually I'm not sure whether every client sends the type, so I kept the original behavior if the type is not sent to keep compatibility.

Assert whether the type is sent in `EXECUTE` statement. If it's sent, use the given `type. If it's not sent, still use the `Blob` to keep compatible with the original behavior.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

I tried to run the program https://gist.github.com/YangKeao/51bbc88600c69691a799227806f32cd5 on TiDB.

Before this PR, the data in table t after running this program is 
```
| utf8     | "你好"      | "你好"                            | 0x22E4BDA0E5A5BD22       |
| utf8     | "你好"      | "base64:type252:IuS9oOWlvSI="     | 0x22E4BDA0E5A5BD22       |
| gbk      | "浣犲ソ"    | "浣犲ソ"                          | 0x22E4BDA0E5A5BD22       |
| gbk      | "浣犲ソ"    | "base64:type252:Iua1o+eKsuOCvSI=" | 0x22E6B5A3E78AB2E382BD22 |
```

After this PR, the data in table t after running this program is
```
| utf8     | "你好"      | "你好"                            | 0x22E4BDA0E5A5BD22       |
| utf8     | "你好"      | "你好"                            | 0x22E4BDA0E5A5BD22       |
| gbk      | "浣犲ソ"    | "浣犲ソ"                          | 0x22E4BDA0E5A5BD22       |
| gbk      | "浣犲ソ"    | "浣犲ソ"                          | 0x22E4BDA0E5A5BD22       |
```

Running this program in MySQL, will get:
```
| utf8     | "你好"      | "你好"      | 0x22E4BDA0E5A5BD22 |
| utf8     | "你好"      | "你好"      | 0x22E4BDA0E5A5BD22 |
| gbk      | "浣犲ソ"    | "浣犲ソ"    | 0x22E4BDA0E5A5BD22 |
| gbk      | "浣犲ソ"    | "浣犲ソ"    | 0x22E4BDA0E5A5BD22 |
```

Which is the same as what we got with this PR.

I'll try to add this manual test to `mysqlclienttest` later :thinking: .

### Release note

```release-note
Fix the issue that param transfered by `SEND_LONG_DATA` is always regarded as `Blob` type.
```

### DIFF
--- a/pkg/server/conn_stmt_params.go
+++ b/pkg/server/conn_stmt_params.go
@@ -40,7 +40,24 @@ func parseBinaryParams(params []param.BinaryParam, boundParams [][]byte, nullBit
 		if boundParams[i] != nil {
 			params[i] = param.BinaryParam{
 				Tp:  mysql.TypeBlob,
-				Val: enc.DecodeInput(boundParams[i]),
+				Val: boundParams[i],
+			}
+
+			// The legacy logic is kept: if the `paramTypes` somehow didn't contain the type information, it will be treated as
+			// BLOB type. We didn't return `mysql.ErrMalformPacket` to keep compatibility with older versions, though it's
+			// meaningless if every clients work properly.
+			if (i<<1)+1 < len(paramTypes) {
+				// Only TEXT or BLOB type will be sent through `SEND_LONG_DATA`.
+				tp := paramTypes[i<<1]
+
+				switch tp {
+				case mysql.TypeVarchar, mysql.TypeVarString, mysql.TypeString, mysql.TypeBit:
+					params[i].Tp = tp
+					params[i].Val = enc.DecodeInput(boundParams[i])
+				case mysql.TypeBlob, mysql.TypeTinyBlob, mysql.TypeMediumBlob, mysql.TypeLongBlob:
+					params[i].Tp = tp
+					params[i].Val = boundParams[i]
+				}
 			}
 			continue
 		}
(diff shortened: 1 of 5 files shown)


YangKeao:
/hold

Need more tests

lcwangchao (approved):

xhebox (approved):

PR #52727, opened 2024-04-18
expression: Make `BuildContext.IsUseCache/SetSkipPlanCache` thread-safe

lcwangchao:
### What problem does this PR solve?

Issue Number: close #52728

Problem Summary:

Because `BuidContext` may be used in concurrency, it is better to make all of its methods thread-safe including `IsUseCache/SetSkipPlanCache`

### What changed and how does it work?

Make `IsUseCache/SetSkipPlanCache` thread-safe to change the undelaying field to atomic.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #52738, opened 2024-04-18
stmtctx: use a standalone `RangeFallbackHandler`

### LINKED ISSUE #52737: Don't use a whole `*StatementContext` as `RangeFallbackHandler`
## Enhancement

We left a TODO for `RangeFallbackHandler`, because we are using a whole `*StatementContext` as the implementation of `RangeFallbackHandler`. We can extract some fields out of the statement context, and create a smaller implementation of `RangeFallbackHandler`.

YangKeao:

Issue Number: close #52737

Problem Summary:

We left a TODO to create a standalone `RangeFallbackHandler` rather than use a whole `*StatementContext`, because it's impossible to clone a whole `*StatementContext`.

In this PR, I prepared a new smaller implementation of `*StatementContext`, and use it in the ranger context.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > 
  > The implementation is blindly copied from `StatementContext` implementation, and it's tested by existing tests about plan cache.

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/util/context/plancache.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (h *PlanCacheTracker) setSkipPlanCache(err error) {
+	h.UseCache = false
+	switch h.CacheType {
+	case DefaultNoCache:
+		h.warnHandler.AppendWarning(errors.NewNoStackError("unknown cache type"))
+	case SessionPrepared:
+		h.warnHandler.AppendWarning(errors.NewNoStackErrorf("skip prepared plan-cache: %s", err.Error()))
+	case SessionNonPrepared:
+		if h.alwaysWarnSkipCache {
```
lcwangchao, line 75: Is it better to rename it to `alwaysWarnSkipNoPreparedCache` because it only works on `SessionNonPrepared` ?

pkg/util/context/plancache.go
```diff
@@ -0,0 +1,122 @@
...
+	DefaultNoCache PlanCacheType = iota
+	// SessionPrepared session prepared plan cache
+	SessionPrepared
+	// SessionNonPrepared session non-prepared plan cache
+	SessionNonPrepared
+)
+
+// PlanCacheTracker PlanCacheTrackerf `PlanCacheTracker`.
+type PlanCacheTracker struct {
+	UseCache            bool
```
lcwangchao, line 37: Could we make `PlanCacheTracker` a thread-safe? I schedule a PR #52727 to make it to avoid some data race if `BuildContext` is used in concurrency.

If we can do it in this PR, I can close #52727

YangKeao: Theoretically, even `BuildContext` may be used concurrently, it only reads the `UseCache` right? Then the `BuildContext` is still thread-safe even if we change nothing :thinking: .

I agree that it's always better to make it thread-safe. Making change here can also make it safe even if the `PlanContext`, `BuildContext` is used concurrently :+1: .

Let me do it.

pkg/util/context/plancache.go
YangKeao: I think the semantic is to warn in any cases, but for other cases there is already a warn so it only applies to `SessionNonPrepared`.

It's used in `explain format='plan_cache'  xxx`. It should give a warning to show the status of any kinds of plan cache.

lcwangchao (approved):

XuHuaiyu (approved):

qw4990 (approved):

PR #52796, opened 2024-04-22
statistics: support global singleflight for sync load

### LINKED ISSUE #52797: global sync load to avoid too much item in the channel
## Enhancement

Now, we need to put all sync load items into the channel, then the workers will get them to sync load through the single flight. we should put all sync load items into a single flight. then worker will deal with them.

hawkingrei:

Issue Number: close #52797

Problem Summary:

#### before

whiteboard_exported_image

All items the query needs will be put into the global task channel( the default capacity is 1000). And workers will merge the same task with the singleflight. So performance depends on the number of tasks that can be concurrently executed. 
Under poor performance conditions, he will not be able to handle a sufficient number of tasks, resulting in slow processing. But there are still many same tasks in the channel. So they still need to check whether to sync load. And They will fill up this channel. Query can not insert the sync load task into it and load necessary stats.

#### after 

whiteboard_exported_image (1)

So we need to refactor this code. We need first to let the query pass through singleflight to merge identical tasks. Then insert the task into the channel to make the worker syncload it.

So why is it that during the startup of Wetech, the queries are very unstable and there are severe timeouts with sync load? The problem lies right here.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/handle/syncload/stats_syncload.go
+++ b/pkg/statistics/handle/syncload/stats_syncload.go
@@ -32,8 +32,10 @@ import (
 	statstypes "github.com/pingcap/tidb/pkg/statistics/handle/types"
 	"github.com/pingcap/tidb/pkg/types"
 	"github.com/pingcap/tidb/pkg/util"
+	"github.com/pingcap/tidb/pkg/util/intest"
 	"github.com/pingcap/tidb/pkg/util/logutil"
 	"go.uber.org/zap"
+	"golang.org/x/sync/singleflight"
 )
 
 // RetryCount is the max retry count for a sync load task.
@@ -43,6 +45,8 @@ type statsSyncLoad struct {
 	statsHandle statstypes.StatsHandle
 	StatsLoad   statstypes.StatsLoad
 }
+
+var globalStatsSyncLoadSingleFlight singleflight.Group
 
 // NewStatsSyncLoad creates a new StatsSyncLoad.
 func NewStatsSyncLoad(statsHandle statstypes.StatsHandle) statstypes.StatsSyncLoad {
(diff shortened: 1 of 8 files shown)


pkg/statistics/handle/syncload/stats_syncload_test.go
```diff
@@ -211,17 +211,18 @@ func TestConcurrentLoadHistWithPanicAndFail(t *testing.T) {
 		task3, err3 := h.HandleOneTask(testKit.Session().(sessionctx.Context), task1, exitCh)
 		require.NoError(t, err3)
 		require.Nil(t, task3)
-
-		task, err3 := h.HandleOneTask(testKit.Session().(sessionctx.Context), nil, exitCh)
```
hawkingrei, line 215: Now, we merge the same request on the top. it has nothing in the task channel. so it will block.

hawkingrei:
the old singleflight still exists, but it is to support timeout for tasks.

pkg/statistics/handle/syncload/stats_syncload_test.go
```diff
@@ -312,11 +323,11 @@ func TestRetry(t *testing.T) {
 	result, err1 := h.HandleOneTask(testKit.Session().(sessionctx.Context), task1, exitCh)
 	require.NoError(t, err1)
 	require.Nil(t, result)
-	select {
```
hawkingrei, line 315: We cannot use the old way to check. the item in the channel will be fetched from ```globalSingleFlight```.

winoros (approved):

pkg/statistics/handle/syncload/stats_syncload.go
```diff
@@ -230,33 +241,17 @@ func (s *statsSyncLoad) HandleOneTask(sctx sessionctx.Context, lastTask *statsty
...
-			return nil, nil
-		}
-		return task, sr.Err
-	case <-time.After(timeout):
-		if !isVaildForRetry(task) {
-			result.Error = errors.New("stats loading timeout")
-			task.ResultCh <- result
-			return nil, nil
-		}
-		task.ToTimeout.Add(time.Duration(sctx.GetSessionVars().StatsLoadSyncWait.Load()) * time.Microsecond)
```
hi-rustin, line 257: Can I ask why do we need to change this behavior?

pkg/statistics/handle/syncload/stats_syncload.go
```diff
@@ -122,25 +128,30 @@ func (*statsSyncLoad) SyncWaitStatsLoad(sc *stmtctx.StatementContext) error {
...
-				errorMsgs = append(errorMsgs, result.ErrorMsg())
+			if result.Err != nil {
+				errorMsgs = append(errorMsgs, result.Err.Error())
 			}
-			delete(resultCheckMap, result.Item)
-			if len(resultCheckMap) == 0 {
-				metrics.SyncLoadHistogram.Observe(float64(time.Since(sc.StatsLoad.LoadStartTime).Milliseconds()))
-				return nil
+			val := result.Val.(stmtctx.StatsLoadResult)
+			if val.HasError() {
```
hi-rustin, line 141: Could you please add some comments to explain why we have one error from the result channel and one error from the value?

hi-rustin (commented):
rest LGTM

pkg/statistics/handle/syncload/stats_syncload.go
hawkingrei: Update.

hi-rustin (approved):

PR #38826, opened 2022-11-02
expression, cmd: fix ColumnSubstitute and allow some cases to substitute

wjhuang2016:
Signed-off-by: wjhuang2016 <huangwenjun1997@gmail.com>

### What problem does this PR solve?

Issue Number: close #38736

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #52812, opened 2024-04-22
expression: don't cast collation for in expression is the new collation is disabled

### LINKED ISSUE #52772: TiDB panic when new collation is disabled and two expressions use different collation
## Bug Report

1. Create a cluster with TiDB version before v4.0.0.
2. Upgrade the cluster to master.
3. Create 2 tables with the same charset but different collations:
```sql
create table t1(code varchar(32)) CHARSET=utf8 COLLATE=utf8_general_ci;
create table t2(code varchar(32)) CHARSET=utf8 COLLATE=utf8_bin;
```
4. Execute a query:
```sql
select * from t1 join t2 on t1.code=t2.code and t1.code in ('1') and t2.code in ('1');
```

Empty query result

TiDB reports error:
```
ERROR 1105 (HY000): runtime error: invalid memory address or nil pointer dereference
```

The stack:
```
github.com/pingcap/tidb/expression.NewFunctionInternal\n\t/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/expression/scalar_function.go:277\ngithub.com/pingcap/tidb/expression.ColumnSubstituteImpl\n\t/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/expression/util.go:515\ngithub.com/pingcap/tidb/expression.ColumnSubstitute\n\t/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/expression/util.go:414\ngithub.com/pingcap/tidb/expression.(*propConstSolver).propagateConstantEQ\n\t/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/expression/constant_propagation.go:214\ngithub.com/pingcap/tidb/expression.(*propConstSolver).solve\n\t/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/expression/constant_propagation.go:352\ngithub.com/pingcap/tidb/expression.(*propConstSolver).PropagateConstant\n\t/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/expression/constant_propagation.go:711\ngithub.com/pingcap/tidb/expression.PropagateConstant\n\t/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/expression/constant_propagation.go:361\ngithub.com/pingcap/tidb/planner/core.(*DataSource).PredicatePushDown\n\t/home/jenkins/agent/workspace/build-common/go/src/github.com/pingcap/tidb/planner/core/rule_predicate_push_down.go:135
```

master

wjhuang2016:

Issue Number: close #52772

Problem Summary:
If the new collation is disabled, we should not cast collation for in expression, which may lead to panic in the issue case.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/expression/collation.go
+++ b/pkg/expression/collation.go
@@ -302,19 +302,6 @@ func deriveCollation(ctx BuildContext, funcName string, args []Expression, retTy
 
 // CheckAndDeriveCollationFromExprs derives collation information from these expressions, return error if derives collation error.
 func CheckAndDeriveCollationFromExprs(ctx BuildContext, funcName string, evalType types.EvalType, args ...Expression) (et *ExprCollation, err error) {
-	//defer func() {
-	//	if err != nil && !collate.NewCollationEnabled() {
-	//		// args[0] can't be nil here.
-	//		// Use the collation of the first argument instead of returning an error.
-	//		et = &ExprCollation{
-	//			Coer:      args[0].Coercibility(),
-	//			Repe:      args[0].Repertoire(),
-	//			Charset:   args[0].GetType().GetCharset(),
-	//			Collation: args[0].GetType().GetCollate(),
-	//		}
-	//		err = nil
-	//	}
-	//}()
 	ec := inferCollation(args...)
 	if ec == nil {
 		return nil, illegalMixCollationErr(funcName, args)
--- a/pkg/expression/collation.go
+++ b/pkg/expression/collation.go
@@ -302,19 +302,19 @@ func deriveCollation(ctx BuildContext, funcName string, args []Expression, retTy
 
 // CheckAndDeriveCollationFromExprs derives collation information from these expressions, return error if derives collation error.
 func CheckAndDeriveCollationFromExprs(ctx BuildContext, funcName string, evalType types.EvalType, args ...Expression) (et *ExprCollation, err error) {
-	defer func() {
-		if err != nil && !collate.NewCollationEnabled() {
-			// args[0] can't be nil here.
-			// Use the collation of the first argument instead of returning an error.
-			et = &ExprCollation{
-				Coer:      args[0].Coercibility(),
-				Repe:      args[0].Repertoire(),
-				Charset:   args[0].GetType().GetCharset(),
-				Collation: args[0].GetType().GetCollate(),
-			}
-			err = nil
-		}
-	}()
+	//defer func() {
+	//	if err != nil && !collate.NewCollationEnabled() {
+	//		// args[0] can't be nil here.
+	//		// Use the collation of the first argument instead of returning an error.
+	//		et = &ExprCollation{
+	//			Coer:      args[0].Coercibility(),
+	//			Repe:      args[0].Repertoire(),
+	//			Charset:   args[0].GetType().GetCharset(),
+	//			Collation: args[0].GetType().GetCollate(),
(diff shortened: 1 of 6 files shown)


pkg/expression/collation.go
```diff
@@ -302,6 +302,19 @@ func deriveCollation(ctx BuildContext, funcName string, args []Expression, retTy
 
 // CheckAndDeriveCollationFromExprs derives collation information from these expressions, return error if derives collation error.
 func CheckAndDeriveCollationFromExprs(ctx BuildContext, funcName string, evalType types.EvalType, args ...Expression) (et *ExprCollation, err error) {
+	defer func() {
+		if err != nil && !collate.NewCollationEnabled() {
+			// args[0] can't be nil here.
+			// Use the collation of the first argument instead of returning an error.
+			et = &ExprCollation{
```
YangKeao, line 309: It'll cause some behavior change :thinking: . For example:

```sql
select convert('1' using utf8) collate utf8_bin = convert('1' using utf8) collate utf8_general_ci;
```

It'll return an error before this PR no matter whether `new_collation` is enabled or not, but after this PR, it'll return without any error.

For this issue, I propose two other options, but none of them is perfect :facepalm: :

1. Avoid constant propogation in this case. We can replace the `NewFunctionInternal` with `NewFunction` in the `ColumnSubstituteImpl`, and check the `err`. If it returns an err, we can avoid `substitute` in this case.

2. Add back the `newExpr.SetCoercibility(v.Coercibility())` removed in https://github.com/pingcap/tidb/pull/38826. It sounds better because actually the coercibility (after `substitute`) should  be the `Column` (2). I'm not sure 100% sure, but maybe we can fix the issue found in #38826 by `Clone` the `newExpr` before `SetCoercibility`.

YangKeao: The proposal 2 is not correct, because the following case will fail:

```
create table t (a char(10) collate utf8mb4_bin, b char(10) collate utf8mb4_general_ci);
insert into t values ('a', 'A');
select * from t t1, t t2 where t1.a=t2.b and t2.b='a' collate utf8mb4_general_ci;
```

pkg/planner/core/expression_rewriter.go
```diff
@@ -1951,6 +1951,9 @@ func (er *expressionRewriter) castCollationForIn(colLen int, elemCnt int, stkLen
 	if colLen != 1 {
 		return
 	}
+	if !collate.NewCollationEnabled() {
```
YangKeao, line 1954: Please, add a comment here to tell why `castCollationForIn` is disabled when new collation is not enabled? It's not intuitive. IIUC, it works because it didn't set coercibility to 0, so that the following panic (mix collation with 0 coercibility) will not happen.

I personally prefer to also enable the collation check in `ColumnSubstituteImpl` for old collation, but it may cause big performance impact and need manually workaround (rewriting SQL), so it's not good.
This solution is also fine for me, as I cannot think up any counter example in old collation.

YangKeao (approved):

winoros (approved):

PR #52890, opened 2024-04-25
txn: handle errors returned from Flush

### LINKED ISSUE #52889: Errors returned from Flush are not equally treated as those returned from Commit
## Bug Report

ref #50215

See the case `TestPipelinedDMLDisableRetry`.

```go
func TestPipelinedDMLDisableRetry(t *testing.T) {
	// the case tests that
	// 1. auto-retry for pipelined dml is disabled
	// 2. the write conflict error message returned from a Flush (instead of from Commit) is correct
	require.Nil(t, failpoint.Enable("tikvclient/pipelinedMemDBMinFlushKeys", `return(1)`))
	require.Nil(t, failpoint.Enable("tikvclient/pipelinedMemDBMinFlushSize", `return(1)`))
	require.Nil(t, failpoint.Enable("tikvclient/pipelinedMemDBForceFlushSizeThreshold", `return(1)`))
	defer func() {
		require.Nil(t, failpoint.Disable("tikvclient/pipelinedMemDBMinFlushKeys"))
		require.Nil(t, failpoint.Disable("tikvclient/pipelinedMemDBMinFlushSize"))
		require.Nil(t, failpoint.Disable("tikvclient/pipelinedMemDBForceFlushSizeThreshold"))
	}()
	tk1 := testkit.NewTestKit(t, store)
	tk1.Session().GetSessionVars().InitChunkSize = 1
	tk1.Session().GetSessionVars().MaxChunkSize = 1
	tk1.MustExec("use test")
	tk2.MustExec("use test")
	tk1.MustExec("drop table if exists t")
	tk1.MustExec("create table t(a int primary key, b int)")
	// we need to avoid inserting *literals* into t, so let t2 be the source table.
	tk1.MustExec("create table t2(a int, b int)")
	tk1.MustExec("insert into t2 values (1, 1), (2, 1)")
	require.Nil(t, failpoint.Enable("tikvclient/beforePipelinedFlush", `pause`))
	tk1.MustExec("set session tidb_dml_type = bulk")
	errCh := make(chan error)
	go func() {
		// we expect that this stmt triggers 2 flushes, each containing only 1 row.
		errCh <- tk1.ExecToErr("insert into t select * from t2 order by a")
	}()
	time.Sleep(500 * time.Millisecond)
	tk2.MustExec("insert into t values (1,2)")
	require.Nil(t, failpoint.Disable("tikvclient/beforePipelinedFlush"))
	err := <-errCh
	require.Error(t, err)
	require.True(t, kv.ErrWriteConflict.Equal(err), fmt.Sprintf("error: %s", err))
	require.ErrorContains(t, err, "tableName=test.t")
}
```

Write conflict and assertion errors should be handled.

They are directly returned to user.

Master(6ed307ec244cb8823f3e1df9c76b862db901c216)

ekexium:

Issue Number: close #52889

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/session/session.go
```diff
@@ -2274,6 +2278,17 @@ func runStmt(ctx context.Context, se *session, s sqlexec.Statement) (rs sqlexec.
 	}
 
 	rs, err = s.Exec(ctx)
+
+	// Pipelined-DML can return assertion error and write conflict here, handle them like we handle
+	// errors returned from commit.
+	if err != nil {
```
cfzjywxk, line 2284: Should there be a condition check like
```
if s.isPipelinedDML {
    if err != nil {
        err = se.handleAssertionFailure(ctx, err)
    }
} else {
    unrechable or errors.
}
```

ekexium: Could be. It also makes the intention clearer.

pkg/session/session.go
```diff
@@ -2274,6 +2276,20 @@ func runStmt(ctx context.Context, se *session, s sqlexec.Statement) (rs sqlexec.
 	}
 
 	rs, err = s.Exec(ctx)
+
+	// check if is pipelined dml
+	if se.txn.Valid() && se.txn.IsPipelined() {
+		// Pipelined-DML can return assertion error and write conflict here, handle them like we handle
+		// errors returned from commit.
```
crazycs520, line 2283: Could you add comment to explain why need check here? 

The comment only says `can`, but doesn't explain `why`.

pkg/session/session.go
```diff
@@ -846,7 +846,9 @@ func (s *session) tryReplaceWriteConflictError(oldErr error) (newErr error) {
 	}
 	originErr := errors.Cause(oldErr)
 	inErr, _ := originErr.(*errors.Error)
-	args := inErr.Args()
+	oldArgs := inErr.Args()
+	args := make([]any, len(oldArgs))
+	copy(args, oldArgs)
```
crazycs520, line 851: Also, could you add some comments about why need a copy here?

crazycs520 (approved):

cfzjywxk (approved):

PR #52941, opened 2024-04-27
Optimizer: Refactor and simplify outer to inner join conversion rule #52941 | tidb-test=pr/2321

ghazalfamilyusa:

Issue Number: Ref #53278

Problem Summary:
Currently, the outer to inner join rewrite is part of the predicate push down. This is causing confusion and difficulty in extending of both rewrites. Also, the current implementation of outer to inner join rewrite has unnecessary special cases.

A new logical optimization is added to the list of rules. Also, a simple solution of null filtering (or NF) predicate check is used.
The solution is simply:
-  NF(A AND B) = NF(A) OR NF(B)
-  NF(A OR B) = NF(A) AND NF(B) 

The scope covers previous solution plus some more with the generalized solution. mysql tests has two diffs and both are OK to promote. A separate PR will be used for that. 

### Check List

Tests 

- [X] Unit test
- [] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/logical_plan_builder.go
```diff
@@ -250,6 +250,7 @@ func (b *PlanBuilder) buildAggregation(ctx context.Context, p base.LogicalPlan,
 	b.optFlag |= flagPredicatePushDown
 	b.optFlag |= flagEliminateAgg
 	b.optFlag |= flagEliminateProjection
+	b.optFlag |= flagConvertOuterToInnerJoin
```
elsa0520, line 253: Do we need to add this flag in buildAgg ? This is only a rule for join node, so maybe we can just flag this rule in buildJoin ?

pkg/planner/core/logical_plan_builder.go
```diff
@@ -1287,6 +1289,7 @@ func (b *PlanBuilder) buildSelection(ctx context.Context, p base.LogicalPlan, wh
 	b.optFlag |= flagPredicatePushDown
 	b.optFlag |= flagDeriveTopNFromWindow
 	b.optFlag |= flagPredicateSimplification
+	b.optFlag |= flagConvertOuterToInnerJoin
```
elsa0520, line 1292: Same as above

pkg/planner/core/logical_plan_builder.go
```diff
@@ -4709,6 +4712,7 @@ func (b *PlanBuilder) buildDataSourceFromCTEMerge(ctx context.Context, cte *ast.
 
 func (b *PlanBuilder) buildDataSource(ctx context.Context, tn *ast.TableName, asName *model.CIStr) (base.LogicalPlan, error) {
 	b.optFlag |= flagPredicateSimplification
+	b.optFlag |= flagConvertOuterToInnerJoin
```
elsa0520, line 4715: Same as above

pkg/planner/core/logical_plan_builder.go
ghazalfamilyusa: Outer to inner has its own push down through nodes and does not rely on predicate push down. if we limit it to joins then we may miss out something like 
with cte as (select count(a2) as cnt,b2 from t1 left outer join t2 on a1=a2 group by b2) select * from cte where b2 > 1

pkg/planner/core/logical_plan_builder.go
ghazalfamilyusa: Same as above as well

pkg/planner/core/rule_outer_to_inner_join.go
```diff
@@ -0,0 +1,185 @@
...
+	case *expression.Constant:
+		return true
+	}
+	return false
+}
+
+// isNullFiltered takes care of complex predicates like this:
+// isNullFiltered(A OR B) = isNullFiltered(A) AND isNullFiltered(B)
+// isNullFiltered(A AND B) = isNullFiltered(A) OR isNullFiltered(B)
+func isNullFiltered(ctx base.PlanContext, innerSchema *expression.Schema, predicate expression.Expression, outerSchema *expression.Schema) bool {
```
elsa0520, line 161: Should we add this part logic into util.IsNullRejected() ?

pkg/planner/core/logical_plan_builder.go
elsa0520: Oh, I understand. We should judge the predicate which is null-rejected firstly furthermore we can judge “outer join to inner join”.
But if we don't have join clause we don't need to judge the predicate which is null-rejected.
So it can cover all situations when we just mark the flag at the "buildjoin" . I don’t know if I understand it correctly?

pkg/planner/core/rule_outer_to_inner_join.go
```diff
@@ -0,0 +1,185 @@
...
+			len(predicates))
+	combinedCond = append(combinedCond, p.LeftConditions...)
+	combinedCond = append(combinedCond, p.RightConditions...)
+	combinedCond = append(combinedCond, expression.ScalarFuncs2Exprs(p.EqualConditions)...)
+	combinedCond = append(combinedCond, p.OtherConditions...)
+	combinedCond = append(combinedCond, predicates...)
+	return combinedCond
+}
+
+// convertOuterToInnerJoin converts outer to inner joins if the unmtaching rows are filtered.
```
elsa0520, line 41: Could you please add more detail comments in here such as the query example, main optimization process (Preorder traversal or subsequent traversal?)?

pkg/planner/core/rule_outer_to_inner_join.go
ghazalfamilyusa: We could but it is a common routine to other parts of the code and could be risky to do so. May be a subsequent PR can do that.

pkg/planner/core/rule_outer_to_inner_join.go
ghazalfamilyusa: OK. Added more comments.

elsa0520:
BTW, your linked issue is incorrect. Could you please link a correct issue? If there is no related issue, it will better to create a new issue for this PR ~

pkg/planner/core/logical_plan_builder.go
ghazalfamilyusa: I see your point. At the build level, for aggregations there is no predicate to pass down for null check. But, why flagPredicatePushDown is used here? I basically, followed the logic of flagPredicatePushDown in setting the flagConvertOuterToInnerJoin.

pkg/planner/core/logical_plan_builder.go
ghazalfamilyusa: Actually, for selection, there is a possibility of passing down predicate from selection to joins.

pkg/expression/schema.go
```diff
@@ -80,6 +80,24 @@ func (s *Schema) Clone() *Schema {
...
+	case *Column:
+		return schema.Contains(v)
+	case *ScalarFunction:
+		for _, arg := range v.GetArgs() {
+			if ExprReferenceSchema(arg, schema) {
+				return true
+			}
+		}
+		return false
+	case *CorrelatedColumn, *Constant:
```
qw4990, line 95: Should we consider `CorrelatedColumn.Column` in this case?

pkg/planner/core/rule_outer_to_inner_join.go
```diff
@@ -0,0 +1,196 @@
...
+// The main logic is applied for joins:
+//  1. Traversal is preorder and the passed down predicate is checked for the left/right after join
+//  2. The ON clause and passed down predicate (from higher selects or joins) are comined and applied to join children.
+//     This logic depends on the join type with the following logic:
+//     - For left/right outer joins, the ON clause an be applied only on the inner side (null producing side)
+//     - For inner/semi joins, the ON clause can be applied on both children
+//     - For anti semi joins, ON clause applied only on left side
+//     - For all other cases, do not pass ON clause.
+func (*convertOuterToInnerJoin) optimize(_ context.Context, p base.LogicalPlan, _ *optimizetrace.LogicalOptimizeOp) (base.LogicalPlan, bool, error) {
+	planChanged := false
```
qw4990, line 57: Why is planChanged always False?

ghazalfamilyusa: This is a place holder for rule engine interaction and we are not leveraging that for all rewrites.

pkg/expression/schema.go
ghazalfamilyusa: > Should we consider `CorrelatedColumn.Column` in this case?

It could be risky to do so since the nullability of correlated  column on the inner side is hard t o figure out.

pkg/planner/core/rule_outer_to_inner_join.go
```diff
@@ -0,0 +1,196 @@
...
+			if isNullFiltered(ctx, innerSchema, expr.GetArgs()[0], outerSchema) {
+				return true
+			}
+			return isNullFiltered(ctx, innerSchema, expr.GetArgs()[0], outerSchema)
+		} else if expr.FuncName.L == ast.LogicOr {
+			if !(isNullFiltered(ctx, innerSchema, expr.GetArgs()[0], outerSchema)) {
+				return false
+			}
+			return isNullFiltered(ctx, innerSchema, expr.GetArgs()[1], outerSchema)
+		} else {
```
qw4990, line 190: Can be simplified to:
```suggestion
		}
	default:
		return util.IsNullRejected(ctx, innerSchema, predicate)
```

pkg/planner/core/rule_outer_to_inner_join.go
```diff
@@ -0,0 +1,196 @@
...
+			return util.IsNullRejected(ctx, innerSchema, expr)
+		}
+	default:
+		return util.IsNullRejected(ctx, innerSchema, predicate)
```
qw4990, line 194: Can be simplified to:
```suggestion
		}
	default:
		return util.IsNullRejected(ctx, innerSchema, predicate)
```

pkg/planner/core/rule_outer_to_inner_join.go
```diff
@@ -0,0 +1,196 @@
...
+	case *expression.Constant:
+		return true
+	}
+	return false
+}
+
+// isNullFiltered takes care of complex predicates like this:
+// isNullFiltered(A OR B) = isNullFiltered(A) AND isNullFiltered(B)
+// isNullFiltered(A AND B) = isNullFiltered(A) OR isNullFiltered(B)
+func isNullFiltered(ctx base.PlanContext, innerSchema *expression.Schema, predicate expression.Expression, outerSchema *expression.Schema) bool {
```
qw4990, line 172: `outerSchema` seems useless? Can we remove this argument from this function?

pkg/planner/core/rule_outer_to_inner_join.go
```diff
@@ -0,0 +1,196 @@
...
+	canBePushed, _ := BreakDownPredicates(p, predicates)
+	child := p.Children()[0]
+	child = child.ConvertOuterToInnerJoin(canBePushed)
+	p.SetChildren(child)
+	return p
+}
+
+// allConstants checks if only the expression has only constants.
+func allConstants(expr expression.Expression) bool {
+	switch v := expr.(type) {
```
qw4990, line 155: To prevent potential bug on Plan Cache, better to add this check as well:
```
	if expression.MaybeOverOptimized4PlanCache(ctx, []expression.Expression{expr}) {
		return false // expression contains non-deterministic parameter
	}
```

pkg/planner/core/rule_outer_to_inner_join.go
qw4990: +1, better to put these 2 functions together

pkg/planner/core/rule_outer_to_inner_join.go
ghazalfamilyusa: Done

ghazalfamilyusa: > To prevent potential bug on Plan Cache, better to add this check as well:
> 
> ```
> 	if expression.MaybeOverOptimized4PlanCache(ctx, []expression.Expression{expr}) {
> 		return false // expression contains non-deterministic parameter
> 	}
> ```

Done

qw4990 (approved):

pkg/planner/core/testdata/plan_suite_unexported_out.json
```diff
@@ -73,16 +73,16 @@
       "Join{DataScan(t1)->DataScan(t2)}(test.t.a,test.t.b)(test.t.d,test.t.d)->Projection",
       "Join{DataScan(ta)->DataScan(tb)}(test.t.d,test.t.d)->Projection",
       "Join{DataScan(ta)->DataScan(tb)}(test.t.d,test.t.d)->Projection",
+      "Join{DataScan(ta)->DataScan(tb)}(test.t.d,test.t.d)->Sel([eq(test.t.a, 0)])->Projection",
```
elsa0520, line 76: Why this output changed a lot ? I saw you didn't change any case of this in.json ?

pkg/planner/core/rule_outer_to_inner_join.go
```diff
@@ -0,0 +1,199 @@
...
+		outerTable = outerTable.ConvertOuterToInnerJoin(combinedCond)
+	} else if p.JoinType == AntiSemiJoin {
+		innerTable = innerTable.ConvertOuterToInnerJoin(predicates)
+		outerTable = outerTable.ConvertOuterToInnerJoin(combinedCond)
+	} else {
+		innerTable = innerTable.ConvertOuterToInnerJoin(predicates)
+		outerTable = outerTable.ConvertOuterToInnerJoin(predicates)
+	}
+
+	if switchChild {
```
elsa0520, line 118: btw, why we need to swith child order ? It seems that the old logic in rule_predicate_push_down not include this part ?

ghazalfamilyusa: There is no child re-ordering here. The switch is just in the naming of innerchild and outerchild to have common code for left/right/inner. The flag is used to put the children in the same original order.

pkg/planner/core/testdata/plan_suite_unexported_out.json
ghazalfamilyusa: @elsa0520 : actually some tests are missing flagConvertOuterToInnerJoin whcih cause some joins not converted and different push down. Will fix the tests by adding flagConvertOuterToInnerJoin to the plan tests.

elsa0520 (approved):
LGTM

XuHuaiyu (approved):

PR #52971, opened 2024-04-29
*: global index support `index_merge` and `mem_index_merge`

### LINKED ISSUE #43013: The global index not work with IndexMerge executor
## Bug Report

```
set tidb_enable_global_index = true;
mysql> show create table tpk2;
+-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                                                                                                                                                                                             |
+-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| tpk2  | CREATE TABLE `tpk2` (
  `a` int(11) DEFAULT NULL,
  `b` int(11) DEFAULT NULL,
  `c` int(11) NOT NULL,
  `d` int(11) NOT NULL AUTO_INCREMENT,
  KEY `idx_bc` (`b`,`c`),
  UNIQUE KEY `uidx_ac` (`a`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin AUTO_INCREMENT=30002
PARTITION BY HASH (`c`) PARTITIONS 2 |
+-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

insert into tpk2 values (1, 1, 1, 1), (2, 2, 2, 2);

mysql> select /*+ use_index_merge(tpk2, uidx_ac, idx_bc) */ * from tpk2 where a=1 or b=2;
+------+------+---+---+
| a    | b    | c | d |
+------+------+---+---+
|    2 |    2 | 2 | 2 |
+------+------+---+---+
1 row in set (0.01 sec)

mysql> select * from tpk2 where a=1 or b=2;
+------+------+---+---+
| a    | b    | c | d |
+------+------+---+---+
|    2 |    2 | 2 | 2 |
|    1 |    1 | 1 | 1 |
+------+------+---+---+
2 rows in set (0.01 sec)
```

The result is same with using indexMerge or not.

The results are different.

Defined2014:

Issue Number: close #43013

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/index_merge_reader.go
+++ b/pkg/executor/index_merge_reader.go
@@ -315,6 +315,23 @@ func (e *IndexMergeReaderExecutor) startIndexMergeProcessWorker(ctx context.Cont
 	}()
 }
 
+func (e *IndexMergeReaderExecutor) getPartitalIndexKeyRanges(indexID int) (ret [][]kv.KeyRange) {
+	if !e.partitionTableMode {
+		return [][]kv.KeyRange{e.keyRanges[indexID]}
+	}
+
+	if idx := e.indexes[indexID]; idx != nil && idx.Global {
+		keyRange, _ := distsql.IndexRangesToKVRanges(e.ctx.GetDistSQLCtx(), e.table.Meta().ID, idx.ID, e.ranges[indexID])
+		ret = [][]kv.KeyRange{keyRange.FirstPartitionRange()}
+	} else {
+		for _, pKeyRanges := range e.partitionKeyRanges { // get all keyRanges related to this PartialIndex
+			ret = append(ret, pKeyRanges[indexID])
+		}
+	}
+
+	return ret
+}
+
 func (e *IndexMergeReaderExecutor) startPartialIndexWorker(ctx context.Context, exitCh <-chan struct{}, fetchCh chan<- *indexMergeTableTask, workID int) error {
 	failpoint.Inject("testIndexMergeResultChCloseEarly", func(_ failpoint.Value) {
 		// Wait for processWorker to close resultCh.
@@ -327,15 +344,7 @@ func (e *IndexMergeReaderExecutor) startPartialIndexWorker(ctx context.Context, 
 		e.dagPBs[workID].CollectExecutionSummaries = &collExec
 	}
 
-	var keyRanges [][]kv.KeyRange
-	if e.partitionTableMode {
-		for _, pKeyRanges := range e.partitionKeyRanges { // get all keyRanges related to this PartialIndex
-			keyRanges = append(keyRanges, pKeyRanges[workID])
-		}
-	} else {
-		keyRanges = [][]kv.KeyRange{e.keyRanges[workID]}
-	}
-
+	keyRanges := e.getPartitalIndexKeyRanges(workID)
 	failpoint.Inject("startPartialIndexWorkerErr", func() error {
 		return errors.New("inject an error before start partialIndexWorker")
 	})
@@ -376,10 +385,6 @@ func (e *IndexMergeReaderExecutor) startPartialIndexWorker(ctx context.Context, 
 						syncErr(ctx, e.finished, fetchCh, err)
 						return
 					}
-				}
-				if is.Index.Global {
-					keyRange, _ := distsql.IndexRangesToKVRanges(e.Ctx().GetDistSQLCtx(), is.Table.ID, is.Index.ID, e.ranges[workID])
-					keyRanges = [][]kv.KeyRange{keyRange.FirstPartitionRange()}
 				}
 				var builder distsql.RequestBuilder
 				builder.SetDAGRequest(e.dagPBs[workID]).
--- a/pkg/executor/index_merge_reader.go
+++ b/pkg/executor/index_merge_reader.go
@@ -103,6 +103,8 @@ type IndexMergeReaderExecutor struct {
 
 	// columns are only required by union scan.
 	columns []*model.ColumnInfo
+	// partitionIDMap are only required by union scan with global index.
+	partitionIDMap map[int64]struct{}
 	*dataReaderBuilder
 
 	// fields about accessing partition tables
@@ -127,9 +129,6 @@ type IndexMergeReaderExecutor struct {
 	memTracker *memory.Tracker
 	paging     bool
 
-	// checkIndexValue is used to check the consistency of the index data.
-	*checkIndexValue // nolint:unused
-
 	partialPlans        [][]base.PhysicalPlan
 	tblPlans            []base.PhysicalPlan
 	partialNetDataSizes []float64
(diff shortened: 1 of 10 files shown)


tests/integrationtest/t/globalindex/mem_index_merge.test
```diff
@@ -0,0 +1,103 @@
...
+
+--echo ## for indexMerge intersection with specified PARTITION
+--replace_regex /in\(_tidb_pid, [0-9]+\)/in(_tidb_pid, pid1)/
+explain select /*+ use_index_merge(tpk2, uidx_a, idx_c) */ * from tpk2 partition(p1) where a > 1 or c > 1;
+--sorted_result
+select /*+ use_index_merge(tpk2, uidx_a, idx_c) */ * from tpk2 partition(p1) where a > 1 or c > 1;
+--sorted_result
+select /*+ use_index_merge(tpk2, uidx_a, idx_c) */ * from tpk2 partition(p1) where a > 0 or c > 0;
+
+commit;
```
mjonss, line 96: Is it intentionally to have the commit here, instead of a rollback in the end? I.e. should the following test not use the mem_reader?

pkg/executor/index_merge_reader.go
```diff
@@ -1525,7 +1537,8 @@ func (w *indexMergeProcessWorker) fetchLoopIntersection(ctx context.Context, fet
 	batchSize := w.indexMerge.Ctx().GetSessionVars().IndexLookupSize
 
 	partCnt := 1
-	if w.indexMerge.partitionTableMode {
+	// To avoid multi-threaded access the handle map, we only use one worker for indexMerge with global index.
```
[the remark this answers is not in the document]
Defined2014: Yes, same as non-partition table.

pkg/executor/mem_reader.go
```diff
@@ -1018,46 +1022,81 @@ func (m *memIndexMergeReader) getMemRowsIter(ctx context.Context) (memRowsIter,
...
+			for _, handle := range handles {
+				if _, ok := handle.(kv.PartitionHandle); !ok && m.partitionMode {
+					pid := m.partitionTables[j].GetPhysicalID()
+					handle = kv.NewPartitionHandle(pid, handle)
+				}
+				if v, ok := hMap.Get(handle); !ok {
+					cnt := 1
+					hMap.Set(handle, &cnt)
+				} else {
+					*(v.(*int))++
```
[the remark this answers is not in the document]
Defined2014: > That only if every memory reader has a match, it should be included if intersect later?
Yes.

> Can it be that a single (reader) handle is found multiple times for the same reader?
No, each handle will only appear once in the same index worker.

tests/integrationtest/t/globalindex/mem_index_merge.test
Defined2014: Maybe I could execute before and after commit. The plan is a little bit different. In `mem_index_merge`, it not support pushdown order by with limit.

pkg/planner/core/indexmerge_path.go
```diff
@@ -446,7 +446,7 @@ func (*DataSource) buildIndexMergePartialPath(indexAccessPaths []*util.AccessPat
 }
 
 // buildIndexMergeOrPath generates one possible IndexMergePath.
-func (ds *DataSource) buildIndexMergeOrPath(
+func (*DataSource) buildIndexMergeOrPath(
```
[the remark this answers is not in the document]
Defined2014: Same as `buildIndexMergePartialPath` func, I think it's acceptable to keep it.

pkg/executor/mem_reader.go
```diff
@@ -1018,46 +1022,78 @@ func (m *memIndexMergeReader) getMemRowsIter(ctx context.Context) (memRowsIter,
...
-		for _, p := range m.partitionTables {
-			tbls = append(tbls, p)
+
+	var err error
+	hMap := kv.NewHandleMap()
+	for i, reader := range m.memReaders {
+		// [partNum][rangeNum]
+		var readerKvRanges [][]kv.KeyRange
+		if m.partitionMode {
+			if r, ok := reader.(*memIndexReader); ok && r.index.Global {
```
AilinKid, line 1032: seems it's part of range build-related logic, can we move this to `buildTableKeyRanges`?

AilinKid: and here we can only start with L1044

pkg/planner/core/indexmerge_path.go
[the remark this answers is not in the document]
AilinKid: better to reduce the pointer-receiver function definition, appreciate it

AilinKid (commented):
Rest LGTM

pkg/executor/mem_reader.go
```diff
@@ -1018,46 +1022,78 @@ func (m *memIndexMergeReader) getMemRowsIter(ctx context.Context) (memRowsIter,
...
+			} else {
+				for _, pKeyRanges := range m.partitionKVRanges { // get all keyRanges related to this PartialIndex
+					readerKvRanges = append(readerKvRanges, pKeyRanges[i])
+				}
+			}
+		} else {
+			readerKvRanges = [][]kv.KeyRange{m.indexMergeReader.keyRanges[i]}
+		}
+
+		for j, kr := range readerKvRanges {
```
AilinKid, line 1044: this function looks so long, why remove the original encapsulation here?

Defined2014: Because the original loop order is 
```
for each partition {
    for each index {
        for each range {
            ....
        }
    }
    do union or intersection
}
```

Now I changed it to
```
for each index {
    for each partition {
        for each range {
            ....
        }
    }
}

do union or intersection
```

The order of loops is same as `indexMergeExec`. The reason is for global index which is in partition mode, but it only has one partition. And it's hard to create table ranges by `distsql.TableHandlesToKVRanges` function.

pkg/executor/mem_reader.go
Defined2014: Done

pkg/executor/mem_reader.go
Defined2014: Done, I divided this function into two functions

AilinKid (approved):

PR #53007, opened 2024-04-30
expression: replace `mock.Context` with `StaticExprContext` in tests

### LINKED ISSUE #53006: replace mock context in `expression` with  `StaticExprContext`
## Enhancement

replace mock context in `expression` with  `StaticExprContext` to reduce the pkg depencies.

lcwangchao:

Issue Number: close #53006

Problem Summary:

After we introduced some static implements for `EvalContext` and `ExprContext`. It's better to replace `mock.Context` with `StaticExprContext` for the following reasons:

- reduce the dependencies of `expression`
- make the test easier to maintain without relying on a hug context 

- `mock.NewContext` is replaced with `mockStmtExprCtx`
- `createContext` is replaced with `mockStmtTruncateAsWarningExprCtx`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/expression/helper.go
```diff
@@ -73,7 +73,7 @@ func getTimeCurrentTimeStamp(ctx EvalContext, tp byte, fsp int) (t types.Time, e
 	}
 	value.SetCoreTime(types.FromGoTime(defaultTime.Truncate(time.Duration(math.Pow10(9-fsp)) * time.Nanosecond)))
 	if tp == mysql.TypeTimestamp || tp == mysql.TypeDatetime || tp == mysql.TypeDate {
-		err = value.ConvertTimeZone(time.Local, ctx.Location())
+		err = value.ConvertTimeZone(defaultTime.Location(), ctx.Location())
```
lcwangchao, line 76: This is the only line of the logic change in the production code.

When calling `ConvertTimeZone`, we use the timezone from `defaultTime` instead of local timezone to make sure the result is correct even if `getStmtTimestamp` returns a time that is not in `Local` timezone.

pkg/expression/helper.go
```diff
@@ -72,11 +72,8 @@ func getTimeCurrentTimeStamp(ctx EvalContext, tp byte, fsp int) (t types.Time, e
 		return value, err
 	}
 	value.SetCoreTime(types.FromGoTime(defaultTime.Truncate(time.Duration(math.Pow10(9-fsp)) * time.Nanosecond)))
-	if tp == mysql.TypeTimestamp || tp == mysql.TypeDatetime || tp == mysql.TypeDate {
-		err = value.ConvertTimeZone(time.Local, ctx.Location())
-		if err != nil {
-			return value, err
-		}
+	if err = value.ConvertTimeZone(defaultTime.Location(), ctx.Location()); err != nil {
+		return value, err
```
lcwangchao, line 76: This is the only file that has some logic change:

1. remove `tp == mysql.TypeTimestamp || tp == mysql.TypeDatetime || tp == mysql.TypeDate` because it seems useless. I check all the places that call `getTimeCurrentTimeStamp`. The tp is always time type. Only converting timezone for time types is also meaningless, so I removed it.

2. Change the first argument of `ConvertTimeZone` from `time.Local` to `defaultTime.Location()` because this the practical meaning for this line. Because `defaultTime` is always in local time zone if it is get from session, it is also right for previous code. But the `StaticEvalContext` returns a current time with the same zone of the context, it's better to make a change.

PTAL @AilinKid

pkg/expression/evaluator_test.go
```diff
@@ -99,44 +102,48 @@ func primitiveValsToConstants(ctx BuildContext, args []any) []Expression {
...
 
 func TestSleep(t *testing.T) {
-	ctx := createContext(t)
-	sessVars := ctx.GetSessionVars()
+	vars := variable.NewSessionVars(nil)
+	ctx := mockStmtTruncateAsWarningExprCtx(t, contextstatic.WithOptionalProperty(
+		contextopt.NewSessionVarsProvider(contextopt.SessionVarsAsProvider(vars)),
+	))
+	vars.TimeZone = ctx.GetEvalCtx().Location()
+	vars.StmtCtx.SetTimeZone(vars.Location())
```
AilinKid, line 110: so heavy to use, hard to say. @Defined2014 @zimulala

pkg/expression/helper.go
AilinKid: here lgtm

YangKeao: How about add a `intest.Assert` here?

(It's also fine to not add it, because IMO it's intuitive to call `getTimeCurrentTimeStamp` with time type.

YangKeao (approved):
Rest LGTM

Issue #53042, opened 2024-05-06
lightning failing the entire bulk load with Error 1205: Lock wait timeout exceeded; try restarting transaction"


mittalrishabh:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

Lightnining is failing the bulk load for intermittent bulk load requests. One common place where we see it failing is while updating the meta tasks. 
### 1. Minimal reproduce step (Required)

It was reproducible internally for intermittent connection issues. 

### 2. What did you expect to see? (Required)

I would expect retry with back off as connection errors are intermittent. 
### 3. What did you see instead (Required)
It is failing the whole bulk load. 
### 4. What is your TiDB version? (Required)
6.5

D3Hunter:
lightning only runs the sql, might be issue of tidb, can you provide more info, like which sql causes the error and tidb logs related to the sql

mittalrishabh:
i have created this PR https://github.com/pingcap/tidb/pull/53043

bb7133:
Change the tag to 'lightning'

lance6716:
closed by https://github.com/pingcap/tidb/pull/55670

PR #53043, opened 2024-05-06
Lightning: add retry if transaction failed while fetching task metas

mittalrishabh:

Issue Number: #53042

Problem Summary:
The transaction may encounter an error message stating "Error 1205: Lock wait timeout exceeded; try restarting transaction". This error can cause the import job to fail. The import job does not always start from the checkpoint, as it depends on the stage at which the failure occurred. If the transaction fails after the import/ingestion is completed, the job will not start from the checkpoint. As a result, the entire bulk load job needs to be restarted, which is a costly operation and can potentially violate our SLOs.
Add retry with back off if transaction fails. It is going to retry up to 5 times with maxbackoff of 30 seconds and baseline backoff of 1 second.
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/lightning/pkg/importer/meta_manager.go
+++ b/lightning/pkg/importer/meta_manager.go
@@ -671,58 +671,65 @@ func (m *dbTaskMetaMgr) CheckTasksExclusively(ctx context.Context, action func(t
 	if err != nil {
 		return errors.Annotate(err, "enable pessimistic transaction failed")
 	}
-	return exec.Transact(ctx, "check tasks exclusively", func(ctx context.Context, tx *sql.Tx) error {
-		rows, err := tx.QueryContext(
-			ctx,
-			common.SprintfWithIdentifiers(`
-				SELECT
-					task_id,
-					pd_cfgs,
-					status,
-					state,
-					tikv_source_bytes,
-					tiflash_source_bytes,
-					tikv_avail,
-					tiflash_avail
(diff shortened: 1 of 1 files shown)


guoshouyan:
Hi, I check the issue @mittalrishabh is reporting, I think we already have a retry mechanism [here](https://github.com/pingcap/tidb/blob/master/lightning/pkg/importer/meta_manager.go#L666), all we need to do is to add the `ErrLockWaitTimeout` to [here](https://github.com/pingcap/tidb/blob/master/pkg/lightning/common/retry.go#L126) so the error can be retried. What do you think @lance6716

lance6716:
> Hi, I check the issue @mittalrishabh is reporting, I think we already have a retry mechanism [here](https://github.com/pingcap/tidb/blob/master/lightning/pkg/importer/meta_manager.go#L666), all we need to do is to add the `ErrLockWaitTimeout` to [here](https://github.com/pingcap/tidb/blob/master/pkg/lightning/common/retry.go#L126) so the error can be retried. What do you think @lance6716

Yes I think make `ErrLockWaitTimeout` a retryable error is easier.

lance6716:
replaced by https://github.com/pingcap/tidb/pull/55670

PR #52993, opened 2024-04-29
*: caller can use a fixed TS for fast reorg

lance6716:
### What problem does this PR solve?

Issue Number: ref #49233 close #46986

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

```
GOCACHE=/tmp/failpoint-cache go build -o bin/tidb-server-failpoint -toolexec failpoint-toolexec ./cmd/tidb-server

tiup playground nightly --tiflash 0 --without-monitor --db 0

rm /tmp/tidb-{0,1}.log

GO_FAILPOINTS="github.com/pingcap/tidb/pkg/ddl/ingest/resignAfterFlush=return();github.com/pingcap/tidb/pkg/ddl/ingest/checkpointLoopExit=return();github.com/pingcap/tidb/pkg/ddl/ingest/forceSyncFlagForTest=return();github.com/pingcap/tidb/pkg/ddl/ownerResignAfterDispatchLoopCheck=return()" ./bin/tidb-server-failpoint -P 4000 --store=tikv --host=127.0.0.1 --status=10080 --path=127.0.0.1:2379 --log-file=/tmp/tidb-0.log

mysql --comments --host 127.0.0.1 --port 4000 -u root -e "set @@global.tidb_enable_dist_task = 0;set @@global.tidb_ddl_reorg_worker_cnt = 1;create table test.t(c int primary key, c2 int);split table test.t between (0) and (10000) regions 10;insert into test.t values (1,1), (3000, 3000), (6000, 6000);"

./bin/tidb-server-failpoint -P 4001 --store=tikv --host=127.0.0.1 --status=10081 --path=127.0.0.1:2379 --log-file=/tmp/tidb-1.log

mysql --comments --host 127.0.0.1 --port 4001 -u root -e "alter table test.t add unique index idx(c2);"
```

before

image

after

image

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

[...]

PR #53072, opened 2024-05-07
ddl: check local file existence before resume checkpoint

### LINKED ISSUE #53009: fast reorg checkpoint has wrong prerequisite
If the checkpoint record shows the it is generated by current node

https://github.com/pingcap/tidb/blob/c24dca58f5ee68941a656ac898136eaf63070703/pkg/ddl/ingest/checkpoint.go#L307-L308
 
we skip the data whose key is smaller that "local storage key watermark" and assume the data is already stay in local storage. 

https://github.com/pingcap/tidb/blob/c24dca58f5ee68941a656ac898136eaf63070703/pkg/ddl/ingest/checkpoint.go#L128-L137

However this is not reliable, especially we are using `/tmp` folder to store the data https://docs.pingcap.com/tidb/stable/command-line-flags-for-tidb-configuration#--temp-dir. If the node restarts in many OS `/tmp` folder will be cleaned.

lance6716:

Issue Number: close #53009

Problem Summary:

- use word "flushed"/"imported" to replace "local"/"global" to unify the terminology with lightning engine
- rename some methods
- remove `FlushModeForceLocalAndCheckDiskQuota` because it's not used
- use `ctx` to detect the closing of `CheckpointManager` and let other goroutine communications check the context
- add a tagged logger to `CheckpointManager`
- remove `updating` field because only one goroutine calls `updateCheckpoint`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/ingest/checkpoint.go
+++ b/pkg/ddl/ingest/checkpoint.go
@@ -34,52 +34,57 @@ import (
 	"go.uber.org/zap"
 )
 
-// CheckpointManager is a checkpoint manager implementation that used by non-distributed reorganization.
+// CheckpointManager is a checkpoint manager implementation that used by
+// non-distributed reorganization. It manages the data as two-level checkpoints:
+// "flush"ed to local storage and "import"ed to TiKV. The checkpoint is saved in
+// a table in the TiDB cluster.
 type CheckpointManager struct {
-	ctx       context.Context
-	flushCtrl FlushController
-	sessPool  *sess.Pool
-	jobID     int64
-	indexIDs  []int64
+	ctx           context.Context
+	cancel        context.CancelFunc
+	flushCtrl     FlushController
+	sessPool      *sess.Pool
+	jobID         int64
+	indexIDs      []int64
+	localStoreDir string
+	logger        *zap.Logger
 
 	// Derived and unchanged after the initialization.
 	instanceAddr     string
 	localDataIsValid bool
 
 	// Live in memory.
-	checkpoints     map[int]*TaskCheckpoint // task ID -> checkpoint
-	mu              sync.Mutex
-	minTaskIDSynced int
-	dirty           bool
+	mu          sync.Mutex
+	checkpoints map[int]*taskCheckpoint // task ID -> checkpoint
+	// we require each task ID to be continuous and start from 1.
+	minTaskIDFinished int
+	dirty             bool
 	// Local meta.
-	pidLocal   int64
-	startLocal kv.Key
-	endLocal   kv.Key
+	pidFlushed      int64
+	startKeyFlushed kv.Key
+	endKeyFlushed   kv.Key
 
 	// Persisted to the storage.
-	minKeySyncLocal  kv.Key
-	minKeySyncGlobal kv.Key
-	localCnt         int
-	globalCnt        int
+	minKeyFlushed  kv.Key
+	minKeyImported kv.Key
+	flushedKeyCnt  int
+	importedKeyCnt int
 	// Global meta.
-	pidGlobal   int64
-	startGlobal kv.Key
-	endGlobal   kv.Key
+	pidImported      int64
+	startKeyImported kv.Key
+	endKeyImported   kv.Key
 
 	// For persisting the checkpoint periodically.
-	updating      bool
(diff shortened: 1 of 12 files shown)


pkg/ddl/ingest/checkpoint.go
```diff
@@ -389,24 +416,24 @@ func (s *CheckpointManager) updateCheckpointLoop() {
...
 			err := s.updateCheckpoint()
 			if err != nil {
-				logutil.DDLIngestLogger().Error("update checkpoint failed", zap.Error(err))
+				s.logger.Error("update checkpoint failed", zap.Error(err))
 			}
-			wg.Done()
+			close(finishCh)
 		case <-ticker.C:
 			s.mu.Lock()
-			if !s.dirty || s.updating {
```
tangenta, line 400: If this is removed, is it possible to update checkpoint to a smaller value? For example,

1. g1 invokes updateCheckpoint, current checkpoint is 1.
2. g2 invokes updateCheckpoint, current checkpoint is 2.
3. g2 updates checkpoint 2 to table.
4. g1 updates checkpoint 1 to table.

lance6716: I see `updateCheckpoint` is only called by one goroutine which is `updateCheckpointLoop`

tangenta (approved):

pkg/ddl/ingest/checkpoint.go
```diff
@@ -34,52 +34,57 @@ import (
...
+	startKeyFlushed kv.Key
+	endKeyFlushed   kv.Key
 
 	// Persisted to the storage.
-	minKeySyncLocal  kv.Key
-	minKeySyncGlobal kv.Key
-	localCnt         int
-	globalCnt        int
+	minKeyFlushed  kv.Key
+	minKeyImported kv.Key
```
D3Hunter, line 68: shouldn't we call it maxKeyImported, it means all keys < it are imported

D3Hunter (approved):
rest lgtm

pkg/ddl/ingest/checkpoint.go
lance6716: ChatGPT still prefer the minXXX

---

You
How to name a variable which means "all keys less than it are flushed, but the keys larger than it can be flushed or not flushed". And it's calculated from the minimum value of some concurrency worker's progress.

ChatGPT
You might consider naming the variable something descriptive yet succinct, like "minFlushedKey" or "flushThreshold". This would indicate that it represents the minimum key value up to which all keys are flushed, but keys larger than it might or might not be flushed depending on their status. Alternatively, you could name it "minFlushedProgressKey" to be more explicit about its relationship to progress.

Here's a breakdown of each potential name:

minFlushedKey: This name clearly indicates that it represents the minimum key value that has been flushed. It suggests that any keys smaller than this value are flushed, while those larger than it may or may not be flushed.
flushThreshold: This name implies that it's a point or level that determines when flushing occurs. It's like a threshold that, once reached, triggers the flushing of keys. "FlushThreshold" conveys that keys below this threshold are flushed, while those above it may or may not be.
minFlushedProgressKey: This name explicitly links the variable to progress and flushing. It suggests that it represents the minimum key based on the progress made, and keys less than or equal to this value are considered flushed.
Choose the name that best fits the context and naming conventions of your project. The most important thing is that the name is clear and understandable to anyone reading your code.

D3Hunter: i don't think `minFlushedKey` suggests that any keys smaller than this value are flushed, it only states that it's the min key that have flushed and any key before or after it may or may not flushed

> And it's calculated from the minimum value of some concurrency worker's progress.

from this description, we can only got a name like min key of on-going task

lance6716: How about rename it to `flushedKeyLowWatermark`? I can do it in https://github.com/pingcap/tidb/pull/52993

D3Hunter: lgtm

Issue #53181, opened 2024-05-10
Support user-implemented authentication plugin within the extension framework


yzhan1:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently, TiDB only supports the built-in authentication and authorization mechanism. The built-in mechanism is not flexible enough to meet the needs of some users. For example, some users may want to use their own authentication and authorization logic, or some users may want to integrate TiDB with their existing authentication and authorization system.

Using the extension system to implement an auth plugin is a good way for users to plug in their own logic.

[Auth plugin](https://dev.mysql.com/doc/extending-mysql/8.0/en/writing-authentication-plugins-server-side.html) is a feature supported in MySQL, so implementing this feature in TiDB will make it more compatible with MySQL.

**Describe the feature you'd like:**

Provide support in the extension system where TiDB users can implement their own authentication plugins with customized authentication (login) and authorization (privilege checking) logic.

Example:

If user implements a plugin called `authentication_my_plugin`, the user should be able to create a user using:
```sql
CREATE USER 'my_user'@'%' IDENTIFIED WITH 'authentication_my_plugin' AS 'optional_authentication_pwd';
```
where authn/authz for `my_user` will go through the customized logic in the implementation of `authentication_my_plugin`.

**Describe alternatives you've considered:**

+ MySQL already supports customized auth plugins
+ Currently, additional privilege checks or access control needs to happen outside of TiDB in a service layer above

**Teachability, Documentation, Adoption, Migration Strategy:**

Design: https://github.com/pingcap/tidb/pull/53182

PR #53182, opened 2024-05-10
docs: design for extension auth plugin

yzhan1:

Issue Number: ref #53181

Problem Summary:

This is the design doc for #53181.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [X] No need to test
  > - [X] I checked and no code files have been changed.
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/docs/design/2024-05-10-extension-authentication-plugin.md
+++ b/docs/design/2024-05-10-extension-authentication-plugin.md
@@ -175,6 +175,10 @@ The above interfaces and parameters should give the users all information they n
 The feature needs to be compatible with existing user related statements and options, such as `CREATE USER`, `ALTER USER`, `SET PASSWORD`.
 However, it will not work with existing functionalities such as password expiry and password validation. Users can implement their own password validation logic inside the `GenerateAuthString` implementation.
 
+It will also not work with `CREATE VIEW SQL SECURITY DEFINER` since the view is created with the privileges of the view creator, and the plugin will not be able to check the privileges of the view creator because it
+does not have the `tls.ConnectionState` information for the definer to pass into `VerifyPrivilege`. This means that if a user is created using an auth plugin, it can only create views using `CREATE VIEW SQL SECURITY INVOKER` since
+at invocation time the plugin can check the privileges of the invoker and pass in the `tls.ConnectionState`.
+
 ## Test Design
 
 This feature requires tests focusing on functional and compatibility. Performance should not be a concern, since that depends on the actual auth plugin implementation and it is up to the users to keep the implementation performant if needed.
@@ -200,6 +204,8 @@ We will implement client to server tests to test out scenarios where MySQL users
 
 This feature should integrate well with all the existing test cases since it is an additional functionality. Existing tests related to authentication and authorization should continue to pass since they are not using auth plugins.
 
+We should add tests to verify that auth plugin users can only create views with `SQL SECURITY INVOKER` as mentioned in the compatibility section.
+
 ### Benchmark Tests
 
 This feature should not affect the performance of the TiDB as it simply provides interfaces for users to inject their own auth plugins via the extension system. The performance implication will come from the implementation of those plugins which depends on how the users of the framework want to use it.
@@ -209,6 +215,8 @@ This feature should not affect the performance of the TiDB as it simply provides
 If a user is using an auth plugin, the SQL privilege status in TiDB might not reflect the actual privileges, since there might be an outside system that decides whether the privilege check can actually pass. However, this is an intentional behavioral change when the user uses an auth plugin since we are providing users customizations on checks to run.
 
 The design does not account for implementing the same level of proxy user support as in MySQL auth plugin, since currently TiDB does not support [proxy users](https://dev.mysql.com/doc/extending-mysql/8.0/en/writing-authentication-plugins-proxy-users.html) yet. Also existing support for password expiry and validation will not work, as mentioned in the earlier section.
+
+Users created with auth plugins can only create views with `SQL SECURITY INVOKER` as well due to the missing `tls.ConnectionState` information for the definer at invocation time.
 
 
 ## Investigation & Alternatives
--- a/docs/design/2024-05-10-extension-authentication-plugin.md
+++ b/docs/design/2024-05-10-extension-authentication-plugin.md
@@ -175,9 +175,8 @@ The above interfaces and parameters should give the users all information they n
 The feature needs to be compatible with existing user related statements and options, such as `CREATE USER`, `ALTER USER`, `SET PASSWORD`.
 However, it will not work with existing functionalities such as password expiry and password validation. Users can implement their own password validation logic inside the `GenerateAuthString` implementation.
 
-It will also not work with `CREATE VIEW SQL SECURITY DEFINER` since the view is created with the privileges of the view creator, and the plugin will not be able to check the privileges of the view creator because it
-does not have the `tls.ConnectionState` information for the definer to pass into `VerifyPrivilege`. This means that if a user is created using an auth plugin, it can only create views using `CREATE VIEW SQL SECURITY INVOKER` since
-at invocation time the plugin can check the privileges of the invoker and pass in the `tls.ConnectionState`.
+It will also not work with `CREATE VIEW SQL SECURITY DEFINER` since the view is created with the privileges of the view creator, and if the creator is an auth plugin user, the plugin will not be able to check the privileges of the view creator because it
+does not have the `tls.ConnectionState` information for the definer to pass into `VerifyPrivilege`. This means we should not allow any users to create a view with `SQL SECURITY DEFINER` where the definer is an auth plugin user.
 
 ## Test Design
 
@@ -204,7 +203,7 @@ We will implement client to server tests to test out scenarios where MySQL users
 
 This feature should integrate well with all the existing test cases since it is an additional functionality. Existing tests related to authentication and authorization should continue to pass since they are not using auth plugins.
 
-We should add tests to verify that auth plugin users can only create views with `SQL SECURITY INVOKER` as mentioned in the compatibility section.
+We should add tests to verify that users cannot create views with `SQL SECURITY DEFINER` where the definer is an auth plugin user as mentioned in the compatibility section.
 
 ### Benchmark Tests
 
@@ -216,8 +215,7 @@ If a user is using an auth plugin, the SQL privilege status in TiDB might not re
 
 The design does not account for implementing the same level of proxy user support as in MySQL auth plugin, since currently TiDB does not support [proxy users](https://dev.mysql.com/doc/extending-mysql/8.0/en/writing-authentication-plugins-proxy-users.html) yet. Also existing support for password expiry and validation will not work, as mentioned in the earlier section.
 
-Users created with auth plugins can only create views with `SQL SECURITY INVOKER` as well due to the missing `tls.ConnectionState` information for the definer at invocation time.
-
+Also, users cannot create views with `SQL SECURITY DEFINER` where the definer is an auth plugin user because we will lack the `tls.ConnectionState` information for the definer to pass into `VerifyPrivilege`.
 
 ## Investigation & Alternatives
 
--- a/docs/design/2024-05-10-extension-authentication-plugin.md
+++ b/docs/design/2024-05-10-extension-authentication-plugin.md
@@ -1,0 +1,231 @@
+# Extension Authentication Plugin
+
+- Author: [Yaoming Zhan](http://github.com/yzhan1)
+- Discussion PR: TBD
+- Tracking Issue: https://github.com/pingcap/tidb/issues/53181
+
+## Table of Contents
+
+* [Introduction](#introduction)
+* [Motivation or Background](#motivation-or-background)
+* [Detailed Design](#detailed-design)
+* [Test Design](#test-design)
+    * [Functional Tests](#functional-tests)
+    * [Scenario Tests](#scenario-tests)
+    * [Compatibility Tests](#compatibility-tests)
+    * [Benchmark Tests](#benchmark-tests)
+* [Impacts & Risks](#impacts--risks)
+* [Investigation & Alternatives](#investigation--alternatives)
+* [Unresolved Questions](#unresolved-questions)
+
+## Introduction
+
+This design aims to introduce the extension auth plugin feature to TiDB. The extension auth plugin is a plugin that allows users to implement their own authentication and authorization logic. The plugin is loaded by TiDB and is responsible for authenticating users and authorizing their operations.
+
+[Auth plugin](https://dev.mysql.com/doc/extending-mysql/8.0/en/writing-authentication-plugins-server-side.html) is a feature supported in MySQL, so implementing this feature in TiDB will make it more compatible with MySQL.
+
+## Motivation or Background
+
+Currently, TiDB only supports the built-in authentication and authorization mechanism. The built-in mechanism is not flexible enough to meet the needs of some users. For example, some users may want to use their own authentication and authorization logic, or some users may want to integrate TiDB with their existing authentication and authorization system. Using the extension system to implement an auth plugin is a good way for users to plug in their own logic.
+
+### Current Code Path
+
+Here we only cover the critical paths involving authentication and authorization.
+
+Authentication:
++ `server.onConn`
+  + A new connection is created and handled using a separate goroutine
++ `conn.handshake`
+  + Perform handshake when a new handshake event is received
++ `conn.readOptionalSSLRequestAndHandshakeResponse`
+  + SSL handshake completed before running custom auth code
+  + This function handles validating and aligning the plugin used for connection on both client and server, and reading the password from client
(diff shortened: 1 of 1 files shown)


docs/design/2024-05-10-extension-authentication-plugin.md
```diff
@@ -0,0 +1,231 @@
...
+    // Name returns the name of the auth plugin. It will be registered as a system variable in TiDB which can be used inside the `CREATE USER ... IDENTIFIED WITH 'plugin_name'` statement.
+    Name() string
+    
+    // PasswordReader is a function to communicate with the client during handshake phase to 
+    // read/write more data to the client-side plugin using authConn. For example, sending extra salt to the client.
+    // If `plugin.PasswordReader()` returns nil, the connection will default to mysql_clear_password plugin which means the password will be passed from the client-side in clear text format. Otherwise, the connection will use `plugin.Name()` as the current plugin
+    // When `plugin.PasswordReader()(ctx, authConn, resp)` is called, it should return the final password received from the client in bytes format. This bytes password will be passed into `plugin.Authenticate` below.
+    // ctx: context passed in from caller
+    // authConn: interface for the plugin to communicate with the client
+    PasswordReader() func(ctx context.Context, authConn conn.AuthConn) ([]byte, error)
```
lcwangchao, line 93: Maybe "tell the client to use a specified plugin proto" and "read the custom password" are two orthogonal functions. Sometimes client may want to keep simple and only use the exist plugins. For example, if we want to introduce a new plugin `caching_sha2_password_enhanced`, it not only checks the passwords from the user as the old way but also does some custom verifications of the SSL certificates or some other checks. We have to implement a client-side plugin because `caching_sha2_password` cannot be used by client in the current design.

How about just give a method: `RequireClientSidePlugin() string` to just indicate the client-side plugin it should use. If it is not set, the plugin in name is used.

`PasswordReader` is not necessary in this way, because the coder can do it in `Authenticate` internally by calling related method of `conn.AuthConn`.

docs/design/2024-05-10-extension-authentication-plugin.md
```diff
@@ -0,0 +1,231 @@
...
+    
+    // Authenticate is called when a client connects to the server as a user and the server authenticates the user.
+    // If an error is returned, the login attempt fails, otherwise it succeeds.
+    // authUser: the username in the connect attempt
+    // storedPwd: the user's password stored in mysql.user table
+    // inputPwd (authentication): the user's password passed in from the connection attempt in bytes (return value of `plugin.PasswordReader()(ctx, authConn, resp)` if `plugin.PasswordReader() != nil`)
+    // salt: randomly generated salt for the current connection
+    // connState: the TLS connection state (contains the TLS certificate) if client is using TLS. It will be nil if the client is not using TLS
+    // authConn: interface for the plugin to communicate with the client
+    Authenticate(authUser string, storedPwd string, inputPwd []byte, salt []byte, connState *tls.ConnectionState, authConn conn.AuthConn) error
```
lcwangchao, line 103: As explained above, we can read the custom password in `Authenticate` without implementing a new method like `PasswordReader`

docs/design/2024-05-10-extension-authentication-plugin.md
```diff
@@ -0,0 +1,231 @@
...
+    
+    // EncodePassword is a function for user to implement customized ways to encode the password (e.g. hash/salt/clear-text). The returned string will be stored as the encoded password in the mysql.user table.
+    // If the input password is considered as invalid, this should return an error.
+    // pwd: User's input password in CREATE/ALTER USER statements in clear-text
+    EncodePassword(pwd string) (string, bool)
+    
+    // IsValidPasswordHash checks if the password hash stored in the mysql.user table is valid.
+    // This is called when retrieving an existing user to make sure the password stored is valid and not modified.
+    // pwdHash: hash of the password stored in the internal user table
+    IsValidPasswordHash(pwdHash string) bool
```
lcwangchao, line 113: Are `GenerateAuthString`/`ValidateAuthString` better names? Because mysql use `generate_authentication_string ` and `validate_authentication_string` as the function names: https://dev.mysql.com/doc/extending-mysql/8.0/en/writing-authentication-plugins-server-side.html

docs/design/2024-05-10-extension-authentication-plugin.md
```diff
@@ -0,0 +1,231 @@
...
+
+Technically, we could implement PAM support in TiDB instead of what is being proposed in this design, but it will just be one standard authentication option which does not utilize the existing extension framework. Implementing auth plugin extension support is more flexible as it allows many different customizations for both authentication and authorization. With the auth plugin extension support in place, we could even consider implementing the PAM auth scheme as an auth plugin.
+
+
+## Unresolved Questions
+
++ Is the `IsValidPasswordHash` function really needed? Why do we want to do a password validity check for passwords fetched from the internal user table?
```
lcwangchao, line 231: I think it is used for `CREATE USER ... AS ...` as the description in MySQL:

> - IDENTIFIED WITH auth_plugin AS 'auth_string'
Sets the account authentication plugin to auth_plugin and stores the 'auth_string' value as is in the mysql.user account row. If the plugin requires a hashed string, the string is assumed to be already hashed in the format the plugin requires.

It is used to check the hashed string from the input.

docs/design/2024-05-10-extension-authentication-plugin.md
```diff
@@ -0,0 +1,231 @@
...
+    // Returns true if user has the requested privilege.
+    // activeRoles: list of active MySQL roles for the current user
+    // user: current user's name
+    // host: the host that the user is connecting from
+    // db: the database to check for privilege
+    // table: the table to check for privilege
+    // column: the column to check for privilege (currently just a placeholder in TiDB as column-level privilege is not supported by TiDB yet)
+    // priv: the privilege type of the SQL statement that will be executed
+    // connState: the TLS connection state (contains the TLS certificate) if client is using TLS. It will be nil if the client is not using TLS
+    VerifyPrivilege(activeRoles []*auth.RoleIdentity, user, host, db, table, column string, priv mysql.PrivilegeType, connState *tls.ConnectionState) bool
```
lcwangchao, line 126: If two connections are identified with the same user and the same plugin, is this method expected to return the same value when the input arguments are also the same?

docs/design/2024-05-10-extension-authentication-plugin.md
```diff
@@ -0,0 +1,231 @@
...
+    + Alter user changes can be applied correctly
+    + Passwords can be encoded with the API defined in the plugin
++ When the connection changes between users, the system should behave correctly
+  + From non-plugin user to plugin user
+  + From plugin user to non-plugin user
+  + From plugin user to plugin user
+
+### Compatibility Tests
+
+This feature should integrate well with all the existing test cases since it is an additional functionality. Existing tests related to authentication and authorization should continue to pass since they are not using auth plugins.
```
lcwangchao, line 203: Maybe we need to define the behavior of some other command like `SHOW GRANTS FOR <someuser>` when "someuser" uses the custom plugin.

lcwangchao: Or the view behavior when a view's definer is a user with a custom plugin. (Whether or not to check a definer's privilege declared in the plugin).

docs/design/2024-05-10-extension-authentication-plugin.md
```diff
@@ -0,0 +1,231 @@
...
+If a user is using an auth plugin, the SQL privilege status in TiDB might not reflect the actual privileges, since there might be an outside system that decides whether the privilege check can actually pass. However, this is an intentional behavioral change when the user uses an auth plugin since we are providing users customizations on checks to run.
+
+The design does not account for implementing the same level of proxy user support as in MySQL auth plugin, since currently TiDB does not support [proxy users](https://dev.mysql.com/doc/extending-mysql/8.0/en/writing-authentication-plugins-proxy-users.html) yet. Also existing support for password expiry and validation will not work, as mentioned in the earlier section.
+
+
+## Investigation & Alternatives
+
+### MySQL
+
+MySQL supports customizing authentication methods with [authentication plugins](https://dev.mysql.com/doc/extending-mysql/8.0/en/writing-authentication-plugins-server-side.html). This design proposes a similar set of interfaces for users to define, compared to those in MySQL. The main differences are that we also allow defining authorization checks (SQL privileges).
```
CbcWestwolf, line 220: I am not pretty sure the necessity of customizing authorization checks, especially `VerifyPrivilege`. Normally for a basic `SELECT` statement, only `SELECT` privilege is required. Checking other privileges, like `INSERT`, for a read-only `SELECT` statement would be weird. Since `mysql.PrivilegeType` is not extensible, is there any scenario that additional privileges would be required by customizing authenrization?

CbcWestwolf (approved):
LGTM

docs/design/2024-05-10-extension-authentication-plugin.md
```diff
@@ -0,0 +1,231 @@
...
+
+After this, `my_user`’s connection verification and privilege verification will be handled by the plugin in addition to the existing default TiDB checks. Users not identified with the plugin will not go through the plugin authorization and authentication checks.
+
+### Interface
+We propose a set of interfaces for users to implement an auth plugin that are similar to what is offered in MySQL’s authentication plugins. We will modify TiDB’s code to call these interfaces during the authentication/authorization flows. Most of these interfaces will receive parameters that are used in the existing auth flows in TiDB, so that the plugin developer can access as much information as possible.
+
+Following is the interface for an auth plugin:
+```go
+// AuthPlugin is an interface for authentication plugins.
+type AuthPlugin interface {
```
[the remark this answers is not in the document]
yzhan1: Good idea. Will make this a struct instead.

One issue is that user might not implement all the functions required, which could cause NPE when invoking them. When we register the plugin in the manifest, we can check whether all required functions are not `nil`, otherwise return error and fail the start up. What do you think?

docs/design/2024-05-10-extension-authentication-plugin.md
yzhan1: That would depend on the implementation of it. The developer could choose to return different values based on the use case.

docs/design/2024-05-10-extension-authentication-plugin.md
yzhan1: There could be use cases where there is a centralized privilege management system which the user would like to integrate TiDB with. In this case, the user could make RPC calls to that system and make a final authorization decision based on that (since based on the design, at this point user already has enough privilege in MySQL).

Basically users can implement their own customized authorization logic on top of the MySQL privilege checks. If the statement requires only `SELECT` privilege, then `VerifyPrivilege` will only receive one call with the `SELECT` privilege according to the current implementation in TiDB, but it's true that there's nothing stopping the user to check for `INSERT` privilege if they somehow actually want to.

docs/design/2024-05-10-extension-authentication-plugin.md
yzhan1: Thanks for the explanation. I will remove this from the question then.

docs/design/2024-05-10-extension-authentication-plugin.md
yzhan1: `EncodePassword` and `IsValidPasswordHash` are picked to be in-sync with current naming in TiDB, but yes I think it's better to match MySQL naming as close as possible. Will update the doc.

docs/design/2024-05-10-extension-authentication-plugin.md
yzhan1: I see. I can add the function `RequiredClientSidePlugin() string` which returns the client-side plugin to use.

I did a quick test and I was able to implement the exact same server-side behavior for `caching_sha2_password` by using `authConn` in `Authenticate`. I will remove `PasswordReader` then. Thanks for the suggestion!

docs/design/2024-05-10-extension-authentication-plugin.md
yzhan1: > define the behavior of some other command like `SHOW GRANTS FOR <someuser>` when "someuser" uses the custom plugin.

I intend to keep the behavior the same as the current implementation, where `SHOW GRANTS` will only display existing grants based on internal MySQL privilege table. The reason is that if `VerifyPrivilege` depends on an outside system, then the privileges granted might require other information such as the SSL certificates, which is not available as a static information when fetching the MySQL grants. Please let me know what do you think.

> view behavior when a view's definer is a user with a custom plugin. (Whether or not to check a definer's privilege declared in the plugin).

I'm not too sure if we need to change anything for this. IIUC view definer's privilege is only checked when running `CREATE VIEW`, which the plugin will need to call `VerifyPrivilege` to check if the definer has `CREATE VIEW` priv. Other operations for the view such as read also only requires the plugin to verify the caller's `SELECT` priv for the view. I don't think these behavior would change depending on the definer's privilege. Please correct me if I'm wrong.

lcwangchao: What I am concerned about is whether it is easy for a coder to avoid some unexpected bugs. This is related to the previous question I asked about the method `VerifyPrivilege/VerifyDynamicPrivilege`. `VerifyPrivilege` is defined as:

```
VerifyPrivilege func(activeRoles []*auth.RoleIdentity, user, host, db, table, column string, priv mysql.PrivilegeType, connState *tls.ConnectionState) bool
```

The last input argument is `connState` with type `*tls.ConnectionState`. I think you added this argument for giving different privileges for connections but with the same user, right? In this way, we can create only one user in TiDB but distinguish the "real users" from the SSL certificate. So different connections may have different privileges even they are using the same "TiDB user".

But this may bring some trouble. If a custom user "myuser1" creates a view, the definer of this view is "myuser1" by default. If another "user2" then logins without any plugin and then select this view, it should also check the privilege of "myuser1". At this time, what is the right `connState` for the method `VerifyPrivilege`? Though we can pass the tls of the current connection, but I think it is not expected. In fact, I'm not sure what is the right behavior for this scene...

lcwangchao: Maybe we can forbid a user to create a view/procedure if it uses custom plugin and  `VerifyPrivilege/VerifyDynamicPrivilege` is specified?

yzhan1: There seems to be [two security modes](https://dev.mysql.com/doc/refman/8.0/en/create-view.html) for checking view privileges:
1. DEFINER
2. INVOKER

We can probably forbid plugin users to create view using `CREATE VIEW SQL SECURITY DEFINER ...` and only allow `CREATE VIEW SQL SECURITY INVOKER ...`, so that the `VerifyPrivilege` will only be invoked for the current connection user who is running the `SELECT` statement using its `connState`.

CbcWestwolf (approved):

docs/design/2024-05-10-extension-authentication-plugin.md
yzhan1: @lcwangchao  I have added the above content to the design doc. Please review. Thank you!

docs/design/2024-05-10-extension-authentication-plugin.md
```diff
@@ -0,0 +1,237 @@
...
+
+The above interfaces and parameters should give the users all information they need to make a decision on the authentication and authorization. For authorizations before each request, the MySQL user should already be granted with the corresponding privileges in MySQL using `GRANT` statements.
+
+### Compatibility
+The feature needs to be compatible with existing user related statements and options, such as `CREATE USER`, `ALTER USER`, `SET PASSWORD`.
+However, it will not work with existing functionalities such as password expiry and password validation. Users can implement their own password validation logic inside the `GenerateAuthString` implementation.
+
+It will also not work with `CREATE VIEW SQL SECURITY DEFINER` since the view is created with the privileges of the view creator, and the plugin will not be able to check the privileges of the view creator because it
+does not have the `tls.ConnectionState` information for the definer to pass into `VerifyPrivilege`. This means that if a user is created using an auth plugin, it can only create views using `CREATE VIEW SQL SECURITY INVOKER` since
```
lcwangchao, line 179: I think we can just forbid creating a view by setting `DEFINER` to a plugin user, that is:

- It is NOT allowed for ANY user to create a view by setting `DEFINER` to a "plugin user".
- It is ALLOWED for ANY user to create a view by setting `DEFINER` to a "normal user".

However, when I checked the mysql's behavior and found that the definer will not be verified when creating a view...

```
mysql> create user testuser1;

mysql> create definer=testuser1 view v1 as select * from t;
```

Maybe we can check the `DEFINER` as mysql's behavior to postpone it to select? Anyway, we only need to forbid a plugin user as the definer, that is enough.

lcwangchao: Moreover, if a plugin does not define `RequestVerification` or `VerifyDynamicPrivilege`, the user that uses this plugin can still be a definer, I think

lcwangchao (commented):
A comment about compatibility. Rest LGTM

docs/design/2024-05-10-extension-authentication-plugin.md
yzhan1: @lcwangchao That makes sense. I have updated the doc to reflect this. Please review again. Thanks!

lcwangchao (approved):

Issue #45133, opened 2023-07-03
Global Index for partitioned tables


mjonss:
## Enhancement

Development tasks for:
Make GLOBAL INDEX General Available

Stage 1 - full implementation, but Experimental

- [x] #48825
- Add tests for all index access plans. Create issues for handling the index access plans currently not supporting Global indexes (first disable the specific plan, then add support).
  - [x] memIndexReader and IndexReader https://github.com/pingcap/tidb/pull/52660
  - [x] memIndexLookUp and IndexLookUp
  - [x] memIndexMerge and IndexMerge https://github.com/pingcap/tidb/pull/52971
  - [x] pointGet and batchPointGet
- Tests 
  - [ ] Add tests for all write transactions (select lock, commit, rollback, different transaction modes, different transaction levels) * (insert, update, delete, select for update, single/multi statement transaction).
- DDL
  - ~~[ ] Support in EXCHANGE PARTITION? #54056~~ Will not support EXCHANGE PARTITION for Global Index, will cause too long state where double writing is needed. Better to add support for CONVERT PARTITION/TABLE instead, possibly later.
  - [x] Support in REORGANIZE PARTITION #53277
  - [x] Support in TRUNCATE PARTITION #55831
  - [x] Support in DROP PARTITION ? + #56082
- Admin
  - [x] #53439
  - [x] #53019
- Miscs
  - [x] #53342
  - [x] #55452 #56028

Stage 2 - Stabilise / General Available
- [x] Add docs https://github.com/pingcap/docs/pull/18543
- [ ] Fix reported bugs
- [x] Adjust syntax if needed https://github.com/pingcap/tidb/pull/55259
- [ ] Add more tests if needed
  - [x] https://github.com/pingcap/tidb/pull/57493
  - [x] #57623
  - [ ] https://github.com/pingcap/tiflow/pull/11651
- [x] Enable by default and remove experimental flag #56169

Optional / future enhancements
- [x] Change syntax, to enable (default) LOCAL index, and (explicit) GLOBAL index, #52994 
- [x] Consider path is global index path or not in `skylinePruning` function, let user easier to select global index. #56196

Defined2014:
Maybe support for `indexMerge` and `UnionScan` can be added, it seems we are not considering global indexes in those related codes.

311ybb:
please make  global index   General Available faster, we  need it !

[...]

PR #53277, opened 2024-05-14
ddl: Reorganize partition supporting global index

mjonss:

Issue Number: ref #45133
REORGANIZE PARTITION is not allowed for tables with GLOBAL INDEX

Problem Summary:
This was disabled, since Global index was not stable. Now when we are stabilizing Global index, this is a functionality blocker.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/ddl/partition.go
```diff
@@ -3415,18 +3494,26 @@ func (w *reorgPartitionWorker) GetCtx() *backfillCtx {
...
-	// First copy all table data to the new partitions
+func (w *worker) reorgPartitionDataAndIndex(t table.Table, reorgInfo *reorgInfo) (err error) {
+	// First copy all table data to the new AddingDefinitions partitions
 	// from each of the DroppingDefinitions partitions.
-	// Then create all indexes on the AddingDefinitions partitions
-	// for each new index, one partition at a time.
+	// Then create all indexes on the AddingDefinitions partitions,
+	// both new local and new global indexes
+	// And last update new global indexes from the non-touched partitions
+	// Note it is hard to update global indexes in-place due to:
```
Defined2014, line 3503: Does it mean even for `reorg partition` op, we will remove all key/value pairs not only relevant partitions?

Defined2014 (commented):
Try SQL below, reports a error.
```
CREATE TABLE `members` (
  `id` int(11) DEFAULT NULL,
  `fname` varchar(255) DEFAULT NULL,
  `lname` varchar(255) DEFAULT NULL,
  `dob` date DEFAULT NULL,
  `data` json DEFAULT NULL,
  UNIQUE KEY `ui` (`id`)
PARTITION BY RANGE (YEAR(`dob`))
(PARTITION `pBefore1950` VALUES LESS THAN (1950),
 PARTITION `p1950` VALUES LESS THAN (1960),
 PARTITION `p1960` VALUES LESS THAN (1970),
 PARTITION `p1970` VALUES LESS THAN (1980),
 PARTITION `p1980` VALUES LESS THAN (1990),
 PARTITION `p1990` VALUES LESS THAN (2000))
```

```
mysql> ALTER TABLE members REORGANIZE PARTITION `p1990` INTO (PARTITION p1995 VALUES LESS THAN (1995),  PARTITION p2000 VALUES LESS THAN (2010));
ERROR 1105 (HY000): partition id not found 126
```

pkg/ddl/ddl_api.go
```diff
@@ -4626,14 +4626,15 @@ func (d *ddl) AlterTablePartitioning(ctx sessionctx.Context, ident ast.Ident, sp
 				return err
 			}
 			if !ck {
-				if ctx.GetSessionVars().EnableGlobalIndex {
-					return dbterror.ErrCancelledDDLJob.GenWithStack("global index is not supported yet for alter table partitioning")
-				}
-				indexTp := "UNIQUE INDEX"
-				if index.Primary {
-					indexTp = "PRIMARY"
+				if !ctx.GetSessionVars().EnableGlobalIndex || index.Primary {
```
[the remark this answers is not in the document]
mjonss: Short answer, we don't need to support it right now, but we should create a feature request for it.

Long answer:
My thoughts about Primary Key and Global key and if it is clustered or non-clustered.

So a primary key cannot be a global index, i.e. the primary key must include all partitioning columns.
Why? To ensure the uniqueness at table level, and, in my opinion, ensure that a primary key lookup is a single index lookup if clustered, and at most 2 lookups if non-clustered; pk index + heap (_tidb_rowid) 

If the PK is clustered and the table is partitioned, then the actual PK would need to be both clustered and include the partitioning column, to ensure a single key lookup.

So if the PK is non-clustered, could not the PK just be the Global Index?
I would say it could, but I don't think we need to support it yet, since one can accomplish the same with an unique index.

Non-Clustered index example, what we would like to have:
CREATE TABLE t (pk_col varchar(255), partition_col int, data_col JSON, primary key (pk_col) nonclustered) PARTITION BY HASH(partition_col) PARTITIONS 3
table/partition data:
Key: _tidb_rowid
Val: pk_col + partition_col + data_col
PK index:
Key: pk_col <= This is currently not supported
Val: partition_col + data_col
Global index, will be the same as the PK:
Key: pk_col
Val: partition_col + data_col

So what we can do is:
CREATE TABLE test.t (pk_col varchar(255) not null, partition_col int, data_col JSON, unique key (pk_col)) PARTITION BY HASH(partition_col) PARTITIONS 3;
Which will create a non-clustered table the way we want it (pk_col is really the PK, just not explicitly named as such!)
Table/partition heap/non-clustered table:
Key: _tidb_rowid
Val: pk_col + partition_col + data_col
Global, Unique Key:
Key: pk_col
Val: Partition ID + _tidb_rowid

But what is missing is to allow partition a table with a PRIMARY KEY, which could be allowed, if it was an non-clustered primary key.



For clustered primary key, what we would prefer to do is:
CREATE TABLE t (pk_col varchar(255), partition_col int, data_col JSON, primary key (pk_col) clustered) PARTITION BY HASH(partition_col) PARTITIONS 3
Clustered index example:
Key: pk_col <= This would not allow a single key lookup, since it would need the partition_col as well.
Val: partition_col + data_col
+ a global index:
Key: pk_col <= This would add the lookup of the Global index.
Val: partition_col
But it has to be:
CREATE TABLE t (pk_col varchar(255), partition_col int, data_col JSON, primary key (pk_col, partition_col) clustered, unique key (pk_col)) PARTITION BY HASH(partition_col) PARTITIONS 3;
Key: pk_col + partition_col
Val: data_col
+ a global index:
Key: pk_col
Val: partition_col (+ pk_col?)
Which means that tidb could enhance the optimizer to read from the actual primary key, skipping the lookup through the global index (only using it for writes, to ensure uniqueness).

What would be needed to further discuss is which decision TiDB should take and which decisions should be left to the user/admin, like:
- Adding a global index, resulting in Partition management commands like drop/truncate/exchange partition is no-longer near instant, but now would take some time to reorganize the global index.
- Adding a global index for a clustered table would really have both the partitioned clustered PK a global unique non-null index.
These questions are also connected to if we should display if an index is global, in SHOW CREATE TABLE, and also accept GLOBAL and possibly LOCAL as index options.

tests/integrationtest/t/ddl/integration.test
```diff
@@ -92,17 +92,17 @@ set tidb_enable_global_index=1;
 create table t3(a int not null, b int, primary key(a) nonclustered, unique idx_b(b)) partition by hash(a) partitions 3;
 drop table if exists t4;
 create table t4(a int not null, b int, primary key(a)) partition by hash(a) partitions 3;
--- error 8214
 alter table t partition by hash(a) partitions 3;
--- error 8214
+alter table t remove partitioning;
```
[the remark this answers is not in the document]
mjonss: Great find! Let us continue the connected issue in the new issue.

pkg/ddl/partition.go
mjonss: Yes, even for reorganize partition, we will recreate a full new global index.

The reason is that there is a DDL state transition where the previous schema version sees the old partitions and the current schema version sees the new partitions, and a single global index cannot represent a row that would match two different partitions (one for old and one for new).

mjonss:
> Try SQL below, reports a error.
> ...
Nice catch, I fixed it and added the test to the integration test.

pkg/ddl/ddl_api.go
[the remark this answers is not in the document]
Defined2014:
```
mysql> create table t(a int, b int) partition by hash(b) partitions 5;

mysql> alter table t add primary key (a);
Query OK, 0 rows affected (0.34 sec)

mysql> alter table t remove partitioning;
Query OK, 0 rows affected (0.20 sec)

mysql> alter table t partition by hash(b) partitions 5;
ERROR 1503 (HY000): A PRIMARY must include all columns in the table's partitioning function
```

Based on the SQLs above, I think we should let the behaviours reasonable.

pkg/ddl/partition.go
Defined2014: Could you provide a state change graph include job state and global index state? Maybe it's more easier to understand it.

mjonss: @Defined2014 please check the comments I added [here](https://github.com/pingcap/tidb/pull/53277/files#diff-a558fe65fa5ae53cf0710c8d700cbfd155d3ba42f4ed85494c01f9c1d24bbe6bR2811-R2875), are they what you were looking for?

pkg/ddl/ddl_api.go
```diff
@@ -4626,18 +4626,27 @@ func (d *ddl) AlterTablePartitioning(ctx sessionctx.Context, ident ast.Ident, sp
...
 			}
 			if !ck {
-				if ctx.GetSessionVars().EnableGlobalIndex {
-					return dbterror.ErrCancelledDDLJob.GenWithStack("global index is not supported yet for alter table partitioning")
+				indexTp := ""
+				if !ctx.GetSessionVars().EnableGlobalIndex {
+					if index.Primary {
+						indexTp = "PRIMARY KEY"
+					} else {
+						indexTp = "UNIQUE KEY"
```
Defined2014, line 4634:
```suggestion
						indexTp = "UNIQUE INDEX"
```

Better same as `checkPartitioningKeysConstraints` function

pkg/ddl/partition.go
```diff
@@ -3892,8 +4076,8 @@ func checkPartitioningKeysConstraints(sctx sessionctx.Context, s *ast.CreateTabl
 	for _, index := range tblInfo.Indices {
 		if index.Unique && !checkUniqueKeyIncludePartKey(partCols, index.Columns) {
 			if index.Primary {
-				// not support global index with clustered index
-				if tblInfo.IsCommonHandle {
+				// global index does not support clustered index
+				if tblInfo.PKIsHandle || tblInfo.IsCommonHandle {
```
Defined2014, line 4080: I think no need check `tblInfo.PKIsHandle` here, because 
> when PKIsHandle, tblInfo.Indices will not contain the primary key.

Defined2014 (approved):
Two tiny comments, reset LGTM

AilinKid:
great work, it seems most of this pull request is related to the DDL module, I'm not quite familiar with this part. @zimulala @tangenta

jiyfhust (approved):

tangenta (commented):
7/11, will review later

pkg/ddl/partition.go
```diff
@@ -2872,6 +2936,33 @@ func (w *worker) onReorganizePartition(d *ddlCtx, t *meta.Meta, job *model.Job)
...
+			}
+			inAllPartitionColumns, err := checkPartitionKeysConstraint(partInfo, index.Columns, tblInfo)
+			if err != nil {
+				return ver, errors.Trace(err)
+			}
+			if index.Global || !inAllPartitionColumns {
+				// Duplicate the unique indexes with new index ids.
+				// If previously was Global or will be Global:
+				// it must be recreated with new index ID
+				newIndex := *index
```
tangenta, line 2954: It is better to use `index.Clone()`, otherwise `index.Columns` reference the same slice, which may be changed unexpectedly.

mjonss: Good catch, updated.

pkg/ddl/partition.go
```diff
@@ -2872,6 +2936,33 @@ func (w *worker) onReorganizePartition(d *ddlCtx, t *meta.Meta, job *model.Job)
...
+				newIndex := index.Clone()
+				newIndex.State = model.StateDeleteOnly
+				newIndex.ID = AllocateIndexID(tblInfo)
+				if inAllPartitionColumns {
+					newIndex.Global = false
+				} else {
+					// If not including all partitioning columns, make it Global
+					newIndex.Global = true
+				}
+				tblInfo.Indices = append(tblInfo.Indices, newIndex)
```
tangenta, line 2963: It seems there is a chance to have two "primary keys", I am not sure if there are problems

```sql
CREATE TABLE `t3` (
  `a` int(11) NOT NULL,
  `b` int(11) DEFAULT NULL,
  PRIMARY KEY (`a`) /*T![clustered_index] NONCLUSTERED */,
  UNIQUE KEY `idx_b` (`b`),
  UNIQUE KEY `idx_a` (`a`)
PARTITION BY HASH (`a`) PARTITIONS 3

alter table t3 partition by hash(b) partitions 3;
```

tangenta: Maybe we can add DML tests during reorganizing partitions in next PR.

mjonss: There should only be one 'PRIMARY KEY' (due to definition of what a primary key is) but multiple candidate keys are possible (i.e. unique indexes of not null columns), both using global index as well as non-partitioned tables.
Also I don't think there are any limitations for having duplicate indexes with different names, are there?

tangenta: > Also I don't think there are any limitations for having duplicate indexes with different names, are there?

+1

I mean there are multiple indexes with `index.Primary == true`, it breaks the assumption of "there is only one primary index at anytime". I can't figure out any scenarios that can cause errors for now

tangenta (approved):

PR #53301, opened 2024-05-15
*: keep in-memory resident for model.TableInfo with special attributes

tiancaiamao:

Issue Number: ref #50959 

Problem Summary:

`SchemaTables()` is a big problem for infoschema v2. It's slow, it cost a lot CPU resource, it make tidb memory thrash ... 

I observe the pattern of call SchemaTables() API in our current codebase look-alike.
For ttl, tiflash and so on, they always iterate the SchemaTables and filter out some attributes.

```
	for _, dbName := range is.AllSchemaNames() {
		for _, tblInfo := range is.SchemaTableInfos(dbName) {
			if tblInfo.State != model.StatePublic || tblInfo.TTLInfo == nil {
				continue
			}
			...
		}
	}
```

So, **what if we keep all the model.TableInfo with special attributes in-memory?** 
This could be a temporary solution, dirty and quick but easy to implement and verify.

In this commit, I add a new `ListTablesWithSpecialAttribute()` api to replace `SchemaTableInfos`.
All the data are in memory, so it is as fast as or even faster than infoschema v1.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test

The same data, memory usage after vs before:

image

We can see some jitters even with this change, it is caused by the statistics calling in this loop:

https://github.com/pingcap/tidb/blob/b9b330a9b6becafaa9be765e57726cb2661885a1/pkg/statistics/handle/bootstrap.go#L88-L109

And this commit only handle the memory in static state, when there are operations like `show xxx` or  `INFORMATION_SCHEMA.TABLES` visiting, there may still be `SchemaTables()` calling cause memory thrash. I'll handle them later.

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


D3Hunter:
> So, what if we keep all the model.TableInfo with special attributes in-memory?

can we only store those special attributes, not the whole table-info, as we have discussed in the coference?

tiancaiamao:
> > So, what if we keep all the model.TableInfo with special attributes in-memory?
> 
> can we only store those special attributes, not the whole table-info, as we have discussed in the coference?

The callee still need the table info. For example, you can check here
https://github.com/pingcap/tidb/blob/3ad726d34cdf25e556c049b699f0bddb894a8e84/pkg/ddl/ddl_tiflash_api.go#L274-L294


Or do you mean just the table ID rather than *model.TableInfo? the pointer size itself is just a machine word.

@ywqzzy

wjhuang2016:
> > > So, what if we keep all the model.TableInfo with special attributes in-memory?
> > 
> > 
> > can we only store those special attributes, not the whole table-info, as we have discussed in the coference?
> 
> The callee still need the table info. For example, you can check here
> 
> https://github.com/pingcap/tidb/blob/3ad726d34cdf25e556c049b699f0bddb894a8e84/pkg/ddl/ddl_tiflash_api.go#L274-L294
> 
> Or do you mean just the table ID rather than *model.TableInfo? the pointer size itself is just a machine word.
> 
> @ywqzzy

Store these infos may consume too much memory.

tiancaiamao:
> > > > So, what if we keep all the model.TableInfo with special attributes in-memory?
> > > 
> > > 
> > > can we only store those special attributes, not the whole table-info, as we have discussed in the coference?
> > 
> > 
> > The callee still need the table info. For example, you can check here
> > https://github.com/pingcap/tidb/blob/3ad726d34cdf25e556c049b699f0bddb894a8e84/pkg/ddl/ddl_tiflash_api.go#L274-L294
> > 
> > Or do you mean just the table ID rather than *model.TableInfo? the pointer size itself is just a machine word.
> > @ywqzzy
> 
> Store these infos may consume too much memory.

If we don't store them, the we go for the Data Dictionary solution. @wjhuang2016

pkg/infoschema/infoschema.go
```diff
@@ -320,6 +320,29 @@ func (is *infoSchema) SchemaTableInfos(schema model.CIStr) []*model.TableInfo {
 	return getTableInfoList(is.SchemaTables(schema))
 }
 
+type tableInfoResult struct {
+	DBName    string
+	TableInfo []*model.TableInfo
+}
+
+func (is *infoSchema) ListTablesWithSpecialAttribute(filter specialAttributeFilter) <-chan tableInfoResult {
```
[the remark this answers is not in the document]
tiancaiamao: Not now. If we need a ctx, then `is.SchemaTableInfos` etc all need it, and a lot other codes.

wjhuang2016 (approved):

pkg/infoschema/infoschema.go
```diff
@@ -320,6 +320,29 @@ func (is *infoSchema) SchemaTableInfos(schema model.CIStr) []*model.TableInfo {
 	return getTableInfoList(is.SchemaTables(schema))
 }
 
+type tableInfoResult struct {
+	DBName    string
+	TableInfo []*model.TableInfo
```
D3Hunter, line 325:
```suggestion
	TableInfos []*model.TableInfo
```

pkg/infoschema/infoschema_v2.go
```diff
@@ -171,11 +188,15 @@ func (isd *Data) add(item tableItem, tbl table.Table) {
...
 	isd.tableCache.Set(tableCacheKey{item.tableID, item.schemaVersion}, tbl)
-	if pi := tbl.Meta().GetPartitionInfo(); pi != nil {
+	ti := tbl.Meta()
+	if pi := ti.GetPartitionInfo(); pi != nil {
 		for _, def := range pi.Definitions {
 			isd.pid2tid.Set(partitionItem{def.ID, item.schemaVersion, tbl.Meta().ID, false})
 		}
 	}
+	if hasSpecialAttributes(ti) {
+		isd.tableInfoResident.Set(tableInfoItem{item.dbName, item.tableID, item.schemaVersion, ti, false})
```
D3Hunter, line 198: prefer we use named style like below, so many fields
```
tableID: xxx,
```

pkg/infoschema/infoschema_v2.go
```diff
@@ -191,6 +212,7 @@ func (isd *Data) remove(item tableItem) {
 	item.tomb = true
 	isd.byID.Set(item)
 	isd.byName.Set(item)
+	isd.tableInfoResident.Set(tableInfoItem{item.dbName, item.tableID, item.schemaVersion, nil, true})
```
D3Hunter, line 215: doitto

pkg/infoschema/infoschema.go
D3Hunter, line 328:
```suggestion
func (is *infoSchema) ListTablesWithSpecialAttribute(dbName string, filter specialAttributeFilter) <-chan tableInfoResult {
```
can we change it to get all tables with matched attributes for a db? so we don't to have a routine to do it. it's more complex now

tiancaiamao: This makes the streaming API possible. For example, if we don't store the special attribute and need to fetch data from tikv, then return all data []XXX may use much more memory then this API setting. @D3Hunter

D3Hunter (approved):

D3Hunter:
/hold

please fix the callsite of the api, you can unhold after change

PR #53337, opened 2024-05-16
context: Add method to detach the `tableReaderExecutor` from the session context

### LINKED ISSUE #53336: Add a method to detach the context used by `TableReaderExecutor`
## Enhancement

We have prepared enough utilities to detach the `TableReacerExecutor`'s context from the session context now! Let's finish the last step before starting to develop a new cursor fetch!

It's not that convenient to copy these contexts again (especially for `WarnHandler`, which is shared in almost every context). To solve this complicated construction issue, a DI framework is usually used. However, I think it over complicates the logic. A better solution (IMO) is to just leave these fields. Re-think that these fields cannot be used across multiple statements because two reasons:

1. They are shared and can be modified by the next statement, so it's not safe to keep using it.
2. The memory is cleared before executing the next statement, so keeping a reference to it is not safe.

For the first situation, we have to copy the internal value. For the second situation, we can just leave it! And allocate/create new one for each statement, so the previous one doesn't need to be cleared (and the GC will help us to free these memory later). Using this strategy, it'll be much simpler to refactor.

YangKeao:

Issue Number: close #53336

Problem Summary:

It's not safe to call `Next` (or other method) of an executor after the session is executing the next statement. We have split out a lot of `Context` and it's clear that the context depended by the `TableReaderExecutor` is safe to be re-used (after proper configuration) while executing another statement is running on the session.

This PR adds `IntoStatic` method for several contexts. After calling this method, the context can be safe to be used while the session continues to run the next statements.

### Check List

Tests 

TODO:

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/backfilling_scheduler.go
+++ b/pkg/ddl/backfilling_scheduler.go
@@ -180,7 +180,7 @@ func newDefaultReorgDistSQLCtx(kvClient kv.Client) *distsqlctx.DistSQLContext {
 		Client:                               kvClient,
 		EnableChunkRPC:                       true,
 		EnabledRateLimitAction:               variable.DefTiDBEnableRateLimitAction,
-		KVVars:                               tikvstore.NewVariables(&sqlKiller.Signal),
+		KVVars:                               *tikvstore.NewVariables(&sqlKiller.Signal),
 		SessionMemTracker:                    memory.NewTracker(memory.LabelForSession, -1),
 		Location:                             time.UTC,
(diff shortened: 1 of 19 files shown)


YangKeao:
Please use https://github.com/pingcap/tidb/pull/54456 instead. It's much better.

PR #53370, opened 2024-05-17
Optimizer: Fix null filtering logic for IN list

### LINKED ISSUE #49476: Unexpected Results of IN expression With NATURAL RIGHT JOIN

```sql
CREATE TABLE t0(c0 INT);
CREATE TABLE t1(c0 BOOLEAN);
INSERT INTO t0 (c0) VALUES (0);

SELECT * FROM t1 NATURAL RIGHT JOIN t0; -- 0
SELECT (false IN (t1.c0, t0.c0)) FROM t1 NATURAL RIGHT JOIN t0; -- 1
SELECT * FROM t1 NATURAL RIGHT JOIN t0 WHERE (false IN (t1.c0, t0.c0));
-- Expected: 0
-- Actual: Empty set
```

The third `SELECT` returns an empty set, which is surprising: If the result of second query is 1 (`TRUE`), the value of the `IN` expression should be `1`, and thus the third query should return `0`.

Not sure if the issue is related to https://github.com/pingcap/tidb/issues/49108

```
Release Version: v7.6.0-alpha
Edition: Community
Git Commit Hash: 918df0ae50568ebc54bd6d89f2d9a49bc9b79b06
Git Branch: heads/refs/tags/v7.6.0-alpha
UTC Build Time: 2023-12-14 04:15:47
GoVersion: go1.21.5
Race Enabled: false
Store: unistore |
```

ghazalfamilyusa:

Issue Number: close #49476

### Problem Summary:

Incorrect outer to inner join conversion for some cases of IN list incorrectly treated as null filtering.
An example in https://github.com/pingcap/tidb/issues/49476 

CREATE TABLE t0(c0 INT);
CREATE TABLE t1(c0 BOOLEAN);
INSERT INTO t0 (c0) VALUES (0);

SELECT * FROM t1 NATURAL RIGHT JOIN t0; -- 0
SELECT (false IN (t1.c0, t0.c0)) FROM t1 NATURAL RIGHT JOIN t0; -- 1
SELECT * FROM t1 NATURAL RIGHT JOIN t0 WHERE (false IN (t1.c0, t0.c0));
-- Expected: 0
-- Actual: Empty set

The problem is in the constant folding logic when it is used to evaluate null rejection logic.
The logic is risky and just fills in constant 1 for non-inner side. In the example above the IN list becomes 
false in (null,1) which is false (null filter is true). 

Fix done at the top level code for null filtering logic used in outer to inner. The fix use the OR logic instead of IN list.
For example, alse IN (t1.c0, t0.c0)) is checked for nullability using (false = t1.c0) or (false = t0.c0) which is not null filter. 
Also, this PR consolidates the null rejection code from outer to inner code and the general utility code. This includes cleans up of previous special cases.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


winoros (approved):

winoros:
I got the idea!

qw4990 (approved):

pkg/planner/util/null_misc.go
```diff
@@ -18,76 +18,98 @@ import (
...
 	return false
 }
 
-// isNullRejectedSpecially handles some null-rejected cases specially, since the current in
-// EvaluateExprWithNull is too strict for some cases, e.g. #49616.
-func isNullRejectedSpecially(ctx context.PlanContext, schema *expression.Schema, expr expression.Expression) bool {
-	return specialNullRejectedCase1(ctx, schema, expr) // only 1 case now
+// isNullRejectedInList checks null filter for IN list using OR logic.
+// Reason is that null filtering through evaluation by isNullRejectedSimpleExpr
+// has problems with IN list.
```
qw4990, line 45: Better to add a case here to explain why we have to treat IN-LIST as OR here. The rest LGTM!

qw4990: So if the in-list contains a column from the non-inner side, then it can never pass the NF check.

pkg/planner/util/null_misc.go
```diff
@@ -18,76 +18,98 @@ import (
...
+	default:
+		return isNullRejectedSimpleExpr(ctx, innerSchema, predicate)
 	}
-	orFunc := isFunc(expr, ast.LogicOr)
-	if orFunc == nil {
+}
+
+// isNullRejectedSimpleExpr check whether a condition is null-rejected
+// A condition would be null-rejected in one of following cases:
+// If it is a predicate containing a reference to an inner table that evaluates to UNKNOWN or FALSE
```
AilinKid, line 97: i think the comment should be refined as well, "inner table" has no ideas with

AilinKid (commented):
Rest LGTM

AilinKid (approved):

pkg/planner/util/null_misc.go
ghazalfamilyusa: The explanation for this case is shown above and in the issue. I guess you guys think it is better to add it in the code. Added that to the comment.

pkg/planner/util/null_misc.go
ghazalfamilyusa: added explanation to inner table as "null producing side"

PR #53444, opened 2024-05-21
config: must set line terminator when use strict-format

### LINKED ISSUE #37338: [Lightning]: Panic on invalid Mertic countter operation
## Bug Report

Version: 6.1.0

Lightning panic on https://github.com/pingcap/tidb/blob/master/br/pkg/lightning/restore/restore.go#L2333

I think it's because float64(currOffset - startOffset) < 0.

```log
panic: counter cannot decrease in value

goroutine 202374 [running]:
github.com/prometheus/client_golang/prometheus.(*counter).Add(0x0?, 0xc069674c80?)
/go/pkg/mod/github.com/prometheus/client_golang@v1.11.0/prometheus/counter.go:109 +0xf4
github.com/pingcap/tidb/br/pkg/lightning/restore.(*chunkRestore).deliverLoop(0xc0618805a0,
{0x445dbd8, 0xc0017c6040}

, 0xc06c80a3c0, 0xc0012af740, 0x3, 0xc002a8ff18?, 0xc002a8ffd0?, 0xc0016e5080)
/mnt/tidb/sql/br/pkg/lightning/restore/restore.go:2255 +0x7a3
github.com/pingcap/tidb/br/pkg/lightning/restore.(*chunkRestore).restore.func1()
/mnt/tidb/sql/br/pkg/lightning/restore/restore.go:2509 +0xab
created by github.com/pingcap/tidb/br/pkg/lightning/restore.(*chunkRestore).restore
/mnt/tidb/sql/br/pkg/lightning/restore/restore.go:2507 +0x31b
```

lance6716:

Issue Number: close #37338

Problem Summary:

as title

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

manually check it from a oncall ticket

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/lightning/mydump/region_test.go
+++ b/pkg/lightning/mydump/region_test.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package mydump
+package mydump_test
 
 import (
 	"context"
@@ -22,6 +22,7 @@ import (
 
 	"github.com/pingcap/tidb/br/pkg/storage"
 	"github.com/pingcap/tidb/pkg/lightning/config"
+	. "github.com/pingcap/tidb/pkg/lightning/mydump"
 	"github.com/pingcap/tidb/pkg/lightning/worker"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -37,18 +38,6 @@ import (
 /*
 TODO : test with specified 'regionBlockSize' ...
 */
-
-func newConfigWithSourceDir(sourceDir string) *config.Config {
-	path, _ := filepath.Abs(sourceDir)
-	return &config.Config{
-		Mydumper: config.MydumperRuntime{
-			SourceDir:        "file://" + filepath.ToSlash(path),
-			Filter:           []string{"*.*"},
-			DefaultFileRules: true,
-		},
-	}
-}
-
 func TestTableRegion(t *testing.T) {
 	cfg := newConfigWithSourceDir("./examples")
 	loader, _ := NewLoader(context.Background(), NewLoaderCfg(cfg))
@@ -499,7 +488,7 @@ func TestSplitLargeFileOnlyOneChunk(t *testing.T) {
 	}
 }
 
-func TestDetectTerminator(t *testing.T) {
+func TestSplitLargeFileSeekInsideCRLF(t *testing.T) {
 	ctx := context.Background()
 	meta := &MDTableMeta{
 		DB:   "csv",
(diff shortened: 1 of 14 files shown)


pkg/lightning/mydump/region_test.go
```diff
@@ -487,3 +487,96 @@ func TestSplitLargeFileOnlyOneChunk(t *testing.T) {
...
+				Separator: ",",
+			},
+			StrictFormat:  true,
+			Filter:        []string{"*.*"},
+			MaxRegionSize: 2,
+		},
+	}
+	divideConfig := NewDataDivideConfig(cfg, 1, ioWorker, store, meta)
+
+	// in fact this is the wrong result, just to show the bug. pos mismatch with offsets
```
D3Hunter, line 531:
```suggestion
	// in fact this is the wrong result, just to show the bug. pos mismatch with offsets. and we might read more rows than expected
```

D3Hunter (approved):
or we can fix it by let parser read the longest terminator when terminator is not given explicitly, import into may have similar issue too

lance6716:
> we can fix it by let parser read the longest terminator when terminator is not given explicitly

this fix has easier logic but require a user behaviour change, I'll ask PM's approval about this fix first. Change the parser behaviour also need to think about what's the position of last field in a line and line terminator in multiple cases, use `==` or `>=`, different lines may end with different terminator, etc. More complex.

> import into may have similar issue too

I also add the check for IMPORT INTO

Frank945946:
@lance6716  Currently, \r, \n, and \r\n can all be used as line delimiters. However, when the user selects the Strict format, an additional check can be performed. This involves reading the first line of the source file to determine the line delimiter used in the source file, such as \r\n. If \r\n is identified as the line delimiter, it will be used to split one large CSV file into multiple smaller CSV files based on the specified size.

lance6716:
> @lance6716 Currently, \r, \n, and \r\n can all be used as line delimiters. However, when the user selects the Strict format, an additional check can be performed. This involves reading the first line of the source file to determine the line delimiter used in the source file, such as \r\n. If \r\n is identified as the line delimiter, it will be used to split one large CSV file into multiple smaller CSV files based on the specified size.

There are many difficulties to implement the additional check. My arguments are

- considering the definition of "strict"-format, it's non-sense to tolerate `\r`, `\n`, `\r\n` all as terminator in one task or even in one file. The old behaviour in fact supports all three terminators in one file, and out detect logic will cause wrong parsing result.
- There are many execution path for this feature, like IMPORT INTO with/without DXF, IMPORT INTO running while upgrading, lightning resuming from checkpoint. Splitting the strict format files and parsing the file chunks are two separate procedures, and `strict-format` is not a sub-item of `csv` in configuration or in code. It's complecated and adding a restriction to user is less error-prone.

tangenta (approved):

PR #53494, opened 2024-05-22
*: implement auth plugin support in the extension framework

### LINKED ISSUE #53181: Support user-implemented authentication plugin within the extension framework

yzhan1:

Implements the extension authentication plugin support. Refer to https://github.com/pingcap/tidb/pull/53182 for design.

Issue Number: close #53181

Problem Summary:

TiDB developers can now implement new authentication plugin using the extension framework to keep proprietary code out of the core TiDB folders.

A new struct `extension.AuthPlugin` is created, where developers can implement their own authentication/authorization logic on top of TiDB. It achieves similar things to MySQL auth plugin by allowing developers implementing customized plugins.

Functions defined in `extension.AuthPlugin` will be executed during
1. `privilege.ConnectionVerification` for authentication check during logins
2. `privilege.RequestVerification` and `privilege.DynamicRequestVerification` for extra authorization checks on top of existing MySQL privilege checks

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Allow developers to implement their own authentication plugin using the extension framework.
```


yzhan1:
@lcwangchao @CbcWestwolf Would like to gather some early feedback from you. Thanks in advance!

yzhan1 (commented):
I saw that in `privilege.go` there are two functions called for verifying privileges of a non-current user:
1. `RequestDynamicVerificationWithUser`
2. `RequestVerificationWithUser`

Where 1 is used for checking system/admin privileges such as `SUPER_USER`, `RESTRICTED_USER_ADMIN` when performing `ALTER USER`/`SET PASSWORD` for super/admin users. 2 is used only for verifying view definer's privilege.

I think it might be too restricted to impose the same restriction on 1, that is - to not allow plugin users to be a super/admin user. Maybe we should allow this, and just call out explicitly that the verify privilege functions in the plugin will only be executed for the current user. And it's up to the plugin owner/operator to revoke unnecessary MySQL privileges.

What do you think? @lcwangchao

lcwangchao:
> I saw that in `privilege.go` there are two functions called for verifying privileges of a non-current user:
> 
> 1. `RequestDynamicVerificationWithUser`
> 2. `RequestVerificationWithUser`
> 
> Where 1 is used for checking system/admin privileges such as `SUPER_USER`, `RESTRICTED_USER_ADMIN` when performing `ALTER USER`/`SET PASSWORD` for super/admin users. 2 is used only for verifying view definer's privilege.
> 
> I think it might be too restricted to impose the same restriction on 1, that is - to not allow plugin users to be a super/admin user. Maybe we should allow this, and just call out explicitly that the verify privilege functions in the plugin will only be executed for the current user. And it's up to the plugin owner/operator to revoke unnecessary MySQL privileges.
> 
> What do you think? @lcwangchao



> I saw that in `privilege.go` there are two functions called for verifying privileges of a non-current user:
> 
> 1. `RequestDynamicVerificationWithUser`
> 2. `RequestVerificationWithUser`
> 
> Where 1 is used for checking system/admin privileges such as `SUPER_USER`, `RESTRICTED_USER_ADMIN` when performing `ALTER USER`/`SET PASSWORD` for super/admin users. 2 is used only for verifying view definer's privilege.
> 
> I think it might be too restricted to impose the same restriction on 1, that is - to not allow plugin users to be a super/admin user. Maybe we should allow this, and just call out explicitly that the verify privilege functions in the plugin will only be executed for the current user. And it's up to the plugin owner/operator to revoke unnecessary MySQL privileges.
> 
> What do you think? @lcwangchao

Do you mean if a user is created with a plugin, some privilege can only be revoked by himself? I think it is also some what restricted. I

A plugin user may have two sets of privileges:

- Privileges stored in database. 
- Privileges provided by the plugin with the session context.
 
The real privileges of a session should be an intersection of those two sets. But in the `ALTER/DROP USER` statements, we can only check the first set for the target user without considering the session privilege. I think it is safer because that means "a user should not change a plugin user's privilege without admin privilege if the plugin user have a admin privilege", even if the plugin user may lost its admin privilege temporarily in some session.

pkg/ddl/ddl_api.go
```diff
@@ -3252,13 +3252,17 @@ func (d *ddl) CreateView(ctx sessionctx.Context, s *ast.CreateViewStmt) (err err
...
 	// Always Use `format.RestoreNameBackQuotes` to restore `SELECT` statement despite the `ANSI_QUOTES` SQL Mode is enabled or not.
 	restoreFlag := format.RestoreStringSingleQuotes | format.RestoreKeyWordUppercase | format.RestoreNameBackQuotes
 	var sb strings.Builder
 	if err := s.Select.Restore(format.NewRestoreCtx(restoreFlag, &sb)); err != nil {
 		return nil, err
 	}
+	// Extension based auth plugin users cannot be the definer of the view.
+	if _, ok := sctx.GetAuthPlugins()[s.Definer.AuthPlugin]; ok && s.Security == model.SecurityDefiner {
+		return nil, dbterror.ErrWrongUsage.GenWithStackByArgs("Definer", s.Definer.AuthPlugin)
+	}
```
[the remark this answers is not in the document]
lcwangchao: Outside the DDL, we also need to check the privilege when selecting a view to forbid this case:

1. create a normal user `user1`
2. create a view with the definer `user1`
3. drop `user1`
4. create a plugin user `user1`   

So, anyway, we need to check whether the user is a plugin user in `RequestVerificationWithUser `. We can just remove this line to allow plugin user to create a view,and it should always report an error when selecting it.

pkg/sessionctx/context.go
```diff
@@ -124,6 +124,9 @@ type Context interface {
 
 	GetSessionManager() util.SessionManager
 
+	// GetAuthPlugins returns all the registered authentication plugins
+	GetAuthPlugins() map[string]*extension.AuthPlugin
```
lcwangchao, line 128: You can move `GetAuthPlugins` to `extension.SessionExtensions` which can be get from `GetExtensions` method of `sessionctx.Context`.

Another question is why we need a `GetAuthPlugins` in session without just call `extension.GetExtensions().GetAuthPlugins()` ?

pkg/parser/ast/misc.go
```diff
@@ -1419,14 +1419,29 @@ func (n *UserSpec) SecurityString() string {
...
 		return "", true
 	}
 
 	opt := n.AuthOpt
+	// If the extension auth plugin is available, use it to encode the password.
+	if pluginEncoder != nil {
+		if opt.ByAuthString {
+			return pluginEncoder(opt.AuthString)
+		}
+		return pluginEncoder(opt.HashString)
```
lcwangchao, line 1443: Is it necessary to encode hash string because it is already hashed.

pkg/privilege/privileges/privileges.go
```diff
@@ -207,14 +212,77 @@ func (p *UserPrivileges) RequestVerificationWithUser(db, table, column string, p
...
+	// Now we know that user has the privilege in MySQL, we need to check the extension auth plugin.
+	if authPlugin, ok := p.authPlugins[p.userRecord.AuthPlugin]; ok {
+		return pluginPrivilegeChecker(authPlugin)
+	}
+	// This normally should not happen since users can only get to here if they have the privilege in MySQL and are using a registered auth plugin.
+	return mysqlAllowed
+}
+
+func (p *UserPrivileges) authenticateWithPlugin(user *auth.UserIdentity, authUser string, authentication, salt []byte, sessionVars *variable.SessionVars, authConn conn.AuthConn, authPlugin *extension.AuthPlugin, pwd string, record *UserRecord, hasPassword string) error {
+	if err := authPlugin.AuthenticateUser(authUser, pwd, authentication, salt, sessionVars.TLSConnectionState, authConn); err != nil {
```
lcwangchao, line 267: I'm not sure we should both pass `user/host` and `authUser/authHost` to method `AuthenticateUser`. They are not all the same.

pkg/privilege/privileges/privileges.go
```diff
@@ -77,13 +77,17 @@ type UserPrivileges struct {
 	host string
 	*Handle
 	extensionAccessCheckFuncs []extension.AccessCheckFunc
+	authPlugins               map[string]*extension.AuthPlugin
+	userRecord                *UserRecord
+	sessionVars               *variable.SessionVars
```
lcwangchao, line 82: These fields are too complex and it's hard to figure out their usage from the field names. We only need to store some plugin-related auth context here, how about just add two functions like `pluginRequestVerification` and `pluginRequestDynamicVerfication` here ?

pkg/sessionctx/context.go
yzhan1: `extension.GetExtensions()` requires locking which could add perf overhead when executing it for every request. Executing it once at connection establishment time is fine.

I will try to move `GetAuthPlugins` into `extension.SessionExtensions`

pkg/privilege/privileges/privileges.go
yzhan1: Removed the `authUser` parameter. That one seems to be only used for LDAP based authentications.

yzhan1:
> > I saw that in `privilege.go` there are two functions called for verifying privileges of a non-current user:
> > 
> > 1. `RequestDynamicVerificationWithUser`
> > 2. `RequestVerificationWithUser`
> > 
> > Where 1 is used for checking system/admin privileges such as `SUPER_USER`, `RESTRICTED_USER_ADMIN` when performing `ALTER USER`/`SET PASSWORD` for super/admin users. 2 is used only for verifying view definer's privilege.
> > I think it might be too restricted to impose the same restriction on 1, that is - to not allow plugin users to be a super/admin user. Maybe we should allow this, and just call out explicitly that the verify privilege functions in the plugin will only be executed for the current user. And it's up to the plugin owner/operator to revoke unnecessary MySQL privileges.
> > What do you think? @lcwangchao
> 
> Do you mean if a user is created with a plugin, some privilege can only be revoked by himself? I think it is also some what restricted. I
> 
> A plugin user may have two sets of privileges:
> 
> * Privileges stored in database.
> * Privileges provided by the plugin with the session context.
> 
> The real privileges of a session should be an intersection of those two sets. But in the `ALTER/DROP USER` statements, we can only check the first set for the target user without considering the session privilege. I think it is safer because that means "a user should not change a plugin user's privilege without admin privilege if the plugin user have a admin privilege", even if the plugin user may lost its admin privilege temporarily in some session.

@lcwangchao Please let me clarify. I meant that for revoking/checking those special privileges for a plugin user, TiDB should only check the priv set 1 (stored in TiDB) since it lacks information about priv set 2 at the execution time, unless the user being verified is the currently connected user itself.

For example: `u1` is a normal user and `u2` is a plugin user. `u2` has select priv on table `t1`. There's a view on `t1` where definer is `u2`. When `u1` wants to run a select on the view, TiDB checks the select priv on `t1` for `u2`, where it will only be able to check `u2`'s priv based on priv set 1. If the DB owner wants to revoke `u2`'s select priv to reject other users from selecting the view, the owner should revoke it for both set 1 and 2 (as opposed to only set 2 for normal query scenarios where we only need to validate priv for current user).

> "a user should not change a plugin user's privilege without admin privilege if the plugin user have a admin privilege"

I think by implementing what you suggested, we are not only doing this, but also not allowing an admin user to change a plugin user's privilege too. That's because inside `RequestDynamicVerificationWithUser` or `RequestVerificationWithUser`, we don't have enough information to check for privilege set 2, and the only thing we can do there seems to be rejecting the priv check if the user being checked is a plugin user.

lcwangchao:
Maybe we can talk about some scenes. Suppose we have below user1:

- `u1`: A plugin user, has `SELECT` privilege for table t1. It also has a privilege of creating a view.
- `u2`: A plugin user, has SUPER privilege.
- `u3`: A normal user with only `CREATE USER` privilege.
- `u4`: A normal user with `CREATE USER` and `SYSTEM_USER` privilege.

And we have some sessions:

- `s1`: logged in with u1, but not permitted to read table t1 in this session.
- `s2`: logged in with u1. All privileges in u1 are permitted.

Currently, my though is:

If `s1` wants to create a view `create view v1 as select * from t1`, it should be denied because it does not a `SELECT` privilege for table t1. However, if `s2` wants to create this view, it should be permitted. After `v1` is created, anyone who have the privilege to select `v1` can select successfully even if `s1` because view only check the stored privilege of `u1` now.

For dropping table:

- `u1` can be dropped by `u2`, `u3`, `u4`.
- `u2` can only be dropped by `u4` because it has the admin privilege. Here we do not need to check `u2`'s privilege in session because it is not valid and not necessary.

yzhan1:
@lcwangchao Thanks for the review!

I have addressed most of the comments. Will add more test cases to cover the scenario you described.

pkg/server/conn.go
```diff
@@ -253,6 +253,13 @@ func (cc *clientConn) authSwitchRequest(ctx context.Context, plugin string) ([]b
 		clientPlugin += "_client"
 	} else if plugin == mysql.AuthLDAPSimple {
 		clientPlugin = mysql.AuthMySQLClearPassword
+	} else if authPluginImpl, ok := cc.extensions.GetAuthPlugins()[plugin]; ok {
+		if authPluginImpl.RequiredClientSidePlugin != "" {
+			clientPlugin = authPluginImpl.RequiredClientSidePlugin
+		} else {
+			// If the plugin does not need special handling of the password, use clear text password.
+			clientPlugin = mysql.AuthMySQLClearPassword
```
lcwangchao, line 261: Seems this behavior is different with the comment of `RequiredClientSidePlugin`:

> If this is empty then `AuthPlugin.Name` is used as the required client-side plugin.

pkg/privilege/privileges/privileges.go
```diff
@@ -213,11 +218,86 @@ func (p *UserPrivileges) RequestVerificationWithUser(db, table, column string, p
...
+		return mysqlAllowed
+	}
+	// If MySQL privilege check failed, skip the extension auth plugin check.
+	if !mysqlAllowed {
+		return false
+	}
+	return p.authPluginRequestVerification(mysqlPriv, user, host, activeRoles, db, table, column, priv)
+}
+
+func (p *UserPrivileges) authenticateWithPlugin(user *auth.UserIdentity, authentication, salt []byte, sessionVars *variable.SessionVars, authConn conn.AuthConn, authPlugin *extension.AuthPlugin, pwd string, record *UserRecord, hasPassword string) error {
```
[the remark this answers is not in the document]
lcwangchao: Same with `authPlugin` and `record` and record is only used for logging plugin name and I think we can remove it too.

pkg/privilege/privileges/privileges.go
```diff
@@ -78,13 +78,18 @@ type UserPrivileges struct {
 	host string
 	*Handle
 	extensionAccessCheckFuncs []extension.AccessCheckFunc
+	authPlugins               map[string]*extension.AuthPlugin
```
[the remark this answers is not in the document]
yzhan1: Functions like `IsValidHash` and `GetAuthPluginForConnection` will need this reference to load the plugins. Would prefer we keep this here since we load the extensions when creating `UserPrivileges` anyways.

yzhan1:
@lcwangchao Could you please take a look at the latest diffs? Thank you!

pkg/executor/simple.go
```diff
@@ -1164,16 +1164,20 @@ func (e *SimpleExec) executeCreateUser(ctx context.Context, s *ast.CreateUserStm
 				return err
 			}
 		}
-		pwd, ok := spec.EncodedPassword()
-
-		if !ok {
-			return errors.Trace(exeerrors.ErrPasswordFormat)
-		}
+		pluginImpl, foundPlugin := e.Ctx().GetExtensions().GetAuthPlugin(authPlugin)
```
lcwangchao, line 1167: It's better to get the plugin in `default` branch and just put a var here:

```
var extPlugin extension. AuthPlugin
```

pkg/executor/simple.go
```diff
@@ -1787,6 +1795,7 @@ func (e *SimpleExec) executeAlterUser(ctx context.Context, s *ast.AlterUserStmt)
 			if spec.AuthOpt.AuthPlugin == "" {
 				spec.AuthOpt.AuthPlugin = currentAuthPlugin
 			}
+			authPluginImpl, foundAuthPlugin := e.Ctx().GetExtensions().GetAuthPlugin(spec.AuthOpt.AuthPlugin)
```
lcwangchao, line 1798: Ditto

pkg/extension/manifest.go
```diff
@@ -65,6 +65,13 @@ func WithCustomAccessCheck(fn AccessCheckFunc) Option {
 	}
 }
 
+// WithCustomAuthPlugins specifies the custom authentication plugins available for the system.
+func WithCustomAuthPlugins(authPlugins map[string]*AuthPlugin) Option {
```
lcwangchao, line 69: Could we make the argument to `plugins []*AuthPlugin` to make it simple ? It's doesn't matter if the input plugins have dup names because we'll check it later

pkg/privilege/privileges/privileges.go
```diff
@@ -591,6 +657,17 @@ func (p *UserPrivileges) ConnectionVerification(user *auth.UserIdentity, authUse
 				zap.String("authentication_string", pwd))
 			return info, ErrAccessDenied.FastGenByArgs(user.Username, user.Hostname, hasPassword)
 		}
+	} else if record.AuthPlugin == mysql.AuthSocket {
```
lcwangchao, line 660: You should remove this branch, seems to be caused by resolving conflict.

lcwangchao:
@bb7133 could you also take a look?

pkg/privilege/privileges/privileges.go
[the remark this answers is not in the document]
lcwangchao: Another option is to move it to `Handle` which is included in line 79. Seems it is shared by all connections and is used to cache some global meta.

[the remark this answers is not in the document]
yzhan1: It might be better to leave it as is to be consistent with `extensionAccessCheckFuncs`. Any concerns on keeping this inside `UserPrivileges`?

lcwangchao (approved):

yzhan1:
@bb7133 Could you please take a look at your convenience? Thank you!

@lcwangchao Could we mark this PR as `ok-to-test` so I can start fixing test issues (if any)? I recall some changes are needed for the bazel files for modified/new tests

bb7133 (approved):
LGTM

PR #53534, opened 2024-05-24
expression,*: remove the session context inside `ParamMarker`

### LINKED ISSUE #53533: Add `EvalContext` argument for `Expression.GetType` and `ParamMarker.GetUserVar`
## Enhancement

Ref https://github.com/pingcap/tidb/issues/53485#issuecomment-2128520761. As the actual type and internal datum depend on the context, we should explicitly pass the context through the argument, rather than store it in the `ParamMarker`. It'll be a big change.

It'll introduce three modification:

1. Remove the `ctx` in `ParamMarker`.
2. Add `EvalContext` argument for `Expression.GetType`.
3. Add `EvalContext` argument for `String` (replace `String` with `StringWithCtx`)

The most concerning issue is that the `.String()` method cannot accept a parameter. We should analyze the usage of it carefully. Sometimes, returning a simple `?` is enough, but it's not always good to do so. If not, we can `Eval` it, or provide another function to get the debug output with context :thinking:  .

YangKeao:

Issue Number: close #53533

Problem Summary:

Previously, the code stores a reference to the session context in `ParamMarker`. It convenient for re-using the expression in the plan cache inside a single session. However, it blocks us from achieving the following goals:

1. We cannot evaluate the expression whiling the params on session is changing. It will happen if we enabled lazy-cursor fetch.
2. The plan cache cannot be re-used across different sessions, because the expression will always reference to the session on which it's cached.

This PR depends on #53922 now.

1. Remove the session context inside `ParamMarker`, and pass it through argument.
2. Add `EvalContext` argument for `GetType` and `GetUserVar`.
4. Add a new method `StringWithCtx` to stringify the expression with a given context.

It's also fine to use the `StringWithCtx` with `nil` argument directly, but it'll print `?` for the params.

The biggest challenge is that the `String` is implicitly used in many places: like formatter and error / warnings. I wrote a linter to detect all the cases, you can find it in `/build/linter/printexpression`. It ensures that all struct/interface (and the reference/slice of them) which implements `StringWithCtx` but doesn't implement `String` cannot be used as an argument of format function.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/importer/import.go
+++ b/pkg/executor/importer/import.go
@@ -809,9 +809,13 @@ func (p *Plan) initParameters(plan *plannercore.ImportInto) error {
 		setClause = sb.String()
 	}
 	optionMap := make(map[string]any, len(plan.Options))
+	var evalCtx expression.EvalContext
+	if plan.SCtx() != nil {
+		evalCtx = plan.SCtx().GetExprCtx().GetEvalCtx()
+	}
 	for _, opt := range plan.Options {
 		if opt.Value != nil {
-			val := opt.Value.String()
+			val := opt.Value.StringWithCtx(evalCtx)
 			if opt.Name == cloudStorageURIOption {
 				val = ast.RedactURL(val)
 			}
--- a/pkg/expression/column.go
+++ b/pkg/expression/column.go
@@ -22,6 +22,7 @@ import (
 	"unsafe"
 
 	"github.com/pingcap/errors"
+	exprctx "github.com/pingcap/tidb/pkg/expression/context"
 	"github.com/pingcap/tidb/pkg/parser/ast"
 	"github.com/pingcap/tidb/pkg/parser/charset"
 	"github.com/pingcap/tidb/pkg/parser/model"
@@ -389,7 +390,7 @@ func (col *Column) VecEvalJSON(ctx EvalContext, input *chunk.Chunk, result *chun
 const columnPrefix = "Column#"
 
 // StringWithCtx implements Expression interface.
-func (col *Column) StringWithCtx(ctx EvalContext) string {
+func (col *Column) StringWithCtx(ctx ParamValues) string {
 	return col.String()
 }
 
@@ -398,7 +399,7 @@ func (col *Column) String() string {
 	if col.IsHidden && col.VirtualExpr != nil {
 		// A hidden column without virtual expression indicates it's a stored type.
 		// a virtual column should be able to be stringified without context.
-		return col.VirtualExpr.StringWithCtx(nil)
+		return col.VirtualExpr.StringWithCtx(exprctx.EmptyParamValues)
 	}
 	if col.OrigName != "" {
 		return col.OrigName
--- a/pkg/expression/column.go
+++ b/pkg/expression/column.go
@@ -388,11 +388,17 @@ func (col *Column) VecEvalJSON(ctx EvalContext, input *chunk.Chunk, result *chun
 
 const columnPrefix = "Column#"
 
+// StringWithCtx implements Expression interface.
+func (col *Column) StringWithCtx(ctx EvalContext) string {
+	return col.String()
+}
+
 // String implements Stringer interface.
 func (col *Column) String() string {
 	if col.IsHidden && col.VirtualExpr != nil {
 		// A hidden column without virtual expression indicates it's a stored type.
-		return col.VirtualExpr.String()
+		// a virtual column should be able to be stringified without context.
+		return col.VirtualExpr.StringWithCtx(nil)
 	}
 	if col.OrigName != "" {
 		return col.OrigName
--- a/pkg/expression/constant.go
+++ b/pkg/expression/constant.go
@@ -19,7 +19,6 @@ import (
 	"unsafe"
 
 	"github.com/pingcap/tidb/pkg/parser/mysql"
-	"github.com/pingcap/tidb/pkg/sessionctx/variable"
 	"github.com/pingcap/tidb/pkg/types"
 	"github.com/pingcap/tidb/pkg/util/chunk"
 	"github.com/pingcap/tidb/pkg/util/codec"
@@ -130,7 +129,6 @@ type Constant struct {
 
 // ParamMarker indicates param provided by COM_STMT_EXECUTE.
 type ParamMarker struct {
-	ctx   variable.SessionVarsProvider
 	order int
 }
 
@@ -139,17 +137,11 @@ func (d *ParamMarker) GetUserVar(ctx EvalContext) types.Datum {
 	return ctx.GetParamValue(d.order)
 }
 
-func (d *ParamMarker) getUserVarWithInternalCtx() types.Datum {
-	// TODO: remove this function in the future
-	sessionVars := d.ctx.GetSessionVars()
-	return sessionVars.PlanCacheParams.GetParamValue(d.order)
-}
-
 // String implements fmt.Stringer interface.
 func (c *Constant) String() string {
 	if c.ParamMarker != nil {
-		dt := c.ParamMarker.getUserVarWithInternalCtx()
-		c.Value.SetValue(dt.GetValue(), c.RetType)
+		// TODO: use a more meaningful value to represent the param marker.
+		return "?"
 	} else if c.DeferredExpr != nil {
 		return c.DeferredExpr.String()
 	}
--- a/pkg/expression/constant.go
+++ b/pkg/expression/constant.go
@@ -19,7 +19,6 @@ import (
 	"unsafe"
 
 	"github.com/pingcap/tidb/pkg/parser/mysql"
-	"github.com/pingcap/tidb/pkg/sessionctx/variable"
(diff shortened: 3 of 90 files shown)


pkg/executor/importer/import.go
```diff
@@ -809,9 +809,13 @@ func (p *Plan) initParameters(plan *plannercore.ImportInto) error {
 		setClause = sb.String()
 	}
 	optionMap := make(map[string]any, len(plan.Options))
+	var evalCtx expression.EvalContext
+	if plan.SCtx() != nil {
+		evalCtx = plan.SCtx().GetExprCtx().GetEvalCtx()
```
lcwangchao, line 814: Any risk for nil exception ？

pkg/executor/sortexec/sort_spill_test.go
```diff
@@ -99,6 +99,8 @@ func (r *resultChecker) initRowPtrs() {
 }
 
 func (r *resultChecker) check(resultChunks []*chunk.Chunk, offset int64, count int64) bool {
+	ctx := mock.NewContext()
```
lcwangchao, line 102: you can also use `StaticEvalContext` here

pkg/expression/column.go
```diff
@@ -388,11 +388,17 @@ func (col *Column) VecEvalJSON(ctx EvalContext, input *chunk.Chunk, result *chun
...
+	return col.String()
+}
+
 // String implements Stringer interface.
 func (col *Column) String() string {
 	if col.IsHidden && col.VirtualExpr != nil {
 		// A hidden column without virtual expression indicates it's a stored type.
-		return col.VirtualExpr.String()
+		// a virtual column should be ablt to be stringified without context.
+		return col.VirtualExpr.StringWithCtx(nil)
```
lcwangchao, line 401: Could we forbid passing a `nil` as a context if possible? It requires all implements to handle the nil value and the coder may not be aware of it.

Could we introduce a standalone interface like:

```
type ParamValues interface {
    GetParamValue(idx int) types.Datum
}
```

And `EvalContext` can inherit it:

```
type  EvalContext interface {
    ParamValues
}
```

We can also have static implement of it:

```
var EmptyParamValues ParamValues
```

If we know one expression does not need any params, we can call it like this:

```
expr.StringWithCtx(EmptyParamValues)
```

pkg/executor/importer/import.go
YangKeao: Yes. The `ImportInto` plan (and `Analyze` plan) didn't initialize the internal base plan, so the `SCtx()` can be `nil`.

pkg/executor/sortexec/sort_spill_test.go
YangKeao: Good! I replaced most new `mock.NewContext()` in this PR into `contextstatic.NewStaticEvalContext()` :+1:.

pkg/expression/column.go
YangKeao: Good! Let me try :+1:

YangKeao: I realized a problem: the implementation of `StringWithCtx` indeed needs to know whether the argument is `nil` (or any empty value). For `Constant`, the expected behavior is that: if the context is empty, return "?"; if the context is not empty, return the corresponding parameter. Then I cannot get an appropriate implementation of `EmptyParamValues`.

I thought two possible implementations:
1. Return the `"?"` datum. It's not good, because the `"?"` value should be a part of the implementation of `Constant`, but not part of `EmptyParamValues`. Also, I don't want to confuse the empty result and a normal `"?"` string datum.
2. Return `nil` for `*types.Datum`. However,  I don't see the difference between "handling the case when `*types.Datum` is `nil`" and "handling the case when `EvalContext` is `nil`".

Therefore, currently, I prefer to keep the current implementation :thinking: .

lcwangchao: Could we add new method `ParamValueCount` to `ParamValues` to solve this ?

pkg/expression/constant.go
```diff
@@ -139,26 +137,25 @@ func (d *ParamMarker) GetUserVar(ctx EvalContext) types.Datum {
...
-
-// String implements fmt.Stringer interface.
-func (c *Constant) String() string {
+// StringWithCtx implements Expression interface.
+func (c *Constant) StringWithCtx(ctx EvalContext) string {
 	if c.ParamMarker != nil {
-		dt := c.ParamMarker.getUserVarWithInternalCtx()
+		intest.Assert(ctx != nil, "context is nil for ParamMarker")
+		if ctx == nil {
+			return "?"
```
qw4990, line 145: How about triggering a panic here? Then if someone forget to pass `ctx`, it can be exposed immediately.

YangKeao: As a result after discussion, we agreed that `panic` here directly will have big risk, so only return a `"?"` and have an assertion for test is good.

Later (not in this PR), I'll try to consider how to extend the tests of binary protocol, because most of the current tests didn't have `param`.

qw4990 (approved):

pkg/expression/constant.go
```diff
@@ -130,35 +132,32 @@ type Constant struct {
...
-}
-
-// String implements fmt.Stringer interface.
-func (c *Constant) String() string {
+// StringWithCtx implements Expression interface.
+func (c *Constant) StringWithCtx(ctx ParamValues) string {
 	if c.ParamMarker != nil {
-		dt := c.ParamMarker.getUserVarWithInternalCtx()
+		dt, err := c.ParamMarker.GetUserVar(ctx)
+		intest.Assert(err == nil, "fail to get param", err)
```
lcwangchao, line 147: The format string: "fail to get param" does not contain a placeholder for err. You can use `intest.AssertNoError`.

YangKeao: Done :+1: .

lance6716 (approved):

pkg/executor/importer/import.go
YangKeao: > I think we can rely on CI to find th problem

Yes. That's why I have added this line (or the unit test will fail).
The fact "The `ImportInto` plan (and `Analyze` plan) didn't initialize the internal base plan" is found by CI :facepalm: .

zimulala (approved):

pkg/expression/column.go
```diff
@@ -388,11 +389,17 @@ func (col *Column) VecEvalJSON(ctx EvalContext, input *chunk.Chunk, result *chun
...
+func (col *Column) StringWithCtx(ctx ParamValues) string {
+	return col.String()
+}
+
 // String implements Stringer interface.
 func (col *Column) String() string {
 	if col.IsHidden && col.VirtualExpr != nil {
 		// A hidden column without virtual expression indicates it's a stored type.
-		return col.VirtualExpr.String()
+		// a virtual column should be ablt to be stringified without context.
```
XuHuaiyu, line 401:
```suggestion
		// a virtual column should be able to be stringified without context.
```

YangKeao: Fixed.

pkg/expression/scalar_function.go
```diff
@@ -141,9 +141,27 @@ func (sf *ScalarFunction) String() string {
 	return buffer.String()
 }
 
-// MarshalJSON implements json.Marshaler interface.
-func (sf *ScalarFunction) MarshalJSON() ([]byte, error) {
```
XuHuaiyu, line 145: Why remove this?

pkg/expression/scalar_function.go
```diff
@@ -141,9 +141,27 @@ func (sf *ScalarFunction) String() string {
...
-	return []byte(fmt.Sprintf("%q", sf)), nil
+// String returns the string representation of the function
+func (sf *ScalarFunction) String() string {
```
XuHuaiyu, line 146: Why not invoke StringWithCtx(nil) ?

pkg/expression/util_test.go
```diff
@@ -558,8 +558,12 @@ func (m *MockExpr) VecEvalJSON(ctx EvalContext, input *chunk.Chunk, result *chun
 	return nil
 }
 
-func (m *MockExpr) String() string               { return "" }
-func (m *MockExpr) MarshalJSON() ([]byte, error) { return nil, nil }
+<<<<<<< Updated upstream
```
XuHuaiyu: need resolve

pkg/expression/scalar_function.go
YangKeao: It's useless. I removed some implementation of `MarshalJSON` in https://github.com/pingcap/tidb/pull/53922, but I forgot to remove the implementation of `ScalarFunction` in that PR :facepalm: .

It's originally introduced to implement `explain` (a long time ago), but it's not used anymore now.

pkg/expression/scalar_function.go
YangKeao: You are right. I'll modify it.

XuHuaiyu (approved):

Issue #53246, opened 2024-05-14
optimize execution of general DDL 


D3Hunter:
## Enhancement
we have optimize `create table` using [fast-create](https://docs.pingcap.com/tidb/dev/accelerated-table-creation), but there are scenarios that we might create a lot of database, or run many DDLs like add/modify column in the maintenance window.
so we want to optimize execution of general DDL, and in a unified way

- [x] https://github.com/pingcap/tidb/pull/53217
- [x] https://github.com/pingcap/tidb/pull/53882
- [ ] query since min job id to mitigate https://github.com/pingcap/tidb/issues/52905, below 2 PRs are reverted in https://github.com/pingcap/tidb/pull/54180
  - [x] https://github.com/pingcap/tidb/pull/54122
  - [x] https://github.com/pingcap/tidb/pull/54142
- [x] move job dependency calculation into memory, to avoid the slower SQL
  - [x] https://github.com/pingcap/tidb/pull/54003
  - [x] https://github.com/pingcap/tidb/pull/54110
- [x] https://github.com/pingcap/tidb/pull/54041
- [x] https://github.com/pingcap/tidb/pull/53747
  - [x] https://github.com/pingcap/tidb/pull/53944
  - [x] https://github.com/pingcap/tidb/pull/53887
- [ ] [low priority] notify the session which is not on DDL owner node when owner finished handling by RPC or other way
- [ ] [hold] add index on `type` for `tidb_ddl_job` to make some query faster
- [ ] other refactors
  - [x] https://github.com/pingcap/tidb/pull/53548
  - [x] https://github.com/pingcap/tidb/pull/53816
  - [x] https://github.com/pingcap/tidb/pull/54141

D3Hunter:
close, as 8.2 code freeze  and will release this part

PR #53548, opened 2024-05-24
ddl: decouple job scheduler from 'ddl' and make it run/exit as owner changes

D3Hunter:

Issue Number: ref #53246 

Problem Summary:

- decouple ddl job scheduler part out of `ddl` partially, some fields are still coupled, and make it a separate and cancellable module. it should make further test easier.
	- all context usage now changed to scheduler/worker context which is only valid during being owner from ddl context
	- current ddl context is non-cancellable, which might cause some issue when owner change, as it cannot be cancelled in time, it might keeps running until we hopefully met some `isOwner` check, so the duration of co-exist of 2 owner might be quite long.
- after this pr, job scheduler related routine will only be start when current node is elected as owner. start `SyncJobSchemaVerLoop` only on owner node.
- fix `add-index` disttask scheduler uses the 'ddlCtx.ctx', it should use the context we pass in.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
  - tiup a cluster with 3 tidb
  - run a script to create 1000 tables with 10 threads, during this time we keeps delete keys of prefix `/tidb/ddl/fg/owner/`, to force owner change
  - ddl should be run success, and no abnormal output in logs.

```sql
mysql> select count(1) from INFORMATION_SCHEMA.tables where table_schema='ff_0';
+----------+
| count(1) |
+----------+
|     1000 |
+----------+
1 row in set (0.01 sec)
```

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/ddl/backfilling_dist_executor.go
```diff
@@ -92,6 +92,8 @@ func (s *backfillDistExecutor) newBackfillSubtaskExecutor(
 	jobMeta := &s.taskMeta.Job
 	ddlObj := s.d
 
+	// TODO getTableByTxn is using DDL ctx which is never cancelled except when shutdown.
+	// we should move this heavy operation out.
```
D3Hunter, line 96: @tangenta ptal

pkg/owner/manager.go
```diff
@@ -62,9 +67,8 @@ type Manager interface {
 	RequireOwner(ctx context.Context) error
 	// CampaignCancel cancels one etcd campaign
 	CampaignCancel()
-
-	// SetBeOwnerHook sets a hook. The hook is called before becoming an owner.
-	SetBeOwnerHook(hook func())
+	// SetListener sets the listener, set before start campaign.
```
lance6716, line 70:
```suggestion
	// SetListener sets the listener, set before CampaignOwner.
```

lance6716 (commented):
will review soon

lance6716 (commented):
7 / 19 files viewed

pkg/ddl/job_table.go
```diff
@@ -82,7 +86,99 @@ const (
...
+	cancel      context.CancelFunc
+	wg          tidbutil.WaitGroupWrapper
+	runningJobs *runningJobs
+
+	// those fields are created on start
+	reorgWorkerPool      *workerPool
+	generalDDLWorkerPool *workerPool
+
+	// those fields are shared with 'ddl' instance
+	// TODO ddlCtx is to larger for here, we should remove dependency on it.
```
tangenta, line 131:
```suggestion
	// TODO ddlCtx is too large for here, we should remove dependency on it.
```

tangenta (approved):

D3Hunter:
the failure of `TestSkipSchemaChecker` is due to https://github.com/pingcap/tidb/issues/53599

lance6716 (dismissed):

pkg/owner/manager_test.go
```diff
@@ -78,6 +79,17 @@ func (ti *testInfo) Close(t *testing.T) {
 	ti.cluster.Terminate(t)
 }
 
+type listener struct {
+	val atomic.Bool
```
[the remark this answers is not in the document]
lance6716: we should not use atomic variables here because the real listeners are not concurrent safe, like l.scheduler is a normal field. If we don't use atomic variables herem, race detector in unit tests can report error when `ownerManager` breaks the no concurrent calling property.

tiancaiamao (approved):

D3Hunter:
TestAddIndexFinishImportError fail due to https://github.com/pingcap/tidb/issues/53641

D3Hunter:
TestFirstLitSlowStart failed part due to https://github.com/pingcap/tidb/issues/53659

tangenta (approved):

pkg/ddl/reorg.go
```diff
@@ -289,11 +289,6 @@ func (w *worker) runReorgJob(reorgInfo *reorgInfo, tblInfo *model.TableInfo,
 		if err != nil {
 			return errors.Trace(err)
 		}
-	case <-w.ctx.Done():
```
lance6716, line 292: I see that we rely on doneCh to find w.ctx.Done(). How about change the signature of `f` to have a context parameter, and in runReorgJob use `w.ctx` as argument. So we have more confident to remove this `case <-w.ctx.Done():`

pkg/ddl/reorg.go
```diff
@@ -289,11 +289,6 @@ func (w *worker) runReorgJob(reorgInfo *reorgInfo, tblInfo *model.TableInfo,
...
-		logutil.DDLLogger().Info("run reorg job quit")
-		d.removeReorgCtx(job.ID)
-		// We return dbterror.ErrWaitReorgTimeout here too, so that outer loop will break.
-		return dbterror.ErrWaitReorgTimeout
```
lance6716, line 296: the error returned to caller is also changed, not sure about its effect.

lance6716 (commented):
20 / 29 files viewed

reviewing

lance6716 (commented):
rest lgtm. You can unhold yourself after address comments.

pkg/ddl/reorg.go
D3Hunter: all current `f`s also calls through `worker`, it should be fine except that we still uses `background` context inside, will change it later

pkg/ddl/reorg.go
D3Hunter: previous ctx only cancells when shutdown, this error shouldn't matter much

PR #53605, opened 2024-05-28
table: memoize TableCommon.getCols

### LINKED ISSUE #53604: TableCommon.getCols() can be memorized
## Enhancement
Every call of getCols() re-calculates the result, which is time-consuming. Since we assume the schema of the table won't change in the lifetime of a TableCommon struct, we can memoize the function.

ekexium:

Issue Number: close #53604 

Problem Summary:

`getCols` does duplicated calculation, which wastes CPU time.

Memoize this function.

Proportion of time in the microbenchmark of AddRecord (with pipelined DML and ART memdb)

Previous:
image

This PR:
image

### Check List

Tests 

- [x] Unit test - micro benchmark + existing test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/table/tables/bench_test.go
+++ b/pkg/table/tables/bench_test.go
@@ -1,0 +1,39 @@
+package tables_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/pingcap/tidb/pkg/parser/model"
+	"github.com/pingcap/tidb/pkg/sessiontxn"
+	"github.com/pingcap/tidb/pkg/testkit"
+	"github.com/pingcap/tidb/pkg/types"
+	_ "github.com/pingcap/tidb/pkg/util/context"
+	"github.com/stretchr/testify/require"
+)
+
+func BenchmarkAddRecordInPipelinedDML(b *testing.B) {
+	store, dom := testkit.CreateMockStoreAndDomain(b)
+	tk := testkit.NewTestKit(b, store)
+	_, err := tk.Session().Execute(context.Background(), "CREATE TABLE test.t (a int primary key auto_increment, b varchar(255))")
+	require.NoError(b, err)
+	ctx := tk.Session()
+	vars := ctx.GetSessionVars()
+	vars.BulkDMLEnabled = true
+	vars.TxnCtx.EnableMDL = true
+	vars.StmtCtx.InInsertStmt = true
+	require.Nil(b, sessiontxn.NewTxn(context.Background(), tk.Session()))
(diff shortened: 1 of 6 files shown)


pkg/table/tables/tables.go
```diff
@@ -82,6 +82,10 @@ type TableCommon struct {
 	// recordPrefix and indexPrefix are generated using physicalTableID.
 	recordPrefix kv.Key
 	indexPrefix  kv.Key
+
+	// avoid duplicated calculations of Cols
+	// mode -> cols
+	getColsCache [][]*table.Column
```
Defined2014, line 88: Maybe we could use `hiddenColumns`, `visibleColumns` and `publicColumns` instead of this one.

ekexium: When are these fields supposed to be set (before this PR)? I don't find anywhere initializing them

Defined2014: Seems never be setted. /cc @zimulala

ekexium: Seems https://github.com/pingcap/tidb/pull/50062 introduced the regression. It removed the default initialization, but mistakenly making them never initialized.

pkg/table/tables/bench_test.go
```diff
@@ -0,0 +1,39 @@
+package tables_test
```
hawkingrei, line 1: Please add a licenser header.

ekexium:
Test failed because `IMPORT INTO` concurrently calls `AddRecord`, which calls `WritableConstraint()`, resulting in data race.
I'd rather let lightning manually initialize these fields than introducing something like sync.Once here.

pkg/table/tables/tables.go
```diff
@@ -300,26 +315,26 @@ func (t *TableCommon) getCols(mode getColsMode) []*table.Column {
 
 // Cols implements table.Table Cols interface.
 func (t *TableCommon) Cols() []*table.Column {
-	if len(t.publicColumns) > 0 {
-		return t.publicColumns
+	if len(t.publicColumns) == 0 {
+		t.publicColumns = t.getCols(full)
```
zimulala, line 319: It seems that there is a parallel call problem here, such as multiple sessions using the same table in the same version of Infoschema.

Defined2014 (approved):

zimulala (approved):
LGTM

PR #53656, opened 2024-05-29
expression, *: Eval param get type with context

### LINKED ISSUE #53653: Add `EvalContext` argument for `Expression.GetType`
## Enhancement

Now, the `Constant` may rely on the internal session context to get the type of itself. It's not expected. It'd be better to add an argument `EvalContext` for `Expression.GetType`, and then we can read the param from it.

YangKeao:

Issue Number: close #53653

Problem Summary:

Ref https://github.com/pingcap/tidb/issues/53533 and https://github.com/pingcap/tidb/pull/53534. This PR splits the first commit of #53534  to make it easier to review.

The goal is to remove the `sessionctx.Session` inside the `Constant`. Instead, we should pass the session context through argument, especially for function `String()` and `GetType()`. This PR modifies the `GetType()` part. To make sure this patch can work standalone, the internal session in `Constant` is not removed yet. It'll be removed in https://github.com/pingcap/tidb/pull/53534 (which is ready, but is too long to review).

This PR adds an argument `EvalType` to the method `GetType` in `Expression`, so that in the future, the 

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > 
  > This refactor should be covered by existing tests. It shouldn't change any logic.

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/aggfuncs/func_group_concat.go
+++ b/pkg/executor/aggfuncs/func_group_concat.go
@@ -242,7 +242,7 @@ func (e *groupConcatDistinct) UpdatePartialResult(sctx AggFuncUpdateContext, row
 
 	collators := make([]collate.Collator, 0, len(e.args))
 	for _, arg := range e.args {
-		collators = append(collators, collate.GetCollator(arg.GetType().GetCollate()))
+		collators = append(collators, collate.GetCollator(arg.GetType(sctx).GetCollate()))
 	}
 
 	for _, row := range rowsInGroup {
@@ -429,6 +429,8 @@ type partialResult4GroupConcatOrder struct {
 
 type groupConcatOrder struct {
 	baseGroupConcat4String
+	ctors []collate.Collator
+	desc  []bool
 }
 
 func (e *groupConcatOrder) AppendFinalResult2Chunk(_ AggFuncUpdateContext, pr PartialResult, chk *chunk.Chunk) error {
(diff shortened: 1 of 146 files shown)


pkg/executor/aggfuncs/func_group_concat.go
```diff
@@ -429,6 +429,8 @@ type partialResult4GroupConcatOrder struct {
 
 type groupConcatOrder struct {
 	baseGroupConcat4String
+	ctors []collate.Collator
```
qw4990, line 432: Why add these 2 new fields?

YangKeao: Because in `func (e *groupConcatOrder) AllocPartialResult()`, it'll use these fields to allocate partial result. Previously, it was generated in `AllocPartialResult`. However, add an argument `EvalContext` to `AllocPartialResult` will lead to tons of changes, so I choose to calculate the `ctors` and `desc` when this struct is constructed.

qw4990 (approved):

lcwangchao (approved):

lance6716 (approved):

pkg/expression/constant.go
```diff
@@ -135,15 +135,20 @@ type ParamMarker struct {
 }
 
 // GetUserVar returns the corresponding user variable presented in the `EXECUTE` statement or `COM_EXECUTE` command.
-func (d *ParamMarker) GetUserVar() types.Datum {
+func (d *ParamMarker) GetUserVar(ctx EvalContext) types.Datum {
+	return ctx.GetParamValue(d.order)
```
zimulala, line 139: Why don't we just consider `EvalContext` as a field of `ParamMarker` like `d.ctx`? Then we can remove the call like `ctx.GetExprCtx().GetEvalCtx()`.

YangKeao: Because the goal of this PR is to finally remove `d.ctx`. To remove the `d.ctx`, it's required to add a `ctx EvalContext` argument to each method. It'll be a big change, so this PR only modifies `GetType` method.

This PR didn't remove the `d.ctx` because the `String()` method is not modified, and is still using the `d.ctx`. It'll finally be removed in https://github.com/pingcap/tidb/pull/53534 (which has two more commits based on this change).

zimulala (approved):
LGTM

XuHuaiyu (approved):

PR #53747, opened 2024-06-03
ddl:  manage job lifecycle in worker

D3Hunter:

Issue Number: ref #53246

Problem Summary:

- manage job life cycle inside worker
- move some methods for querying DDL related system tables to a separate pkg
- replace some `background` contexts to DDL worker ctx
- add context param to `begin/commit`, `rollback` is not changed as it doesn't return err and the main usages of ctx is to extract value

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


lance6716 (commented):
14 / 24 files viewed

pkg/ddl/ddl_worker.go
```diff
@@ -637,41 +637,25 @@ func cleanMDLInfo(pool *sess.Pool, job *model.Job, ec *clientv3.Client, ownerID
...
 	se := sess.NewSession(sctx)
 	se.GetSessionVars().SetDiskFullOpt(kvrpcpb.DiskFullOpt_AllowedOnAlmostFull)
-	_, err := se.Execute(context.Background(), sql, "delete-mdl-info")
+	_, err := se.Execute(s.schCtx, sql, "delete-mdl-info")
 	if err != nil {
 		logutil.DDLLogger().Warn("unexpected error when clean mdl info", zap.Int64("job ID", job.ID), zap.Error(err))
 		return
 	}
-	if cleanETCD && ec != nil {
+	// TODO do we need clean it for JobStateRollbackDone?
```
tangenta, line 649: `JobStateRollbackDone` does trigger schema version changes for some DDLs, if we don't clean keys here, there may be a leak?

D3Hunter: yes, that's why i want to add a JobStateRollbackSynced, so we can check it easily

pkg/ddl/ddl_running_jobs.go
```diff
@@ -53,40 +53,38 @@ func (j *runningJobs) clear() {
...
 
-func (j *runningJobs) remove(job *model.Job) {
+func (j *runningJobs) remove(jobID int64, involvedSchemaInfos []model.InvolvingSchemaInfo) {
 	j.Lock()
 	defer j.Unlock()
-	delete(j.processingIDs, job.ID)
+	delete(j.processingIDs, jobID)
 	j.updateInternalRunningJobIDs()
 
-	if job.IsFinished() || job.IsSynced() {
```
D3Hunter, line 81: reorg jobs will check whether there are jobs of same table-id are `in-processing` in `getReorgJob`, so for a table-id there shouldn't be 2 jobs running concurrently, so it's ok to delete those unfinishedXXX all the time

pkg/infoschema/test/clustertablestest/cluster_tables_test.go
```diff
@@ -1776,14 +1777,14 @@ func TestMDLViewIDConflict(t *testing.T) {
 	wg := &sync.WaitGroup{}
 	wg.Add(1)
 	go func() {
-		ddlTK1.MustExec("ALTER TABLE t ADD COLUMN b INT;")
+		ddlTK1.MustExec("ALTER TABLE t ADD index(a);")
```
D3Hunter, line 1780: now the routine will hold worker until it exist, so we change to reorg job to make this test work

D3Hunter:
@lance6716 @tangenta test pass, ptal

pkg/ddl/reorg.go
```diff
@@ -892,13 +893,13 @@ func (r *reorgInfo) UpdateReorgMeta(startKey kv.Key, pool *sess.Pool) (err error
 	defer pool.Put(sctx)
 
 	se := sess.NewSession(sctx)
-	err = se.Begin()
+	err = se.Begin(context.Background())
```
lance6716, line 896: maybe use r.d.ctx

lance6716 (approved):

pkg/ddl/reorg.go
D3Hunter: that's no much difference with background, it's only cancelled at shutdown. we should refactor to use job worker  ctx or disttask ctx later

pkg/ddl/ddl_running_jobs.go
tangenta: @D3Hunter Is it not necessary to distinguish `processingIDs` from `unfinishedIDs`?

pkg/ddl/job_table.go
```diff
@@ -507,93 +510,123 @@ func (d *ddl) delivery2LocalWorker(pool *workerPool, task *limitJobTask) {
...
+	metrics.DDLRunningJobCount.WithLabelValues(pool.tp().String()).Inc()
+	s.wg.RunWithLog(func() {
 		defer func() {
+			pool.put(wk)
 			failpoint.InjectCall("afterDelivery2Worker", job)
-			s.runningJobs.remove(job)
+			s.runningJobs.remove(jobID, involvedSchemaInfos)
 			asyncNotify(s.ddlJobNotifyCh)
 			metrics.DDLRunningJobCount.WithLabelValues(pool.tp().String()).Dec()
 			if wk.ctx.Err() != nil && ingest.LitBackCtxMgr != nil {
```
tangenta, line 523: At this point of time, `wk` is put back into the pool and may be get by someone else. But it is still used by here, will there be any problems?

D3Hunter: wk.ctx is same as scheduler.schCtx now, it should be ok，but let's move `push back` after last usage

D3Hunter: https://github.com/pingcap/tidb/pull/53747/commits/1e89963a3403428fb7df4ac91fa6eb889e91ac77

pkg/ddl/ddl_running_jobs.go
D3Hunter: like what we have discussed, `runningJobs.remove` will only called when the job is in final state like synced/paused/cancelled or it's done by another owner, before that another job operating the same table-id or name should be blocked

PR #53793, opened 2024-06-04
br: set memory limit

### LINKED ISSUE #53777: Automatically adjust `GOMEMLIMIT` for br clp

Leavrth:

Issue Number: close #53777

Problem Summary:
need to automatically adjust `GOMEMLIMIT` for br clp
automatically adjust `GOMEMLIMIT` for br clp

```
[INFO] [meminfo.go:196] ["use physical memory hook"] [cgroupMemorySize=9223372036854771712] [physicalMemorySize=33731502080]
[INFO] [cmd.go:168] ["memory limit"] [limit=9223372036854775807]
[INFO] [cmd.go:189] [memory] [memtotal=33731502080] [memused=5004333056]
[INFO] [cmd.go:191] ["set memory limit"] [limit=28190298112]
```

```
[INFO] [meminfo.go:179] ["use cgroup memory hook because TiDB is in the container"]
[INFO] [cmd.go:168] ["memory limit"] [limit=9223372036854775807]
[INFO] [cmd.go:189] [memory] [memtotal=8589934592] [memused=971722752]
[INFO] [cmd.go:191] ["set memory limit"] [limit=7081340928]
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


br/cmd/br/cmd.go
```diff
@@ -162,6 +165,32 @@ func Init(cmd *cobra.Command) (err error) {
...
+			memleft := memtotal - memused - halfGB
+			// 0       512MB               +inf
+			// +--------+-------------------->
+			//          ^____________________>
+			//             GOMEMLIMIT range
+			log.Info("calculate the rest memory", zap.Uint64("memtotal", memtotal), zap.Uint64("memused", memused))
+			if memleft > halfGB {
+				log.Info("set memory limit", zap.Uint64("limit", memleft))
+				debug.SetMemoryLimit(int64(memleft))
+			}
```
kennytm, line 192: if `memtotal - memused < halfGB` the subtraction in `memleft` will overflow causing BR to set GOMEMLIMIT to an absurdly large value.

Leavrth: the first `halfGB` is aimed to keep the reserved memory to avoid OOM.
the second `halfGB` is aimed to give up use memory limit because the `memleft` may be too small for br. I am worried that there is a high probability that it triggers frequent GC and then oom.

Besides, the current method of calculating reserved memory is too coarse, I will modify it.

br/cmd/br/cmd.go
```diff
@@ -107,6 +110,23 @@ func AddFlags(cmd *cobra.Command) {
 	_ = cmd.PersistentFlags().MarkHidden(FlagRedactLog)
 }
 
+func calculateMemoryLimit(memleft uint64) uint64 {
```
BornChanger, line 113: Do we follow the logic of tidb? if yes, can we reuse the their code?

Leavrth: the tidb go memory limit tuner would reset the memory limit by TiDB node's `tidb_server_memory_limit`, which may large than BR node's memory.
For BRIE via SQL, BR task runs on a TiDB node, so it can use the memory limit tuner. 
For br clp, BR has forbidden the memory limit tuner. Besides, for br clp, the frequent GC will not affect the work of BR too much. Therefore, for simplicity, we can just fix the memory limit at the beginning.

BornChanger (approved):

br/cmd/br/cmd.go
```diff
@@ -162,6 +182,23 @@ func Init(cmd *cobra.Command) (err error) {
...
+			memused, e := memory.MemUsed()
+			if e != nil {
+				err = e
+				return
+			}
+			memleft := memtotal - memused
+			memlimit := calculateMemoryLimit(memleft)
+			log.Info("calculate the rest memory",
+				zap.Uint64("memtotal", memtotal), zap.Uint64("memused", memused), zap.Uint64("memlimit", memlimit))
+			debug.SetMemoryLimit(int64(memlimit))
```
Tristan1900, line 200: to be extra safe, can we do a safe cast from uint64 to int64, something like
```
func safeCastUint64ToInt64(u uint64) int64 {
	if u > uint64(math.MaxInt64) {
		return math.MaxInt64
	}
	return int64(u)
}
```
and make it a util method?
cuz in the doc
 `
A zero limit or a limit that's lower than the amount of memory
used by the Go runtime may cause the garbage collector to run
 nearly continuously. However, the application may still make
 progress.
`

br/cmd/br/cmd.go
```diff
@@ -107,6 +110,23 @@ func AddFlags(cmd *cobra.Command) {
 	_ = cmd.PersistentFlags().MarkHidden(FlagRedactLog)
 }
 
+func calculateMemoryLimit(memleft uint64) uint64 {
+	const halfGB = uint64(512 * 1024 * 1024)
+	const fourGB = 8 * halfGB
+	// memreserved = f(memleft) = 512MB * memleft / (memleft + 4GB)
```
Tristan1900, line 116: just for my knowledge, what is our memory footprint like in br cli, is max 512MB large enough?

br/cmd/br/cmd.go
```diff
@@ -107,6 +110,23 @@ func AddFlags(cmd *cobra.Command) {
...
+	//  * f(0) = 0
```
Tristan1900, line 117: if the mem unused is 0, do we even bother to start the BR? I guess a correctly configured machine should not use up all its mem and if they do they are facing a bigger problem and we probably don't want to start a BR to make things more complicated. Let me know your thoughts!

br/cmd/br/cmd.go
```diff
@@ -107,6 +110,23 @@ func AddFlags(cmd *cobra.Command) {
 	_ = cmd.PersistentFlags().MarkHidden(FlagRedactLog)
 }
 
+func calculateMemoryLimit(memleft uint64) uint64 {
+	const halfGB = uint64(512 * 1024 * 1024)
```
Tristan1900, line 114: I feel like it would be better to put it into some util file for reuse.

br/cmd/br/main_test.go
```diff
@@ -75,3 +76,18 @@ func TestRunMain(*testing.T) {
 
 	<-waitCh
 }
+
+func TestCalculateMemoryLimit(t *testing.T) {
```
Tristan1900, line 80: Could we add a test case for invalid input, like -1024 or something. just in case 
`meqleft := memtotal - memused` returns a negative value,

Tristan1900 (commented):
Overall LGTM, thanks for the change. just a few questions and suggestions,

br/cmd/br/cmd.go
```diff
@@ -107,6 +110,23 @@ func AddFlags(cmd *cobra.Command) {
...
+	const fourGB = 8 * halfGB
+	// memreserved = f(memleft) = 512MB * memleft / (memleft + 4GB)
+	//  * f(0) = 0
+	//  * f(4GB) = 256MB
+	//  * f(+inf) -> 512MB
+	memreserved := halfGB / (1 + fourGB/(memleft+1))
```
Tristan1900, line 120: extreme case, if memleft is unit64 max it's going to overflow I guess

br/cmd/br/cmd.go
```diff
@@ -162,6 +182,23 @@ func Init(cmd *cobra.Command) (err error) {
...
+			if e != nil {
+				err = e
+				return
+			}
+			memused, e := memory.MemUsed()
+			if e != nil {
+				err = e
+				return
+			}
+			memleft := memtotal - memused
```
Tristan1900, line 196: do we need to add a safety check here to make sure `memtotal-memused` won't underflow? since we own the memory package and if might have a bug that returned memtotal is actually smaller than memused

br/cmd/br/cmd.go
Leavrth: Actually, the `memreserved` is not the memory reserved for BR, but it is the memory reserved for system. Make sure the system still has `memreserved` memory to avoid to triggering OOM.
The max memory reserved for BR is `memleft - memreserved`.

For backup, the memory footprint is
```
Domain:
  many tables Info (about 4GB if 1 million tables)

Ranges:
  no more than 1 GB (about 2 million ranges)

Temporary Table Infos BR dealing with:
  no more than 256 MB (kv batch default cache size)

Temporary Schema data to be flushed to external storage:
  no more than 256 MB

(Optional) Statistic:
  no more than schema-concurrency * max memory size of partition/table's statistic

Other Grpc memory: ...
```

For restore, the memory footprint is
```
Domain:
  many tables Info (about 4GB if 1 million tables)

All the backupmeta from external storage:
  many tables info (about 4GB if 1 million tables)
  no more than 1 GB (about 2 million ranges)

(Optional) Statistic: 
  no more than schema-concurrency * max memory size of partition/table's statistic

Goroutines restoring SSTs files: (The workpool is very large)
  some scan region information
  some sst meta
  some grpc memory
  goroutine stack itself
```

br/cmd/br/cmd.go
Leavrth: I think maybe it is already OOM. Maybe we can directly limit the minimum remaining memory.

br/cmd/br/cmd.go
kennytm: in what case would memleft be uint64max (= 16 EiB) 👀 

anyway changing `memleft + 1` to `memleft | 1` should be good enough

br/cmd/br/cmd.go
Tristan1900: thanks! That's really useful. 
oh I see, yeah I thought it's the memory reserved for the BR cli.

br/cmd/br/cmd.go
Tristan1900: yeah, I feel like we can have some minimum memory not starting from 0, if below that we just throw error and exit.

br/cmd/br/cmd.go
Tristan1900: in the unexpected error case I would say, the `memleft` is passed down by  `memleft := memtotal - memused`, if `memtotal < memused`, it can underflow as in my other comment. I feel like we probably should explicitly check and make sure it won't go wrong in any extreme case, cuz a lot of assumptions are false when running in production.

br/cmd/br/cmd.go
Leavrth: done. If underflow, output a warning log and skip set memory limit.

br/cmd/br/cmd.go
Leavrth: done. And add check `memtotal < memused` in another comment.

br/cmd/br/cmd.go
Leavrth: done. Reuse the `tidb/pkg/util/size.MB`.

br/cmd/br/cmd.go
Leavrth: I add the minimum value (`256MiB`) of memory limit. It's hard to determine the minimum memory br usage. In my test, br uses only about 200 MB~300 MB memory to backup data generated by `tiup bench tpcc prepare --warehouses 3`.

br/cmd/br/cmd.go
Leavrth: done. `debug.SetMemoryLimit(int64(memlimit & math.MaxInt64))`

br/cmd/br/cmd.go
Tristan1900: thanks for digging into that, yeah having an estimation of memory usage very useful.

br/cmd/br/cmd.go
```diff
@@ -162,6 +186,34 @@ func Init(cmd *cobra.Command) (err error) {
...
+				return
+			}
+			if memused >= memtotal {
+				log.Warn("failed to obtain memory size, skip setting memory limit",
+					zap.Uint64("memused", memused), zap.Uint64("memtotal", memtotal))
+			} else {
+				memleft := memtotal - memused
+				memlimit := calculateMemoryLimit(memleft)
+				// BR command needs 256 MiB at least, if the left memory is less than 256 MiB,
+				// the memory limit cannot limit anyway and then finally OOM.
```
Tristan1900, line 207: It seems at here the code will keep running and let the system OOM and stop. 
In my mind it's better to explicitly panic or throw error by ourselves and exit early instead of actually OOM the system, it might cause unexpected issue to the other processes running on the same machine. what do you think?

Tristan1900: rest LGTM!

Leavrth: The memory usage of br depends in part on the cluster tables' information. Only when br starts a domain to loads all the cluster's table's information into memory, it will know how much memory it will actually use at least. Maybe it is 200 MB or 1 GB, so I think it's hard to determine the memory usage here. Besides, setting go memory limit in this PR just makes the process runtime free the unused memory back to system in time, I think estimate the process memory usage is another issue.

Leavrth: If the command is `br debug ...`, it will use a little memory. Maybe more work is needed.

Issue #53713, opened 2024-05-31
Incorrect swaping column leads to nil pointer


xzhangxian1008:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

1. `source valid.txt`, 
[valid.txt](https://github.com/user-attachments/files/15510266/valid.txt)
2. `source error.txt`, 
[error.txt](https://github.com/user-attachments/files/15510272/error.txt)

If no error happens, try it again.

### 2. What did you expect to see? (Required)

no error

### 3. What did you see instead (Required)

`ERROR 1105 (HY000): runtime error: index out of range [0] with length 0`

### 4. What is your TiDB version? (Required)

0afe54d7562f31bbcc

xzhangxian1008:
panic stack:
```
runtime error: index out of range [0] with length 0
github.com/pingcap/errors.AddStack
	/DATA/disk3/xzx/go/path/pkg/mod/github.com/pingcap/errors@v0.11.5-0.20240318064555-6bd07397691f/errors.go:178
github.com/pingcap/errors.Trace
	/DATA/disk3/xzx/go/path/pkg/mod/github.com/pingcap/errors@v0.11.5-0.20240318064555-6bd07397691f/juju_adaptor.go:15
github.com/pingcap/tidb/pkg/util.GetRecoverError
	/DATA/disk3/xzx/tidb/pkg/util/util.go:304
github.com/pingcap/tidb/pkg/executor/aggregate.recoveryHashAgg
	/DATA/disk3/xzx/tidb/pkg/executor/aggregate/agg_util.go:91
github.com/pingcap/tidb/pkg/executor/aggregate.(*HashAggPartialWorker).run.func1
	/DATA/disk3/xzx/tidb/pkg/executor/aggregate/agg_hash_partial_worker.go:194
runtime.gopanic
	/DATA/disk3/xzx/go/src/runtime/panic.go:920
runtime.goPanicIndex
	/DATA/disk3/xzx/go/src/runtime/panic.go:114
github.com/pingcap/tidb/pkg/util/chunk.(*Column).IsNull
	/DATA/disk3/xzx/tidb/pkg/util/chunk/column.go:180
github.com/pingcap/tidb/pkg/util/codec.HashGroupKey
	/DATA/disk3/xzx/tidb/pkg/util/codec/codec.go:1247
github.com/pingcap/tidb/pkg/executor/aggregate.GetGroupKey
[...]

PR #53794, opened 2024-06-04
planner: fix column evaluator can not detect input's column-ref and thus swapping and destroying later column ref projection logic

AilinKid:

Issue Number: close https://github.com/pingcap/tidb/issues/53713

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
use the valid SQL and error SQL inside issue to produce result
and compare it with MySQL output, both them output like below:
```
mysql> source /Users/arenatlx/Downloads/error.txt
+------+------+------+------+
| c1   | c3   | c4   | c5   |
+------+------+------+------+
| NULL | NULL | 4    | NULL |
+------+------+------+------+
1 row in set (0.04 sec)
```

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fixed an issue where the column evaluator could not detect column references in the input chunk. This bug caused the column evaluator to potentially lose the original input column data during projection, disrupting subsequent reference projection logic.
修复了 column evaluator 无法识别输入 chunk 中的列引用的问题。该问题会导致在投影过程中，column evaluator 可能丢失原始输入列的内容，从而破坏后续引用投影逻辑。


```

### DIFF
--- a/pkg/expression/evaluator.go
+++ b/pkg/expression/evaluator.go
@@ -19,12 +19,13 @@ import (
 	"github.com/pingcap/tidb/pkg/util/chunk"
 	"github.com/pingcap/tidb/pkg/util/disjointset"
 	"github.com/pingcap/tidb/pkg/util/intest"
+	"sync/atomic"
 )
 
 type columnEvaluator struct {
 	inputIdxToOutputIdxes map[int][]int
 	// mergedInputIdxToOutputIdxes is only determined in runtime when saw the input chunk.
-	mergedInputIdxToOutputIdxes map[int][]int
+	mergedInputIdxToOutputIdxes atomic.Pointer[map[int][]int]
 }
 
 // run evaluates "Column" expressions.
@@ -33,10 +34,10 @@ type columnEvaluator struct {
 //	since it will change the content of the input Chunk.
 func (e *columnEvaluator) run(ctx EvalContext, input, output *chunk.Chunk) error {
 	// mergedInputIdxToOutputIdxes only can be determined in runtime when we saw the input chunk structure.
-	if e.mergedInputIdxToOutputIdxes == nil {
+	if e.mergedInputIdxToOutputIdxes.Load() == nil {
 		e.mergeInputIdxToOutputIdxes(input, e.inputIdxToOutputIdxes)
 	}
-	for inputIdx, outputIdxes := range e.mergedInputIdxToOutputIdxes {
+	for inputIdx, outputIdxes := range *e.mergedInputIdxToOutputIdxes.Load() {
 		if err := output.SwapColumn(outputIdxes[0], input, inputIdx); err != nil {
 			return err
 		}
@@ -129,7 +130,9 @@ func (e *columnEvaluator) mergeInputIdxToOutputIdxes(input *chunk.Chunk, inputId
 		mergedOutputIdxes = append(mergedOutputIdxes, inputIdxToOutputIdxes[inputIdx]...)
 		newInputIdxToOutputIdxes[originalVal] = mergedOutputIdxes
 	}
-	e.mergedInputIdxToOutputIdxes = newInputIdxToOutputIdxes
+	// Update the merged inputIdxToOutputIdxes automatically.
+	// Once failed, it means other worker has done this job at meantime.
+	e.mergedInputIdxToOutputIdxes.CompareAndSwap(nil, &newInputIdxToOutputIdxes)
 }
 
 type defaultEvaluator struct {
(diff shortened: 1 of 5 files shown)


pkg/expression/evaluator.go
```diff
@@ -38,6 +44,93 @@ func (e *columnEvaluator) run(ctx EvalContext, input, output *chunk.Chunk) error
...
+		}
+		// if we found, root should be redirected to real root of inputIdx (link the real root)
+		// if we didn't, root should be the original inputIdx as it was. (inputIdx itself is a root)
+		if _, ok := rootIdxToOutputIdxes[root]; ok {
+			rootIdxToOutputIdxes[root] = append(rootIdxToOutputIdxes[root], outputIdxes...)
+		} else {
+			// if not find in any value set, it means input idx itself is a root.
+			rootIdxToOutputIdxes[root] = append(rootIdxToOutputIdxes[root], outputIdxes...)
+		}
+	}
```
winoros, line 130: I think we're doing things like the disjoint set? There's already a implemented disjoint set in `util`

AilinKid: Make sense

pkg/expression/evaluator.go
```diff
@@ -39,6 +46,82 @@ func (e *columnEvaluator) run(ctx EvalContext, input, output *chunk.Chunk) error
...
+// ___________________//  \\                      //  \\
+//
+// proj2:   a3(addr:0xe) a4(addr:0xe)   a3(addr:???) a4(addr:???)
+//
+// then when we start the projection for second <1, [2,3]>, the targeted column a2 addr has already been
+// swapped as an invalid one. so swapping between a2 <-> a3 and a4 is not safe anymore.
+//
+// keypoint: we should identify the original column ref from input chunk, and merge current inputIdxToOutputIdxes
+// as soon as possible. since input chunk's 0 and 1 is column referred. the final inputIdxToOutputIdxes should
+// be like: <0, [0,1,3,4]>.
```
XuHuaiyu, line 85: Optimized by chatgpt:

```suggestion
// mergeInputIdxToOutputIdxes merges separate inputIdxToOutputIdxes entries when column references
// are detected within the input chunk. This process ensures consistent handling of columns derived
// from the same original source.
//
// Consider the following scenario:
//
// Initial scan operation produces a column 'a':
//
// scan:                       a (addr: ???)
//
// This column 'a' is used in the first projection (proj1) to create two columns a1 and a2, both referencing 'a':
//
//                         proj1
//                        /     \
//                       /       \
//                      /         \
//        a1 (addr: 0xe)           a2 (addr: 0xe)
//        /                         \
//       /                           \
//      /                             \
//     proj2                          proj2
//     /     \                       /     \
//    /       \                     /       \
//   a3        a4                  a3        a4
// (addr: 0xe) (addr: 0xe)      (addr: 0xe) (addr: 0xe)
//
// Here, a1 and a2 share the same address (0xe), indicating they reference the same data from the original 'a'.
//
// When moving to the second projection (proj2), the system tries to project these columns further:
// - The first set (left side) consists of a3 and a4, derived from a1, both retaining the address (0xe).
// - The second set (right side) consists of a3 and a4, derived from a2, also starting with address (0xe).
//
// When proj1 is complete, the output chunk contains two columns [a1, a2], both derived from the single column 'a' from the scan.
// Since both a1 and a2 are column references with the same address (0xe), they are treated as referencing the same data.
//
// In proj2, two separate <inputIdx, []outputIdxes> items are created:
// - <0, [0,1]>: This means the 0th input column (a1) is projected twice, into the 0th and 1st columns of the output chunk.
// - <1, [2,3]>: This means the 1st input column (a2) is projected twice, into the 2nd and 3rd columns of the output chunk.
//
// Due to the column swapping logic in each projection, after applying the <0, [0,1]> projection,
// the addresses for a1 and a2 may become swapped or invalid:
//
// proj1:          a1 (addr: invalid)             a2 (addr: invalid)
//
// This can lead to issues in proj2, where further operations on these columns may be unsafe:
// 
// proj2:   a3 (addr: 0xe) a4 (addr: 0xe)   a3 (addr: ???) a4 (addr: ???)
//
// Therefore, it's crucial to identify and merge the original column references early, ensuring
// the final inputIdxToOutputIdxes mapping accurately reflects the shared origins of the data.
// For instance, <0, [0,1,2,3]> indicates that the 0th input column (original 'a') is referenced
// by all four output columns in the final output.

```

XuHuaiyu:
release note provide by gpt:

### Release Note - Chinese Version

**修复：**

修复了 column evaluator 无法识别输入 chunk 中的列引用的问题。该问题会导致在投影过程中，column evaluator 可能丢失原始输入列的内容，从而破坏后续引用投影逻辑。

### Release Note - English Version

**Fix:**

Fixed an issue where the column evaluator could not detect column references in the input chunk. This bug caused the column evaluator to potentially lose the original input column data during projection, disrupting subsequent reference projection logic.

XuHuaiyu (approved):

AilinKid:
```
tidb> source /Users/arenatlx/Downloads/error.txt
+------+------+----+------+
| c1   | c3   | c4 | c5   |
+------+------+----+------+
| NULL | NULL | 4  | NULL |
+------+------+----+------+
1 row in set (0.01 sec)
```
should be ok

winoros (approved):

PR #16367, opened 2020-04-14
expression: mark `random_bytes()` as deferred function

eurekaka:
### What problem does this PR solve?

Issue Number: close https://github.com/pingcap/tidb/issues/16205

Problem Summary:

wrong result for `random_bytes()` when plan cache is enable.

### What is changed and how it works?

What's Changed:

Mark it as `DeferredFunctions`.

### Related changes

- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test

Side effects

N/A

PR #53795, opened 2024-06-04
expression: move random_bytes from deferred to unfoldable functions

### LINKED ISSUE #37412: RANDOM_BYTES function is not random
## Bug Report

This is the easiest SQL to reproduce:

```sql
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `col1` varbinary(1024) NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY (`col1`)
);

INSERT INTO t1
 SELECT NULL, RANDOM_BYTES(1024) FROM dual;
INSERT INTO t1
 SELECT NULL, RANDOM_BYTES(1024) FROM t1;
INSERT INTO t1
 SELECT NULL, RANDOM_BYTES(1024) FROM t1;
```

The value is cached incorrectly. See `SELECT RANDOM_BYTES(n)` for another example: every row always has the same value.

All inserts should succeed

```sql
mysql> INSERT INTO t1
    ->  SELECT NULL, RANDOM_BYTES(1024) FROM dual;
Records: 1  Duplicates: 0  Warnings: 0

mysql> INSERT INTO t1
    ->  SELECT NULL, RANDOM_BYTES(1024) FROM t1;
Records: 1  Duplicates: 0  Warnings: 0

mysql> INSERT INTO t1
    ->  SELECT NULL, RANDOM_BYTES(1024) FROM t1;
ERROR 1062 (23000): Duplicate entry 't��Gt�IgN��C5
```

```sql
mysql> SELECT tidb_version()\G
tidb_version(): Release Version: v6.3.0-alpha-85-ga8f524b31
Edition: Community
Git Commit Hash: a8f524b31b26cfa4073f2c687bc6ffe55eade37c
Git Branch: master
UTC Build Time: 2022-08-26 15:29:13
GoVersion: go1.19
Race Enabled: false
TiKV Min Version: 6.2.0-alpha
Store: unistore
1 row in set (0.00 sec)
```

kennytm:

Issue Number: close #37412

Problem Summary: `RANDOM_BYTES` was marked as deferred by #16367. However, this is insufficient to prevent const-folding across rows. 

Mark `RANDOM_BYTES` to unfoldable rather than deferred, in the same league as `RAND` and `UUID`.

(Thanks to suggestion from [@‌unconsolable](https://github.com/pingcap/tidb/issues/37412#issuecomment-1230499189))

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
When a query involving RANDOM_BYTES() returned multiple rows, the function is re-evaluated for every row in every invocation, ensuring randomness across the entire result set.
```


kennytm:
🤔 The failure is legit. The `typeinfer_test` requires that:

* `RANDOM_BYTES(5)` has type `VARBINARY(1024) NOT NULL`
* `RANDOM_BYTES('123')` has type `VARBINARY(1024) NOT NULL`
* `RANDOM_BYTES('abc')` has type <code>VARBINARY(1024) <strong>NULL</strong></code>

Because this PR removed the const-folding, #23172 is not triggered and thus cannot mark `RANDOM_BYTES(5)` and `RANDOM_BYTES('123')` as NOT NULL (note that on MySQL the type is always `BLOB NULL`).

There are two paths to go next:

1. We declare that `RANDOM_BYTES()` always return `VARBINARY(1024) NULL` and change the test case
    * Because `RANDOM_BYTES(NULL)` is a thing, we *cannot* declare the function return type to be `VARBINARY(1024) NOT NULL` always.
2. We maintain the test case, which in turn requires creating a new category between "unfoldable" and "deferred" --- the *type information* is foldable but the *value* is not.

pkg/expression/integration_test/BUILD.bazel
```diff
@@ -8,7 +8,7 @@ go_test(
         "main_test.go",
     ],
     flaky = True,
-    shard_count = 50,
+    shard_count = 51,
```
kennytm:
```suggestion
    shard_count = 50,
```

PR #53832, opened 2024-06-05
tables: optimize memory allocation in tables' hot path

### LINKED ISSUE #53837: Optimize slice allocations in the hot path
## Enhancement
Some slice allocations can be costly. We can reuse old slices to save CPU time.

ekexium:

Issue Number: close #53837

Problem Summary:

Optimize memory allocations to save CPU time in the hot path.

Master:
```
go test -run=^$ -bench=Record -benchmem -benchtime=30s
goos: linux
goarch: amd64
pkg: github.com/pingcap/tidb/pkg/table/tables
cpu: AMD Ryzen 5 3600 6-Core Processor
BenchmarkAddRecordInPipelinedDML-12       	    1050	  28595471 ns/op	      2860 ns/record	 7818797 B/op	  161352 allocs/op
BenchmarkRemoveRecordInPipelinedDML-12    	    2382	  12638043 ns/op	      1264 ns/record	 2848806 B/op	   90455 allocs/op
BenchmarkUpdateRecordInPipelinedDML-12    	    1844	  16815817 ns/op	      1682 ns/record	 8296254 B/op	  110608 allocs/op
PASS
ok  	github.com/pingcap/tidb/pkg/table/tables	261.149s
```

This PR:
```
go test -run=^$ -bench=Record -benchmem -benchtime=30s
goos: linux
goarch: amd64
pkg: github.com/pingcap/tidb/pkg/table/tables
cpu: AMD Ryzen 5 3600 6-Core Processor
BenchmarkAddRecordInPipelinedDML-12       	    1569	  24222621 ns/op	      2422 ns/record	 5050033 B/op	  121106 allocs/op
BenchmarkRemoveRecordInPipelinedDML-12    	    3177	  10820754 ns/op	      1082 ns/record	 1736882 B/op	   70328 allocs/op
BenchmarkUpdateRecordInPipelinedDML-12    	    2612	  13296768 ns/op	      1330 ns/record	 3345944 B/op	   80497 allocs/op
PASS
ok  	github.com/pingcap/tidb/pkg/table/tables	288.480s
```

### Check List

Tests 

- [x] Unit test: existing tests and micro benchmark
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/sessionctx/variable/session.go
```diff
@@ -1618,6 +1618,38 @@ type SessionVars struct {
 
 	// GroupConcatMaxLen represents the maximum length of the result of GROUP_CONCAT.
 	GroupConcatMaxLen uint64
+
+	// TablesBufferPool is a memory pool for table related memory allocation that aims to reuse memory
+	// and saves allocation
+	TablesBufferPool TablesBufferPool
```
cfzjywxk, line 1624: Better to comment about the lifetime of the reused-buffer or how to use it, like staging, statement or the whole session.

Since it's not related to a session variable, could we put it somewhere else considering its lifetime?

pkg/table/tables/tables.go
```diff
@@ -581,18 +595,18 @@ func (t *TableCommon) UpdateRecord(ctx context.Context, sctx table.MutateContext
...
+		buffer.RowToCheck = append(buffer.RowToCheck, value)
 		if shouldWriteBinlog(sctx.GetSessionVars(), t.meta) && !t.canSkipUpdateBinlog(col, value) {
 			binlogColIDs = append(binlogColIDs, col.ID)
 			binlogOldRow = append(binlogOldRow, oldData[col.Offset])
 			binlogNewRow = append(binlogNewRow, value)
 		}
 	}
 	// check data constraint
-	err = t.CheckRowConstraint(sctx, rowToCheck)
+	err = t.CheckRowConstraint(sctx, buffer.RowToCheck)
```
cfzjywxk, line 609: Need to be cautious these slices are not referenced after the `UpdateRecord` call, a possible constraint is to abstarct methods explicitly like:
```
// Only use this when the slice is referenced locally and would NOT be used latter.
func GetRowToCheck() 
// Use this if the slices would be referenced later.
func GetRowToCheckCloned()
```
and comments about the usages.

Using the buffered slice fields directly could make the future maintenance complicated and error-prone.

pkg/sessionctx/variable/session.go
ekexium: By convention such data structures are directly put in `SessionVars`, like
https://github.com/pingcap/tidb/blob/ec6f56e82609659edb30dc309d2eb7615eb0f7b8/pkg/sessionctx/variable/session.go#L1324
https://github.com/pingcap/tidb/blob/ec6f56e82609659edb30dc309d2eb7615eb0f7b8/pkg/sessionctx/variable/session.go#L1198
https://github.com/pingcap/tidb/blob/ec6f56e82609659edb30dc309d2eb7615eb0f7b8/pkg/sessionctx/variable/session.go#L1490

pkg/table/tables/tables.go
ekexium: But there is no need for `GetRowToCheckCloned()`. All these slices are temporary variables. Even if a slice is obtained from `GetRowToCheck`, it still can be mistakenly accessed outside its expected lifetime. 
The best option I could think of is to use a clear name that everyone would understand this is a temporary slice.

cfzjywxk: > use a clear name that everyone would understand

Using clearer naming is also feasible. If there are no better hard constraint methods, I believe we need to add comments and warnings in all places where these buffered slices are used, to avoid additionally referencing the input slice and to highlight the related lifetime risks. 

This is actually the most difficult part (or the limitation of Golang). The way mentioned above using functions is essentially an attempt to achieve this goal as much as possible.

pkg/sessionctx/variable/session.go
cfzjywxk: The `SessionVars` seems very bloated, which might not be a good design. We can leave the new fields here for now. Refactoring this part might be a matter for another PR.

ekexium: +1. That should be part of the session refactoring work.

pkg/table/tables/tables.go
```diff
@@ -512,20 +512,38 @@ func (t *TableCommon) UpdateRecord(ctx context.Context, sctx table.MutateContext
...
+		buffer.ColIDs = make([]int64, 0, numColsCap)
+	} else {
+		buffer.ColIDs = buffer.ColIDs[:0]
+	}
+
+	if cap(buffer.Row) < numColsCap {
+		buffer.Row = make([]types.Datum, 0, numColsCap)
+	} else {
+		buffer.Row = buffer.Row[:0]
+	}
```
zyguan, line 534: How about extract a function for these boilerplate code. eg:
```go
func resizeSlice[T any](slice []T, size int, optCap ...int) []T {
	capacity := size
	if len(optCap) > 0 {
		capacity = optCap[0]
	}
	if cap(slice) < capacity {
		return make([]T, size, capacity)
	} else {
		return slice[:size]
	}
}

buffer.Row = resizeSlice(buffer.Row, 0, numColsCap)
```

pkg/sessionctx/variable/session.go
```diff
@@ -1618,6 +1618,40 @@ type SessionVars struct {
...
+	// It's users duty to reset them before use.
+	TablesBufferPool TablesBufferPool
+}
+
+// TablesBufferPool is a memory pool for table related memory allocation that aims to reuse memory
+// and saves allocation
+type TablesBufferPool struct {
+	Update *UpdatePool
+	Add    *AddPool
+	Remove *RemovePool
```
zyguan, line 1634: Just a question, why use pointer types here? Besides, these pools share some common fields (`[]int64`, `[]Datum`, `[]ColSize`), is it possible/safe to use a single union pool?

ekexium: > Just a question, why use pointer types here?

We use pointer types because we need to modify the fields in each pool. Without pointers, we would have to manually set the fields after usage, which would be less efficient and more error-prone.

> Besides, these pools share some common fields ([]int64, []Datum, []ColSize), is it possible/safe to use a single union pool?

Theoretically possible I think. But I don't see obvious benefits. The common slices are more of a coincidence rather than an intended design. Therefore, maintaining separate pools ensures clarity and avoids potential complications.

pkg/table/tables/tables.go
ekexium: Good idea. We may not need the generalization of `size` and `optCap`

zyguan (approved):

cfzjywxk (approved):

cfzjywxk:
@ekexium 
Need approvements from
-  [pkg/lightning/OWNERS](https://github.com/pingcap/tidb/blob/master/pkg/lightning/OWNERS)
-  [pkg/sessionctx/variable/OWNERS](https://github.com/pingcap/tidb/blob/master/pkg/sessionctx/variable/OWNERS)

ekexium:
@glorv FYI, this PR rewrites your previous optimization for lightning. cc @lance6716

lance6716 (approved):

pkg/table/tables/tables.go
```diff
@@ -881,6 +866,7 @@ func (t *TableCommon) AddRecord(sctx table.MutateContext, r []types.Datum, opts
 		return nil, err
 	}
 
+	// TODO: optimize the allocation (and calculation) of opt.
```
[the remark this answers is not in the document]
ekexium: Yes. IIRC allocating the opt takes some time.

pkg/sessionctx/variable/session.go
```diff
@@ -1618,6 +1618,40 @@ type SessionVars struct {
 
 	// GroupConcatMaxLen represents the maximum length of the result of GROUP_CONCAT.
 	GroupConcatMaxLen uint64
+
+	// TablesBufferPool is a memory pool for table related memory allocation that aims to reuse memory
+	// and saves allocation
+	// The buffers are supposed to be used inside AddRecord/UpdateRecord/RemoveRecord.
+	// It's users duty to reset them before use.
+	TablesBufferPool TablesBufferPool
```
lcwangchao, line 1626: How about just add a new method in `table.MutateContext` without modify `SessionVars`: https://github.com/pingcap/tidb/blob/1a24c032126dce4a79ea14b51108dbf4caf02a03/pkg/table/context/table.go#L32C6-L32C19

```
type MutateContext interface {
	// ...
	GetTablesBufferPool() * TablesBufferPool
	// ...
}

lcwangchao: You can implement this method here:

https://github.com/pingcap/tidb/blob/master/pkg/table/contextimpl/table.go

PR #53860, opened 2024-06-06
planner: improve range underestimation

### LINKED ISSUE #53907: Range estimate not to go below equivalent point get estimate
## Enhancement
Range predicate estimate can be much lower than an equivalent point get estimate if the high/low range of a histogram covers a broader range of values. This is because range estimation uses an interpolation formula whereas point get estimates will default to 1/NDV. These should be relatively consistent.

Example below shows the same query written as between vs an IN predicate. The example used comes from mysql_test/t/select_all.test. And any predicate values can be used where the histograms show a large range.

```
tidb> explain select fld1 from t2 where fld1 between 228313 and 228314;
+------------------------+---------+-----------+----------------------------+-----------------------------------------+
| id                     | estRows | task      | access object              | operator info                           |
+------------------------+---------+-----------+----------------------------+-----------------------------------------+
| IndexReader_6          | 0.05    | root      |                            | index:IndexRangeScan_5                  |
| └─IndexRangeScan_5    | 0.05    | cop[tikv] | table:t2, index:fld1(fld1) | range:[228313,228314], keep order:false |
+------------------------+---------+-----------+----------------------------+-----------------------------------------+

tidb> explain select fld1 from t2 where fld1 in (228313, 228314);
+-------------------+---------+------+----------------------------+------------------------------+
| id                | estRows | task | access object              | operator info                |
+-------------------+---------+------+----------------------------+------------------------------+
| Batch_Point_Get_1 | 2.00    | root | table:t2, index:fld1(fld1) | keep order:false, desc:false |
+-------------------+---------+------+----------------------------+------------------------------+

```

terry1purcell:
Issue Number: close #53907 

Problem Summary:

Range predicate estimation is changed to default to a minimum of 1/NDV for each of the low & high values (thus 2/NDV) which is consistent with the estimation used for 2 point-get values.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/statistics/histogram.go
```diff
@@ -593,14 +593,16 @@ func (hg *Histogram) LessRowCount(sctx context.PlanContext, value types.Datum) f
 func (hg *Histogram) BetweenRowCount(sctx context.PlanContext, a, b types.Datum) float64 {
 	lessCountA := hg.LessRowCount(sctx, a)
 	lessCountB := hg.LessRowCount(sctx, b)
-	// If lessCountA is not less than lessCountB, it may be that they fall to the same bucket and we cannot estimate
-	// the fraction, so we use `totalCount / NDV` to estimate the row count, but the result should not greater than
-	// lessCountB or notNullCount-lessCountA.
-	if lessCountA >= lessCountB && hg.NDV > 0 {
+	rangeEst := lessCountB - lessCountA
+	aEqual, _ := hg.EqualRowCount(sctx, a, false)
+	bEqual, _ := hg.EqualRowCount(sctx, b, false)
```
qw4990, line 598: Not sure whether I understand this correctly. Can we use `bEqual` here? I see the comment of this function, it returns the number of row count where the column is LESS than value.

winoros, line 598: Before we enter the `BetweenRowCount`, we'll modify the `Range` struct to ensure that the passed range here is always in the form `[a, x)`.

If the given filter is `a >= 1 and a <= 3` => the range is [1, 3], the range passed to the `BetweenRowCount` will be something like `[1, 4)`.

So directly using `EqualRowCount(a)` and `EqualRowCount(b)` may also lead to wrong estimation.

So my suggestion is that:
- If the Range.Lower.Exclude is false and the `Lower` itself is the upper bound of one bucket, use `Bucket.Repeat` which is the occurrence of the value.
- For any other cases for the `Range.Lower`, use `notNullCount()/NDV`.
- For `Range.Upper`, it's really hard to do the equal row count estimation since it's always excluded from the range we do the estimation. I actually have no good idea about this part, just suggest using the `notNullCount()/NDV`.

terry1purcell: This comment is encompassed by the next comment. I will resolve both comments with one solution as proposed below.

terry1purcell: If range lower is false - the call for equals estimation will already return notNullCount()/NDV. For Range.Upper - as per the review comment here - I have replaced it with notNullCount()/NDV.

winoros (approved):

qw4990 (approved):

Issue #41598, opened 2023-02-20
optimizer cannot handle cases like ((a = 1 and b = 2 and c > 3) or (a = 4 and b = 5 and c > 6)) and d > 3


xuyifangreeneyes:
## Enhancement

```
mysql> create table t1(a int, b int, c int, d int, index idx(a, b, c));
Query OK, 0 rows affected (0.06 sec)

mysql> explain select * from t1 where (a = 1 and b = 2 and c > 3) or (a = 4 and b = 5 and c > 6)
    -> ;
+-------------------------------+---------+-----------+------------------------------+--------------------------------------------------------------------------+
| id                            | estRows | task      | access object                | operator info                                                            |
+-------------------------------+---------+-----------+------------------------------+--------------------------------------------------------------------------+
| IndexLookUp_10                | 0.67    | root      |                              |                                                                          |
| ├─IndexRangeScan_8(Build)     | 0.67    | cop[tikv] | table:t1, index:idx(a, b, c) | range:(1 2 3,1 2 +inf], (4 5 6,4 5 +inf], keep order:false, stats:pseudo |
| └─TableRowIDScan_9(Probe)     | 0.67    | cop[tikv] | table:t1                     | keep order:false, stats:pseudo                                           |
+-------------------------------+---------+-----------+------------------------------+--------------------------------------------------------------------------+
3 rows in set (0.00 sec)

mysql> explain select * from t1 where ((a = 1 and b = 2 and c > 3) or (a = 4 and b = 5 and c > 6)) and d > 3;
+----------------------------+---------+-----------+------------------------------+------------------------------------------------------------------------------------------------------------------------------------+
| id                         | estRows | task      | access object                | operator info                                                                                                                      |
+----------------------------+---------+-----------+------------------------------+------------------------------------------------------------------------------------------------------------------------------------+
| IndexLookUp_12             | 0.22    | root      |                              |                                                                                                                                    |
| ├─Selection_10(Build)      | 0.22    | cop[tikv] |                              | or(and(eq(test.t1.a, 1), and(eq(test.t1.b, 2), gt(test.t1.c, 3))), and(eq(test.t1.a, 4), and(eq(test.t1.b, 5), gt(test.t1.c, 6)))) |
| │ └─IndexRangeScan_8       | 20.00   | cop[tikv] | table:t1, index:idx(a, b, c) | range:[1,1], [4,4], keep order:false, stats:pseudo                                                                                 |
[...]

PR #53908, opened 2024-06-10
Optimizer: Fix range extraction for CNF(conjunctive normal form) | tidb-test=pr/2341

ghazalfamilyusa:

Issue Number: Ref #41598

### Problem Summary:
Index range extraction in the optimizer can handle disjunctions like 
((a = 1 and b = 2 and c > 3) or (a = 4 and b = 5 and c > 6)) and builds the proper index range like "(1 2 3,1 2 +inf], (4 5 6,4 5 +inf]" for this example. The problem is that when we add another conjunct like d > 3 then the optimizer takes another path which find only the point ranges "[1,1], [4,4]" in this example. Finding point ranges is OK for complex CNF where we only pick ranges from only one conjunct, But, in this case, the best conjunct is also better than point ranges. 

 We added code that picks the best conjunct ranges if it is more selective than the point ranges. For example, "(1 2 3,1 2 +inf], (4 5 6,4 5 +inf]" is more selective than "[1,1], [4,4]" . We implemented that by checking if the best CNF ranges is a subset of the point ranges. 

### Check List

Tests 

- [X] Unit test
- [] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/util/ranger/detacher.go
+++ b/pkg/util/ranger/detacher.go
@@ -463,6 +463,9 @@ func (d *rangeDetacher) detachCNFCondAndBuildRangeForIndex(conditions []expressi
 		// TODO: we will optimize it later.
 		res.RemainedConds = AppendConditionsIfNotExist(res.RemainedConds, remainedConds)
 		res.Ranges = ranges
+		// Choosing between point ranges and bestCNF is needed since bestCNF does not cover the intersection
+		// of all conjuncts. Even when we add support for intersection, it could be turned off by a flag or it could be
+		// incomplete due to a long list of conjuncts.
 		if bestCNFItemRes != nil && res != nil && len(res.Ranges) != 0 {
 			bestCNFIsSubset := bestCNFItemRes.rangeResult.Ranges.Subset(d.sctx.TypeCtx, res.Ranges)
 			pointRangeIsSubset := res.Ranges.Subset(d.sctx.TypeCtx, bestCNFItemRes.rangeResult.Ranges)
--- a/pkg/util/ranger/detacher.go
+++ b/pkg/util/ranger/detacher.go
@@ -277,6 +277,7 @@ func extractBestCNFItemRanges(sctx *rangerctx.RangerContext, conds []expression.
 			bestRes = curRes
 		}
 	}
+
 	if bestRes != nil && bestRes.rangeResult != nil {
 		bestRes.rangeResult.IsDNFCond = false
 	}
@@ -462,6 +463,18 @@ func (d *rangeDetacher) detachCNFCondAndBuildRangeForIndex(conditions []expressi
 		// TODO: we will optimize it later.
 		res.RemainedConds = AppendConditionsIfNotExist(res.RemainedConds, remainedConds)
 		res.Ranges = ranges
+		if bestCNFItemRes != nil {
+			bestCNFIsSubset := bestCNFItemRes.rangeResult.Ranges.Subset(d.sctx.TypeCtx, res.Ranges)
+			pointRangeIsSubset := res.Ranges.Subset(d.sctx.TypeCtx, bestCNFItemRes.rangeResult.Ranges)
+			// Pick bestCNFIsSubset if it is more selective than point ranges(res).
+			// Only optimization if it is a proper subset bestCNFIsSubset and !pointRangeIsSubset.
+			if bestCNFIsSubset && !pointRangeIsSubset {
+				// Update final result and just update: Ranges, AccessConds and RemainedConds
+				res.RemainedConds = removeConditions(res.RemainedConds, bestCNFItemRes.rangeResult.AccessConds)
+				res.Ranges = bestCNFItemRes.rangeResult.Ranges
+				res.AccessConds = bestCNFItemRes.rangeResult.AccessConds
+			}
+		}
 		return res, nil
 	}
 	for _, cond := range newConditions {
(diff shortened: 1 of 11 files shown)


zanmato1984 (approved):

ghazalfamilyusa:
cc @XuHuaiyu for expression changes

ghazalfamilyusa:
cc @qw4990

elsa0520:
> ### What problem does this PR solve?
> Issue Number: Ref #41598
> 
> ### Problem Summary:
> Index range extraction in the optimizer can handle disjunctions like ((a = 1 and b = 2 and c > 3) or (a = 4 and b = 5 and c > 6)) and builds the proper index range like "(1 2 3,1 2 +inf], (4 5 6,4 5 +inf]" for this example. The problem is that when we add another conjunct like d > 3 then the optimizer takes another path which find only the point ranges "[1,1], [4,4]" in this example. Finding point ranges is OK for complex CNF where we only pick ranges from only one conjunct, But, in this case, the best conjunct is also better than point ranges.
> 
> ### What changed and how does it work?
> We added code that picks the best conjunct ranges if it is more selective than the point ranges. For example, "(1 2 3,1 2 +inf], (4 5 6,4 5 +inf]" is more selective than "[1,1], [4,4]" . We implemented that by checking if the best CNF ranges is a subset of the point ranges.
> 
> ### Check List
> Tests
> 
> * [x]  Unit test
> * [] Integration test
> * [ ]  Manual test (add detailed scripts or steps below)
> * [ ]  No need to test
>   > * [ ]  I checked and no code files have been changed.
> 
> Side effects
> 
> * [ ]  Performance regression: Consumes more CPU
> * [ ]  Performance regression: Consumes more Memory
> * [ ]  Breaking backward compatibility
> 
> Documentation
> 
> * [ ]  Affects user behaviors
> * [ ]  Contains syntax changes
> * [ ]  Contains variable changes
> * [ ]  Contains experimental features
> * [ ]  Changes MySQL compatibility
> 
> ### Release note
> Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.
> 
> ```
> None
> ```

What's different between with ```and d>3``` and without ```and d>3``` ? 
Whatever with or without ``` and d>3``` , the range of DNF ```((a = 1 and b = 2 and c > 3) or (a = 4 and b = 5 and c > 6)) ``` should be  "(1 2 3,1 2 +inf], (4 5 6,4 5 +inf]".
Why the range of DNF changed to point range when there is additional predicate ```and d>3``` ?

ghazalfamilyusa:
> What's different between with `and d>3` and without `and d>3` ? Whatever with or without ` and d>3` , the range of DNF `((a = 1 and b = 2 and c > 3) or (a = 4 and b = 5 and c > 6)) ` should be "(1 2 3,1 2 +inf], (4 5 6,4 5 +inf]". Why the range of DNF changed to point range when there is additional predicate `and d>3` ?
------------
@elsa0520 ; this is what I talked about in the design document and in the summary above. The code will take the point ranges path when you have multiple conjuncts regardless if these conjuncts are relevant or not like (d > 3).

XuHuaiyu (approved):
The expression change LGTM

elsa0520:
> > What's different between with `and d>3` and without `and d>3` ? Whatever with or without ` and d>3` , the range of DNF `((a = 1 and b = 2 and c > 3) or (a = 4 and b = 5 and c > 6)) ` should be "(1 2 3,1 2 +inf], (4 5 6,4 5 +inf]". Why the range of DNF changed to point range when there is additional predicate `and d>3` ?
> 
> @elsa0520 ; this is what I talked about in the design document and in the summary above. The code will take the point ranges path when you have multiple conjuncts regardless if these conjuncts are relevant or not like (d > 3).

But in the design we will deal with the CNF intersection. So after you support Computing intersection for CNF . We can merge the ((a = 1 and b = 2 and c > 3) or (a = 4 and b = 5 and c > 6)) and d>3 . Then we don't need to support PR alone, isn't it?

ghazalfamilyusa:
> > > What's different between with `and d>3` and without `and d>3` ? Whatever with or without ` and d>3` , the range of DNF `((a = 1 and b = 2 and c > 3) or (a = 4 and b = 5 and c > 6)) ` should be "(1 2 3,1 2 +inf], (4 5 6,4 5 +inf]". Why the range of DNF changed to point range when there is additional predicate `and d>3` ?
> > 
> > 
> > @elsa0520 ; this is what I talked about in the design document and in the summary above. The code will take the point ranges path when you have multiple conjuncts regardless if these conjuncts are relevant or not like (d > 3).
> 
> But in the design we will deal with the CNF intersection. So after you support Computing intersection for CNF . We can merge the ((a = 1 and b = 2 and c > 3) or (a = 4 and b = 5 and c > 6)) and d>3 . Then we don't need to support PR alone, isn't it?

@elsa0520 : seems my previous reply is not clear. The intersection is not 100% guaranteed at least for a while. Initially, we will have a flag and later on we may implement a heuristic solution. So, we need the point ranges in case the flag is off or if the intersection is not done for all the conjuncts.  Now, with the reality of having point ranges (just in case) we still need to choose bestCNF over range. we can remove point ranges support and related PRs once we have: 100% intersection applied with no flag.  Hope this makes sense and if not we can chat

pkg/util/ranger/detacher.go
```diff
@@ -462,6 +463,18 @@ func (d *rangeDetacher) detachCNFCondAndBuildRangeForIndex(conditions []expressi
 		// TODO: we will optimize it later.
 		res.RemainedConds = AppendConditionsIfNotExist(res.RemainedConds, remainedConds)
 		res.Ranges = ranges
+		if bestCNFItemRes != nil && res != nil && len(res.Ranges) != 0 {
```
elsa0520, line 466: Could you please comment this case " (heuristics applied for long lists or we turn off the intersection)" in front of here?

ghazalfamilyusa: done

elsa0520 (approved):
LGTM

Issue #53677, opened 2024-05-30
Support merge adding unique and non-unique index into one pass in multi-schema change


lance6716:
## Enhancement

as title

PR #53632, opened 2024-05-28
ddl: unify merging unique and non-unique index for multi-schema change

lance6716:
### What problem does this PR solve?

Issue Number: close https://github.com/pingcap/tidb/issues/53677

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #53975, opened 2024-06-12
ddl: record the used indexID for global-sort reorg

lance6716:

Issue Number: ref #53677

Problem Summary:

This is a follow-up of PR https://github.com/pingcap/tidb/pull/53632 to resolve a TODO. Before this PR, we don't know which index does a ingest subtask belong, so we can't generate the index name of "duplicate entry" error.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/backfilling_import_cloud.go
+++ b/pkg/ddl/backfilling_import_cloud.go
@@ -78,10 +78,31 @@ func (m *cloudImportExecutor) RunSubtask(ctx context.Context, subtask *proto.Sub
 	if local == nil {
 		return errors.Errorf("local backend not found")
 	}
-	// TODO(lance6716): find the correct index, so "duplicate entry" error can have index name
-	currentIdx := m.indexes[0]
 
-	_, engineUUID := backend.MakeUUID(m.ptbl.Meta().Name.L, currentIdx.ID)
+	var (
+		currentIdx *model.IndexInfo
+		idxID      int64
+	)
+	switch len(sm.EleIDs) {
+	case 1:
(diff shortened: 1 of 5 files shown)


pkg/ddl/backfilling_import_cloud.go
```diff
@@ -78,10 +78,31 @@ func (m *cloudImportExecutor) RunSubtask(ctx context.Context, subtask *proto.Sub
...
+	switch len(sm.EleIDs) {
+	case 1:
+		for _, idx := range m.indexes {
+			if idx.ID == sm.EleIDs[0] {
+				currentIdx = idx
+				idxID = idx.ID
+				break
+			}
+		}
+	case 0:
```
D3Hunter, line 95: maybe we should write in doc that we don't support using global sort during upgrade, so we don't need to maintain this in-compatibility

D3Hunter (commented):
rest lgtm

pkg/ddl/backfilling_import_cloud.go
Frank945946: From a user perspective, it would be best if we could accommodate this case and not block users from upgrading. Otherwise, a lot of restrictions would hinder the promotion and usage of this feature. Therefore, I suggest we optimize internally and avoid blocking upgrades.

D3Hunter (approved):

tangenta (approved):

Issue #52795, opened 2024-04-22
some schema level DDLs are not considered in dependency check


D3Hunter:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

currently we only consider `DropSchema` for schema level check, but not for `RecoverSchema`, and there are `ModifySchemaCharsetAndCollate` and `ModifySchemaDefaultPlacement` that might need to consider too.
https://github.com/pingcap/tidb/blob/79c1499bec7e609a051634c87460511b203c5389/pkg/ddl/job_table.go#L232-L236

### 4. What is your TiDB version? (Required)
master

lance6716:
I suspect `runningJobs` already use `unfinishedSchema` to filter them. How about we remove these SQL based check? I can open a PR if you agree with it @wjhuang2016 @tangenta

D3Hunter:
> I suspect `runningJobs` already use `unfinishedSchema` to filter them. How about we remove these SQL based check? I can open a PR if you agree with it @wjhuang2016 @tangenta

i record it in the doc, the SQL based check will be removed

PR #54003, opened 2024-06-13
ddl: only use runningJobs to check job dependency

lance6716:

Issue Number: ref #52795, #53246

Problem Summary:

assuming there's one DDL owner at any time, we don't need to query persistent data of running DDL jobs. Instead, we can check the memory cache of running DDL jobs.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/ddl/ddl_running_jobs.go
```diff
@@ -19,119 +19,158 @@
...
+		if info.Database == model.InvolvingNone {
+			// should not happen
+			if intest.InTest {
+				if info.Table != model.InvolvingNone {
+					panic(fmt.Sprintf(
+						"job %d is invalid, involved table name is not empty: %s",
+						jobID, info.Table,
+					))
+				}
+			}
```
tangenta, line 74: There are some invalid jobs in tests(like `doDDLJobErr`). Maybe you can search `&model.Job{` and fix all usages.

lance6716: I only check for db is None but table is not None. Not found in CI I think.

pkg/ddl/ddl_api.go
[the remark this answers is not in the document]
lance6716: For placement policy and resource group, do you think we can add another field for `InvolvingSchemaInfo` to record the dependency or other implementation? also cc @tangenta 

I'll consider FK in next commit.

pkg/ddl/ddl_api.go
```diff
@@ -7949,7 +7969,17 @@ func (d *ddl) CreateForeignKey(ctx sessionctx.Context, ti ast.Ident, fkName mode
 		BinlogInfo:     &model.HistoryInfo{},
 		Args:           []any{fkInfo, fkCheck},
 		CDCWriteSource: ctx.GetSessionVars().CDCWriteSource,
-		SQLMode:        ctx.GetSessionVars().SQLMode,
+		InvolvingSchemaInfo: []model.InvolvingSchemaInfo{
```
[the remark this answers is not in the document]
lance6716: I see that in the execution code `dropForeignKey`, it does not check the existence of RefSchema / RefTable, maybe in this way it's more rubust. 

@crazycs520 Do you think we should check dependency for the reference table of drop foreign key?

pkg/ddl/ddl_api.go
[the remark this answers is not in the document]
D3Hunter: now placement policy and resource group will set its name to job.Schema, if they have same namespace as database, seems ok to set `InvolvingSchemaInfo = xx.*`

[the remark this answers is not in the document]
lance6716: Just in case there's a database with same name. In meta KV policy has the prefix of `mPolicyPrefix` which is different with schema.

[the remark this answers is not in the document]
D3Hunter: placement policy and resource group related ddl all have a `none, none` InvolvingSchemaInfo😓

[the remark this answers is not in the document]
tangenta: Adding another field maybe better. It is weird if a policy job blocks a create-database job..

pkg/ddl/multi_schema_change.go
```diff
@@ -34,19 +38,52 @@ func (d *ddl) MultiSchemaChange(ctx sessionctx.Context, ti ast.Ident, info *mode
...
+			}
+
+			if len(subjob.Args) < 1 {
+				ddllogutil.DDLLogger().Warn(
+					"expected add foreign key subjob args length",
+					zap.Any("args", subjob.Args))
+				continue
+			}
+			fkInfo, ok := subjob.Args[0].(*model.FKInfo)
+			if !ok {
```
tangenta, line 60: How about using `intest` and panic?

pkg/ddl/ddl_running_jobs.go
```diff
@@ -19,122 +19,158 @@
...
-			delete(j.unfinishedSchema, info.Database)
+		if len(j.schemas[info.Database]) == 0 {
+			delete(j.schemas, info.Database)
 		}
 	}
 }
 
+func (j *runningJobs) updateIDsStrGetter() {
+	var (
+		once   sync.Once
```
tangenta, line 110: `j.idsStrGetter()` has been protected by `j.RLock()`, it is still necessary to use `sync.Once`? I mean why not:

```
if idStr == nil {
    // construct idstr
}
return idsStr
```

pkg/ddl/multi_schema_change.go
lance6716: Not sure we have a upgrading integration test and will see the old version DDL jobs while `intest` is true.

pkg/ddl/ddl_running_jobs.go
lance6716: RLock will let multiple lock holders write to idsStr, which is a data race

pkg/ddl/multi_schema_change.go
tangenta: OK. Anyway we should always turn on `intest` for all integration tests.

D3Hunter (approved):

lance6716:
/hold

adding some code

pkg/ddl/ddl_api.go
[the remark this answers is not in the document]
lance6716: I found it's a bit complex and I only want to add a TODO in this PR. Because there're exclusive jobs and shared jobs. Like CreatePlacementPolicy that directly modifies PlacementPolicy, and AlterTablePlacement that referring PlacementPolicy. Two AlterTablePlacement jobs should not have dependency.

And I think CREATE TABLE LIKE and FK are simliar. I want to leave this PR to this situation: 

- When two jobs only referring to an object, in other words, does not modify it, the jobs have no dependency
- A referring job and a modifying job have no dependency. (I'll add TODO and improve it in next PR)
- Two modifying jobs have dependency.

I'll commit soon

pkg/ddl/ddl_api.go
```diff
@@ -2829,31 +2829,16 @@ func (d *ddl) createTableWithInfoJob(
 		args = append(args, ctx.GetSessionVars().ForeignKeyChecks)
 	}
 
-	involvingSchemaInfo := make([]model.InvolvingSchemaInfo, 0, len(tbInfo.ForeignKeys)+1)
```
D3Hunter: why it's removed? and below FK for multi schema change

lance6716: I want to allow CREATE TABLE t1 FOREIGN KEY t0.c0 and CREATE TABLE t2 FOREIGN KEY t0.c0 running concurrently in this PR. See below.

https://github.com/pingcap/tidb/pull/54003#discussion_r1644067821

lance6716: Oh, CI is failed because of this...

TestTruncateOrDropTableWithForeignKeyReferred2

pkg/ddl/ddl_api.go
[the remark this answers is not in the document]
D3Hunter: > When two jobs only referring to an object, in other words, does not modify it, the jobs have no dependency

we should take them as depends on each other now as when calculate dependency we don't consider skipped jobs now, suppose this
- job 1: create table a ...foreign key refer test.t
- job 2: alter table test.t
- job 3: create table b ...foreign key refer test.t
job 1 can run, but not job 2, when we check job 3, we need consider job 2, not run it directly.

> A referring job and a modifying job to the same object have no dependency.

do you mean `job 1 create table a foreign key test.t; job 2 alter table test.t` have no dependency? i think it does if job 2 modify the referred column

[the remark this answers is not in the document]
lance6716: > job 1: create table a ...foreign key refer test.t
> job 2: alter table test.t
> job 3: create table b ...foreign key refer test.t
> job 1 can run, but not job 2, when we check job 3, we need consider job 2, not run it directly.

I think this is the case of fair lock. I'll keep this property in next PR

> do you mean job 1 create table a foreign key test.t; job 2 alter table test.t have no dependency? i think it does if job 2 modify the referred column

It should have dependency, but I also want to leave it for next PR. In next PR I'll add `exclusive` and `shared` fields to `runningJobs` and `InvolvingSchemaInfo` to implement it.

tangenta (approved):

Issue #54022, opened 2024-06-14
TiDB memory leak after meets the `fail to get stats version for this histogram`


hi-rustin:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

1. Change the async load stats code:
```diff
diff --git a/pkg/sessionctx/variable/tidb_vars.go b/pkg/sessionctx/variable/tidb_vars.go
index e10aa9c5bbb9f..39153b4433448 100644
--- a/pkg/sessionctx/variable/tidb_vars.go
+++ b/pkg/sessionctx/variable/tidb_vars.go
@@ -1287,7 +1287,7 @@ const (
 	DefTiDBTableCacheLease                         = 3 // 3s
 	DefTiDBPersistAnalyzeOptions                   = true
 	DefTiDBEnableColumnTracking                    = false
-	DefTiDBStatsLoadSyncWait                       = 100
+	DefTiDBStatsLoadSyncWait                       = 0
 	DefTiDBStatsLoadPseudoTimeout                  = true
 	DefSysdateIsNow                                = false
 	DefTiDBEnableMutationChecker                   = false
diff --git a/pkg/statistics/handle/storage/read.go b/pkg/statistics/handle/storage/read.go
index bacf43abef60f..e1f43e8ecdc65 100644
--- a/pkg/statistics/handle/storage/read.go
+++ b/pkg/statistics/handle/storage/read.go
@@ -539,7 +539,7 @@ func loadNeededColumnHistograms(sctx sessionctx.Context, statsCache util.StatsCa
 	if err != nil {
 		return errors.Trace(err)
 	}
-	if len(rows) == 0 {
+	if true {
 		logutil.BgLogger().Error("fail to get stats version for this histogram", zap.Int64("table_id", col.TableID), zap.Int64("hist_id", col.ID))
 		return errors.Trace(fmt.Errorf("fail to get stats version for this histogram, table_id:%v, hist_id:%v", col.TableID, col.ID))
 	}
@@ -599,7 +599,7 @@ func loadNeededIndexHistograms(sctx sessionctx.Context, statsCache util.StatsCac
 	if err != nil {
 		return errors.Trace(err)
 	}
-	if len(rows) == 0 {
+	if true {
 		logutil.BgLogger().Error("fail to get stats version for this histogram", zap.Int64("table_id", idx.TableID), zap.Int64("hist_id", idx.ID))
[...]

PR #54060, opened 2024-06-17
statistics: do not load unnecessary index statistics

hi-rustin:

Issue Number: close https://github.com/pingcap/tidb/issues/54022

Problem Summary:

See more at the issue. 

A quick summary:
If you have `tidb_opt_objective='determine'` enabled, it will trigger some async stats load for some tables. If there tables histogram isn't in the system, it would cause some problems when we try to load it.

Check that the index statistics are really needed before loading them.

If we don't check it here, it will cause some problems if the index statistics don't exist.

**Note:** We still don't know why the histogram record for the index is missing. There are some possible issues:

1. When users create this table, they may not have created this index yet, so we don't create the record for it automatically. (But if this is the case, I don't why do we have the index in the stats cache information)
2. It may have been accidentally dropped, but I can't find any scenario where we would only drop one index for a table.

What happened in the test case:
1. We create a table and the `statsHandle.Update(do.InfoSchema())` will load this table into the stats cache. You can test it by breakpoint here:
image
3. We drop the stats of the stats, it will clean up all system table records for this table.
4. We insert some data and wait for the modfiy_count and the count is not null in the `mysql.stats_meta`.
5. Try to select some data from this table by ID, it would trigger an async load. Then we get that bug.
https://github.com/hi-rustin/tidb/blob/648fc6386aa9aabb365baf0912f1e18c4ab74d95/pkg/planner/core/logical_plan_builder.go#L4824
6. So in the test case we check if the index is still in the needed item array after we fix it.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [x] [Manual test](https://github.com/pingcap/tidb/pull/54060#issuecomment-2174949286)
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix a memory leak when loading index statistics.
修复加载 index 统计信息可能会造成内存泄漏的问题
```

### DIFF
--- a/pkg/statistics/handle/storage/read.go
+++ b/pkg/statistics/handle/storage/read.go
@@ -576,7 +576,15 @@ func loadNeededIndexHistograms(sctx sessionctx.Context, statsCache util.StatsCac
 		return nil
 	}
 	index, ok := tbl.Indices[idx.ID]
-	if !ok {
+	// Double check if the index is really needed to load.
+	// See: https://github.com/pingcap/tidb/issues/54022
+	if !ok || !index.IsLoadNeeded() {
+		if !index.IsLoadNeeded() {
+			logutil.BgLogger().Warn(
+				"Although the index stats is not required to load, an attempt is still made to load it. Please verify if this index has any histogram records in the `mysql.stats_histograms` table.",
+				zap.Int64("table_id", idx.TableID), zap.Int64("hist_id", idx.ID),
+			)
+		}
 		statistics.HistogramNeededItems.Delete(idx)
 		return nil
 	}
--- a/pkg/statistics/handle/storage/read.go
+++ b/pkg/statistics/handle/storage/read.go
@@ -581,7 +581,7 @@ func loadNeededIndexHistograms(sctx sessionctx.Context, statsCache util.StatsCac
 	if !ok || !index.IsLoadNeeded() {
 		if !index.IsLoadNeeded() {
 			logutil.BgLogger().Warn(
-				"Although the index stats is not required to load, an attempt is still made to load it. Please verify if this index has any histogram records in the `mysql.stats_histograms` table.",
+				"Although the index stats is not required to load, an attempt is still made to load it.",
 				zap.Int64("table_id", idx.TableID), zap.Int64("hist_id", idx.ID),
 			)
 		}
--- a/pkg/statistics/handle/storage/read.go
+++ b/pkg/statistics/handle/storage/read.go
@@ -577,6 +577,7 @@ func loadNeededIndexHistograms(sctx sessionctx.Context, statsCache util.StatsCac
 	}
 	index, ok := tbl.Indices[idx.ID]
 	// Double check if the index is really needed to load.
+	// If we don't do this it might cause a memory leak.
 	// See: https://github.com/pingcap/tidb/issues/54022
 	if !ok || !index.IsLoadNeeded() {
 		if !index.IsLoadNeeded() {
--- a/pkg/statistics/handle/storage/read.go
+++ b/pkg/statistics/handle/storage/read.go
@@ -540,8 +540,8 @@ func loadNeededColumnHistograms(sctx sessionctx.Context, statsCache util.StatsCa
 		return errors.Trace(err)
 	}
 	if len(rows) == 0 {
-		logutil.BgLogger().Error("fail to get stats version for this histogram", zap.Int64("table_id", col.TableID), zap.Int64("hist_id", col.ID))
-		return errors.Trace(fmt.Errorf("fail to get stats version for this histogram, table_id:%v, hist_id:%v", col.TableID, col.ID))
+		logutil.BgLogger().Error("fail to get stats version for this histogram, normally this wouldn't happen, please check if this column or index has a histogram record in `mysql.stats_histogram`", zap.Int64("table_id", col.TableID), zap.Int64("column_id", col.ID))
+		return errors.Trace(fmt.Errorf("fail to get stats version for this histogram, normally this wouldn't happen, please check if this column or index has a histogram record in `mysql.stats_histogram`, table_id:%v, column_id:%v", col.TableID, col.ID))
 	}
 	statsVer := rows[0].GetInt64(0)
 	colHist := &statistics.Column{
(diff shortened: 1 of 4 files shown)


hi-rustin:
Tested locally:
After:
1. Start the cluster with patch: `tiup playground v7.5.1  --db.binpath /Volumes/t7/code/tidb/bin/tidb-server`
2. Create a table: `create table if not exists t(a int, b int, index ia(a));`
3. Drop the stats: `drop stats t;`
4. Insert some data: `insert into t value(1,1), (2,2);`
5. Wait for a while until we can see the stats: `show stats_meta`
```sql
mysql> show stats_meta
    -> ;
| test    | t          |                | 2024-06-18 12:07:12 |            0 |         0 |
1 row in set (0.01 sec)
```
6. Enable determinate: `set tidb_opt_objective='determinate';`
7. Try select data: `select * from t where a = 1 and b = 1;`
```sql
mysql> select * from t where a = 1 and b = 1;
+------+------+
| a    | b    |
+------+------+
|    1 |    1 |
+------+------+
1 row in set (0.01 sec)
```
8. Check logs:
```log
[2024/06/18 12:09:47.346 +08:00] [WARN] [read.go:583] ["Although the index stats is not required to load, an attempt is still made to load it."] [table_id=102] [hist_id=1]
```

Before:

hi-rustin (commented):

pkg/statistics/handle/storage/read.go
```diff
@@ -576,7 +576,15 @@ func loadNeededIndexHistograms(sctx sessionctx.Context, statsCache util.StatsCac
 		return nil
 	}
 	index, ok := tbl.Indices[idx.ID]
-	if !ok {
+	// Double check if the index is really needed to load.
```
elsa0520, line 579:
```suggestion
	// Double check if the index is really needed to load.
	// Sometimes, the stats_meta is missing which will cause the memory  leaks
```

elsa0520, line 579: Or can you print out the index names that do not exist in stats_histogram records? This will be handy for later troubleshooting.

hi-rustin: We already have this log in other places.

hi-rustin: Updated some logs.

hi-rustin (commented):

elsa0520 (approved):
LGTM

hawkingrei:
> Tested locally: After:
> 
>     1. Start the cluster with patch: `tiup playground v7.5.1  --db.binpath /Volumes/t7/code/tidb/bin/tidb-server`
> 
>     2. Create a table: `create table if not exists t(a int, b int, index ia(a));`
> 
>     3. Drop the stats: `drop stats t;`
> 
>     4. Insert some data: `insert into t value(1,1), (2,2);`
> 
>     5. Wait for a while until we can see the stats: `show stats_meta`
> 
> 
> ```sql
> mysql> show stats_meta
>     -> ;
> +---------+------------+----------------+---------------------+--------------+-----------+
> | Db_name | Table_name | Partition_name | Update_time         | Modify_count | Row_count |
> +---------+------------+----------------+---------------------+--------------+-----------+
> | test    | t          |                | 2024-06-18 12:07:12 |            0 |         0 |
> +---------+------------+----------------+---------------------+--------------+-----------+
> 1 row in set (0.01 sec)
> ```
> 
>     6. Enable determinate: `set tidb_opt_objective='determinate';`
> 
>     7. Try select data: `select * from t where a = 1 and b = 1;`
> 
> 
> ```sql
> mysql> select * from t where a = 1 and b = 1;
> +------+------+
> | a    | b    |
> +------+------+
> |    1 |    1 |
> +------+------+
> 1 row in set (0.01 sec)
> ```
> 
>     8. Check logs:
> 
> 
> ```
> [2024/06/18 12:09:47.346 +08:00] [WARN] [read.go:583] ["Although the index stats is not required to load, an attempt is still made to load it."] [table_id=102] [hist_id=1]
> ```
> 
> Before:
> 
>     1. Start the cluster with patch: `tiup playground v7.5.1  --db.binpath /Volumes/t7/code/tidb/bin/tidb-server`
> 
>     2. Create a table: `create table if not exists t(a int, b int, index ia(a));`
> 
>     3. Drop the stats: `drop stats t;`
> 
>     4. Insert some data: `insert into t value(1,1), (2,2);`
> 
>     5. Wait for a while until we can see the stats: `show stats_meta`
> 
> 
> ```sql
> mysql> show stats_meta
>     -> ;
> +---------+------------+----------------+---------------------+--------------+-----------+
> | Db_name | Table_name | Partition_name | Update_time         | Modify_count | Row_count |
> +---------+------------+----------------+---------------------+--------------+-----------+
> | test    | t          |                | 2024-06-18 12:07:12 |            0 |         0 |
> +---------+------------+----------------+---------------------+--------------+-----------+
> 1 row in set (0.01 sec)
> ```
> 
>     6. Enable determinate: `set tidb_opt_objective='determinate';`
> 
>     7. Try select data: `select * from t where a = 1 and b = 1;`
> 
> 
> ```sql
> mysql> select * from t where a = 1 and b = 1;
> +------+------+
> | a    | b    |
> +------+------+
> |    1 |    1 |
> +------+------+
> 1 row in set (0.01 sec)
> ```
> 
>     8. Check logs:
> 
> 
> ```
> [2024/06/18 12:15:01.729 +08:00] [ERROR] [read.go:603] ["fail to get stats version for this histogram"] [table_id=102] [hist_id=1]
> [2024/06/18 12:15:04.728 +08:00] [ERROR] [read.go:603] ["fail to get stats version for this histogram"] [table_id=102] [hist_id=1]
> ```

You can add this desciption into ```Manual test```.

tests/realtikvtest/statisticstest/statistics_test.go
```diff
@@ -232,3 +236,67 @@ func checkFMSketch(tk *testkit.TestKit) {
...
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test;")
+	tk.MustExec("drop table if exists t;")
+	tk.MustExec("create table if not exists t(a int, b int, index ia(a));")
+	tk.MustExec("drop stats t;")
+	tk.MustExec("insert into t value(1,1), (2,2);")
+	require.Eventually(t, func() bool {
+		rows := tk.MustQuery("show stats_meta").Rows()
+		return len(rows) > 0
+	}, 2*time.Minute, 5*time.Millisecond)
```
time-and-fate, line 251: We can use this to trigger it immediately.
```suggestion
	h := dom.StatsHandle()
	require.NoError(t, h.DumpStatsDeltaToKV(true))
	require.NoError(t, h.Update(dom.InfoSchema()))
```

tests/realtikvtest/statisticstest/statistics_test.go
```diff
@@ -232,3 +236,67 @@ func checkFMSketch(tk *testkit.TestKit) {
...
+					case <-ticker.C:
+						items := statistics.HistogramNeededItems.AllItems()
+						found := false
+						for _, item := range items {
+							if item.TableID == tableID {
+								found = true
+							}
+						}
+						if !found {
+							done <- true
```
time-and-fate, line 285: Looks like we should also `return` here?

tests/realtikvtest/statisticstest/statistics_test.go
hi-rustin: Nice! Updated. I forgot we can directly call an update here.

hi-rustin: I tested it locally. It seems we can't trigger the bug in this way.

hi-rustin: Updated.

time-and-fate (approved):

Issue #38911, opened 2022-11-04
Index lookup can be optimized in delete statement


tangenta:
## Enhancement

```sql
mysql> create table t (a varchar(100), b int, c int, primary key(a, b) nonclustered);
Query OK, 0 rows affected (0.17 sec)

mysql> explain delete from t where a in ('aaaaa', 'b');
+----------------------------------+---------+-----------+------------------------------+--------------------------------------------------------------------+
| id                               | estRows | task      | access object                | operator info                                                      |
+----------------------------------+---------+-----------+------------------------------+--------------------------------------------------------------------+
| Delete_4                         | N/A     | root      |                              | N/A                                                                |
| └─IndexLookUp_11                 | 20.00   | root      |                              |                                                                    |
|   ├─IndexRangeScan_9(Build)      | 20.00   | cop[tikv] | table:t, index:PRIMARY(a, b) | range:["aaaaa","aaaaa"], ["b","b"], keep order:false, stats:pseudo |
|   └─TableRowIDScan_10(Probe)     | 20.00   | cop[tikv] | table:t                      | keep order:false, stats:pseudo                                     |
+----------------------------------+---------+-----------+------------------------------+--------------------------------------------------------------------+
4 rows in set (0.01 sec)
```

For the delete operation, only `_tidb_rowid` and the value `a` and `b` are required. In this case, `IndexLookUp` can be eliminated to a `IndexRangeScan`.

fixdb:
Perhaps in the column pruning step, planner doesn't prune column for delete operator, hence causing the generation of IndexLookUp?

Defined2014:
In `buildDelete` func, we always set a schema which contained all columns and indexes for it. Then during the `prune Column` function, nothing can be done.

https://github.com/pingcap/tidb/blob/df249e9f50d3b8fa1b161bb778723ed322f0e84c/planner/core/logical_plan_builder.go#L5883

winoros:
Get/BatchGet might be difficult to support all the cases, planner can disable the Get/BatchGet choice if the table contains too much columns.

PR #54065, opened 2024-06-17
*: let planner decide the col refs for non-partition and non-foreignkey table's delete

winoros:

Issue Number: ref https://github.com/pingcap/tidb/issues/38911 

Problem Summary:

This pull tries to change the column references of rows to be deleted to the ones constructed by planner. Make it possible to prune columns for deletion.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


winoros:
I want this pull to get at least three LGTM.

So add the hold label to make sure that it would be merged accidentally.

winoros:
Multi-table delete's change will be added soon.

pkg/table/tables/partition.go
```diff
@@ -1684,6 +1684,33 @@ func (t *partitionedTable) RemoveRecord(ctx table.MutateContext, h kv.Handle, r
...
+		return errors.Trace(err)
+	}
+
+	if _, ok := t.reorganizePartitions[pid]; ok {
+		pid, err = t.locateReorgPartition(ectx.GetEvalCtx(), r)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		tbl = t.GetPartition(pid)
+		err = tbl.RemoveRecord(ctx, h, r)
```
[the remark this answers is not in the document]
winoros: partition case is not finished yet, we also need to deal with the partition expression.

pkg/executor/delete.go
```diff
@@ -62,16 +62,17 @@ func (e *DeleteExec) Next(ctx context.Context, req *chunk.Chunk) error {
...
 		end--
 	}
-	handle, err := handleCols.BuildHandleByDatums(row)
+	handle, err := colInfo.HandleCols.BuildHandleByDatums(row)
 	if err != nil {
 		return err
 	}
-	err = e.removeRow(e.Ctx(), tbl, handle, row[:end])
+	err = e.removeRow(e.Ctx(), tbl, handle, row[:end], colInfo)
+	// err = e.removeRow(e.Ctx(), tbl, handle, row[:end])
```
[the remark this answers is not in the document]
winoros: Yes, it's used to compare the new impl with old one.

pkg/table/tables/tables.go
```diff
@@ -1451,6 +1451,99 @@ func (t *TableCommon) RemoveRecord(ctx table.MutateContext, h kv.Handle, r []typ
...
+	txn, err := ctx.Txn(true)
+	if err != nil {
+		return err
+	}
+
+	memBuffer := txn.GetMemBuffer()
+	sh := memBuffer.Staging()
+	defer memBuffer.Cleanup(sh)
+
+	logutil.BgLogger().Debug("RemoveRecord",
```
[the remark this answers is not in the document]
winoros: yes

winoros:
QQ_1720456173242
FK is checked before the `recordSet.Close()` is called. I think it's a very reasonable behavior. So the single table delete case also needs to set and reset the extra info for each row.

pkg/table/context/table.go
```diff
@@ -60,6 +60,19 @@ type MutateContext interface {
...
 	GetMutateBuffers() *MutateBuffers
+
+	// SetExtraIndexKeyPosInfo sets the extra index key pos info.
+	// It's used when the given row is not a full row, but only contains some limit columns
+	SetExtraIndexKeyPosInfo(indexKeyPos map[int64][]int)
+
+	// GetExtraIndexKeyPosInfo gets the extra index key pos info.
+	GetExtraIndexKeyPosInfo(idxID int64) []int
+
+	// ResetExtraInfo resets the extra index key pos info.
```
cfzjywxk, line 71: Better to explain when it should be called, for example reseting statement context?

pkg/executor/delete.go
```diff
@@ -231,35 +249,41 @@ func (e *DeleteExec) deleteMultiTablesByChunk(ctx context.Context) error {
...
+		e.Ctx().GetTableCtx().ResetExtraInfo()
 		return err
 	}
 	tid := t.Meta().ID
 	err = onRemoveRowForFK(ctx, data, e.fkChecks[tid], e.fkCascades[tid])
 	if err != nil {
+		e.Ctx().GetTableCtx().ResetExtraInfo()
 		return err
 	}
+	e.Ctx().GetTableCtx().ResetExtraInfo()
```
cfzjywxk, line 286: It's easy to forget to reset in this way(so is L267, L277, L283), better to abstract a unified place which is guanrateed to be called, and do the cleanup or reset job there.

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5718,13 +5721,109 @@ func buildColumns2Handle(
...
+		found := -1
+		for j, col := range deletableCols {
+			if col.Name.L == name.ColName.L {
+				offsetMap[j] = i
+				found = j
+				break
+			}
+		}
+		if found == -1 {
+			return TblColPosInfo{}, plannererrors.ErrDeleteNotFoundColumn.GenWithStackByArgs(name.ColName.O, tblInfo.Name.O)
```
cfzjywxk, line 5798: Could the util method like `FindTableByName` be used here?

pkg/table/tables/tables.go
```diff
@@ -1478,11 +1478,43 @@ func (t *TableCommon) removeRowIndices(ctx table.MutateContext, h kv.Handle, rec
...
+		if err != nil {
+			logutil.BgLogger().Info("remove row index failed", zap.Any("index", v.Meta()), zap.Uint64("txnStartTS", txn.StartTS()), zap.String("handle", h.String()), zap.Any("record", rec), zap.Error(err))
+			return err
+		}
+		if err = v.Delete(ctx, txn, vals, h); err != nil {
+			if v.Meta().State != model.StatePublic && kv.ErrNotExist.Equal(err) {
+				// If the index is not in public state, we may have not created the index,
+				// or already deleted the index, so skip ErrNotExist error.
+				logutil.BgLogger().Debug("row index not exists", zap.Any("index", v.Meta()), zap.Uint64("txnStartTS", txn.StartTS()), zap.String("handle", h.String()))
+				continue
```
cfzjywxk, line 1495: Perhaps need review by the DDL owner, why the `kv.ErrNotExist` would be returned from the deletable index? Not quite sure about it.

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5718,13 +5721,109 @@ func buildColumns2Handle(
...
+			}
+			cols2PosInfos = append(cols2PosInfos, curColPosInfo)
+		}
+	}
+	sort.Sort(cols2PosInfos)
+	return cols2PosInfos, nil
+}
+
+// buildSingleTableColPosInfoForDelete builds columns mapping for delete.
+func buildSingleTableColPosInfoForDelete(
```
cfzjywxk, line 5768: Better to add unit test cases to cover the two new methods `buildColPositionInfoForDelete` and `buildSingleTableColPosInfoForDelete`.

pkg/planner/core/logical_plan_builder.go
winoros: No, most common cases are focused on `*model.ColumnInfo`, not `*table.Column`

pkg/table/contextimpl/table.go
```diff
@@ -73,3 +76,25 @@ func (ctx *TableContextImpl) GetMutateBuffers() *context.MutateBuffers {
...
+
+// ResetExtraInfo resets the extra index key pos info.
+func (ctx *TableContextImpl) ResetExtraInfo() {
+	ctx.hasExtraPosInfo = false
+	ctx.extraIndexKeyPosInfo = nil
+}
+
+// HasExtraInfo returns whether the context has extra index key pos info.
+func (ctx *TableContextImpl) HasExtraInfo() bool {
+	return ctx.hasExtraPosInfo
```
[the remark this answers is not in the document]
winoros: It's okay to do it. And later we'll support PartitionExpr, so it will become `len(extraIndexKeyPosInfo) == 0 && len(extraPartitionExpr) == 0` if we change to that way.

pkg/executor/delete.go
```diff
@@ -231,35 +248,41 @@ func (e *DeleteExec) deleteMultiTablesByChunk(ctx context.Context) error {
...
 	}
 
+	e.Ctx().GetTableCtx().ResetExtraInfo()
+
 	return nil
 }
 
-func (e *DeleteExec) removeRow(ctx sessionctx.Context, t table.Table, h kv.Handle, data []types.Datum) error {
+func (e *DeleteExec) removeRow(ctx sessionctx.Context, t table.Table, h kv.Handle, data []types.Datum, posInfo *plannercore.TblColPosInfo) error {
+	e.Ctx().GetTableCtx().SetExtraIndexKeyPosInfo(posInfo.IndexesForDelete)
```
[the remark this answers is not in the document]
AilinKid: seems multi-table delete statement cannot pin the posInfo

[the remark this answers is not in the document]
AilinKid: comment closed

winoros:
```go
//go:noinline
func doNoDefer(t *int) {
	*t++
	*t++
}

//go:noinline
func doDefer(t *int) {
	defer func() {
		*t++
	}()
	*t++
}

func BenchmarkDeferYes(b *testing.B) {
	t := 0
	for i := 0; i < b.N; i++ {
		doDefer(&t)
	}
}
func BenchmarkDeferNo(b *testing.B) {
	t := 0
	for i := 0; i < b.N; i++ {
		doNoDefer(&t)
	}
}
```

```
goos: linux
goarch: amd64
pkg: github.com/pingcap/tidb/pkg/planner/core
cpu: Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz
BenchmarkDeferYes
BenchmarkDeferYes-72            255241609                4.967 ns/op           0 B/op          0 allocs/op
BenchmarkDeferNo
BenchmarkDeferNo-72             729406417                1.646 ns/op           0 B/op          0 allocs/op
PASS
```

The performance has improved a lot. I think it's ok to use defer in this pull.

pkg/executor/delete.go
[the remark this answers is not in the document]
winoros: MySQL's multi table delete allows one table to be used multiple times. So we cannot change it here.

pkg/table/context/table.go
cfzjywxk: @winoros

pkg/table/contextimpl/table.go
```diff
@@ -34,6 +34,9 @@ type TableContextImpl struct {
 	// and saves allocation
 	// The buffers are supposed to be used inside AddRecord/UpdateRecord/RemoveRecord.
 	mutateBuffers *context.MutateBuffers
+
+	hasExtraPosInfo      bool
+	extraIndexKeyPosInfo map[int64][]int
```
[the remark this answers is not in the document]
cfzjywxk: @winoros

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5304,13 +5306,100 @@ func buildColumns2Handle(
...
+	}
+	indexColMap := make(map[int64][]int)
+	for _, idx := range deletableIdxs {
+		idxCols := idx.Meta().Columns
+		colPos := make([]int, 0, len(idxCols))
+		for _, col := range idxCols {
+			if col.Offset == len(pubCols) {
+				colPos = append(colPos, len(pubCols))
+				continue
+			}
```
crazycs520, line 5394: I don't understand this, could you add some comments?

pkg/executor/delete.go
```diff
@@ -164,15 +167,29 @@ func (e *DeleteExec) doBatchDelete(ctx context.Context) error {
...
 }
 
+// fixHandlePosInfoForMultiDelete fixes the position info by the current execution logic.
+// We will cut the mixed row into single table row, so the handle position info should be fixed.
+func (e *DeleteExec) fixHandlePosInfoForMultiDelete() {
+	for i := range e.tblColPosInfos {
+		info := e.tblColPosInfos[i]
+		for _, idxCols := range info.IndexesForDelete {
+			for i := range idxCols {
+				idxCols[i] -= info.Start
```
crazycs520, line 177: Could you add some comments about this?

pkg/table/tables/tables.go
```diff
@@ -1027,11 +1027,7 @@ func (t *TableCommon) addIndices(sctx table.MutateContext, recordID kv.Handle, r
 		if t.meta.IsCommonHandle && v.Meta().Primary {
 			continue
 		}
-		// We declared `err` here to make sure `indexVals` is assigned with `=` instead of `:=`.
-		// The latter one will create a new variable that shadows the outside `indexVals` that makes `indexVals` outside
-		// always nil, and we cannot reuse it.
-		var err error
-		indexVals, err = v.FetchValues(r, indexVals)
+		indexVals, err := v.FetchValues(sctx, r, indexVals)
```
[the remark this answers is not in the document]
winoros: oh, it's by mistake. The git merge was broken at that time...

pkg/planner/core/logical_plan_builder.go
winoros: ... Maybe useless.
Let me delete them and have a try

pkg/table/tables/tables.go
```diff
@@ -1211,7 +1211,7 @@ func (t *TableCommon) RemoveRecord(ctx table.MutateContext, h kv.Handle, r []typ
 	}
 
 	// The table has non-public column and this column is doing the operation of "modify/change column".
-	if len(t.Columns) > len(r) && t.Columns[len(r)].ChangeStateInfo != nil {
+	if !ctx.HasExtraInfo() && len(t.Columns) > len(r) && t.Columns[len(r)].ChangeStateInfo != nil {
```
crazycs520, line 1214: Look like `ctx.HasExtraInfo()` will always be true in `DELETE` DML, Does this affect the logic of the original if judgment?

pkg/table/context/table.go
```diff
@@ -133,6 +133,19 @@ type MutateContext interface {
...
+	ResetExtraInfo()
```
time-and-fate, line 145: I think it's better to have a consistent naming.
"ExtraInfo" is a bit confusing. We already call it "ExtraIndexKeyPosInfo", let's just call it "ExtraIndexKeyPosInfo" everywhere.

pkg/executor/delete.go
```diff
@@ -164,15 +167,31 @@ func (e *DeleteExec) doBatchDelete(ctx context.Context) error {
 	return nil
 }
 
+// fixHandlePosInfoForMultiDelete fixes the position info by the current execution logic.
+// We will cut the mixed row into single table row, so the handle position info should be fixed.
+//   e.g. The multi-delete case is [t1.a, t1.b, t2.a, t2.b] and There's a index [t2.b].
+//        The idxCols is [3]. And we should fix it to [1].
+func (e *DeleteExec) fixHandlePosInfoForMultiDelete() {
```
time-and-fate, line 174: Why don't we do this in the planner? We already have enough information and are doing similar things there.

pkg/table/tables/tables.go
winoros: Hahaha... insert on duplicate update will do the deletion. And this if branch is just designed for that case...

Insert on duplicate will use the whole columns(the same as deletable columns). It just uses writable columns.

pkg/executor/delete.go
winoros: Added more header comments

crazycs520 (commented):
REST LGTM, please address comments.

AilinKid (commented):
come conflicts need to be resolved

crazycs520 (approved):

pkg/table/context/table.go
```diff
@@ -130,6 +130,19 @@ type MutateContext interface {
 	// which is a buffer for table related structures that aims to reuse memory and
 	// saves allocation.
 	GetMutateBuffers() *MutateBuffers
+
+	// SetExtraIndexKeyPosInfo sets the extra index key pos info.
+	// It's used when the given row is not a full row, but only contains some limit columns
+	SetExtraIndexKeyPosInfo(indexKeyPos map[int64][]int)
+
+	// GetExtraIndexKeyPosInfo gets the extra index key pos info.
+	GetExtraIndexKeyPosInfo(idxID int64) []int
```
lcwangchao, line 139: I don't think it's a good way to put these "temporary" state to `MutateContext` for reasons:

1. In the current implementation, the `MutateContext` is kept in the session instead of building it in place and discarded after use. That means it can be shared across scenes and it's better to expose some stable facts only to make it easy to use, such as system variable, etc.
2. I think `HasExtraInfo` and `GetExtraIndexKeyPosInfo` is only used to describe the input datums for `RemoveRecord`, right? So it's better to keep them together. `MutateContext` is also used by `AddRecord` and `UpdateRecord` and other scenes like lightning import or ddl reorging. These new methods will make the context confusing.

Why not to introduce a new struct to replace `[]types.Datum` as the input of `RemoveRecord`, for example:

```
type PartialRow struct {
    // d may contains the part of column datums of a row
    d []types.Datum
    // offsetMapping has a same size with d, it indicates the offset mapping the offset in `d` to the real offset
    offset mapping []int
}
```

So we can change signature the method `RemoveRecord` to:

```
RemoveRecord(ctx MutateContext, txn kv.Transaction, h kv.Handle, row PartialRow) error
```

And `FetchValues`

```
FetchValues(row PartialRow, columns []types.Datum) ([]types.Datum, error)
```

pkg/table/index.go
```diff
@@ -102,7 +102,7 @@ type Index interface {
 	// FetchValues fetched index column values in a row.
 	// Param columns is a reused buffer, if it is not nil, FetchValues will fill the index values in it,
 	// and return the buffer, if it is nil, FetchValues will allocate the buffer instead.
-	FetchValues(row []types.Datum, columns []types.Datum) ([]types.Datum, error)
+	FetchValues(ctx MutateContext, row []types.Datum, columns []types.Datum) ([]types.Datum, error)
```
lcwangchao, line 105: It's strange to pass `MutateContext` to `FetchValues` as an argument because this method  does not mutate any data.

pkg/table/tables/tables.go
lcwangchao: My confusion is why we skip to handle the non-public column in normal deletion. Because the missing column data has already been processed outside `RemoveRecord` ? If so, could we move the below codes totally outside, it is really confusing here...

pkg/table/context/table.go
lcwangchao: Another way is to introduce `DeleteRecordOption` like `AddRecord` or `UpdateRecord`, if the input is a partial row, we can invoke `RemoveRecord` like this:

```
t.RemoveRecord(ctx.GetTableCtx(), txn, h, data, table.WithRowIndexMapping(posInfo.IndexesForDelete))
```

pkg/table/tables/tables.go
```diff
@@ -1188,7 +1192,10 @@ func (t *TableCommon) RemoveRecord(ctx table.MutateContext, txn kv.Transaction,
 	}
 
 	// The table has non-public column and this column is doing the operation of "modify/change column".
-	if len(t.Columns) > len(r) && t.Columns[len(r)].ChangeStateInfo != nil {
+	// DELETE will use deletable columns, which is the same as the full columns of the table.
+	// INSERT and UPDATE will only use the wrtiable columns. So they will not see column under MOIDFY/CHANGE state.
+	// This if block is for for the INSERT and UPDATE.
+	if !opt.HasIndexesLayout() && len(t.Columns) > len(r) && t.Columns[len(r)].ChangeStateInfo != nil {
```
lcwangchao, line 1198: Please add a comment that `opt.HasIndexesLayout()` returns `true` in `DELETE` statement and `false` in `INSERT` and `UPDATE`

lcwangchao (commented):
LGTM for reset context part

pkg/executor/delete.go
```diff
@@ -83,10 +83,13 @@ func (e *DeleteExec) deleteSingleTableByChunk(ctx context.Context) error {
 		tbl           table.Table
 		isExtrahandle bool
 		handleCols    util.HandleCols
+		colPosInfo    *plannercore.TblColPosInfo
 		rowCount      int
 	)
-	for _, info := range e.tblColPosInfos {
+	for i := range e.tblColPosInfos {
```
lcwangchao, line 89: why make this change？

lcwangchao: I think `for i, info := range e.tblColPosInfos` is better

lcwangchao (approved):

pkg/executor/delete.go
winoros: removed the loop for this single table deletion case

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5252,13 +5254,97 @@ func buildColumns2Handle(
...
+				break
+			}
+		}
+		if found == -1 {
+			return TblColPosInfo{}, plannererrors.ErrDeleteNotFoundColumn.GenWithStackByArgs(name.ColName.O, tblInfo.Name.O)
+		}
+		// For multi-delete case, the offset is recorded for the mixed row. We need to restore it to single table status.
+		//	e.g. The multi-delete case is [t1.a, t1.b, t2.a, t2.b] and There's a index [t2.b].
+		//	     The idxCols is [3]. And we should fix it to [1].
+		offsetMap[found] = i - offset
```
time-and-fate, line 5333: Why don't we use `names[offset:end]` for the `for` loop so we don't need this?

pkg/planner/core/point_get_plan.go
```diff
@@ -2130,6 +2130,18 @@ func buildHandleCols(ctx base.PlanContext, tbl *model.TableInfo, schema *express
...
+	if schema.Len() > 0 {
+		tableAliasName = pointget.OutputNames()[0].TblName
+	}
+	newOutputNames = append(newOutputNames, &types.FieldName{
+		DBName:      pmodel.NewCIStr(dbName),
+		TblName:     tableAliasName,
+		OrigTblName: tbl.Name,
+		ColName:     model.ExtraHandleName,
+	})
+	pointget.SetOutputNames(newOutputNames)
```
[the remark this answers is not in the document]
winoros: The new `buildSingleTableColPosInfoForDelete` will need it.

pkg/planner/core/point_get_plan.go
```diff
@@ -2064,21 +2064,20 @@ func buildPointDeletePlan(ctx base.PlanContext, pointPlan base.PhysicalPlan, dbN
...
-				End:        pointPlan.Schema().Len(),
-				HandleCols: handleCols,
-			},
-		},
-	}.Init(ctx)
+	handleCols := buildHandleCols(ctx, dbName, tbl, pointPlan)
 	var err error
 	is := ctx.GetInfoSchema().(infoschema.InfoSchema)
 	t, _ := is.TableByID(context.Background(), tbl.ID)
+	idxs := t.DeletableIndices()
```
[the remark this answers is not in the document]
winoros: Use intest to check it.

pkg/planner/core/logical_plan_builder.go
winoros: I'll add some in the next pr that supports the column pruning.

time-and-fate (approved):

PR #54091, opened 2024-06-18
executor, session: Add Detach() method to detach the record set

### LINKED ISSUE #54090: Add a `Detach` method for `RecordSet` to detach a record set from the session context
## Enhancement

To achieve the lazy cursor fetch functionality, we'll need to detach the `executor -> recordSet -> resultSet` from the session context. The PR to add `Detach` method for executor is WIP https://github.com/pingcap/tidb/issues/53336. At the same time, we can add a method for `recordSet` to detach the content of it from the session context.

We need to carefully handle the following things:

1. `*ExecStmt`, which will be used to upload the statistical information after executing the statement.
2. `*execStmtResult`, which will auto-commit the transaction after executing the statement.

For this issue, we'll create a simple `RecordSet`, to run read-only statement without statistical information and not allow running in an explicit transaction.

YangKeao:

Issue Number: close #54090

Problem Summary:

The original `recordSet` and `execStmtResult` are too complicated to implement the lazy cursor fetch feature. We do need a much simpler one: without explicit transaction, without metrics/summaries and only supports read-only statements.

This PR adds a method `Detach` to create a new record set from existing `recordSet` or `execStmtResult`. The new record set can run without depending on the session context.

However, currently the executor still cannot be detached, so this PR specially handled the `TableReaderExecutor` and left a `TODO` here.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/adapter.go
+++ b/pkg/executor/adapter.go
@@ -35,6 +35,7 @@ import (
 	"github.com/pingcap/tidb/pkg/domain"
 	"github.com/pingcap/tidb/pkg/executor/internal/exec"
 	executor_metrics "github.com/pingcap/tidb/pkg/executor/metrics"
+	"github.com/pingcap/tidb/pkg/executor/staticrecordset"
 	"github.com/pingcap/tidb/pkg/expression"
 	"github.com/pingcap/tidb/pkg/infoschema"
 	"github.com/pingcap/tidb/pkg/keyspace"
@@ -239,6 +240,17 @@ func (a *recordSet) Close() error {
 // OnFetchReturned implements commandLifeCycle#OnFetchReturned
 func (a *recordSet) OnFetchReturned() {
 	a.stmt.LogSlowQuery(a.txnStartTS, len(a.lastErrs) == 0, true)
+}
+
+// Detach creates a new `RecordSet` which doesn't depend on the current session context.
+func (a *recordSet) TryDetach() (sqlexec.RecordSet, bool, error) {
+	// TODO: also detach the executor. Currently, the executor inside may contain the session context. Once
+	// the executor itself supports detach, we should also detach it here.
+	e, ok := a.executor.(*TableReaderExecutor)
+	if !ok {
+		return nil, false, nil
+	}
+	return staticrecordset.New(a.Fields(), e, a.stmt.GetTextToLog(false)), true, nil
 }
 
 // ExecStmt implements the sqlexec.Statement interface, it builds a planner.Plan to an sqlexec.Statement.
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -2379,6 +2379,40 @@ func (rs *execStmtResult) Close() error {
 	return err2
 }
 
+func (rs *execStmtResult) TryDetach() (sqlexec.RecordSet, bool, error) {
+	if !rs.sql.IsReadOnly(rs.se.GetSessionVars()) {
+		return nil, false, nil
+	}
+	if !plannercore.IsAutoCommitTxn(rs.se.GetSessionVars()) {
+		return nil, false, nil
+	}
+
+	drs, ok := rs.RecordSet.(sqlexec.DetachableRecordSet)
+	if !ok {
+		return nil, false, nil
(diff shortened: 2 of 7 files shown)


pkg/session/session.go
```diff
@@ -2379,6 +2380,39 @@ func (rs *execStmtResult) Close() error {
...
+	}
+	detachedRS, err := drs.Detach()
+	if err != nil {
+		return nil, err
+	}
+
+	// FIXME: if the following code meets error, the `detachedRS` will be leaked.
+	// The error may occur during the begin/commit of the newly created transaction on the given `startTS``.
+
+	// FIXME: block the min-start-ts. Now the `min-start-ts` will advance and exceed the `startTS` used by
```
YangKeao, line 2403: This `FIXME` will be fixed in https://github.com/pingcap/tidb/pull/54114

pkg/util/sqlexec/restricted_sql_executor.go
```diff
@@ -199,6 +199,14 @@ type RecordSet interface {
 	Close() error
 }
 
+// DetachableRecordSet extends the `RecordSet` to support detaching from current session context
+type DetachableRecordSet interface {
+	RecordSet
+
+	// Detach detaches the record set from the current session context.
+	Detach() (RecordSet, error)
```
lcwangchao, line 207: Could we make this function to `TryDetach() (RecordSet, bool)` because and recordSet that cannot be detach may be a normal case

pkg/executor/adapter.go
```diff
@@ -241,6 +242,21 @@ func (a *recordSet) OnFetchReturned() {
 	a.stmt.LogSlowQuery(a.txnStartTS, len(a.lastErrs) == 0, true)
 }
 
+// Detach creates a new `RecordSet` which doesn't depend on the current session context.
+func (a *recordSet) Detach() (sqlexec.RecordSet, error) {
+	cteStorage := a.stmt.Ctx.GetSessionVars().StmtCtx.CTEStorageMap
```
lcwangchao, line 247: Can we return an error when `cteStorage` is not empty because CTE is not currently supported.

In the future, if we want to support CTE,  I think we should copy this map

pkg/session/session.go
```diff
@@ -2379,6 +2379,39 @@ func (rs *execStmtResult) Close() error {
...
+	// this detached record set, and may cause an error if `GC` runs.
+
+	// Now, a transaction is not needed for the detached record set, so we commit the transaction and cleanup
+	// the session state.
+	err = rs.se.CommitTxn(context.Background())
```
lcwangchao, line 2407: And what is the expected operation for an old RecordSet if a developer called `Detach`? Does he still need to call the `Close` method of the old one? If it is yes, I think we do not need to call `CommitTxn` because the old one does it.

YangKeao: > Does he still need to call the `Close` method of the old one?

No. If the `Detach` return successfully, the old `RecordSet` shouldn't be used anymore (including `Close()`).

I think it's not easy to write a `Close` suitable for both detached and not-detached record set. Therefore, we should clean up TXN states here.

pkg/executor/adapter.go
YangKeao: Done.

pkg/util/sqlexec/restricted_sql_executor.go
YangKeao: I'll return `TryDetach() (RecordSet, bool, error)` to indicate whether the statement is suitable to detach, and whether it detaches successfully. I sounds strange but I didn't come up a better way :facepalm:.

If the caller receives `nil, false, _`, it means the original record set can still be used. If the caller receives `_, _, err`, it means the state of session/recordSet is in a chaos, and usually the connection will be halted.

pkg/session/session.go
YangKeao: I'll use `finishStmt` directly.

lcwangchao (approved):

PR #54110, opened 2024-06-19
ddl: support referring objects in runningJobs

lance6716:

Issue Number: close https://github.com/pingcap/tidb/issues/52795, ref https://github.com/pingcap/tidb/issues/53246

Problem Summary:

a DDL job will modify and refer to multiple objects (database, table, placement policy, resource group). Treat it like the job will lock these objects, in exclusive or shared lock mode. If the job finished all locking, then it can run now. In other words, the job has no dependency on other running jobs. This PR implements this lock-like behaviour.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/ddl_api.go
+++ b/pkg/ddl/ddl_api.go
@@ -2853,7 +2853,7 @@ func (d *ddl) createTableWithInfoJob(
 		args = append(args, ctx.GetSessionVars().ForeignKeyChecks)
 	}
 
-	var involvingSchemas []model.InvolvingSchemaInfo
+	involvingSchemas := make([]model.InvolvingSchemaInfo, 0, len(involvingRef)+len(tbInfo.ForeignKeys)+1)
 	if len(involvingRef) > 0 {
 		involvingSchemas = append(involvingSchemas, model.InvolvingSchemaInfo{
 			Database: schema.Name.L,
@@ -3249,13 +3249,13 @@ func (d *ddl) FlashbackCluster(ctx sessionctx.Context, flashbackTS uint64) error
 		Args: []any{
 			flashbackTS,
 			map[string]any{},
-			true,         /* tidb_gc_enable */
-			variable.On,  /* tidb_enable_auto_analyze */
-			variable.Off, /* tidb_super_read_only */
-			0,            /* totalRegions */
-			0,            /* startTS */
-			0,            /* commitTS */
-			variable.On,  /* tidb_ttl_job_enable */
+			true,           /* tidb_gc_enable */
+			variable.On,    /* tidb_enable_auto_analyze */
+			variable.Off,   /* tidb_super_read_only */
+			0,              /* totalRegions */
+			0,              /* startTS */
+			0,              /* commitTS */
+			variable.On,    /* tidb_ttl_job_enable */
 			[]kv.KeyRange{} /* flashback key_ranges */},
 		CDCWriteSource: ctx.GetSessionVars().CDCWriteSource,
 		// FLASHBACK CLUSTER affects all schemas and tables.
--- a/pkg/ddl/ddl_api.go
+++ b/pkg/ddl/ddl_api.go
@@ -229,6 +229,11 @@ func (d *ddl) CreateSchemaWithInfo(
 		}},
 		SQLMode: ctx.GetSessionVars().SQLMode,
 	}
+	if ref := dbInfo.PlacementPolicyRef; ref != nil {
+		job.InvolvingSchemaInfo[0].Shared = &model.InvolvingSharedSchemaInfo{
+			Policy: ref.Name.L,
+		}
+	}
 
 	err = d.DoDDLJob(ctx, job)
 	err = d.callHookOnChanged(job, err)
@@ -305,6 +310,7 @@ func (d *ddl) ModifySchemaDefaultPlacement(ctx sessionctx.Context, stmt *ast.Alt
 		InvolvingSchemaInfo: []model.InvolvingSchemaInfo{{
 			Database: dbInfo.Name.L,
 			Table:    model.InvolvingAll,
+			Shared:   &model.InvolvingSharedSchemaInfo{Policy: placementPolicyRef.Name.L},
 		}},
 		SQLMode: ctx.GetSessionVars().SQLMode,
 	}
@@ -525,16 +531,26 @@ func (d *ddl) AlterTablePlacement(ctx sessionctx.Context, ident ast.Ident, place
 		return err
 	}
 
+	var involvingSchemaInfo []model.InvolvingSchemaInfo
+	if placementPolicyRef != nil {
+		involvingSchemaInfo = []model.InvolvingSchemaInfo{{
+			Database: schema.Name.L,
+			Table:    tblInfo.Name.L,
+			Shared:   &model.InvolvingSharedSchemaInfo{Policy: placementPolicyRef.Name.L},
+		}}
+	}
+
 	job := &model.Job{
-		SchemaID:       schema.ID,
-		TableID:        tblInfo.ID,
-		SchemaName:     schema.Name.L,
-		TableName:      tblInfo.Name.L,
-		Type:           model.ActionAlterTablePlacement,
-		BinlogInfo:     &model.HistoryInfo{},
-		CDCWriteSource: ctx.GetSessionVars().CDCWriteSource,
-		Args:           []any{placementPolicyRef},
-		SQLMode:        ctx.GetSessionVars().SQLMode,
+		SchemaID:            schema.ID,
+		TableID:             tblInfo.ID,
+		SchemaName:          schema.Name.L,
+		TableName:           tblInfo.Name.L,
+		Type:                model.ActionAlterTablePlacement,
+		BinlogInfo:          &model.HistoryInfo{},
+		CDCWriteSource:      ctx.GetSessionVars().CDCWriteSource,
+		Args:                []any{placementPolicyRef},
+		InvolvingSchemaInfo: involvingSchemaInfo,
+		SQLMode:             ctx.GetSessionVars().SQLMode,
 	}
 
 	err = d.DoDDLJob(ctx, job)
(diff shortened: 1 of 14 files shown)


pkg/ddl/ddl_api.go
```diff
@@ -3290,11 +3341,13 @@ func (d *ddl) CreateView(ctx sessionctx.Context, s *ast.CreateViewStmt) (err err
 		onExist = OnExistReplace
 	}
 
-	return d.CreateTableWithInfo(ctx, s.ViewName.Schema, tbInfo, onExist)
+	var involvingRef []model.InvolvingSchemaInfo
+	// TODO(lance6716): visit the s.Select to get all TableNames?
```
D3Hunter, line 3345: i think it's not needed, view is only a stored sql, it's not checked when operating underly table

D3Hunter (commented):
9/14

pkg/parser/model/ddl.go
```diff
@@ -585,15 +585,36 @@ type Job struct {
...
 
-// InvolvingSchemaInfo returns the schema info involved in the job.
-// The value should be stored in lower case.
+// InvolvingSchemaInfo returns the schema info involved in the job. The value
+// should be stored in lower case. The top level fields are exclusively used for
+// the job, and the shared objects are saved in Shared field.
+//
+// Exclusive and shared objects are considered like the exclusive lock and shared
+// lock when calculate DDL job dependencies. And we also implement the fair lock
+// semantic which means if job B (exclusive request object 0) is waiting for the
```
D3Hunter, line 594:
```suggestion
// semantic which means if we have job A/B/C arrive in order, and job B (exclusive request object 0) is waiting for the
```

pkg/parser/model/ddl.go
```diff
@@ -585,15 +585,36 @@ type Job struct {
...
+	Table         string `json:"table,omitempty"`
+	Policy        string `json:"policy,omitempty"`
+	ResourceGroup string `json:"resource_group,omitempty"`
 }
 
 const (
-	// InvolvingAll means all schemas/tables are affected.
+	// InvolvingAll means all schemas/tables are affected. If it's used in
+	// InvolvingSchemaInfo.Database, it also means all placement policies and
+	// resource groups are affected. Currently the only usage is FLASHBACK CLUSTER.
```
D3Hunter, line 617:
```suggestion
	// InvolvingSchemaInfo.Database/Table. It also means all placement policies and
	// resource groups are affected for FLASHBACK CLUSTER.
```

pkg/parser/model/ddl.go
```diff
@@ -585,15 +585,36 @@ type Job struct {
...
+// lock when calculate DDL job dependencies. And we also implement the fair lock
+// semantic which means if job B (exclusive request object 0) is waiting for the
+// running job A (shared request object 0), and job C (shared request object 0)
+// arrives, job C should also be blocked until job B is finished although job A &
+// C has no dependency.
 type InvolvingSchemaInfo struct {
-	Database string `json:"database"`
-	Table    string `json:"table"`
+	Database      string                     `json:"database"`
+	Table         string                     `json:"table"`
```
D3Hunter, line 600:
```suggestion
	Database      string                     `json:"database,omitempty"`
	Table         string                     `json:"table,omitempty"`
```

D3Hunter (commented):
11/14

pkg/parser/model/ddl.go
```diff
@@ -585,15 +585,36 @@ type Job struct {
...
+	Policy        string                     `json:"policy,omitempty"`
+	ResourceGroup string                     `json:"resource_group,omitempty"`
+	Shared        *InvolvingSharedSchemaInfo `json:"shared,omitempty"`
+}
+
+// InvolvingSharedSchemaInfo is used by InvolvingSchemaInfo.
+type InvolvingSharedSchemaInfo struct {
```
D3Hunter, line 607: those 2 struct is so similar, i prefer unify them and add a enum field to mean it's exclusive or shared

pkg/parser/model/ddl.go
lance6716: Not sure we should keep compatibility, but seems no harm.

pkg/ddl/job_table.go
```diff
@@ -184,7 +184,10 @@ func (s *jobScheduler) close() {
 	}
 }
 
+// getJob reads tidb_ddl_job and returns the first runnable DDL job.
 func (s *jobScheduler) getJob(se *sess.Session, tp jobType) (*model.Job, error) {
```
D3Hunter, line 188: we need merge 2 `loadDDLJobAndRun` into 1 to hand a 'fair' case below:
- job 1: add column to test.t
- job 2: truncate test.t: skipped scan since job 1 is running
- job 3: add index to test.t: before we check it, job 1 finished, so it can be run, but it should be run after job 2

since it's already a large pr, i add a task item to track and we can change it later, and can do it together with remove the sub-query inside getJobSQL to speedup.

pkg/ddl/job_table.go
```diff
@@ -235,6 +240,7 @@ func (s *jobScheduler) getJob(se *sess.Session, tp jobType) (*model.Job, error)
 				"[ddl] handle ddl job failed: mark job is processing meet error",
 				zap.Error(err),
 				zap.Stringer("job", &job))
+			s.runningJobs.addPending(involving)
```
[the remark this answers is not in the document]
lance6716: Need more reviewers to careful think about that, if it's essential to persist the result of which job is chosen as running.

pkg/ddl/ddl_running_jobs.go
```diff
@@ -37,57 +37,180 @@ type runningJobs struct {
...
+		return false
+	}
+	// Currently flashback cluster is the only DDL that involves ALL schemas.
+	if _, ok := j.exclusive.schemas[model.InvolvingAll]; ok {
+		return false
+	}
+
+	for _, info := range involves {
+		if intest.InTest {
+			if info.Database == model.InvolvingNone && info.Table != model.InvolvingNone {
```
D3Hunter, line 99: can use `intest.Assert`

pkg/ddl/ddl_running_jobs.go
```diff
@@ -37,57 +37,180 @@ type runningJobs struct {
...
+					jobID, info.Database,
+				))
+			}
+		}
+
+		if info.Database == model.InvolvingAll && info.Table == model.InvolvingAll &&
+			info.Mode == model.ExclusiveInvolving {
+			// check for involving all databases and tables, where the only case is FLASHBACK
+			// CLUSTER
+			if len(j.exclusive.schemas) != 0 || len(j.shared.schemas) != 0 ||
```
D3Hunter, line 117: maybe move the `empty` check above, if it's empty, then any job can be run directly.

pkg/ddl/ddl_running_jobs.go
```diff
@@ -37,57 +37,180 @@ type runningJobs struct {
...
+				return false
+			}
+			continue
+		}
+
+		var toCheck []*objects
+		switch info.Mode {
+		case model.ExclusiveInvolving:
+			// Exclusive objects conflicts with running exclusive and shared objects.
+			toCheck = []*objects{j.exclusive, j.shared}
```
D3Hunter, line 129: check `pending`?

D3Hunter (commented):
11/13

pkg/ddl/ddl_running_jobs.go
```diff
@@ -37,57 +37,180 @@ type runningJobs struct {
...
+	schemas           map[string]map[string]struct{}
+	placementPolicies map[string]struct{}
+	resourceGroups    map[string]struct{}
+}
+
+func newObjects() *objects {
+	return &objects{
+		schemas:           make(map[string]map[string]struct{}),
+		placementPolicies: make(map[string]struct{}),
+		resourceGroups:    make(map[string]struct{}),
```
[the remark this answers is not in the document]
lance6716: I see the initial size is chosen by runtime https://pkg.go.dev/builtin#make

> The size may be omitted, in which case a small starting size is allocated.

pkg/parser/model/ddl.go
```diff
@@ -585,15 +585,38 @@ type Job struct {
 	SQLMode mysql.SQLMode `json:"sql_mode"`
 }
 
-// InvolvingSchemaInfo returns the schema info involved in the job.
-// The value should be stored in lower case.
+// InvolvingSchemaInfo returns the schema info involved in the job. The value
+// should be stored in lower case. No need to use this type as one-of
+// Database/Table, Policy, ResourceGroup.
```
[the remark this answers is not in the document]
D3Hunter: i cannot think of a case of mixing them, when they do they have different involve mode.

pkg/ddl/ddl_running_jobs.go
lance6716: You mean add a check that all fields of `info` is empty above here?

D3Hunter: when both j.exclusive and j.shared are empty, above the for loop

pkg/ddl/ddl_running_jobs.go
lance6716: I think `pending` is caused by previous `shared` objects, so check `shared` is enough. But then I reallised `shared` objects may be removed concurrently during a round of `getJob`. I'll add `pending` in next commit

pkg/ddl/ddl_api.go
```diff
@@ -2977,7 +3039,7 @@ func (d *ddl) BatchCreateTableWithInfo(ctx sessionctx.Context,
 			}
 		}
 
-		job, err := d.createTableWithInfoJob(ctx, dbName, info, c.OnExist, true)
+		job, err := d.createTableWithInfoJob(ctx, dbName, info, nil, c.OnExist, true)
```
D3Hunter, line 3042: for batch create tables we should append their involved schema to the final job

D3Hunter (commented):
rest lgtm

pkg/parser/model/ddl.go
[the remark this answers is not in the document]
D3Hunter: and this too

pkg/ddl/ddl_api.go
lance6716: I have confirmed with BR dev that BatchCreateTable is only used in BR, and BR makes sure no concurrent DDL API will be called. Apart from that, BatchCreateTable has no SQL statement and should not happen in TiDB DDL use case. I will add a TODO for it.

pkg/parser/model/ddl.go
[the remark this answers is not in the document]
lance6716: OK, I'll make it one-of type

pkg/ddl/ddl_api.go
lance6716: Oh I understand you mean add involved schema `db.*`

D3Hunter (approved):

lance6716:
ping @tangenta

tangenta (approved):

PR #54114, opened 2024-06-19
session, infosync: add a cursor tracker to the session to keep tracking all needed information of the cursor.

### LINKED ISSUE #54113: Add a `CursorTracker` to track the `min-start-ts` of each cursor.
## Enhancement

We should track all existing cursors on a session, and store the state (`min-start-ts`, and any other things needed), so that the cursor can be observed.

YangKeao:

Issue Number: close #54113

Problem Summary:

The `min-start-ts` of the cursor is not tracked, so that the safe point may advance to be greater than the `startTS` used by the cursor.

Add a cursor tracker on the session, and expose a read-only interface.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/testkit/mocksessionmanager.go
```diff
@@ -157,11 +157,13 @@ func (msm *MockSessionManager) GetInternalSessionStartTSList() []uint64 {
 	ret := make([]uint64, 0, len(msm.internalSessions))
 	for internalSess := range msm.internalSessions {
 		se := internalSess.(sessionctx.Context)
-		sessVars := se.GetSessionVars()
-		sessVars.TxnCtxMu.Lock()
-		startTS := sessVars.TxnCtx.StartTS
-		sessVars.TxnCtxMu.Unlock()
-		ret = append(ret, startTS)
```
YangKeao, line 164: The original implementation is not correct. The `sessVars.TxnCtx.StartTS` is not cleared after `Commit`, so that a pending session will always have a small start ts.

pkg/session/cursor/state.go
```diff
@@ -0,0 +1,36 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cursor
+
+// State is a read-only interface to access the state of cursor
+type State interface {
```
lcwangchao, line 18: Why we need to add a new interface `State` instead of just using a `uint64` or `struct` as the state

lcwangchao (approved):

pkg/session/cursor/state.go
YangKeao: I agree. The interface is over-complicated. I'll use a struct directly as instead.

lance6716 (commented):
/approve

for domain part

xhebox (approved):

PR #54142, opened 2024-06-20
ddl: query since min job id for tidb_mdl_info

D3Hunter:

Issue Number: ref #53246

Problem Summary:

- similar to https://github.com/pingcap/tidb/pull/54122, we change how we query tidb_mdl_info

### Check List

Tests 

- [x] Unit test
- [ ] Integration test

case is same as https://github.com/pingcap/tidb/pull/54122, time takes 34m58 -> 24m44. this pr have more speedup might be caused by that job schedule part is faster infoschema reload, and all jobs need wait infoschema reload, so it have more effect even through that they are all single thread

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/ddl/systable/min_job_id.go
```diff
@@ -0,0 +1,63 @@
...
+		return
+	}
+	r.lastRefreshTime = now
+	minID, err := r.sysTblMgr.GetMinJobID(ctx, r.currMinJobID)
+	if err != nil {
+		logutil.DDLLogger().Info("get min job ID failed", zap.Error(err))
+		return
+	}
+	// use max, in case all job are finished to avoid the currMinJobID go back.
+	r.currMinJobID = max(r.currMinJobID, minID)
```
tangenta, line 62: Is it possible to write a job with a smaller ID into `tidb_ddl_job` due to some network issues?

D3Hunter: job id is allocated from meta `NextGlobalID` which is increased in txn. if network issue after alloc, then those allocated id are a hole and will not be reused, see
https://github.com/pingcap/tidb/blob/f4ae757aaeb80abc0ec199d58330da3c60544f23/pkg/ddl/ddl_worker.go#L432

tangenta (approved):

wjhuang2016 (approved):

PR #16798, opened 2020-04-24
*: fix the way to use `recover` in ddl, and unify using `recover` in domain and ddl

zimulala:
### What problem does this PR solve?

Problem Summary:
If `limitDDLJobs` is panic and after recovery in the package of ddl, this TiDB can't put the DDL job to storage and don't return errors.
If `start`  is panic and after recovery in the package of ddl, this TiDB can't handle the DDL job in the background even if it is the owner.

### What is changed and how it works?

What's Changed:

- If `limitDDLJobs` and `start` are panic and after recovery in the package of ddl, print logs and exit this TiDB.

- Unify using `recover` in the packages of domain and ddl.

### Related changes

- Need to cherry-pick to the release branch

### Check List 

Tests 

- No code

### Release note 

- Fix the way to use `recover` in ddl, and unify using `recover` in domain and ddl

PR #54166, opened 2024-06-22
Optimizer: Fix range intersection for CNF(conjunctive normal form)

ghazalfamilyusa:

Issue Number: Ref #16798

### Problem Summary:
Current code does not optimally find index range when we have the filter as complex conjuncts. A complex cojunct in this context is DNF. The code does well for each DNF and builds a union of ranges but for the CNF part, it does not do full intersection. The code does intersection only for point ranges which may not as selective as the full intersection.  There is a simple example under https://github.com/tikv/tikv/issues/16798 that shows the optimization limitation. 

 We added code  does full intersection of ranges. For example, if we have ((a1>1) or (a1=1 and b1 > 10)) and ((a1<10) or (a1=10 and b1 < 20))
then there are two conjuncts : 
- ((a1>1) or (a1=1 and b1 > 10)) 
- ((a1<10) or (a1=10 and b1 < 20))
The first produces intervals : (1 10,1 +inf], (1,+inf] and second intervals is [-inf,10), [10 -inf,10 20) and their intersection is 
(1 10,1 +inf], (1,10), [10 -inf,10 20).  Before this fix, the code just produce point ranges as [10,10], (1,1].

### Check List

Tests 

- [X] Unit test
- [] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/casetest/index/testdata/index_range_in.json
+++ b/pkg/planner/core/casetest/index/testdata/index_range_in.json
@@ -19,5 +19,25 @@
 	"select /*+ USE_INDEX(t1char,PKX) */ count(*) from t1char where  ((a1<'10') or (a1='10' and b1 < '20')) -- upper bound on (a1,b1)",
 	"select /*+ USE_INDEX(t1char,PKX) */ count(*) from t1char where ((a1>'1') or (a1='1' and b1 > '10')) and ((a1<'10') or (a1='10' and b1 < '20')) -- upper and lower bound on (a1,b1)"
     ]
+  },
+  {
+    "name": "TestRangeIntersection",
+    "cases": [
+	"select /*+ USE_INDEX(t1,PKX) */ count(*) from t1 where ((a1>1) or (a1=1 and b1 > 10));",
+	"select /*+ USE_INDEX(t1,PKX) */ count(*) from t1 where ((a1<10) or (a1=10 and b1 < 20));",
+	"select /*+ USE_INDEX(t1,PKX) */ count(*) from t1 where ((a1>1) or (a1=1 and b1 > 10)) and ((a1<10) or (a1=10 and b1 < 20));",
+	"select /*+ USE_INDEX(t1,PKX) */ count(*) from t1 where (a1>1 and a1 < 10) or (a1=1 and b1>10) or (a1=10 and b1<20);",
+	"select /*+ USE_INDEX(t11,PKX) */ count(*) from t11 where ((a1>1) or (a1=1 and b1 > 10));",
+	"select /*+ USE_INDEX(t11,PKX) */ count(*) from t11 where ((a1<10) or (a1=10 and b1 < 20));",
+	"select /*+ USE_INDEX(t11,PKX) */ count(*) from t11 where ((a1>1) or (a1=1 and b1 > 10)) and ((a1<10) or (a1=10 and b1 < 20));",
+	"select /*+ USE_INDEX(t11,PKX) */ count(*) from t11 where (a1>1 and a1 < 10) or (a1=1 and b1>10) or (a1=10 and b1<20);",
+	"select /*+ USE_INDEX(t,PKX) */ * from t where (a,b) in ((1,1),(2,2)) and c > 2;",
+	"select /*+ USE_INDEX(t,PKX) */ * from t where c > 2 and (a,b,c) in ((1,1,1),(2,2,3));",
+	"select /*+ USE_INDEX(t,PKX) */ * from t where (a,b) in ((1,1),(2,2)) and c > 2 and (a,b,c) in ((1,1,1),(2,2,3));",
+	"select /*+ USE_INDEX(t,PKX) */ * from tt where (a,b) in ((1,1),(2,2)) and c > 2;",
+	"select /*+ USE_INDEX(t,PKX) */ * from tt where c > 2 and (a,b,c) in ((1,1,1),(2,2,3));",
+	"select /*+ USE_INDEX(t,PKX) */ * from tt where (a,b) in ((1,1),(2,2)) and c > 2 and (a,b,c) in ((1,1,1),(2,2,3));",
+	"select /*+ USE_INDEX(tablename,PKK) */ count(*) from tablename where (primary_key > '1primary_key_start' or (primary_key = '1primary_key_start' and secondary_key > '3secondary_key_start') or (primary_key = '1primary_key_start' and secondary_key = '3secondary_key_start' and timestamp >= '1707885658544000000')) and (primary_key < '2primary_key_end' or (primary_key = '2primary_key_end' and secondary_key < '4secondary_key_end') or (primary_key = '2primary_key_end' and secondary_key = '4secondary_key_end' and timestamp <= '2707885658544000000'));"
+    ]
   }
 ]
--- a/pkg/util/ranger/detacher.go
+++ b/pkg/util/ranger/detacher.go
@@ -273,8 +273,32 @@ func extractBestCNFItemRanges(sctx *rangerctx.RangerContext, conds []expression.
 			continue
 		}
 		curRes := getCNFItemRangeResult(sctx, res, i)
-		if bestRes == nil || compareCNFItemRangeResult(curRes, bestRes) {
+		if bestRes == nil {
 			bestRes = curRes
+		} else {
+			if curRes != nil && bestRes.rangeResult != nil {
+				bestResIsSubset := bestRes.rangeResult.Ranges.Subset(sctx.TypeCtx, curRes.rangeResult.Ranges)
+				// if bestRes is a subset then do nothing
+				if !bestResIsSubset {
+					curResIsSubset := curRes.rangeResult.Ranges.Subset(sctx.TypeCtx, bestRes.rangeResult.Ranges)
+					// if curRes is subset (more selective) then make it bestRes.
+					if curResIsSubset {
+						bestRes = curRes
+					} else {
+						// Try intersecting result of different conjuncts.
+						intersection := curRes.rangeResult.Ranges.IntersectRanges(sctx.TypeCtx, bestRes.rangeResult.Ranges)
+						if intersection == nil {
+							if compareCNFItemRangeResult(curRes, bestRes) {
+								bestRes = curRes
+							}
+						} else {
+							bestRes.rangeResult.Ranges = intersection
+							bestRes.rangeResult.AccessConds =
+								AppendConditionsIfNotExist(bestRes.rangeResult.AccessConds, []expression.Expression{cond})
+						}
+					}
+				}
+			}
 		}
 	}
 
--- a/pkg/util/ranger/types.go
+++ b/pkg/util/ranger/types.go
@@ -282,6 +282,88 @@ func formatDatum(d types.Datum, isLeftSide bool) string {
 	return fmt.Sprintf("%v", d.GetValue())
 }
 
+// compareLexicographically compares two bounds from two ranges and returns 0, 1, -1
+// for equal, greater than or less than respectively. It gets the two bounds,
+// collations and if each bound is open or closed. In addition, it also gets what type of bound it is.
+// Lower bounds logically can be extended with -infinity and upper bounds can be extended with +infinity.
+func compareLexicographically(tc types.Context, bound1, bound2 []types.Datum, collators []collate.Collator,
+	open1, open2, low1, low2 bool) (int, error) {
+	n1 := len(bound1)
+	n2 := len(bound2)
+	n := min(n1, n2)
+
+	for i := 0; i < n; i++ {
+		cmp, err := bound1[i].Compare(tc, &bound2[i], collators[i])
+		if err != nil {
+			return 0, err
+		}
+		if cmp != 0 {
+			return cmp, nil
+		}
+	}
+
+	// Handle interval types
+	if n1 == n2 {
+		switch {
+		case !open1 && !open2:
+			return 0, nil
+		case open1 == open2:
+			if low1 == low2 {
+				return 0, nil
+			} else if low1 {
+				return 1, nil
+			} else {
+				return -1, nil
+			}
+		case open1:
+			if low1 {
+				return 1, nil
+			} else {
+				return -1, nil
+			}
+		case open2:
+			if low2 {
+				return -1, nil
+			} else {
+				return 1, nil
+			}
+		}
+	}
+
+	// Unequal length ranges. We use -infinity for lower bounds and +infinity for upper counds.
+	if n1 < n2 {
+		if low1 {
+			// -infinity is less than anything
+			return -1, nil
+		} else {
+			// +infinity is higher than anything
+			return 1, nil
+		}
+	}
+	// n2 > n1
+	if low2 {
+		// anything is larger than -infinity.
+		return 1, nil
+	} else {
+		// anything is less than +infinity
+		return -1, nil
+	}
+}
+
+// Compare two lists of Datum. This is useful for lower/upper bounds of Range.
+func compare(tc types.Context, bound1 []types.Datum, bound2 []types.Datum, length int, collators []collate.Collator, openInterval bool) (int, error) {
+	for i := 0; i < length; i++ {
+		cmp, err := bound1[i].Compare(tc, &bound2[i], collators[i])
+		if err != nil {
+			return 0, err
+		}
+		if cmp != 0 || (cmp == 0 && openInterval) {
+			return cmp, nil
+		}
+	}
+	return 0, nil
+}
+
 // Check if a list of Datum is a prefix of another list of Datum. This is useful for checking if
 // lower/upper bound of a range is a subset of another.
 func prefix(tc types.Context, superValue []types.Datum, supValue []types.Datum, length int, collators []collate.Collator) bool {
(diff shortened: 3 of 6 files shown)


elsa0520:
Please supply the release note on the PR comment 
image

pkg/util/ranger/detacher.go
```diff
@@ -273,8 +273,37 @@ func extractBestCNFItemRanges(sctx *rangerctx.RangerContext, conds []expression.
 			continue
 		}
 		curRes := getCNFItemRangeResult(sctx, res, i)
-		if bestRes == nil || compareCNFItemRangeResult(curRes, bestRes) {
+		if bestRes == nil {
 			bestRes = curRes
+		} else {
+			if curRes != nil && bestRes.rangeResult != nil {
+				if fixcontrol.GetBoolWithDefault(sctx.OptimizerFixControl, fixcontrol.Fix54166, false) {
```
elsa0520, line 280: Should we enable it be default and don't need a control variable ?
It seems that intersection range is actually an always good way ?

ghazalfamilyusa: You are right that the intersection is always better. The control is to minimize the risk initially for the high complexity of the intersection. We should take it out in few sprints and leave the solution as is or add a heuristic per the design for the complex cases (too many ranges to intersect).

ghazalfamilyusa:
> Please supply the release note on the PR comment image

@elsa0520 : this optimization is transparent to the user and not sure if we need a release note.

pkg/util/ranger/types.go
```diff
@@ -329,34 +407,124 @@ func (rs Ranges) Subset(tc types.Context, superRanges Ranges) bool {
...
+//     Thus include checking open/closed inetrvals.
+func (ran *Range) Subset(tc types.Context, otherRange *Range) bool {
+	if len(ran.LowVal) < len(otherRange.LowVal) {
+		return false
+	}
 
-	// Either superRange is closed or both ranges have the same open/close setting.
-	lowExcludeOK := !superRange.LowExclude || ran.LowExclude == superRange.LowExclude
-	highExcludeOK := !superRange.HighExclude || ran.HighExclude == superRange.HighExclude
+	if !checkCollators(ran, otherRange, min(len(otherRange.LowVal), len(ran.LowVal))) {
```
elsa0520, line 434:
```suggestion
	if !checkCollators(ran, otherRange, len(otherRange.LowVal)) {
```

pkg/planner/core/casetest/index/testdata/index_range_in.json
```diff
@@ -19,5 +19,31 @@
...
+  },
+  {
+    "name": "TestRangeIntersection",
+    "cases": [
+	"select /*+ USE_INDEX(t1,PKX) */ count(*) from t1 where ((a1>1) or (a1=1 and b1 > 10));",
+	"select /*+ USE_INDEX(t1,PKX) */ count(*) from t1 where ((a1<10) or (a1=10 and b1 < 20));",
+	"select /*+ USE_INDEX(t1,PKX) */ count(*) from t1 where ((a1>1) or (a1=1 and b1 > 10)) and ((a1<10) or (a1=10 and b1 < 20));",
+	"select /*+ USE_INDEX(t1,PKX) */ count(*) from t1 where (a1,b1) > (1,10) and (a1,b1) < (10,20); -- row format of previous test",
+	"select /*+ USE_INDEX(t1,PKX) */ count(*) from t1 where ((a1>1) or (a1=1 and b1 >= 10)) and ((a1<10) or (a1=10 and b1 <= 20));",
+	"select /*+ USE_INDEX(t1,PKX) */ count(*) from t1 where (a1,b1) >= (1,10) and (a1,b1) <= (10,20); -- row format of previous test",
```
elsa0520, line 31: Could you please add some no intersection range case ? Such as:
``` 
select count(*) from t1 where (a1, b1) > (1, 10) and (a1, b1) < (0, 20) 
select count(*) from t1 where (a1, b1) > (1, 10) and (a1, b1) < (2, 20) and b1 <5
select count(*) from t1 where (a1, b1) > (1, 10) and (a1, b1) < (4, 20) and a1 < 0
```

pkg/util/ranger/detacher.go
ghazalfamilyusa: I would rather keep it off initially and turn it on by default after couple of sprints.

pkg/planner/util/fixcontrol/get.go
```diff
@@ -54,6 +54,8 @@ const (
 	// Fix52869 controls whether to disable the limitation that index merge path won't be generated automatically when
 	// there exist other single-index access paths that do range scan.
 	Fix52869 uint64 = 52869
+	// Fix54166 controls whether to apply ior not range intersection for index access.
```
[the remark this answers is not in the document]
ghazalfamilyusa: done

pkg/planner/util/fixcontrol/get.go
```diff
@@ -54,6 +54,8 @@ const (
...
+	Fix54166 uint64 = 54166
```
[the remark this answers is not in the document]
ghazalfamilyusa: I used the PR number and fixed it to be the issue number.

pkg/util/ranger/types.go
ghazalfamilyusa: done

pkg/planner/core/casetest/index/testdata/index_range_in.json
ghazalfamilyusa: The unit test for intersection covers empty intersections. But, added those as well.

ant
```diff
@@ -0,0 +1 @@
+apache-ant-1.9.16
```
time-and-fate, line 1: Looks like this needs to be removed.

pkg/util/ranger/types.go
```diff
@@ -282,6 +282,84 @@ func formatDatum(d types.Datum, isLeftSide bool) string {
 	return fmt.Sprintf("%v", d.GetValue())
 }
 
+// compareLexicographically compares two bounds from two ranges and returns 0, 1, -1
+// for equal, greater than or less than respectively. It gets the two bounds,
+// collations and if each bound is open or closed. In addition, it also gets what type of bound it is.
+// Lower bounds logically can be extended with -infinity and upper bounds can be extended with +infinity.
+func compareLexicographically(tc types.Context, bound1, bound2 []types.Datum, collators []collate.Collator,
+	open1, open2, low1, low2 bool) (int, error) {
```
time-and-fate, line 290: What does `low1` and `low2` mean here?

pkg/util/ranger/detacher.go
```diff
@@ -273,8 +273,37 @@ func extractBestCNFItemRanges(sctx *rangerctx.RangerContext, conds []expression.
...
+								bestRes.rangeResult.Ranges = intersection
+								bestRes.rangeResult.AccessConds =
+									AppendConditionsIfNotExist(bestRes.rangeResult.AccessConds, []expression.Expression{cond})
+							}
+						}
+					}
+				} else if compareCNFItemRangeResult(curRes, bestRes) {
+					bestRes = curRes
+				}
+			}
```
time-and-fate, line 306: I think we can make the entire logic of getting the best from `curRes` and `bestRes` a separate function.
That should make the new function have a clear input and output, and also reduce the deep indentations, also looks like the repeated `compareCNFItemRangeResult` can be removed.

pkg/planner/util/fixcontrol/get.go
```diff
@@ -54,6 +54,8 @@ const (
 	// Fix52869 controls whether to disable the limitation that index merge path won't be generated automatically when
 	// there exist other single-index access paths that do range scan.
 	Fix52869 uint64 = 52869
+	// Fix16798 controls whether to apply or not range intersection for index access.
+	Fix16798 uint64 = 16798
```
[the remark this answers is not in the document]
ghazalfamilyusa: Ideally, yes but the issue creator wrote it as TiKV issue.

pkg/util/ranger/types.go
```diff
@@ -282,6 +282,84 @@ func formatDatum(d types.Datum, isLeftSide bool) string {
...
+	if low2 {
+		// anything is larger than -infinity.
+		return 1, nil
+	}
+	// anything is less than +infinity
+	return -1, nil
+}
+
+// Compare two lists of Datum. This is useful for lower/upper bounds of Range.
+func compare(tc types.Context, bound1 []types.Datum, bound2 []types.Datum, length int, collators []collate.Collator, openInterval bool) (int, error) {
```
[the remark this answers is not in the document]
ghazalfamilyusa: Good catch. Removed.

ant
ghazalfamilyusa: done

pkg/util/ranger/detacher.go
ghazalfamilyusa: Makes sense. Done

pkg/util/ranger/detacher.go
elsa0520: OK

pkg/planner/core/casetest/index/testdata/index_range_in.json
elsa0520: So, if the condition 'b1<5' doesn't begin as a1, it will not be merged with other condition range in our logic, right ?
But the actually range of "(a1, b1) > (1, 10) and (a1, b1) < (2, 20) and b1 <5"  should be empty.
So, we can't due with this situation right now ?

pkg/planner/util/fixcontrol/get.go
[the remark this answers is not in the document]
elsa0520: Not a big issue. But the better way is create a new tidb issue related the TiKV issue and linked them.
Actually we do have a similar issue #https://github.com/pingcap/tidb/issues/41598 , you can modify this issue slightly and make it the linked issue of your PR

pkg/planner/core/casetest/index/testdata/index_range_in.json
ghazalfamilyusa: Not really. The predicate is actually not empty. The row function is lexicographically ordered and (a1,b1) > (1,10) **does not** mean a1 > 1 and b1 > 10 which makes it with b1 < 5 empty.   (a1,b1) > (1,10) means (a1 > 1) or (a1=1 and b1 > 10) which does not contradict with b1 < 5. But, it does contradict with a1 < 0.

pkg/planner/util/fixcontrol/get.go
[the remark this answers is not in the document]
ghazalfamilyusa: https://github.com/pingcap/tidb/issues/41598 is fixed by my earlier PR and not this one. The linked issue is listed here anyway so let us keep it as is.

elsa0520 (approved):
LGTM

pkg/util/ranger/detacher.go
```diff
@@ -236,6 +236,53 @@ func compareCNFItemRangeResult(curResult, bestResult *cnfItemRangeResult) (curIs
...
+// - Result = intersection of the two ranges
+// - Try heuristic to pick which range is better if intersections fails or if feature is off.
+func mergeTwoCNFRanges(sctx *rangerctx.RangerContext, cond expression.Expression,
+	rangeResult, otherRangeResult *cnfItemRangeResult) *cnfItemRangeResult {
+	if rangeResult == nil {
+		return otherRangeResult
+	}
+	tryHeuristic := false
+	mergedResult := rangeResult
+	if otherRangeResult != nil && mergedResult.rangeResult != nil {
```
time-and-fate, line 254: Looks like we can move this nil check to the beginning, like
```go
if otherRangeResult == nil || rangeResult.rangeResult == nil {
	return rangeResult
}
```

pkg/util/ranger/detacher.go
```diff
@@ -236,6 +236,53 @@ func compareCNFItemRangeResult(curResult, bestResult *cnfItemRangeResult) (curIs
...
+		if fixcontrol.GetBoolWithDefault(sctx.OptimizerFixControl, fixcontrol.Fix54337, false) {
+			mergedResultIsSubset := mergedResult.rangeResult.Ranges.Subset(sctx.TypeCtx, otherRangeResult.rangeResult.Ranges)
+			// if mergedResult is a subset then do nothing
+			if !mergedResultIsSubset {
```
time-and-fate, line 258: Looks like we can directly return `mergedResult` here if `mergedResultIsSubset` is `true`.

pkg/util/ranger/detacher.go
```diff
@@ -236,6 +236,53 @@ func compareCNFItemRangeResult(curResult, bestResult *cnfItemRangeResult) (curIs
...
+				otherRangeResultIsSubset := otherRangeResult.rangeResult.Ranges.Subset(sctx.TypeCtx, mergedResult.rangeResult.Ranges)
+				// if otherRangeResult is subset (more selective) then make it mergedResult.
+				if otherRangeResultIsSubset {
```
time-and-fate, line 261: Ditto. Looks like we can directly return here.

pkg/util/ranger/types.go
```diff
@@ -329,34 +394,124 @@ func (rs Ranges) Subset(tc types.Context, superRanges Ranges) bool {
...
+	}
+	if lowVsLow == -1 {
+		result.LowVal = otherRange.LowVal
+		result.LowExclude = otherRange.LowExclude
+	} else {
+		result.LowVal = ran.LowVal
+		result.LowExclude = ran.LowExclude
+	}
+
+	highVshigh, err := compareLexicographically(tc, ran.HighVal, otherRange.HighVal,
```
time-and-fate, line 483:
```suggestion
	highVsHigh, err := compareLexicographically(tc, ran.HighVal, otherRange.HighVal,
```

pkg/util/ranger/types.go
```diff
@@ -329,34 +394,124 @@ func (rs Ranges) Subset(tc types.Context, superRanges Ranges) bool {
...
+		result.HighExclude = ran.HighExclude
+	}
+	return result, nil
+}
+
+// IntersectRanges computes pairwise intersection between each element in rs and otherRangeList.
+func (rs Ranges) IntersectRanges(tc types.Context, otherRanges Ranges) Ranges {
+	result := Ranges{}
+	for _, rsRange := range rs {
+		for _, otherRange := range otherRanges {
```
time-and-fate, line 502: Ideally, we can order the two `Ranges` and avoid calculating the intersection for every pair between them, right?

pkg/util/ranger/types.go
```diff
@@ -282,6 +282,71 @@ func formatDatum(d types.Datum, isLeftSide bool) string {
...
+			return -1, nil
+		case open2:
+			if low2 {
+				return -1, nil
+			}
+			return 1, nil
+		}
+	}
+
+	// Unequal length ranges. We use -infinity for lower bounds and +infinity for upper counds.
```
time-and-fate, line 332:
```suggestion
	// Unequal length ranges. We use -infinity for lower bounds and +infinity for upper bounds.
```

pkg/util/ranger/detacher.go
ghazalfamilyusa: This comment and subsequent ones are about returning results rather than storing the result with a single return at the end. It is style issue and not sure which one is better. In any case, I would rather keep it like this and avoid the code for the case rangeResult == nil

pkg/util/ranger/detacher.go
ghazalfamilyusa: Same as above

pkg/util/ranger/detacher.go
ghazalfamilyusa: same as above

pkg/util/ranger/types.go
ghazalfamilyusa: Correct and that will be done in subsequent PR when we optimize the intersections and remove the flag.

pkg/util/ranger/types.go
ghazalfamilyusa: OK

pkg/util/ranger/types.go
ghazalfamilyusa: OK

pkg/util/ranger/types.go
```diff
@@ -282,6 +282,71 @@ func formatDatum(d types.Datum, isLeftSide bool) string {
...
+	}
+
+	// Unequal length ranges. We use -infinity for lower bounds and +infinity for upper bounds.
+	if n1 < n2 {
+		if low1 {
+			// -infinity is less than anything
+			return -1, nil
+		}
+		// +infinity is higher than anything
+		return 1, nil
```
time-and-fate, line 339: It seems different `open1` should result in different results.

pkg/util/ranger/types.go
time-and-fate: Could you add some examples for this function here, or in the code below?
I think it's hard for one at first sight to grasp the idea behind `open1`, `open2`, `low1`, and `low2`, and how these values lead to different results.

pkg/util/ranger/types.go
ghazalfamilyusa: it will be the same here if it is open or close because it is -infinity. -infinity is lower than any other value.

pkg/util/ranger/types.go
ghazalfamilyusa: The test cases cover these combinations. I do not think it is a good idea to add examples as comments unless it is something simple and not covering combinations like this. We should have comments in the code and I hope I did that. The logic is no complicated. If one value is not equal to the other in the suffix then it is simple. The extra logic for cases where they are equal in the common length and it goes through cases of n1 == n2, n1 > n2 and n1 < n2. For each case, there is additional logic when the range is open or closed.

ghazalfamilyusa:
for, "https://github.com/pingcap/tidb/pull/54166#discussion_r1662913214"

The test cases cover these combinations. I do not think it is a good idea to add examples as comments unless it is something simple and not covering combinations like this. We should have comments in the code and I hope I did that. The logic is no complicated. If one value is not equal to the other in the suffix then it is simple. The extra logic for cases where they are equal in the common length and it goes through cases of n1 == n2, n1 > n2 and n1 < n2. For each case, there is additional logic for lower/upper bounds.

pkg/util/ranger/types.go
time-and-fate: It becomes a bit complicated here and I think it relates to the definition of what a range with fewer columns means.
I come up with one example: `[1 1,1 5) [1 5 1,1 6 5]`. I think `1 5)` should be less than `[1 5 1` because the first range should come from expressions like `a = 1 and b >= 1 and b < 5`.

time-and-fate: Looks like the key issue here is that does `[1 1,1 5)` mean `a = 1 and b >= 1 and b < 5` or `a = 1 and b >= 1 and b < 6`.

pkg/util/ranger/types.go
```diff
@@ -282,6 +282,71 @@ func formatDatum(d types.Datum, isLeftSide bool) string {
...
+	}
+	// n2 > n1
```
time-and-fate, line 341:
```suggestion
	// n2 < n1
```

pkg/util/ranger/types_test.go
```diff
@@ -238,3 +252,366 @@ func TestRangeMemUsage(t *testing.T) {
...
+		Collators: collate.GetBinaryCollatorSlice(3),
+	}
+	intersection1, _ = range1.IntersectRange(ctx.GetRangerCtx().TypeCtx, &range2)
+	actualResult[range1.String()+" "+range2.String()] = rangeToString(intersection1)
+	intersection2, _ = range2.IntersectRange(ctx.GetRangerCtx().TypeCtx, &range1)
+	require.Equal(t, intersection1, intersection2)
+
+	// Compare actual vs expected
+	for k, v := range actualResult {
+		require.Equal(t, v, expectedResult[k])
```
time-and-fate, line 613: The expected should be the first argument.
```suggestion
		require.Equal(t, expectedResult[k], v)
```

ghazalfamilyusa: OK. done

pkg/util/ranger/types.go
ghazalfamilyusa: OK and done

pkg/util/ranger/types.go
ghazalfamilyusa: The example above is not applicable, You are using a general n dimensional range and not the lexiographically ordered index. [1 5 1,1 6 5] is not a valid index range. All the prefix should be single point. The open and closed is applied to all the columns. So do you mean (1 5 compared to [1 5 1, yes it is less than and consistent with the code.

ghazalfamilyusa: [1 1,1 5) mean a = 1 and b >= 1 and b < 5

time-and-fate: > [1 5 1,1 6 5] is not a valid index range

This can be a valid range. If the user writes`(a,b,c) >= (1,5,1) and (a,b,c) <= (1,6,5)`, that's exactly what we want.

> If you mean comparing the upper bound 1 5) to lower bound [1 5 1 then you are comparing 1 5 +infinity) and [1 5 1 and 1 5 +infinity) is higher.

If `[1 1,1 5)` means `a = 1 and b >= 1 and b < 5`, then `1 5)` means `1 5 -infinity)` instead of `1 5 +infinity)`, right? Because `a = 1 and b = 5 and c = -999` doesn't satisfy `a = 1 and b >= 1 and b < 5`.
Then `1 5)` should be lower than `[1 5 1`.

ghazalfamilyusa: 1 5) as an upper bound means a=1 and b < 5 and c anything which means c < +infinity. If you take a simple example of [1 5, 1 5] it means a=1 and b=1 and all values of c which means c >= -infinity for low range and c <= +infinity for the upper bound.

time-and-fate (approved):

Issue #53634, opened 2024-05-28
The real data type does not match the current column type


zimulala:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

daily run in TiFlash
**DDLs:**
```
[sql="\nCREATE TABLE IF NOT EXISTS stock (\n\ts_i_id INT NOT NULL,\n\ts_w_id INT NOT NULL,\n\ts_quantity INT,\n\ts_dist_01 CHAR(24), \n\ts_dist_02 CHAR(24),\n\ts_dist_03 CHAR(24),\n\ts_dist_04 CHAR(24), \n\ts_dist_05 CHAR(24), \n\ts_dist_06 CHAR(24), \n\ts_dist_07 CHAR(24), \n\ts_dist_08 CHAR(24), \n\ts_dist_09 CHAR(24), \n\ts_dist_10 CHAR(24), \n\ts_ytd INT, \n\ts_order_cnt INT, \n\ts_remote_cnt INT,\n\ts_data VARCHAR(50),\n\tPRIMARY KEY(s_w_id, s_i_id)\n)"]
[sql="alter table `tpcc`.`stock` set tiflash replica 2"]
[sql="alter table `tpcc`.`stock` set tiflash replica 2"]
[sql="alter table stock drop column cct_1"]
[sql="alter table stock add column cct_1 int default 10"]
[sql="alter table stock modify cct_1 json"]
[sql="alter table stock drop column adc_1"]
[sql="alter table stock add column adc_1 smallint"]
[sql="alter table stock drop column cct_1"]
```

**DMLs:**
Execute the following DMLs when finished `alter table stock add column adc_1 smallint`.

`["wait latest schema version changed(get the metadata lock if tidb_enable_metadata_lock is true)"] [category=ddl] **[ver=237]** ["take time"=127.144578ms] [job="ID:203, Type:add column, State:done, SchemaState:public, SchemaID:104, TableID:120, RowCount:0, ArgLen:4, start time: 2024-05-22 11:35:43.835 +0800 CST, ts:449930577532682272, Err:<nil>, ErrCount:0, SnapshotVersion:0, LocalMode: false"]`

```
[...]

PR #51897, opened 2024-03-19
*: add metadata lock when using the plan cache

wjhuang2016:
### What problem does this PR solve?

Issue Number: close #51407

Problem Summary:
When executing the prepared statement from the plan cache, tidb doesn't add the metadata lock, which may cause data-index inconsistency. 

### What changed and how does it work?
1. Collect the related table information when `PREPARE`.
2. Add metadata lock during planCachePreprocess. If the schema is changed, can't use the cache plan anymore.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
fix a bug that metadata lock is missing when using the plan cache
```

PR #54183, opened 2024-06-24
*: fix a bug that update statement uses point get and update plan with different tblInfo

zimulala:

Issue Number: close https://github.com/pingcap/tidb/issues/53634

Problem Summary:
#### Case
Init SQLs：
```
CREATE TABLE stock ( a int NOT NULL, b char(30) NOT NULL,  c int, d char(64), PRIMARY KEY(a,b)) ;
insert into stock values(1, 'a', 11, 'x'), (2, 'b', 22, 'y');
alter table stock add column cct_1 int default 10;
alter table stock modify cct_1 json;
alter table stock add column adc_1 smallint;
```

* Step1. using conn1
prepare statements:
```
    begin;
    SELECT a, c, d from stock where (a, b) IN ((?, ?),(?, ?)) FOR UPDATE;
    UPDATE stock SET c = ? WHERE a= ? AND b = 'a';
    UPDATE stock SET c = ?, d = 'z' WHERE a= ? AND b = 'b';
    commit;
```
run a statement:
    `begin;`

* Step2. using conn2
 do DDL:
```
    alter table stock drop column cct_1;
```

* Step3. using conn1
When the DDL is in Write-Only state
exec statements:
```
    SELECT a, c, d from stock where (a, b) IN ((?, ?),(?, ?)) FOR UPDATE;
    UPDATE stock SET c = ? WHERE a= ? AND b = 'a';
    UPDATE stock SET c = ?, d = 'z' WHERE a= ? AND b = 'b';
    commit;
```

* Step4. using conn1
Check the result
`select * from stock;`

* Statement execution order table 

| conn1                       | conn2                                                                    |
|-----------------------------|--------------------------------------------------------------------------|
| step1. prepare stmts        |                                                                          |
| step1. begin;               |                                                                          |
|                             | step2. alter table stock drop column cct_1; (state public -> write-only) |
| step3. exec stmts           |                    write-only                                     |
|                             | step2. alter table stock drop column cct_1; (finish)                     |
| step4. select * from stock; |                                                                          |

#### Conclusion
The update statement in step3 uses point get and update plan, but the tblInfo used by the two plans is inconsistent, resulting in incorrect data in real storage.

* After executing step3. select statement, the stock is locked in `GetRelatedTableForMDL`, so `stmt.tbls[i].Meta().Revision != newTbl.Meta().Revision` is false. It means `schemaNotMatch` is false.  So we needn't to `Preprocess`. 
https://github.com/pingcap/tidb/blob/d5b89f8bccdf13cf664ba8bb65a87b7342676d49/pkg/planner/core/plan_cache.go#L114-L119

* Step3. update statement using `tblName.TableInfo`(get it when preparing statements, `cct_1` is public) in `newPointGetPlan`. 
https://github.com/pingcap/tidb/blob/d5b89f8bccdf13cf664ba8bb65a87b7342676d49/pkg/planner/core/point_get_plan.go#L1317

*  Step3. update statement using `t` gets from `is.TableByID(tbl.ID)`( `cct_1` is write-only)
https://github.com/pingcap/tidb/blob/d5b89f8bccdf13cf664ba8bb65a87b7342676d49/pkg/planner/core/point_get_plan.go#L1945-L1950

Add the `Revision` field comparison of `tbl`(get from txn infoschema) and `newTbl` to confirm whether reprocess is required.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/plan_cache.go
+++ b/pkg/planner/core/plan_cache.go
@@ -101,7 +101,7 @@ func planCachePreprocess(ctx context.Context, sctx sessionctx.Context, isNonPrep
 	// step 3: add metadata lock and check each table's schema version
 	schemaNotMatch := false
 	for i := 0; i < len(stmt.dbName); i++ {
-		_, ok := is.TableByID(stmt.tbls[i].Meta().ID)
+		tbl, ok := is.TableByID(stmt.tbls[i].Meta().ID)
 		if !ok {
 			tblByName, err := is.TableByName(stmt.dbName[i], stmt.tbls[i].Meta().Name)
 			if err != nil {
@@ -116,7 +116,7 @@ func planCachePreprocess(ctx context.Context, sctx sessionctx.Context, isNonPrep
 			schemaNotMatch = true
 			continue
 		}
-		if stmt.tbls[i].Meta().Revision != newTbl.Meta().Revision {
+		if stmt.tbls[i].Meta().Revision != newTbl.Meta().Revision || tbl.Meta().Revision != newTbl.Meta().Revision {
 			schemaNotMatch = true
 		}
 		stmt.tbls[i] = newTbl
--- a/pkg/planner/core/plan_cache.go
+++ b/pkg/planner/core/plan_cache.go
@@ -116,6 +116,11 @@ func planCachePreprocess(ctx context.Context, sctx sessionctx.Context, isNonPrep
 			schemaNotMatch = true
 			continue
 		}
+		// The revision of tbl and newTbl may not be the same.
+		// Example:
+		// The version of stmt.tbls[i] is taken from the prepare statement and is revision v1.
+		// When stmt.tbls[i] is locked in MDL, the revision of newTbl is also v1.
+		// The revision of tbl is v2. The reason may have other statements trigger "tryLockMDLAndUpdateSchemaIfNecessary" before, leading to tbl revision update.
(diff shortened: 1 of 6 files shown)


wjhuang2016 (approved):

JaySon-Huang:
Great to see the root cause is located and fixed! :)

pkg/planner/core/plan_cache.go
```diff
@@ -116,7 +116,7 @@ func planCachePreprocess(ctx context.Context, sctx sessionctx.Context, isNonPrep
 			schemaNotMatch = true
 			continue
 		}
-		if stmt.tbls[i].Meta().Revision != newTbl.Meta().Revision {
+		if stmt.tbls[i].Meta().Revision != newTbl.Meta().Revision || (tbl != nil && tbl.Meta().Revision != newTbl.Meta().Revision) {
```
JaySon-Huang, line 119: IMO, add some comments about why not only `stmt.tbls` but also `tbl` need to be check is better

zimulala: This scene is quite difficult to describe, you see my PR description written quite a lot. Do you have any suggestions?

qw4990 (approved):

zimulala:
Related issue: https://github.com/pingcap/tidb/pull/51897

PR #54185, opened 2024-06-24
*: optimize the speed of full load

wjhuang2016:

Issue Number: ref #50959

Problem Summary:
Full load is slow if there are too many tables.

Use regexp to filter the special table and get the table name to ID mapping.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/meta/meta.go
```diff
@@ -1004,6 +1006,57 @@ func (m *Meta) GetMetasByDBID(dbID int64) ([]structure.HashPair, error) {
 	return res, nil
 }
 
+// CheckSpecialAttributes checks if the special attributes are in the table info.
+// Make it same as hasSpecialAttributes.
+// Exported for testing.
+func CheckSpecialAttributes(str string) bool {
+	if strings.Contains(str, "\"tiflash_replica\":null") && strings.Contains(str, "\"policy_ref_info\":null") &&
```
lance6716, line 1013: sometimes `str` is large, not sure it will run faster using this https://pkg.go.dev/index/suffixarray#New . Or we have prior knowledge about the order of these 4 substring so we can use strings/bytes.Index for one matching and use the index to skip the prefix when continue next matching. bytes package may be more suitable if strings package will copy and generate intermediate string objects.

pkg/meta/meta.go
```diff
@@ -1004,6 +1006,57 @@ func (m *Meta) GetMetasByDBID(dbID int64) ([]structure.HashPair, error) {
...
+
+// GetAllNameToIDAndSpecialAttributeInfo gets all the fields and values and table info for special attributes in a hash.
+func GetAllNameToIDAndSpecialAttributeInfo(m *Meta, dbID int64) (map[string]int64, []*model.TableInfo, error) {
+	dbKey := m.dbKey(dbID)
+	if err := m.checkDBExists(dbKey); err != nil {
+		return nil, nil, errors.Trace(err)
+	}
+
+	res := make(map[string]int64)
+	idRegex := regexp.MustCompile(`"id":(\d+)`)
```
lance6716, line 1028: can move them as global variables

pkg/meta/meta.go
```diff
@@ -1004,6 +1006,57 @@ func (m *Meta) GetMetasByDBID(dbID int64) ([]structure.HashPair, error) {
...
+	tableInfos := make([]*model.TableInfo, 0)
+
+	err := m.txn.IterateHash(dbKey, func(field []byte, value []byte) error {
+		if !strings.HasPrefix(string(hack.String(field)), "Table") {
+			return nil
+		}
+
+		idMatch := idRegex.FindStringSubmatch(string(hack.String(value)))
+		nameLMatch := nameLRegex.FindStringSubmatch(string(hack.String(value)))
+		id, err := strconv.Atoi(idMatch[1])
```
lance6716, line 1040: please add comment or unit tests to explain why use `[1]`

lance6716 (commented):
5 / 10 files viewed

pkg/meta/meta.go
```diff
@@ -1012,6 +1014,68 @@ func (m *Meta) GetMetasByDBID(dbID int64) ([]structure.HashPair, error) {
 	return res, nil
 }
 
+// CheckSpecialAttributes checks if the special attributes are in the table info.
+// Make it same as hasSpecialAttributes.
+// Exported for testing.
+func CheckSpecialAttributes(str string) bool {
+	if strings.Contains(str, "\"tiflash_replica\":null") && strings.Contains(str, "\"policy_ref_info\":null") &&
```
tiancaiamao, line 1021: Will the json object skip displaying the null field?
For example, the table info has not partition, so instead of an explicit "partition:null", there is nothing.

wjhuang2016: No, it's tested and I don't think it would change.

tiancaiamao (approved):

pkg/meta/meta.go
lance6716: please address this comment. Do you think it's negligible or will change it?

lance6716 (approved):

winoros (approved):

pkg/parser/model/model.go
```diff
@@ -1687,13 +1687,14 @@ func (fk *FKInfo) Clone() *FKInfo {
...
-	State              SchemaState    `json:"state"`
-	PlacementPolicyRef *PolicyRefInfo `json:"policy_ref_info"`
+	ID                 int64            `json:"id"`      // Database ID
+	Name               CIStr            `json:"db_name"` // DB name.
+	Charset            string           `json:"charset"`
+	Collate            string           `json:"collate"`
+	Tables             []*TableInfo     `json:"-"` // Tables in the DB.
+	State              SchemaState      `json:"state"`
+	PlacementPolicyRef *PolicyRefInfo   `json:"policy_ref_info"`
+	Name2ID            map[string]int64 `json:"-"`
```
D3Hunter, line 1697: tableName 2 id? if it's true, i prefer rename it as is.

pkg/meta/meta.go
```diff
@@ -1012,6 +1014,68 @@ func (m *Meta) GetMetasByDBID(dbID int64) ([]structure.HashPair, error) {
...
+const NameExtractRegexp = `"L":"([^"\\]*(?:\\.[^"\\]*)*)"}`
+
+// Unescape is exported for testing.
+func Unescape(s string) string {
+	s = strings.ReplaceAll(s, `\"`, `"`)
+	s = strings.ReplaceAll(s, `\\`, `\`)
+	return s
+}
+
+// GetAllNameToIDAndSpecialAttributeInfo gets all the fields and values and table info for special attributes in a hash.
```
D3Hunter, line 1038: please comment on why we do it using this way, and future plans, it's more looks like a hacky optimization to use regex to check a structured json.

D3Hunter (commented):
rest lgtm

windtalker (approved):
LGTM

D3Hunter (approved):

PR #54211, opened 2024-06-25
lighting: lightning configuration tidb.tls = false should not affect cluster certificate (PD's connection) 

### LINKED ISSUE #54172: lightning configuration `tidb.tls = false` should not affect cluster certificate (PD's connection)
When user only set `security` and not set `tidb.security`, TiDB connection's certificate will fallback to cluster certificate. However, TiDB connection has a separate `tidb.tls` to control other TLS behaviours. We should not let it affects cluster certificate, like in the case when fallack TiDB has `tidb-tls = false` but PD needs TLS

https://github.com/pingcap/tidb/blob/f57d2b9d379b188e46b187bf28db7ce30e67f0d5/pkg/lightning/config/config.go#L177-L179

ei-sugimoto:

Issue Number: close #54172

Problem Summary:
Based on the issue, my suggestion to fix the problem is to treat the `d.Security` and `d.TLS` settings separately in the `DBStore.adjust` function. Specifically, the `d.Security` setting should not be modified by the `d.TLS` value, so that even if the `d.TLS` setting is `false`, the cluster certificate (`s.CAPath`, `s.CertPath`, `s.KeyPath`, etc.) is Security to be set to `d.Security`. This ensures that the TLS setting for the TiDB connection does not affect the cluster certificate setting.

Translated with DeepL.com (free version)

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/lightning/config/config.go
+++ b/pkg/lightning/config/config.go
@@ -176,14 +176,12 @@ func (d *DBStore) adjust(
 
 	if d.Security == nil {
 		d.Security = &Security{
-			CAPath:                   s.CAPath,
-			CertPath:                 s.CertPath,
-			KeyPath:                  s.KeyPath,
-			CABytes:                  s.CABytes,
-			CertBytes:                s.CertBytes,
-			KeyBytes:                 s.KeyBytes,
-			RedactInfoLog:            s.RedactInfoLog,
-			AllowFallbackToPlaintext: s.AllowFallbackToPlaintext,
+			CAPath:    s.CAPath,
+			CertPath:  s.CertPath,
+			KeyPath:   s.KeyPath,
+			CABytes:   s.CABytes,
+			CertBytes: s.CertBytes,
+			KeyBytes:  s.KeyBytes,
 		}
 	}
 
--- a/pkg/lightning/config/config.go
+++ b/pkg/lightning/config/config.go
@@ -175,7 +175,16 @@ func (d *DBStore) adjust(
 	}
 
 	if d.Security == nil {
-		d.Security = s
+		d.Security = &Security{
+			CAPath:                   s.CAPath,
+			CertPath:                 s.CertPath,
+			KeyPath:                  s.KeyPath,
+			CABytes:                  s.CABytes,
+			CertBytes:                s.CertBytes,
+			KeyBytes:                 s.KeyBytes,
+			RedactInfoLog:            s.RedactInfoLog,
+			AllowFallbackToPlaintext: s.AllowFallbackToPlaintext,
+		}
 	}
 
 	switch d.TLS {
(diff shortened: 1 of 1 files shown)


pkg/lightning/config/config.go
```diff
@@ -200,12 +207,6 @@ func (d *DBStore) adjust(
 	case "":
 	case "false":
 		d.Security.TLSConfig = nil
-		d.Security.CAPath = ""
```
lance6716, line 203: We should keep them. Because the real effective object is `TLSConfig`, and it's built from these fields in https://github.com/pingcap/tidb/blob/761b5926a73d15ef165f525fdada5a0f7fa3e221/lightning/pkg/server/lightning.go#L394 and https://github.com/pingcap/tidb/blob/761b5926a73d15ef165f525fdada5a0f7fa3e221/lightning/pkg/server/lightning.go#L503, while this `adjust` is called at https://github.com/pingcap/tidb/blob/761b5926a73d15ef165f525fdada5a0f7fa3e221/lightning/pkg/server/lightning.go#L355 which is before `run`. So only assigning `Security.TLSConfig` to `nil` is not enough, because `BuildTLSConfig` will re-create the `TLSConfig`

ei-sugimoto: @lance6716 Thanks for the review.
I was mistaken. I should have discussed this issue before I started working on it.
I have two questions.
Is TLS always enabled for connections to PD?
Do we need to add a CAPath property dedicated to connections to TiDB to solve this problem?

lance6716: > Is TLS always enabled for connections to PD?

it's controlled by `[security]` section in configuration file

> Do we need to add a CAPath property dedicated to connections to TiDB to solve this problem?

We already have it, it's `[tidb.security]`. And when `[tidb.security]` is not set, the fallback value is `[security]`. The original issue is related to `tidb.tls` configuration which will modify the behaviour of `[tidb.security]`. But when fallback happens, we don't want `tidb.tls` affects `[security]`.

https://docs.pingcap.com/tidb/stable/tidb-lightning-configuration

pkg/lightning/config/config_test.go
```diff
@@ -445,6 +445,17 @@ func TestAdjustSecuritySection(t *testing.T) {
 			hasTLS:         false,
 			fallback2NoTLS: false,
 		},
+		{
+			input: `
+				[security]
+				ca-path = "/path/to/ca2.pem"
```
lance6716, line 451: can you also add a unit test to check `security.ca-path` (`cfg.Security`) is unchanged? The new unit test should fail with the old code and should pass with this new PR.

ei-sugimoto: https://github.com/pingcap/tidb/pull/54211/files#diff-ebcda57fdf1bea10676a21181bc388b6dd8f95c8a2013438fd826cfec040e4c7R479-R490
I add a unit test. 
I did a comparison of the ca-path before and after adjusting, was this the right way to do it?

lance6716: I'll check it later. We have some flaky tests and they causes the CI failure, don't worry.

lance6716 (approved):

lyzx2001:
pls fix unit test

lyzx2001 (commented):
pls fix unit test

lyzx2001:
> pls fix unit test

Sorry I missed the comment here
https://github.com/pingcap/tidb/pull/54211#discussion_r1660650080
You can ignore the previous comment

lyzx2001 (approved):

PR #54279, opened 2024-06-27
*: make changes to tidb_schema_cache_size effective immediately

tangenta:

Issue Number: ref #50959

Problem Summary:

Previously, if we change `tidb_schema_cache_size` to a different number, we still need to execute something(like DDL) to trigger `loadInfoSchema` to make it effective.

This PR makes changes to tidb_schema_cache_size effective immediately.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
  image
  ```
  mysql> set global tidb_schema_cache_size = 1000;
  
  mysql> set global tidb_schema_cache_size = 10000;
  Query OK, 0 rows affected (0.04 sec)
  
  mysql> set global tidb_schema_cache_size = 100000;
  Query OK, 0 rows affected (0.04 sec)
  ```
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/infoschema/sieve.go
+++ b/pkg/infoschema/sieve.go
@@ -76,10 +76,13 @@ func (s *Sieve[K, V]) SetCapacity(capacity uint64) {
 }
 
 func (s *Sieve[K, V]) SetCapacityAndWaitEvict(capacity uint64) {
-	s.mu.Lock()
-	defer s.mu.Unlock()
+	s.SetCapacity(capacity)
 	for s.size > s.capacity {
-		s.evict()
+		s.mu.Lock()
+		for i := 0; s.size > s.capacity && i < 10; i++ {
+			s.evict()
+		}
+		s.mu.Unlock()
 	}
 }
 
--- a/pkg/infoschema/sieve.go
+++ b/pkg/infoschema/sieve.go
@@ -75,6 +75,14 @@ func (s *Sieve[K, V]) SetCapacity(capacity uint64) {
 	s.capacity = capacity
 }
 
+func (s *Sieve[K, V]) SetCapacityAndWaitEvict(capacity uint64) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	for s.size > s.capacity {
+		s.evict()
+	}
+}
+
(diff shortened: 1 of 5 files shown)


pkg/infoschema/sieve.go
```diff
@@ -75,6 +75,14 @@ func (s *Sieve[K, V]) SetCapacity(capacity uint64) {
 	s.capacity = capacity
 }
 
+func (s *Sieve[K, V]) SetCapacityAndWaitEvict(capacity uint64) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	for s.size > s.capacity {
+		s.evict()
```
tangenta, line 82: This may block the access to cache if the difference is large, is there any suggestion for this risk?

lance6716: maybe use a loop and in each round, acuqire the lock, evict some number of items (like in line 102 ~ 104), release the lock.

wjhuang2016 (approved):

pkg/infoschema/sieve.go
```diff
@@ -75,6 +75,17 @@ func (s *Sieve[K, V]) SetCapacity(capacity uint64) {
 	s.capacity = capacity
 }
 
+func (s *Sieve[K, V]) SetCapacityAndWaitEvict(capacity uint64) {
```
lance6716, line 78: I'm not sure if there will be two call, one wants to increase capacity and the other wants to decrease. Maybe we should add another lock to make sure `SetCapacityAndWaitEvict` is not concurrently executing.

lance6716:
rest lgtm

pkg/infoschema/sieve.go
tangenta: Do you mean the access of `s.size` and `s.capacity` is not protected by lock?

lance6716 (approved):

tiancaiamao:
This works when it's infoschema v2 already ...
But in case it's infoschema v1 currently, that change does not make it run v2 immediately.

PR #54420, opened 2024-07-03
analyze: fix AnalyzeV2 with special global index

### LINKED ISSUE #54233: Statistics cannot be generated correctly for a global index that has an expression index
## Bug Report

```
set tidb_enable_global_index=true;
drop table if exists t;
CREATE TABLE `t` (
  `a` int(11) DEFAULT NULL,
  `b` int(11) DEFAULT NULL,
  `c` int(11) DEFAULT NULL,
  UNIQUE KEY `idx` ((`c` + 1)),
  UNIQUE KEY `idx1` (`c`)
PARTITION BY HASH (`b`) PARTITIONS 2;
insert into t values (1,1,1),(2,2,2),(3,3,3),(4,4,4),(5,5,5);
analyze table t;
explain select c+1 from t where (c+1)>3;
```

```
mysql> explain select c+1 from t where (c+1)>3;
+------------------------+---------+-----------+------------------------------+--------------------------------------------------------------------------+
| id                     | estRows | task      | access object                | operator info                                                            |
+------------------------+---------+-----------+------------------------------+--------------------------------------------------------------------------+
| IndexReader_6          | 3.00    | root      |                              | index:IndexRangeScan_5                                                   |
| └─IndexRangeScan_5     | 3.00    | cop[tikv] | table:t, index:idx(`c` + 1) | range:(3,+inf], keep order:false, stats:partial[_v$_idx_0:unInitialized] |
+------------------------+---------+-----------+------------------------------+--------------------------------------------------------------------------+
2 rows in set (0.01 sec)
```

```
mysql> explain select c+1 from t where (c+1)>3;
+------------------------+---------+-----------+-----------------------------+--------------------------------------------------------------------------+
| id                     | estRows | task      | access object               | operator info                                                            |
+------------------------+---------+-----------+-----------------------------+--------------------------------------------------------------------------+
| IndexReader_7          | 2.08    | root      | partition:all               | index:IndexRangeScan_5                                                   |
| └─IndexRangeScan_5     | 2.08    | cop[tikv] | table:t, index:idx(`c` + 1) | range:(3,+inf], keep order:false, stats:partial[_v$_idx_0:unInitialized] |
+------------------------+---------+-----------+-----------------------------+--------------------------------------------------------------------------+
2 rows in set (0.01 sec)
```

The estRows are never changed, no matter what conditions are used to query, e.g. `explain select c+1 from t where (c+1)>0;`

Defined2014:

Issue Number: close #54233

Problem Summary: Supports processing global index in analyze version V2. Because the data of the global index is not scattered in every partitions, only one analyze is required for the whole table.

Because we use an independent index analyze task, there will be results of the same version when saving, which needs to be processed like mv index.

After the analysis is complete, the global index will only have global stats.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/analyze_idx.go
```diff
@@ -88,8 +88,8 @@ func analyzeIndexPushdown(idxExec *AnalyzeIndexExec) *statistics.AnalyzeResults
 		Count:    cnt,
 		Snapshot: idxExec.snapshot,
 	}
-	if idxExec.idxInfo.MVIndex {
-		result.ForMVIndex = true
+	if idxExec.idxInfo.MVIndex || (idxExec.idxInfo.Global && statsVer == statistics.Version2) {
```
hi-rustin, line 91: May I ask what is the current behavior when the stats version is 1?

pkg/planner/core/planbuilder.go
```diff
@@ -2181,6 +2181,10 @@ func getModifiedIndexesInfoForAnalyze(
 		if originIdx.State != model.StatePublic {
 			continue
 		}
+		// For global index, we also need to analyze it as independent index analyze task.
```
hi-rustin, line 2184: Please update the comments above this function.

pkg/planner/core/planbuilder.go
```diff
@@ -2327,6 +2331,13 @@ func (b *PlanBuilder) buildAnalyzeFullSamplingTask(
 		}
 	}
 
+	// Generate global indexes idxTask.
+	for _, indexInfo := range tbl.TableInfo.Indices {
```
hi-rustin, line 2335: How about we collect this information in the getModifiedIndexesInfoForAnalyze. Because in that function we:
1. collect the normal indexes
2. skip and collect the MVIndexes
3. skip the global index

How about we also collect the global indexes in this function? Then we don't need to iterate the indexes again. And also helps us do the index check in one place.

pkg/statistics/analyze.go
```diff
@@ -108,7 +108,10 @@ type AnalyzeResults struct {
 	// take care of those table-level fields.
 	// In conclusion, when saving the analyze result for mv index, we need to store the index stats, as for the
 	// table-level fields, we only need to update the version.
-	ForMVIndex bool
```
hi-rustin, line 111: I believe there are still some comments using `ForMVIndex`.

hi-rustin (commented):
Thanks! 🤟 🖤

I have some questions:
1. What happens if the stats version is 1?
2. What happens if we only analyze a few partitions of one table in one analyze statement? Do we analyze the global index every time? What happens if the partition's stats are outdated, but the global index stats are updated?

I would like to ask @winoros and @time-and-fate those experts to help review this PR. 👀

Defined2014:
> Thanks! 🤟 🖤
> 
> I have some questions(not only for you, I am still learning, so any comments from anyone all help):
> 
> 1. What happens if the stats version is 1?
> 2. What happens if we only analyze a few partitions of one table in one analyze statement? Do we analyze the global index every time? What happens if the partition's stats are outdated, but the global index stats are updated?
> 
> I would like to ask @winoros and @time-and-fate those experts to help review this PR. 👀

1. For stats version 1, it also calls `generateIndexTasks` funtion to generate global index tasks. Except for the key range, everything else is the same as the normal index, so there is not much difference.

3. We will analyze the global index every time.

pkg/executor/analyze_idx.go
Defined2014: The `ForMVIndex` flag is used in update `stats_meta` table. And for stats version V1, it should update `stats_meta` table directly no matter it already have a record or not.

winoros:
QQ_1720533123912
For example, the global scope is conflicted with `for each partition`.

pkg/statistics/handle/autoanalyze/priorityqueue/calculator.go
```diff
@@ -56,6 +56,13 @@ func (pc *PriorityCalculator) CalculateWeight(job AnalysisJob) float64 {
 	// when we apply the log10 function, the resulting value is more meaningful and reasonable.
 	indicators := job.GetIndicators()
 	changeRatio := 100 * indicators.ChangePercentage
+	if j, ok := job.(*DynamicPartitionedTableAnalysisJob); ok {
```
hi-rustin, line 59: This is so hack. I think we should consider introducing a new job for the global index instead of lacking this information here.

pkg/statistics/handle/autoanalyze/priorityqueue/calculator.go
```diff
@@ -56,6 +56,13 @@ func (pc *PriorityCalculator) CalculateWeight(job AnalysisJob) float64 {
...
+		tblSize := indicators.TableSize * (1 + float64(len(j.SpecialGlobalIndexes)))
```
hi-rustin, line 60: I am not sure why we need to use `indicators.TableSize(rows * cols)` * (1 + SpecialGlobalIndexes) here. I guess it would be rows * (normal cols count + SpecialGlobalIndexes count)?

pkg/statistics/handle/autoanalyze/autoanalyze.go
hi-rustin, line 1: I am not sure if we really need to change this old implementation anymore, because we have already informed users that it will be deprecated soon.

pkg/statistics/handle/autoanalyze/priorityqueue/calculator.go
hi-rustin: Or at least keep it in the refresher.

hi-rustin (commented):
Thanks!

PR #54456, opened 2024-07-04
context, executor: add a function to detach the `TableReaderExecutor`

YangKeao:

Issue Number: close https://github.com/pingcap/tidb/issues/53336

Problem Summary:

It's not safe to call `Next` (or other method) of an executor after the session is executing the next statement. We have split out a lot of `Context` and it's clear that the context depended by the `TableReaderExecutor` is safe to be re-used (after proper configuration) while executing another statement is running on the session.

This PR adds a `Detach` function to the executor package. After calling `Detach`, the session can be used to run other statements. It depends on the following assumption:

- The `StmtCtx` will **not** be re-used. (For cursor-fetch scenario, it guaranteed by the logic inside `ResetContextOfStmt`).

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```


pkg/expression/contextstatic/exprctx.go
```diff
@@ -171,8 +163,9 @@ func NewStaticExprContext(opts ...StaticExprCtxOption) *StaticExprContext {
 		},
 	}
 
-	ctx.canUseCache = &atomic.Bool{}
-	ctx.canUseCache.Store(true)
+	cacheTracker := contextutil.NewPlanCacheTracker(ctx.evalCtx)
+	ctx.planCacheTracker = &cacheTracker
+	ctx.planCacheTracker.EnablePlanCache()
```
lcwangchao, line 168: How about to move this assign range apply opts:

```
if ctx. planCacheTracker == nil {
	cacheTracker := contextutil.NewPlanCacheTracker(ctx.evalCtx)
	ctx.planCacheTracker = &cacheTracker
	ctx.planCacheTracker.EnablePlanCache()
}
```

pkg/executor/detach.go
```diff
@@ -0,0 +1,74 @@
...
+	copy(newExecutor.AllChildren(), newChildren)
+
+	return newExecutor, true
+}
+
+func (treCtx tableReaderExecutorContext) Detach() tableReaderExecutorContext {
+	newCtx := treCtx
+
+	if ctx, ok := treCtx.ectx.(*contextsession.SessionExprContext); ok {
+		staticExprCtx := ctx.IntoStatic()
```
lcwangchao, line 54: Is it possible to make `IntoStatic` a static function and receives `exprctx.BuildContext` as input? So that we will need to try to cast the expr ctx to `SessionExprContext`

YangKeao: It's a little weird. I think it'd be better to announce that it can only handle the `SessionExprContext`, as we may have many other implementations for `exprctx.BuildContext`.

For example, the `NullRejectCheckExprContext { *SessionExprContext }` also implements the `exprctx.BuildContext`. If we are going to implement a standalone static `IntoStatic` function, it'll have to handle this case. It's not easy to maintain all possible implementations in that function (though the only possible branch is the `*contextsession.SessionExprContext`).

pkg/expression/contextstatic/exprctx.go
YangKeao: Done.

lcwangchao (approved):

qw4990 (approved):

zanmato1984:
@XuHuaiyu PTAL, thanks.

tangenta (approved):

XuHuaiyu (approved):

PR #54480, opened 2024-07-05
statistics: avoid frequantly syncing stats simultaneously

### LINKED ISSUE #54481: avoid frequantly syncing stats simultaneously
## Enhancement

Now, ```StatsCacheImpl.Update``` can sync stats by version from the storage. we have many scenary to call it. such as ```domain.loadStatsWorker``` and ```Analyze```. When customers use lightning, lightning will trigger the multi-analyzed tasks simultaneously. so it also trigger multi ``````StatsCacheImpl.Update``` simultaneously. it is unnecessary.

hawkingrei:

Issue Number: close #54481

Problem Summary:

Now, ```StatsCacheImpl.Update``` can sync stats by version from the storage. we have many scenarios to call it. such as ```domain.loadStatsWorker``` and ```Analyze```. When customers use lightning, lightning will trigger the multi-analyzed tasks simultaneously. so it also triggers multi  ```StatsCacheImpl.Update``` simultaneously. it is unnecessary.

add a singleflight for ```StatsCacheImpl.Update```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/handle/cache/statscache.go
+++ b/pkg/statistics/handle/cache/statscache.go
@@ -63,8 +63,20 @@ func NewStatsCacheImplForTest() (types.StatsCache, error) {
 	return NewStatsCacheImpl(nil)
 }
 
+// StatsCacheUpdateChan is a channel for updating stats cache.
+var StatsCacheUpdateChan = make(chan struct{}, 1)
+
 // Update reads stats meta from store and updates the stats map.
-func (s *StatsCacheImpl) Update(ctx context.Context, is infoschema.InfoSchema) error {
+func (*StatsCacheImpl) Update() error {
+	select {
+	case StatsCacheUpdateChan <- struct{}{}:
+	default:
+	}
+	return nil
+}
+
+// UpdateWorker reads stats meta from store and updates the stats map.
+func (s *StatsCacheImpl) UpdateWorker(ctx context.Context, is infoschema.InfoSchema) error {
 	start := time.Now()
 	lastVersion := s.getLastVersion()
 	var (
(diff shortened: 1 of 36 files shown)


hi-rustin:
We had some offline discussions that this change could potentially cause a table to be re-analyzed after the analysis is executed. The update method is called after the analysis, but if we only do it once, there is a big chance that the stats cache of the analyzed table will not be updated, which will result in the table being re-queued for analysis.

pkg/statistics/handle/cache/statscache.go
```diff
@@ -62,8 +62,20 @@ func NewStatsCacheImplForTest() (types.StatsCache, error) {
...
 
+// StatsCacheUpdateChan is a channel for updating stats cache.
+var StatsCacheUpdateChan = make(chan struct{}, 1)
+
 // Update reads stats meta from store and updates the stats map.
-func (s *StatsCacheImpl) Update(is infoschema.InfoSchema) error {
+func (*StatsCacheImpl) Update() error {
+	select {
+	case StatsCacheUpdateChan <- struct{}{}:
+	default:
```
hi-rustin, line 72: I think it doesn't solve the issue I mentioned https://github.com/pingcap/tidb/pull/54480#issuecomment-2210670759.

I believe we need to ensure that once the analysis is finished, we can always load the latest stats for the analyzed tables. Otherwise, there is a chance we will reanalyze them.

hi-rustin (changes requested):

pkg/statistics/handle/cache/statscache.go
hawkingrei: I agree with you. Therefore, prior to commencing auto analyze, I will enforce a synchronization of the stats cache.

Issue #20484, opened 2020-10-16
Make @@tidb_constraint_check_in_place session variable work for unique index in the sysbench `oltp_update_index` workload


tiancaiamao:
## Description

![image](https://user-images.githubusercontent.com/1420062/96205025-3b57fd80-0f98-11eb-9334-3fb468444db9.png)

From the trace result, we can see that update the unique index needs to `getSnapshot` and visit TiKV (`SendReqCtx`) to ensure the new KV entry does not exist.

It should be possible to make this check lazy, just make an assumption the new value does not exist and delay the check to 2PC commit.

This is exactly what `@@tidb_constraint_check_in_place` do, so this task is to verify it could work, fix the code if it's broken.

## Score

* 600

## Mentor(s)

* @tiancaiamao

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Go, TiDB source code

tiancaiamao:
@xiaodong-ji would you like to take it? it seems the bot has a `/pick` command

xiaodong-ji:
> @xiaodong-ji would you like to take it? it seems the bot has a `/pick` command

OK，Let me try.

xiaodong-ji:
It looks like the `tidb_constraint_check_in_place` parameter is valid for INSERT and invalid for UPDATE.
```
mysql> show variables like 'tidb_constraint_check_in_place';
+--------------------------------+-------+
| Variable_name                  | Value |
+--------------------------------+-------+
| tidb_constraint_check_in_place | 0     |
+--------------------------------+-------+
1 row in set (0.00 sec)

// insert statement
mysql> begin optimistic;
Query OK, 0 rows affected (0.00 sec)

mysql> insert into sbtest1(k,c) values(4993,'83868641912-28773972837-60736120486-75162659906-27563526494-20381887404-41576422241-93426793964-56405065102-33518432330');
Query OK, 1 row affected (0.00 sec)

mysql> commit;
ERROR 1062 (23000): Duplicate entry '4993-83868641912-28773972837-60736120486-75162659906-27563526494' for key 'k_1'
[...]

PR #54495, opened 2024-07-08
executor: fix update ignore still report dup-key error in statement retry

### LINKED ISSUE #54489: The `update ignore` may still report dup-key error when retry
## Bug Report

This unit test fails

```golang
func TestUpdateRowRetryAndThenDupKey(t *testing.T) {
	tk := testkit.NewSteppedTestKit(t, store)
	tk.MustExec("use test")
	tk.MustExec("create table t(id int primary key, u int unique)")
	tk.MustExec("insert into t values(1, 1)")

	// session 1 update u=2 for id=1 and halt before executor first runs.
	tk.SetBreakPoints(sessiontxn.BreakPointBeforeExecutorFirstRun)
	tk.SteppedMustExec("update ignore t set u = 2 where id = 1").
		ExpectStopOnBreakPoint(sessiontxn.BreakPointBeforeExecutorFirstRun)

	// session 2  insert a new row (2, 2) to make the unique key conflict.
	tk2.MustExec("use test")
	tk2.MustExec("insert into t values(2, 2)")

	// Continue the execution of session1, it should meet an optimistic conflict and retry.
	// The second execution is still failed because of the unique key conflict.
	// But the `update ignore` statement should not give any error.
	tk.Continue().ExpectStopOnBreakPoint(sessiontxn.BreakPointBeforeExecutorFirstRun)
	tk.Continue().ExpectIdle()
	// Should only a dup-key warning and the row 1 is not updated.
	tk.MustQuery("show warnings").Check(testkit.Rows("Warning 1062 Duplicate entry '2' for key 't.u'"))
	tk.MustQuery("select * from t order by id").Check(testkit.Rows("1 1", "2 2"))
}
```

The test should pass.

The test fails with error:

```
        	            	[kv:1062]Duplicate entry '2' for key 't.u'
```

### LINKED ISSUE #54492: optimistic transaction does not respect `tidb_constraint_check_in_place`
## Bug Report

```
create table t(id int primary key, u int unique);
insert into t values (1, 1), (2, 2);
set @@tidb_constraint_check_in_place=0;
begin optimistic;
update t set u=2 where id=1;
```

update should not report any error in optimistic transaction because `tidb_constraint_check_in_place` is off

```
> update t set u=2 where id=1;
(1062, "Duplicate entry '2' for key 't.u'")
```

lcwangchao:

Issue Number: close #54489

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
fix some dup-key cases in UpdateRecord
```

### DIFF
--- a/pkg/table/tables/tables.go
+++ b/pkg/table/tables/tables.go
@@ -576,21 +576,9 @@ func (t *TableCommon) UpdateRecord(ctx context.Context, sctx table.MutateContext
 	}
 	sessVars := sctx.GetSessionVars()
 	// rebuild index
-	if !sessVars.InTxn() {
-		savePresumeKeyNotExist := sessVars.PresumeKeyNotExists
-		if !sessVars.ConstraintCheckInPlace && sessVars.TxnCtx.IsPessimistic {
-			sessVars.PresumeKeyNotExists = true
-		}
-		err = t.rebuildIndices(sctx, txn, h, touched, oldData, newData, table.WithCtx(ctx))
-		sessVars.PresumeKeyNotExists = savePresumeKeyNotExist
-		if err != nil {
-			return err
-		}
-	} else {
-		err = t.rebuildIndices(sctx, txn, h, touched, oldData, newData, table.WithCtx(ctx))
-		if err != nil {
-			return err
-		}
+	err = t.rebuildIndices(sctx, txn, h, touched, oldData, newData, table.WithCtx(ctx))
+	if err != nil {
+		return err
 	}
 
 	key := t.RecordKey(h)
--- a/pkg/executor/insert_common.go
+++ b/pkg/executor/insert_common.go
@@ -1413,11 +1413,15 @@ func (e *InsertValues) addRecordWithAutoIDHint(
 	ctx context.Context, row []types.Datum, reserveAutoIDCount int,
 ) (err error) {
 	vars := e.Ctx().GetSessionVars()
+	if !vars.ConstraintCheckInPlace {
+		vars.PresumeKeyNotExists = true
+	}
 	if reserveAutoIDCount > 0 {
 		_, err = e.Table.AddRecord(e.Ctx().GetTableCtx(), row, table.WithCtx(ctx), table.WithReserveAutoIDHint(reserveAutoIDCount))
 	} else {
 		_, err = e.Table.AddRecord(e.Ctx().GetTableCtx(), row, table.WithCtx(ctx))
 	}
+	vars.PresumeKeyNotExists = false
 	if err != nil {
 		return err
 	}
--- a/pkg/executor/insert_common.go
+++ b/pkg/executor/insert_common.go
@@ -1413,15 +1413,11 @@ func (e *InsertValues) addRecordWithAutoIDHint(
 	ctx context.Context, row []types.Datum, reserveAutoIDCount int,
 ) (err error) {
 	vars := e.Ctx().GetSessionVars()
-	if !vars.ConstraintCheckInPlace {
-		vars.PresumeKeyNotExists = true
-	}
 	if reserveAutoIDCount > 0 {
 		_, err = e.Table.AddRecord(e.Ctx().GetTableCtx(), row, table.WithCtx(ctx), table.WithReserveAutoIDHint(reserveAutoIDCount))
 	} else {
 		_, err = e.Table.AddRecord(e.Ctx().GetTableCtx(), row, table.WithCtx(ctx))
 	}
-	vars.PresumeKeyNotExists = false
 	if err != nil {
 		return err
 	}
--- a/pkg/executor/update_test.go
+++ b/pkg/executor/update_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/pingcap/tidb/pkg/kv"
 	sessiontypes "github.com/pingcap/tidb/pkg/session/types"
 	"github.com/pingcap/tidb/pkg/sessionctx/variable"
+	"github.com/pingcap/tidb/pkg/sessiontxn"
 	"github.com/pingcap/tidb/pkg/testkit"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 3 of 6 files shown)


pkg/table/tables/tables.go
```diff
@@ -576,21 +576,9 @@ func (t *TableCommon) UpdateRecord(ctx context.Context, sctx table.MutateContext
...
-		sessVars.PresumeKeyNotExists = savePresumeKeyNotExist
-		if err != nil {
-			return err
-		}
-	} else {
-		err = t.rebuildIndices(sctx, txn, h, touched, oldData, newData, table.WithCtx(ctx))
-		if err != nil {
-			return err
-		}
+	err = t.rebuildIndices(sctx, txn, h, touched, oldData, newData, table.WithCtx(ctx))
```
ekexium, line 579: Why change these lines? The test still passes if these lines are unchanged.

lcwangchao: Because this code is unnecessary now. The current `LazyCheckKeyNotExists` is:

```
func (s *SessionVars) LazyCheckKeyNotExists() bool {
	if s.StmtCtx.ErrGroupLevel(errctx.ErrGroupDupKey) != errctx.LevelError {
		// This branch means we are in `insert/update ignore`.
		// The executor will handle the dup-key error and ignore it in executor,
		// so we must check the dup-key error in place to make sure the executor can get the error.
		return false
	}
	return s.PresumeKeyNotExists || (s.TxnCtx != nil && s.TxnCtx.IsPessimistic)
}
```

You can see that if `sessVars.TxnCtx.IsPessimistic` is true, `s.PresumeKeyNotExists || (s.TxnCtx != nil && s.TxnCtx.IsPessimistic)` will always return true, we do not need to set `PresumeKeyNotExists`.

ekexium (commented):
Consider a pessimistic txn, constraint_check_in_place=off, and error level is LevelWarn. In original code `LazyCheckKeyNotExists` will return true, in this PR it will return false, right?

Besides, `s.PresumeKeyNotExists` only makes the reasoning more complex. Since it's only set in one place, I think we should consider removing it.

lcwangchao:
> Consider a pessimistic txn, constraint_check_in_place=off, and error level is LevelWarn. In original code `LazyCheckKeyNotExists` will return true, in this PR it will return false, right? If we let the original value of `vars.PresumeKeyNotExists` be `!vars.ConstraintCheckInPlace` we will have the table:
> 
> InTxn	constraint_check_in_place	mode	levelError	master, lazy?	PR, lazy?	Equivalent
> FALSE	FALSE	opt	FALSE	TRUE	FALSE	FALSE
> FALSE	FALSE	opt	TRUE	TRUE	TRUE	TRUE
> FALSE	FALSE	pes	FALSE	TRUE	FALSE	FALSE
> FALSE	FALSE	pes	TRUE	TRUE	TRUE	TRUE
> FALSE	TRUE	opt	FALSE	FALSE	FALSE	TRUE
> FALSE	TRUE	opt	TRUE	FALSE	FALSE	TRUE
> FALSE	TRUE	pes	FALSE	FALSE	FALSE	TRUE
> FALSE	TRUE	pes	TRUE	TRUE	TRUE	TRUE
> TRUE	FALSE	opt	FALSE	TRUE	FALSE	FALSE
> TRUE	FALSE	opt	TRUE	TRUE	TRUE	TRUE
> TRUE	FALSE	pes	FALSE	TRUE	FALSE	FALSE
> TRUE	FALSE	pes	TRUE	TRUE	TRUE	TRUE
> TRUE	TRUE	opt	FALSE	FALSE	FALSE	TRUE
> TRUE	TRUE	opt	TRUE	FALSE	FALSE	TRUE
> TRUE	TRUE	pes	FALSE	FALSE	FALSE	TRUE
> TRUE	TRUE	pes	TRUE	TRUE	TRUE	TRUE
> Besides, `s.PresumeKeyNotExists` only makes the reasoning more complex. Since it's only set in one place, I think we should consider removing it.

This PR only changes `LazyCheckKeyNotExists` in the pessimistic mode. In optimistic mode, it's return value is affected by `PresumeKeyNotExists`. In the update statement, it is not set anyway, so this PR does not change the return value here. In the insert statement, it only changes the return value when error level is `LevelWarn`, however, this only happens in `insert ignore`. The `LazyCheckKeyNotExists` is not used because `BatchCheck` is set to true...

Yes, I think `PresumeKeyNotExists` is boring and we should remove it. But I think we should also remove `LazyCheckKeyNotExists` and use a new option `WithDupKeyCheckMode` (default is checkInPlace) to indicate it. Each scene should compute the DupKeyCheckMode separately.

ekexium:
So the correctness of this fix depends on how upper layer uses it. I suggest that we make `LazyCheckKeyNotExists` unchanged for optimistic transactions, instead of depending on the callers. It seems a more safer approach.

pkg/table/tables/tables.go
lcwangchao: reverted diffs in `LazyCheckKeyNotExists` can also fix this bug...

ekexium (commented):
Have you verified https://github.com/pingcap/tidb/issues/20484? Seems the part was introduced to solve that issue.

lcwangchao:
> Have you verified #20484? Seems the part was introduced to solve that issue.

Seems it is trying to make a lazy check for update... But I don't understand this PR, if a transaction is pessimistic and the statement is not `update ignore`, `LazyCheckKeyNotExists` should always return true. If the statement is `update ignore`, this bug will occur...

tiancaiamao (approved):
The current master branch does not work as expected.
If there is no `begin`, the expected behavior is that `index.Create`  does not need tikvSnapshotGet to check unique key exist or not.  
1WkBxs4akW

If there is a begin, both master branch and this PR don't use `tikvSnapshotGet` under `index.Create`:

QpFEM3wbyh


This PR does not make thing change, and it simplify the logic, so LGTM
As for why the current master branch does not work as expected, we can take a separate thread to trace...

pkg/table/tables/tables.go
jackysp: ‌‌Let's go back to the original place. The purpose of `tidb_constraint_check_in_place` is only to make the lazy unique index check for insert effective. Historically, only inserts had a lazy unique index check. `tidb_constraint_check_in_place` was actually designed to remove this exclusive check for inserts, so that all statements execute consistent in-place checks. Therefore, update statements have never had a lazy check. So even if `tidb_constraint_check_in_place` is off, updates should still report an error. @lcwangchao @ekexium

jackysp: As for why the name was designed this way, that was the product manager's decision at the time. However, I don't think it's necessary to make updates lazy check as well. The reason insert needs lazy check is because it performs more efficiently during bulk load. Bulk loading should rarely involve updates, right?

jackysp: If you want the update to also perform a lazy check, handling it only at https://github.com/pingcap/tidb/blob/08147e7e2586b1db7162ad9c2dfc134b5a5f6b43/pkg/table/tables/tables.go#L579-L589 is not enough. It also requires the 2PC phase to first check for unique constraints like insert before committing. This way, for https://github.com/pingcap/tidb/issues/54489, there will be no retry during an update; instead, a duplicate entry error reported by TiKV will be received.

cfzjywxk (approved):
It's difficult to verify the correctness for all the combinations of optimistic/pessimistic mods and check in-place or not directly.

We may need to file a refactor and test coverage task to clarify the related code path, especially the `optimistic mode` would be deprecated in the future. It could be also considered as one of the sub-tasks of the deprecation.
/cc @lcwangchao @ekexium

PR #54527, opened 2024-07-09
server: add protocol support for lazy cursor fetch

### LINKED ISSUE #54526: Enable lazy cursor fetch in `ResultSet` and protocol layer
## Enhancement

It'll need to have two changes:

1. Add a new variable `tidb_enable_lazy_cursor_fetch`.
2. Modify the existing code to try to detach the `recordSet`, and store needed information in the `ResultSet`.

YangKeao:

Issue Number: close #54526

Problem Summary:

Previously, all cursor fetch requests will load the data to the TiDB. Now, we can support lazily fetch results in a limited scope.

1. Add a new system variable `tidb_enable_lazy_cursor_fetch`.
2. Add a new implementation of `ResultSet` to give similar interface like the one for eager cursor fetch.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
Add a new system variable `tidb_enable_lazy_cursor_fetch` to indicate whether to try to fetch data lazily for cursor fetch requests.
```

### DIFF
--- a/pkg/server/conn_stmt.go
+++ b/pkg/server/conn_stmt.go
@@ -276,8 +276,8 @@ func (cc *clientConn) executePreparedStmtAndWriteResult(ctx context.Context, stm
 
 	// first, try to clear the left cursor if there is one
 	if useCursor && stmt.GetCursorActive() {
-		if stmt.GetResultSet() != nil && stmt.GetResultSet().GetRowContainerReader() != nil {
-			stmt.GetResultSet().GetRowContainerReader().Close()
+		if stmt.GetResultSet() != nil && stmt.GetResultSet().GetRowIterator() != nil {
+			stmt.GetResultSet().GetRowIterator().Close()
 		}
 		if stmt.GetRowContainer() != nil {
 			stmt.GetRowContainer().GetMemTracker().Detach()
@@ -304,8 +304,13 @@ func (cc *clientConn) executePreparedStmtAndWriteResult(ctx context.Context, stm
 	}
 	execStmt.SetText(charset.EncodingUTF8Impl, sql)
 	rs, err := (&cc.ctx).ExecuteStmt(ctx, execStmt)
+	var lazy bool
 	if rs != nil {
-		defer rs.Close()
+		defer func() {
+			if !lazy {
+				rs.Close()
+			}
+		}()
 	}
 	if err != nil {
 		// If error is returned during the planner phase or the executor.Open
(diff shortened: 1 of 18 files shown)


pkg/server/tests/cursor/cursor_test.go
```diff
@@ -226,3 +226,66 @@ outerLoop:
 		}
 	}
 }
+
+func TestSerialLazyExecuteAndFetch(t *testing.T) {
```
YangKeao, line 230: Now we only have a serial test here. This PR won't test concurrent cursor fetch with normal execution (or another cursor fetch). After both https://github.com/pingcap/tidb/pull/54456 and this one are merged, I'll add more tests about the concurrent lazy cursor fetch.

pkg/server/internal/resultset/cursor.go
```diff
@@ -53,3 +97,81 @@ type FetchNotifier interface {
...
+type lazyRowIterator struct {
+	rs       ResultSet
+	err      error
+	chk      *chunk.Chunk
+	idxInChk int
+	started  bool
+}
+
+func (iter *lazyRowIterator) Next(ctx context.Context) chunk.Row {
+	iter.idxInChk++
```
lcwangchao, line 134: If we call `Next` before `Current`, seems `Current` will returns a different row with the row returned by `Next`.

Should we init the first row when we find `!started` here?

Another way to implement this iterator is to init the first when `lazyRowIterator` is created so that we do not need to add a extra ctx argument to `Current`

pkg/server/internal/resultset/resultset.go
```diff
@@ -142,3 +143,25 @@ func (trs *tidbResultSet) FieldTypes() []*types.FieldType {
...
+	if !ok {
+		return nil, false, nil
+	}
+
+	recordSet, detached, err := detachableRecordSet.TryDetach()
+	if !detached {
+		return nil, false, nil
+	}
+	if err != nil {
+		return nil, true, err
```
lcwangchao, line 159: why return true here?

pkg/server/conn_stmt.go
```diff
@@ -331,97 +336,143 @@ func (cc *clientConn) executePreparedStmtAndWriteResult(ctx context.Context, stm
...
 
-		if err = cc.writeColumnInfo(crs.Columns()); err != nil {
-			return false, err
+	reader := chunk.NewRowContainerReader(rowContainer)
+	defer func() {
+		if err != nil {
+			reader.Close()
 		}
+	}()
+	crs.StoreRowContainerReader(reader)
```
lcwangchao, line 425: Seems `StoreRowContainerReader` is the only extra method for `EagerCursorResultSet` and is only used here. Why not to remove `EagerCursorResultSet` and add a second argument `reader` to `WrapWithCursor`.

We can rename `WrapWithCursor` to `WrapWithRowContainerCursor`

pkg/server/conn_stmt.go
```diff
@@ -331,97 +336,143 @@ func (cc *clientConn) executePreparedStmtAndWriteResult(ctx context.Context, stm
...
+	retryable, err := cc.writeResultSet(ctx, rs, true, cc.ctx.Status(), 0)
+	if err != nil {
+		return retryable, errors.Annotate(err, cc.preparedStmt2String(uint32(stmt.ID())))
+	}
+	return false, nil
+}
 
-		for {
-			chk := crs.NewChunk(nil)
+func (cc *clientConn) executeWithCursor(ctx context.Context, stmt PreparedStatement, rs resultset.ResultSet) (lazy bool, retryable bool, err error) {
```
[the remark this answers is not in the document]
YangKeao: Done.

pkg/server/conn_stmt.go
YangKeao: Done.

pkg/server/internal/resultset/cursor.go
YangKeao: I'll add a `iter.started = true` in the `.Next()` function. Then it'll fix the difference between the first `Next` and the following `Current`. Their results are different because the second `Current()` will call an extra `Next()` because the `iter.started` is not set to `true`.

> Should we init the first row when we find `!started` here?

In `.Next()`, it'll always init the first row, because initially the chunk is empty and `if iter.idxInChk >= iter.chk.NumRows() {` always meets.

> Another way to implement this iterator is to init the first when lazyRowIterator is created so that we do not need to add a extra ctx argument to Current

Then the first `.Next()` of the executor will be called in the `EXECUTE` but not `FETCH` :thinking:. It's OK, but I don't think remove one argument of the `Current` is valuable enough to bring this change.

pkg/server/internal/resultset/resultset.go
YangKeao: It follows the second return value from `detachableRecordSet.TryDetach()`. I'll modify it to `return nil, detached, err` to make it clearer. It's always `true` in this case.

pkg/server/internal/resultset/resultset.go
```diff
@@ -142,3 +143,25 @@ func (trs *tidbResultSet) FieldTypes() []*types.FieldType {
...
+
+// TryDetach creates a new `ResultSet` which doesn't depend on the current session context.
+func (trs *tidbResultSet) TryDetach() (ResultSet, bool, error) {
+	detachableRecordSet, ok := trs.recordSet.(sqlexec.DetachableRecordSet)
+	if !ok {
+		return nil, false, nil
+	}
+
+	recordSet, detached, err := detachableRecordSet.TryDetach()
+	if !detached {
```
lcwangchao, line 155: Do we need to process errors firstly? Or just:

```
if !detached || err != nil {
    return recordSet, detached, err
}
```

YangKeao: Done.

pkg/server/internal/resultset/resultset.go
YangKeao: I realized that the implementation of `execStmtResult.TryDetach()` has a little mistake: it should return `nil, true, err` when facing the error returned by `finishStmt` :facepalm: . I've added a test for this situation.

lcwangchao (approved):

pkg/server/conn_stmt.go
```diff
@@ -331,97 +336,143 @@ func (cc *clientConn) executePreparedStmtAndWriteResult(ctx context.Context, stm
...
+		return retryable, errors.Annotate(err, cc.preparedStmt2String(uint32(stmt.ID())))
+	}
+	return false, nil
+}
 
-		for {
-			chk := crs.NewChunk(nil)
+func (cc *clientConn) executeWithCursor(ctx context.Context, stmt PreparedStatement, rs resultset.ResultSet) (lazy bool, err error) {
+	vars := (&cc.ctx).GetSessionVars()
+	if vars.EnableLazyCursorFetch {
```
YangKeao, line 351: Most of the following logic in this function is copied from the previous deleted lines. The only difference is this branch.

xhebox (approved):

pkg/server/conn_stmt.go
```diff
@@ -331,97 +336,143 @@ func (cc *clientConn) executePreparedStmtAndWriteResult(ctx context.Context, stm
...
+	return false, err
+}
+
+// executeWithLazyCursor tries to detach the `ResultSet` and make it suitable to execute lazily.
+// Be careful that the return value `(bool, error)` has different meaning with other similar functions. The first `bool` represent whether
+// the `ResultSet` is suitable for lazy execution. If the return value is `(false, _)`, the `rs` in argument can still be used. If the
+// first return value is `true` and `err` is not nil, the `rs` cannot be used anymore and should return the error to the upper layer.
+func (cc *clientConn) executeWithLazyCursor(ctx context.Context, stmt PreparedStatement, rs resultset.ResultSet) (ok bool, err error) {
+	drs, ok, err := rs.TryDetach()
+	if !ok {
```
qw4990, line 439: Should we use "!ok || err!=nil", which seems safer?

qw4990 (approved):

pkg/server/conn_stmt.go
YangKeao: You are right!

Issue #52905, opened 2024-04-26
`select` very slow on an empty table from `delete from xx`


D3Hunter:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
- create cluster with `tiup playground`
- do the following, we can see query from the empty table takes about `60ms`. also test another table with `524288` rows, then empty it with `delete from`, it takes `440ms` to query it.
```sql
mysql> create table t(a int);
Query OK, 0 rows affected (0.10 sec)

.... insert data

mysql> delete from t;
Query OK, 65536 rows affected (0.38 sec)

mysql> select count(1) from t;
+----------+
| count(1) |
+----------+
|        0 |
+----------+
1 row in set (0.06 sec)

mysql> show table t regions;
+-----------+-----------+--------------------+-----------+-----------------+-------+------------+---------------+------------+----------------------+------------------+------------------------+------------------+
| REGION_ID | START_KEY | END_KEY            | LEADER_ID | LEADER_STORE_ID | PEERS | SCATTERING | WRITTEN_BYTES | READ_BYTES | APPROXIMATE_SIZE(MB) | APPROXIMATE_KEYS | SCHEDULING_CONSTRAINTS | SCHEDULING_STATE |
+-----------+-----------+--------------------+-----------+-----------------+-------+------------+---------------+------------+----------------------+------------------+------------------------+------------------+
|        14 | t_107_    | t_281474976710649_ |        15 |               1 | 15    |          0 |      23118061 |   16450774 |                   24 |           133569 |                        |                  |
+-----------+-----------+--------------------+-----------+-----------------+-------+------------+---------------+------------+----------------------+------------------+------------------------+------------------+

mysql> select now();
+---------------------+
| now()               |
+---------------------+
| 2024-04-26 11:34:15 |
+---------------------+
1 row in set (0.00 sec)

```
- at first, i thought it related to MVCC, so wait GC and query again, still very slow
```sql
...... wait gc finished on table t


mysql> select count(1) from t;
+----------+
| count(1) |
+----------+
|        0 |
+----------+
1 row in set (0.07 sec)

mysql> select now();
+---------------------+
| now()               |
+---------------------+
| 2024-04-26 14:18:50 |
+---------------------+
1 row in set (0.00 sec)
```

- for a new created empty table, query time < 10ms
```sql
mysql> create table tt(a int);
Query OK, 0 rows affected (0.11 sec)

mysql> select count(1) from tt;
[...]

PR #54547, opened 2024-07-10
ddl: pessimistic lock global id, alloc id & insert ddl job in one txn

### LINKED ISSUE #54436: ddl code refactor/optimize
## Enhancement
We have listed a lot of items that we want to refactor, to make the structure more clear and code easier to maintain and extend, and some optimize tasks to make general DDL run faster. And in the last 8.2 sprint, we optimized for general DDL, and done some refactor for job scheduling and have left some TODOs during the refactor, https://github.com/pingcap/tidb/issues/53246.
In 8.3 sprint, we have the same goal as in 8.2 and we also want to enhance testibility of DDL component, by decoupling different part of DDL(TODO), and replacing ddl.Hook with simplier failpoint.

### Decouple components of DDL
Currently, ddl/ddlCtx is quite large and has a lot of responsibility, including:
- handle job submission which come from sql executor, this is the biggest part
  - This part shares the same interface as schema-tracker, maybe extract this interface as something called DDLExecutor.
- Handle local ddl job execution for fast creating tables. It runs on every node
- owner election, start job scheduler
- Enable/disable some functions, such as whether this node can attend owner election, and fast-create
- It contains a lot of methods related to job execution, such as writePhysicalTableRecord
- Manages fields which are shared with other components, such as below. They only use part of their function, and there are no intersections. Maybe we can create those fields on each side or separate the function.
  - SchemaSyncer, or should be called schema-version-syncer, shared with domain
  - StateSyncer, shared with SyncUpgrade api handling.
- Manages fields that should be part of job scheduler: reorg-ctx/job-ctx/ddlSeqNum/waitSchemaSyncedController
- Some fields are used for unit tests: hook/interceptor inside ddlCtx. We can replace them with failpoints.
- Some util methods which should be moved out: GetTableMaxHandle

### Tasks
- [ ] Optimize job execution
  - [ ] remove this markJobProcessing completely, we have move job dependency calculation in memory, and as long as the job resides in the table, we should try run it, so this fields is useless now.
  - [ ] merge loadDDLJobAndRun by job type into one, and remove the sub-query in getJobSQL, and handle this case https://github.com/pingcap/tidb/pull/54110#discussion_r1647002871 . And schedule as many jobs as possible in one round.
  - [ ] Make job id allocation and insert job run in one transaction, and test QPS of it when run in parallel. If it outperforms job execution, we can re-add the reverted optimization that query from the min job id.
  - [ ] Use RPC to notify job done when DDL job is submitted on non-owner node.
- [ ] Decouple components of DDL
  - [ ] Move waitSchemaSyncedController / reorgCtx / jobCtx / ddlSeqNum/waitSchemaSyncedController to job scheduler from ddlCtx
  - [ ] Separate DDLExecutor interface out of DDL interface
  - [ ] separate local job execution out
  - [ ] Moving methods related to job execution out
  - [ ] Avoid exposing SchemaSyncer and StateSyncer from DDL
  - [ ] Replace fields only used for test with failpoint: such as hook/interceptor inside ddlCtx
  - [ ] Move GetTableMaxHandle out
- [ ] Refactor for reorg jobs (TODO, only list a few)
  - [ ] make all global variables like ingest.LitBackCtxMgr local to job scheduler
  - [ ] ReorgCtx is stored at ddlCtx, and we are waiting for the reorg routine repeatedly by entering/exit the ddl worker. If after the check due to waitTimeout, owner changes, we might have no chance to clean it up. We better wait for the routine inside runReorgJob
  - [ ] Remove dependency on lightning config, we should use local.BackendConfig directly.

D3Hunter:

Issue Number: ref #54436

Problem Summary:

-  pessimistic lock global ID key before alloc it to avoid write conflict
- job id allocation and job insertion are in the same transaction, as we want to make sure DDL jobs are inserted in id order, then we can query from a min job ID when scheduling DDL jobs to mitigate https://github.com/pingcap/tidb/issues/52905.
- 🟥 **THIS PR hurts DDL performance, it requires another 2 PR to out perform current master: 1 to combine table id allocation with job id, and 2 to query since min job id**. they will be filed later.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test

when len(job_meta) < 8k, with [1, 1000] routines to run TestGenIDAndInsertJobsWithRetryQPS on 1pd/3tikv environment, QPS is about [300, 400], much larger than general DDL execution QPS

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


lcwangchao:
LGTM for txn part

D3Hunter:
/hold

wait a another review from txn team

GMHDBJD (approved):
LGTM

pkg/ddl/ddl_worker.go
```diff
@@ -492,28 +484,150 @@ func (d *ddl) addBatchDDLJobs(tasks []*limitJobTask) error {
...
+	}
+
+	return nil
+}
+
+// GenIDAndInsertJobsWithRetry generate job ID and inserts DDL jobs to the DDL job
+// table with retry. job id allocation and job insertion are in the same transaction,
+// as we want to make sure DDL jobs are inserted in id order, then we can query from
+// a min job ID when scheduling DDL jobs to mitigate https://github.com/pingcap/tidb/issues/52905.
+// so this function has side effect, it will set the job id of 'tasks'.
```
lance6716, line 518:
```suggestion
// so this function has side effect, it will set the job id of 'jobs'.
```

pkg/ddl/ddl_worker.go
```diff
@@ -492,28 +484,150 @@ func (d *ddl) addBatchDDLJobs(tasks []*limitJobTask) error {
...
+}
+
+// lockGlobalIDKey locks the global ID key in the meta store. it keeps trying if
+// meet write conflict, we cannot have a fixed retry count for this error, see this
+// https://github.com/pingcap/tidb/issues/27197#issuecomment-2216315057.
+// this part is same as how we implement pessimistic + repeatable read isolation
+// level in SQL executor, see doLockKeys.
+// TODO maybe we can unify the lock mechanism with SQL executor in the future, or
+// implement it inside TiKV client-go.
+func lockGlobalIDKey(ctx context.Context, ddlSe *sess.Session, txn kv.Transaction) (uint64, error) {
```
lance6716, line 596: not a big problem. We can get `txn` from `ddlSe`

pkg/ddl/ddl_worker.go
```diff
@@ -492,28 +484,150 @@ func (d *ddl) addBatchDDLJobs(tasks []*limitJobTask) error {
...
+	var (
+		iteration   uint
+		forUpdateTs = txn.StartTS()
+		ver         kv.Version
+		err         error
+	)
+	waitTime := ddlSe.GetSessionVars().LockWaitTimeout
```
lance6716, line 603: this session (ddlSe) is get from session pool, not user's session. Maybe get the lock wait value from user's session.

D3Hunter: this is part of internal DDL execution, internal not user setting should be used

pkg/ddl/ddl_worker.go
```diff
@@ -492,28 +484,150 @@ func (d *ddl) addBatchDDLJobs(tasks []*limitJobTask) error {
...
+			if err != nil {
+				return errors.Trace(err)
+			}
+			if err = fn(ids); err != nil {
+				return errors.Trace(err)
+			}
+			return ddlSe.Commit(ctx)
+		}()
+
+		if resErr != nil && kv.IsTxnRetryableError(resErr) {
```
lance6716, line 579: not sure what's the meaning of `ErrLockExpire`. I just want to cover the case that pessimistic lock is cleaned by other readers of GlobalIDKey. (If the only reader is the new DDL owner, we don't need to retry because it means current node should retire)

pkg/ddl/ddl_worker.go
```diff
@@ -492,28 +484,150 @@ func (d *ddl) addBatchDDLJobs(tasks []*limitJobTask) error {
...
-		}
 	}
 
 	se.GetSessionVars().SetDiskFullOpt(kvrpcpb.DiskFullOpt_AllowedOnAlmostFull)
-
-	if tasks[0].job.LocalMode {
+	ddlSe := sess.NewSession(se)
+	localMode := tasks[0].job.LocalMode
+	if localMode {
+		if err = fillJobIDs(ctx, ddlSe, tasks); err != nil {
```
lance6716, line 495: local mode jobs are not written to ddl_job table, no need to lock the global ID? I'm afraid the performance for local job is more important.

lance6716 (commented):
will review soon

pkg/ddl/ddl_worker.go
lance6716: Not sure if it's more reasonable that the setting of DDL's SQL connection can affect the timeout.

lance6716 (commented):
rest lgtm

pkg/ddl/ddl_worker.go
```diff
@@ -492,28 +484,150 @@ func (d *ddl) addBatchDDLJobs(tasks []*limitJobTask) error {
...
+	}
+	return resErr
+}
+
+// lockGlobalIDKey locks the global ID key in the meta store. it keeps trying if
+// meet write conflict, we cannot have a fixed retry count for this error, see this
+// https://github.com/pingcap/tidb/issues/27197#issuecomment-2216315057.
+// this part is same as how we implement pessimistic + repeatable read isolation
+// level in SQL executor, see doLockKeys.
+// TODO maybe we can unify the lock mechanism with SQL executor in the future, or
```
cfzjywxk, line 594: The kv-client is aware of key-value semantics but not SQL execution semantics, it has only key-value APIs.

Actually if the meta key could be abstracted as an internal table, the lock operation could be achived using internal SQL statements like 
```
select * from t_global_id_key for update;
or
update t_global_id_key set val = val + 1;
```

The lock operations could be saved and the execution is just like normal `select for update` and `update` statemetns.

pkg/ddl/ddl_worker.go
```diff
@@ -492,28 +484,150 @@ func (d *ddl) addBatchDDLJobs(tasks []*limitJobTask) error {
...
+		task.job.ID = allocatedIDs[i]
+	}
+	return nil
+}
+
+// genIDAndCallWithRetry generates global IDs and calls the function with retry.
+// generate ID and call function runs in the same transaction.
+func genIDAndCallWithRetry(ctx context.Context, ddlSe *sess.Session, count int, fn func(ids []int64) error) error {
+	var resErr error
+	for i := uint(0); i < kv.MaxRetryCnt; i++ {
```
cfzjywxk, line 547: Not that much transaction retry is needed, as the conflict error is already handled by the in-transaction statements or operations.

pkg/ddl/ddl_worker.go
```diff
@@ -492,28 +484,150 @@ func (d *ddl) addBatchDDLJobs(tasks []*limitJobTask) error {
...
+		ver         kv.Version
+		err         error
+	)
+	waitTime := ddlSe.GetSessionVars().LockWaitTimeout
+	m := meta.NewMeta(txn)
+	idKey := m.GlobalIDKey()
+	for {
+		lockCtx := tikv.NewLockCtx(forUpdateTs, waitTime, time.Now())
+		err = txn.LockKeys(ctx, lockCtx, idKey)
+		if err == nil || !terror.ErrorEqual(kv.ErrWriteConflict, err) {
```
cfzjywxk, line 609: It's not a good idea to expost these details outside functions like `handlePessimitiLockError`, these errors and types could be changed. As commented above, it's better to use internal SQL statements to operate the internal states.

Actually, we have seen compatible problems from the direct use of `txn.xxx` api with optimistic internal transactions before.

pkg/ddl/ddl_worker.go
D3Hunter: > no need to lock the global ID?

memory lock is used to reduce write conflict, after later pr to "combine table id allocation with job id", there will be no write conflict in 1 node.

with the 2 pr mentioned in the pr, we can create 100k tables in about 13 minutes, and there is very little speed degradation, i will test 1M tables later, if everything goes ok, i still suggest deprecate fast-create in later version
image

pkg/ddl/ddl_worker.go
D3Hunter: yes, @lcwangchao ever suggest using advisory lock which uses the same way as you suggest, but we are in DDL module already, using those systable create cycle dependency. also we cannot change global ID to a table row now to be compatible.

pkg/ddl/ddl_worker.go
D3Hunter: > expost these details outside functions like handlePessimitiLockError, these errors and types could be changed

explained in prev comment, and added a todo

> txn.xxx api with optimistic internal transactions before.

we have set the transaction to pessimistic.

pkg/ddl/ddl_worker.go
D3Hunter: it's reusing retry count of `kv.RunInNewTxn`, in case of other retryable error

pkg/ddl/ddl_worker.go
D3Hunter: i will add this reason to comments

pkg/ddl/ddl_worker.go
D3Hunter: see https://github.com/tikv/client-go/blob/d73cc1ed6503925dfc7226e8d5677ceb4c2fd6f1/txnkv/transaction/2pc.go#L1227-L1230, and handled in tidb with below, seems no special handling https://github.com/pingcap/tidb/blob/06e0e17d98abbe3f88f82cee7746c0940384364f/pkg/session/session.go#L1029-L1040

i guess the reason is: T1 lock it so no one can change it before forUpdateTS, after the lock is expired & it hasn't commit, if there is T2 lock it later, T1 commit > T2 forUpdateTS  > T1 forUpdateTS, one of them will report `write conflict` on commit

pkg/ddl/ddl_worker.go
D3Hunter: will keep it, don't want to get it again, this method is internal anyway.

pkg/ddl/ddl_worker.go
D3Hunter: https://github.com/pingcap/tidb/pull/54547/commits/96ebfe6f49516f9411117f93db7e212f87cadd85

pkg/ddl/ddl_worker.go
D3Hunter: seems only affects txn for DML in mysql, mysql don't have a ddl_job table like us, so it shouldn't affect DDL
https://dev.mysql.com/doc/refman/8.4/en/innodb-parameters.html#sysvar_innodb_lock_wait_timeout

pkg/ddl/ddl_worker.go
lance6716: I found this type of error `ERROR 1105 (HY000): tikv aborts txn: Error(Txn(Error(Mvcc(Error(PessimisticLockNotFound` which likely standing for the case that pessimistic lock is cleaned by other transactions. But I don't know if it's handled by `kv.IsTxnRetryableError`. Transaction is very complex 😂

D3Hunter: > But I don't know if it's handled by kv.IsTxnRetryableError

https://github.com/pingcap/tidb/blob/9044acbff3a189b26b52ffd2990ce7b4f59021e0/pkg/kv/error.go#L83-L85

only ErrTxnRetryable except write conflict

> Transaction is very complex 😂

yes, that's why i ask @lcwangchao @cfzjywxk to review this part

D3Hunter: default `MaxTxnTTL:             60 * 60 * 1000, // 1hour`, seems ok to let user retry it themself in this case, even DML transaction will abort

lance6716: I see `ManagedLockTTL` in client-go is 20s. Does it means if pessimistic lock left in TiKV while the node crashes, other node must wait at least 20s to clean up the lock? If lock owner crashes, submitting DDL will be paused for 20s, it's not friendly.

D3Hunter: seems we cannot avoid this on crash
- create table t(id int primary key, v int); insert into t values(1,1);
- mysql -uroot -h 127.0.0.1 -P4000 test -e "begin; update t set v=2 where id=1; select sleep(100);";
```
mysql -uroot -h 127.0.0.1 -P4000 test -e "select now(); begin; update t set v=2 where id=1; select sleep(100);";
+---------------------+
| now()               |
+---------------------+
| 2024-07-15 21:59:56 |
+---------------------+
ERROR 2013 (HY000) at line 1: Lost connection to MySQL server during query
```
- kill -9 tidb, immediately after previous step
- restart
- run immediately after resart 
```
mysql -uroot -h 127.0.0.1 -P4000 test -e "begin; update t set v=3 where id=1; commit; select now();"
+---------------------+
| now()               |
+---------------------+
| 2024-07-15 22:00:17 |
+---------------------+
```

cfzjywxk (approved):
LGTM

pkg/ddl/ddl_worker.go
```diff
@@ -492,28 +484,155 @@ func (d *ddl) addBatchDDLJobs(tasks []*limitJobTask) error {
...
+	return nil
+}
+
+// genIDAndCallWithRetry generates global IDs and calls the function with retry.
+// generate ID and call function runs in the same transaction.
+func genIDAndCallWithRetry(ctx context.Context, ddlSe *sess.Session, count int, fn func(ids []int64) error) error {
+	var resErr error
+	for i := uint(0); i < kv.MaxRetryCnt; i++ {
+		resErr = func() (err error) {
+			if err := ddlSe.Begin(ctx); err != nil {
```
tangenta, line 549: If we `Begin()` here, previous statements executed by `ddlSe` will be committed implicitly. For example, the flashback jobs checking maybe outdated.

D3Hunter: > If we Begin() here, previous statements executed by ddlSe will be committed implicitly.

what do you mean, we don't have transaction before this one

> flashback jobs checking maybe outdated

it's the same behavior as our previous impl. if such job submitted, job scheduler will stop it from running if there is a flashback job before

D3Hunter: i ever thought remove this checking, as job schedule will calculate dependency between jobs, but a little different with current behavior

tangenta: > it's the same behavior as our previous impl.

OK.

> job scheduler will stop it from running

I guess the purpose of checking flashback job before submitting is to prevent having potential wrong info in `job.Args`.

D3Hunter: previously we cannot make sure there is no flashback job before this job, they might be submitted on different node

now we can do it by check inside the submit transaction, but i don't want to change this part now, and it might hurt performance

tangenta (approved):

PR #54600, opened 2024-07-12
*: Add GLOBAL as IndexOption and require it for creating Global Index

### LINKED ISSUE #52994: Add GLOBAL as IndexOption and require it explicitly for GLOBAL INDEX
## Enhancement
For Global index (for partitioned tables) the syntax is currently not explicit, but implicitly will create an global index if:
- the table is partitioned
- the index is unique and does not include all the columns in the partitioning expression.

To make it easier for users to see the difference we should make both the index creation/alter and the output from `SHOW CREATE TABLE` to explicitly show if it is an GLOBAL or LOCAL index.

Definition:
- a LOCAL index has the partition id as the key prefix (i.e. is local to the partition, and is maintained together with the partition)
- a GLOBAL index has the table id as the key prefix (i.e. covers the whole/global table data, and can enforce uniqueness for the whole table, but needs separate maintenance when doing partitioning management, like removing all index entries matching a dropped/truncated partition).

Reasoning:
GLOBAL or LOCAL index for partitioned tables should be an explicit choice, since before GLOBAL INDEX is supported, only LOCAL will be implicitly chosen.
GLOBAL index will allow removing the limitation that "all unique indexes needs to include all partitioning columns" and it will also require data reorganization (costly index maintenance) for partitioning operations, such as DROP/TRUNCATE PARTITION.

Limitations:
Oracle also allows partitioning a GLOBAL index (in a different way as the table is partitioned), that is not considered here. GLOBAL INDEX means a single non-partitioned index in this context.

mjonss:

Issue Number: close #52994

Problem Summary:

This also makes it easier to allow non-unique indexes as well as unique indexes already containing all partitioning columns  to be GLOBAL.

May be discussed further: how to handle `ALTER TABLE t PARTITION BY ...` when it would implicitly create a Global Index of UNIQUE Indexes if they don't include all partitioning columns:
- require a new ALTER option, like CONVERT TO GLOBAL INDEX (implemented in this PR, syntax still possible to change)
- just implicitly make the index Global
- Issue a warning or not when making implicitly making it Global

Parser:
- Added "GLOBAL" as IndexOption
- Added "GLOBAL" as option for ColumnOption in combination with {PRIMARY|UNIQUE} KEY, where it will use the StrValue set to "Global" instead of adding yet another struct member.
- Added "CONVERT TO GLOBAL INDEX" as option for ALTER TABLE t PARTITION BY ... [CONVERT TO GLOBAL INDEX] to allow [re]partitioning a table and convert unique indexes that does not include all partititioning columns into global indexes.

DDL:
- Checking and propagating the options from the parser, so that we require GLOBAL to be explicitly mentioned when creating a global index.

Executor:

Table:
- Just updated test to use GLOBAL IndexOption.

Added a new error for needing Global Index, but is not explicitly given.
Most changes are just updated tests and results, to comply with the new requirement of explicitly mention GLOBAL.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


bb7133:
> require a new ALTER option, something like {WITH|WITHOUT} GLOBAL INDEX

Personally I prefer this since it would be always better to use the explicit syntax for Global Index IMHO.

If we introduce something `WITH GLOBAL INDEX`, will this syntax required(report an error otherwise) if we have the unique index defined already for a table?

Defined2014:
Could you split it to several PRs, especially the parser part? Because if we need add a new syntax, should be approved by PM. And I think have a separate PR makes it easier to review and discuss syntax related issues. @mjonss

mjonss:
I will split this PR into at least a parser PR (including SHOW CREATE TABLE) and one for the logic to require GLOBAL IndexOption.

PR #54615, opened 2024-07-15
try balance lighting throughput to all tikvs

guoshouyan:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/lightning/backend/local/local.go
+++ b/br/pkg/lightning/backend/local/local.go
@@ -1294,6 +1294,69 @@ func (local *local) readAndSplitIntoRange(ctx context.Context, engine *Engine, r
 	return ranges, nil
 }
 
+func (local *local) scanRegion(
+	ctxt context.Context,
+	engine *Engine,
+	start, end []byte,
+	regionSplitSize int64,
+	regionSplitKeys int64,
+) ([]*split.RegionInfo, error) {
+	ito := &pebble.IterOptions{
+		LowerBound: start,
+		UpperBound: end,
+	}
+
+	iter := engine.newKVIter(ctxt, ito)
+	//nolint: errcheck
+	defer iter.Close()
+	// Needs seek to first because NewIter returns an iterator that is unpositioned
+	hasKey := iter.First()
+	if iter.Error() != nil {
+		return nil, errors.Annotate(iter.Error(), "failed to read the first key")
+	}
+	if !hasKey {
+		log.FromContext(ctxt).Info("There is no pairs in iterator",
+			logutil.Key("start", start),
+			logutil.Key("end", end))
+		engine.finishedRanges.add(Range{start: start, end: end})
+		return nil, nil
+	}
+	pairStart := append([]byte{}, iter.Key()...)
+	iter.Last()
+	if iter.Error() != nil {
+		return nil, errors.Annotate(iter.Error(), "failed to seek to the last key")
+	}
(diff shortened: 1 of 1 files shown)


lance6716:
Thanks for the contribution. Here's another point that relavant to import, lightning will use "write" RPC to send data to all peers of a region

https://github.com/pingcap/tidb/blob/2c71fbb84af72cf91127a4704cfe123e8e4622ca/br/pkg/lightning/backend/local/local.go#L1014-L1020

> pick a region with different leader store

Maybe it's helpful that lightning also considering peers' store.

br/pkg/lightning/backend/local/local.go
```diff
@@ -1541,53 +1604,89 @@ func (local *local) writeAndIngestByRanges(ctx context.Context, engine *Engine,
...
+				go func(w *worker.Worker) {
+					defer func() {
+						local.rangeConcurrency.Recycle(w)
+						wg.Done()
+					}()
+					var err error
+					// max retry backoff time: 2+4+8+16+30*26=810s
+					backOffTime := time.Second
+					for i := 0; i < maxWriteAndIngestRetryTimes; i++ {
+						err = local.writeAndIngestByRange(ctx, engine, startKey, endKey, regionSplitSize, regionSplitKeys)
```
lance6716, line 1656: there's also a PaginateScanRegion inside writeAndIngestByRange. Because we just get the region of every key range in the new scanRegion, I think we can skip the PaginateScanRegion inside writeAndIngestByRange.

It may be a bit hard for you to change the related codes, and we expect the code changes should be made on master branch first. Maybe we can implement your idea on master branch in out separate PR. We will discuss soon.

br/pkg/lightning/backend/local/local.go
```diff
@@ -1541,53 +1604,89 @@ func (local *local) writeAndIngestByRanges(ctx context.Context, engine *Engine,
...
-				if backOffTime > maxRetryBackoffTime {
-					backOffTime = maxRetryBackoffTime
-				}
-				select {
-				case <-time.After(backOffTime):
-				case <-ctx.Done():
-					return
-				}
+		// only take to leader's store id fo the first region
+		if len(regions) > 0 {
```
lance6716, line 1619: one key range may have multiple regions, because regions are split by other logic previously. That's why writeAndIngestByRange we have a separate PaginateScanRegion and handle each region in writeAndIngestPairs.

So your current PR is not very precise. It's still an improvement.

br/pkg/lightning/backend/local/local.go
guoshouyan: cool, if you could help to implement on master branch, I could cherry pick it to our 6.5 branch

br/pkg/lightning/backend/local/local.go
guoshouyan: yeah, I just want to make the logic simple so I only consider the leader store id of the first region. I'm trying a solution which consider all peer id of all regions. I'm currently testing its performance

guoshouyan: updated in : https://github.com/pingcap/tidb/pull/54976

PR #54627, opened 2024-07-15
meta: fix memory leak when loading infoschema v2

tiancaiamao:

Issue Number: ref #50959 

Problem Summary:

Start tidb with 1M tables, I found some unexpected memory usage.
That could be a leak.

image

When TiDB start up using infoschema v2, it builds the name-> id mapping by loading the meta data, and store the mapping in memory.
The map key (name) is a []byte slice taken from the original kvpair, and we use hack.String for optimization.
So the key references the original  kvrpcpb.Response causing the Go GC can't release it, result in a memory leak.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

Get the heap profile and compare them.
After this commit, the occpuation of kvrpcpb.XXX is gone:

image

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -1062,7 +1062,8 @@ func GetAllNameToIDAndSpecialAttributeInfo(m *Meta, dbID int64) (map[string]int6
 			return errors.Trace(err)
 		}
 
-		res[Unescape(nameLMatch[1])] = int64(id)
+		key := Unescape(nameLMatch[1])
+		res[strings.Clone(key)] = int64(id)
 		if CheckSpecialAttributes(string(hack.String(value))) {
 			tbInfo := &model.TableInfo{}
 			err = json.Unmarshal(value, tbInfo)


pkg/meta/meta.go
```diff
@@ -1062,7 +1062,8 @@ func GetAllNameToIDAndSpecialAttributeInfo(m *Meta, dbID int64) (map[string]int6
 			return errors.Trace(err)
 		}
 
-		res[Unescape(nameLMatch[1])] = int64(id)
+		key := Unescape(nameLMatch[1])
+		res[strings.Clone(key)] = int64(id)
```
lance6716, line 1066: if `Unescape` really modified the string, we don't need to clone it anymore. How about we change line 1059 to FindSubmatch and other variables to `[]byte`, and we let `Unescape` must clone the input `[]byte` even if it's untouched?

tiancaiamao: We clone it because after several complex operation, the  actual memory of `key` is a slice from the kvrpcpb.Response packet ... 

=> kvrpcpb.Response ...   
=> kvpair
=> key.   value  
=> ....       value []byte
=>  regexp on value to get nameLMatch
=>  nameLMatch[1]
=>  Unmarshal(nameLMatch[1]).    
=> map[key]..........   key is still referencing kvrpcpb.Response

lance6716 (approved):

wjhuang2016 (approved):

PR #54630, opened 2024-07-15
executor: Optimize slow log parsing's splitByColon function

### LINKED ISSUE #54538: Slow log parse performance
## Enhancement
Do cpu profile for query like:
```
SELECT
  Digest,
  Query,
  Conn_ID,
  (UNIX_TIMESTAMP(Time) + 0E0) AS timestamp,
  Query_time,
  Mem_max,
  Process_keys
FROM
  `INFORMATION_SCHEMA`.`CLUSTER_SLOW_QUERY`
WHERE
  Time BETWEEN FROM_UNIXTIME(?)
  AND FROM_UNIXTIME(?)
ORDER BY
  Query_time DESC
LIMIT 100;
```
Find the splitByColon takes about 80% cpu times:
img_v3_02cl_7c4b6024-a3b3-4168-b5c4-86858cb0605g

yibin87:

Issue Number: close #54538

Problem Summary:

Replace regexp matching with simple string comparison operations. Besides, previously, ":=" inside "{}" are not handled correctly, fix it in this PR also.
In local mannual test, performance will improve about 10x, from 24s to 2.2s for the following sql:
```SELECT   Digest,   Query,   Conn_ID,   (UNIX_TIMESTAMP(Time) + 0E0) AS timestamp,   Query_time,   Mem_max,   Process_keys FROM   `INFORMATION_SCHEMA`.`CLUSTER_SLOW_QUERY` WHERE   Time BETWEEN FROM_UNIXTIME(1720471890)   AND FROM_UNIXTIME(1720515091) ORDER BY   Query_time DESC LIMIT   100;```
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/slow_query.go
+++ b/pkg/executor/slow_query.go
@@ -525,9 +525,9 @@ func getLineIndex(offset offset, index int) int {
 // findMatchedRightBracket returns the rightBracket index which matchs line[leftBracketIdx]
 // leftBracketIdx should be valid string index for line
 // Returns -1 if invalid inputs are given
-func findMatchedRightBracket(line string, leftBracketIdx int) int {
+func findMatchedRightBracket(line []rune, leftBracketIdx int) int {
 	leftBracket := line[leftBracketIdx]
-	rightBracket := byte('}')
+	rightBracket := '}'
 	if leftBracket == '[' {
 		rightBracket = ']'
 	} else if leftBracket != '{' {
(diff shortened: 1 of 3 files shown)


pkg/executor/slow_query.go
```diff
@@ -523,32 +522,94 @@ func getLineIndex(offset offset, index int) int {
...
+func isLetterOrNumeric(b byte) bool {
+	return ('A' <= b && b <= 'Z') || ('a' <= b && b <= 'z') || ('0' <= b && b <= '9')
+}
 
 // splitByColon split a line like "field: value field: value..."
+// Note:
+// 1. Both field and value string contain only ANSI characters
+// 2. value string may be surrounded by brackets, allowed brackets includes "[]" and "{}",  like {key: value,{key: value}}
+// "[]" can only be nested inside "[]"; "{}" can only be nested inside "{}"
+// 3. value string can't contain ' ' character unless it is inside brackets
```
xzhangxian1008, line 568: Is origin implementation also contains these restriction?

yibin87: No, while current slow log satisfy these restrictions and it doesn't seem to be changed frequently in future. Previous implementation provides a broader functionality.

pkg/executor/slow_query_test.go
```diff
@@ -550,6 +550,26 @@ func TestSplitbyColon(t *testing.T) {
...
+		{
+			"1a: {2b,3c} 4d: 5e",
+			[]string{"1a", "4d"},
+			[]string{"{2b,3c}", "5e"},
+		},
+		{
+			"1a: {2b,{3c: 3cc}} 4d: 5e",
+			[]string{"1a", "4d"},
+			[]string{"{2b,{3c: 3cc}}", "5e"},
+		},
```
xzhangxian1008, line 572: - Add some invalid cases?
- Maybe we can add some cases that multi left brackets stay together such as "{{{" or "[[["

pkg/executor/slow_query.go
```diff
@@ -523,32 +522,94 @@ func getLineIndex(offset offset, index int) int {
...
+				return -1
+			}
+		} else {
+			current++
+		}
+	}
+	return -1
+}
+
+func isLetterOrNumeric(b byte) bool {
```
xzhangxian1008, line 559: Maybe we can put this function in pkg/util/stringutil file.

pkg/executor/slow_query.go
```diff
@@ -523,32 +522,94 @@ func getLineIndex(offset offset, index int) int {
...
+	return ('A' <= b && b <= 'Z') || ('a' <= b && b <= 'z') || ('0' <= b && b <= '9')
+}
 
 // splitByColon split a line like "field: value field: value..."
+// Note:
+// 1. Both field and value string contain only ANSI characters
```
xhebox, line 565: That is too restricted. E.g. tidb_redact_log may output non-ansi characters later sometime.
I am thinking if we could just use `strings.Index(a, b)` to mimic the original `regexp`.

yibin87: I see, this rule can be removed for current algorithm, I'll change it.
BTW, splitByColon is not the root entry for slow log parsing, for logs that may contain non-ansi characters we can handle it separately if we doneed it.

pkg/executor/slow_query_test.go
yibin87: Good suggestion, will add it.

pkg/executor/slow_query.go
yibin87: Checked that whitespace character can't be a part of a valid non-whitespace utf8 character. Thus value string can contain non ascii characters except ' '.  And whitespace can be contained inside "{}"/"[]" in value string. For field string, first character should be letters or digits.

pkg/executor/slow_query_test.go
yibin87: Done

pkg/executor/slow_query.go
yibin87: Just leave it in the local file, since it may be changed for slow log parsing only.

xhebox (approved):

xzhangxian1008 (approved):
LGTM

PR #54670, opened 2024-07-17
session: bind extensions to bootstrap session

### LINKED ISSUE #54671: Cannot use extension auth plugin in init sql script
## Bug Report
Currently, users cannot create users with extension auth plugin in the init sql script. If they do so, TiDB will complain about plugin not loaded during the execution and causing the SQL statements to fail.

1. Write an auth plugin using the extension system following the example in https://github.com/pingcap/tidb/blob/master/pkg/extension/README.md
2. Create an init sql file with content:
```sql
CREATE USER myuser IDENTIFIED WITH 'my_auth_plugin' AS 'mypwd';
```
3. Start the server with this init SQL file

Logging in as the user should succeed since the user should be created successfully.

Logging in fails because the user is not created. The `CREATE USER` statement hits the following error:
```
[2024/07/16 21:54:04.259 -07:00] [INFO] [session.go:4008] ["CRUCIAL OPERATION"] [conn=0] [schemaVersion=51] ["secure text"="create user {myuser@% password = ***}"] [user=]
[2024/07/16 21:54:04.259 -07:00] [INFO] [tidb.go:269] ["rollbackTxn called due to ddl/autocommit failure"]
[2024/07/16 21:54:04.259 -07:00] [WARN] [session.go:2162] ["run statement failed"] [schemaVersion=51] [error="[executor:1524]Plugin 'my_auth_plugin' is not loaded"] [session="{\n  \"currDBName\": \"\",\n  \"id\": 0,\n  \"status\": 2,\n  \"strictMode\": true,\n  \"user\": null\n}"]
[2024/07/16 21:54:04.259 -07:00] [WARN] [bootstrap.go:3249] ["InitializeSQLFile error"] [error="[executor:1524]Plugin 'my_auth_plugin' is not loaded"]
```

```
mysql> SELECT tidb_version();
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| tidb_version()                                                                                                                                                                                                                                                                       |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Release Version: v8.2.0-alpha-606-gecf9031c1b-dirty
Edition: Community
Git Commit Hash: ecf9031c1b8e8243ac9dd7cc1439818d4d85e160
Git Branch: extbootstrap
UTC Build Time: 2024-07-17 04:53:08
GoVersion: go1.21.6
Race Enabled: false
Store: unistore |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

yzhan1:

Issue Number: close #54671

Problem Summary:

When bootstrapping with init SQL file, all extensions are not bind to the session object used for bootstrapping. This means if the init SQL file has any SQL statements which creates users using extension auth plugins, it will fail to execute because the executor can't find the plugin implementation.

This PR fixes it by properly binding all extensions to the session context used for bootstrapping.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
  > Repeat the steps in #54671 and observe that the errors are gone and user is created
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue where extension auth plugin used in init sql file will cause the file execution to fail
```

### DIFF
--- a/pkg/session/bootstrap.go
+++ b/pkg/session/bootstrap.go
@@ -35,6 +35,7 @@ import (
 	"github.com/pingcap/tidb/pkg/domain"
 	"github.com/pingcap/tidb/pkg/domain/infosync"
 	"github.com/pingcap/tidb/pkg/expression"
+	"github.com/pingcap/tidb/pkg/extension"
 	"github.com/pingcap/tidb/pkg/infoschema"
 	"github.com/pingcap/tidb/pkg/kv"
 	"github.com/pingcap/tidb/pkg/meta"
@@ -3216,6 +3217,13 @@ func doDDLWorks(s sessiontypes.Session) {
 // doBootstrapSQLFile executes SQL commands in a file as the last stage of bootstrap.
 // It is useful for setting the initial value of GLOBAL variables.
 func doBootstrapSQLFile(s sessiontypes.Session) error {
+	if s.GetExtensions() == nil {
+		extensions, err := extension.GetExtensions()
+		if err != nil {
+			return err
+		}
+		s.SetExtensions(extensions.NewSessionExtensions())
+	}
 	sqlFile := config.GetGlobalConfig().InitializeSQLFile
 	ctx := kv.WithInternalSourceType(context.Background(), kv.InternalTxnBootstrap)
 	if sqlFile == "" {
--- a/pkg/session/bootstrap.go
+++ b/pkg/session/bootstrap.go
@@ -35,7 +35,6 @@ import (
 	"github.com/pingcap/tidb/pkg/domain"
 	"github.com/pingcap/tidb/pkg/domain/infosync"
 	"github.com/pingcap/tidb/pkg/expression"
-	"github.com/pingcap/tidb/pkg/extension"
 	"github.com/pingcap/tidb/pkg/infoschema"
 	"github.com/pingcap/tidb/pkg/kv"
 	"github.com/pingcap/tidb/pkg/meta"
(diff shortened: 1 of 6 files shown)


yzhan1:
@lcwangchao @CbcWestwolf @bb7133 PTAL. Thanks!

CbcWestwolf:
Please add a new UT to cover this scenario. You can take https://github.com/pingcap/tidb/blob/ca7ba14da87abd060922b32c893b121c4c3d6373/pkg/session/test/session_test.go#L788 as a reference. :-)

pkg/session/bootstrap.go
```diff
@@ -3216,6 +3217,13 @@ func doDDLWorks(s sessiontypes.Session) {
 // doBootstrapSQLFile executes SQL commands in a file as the last stage of bootstrap.
 // It is useful for setting the initial value of GLOBAL variables.
 func doBootstrapSQLFile(s sessiontypes.Session) error {
+	if s.GetExtensions() == nil {
```
lcwangchao, line 3220: This fix seems a little tricky. Before this PR, all internal session seems do not have a `SessionExtension` set. This PR only sets the session's inner extension when we use it to bootstrap the SQL file that makes some differences of sessions in the pool. Though it will not cause any bug now, but in the future if the extension in internal session has some bug, it will be hard to locate and debug.

A better way is to make sure all internal SQLs are created with internal extension, but seems it has many code to change and to test.

How about using a global extension by `extension.GetExtensions()` when create/alter a user?

yzhan1: > How about using a global extension by extension.GetExtensions() when create/alter a user?

My concern for this is `extension.GetExtensions()` requires locking every time and may cause performance issue when create/alter user is getting called with high concurrency.

Maybe instead of using the shared session `ses[9]` for bootstrapping, we can create a new session inside https://github.com/pingcap/tidb/blob/c4fd1b212c0ca6572e3ad6f18fdcdd58354f629c/pkg/session/session.go#L3559
with extensions bind to it specifically for bootstrap usage?

lcwangchao: Performance isn't a big deal, IMO. You can see that it uses a read-write lock:

https://github.com/pingcap/tidb/blob/2da324d1a1831c7be31ef17a164ff511757cdbb5/pkg/extension/registry.go#L45-L57

For most times, the calls of `GetExtensions` will not block each other because the inner state has been setup and the only read lock is used.

Compared with other parts of DDL such as the network, the cost of `GetExtensions` is insignificant.

lcwangchao (approved):

yzhan1:
@CbcWestwolf could you take a look again? Thanks!

CbcWestwolf (approved):

PR #54581, opened 2024-07-11
infoschema: use maps.Clone to replace iterating

D3Hunter:
### What problem does this PR solve?

Issue Number: ref #54436

Problem Summary:

### What changed and how does it work?
- in https://go-review.googlesource.com/c/go/+/471400, maps.Clone is optimized to clone underlying buckets, not iterating each item as before. our info schema load part is the perfect place for this
- remove the Reload inside OnChangedof DefaultCallback which is used in normal code path, we have already waited each instance to reload, no need to call it again, will remove it in later refactor. reload inside OnSchemaStateChanged is for test, but i don't see the need to do so, as we have waited schema change before this call, so remove it too. And as fast-create depends on reload after DDL execution, we add a reload for it separately
https://github.com/pingcap/tidb/blob/a4908820a38f6e893c1882fdcf2c3a6d56141dba/pkg/ddl/job_table.go#L625-L636

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)

same test as in https://github.com/pingcap/tidb/pull/54438, run 3 times, takes 22m30/21m40/20m44. unlike prev PR, the speed is quite stable. 

compare the time to load schema diff, before
image
after
img_v3_02cm_0b13db42-e823-43bc-a792-8f0df25baccg

- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #54693, opened 2024-07-17
ddl: query since min job id

D3Hunter:

Issue Number: ref #54436

Problem Summary:

second pr mentioned in https://github.com/pingcap/tidb/pull/54547/, now we pessimistic lock global id, and generate global id and insert job in one txn, so we can make sure when job of ID A which is the minimum in a txn submitted, no jobs of ID < A can be inserted, so it's safe to query since min job id
- query since min job id to mitigate slowness due to un-compacted deleted kv, see this issue https://github.com/pingcap/tidb/issues/52905. by query since min job id, tikv can directly seek to the starting KV and skip all those that have deleted but not compacted without scanning

### Check List

Tests 

- [x] Unit test
- [ ] Integration test

same case as in https://github.com/pingcap/tidb/pull/54581, takes 11m24, and there is very little speed degradation now

image

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/ddl/ddl_worker.go
```diff
@@ -411,16 +411,17 @@ func (d *ddl) addBatchDDLJobs(jobWs []*JobWrapper) error {
 		return nil
 	}
 
+	ctx := kv.WithInternalSourceType(context.Background(), kv.InternalTxnDDL)
```
lance6716, line 414:
```suggestion
	ctx := kv.WithInternalSourceType(d.ctx, kv.InternalTxnDDL)
```

pkg/ddl/job_table.go
```diff
@@ -97,12 +97,14 @@ var _ owner.Listener = (*ownerListener)(nil)
...
 	l.scheduler = &jobScheduler{
-		schCtx:       ctx,
-		cancel:       cancelFunc,
-		runningJobs:  newRunningJobs(),
-		sysTblMgr:    systable.NewManager(l.ddl.sessPool),
-		schemaLoader: l.ddl.schemaLoader,
+		schCtx:            ctx,
+		cancel:            cancelFunc,
+		runningJobs:       newRunningJobs(),
+		sysTblMgr:         sysTblMgr,
```
lance6716, line 105: can we use ddl.sysTblMgr?

lance6716 (approved):

pkg/ddl/job_table.go
D3Hunter: it's just a wrapper around session pool, both lgtm

tangenta (approved):

PR #54743, opened 2024-07-18
infoschema: optimize the performance of TableByName for infoschemav2

tiancaiamao:

Issue Number: ref #50959 

Problem Summary:

I find some performance issue of the TableByName API.
As the flamegraph shows, all the calling of `TableByName` goes to `mallocgc`
image

It's cause by the old style of writting like:

```
var end tableItem
var schemaVersion
var ok bool
var itm tableItem

is.byName.Descend(end, func() {             //   <-  this closure is allocated every time!
       if item.dbName != h.end.dbName || item.tableName != end.tableName {
               ok = false
               return false
       }
       if item.schemaVersion <= schemaVersion {
               if !item.tomb { // If the item is a tomb record, the database is dropped.
                       ok = true
                       itm = item
               }
               return false
       }
       return true
}
```

This closure function is a large burden for GC, it need to be allocated every time.

In this PR, we use and the closure allocation is avoid.

```
is.byName.Descend(end, helper.iter)
```

### Check List

Tests 

- [X] Unit test

```
cd pkg/infoschema/test/infoschemav2test
go test -tags intest -run XXX -bench BenchmarkTableByName -benchmem
```

Before VS After:

```
 2533712               450.9 ns/op           195 B/op          3 allocs/op
 3274862               360.9 ns/op             2 B/op          0 allocs/op
```

- [ ] Integration test

You might think the differ is not much? No!
I tested some integration test scenario, one goroutine run "truncate table t" in a loop, 
the other goroutine run "select * from t" and observe the latency.

The QPS before and after this commit:

image

TableByName metrics before fix:

image

TableByName metrics after fix:

image

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/infoschema/infoschema_v2.go
+++ b/pkg/infoschema/infoschema_v2.go
@@ -455,6 +455,28 @@ func (is *infoschemaV2) EvictTable(schema, tbl string) {
 	is.tableCache.Remove(tableCacheKey{itm.tableID, itm.schemaVersion})
 }
 
+type helper struct {
+	end           tableItem
+	schemaVersion int64
+	ok            bool
+	itm           tableItem
+}
+
+func (h *helper) iter(item tableItem) bool {
+	if item.dbName != h.end.dbName || item.tableName != h.end.tableName {
+		h.ok = false
+		return false
+	}
+	if item.schemaVersion <= h.schemaVersion {
+		if !item.tomb { // If the item is a tomb record, the database is dropped.
+			h.ok = true
+			h.itm = item
+		}
+		return false
+	}
+	return true
+}
+
 func (is *infoschemaV2) TableByName(ctx context.Context, schema, tbl model.CIStr) (t table.Table, err error) {
 	if IsSpecialDB(schema.L) {
 		if raw, ok := is.specials.Load(schema.L); ok {
@@ -467,11 +489,16 @@ func (is *infoschemaV2) TableByName(ctx context.Context, schema, tbl model.CIStr
 	}
 
 	start := time.Now()
-	eq := func(a, b *tableItem) bool { return a.dbName == b.dbName && a.tableName == b.tableName }
-	itm, ok := search(is.byName, is.infoSchema.schemaMetaVersion, tableItem{dbName: schema.L, tableName: tbl.L, schemaVersion: math.MaxInt64}, eq)
-	if !ok {
+
+	var h helper
+	h.end = tableItem{dbName: schema.L, tableName: tbl.L, schemaVersion: math.MaxInt64}
+	h.schemaVersion = is.infoSchema.schemaMetaVersion
+	is.byName.Descend(h.end, h.iter)
+
+	if !h.ok {
 		return nil, ErrTableNotExists.FastGenByArgs(schema, tbl)
 	}
+	itm := h.itm
 
 	// Get from the cache with old key
 	oldKey := tableCacheKey{itm.tableID, itm.schemaVersion}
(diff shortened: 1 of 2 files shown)


pkg/infoschema/infoschema_v2.go
```diff
@@ -455,6 +455,28 @@ func (is *infoschemaV2) EvictTable(schema, tbl string) {
 	is.tableCache.Remove(tableCacheKey{itm.tableID, itm.schemaVersion})
 }
 
+type helper struct {
+	end           tableItem
+	schemaVersion int64
+	ok            bool
```
ngaut, line 461: ok as a field name is confusing, what does it mean?

tiancaiamao: `ok` here is used to check whether the object is found or not.
Since the return object is a `tableItem` type, not a `*tableItem`, we need this extra flag.

```
var ok bool
var res tableItem

btree.Descend(end, func(item tableItem) {
if .. {
	ok = true
	res = item
})
```

You can view it as 

```
	res, ok := searchTheBtree(...)
	if !ok {
		...
	}
```

lance6716 (approved):

pkg/infoschema/infoschema_v2.go
```diff
@@ -455,6 +455,28 @@ func (is *infoschemaV2) EvictTable(schema, tbl string) {
 	is.tableCache.Remove(tableCacheKey{itm.tableID, itm.schemaVersion})
 }
 
+type helper struct {
```
D3Hunter, line 458:
```suggestion
type tableByNameHelper struct {
```

pkg/infoschema/infoschema_v2.go
D3Hunter: prefer we name it as `found`, more clear

pkg/infoschema/infoschema_v2.go
```diff
@@ -455,6 +455,28 @@ func (is *infoschemaV2) EvictTable(schema, tbl string) {
...
+	end           tableItem
+	schemaVersion int64
+	ok            bool
+	itm           tableItem
+}
+
+func (h *helper) iter(item tableItem) bool {
```
D3Hunter, line 465:
```suggestion
func (h *helper) onItem(item tableItem) bool {
```

D3Hunter: i don't like the name in `btree` pkg, `iter` mostly means something with open/next/close

pkg/infoschema/infoschema_v2.go
bb7133: +1 for `found`.

pkg/infoschema/infoschema_v2.go
```diff
@@ -455,6 +455,28 @@ func (is *infoschemaV2) EvictTable(schema, tbl string) {
...
+	if item.dbName != h.end.dbName || item.tableName != h.end.tableName {
+		h.ok = false
+		return false
+	}
+	if item.schemaVersion <= h.schemaVersion {
```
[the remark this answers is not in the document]
tiancaiamao: It might happen when we're using historial read, or set tidb_snapshot, we may ask for the old infoschema, and the latest schema version has bumped. The btree records all the history version.

pkg/infoschema/infoschema_v2.go
tiancaiamao: Done

D3Hunter (approved):

PR #54831, opened 2024-07-23
*: reset infoschema v2 data before a full load

### LINKED ISSUE #54796: Test TestExchangePartitionWithPlacement fails after injecting full reload
## Bug Report

Enable info schema cache V2. 

Modify codes
```
func (do *Domain) tryLoadSchemaDiffs(m *meta.Meta, usedVersion, newVersion int64, startTS uint64) (infoschema.InfoSchema, *transaction.RelatedSchemaChange, []string, error) {
	var diffs []*model.SchemaDiff
	for usedVersion < newVersion {
		usedVersion++
		diff, err := m.GetSchemaDiff(usedVersion)
		if err != nil {
			return nil, nil, nil, err
		}
		if diff == nil {
			// Empty diff means the txn of generating schema version is committed, but the txn of `runDDLJob` is not or fail.
			// It is safe to skip the empty diff because the infoschema is new enough and consistent.
			logutil.BgLogger().Info("diff load InfoSchema get empty schema diff", zap.Int64("version", usedVersion))
			do.infoCache.InsertEmptySchemaVersion(usedVersion)
			continue
		}
		diffs = append(diffs, diff)
	}
	if diffs[0].Type == model.ActionExchangeTablePartition {
		return nil, nil, nil, errors.New("mock error")
	}
	builder, err := infoschema.NewBuilderV3(do, do.sysFacHack, do.infoCache.Data).InitWithOldInfoSchema(do.infoCache.GetLatest())
	if err != nil {
		return nil, nil, nil, errors.Trace(err)
	}
```
Inject mock error if DDL type is ActionExchangeTablePartition.
Test TestExchangePartitionWithPlacement passes.
Test TestExchangePartitionWithPlacement fails.
master

tiancaiamao:

Issue Number: close #54796

close https://github.com/pingcap/tidb/issues/54856

Problem Summary:

As the comment says:

https://github.com/pingcap/tidb/compare/master...tiancaiamao:issue54796?expand=1#diff-e33e11641dcb529d51aab53c02179aa98f1261cf704d4dfb6189e96489b4f2caR869-R887

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

- [X] Performance regression: Consumes more CPU

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/infoschema/infoschema_v2.go
```diff
@@ -250,6 +250,121 @@ func (isd *Data) deleteDB(dbInfo *model.DBInfo, schemaVersion int64) {
 	isd.schemaID2Name.Set(schemaIDName{schemaVersion: schemaVersion, id: dbInfo.ID, name: dbInfo.Name.O, tomb: true})
 }
 
+// resetBeforeFullLoad is called before a full recreate operation within builder.InitWithDBInfos().
+// TODO: write a generics version to avoid repeated code.
```
tiancaiamao, line 254: In fact, I tried to do that  and failed, just wast triple time than the non-generic version.

tiancaiamao:
/retest

https://github.com/pingcap/tidb/issues/54846

pkg/domain/domain.go
```diff
@@ -548,6 +548,24 @@ func (do *Domain) tryLoadSchemaDiffs(m *meta.Meta, usedVersion, newVersion int64
 		}
 		diffs = append(diffs, diff)
 	}
+
+	failpoint.Inject("MockTryLoadDiffError", func(val failpoint.Value) {
+		switch val.(string) {
+		case "exchangepartition":
```
Benjamin2037, line 554: Do we need consider other scenarios?

tiancaiamao: Yes, but we can add more tests after we find something.
This PR is mainly for https://github.com/pingcap/tidb/issues/54796

Benjamin2037:
Will the reset operations cause performance problems when there are 1M tables in the cluster?

pkg/infoschema/infoschema_v2.go
```diff
@@ -251,6 +251,206 @@ func (isd *Data) deleteDB(dbInfo *model.DBInfo, schemaVersion int64) {
...
+		items = items[:0]
+	}
+}
+
+func resetTableInfoResidentBeforeFullLoad(bt *btree.BTreeG[tableInfoItem], schemaVersion int64) {
+	pivot, ok := bt.Max()
+	if !ok {
+		return
+	}
+	items := make([]tableInfoItem, 0, bt.Len())
```
Benjamin2037, line 349: Why do not use batch-size to implement multi-round reset?

Benjamin2037 (approved):
rest LGTM

Benjamin2037:
I hold because there are some comments need to be reply.

pkg/infoschema/infoschema_v2.go
tiancaiamao: We know some btree may be very large, for example the byName and byID could have 1M entries.
And we assume some btree should not be too large, for example here the TableInfoResident are those table with special attributes.

Using batch-size we have an extra loop -- we need to visit Descend() *twice at least* to break the for-loop.
So that is not always better than the non-batch-size way.

The first version I wrote is the non-batch way, and copy the code several times.  
Then the batch way, only some of them are updated when I think it's necessary.  @Benjamin2037

pkg/infoschema/builder.go
```diff
@@ -866,6 +866,25 @@ func (b *Builder) InitWithDBInfos(dbInfos []*model.DBInfo, policies []*model.Pol
...
+		// We must not clear the historial versions like b.infoData = NewData(), because losing
+		// the historial versions would cause applyDiff get db not exist error and fail, then
+		// infoschema reloading retries with full load every time.
+		// See https://github.com/pingcap/tidb/issues/53442
+		//
+		// We must reset it, otherwise the stale tables remain and cause bugs later.
+		// For example, schema version 59:
+		//         107: t1
+		//         112: t2 (partitions p0=113, p1=114, p2=115)
+		// operation: alter table t2 exchange partition p1 with table t1
```
D3Hunter, line 879:
```suggestion
		// operation: alter table t2 exchange partition p0 with table t1
```

PR #53831, opened 2024-06-05
config: Change tidb_opt_projection_push_down default value to true | tidb-test=pr/2359

yibin87:
### What problem does this PR solve?

Issue Number: ref #51876 

Problem Summary:

Described in #51876 

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

benchbot test performance, no significant performance differences:
2irsq7zFnl

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
Change the default value of tidb_opt_projection_push_down to true so that TIDB can choose to push down projections to TIKV to reduce network bandwith usage. 
```

Issue #51876, opened 2024-03-19
Projection pushdown to tikv feature GA


yibin87:
## Enhancement
tidb_opt_projection_push_down is introduced by #27029 , and is set 'off' by default.
It aimed to improve performance in two sides:

1. Utilize distributed tikv nodes to perform calculations parallelly
2. Provide a elegantly way to reduce unused columns which are transformed back to tidb from tikv #44902 

However, there are still some functionality issues when this feature is set 'on':
- [ ] #51874

Issue #34993, opened 2022-05-27
Projection push down make tikv crashed.


wshwsh12:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
Please set **tidb_opt_projection_push_down = on**;

```
use test;
drop table if exists UK_GCOL_STORED_176092;
CREATE TABLE `UK_GCOL_STORED_176092` (
  `COL1` bit(16) DEFAULT NULL,
  UNIQUE KEY `UK_COL1` (`COL1`) /*!80000 INVISIBLE */
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
insert into UK_GCOL_STORED_176092 values(0x15AC), (0xB0B), (0xBC9), (0x11A3), (0x1055), (0x5CB), (0xC3C), (0xB25), (0x88D), (0x1220), (0x156F), (0x975), (0x190F), (0x11E6), (0x93C), (0x14B1), (0x1889), (0x71B), (0x1484), (0x133C), (0x577), (0x907), (0x8A3), (0x1598), (0xE45), (0x4C6), (0x12A2), (0x1917), (0xE57), (0x1227), (0x4D2), (0x16B8), (0x7E0), (0x990), (0x18CF), (0x233), (0x357), (0x351), (0xF01), (0x13D5), (0xCED), (0x124D), (0xB01), (0x1992), (0x7D5), (0x11AD), (0x804), (0x830), (0xBA0), (0x13FF), (0x144F), (0xE7D);
set tidb_opt_projection_push_down = on;
select cast(col1 as signed) from  UK_GCOL_STORED_176092 where cast(col1 as signed) > 3709;

[2022/05/26 19:57:46.386 +08:00] [INFO] [scheduler.rs:548] ["get snapshot failed"] [err="Error(Request(message: \"EpochNotMatch current epoch of region 2 is conf_ver: 1 version: 34, but you sent conf_ver: 1 version: 33\" epoch_not_match { current_regions { id: 2 start_key: 7480000000000000FF4500000000000000F8 region_epoch { conf_ver: 1 version: 34 } peers { id: 3 store_id: 1 } } current_regions { id: 68 start_key: 7480000000000000FF4300000000000000F8 end_key: 7480000000000000FF4500000000000000F8 region_epoch { conf_ver: 1 version: 34 } peers { id: 69 store_id: 1 } } }))"] [cid=350]
[...]

PR #54861, opened 2024-07-24
session, config, sessionctx, executor: Set tidb_opt_projection_push_down to false when tidb upgrades from versions older than 8.3.0 to 8.3.0 and later

yibin87:

Issue Number: ref #51876 #53831

Problem Summary:

In #53831, we change tidb_opt_projection_push_down session variable's default value to 'on'. After discussion with PM, we need to keep consistency with older versions and avoid any unknown potential problems, we decide to keep the 'off' value for tidb cluster upgradation. Thus we change the variable to both session and global scope. Besides, this PR deprecates the previous config item "projection-push-down", we can do this because this config would have never been used by customers due to #34993 issue. 
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/sessionctx/variable/sysvar.go
+++ b/pkg/sessionctx/variable/sysvar.go
@@ -211,7 +211,7 @@ var defaultSysVars = []*SysVar{
 		}
 		return nil
 	}},
-	{Scope: ScopeSession, Name: TiDBOptProjectionPushDown, Value: BoolToOnOff(config.GetGlobalConfig().Performance.ProjectionPushDown), Type: TypeBool, SetSession: func(s *SessionVars, val string) error {
+	{Scope: ScopeGlobal | ScopeSession, Name: TiDBOptProjectionPushDown, Value: BoolToOnOff(DefOptEnableProjectionPushDown), Type: TypeBool, SetSession: func(s *SessionVars, val string) error {
 		s.AllowProjectionPushDown = TiDBOptOn(val)
 		return nil
 	}},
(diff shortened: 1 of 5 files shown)


windtalker (approved):
LGTM

hawkingrei (approved):

pkg/sessionctx/variable/sysvar.go
```diff
@@ -211,7 +211,7 @@ var defaultSysVars = []*SysVar{
 		}
 		return nil
 	}},
-	{Scope: ScopeSession, Name: TiDBOptProjectionPushDown, Value: BoolToOnOff(config.GetGlobalConfig().Performance.ProjectionPushDown), Type: TypeBool, SetSession: func(s *SessionVars, val string) error {
+	{Scope: ScopeGlobal | ScopeSession, Name: TiDBOptProjectionPushDown, Value: BoolToOnOff(DefOptEnableProjectionPushDown), Type: TypeBool, SetSession: func(s *SessionVars, val string) error {
```
hawkingrei, line 214: Why do we need to change this? It appears that the old behavior has changed.

yibin87: To allow different init values for new deployed 8.3.0 tidb cluster and upgraded 8.3.0 tidb cluster, need to change to global session scope. Besides, it makes sense to be a global session scope system variable also.

winoros (approved):

Issue #53460, opened 2024-05-22
concurrency auto analyze


hawkingrei:
## Enhancement

Now, TiDB's analyze can only be executed by a single goroutine. so when we use millions of tables, it will be very slow, so we need to implement the concurrently auto analyze for raising the QPS of auto analyze.

PR #54902, opened 2024-07-25
*: remove limiting process id for auto analyze

hawkingrei:

Issue Number: ref #53460

Problem Summary:

- Add ```GlobalAutoAnalyzeProcessList``` to collect all auto analyze process IDs.
- Remove the restriction of having only one process ID for auto analyze and release the used process ID.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/analyze_col_v2.go
```diff
@@ -854,7 +855,9 @@ func readDataAndSendTask(ctx sessionctx.Context, handler *tableResultHandler, me
 	for {
 		failpoint.Inject("mockKillRunningV2AnalyzeJob", func() {
```
hawkingrei, line 856: Here are the tests.

pkg/domain/domain.go
```diff
@@ -2840,9 +2840,8 @@ func (do *Domain) ReleaseConnID(connID uint64) {
 }
 
 // GetAutoAnalyzeProcID returns processID for auto analyze
-// TODO: support IDs for concurrent auto-analyze
 func (do *Domain) GetAutoAnalyzeProcID() uint64 {
```
winoros, line 2843: This function is not used after this patch?

pkg/statistics/handle/util/auto_analyze_proc_id_generator.go
```diff
@@ -14,27 +14,112 @@
...
 // AutoAnalyzeProcIDGenerator is used to generate auto analyze proc ID.
 type AutoAnalyzeProcIDGenerator interface {
 	// AutoAnalyzeProcID generates an analyze ID.
 	AutoAnalyzeProcID() uint64
+	ReleaseAutoAnalyzeProcID(uint64)
 }
 
 var _ AutoAnalyzeProcIDGenerator = (*generator)(nil)
 
 type generator struct {
```
winoros, line 33: Why is it a generator? Seems that the process is generated by the `domain.NextConnID`

pkg/domain/domain.go
hawkingrei: Yes, have removed.

pkg/statistics/handle/util/auto_analyze_proc_id_generator.go
hawkingrei: I think it just wants to isolate the dependency on the domain and provide a container to reference the application and release of the connect id.

xhebox (approved):

pingyu (approved):
[`reservedConnAnalyze`](https://github.com/pingcap/tidb/blob/56710aeaef70943186d4e0fa7bc233b9834fe8cb/pkg/domain/domain.go#L2859) can be removed. Rest LGTM~

hawkingrei:
> [`reservedConnAnalyze`](https://github.com/pingcap/tidb/blob/56710aeaef70943186d4e0fa7bc233b9834fe8cb/pkg/domain/domain.go#L2859) can be removed. Rest LGTM~

I have removed it. Thanks.

lance6716 (approved):

pkg/statistics/handle/util/auto_analyze_proc_id_generator.go
```diff
@@ -14,27 +14,111 @@
...
+// ReleaseAutoAnalyzeProcID implements AutoAnalyzeProcIDGenerator.
+func (g *generator) ReleaseAutoAnalyzeProcID(id uint64) {
+	g.releaseAutoAnalyzeProcID(id)
+}
+
+// GlobalAutoAnalyzeProcessList is used to track the auto analyze process.
+var GlobalAutoAnalyzeProcessList = newGlobalAutoAnalyzeProcessList()
+
+type globalAutoAnalyzeProcessList struct {
+	processes map[uint64]struct{}
```
hawkingrei, line 61: We use global thread-safe map to record the running analyze task.

hawkingrei:
@winoros PTAL

qw4990 (approved):

pkg/statistics/handle/util/auto_analyze_proc_id_generator.go
AilinKid: why not use a `sync.map`

hawkingrei: The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex. 

so two common use cases are not the same as us.

Issue #55063, opened 2024-07-30
Tracking Issue For Maintaining Priority Queue In Memory


Rustin170506:
## Enhancement

Currently, we use a priority queue at TiDB to automatically update statistics. However, we currently maintain this queue by rebuilding it every time. This causes the following problems:
1. When there are many tables, scanning all tables' schema information can have a big impact on the meta information cache. For example, it will load all the tables at once, which is a very slow process and affects other modules' use of the schema cache.
2. Because we scan it too frequently, we'll spend a lot of CPU on building this queue, since we'll be scanning all the stats cache and the schema cache. See more: https://github.com/pingcap/tidb/issues/49972
To solve the above issues, we can try to keep the priority queue in memory instead of rebuilding it every time.

PR #55064, opened 2024-07-30
statistics: add a thread-safe heap

hi-rustin:

Issue Number: ref https://github.com/pingcap/tidb/issues/55063

Problem Summary:

This PR added a thread-safe heap for the priority queue.
Because it doesn't involve any business logic to the priority queue, it is a pure data structure, so I think we can add it in advance.

Maybe you will ask why we don't use Fibonacci heap here. The constants for Fibonacci heaps are actually quite large, and their complexity is amortized. Therefore, it is very likely that their performance will be similar to that of regular binary heaps in this scenario.

The original implementation:
https://github.com/kubernetes/client-go/blob/master/tools/cache/heap.go

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/handle/autoanalyze/internal/heap/heap.go
+++ b/pkg/statistics/handle/autoanalyze/internal/heap/heap.go
@@ -1,0 +1,306 @@
+// Copyright 2017 The Kubernetes Authors.
+// Copyright 2024 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


ngaut:
The Golang standard library has builtin heap implementation. https://pkg.go.dev/container/heap

pkg/statistics/handle/autoanalyze/internal/heap/heap.go
```diff
@@ -0,0 +1,306 @@
+// Copyright 2017 The Kubernetes Authors.
+// Copyright 2024 PingCAP, Inc.
```
qw4990, line 2: Better to add some comments to explain why we copy this implementation instead of using other third-party libs.

qw4990 (approved):

AilinKid (approved):

PR #55088, opened 2024-07-31
infoschema: rewrite SchemaSimpleTableInfos for infoschema v2 to avoid network

tiancaiamao:

Issue Number: ref #50959 

Problem Summary:

HTTP API for schema table is a burden for infoschema v2 when there are a lot of tables.
In the past, we have change SchemaTable to SimpleSchemaTableInfos, i.e. from full `model.TableInfo` to `model.TableNameInfo` to reduce the cost.

SimpleSchemaTableInfos fetch the table name and id using meta API, fetch data from tikv.
As we store all the name->id mapping now, we can avoid the network API.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/infoschema/infoschema_v2.go
```diff
@@ -795,33 +795,34 @@ func (is *infoschemaV2) SchemaSimpleTableInfos(ctx context.Context, schema model
 		return nil, nil // something wrong?
 	}
 
-retry:
-	dbInfo, ok := is.SchemaByName(schema)
-	if !ok {
+	// Ascend is much more difficult than Descend.
```
lance6716, line 798: what do you mean by `difficult`, difficult to implement?

lance6716 (approved):

pkg/infoschema/infoschema_v2.go
tiancaiamao: Handle it in Ascend order is more difficult than handle it in Descend order.

The data organize like this:

```
| t1                      | t2     | t3    |
| v5  v6  v7  v8 v10  v12 | v7 v13 | v2 v5 |
```

If we want infoschma snapshot at v10, the result should be `[t1.v10     t2.v7]`

In Descend order, the total test condition could be:

```
if is.infoSchema.schemaMetaVersion >= item.schemaVersion {
    if curr == nil || curr.tableName != tableItems[i].tableName {
          // add curr to the result
    }
}
```

But in Ascend order, you may have a try to see how much more difficult

 :-)

wjhuang2016 (approved):

PR #55101, opened 2024-07-31
infoschema,executor: change TableByID to skip refill infoschema v2 cache

tiancaiamao:

Issue Number: ref #50959 

Problem Summary:

During "select * from INFORMATION_SCHEMA.TABLES where table_schema = 'test'", the schema cache hit ratio drop to nearly 0

image

It's cause by that the result need to show the next autoid info, so it calls TableByName.
And since this is a list API, the infoschema v2 cache is affect by the cache refill.

In this commit I change it to use TableByID, and change TableByID to skip refill infoschema v2 cache.
The observation is that: when we call TableByID, we must know the table ID ... how do we get the ID? something like TableByName may have been called, so we do not need to update cache again in TableByID because TableByName already do it.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

Verify the scanerio and check cache hit ratio.

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


lance6716 (commented):
Should we let caller decide `noRefill` cache? I think there are some cases that full tables load is trigger but we still need to skip refill, like keyvis queries all tables.

tiancaiamao:
> Should we let caller decide `noRefill` cache? I think there are some cases that full tables load is trigger but we still need to skip refill, like keyvis queries all tables.

Change interface method involves much more code changes.
We need to evoluate it carefully before doing any actual actions.
Let's keep thing simple first.   @lance6716

lance6716 (approved):

GMHDBJD (approved):
LGTM

PR #55133, opened 2024-08-01
context, hack: add more comments/tests to warn the developers to update `Detach` and `IntoStatic`.

### LINKED ISSUE #55130: make sure developers will update`IntoStatic` and `Detach` after adding new fields in the context
## Enhancement

Currently, we have `IntoStatic` on `SessionExprContext` and `SessionEvalContext`, and the `Detach` method on many other smaller contexts. These methods rely on the fields of themselves. If some developers add new fields in the future, they'll need to consider whether they need to copy the new field.

As I've tried, it's hard to mock these structs, and build a good test for deep copy, because actually most of these fields don't need to be deep copied. It would be easier for the developers to make their own decision, rather than assert the deep copy. Another reason is that the session is too hard to mock :facepalm: .

I think maybe it'd be better to add a test and checksum the type of these context, to make sure the developers have read these comments :thinking: .

YangKeao:

Issue Number: close #55130

Problem Summary:

As described in the issue, we need to make sure the developers know that the `Detach` or `IntoStatic` should be changed if the interface or struct structure changed. 

1. Implement a function `hack.CalcTypeChecksum` to calculate the checksum of a type to fail a test when the type changed.
2. Add a test for many context types.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```

### DIFF
--- a/pkg/util/hack/type_checksum.go
+++ b/pkg/util/hack/type_checksum.go
@@ -1,0 +1,96 @@
+// Copyright 2024 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hack
+
+import (
+	"encoding/binary"
+	"hash"
+	"hash/crc64"
+	"reflect"
+)
+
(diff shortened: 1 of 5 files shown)


pkg/util/hack/type_checksum.go
```diff
@@ -0,0 +1,96 @@
...
+	"encoding/binary"
+	"hash"
+	"hash/crc64"
+	"reflect"
+)
+
+// CalcTypeChecksum return a value which can represent the change of the type T
+// It's used in the test to make sure the developer can read some comments or
+// make some related changes when the type T is changed.
+func CalcTypeChecksum[T any]() uint64 {
```
YangKeao, line 27: It seems that the result of this function depends on the go version :thinking:. I don't know why, every steps look consistent. Let's hold this PR for a while.

YangKeao:
We have a better one now https://github.com/pingcap/tidb/pull/55311

Close this.

Issue #54434, opened 2024-07-04
Enhance runaway query management features


JmPotato:
This is the tracking issue for improvements to runway management.

- [x] https://github.com/pingcap/tidb/issues/51325
- [x] https://github.com/pingcap/tidb/issues/54413
- [x] https://github.com/pingcap/tidb/issues/54417
- [x] https://github.com/pingcap/tidb/issues/54453
- [x] Provide additional rules for identifying runaway queries, such as based on `PROCESSED_KEYS` and `REQUEST_UNIT` produced.
    - [x] https://github.com/pingcap/kvproto/pull/1261
    - [x]   https://github.com/pingcap/tidb/pull/55147
- [x] Support `SWITCH_GROUP` action to individually switch the specified resource group for a runaway query.
    - [x] https://github.com/pingcap/tidb/pull/56145
    - [x] https://github.com/pingcap/tidb/pull/54804

PR #55147, opened 2024-08-02
runaway: Add processed_keys and request_ru to check

HuSharp:

Issue Number: ref #54434, merged after https://github.com/pingcap/kvproto/pull/1261#pullrequestreview-2274801129

Problem Summary:

supports PROCESSED_KEYS and REQUEST_UNIT to check runaway

#### Explain 

Can check doc https://github.com/pingcap/docs/pull/18858/files

Processed_keys: The number of keys that Coprocessor has processed. Compared with total_keys, processed_keys does not include the old versions of MVCC. A great difference between processed_keys and total_keys indicates that many old versions exist.
> can check in [`Process_keys`  of slow query log](https://docs.pingcap.com/tidb/stable/identify-slow-queries#fields-description)

REQUEST_UNIT: The number of request units produced by the query. It's the total number of Request_unit_read and Request_unit_write.

> can check in [`REQUEST_UNIT` of statement fields desciption](https://docs.pingcap.com/tidb/stable/statement-summary-tables#fields-description)

#### Display

```bash
mysql> ALTER RESOURCE GROUP default QUERY_LIMIT=(EXEC_ELAPSED='2s', PROCESSED_KEYS=100, ACTION=KILL,  WATCH=EXACT DURATION='10m');

mysql> SELECT * FROM information_schema.resource_groups;
+---------+------------+----------+-----------+--------------------------------------------------------------------------------+------------+
| NAME    | RU_PER_SEC | PRIORITY | BURSTABLE | QUERY_LIMIT                                                                    | BACKGROUND |
+---------+------------+----------+-----------+--------------------------------------------------------------------------------+------------+
| default | UNLIMITED  | MEDIUM   | YES       | EXEC_ELAPSED='2s', PROCESSED_KEYS=100, ACTION=KILL, WATCH=EXACT DURATION='10m0s' | NULL       |
+---------+------------+----------+-----------+--------------------------------------------------------------------------------+------------+
1 row in set (0.00 sec)
```

can check in log 
```bash
[ERROR] [domain.go:1908] ["LoadSysVarCacheLoop failed"] [error="[executor:8253]Query execution was interrupted, identified as runaway query"]
```

check query result
```bash
mysql> select * from mysql.tidb_runaway_watch limit 3 \G;
                 id: 2
resource_group_name: default
         start_time: 2024-09-06 08:54:01.886568
           end_time: 2024-09-06 09:04:01.886568
              watch: 1
         watch_text: SELECT variable_name, variable_value FROM mysql.global_variables
             source: 127.0.0.1:4000
             action: 3
  switch_group_name:
               rule: ProcessedKeys = 666(10)

mysql> SELECT * FROM mysql.tidb_RUNAWAY_queries limit 10 \G;
resource_group_name: default
               time: 2024-09-06 17:17:42
         match_type: watch
             action: kill
       original_sql: SELECT variable_name, variable_value FROM mysql.global_variables
        plan_digest: e5003a13344e26b749329d8e96cae94e844a11bb7e027bd3ea5c2a39f7c07ee7
        tidb_server: 127.0.0.1:4000
               rule: ProcessedKeys = 666(10)
```

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Documentation

- [x] Contains experimental features

### Release note

```release-note
runaway: Add processed_keys and request_ru to check
```


HuSharp:
@JmPotato @nolouch @glorv PTAL, thx!

pkg/parser/parser.y
```diff
@@ -1837,10 +1839,22 @@ DirectResourceGroupRunawayOption:
...
+		}
+	}
+|	"PROCESSED_KEYS" EqOpt intLit
+	{
+		$$ = &ast.ResourceGroupRunawayOption{
+			Tp:         model.RunawayRule,
+			RuleOption: &ast.ResourceGroupRunawayRuleOption{Tp: ast.RunawayRuleProcessedKeys, ProcessedKeys: $3.(int64)},
+		}
+	}
+|	"REQUEST_UNIT" EqOpt intLit
```
nolouch, line 1853: maybe we need should unified to "RU". cc @songrijie

pkg/ddl/resource_group.go
```diff
@@ -221,6 +221,13 @@ func SetDirectResourceGroupSettings(groupInfo *model.ResourceGroupInfo, opt *ast
 		if len(opt.RunawayOptionList) == 0 {
 			resourceGroupSettings.Runaway = nil
 		}
+		if resourceGroupSettings.Runaway == nil {
+			resourceGroupSettings.Runaway = &model.ResourceGroupRunawaySettings{}
+		} else { // reset the runaway rule settings
+			resourceGroupSettings.Runaway.ExecElapsedTimeMs = 0
+			resourceGroupSettings.Runaway.ProcessedKeys = 0
+			resourceGroupSettings.Runaway.RequestUnit = 0
+		}
```
glorv, line 230:
```suggestion
		resourceGroupSettings.Runaway = &model.ResourceGroupRunawaySettings{}
```
Based on the current logic, runaway settings show be set as a whole. Are there any fields need to be inherited?

pkg/ddl/resource_group.go
```diff
@@ -258,18 +265,23 @@ func SetDirectResourceGroupRUSecondOption(resourceGroupSettings *model.ResourceG
...
+			if err != nil {
+				return err
+			}
+			settings.ExecElapsedTimeMs = uint64(dur.Milliseconds())
+		}
+		if opt.RuleOption.ProcessedKeys != 0 {
+			settings.ProcessedKeys = opt.RuleOption.ProcessedKeys
+		}
+		if opt.RuleOption.RequestUnit != 0 {
+			settings.RequestUnit = opt.RuleOption.RequestUnit
```
glorv, line 283:
```suggestion
		settings.ProcessedKeys = opt.RuleOption.ProcessedKeys
		settings.RequestUnit = opt.RuleOption.RequestUnit
```
Seems no need to check the if condition

pkg/store/copr/coprocessor.go
```diff
@@ -1825,18 +1832,32 @@ func (worker *copIteratorWorker) collectCopRuntimeStats(copStats *CopRuntimeStat
 	}
 	copStats.ScanDetail = sd
 	copStats.TimeDetail = td
+
+	if worker.req.RunawayChecker != nil {
+		var ruDetail *util.RUDetails
+		if ruDetailRaw := bo.GetCtx().Value(util.RUDetailsCtxKey); ruDetailRaw != nil {
+			ruDetail = ruDetailRaw.(*util.RUDetails)
+		}
+		if err := worker.req.RunawayChecker.CheckThresholds(ruDetail, sd.ProcessedKeys, nil); err != nil {
```
glorv, line 1841: `sd.ProcessedKeys` is the processed keys of the current copr request, but I suppose we should compare the total processed_keys of all copr tasks.

pkg/ddl/resource_group.go
HuSharp: Since ProcessedKeys/RequestUnit/ExecElapsedTimeMs belongs to `RunawayOptionList` and `RunawayOptionList ` is a for loop, each Option has only one setting. Hence, we need to check here to avoid the last setting being useless.

glorv: Ok. Then I suggest change `ResourceGroupRunawayRuleOption` to something like follows:
```
type ResourceGroupRunawayRuleOption {
    OpType RuleOptionType
    Value uint64
}
```

HuSharp: How about just modifying `if` to `switch type` here?

glorv: Another thought, as we only support add/update `RunawayRule` as a whole, why not use a unique `ResourceGroupRunawaySettings` in parser instead of the option list, I wonder what the benefit that this list bring us

HuSharp: - `ExecElapsedTimeMs`, `ProcessedKeys`, `RequestUnit`, `ACTION` and `WATCH` belong `ResourceGroupRunawayOptionList`
- `RU_PER_SEC`, `PRIORITY`, ... , `QUERY_LIMIT` belongs the same level which using `ResourceGroupOptionList`
- `ResourceGroupRunawayOptionList` belongs `ResourceGroupRunawayOptionList`

With all splitted as `,` characters, it is difficult for the parser to determine the next character whether `ResourceGroupOptionList` or `ResourceGroupRunawayOptionList`.

HuSharp: I think this is to support multiple out-of-order options?

pkg/parser/parser.y
HuSharp: After discussing with PM, we will use RU

pkg/store/copr/coprocessor.go
```diff
@@ -102,6 +102,7 @@ func (c *CopClient) Send(ctx context.Context, req *kv.Request, variables any, op
 		return errRes
 	}
 	ctx = context.WithValue(ctx, tikv.RPCCancellerCtxKey{}, it.rpcCancel)
+	ctx = context.WithValue(ctx, util.RUDetailsCtxKey, util.NewRUDetails())
```
nolouch, line 105: why new `RUDetailsCtxKey` here, will it collect the stat with old path?, what is different with `clientutil.RUDetailsCtxKey`.

HuSharp: > will it collect the stat with old path

In some places the incoming ctx does not have ru(Analyze), in others it does(executeKillStmt). I added the judgment

> no different with `different with clientutil.RUDetailsCtxKey.`

just the go import name is different

nolouch (approved):
lgtm

HuSharp:
@glorv @JmPotato PTAL, thx~

pkg/parser/go.mod
```diff
@@ -20,7 +20,6 @@ require (
 )
 
 require (
-	github.com/benbjohnson/clock v1.3.5 // indirect
```
JmPotato, line 23: Why do we have this change?

pkg/ddl/resource_group.go
```diff
@@ -222,6 +222,8 @@ func SetDirectResourceGroupSettings(groupInfo *model.ResourceGroupInfo, opt *ast
 	case ast.ResourceGroupRunaway:
 		if len(opt.RunawayOptionList) == 0 {
 			resourceGroupSettings.Runaway = nil
+		} else {
+			resourceGroupSettings.Runaway = &model.ResourceGroupRunawaySettings{}
```
JmPotato, line 226: This part changed the behavior, for the old code, if `resourceGroupSettings.Runaway` is not nil, it won't be overwritten in `SetDirectResourceGroupRunawayOption`, but now it will be reset to an empty new one if `len(opt.RunawayOptionList) > 0`. Will this be a problem?

pkg/store/copr/coprocessor.go
```diff
@@ -915,6 +918,10 @@ func (sender *copIteratorTaskSender) run(connID uint64) {
 	if sender.respChan != nil {
 		close(sender.respChan)
 	}
+	if checker != nil {
```
JmPotato, line 921: No need to check since `ResetTotalProcessedKeys` will do it also.

glorv (approved):

pkg/ddl/resource_group.go
HuSharp: Nice catch.

pkg/parser/go.mod
HuSharp: I just used gomod :(

pkg/store/copr/coprocessor.go
HuSharp: struct's checker field is of type `RunawayChecker`, which is an interface, but it hasn't been initialized with a instance. When calling `checker.ResetTotalProcessedKeys(),` the value of `checker` is nil, leading to a runtime error: invalid memory address or nil pointer dereference.

You can check this [playground](https://go.dev/play/p/MF9M51vdLbo)

Instead of the following approach, maybe we can judge in `ResetTotalProcessedKeys`?

```bash
if checker != nil {
    if rcChecker, ok := checker.(*Checker); ok && rcChecker != nil {
        rcChecker.ResetTotalProcessedKeys()
	}
```

tangenta (approved):

yudongusa (approved):
LGTM

Issue #54397, opened 2024-07-03
Remove `GetSessionVars` in `table.MutateContext`


lcwangchao:
## Enhancement

`GetSessionVars` returns `variable.SessionVars` which exposes too many fields and it's hard to build correctly. It's better to remove it and exposes some fields we really needed outside to `table.MutateContext`.

- #54394
- #54433
- #54521
- #54673
- #54715
- #54767
- #54789
- #54900
- #55125
- #55192
- #55194
- #55246

PR #55192, opened 2024-08-05
table: Add option `SkipWriteUntouchedIndices` for `Table.UpdateRecord`

lcwangchao:

Issue Number: ref #54397

`table.SkipWriteUntouchedIndices` is introduced as a new option for `Table.UpdateRecord`. It is used to decouple the code of `UpdateRecord` and the executor logic. It makes the code more clear and easier to read.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/table/tables/tables.go
```diff
@@ -532,7 +532,7 @@ func (t *TableCommon) updateRecord(sctx table.MutateContext, h kv.Handle, oldDat
 		}
 	}
 	// rebuild index
-	err = t.rebuildIndices(sctx, txn, h, touched, oldData, newData, opt.GetCreateIdxOpt())
+	err = t.rebuildIndices(sctx, txn, h, touched, oldData, newData, opt.SkipWriteUntouchedIndices, opt.GetCreateIdxOpt())
```
crazycs520, line 535: How about put `SkipWriteUntouchedIndices` into `CreateIdxOpt`?

lcwangchao: I think It's not a good way. Because the coder can then call `Index.Create` like this:

```
index.Create(..., table.SkipWriteUntouchedIndices)
```

It is meaningless because `SkipWriteUntouchedIndices` only works for `UpdateRecord`

lcwangchao: If you think `rebuildIndices` has too many arguments, it maybe better to change `rebuildIndices` to `rebuildIndicesForUpdate` and pass `opt *UpdateRecordOpt` as an argument:

```go
func (t *TableCommon) rebuildIndicesForUpdate(ctx table.MutateContext, ..., opt *table.UpdateRecordOpt) error {
	...
	createIdxOpt := opt.GetCreateIdxOpt()
	for _, idx := range t.Indices() {
		...
		if untouched && opt.SkipWriteUntouchedIndices {
			continue
		}
		newVs, err := idx.FetchValues(newData, nil)
		if err != nil {
			return err
		}
		if err := t.buildIndexForRow(ctx, h, newVs, newData, asIndex(idx), txn, untouched, createIdxOpt); err != nil {
			return err
		}
	}
	return nil
}
```

crazycs520 (approved):

tangenta (approved):

PR #54333, opened 2024-07-01
infoschema: speed up reader for some tables

ywqzzy:
### What problem does this PR solve?

Issue Number: ref #50305 

Problem Summary:

### What changed and how does it work?
Push down some filters for：
1. infoschema.TableStatistics
2. infoschema.TablePartitions
3. infoschema.TableSequences
4. infoschema.TableCheckConstraints
5. infoschema.TableTiDBCheckConstraints
6. infoschema.TableTiDBIndexUsage
7. infoschema.TableTiDBIndexes
8. infoschema.TableViews
9. infoschema.TableConstraints
10. infoschema.TableSchemata

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #55162, opened 2024-08-02
executor: filter on `table_schema` for `INFORMATION_SCHEMA.KEY_COLUMN_USAGE` table

YangKeao:
### What problem does this PR solve?

Issue Number: close #55156

Problem Summary:

Previously, the condition `table_schema = ...` has no effect on table `INFORMATION_SCHEMA.KEY_COLUMN_USAGE`. This PR fixed it.

### What changed and how does it work?

Also filter on `table_schema` condition.

Maybe I can add more automatically generated tests for `InfoSchemaTablesExtractor` :thinking: later. We should verify that all these extracted columns work on all the information schema tables.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #55236, opened 2024-08-06
executor: fix the issue that some extracted conditions are not used in information_schema

### LINKED ISSUE #55235: `constraint_name` condition is ignored in `KEY_COLUMN_USAGE`
This issue is similar to https://github.com/pingcap/tidb/issues/55156, and is introduced in https://github.com/pingcap/tidb/pull/54333.

I'd like to introduce a better way to test these functions :thinking: . It's really easy to make mistakes by not using one of the extracted conditions.

YangKeao:

Issue Number: close #55235

Problem Summary:

Some conditions are just not used in the retriever of information schema tables. This PR adds a test for them, and fixes these issues. These issues are introduced in https://github.com/pingcap/tidb/pull/54333.

No matter whether an extracted condition is used in the real implementation, the parent executor will not use it to run a `SELECTION` again, and that's why these issues appear. It's hard to ensure all the conditions are used correctly, so I added a test to automatically test these lines.

1. Add a test to automatically generate tables and selecting rows in the information schema to see whether the condition behaves well.
2. Fix the case when the condition has no effect.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
Fix the issue that some SELECT conditions have no effect for some tables in the information_schema.
```

### DIFF
--- a/pkg/executor/infoschema_reader.go
+++ b/pkg/executor/infoschema_reader.go
@@ -1752,7 +1752,7 @@ func (e *memtableRetriever) setDataFromKeyColumnUsage(ctx context.Context, sctx 
 			if checker != nil && !checker.RequestVerification(sctx.GetSessionVars().ActiveRoles, schema.L, table.Name.L, "", mysql.AllPrivMask) {
 				continue
 			}
-			rs := keyColumnUsageInTable(schema, table)
+			rs := keyColumnUsageInTable(schema, table, extractor)
 			rows = append(rows, rs...)
 		}
 	}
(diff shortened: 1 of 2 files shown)


YangKeao:
ref https://github.com/pingcap/tidb/pull/55162.

Similar issue I fixed before.

pkg/planner/core/memtable_infoschema_extractor.go
```diff
@@ -59,8 +59,7 @@ var extractableColumns = map[string][]string{
 	// See infoschema.statisticsCols for full columns.
 	// Used by InfoSchemaStatisticsExtractor and setDataForStatistics.
 	infoschema.TableStatistics: {
-		_tableSchema, _tableName,
```
YangKeao, line 62: `infoschema.TableStatistics` doesn't have `_tableSchema` column.

YangKeao: I'm wrong.

tangenta (approved):

tiancaiamao (approved):

PR #55259, opened 2024-08-07
parser: support GLOBAL IndexOption

mjonss:

Issue Number: ref #52994

Problem Summary:

This also makes it easier to allow non-unique indexes as well as unique indexes already containing all partitioning columns to be GLOBAL.

Currently ALTER TABLE t PARTITION BY ... would implicitly create a Global Index of UNIQUE Indexes if they don't include all partitioning columns.
With this PR we are adding an optional UPDATE INDEXES clause after the partitioning definitions:
UPDATE INDEXES ( index_name {GLOBAL|LOCAL} [, index_name {GLOBAL|LOCAL}[,...]] ) to make every index change explicit.

Parser:
- Added "GLOBAL" as IndexOption
- Added "GLOBAL" as option for ColumnOption in combination with {PRIMARY|UNIQUE} KEY, where it will use the StrValue set to "Global" instead of adding yet another struct member.
- Added "UPDATE INDEXES (...)" as option for ALTER TABLE t PARTITION BY ... [UPDATE INDEXES (...)] to allow [re]partitioning a table and convert unique indexes that does not include all partitioning columns into global indexes.

Executor:

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [x] Contains syntax changes

### Release note

```release-note
None
```


tangenta (approved):

tangenta:
Is there a convenient way to convert between local and global index like

```sql
ALTER TABLE t ALTER INDEX c1 [GLOBAL | LOCAL];
```
?

mjonss:
> Is there a convenient way to convert between local and global index like
> 
> ```sql
> ALTER TABLE t ALTER INDEX c1 [GLOBAL | LOCAL];
> ```
> 
> ?

@tangenta Good question, I have not thought of it, I would consider the indexes to be different, so you should probably create a new index and drop the old one instead, which would also have less impact of DDL handling, since it would need to recreate the full index and it is not a simple metadata only change like VISIBLE/INVISIBLE IndexOption.

mjonss:
After some consideration and discussion with the PM, I changed from:
ALTER TABLE t PARTITION BY ... [CONVERT TO GLOBAL INDEX]
to
ALTER TABLE t PARTITION BY ... [UPDATE INDEXES (<list of index changes>)]
where list of index changes is a comma separated list of:
IndexName {GLOBAL|LOCAL}

to be able to more explicitly change which indexes to change to GLOBAL (and to LOCAL if you repartition a table).
This will also allow to make non-unique indexes and unique indexes that includes all partitioning columns to be GLOBAL in the future.

pkg/parser/ast/ddl.go
```diff
@@ -921,7 +919,6 @@ func (n *Constraint) Restore(ctx *format.RestoreCtx) error {
 	}
 
 	if n.Option != nil {
-		ctx.WritePlain(" ")
 		if err := n.Option.Restore(ctx); err != nil {
```
tangenta, line 922: The convention is that "Restore()" always generates a string with whitespace trimming :)

pkg/parser/parser_test.go
```diff
@@ -2811,7 +2811,7 @@ func TestDDL(t *testing.T) {
...
+		) ENGINE=InnoDB AUTO_INCREMENT=31 DEFAULT CHARACTER SET UTF8 COLLATE UTF8_GENERAL_CI ROW_FORMAT=COMPACT COMMENT='' CHECKSUM=0 DELAY_KEY_WRITE=0;`, true, "CREATE TABLE `address` (`id` BIGINT(20) NOT NULL AUTO_INCREMENT,`create_at` DATETIME NOT NULL,`deleted` TINYINT(1) NOT NULL,`update_at` DATETIME NOT NULL,`version` BIGINT(20) DEFAULT NULL,`address` VARCHAR(128) NOT NULL,`address_detail` VARCHAR(128) NOT NULL,`cellphone` VARCHAR(16) NOT NULL,`latitude` DOUBLE NOT NULL,`longitude` DOUBLE NOT NULL,`name` VARCHAR(16) NOT NULL,`sex` TINYINT(1) NOT NULL,`user_id` BIGINT(20) NOT NULL,PRIMARY KEY(`id`),CONSTRAINT `FK_7rod8a71yep5vxasb0ms3osbg` FOREIGN KEY (`user_id`) REFERENCES `waimaiqa`.`user`(`id`),INDEX `FK_7rod8a71yep5vxasb0ms3osbg`(`user_id`)) ENGINE = InnoDB AUTO_INCREMENT = 31 DEFAULT CHARACTER SET = UTF8 DEFAULT COLLATE = UTF8_GENERAL_CI ROW_FORMAT = COMPACT COMMENT = '' CHECKSUM = 0 DELAY_KEY_WRITE = 0"},
```
Defined2014, line 2814: Why the value of auto_increment changed?

pkg/parser/parser.y
```diff
@@ -3567,27 +3585,31 @@ ColumnOption:
...
-|	PrimaryOpt "KEY" WithClustered
+|	PrimaryOpt "KEY" WithClustered GlobalOrLocalOpt
 	{
 		// KEY is normally a synonym for INDEX. The key attribute PRIMARY KEY
 		// can also be specified as just KEY when given in a column definition.
 		// See http://dev.mysql.com/doc/refman/5.7/en/create-table.html
-		$$ = &ast.ColumnOption{Tp: ast.ColumnOptionPrimaryKey, PrimaryKeyTp: $3.(model.PrimaryKeyType)}
+		$$ = &ast.ColumnOption{Tp: ast.ColumnOptionPrimaryKey, PrimaryKeyTp: $3.(model.PrimaryKeyType), StrValue: $4}
+	}
+|	"UNIQUE" "GLOBAL"
```
Defined2014, line 3602: Not tested in UT and should this be `GlobalOrLocalOpt` instead of `GLOBAL`?

mjonss: Right, I will update it or add the possibility of LOCAL as well in case of conflicts.

pkg/parser/parser_test.go
mjonss: Only to make it easier to find the failing tests, since there were multiple similar tests.

pkg/parser/ast/ddl.go
mjonss: I can revert this change if you like, I did it so that the full restore would not add additional spaces in case there were a default IndexOption, like LOCAL.

tangenta: I think we can add `IndexOption.isEmpty()` to help determine whether it is a default value:

```
if n.Option != nil && !n.Option.isEmpty() {
		ctx.WritePlain(" ")
		if err := n.Option.Restore(ctx); err != nil {
```

Then it is consist with other AST nodes, and the change in `runNodeRestoreTestWithFlags` is not necessary.

pkg/parser/parser.y
mjonss: I explicitly added `"GLOBAL"` and `"LOCAL"`, as well as kept the `"UNIQUE" %prec lowerThanKey` to avoid conflicts.

pkg/parser/ast/ddl.go
mjonss: Thanks for the suggestion, done :)

Defined2014 (approved):

tangenta (approved):

bb7133 (approved):
LGTM

Defined2014:
/hold
Should we merge it now or after branch v8.3 created?

PR #55264, opened 2024-08-07
*: Require GLOBAL IndexOption for creating Global Index

### LINKED ISSUE #52994: Add GLOBAL as IndexOption and require it explicitly for GLOBAL INDEX

mjonss:

Issue Number: close #52994

Building upon  #55259, which is the Parser change needed for this PR.

Problem Summary:

This also makes it easier to allow non-unique indexes as well as unique indexes already containing all partitioning columns to be GLOBAL.

Currently ALTER TABLE t PARTITION BY ... would implicitly create a Global Index of UNIQUE Indexes if they don't include all partitioning columns.
With this PR we are adding an optional UPDATE INDEXES clause after the partitioning definitions:
UPDATE INDEXES ( index_name {GLOBAL|LOCAL} [, index_name {GLOBAL|LOCAL}[,...]] ) to make every index change explicit.

DDL:
- Checking and propagating the options from the parser, so that we require GLOBAL to be explicitly mentioned when creating a global index.

Executor:

Table:
- Just updated test to use GLOBAL IndexOption.

Added a new error for needing Global Index, but is not explicitly given.
Most changes are just updated tests and results, to comply with the new requirement of explicitly mention GLOBAL.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [x] Contains syntax changes

### Release note

```release-note
Require GLOBAL IndexOption for creating a Global Index (experimental).
```


Defined2014:
@mjonss Please rebase master branch and resolve confilcts.

tests/integrationtest/t/ddl/db_partition.test
```diff
@@ -2270,11 +2271,12 @@ create temporary table nt34 (id int);
 -- error 1733
 alter table pt34 exchange partition p0 with table nt34;
 
---echo ## global index
-create table pt35 (a int, b int, unique index(b)) partition by hash(a) partitions 1;
-create table nt35 (a int, b int, unique index(b));
--- error 1731
-alter table pt35 exchange partition p0 with table nt35;
+# TODO: support EXCHANGE PARTITION with GLOBAL INDEX
```
Defined2014, line 2274: Why comment it? What's the exchange partition behaviour after this PR?

pkg/ddl/tests/partition/db_partition_test.go
```diff
@@ -3673,7 +3676,7 @@ func TestPrimaryNoGlobalIndex(t *testing.T) {
 	// get the full row with a single primary key lookup.
 	// Since a global index would first need to be checked to get the partition to
 	// do a second index lookup for getting the full row.
-	tk.MustContainErrMsg(`create table t (a int primary key clustered, b varchar(255)) partition by key(b) partitions 3`, `A CLUSTERED INDEX must include all columns in the table's partitioning function`)
+	tk.MustContainErrMsg(`create table t (a int primary key clustered, b varchar(255)) partition by key(b) partitions 3`, `A PRIMARY KEY must include all columns in the table's partitioning function`)
```
Defined2014, line 3679: Why this changed? Use CLUSTERED INDEX is more precise.

pkg/ddl/executor.go
```diff
@@ -4554,6 +4512,9 @@ func (e *executor) CreatePrimaryKey(ctx sessionctx.Context, ti ast.Ident, indexN
 				return dbterror.ErrUniqueKeyNeedAllFieldsInPf.GenWithStackByArgs("PRIMARY")
 			}
 			// index columns does not contain all partition columns, must set global
+			if indexOption == nil || !indexOption.Global {
+				return dbterror.ErrGlobalIndexNotExplicitlySet.GenWithStackByArgs("PRIMARY")
+			}
 			global = true
```
Defined2014, line 4518: Param global can be removed, just like createIndex

tests/integrationtest/t/ddl/db_partition.test
mjonss: Thanks for the suggestion, I reverted the change, added GLOBAL ColumnOption and corrected the error code.

pkg/ddl/tests/partition/db_partition_test.go
mjonss: Fixed it :) Thanks for noticing!

pkg/ddl/executor.go
mjonss: Nice find, I removed it and also removed the handling of the 'global' arg in the `job.Args`/`decodeAddIndexArgs`

pkg/ddl/partition.go
```diff
@@ -623,16 +623,83 @@ func buildTablePartitionInfo(ctx sessionctx.Context, s *ast.PartitionOptions, tb
...
-	if err != nil {
-		return errors.Trace(err)
+	if len(s.UpdateIndexes) > 0 {
+		updateIndexes := make([]model.UpdateIndexInfo, 0, len(s.UpdateIndexes))
+		dupCheck := make(map[string]struct{})
+		for _, idxUpdate := range s.UpdateIndexes {
+			idxOffset := -1
+			for i := range tbInfo.Indices {
+				if strings.EqualFold(tbInfo.Indices[i].Name.L, idxUpdate.Name) {
+					idxOffset = i
```
Defined2014, line 633: Should we check is commonHandle here?

pkg/ddl/executor.go
```diff
@@ -4576,7 +4535,7 @@ func (e *executor) CreatePrimaryKey(ctx sessionctx.Context, ti ast.Ident, indexN
 		Type:           model.ActionAddPrimaryKey,
 		BinlogInfo:     &model.HistoryInfo{},
 		ReorgMeta:      nil,
-		Args:           []any{unique, indexName, indexPartSpecifications, indexOption, sqlMode, nil, global},
+		Args:           []any{unique, indexName, indexPartSpecifications, indexOption, sqlMode, nil},
```
Defined2014, line 4538: Will deleting this affect the upgrade? /cc @tangenta

mjonss: No, removing arguments from job.Args is handled gracefully (simply ignored and DecodeArgs will regenerate RawArgs IIRC. So the result should be that old queued DDLs should fail due to lack of explicitly set GLOBAL (since it did not have IndexOption set).

pkg/ddl/partition.go
mjonss: Not sure, it is checked in the next loop over tbInfo.Indices, and will fail if CommonIndex and global index is set for the same index.

Defined2014 (approved):

pkg/ddl/executor.go
tangenta: I think we should leave a placeholder here to distinguish it from new argument introduced in the future. Otherwise, there may be an unmarshal error during upgrade.

pkg/ddl/index_modify_test.go
```diff
@@ -749,19 +749,19 @@ func TestAddGlobalIndex(t *testing.T) {
...
-	tk.MustExec("alter table t add unique index idx(a)")
+	tk.MustExec("alter table t add unique index idx(a) global")
 
 	// meets duplicate
 	tk.MustExec("drop table t")
 	tk.MustExec("create table t(a int, b int) partition by hash(b) partitions 64")
 	tk.MustExec("insert into t values (1, 2), (1, 3)")
 	// Duplicate
-	tk.MustExecToErr("alter table t add unique index idx(a)")
+	tk.MustExecToErr("alter table t add unique index idx(a)i global")
```
tangenta, line 759: Is this a valid syntax? or typo?

pkg/ddl/executor.go
mjonss: Ah, right if it is queued with a newer version and unqueued by an old version there might be issues, I will add a placeholder.

pkg/ddl/index_modify_test.go
mjonss: Nice catch, it is a typo and hidden since it just want to get an error... Will fix it and change from MustExecToErr to MustContainErrMsg

qw4990 (approved):
LGTM for the optimizer part

tangenta (approved):

PR #55311, opened 2024-08-08
context: add tests to avoid making mistakes by adding new fields but not modifying `IntoStatic`.

### LINKED ISSUE #55130: make sure developers will update`IntoStatic` and `Detach` after adding new fields in the context

YangKeao:

Issue Number: close #55130

Problem Summary:

It's easy to make mistake by adding new fields to the complex context but forgot to modify the `IntoStatic` and `Detach` methods.

This PR makes several changes to make it easy to test:

1. Refactor the `IntoStatic` of the `ExprContext` and `EvalContext` into functions `MakeEvalContextStatic` and `MakeExprContextStatic`, which accepts `ExprContext` and `EvalContext` (but not only `Session.*`). It's much easier to generate a non-empty `Static.*Context` than a whole session.
2. Add several tests. The tests are organized in the following pattern:
   1. Require the developer to provide a new struct, and all fields of it should be non-empty. We provide a helper function to test whether it meets the condition.
   2. Use `Make...Static` or `Detach` to try to detach it. And check the new one and the old one is deep equal.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```


pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,142 @@
...
+	for _, reg := range h.ignorePath {
+		if reg.MatchString(path) {
+			return true
+		}
+	}
+	return false
+}
+
+// AssertNotEmpty asserts that the value and its every fields are not empty.
+func (h *staticTestHelper) AssertNotEmpty(t *testing.T, valObj reflect.Value, path string) {
```
lcwangchao, line 49: How about to introduce a method `AssertEveryFieldNotEqual(t *testing.T, val1, val2 reflect.Value, path)` ? For example, `NewStaticExprContext()` does not return an object with the "go default fields", it returns "variable default fields" with some fields set to `variable.DefaultXXX`.

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,142 @@
...
+		h.AssertNotEmpty(t, valObj.Elem(), path)
+	case reflect.Func:
+		return
+	default:
+		require.Fail(t, "unsupported type", path)
+	}
+}
+
+// Equal tells whether a and b are deeply equal.
+func (h *staticTestHelper) Equal(t *testing.T, valA, valB reflect.Value, path string) {
```
lcwangchao, line 93: Could we use generic to make it easier to use? For example:

```go
func (h * staticTestHelper[T]) AssertEqual(t *testing.T, obj *T) {
	val := reflect.ValueOf(objA)
        var empty T
	valEmpty := reflect.ValueOf(&empty)
	....
}

func (h * staticTestHelper[T]) AssertEveryFieldNotEqual(t *testing.T, objA, objB *T) {
	valA := reflect.ValueOf(objA)
	valB := reflect.ValueOf(objB)
	....
}

func (h * staticTestHelper[T]) AssertNotEmpty(t *testing.T, objA, objB *T) {
	valA := reflect.ValueOf(objA)
	valB := reflect.ValueOf(objB)
	....
}
```

I think we can also remove the last argument `path` because it seems the outside code always use `$` as an input.

You can introduce a private method `equal(t *testing.T, objA, objB *T, path string)` to be used it internally.

lcwangchao: Also, we can just use a function instead of an helper object to do the assert:

```
func AssertEqual(objA, objB T, opts ...Option) {
	...
}

func WithIgnorePath(path ...string) Option {
	...
}

// We can use it like this
AssertEqual(t, ctx1, ctx2, WithIgnorePath(path1, path2))
```

pkg/expression/contextstatic/exprctx_test.go
```diff
@@ -145,16 +147,47 @@ func TestExprCtxColumnIDAllocator(t *testing.T) {
...
+	)
+
+	ignorePath := []string{
+		"^\\$\\.staticExprCtxState\\.evalCtx\\..*",
+		"^\\$\\.staticExprCtxState\\.planCacheTracker\\..*",
+		"^\\$\\.staticExprCtxState\\.columnIDAllocator\\..*",
+		"^\\$.*\\.mu",
+	}
+	h := deeptest.NewStaticTestHelper(ignorePath)
+	h.AssertNotEmpty(t, reflect.ValueOf(obj), "$")
```
lcwangchao, line 189: We should also check obj's fields is not the same value of `NewStaticExprContext()` without any options

pkg/expression/context/context.go
```diff
@@ -234,3 +243,22 @@ func AssertLocationWithSessionVars(ctxLoc *time.Location, vars *variable.Session
 		ctxLocStr, varsLocStr, stmtLocStr,
 	)
 }
+
+// StaticalizableExprContext provides more methods to implement the clone of a `ExprContext`
+type StaticalizableExprContext interface {
+	ExprContext
+
+	GetStaticalizableEvalContext() StaticalizableEvalContext
```
lcwangchao, line 251: Could we only give one method like `GetSpecailStaticalizableExprCtxOptions() ([]StaticExprCtxOption, []StaticEvalCtxOption)` to return some special fields to staticalize.

I think its can reduce same misusages like `GetPlanCacheTracker` which should only be used in `Detach`

pkg/util/deeptest/statictesthelper.go
YangKeao: Good. I've changed to `AssertEveryFieldNotEqual`, and it actually found one issue :facepalm:

pkg/util/deeptest/statictesthelper.go
YangKeao: Yes. But I still kept the helper for internal use to share the `ignorePath` recursively. The option is introduced to change the internal configuration of the helper.

pkg/expression/contextstatic/evalctx.go
```diff
@@ -203,6 +203,13 @@ func WithParamList(params *variable.PlanCacheParamList) StaticEvalCtxOption {
 	}
 }
 
+// WithEnableRedactLog sets the value of the 'tidb_redact_log' system variable.
+func WithEnableRedactLog(enableRedactLog string) StaticEvalCtxOption {
```
YangKeao, line 207: This PR actually finds one bug :beers: . The `enableRedactLog` is not copied to the detached expr context.

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,181 @@
...
+func WithIgnorePath(ignorePath []string) Option {
+	return ignorePathOption{ignorePath}
+}
+
+// AssertEveryFieldNotEqual asserts that every field of `a` is not equal to the corresponding field of `b` deeply.
+func AssertEveryFieldNotEqual[T any](t *testing.T, valA, valB T, opts ...Option) {
+	h := &staticTestHelper{}
+	for _, opt := range opts {
+		opt.apply(h)
+	}
```
lcwangchao, line 168: These codes can be wrapped to a method `staticTestHelper.applyOptions(opts ...Option)` for reuse.

lcwangchao (commented):
LGTM

lcwangchao (approved):

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,256 @@
...
+	// This function assumes that `a` and `b` are the same type
+	switch valA.Type().Kind() {
+	case reflect.Struct:
+		for i := 0; i < valA.NumField(); i++ {
+			h.assertDeepClonedEqual(t, valA.Field(i), valB.Field(i), path+"."+valA.Type().Field(i).Name)
+		}
+	case reflect.Ptr:
+		// both of them are not nil
+		require.NotEqual(t, 0, valA.Pointer(), path+" should not be nil")
+		require.NotEqual(t, 0, valB.Pointer(), path+" should not be nil")
```
lcwangchao, line 145: I think both nil values should be allowed?

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,256 @@
...
+
+		if h.shouldComparePointer(path) {
+			require.Equal(t, valA.Pointer(), valB.Pointer(), path+" should be the same")
+		} else {
+			require.NotEqual(t, valA.Pointer(), valB.Pointer(), path+" should be different")
+			h.assertDeepClonedEqual(t, valA.Elem(), valB.Elem(), path)
+		}
+	case reflect.Slice:
+		if h.shouldComparePointer(path) {
+			require.Equal(t, valA.Pointer(), valB.Pointer(), path+" should be the same")
```
lcwangchao, line 155: Do we need also to compare the lenght of these two slice?

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,256 @@
...
+			require.Equal(t, valA.Len(), valB.Len(), path+" should have the same length")
+			for _, key := range valA.MapKeys() {
+				h.assertDeepClonedEqual(t, valA.MapIndex(key), valB.MapIndex(key), path+fmt.Sprintf("[%v]", key))
+			}
+		}
+	case reflect.Interface:
+		if valA.IsNil() && valB.IsNil() {
+			return
+		}
+		h.assertDeepClonedEqual(t, valA.Elem(), valB.Elem(), path)
```
lcwangchao, line 192: should we assert valA.Elem() and valB.Elem() should have the same type?

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,256 @@
...
+	case reflect.Func:
+		if !h.shouldComparePointer(path) && !h.shouldIgnore(path) {
```
lcwangchao, line 194:
```go
if h.shouldComparePointer(path) {
    require.Equal(t, valA.Pointer(), valB.Pointer(), path+" should be different")
} else {
    require.Fail(...)
}
```

seems more easy to understand to indicate that function can only be compared with pointer

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,256 @@
...
+
+		if !h.shouldComparePointer(path) {
+			h.assertEveryFieldNotEqual(t, valA.Elem(), valB.Elem(), path)
+		}
+	case reflect.Slice:
+		require.NotEqual(t, valA.Pointer(), valB.Pointer(), path+" should not be the same")
+
+		if !h.shouldComparePointer(path) {
+			if valA.Len() == valB.Len() {
+				for i := 0; i < valA.Len(); i++ {
```
lcwangchao, line 82: If valLen()  != valB.Len() and valA[0] == valB[0] that will returns true. Do we need to check their common part?

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,256 @@
...
+					h.assertEveryFieldNotEqual(t, valA.Index(i), valB.Index(i), path+fmt.Sprintf("[%d]", i))
+				}
+			}
+		}
+	case reflect.Array:
+		if valA.Len() == valB.Len() {
```
lcwangchao, line 88: Same with slice

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,256 @@
...
+					h.assertDeepClonedEqual(t, valA.MapIndex(key), valB.MapIndex(key), path+fmt.Sprintf("[%v]", key))
+				}
+			}
+		}
+	case reflect.Interface:
+		if valA.IsNil() || valB.IsNil() {
+			require.False(t, valA.IsNil() && valB.IsNil(), path+" should not be nil at the same time")
+			return
+		}
+		h.assertEveryFieldNotEqual(t, valA.Elem(), valB.Elem(), path)
```
lcwangchao, line 118: `valA.Elem()` and `valB.Elem()` may not be a same type

pkg/util/deeptest/statictesthelper.go
YangKeao: Yes.

pkg/util/deeptest/statictesthelper.go
YangKeao: I add a `require.Equal(t, valA.Type(), valB.Type(), path+" should have the same type")` at the top of `assertDeepClonedEqual` :thinking: . I hope it work (but not sure, asserting type can be a little difficult).

YangKeao: Also, I make it pass when the types are different in `assertEveryFieldNotEqual`.

pkg/util/deeptest/statictesthelper.go
YangKeao: Done. :+1:

qw4990 (approved):

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,262 @@
...
+	if h.shouldIgnore(path) {
+		return
+	}
+
+	if !valA.IsValid() || !valB.IsValid() {
+		require.False(t, !valA.IsValid() && !valB.IsValid(), path+" should not be zero value at the same time")
+		return
+	}
+
+	if valA.Type() != valB.Type() {
```
XuHuaiyu, line 64: Why not use 	`require.Equal(t, valA.Type(), valB.Type(), path+" should have the same type")`

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,262 @@
...
+		if h.shouldComparePointer(path) {
+			require.Equal(t, valA.Pointer(), valB.Pointer(), path+" should be the same")
+			require.Equal(t, valA.Len(), valB.Len(), path+" should have the same length")
+		} else {
+			require.NotEqual(t, valA.Pointer(), valB.Pointer(), path+" should not be the same")
+			require.Equal(t, valA.Len(), valB.Len(), path+" should have the same length")
+			for i := 0; i < valA.Len(); i++ {
+				h.assertDeepClonedEqual(t, valA.Index(i), valB.Index(i), path+fmt.Sprintf("[%d]", i))
+			}
+		}
```
XuHuaiyu, line 169:
```suggestion
		require.Equal(t, valA.Len(), valB.Len(), path+" should have the same length")

		if h.shouldComparePointer(path) {
			require.Equal(t, valA.Pointer(), valB.Pointer(), path+" should be the same")
		} else {
			require.NotEqual(t, valA.Pointer(), valB.Pointer(), path+" should not be the same")
			for i := 0; i < valA.Len(); i++ {
				h.assertDeepClonedEqual(t, valA.Index(i), valB.Index(i), path+fmt.Sprintf("[%d]", i))
			}
		}
```

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,262 @@
...
+	case reflect.Float32, reflect.Float64:
+		require.Equal(t, valA.Float(), valB.Float(), path+" should be the same")
+	case reflect.String:
+		require.Equal(t, valA.String(), valB.String(), path+" should be the same")
+	case reflect.Map:
+		if h.shouldComparePointer(path) {
+			require.Equal(t, valA.Pointer(), valB.Pointer(), path+" should be the same")
+		} else {
+			require.NotEqual(t, valA.Pointer(), valB.Pointer(), path+" should not be the same")
+			require.Equal(t, valA.Len(), valB.Len(), path+" should have the same length")
```
XuHuaiyu, line 190: move this line outside of this if-else block?

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+		h.assertDeepClonedEqual(t, valA.Elem(), valB.Elem(), path)
+	case reflect.Func:
+		if h.shouldComparePointer(path) {
+			require.Equal(t, valA.Pointer(), valB.Pointer(), path+" should be different")
+		} else {
+			require.Fail(t, "a function should be compared by pointer or ignored, because there's no way to compare the content of a function", path)
+		}
+	default:
+		require.Fail(t, "unsupported type", path)
```
XuHuaiyu, line 207:
```suggestion
		require.Fail(t, "unsupported type: "+valA.Type().String(), path)
```

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,262 @@
...
+			}
+		}
+	case reflect.Interface:
+		if valA.IsNil() && valB.IsNil() {
+			return
+		}
+		h.assertDeepClonedEqual(t, valA.Elem(), valB.Elem(), path)
+	case reflect.Func:
+		if h.shouldComparePointer(path) {
+			require.Equal(t, valA.Pointer(), valB.Pointer(), path+" should be different")
```
XuHuaiyu, line 202:
```suggestion
			require.Equal(t, valA.Pointer(), valB.Pointer(), path+" should be the same")
```

pkg/util/deeptest/statictesthelper.go
```diff
@@ -0,0 +1,262 @@
...
+// AssertRecursivelyNotEqual asserts that every field of `a` is not equal to the corresponding field of `b` deeply.
+func AssertRecursivelyNotEqual[T any](t require.TestingT, valA, valB T, opts ...Option) {
+	h := &staticTestHelper{}
+	h.applyOptions(opts...)
+
+	h.assertRecursivelyNotEqual(t, reflect.ValueOf(valA), reflect.ValueOf(valB), "$")
+}
+
+// AssertDeepClonedEuqal tells whether a and b are deeply equal.
+func AssertDeepClonedEuqal[T any](t require.TestingT, valA, valB T, opts ...Option) {
```
XuHuaiyu, line 257:
```suggestion
func AssertDeepClonedEqual[T any](t require.TestingT, valA, valB T, opts ...Option) {
```

pkg/util/deeptest/statictesthelper.go
YangKeao: If there are two different implementations for an `interface`, it'd be expected to have different types here. They are also considered not equal.

pkg/util/deeptest/statictesthelper.go
YangKeao: Done.

pkg/util/deeptest/statictesthelper.go
YangKeao: Done.

pkg/util/deeptest/statictesthelper.go
YangKeao: Done.

pkg/util/deeptest/statictesthelper.go
YangKeao: Done.

pkg/util/deeptest/statictesthelper.go
YangKeao: Done.

XuHuaiyu (approved):

YangKeao:
:beers:

PR #55331, opened 2024-08-09
domain/rg: organize and enhance the code of RunawayChecker

JmPotato:

Issue Number: ref #54434.

- Delete some unnecessary code.
- Add more comments.
- Organize and enhance some functions.

### Check List

Tests 

- [x] Unit test
- [x] Integration test

### Release note

```release-note
None
```

### DIFF
--- a/pkg/domain/resourcegroup/runaway.go
+++ b/pkg/domain/resourcegroup/runaway.go
@@ -54,27 +54,6 @@ const (
 // NullTime is a zero time.Time.
 var NullTime time.Time
 
-// RunawayMatchType is used to indicate whether query was interrupted by runaway identification or quarantine watch.
-type RunawayMatchType uint
-
-const (
-	// RunawayMatchTypeWatch shows quarantine watch.
-	RunawayMatchTypeWatch RunawayMatchType = iota
-	// RunawayMatchTypeIdentify shows identification.
-	RunawayMatchTypeIdentify
-)
-
-func (t RunawayMatchType) String() string {
-	switch t {
-	case RunawayMatchTypeWatch:
-		return "watch"
-	case RunawayMatchTypeIdentify:
-		return "identify"
-	default:
-		panic("unknown type")
-	}
-}
-
 // RunawayRecord is used to save records which will be insert into mysql.tidb_runaway_queries.
 type RunawayRecord struct {
 	ResourceGroupName string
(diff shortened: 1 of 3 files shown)


pkg/domain/resourcegroup/runaway.go
```diff
@@ -499,50 +483,48 @@ func (r *RunawayChecker) BeforeExecutor() error {
...
-			r.markRunaway(RunawayMatchTypeWatch, action, &now)
-			// If no match action, it will do nothing.
-			switch action {
-			case rmpb.RunawayAction_Kill:
-				return exeerrors.ErrResourceGroupQueryRunawayQuarantine
-			case rmpb.RunawayAction_CoolDown:
-				// This action should be done in BeforeCopRequest.
-				return nil
-			case rmpb.RunawayAction_DryRun:
-				return nil
```
HuSharp, line 520: why removed `RunawayAction_DryRun`?

pkg/domain/resourcegroup/runaway.go
```diff
@@ -596,25 +569,43 @@ func (r *RunawayChecker) CheckCopRespError(err error) error {
...
+	r.watchAction = action
+	now := time.Now()
+	r.markRunaway("watch", action, &now)
 }
 
-func (r *RunawayChecker) markRunaway(matchType RunawayMatchType, action rmpb.RunawayAction, now *time.Time) {
-	actionStr := strings.ToLower(rmpb.RunawayAction_name[int32(action)])
-	metrics.RunawayCheckerCounter.WithLabelValues(r.resourceGroupName, matchType.String(), actionStr).Inc()
+func (r *RunawayChecker) markRunaway(matchType string, action rmpb.RunawayAction, now *time.Time) {
+	actionStr := strings.ToLower(action.String())
```
HuSharp, line 599: please change this line as well :)
https://github.com/pingcap/tidb/blob/7291829726b1c112e1179469c679b3bcd684eabd/pkg/executor/infoschema_reader.go#L3626

pkg/domain/resourcegroup/runaway.go
JmPotato: It can be covered by the `default` case, there is no need to explicitly write it here I think

HuSharp: I agree that this action can be similar to the default case.
> Previous behavior may have been wrong:
> Previously, the dryRun branch would return directly, but now it iterates over all identifiers in a for loop.

There is a concept of [dryRun](https://docs.pingcap.com/tidb/stable/tidb-resource-control#query_limit-parameters) in the documentation, but after removing this `RunawayAction_DryRun` in the code,  there is no dryRun to be found in the code (which may be confusing?). Do we need to put it here explicitly? Just like: 
```
		case rmpb.RunawayAction_DryRun:
		default:
```

JmPotato: Good catch, you're right!

JmPotato:
/cc @nolouch @glorv @HuSharp 

PTAL

HuSharp (approved):

glorv (approved):

D3Hunter:
/approve

for domain pkg

nolouch (approved):

PR #55400, opened 2024-08-14
table: remove method `Txn()` in `MutateContext`

### LINKED ISSUE #55378: Remove `Txn(bool)` in `table.MutateContext`
## Enhancement

For some methods like `Index.Create` or `Index.Delete`, a txn is already provided in the argument:

https://github.com/pingcap/tidb/blob/9a7e5cc2f7c433ce238a69a933251a565e089925/pkg/table/index.go#L91-L93

However, because `MutateContext` also provides a `Txn(bool)` method, the coder will be confused about which to choose. Furthur more, in some scenes, the argument `txn` and the return value of `MutateContext.Txn(true)` are not the same, for example in `RecoverIndexExec.backfillIndex`:

https://github.com/pingcap/tidb/blob/9a7e5cc2f7c433ce238a69a933251a565e089925/pkg/executor/admin.go#L324-L329

And in `RecoverIndexExec.backfillIndexInTxn`

https://github.com/pingcap/tidb/blob/9a7e5cc2f7c433ce238a69a933251a565e089925/pkg/executor/admin.go#L527

A separated txn is created by `kv.RunInNewTxn` not the txn in session.

But in implementation, the `index.create` uses the `ctx.Txn(true)` by mistake:

https://github.com/pingcap/tidb/blob/9a7e5cc2f7c433ce238a69a933251a565e089925/pkg/table/tables/index.go#L209-L213

Though there is no bug here because `untouched` is always true, we still need to care about it.

Maybe remove `Txn(bool)` from `MutateContext` is a better choice because then we can only use txn in argument .

lcwangchao:

Issue Number: close #55378

Problem Summary:

See descriptions in issue.

- Removed `Txn()` in `MuateContext`
- Add new parameter `txn kv.Transaction` in `AddRecord/UpdateRecord/RemoveRecord`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/table/tables/index.go
```diff
@@ -206,10 +206,6 @@ func (c *index) create(sctx table.MutateContext, txn kv.Transaction, indexedValu
 		}
 
 		if untouched {
-			txn, err1 := sctx.Txn(true)
```
lcwangchao, line 209: We should not use `sctx.Txn()` here because it can be different with the txn in argument.

tangenta: Is there a test covering this?

tangenta (approved):

lance6716 (commented):
/approve

pkg/table/tables/index.go
lcwangchao: There is not bug now because untouched is always false in DDL reorg and `admin recover index` scene. But it is a potential risk...

YangKeao (approved):
LGTM

Issue #53127, opened 2024-05-09
Add hash join v2 to make hash join more efficient


windtalker:
## Enhancement
Our current hash join's implementation is relatively slow, this issue aims at adding a new hash join implementation. The new hash join implementation will at least
* support concurrent build
* simplify probe code, and make join probe more efficient
* support use left side to build for semi join

PR #55470, opened 2024-08-16
executor: introduce tagged pointer in hash join v2

windtalker:

Issue Number: ref #53127

Problem Summary:

This pr add tagged pointer in hash join v2.
The basic idea of tagged pointer is for all the value of `unsafe.Pointer`, the first N(usually N is at least 16) MSB is all zero. So we can use these N bit to store extra information.
In Hash join v2, we use the N bit to store part of the value of hash value:
During hash join v2 build, we store the row address to the related slot in hash table. In this pr, instead of store the row address in hash table, we store the tagged pointer in hash table, and the tagged value is extracted from hash value.
For example
raw address is: `0x0000xxxxxxxxxxxx`
hash value is `0xabcdxxxxxxxxxxxx`
then before this pr, we save `0x0000xxxxxxxxxxxx` into hash table, after this pr we save `0xabcdxxxxxxxxxxxx` into hash table
During the probe stage, we can first use `0xabcd000000000000` to test against the hash value:
* if `0xabcd000000000000 & hashvalue` == `0xabcd000000000000`, it means the hash value might match, we need further compare the join key
* if `0xabcd000000000000 & hashvalue` != `0xabcd000000000000`, it means the hash value is different, then the probe result is definitely false, there is no need to further compare of the join key.

Test tag pointer on TPCH-50 dataset:
Build time: almost the same with/without tag pointer
Probe time: probe time is reduced by 15% with tag pointer
Probe collision: probe collision is reduce greatly with tag pointer(from `1594141458` to `232887049`)
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/join/tagged_ptr.go
```diff
@@ -0,0 +1,84 @@
...
+type taggedPtr uintptr
+
+type tagPtrHelper struct {
+	// mask to get the tagged value in taggedPtr
+	taggedMask uint64
+}
+
+func (th *tagPtrHelper) init(taggedBits uint8) {
+	hashValueTaggedMask := uint64(1<<taggedBits) - 1
+	hashValueTaggedOffset := uint64(64 - taggedBits)
```
XuHuaiyu, line 46:
```suggestion
	hashValueTaggedOffset := 64 - taggedBits
```

pkg/executor/join/tagged_ptr.go
```diff
@@ -0,0 +1,84 @@
...
+	*(*unsafe.Pointer)(unsafe.Pointer(&ret)) = ptr
+	// tag the value
+	ret = taggedPtr(uint64(ret) | taggedValue)
+	return ret
+}
+
+func (th *tagPtrHelper) toUnsafePointer(tPtr taggedPtr) unsafe.Pointer {
+	// clear the tagged value
+	tPtr = taggedPtr(uint64(tPtr) & ^th.taggedMask)
+	return *(*unsafe.Pointer)(unsafe.Pointer(&tPtr))
```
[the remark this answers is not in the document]
windtalker: It is by design that we `save the unsafe.Pointer into the memory of tPtr`

pkg/executor/join/tagged_ptr.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (th *tagPtrHelper) getTaggedValue(hashValue uint64) uint64 {
+	return hashValue & th.taggedMask
+}
+
+func (*tagPtrHelper) toTaggedPtr(taggedValue uint64, ptr unsafe.Pointer) taggedPtr {
+	ret := taggedPtr(0)
+	// first save ptr into taggedPtr
+	*(*unsafe.Pointer)(unsafe.Pointer(&ret)) = ptr
+	// tag the value
```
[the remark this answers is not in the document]
windtalker: It is by design that we `save the unsafe.Pointer into the memory of tPtr`

pkg/executor/join/tagged_ptr.go
windtalker: Done

pkg/executor/join/base_join_probe.go
```diff
@@ -106,8 +106,9 @@ type baseJoinProbe struct {
 	selRows  []int
 	usedRows []int
 	// matchedRowsHeaders, serializedKeys is indexed by logical row index
-	matchedRowsHeaders []uintptr // the start address of each matched rows
-	serializedKeys     [][]byte  // used for save serialized keys
+	matchedRowsHeaders   []taggedPtr // the start address of each matched rows
+	matchedRowsHashValue []uint64    // the start address of each matched rows
```
XuHuaiyu, line 110: unmatched comment

wshwsh12 (approved):
LGTM

XuHuaiyu (approved):

D3Hunter:
what's the overall performance gain when running some SQL with hash job?

windtalker:
> what's the overall performance gain when running some SQL with hash job?

you mean hash join v2 vs hash join v1 or  hash join v2 with tagged pointer vs  hash join v2 without tagged pointer?

D3Hunter:
> > what's the overall performance gain when running some SQL with hash job?
> 
> you mean hash join v2 vs hash join v1 or hash join v2 with tagged pointer vs hash join v2 without tagged pointer?

hash job v2 I suppose? before and after this PR

windtalker:
> > > what's the overall performance gain when running some SQL with hash job?
> > 
> > 
> > you mean hash join v2 vs hash join v1 or hash join v2 with tagged pointer vs hash join v2 without tagged pointer?
> 
> hash job v2 I suppose? before and after this PR

For this pr, test on TPCH-50 shows that there is no significant overall performance gains between hash join v2 with tagged pointer and hash join v2 without tagged pointer. When looking into the hash join itself, the probe time is 15% faster with tagged pointer, and the probe collision is reduced by 10 times, which I think can effectively reduce the CPU time of hash join

D3Hunter:
> which I think can effectively reduce the CPU time of hash join

any compare result about this part? I personally prefer other ways of optimization to bit manipulation, unless we have no choice.

D3Hunter:
/hold

hold for a while to get more test result.

windtalker:
> > which I think can effectively reduce the CPU time of hash join
> 
> any compare result about this part? I personally prefer other ways of optimization to bit manipulation, unless we have no choice and it does show significant improve

Probe collision: probe collision is reduce from 1594141458 to 232887049
For each probe collision, probe need to do:
https://github.com/pingcap/tidb/blob/eb19c100db5effe6ca86c8cd75ddfac1dde61e38/pkg/executor/join/inner_join_probe.go#L48-L58
1. compare the join key(`isKeyMatched`)
2. advance current row to the next row(`getNextRowAddress`)

So I think if we can reduce the probe collision, it is for sure that we can reduce the cpu time.

Since hash join v2 use a much simple hash table, it is by design that it may meet many probe colision, especially for the cases that the build side has many duplicate keys.
The problem this pr tries to solve is actually
* For extrame cases(build side has many duplicated keys), avoid big performance regression for the extrame cases
* For normal cases, the overhead should be as small as possible, so it will not introduce performance regression

In order to avoid probe colision, the basic idea is to use some quick filter. 
There is at least 3 intuitive approaches
* use tagged pointer
* similar with tagged pointer, but instead of save part of hash value into pointer, each row allocated an extra 8 bytes to save information of the hash value
* for each hash table slot, use a bloom filter

For all the 3 candidates, tagged pointer has the smallest cost, and according to the tests results, it will not cause performance regression. For method 2, it need extra 8 byte for each build row, it cause more memory usage and may cause performance regression for some simple query. For method 3, it need extra memory for each hash table slot, and the cost of construct bloom filter is relatively big and may cause performance regression for simple query.

D3Hunter:
> So I think if we can reduce the probe collision, it is for sure that we can reduce the cpu time.

Thanks for the detailed explanation. it makes sense, but it would be more helpful if we can have a compare result of how much CPU this PR can reduce

windtalker:
> > So I think if we can reduce the probe collision, it is for sure that we can reduce the cpu time.
> 
> Thanks for the detailed explanation. it makes sense, but it would be more helpful if we can have a compare result of how much CPU this PR can reduce

For a join, the explain analyze result will be something like this
```
| └─HashJoin_11                  | 12487.50 | 1       | root      |               | time:533.9µs, loops:2, build_hash_table:{total:484.2µs, fetch:475.4µs, build:500ns}, probe:{concurrency:5, total:2.45ms, max:496.3µs, probe:10.2µs, fetch and wait:2.44ms}     
```
If we simply treated the `probe: 10.2us` as the probe cpu time, then with this pr, the probe cpu time can be reduced by 15%. I don't have the original data now, I can re-test and record the original data later.

windtalker:
> > So I think if we can reduce the probe collision, it is for sure that we can reduce the cpu time.
> 
> Thanks for the detailed explanation. it makes sense, but it would be more helpful if we can have a compare result of how much CPU this PR can reduce

Test Dataset TPCH-50
Test query: tpch query 1 - tpch query 22
| |hash join build time | hash join probe time|
|--|--|--|
| without tagged pointer| 50341ms| 473064ms| 
|with tagged pointer| 51836ms| 408343ms|

Issue #46607, opened 2023-09-02
Control the max number of row and bytes lighting sends to tidb


guoshouyan:
## Feature Request

**Is your feature request related to a problem? Please describe:**

In lightning physical backend, we can set the store-write-bwlimit to control how much data lightning send to tikv.
Can we have similar feature in logical mode?
We already have minDeliverBytes [here](https://github.com/pingcap/tidb/blob/b524a2316f08dbaade3b8cd3c7189f90895ea95e/br/pkg/lightning/importer/import.go#L2204) but it's not configurable. 
Can we make it configurable and also have another config like minDeliverRows to control the rows it sends?

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

lance6716:
Hi, not sure if I understand it correctly. In logical backend lightning will try to send data every 1MB. (`be.MaxChunkSize()` always return 1MB)

https://github.com/pingcap/tidb/blob/a4271136ecfdc8352f9ee30ddd6b707b04484236/br/pkg/lightning/backend/tidb/tidb.go#L614-L617

with the concurrency of `lightning.region-concurrency` from the configuration file. I think you want to make 1MB configurable?

And can you explain why the default value is not good enough?

D3Hunter:
> Hi, not sure if I understand it correctly. In logical backend lightning will try to send data every 1MB. (`be.MaxChunkSize()` always return 1MB)
> 
> https://github.com/pingcap/tidb/blob/a4271136ecfdc8352f9ee30ddd6b707b04484236/br/pkg/lightning/backend/tidb/tidb.go#L614-L617
> 
> with the concurrency of `lightning.region-concurrency` from the configuration file. I think you want to make 1MB configurable?
> 
> And can you explain why the default value is not good enough?

this 1M is useless, as deliverLoop send in 96K batch, see https://github.com/pingcap/tidb/pull/46997#issuecomment-1880744664

PR #55482, opened 2024-08-17
lightning: support prepared statement and client stmt cache in logical import mode

### LINKED ISSUE #54850: lightning logical mode import should support persistent connection pooling and prepared statement and client statement cache
## Enhancement
during the tidb mode import, seems lightning destroy the connection after the insert, making it impossible to reuse the connection and plan cache hit.

image

high overhead on sql parse and compile
image

dbsid:

Issue Number: close #54850

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
add doc for logical-import-prep-stmt
```

Test logical import on a table with 1200 columns

Baseline
image

With this PR
image

image

image


lance6716:
please merge master to fix lightning CI @dbsid

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -309,6 +334,11 @@ type tidbBackend struct {
 	// affecting the cluster too much.
 	maxChunkSize uint64
 	maxChunkRows int
+	// implement stmtCache to improve performance, especially when the downstream is TiDB
```
D3Hunter, line 337:
```suggestion
	// implement stmtCache to improve performance
```
we don't support downstream other than TiDB

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -309,6 +334,11 @@ type tidbBackend struct {
...
+	stmtCache      *kvcache.SimpleLRUCache
```
D3Hunter, line 338: as long as the table is importing, the prepared stmt is needed, no need such LRU

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -564,16 +602,24 @@ func (enc *tidbEncoder) Encode(row []types.Datum, _ int64, columnPermutation []i
...
+	}
 	cnt := 0
 	for i, field := range row {
 		if enc.columnIdx[i] < 0 {
 			continue
 		}
 		if cnt > 0 {
 			encoded.WriteByte(',')
+			if enc.logicalImportPrepStmt {
+				preparedInsertStmt.WriteByte(',')
```
D3Hunter, line 621: only need keep one of them

pkg/lightning/config/config.go
```diff
@@ -1101,6 +1101,7 @@ type TikvImporter struct {
 	StoreWriteBWLimit       ByteSize `toml:"store-write-bwlimit" json:"store-write-bwlimit"`
 	LogicalImportBatchSize  ByteSize `toml:"logical-import-batch-size" json:"logical-import-batch-size"`
 	LogicalImportBatchRows  int      `toml:"logical-import-batch-rows" json:"logical-import-batch-rows"`
+	LogicalImportPrepStmt   bool     `toml:"logical-import-prep-stmt" json:"logical-import-prep-stmt"`
```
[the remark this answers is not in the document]
dbsid: I prefer keep the config, in case we need to fallback to raw query.

pkg/lightning/backend/tidb/tidb.go
dbsid: The reason to use LRU here is that the batch insert prepared statement may not be the same. where the insert/replace statement is constrained by `logical-import-batch-size` , the actual SQL executed might be longer or shorter, depending on the actual content imported.

pkg/lightning/backend/tidb/tidb.go
dbsid: Currently the `splitIntoChunks` and `tidbRow.Size()` is based on the length of  `tidbRow.insertStmt`, I would like to resue this logic so keep the insertStmt here.

D3Hunter:
/hold

need confirm from PM

pkg/lightning/backend/tidb/tidb.go
D3Hunter: depends on row size is unstable for prepared stmt, in worse case, every prepared stmt is different. we can ignore batch-size in this case, and only honors batch-rows, cc @lance6716

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -715,7 +778,7 @@ func (be *tidbBackend) WriteRowsToDB(ctx context.Context, tableName string, colu
 		var finalInsertStmt strings.Builder
 		finalInsertStmt.WriteString(is)
 		finalInsertStmt.WriteString(row.insertStmt)
-		stmtTasks = append(stmtTasks, stmtTask{[]tidbRow{row}, finalInsertStmt.String()})
+		stmtTasks = append(stmtTasks, stmtTask{[]tidbRow{row}, finalInsertStmt.String(), []any{}})
```
D3Hunter, line 781: `stmt` in this task will always be a non-prepared stmt

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -753,8 +816,27 @@ stmtLoop:
...
+				key := &stmtKey{query: query}
+				be.stmtCacheMutex.RLock()
+				stmt, ok := be.stmtCache.Get(key)
+				be.stmtCacheMutex.RUnlock()
+				if ok {
+					prepStmt = stmt.(*sql.Stmt)
+				} else if stmt, err := be.db.Prepare(query); err == nil {
+					prepStmt = stmt
+					be.stmtCacheMutex.Lock()
+					be.stmtCache.Put(key, stmt)
```
D3Hunter, line 831: `execStmts` is called concurrently, this might override existing stmt without close it.

pkg/lightning/config/config.go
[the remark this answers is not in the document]
D3Hunter: please add it to tidb-lightning.toml

pkg/lightning/config/config.go
D3Hunter, line 1: add some integration test for new config

pkg/lightning/backend/tidb/tidb.go
dbsid: the default `logical-import-batch-rows` is 65536, usually this batch size is too large, should we lower the default value of `logical-import-batch-rows` to some reasonable number like 128/256/512?

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -753,8 +816,27 @@ stmtLoop:
...
+			query := stmtTask.stmt
+			if be.cachePrepStmts {
+				var prepStmt *sql.Stmt
+				key := &stmtKey{query: query}
+				be.stmtCacheMutex.RLock()
+				stmt, ok := be.stmtCache.Get(key)
+				be.stmtCacheMutex.RUnlock()
+				if ok {
+					prepStmt = stmt.(*sql.Stmt)
+				} else if stmt, err := be.db.Prepare(query); err == nil {
```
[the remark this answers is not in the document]
dbsid: the default logical-import-batch-rows is 65536, usually this batch size is too large, should we lower the default value of logical-import-batch-rows to some reasonable number like 128/256/512?

pkg/lightning/backend/tidb/tidb.go
dbsid: yes, cached prepstmt will be concurrent use by multiple goroutines. any idea how to avoid override existing stmt without close it?

pkg/lightning/common/util.go
```diff
@@ -51,6 +51,9 @@ const (
 	retryTimeout = 3 * time.Second
 
 	defaultMaxRetry = 3
+
+	// DefaultMaxIdleConns is the default value for sql.DB's MaxIdleConns.
+	defaultMaxIdleConns = 128
```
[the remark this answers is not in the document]
lance6716: I'm not sure we should keep more connections because it's consume like network IO resource rather than CPU resource.

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -287,6 +296,22 @@ func (*targetInfoGetter) CheckRequirements(ctx context.Context, _ *backend.Check
...
 
+// stmtKey defines key for stmtCache.
+type stmtKey struct {
+	query string
+	// `hash` is the hash value of this object.
+	hash []byte
+}
+
+// Hash implements SimpleLRUCache.Key.
+func (k *stmtKey) Hash() []byte {
```
lance6716, line 307: (I'm not sure why the `kvcache.Key` interface call it `Hash`. Hash means a fixed-length digest)

pkg/lightning/backend/tidb/tidb.go
lance6716: Maybe we can ask the user that request this feature what's the root cause https://github.com/pingcap/tidb/issues/46607 . If the prepared statement can solve that problem, we can ignore these configurations. @D3Hunter and I are busy this week, please help to contact the issue author.

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -333,13 +363,21 @@ func NewTiDBBackend(
 		log.FromContext(ctx).Warn("unsupported conflict strategy for TiDB backend, overwrite with `error`")
 		onDuplicate = config.ErrorOnDup
 	}
+	stmtCache := kvcache.NewSimpleLRUCache(prepStmtCacheSize, 0, 0)
```
lance6716, line 366: I don't know what's the meaning of the parameters `guard` and `quota` 😂 I'll learn them later.

And no need to create the LRU cache when `LogicalImportPrepStmt` is false? Or we can use nilness of `stmtCache` as the bool `cachePrepStmts`?

lance6716 (commented):
will review later

pkg/lightning/backend/tidb/tidb.go
lance6716: after obtaining the lock, check again the existence of cache entry.

pkg/lightning/backend/tidb/tidb.go
kennytm: Not the author of #46607, but we've got customer which database is unable to handle 96 KB transactions thus requiring the limit to be lowered.

The `-batch-size` does not need to be very precise. It is mainly used to ensure (number of rows) &times; (size of row) to be maintained at a constant limit. You should *not* ignore the batch-size entirely because the same limit is shared among all tables and the constraint should be reasonable for both wide and narrow tables.

pkg/lightning/common/util.go
```diff
@@ -142,6 +143,11 @@ func (param *MySQLConnectParam) Connect() (*sql.DB, error) {
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
+	// The actual number of alive connections is controlled by the region concurrency
+	// The setting is required to avoid frequent connection creation and close
+	if db != nil {
```
lance6716, line 148: I think we don't need to check nilness. We have make sure it's not nil  when `err` is nil

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -287,6 +296,22 @@ func (*targetInfoGetter) CheckRequirements(ctx context.Context, _ *backend.Check
...
+type stmtKey struct {
+	query string
+	// `hash` is the hash value of this object.
+	hash []byte
+}
+
+// Hash implements SimpleLRUCache.Key.
+func (k *stmtKey) Hash() []byte {
+	if len(k.hash) == 0 {
+		k.hash = make([]byte, 0, len(k.query))
```
lance6716, line 309: Maybe use table name + number of rows as key? I'm not sure if comparing the bytes is a bottleneck in lightning's CPU profile.

pkg/lightning/backend/tidb/tidb.go
lance6716: Please address this comment. Currently caching prep statement and using prep statement are the same thing, we don't need two members.

pkg/lightning/common/util.go
dbsid: If I remove the nilness check, the unit test failed, not sure why https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fghpr_unit_test/detail/ghpr_unit_test/17153/pipeline

lance6716: Sorry we have a failpoint that returns (nil, nil)

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -287,6 +296,22 @@ func (*targetInfoGetter) CheckRequirements(ctx context.Context, _ *backend.Check
...
+		k.hash = append(k.hash, hack.Slice(k.query)...)
```
lance6716, line 310: we can directly return `hack.Slice(k.query)`? No need to use k.hash field.

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -753,8 +825,31 @@ stmtLoop:
...
+				stmt, ok := be.stmtCache.Get(key)
+				be.stmtCacheMutex.RUnlock()
+				if ok {
+					prepStmt = stmt.(*sql.Stmt)
+				} else if stmt, err := be.db.PrepareContext(ctx, query); err == nil {
+					prepStmt = stmt
+					be.stmtCacheMutex.Lock()
+					// check again if the key is already in the cache
+					// to avoid override existing stmt without closing it
+					if _, ok := be.stmtCache.Get(key); !ok {
```
lance6716, line 842: if the cache entry exists, we need to find a place to close this `stmt` because it can't be closed by `SetOnEvict`

lance6716 (commented):
rest lgtm

pkg/lightning/common/util.go
dbsid: should I get back the check `if db != nil `?

lance6716: yes, sorry for my previous comment

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -287,6 +296,21 @@ func (*targetInfoGetter) CheckRequirements(ctx context.Context, _ *backend.Check
 	return nil
 }
 
+// stmtKey defines key for stmtCache.
+type stmtKey struct {
+	query string
+	// `hash` is the hash value of this object.
+	hash []byte
```
lance6716, line 303: I think this field is not used, we can directly use `hack.Slice(k.query)`

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -753,8 +821,33 @@ stmtLoop:
...
+					prepStmt = stmt.(*sql.Stmt)
+				} else if stmt, err := be.db.PrepareContext(ctx, query); err == nil {
+					prepStmt = stmt
+					be.stmtCacheMutex.Lock()
+					// check again if the key is already in the cache
+					// to avoid override existing stmt without closing it
+					if _, ok := be.stmtCache.Get(key); !ok {
+						be.stmtCache.Put(key, stmt)
+					} else {
+						defer prepStmt.Close()
```
lance6716, line 841: we can let `prepStmt` be assigned to the cached entry, and directly Close `stmt` here.

lance6716 (commented):
rest lgtm

lance6716 (approved):

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -712,11 +769,17 @@ func (be *tidbBackend) WriteRowsToDB(ctx context.Context, tableName string, colu
 	}
 	is := insertStmt.String()
 	stmtTasks := make([]stmtTask, 0, len(rows))
+	var values []any
```
D3Hunter, line 772: we can alloc with `len(rows) * len(rows[0].values)`

pkg/lightning/config/config.go
```diff
@@ -1101,6 +1102,7 @@ type TikvImporter struct {
...
```
D3Hunter, line 1105: please add a integration test for it

pkg/lightning/backend/tidb/tidb.go
```diff
@@ -712,11 +769,17 @@ func (be *tidbBackend) WriteRowsToDB(ctx context.Context, tableName string, colu
...
 	for _, row := range rows {
 		var finalInsertStmt strings.Builder
 		finalInsertStmt.WriteString(is)
-		finalInsertStmt.WriteString(row.insertStmt)
-		stmtTasks = append(stmtTasks, stmtTask{[]tidbRow{row}, finalInsertStmt.String()})
+		if be.stmtCache != nil {
+			finalInsertStmt.WriteString(row.preparedInsertStmt)
+			values = append(values, row.values...)
```
D3Hunter, line 778: please add ut to cover the case that Batch failed and fallback to single row insert with prepared-stmt

D3Hunter (commented):
rest lgtm

pkg/lightning/config/config.go
dbsid: does the test of TestAdjustLogicalImportPrepStmt look good to you?

D3Hunter: ut it not enough for new features that's visible to user

D3Hunter:
please prepare a doc pr to this feature

dbsid:
> please prepare a doc pr to this feature

https://github.com/pingcap/docs/pull/18902
https://github.com/pingcap/docs-cn/pull/18627

D3Hunter (approved):

pkg/lightning/config/config.go
```diff
@@ -78,6 +78,7 @@ const (
 	DefaultRegionSplitBatchSize       = 4096
 	defaultLogicalImportBatchSize     = 96 * units.KiB
 	defaultLogicalImportBatchRows     = 65536
+	defaultLogicalImportPrepStmt      = false
```
qw4990, line 81: Why not enable this by default?

qw4990 (approved):

yudongusa (approved):
LGTM

PR #55494, opened 2024-08-19
server: wait for auto commit queries before shutdown

### LINKED ISSUE #55464: TiDB will not wait for auto-commit statements during graceful shutdown
## Bug Report

In the following logic, `conn.getCtx().GetSessionVars().InTxn()` is only true for explictly transactions. Therefore, the `DrainClients` will not wait for auto-commit statements, and results in leaking locks.

```
// DrainClients drain all connections in drainWait.
// After drainWait duration, we kill all connections still not quit explicitly and wait for cancelWait.
func (s *Server) DrainClients(drainWait time.Duration, cancelWait time.Duration) {
...
	go func() {
		defer close(allDone)
		for _, conn := range conns {
			if !conn.getCtx().GetSessionVars().InTxn() {
				continue
			}
			select {
			case <-conn.quit:
			case <-quitWaitingForConns:
				return
			}
		}
	}()
...
}
```

It'd be better to also wait for all auto-commit statements to finish :thinking: .

YangKeao:

Issue Number: close #55464

Problem Summary:

Previously, TiDB will not wait for auto-commit queries. These queries are directly killed. It'd be better to wait for them to finish.

Also checking the connection status when filtering the connection to wait.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Run a workload with heavy auto-commit traffic, and you'll see the connections are closed with little leaked lock.

### Release note

```release-note
None
```

### DIFF
--- a/pkg/server/conn.go
+++ b/pkg/server/conn.go
@@ -1126,15 +1126,20 @@ func (cc *clientConn) Run(ctx context.Context) {
 			return
 		}
 
-		// Should check InTxn() to avoid execute `begin` stmt.
+		// It should be CAS before checking the `inShutdownMode` to avoid the following scenario:
+		// 1. The connection checks the `inShutdownMode` and it's false.
+		// 2. The server sets the `inShutdownMode` to true. The `DrainClients` process ignores this connection
+		//   because the connection is in the `connStatusReading` status.
+		// 3. The connection changes its status to `connStatusDispatching` and starts to execute the command.
+		if !cc.CompareAndSwapStatus(connStatusReading, connStatusDispatching) {
+			return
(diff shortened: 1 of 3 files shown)


YangKeao:
/hold

Testing with a big cluster now.

Defined2014 (approved):

YangKeao:
/unhold

The bot ignored my commands :angry:

YangKeao:
/hold

Testing now

YangKeao:
/unhold

No problem. I made some mistakes while testing.

xhebox (approved):

Issue #55540, opened 2024-08-20
Introduce additional labels to TOP SQL


nolouch:
## Enhancement
add `table_id`, schema to `TOPSQL`, then we can know the cost by the table_id or schema

- [x] add kvproto https://github.com/pingcap/tipb/pull/338 
- [x] add the label for the request https://github.com/pingcap/tidb/pull/55555
- [x] add the new label for the scraper https://github.com/pingcap/ng-monitoring/pull/260/
- [x]  topsql backend API  https://github.com/pingcap/ng-monitoring/pull/262
- [x] UI part https://github.com/pingcap/tidb-dashboard/pull/1732
    - ![demo2](https://github.com/user-attachments/assets/1e3af788-1096-4253-9c37-25a32b6b9c5d)
- [x]  Update the dashboard version in PD https://github.com/tikv/pd/pull/8656

docs:
- [ ] https://github.com/pingcap/docs-cn/pull/18664 
- [ ] https://github.com/pingcap/docs/pull/18982

PR #55555, opened 2024-08-21
TopSQL: add table id in resource tag

nolouch:

Issue Number:  ref https://github.com/pingcap/tidb/issues/55540
Problem Summary:
*: add table id in resource tag
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```


pkg/kv/kv.go
```diff
@@ -766,3 +770,87 @@ const (
...
+		}
+		key = k
+		if !key.HasPrefix(tablePrefix) {
+			return 0
+		}
+	}
+	key = key[len(tablePrefix):]
+	_, tableID, _ := codec.DecodeInt(key)
+	return tableID
+}
```
XuHuaiyu, line 856:
```suggestion
var DecodeTableIDFunc func(Key) int64

func decodeTableID(key Key) int64 {
	if DecodeTableIDFunc != nil {
		return DecodeTableIDFunc(key)
	}
	return 0
}
```

``` diff
diff --git a/pkg/tablecodec/tablecodec.go b/pkg/tablecodec/tablecodec.go
index 042f620f2e..698fdfcecc 100644
--- a/pkg/tablecodec/tablecodec.go
+++ b/pkg/tablecodec/tablecodec.go
@@ -1902,3 +1902,7 @@ func VerifyTableIDForRanges(keyRanges *kv.KeyRanges) ([]int64, error) {
        err := keyRanges.ForEachPartitionWithErr(collectFunc)
        return tids, err
 }
+
+func init(){
+       kv.DecodeTableIDFunc = DecodeTableID
+}
```

pkg/kv/kv.go
```diff
@@ -766,3 +770,87 @@ const (
...
+func (b *ResourceGroupTagBuilder) EncodeTagWithKey(key []byte) []byte {
+	tag := &tipb.ResourceGroupTag{}
+	if b.sqlDigest != nil {
+		tag.SqlDigest = b.sqlDigest.Bytes()
+	}
+	if b.planDigest != nil {
+		tag.PlanDigest = b.planDigest.Bytes()
+	}
+	if len(key) > 0 {
+		tag.TableId = decodeTableID(key)
```
yibin87, line 815: Do we need to consider any error cases here?

pkg/util/resourcegrouptag/resource_group_tag.go
```diff
@@ -125,6 +125,12 @@ func GetFirstKeyFromRequest(req *tikvrpc.Request) (firstKey []byte) {
 				}
 			}
 		}
+	case *kvrpcpb.PessimisticLockRequest:
```
yibin87, line 128: Why add this new case?

pkg/kv/kv.go
```diff
@@ -766,3 +770,87 @@ const (
...
+	accessKey  []byte
+}
+
+// NewResourceGroupTagBuilder creates a new ResourceGroupTagBuilder.
+func NewResourceGroupTagBuilder() *ResourceGroupTagBuilder {
+	return &ResourceGroupTagBuilder{}
+}
+
+// SetSQLDigest sets the sql digest for the request.
+func (b *ResourceGroupTagBuilder) SetSQLDigest(digest *parser.Digest) *ResourceGroupTagBuilder {
```
yibin87, line 787: It seems not common to return self for SetXXX function in tidb's existing go code

pkg/kv/kv.go
nolouch: If the decoded fails, it will be tagged as 0.

pkg/kv/kv.go
nolouch: Here example are: https://github.com/pingcap/tidb/blob/master/pkg/distsql/request_builder.go#L177-L200

pkg/util/resourcegrouptag/resource_group_tag.go
nolouch: I just think it was missing, so I added it.

pkg/kv/kv.go
yibin87: OK

pkg/kv/kv.go
yibin87: I've no idea, do we need to distinguish real '0' with the error '0'?

nolouch: I think no need, and the real `table_id` must be greater than 0, https://github.com/pingcap/tidb/blob/master/pkg/meta/meta.go#L253-L254

yibin87: Got it

yibin87 (approved):
LGTM

nolouch:
PTAL @XuHuaiyu

crazycs520 (approved):

XuHuaiyu (approved):

tangenta (approved):

PR #55570, opened 2024-08-21
planner: add canonical hasher to take in primitive type directly for hashing.

AilinKid:
### What problem does this PR solve?

Issue Number: ref https://github.com/pingcap/tidb/issues/51664

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #55652, opened 2024-08-26
planner: integrate hashEqual interface into LogicalPlan and expression.Expression.

AilinKid:

Problem Summary:

in the https://github.com/pingcap/tidb/pull/55570 we introduced a new kind of HashEqual interface, for simplest hash uint64 digest from some primitive types.

And this kind of interface should be implemented by logicalPlan and expression.Expression to quick hash themselves and to compare the same type structure if they are equivalent.  

The first thing came into my mind is that, we should let comprehensive LogicalPlan interface and expression.Expression go include this small trait, why i find it will cause too much change for interface's original implementation. Because it's a generic interface, when it is integrated in: it will make this two to be generic

3eec3c1e-76bf-48f5-aab3-ce345c28bf9f

9b7cf491-f5c8-403c-88ef-594ffb6c2e56

which will cause all the place that referred to this two big interface to be redefined as xxx[any] as well. it's not that clean work. So we think about a way to define two new interface as:
```
// LogicalOperator is the interface for logical operators.
// It is used to represent the logical operators in the cascades framework.
// Since the LogicalPlan are the only implementation of normal interface{}
// with generic. Once we introduce a new generic HashEquals[T] inside it,
// like:
//
//	type LogicalPlan interface {
//			Plan
//	 	HashEquals[T]
//	 	...
//	}
//
// it will make the every implementor of LogicalPlan changed a lot. That's
// why this upper kind of LogicalOperator interface is introduced. And it
// will be used in the cascades framework.
//
// With this interface, every original LogicalPlan implementor don't need
// to change any old implementations, only have to add the two functions of
// Hash64 and Equals to become a LogicalOperator.
type LogicalOperator[T any] interface {
	base.LogicalPlan
	HashEquals[T]
}

// ScalarOperator is the interface for scalar operators.
// It is used to represent the scalar operators in the cascades framework.
// It is named for expression.Expression with HashEquals[T] interface. For
// the same reason as above LogicalOperator, we introduce this interface to
// avoid changing the original implementation of expression.Expression.
type ScalarOperator[T any] interface {
	expression.Expression
	HashEquals[T]
}

```
the logical expression that will be used in memo, we will referred them as LogicalOperator, which kept all the elements that LogicalExpression has while requiring HashEquals implementations on it. With that we don't need to do much change about current implementation function, only adding these two addition functional is enough.

Finally, after discussion, we derive a solution like below, getting rid of generic and use any instead.
```
// Hash64 is the interface for hashcode.
// It is used to calculate the lossy digest of an object to return uint64
// rather than compacted bytes from cascaded operators bottom-up.
type Hash64 interface {
	// Hash64 returns the uint64 digest of an object.
	Hash64(h Hasher)
}

// Equals is the interface for equality check.
// When we need to compare two objects when countering hash conflicts, we can
// use this interface to check whether they are equal.
type Equals interface {
	// Equals checks whether two base objects are equal.
	Equals(any) bool
}

// HashEquals is the interface for hash64 and equality check.
type HashEquals interface {
	Hash64
	Equals
}
```
And we did a benchmark for this, the any usage seem adequate: https://github.com/pingcap/tidb/pull/55652/files#diff-ef9b611a397bd5133fb3b00e5fca906beee78179821ee20a308e22110ead85aeR93

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/operator/logicalop/logical_projection.go
```diff
@@ -55,7 +57,26 @@ func (p LogicalProjection) Init(ctx base.PlanContext, qbOffset int) *LogicalProj
 	return &p
 }
 
-// *************************** start implementation of Plan interface ***************************
+// *************************** start implementation of HashEquals interface ****************************
+
+func (p *LogicalProjection) Hash64(h mutil.Hasher) {
```
AilinKid, line 62: just a future example

pkg/planner/core/operator/logicalop/base_logical_plan.go
```diff
@@ -50,6 +55,21 @@ type BaseLogicalPlan struct {
...
+func (p *BaseLogicalPlan) Hash64(h mutil.Hasher) {
+	h.HashInt(p.ID())
+}
+
+// Equals implements HashEquals.<1st> interface.
+func (p *BaseLogicalPlan) Equals(other *BaseLogicalPlan) bool {
+	if other != nil {
+		return p.ID() == other.ID()
+	}
+	return false
```
winoros, line 70: I think we can add `intest.Assert(false, "should not be called")`

AilinKid: yes, great

pkg/planner/core/operator/logicalop/logical_projection.go
```diff
@@ -55,7 +56,52 @@ func (p LogicalProjection) Init(ctx base.PlanContext, qbOffset int) *LogicalProj
...
+	if other == nil {
+		return false
+	}
+	proj, ok := other.(*LogicalProjection)
+	if !ok {
+		return false
+	}
+	// todo: LogicalSchemaProducer should implement HashEquals interface, otherwise, its self elements
+	// like schema and names are lost.
+	if !p.LogicalSchemaProducer.Equals(&proj.BaseLogicalPlan) {
```
elsa0520, line 90: Have you ever implemented the equals function of LogicalSchemaProducer ?

AilinKid: currently not, this is is just a example and reminder for next few pull requests.

pkg/planner/core/operator/logicalop/logical_projection.go
```diff
@@ -55,7 +56,52 @@ func (p LogicalProjection) Init(ctx base.PlanContext, qbOffset int) *LogicalProj
...
+		return false
+	}
+	//for i, one := range p.Exprs {
```
hawkingrei, line 93: Can we remove it?

pkg/planner/cascades/base/base_test.go
```diff
@@ -0,0 +1,99 @@
+// Copyright 2024 PingCAP, Inc.
```
hawkingrei, line 1: Can we add this bench into daily-bench？ so that we can measure the benchmark result.

pkg/planner/cascades/base/base_test.go
```diff
@@ -0,0 +1,99 @@
...
+// cpu: 12th Gen Intel(R) Core(TM) i7-12700KF
+// BenchmarkSuperEquals
+// BenchmarkSuperEquals-20    	1000000000	         0.9047 ns/op
+//func BenchmarkSuperEquals(b *testing.B) {
+//	tc1 := &testcase{a: 1, b: 2, c: "3"}
+//	var tc2 TestSuperEquals = &testcase{a: 1, b: 2, c: "3"}
+//	for i := 0; i < b.N; i++ {
+//		tc1.EqualsT(tc2)
+//	}
+//}
```
hawkingrei, line 69: Remove it ?

pkg/planner/core/operator/logicalop/logical_projection.go
AilinKid: no, next pull request will introduce hashEqual to expression.Expression, which will enable the comment line here.

Makefile
```diff
@@ -372,6 +372,7 @@ bench-daily:
 	go test github.com/pingcap/tidb/pkg/util/codec -run TestBenchDaily -bench Ignore --outfile bench_daily.json
 	go test github.com/pingcap/tidb/pkg/distsql -run TestBenchDaily -bench Ignore --outfile bench_daily.json
 	go test github.com/pingcap/tidb/pkg/statistics -run TestBenchDaily -bench Ignore --outfile bench_daily.json
+	go test github.com/pingcap/tidb/pkg/planner/cascades/base -run TestBenchDaily -bench Ignore --outfile bench_daily.json
```
elsa0520, line 375: Maybe we need to remove this bench test

pkg/planner/cascades/base/base_test.go
AilinKid: removed after discussion with elsa, it not about SQL optimization cycle, just a based func, it shouldn't be changed from a long term

Makefile
AilinKid: make sense, done

elsa0520 (approved):
LGTM

pkg/planner/core/operator/logicalop/logical_projection.go
hawkingrei: OK

hawkingrei (approved):

PR #55664, opened 2024-08-26
runaway: Mitigation  tidb_runaway_queries flooding

HuSharp:

Issue Number: ref #54434

Problem Summary: There could be a large number of runaway queries that are identified in a short period. 
Currently, the TiDB logs each occurrence which flood the log table. Since it's likely the most queries are identical, we shall introduce a way to compress the logs.  

#### changed
- Include `sql_digest` of this query in this table. If users add manual rules with sql_digest, they can easily identify the hits by filtering sql_digest.
- There's a potential risk that table`tidb_runaway_queries` could be flooded if massive queries are identified as runaway queries in a short period. For example, SQL injections. Since this table is informational, we can afford to lose details.
	1. modify flushes the records to KV at an interval of 10 seconds
	2. Provide a record map with hash (resouce_name, sql_digest, plan_digest, match_type). This map will be flushed to tidb_runaway_queries table every 10s. (This is because flush inserts all the records, so the map is added to minimize the number of inserts)
	3. Add an UNIQUE KEY runaway_task(resource_group_name, sql_digest, plan_digest), time represents the timestamp of the first occurrence. repeat represents the times that the message repeats since the last flush.

#### result

```bash
mysql> ALTER RESOURCE GROUP default QUERY_LIMIT=(EXEC_ELAPSED='1s', ACTION=KILL,  WATCH=EXACT DURATION='10m');
Query OK, 0 rows affected (0.22 sec)

// now is empty
mysql> select * from mysql.tidb_runaway_queries limit 3 \G;
Empty set (0.00 sec)

mysql> select sleep(2) from t;
ERROR 8253 (HY000): Query execution was interrupted, identified as runaway query

mysql> select * from mysql.tidb_runaway_queries limit 3 \G;
resource_group_name: default
         start_time: 2024-09-05 11:59:57
            repeats: 1
         match_type: identify
             action: kill
         sample_sql: select sleep(2) from t
         sql_digest: 4adbc838b86c573265d4b39a3979d0a362b5f0336c91c26930c83ab187701a55
        plan_digest: 5d094f78efbce44b2923733b74e1d09233cb446318293492901c5e5d92e27dbc
        tidb_server: 127.0.0.1:4000
1 row in set (0.00 sec)
```

And then execute `select sleep(3) from t;` which has same UNIQUE KEY runaway_task(resource_group_name, sql_digest, plan_digest)  as `select sleep(3) from t;`

```
mysql> select sleep(3) from t;
ERROR 8253 (HY000): Query execution was interrupted, identified as runaway query

// We can find time and repeats have been updated
mysql> select * from mysql.tidb_runaway_queries limit 3 \G;
resource_group_name: default
         start_time: 2024-09-05 11:59:57
            repeats: 2
         match_type: identify
             action: kill
         sample_sql: select sleep(2) from t
         sql_digest: 4adbc838b86c573265d4b39a3979d0a362b5f0336c91c26930c83ab187701a55
        plan_digest: 5d094f78efbce44b2923733b74e1d09233cb446318293492901c5e5d92e27dbc
        tidb_server: 127.0.0.1:4000
```

And then execute `select sleep(2) from t;` which will be regarded as `watch` type rather than `kill`.

```
mysql> select sleep(2) from t;
ERROR 8253 (HY000): Query execution was interrupted, identified as runaway query

// We can find time and repeats have been updated
mysql> select * from mysql.tidb_runaway_queries limit 3 \G;
resource_group_name: default
         start_time: 2024-09-05 11:59:57
            repeats: 2
         match_type: identify
             action: kill
         sample_sql: select sleep(2) from t
         sql_digest: 4adbc838b86c573265d4b39a3979d0a362b5f0336c91c26930c83ab187701a55
        plan_digest: 5d094f78efbce44b2923733b74e1d09233cb446318293492901c5e5d92e27dbc
        tidb_server: 127.0.0.1:4000
*************************** 2. row ***************************
resource_group_name: default
         start_time: 2024-09-05 12:00:13
            repeats: 1
         match_type: watch
             action: kill
         sample_sql: select sleep(2) from t
         sql_digest: 4adbc838b86c573265d4b39a3979d0a362b5f0336c91c26930c83ab187701a55
        plan_digest: 5d094f78efbce44b2923733b74e1d09233cb446318293492901c5e5d92e27dbc
        tidb_server: 127.0.0.1:4000
2 rows in set (0.00 sec)
```

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
runaway: Mitigation  tidb_runaway_queries flooding
```


nolouch (commented):
can u add some tests?

pkg/resourcegroup/runaway/checker.go
```diff
@@ -275,7 +275,7 @@ func (r *Checker) markRunawayByWatch(action rmpb.RunawayAction) {
 func (r *Checker) markRunaway(matchType string, action rmpb.RunawayAction, now *time.Time) {
 	actionStr := strings.ToLower(action.String())
 	metrics.RunawayCheckerCounter.WithLabelValues(r.resourceGroupName, matchType, actionStr).Inc()
-	r.manager.markRunaway(r.resourceGroupName, r.originalSQL, r.planDigest, actionStr, matchType, now)
+	r.manager.markRunaway(r.resourceGroupName, r.originalSQL, r.planDigest, r.sqlDigest, actionStr, matchType, now)
```
JmPotato, line 278: Consider passing `*Checker` directly to reduce the number of parameters.

HuSharp: replaced

pkg/session/bootstrap_test.go
```diff
@@ -270,6 +270,13 @@ func revertVersionAndVariables(t *testing.T, se sessiontypes.Session, ver int) {
 		// for version <= version195, tidb_enable_dist_task should be disabled before upgrade
 		MustExec(t, se, "update mysql.global_variables set variable_value='off' where variable_name='tidb_enable_dist_task'")
 	}
+	if ver < version212 {
```
nolouch, line 273: only for tests?  if in real environment, do we have a way or need a way to revert?

nolouch (commented):
rest lgtm

pkg/session/bootstrap_test.go
HuSharp: Yes, for testing purposes only. We will not encounter this condition in the real world environment. PTAL, thx~ @tangenta

pkg/resourcegroup/runaway/manager.go
```diff
@@ -176,11 +178,21 @@ func (rm *Manager) RunawayRecordFlushLoop() {
...
+			if _, exists := recordMap[key]; exists {
+				recordMap[key].Repeats++
+			} else {
+				recordMap[key] = r
+			}
 			failpoint.Inject("FastRunawayGC", func() {
 				flushRunawayRecords()
 			})
-			if len(records) >= flushThreshold {
+			if len(recordMap) >= flushThreshold {
```
JmPotato, line 195: Should we consider the repeats here? For example, if we have only one key but repeat for `flushThreshold` times.

HuSharp: Yes, this pr has considered this situation.
https://github.com/pingcap/tidb/blob/6b76e24e450273bf9f6fa3c0ca94c5f9f3d06b2c/pkg/resourcegroup/runaway/manager.go#L187-L189
We only update the repeat field of the value, not adding a new record.

> if we have only one key but repeat for flushThreshold times.

So for this situation, only one record in the map

HuSharp: make sense, nice catch!

HuSharp: added

JmPotato (approved):

pkg/session/bootstrap.go
```diff
@@ -3090,8 +3096,26 @@ func upgradeToVer212(s sessiontypes.Session, ver int64) {
...
+	// 2. modify column `plan_digest` type, modify column `time` to `start_time,
+	// modify column `original_sql` to `sample_sql` and unique union key to `mysql.tidb_runaway_queries`.
+	// add column `sql_digest`.
+	doReentrantDDL(s, "ALTER TABLE mysql.tidb_runaway_queries ADD COLUMN `sql_digest` varchar(64) DEFAULT '' AFTER `original_sql`;", infoschema.ErrColumnExists)
+	// add column `repeats`.
+	doReentrantDDL(s, "ALTER TABLE mysql.tidb_runaway_queries ADD COLUMN `repeats` int DEFAULT 0 AFTER `time`;", infoschema.ErrColumnExists)
+	// modify column `time` to `start_time` and index(time).
+	doReentrantDDL(s, "ALTER TABLE mysql.tidb_runaway_queries CHANGE COLUMN `time` `start_time` TIMESTAMP NOT NULL", infoschema.ErrColumnExists)
+	doReentrantDDL(s, "ALTER TABLE mysql.tidb_runaway_queries DROP INDEX `time_index`", dbterror.ErrCantDropFieldOrKey)
+	doReentrantDDL(s, "ALTER TABLE mysql.tidb_runaway_queries ADD INDEX time_index(start_time) COMMENT 'accelerate the speed when querying with active watch'", dbterror.ErrDupKeyName)
```
glorv, line 3112: Since add index can be very slow if the table is large, should this kind of actions should be done in bootstrap? /cc @wjhuang2016

pkg/session/bootstrap.go
```diff
@@ -3090,8 +3096,26 @@ func upgradeToVer212(s sessiontypes.Session, ver int64) {
...
+	// modify column `original_sql` to `sample_sql`.
+	doReentrantDDL(s, "ALTER TABLE mysql.tidb_runaway_queries CHANGE COLUMN `original_sql` `sample_sql` TEXT NOT NULL", infoschema.ErrColumnExists)
+	// modify column type of `plan_digest`.
+	doReentrantDDL(s, "ALTER TABLE mysql.tidb_runaway_queries MODIFY COLUMN `plan_digest` varchar(64) DEFAULT '';", infoschema.ErrColumnExists)
+	// add union key(resource_group_name, sql_digest, plan_digest, match_type)
```
glorv, line 3117:
```suggestion
	// add unique key(resource_group_name, sql_digest, plan_digest, match_type)
```

pkg/session/bootstrap.go
```diff
@@ -3090,8 +3096,26 @@ func upgradeToVer212(s sessiontypes.Session, ver int64) {
...
+	doReentrantDDL(s, "ALTER TABLE mysql.tidb_runaway_queries ADD UNIQUE KEY runaway_task(resource_group_name, sql_digest, plan_digest, match_type);", dbterror.ErrDupKeyName)
```
glorv, line 3118: Why this key can be `unique`? What if 2 batch of record are flushed within 20s?

HuSharp: > Why this key can be unique?

After discussing with the PM, we can use these params as unique cc @songrijie 

> What if 2 batch of record are flushed within 20s?

When refreshed within 20s, it will be updated by `ON DUPLICATE KEY UPDATE` which means the second batch will overlap previous's record's repeats.
https://github.com/pingcap/tidb/pull/55664/files#diff-d740f8f40d52887b1bee7292178ab85c36b151391a2afce9b58de3803e8d3af0R107
```bash
builder.WriteString(" ON DUPLICATE KEY UPDATE repeats = repeats + 1 + VALUES(repeats)")
```

pkg/session/bootstrap.go
tangenta: For new clusters, this index has been included in `CREATE TABLE` statements.
For old clusters, TiDB only run this upgrading operation once, I think it is acceptable.

pkg/session/bootstrap.go
glorv: Ok. Got it.
Then, another question: if all the runaway records of the same query are flushed into the same row, how do you expect our user to use this row? E.g, how can I know how many new records are caught in the last 1 hour?

tangenta (approved):

pkg/session/bootstrap.go
HuSharp: After discussing with the PM, we will flush every 30 seconds and delete the union key, writing a new record(with compressing and repeats field) each time

pkg/resourcegroup/runaway/manager.go
```diff
@@ -176,11 +178,21 @@ func (rm *Manager) RunawayRecordFlushLoop() {
 			flushRunawayRecords()
 			fired = true
 		case r := <-recordCh:
-			records = append(records, r)
+			key := recordKey{
+				ResourceGroupName: r.ResourceGroupName,
+				SQLDigest:         r.SQLDigest,
+				PlanDigest:        r.PlanDigest,
+				Match:             r.Match,
```
glorv, line 185: Should `Repeats` initialized with `1` so it means how many queries has been caught during that time?

HuSharp: Indeed, maybe 1 is more intuitive, I was wrong about 0 as the default value.

pkg/resourcegroup/runaway/manager.go
HuSharp: After discussing, it is unnecessary to add a repeats check.

glorv (approved):

pkg/session/bootstrap.go
```diff
@@ -3090,8 +3095,24 @@ func upgradeToVer212(s sessiontypes.Session, ver int64) {
...
+	// 2. modify column `plan_digest` type, modify column `time` to `start_time,
+	// modify column `original_sql` to `sample_sql` and unique union key to `mysql.tidb_runaway_queries`.
+	// add column `sql_digest`.
+	doReentrantDDL(s, "ALTER TABLE mysql.tidb_runaway_queries ADD COLUMN `sql_digest` varchar(64) DEFAULT '' AFTER `original_sql`;", infoschema.ErrColumnExists)
+	// add column `repeats`.
+	doReentrantDDL(s, "ALTER TABLE mysql.tidb_runaway_queries ADD COLUMN `repeats` int DEFAULT 1 AFTER `time`;", infoschema.ErrColumnExists)
+	// modify column `time` to `start_time` and index(time).
+	doReentrantDDL(s, "ALTER TABLE mysql.tidb_runaway_queries CHANGE COLUMN `time` `start_time` TIMESTAMP NOT NULL", infoschema.ErrColumnExists)
+	doReentrantDDL(s, "ALTER TABLE mysql.tidb_runaway_queries DROP INDEX `time_index`", dbterror.ErrCantDropFieldOrKey)
+	doReentrantDDL(s, "ALTER TABLE mysql.tidb_runaway_queries ADD INDEX time_index(start_time) COMMENT 'accelerate the speed when querying with active watch'", dbterror.ErrDupKeyName)
```
lcwangchao, line 3111: Do we need this operation?

lcwangchao, line 3111: Why we need to drop index and create again?

HuSharp: changed to `ALTER TABLE [table_name] RENAME COLUMN [old_column_name] TO [new_column_name]; `

we can check index
```bash
mysql> show indexes from mysql.tidb_runaway_queries\G
...
*************************** 2. row ***************************
        Table: tidb_runaway_queries
   Non_unique: 1
     Key_name: time_index
 Seq_in_index: 1
  Column_name: start_time			<-  column name changed
    Collation: A
  Cardinality: 0
     Sub_part: NULL
       Packed: NULL
         Null:
   Index_type: BTREE
      Comment:
Index_comment: accelerate the speed when querying with active watch
      Visible: YES
   Expression: NULL
    Clustered: NO
```

lcwangchao (approved):
LGTM for session part

yudongusa (approved):
LGTM

PR #55691, opened 2024-08-27
planner: introduce hashEquals for expression.Column/collationInfo/fieldType

AilinKid:

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/expression/column.go
+++ b/pkg/expression/column.go
@@ -27,10 +27,15 @@ import (
 	"github.com/pingcap/tidb/pkg/parser/charset"
 	"github.com/pingcap/tidb/pkg/parser/model"
 	"github.com/pingcap/tidb/pkg/parser/mysql"
+	"github.com/pingcap/tidb/pkg/planner/cascades/base"
 	"github.com/pingcap/tidb/pkg/types"
 	"github.com/pingcap/tidb/pkg/util/chunk"
 	"github.com/pingcap/tidb/pkg/util/codec"
 	"github.com/pingcap/tidb/pkg/util/size"
+)
+
+var (
+	_ base.HashEquals = &Column{}
 )
 
 // CorrelatedColumn stands for a column in a correlated sub query.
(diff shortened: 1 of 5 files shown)


pkg/parser/types/field_type.go
```diff
@@ -40,6 +40,21 @@ var (
 	TiDBStrictIntegerDisplayWidth bool
 )
 
+// IHasher is internal usage represent cascades/base.Hasher
+type IHasher interface {
```
AilinKid, line 44: parser pkg sub-mod couldn't import back tidb pkg, which is undesirable, while for base.hasher, we can define the same even anonimous interface with same func signatures

AilinKid: then, when calling from tidb mod, we can directly pass the base.Hasher as it's param here, implicit convertion will be conducted inside.

pkg/expression/column.go
```diff
@@ -442,6 +447,48 @@ func (col *Column) GetStaticType() *types.FieldType {
...
+	case *Column:
+		col2 = x
+	default:
+		return false
+	}
+	return col.RetType.Equal(col2.RetType) &&
+		col.ID == col2.ID &&
+		col.UniqueID == col2.UniqueID &&
+		col.Index == col2.Index &&
+		//col.VirtualExpr.Equals(col2.VirtualExpr) &&
```
hawkingrei, line 483: Is it to keep or remove?

AilinKid: VirtualExpr is scalar expression，which should introduce HashEquals interface, which i haven't done yet

AilinKid: that's why i just comment it for later opening

hawkingrei (approved):

pkg/parser/types/field_type.go
```diff
@@ -61,6 +76,68 @@ type FieldType struct {
...
+	}
+	h.HashInt(len(ft.elemsIsBinaryLit))
+	for _, elem := range ft.elemsIsBinaryLit {
+		h.HashBool(elem)
+	}
+	h.HashBool(ft.array)
+}
+
+// Equals implements the cascades/base.Hasher.<1th> interface.
+func (ft *FieldType) Equals(other any) bool {
```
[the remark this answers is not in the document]
AilinKid: its same for type downcast
say logicalProjection has one children stored as [child:logicalPlan]
when we want to check whether proj.child is equal to the other, we need to switch case to tell what it's children actual type is.

the type check is unavoidable, it depends on how you are ganna implement it

tangenta (approved):

pkg/parser/types/field_type.go
```diff
@@ -61,6 +76,68 @@ type FieldType struct {
...
+		ft.collate == ft2.collate &&
+		ft.array == ft2.array
+	if !ok {
+		return false
+	}
+	if len(ft.elems) != len(ft2.elems) {
+		return false
+	}
+	for i, one := range ft.elems {
+		if one != ft2.elems[i] {
```
windtalker, line 126: do we need to consider collation here?

windtalker: but if you update the `hash64` method to consider the collation, it is possible that under ci collation, two type with "a" and "A" is equal and has the same hash6. So the problem is, by you original define and purpose, enum should we define type("a","b") euqal to enum type("A","B")  when collation is ci?

AilinKid: “by you original definition and purpose, should we define enum type("a","b") equals to enum type("A","B") when collation is ci?” 
this is not in consideration here, high-level equivalence is guaranteed by datum-compare liked infra, hash64 is based from the basic representation of data structure.

windtalker (approved):
lgtm

Issue #55725, opened 2024-08-28
query information_schema.ddl_jobs is very slow


3AceShowHand:
## Feature Request

**Is your feature request related to a problem? Please describe:**

We'd like to query the tidb to know the ddl job status, by running the following query, it takes more than 20 seconds return the result, it's too slow.

```
SELECT JOB_ID, JOB_TYPE, SCHEMA_STATE, SCHEMA_ID, TABLE_ID, STATE, QUERY
FROM information_schema.ddl_jobs
WHERE DB_NAME = "%s" 
    AND TABLE_NAME = "%s"
    AND JOB_TYPE LIKE "add index%%"
    AND (STATE = "running" OR STATE = "queueing")
LIMIT 1;
```

**Describe the feature you'd like:**

The query above should return result very quick.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

3AceShowHand:
ref https://github.com/pingcap/tiflow/issues/11521

joechenrh:
Maybe the root cause is still #52905? Since we cannot use job_id to skip delete kvs.

PR #55748, opened 2024-08-29
*: apply prefiltering before `json.Unmarshal` for information_schema.ddl_jobs table

joechenrh:

Issue Number: ref #55725

Problem Summary:

Querying `information_schema.ddl_jobs` table mainly includes these steps:

1. Scan the whole `tidb_ddl_job` table to get all running job metas.
2. Get all history job metas from tikv.
3. Unmarshal all metas obtained from the first two steps.
4. Finally, all data will be filtered by `PhysicalSelection`.

This process may arise two problems:

1. If there were lots of jobs finished right before this query, scanning `tidb_ddl_job` table which contains lots of deleted kvs might be a little bit slower. (Maybe about several hundreds of milliseconds)
2. All ddl job metas must be unmarshalled before filtering.

This PR tries to solve the second problem, i.e.

1. Do some filtering before unmarshalling the whole structure.
2. Use state to determine whether to query running/historical ddl jobs.

I tested it on my local machine with 100k DDL jobs (create 100k tables randomly in 32 databases)

Before:

```mysql
mysql> select count(*) from information_schema.ddl_jobs where db_name = "d1";
+----------+
| count(*) |
+----------+
|     3170 |
+----------+
1 row in set (4.64 sec)

mysql> select count(*) from information_schema.ddl_jobs where table_name = "t1";
+----------+
| count(*) |
+----------+
|        4 |
+----------+
1 row in set (4.54 sec)
mysql> select count(*) from information_schema.ddl_jobs where state = "running";
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (4.23 sec)
```

After
```mysql
mysql> select count(*) from information_schema.ddl_jobs;
+----------+
| count(*) |
+----------+
|   100109 |
+----------+
1 row in set (4.36 sec)

mysql> select count(*) from information_schema.ddl_jobs where db_name = "d3";
+----------+
| count(*) |
+----------+
|     3157 |
+----------+
1 row in set (1.14 sec)

mysql> select count(*) from information_schema.ddl_jobs where table_name = "t43";
+----------+
| count(*) |
+----------+
|        1 |
+----------+
1 row in set (1.08 sec)

mysql> select count(*) from information_schema.ddl_jobs where state = "running"; -- equivalent to query tidb_ddl_job
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.01 sec)
```

Below is the pref result of  SQL: `select count(*) from information_schema.ddl_jobs where db_name = "d3";` 

Screenshot 2024-09-02 at 15 15 39

An updated results with 160k history jobs:

Before and after

```mysql
mysql> select count(*) from information_schema.ddl_jobs where state = "synced" and db_name = "d10";
+----------+
| count(*) |
+----------+
|     3172 |
+----------+
1 row in set (13.00 sec)

mysql> select count(*) from information_schema.ddl_jobs where state = "synced" and db_name = "d10";                                                                                        
+----------+                                                                                                                                                                               
| count(*) |                                                                                                                                                                               
+----------+                                                                                                                                                                               
|     3172 |                                                                                                                                                                               
+----------+                                                                                                                                                                               
1 row in set (2.66 sec)  
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/ddl/ddl_history.go
```diff
@@ -123,6 +124,15 @@ func GetLastHistoryDDLJobsIterator(m *meta.Meta) (meta.LastJobIterator, error) {
 	return m.GetLastHistoryDDLJobsIterator()
 }
 
+// GetLastHistoryDDLJobsIteratorWithFilter gets latest N history DDL jobs iterator.
```
lance6716, line 127: need to update comment

pkg/planner/core/memtable_infoschema_extractor.go
```diff
@@ -260,6 +262,52 @@ func NewInfoSchemaTablesExtractor() *InfoSchemaTablesExtractor {
...
+// So the Selection Operator is still needed.
+func (e *InfoSchemaDDLExtractor) Extract(
+	ctx base.PlanContext,
+	schema *expression.Schema,
+	names []*types.FieldName,
+	predicates []expression.Expression,
+) (remained []expression.Expression) {
+	c := predicates
+	e.InfoSchemaBaseExtractor.Extract(ctx, schema, names, predicates)
+	return c
```
lance6716, line 284:
```suggestion
	return predicates
```

pkg/planner/core/memtable_infoschema_extractor.go
```diff
@@ -260,6 +262,52 @@ func NewInfoSchemaTablesExtractor() *InfoSchemaTablesExtractor {
...
+	for _, s := range schemas {
+		if n, ok := is.SchemaByName(s); ok {
+			ret = append(ret, n.ID)
+		}
+	}
+	return ret, len(schemas) > 0
+}
+
+// NewInfoSchemaDDLExtractor creates a new InfoSchemaDDLExtractor.
+func NewInfoSchemaDDLExtractor() *InfoSchemaDDLExtractor {
```
lance6716, line 301: I slightly prefer the NewXXX function be put right after the declaration of XXX

pkg/executor/executor.go
```diff
@@ -410,19 +411,65 @@ type DDLJobRetriever struct {
...
+		if sIDs, hasPredicates := ex.ListSchemaIDs(e.is); len(sIDs) > 0 {
+			var builder strings.Builder
+			for _, dbid := range sIDs {
+				schemaIDSet.Insert(dbid)
+				builder.WriteString(fmt.Sprintf("schema_ids LIKE '%%%d%%' OR ", dbid))
+			}
+			schemaFilter = builder.String()
+			schemaFilter = schemaFilter[:len(schemaFilter)-4]
+		} else if hasPredicates {
+			// This means all schemas in predicates are not found, just skip
```
lance6716, line 449: Does it mean the schemas are dropped, or not yet created? So we should not skip them

pkg/meta/meta.go
```diff
@@ -1546,6 +1547,21 @@ func (m *Meta) GetLastHistoryDDLJobsIterator() (LastJobIterator, error) {
 	}, nil
 }
 
+// GetLastHistoryDDLJobsIteratorWithFilter gets latest N history ddl jobs iterator.
+func (m *Meta) GetLastHistoryDDLJobsIteratorWithFilter(
+	schemaIDs set.Int64Set,
+	tableNames set.StringSet) (LastJobIterator, error) {
```
lance6716, line 1553:
```suggestion
	tableNames set.StringSet，
) (LastJobIterator, error) {
```

pkg/meta/meta.go
```diff
@@ -1560,7 +1576,50 @@ func (m *Meta) GetHistoryDDLJobsIterator(startJobID int64) (LastJobIterator, err
...
+// extractSchemaIDAndTableName extract schema_id and table_id from encoded Job structure
+// Note, here we strongly rely on the order of fields in marshalled string, just like checkSubstringsInOrder
+func extractSchemaIDAndTableName(s string) (int64, string, error) {
+	pos := strings.Index(s, `"schema_id":`)
+	if pos == -1 {
+		return 0, "", fmt.Errorf("schema_id not found in model.Job json")
+	}
+
+	start := pos + len(`"schema_id":`)
+	substr := s[start:]
```
lance6716, line 1593: I'm not sure if we should use `[]byte` rather than `string` type because `string` type may create another instance after slicing.

pkg/meta/meta.go
```diff
@@ -1560,7 +1576,50 @@ func (m *Meta) GetHistoryDDLJobsIterator(startJobID int64) (LastJobIterator, err
...
+	}
+
+	pos = strings.Index(substr, `"table_name":`)
+	if pos == -1 {
+		return 0, "", fmt.Errorf("table_name not found in model.Job json")
+	}
+
+	start = pos + len(`"table_name":`)
+	substr = substr[start:]
+	end = strings.Index(substr, ",")
```
lance6716, line 1612: the table name may contain `,`. Please add unit test to cover these corner cases:

- table name is `` `t1,2` ``
- `schema_id` and `table_name` appears in column names, table comments and many places

pkg/meta/meta.go
```diff
@@ -1560,7 +1576,50 @@ func (m *Meta) GetHistoryDDLJobsIterator(startJobID int64) (LastJobIterator, err
...
+	if end == -1 {
+		end = len(substr)
+	}
+
+	tableName := strings.TrimSpace(substr[:end])
+	if strings.HasPrefix(tableName, `"`) && strings.HasSuffix(tableName, `"`) {
```
lance6716, line 1618: Does it always have `"` because it's a JSON string? we don't need to check it.

pkg/meta/meta.go
lance6716: seems OK https://stackoverflow.com/questions/52395730/does-slice-of-string-perform-copy-of-underlying-data

pkg/executor/executor.go
joechenrh: You are right, for `create table` and `create database`, the corresponding info of schema/table cannot be found in infoschema. Maybe we can't apply any filtering on running job table.

pkg/meta/meta.go
joechenrh: - I wonder when will the `table_name` contains comma, just curious.
- Seems like `table_name` and `schema_name` only appear once. 
https://github.com/pingcap/tidb/blob/eaa75a81d56471b602efa3f4ee19bc5345f2cdff/pkg/parser/model/ddl.go#L490-L501
BTW, I'll add tests to cover these cases.

pkg/meta/meta.go
joechenrh: I'll check it and add some tests too.

pkg/meta/meta.go
lance6716: > I wonder when will the table_name contains comma, just curious.

It's decided bu the user. Maybe user will create such table.

> Seems like table_name and schema_name only appear once.

It mean appears in column names, table comments and many places. For example, `CREATE TABLE ... COMMENT 'table_name'`

pkg/executor/executor.go
```diff
@@ -410,19 +411,58 @@ type DDLJobRetriever struct {
...
+		e.runningJobs, err = ddl.GetNeededDDLJobs(sess, schemaNames, tableNames)
+		if err != nil {
+			return err
+		}
 	}
-	e.runningJobs = jobs
+
+	if !skipHistoryJobs {
+		// For the similar reason, we can only use schema_name and table_name to do filtering here.
+		m := meta.NewMeta(txn)
```
lance6716, line 459: I'm not sure we should provide a consistent snapshot of `e.runningJobs` and `e.historyJobIter` or let caller to resolve the duplication. Because a job may get finished after line 451

pkg/meta/meta.go
```diff
@@ -1560,7 +1578,71 @@ func (m *Meta) GetHistoryDDLJobsIterator(startJobID int64) (LastJobIterator, err
...
-	iter *structure.ReverseHashIterator
+	iter        *structure.ReverseHashIterator
+	schemaNames set.StringSet
+	tableNames  set.StringSet
+}
+
+func findNextNonEscapedQuote(s string) int {
+	findEscape := false
+	for i, ch := range s {
+		if ch == '\\' {
```
lance6716, line 1589: for two consecutive `\`, the second one should not trigger `findEscape` (they are one escaped `\` in JSON string). Please add this test case.

pkg/meta/meta.go
```diff
@@ -1571,8 +1653,21 @@ func (i *HLastJobIterator) GetLastJobs(num int, jobs []*model.Job) ([]*model.Job
...
+			err := iter.Next()
+			if err != nil {
+				return nil, errors.Trace(err)
+			}
+			continue
+		}
+
 		job := &model.Job{}
-		err := job.Decode(iter.Value())
+		err = job.Decode(iter.Value())
```
lance6716, line 1670: The filter only skip the `job.Decode`. Is this the bottleneck in original problem? Please run a simple benchmark, I'm not sure it's worthy considering we are going to add many complex codes.

joechenrh: I only did a test on my local machine with few ddl jobs(around 100k). I'll test it with more tables on TCMS later.

lance6716: I think 100k is enough. You can paste the result in PR description. For different filtering how much it will improve the performance.

joechenrh: Since I only used 8 goroutine to create tables concurrently, the number of records in `tidb_ddl_jobs` is quite small. So I mainly tested the performance of querying history ddl jobs. I'll paste it later.

pkg/executor/executor.go
joechenrh: If we want to query running job only, this doesn't matter?

lance6716: If the user queries both running job and finished job, it may appear twice.

pkg/meta/meta.go
lance6716: Please check the performance of `where state = "synced" and db_name = "xxx"`. If there's not a huge improvement, I prefer we delete these optimization based on ExtractSchemaAndTableNameFromJob

joechenrh: In the above tests using 100k jobs , the state of all jobs are "synced".（Please ignore the following differences. Because I tried to create more jobs after the test and I terminated the program, leading to 4 cancelled jobs.)
```mysql
mysql> select count(*) from information_schema.ddl_jobs;
+----------+
| count(*) |
+----------+
|   164156 |
+----------+
1 row in set (12.29 sec)

mysql> select count(*) from information_schema.ddl_jobs where state = "synced";
+----------+
| count(*) |
+----------+
|   164152 |
+----------+
1 row in set (9.59 sec)
```

lance6716: I mean check the effectiveness of filtering DB name and table name on history job, like run `... where state = "synced" and db_name = "any_db"` on old code and new code.

pkg/executor/executor.go
joechenrh: If "Remove job from tidb_ddl_job` and "Add job info into history jobs` is not atomic, this problem would appear. IMO, it's not related to this optimization? 
If we want to solve this problem, maybe we need to rewrite the query itself, for example, use `distinct`.

lance6716: Oh I see that `sess` and `txn` are at the same snapshot, there's no problem. I was thinking they are two independent variables.

pkg/ddl/ddl_history.go
lance6716: > gets latest N history DDL jobs iterator.

There's no `N`?

pkg/meta/meta.go
```diff
@@ -1584,6 +1585,23 @@ func (m *Meta) GetLastHistoryDDLJobsIterator() (LastJobIterator, error) {
 	}, nil
 }
 
+// GetLastHistoryDDLJobsIteratorWithFilter gets latest N history ddl jobs iterator
```
lance6716, line 1588: ditto

lance6716 (approved):

CbcWestwolf (approved):

pkg/executor/executor.go
```diff
@@ -410,19 +411,57 @@ type DDLJobRetriever struct {
...
-		return err
+
+	var err error
+
+	if !skipRunningJobs {
+		// We cannot use table_id and schema_id to construct predicates for the tidb_ddl_job table.
+		// For instance, in the case of the SQL like `create table t(id int)`,
+		// the tableInfo for 't' will not be available in the infoschema until the job is completed.
+		// As a result, we cannot retrieve its table_id.
+		e.runningJobs, err = ddl.GetAllDDLJobs(sess)
```
AilinKid, line 450: we can think about how to make this part for streaming, like not draining the result at first, cause we don't know whether there is jobNum limit outside as we say.

the below historyJobIter is a quite good example

pkg/executor/infoschema_reader_test.go
```diff
@@ -839,6 +842,77 @@ func TestSameTableNameInTwoSchemas(t *testing.T) {
...
+		tk.MustExec(fmt.Sprintf("create database d%d", i))
+		tk.MustExec(fmt.Sprintf("use d%d", i))
+		for j := 0; j < 4; j++ {
+			tk.MustExec(fmt.Sprintf("create table t%d(id int, col1 int, col2 int)", j))
+			tk.MustExec(fmt.Sprintf("alter table t%d add index (col1)", j))
+		}
+	}
+
+	tk2 := testkit.NewTestKit(t, store)
+	tk2.MustQuery(`SELECT JOB_ID, JOB_TYPE, SCHEMA_STATE, SCHEMA_ID, TABLE_ID, table_name, STATE
```
AilinKid, line 859: better add a explain case like
img_v3_02ee_4f9e19c0-b315-4630-833c-13f9cf9391bg

pkg/meta/meta.go
```diff
@@ -1599,7 +1617,50 @@ func (m *Meta) GetHistoryDDLJobsIterator(startJobID int64) (LastJobIterator, err
...
+	m, err := partialjson.ExtractTopLevelMembers(data, jobExtractFields)
+
+	schemaNameToken, ok := m["schema_name"]
+	if !ok || len(schemaNameToken) != 1 {
+		return "", "", errors.New("name field not found in JSON")
+	}
+	schemaName, ok = schemaNameToken[0].(string)
+	if !ok {
+		return "", "", errors.Errorf("unexpected name field in JSON, %v", schemaNameToken)
+	}
```
D3Hunter, line 1640: in current usecase, we are always extract top level basic elements, we can abstract this part into a generic method

D3Hunter:
/hold

please unhold after fix comments

D3Hunter:
/unhold

FastUnmarshalTableNameInfo read token directly 😓

pkg/meta/meta.go
```diff
@@ -1599,7 +1617,50 @@ func (m *Meta) GetHistoryDDLJobsIterator(startJobID int64) (LastJobIterator, err
...
+	tableName, ok = tableNameToken[0].(string)
+	if !ok {
+		return "", "", errors.Errorf("unexpected name field in JSON, %v", tableNameToken)
+	}
+	return
+}
+
+// IsJobMatch examines whether given job's table/schema name matches.
+func IsJobMatch(job []byte, schemaNames, tableNames set.StringSet) (match bool, err error) {
+	schemaName, tableName, err := ExtractSchemaAndTableNameFromJob(job)
```
D3Hunter, line 1655: we can return `true` directly when schemaNames and TableNames are empty

PR #55831, opened 2024-09-03
*: Truncate partition with global index improvement

### LINKED ISSUE #55819: Truncate partition with global index blocks inserts also during delete reorganization state
## Enhancement
ALTER TABLE t TRUNCATE PARTITION will block inserts to the truncated partition during both "delete only" and "delete reorganization" regardless if there is a duplicate key or not.

How to test:
```sql
-- client 1
create table t (a int primary key nonclustered global, b int) partition by hash (b) partitions 3;
insert into t values (1,1),(2,2),(3,3),(4,4),(5,5),(6,6);
-- client 2
begin;
select count(*) from t;
-- client 1
alter table t truncate partition p0;
-- client 3
begin;
insert into t values (9,9); -- fails with "ERROR 8210 (HY000): the partition is in not in public"
select * from information_schema.DDL_JOBS where table_name = 't' order by job_id desc limit 1; -- shows "delete only" schema state
-- client 2
rollback;
begin; 
insert into t values (12,12); -- fails with "ERROR 8210 (HY000): the partition is in not in public"
select * from information_schema.DDL_JOBS where table_name = 't' order by job_id desc limit 1; -- shows "delete only" schema state
-- client 3
rollback;
begin;
insert into t values (15,15); -- fails with "ERROR 8210 (HY000): the partition is in not in public"
select * from information_schema.DDL_JOBS where table_name = 't' order by job_id desc limit 1; -- shows "delete reorganization" schema state
-- client 2
rollback;
begin;
insert into t values (18,18);
select * from information_schema.DDL_JOBS where table_name = 't' order by job_id desc limit 1; -- shows "delete reorganization" schema state
-- client 3
rollback;
begin;
insert into t values (21,21); -- Finally succeeds!
select * from information_schema.DDL_JOBS where table_name = 't' order by job_id desc limit 1; -- shows "none" schema state, done state
-- client 2
insert into t values (24,24); -- Still fails with "ERROR 8210 (HY000): the partition is in not in public"
select * from information_schema.DDL_JOBS where table_name = 't' order by job_id desc limit 1; -- shows "none" schema state, done state
rollback; -- now client 1 will finish and the job will be synced
insert into t values (27,27); -- Finally succeeds!
select * from information_schema.DDL_JOBS where table_name = 't' order by job_id desc limit 1; -- shows "none" schema state, synced state
```

I would expect that during "delete reorganize" state, which can take a long time if the partition was big, that inserts/updates/deletes would work without errors (not even duplicate key errors).
During the first schema state "delete only" I would accept that "duplicate key" errors were given.

mjonss:

Issue Number: close #55819

Problem Summary:
During TRUNCATE PARTITION writes was blocked to partitions that was being truncated, also during `StateDeleteReorganization`, meaning the DDL would block writes to all truncated partitions for the full duration of the DDL, in case of a Global Index.

Instead of simply blocking writes to partitions being truncated, we are hiding the new partitions from Global Indexes `StateWriteOnly` and old partitions from `StateDeleteOnly` so that they cannot be seen, and not cause any anomaly, where you can find data or not depending on access method, i.e. it may be found if using Global Index, but not for table scan.

A side effect of this is that we are still blocking duplicate keys, if a write to a global unique index from a `StateDeleteOnly` client that collides with a not-yet cleaned up Global Index entry, or a write to a global unique index from `StateWriteOnly` (old partition, to be truncated) collides with a newly written entry in `StateDeleteOnly` (using the new emptied partition).

To not generate these kind of duplicate errors for too long period, duplicate key errors are not generated during `StateDeleteReorganization` if the duplicate comes from an old truncated partition, instead that global unique index entry is overwritten.

Also refactored/deduplicated code for cleaning up the global index entries from dropped/truncated partitions, so both DROP and TRUNCATE PARTITION uses the same code.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Removed write blocking partitions involved in TRUNCATE PARTITIONs if they had a Global Index.
```


pkg/testkit/testkit.go
```diff
@@ -167,6 +167,22 @@ func (tk *TestKit) MustExecWithContext(ctx context.Context, sql string, args ...
 	}
 }
 
+// SkipIfFailpointDisabled will skip the current test if failpoint is not enabled
+func SkipIfFailpointDisabled(t *testing.T) {
```
Defined2014, line 171: Why add this function? I think we always enable failpoint in ci env.

mjonss: When I develop, I want to see the `git diff` which is full of changes during `failpoint-enable` and running tests without it makes them fail, so I rather skip them with a message if I have not enabled the failpoints.

mjonss: Function removed.

winoros (approved):
for planner part

pkg/table/tables/index.go
```diff
@@ -551,6 +583,32 @@ func FetchDuplicatedHandle(ctx context.Context, key kv.Key, distinct bool,
 	return true, nil, nil
 }
 
+// GetPartitionHandleFromVal creates a PartitionHandle from a global index value
+func GetPartitionHandleFromVal(val []byte) (*kv.PartitionHandle, error) {
```
Defined2014, line 587: Why not just use `DecodeHandleInIndexValue` function?

pkg/ddl/partition.go
```diff
@@ -2277,7 +2248,7 @@ func (w *worker) onDropTablePartition(jobCtx *jobContext, job *model.Job) (ver i
 		// Since the previous state do not use the dropping partitions,
 		// we can now actually remove them, allowing to write into the overlapping range
 		// of the higher range partition or LIST default partition.
-		physicalTableIDs = updateDroppingPartitionInfo(tblInfo, partNames)
+		_ = updateDroppingPartitionInfo(tblInfo, partNames)
```
Defined2014, line 2251:
```suggestion
		updateDroppingPartitionInfo(tblInfo, partNames)
```

pkg/ddl/partition.go
```diff
@@ -2266,7 +2237,7 @@ func (w *worker) onDropTablePartition(jobCtx *jobContext, job *model.Job) (ver i
 		// Reason, see https://github.com/pingcap/tidb/issues/55888
 		// Only mark the partitions as to be dropped, so they are not used, but not yet removed.
 		originalDefs := tblInfo.Partition.Definitions
-		physicalTableIDs = updateDroppingPartitionInfo(tblInfo, partNames)
+		_ = updateDroppingPartitionInfo(tblInfo, partNames)
```
Defined2014, line 2240:
```suggestion
		updateDroppingPartitionInfo(tblInfo, partNames)
```

pkg/ddl/partition.go
```diff
@@ -2428,236 +2357,245 @@ func removeTiFlashAvailablePartitionIDs(tblInfo *model.TableInfo, pids []int64)
...
+		return updateVersionAndTableInfo(jobCtx, job, tblInfo, true)
 	case model.StateDeleteOnly:
-		// This state is not a real 'DeleteOnly' state, because tidb does not maintaining the state check in partitionDefinition.
-		// Insert this state to confirm all servers can not see the old partitions when reorg is running,
-		// so that no new data will be inserted into old partitions when reorganizing.
+		// Now we don't see the old partitions, but other sessions may still use them.
+		// So to keep the Global Index consistent, we will still keep it up-to-date with
+		// the old partitions, as well as the new partitions.
+		// Also ensures that no writes will happen after GC in DeleteRanges.
+
```
Defined2014, line 2543: I don't understand why this intermediate state is needed. When L2544 is executed, other TiDBs must have switched to the StateDeleteOnly state. This state seems do nothing here.

mjonss: No, when L2544 is executed, it only means that the current go routine has updated the state/other TiDBs must have switched to the `StateWriteOnly` state. The tableInfo will be written to the KV store and either patched by ApplyDiff or fully reloaded, which will happen asyncronously between different TiDB servers (domains). The goroutine executing this DDL may wait until there are no transactions using the previous schema version (see how MDL is implemented).

But it still means that as soon as updateVersionAndTableInfo + commit in calling function is done, then some TiDB servers can see the new SchemaVersion and some will be using the 'old' SchemaVersion.
And in that case if `StateDeleteOnly` was not used, then `StateWriteOnly` could write to the old set of partitions during `StateDeleteReorganization`, meaning that depending on timing, it may also write to the Global Index, after DeleteReorganization has already cleaned that record range, leaving an orphan entry in the Global index.

pkg/ddl/partition.go
```diff
@@ -2428,236 +2357,245 @@ func removeTiFlashAvailablePartitionIDs(tblInfo *model.TableInfo, pids []int64)
...
+		pi.NewPartitionIDs = nil
+		pi.DDLState = model.StateNone
+		pi.DDLAction = model.ActionNone
+		// Done, continue after the SchemaState switch!
 	default:
-		err = dbterror.ErrInvalidDDLState.GenWithStackByArgs("partition", job.SchemaState)
+		return ver, dbterror.ErrInvalidDDLState.GenWithStackByArgs("partition", job.SchemaState)
 	}
 
+	// used by ApplyDiff in updateSchemaVersion
```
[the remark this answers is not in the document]
mjonss: Ah, I had an optimization for non-global index tables, where it could skip one state, but decided to remove that optimization, to have the same code path regardless of global index or not.
I will move these lines back into `StateDeleteReorganization`

pkg/ddl/partition.go
Defined2014: We have `waitVersionSynced` function which will wait all TiDB schemas are synced. So when  `cleanGlobalIndexEntriesFromDroppedPartitions` execute, all TiDB nodes should in `model.StateDeleteReorganization` state.

mjonss: Ah, thank you for the correction!
After trying to remove the StateDeleteOnly mode, I realized that there is one more reason to keep it:
If going directly from StateWriteOnly (where it would not read the new partitions, but still get duplicate key errors on the Global Index) to StateDeleteReorganization (where it would not read the old partitions, and directly overwrite Global Index entries pointing to old partitions) there would be an anomaly like:
StateWriteOnly session 'wo> '
StateDeleteReorganization session 'dr> '
```mysql
init> create table t (a int primary key, b varchar(255), unique index (`b`) global) partition by hash (a) partitions 2;
wo> insert into t values (1,1,"WO"); -- success
dr> insert into t values (1,1,"DR"); -- success
wo> select * from t where a = 1; -- 1 1 WO
dr> select * from t where a = 1; -- 1 1 DR
```
Regardless of global index or not, the above would happen (with current PR implementation).
We could also remove the optimization for overwriting old global index entries in StateDeleteReorganization, which would then fail the 'dr> ' insert due to duplicated global index entry. But that state/error would potentially exists for a long time, if the truncated partition had a lot of rows.

@Defined2014 I would still propose to keep the StateDeleteOnly where the three states would mean:
- StateWriteOnly
  - Aware of new partitions, so it would filter out those partitions from Global Indexes on read, and fail on duplicated key on write.
- StateDeleteOnly
  - Using the new (initally empty) partitions, but will give duplicate errors on the Global Index if writes collide with entries from old (dropped) partitions, since those entries may still be used by sessions in StateWriteOnly.
  - This state will only take as long as it takes for updating the Schema Version across a cluster, so limited time not depending on how many rows the dropped partitions has.
- StateDeleteReorganization
  - Using the new partitions, and if there is a duplicate Global Index entry for an old partition, it will simply overwrite it, since those entries are no-longer used.

What do you think?
  - During this state all Global Index entries pointing to old partitions will be deleted by a reorg DDL worker.

Defined2014: There is another small question. If the `StateDeleteOnly` state is used, will the above situation still occur for the unique key of non-global index?

pkg/ddl/partition.go
```diff
@@ -2428,236 +2357,245 @@ func removeTiFlashAvailablePartitionIDs(tblInfo *model.TableInfo, pids []int64)
...
-		ver, err = updateVersionAndTableInfo(jobCtx, job, tblInfo, true)
+		return updateVersionAndTableInfo(jobCtx, job, tblInfo, true)
 	case model.StateDeleteOnly:
-		// This state is not a real 'DeleteOnly' state, because tidb does not maintaining the state check in partitionDefinition.
-		// Insert this state to confirm all servers can not see the old partitions when reorg is running,
-		// so that no new data will be inserted into old partitions when reorganizing.
+		// Now we don't see the old partitions, but other sessions may still use them.
+		// So to keep the Global Index consistent, we will still keep it up-to-date with
+		// the old partitions, as well as the new partitions.
+		// Also ensures that no writes will happen after GC in DeleteRanges.
```
Defined2014, line 2542: Could add a comment point to `IDsInDDLToIgnore` function.

Defined2014 (approved):
LGTM, please move some lines back into `StateDeleteReorganization`

pkg/ddl/partition.go
mjonss: Yes, non-global indexes (local indexes) will be able to have 'non-visible' duplicates, since two sessions can have different sets of partitions between `StateWriteOnly` and `StateDeleteOnly`. But it is not really related to the specific state, just that the transition between having the old set of partitions visible, to have the new set of partitions visible. A less user friendly way would be to either have a state where the truncated partitions are completely blocked or check both old and new partitions on write (which is kind of what Global Index does).

pkg/ddl/partition.go
mjonss: Done.

pkg/ddl/partition.go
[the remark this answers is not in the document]
mjonss: Done.

pkg/ddl/partition.go
```diff
@@ -2061,7 +2061,6 @@ func updateDroppingPartitionInfo(tblInfo *model.TableInfo, partLowerNames []stri
 
 	tblInfo.Partition.Definitions = newDefs
 	tblInfo.Partition.DroppingDefinitions = droppingDefs
-	return pids
```
tangenta, line 2064: Maybe remove all the usages of `pids` as well

pkg/ddl/partition.go
mjonss: Actually done in the follow up pr, #56029.

tangenta (approved):

PR #55842, opened 2024-09-04
planner: fix incorrect estRows with global index and json column

hawkingrei:

Issue Number: close https://github.com/pingcap/tidb/issues/55818

Problem Summary:

Because the JSON column will be without stats by default, the row count will be 0. But when calculating selectivity, we cannot divide by zero. it will get NaN row count.

image

#### why is the result wrong?

image

It will return ```NaN``` from ```Selectivity``` and break all stats info by ```ds.TableStats.Scale(selectivity)```

image

image

Final, The problem will be created at the ```joinReorderGreedySolver.solve```-> ```constructConnectedJoinTree```

image

```checkConnectionAndMakeJoin``` will create a right join plan. but when to calculate the cost, it will get ```NaN``` because of the pollute stats info. we cannot get best join because of ```Nan > math.MaxFloat64```. ```makeBushyJoin``` will not consider the ```EqCondtion```. Many condition will lose. This is the reason why it will get wrong result.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test

Due to incorrect statistics, the filter condition was completely removed.

before

```


+---------------------------+---------+-----------+-----------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| id                        | estRows | task      | access object   | operator info                                                                                                                                                                            |
+---------------------------+---------+-----------+-----------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Projection_12             | NaN     | root      |                 | 1->Column#18                                                                                                                                                                             |
| └─HashJoin_13             | NaN     | root      |                 | CARTESIAN inner join                                                                                                                                                                     |
|   ├─TableReader_19(Build) | 38.00   | root      |                 | data:TableFullScan_18                                                                                                                                                                    |
|   │ └─TableFullScan_18    | 38.00   | cop[tikv] | table:t61a85298 | keep order:false, stats:pseudo                                                                                                                                                           |
|   └─TableReader_17(Probe) | NaN     | root      | partition:all   | data:Selection_16                                                                                                                                                                        |
|     └─Selection_16        | NaN     | cop[tikv] |                 | or(json_memberof(cast(16739493649928310215, json BINARY), test.tceb7972c.col_17), not(istrue_with_null(json_contains(test.tceb7972c.col_17, cast("6019730272580550835", json BINARY))))) |
|       └─TableFullScan_15  | 12.00   | cop[tikv] | table:tceb7972c | keep order:false, stats:partial[col_17:unInitialized]                                                                                                                                    |
+---------------------------+---------+-----------+-----------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+

```

after 

```
id  estRows task  access object operator info
Projection_11 6.00  root    1->Column#18
└─HashJoin_13 6.00  root    inner join, equal:[eq(test.tceb7972c.col_19, Column#19)]
  ├─TableReader_16(Build) 4.80  root  partition:all data:Selection_15
  │ └─Selection_15  4.80  cop[tikv]   or(json_memberof(cast(16739493649928310215, json BINARY), test.tceb7972c.col_17), not(istrue_with_null(json_contains(test.tceb7972c.col_17, cast("6019730272580550835", json BINARY)))))
  │   └─TableFullScan_14  6.00  cop[tikv] table:tceb7972c keep order:false, stats:partial[col_17:missing]
  └─Projection_17(Probe)  10000.00  root    cast(test.t61a85298.col_71, double BINARY)->Column#19
    └─TableReader_19  10000.00  root    data:TableFullScan_18
      └─TableFullScan_18  10000.00  cop[tikv] table:t61a85298 keep order:false, stats:pseudo
```

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/cardinality/selectivity.go
```diff
@@ -512,6 +512,9 @@ func CalcTotalSelectivityForMVIdxPath(
 				realtimeCount, _ = coll.GetScaledRealtimeAndModifyCnt(coll.GetIdx(path.Index.ID))
 			}
 		}
+		if realtimeCount == 0 {
+			realtimeCount = 1
```
qw4990, line 516: Please add some comments as well

qw4990 (approved):

pkg/planner/cardinality/selectivity.go
hawkingrei: has been updated.

time-and-fate (commented):
I think the root cause here is that the global index (`idx_10` and `idx_9`) here is not analyzed actually, but they are marked as analyzed (according to `ColAndIdxExistenceMap`) and fully loaded (according to `StatsLoadedStatus`).
So we will try to use this here, then the `idxStats.TotalRowCount()` is 0 in `GetScaledRealtimeAndModifyCnt()`, causing the problem.

time-and-fate:
And I think we can't close the issue in this PR, right? Since the wrong result is not fixed.

hawkingrei:
> And I think we can't close the issue in this PR, right? Since the wrong result is not fixed.

The wrong result has been fixed. 

BTW, ```col_17``` is the JSON column. We cannot analyze it by default. so the index is related with ```col_17``` will be without stats.

hawkingrei:
> MV index has stats, but `idx_10` and `idx_9` don't have stats here. It's likely because they are global index. I think the root cause is still that the not analyzed stats are marked analyzed and fully loaded. For a temporary fix, I think we can add a check for `idxStats.TotalRowCount() == 0` in `GetScaledRealtimeAndModifyCnt()`.

I have updated it again.

time-and-fate (approved):
Please restore `tidb_enable_global_index` after the tests.

Issue #54906, opened 2024-07-25
full load doesn't load non-public table


wjhuang2016:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

loadTableInfo will load non-public tables which fetchSchemasWithTables doesn't. Which is inconsistent.

### 2. What did you expect to see? (Required)
N/A
### 3. What did you see instead (Required)
N/A
### 4. What is your TiDB version? (Required)
master

PR #55891, opened 2024-09-06
ddl, planner: avoid running queries on the non-Public state table

### LINKED ISSUE #55813: FK constraint might not enforced when child table is in state write_only
## Bug Report

use the case in the design doc: https://github.com/pingcap/tidb/blob/b71f6c33d69ff8884d4ef3d15f9db485fba9926f/docs/design/2022-06-22-foreign-key.md#L104-L115

steps:
1. apply this diff to code of v8.1.0, after that we will wait when create child table and it's in write-only state, and on 4001 node, we skip load schema diff of child table, so we can make a case that 4000 and 4001 have different version
```diff
diff --git a/pkg/ddl/ddl_worker.go b/pkg/ddl/ddl_worker.go
index 916458f1d0..1dc34f79f0 100644
--- a/pkg/ddl/ddl_worker.go
+++ b/pkg/ddl/ddl_worker.go
@@ -1430,6 +1430,9 @@ func waitSchemaChanged(d *ddlCtx, waitTime time.Duration, latestSchemaVersion in
                         return nil
                 }
         }
+        if job.TableName == "child" {
+                time.Sleep(time.Hour)
+        }

         return checkAllVersions(d, job, latestSchemaVersion, timeStart)
 }
diff --git a/pkg/infoschema/builder.go b/pkg/infoschema/builder.go
index d03b07d7d1..62347c888e 100644
--- a/pkg/infoschema/builder.go
+++ b/pkg/infoschema/builder.go
@@ -659,6 +659,9 @@ func applyCreateTable(b *Builder, m *meta.Meta, dbInfo *model.DBInfo, tableID in
                         fmt.Sprintf("(Table ID %d)", tableID),
                 )
         }
+        if tblInfo.Name.L == "child" && config.GetGlobalConfig().Port == 4001 {
+                return []int64{}, nil
+        }

         b.updateBundleForCreateTable(tblInfo, tp)
```
2. start 2 tidb on port 4000 and 4001
3. connect 4000, and run below, it will block on create `child`, check `admin show ddl jobs 1` to make sure child is in write_only state.
```sql
CREATE TABLE parent (
    id INT KEY
);

insert into parent values(1)

CREATE TABLE child (
    id INT,
    pid INT,
    INDEX idx_pid (pid),
    FOREIGN KEY (pid) REFERENCES parent(id) ON DELETE CASCADE
);
```
3. start 2 sessions, and run the case in the design doc

|connect 4000 session| 4001 session|
|---|---|
|begin||
|insert into child values(1,1)||
||begin|
||delete from parent where id=1; // it blocks until prev txn commit|
|commit||
||commit|
|select * from child; // we can see 1 row||
||select * from parent;// empty result|

FK constraint enforced
it's not
8.1.0

YangKeao:

Issue Number: close #55813

Problem Summary:

When the table is in `WriteOnly` state (and hasn't been synchronized to other tidb instances), it'll cause inconsistency:

|WriteOnly TiDB| None TiDB |
|---|---|
|begin||
|insert into child values(1,1)||
||begin|
||delete from parent where id=1; // it blocks until prev txn commit|
|commit||
||commit|
|select * from child; // we can see 1 row||
||select * from parent;// empty result|

Make the table invisible when it's in `WriteOnly/DeleteOnly` state. Previously, the users can see/write/delete any data for any table without considering the state. I think it's because the key range of the table is not affected by newly written things, so the `DROP TABLE` always has a good behavior.

In conclusion, the behavior of foreign key in ` `WriteOnly` (or `DeleteOnly` is that:

1. The table is not visible for any explicit DML (or any query? Maybe only DML is enough 🤔, but make it always invisible is clearer).
2. The table will be modified by implicit modification caused by ON DELETE CASCADE or ON UPDATE CASCADE

For more detail, see #55813

This PR also modified the state from `WriteOnly` to `DeleteOnly`. Both state have the same effect, but using `DeleteOnly` is more suitable considering the semantic.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

TODO

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


YangKeao:
/hold

I'm doing some manual test.

YangKeao:
Another concern is that: when the table is in `WriteOnly` state, should it show in the `show tables`? The current behavior is that it'll show in the `show tables`, and I didn't change the behavior in this PR.

pkg/planner/core/preprocess.go
```diff
@@ -1634,6 +1634,10 @@ func (p *preprocessor) handleTableName(tn *ast.TableName) {
 			return
 		}
 	}
+	if err := p.checkTableVisible(tableInfo); err != nil {
```
crazycs520, line 1637: I am not sure about this modification, please ask the DDL owner to take a look.

REST LGTM.

pkg/infoschema/builder.go
```diff
@@ -341,6 +341,18 @@ func (b *Builder) getTableIDs(diff *model.SchemaDiff) (oldTableID, newTableID in
 		newTableID = diff.TableID
 	case model.ActionDropTable, model.ActionDropView, model.ActionDropSequence:
 		oldTableID = diff.TableID
+
+		// Still keep the table in infoschema until when the state of table reaches StateNone. This is because
+		// the `ON DELETE/UPDATE CASCADE` function of foreign key may need to find the table in infoschema. Not
+		// only the table with foreign keys are stored in the infoschema to keep a consistent behavior for all
+		// tables.
+		tblInfo, err := m.GetTable(diff.SchemaID, oldTableID)
```
YangKeao, line 349: Here is the key change to avoid removing `WriteOnly` table from the infoschema.

wjhuang2016 (approved):

pkg/executor/infoschema_reader_test.go
```diff
@@ -1066,3 +1066,43 @@ func TestInfoschemaTablesSpecialOptimizationCovered(t *testing.T) {
...
+
+func testInfoschemaTablesWhenTableIsNotPublic(t *testing.T, enableV2 bool) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	if enableV2 {
+		tk.MustExec("set @@global.tidb_schema_cache_size = 512 * 1024 * 1024")
+	} else {
+		tk.MustExec("set @@global.tidb_schema_cache_size = 0")
+	}
```
tiancaiamao, line 1078: This is unnecessary now?  We 70% chance run v2 and 30% chance v1 automatically (in testkit).
So there is no need to do it twice.

pkg/infoschema/builder.go
tiancaiamao: FYI: In the past, there is a bug https://github.com/pingcap/tidb/issues/54906 and the behaviour is:
infoschema full load do not load any non-public table
apply diff load the tables in the intermediate state.

And it's fixed to load all non-publice tables recently.

pkg/infoschema/builder.go
```diff
@@ -341,6 +341,18 @@ func (b *Builder) getTableIDs(diff *model.SchemaDiff) (oldTableID, newTableID in
...
+		if err != nil {
+			return 0, 0, err
```
tiancaiamao, line 351: When will this code branch happen?
If an error is returned, it cause full reload ... and not easy to notice the case.

tiancaiamao: Because everything still works, just slower.

pkg/infoschema/infoschema_v2.go
```diff
@@ -49,6 +49,7 @@ type tableItem struct {
 	tableID       int64
 	schemaVersion int64
 	tomb          bool
+	state         model.SchemaState
```
tiancaiamao, line 52: This should be the value field instead of the key field?

tiancaiamao: The non-public state is also a schema version ...

pkg/infoschema/infoschema_v2_test.go
```diff
@@ -657,15 +657,34 @@ func TestDataStructFieldsCorrectnessInSchemaChange(t *testing.T) {
...
+	require.False(t, tblItem.tomb)
+	tblItem, ok = v2.Data.byID.Get(tableItem{tableID: tblInfo.ID, schemaVersion: 5})
+	require.True(t, ok)
+	require.False(t, tblItem.tomb)
+	_, ok = v2.Data.tableCache.Get(tableCacheKey{tableID: tblInfo.ID, schemaVersion: 5})
+	require.True(t, ok)
+	require.Equal(t, v2.Data.pid2tid.Len(), 5) // tomb partition info
+	tblInfoItem, ok = v2.Data.pid2tid.Get(partitionItem{partitionID: 1, schemaVersion: 5})
+	require.True(t, ok)
+	require.False(t, tblInfoItem.tomb)
```
tiancaiamao, line 675: Do you mean it's a drop table but the state is not public still, so not tomb?

pkg/executor/infoschema_reader_test.go
YangKeao: It may be treated as a flaky test if it has the opportunity to success :facepalm:. BTW, by default, bazel will retry the failed tests (marked with `flaky = True`) three times, so it has more chances to success.

Therefore, I think it's better to actually run it twice.

pkg/infoschema/infoschema_v2_test.go
YangKeao: Yes.
The expected behavior is that all functions should return all tables in any state (except tables in `None` state), then the callers can filter out the needed tables.

If it's marked as `tomb`, it seems that some functions (like `IterateAllTableItems` and `SchemaSimpleTableInfos`) will not return the table, then it'll be easy for the callers to make mistakes. For example, during `drop schema`, tidb will call `checkDatabaseHasForeignKeyReferred -> is.SchemaSimpleTableInfos` to check whether a database has foreign key referred. If a table is marked `tomb`, it'll give unexpected result.

Though it's actually not a big issue because everything will finally be dropped, it's only an example to show that it's easy to make mistakes / the logic becomes hard to think or understand if it's marked as `tomb` during dropping tables.

pkg/infoschema/infoschema_v2.go
YangKeao: Removed.

pkg/executor/infoschema_reader_test.go
YangKeao: Not needed now.

tiancaiamao (approved):

pkg/planner/core/preprocess.go
```diff
@@ -1839,14 +1839,39 @@ func (p *preprocessor) hasAutoConvertWarning(colDef *ast.ColumnDef) bool {
 	return false
 }
 
-func tryLockMDLAndUpdateSchemaIfNecessary(ctx context.Context, sctx base.PlanContext, dbName pmodel.CIStr, tbl table.Table, is infoschema.InfoSchema) (table.Table, error) {
+func tryLockMDLAndUpdateSchemaIfNecessary(ctx context.Context, sctx base.PlanContext, dbName pmodel.CIStr, tbl table.Table, is infoschema.InfoSchema) (retTbl table.Table, err error) {
+	skipLock := false
+	shouldLockMDL := false
+	var lockedID int64
+
+	defer func() {
```
winoros, line 1847: Why it's defined as a defer instead of directly checking it?

YangKeao: Because we need to check `retTbl`. The only way to check the `retTbl` without modifying every `return` statement is to check it in `defer` :thinking:  .

The `retTbl` can be different with the `tbl` in argument, so we cannot check it only once at the entry of this function.

winoros (approved):

hawkingrei (approved):

AilinKid (approved):

PR #55903, opened 2024-09-06
docs: Add maintaining the priority queue in memory design

Rustin170506:

Issue Number: close None

Problem Summary:

I have added the documentation for maintaining the priority queue in the memory design.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > - [x] I checked and no code files have been changed.
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,556 @@
...
+| -------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------- |
+| Percentage of Change | The percentage of change since the last analysis.                                                                                                                                             | log10(1 + Change Ratio). [Check the graph](https://www.desmos.com/calculator/gb3i2cropz). <br><br> Note: For unanalyzed tables, we set the percentage of changes to 100%                                                                                                                                            | Stats Cache                  |
```
lance6716, line 67: "Change Ratio" and "Percentage of Change" are the same thing?

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,556 @@
...
+| Table Size           | The size is equal to the number of rows multiplied by the number of columns in the table that are subject to analysis. The smaller tables should have a higher priority than the bigger ones. | Applying a logarithmic transformation, namely log10(1 + Table Size), and its 'penalty' is calculated as 1 - log10(1 + Table Size). [Check the graph.](https://www.desmos.com/calculator/x3fq76w8sb)                                                                                                                 | Stats Cache + Table Info     |
```
lance6716, line 68: maybe directly descibe the weight as "1 - log10(1 + Table Size)". 

BTW I guess the final score of a table is sum_for_factors(weight)? Then I think no need to use `1` in this formula, "-log10(1 + Table Size)" should be equivalent.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,556 @@
...
+| Analysis Interval    | Time since the last analysis execution for the table. The bigger interval should have a higher priority than the smaller interval.                                                            | Applying a logarithmic transformation, namely log10(1 + Analysis Interval). To further compress the rate of growth for larger values, we can consider taking the logarithmic square root of x. The final formula is log10(1 + √Analysis Interval). [Check the graph.](https://www.desmos.com/calculator/plhtkfqhx9) | Stats Cache                  |
```
lance6716, line 69: what's the unit of interval? If Analysis Interval >> 1, "log10(1 + √Analysis Interval)" can be approximately equal to "0.5 * log10(Analysis Interval)", it does not compress the rate of growth for larger values well.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,556 @@
...
+2. Because we scan it too frequently, we'll spend a lot of CPU on building this queue, since we'll be scanning all the stats cache and the schema cache. See more: <https://github.com/pingcap/tidb/issues/49972>
+We can try to keep the priority queue in memory to solve the above issues instead of rebuilding it every time.
+
+## Current Implementation
+
+### Calculate the score
+
+In the current implementation, we use a weighted scoring approach to maintain the queue, which is our current scoring criteria.
+
+Weight table:
```
lance6716, line 63: I'm not sure we should call it "weight" or "scroe". Weight means it will be multiplied to another number. But seems we directly add them up?

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,556 @@
+# TiDB Design Documents
+
+- Author(s): [Rustin Liu](http://github.com/Rustin170506), [Lance Tian](http://github.com/lance6716)
+- Discussion PR: <https://github.com/pingcap/tidb/pull/55903>
```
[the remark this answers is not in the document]
Rustin170506: It is suggested by the markdown lint. See https://github.com/updownpress/markdown-lint/blob/master/rules/034-no-bare-urls.md

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,557 @@
...
+For DDL, we need a reliable way to get the DDL events that have occurred and recalculate the score as a result of the change in the table structure.
+
+### Incremental Update
+
+#### DML
+
+Currently, the stats module maintains the most basic information for all tables (modify_count and count) in the stats cache, and we can maintain the priority queue along the lines of incrementally updating it.
+
+1. We start a background worker to call the statsHandle.Update(do.InfoSchema()) function every 3 seconds(stats lease).
+2. In the stats cache struct, we maintain an atomic timestamp called: maxTblStatsVer
```
lance6716, line 97: maybe mention what's the definition of `maxTblStatsVer` here.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,557 @@
...
+3. Whenever we call the statsHandle.Update(do.InfoSchema()) function, we query the new update from incrementally updating.
```
lance6716, line 98: maybe mention that `maxTblStatsVer` is the argument

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: > BTW I guess the final score of a table is sum_for_factors(weight)? Then I think no need to use `1` in this formula, "-log10(1 + Table Size)" should be equivalent.

If I understand correctly, they have the same slope, but you can look at their graphs, they have different value domains, the whole function is shifted up(down) on the image, so it will produce a different result for the formula, it will produce a different result for the same table size, and we want to use the first one, we don't want to penalize very much when the table size is very small.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: The unit is second.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,557 @@
...
+
+    ```sql
+    SELECT version, table_id, modify_count, count from mysql.stats_meta where version > %? order by version
```
lance6716, line 101: maybe to mention that prerequisite property of the `version` column, I think the writer to `version` should be serialized and `version` is monotonic.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: The final formula is:
```go
//	priority_score = (0.6 * math.Log10(1 + ChangeRatio) +
//	                  0.1 * (1 - math.Log10(1 + TableSize)) +
//	                  0.3 * math.Log10(1 + math.Sqrt(AnalysisInterval)) +
//	                  special_event[event])
```

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
lance6716: I have see the final formula `priority_score = ...` in your below comment. If every score minus `0.1` which means use -log10(...), the relative order is not changed.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,557 @@
...
+Because we are constantly adding, removing, and updating the heap, we need to use a thread-safe heap to store analysis jobs.
+The required interfaces look like the following:
+
+| Operation | Time complexity |
+| --------- | --------------- |
+| Add       | O(log n)        |
+| Update    | O(log n)        |
+| Delete    | O(log n)        |
+| Pop       | O(1)            |
+| Get       | O(1)            |
```
lance6716, line 128: Get means GetTop?

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,557 @@
...
+```go
+type Indicators struct {
+        // ChangePercentage is the percentage of the changed rows.
+        // Usually, the more the changed rows, the higher the priority.
+        // It is calculated by modifiedCount / last time analysis count.
+        ChangePercentage float64
+        // TableSize is the table size in rows * len(columns).
+        TableSize float64
+        // LastAnalysisDuration is the duration from the last analysis to now.
+        // In seconds.
```
lance6716, line 141: "In seconds" means 1 second should be store as `time.Duration(1)` rather than `1 * time.Second` which is `1000000000`? Easy to make mistake.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,557 @@
...
+#### Information to be stored
+
+```go
+type Indicators struct {
+        // ChangePercentage is the percentage of the changed rows.
+        // Usually, the more the changed rows, the higher the priority.
+        // It is calculated by modifiedCount / last time analysis count.
+        ChangePercentage float64
+        // TableSize is the table size in rows * len(columns).
+        TableSize float64
```
lance6716, line 139: it's int?

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: Yes, for DML changes, the relative order remains unchanged, but the value domain does change. I guess EventNewIndex will take a higher rank in some cases because it uses a fixed number score, which is 2.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: Updated.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: I have added a link for it.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: Updated.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: We use start ts as the version. I have mentioned it in the doc in my new commit.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: Yes, It should be an int, here we use float64 to simplify the code when we pass it to `math.Log10`.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: Oh, good point. It shouldn't say "In seconds" here. In code, we actually use nanoseconds.

Rustin170506: Updated.

Rustin170506: I will not remove the code comment in this PR. This PR is only for the design doc.

Rustin170506: I will fix it in another PR.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,567 @@
...
+```go
+type Heap[T any, K comparable] interface {
+    // Add/Update elements
+    Add(obj T) error
+    BulkAdd(list []T) error
+    AddIfNotPresent(obj T) error
+    Update(obj T) error // Same as Add
+
+    // Remove element
+    Delete(obj T) error
```
lance6716, line 181: Maybe add DeleteByKey

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,567 @@
...
+
+        // GetIndicators gets the indicators of the job.
+        GetIndicators() Indicators
+}
+```
+
+#### Basic API
+
+```go
+type Heap[T any, K comparable] interface {
```
lance6716, line 173: maybe to mention in comment what's the meaning of type `K`. It's the ID of `T` or the sort key in heap (I know it's the former after reading KeyFunc and LessFunc)

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,567 @@
...
+    // State operations
+    Close()
+    IsClosed() bool
+}
+
+// Constructor
+func NewHeap[T any, K comparable](keyFunc KeyFunc[T, K], lessFunc LessFunc[T]) *Heap[T, K]
+
+// Helper function type definitions
+type KeyFunc[T any, K comparable] func(T) (K, error)
```
lance6716, line 202: Why it will return error? I think ID of `T` is preset in `T` and will not have error when reading it.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,567 @@
...
+**A possible solution would be to start DDLNotifier on each TiDB node, with different subscribers on each node. This would allow us to solve the problem.**
+However, this solution introduces more complexities, such as how to manage subscribers, and ensuring that an event can be deleted after it has been processed by subscribers on all the different nodes becomes more complicated. Even the granularity of the subscribers needs to be controlled. For example, when the same subscriber logic is used, but they are running on every node, whether we consider it to be a single subscriber or multiple subscribers is something that needs to be discussed.
+**Since this issue is already far from our goal this time, I decided to consider it as a future possibility rather than in the pursuit of completing it in this design.**
+
+#### Reflect on Time Change
+
+We need to consider a time factor here because one of the metrics used to calculate our scores is the analysis interval, which is the time interval between the last analysis and the present. Obviously, it changes over time and we need to keep recalculating the analysis interval for these tables.
```
lance6716, line 476: I think here you want to mention dynamic polling interval? Can you explain how we should change it? For exmaple, when the heap gets larger we should use smaller interval

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,567 @@
...
+#### Failover
+
+Currently, there are only two scenarios where failover needs to be handled.
+
+1. the node is restarted and
+2. an owner switch occurs.
+
+In these two cases, we need to rebuild the priority queue, that is, to rescan and reconstruct the entire priority queue.
+For DML events, we scan the entire stats cache for the latest `modify_count` and `count`.
+For DDL events, since all our operations are processed in transactions. So after waiting for the priority queue to be ready, we just let the notifier restart processing.
```
lance6716, line 488: I only understand that the transaction consumes the mysql.ddl_notifier` table, but where is the table that the results are saved? Do we persist the enrties of heap in a table?

lance6716 (commented):
rest lgtm

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,567 @@
...
+For DML, we can get it from the stats cache as we load all table's modify_count and count by default, so we can just recalculate the score when table data changes.
+For DDL, we need a reliable way to get the DDL events that have occurred and recalculate the score as a result of the change in the table structure.
+
+### Incremental Update
+
+#### DML
+
+Currently, the stats module maintains the most basic information for all tables (modify_count and count) in the stats cache, and we can maintain the priority queue along the lines of incrementally updating it.
+
+Every time when there are DML changes, we will dump the update details into `mysql.stats_meta` with the new version. The version is the start timestamp of the transaction used to write the record.
```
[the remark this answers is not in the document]
Rustin170506: Yes, we have this problem, but it is acceptable because if there are new updates, then we load them correctly.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: Just for the sake of scalability, it's not always a simple number.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: No, this is specifically for the Analysis Interval indicator. Time passes, so we need to recalculate the time duration based on the latest start timestamp from the system. However, because we want to keep it in memory, we need to update it periodically with the new stats timestamp.

Because it consists of all memory operations, I suggest using a fixed interval to refresh it, for example: every 5 minutes. I believe that would be acceptable.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: No, we just rebuilding it with the latest schema and the stats cache.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
```diff
@@ -0,0 +1,567 @@
...
+
+We can try to keep the priority queue in memory to solve the above issues instead of rebuilding it every time.
+
+## Current Implementation
+
+### Calculate the score
+
+In the current implementation, we use a weighted scoring approach to maintain the queue, which is our current scoring criteria.
+
+Score table:
```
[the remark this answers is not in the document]
Rustin170506: Sounds good. Updated.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: Added.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: Added.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: Discussion offline: We don't want to change the formula in this design. However, this change will affect the order between the newly added index job and other jobs. Therefore, we will need to adjust the value of the newly added index event as well.

docs/design/2024-09-06-maintain-priority-queue-in-memory.md
Rustin170506: Offline discussion:
Because pq is time-consuming in the rebuild process, its api returns errors that can be retried until it finishes rebuilding, and then the notifier holds those events. Waiting until PQ rebuilds before sending them to it is acceptable for PQ to receive a duplicate ddl event since we always go with the latest schema.

lance6716 (approved):

qw4990 (approved):

PR #55922, opened 2024-09-06
executor: support memRowsIter for memIndexLookUpReader

tiancaiamao:

Issue Number: ref #43249

Problem Summary:

We have to change the memReader to an iterator API.
That is, get the mem rows inside the Next() call.

The old code will get all the mem buffer rows in `Open()` function, so there is no way to optimize it when the Next() call exit early. For example, 'select ... limit 1'

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

Use the following case:

```
func BenchmarkUnionScanRead(b *testing.B) {
	store := testkit.CreateMockStore(b)

	tk := testkit.NewTestKit(b, store)
	tk.MustExec("use test")
	tk.MustExec(`create table t_us (
a int,
c1 varchar(10),
c2 varchar(30),
c3 varchar(1),
c4 varchar(12),
c5 varchar(10),
c6 datetime,
index k(a));`)
	tk.MustExec(`begin;`)
	for i := 0; i < 8000; i++ {
		tk.MustExec(fmt.Sprintf("insert into t_us values (%d, '54321', '1234', '1', '000000', '7518', '2014-05-08')", i))
	}

	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		// tk.MustQuery("select * from t_us where c1 = '12345'").Check(testkit.Rows())
		tk.MustQuery("select * from t_us use index(k) order by a limit 1")
	}
	b.StopTimer()
	tk.MustExec("rollback")
}
```

Before vs After:

```
 go test -run XXX -bench BenchmarkUnionScanRead -cpuprofile cpu1.out -benchtime 45s
    3966          12947499 ns/op         9695363 B/op     105946 allocs/op

   22173           2408899 ns/op         1836421 B/op      17761 allocs/op
```

This PR does not make the full optimization for IndexLookup memReader, because when constructing the index to handle keyRanges, it's still handling all data in `Open()`.  But it's much better than before anyway.

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


tiancaiamao:
IndexLookup full scan also get better using this test:

```
func BenchmarkUnionScanRead(b *testing.B) {
	store := testkit.CreateMockStore(b)

	tk := testkit.NewTestKit(b, store)
	tk.MustExec("use test")
	tk.MustExec(`create table t_us (
a int,
c1 varchar(10),
c2 varchar(30),
c3 varchar(1),
c4 varchar(12),
c5 varchar(10),
c6 datetime,
index k(a));`)
	tk.MustExec(`begin;`)
	for i := 0; i < 8000; i++ {
		tk.MustExec(fmt.Sprintf("insert into t_us values (%d, '54321', '1234', '1', '000000', '7518', '2014-05-08')", i))
	}

	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		// tk.MustQuery("select * from t_us where c1 = '12345'").Check(testkit.Rows())
		tk.MustQuery("select * from t_us use index(k) where c2 in ('12345') order by a ASC")
	}
	b.StopTimer()
	tk.MustExec("rollback")
}
```

Before vs After:

```
 go test -run XXX -bench BenchmarkUnionScanRead -cpuprofile cpu1.out -benchtime 45s

    4092          12875574 ns/op         4809832 B/op      97943 allocs/op
    8182           6604521 ns/op         2155012 B/op      25666 allocs/op

```

pkg/executor/mem_reader.go
```diff
@@ -726,11 +728,62 @@ func buildMemIndexLookUpReader(ctx context.Context, us *UnionScanExec, idxLookUp
 }
 
 func (m *memIndexLookUpReader) getMemRowsIter(ctx context.Context) (memRowsIter, error) {
-	data, err := m.getMemRows(ctx)
-	if err != nil {
-		return nil, errors.Trace(err)
+	kvRanges := [][]kv.KeyRange{m.idxReader.kvRanges}
```
tiancaiamao, line 731: No new code are added.
This is indeed copied from `memRowsIterForTable` function

https://github.com/pingcap/tidb/pull/55922/files#diff-4745e9eb77a8450aa7d598691e57074ad2763cbb5570ad12493adba31bfb9310L955-L1010

bb7133: Do u mean copied from https://github.com/pingcap/tidb/blob/f9c4773ba34854cec93dec3c5cae8a625925f9e9/pkg/executor/mem_reader.go#L736-L794 ?

tiancaiamao: Oh, yes!

pkg/executor/mem_reader.go
```diff
@@ -726,11 +728,62 @@ func buildMemIndexLookUpReader(ctx context.Context, us *UnionScanExec, idxLookUp
...
+			handleBytes: make([]byte, 0, 16),
+			rd:          rd,
+			cd:          cd,
+		},
+		cacheTable:  m.cacheTable,
+		keepOrder:   m.keepOrder,
+		compareExec: m.compareExec,
+	}
+
+	return memTblReader.getMemRowsIter(ctx)
```
lcwangchao, line 786: Why not to delete `getMemRows` to reduce dup code, I think it is not used any more. Or we can update `getMemRows` implementation to use `getMemRowsIter` to read all rows.

lcwangchao (commented):
Rest LGTM

pkg/executor/mem_reader.go
tiancaiamao: OK, I've removed the getMemRows for memIndexLookupReader @lcwangchao

lcwangchao (approved):

bb7133 (approved):
LGTM

Issue #54245, opened 2024-06-26
Vector Search


breezewish:
This is the tracking issue of adding Vector Search to TiDB.

TiFlash tracking issue: https://github.com/pingcap/tiflash/issues/9032

PR #55934, opened 2024-09-09
planner: Display truncate vector in EXPLAIN

EricZequan:

Issue Number: ref #54245 

Problem Summary:

Before this PR, when we used `explain sql` to view the execution plan, it might be affected by the long string or vector entered by the user, making the execution plan difficult to read. This PR brings long constant truncation to more kind of plans.

We added `TruncatedStringify` in `pkg/types/datum.go` to handle the explain problem of `constant` type, interpreting variables with length longer than 64 as `xxx...len(x)`.

At the same time, the truncation of `constant` will affect the redact processing of some private data in the `import into` statement (data is lost after truncation). We processed the `initParameters` function in `pkg/executor/importer/import.go` so that it will not be affected by truncation.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/importer/import.go
+++ b/pkg/executor/importer/import.go
@@ -813,9 +813,12 @@ func (p *Plan) initParameters(plan *plannercore.ImportInto) error {
 	optionMap := make(map[string]any, len(plan.Options))
 	for _, opt := range plan.Options {
 		if opt.Value != nil {
-			var val string
+			// The option attached to the import statement here are all
+			// parameters entered by the user. TiDB will process the
+			// parameters entered by the user as constant. so we can
+			// directly convert it to constant.
 			cons := opt.Value.(*expression.Constant)
-			val = fmt.Sprintf("%v", cons.Value.GetValue())
+			val := fmt.Sprintf("%v", cons.Value.GetValue())
 			if opt.Name == cloudStorageURIOption {
 				val = ast.RedactURL(val)
 			}
--- a/pkg/executor/importer/import.go
+++ b/pkg/executor/importer/import.go
@@ -816,7 +816,7 @@ func (p *Plan) initParameters(plan *plannercore.ImportInto) error {
 	}
 	for _, opt := range plan.Options {
 		if opt.Value != nil {
-			val := opt.Value.StringWithCtxForImport(evalCtx, errors.RedactLogDisable)
+			val := opt.Value.StringifyWithoutTruncate(evalCtx, errors.RedactLogDisable)
 			if opt.Name == cloudStorageURIOption {
 				val = ast.RedactURL(val)
 			}
--- a/pkg/executor/importer/import.go
+++ b/pkg/executor/importer/import.go
@@ -52,7 +52,6 @@ import (
 	"github.com/pingcap/tidb/pkg/sessionctx"
 	"github.com/pingcap/tidb/pkg/sessionctx/variable"
 	"github.com/pingcap/tidb/pkg/table"
-	"github.com/pingcap/tidb/pkg/types"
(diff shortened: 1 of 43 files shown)


pkg/expression/expression.go
```diff
@@ -258,6 +258,8 @@ type Expression interface {
 	MemoryUsage() int64
 
 	StringerWithCtx
+
+	StringerWithCtxForImport
```
breezewish, line 262: Too ugly, because expression should not care anything about "import".

pkg/executor/importer/import.go
```diff
@@ -816,7 +818,23 @@ func (p *Plan) initParameters(plan *plannercore.ImportInto) error {
...
+			if constValue, ok := opt.Value.(*expression.Constant); ok {
+				// we can stringify 'constant' directly because
+				// redact is set to 'RedactLogDisable' by default
+				switch constValue.Value.Kind() {
+				case types.KindString, types.KindBytes:
+					val = constValue.Value.GetString()
+				case types.KindMysqlJSON:
+					val = constValue.Value.GetMysqlJSON().String()
+				case types.KindVectorFloat32:
+					val = constValue.Value.GetVectorFloat32().String()
```
EricZequan, line 831: @breezewish The option here is unlikely to process vectors. So maybe we can remove it ?

pkg/executor/importer/import.go
```diff
@@ -816,7 +816,7 @@ func (p *Plan) initParameters(plan *plannercore.ImportInto) error {
 	}
 	for _, opt := range plan.Options {
 		if opt.Value != nil {
-			val := opt.Value.StringWithCtx(evalCtx, errors.RedactLogDisable)
+			val := opt.Value.StringWithCtxForImport(evalCtx, errors.RedactLogDisable)
```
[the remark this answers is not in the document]
breezewish: So opt.Value can be an expression.Expression other than expression.Constant? What kind of SQL will cause this case? Could you come up an example SQL shows that we need to deal with this?

pkg/executor/importer/import.go
```diff
@@ -810,13 +811,11 @@ func (p *Plan) initParameters(plan *plannercore.ImportInto) error {
...
-	var evalCtx expression.EvalContext
-	if plan.SCtx() != nil {
-		evalCtx = plan.SCtx().GetExprCtx().GetEvalCtx()
-	}
 	for _, opt := range plan.Options {
 		if opt.Value != nil {
-			val := opt.Value.StringWithCtx(evalCtx, errors.RedactLogDisable)
+			var val string
+			cons := opt.Value.(*expression.Constant)
+			val = fmt.Sprintf("%v", cons.Value.GetValue())
```
breezewish, line 818:
```suggestion
			cons := opt.Value.(*expression.Constant)
			val := fmt.Sprintf("%v", cons.Value.GetValue())
```

Also please add some comment about why it is safe to assert opt.Value is a constant.

breezewish (commented):
The rest looks good!

breezewish:
Please also refine the description of the PR, cause there is no vector truncation at all in the master branch previously.

EricZequan:
Update.

hawkingrei (approved):

wjhuang2016 (approved):

JaySon-Huang (approved):
lgtm

windtalker (approved):
lgtm

GMHDBJD (approved):
LGTM

PR #55938, opened 2024-09-09
statistics: avoid frequantly syncing stats simultaneously

hawkingrei:

Issue Number: close #xxx

Problem Summary:

Now, ```StatsCacheImpl.Update``` can sync stats by version from the storage. we have many scenarios to call it. such as ```domain.loadStatsWorker``` and ```Analyze```. When customers use lightning, lightning will trigger the multi-analyzed tasks simultaneously. so it also triggers multi  ```StatsCacheImpl.Update``` simultaneously. it is unnecessary.

add a singleflight for ```StatsCacheImpl.Update```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/analyze.go
```diff
@@ -173,7 +173,8 @@ TASKLOOP:
 	if err != nil {
 		sessionVars.StmtCtx.AppendWarning(err)
 	}
-	return statsHandle.Update(ctx, infoSchema)
+	statsHandle.TriggerSyncStats()
```
Rustin170506, line 176: Here we need to make sure that it definitely executes successfully, otherwise we have the problem of a table being analyzed but immediately analyzed again. This is because our auto-analysis ticker will be triggered again during the analysis, and then because it didn't update the stats cache correctly to load the count and modify_count of the analyzed table, it will use the last count and modify_count to calculate the change ratio, which will ultimately cause it to queue up again.

hawkingrei: add force sync stats before the auto-analyze.

hawkingrei: Generally, the load time for this is in the millisecond range, and the gap time can be considered negligible.

Issue #55452, opened 2024-08-15
Add "Global" to SHOW INDEXES output


mjonss:
## Enhancement
Global index is currently in experimental state. `SHOW INDEXES IN t` does not have a column for GLOBAL, to indicate if it is a global index or not.

samba-rgb:
Hi @mjonss  I would like to work on this issue.

mjonss:
> Hi @mjonss I would like to work on this issue.

@samba-rgb I have assigned you, if you need any help, just ask! Looking at previous PRs may help as well :)

samba-rgb:
Hi @mjonss , i am done with code but unable to create a pr, Doesnt have permission to push any new branch
can you please guide me in this

mjonss:
Fixed by #56028

PR #56028, opened 2024-09-11
executor: Add Global to SHOW INDEX output | tidb-test=pr/2391

samba-rgb:
No info about index if its global or not , added in show index in this pr

Issue Number: ref #55452

Problem Summary:

added Global index from index info

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/show.go
+++ b/pkg/executor/show.go
@@ -792,6 +792,8 @@ func (e *ShowExec) fetchShowIndex() error {
 		return e.tableAccessDenied("SELECT", tb.Meta().Name.O)
 	}
 
+	isGlobalIndex := "NO"
+
 	if tb.Meta().PKIsHandle {
 		var pkCol *table.Column
 		for _, col := range tb.Cols() {
@@ -805,6 +807,14 @@ func (e *ShowExec) fetchShowIndex() error {
 		if colStats != nil {
 			ndv = colStats.NDV
 		}
+
+		pi := tb.Meta().GetPartitionInfo()
+		isPartitioned := pi != nil && pi.Type != pmodel.PartitionTypeNone
+
+		if isPartitioned && e.Ctx().GetSessionVars().EnableGlobalIndex {
+			isGlobalIndex = "YES"
+		}
+
 		e.appendRow([]any{
 			tb.Meta().Name.O, // Table
 			0,                // Non_unique
@@ -822,6 +832,7 @@ func (e *ShowExec) fetchShowIndex() error {
 			"YES",            // Index_visible
 			nil,              // Expression
 			"YES",            // Clustered
+			isGlobalIndex,    // Global_index
 		})
 	}
 	for _, idx := range tb.Indices() {
@@ -833,6 +844,12 @@ func (e *ShowExec) fetchShowIndex() error {
 		if tb.Meta().IsCommonHandle && idxInfo.Primary {
 			isClustered = "YES"
 		}
+
+		isGlobalIndex = "NO"
+		if idxInfo.Global {
+			isGlobalIndex = "YES"
+		}
+
 		for i, col := range idxInfo.Columns {
 			nonUniq := 1
 			if idx.Meta().Unique {
@@ -885,6 +902,7 @@ func (e *ShowExec) fetchShowIndex() error {
 				visible,                // Index_visible
 				expression,             // Expression
 				isClustered,            // Clustered
+				isGlobalIndex,          // Global_index
 			})
 		}
 	}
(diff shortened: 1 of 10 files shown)


pkg/executor/show.go
```diff
@@ -805,6 +807,14 @@ func (e *ShowExec) fetchShowIndex() error {
 		if colStats != nil {
 			ndv = colStats.NDV
 		}
+
+		pi := tb.Meta().GetPartitionInfo()
+		isPartitioned := pi != nil && pi.Type != pmodel.PartitionTypeNone
+
+		if isPartitioned && e.Ctx().GetSessionVars().EnableGlobalIndex {
+			isGlobalIndex = "YES"
```
mjonss, line 815: If PKIsHandle, then it is by definition a LOCAL index, i.e. each partition will have the clustered index, and the PK column must be a part of the partitioning expression.

Test by:
```sql
set tidb_enable_global_index = 0;
create table t (a int primary key clustered, b int) partition by hash (a) partitions 2;
show create table t;
show index from t;
set tidb_enable_global_index = 1;
show index from t;
```

pkg/executor/show.go
```diff
@@ -822,6 +832,7 @@ func (e *ShowExec) fetchShowIndex() error {
 			"YES",            // Index_visible
 			nil,              // Expression
 			"YES",            // Clustered
+			isGlobalIndex,    // Global_index
```
mjonss, line 835: So here you can just use "NO"
```suggestion
			"NO",    // Global_index
```

pkg/executor/show_test.go
```diff
@@ -97,3 +97,32 @@ func TestShow(t *testing.T) {
...
 }
+
+func TestShowIndex(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+
+	tk.MustExec("use test")
+	tk.MustExec("create table t(id int, abclmn int);")
+
+	defer tk.MustExec("drop table t;")
```
mjonss, line 108: No need to remove tables, each MockStore will be created fresh
```suggestion 
```

mjonss (commented):
Looks good, with some comments.
There are also more tests that needs to update (since the SHOW INDEX output changed).
You can run all unit tests with `make dev` and you can run integration tests by first building the `tidb-server` binary with `make` and then `(cd tests/integrationtest ; ./run-tests.sh -s ../../bin/tidb-server)`

mjonss:
@samba-rgb I can fix the idc-jenkins-ci-tidb/mysql-test test failures. Can you fix the other failed tests?

samba-rgb:
@mjons Thank you for helping with my SQL test fix. I will now resume working on the fixes for integration and unit tests, along with addressing the comments mentioned above.

pkg/executor/show.go
mjonss: Ah, sorry, I had not indented it correctly :(

samba-rgb:
Hi @mjonss i added the  Global column in the integrationtest/r/*/*.result path
 example : tidb/tests/integrationtest/r/session/clustered_index.result
 but still in the check_dev its not reflecting 
 is their any place that i need to change/add
 
 BTW when i ran make dev no test case is running for me 
 find: md5sum: No such file or directory

mjonss:
> Hi @mjonss i added the Global column in the integrationtest/r/_/_.result path example : tidb/tests/integrationtest/r/session/clustered_index.result but still in the check_dev its not reflecting is their any place that i need to change/add
> 
> BTW when i ran` make dev` no test case is running for me find: md5sum: No such file or directory

@samba-rgb 
For the integration tests, you can simply re-record the result files, it seems like there is some formatting issues in your change, likely spaces instead of tabs. Try `./run-tests.sh -s ../../bin/tidb-server -r session/clustered_index` and it will overwrite the .result file, and you can try with `./run-tests.sh -s ../../bin/tidb-server -t session/clustered_index` to verify that it passes. 

I assume you are using a mac? md5sum is md5 in macOs..., you can also run only the unit tests with `make ut` IIRC.
Also if there are a few integration tests that fails on mac, it is due to floating point difference between arm and i386, so leave them as-is.

I will try to find which unit tests still needing to be updated.

mjonss:
pkg/ddl/db_change_test.go:857
        	            	need [[t 0 PRIMARY 1 c1 A 0 <nil> <nil>  BTREE   YES <nil> NO]], but got [[t 0 PRIMARY 1 c1 A 0 <nil> <nil>  BTREE   YES <nil> NO NO]]
        	Test:       	TestShowIndex

To run the test you must also enable failpoints, `make failpoint-enable`, which you then can disable before committing, `make failpoint-disable`
Maybe this patch file could help?
[56028.patch](https://github.com/user-attachments/files/17050432/56028.patch)

lance6716 (approved):

winoros (approved):

Issue #53388, opened 2024-05-20
remove `mock.Context` usage in production code


lcwangchao:
## Enhancement

It's better to remove the `mock.Context` usages in TiDB production code. The reason is that `mock.Context` is not designed for production code. It is designed for testing, that means some values in it are not reasonable in production and some values will be modified by the developer without any care.

We can replace some code that uses `mock.Context` with some "static" context we introduced before such as `StaticExprContext` or `DistSQLContext`

## Tasks

- #53378
- #53389
- #53441
- #53467
- #53479
- #53642
- #53671
- #53721
- #55823
- #55873
- #56038
- #56174
- #56176
- #56348
- #56355
- #56654

PR #55955, opened 2024-09-09
* : Remove binlog related codes

Benjamin2037:
### What problem does this PR solve?
Remove TiDB-binlog related code.

Issue Number: close #55949

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [X] No need to test
  > Only remove code. 
     

Documentation

- [X] Affects user behaviors
- [X] Contains syntax changes
- [X] Contains variable changes
- [ ] Contains experimental features
- [X] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
TiDB-Binlog related code has been deleted, since TiDB-Binlog has been deprecated.
```

PR #56038, opened 2024-09-12
lighting: replace lightning inner contexts with manged ones

lcwangchao:

Issue Number: ref #53388

This RP replaces some internal contexts for lighting/import into to make sure to contexts are simple and easier to maintain, including:

- `litExprContext` which wraps `StaticExprContext` to provide expression context.
- `litTableMutateContext` provides table mutate context

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/importer/kv_encode.go
```diff
@@ -62,8 +62,6 @@ func NewTableKVEncoder(
 	if err != nil {
 		return nil, err
 	}
-	// we need a non-nil TxnCtx to avoid panic when evaluating set clause
-	baseKVEncoder.SessionCtx.SetTxnCtxNotNil()
```
lcwangchao, line 66: This is not needed any more because we have removed `SessionVars` in the new session.

pkg/lightning/backend/kv/session.go
```diff
@@ -312,6 +311,7 @@ func (*sessEvalContext) GetOptionalPropProvider(exprctx.OptionalEvalPropKey) (ex
 	return nil, false
 }
 
+// Deprecated: `session` will be removed soon.
```
lcwangchao, line 314: I have decided to remove the old `session` from the next PR. In this PR, it is used in some tests to check the old and new session's behavior are the same

pkg/lightning/backend/kv/context.go
```diff
@@ -0,0 +1,259 @@
...
+
+func (*litTableMutateContext) GetTemporaryTableSupport() (tbctx.TemporaryTableSupport, bool) {
+	// lightning import does not support temporary table.
+	return nil, false
+}
+
+func (*litTableMutateContext) GetExchangePartitionDMLSupport() (tbctx.ExchangePartitionDMLSupport, bool) {
+	// lightning import is not in a DML query, we do not need to support it.
+	return nil, false
+}
```
lcwangchao, line 230: PTAL @lance6716 @D3Hunter . some features not supported in lighthing/import into

lcwangchao:
/hold

Should waiting for #56059 merged first

pkg/lightning/backend/kv/context.go
```diff
@@ -0,0 +1,260 @@
...
+	"github.com/pingcap/tidb/pkg/sessionctx/stmtctx"
+	"github.com/pingcap/tidb/pkg/sessionctx/variable"
+	tbctx "github.com/pingcap/tidb/pkg/table/context"
+	"github.com/pingcap/tidb/pkg/types"
+	contextutil "github.com/pingcap/tidb/pkg/util/context"
+	"github.com/pingcap/tidb/pkg/util/intest"
+	"github.com/pingcap/tidb/pkg/util/timeutil"
+)
+
+// litExprContext implements the `exprctx.ExprContext` interface for lightning import.
```
lance6716, line 38: maybe add a `var _ exprctx.ExprContext = &litExprContext{}`

pkg/lightning/backend/kv/context.go
```diff
@@ -0,0 +1,260 @@
...
+	ctx = ctx.Apply(contextstatic.WithEvalCtx(evalCtx))
+
+	return &litExprContext{
+		StaticExprContext: ctx,
+		userVars:          userVars,
+	}, nil
+}
+
+// SetUserVarVal sets the value of a user variable.
+func (ctx *litExprContext) SetUserVarVal(name string, dt types.Datum) {
```
lance6716, line 101: no need to be public. Also for UnsetUserVar

pkg/lightning/backend/kv/context.go
```diff
@@ -0,0 +1,260 @@
...
+	ctx.userVars.SetUserVarVal(name, dt)
+}
+
+// UnsetUserVar unsets a user variable.
+func (ctx *litExprContext) UnsetUserVar(varName string) {
+	ctx.userVars.UnsetUserVar(varName)
+}
+
+// litTableMutateContext implements the `table.MutateContext` interface for lightning import.
```
lance6716, line 110: ditto

pkg/lightning/backend/kv/context.go
```diff
@@ -0,0 +1,260 @@
...
+
+// InRestrictedSQL implements the `table.MutateContext` interface.
+func (*litTableMutateContext) InRestrictedSQL() bool {
+	// Just return false because lightning import does not in any SQL.
+	return false
+}
+
+// TxnAssertionLevel implements the `table.MutateContext` interface.
+func (ctx *litTableMutateContext) TxnAssertionLevel() variable.AssertionLevel {
+	return ctx.assertionLevel
```
lance6716, line 150: just curious, can user change `assertionLevel` and below settings?

pkg/lightning/backend/kv/context.go
```diff
@@ -0,0 +1,260 @@
...
+func (ctx *litTableMutateContext) GetRowIDShardGenerator() *variable.RowIDShardGenerator {
+	return ctx.shardID
+}
+
+// GetReservedRowIDAlloc implements the `table.MutateContext` interface.
+func (ctx *litTableMutateContext) GetReservedRowIDAlloc() (*stmtctx.ReservedRowIDAlloc, bool) {
+	return &ctx.reservedRowIDAlloc, true
+}
+
+// GetBinlogSupport implements the `table.MutateContext` interface.
```
lance6716, line 178: this method will be soon removed by https://github.com/pingcap/tidb/pull/55955

pkg/lightning/backend/kv/context.go
```diff
@@ -0,0 +1,260 @@
...
+	return nil, false
+}
+
+// GetStatisticsSupport implements the `table.MutateContext` interface.
+func (ctx *litTableMutateContext) GetStatisticsSupport() (tbctx.StatisticsSupport, bool) {
+	return ctx, true
+}
+
+// UpdatePhysicalTableDelta implements the `table.StatisticsSupport` interface.
+func (ctx *litTableMutateContext) UpdatePhysicalTableDelta(
```
lance6716, line 190: I think lightning don't need implement whole logic. It only need column size by GetColumnSize. Or if we can maintain column size elsewhere this function can be a no-op.

pkg/lightning/backend/kv/session.go
```diff
@@ -438,34 +438,75 @@ func (*session) GetStmtStats() *stmtstats.StatementStats {
...
 }
 
 // GetTableCtx returns the table MutateContext.
 func (s *Session) GetTableCtx() tbctx.MutateContext {
-	return s.sctx.tblctx
+	return s.tblCtx
 }
 
 // TakeKvPairs returns the current Pairs and resets the buffer.
 func (s *Session) TakeKvPairs() *Pairs {
```
lance6716, line 508: lightning use `TakeKvPairs` to get encoded KV paris. Can we use `GetMutateBuffers` in future?

pkg/lightning/backend/kv/context.go
lcwangchao: Yes, if #55955 merged first, I'll remove this implementation. Otherwise, #55955 should remove this ...

pkg/lightning/backend/kv/context.go
lcwangchao: I think lightning does not forbid configurationing these variables: https://github.com/pingcap/tidb/blob/4371a1d4850f8924ba922f179daa7680ef2976da/lightning/pkg/importer/get_pre_info.go#L229-L234

This PR just keep the same behavior with the previous implementation.

pkg/lightning/backend/kv/context.go
lcwangchao: I've simplified the implementation, please take a look again.

pkg/lightning/backend/kv/session.go
lcwangchao: Currently, `GetMutateBuffers` is only designed for implementation in `tables` package so it does not provide general object pools. However, if you want to reuse the memory, the simplest way is to maintain a memory buffer in `Session` itself.

YangKeao (approved):
LGTM

pkg/lightning/backend/kv/session_internal_test.go
```diff
@@ -124,3 +130,46 @@ func TestKVMemBufBatchAllocAndRecycle(t *testing.T) {
...
+			"block_encryption_mode":   "aes-256-ecb",
+			"group_concat_max_len":    "2048",
+			"tidb_backoff_weight":     "6",
+			"tidb_row_format_version": "2",
+		},
+		Timestamp: 123456,
+	}, log.L())
+	require.NoError(t, err)
+	// some system vars should be loaded
+	require.Equal(t, uint64(40960), se.GetExprCtx().GetEvalCtx().GetMaxAllowedPacket())
```
D3Hunter, line 153: why we need a ExprCtx? it sounds like a static version of EvalCtx

pkg/lightning/backend/kv/context.go
```diff
@@ -0,0 +1,260 @@
...
+func (ctx *litExprContext) setUserVarVal(name string, dt types.Datum) {
+	ctx.userVars.SetUserVarVal(name, dt)
+}
+
+// UnsetUserVar unsets a user variable.
+func (ctx *litExprContext) unsetUserVar(varName string) {
+	ctx.userVars.UnsetUserVar(varName)
+}
+
+var _ table.MutateContext = &litTableMutateContext{}
```
D3Hunter, line 113: not related to this PR, I prefer put MutateContext into a pkg with more semantic meaning, a pkg of name `context` is so common, it's hard to infer what's inside

a example would be: put it into pkg named as `mutate`, and its name is `Context`, then we can use it as `mutate.Context`, no need for type alias

D3Hunter (commented):
rest lgtm

pkg/lightning/backend/kv/session_internal_test.go
lcwangchao: `ExprCtx` is used to build a context and `EvalCtx` is used for evaluating. `ExprCtx` contains `EvalCtx` because sometimes we need to evaluations when building expressions (It is hard to eliminate these codes because the concurrent implementation is so complex...).

pkg/lightning/backend/kv/context.go
lcwangchao: ok, I'll do it in other PRs

D3Hunter (approved):

PR #56060, opened 2024-09-13
ddl: fix admin check table error after rename index with scalar function 

### LINKED ISSUE #56045: panic when admin check table
## Bug Report

``` SQL
set global tidb_enable_global_index=1;
create table t ( col_1 json ,col_2 tinyint  unsigned not null default 57 , key idx_1 ( col_2 ,(cast(col_1 as CHAR(64) array)) ) ,unique key idx_2 ( (cast(col_1 as CHAR(64) array)) ) global ,primary key  ( col_2 )  /*T![clustered_index] clustered */ ) charset utf8 collate utf8_general_ci partition by hash ( col_2 ) partitions 5;

alter table t rename index idx_1 to idx_1_1;

admin check table t;
```

No panic
``` SQL

mysql> admin check table t;
ERROR 1105 (HY000): runtime error: index out of range [1] with length 1
```
master

joechenrh:

Issue Number: close #56045

Problem Summary:

`renameIndexes` didn't rename generated columns previously, causing error during `buildPhysicalIndexLookUpReader`.

To make the newly added test works, we don't clear `idxPart.Expr` (which is shallow copied from `stmt`) in `BuildHiddenColumnInfo`. Since this stmt will be used twice.

https://github.com/pingcap/tidb/blob/027f01b7ef88e41fd982996b675ba1cad94f715b/pkg/ddl/schematracker/checker.go#L225-L235

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/column.go
+++ b/pkg/ddl/column.go
@@ -1304,8 +1304,8 @@ func getChangingColumnOriginName(changingColumn *model.ColumnInfo) string {
 	return columnName[:pos]
 }
 
-func getExpressionIndexOriginName(expressionIdx *model.ColumnInfo) string {
-	columnName := strings.TrimPrefix(expressionIdx.Name.O, expressionIndexPrefix+"_")
+func getExpressionIndexOriginName(originalName pmodel.CIStr) string {
+	columnName := strings.TrimPrefix(originalName.O, expressionIndexPrefix+"_")
 	var pos int
 	if pos = strings.LastIndex(columnName, "_"); pos == -1 {
 		return columnName
--- a/pkg/ddl/create_table.go
+++ b/pkg/ddl/create_table.go
@@ -1432,6 +1432,7 @@ func BuildHiddenColumnInfo(ctx sessionctx.Context, indexPartSpecifications []*as
 				return nil, errors.Trace(err)
 			}
 		}
+		idxPart.Expr = nil
 		hiddenCols = append(hiddenCols, colInfo)
 	}
 	return hiddenCols, nil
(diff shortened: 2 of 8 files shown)


lance6716:
> To make the newly added test works, we don't clear idxPart.Expr (which is shallow copied from stmt) in BuildHiddenColumnInfo. Since this stmt will be used twice.

I think it's not worthy to change the code logic. You can call Checker.Disable / Enable for this test. There are other places that the original statement is changed inside DDL so we need to disable checker. And please write a unit test in pkg/ddl/schematracker to make sure the skipped SQL can be executed and SHOW CREATE TABLE is correct. You can refer to TestBitDefaultValues and checkShowCreateTable

wjhuang2016 (approved):

joechenrh:
> I think it's not worthy to change the code logic. You can call Checker.Disable / Enable for this test. There are other places that the original statement is changed inside DDL so we need to disable checker. And please write a unit test in pkg/ddl/schematracker to make sure the skipped SQL can be executed and SHOW CREATE TABLE is correct. You can refer to TestBitDefaultValues and checkShowCreateTable

May it's not necessary to write test in pkg/ddl/schematracker. Because `rename index` can be executed without error, and `SHOW CREATE TABLE` will also return correct info. The error is raised from `admin check table`.

lance6716:
Anyway, skipped DDL of checker should have a place to test. Otherwise we won't make sure it will always be correct in future

lance6716 (approved):

PR #56077, opened 2024-09-13
planner: support to prune column for simple table's DELETE | tidb-test=pr/2392

winoros:

Issue Number: ref #38911 

Problem Summary:

Support pruning columns for non-partitioned, non-fk tables in the DELETE statement. (point get will not prune also).

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Support column pruning for tables that are non-partitioned and don't have foreign keys in the DELETE statement, improving performance and saving network costs
支持对 DELETE 语句中没有外键的非分区表进行列裁剪，提升执行效率以及减少网络带宽
```


pkg/planner/core/point_get_plan.go
```diff
@@ -2071,7 +2071,11 @@ func buildPointDeletePlan(ctx base.PlanContext, pointPlan base.PhysicalPlan, dbN
 	intest.Assert(t != nil, "The point get executor is accessing a table without meta info.")
 	idxs := t.DeletableIndices()
 	deletableCols := t.DeletableCols()
-	colPosInfo, err := buildSingleTableColPosInfoForDelete(pointPlan.OutputNames(), handleCols, idxs, deletableCols, tbl)
+	colPosInfo, err := initColPosInfo(tbl.ID, pointPlan.OutputNames(), handleCols)
+	if err != nil {
+		return nil
+	}
+	_, err = pruneAndBuildSingleTableColPosInfoForDelete(pointPlan.OutputNames(), idxs, deletableCols, tbl, &colPosInfo, 0, nil)
```
cfzjywxk, line 2078: The https://github.com/pingcap/tidb/blob/master/pkg/planner/optimize.go#L233 fast path plan building should also be considered.

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5266,76 +5268,143 @@ func buildColumns2HandleWithWrtiableColumns(
 // This function records the following things:
 //  1. The position of the columns used by indexes in the DELETE's select's output.
 //  2. The row id's position in the output.
```
cfzjywxk, line 5270: Better to update the comments too as the returned result is changed.

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5266,76 +5268,143 @@ func buildColumns2HandleWithWrtiableColumns(
...
 			}
+			visitedCols[col.Offset] = struct{}{}
 		}
-		if !found {
-			return TblColPosInfo{}, plannererrors.ErrDeleteNotFoundColumn.GenWithStackByArgs(name.ColName.O, tblInfo.Name.O)
+	}
+
+	// Fix the column offsets.
+	pruned := 0
+	fixedPos := make(map[int]int, len(deletableCols))
```
cfzjywxk, line 5376: Could the memory be reused for `visitedCols` and `fixedPos` in some way?  As it's a critical path for each statement.

pkg/planner/core/testdata/plan_suite_unexported_in.json
```diff
@@ -595,5 +595,14 @@
...
       "select * from t as t1 left join t as t2 on t1.b = t2.b where (t1.c=1 and ((t1.a=3 and t2.a=3) or (t1.a=4 and t2.a=4))) or (t1.a=2 and t2.a is null)"
     ]
+  },
+  {
+    "name": "TestPruneColumnsForDelete",
+    "cases": [
+      "delete from t where a = 1",
+      "delete from t3 where b = 1",
+      "delete from pt2",
+      "delete t, t3 from t, t3 where t.a = t3.a"
```
cfzjywxk, line 605: Need to cover the fast path and the batch point get path.

pkg/planner/core/point_get_plan.go
winoros: the delete op should finally fall into this place. I think it should be safe. I'll add some test

AilinKid (approved):

AilinKid (approved):

pkg/table/tables/tables.go
```diff
@@ -1254,8 +1254,23 @@ func (t *TableCommon) removeRecord(ctx table.MutateContext, txn kv.Transaction,
...
+			columnOffset := id
+			if columnSizeOpt != nil {
+				if !columnSizeOpt.NotPruned.Test(uint(id)) {
+					size = int(columnSizeOpt.AvgSizes[pruned])
+					pruned++
+					colSizeBuffer.AddColSizeDelta(col.ID, -int64(size-1))
+					s.UpdatePhysicalTableDelta(t.physicalTableID, -1, 1, colSizeBuffer)
+					continue
+				}
+				columnOffset = columnSizeOpt.PublicColsLayout[notPruned]
```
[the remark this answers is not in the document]
winoros: The public columns here actually drop the ones which are `hidden`. So the offset is not equal.

pkg/planner/core/logical_plan_builder.go
winoros: I merged the two maps into one. Now the memory pressure is the same as before.

pkg/table/table.go
```diff
@@ -251,10 +274,31 @@ type IndexRowLayoutOption []int
...
 }
 
+// ColumnSizeOption records the column size information.
+// If the column is not pruned, we can calculate its acurate size by the real data.
+// Otherwise, we use the estimated avg size given by table statistics and field type information.
+type ColumnSizeOption struct {
+	NotPruned        *bitset.BitSet
+	AvgSizes         []float64
+	PublicColsLayout []int
+}
```
time-and-fate, line 296: You didn't mention why do we need this and what's this for.
There are also no comments for each field, which is not easy to understand.

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5266,76 +5268,184 @@ func buildColumns2HandleWithWrtiableColumns(
...
+) (int, error) {
 	// Columns can be seen by DELETE are the deletable columns.
 	tblLen := len(deletableCols)
-	offset, err := getTableOffset(names, names[handleCol.GetCol(0).Index])
-	if err != nil {
-		return TblColPosInfo{}, err
+
+	// If it's partitioned table, or has foreign keys, or is point get plan, we can't prune the columns, currently.
+	// nonPrunedSet will be nil if it's a point get or has foreign keys.
+	if tblInfo.GetPartitionInfo() != nil || nonPrunedSet == nil {
```
time-and-fate, line 5350: Why are we using two different methods to pass the two similar kinds of information?

time-and-fate: And since this `if` block doesn't reuse any other logic in this function, why do we make this part a separate function? This function is already too large and lacks comments.

go.mod
```diff
@@ -185,6 +186,7 @@ require (
 	github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d // indirect
 	github.com/apache/thrift v0.16.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
+	github.com/bits-and-blooms/bitset v1.14.3
```
time-and-fate, line 189: Why do we need another one after we already have `FastIntSet`?

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5266,76 +5270,179 @@ func buildColumns2HandleWithWrtiableColumns(
...
-	return cols2PosInfos, nil
+	// Sort by start position. To do the later column pruning.
+	slices.SortFunc(cols2PosInfos, func(i, j TblColPosInfo) int {
+		if i.Start < j.Start {
+			return -1
+		}
+		if i.Start > j.Start {
+			return 1
+		}
+		return 0
```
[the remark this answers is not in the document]
time-and-fate: Does "later" mean this PR or another PR?

go.mod
winoros: That one is for the sparsed case. When the elements are dense, it's not very suitable.
The bitset has an operation like `SetAll` which the sparse set will not have.

pkg/planner/core/logical_plan_builder.go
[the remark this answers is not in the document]
winoros: Another pr, i need to take a careful look to remove the binary search based on the old `sort` package.

go.mod
winoros: Or we just say that
- if we want to use a more high-performance struct to replace `[]bool`, we need the bitset.
- if we want to use a more high-performance struct to replace `map[int]struct{}`, we need the `FastIntSet`

cfzjywxk (approved):

pkg/table/table.go
```diff
@@ -243,6 +255,17 @@ type RemoveRecordOption interface {
 	applyRemoveRecordOpt(*RemoveRecordOpt)
 }
 
+// ExtraPartialRowOption is the combined one of IndexesLayout and ColumnSizeOption.
+type ExtraPartialRowOption struct {
+	IndexesRowLayout  IndexesLayout
+	ColumnsSizeHelper *ColumnsSizeHelper
+}
```
time-and-fate, line 262: `ExtraPartialRowOption` and `RemoveRecordOpt` have exactly the same definition. Why don't we just pass a `RemoveRecordOpt` struct.

pkg/table/table.go
```diff
@@ -251,8 +274,25 @@ type IndexRowLayoutOption []int
...
+// GetIndexLayout returns the layout of the specified index.
+func (idx IndexesLayout) GetIndexLayout(idxID int64) IndexRowLayoutOption {
+	if idx == nil {
+		return nil
+	}
+	return idx[idxID]
+}
+
+// ColumnsSizeHelper records the column size information.
+// If the column is pruned when doing DML, we can't get the accurate size of the column. So we need the estimated avg size.
```
time-and-fate, line 286: Please add why do we need to get the size of the column.

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5256,86 +5258,209 @@ func buildColumns2HandleWithWrtiableColumns(
...
+	for i, col := range publicCols {
+		// If the column is not pruned, we can use the column data to get a more accurate size.
+		// We just need to record its position info.
+		if _, ok := fixedPos[col.Offset]; ok {
+			colPosInfo.ColumnsSizeHelper.PublicColsLayout = append(colPosInfo.ColumnsSizeHelper.PublicColsLayout, fixedPos[col.Offset])
+			continue
 		}
+		// Otherwise we need to get the average size of the column by its field type.
+		// TODO: use statistics to get a maybe more accurate size.
+		colPosInfo.ColumnsSizeHelper.NotPruned.Clear(uint(i))
```
crazycs520, line 5429: The above comment is strange for this line, how about move this line after next line?

pkg/planner/core/logical_plan_builder.go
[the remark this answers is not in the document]
time-and-fate: Then we should leave this change to the next PR instead of letting these two places do the same thing.

pkg/table/table.go
```diff
@@ -251,8 +274,25 @@ type IndexRowLayoutOption []int
...
+//   - If the column is not pruned, we can calculate its acurate size by the real data.
+//   - Otherwise, we use the estimated avg size given by table statistics and field type information.
+type ColumnsSizeHelper struct {
+	// NotPruned is a bitset to record the columns that are not pruned.
+	NotPruned *bitset.BitSet
+	// If the column is pruned, we use the estimated avg size. They are stored by their ordinal in the table.
+	AvgSizes []float64
+	// If the column is not pruned, we use the accurate size. They are stored by their ordinal in the pruned row.
+	PublicColsLayout []int
```
time-and-fate, line 295: Please add what do they mean directly and explicitly.

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5256,86 +5258,209 @@ func buildColumns2HandleWithWrtiableColumns(
...
 //
 // The two kind of columns forms the whole columns of the table. Public part first.
-// This function records the following things:
-//  1. The position of the columns used by indexes in the DELETE's select's output.
-//  2. The row id's position in the output.
-func buildColPositionInfoForDelete(
+// This function does the following things:
+//  1. Prune the columns that are not used by indexes and row id(and the partition expression and foreign keys in the near future).
+//  2. The position of the columns used by indexes in the DELETE's select's output.
+//  3. The row id's position in the final output.
```
time-and-fate, line 5271: These comments don't make any sense.

crazycs520 (approved):
REST LGTM

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5923,23 +6042,49 @@ func (b *PlanBuilder) buildDelete(ctx context.Context, ds *ast.DeleteStmt) (base
...
+		finalProjNames = make(types.NameSlice, 0, oldLen/2)
+		for i, found := nonPruned.NextSet(0); found; i, found = nonPruned.NextSet(i + 1) {
+			finalProjCols = append(finalProjCols, p.Schema().Columns[i])
+			finalProjNames = append(finalProjNames, p.OutputNames()[i])
+		}
+	} else {
+		finalProjCols = make([]*expression.Column, oldLen)
+		copy(finalProjCols, p.Schema().Columns[:oldLen])
+		finalProjNames = preProjNames.Shallow()
+	}
```
time-and-fate, line 6079: There are no comments explaining 
1. What does `nonPruned != nil` mean
2. What's the relationship between `nonPruned` and `Schema`/`OutputNames`

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5923,23 +6042,49 @@ func (b *PlanBuilder) buildDelete(ctx context.Context, ds *ast.DeleteStmt) (base
...
+	var nonPruned *bitset.BitSet
+	del.TblColPosInfos, nonPruned, err = pruneAndBuildColPositionInfoForDelete(preProjNames, tblID2Handle, tblID2table, len(del.FKCascades) > 0 || len(del.FKChecks) > 0)
 	if err != nil {
 		return nil, err
 	}
+	var (
+		finalProjCols  []*expression.Column
+		finalProjNames types.NameSlice
+	)
+	// Use a very relax check for foreign key cascades and checks.
```
time-and-fate, line 6066: What does "relax check" mean?

go.mod
time-and-fate: `FastIntSet` provides optimization for sizes less than 64, which is just essentially a bitset.
And it also provides `AddRange()` method, which is even easier than `SetAll`.

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5256,86 +5258,209 @@ func buildColumns2HandleWithWrtiableColumns(
...
+	return nil
+}
+
+// pruneAndBuildSingleTableColPosInfoForDelete builds columns mapping for delete.
+// And it will try to prune columns that not used in pk or indexes.
+func pruneAndBuildSingleTableColPosInfoForDelete(
+	t table.Table,
 	tblInfo *model.TableInfo,
-) (TblColPosInfo, error) {
+	names []*types.FieldName,
```
time-and-fate, line 5370: We don't need to pass `TableInfo` if we already pass `table.Table`.

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5256,86 +5258,209 @@ func buildColumns2HandleWithWrtiableColumns(
...
+		indexColMap[idx.Meta().ID] = colPos
+	}
+	colPosInfo.End = colPosInfo.Start + tblLen
+	colPosInfo.IndexesRowLayout = indexColMap
+	return nil
+}
+
+// pruneAndBuildSingleTableColPosInfoForDelete builds columns mapping for delete.
+// And it will try to prune columns that not used in pk or indexes.
+func pruneAndBuildSingleTableColPosInfoForDelete(
```
time-and-fate, line 5367: This function is too large, unnecessarily complicated, and hard to understand.
We should initialize `TblColPosInfo`, do pruning, and collect information for the tables one by one, instead of initializing them as if there's no pruning, and then "fix" everything when we prune them.

pkg/executor/test/oomtest/oom_test.go
```diff
@@ -151,6 +152,7 @@ func TestMemTracker4InsertAndReplaceExec(t *testing.T) {
 }
 
 func TestMemTracker4DeleteExec(t *testing.T) {
+	t.Skip("don't know why, skip first to test others")
```
time-and-fate, line 155: What about this case?

pkg/table/tables/tables.go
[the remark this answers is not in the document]
time-and-fate: I agree that `PublicColsLayout` is not necessary.
We can use a map of "`Column.ID` -> index in the row" or "`Column.Offset` -> index in the row" to replace the `PublicColsLayout` and `IndexesLayout`.

pkg/table/table.go
winoros: This one is to be consistent with @lcwangchao 's new API changes

pkg/planner/core/logical_plan_builder.go
winoros: The current implementation achieves the optimal time complexity.

go.mod
winoros: No, they are two different things.
image
The `FastIntSet` is actually a set. Not a bitmask compression for `[]bool`.

pkg/planner/core/logical_plan_builder.go
winoros: If there are two tables and just one table has fk, only that table can not prune columns. But here we directly reject pruning for the two tables. So the check is a very relax one, not a strict check.

pkg/executor/test/oomtest/oom_test.go
```diff
@@ -156,8 +157,8 @@ func TestMemTracker4DeleteExec(t *testing.T) {
 	tk := testkit.NewTestKit(t, store)
 	tk.MustExec("use test")
 	tk.MustExec("set tidb_cost_model_version=1")
-	tk.MustExec("create table MemTracker4DeleteExec1 (id int, a int, b int, index idx_a(`a`))")
-	tk.MustExec("create table MemTracker4DeleteExec2 (id int, a int, b int, index idx_a(`a`))")
```
winoros, line 160: The test needs IndexLookUp to trigger some action.
So we need to add an index.

PR #56082, opened 2024-09-13
*: Drop partition DDL handling for overlapping partitions during State Changes

### LINKED ISSUE #55888: DROP PARTITION has an anomaly during DELETE ONLY state, which looks like PK violation.
## Bug Report

Repeatable test case [here](https://github.com/mjonss/tidb/blob/b8a37f29bdfd215656e3917a62177b669c947b1f/pkg/ddl/tests/partition/multi_domain_test.go#L42-L48). Check out the branch, `make failpoint-enable` then run the test TestMultiSchemaDropPartition.

It looks like this in a multi domain (tidb node) scenario:
```sql
-- DDL Owner node
create table t (a int primary key, b varchar(255)) partition by range (a) (partition p0 values less than (100), partition p1 values less than (200));
insert into t values (1,1);
-- Schema version v0
alter table t drop partition p0;
-- new client1 seeing v0+1 StateDeleteOnly, most likely in the DDL Owner domain/node
-- valid, since the table does no longer contain partition p0
insert into t values (1,1);
-- in the other domain, client 2, still in v0, still seeing partition p0
-- looks like PK violation!!!
select * from t; -- "1 1", "1 1"
-- Original row
select * from t partition (p0); -- "1 1"
-- This is not consistent with the view of the partition definitions!
-- sees new row from client1
select * from t partition (p1); -- "1 1"
```

This is not completely well defined with TiDB's DDL model, since it allows two schema versions with the same data.
But the most consistent view would probably be to not allow duplicate keys when the dropping partitions still can be used.
How to handle the anomaly that rows can exists in an overlapping partition (this case p1, when p0 is being dropped) is harder to define, since it would also be strange to not allow inserts/updates in p1 (that may match dropped p0) by client1 during StateDeleteOnly just because it may overlap with possible sessions seeing v0 schema versions

See repeatable test case.
Duplicate rows and rows in p1 that matches the partition definition of p0 in client2.

b8a37f29bdfd215656e3917a62177b669c947b1f based on current master 9c7f4eb455a4e781e05da4f4125c49fda29a74f5
i.e. v8.4.0-alpha something

mjonss:

Issue Number: close #55888

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/casetest/partition/testdata/partition_pruner_in.json
```diff
@@ -218,11 +218,11 @@
       },
       {
         "SQL": "select * from t1 where t1.a in (select b from t2 where a in (1,2)) order by a",
-        "Pruner": "t1: p0,p1; t2: p0"
+        "Pruner": "t1: all; t2: p0"
       },
       {
         "SQL": "select * from t1 where t1.a in (select b from t1 where a in (1,2)) order by a",
-        "Pruner": "t1: p0; t1: p0,p1"
+        "Pruner": "t1: all; t1: p0"
```
Defined2014, line 225: I think should be `t1: p0; t1: p0`. Am I right?

pkg/planner/core/point_get_plan.go
```diff
@@ -682,6 +673,12 @@ func (p *BatchPointGetPlan) getPartitionIdxs(sctx sessionctx.Context) []int {
 			rows[i][j].Copy(&r[p.IndexInfo.Columns[j].Offset])
 		}
 		pIdx, err := pTbl.GetPartitionIdxByRow(sctx.GetExprCtx().GetEvalCtx(), r)
+		if err != nil && pIdx >= 0 {
+			pIdx = pTbl.Meta().Partition.GetOverlappingDroppingPartitionIdx(pIdx)
+			if pIdx >= 0 {
+				err = nil
+			}
```
Defined2014, line 680: Maybe we could combine those lines as a new function.

pkg/planner/core/casetest/partition/testdata/partition_pruner_in.json
mjonss: This indeed looks weird, I will investigate and see what changed...

mjonss: The pruning or plan has not changed, but when extracting the Pruner string, the strings are sorted first by table, then by partitions, so 'all' is less than 'p...', so since it is the same table, "t1: all" < "t1: p..." resulting in the need to change the order :)

pkg/meta/model/table.go
```diff
@@ -890,6 +890,18 @@ func (pi *PartitionInfo) CanHaveOverlappingDroppingPartition() bool {
 	return false
 }
 
+// ReplaceWithOverlappingPartitionIdx returns the overlapping partition
+// if
```
Defined2014, line 894: Miss comments

pkg/planner/core/partition_prune.go
```diff
@@ -38,6 +38,53 @@ func PartitionPruning(ctx base.PlanContext, tbl table.PartitionedTable, conds []
...
+						if len(newRet) == len(pi.Definitions) {
+							return []int{FullRange}, nil
+						}
+						return newRet, nil
+					}
+					break
+				}
+				if len(partitionNames) == 0 || s.findByName(partitionNames, pi.Definitions[idx].Name.L) {
+					newRet = append(newRet, idx)
+				}
```
Defined2014, line 77: When will this happen?

pkg/planner/core/partition_prune.go
```diff
@@ -38,6 +38,53 @@ func PartitionPruning(ctx base.PlanContext, tbl table.PartitionedTable, conds []
...
+						continue
+					}
+					if ret[end] == idx {
+						// overlapping partition already included, append it together with the following
+						newRet = append(newRet, ret[end:]...)
+						if len(newRet) == len(pi.Definitions) {
+							return []int{FullRange}, nil
+						}
+						return newRet, nil
+					}
```
Defined2014, line 72: Maybe just break here. It seems it already covered by outside loop.

pkg/planner/core/partition_prune.go
mjonss: This is when the overlapping partition was not included in the pruning result, so it should be added, unless explicit partition pruning is given in partitionNames, and it is not included there.
I have reformatted the code a little, with your suggestion above, and added comments to make it more clear.

Defined2014 (approved):

pkg/meta/model/table.go
```diff
@@ -854,6 +858,96 @@ func (pi *PartitionInfo) GetPartitionIDByName(partitionDefinitionName string) in
...
+	}
+	return false
+}
+
+// ReplaceWithOverlappingPartitionIdx returns the overlapping partition
+// if there is one and a previous error.
+// Functions based on locatePartitionCommon, like GetPartitionIdxByRow
+// will return the found partition, with an error, since it is being dropped.
+// This function will correct the partition index and error if it can.
+func (pi *PartitionInfo) ReplaceWithOverlappingPartitionIdx(idx int, err error) (int, error) {
```
qw4990, line 898: Could you add a concrete example here to help understand what this function is doing?

pkg/planner/core/casetest/partition/testdata/integration_partition_suite_out.json
```diff
@@ -1440,7 +1440,7 @@
       {
         "SQL": "select * from tlist2 where a = 1 and b in (1,2) order by b",
         "DynamicPlan": [
-          "TableReader 2.00 root partition:p0,p1 data:TableRangeScan",
+          "TableReader 2.00 root partition:all data:TableRangeScan",
```
qw4990, line 1443: Will this PR cause some performance regression since I see in some cases that partition pruning can work well as before?

mjonss: This difference is just making it more clear, that `p0,p1` is actually `all` partitions, so there should be no performance regression related to partition pruning.
The impact in general should be minimal, it will need to check if the table's PartitionInfo is in a state that can have overlapping partitions (very rare to hit, only during one State change for DROP PARTITION), and if it is, then some smaller CPU usage for handling the partitions.

pkg/meta/model/table.go
qw4990: Especially more comments to explain the term `OverlappingDropping` here.

pkg/table/tables/partition.go
```diff
@@ -1312,30 +1317,46 @@ func (t *partitionedTable) locatePartitionCommon(ctx expression.EvalContext, tp
 		} else {
 			idx, err = t.locateRangePartition(ctx, partitionExpr, r)
 		}
+		if err != nil {
+			return -1, err
+		}
+		pi := t.Meta().Partition
+		if pi.CanHaveOverlappingDroppingPartition() {
+			pid := pi.Definitions[idx].ID
+			for _, droppingDef := range pi.DroppingDefinitions {
```
qw4990, line 1326: It seems better to create a function for this operation like `pi.IsDropping(idx)` to help simplify this code.

pkg/table/tables/partition.go
```diff
@@ -1312,30 +1317,46 @@ func (t *partitionedTable) locatePartitionCommon(ctx expression.EvalContext, tp
...
+				if pid == droppingDef.ID {
+					// idx != -1, so the correct partition, but give an error, since it should
+					// not be written to! For read it can check the Overlapping partition and ignore the error.
+					// One should use the next non-dropping partition for range, or the default
+					// partition for list partitioned table with default partition, for read.
+					return idx, table.ErrNoPartitionForGivenValue.GenWithStackByArgs(fmt.Sprintf("matching a partition being dropped, '%s'", droppingDef.Name.String()))
```
qw4990, line 1332: So if the query matches a dropping partition, it will return an error directly?

qw4990, line 1332: Why don't we call `pi.ReplaceOverlappingDroppingPartitionIdx` here?

pkg/meta/model/table.go
```diff
@@ -854,6 +858,96 @@ func (pi *PartitionInfo) GetPartitionIDByName(partitionDefinitionName string) in
...
+func (pi *PartitionInfo) GetOverlappingDroppingPartitionIdx(idx int) int {
+	if idx < 0 || idx >= len(pi.Definitions) {
+		return -1
+	}
+	if pi.CanHaveOverlappingDroppingPartition() {
+		switch pi.Type {
+		case model.PartitionTypeRange:
+		DEFINITIONS_LOOP:
+			for i := idx; i < len(pi.Definitions); i++ {
+				for _, def := range pi.DroppingDefinitions {
```
qw4990, line 925: Same as the prior comment, better to create a function for this operation like `pi.IsDropping(idx)`

pkg/planner/core/partition_prune.go
```diff
@@ -38,6 +38,48 @@ func PartitionPruning(ctx base.PlanContext, tbl table.PartitionedTable, conds []
...
+				if end >= len(ret) || ret[end] != idx {
+					// It must also match partitionNames if explicitly given
+					if len(partitionNames) == 0 || s.findByName(partitionNames, pi.Definitions[idx].Name.L) {
+						newRet = append(newRet, idx)
+					}
+				}
+				if end < len(ret) {
+					newRet = append(newRet, ret[end:]...)
+				}
+				break
```
qw4990, line 76: Emm this part seems tricky, could you please add a concrete example here?

pkg/planner/core/partition_prune.go
```diff
@@ -38,6 +38,48 @@ func PartitionPruning(ctx base.PlanContext, tbl table.PartitionedTable, conds []
...
+			}
+			if len(newRet) == len(pi.Definitions) {
+				return []int{FullRange}, nil
+			}
+			return newRet, nil
+		}
```
qw4990, line 82: How about moving this part into a separate function like `handleDroppingForRange(idx)` or something like this?

pkg/table/tables/partition.go
mjonss: Yes, so that writes to that partition (in that specific DDL State) will fail, i.e. being blocked.

pkg/planner/core/partition_prune.go
mjonss: Yes it is a bit tricky, not sure how to express the code in the best way.
But what it should do is:
from a list of indexes/offsets of pi.Definitions (i.e partitions to be used after partition pruning)
- if full range, replace the 'full range' marker with a range of all partitions
- remove the partitions that are being dropped
- if not already included add the 'overlapping' partition (also check explicit partition pruning in partitionNames)

How it is implemented is by looping over all used partitions to create a new set of used partitions
- add the non-dropping partitions
- Until an 'overlapping' partition is found, then
  - skip the full range of dropping partitions (for range partitioning, one can only drop a consecutive set of partitions).
  - if not already included, add the 'overlapping' partition, with a check that it also matches explicit partition pruning/partitionNames if given.
  - append all following used partitions.
And before returning, check if it matches the full range, if so return the simpler FullRange marker.

Actual example:
```sql
CREATE TABLE t (a int, b varchar(255))
PARTITION BY RANGE (a)
(PARTITION p0 VALUES LESS THAN (10),
 PARTITION p1 VALUES LESS THAN (20),
 PARTITION p2 VALUES LESS THAN (30),
 PARTITION p3 VALUES LESS THAN (40));
-- various DML...
ALTER TABLE t DROP PARTITION p1,p2;
```

During DDL state StateWriteOnly, the dropping partitions will still be there, but 'blocked' for read and write, so for write it will simply give an error, but for read we must take care for the possibility that other sessions might see the next DDL state StateDeleteOnly, where the partition p1 and p2 are gone, so p3 will accept values into the same range as p1 and p2 previously had.

So to not hide any values or have inconsistency depending on if you do pruning or not we need to replace dropping partitions with the 'overlapping' partition when doing partition pruning (filtering etc. will then make sure that we still return the matching rows.

@qw4990 do you want me to go through the code what happens in this scenario?

pkg/meta/model/table.go
mjonss: Hi, I have updated the function comments, with an example of OverlappingDropping connected to CanHaveOverlappingDroppingPartition. Please read through it if it makes sense?

pkg/table/tables/partition.go
mjonss: Done, thanks for the suggestion!

mjonss: oh, this place is not using it yet, fixing...

qw4990 (approved):

pkg/ddl/partition.go
```diff
@@ -2144,75 +2129,115 @@ func dropLabelRules(ctx context.Context, schemaName, tableName string, partNames
 	return infosync.UpdateLabelRules(ctx, patch)
 }
 
-// onDropTablePartition deletes old partition meta.
-func (w *worker) onDropTablePartition(jobCtx *jobContext, job *model.Job) (ver int64, _ error) {
+func (w *worker) rollbackLikeDropPartition(jobCtx *jobContext, job *model.Job) (ver int64, _ error) {
```
Benjamin2037, line 2132: Can you add some comments to describe this func's usage?

pkg/ddl/partition.go
```diff
@@ -97,7 +97,7 @@ func checkAddPartition(t *meta.Mutator, job *model.Job) (*model.TableInfo, *mode
 func (w *worker) onAddTablePartition(jobCtx *jobContext, job *model.Job) (ver int64, _ error) {
 	// Handle the rolling back job
 	if job.IsRollingback() {
-		ver, err := w.onDropTablePartition(jobCtx, job)
+		ver, err := w.rollbackLikeDropPartition(jobCtx, job)
```
Benjamin2037, line 100: As DDL execution state machine the rollbackone statement still need go through some state to Public that does in onDropTablePartition func does. In rollbackLikeDropPartition I did not see this kind of state machine path execution?

mjonss: The rollback logic has not been changed in this PR, just moved the code to a separat function.

For ADD partition, there is only two states, so the first state transition should cancel directly, without need for rollback, the second state transition it can do rollback directly in a single state transition.

For REORGANIZE partition, it will normally be called after `convertReorgPartitionJob2RollbackJob()`, which does handle state transitions where needed, to then allow `rollbackLikeDropPartition` to do the final cleanup.

Issue #55722, opened 2024-08-28
ddl, planner: introduce a DDL notifier to allow the statistics module to subscribe DDL events


fzzf678:
## Background
In statistic module, we need all table information to rebuild the priority queue to do auto analyze every time,  but in information schema v2, TiDB no longer stores all table information in memory(load as required), so we need a method to avoid traverse all tables as soon as possible, so we can try to keep the priority queue in memory to solve this problem.

In order to cooperate with the new priority queue we need a method to notify the statistics module handle DDL events.

## Development tasks

1. Introduce a new ddl change structure `SchemaChangeEvent` to replace the old `DDLEvent`, use getters and setters to avoid direct access to the struct's fields.
2. Add ddl notifier table and support ddl changes writing (write).
3. Implementation of the ddl notifier struct with subscription support (read).

PR #56089, opened 2024-09-14
ddl_notifier: implement serialization and deserialization for `SchemaChangeEvent`

fzzf678:

Issue Number: ref #55722

Problem Summary:

Add `Marshal` and `UnMarshal` method for `SchemaChangeEvent`.
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/ddl/notifier/events.go
```diff
@@ -380,3 +381,44 @@ func NewFlashbackClusterEvent() *SchemaChangeEvent {
...
+	AddedPartInfo        *model.PartitionInfo
+	DroppedPartInfo      *model.PartitionInfo
+	ColumnInfos          []*model.ColumnInfo
+	OldTableID4Partition int64
+
+	Tp model.ActionType
+}
+
+// MarshalJSON implements the json.Marshaler interface.
+func (s *SchemaChangeEvent) MarshalJSON() ([]byte, error) {
```
lance6716, line 397: use 

```golang
type SchemaChangeEvent struct {
  inner jsonSchemaChangeEvent
}
```

so we don't need to copy all fields, while the `SchemaChangeEvent` still has private members.

fzzf678: Oh, I tried, it really works. But if so, when we want to access the filed in `SchemaChangeEvent`, we need to use `s.inner.TableInfo` or `s.inner.Tp` to do this, the code becomes loner. It looks like a trade off, if we don't want to copy all field, we need more code access the field.Do you have any
advice on this？

lance6716: It's expected to use `s.inner.XXX` to access them. The main reason is if in future we add more fields, we may forget to copy them between `SchemaChangeEvent` and `jsonSchemaChangeEvent`. This tradeoff is also agreed in the pre-RFC document.

pkg/ddl/notifier/publish_testkit_test.go
```diff
@@ -33,7 +35,7 @@ func TestPublishToTableStore(t *testing.T) {
 CREATE TABLE ddl_notifier (
 	ddl_job_id BIGINT,
 	multi_schema_change_seq BIGINT COMMENT '-1 if the schema change does not belong to a multi-schema change DDL. 0 or positive numbers representing the sub-job index of a multi-schema change DDL',
-	schema_change JSON COMMENT 'SchemaChange at rest',
+	schema_change LONGBLOB COMMENT 'SchemaChange at rest',
```
lance6716, line 38: What's the reason to change the column type?

fzzf678: I tried the `JSON` type and use `row.GetJSON()` and `row.GetBytes()` get the serialized data, but it can't be deserialized. I took a look at other system tables like `tidb_global_task` and `tidb_background_subtask`, they use `LONGBLOB ` type save the meta and `row.GetBytes()` get the serialized data and I did so. Is it my way of operating the `JSON` type is wrong?

lance6716:
```diff
diff --git a/pkg/ddl/notifier/publish_testkit_test.go b/pkg/ddl/notifier/publish_testkit_test.go
index 30f0c1170f..d5e01cbabf 100644
--- a/pkg/ddl/notifier/publish_testkit_test.go
+++ b/pkg/ddl/notifier/publish_testkit_test.go
@@ -35,7 +35,7 @@ func TestPublishToTableStore(t *testing.T) {
 CREATE TABLE ddl_notifier (
        ddl_job_id BIGINT,
        multi_schema_change_seq BIGINT COMMENT '-1 if the schema change does not belong to a multi-schema change DDL. 0 or positive numbers representing the sub-job index of a multi-schema change DDL',
-       schema_change LONGBLOB COMMENT 'SchemaChange at rest',
+       schema_change JSON COMMENT 'SchemaChange at rest',
        processed_by_flag BIGINT UNSIGNED DEFAULT 0 COMMENT 'flag to mark which subscriber has processed the event',
        PRIMARY KEY(ddl_job_id, multi_schema_change_seq)
 )
diff --git a/pkg/ddl/notifier/store.go b/pkg/ddl/notifier/store.go
index c652fd0052..f259277e0b 100644
--- a/pkg/ddl/notifier/store.go
+++ b/pkg/ddl/notifier/store.go
@@ -94,7 +94,12 @@ func (t *tableStore) List(ctx context.Context, se *sess.Session, limit int) ([]*
        ret := make([]*schemaChange, 0, len(rows))
        for _, row := range rows {
                event := SchemaChangeEvent{}
-               err = json.Unmarshal(row.GetBytes(2), &event)
+               binaryJSON := row.GetJSON(2)
+               jsonStr, err := binaryJSON.MarshalJSON()
+               if err != nil {
+                       return nil, errors.Trace(err)
+               }
+               err = json.Unmarshal(jsonStr, &event)
                if err != nil {
                        return nil, errors.Trace(err)
                }
```

However I don't have strong opinion about the column type. JSON type can add JSON syntax check in the SQL layer and readable SELECT output, LONGTEXT has readable SELECT output. LONGBLOB has no advantage. Maybe wait other reviewers' opinions.

fzzf678: I tried again and found that the `JSON` type in table is not equal to our internal `json` type in code, we need a conversion to use the `JSON` type in table like this:
```
j, err := row.GetJSON(2).MarshalJSON()
err = json.Unmarshal(j, &event)
```
 https://github.com/pingcap/tidb/blob/master/pkg/types/json_binary.go#L151-L180

pkg/ddl/notifier/events.go
```diff
@@ -370,55 +385,46 @@ func (s *SchemaChangeEvent) GetRemovePartitioningInfo() (
...
 
+// jsonSchemaChangeEvent is used by SchemaChangeEvent when needed to (un)marshal data,
+// we want to hide the details to subscribers, so SchemaChangeEvent contain this struct.
 type jsonSchemaChangeEvent struct {
-	TableInfo            *model.TableInfo
-	OldTableInfo         *model.TableInfo
-	AddedPartInfo        *model.PartitionInfo
-	DroppedPartInfo      *model.PartitionInfo
-	ColumnInfos          []*model.ColumnInfo
+	TableInfo       *model.TableInfo
```
lance6716, line 405: add omitempty in json tag

pkg/ddl/notifier/publish_testkit_test.go
lance6716, line 38:
```suggestion
	schema_change LONGBLOB COMMENT 'SchemaChangeEvent at rest',
```

lance6716 (commented):
rest lgtm

pkg/ddl/notifier/events.go
fzzf678: updated in https://github.com/pingcap/tidb/pull/56089/commits/dbd3445e1294feed31627a874061ec8a15b79029 and https://github.com/pingcap/tidb/pull/56089/commits/fe72a490bdff4ed6aa5e335091573416af659b93

CbcWestwolf (approved):

pkg/ddl/notifier/publish_testkit_test.go
```diff
@@ -33,7 +35,7 @@ func TestPublishToTableStore(t *testing.T) {
 CREATE TABLE ddl_notifier (
 	ddl_job_id BIGINT,
 	multi_schema_change_seq BIGINT COMMENT '-1 if the schema change does not belong to a multi-schema change DDL. 0 or positive numbers representing the sub-job index of a multi-schema change DDL',
-	schema_change JSON COMMENT 'SchemaChange at rest',
+	schema_change LONGBLOB COMMENT 'SchemaChangeEvent at rest',
```
lance6716, line 38: Because other reviewers and me has no comment, I think you can choose the type as you like. Please `/unhold` after finished.

pkg/ddl/notifier/events.go
```diff
@@ -369,14 +385,46 @@ func (s *SchemaChangeEvent) GetRemovePartitioningInfo() (
...
+		inner: &jsonSchemaChangeEvent{
+			Tp: model.ActionFlashbackCluster,
+		},
+	}
+}
+
+// jsonSchemaChangeEvent is used by SchemaChangeEvent when needed to (un)marshal data,
+// we want to hide the details to subscribers, so SchemaChangeEvent contain this struct.
+type jsonSchemaChangeEvent struct {
+	TableInfo       *model.TableInfo     `json:"tableInfo"`
```
lance6716, line 405: 1. add `omitempty` JSON tag
2. seems in other places of ddl package we use snake_case

fzzf678: PTAL  again https://github.com/pingcap/tidb/pull/56089/commits/be7d271e89af9ef7bb09c8c38156d1b6dc77f007

lance6716 (approved):

PR #56112, opened 2024-09-18
lightning: buffer all region jobs and consider store balance

### LINKED ISSUE #54886: Lightning import throughput should be balanced across all TiKVs when the table is pre-split
## Enhancement
When using Lightning to import data, if pre-split is performed on the table and node rate limiting is enabled on TiKV, it cause  continuous range regions to be distributed on the same store. This results in store-level hotspots during the import process, causing the import speed to fall far short of expectations. Currently, PD only determines balance and scatter regions based on the number of regions per store. Therefore, even if Lightning calls scatter regions before data import, it does not have the desired effect.

### LINKED ISSUE #56113: balance lighting throughput to all tikvs
## Enhancement

as https://github.com/pingcap/tidb/pull/54976 said

lance6716:

Issue Number: close #56113 close #54886

Problem Summary:

add `storeBalancer`

```
	 [prepareAndSendJob]---jobToWorkerCh->[storeBalancer(optional)]->[workers]
	                     ^                                             |
	                     |                                             v
	                     |                           [storeBalancer(optional)]
	                     |                                             |
	                     |                                     jobFromWorkerCh
	                     |                                             |
	                     |                                             v
	               [regionJobRetryer]<-------------[dispatchJobGoroutine]-->done
```

and refine the components workflow. Now the behaviour of workflow is described in code comment

```
	// Above is the happy path workflow of region jobs. A job is generated by
	// prepareAndSendJob and terminated to "done" state by dispatchJobGoroutine. We
	// maintain an invariant that the number of generated jobs (after job.ref())
	// minus the number of "done" jobs (after job.done()) equals to jobWg. So we can
	// use jobWg to wait for all jobs to be finished.
	...
```

Previously I don't want to write many `select {case <- ctx.Done: ..., case ...}` for every channel action, so in the old code I assume the output channel must not be blocked (which means its reader/receiver must not exit earlier than writer/sender). But after I come back to this function after some months, I found it's hard to remember the different exit behaviour of components. So I changed them to new behaviours like comment said.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/lightning/backend/local/local.go
```diff
@@ -1364,63 +1367,104 @@ func (local *Backend) doImport(
...
+	                     |                                             v
+	                     |                           [storeBalancer(optional)]
+	                     |                                             |
+	                     |                                     jobFromWorkerCh
+	                     |                                             |
+	                     |                                             v
+	               [regionJobRetryer]<-------------[dispatchJobGoroutine]-->done
 	*/
 
+	// Above is the happy path workflow of region jobs. A job is generated by
```
D3Hunter, line 1381: maybe put all those comment at method level, so i can see it when cursor hover

D3Hunter (commented):
will review later

pkg/lightning/backend/local/local.go
lance6716: These are the inner logic of `doImport`. Do you think it's also helpful to see them at the caller of `doImport`? If you can confirm it I will move them.

pkg/lightning/backend/local/region_job.go
```diff
@@ -975,14 +983,33 @@ func (q *regionJobRetryer) run(ctx context.Context) {
 			select {
 			case <-ctx.Done():
 				return
+			case <-q.done:
+				return
```
D3Hunter, line 987: use sub context? it duplicates with `case <-ctx.Done():`

pkg/lightning/backend/local/region_job.go
```diff
@@ -1013,3 +1040,235 @@ func (q *regionJobRetryer) push(job *regionJob) bool {
...
+		innerJobToWorkerCh:   make(chan *regionJob),
+		innerJobFromWorkerCh: make(chan *regionJob),
+		wakeSendToWorker:     make(chan struct{}, 1),
+		done:                 make(chan struct{}),
+		jobWg:                jobWg,
+	}
+}
+
+func (b *storeBalancer) run(workerCtx context.Context) {
+	var wg sync.WaitGroup
```
D3Hunter, line 1084: use util.WaitGroupWrapper in case it panics

pkg/lightning/backend/local/region_job.go
```diff
@@ -1013,3 +1040,235 @@ func (q *regionJobRetryer) push(job *regionJob) bool {
...
+	jobIdx int
+	jobWg  *sync.WaitGroup
+
+	jobToWorkerCh        <-chan *regionJob
+	jobFromWorkerCh      chan<- *regionJob
+	innerJobToWorkerCh   chan *regionJob
+	innerJobFromWorkerCh chan *regionJob
+
+	wakeSendToWorker chan struct{}
+	done             chan struct{}
```
D3Hunter, line 1061: do it too

pkg/lightning/backend/local/region_job.go
```diff
@@ -1013,3 +1040,235 @@ func (q *regionJobRetryer) push(job *regionJob) bool {
...
+
+	b.jobs.Delete(bestIdx)
+	// in unit tests, the fields of job may not set
+	if best.region == nil || best.region.Region == nil {
+		return best
+	}
+
+	for _, p := range best.region.Region.Peers {
+	retry:
+		if val, ok := b.storeLoadMap.Load(p.StoreId); !ok {
```
D3Hunter, line 1218: use LoadOrStore?

D3Hunter (commented):
rest lgtm

pkg/lightning/backend/local/region_job.go
lance6716: in current code, `run` and `close` are executed in two goroutines. If we use sub context, `close` need to wait `run` is called to initialize the sub context. The caller will be more complicated. I prefer current `done` channel

pkg/lightning/backend/local/region_job.go
```diff
@@ -1013,3 +1040,235 @@ func (q *regionJobRetryer) push(job *regionJob) bool {
...
+		select {
+		case <-workerCtx.Done():
+			return
+		case job, ok := <-b.innerJobFromWorkerCh:
+			if !ok {
+				close(b.jobFromWorkerCh)
+				return
+			}
+			// in unit tests, the fields of job may not set
+			if job.region != nil && job.region.Region != nil {
```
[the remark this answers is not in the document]
lance6716: I'll check this comment soon

pkg/lightning/backend/local/region_job.go
D3Hunter: we can init the context on `new`

lance6716: https://github.com/pingcap/tidb/pull/56112/commits/3f8f6c5b27ec0292fb156563e1be2a1e03287614

pkg/lightning/backend/local/region_job.go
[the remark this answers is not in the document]
lance6716: https://github.com/pingcap/tidb/pull/56112/commits/c4ee8f31728892aca58e242fdbc581b4973403ea

pkg/lightning/backend/local/region_job.go
```diff
@@ -1075,33 +1075,38 @@ func newStoreBalancer(
...
+	wg.Go(func() error {
+		b.runReadFromWorker(ctx2)
+		return nil
+	})
+
+	if err := wg.Wait(); err != nil {
+		return err
+	}
 
 	b.jobs.Range(func(_, value any) bool {
```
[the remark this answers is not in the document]
lance6716: when `wg` has error, the caller `doImport` will use error case to exit. `jobWg` is only used in happy path. I'll check if the comment can be refined later.

pkg/lightning/backend/local/local.go
```diff
@@ -1473,16 +1530,28 @@ func (local *Backend) doImport(
 		}
 
 		jobWg.Wait()
-		workerCancel()
+		if balancer != nil {
+			intest.AssertFunc(func() bool {
+				allZero := true
+				balancer.storeLoadMap.Range(func(key, value any) bool {
```
lance6716, line 1536:
```suggestion
				balancer.storeLoadMap.Range(func(_, value any) bool {
```

D3Hunter (approved):

tangenta (approved):

PR #56136, opened 2024-09-19
Optimizer : Add OR pruning to predicate simplification | tidb-test=pr/2397 

### LINKED ISSUE #56005: optimizer lacks feature of OR expansion and predicate elimination
## Bug Report


```sql
CREATE TABLE `dt` (
  `a` bigint(20) unsigned NOT NULL,
  `pk` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `b` longblob DEFAULT NULL,
  `c` int(11) unsigned NOT NULL,
  PRIMARY KEY (`pk`)
);

CREATE TABLE `it` (
  `pk` bigint(20) unsigned NOT NULL,
  `a` varbinary(767) DEFAULT NULL,
  `b` varbinary(767) DEFAULT NULL,
  `c` varbinary(767) DEFAULT NULL,
  `d` bigint(20) DEFAULT NULL,
  `e` varbinary(767) DEFAULT NULL,
  `f` int(11) unsigned NOT NULL,
  PRIMARY KEY (`pk`) ,
  KEY `a` (`a`,`d`,`c`,`pk`),
  KEY `b` (`b`,`pk`),
  KEY `c` (`d`,`pk`),
  KEY `d` (`e`,`pk`),
  KEY `e` (`c`,`pk`),
  KEY `f` (`a`,`pk`)
);

explain SELECT
  dt.*
FROM
  it 
  LEFT JOIN dt ON it.pk = dt.pk
WHERE
  it.a = "a"
  AND (
    (
      it.a > "a"
    )
    OR (
      it.a = "a" AND it.pk > 1
    )
  )
ORDER BY
  it.pk
LIMIT
  240;
```

There is an index `KEY `f` (`a`,`pk`)`, and there is where predicate `it.a = "a" AND it.pk > 1`. We expect to see `range:("a" 1,"a" +inf]` in execution plan

This is the full execution plan
```sql
+---------------------------------------+---------+-----------+--------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| id                                    | estRows | task      | access object            | operator info                                                                                                                                                                                                                                                                                      |
+---------------------------------------+---------+-----------+--------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Projection_12                         | 4.19    | root      |                          | test.dt.a, test.dt.pk, test.dt.b, test.dt.c                                                                                                                                                                                                    |
| └─Limit_19                            | 4.19    | root      |                          | offset:0, count:240                                                                                                                                                                                                                                                                                |
|   └─IndexHashJoin_80                  | 4.19    | root      |                          | left outer join, inner:TableReader_75, outer key:test.it.pk, inner key:test.dt.pk, equal cond:eq(test.it.pk, test.dt.pk), other cond:or(gt(test.it.a, "a"), and(eq(test.it.a, "a"), gt(test.dt.pk, 1))) |
|     ├─Limit_36(Build)                 | 3.36    | root      |                          | offset:0, count:240                                                                                                                                                                                                                                                                                |
|     │ └─IndexReader_54                | 3.37    | root      |                          | index:Limit_53                                                                                                                                                                                                                                                                                     |
|     │   └─Limit_53                    | 3.37    | cop[tikv] |                          | offset:0, count:240                                                                                                                                                                                                                                                                                |
|     │     └─Selection_52              | 3.37    | cop[tikv] |                          | or(gt(test.it.a, "a"), and(eq(test.it.a, "a"), gt(test.it.pk, 1)))                                                                                                                                                                                          |
|     │       └─IndexRangeScan_51       | 10.00   | cop[tikv] | table:it, index:f(a, pk) | range:["a","a"], keep order:true, stats:pseudo                                                                                                                                                                                                                                                     |
|     └─TableReader_75(Probe)           | 3.36    | root      |                          | data:TableRangeScan_74                                                                                                                                                                                                                                                                             |
|       └─TableRangeScan_74             | 3.36    | cop[tikv] | table:dt                 | range: decided by [test.it.pk], keep order:false, stats:pseudo                                                                                                                                                                                                                        |
+---------------------------------------+---------+-----------+--------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```
We see `range:["a","a"]`. This is causing unnecessary additional data access and poor performance.
```sql
mysql> select @@version;
+--------------------+
| @@version          |
+--------------------+
| 8.0.11-TiDB-v7.5.1 |
+--------------------+
1 row in set (0.00 sec)
```

ghazalfamilyusa:

Issue Number: close #56005

### Problem Summary:
For this predicate it.a = "a" AND (it.a > "a" OR it.a = "a" AND it.pk > 1), the condition  it.a > "a" in the OR can be pruned since 
- This expression can be rewritten as (it.a = "a" AND it.a > "a") OR (it.a = "a" AND it.a = "a" AND it.pk > 1) 
- (it.a = "a" AND it.a > "a") is false
- false OR (it.a = "a" AND it.a = "a" AND it.pk > 1)  is (it.a = "a" AND it.a = "a" AND it.pk > 1) . 
This logic is already implemented in the optimizer but it is limited when the collation/type of column and constant are not the same. In this example, it.a is binary and "a" is string. The restriction is too strict since the string value is treated like binary and not casting is applied.

We can safely apply the pruning logic in this case since no casting is done and we can easily drive "a" > "a" which is false.
The change is applied to the new predicate simplification code and not the legacy code which need to be refactored in another PR.  The main idea for P & (P1 OR P2, ... OR Pn), we prune Pi when P &Pi is false. If all Pi are pruned then the whole predicate is false. 

### Check List

Tests 

- [X] Unit test
- [] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/rule_predicate_simplification.go
```diff
@@ -147,6 +190,116 @@ func applyPredicateSimplification(sctx base.PlanContext, predicates []expression
...
+		equalPred = p1
+		otherPred = p2
+	} else if p2Type == equalPredicate {
+		equalPred = p2
+		otherPred = p1
+	}
+	if equalPred == nil || otherPred == nil {
+		return false
+	}
+	// Copy constant from equal predicate into other predicate.
```
AilinKid, line 221: just a easy explanation for this
```
expr1: col=$1
expr2: myfunc(col, z)
===> myfunc($1,z)
```
then use PropagateConstant to converge new expr: like(a>1 and a<1) ===> false

tests/integrationtest/t/planner/core/casetest/predicate_simplification.test
```diff
@@ -31,3 +31,112 @@ explain format = 'brief' select count(*) cnt from ts where f <> '1' and f in ('1
...
+      it.a = "a" AND it.pk > 1
+    )
+  )
+ORDER BY
+  it.pk
+LIMIT
+  240;
+
+
+explain SELECT * FROM
```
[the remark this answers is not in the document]
ghazalfamilyusa: OK. Done.

pkg/planner/core/rule_predicate_simplification.go
```diff
@@ -147,6 +190,116 @@ func applyPredicateSimplification(sctx base.PlanContext, predicates []expression
 	return newValues
 }
 
+// Check for constant false condition.
+func unsatisfiableExpression(ctx base.PlanContext, p expression.Expression) bool {
+	if constExpr, ok := p.(*expression.Constant); ok {
+		if b, err := constExpr.Value.ToBool(ctx.GetSessionVars().StmtCtx.TypeCtx()); err == nil && b == 0 {
```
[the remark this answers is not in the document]
ghazalfamilyusa: This is covered I think in the old code path. To make sure, the objective of this PR is to support the airbnb use case in an extensible manner using the new predicatee simplification rule. A subsequent effort will be refactoring and more general solutions.

pkg/planner/core/rule_predicate_simplification.go
```diff
@@ -34,35 +38,57 @@ type predicateType = byte
...
 			if _, ok := args[1].(*expression.Constant); !ok {
 				return nil, otherPredicate
 			}
-			return col, notEqualPredicate
 		}
-		if v.FuncName.L == ast.In {
+		if v.FuncName.L == ast.NE {
+			return col, notEqualPredicate
+		} else if v.FuncName.L == ast.EQ {
+			return col, equalPredicate
```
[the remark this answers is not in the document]
ghazalfamilyusa: Correct for something like 5 = x. But, per my comment above the purpose is not a full scope predicate simplification solution here for corner cases liek this (users and tools typically do not write queries like this). In any case, a subsequent effort will be refactoring old code path and providing a general solution.

pkg/planner/core/rule_predicate_simplification.go
ghazalfamilyusa: This case actually is not covered by PropagateConstant if the type of col and constant are not the same.

AilinKid: i see.

AilinKid (commented):
from overall point of view, it works for some special cases by now, while we still a little bit hacky in some special judgement like what we do in findPredicateType() and the chained caller, we could refine and make it pretty in future efforts. LGTM

tests/integrationtest/t/planner/core/casetest/predicate_simplification.test
```diff
@@ -31,3 +31,112 @@ explain format = 'brief' select count(*) cnt from ts where f <> '1' and f in ('1
 explain format = 'brief' select f from t where f <> NULL  and f in (1,2,3) -- Special case of NULL with no simplification.;
 explain format = 'brief' select f from t where f != NULL  and f in (NULL,2,3) -- Special case of NULL with no simplification.;
 
+# Test OR simplification start
+# Airbnb test case
```
hawkingrei, line 35: It is best not to include the customer's name

ghazalfamilyusa: Agree and removed.

AilinKid (approved):

pkg/planner/core/rule_predicate_simplification.go
```diff
@@ -34,35 +38,57 @@ type predicateType = byte
...
 					return nil, otherPredicate
 				}
 			}
 			return col, inListPredicate
+		} else {
+			return nil, otherPredicate
 		}
 	default:
 		return nil, otherPredicate
 	}
```
hawkingrei, line 91:
```suggestion
		}
	default:
	}
	return nil, otherPredicate
```

pkg/planner/core/rule_predicate_simplification.go
```diff
@@ -147,6 +190,116 @@ func applyPredicateSimplification(sctx base.PlanContext, predicates []expression
...
+	// Copy constant from equal predicate into other predicate.
+	equalValue := equalPred.(*expression.ScalarFunction)
+	otherValue := otherPred.(*expression.ScalarFunction)
+	newPred := expression.NewFunctionInternal(ctx.GetExprCtx(), otherValue.FuncName.L, otherValue.RetType, equalValue.GetArgs()[1], otherValue.GetArgs()[1])
+	newPredList := make([]expression.Expression, 0, 1)
+	newPredList = append(newPredList, newPred)
+	newPredList = expression.PropagateConstant(ctx.GetExprCtx(), newPredList)
+	return unsatisfiableExpression(ctx, newPredList[0])
+}
+func comparisonPred(predType predicateType) predicateType {
```
hawkingrei, line 230:
```suggestion
}

func comparisonPred(predType predicateType) predicateType {
```

pkg/planner/core/rule_predicate_simplification.go
```diff
@@ -147,6 +190,116 @@ func applyPredicateSimplification(sctx base.PlanContext, predicates []expression
...
+			_, jType := findPredicateType(jthPredicate)
+			iType = comparisonPred(iType)
+			jType = comparisonPred(jType)
+			if iType == scalarPredicate && jType == orPredicate {
+				predicates[j] = updateOrPredicate(sctx, jthPredicate, ithPredicate)
+				sctx.GetSessionVars().StmtCtx.SetSkipPlanCache("OR predicate simplification is triggered")
+			} else if iType == orPredicate && jType == scalarPredicate {
+				predicates[i] = updateOrPredicate(sctx, ithPredicate, jthPredicate)
+				sctx.GetSessionVars().StmtCtx.SetSkipPlanCache("OR predicate simplification is triggered")
+			}
```
hawkingrei, line 298:
```suggestion
			if iType == scalarPredicate && jType == orPredicate {
				predicates[j] = updateOrPredicate(sctx, jthPredicate, ithPredicate)
			} else if iType == orPredicate && jType == scalarPredicate {
				predicates[i] = updateOrPredicate(sctx, ithPredicate, jthPredicate)
				
			}
		  sctx.GetSessionVars().StmtCtx.SetSkipPlanCache("OR predicate simplification is triggered")
```

hawkingrei: These two SetSkipPlanCache statements are identical; they can be moved outside.

pkg/planner/core/rule_predicate_simplification.go
hawkingrei: There should be blank lines between the functions.

ghazalfamilyusa: OK

pkg/planner/core/rule_predicate_simplification.go
ghazalfamilyusa: SetSkipPlanCache is applied in "if" and "else if" blocks and not "else". Moving it outside requires another check and does not help.

pkg/planner/core/rule_predicate_simplification.go
```diff
@@ -147,6 +190,116 @@ func applyPredicateSimplification(sctx base.PlanContext, predicates []expression
...
+	_, orPredicateType := findPredicateType(orPredicateList)
+	_, scalarPredicateType := findPredicateType(scalarPredicatePtr)
+	scalarPredicateType = comparisonPred(scalarPredicateType)
+	if orPredicateType != orPredicate || scalarPredicateType != scalarPredicate {
+		return orPredicateList
+	}
+	v := orPredicateList.(*expression.ScalarFunction)
+	firstCondition := v.GetArgs()[0]
+	secondCondition := v.GetArgs()[1]
+	_, firstConditionType := findPredicateType(firstCondition)
```
hawkingrei, line 252: This section seems to contain identical code segments. Could a single function, or even an anonymous function, be used to simplify the code?

hawkingrei: 252-267L

ghazalfamilyusa: Agree the code is repeated for argument 0 and 1 and I considered factoring those. But, the common code is recursive and I would rather avoid this indirect recursion. Let me see if I can refactor and keep the direct recursion.

pkg/planner/core/rule_predicate_simplification.go
hawkingrei: Okay, I was mistaken here.

tokens used: 313

pkg/planner/core/rule_predicate_simplification.go
hawkingrei: Actually, it's not a big issue. Sometimes Golang will inline this part of the code by itself.

hawkingrei (approved):

PR #56145, opened 2024-09-19
executor: verify the existence of the resource group before switching

JmPotato:

Issue Number: ref #54434.

Before switching, verify the existence of the resource group. If the specified resource group does not exist, take no action.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [x] Manual test

Before:

image

After:

image

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/adapter.go
+++ b/pkg/executor/adapter.go
@@ -555,7 +555,12 @@ func (a *ExecStmt) Exec(ctx context.Context) (_ sqlexec.RecordSet, err error) {
 			return nil, err
 		}
 		if len(switchGroupName) > 0 {
-			stmtCtx.ResourceGroupName = switchGroupName
+			group, err := rm.ResourceGroupCtl.GetResourceGroup(switchGroupName)
+			if err != nil || group == nil {
+				logutil.BgLogger().Debug("invalid switch resource group", zap.String("switch-group-name", switchGroupName), zap.Error(err))
+			} else {
+				stmtCtx.ResourceGroupName = switchGroupName
+			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


JmPotato:
@nolouch @glorv PTAL

nolouch (commented):
do we need some tests?

pkg/executor/adapter.go
```diff
@@ -555,7 +555,12 @@ func (a *ExecStmt) Exec(ctx context.Context) (_ sqlexec.RecordSet, err error) {
 			return nil, err
 		}
 		if len(switchGroupName) > 0 {
-			stmtCtx.ResourceGroupName = switchGroupName
+			group, err := rm.ResourceGroupCtl.GetResourceGroup(switchGroupName)
```
nolouch, line 558: BTW, Have we already checked when setting up a non-existent group?

JmPotato: Considering that the records of `QUERY WATCH` and the resource group are decoupled, there is currently no check on the existence of the resource group when setting up. This allows for a more flexible configuration of `QUERY WATCH` and simplifies the code. Here are a few typical scenarios: we can set the corresponding resource group after adding `SWITCH_GROUP`; or after deleting a resource group, the corresponding runaway statements will not fail immediately.

JmPotato:
> do we need some tests?

After trying, I found that the existing unit tests and integration tests could not test this situation well. I first completed the test manually and posted the test results in the PR description. PTAL.

glorv (approved):

nolouch (approved):

Issue #55184, opened 2024-08-05
region distribution not balance in create table with enable `tidb_scatter_region` in cluster level


River2000i:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
1. enable `tidb_scatter_region`
2. create table to split region

### 2. What did you expect to see? (Required)
region distribution should balance in cluster level
### 3. What did you see instead (Required)
![image](https://github.com/user-attachments/assets/09f60a5e-a426-4f89-bc27-148beab3aaf9)
1. create table t1 (a int, b int,index idx1(a)) shard_row_id_bits = 10 pre_split_regions=8;
2. create table t1 (a int, b int,index idx1(a));
3. create table t1 (a int, b int,index idx1(a)) PARTITION BY RANGE(b) PARTITION p0 VALUES LESS THAN (10),PARTITION p1 VALUES LESS THAN (20)); 
### 4. What is your TiDB version? (Required)
```sql
mysql> select version();
+--------------------+
| version()          |
+--------------------+
| 8.0.11-TiDB-v8.2.0 |
+--------------------+
1 row in set (0.04 sec)
```

PR #56157, opened 2024-09-19
ddl: support scatter region in cluster/global level

River2000i:

Issue Number: close https://github.com/pingcap/tidb/issues/55184 ref https://github.com/tikv/pd/issues/8424 ref #54886

Problem Summary:
region distribution not balance in create table with enable `tidb_scatter_region` in cluster level.

1. `tidb_scatter_region` support `""`, `"table"`, `"global"` instead of `on` and `off`.
2. In `global`, trigger PD scatter region is achieved by passing `tableID` as nil. When `tableID` is nil, PD selects new peers based on the cluster level by https://github.com/tikv/pd/blob/13174b5d4cab4d958f0b4ea2718ea7bb74992bc7/pkg/schedule/scatter/region_scatterer.go#L444

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
refer: https://github.com/pingcap/docs-cn/pull/18626/files
```


River2000i:
@D3Hunter @Benjamin2037 PTAL~

pkg/session/session.go
```diff
@@ -3212,7 +3212,7 @@ func splitAndScatterTable(store kv.Storage, tableIDs []int64) {
 		for _, id := range tableIDs {
 			regionIDs = append(regionIDs, ddl.SplitRecordRegion(ctxWithTimeout, s, id, id, variable.DefTiDBScatterRegion))
 		}
-		if variable.DefTiDBScatterRegion {
+		if variable.DefTiDBScatterRegion != "" {
```
Benjamin2037, line 3215: Suggest use enum type to represent value list。

pkg/ddl/executor.go
```diff
@@ -1382,21 +1382,21 @@ func preSplitAndScatter(ctx sessionctx.Context, store kv.Storage, tbInfo *model.
...
-		scatterRegion bool
+		preSplit func()
+		scope    scatterScope
 	)
-	val, err := ctx.GetSessionVars().GetGlobalSystemVar(context.Background(), variable.TiDBScatterRegion)
-	if err != nil {
-		logutil.DDLLogger().Warn("won't scatter region", zap.Error(err))
+	val, ok := ctx.GetSessionVars().GetSystemVar(variable.TiDBScatterRegion)
+	if !ok {
+		logutil.DDLLogger().Warn("won't scatter region")
```
Benjamin2037, line 1390:
```suggestion
		logutil.DDLLogger().Warn("get system variable met problem, won't scatter region")
```

pkg/ddl/executor.go
```diff
@@ -1382,21 +1382,21 @@ func preSplitAndScatter(ctx sessionctx.Context, store kv.Storage, tbInfo *model.
...
 	}
 	if len(parts) > 0 {
-		preSplit = func() { splitPartitionTableRegion(ctx, sp, tbInfo, parts, scatterRegion) }
+		preSplit = func() { splitPartitionTableRegion(ctx, sp, tbInfo, parts, scope) }
 	} else {
-		preSplit = func() { splitTableRegion(ctx, sp, tbInfo, scatterRegion) }
+		preSplit = func() { splitTableRegion(ctx, sp, tbInfo, scope) }
 	}
-	if scatterRegion {
+	if scope != scatterOff {
```
Benjamin2037, line 1399: Why here "scope == scatteOff" we need start new go routine for preSplit?

pkg/ddl/partition.go
```diff
@@ -3230,7 +3230,7 @@ func (w *worker) onReorganizePartition(jobCtx *jobContext, t *meta.Meta, job *mo
 		// and we will soon start writing to the new partitions.
 		if s, ok := jobCtx.store.(kv.SplittableStore); ok && s != nil {
 			// partInfo only contains the AddingPartitions
-			splitPartitionTableRegion(w.sess.Context, s, tblInfo, partInfo.Definitions, true)
+			splitPartitionTableRegion(w.sess.Context, s, tblInfo, partInfo.Definitions, scatterTable)
```
Benjamin2037, line 3233: Why here use "table" level scatter stratage?

pkg/ddl/split_region.go
```diff
@@ -107,20 +115,54 @@ func preSplitPhysicalTableByShardRowID(ctx context.Context, store kv.SplittableS
...
+	regionIDs = append(regionIDs, splitIndexRegion(store, tbInfo, scatter, tableID)...)
 	return regionIDs
 }
 
 // SplitRecordRegion is to split region in store by table prefix.
-func SplitRecordRegion(ctx context.Context, store kv.SplittableStore, physicalTableID, tableID int64, scatter bool) uint64 {
+func SplitRecordRegion(ctx context.Context, store kv.SplittableStore, physicalTableID, tableID int64, scope scatterScope) uint64 {
 	tableStartKey := tablecodec.GenTablePrefix(physicalTableID)
-	regionIDs, err := store.SplitRegions(ctx, [][]byte{tableStartKey}, scatter, &tableID)
+	var (
```
Benjamin2037, line 148: Can we merge line 148 ~ 164 into one funcs? since same code segments in multi place.

pkg/ddl/table_split_test.go
```diff
@@ -45,7 +45,7 @@ func TestTableSplit(t *testing.T) {
 	tk := testkit.NewTestKit(t, store)
 	tk.MustExec("use test")
 	// Synced split table region.
-	tk.MustExec("set global tidb_scatter_region = 1")
+	tk.MustExec("set @@session.tidb_scatter_region = 1")
```
Benjamin2037, line 48: Need also keep set global level sys var testcase, that means we need add session level test case and keep global level tests.

pkg/ddl/executor.go
River2000i: When `scope != scatterOff `, the `CREATE TABLE` statement need to wait all split/scatter region finish. IOW, `scope == scatteOff` indicate split region will not block the `CREATE TABLE` statement return.

> the `CREATE TABLE` statement returns success only after the Regions are successfully scattered.
refer: https://docs.pingcap.com/tidb/stable/system-variables#tidb_scatter_region

pkg/ddl/table_split_test.go
River2000i: In this case, the reason for setting `set @@session.tidb_scatter_region = 1` is waiting split region finish before create table statement return. However, we do indeed need to add a test case at the global level.

pkg/executor/test/showtest/show_test.go
```diff
@@ -807,7 +807,7 @@ func TestAutoRandomWithLargeSignedShowTableRegions(t *testing.T) {
 	tk.MustExec("drop table if exists t;")
 
 	tk.MustExec("create table t (a bigint unsigned auto_random primary key clustered);")
-	tk.MustExec("set @@global.tidb_scatter_region=1;")
+	tk.MustExec("set @@session.tidb_scatter_region=1;")
```
Benjamin2037, line 810: Why here still use "1"?

pkg/session/session.go
Benjamin2037: I say the enum type value is only use in package level, here we still use string directly.

pkg/ddl/partition.go
River2000i: 1. The DDL statements that trigger split region need to wait for the scatter region to complete. 
2. yes, 'on' equal 'scatterTable' here, i will add some comment.

pkg/executor/test/showtest/show_test.go
River2000i: since using enum indicate three options for `tidb_scatter_region` .
https://github.com/pingcap/tidb/blob/5986baca1e4d75b7d7959fa03a1f0f91cbc4e8c0/pkg/sessionctx/variable/sysvar.go#L778
1. `0` equal `""`
2. `1` equal `"table"`
3. `2` equal `"global"`

suggesting use `"table"` instead of `1`?

Benjamin2037: I know，what I mean is since we use enum, so can we only accept string input? 0,1,2 is not good for understand.

pkg/ddl/split_region.go
```diff
@@ -30,42 +30,64 @@ import (
...
+	if scope != scatterOff {
 		WaitScatterRegionFinish(ctxWithTimeout, store, regionIDs...)
 	}
 }
 
-func preSplitPhysicalTableByShardRowID(ctx context.Context, store kv.SplittableStore, tbInfo *model.TableInfo, physicalID int64, scatter bool) []uint64 {
+// PD controls the scope of scatter region through `tID`.
+// 1. nil means PD scatter region at cluster level.
+// 2. not nil means PD scatter region at table level when `scatter` is true.
+func getScatterConfig(scope scatterScope, tableID int64) (scatter bool, tID *int64) {
```
GMHDBJD, line 79: maybe use 0/-1 instead of pointer of int64?

Benjamin2037 (approved):
LGTM

pkg/ddl/table_split_test.go
```diff
@@ -65,6 +73,46 @@ func TestTableSplit(t *testing.T) {
...
+	tbl, err = infoSchema.TableByName(context.Background(), model.NewCIStr("test"), model.NewCIStr("t_part_2"))
+	require.NoError(t, err)
+	pi = tbl.Meta().GetPartitionInfo()
+	require.NotNil(t, pi)
+	for _, def := range pi.Definitions {
+		checkRegionStartWithTableID(t, def.ID, store.(kvStore))
+	}
+}
+
+func TestScatterRegion(t *testing.T) {
```
Benjamin2037, line 85: Please add comment to describe the testcase target

wjhuang2016 (approved):

benmeadowcroft (approved):

pkg/ddl/split_region.go
```diff
@@ -107,20 +121,21 @@ func preSplitPhysicalTableByShardRowID(ctx context.Context, store kv.SplittableS
...
 	return regionIDs
 }
 
 // SplitRecordRegion is to split region in store by table prefix.
-func SplitRecordRegion(ctx context.Context, store kv.SplittableStore, physicalTableID, tableID int64, scatter bool) uint64 {
+func SplitRecordRegion(ctx context.Context, store kv.SplittableStore, physicalTableID, tableID int64, scatterScope string) uint64 {
 	tableStartKey := tablecodec.GenTablePrefix(physicalTableID)
-	regionIDs, err := store.SplitRegions(ctx, [][]byte{tableStartKey}, scatter, &tableID)
+	scatter, tID := getScatterConfig(scatterScope, tableID)
+	regionIDs, err := store.SplitRegions(ctx, [][]byte{tableStartKey}, scatter, &tID)
```
GMHDBJD, line 138: so -1 is ok or it should be nil here?

GMHDBJD (commented):
rest LGTM

pkg/ddl/split_region.go
River2000i: Both is ok. In client-go, the `tableID` will convert from *int64 to string. If it's nil, it will equal `"0"`. The -1 is used to distinguish it from the default setting of 0 in client-go.
refer: https://github.com/tikv/client-go/blob/6ba909c4ad2de65b5b36d0e5036d0a85f3154cc0/tikv/split_region.go#L245-L248

GMHDBJD (approved):
LGTM

PR #56168, opened 2024-09-19
expression: extend locale support or format()

### LINKED ISSUE #56167: Support more locales for FORMAT()
## Enhancement

The `FORMAT(x,d,locale)` function now only supports `en_US`. Would be good to extend the support for other locales.

https://github.com/pingcap/tidb/blob/c13eb90fcf0ef3d4fe0fdb32851458651f7bc0a2/pkg/expression/builtin_string.go#L3366-L3386

dveeden:

Issue Number: close #56167 

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [x] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
The FORMAT() function now supports more locales
```


kennytm:
:eyes: i'm afraid the test cases is too simple, since `format()` is supposed to handle thousands separators as well so you should at least test

```sql
mysql> select format('12345678.9', 3, 'en_US');
12,345,678.900
mysql> select format('12345678.9', 3, 'nl_NL');
12345678,900
mysql> select format('12345678.9', 3, 'id_ID');
12.345.678,900
mysql> select format('12345678.9', 3, 'it_CH'); -- Switzerland
12'345'678,900
mysql> select format('12345678.9', 3, 'de_CH'); -- Switzerland
12'345'678.900
mysql> select format('12345678.9', 3, 'en_IN'); -- India
1,23,45,678.900
```

pkg/expression/builtin_string.go
```diff
@@ -3377,11 +3380,17 @@ func (b *builtinFormatWithLocaleSig) evalString(ctx EvalContext, row chunk.Row)
...
 	}
-	locale = "en_US"
-	formatString, err := mysql.GetLocaleFormatFunction(locale)(x, d)
+
+	lang, err := language.Parse(locale)
+	if err != nil {
+		return "", false, err
+	}
+	p := message.NewPrinter(lang)
+	xint, _ := strconv.ParseFloat(x, 64)
```
lcwangchao, line 3390: Why ignore error here? I think it's better to handle error even if we suppose it should not happen

tests/integrationtest/r/expression/format.result
```diff
@@ -16,47 +16,56 @@ FORMAT('1.234',1,'xxxxxx')
...
+12,345,678.9000
+SELECT FORMAT(12345678.9,4,'ps_AF');
+FORMAT(12345678.9,4,'ps_AF')
+12.345.678,9000
+SELECT FORMAT(12345678.9,4,'kv_RU');
+FORMAT(12345678.9,4,'kv_RU')
+12,345,678.9000
+SELECT FORMAT(12345678.9,999999999999999999,'de_CH');
+FORMAT(12345678.9,999999999999999999,'de_CH')
+12’345’678.900000000372529029846191406250
```
kennytm, line 54: why will it end with ....372529029846191406250? the input should be a decimal number with exact decimal representation :thinking: 

```
mysql> select FORMAT(12345678.9, 99999999999999999, 'de_CH');
+------------------------------------------------+
| FORMAT(12345678.9, 99999999999999999, 'de_CH') |
+------------------------------------------------+
| 12'345'678.900000000000000000000000000000      |
+------------------------------------------------+
1 row in set (0.00 sec)
```

also why is it separated using `’` U+2019 RIGHT SINGLE QUOTATION MARK rather than `'` U+0027 APOSTROPHE? is this OS-dependent?

dveeden: # Go

```go
package main

import (
	"golang.org/x/text/language"
	"golang.org/x/text/message"
	"golang.org/x/text/number"
)

func main() {
	p := message.NewPrinter(language.MustParse("de_CH"))
	p.Printf("%f", number.Decimal(12345678.9, number.Scale(1)))
}

```
https://go.dev/play/p/7Iv3SA0R-bR

output:
```
12’345’678.9
```

# C with ICU

```c
#include <stdio.h>
#include <unicode/unum.h>
#include <unicode/uloc.h>
#include <unicode/ustring.h>
#include <unicode/ustdio.h>

int main() {
    double number = 12345678.9;
    const char* locale = "de_CH";
    UChar result[100];

    UErrorCode status = U_ZERO_ERROR;
    UNumberFormat* fmt = unum_open(UNUM_DECIMAL, NULL, 0, locale, NULL, &status);
    int32_t resultLength = sizeof(result) / sizeof(result[0]);
    resultLength = unum_formatDouble(fmt, number, result, resultLength, NULL, &status);
    char resultUTF8[100];
    u_strToUTF8(resultUTF8, sizeof(resultUTF8), NULL, result, resultLength, &status);
    printf("%s\n", resultUTF8);
    unum_close(fmt);
    return 0;
}
```

output:
```
12’345’678.9
```

# Python

```python
#!/bin/python3
import locale

locale.setlocale(locale.LC_NUMERIC, "de_CH")
print(locale.format_string("%.2f", 12345678.9, grouping=True))
```

output:
```
12'345'678.90
```

# Conclusion

Depending on the OS, programming language and libraries either `’` or `'` are used. Looks like both are acceptable.

Note that this results in minor differences between TiDB and MySQL (tested with 9.0.1):

| locale | TiDB | MySQL | Difference? |
|---|---|---|---|
|nl_BE|12.345.678,9000|12345678,9000|Yes|
|fr_BE|12 345 678,9000|12345678,9000|Yes|
|en_IN|1,23,45,678.9000|1,23,45,678.9000|No|
|de_CH|12’345’678.9000|12'345'678.9000|Yes|
|no_NO|12,345,678.9000|12.345.678,9000|Yes (big difference)|
|bg_BG|12 345 678,9000|12 345 678,9000|No|
|ar_SA|12,345,678.9000|12345678.9000|Yes|
|es_MX|12,345,678.9000|12,345,678.9000|No|

dveeden: The `no_NO` difference is probably not a real issue as it should have been `nb_NO` or `nn_NO` instead of `no_NO`. See also https://bugzilla.redhat.com/show_bug.cgi?id=532487#c0

dveeden:
I'm going to part this pull request for now. The reason for this is that `golang.org/x/text` can do nice formatting for a large number of locales, but it only accepts integer and float types, which cause incorrect results due to how floats are handled.

I've filed a proposal to allow direct conversion from a string instead of relying on first converting the value to a float:
https://github.com/golang/go/issues/69740

The options we have now are:
1. Wait for golang.org/x/text improvements
2. Do the locale formatting without an external library. This is quite a bit of work to do it correctly. However the benefit is that we more easily follow the output of MySQL more closely.
3. Find another library that does this

cc @kennytm @lcwangchao @YangKeao

Issue #56271, opened 2024-09-24
potential memory leak when modify password 


tiancaiamao:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

I create 2M users, and then call "alter user 'xxx' identified by 'yyy'" for every user. 

### 2. What did you expect to see? (Required)

No memory leak.

### 3. What did you see instead (Required)

The memory usage is abnormal

![image](https://github.com/user-attachments/assets/2dc1433c-50a3-452b-bfec-2ac0ef9ef649)

I add a internal sessions metrics and it shows that the internal session leaks:

![image](https://github.com/user-attachments/assets/505534b7-74a9-4789-9a8b-34eac47dfd7b)

Profile also shows the `ExecuteInternal` takes a lot inuse memory:

![image](https://github.com/user-attachments/assets/2cf7ff59-2422-4112-b414-922c07b57670)

### 4. What is your TiDB version? (Required)

master ... this should affect a lot versions.

PR #56272, opened 2024-09-24
server,metrics: add a metrics for internal session count

tiancaiamao:

Issue Number: ref #56271 

Problem Summary:

If we get some internal session leak, this new added panel can catch it.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

image

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1055,23 +1055,19 @@ func (s *Server) ServerID() uint64 {
 // StoreInternalSession implements SessionManager interface.
 // @param addr	The address of a session.session struct variable
 func (s *Server) StoreInternalSession(se any) {
-	var count int
 	s.sessionMapMutex.Lock()
 	s.internalSessions[se] = struct{}{}
-	count = len(s.internalSessions)
+	metrics.InternalSessions.Set(float64(len(s.internalSessions)))
 	s.sessionMapMutex.Unlock()
-	metrics.InternalSessions.Set(float64(count))
 }
 
 // DeleteInternalSession implements SessionManager interface.
 // @param addr	The address of a session.session struct variable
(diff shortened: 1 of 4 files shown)


pkg/server/server.go
```diff
@@ -1055,17 +1055,23 @@ func (s *Server) ServerID() uint64 {
 // StoreInternalSession implements SessionManager interface.
 // @param addr	The address of a session.session struct variable
 func (s *Server) StoreInternalSession(se any) {
+	var count int
 	s.sessionMapMutex.Lock()
 	s.internalSessions[se] = struct{}{}
+	count = len(s.internalSessions)
 	s.sessionMapMutex.Unlock()
+	metrics.InternalSessions.Set(float64(count))
```
lance6716, line 1063: I hope we can move it before `Unlock`. So that the last executed `InternalSessions.Set` is always the latest value guarded by `sessionMapMutex` when `StoreInternalSession` is concurrency called.

tiancaiamao: It's not need to be very accurate, at the cost of performance.
Allocation after lock / or visit network after lock ... are all costly.  We can avoid those as much as possible.

This value is constantly changing, and prometheus update the data every 15s (if I remember correctly)
So the accuracy does not make much difference -- you can get hundreds or thousands values in the 15s interval, and all of them are discarded (not collected)

@lance6716

hawkingrei (approved):

pkg/server/server.go
lance6716: It's just an atomic operation, I don't think it will affect performance.

lance6716 (approved):

PR #56292, opened 2024-09-25
*: Add sysvar `tidb_tso_client_rpc_mode` for setting TSO RPC concurrency

### LINKED ISSUE #54960: TSO Request Parallelizing
## Enhancement

- Corresponding issue in PD repo:  https://github.com/tikv/pd/issues/8432

We used to find that in some OLTP workloads where the QPS is high and the queries are simple, the TSO Wait duration usually become a significant portion of the total duration of queries. In TiDB, TSO loading is already made concurrent with some other works such as compiling. In cases that the queries are simple, it would be hard to further optimize it by making it concurrent with more phases of the SQL execution. But we found a practical way to optimize it is to do it from the TSO client.

Currently, a TSO client object has a goroutine that collects `GetTS` (and `GetTSAsync`) calls (`tsoRequest`s) as a batch, send it to PD, wait for the response, and dispatch the results to these `tsoRequests`, serially. As a result, each `GetTS` calls may need to spend up to 1x TSO RPC time to wait for being collected to the next batch.

Considering the case that PD's TSO allocator is not the bottle neck and can deal with more TSO requests (so that the majority part of TSO RPC's time cost is on the network), we find that it's possible to start collecting the next batch and send it before receiving the response of the previous batch. So that each `GetTS` call needs to wait for less time to be batched, and gets a shorter total duration.

So this is an approach that reduces the duration of `GetTS` & `GetTSAsync - Wait` at the expense of higher TSO RPC OPS and higher pressure to PD. It's not suitable to be enabled by default, but we can provide such an option when the TSO Wait duration becomes a problem.

- [ ] Implementation of TSO request parallelizing of PDClient in PD repo
    - Ref: https://github.com/tikv/pd/issues/8432
- [ ] Add system variable for configuring the optimization
    - https://github.com/pingcap/tidb/pull/56292

MyonKeminta:

Issue Number: close #54960

Requires: https://github.com/tikv/pd/pull/8633

Problem Summary:

* Update pd client to support concurrent TSO RPC
* Add sysvar `tidb_tso_client_rpc_mode` for setting TSO RPC concurrency
* Add some metrics to the grafana dashboard that would be helpful for concurrent TSO RPC.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Add support for multiple TSO RPC requests to be done in parallel. `tidb_tso_client_rpc_mode` is now available for enabling this feature, to reduce GetTS latency, at the expense of higher ops of TSO RPC.
```


joey-yez (approved):

crazycs520 (approved):

ekexium (approved):

yudongusa (approved):
LGTM

tangenta (approved):

PR #56382, opened 2024-09-27
*: Allow Point_Get during DDL with Global Index

### LINKED ISSUE #56819: Row not found for Global Index during DeleteOnly state of Reorganize Partition
## Bug Report

Apply this patch:
```diff
diff --git a/pkg/ddl/tests/partition/multi_domain_test.go b/pkg/ddl/tests/partition/multi_domain_test.go
index 1c567c9a4a..3a17df219d 100644
--- a/pkg/ddl/tests/partition/multi_domain_test.go
+++ b/pkg/ddl/tests/partition/multi_domain_test.go
@@ -27,6 +27,28 @@ import (
        "go.uber.org/zap"
 )
 
+func TestMultiSchemaReorganizePartition(t *testing.T) {
+       createSQL := `create table t (a int primary key, b varchar(255), unique index idx_b_global (b) global) partition by range (a) (partition p1 values less than (200), partition pMax values less than (maxvalue))`
+       initFn := func(tkO *testkit.TestKit) {
+               tkO.MustExec(`insert into t values (1,1),(2,2)`)
+       }
+       alterSQL := `alter table t reorganize partition p1 into (partition p0 values less than (100), partition p1 values less than (200))`
+       loopFn := func(tkO, tkNO *testkit.TestKit) {
+               res := tkO.MustQuery(`select schema_state from information_schema.DDL_JOBS where table_name = 't' order by job_id desc limit 1`)
+               schemaState := res.Rows()[0][0].(string)
+               switch schemaState {
+               case model.StateDeleteOnly.String():
+                       tkNO.MustExec(`insert into t values (4,4)`)
+                       tkNO.MustQuery(`select * from t where b = "4"`).Sort().Check(testkit.Rows("4 4"))
+                       tkO.MustQuery(`select * from t where b = "4"`).Sort().Check(testkit.Rows("4 4"))
+               }
+       }
+       postFn := func(tkO *testkit.TestKit) {
+               // nothing
+       }
+       runMultiSchemaTest(t, createSQL, alterSQL, initFn, postFn, loopFn)
+}
+
 func TestMultiSchemaDropRangePartition(t *testing.T) {
        createSQL := `create table t (a int primary key, b varchar(255)) partition by range (a) (partition p0 values less than (100), partition p1 values less than (200))`
        initFn := func(tkO *testkit.TestKit) {
```

use `make failpoint-enable` and run the test.

test passing

Read from the second SELECT did not return anything! I.e. the row "4 4" could not be found!

```
tidb_version(): Release Version: v8.4.0-alpha-463-ged9a909395
Edition: Community
Git Commit Hash: ed9a9093954f6636f8d3c9d09686069626e5b5ce
Git Branch: master
UTC Build Time: 2024-10-24 13:20:54
GoVersion: go1.23.2
Race Enabled: false
Store: unistore
```

mjonss:

Issue Number: ref #55819 , ref #45133 , close #56819

Problem Summary:
During DDL that affects Global Index, the PointGet access method is blocked, which may cause temporary bad query plans during partitioning management commands, like DROP/TRUNCATE/REORGANIZE PARTITION.

Allow PointGet, by filter away the partitions that should not be seen (like when the cleanup of the global index is ongoing).

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/meta/model/table.go
```diff
@@ -867,6 +872,69 @@ func (pi *PartitionInfo) SetOriginalPartitionIDs() {
 	pi.OriginalPartitionIDsOrder = ids
 }
 
+// IDsInDDLToIgnore returns a list of IDs that the current
+// session should not see (may be duplicate errors on insert/update though)
+// For example during truncate or drop partition.
+func (pi *PartitionInfo) IDsInDDLToIgnore() []int64 {
+	// TODO:
```
mjonss, line 879: There are multiple PRs/enhancements that will build upon this, especially for DROP and TRUNCATE partition.

mjonss:
Splitted this part from https://github.com/pingcap/tidb/issues/55819 since it can be reviewed separately.

pkg/planner/core/find_best_task.go
```diff
@@ -2264,26 +2254,33 @@ func (is *PhysicalIndexScan) addSelectionConditionForGlobalIndex(p *logicalop.Da
...
 		}
 	} else if len(idxArr) == 0 {
-		// add an invalid pid as param for `IN` function
+		// TODO: Can we change to Table Dual somehow?
+		// Add an invalid pid as param for `IN` function
 		args = append(args, expression.NewInt64Const(-1))
 	} else {
-		// `PartitionPruning`` func does not return adding and dropping partitions
+		// TODO: When PartitionPruning is guaranteed to not
+		// return old/blocked partition ids then ignoreMap can be removed.
```
Defined2014, line 2269: Does it will happen? If the answer is yes, please add a related test case for it.

pkg/planner/core/find_best_task.go
```diff
@@ -2264,26 +2254,33 @@ func (is *PhysicalIndexScan) addSelectionConditionForGlobalIndex(p *logicalop.Da
...
+		}
 		for _, idx := range idxArr {
-			args = append(args, expression.NewInt64Const(pInfo.Definitions[idx].ID))
+			id := pInfo.Definitions[idx].ID
+			if _, ok := ignoreMap[id]; !ok {
+				args = append(args, expression.NewInt64Const(id))
+			}
 		}
 	}
+	if len(args) == 1 {
```
Defined2014, line 2281: I think it only happens in the first if-branch, maybe we could move into it.

pkg/ddl/partition.go
```diff
@@ -2265,13 +2266,16 @@ func (w *worker) onDropTablePartition(jobCtx *jobContext, t *meta.Mutator, job *
 		}
 
 		job.SchemaState = model.StateDeleteOnly
-		ver, err = updateVersionAndTableInfo(jobCtx, t, job, tblInfo, originalState != job.SchemaState)
+		tblInfo.Partition.DDLState = job.SchemaState
+		tblInfo.Partition.DDLAction = job.Type
+		ver, err = updateVersionAndTableInfo(jobCtx, t, job, tblInfo, true)
```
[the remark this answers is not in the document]
mjonss: It will always be true, since `originalState` is set to `job.SchemaState` before the switch, and is changed in each of the switch cases before the call. And it should update the Schema Version for each of the state changes.

pkg/planner/core/find_best_task.go
```diff
@@ -1466,16 +1466,6 @@ func findBestTask4LogicalDataSource(lp base.LogicalPlan, prop *property.Physical
 			if canConvertPointGet && path.IsIntHandlePath && !ds.Table.Meta().PKIsHandle && len(ds.PartitionNames) != 1 {
 				canConvertPointGet = false
 			}
-			if canConvertPointGet {
-				if path != nil && path.Index != nil && path.Index.Global {
-					// Don't convert to point get during ddl
-					// TODO: Revisit truncate partition and global index
```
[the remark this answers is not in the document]
mjonss: Since it may have used partitions that should not have been seen, when it comes to global index, since the global index is unique it can be used for Point Get, and during DDL there might be old entries in the Global Index pointing to dropped partitions or new entries pointing to partitions being added or replacing old ones that are not yet seen by all sessions. Meaning a full table scan may see different data from a Global Index read.

pkg/planner/core/find_best_task.go
mjonss: I'll add a test, I think there may be possible cases, until DROP/TRUNCATE/ADD PARTITION are fully updated for Global Index. Like https://github.com/pingcap/tidb/pull/55831 and https://github.com/pingcap/tidb/pull/56082.

pkg/planner/core/find_best_task.go
mjonss: If one truncates all partitions or partition pruning don't find any matching partition, it could also happen in last if-branch.

pkg/planner/core/find_best_task.go
Defined2014: I think `idxArr` will not contained any paritition in `AddingDefinitions` and `DroppingDefinitions`.

mjonss: You are probably right, but I prefer to have all partitioning DDLs checked and tested for Global Index consistency and proper visibility for each state combination. So until then I prefer to have this check here. Or do you prefer to add an intest-assert in PartitionPruning instead?

Defined2014: Add an intest-assert is better for me.

pkg/executor/point_get.go
```diff
@@ -371,9 +371,15 @@ func (e *PointGetExecutor) Next(ctx context.Context, req *chunk.Chunk) error {
...
-				if !matchPartitionNames(tblID, e.partitionNames, e.tblInfo.GetPartitionInfo()) {
+				pi := e.tblInfo.GetPartitionInfo()
+				if !matchPartitionNames(tblID, e.partitionNames, pi) {
 					return nil
 				}
+				for _, id := range pi.IDsInDDLToIgnore() {
+					if id == pid {
+						return nil
+					}
+				}
```
qw4990, line 382: It seems better to converge all partition logic together into the `PartitionProcessor`.

pkg/planner/core/find_best_task.go
Defined2014: If partition pruning don't find any matching partition, it will in the second if-branch.

And I don't know what is `one truncates all partitions` meaning.

pkg/ddl/partition.go
```diff
@@ -3435,10 +3446,10 @@ func (w *worker) onReorganizePartition(jobCtx *jobContext, job *model.Job) (ver
 				tblInfo.Indices[i].State = model.StateWriteReorganization
 			}
 		}
-		tblInfo.Partition.DDLState = model.StateWriteReorganization
+		job.SchemaState = model.StateWriteReorganization
+		tblInfo.Partition.DDLState = job.SchemaState
 		metrics.GetBackfillProgressByLabel(metrics.LblReorgPartition, job.SchemaName, tblInfo.Name.String()).Set(0.3 / float64(math.MaxUint64))
 		ver, err = updateVersionAndTableInfo(jobCtx, job, tblInfo, true)
-		job.SchemaState = model.StateWriteReorganization
```
Defined2014, line 3441: Does this line caused the https://github.com/pingcap/tidb/issues/56819?

pkg/planner/core/find_best_task.go
mjonss: The problem was that the previous logic was not correct I think. Always ignoring both `AddingDefinitions` and `DroppingDefinitions` would miss entries during REORGANIZE PARTITION for example (first if-branch).
Now it will always check `IDsInDDLToIgnore()` and if there are no IDs to ignore, then `args` will have len 1, and we should just keep the conditions as-is.

pkg/meta/model/table.go
```diff
@@ -856,6 +856,8 @@ func (pi *PartitionInfo) ClearReorgIntermediateInfo() {
 	pi.DDLExpr = ""
 	pi.DDLColumns = nil
 	pi.NewTableID = 0
+	pi.DDLState = StateNone
```
Defined2014, line 859: Dup with L853-L854

Defined2014 (approved):

pkg/planner/core/find_best_task.go
```diff
@@ -2278,26 +2268,35 @@ func (is *PhysicalIndexScan) addSelectionConditionForGlobalIndex(p *logicalop.Da
...
+		for _, id := range pInfo.IDsInDDLToIgnore() {
+			ignoreMap[id] = struct{}{}
+		}
 		for _, idx := range idxArr {
-			args = append(args, expression.NewInt64Const(pInfo.Definitions[idx].ID))
+			id := pInfo.Definitions[idx].ID
+			if _, ok := ignoreMap[id]; !ok {
+				args = append(args, expression.NewInt64Const(id))
+			} else if intest.InTest {
+				panic("PartitionPruning returns partitions which should be ignored!")
```
winoros, line 2293: use `intest.Assert` instead since `intest.Assert` will work in not only unit test but also integration test.

winoros (approved):

tangenta (approved):

PR #56485, opened 2024-10-08
linter, executor: fix the `constructor` linter and add it to `BaseExecutor/BaseExecutorV2`

### LINKED ISSUE #56484: Fix the `constructor` linter
## Enhancement

The `constructor` linter is not valid after migrating the packages from `/util/linter/constructor` to `/pkg/util/linter/constructor`. We need to fix it.

YangKeao:

Issue Number: close #56484

Problem Summary:

1. This PR adds the `constructor` annotation to the struct `BaseExecutor/BaseExecutorV2` to avoid constructing the struct unexpectedly.
2. This PR fixes the invalid `constructor` linter, and correct some behaviors.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


build/linter/constructor/analyzer_test.go
```diff
@@ -28,6 +28,6 @@ import (
 
 func Test(t *testing.T) {
 	testdata := analysistest.TestData()
-	pkgs := []string{"t", "github.com/pingcap/tidb/util/linter/constructor"}
+	pkgs := []string{"t", "github.com/pingcap/tidb/pkg/util/linter/constructor"}
```
lance6716, line 31: maybe declare a const for the package prefix to reduce duplication. And maybe use some runtime reflections to ensure it's updated after the folder is changed.

build/linter/constructor/analyzer.go
```diff
@@ -67,7 +67,12 @@ func getConstructorList(t types.Type) []string {
 		if !ok {
 			continue
 		}
-		if named.Obj().Name() == "Constructor" && named.Obj().Pkg().Path() == "github.com/pingcap/tidb/util/linter/constructor" {
+		if ignoreFields != nil {
+			if _, ok := ignoreFields[field.Name()]; ok {
```
lance6716, line 71: should we don't recursively check the fields in line 90 instead of adding `ignoreFields` argument? Not sure `(*inspector.Inspector).WithStack` still can do a recursively check.

yibin87 (approved):
LGTM

build/linter/constructor/analyzer_test.go
YangKeao: Done.

build/linter/constructor/analyzer.go
YangKeao: `(*inspector.Inspector).WithStack` will do the recursively check. I've added the test case:

```
	// specified field with manually constructed struct is not allowed
	_ = compositeImplicitInitiate1{
		StructWithSpecificConstructor: StructWithSpecificConstructor{}, // want `struct can only be constructed in constructors NewStructWithSpecificConstructor, AnotherConstructor`
	}
```

Even though the `StructWithSpecificConstructor` is ignored, constructing `StructWithSpecificConstructor{}` is still not allowed.

build/linter/constructor/analyzer.go
```diff
@@ -82,7 +90,7 @@ func getConstructorList(t types.Type) []string {
 		}
 
 		if fieldStruct, ok := named.Underlying().(*types.Struct); ok {
-			ctors = append(ctors, getConstructorList(fieldStruct)...)
+			ctors = append(ctors, getConstructorList(fieldStruct, nil)...)
```
lance6716, line 93: Let me clarify. In this PR you added `ignoreFields` to `getConstructorList` to skip some fields when recursively process `struct`. And I think this line is the cause of recursively processing `struct`. We don't need this line because `WithStack` still access the member of `struct`. Can we delete this line and delete the `ignoreFields` logic of `getConstructorList`?

YangKeao: `getConstructorList` should be called recursively to avoid implicitly constructing the embedded structs:

```go
type A struct { _ constructor.Constructor `ctor:"NewA"`}
type B struct {A}
type C struct {B}

var c C
```

The `var c C` is not allowed, because it will implicitly create a `A` outside of its constructor. The `WithStack` can do nothing for it as it doesn't appear in the AST.

The `ignoreFields` is added in this PR to allow this situation:

```go
func XFunc() {
    b := B {
        A: NewA()
    }
}
```

Preivously, the `getConstructorList()` of `B` will return `NewA`, and linter will give an error because it's constructed in `XFunc`. However, it should be allowed as the `A` is actually not constructed in `XFunc`, but in `NewA`.

hawkingrei (approved):

PR #56516, opened 2024-10-09
ttl: use a pessimistic transaction to finish the job

### LINKED ISSUE #56422: TTL Task will not be canceled if facing write conflict
## Bug Report

See the following function:

```
// finish turns current job into last job, and update the error message and statistics summary
func (job *ttlJob) finish(se session.Session, now time.Time, summary *TTLSummary) {
	// at this time, the job.ctx may have been canceled (to cancel this job)
	// even when it's canceled, we'll need to update the states, so use another context
	err := se.RunInTxn(context.TODO(), func() error {
		sql, args := finishJobSQL(job.tbl.ID, now, summary.SummaryText, job.id)
...
		sql, args = removeTaskForJob(job.id)
...
		sql, args = finishJobHistorySQL(job.id, now, summary)
,,,
		return nil
	}, session.TxnModeOptimistic)

	if err != nil {
		logutil.BgLogger().Error("fail to finish a ttl job", zap.Error(err), zap.Int64("tableID", job.tbl.ID), zap.String("jobID", job.id))
	}
}
```

It'll update records in three tables: `tidb_ttl_table_status`, `tidb_ttl_task` and `tidb_ttl_job_history`. It's usually fine to write to `tidb_ttl_table_status` and `tidb_ttl_job_history`, because the `job_manager ` always holds the owner and runs as a single goroutine, but it may have conflict to write to `tidb_ttl_task`.

When it faces a write conflict error, the job will be forgot by the job manager, and nobody will update the status, clean the tasks and update the history again.

The `DoGC` didn't work because it cleans the tasks using the following SQL:

```sql
DELETE task FROM
		mysql.tidb_ttl_task task
	left join
		mysql.tidb_ttl_table_status job
	ON task.job_id = job.current_job_id
	WHERE job.table_id IS NULL
```

If the `tidb_ttl_table_status` is not updated, the `tidb_ttl_task` will also not be cleaned.

Therefore, I propose to execute this transaction in pessimistic mode. (Actually, it's not necessary to execute them in a single transaction. They are safe to be executed one-by-one.)

YangKeao:

Issue Number: close #56422

Problem Summary:

When the job is finished, and at the same time the task is assigned to some TiDB node, it may fail to finish the job.

Use the pessimistic transaction to finish job. Then the `DELETE ... FROM tidb_ttl_tasks` will block `UPDATE mysql.tidb_ttl_task SET owner_id = ?...` (or in another direction).

If the ttl task is removed, it'll finally be removed by `(m *taskManager) checkInvalidTask` in the memory.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that a TTL job cannot be finished in some situation.
```

### DIFF
--- a/pkg/ttl/ttlworker/job.go
+++ b/pkg/ttl/ttlworker/job.go
@@ -20,11 +20,10 @@ import (
 	"time"
 
 	"github.com/pingcap/errors"
+	"github.com/pingcap/failpoint"
 	"github.com/pingcap/tidb/pkg/ttl/cache"
 	"github.com/pingcap/tidb/pkg/ttl/session"
 	"github.com/pingcap/tidb/pkg/util/intest"
-	"github.com/pingcap/tidb/pkg/util/logutil"
-	"go.uber.org/zap"
 )
 
 const finishJobTemplate = `UPDATE mysql.tidb_ttl_table_status
@@ -122,8 +121,9 @@ type ttlJob struct {
 }
 
 // finish turns current job into last job, and update the error message and statistics summary
-func (job *ttlJob) finish(se session.Session, now time.Time, summary *TTLSummary) {
+func (job *ttlJob) finish(se session.Session, now time.Time, summary *TTLSummary) error {
 	intest.Assert(se.GetSessionVars().Location().String() == now.Location().String())
+
 	// at this time, the job.ctx may have been canceled (to cancel this job)
 	// even when it's canceled, we'll need to update the states, so use another context
 	err := se.RunInTxn(context.TODO(), func() error {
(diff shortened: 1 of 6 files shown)


pkg/ttl/ttlworker/job.go
```diff
@@ -146,8 +146,9 @@ func (job *ttlJob) finish(se session.Session, now time.Time, summary *TTLSummary
 		}
 
 		return nil
-	}, session.TxnModeOptimistic)
+	}, session.TxnModePessimistic)
```
lcwangchao, line 149: Tough it can reduce the issue but cannot fully fix it because some other cases can also cause the sql execution failed, i.e network failure. Should we also finish the job first before we remove it from memory, if the finish fails, it will try it again next time.

https://github.com/pingcap/tidb/blob/1e24d39608214cc463c1de913e9a308be1c8193c/pkg/ttl/ttlworker/job_manager.go#L561-L562

YangKeao: Done.

lcwangchao (approved):

hawkingrei (approved):

PR #56542, opened 2024-10-10
ddl: move insert job to table to job submitter

D3Hunter:

Issue Number: ref #54436

Problem Summary:

as title, they are only used there
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > - [x] I checked and no code files have been changed. only **move** code
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/job_scheduler.go
+++ b/pkg/ddl/job_scheduler.go
@@ -15,13 +15,11 @@
 package ddl
 
 import (
-	"bytes"
 	"context"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"runtime"
-	"slices"
 	"strconv"
 	"strings"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


lance6716 (approved):

pkg/ddl/job_scheduler.go
```diff
@@ -764,7 +666,7 @@ func updateDDLJob2Table(
 	if err != nil {
 		return err
 	}
-	sql := fmt.Sprintf(updateDDLJobSQL, util.WrapKey2String(b), job.ID)
+	sql := fmt.Sprintf("update mysql.tidb_ddl_job set job_meta = %s where job_id = %d", util.WrapKey2String(b), job.ID)
```
GMHDBJD, line 669: I think original updataDDLJobSQL variable would be better?

GMHDBJD (commented):
rest LGTM

pkg/ddl/job_scheduler.go
D3Hunter: the reason I inline it is nearly all other place use string literal directly, and it's only used once

I do agree to define it as const if it's or will be used multiple times, but seems we won't

GMHDBJD (approved):
LGTM

Issue #53930, opened 2024-06-11
improve the usability of DDL Job.Args


lance6716:
## Enhancement

Currently Job.Args is like a type-erasure field https://github.com/pingcap/tidb/blob/fedf3b7ba9a841e90de55cc3c4c1b037343f2339/pkg/parser/model/ddl.go#L514-L522
It's hard to find all usage of a DDL job type arguments and maintain them.

![image](https://github.com/pingcap/tidb/assets/1689766/ac9a756d-efa0-4949-bbf3-589ee14debaf)

idea 1: I think it's better to have a specialized `struct` for each DDL job type. So the developer can directly find all usage of a DDL job type among all types.

idea 2: don't change old format of the serialized Job.RawArgs, the new `struct`s should handle the serialization/deserialization in the old format. The advantage is we don't need to handle the compatibility of DDL jobs written by old version clusters.

# Tasks
- [x] https://github.com/pingcap/tidb/pull/55634
- [x] https://github.com/pingcap/tidb/pull/55754
- [x] https://github.com/pingcap/tidb/pull/55854
- [x] https://github.com/pingcap/tidb/pull/55919
- [x] https://github.com/pingcap/tidb/pull/55964
- [x] https://github.com/pingcap/tidb/pull/56153
- [x] https://github.com/pingcap/tidb/pull/56163
- [x] https://github.com/pingcap/tidb/pull/56219
- [x] https://github.com/pingcap/tidb/pull/56222
- [x] https://github.com/pingcap/tidb/pull/56242
- [x] https://github.com/pingcap/tidb/pull/56269
- [x] https://github.com/pingcap/tidb/pull/56400
- [x] https://github.com/pingcap/tidb/pull/56417
- [x] https://github.com/pingcap/tidb/pull/55946
- [x] https://github.com/pingcap/tidb/pull/56010
- [x] https://github.com/pingcap/tidb/pull/56130
- [x] https://github.com/pingcap/tidb/pull/56536
- [x] https://github.com/pingcap/tidb/pull/56186
- [x] https://github.com/pingcap/tidb/pull/56194
- [x] https://github.com/pingcap/tidb/pull/56290
- [x] https://github.com/pingcap/tidb/pull/55959
- [x] https://github.com/pingcap/tidb/pull/56021
- [x] https://github.com/pingcap/tidb/pull/56164
- [x] https://github.com/pingcap/tidb/pull/56206
- [x] https://github.com/pingcap/tidb/pull/56241
[...]

PR #56586, opened 2024-10-11
ddl: pass runtime info through jobContext for rename table(s)

D3Hunter:

Issue Number: ref #53930

Problem Summary:

as title
### Check List

Tests 

- [x] Unit test, existing cases should be enough to cover it
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/table.go
+++ b/pkg/ddl/table.go
@@ -16,7 +16,6 @@ package ddl
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strconv"
 	"sync/atomic"
@@ -747,6 +746,7 @@ func onRenameTable(jobCtx *jobContext, job *model.Job) (ver int64, _ error) {
 		job.State = model.JobStateCancelled
 		return ver, errors.Trace(err)
 	}
+	jobCtx.jobArgs = args
 
 	oldSchemaID, oldSchemaName, tableName := args.OldSchemaID, args.OldSchemaName, args.NewTableName
 	if job.SchemaState == model.StatePublic {
@@ -791,9 +791,10 @@ func onRenameTables(jobCtx *jobContext, job *model.Job) (ver int64, _ error) {
 		job.State = model.JobStateCancelled
 		return ver, errors.Trace(err)
 	}
+	jobCtx.jobArgs = args
 
 	if job.SchemaState == model.StatePublic {
-		return finishJobRenameTables(jobCtx, job, args)
+		return finishJobRenameTables(jobCtx, job)
 	}
 
 	fkh := newForeignKeyHelper()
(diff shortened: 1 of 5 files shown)


pkg/ddl/table.go
```diff
@@ -970,22 +969,13 @@ func finishJobRenameTables(jobCtx *jobContext, job *model.Job, args *model.Renam
...
-	var err error
-	oldRawArgs := job.RawArgs
 	for _, info := range infos {
-		info.OldSchemaID = info.NewSchemaID
-	}
-	job.FillArgs(args)
-	job.RawArgs, err = json.Marshal(job.Args)
-	if err != nil {
-		return 0, errors.Trace(err)
+		info.OldSchemaIDForSchemaDiff = info.NewSchemaID
```
lance6716, line 973: after reading the comment I still don't understand why we should set `OldSchemaIDForSchemaDiff` instead of `OldSchemaID`. What does the unchanged `OldSchemaID` use for?

D3Hunter: see this https://github.com/pingcap/tidb/pull/44585/files#diff-9dd5a0976146b94bb3074ca04e8131dffc03096505755440e64760b42bbde0fdR1181-R1183

that fix is backup RawArgs, change args and put into RawArgs, so later decode can see it, then restore

here we use OldSchemaIDForSchemaDiff to do the same thing by pass is directly

D3Hunter: the fix in #44585 should better be done in diff apply, not here, but I won't change it here

lance6716 (approved):

pkg/ddl/table.go
```diff
@@ -937,21 +938,19 @@ func finishJobRenameTable(jobCtx *jobContext, job *model.Job) (int64, error) {
...
 	if err != nil {
 		return ver, errors.Trace(err)
 	}
-	job.RawArgs = oldRawArgs
 	job.FinishTableJob(model.JobStateDone, model.StatePublic, ver, tblInfo)
 	return ver, nil
 }
 
-func finishJobRenameTables(jobCtx *jobContext, job *model.Job, args *model.RenameTablesArgs) (int64, error) {
+func finishJobRenameTables(jobCtx *jobContext, job *model.Job) (int64, error) {
```
joechenrh, line 952: `finishJobRenameTable` also contains same logic needed to be removed.

tangenta (approved):

pkg/ddl/table.go
tangenta: Maybe remove the comments above since job args are not changed?

D3Hunter: those comments explains why we do it, so I keep it

PR #56613, opened 2024-10-12
ddl: set a lower limit of regions for distributed local sort

### LINKED ISSUE #56501: Regions number of one add-index subtask should be limited
## Enhancement

If there are many TiDB instances, even if the task of adding indexes is small, TiDB will split at least one subtask for each TiDB instance. The result may bring heavy workload to TiKV and reduce the performance of the entire cluster.

We should set a lower limit for regions of each adding index subtask.

tangenta:

Issue Number: close #56501

Problem Summary: See #56501

```
// Assuming the region size is 96MiB and the index data size accounts for 1/10 of the row data.
// Row data take 960GiB for 10000 regions, and index data take 96GiB.
// This basically matches the default value of tidb_ddl_disk_quota.
// Estimates don't necessarily have to be accurate.
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


CbcWestwolf (approved):

pkg/util/dbterror/ddl_terror.go
```diff
@@ -531,4 +531,5 @@ var ReorgRetryableErrCodes = map[uint16]struct{}{
 var ReorgRetryableErrMsgs = []string{
 	"context deadline exceeded",
 	"requested lease not found",
+	"create TSO stream failed",
```
lance6716, line 534: I don't want to hide the problem for PD client 🤔 We have fixed at caller for many times and it delayed the real fix at PD client side.

lance6716 (approved):
/hold

Free to unhold

PR #56661, opened 2024-10-15
executor: fix the issue that `INSERT IGNORE` doesn't lock the parent table

### LINKED ISSUE #56655: `INSERT IGNORE` didn't lock the referenced row in parent table
## Bug Report

Prepare:

```sql
create table parent (id int primary key);
create table child(id int primary key, ref int, foreign key (ref) references parent(id));

insert into parent values (1);
```

Execute:

| Transaction 1    | Transaction 2 |
| -------- | ------- |
| BEGIN;  | BEGIN;    |
| INSERT IGNORE INTO child VALUES (1, 1); |     |
|     | UPDATE parent SET id = 2 WHERE id = 1    |
| COMMIT; | |
| |  COMMIT;|

The statement `UPDATE` in transaction 2 is blocked until the transaction 1 `COMMIT`, and it will finally return an error.

Both transactions are not blocked. The constraint is broken:

Then the constraint is not consistent:

```sql
mysql> select * from parent;
+----+
| id |
+----+
|  2 |
+----+
1 row in set (0.00 sec)

mysql> select * from child;
+----+------+
| id | ref  |
+----+------+
|  1 |    1 |
+----+------+
1 row in set (0.01 sec)
```

```
mysql> select tidb_version();
+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| tidb_version()                                                                                                                                                                                                  |
+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Release Version: v8.4.0-this-is-a-placeholder
Edition: Community
Git Commit Hash: None
Git Branch: None
UTC Build Time: None
GoVersion: go1.23.0
Race Enabled: false
Store: tikv |
+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

YangKeao:

Issue Number: close #56655 

Problem Summary:

`INSERT IGNORE` cannot lock the rows in parent, so the constraint may be broken.

Try to lock the rows even when the `toBeCheckedKeys` and `toBeCheckedPrefixKeys` are empty.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/foreign_key.go
+++ b/pkg/executor/foreign_key.go
@@ -223,7 +223,7 @@ func (fkc *FKCheckExec) doCheck(ctx context.Context) error {
 		fkc.stats = &FKCheckRuntimeStats{}
 		defer fkc.ctx.GetSessionVars().StmtCtx.RuntimeStatsColl.RegisterStats(fkc.ID(), fkc.stats)
 	}
-	if len(fkc.toBeCheckedKeys) == 0 && len(fkc.toBeCheckedPrefixKeys) == 0 {
+	if len(fkc.toBeCheckedKeys) == 0 && len(fkc.toBeCheckedPrefixKeys) == 0 && len(fkc.toBeLockedKeys) == 0 {
 		return nil
 	}
 	start := time.Now()
@@ -248,6 +248,7 @@ func (fkc *FKCheckExec) doCheck(ctx context.Context) error {
 	if fkc.stats != nil {
 		fkc.stats.Check = time.Since(start)
 	}
+
 	if len(fkc.toBeLockedKeys) == 0 {
 		return nil
 	}
@@ -542,7 +543,7 @@ type fkCheckKey struct {
 	isPrefix bool
 }
 
-func (fkc FKCheckExec) checkRows(ctx context.Context, sc *stmtctx.StatementContext, txn kv.Transaction, rows []toBeCheckedRow) error {
+func (fkc *FKCheckExec) checkRows(ctx context.Context, sc *stmtctx.StatementContext, txn kv.Transaction, rows []toBeCheckedRow) error {
 	if fkc.ctx.GetSessionVars().StmtCtx.RuntimeStatsColl != nil {
 		fkc.stats = &FKCheckRuntimeStats{}
 		defer fkc.ctx.GetSessionVars().StmtCtx.RuntimeStatsColl.RegisterStats(fkc.ID(), fkc.stats)
--- a/pkg/executor/test/fktest/BUILD.bazel
+++ b/pkg/executor/test/fktest/BUILD.bazel
@@ -8,7 +8,7 @@ go_test(
         "main_test.go",
     ],
     flaky = True,
-    shard_count = 26,
+    shard_count = 27,
     deps = [
         "//pkg/config",
         "//pkg/executor",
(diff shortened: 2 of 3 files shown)


tests/realtikvtest/txntest/txn_test.go
```diff
@@ -595,3 +595,42 @@ func TestDMLWithAddForeignKey(t *testing.T) {
 
 	require.True(t, errDML != nil || errDDL != nil)
 }
+
+func TestLockKeysInInsertIgnore(t *testing.T) {
+	// FIXME: this test will fail in the unistore. After fixing the issue in unistore, please move
+	// this test to `foreign_key_test.go`.
```
YangKeao, line 601: I still don't know why this case will fail in the unistore. Maybe the transaction implementation has some bugs (or just different behaviors).

I tried this case with/without `IGNORE`. It seems that the foreign key behaves strangely in unistore :thinking: :


Prepare:

```sql
create table parent (id int primary key);
create table child(id int primary key, ref int, foreign key (ref) references parent(id));

insert into parent values (1);
```

Execute:

| Transaction 1    | Transaction 2 |
| -------- | ------- |
| BEGIN;  | BEGIN;    |
| INSERT INTO child VALUES (1, 1); |     |
|     | UPDATE parent SET id = 2 WHERE id = 1    |
| COMMIT; | COMMIT; |

YangKeao: Maybe it should get a snapshot with `ForUpdateTS`?

YangKeao: I record this issue in https://github.com/pingcap/tidb/issues/56663.

YangKeao:
```
run test [planner/core/indexmerge_path] err: sql:explain select * from t use index (iad) where a = 1;: failed to run query 
"explain select * from t use index (iad) where a = 1;" 
 around line 233, 
we need(299):
explain select * from t use index (iad) where a = 1;
TableReader_7   1.00    root            data:Selection_6
└─Selection_6   1.00    cop[tikv]               eq(planner__core__indexmerge_path.t.a, 1)
  └─TableFullScan_5     2.00    cop[tikv]       table:t keep order:false
explain select * fro
but got(299):
explain select * from t use index (iad) where a = 1;
TableReader_7   10.00   root            data:Selection_6
└─Selection_6   10.00   cop[tikv]               eq(planner__core__indexmerge_path.t.a, 1)
  └─TableFullScan_5     10000.00        cop[tikv]       table:t keep order:false, stats:pseudo
```

This test failure cannot be reproduced locally :thinking: . Investigating.

pkg/executor/foreign_key.go
```diff
@@ -223,10 +223,19 @@ func (fkc *FKCheckExec) doCheck(ctx context.Context) error {
...
+		// It's possible that the `toBeLockedKeys` is not empty. Because the `INSERT IGNORE` will use
+		// `checkRows` to check each row, which doesn't need this function to check the rows again, but
+		// it still needs to lock the rows.
+		err := fkc.doLock(ctx)
+		if len(fkc.toBeLockedKeys) > 0 && fkc.stats != nil {
+			fkc.stats.Lock = time.Since(start)
+			fkc.stats.Total = fkc.stats.Lock
+		}
+
+		return err
```
crazycs520, line 237:
```suggestion
    if len(fkc.toBeCheckedKeys) == 0 && len(fkc.toBeCheckedPrefixKeys) == 0 && len(fkc.toBeLockedKeys) == 0 {
        return nil
    }
```

YangKeao: Done.

crazycs520 (approved):

tests/realtikvtest/txntest/txn_test.go
YangKeao: It'll be fixed in https://github.com/pingcap/tidb/pull/56821

bb7133 (approved):
LGTM

Issue #55563, opened 2024-08-21
Modify user or privilege will consume a significant amount of time and resources in million users scenario


lilinghai:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

1. Create 1 million databases with 2 tables per database.
2. Create 2 users for each schema and grant privileges on the schema to the users.

From the monitoring of TiDB, it can be observed that the QPS for creating users and granting permissions is continuously declining. When the number of users reaches 200,000, the QPS is nearly 1. 
TiDB maintains a full set of user and permission information in memory. When there are changes to users or permissions (triggered by statements such as create/alter user, grant, etc.), TiDB calls the NotifyUpdatePrivilege function which loads the full user and permission information from TiKV to refresh the cache. As the user information increases, this operation will become increasingly time-consuming and resource-intensive. Actual tests show CPU resource consumption at over 80%, and Golang GC causes memory fluctuations exceeding 10GB.

![image](https://github.com/user-attachments/assets/a2683437-44ef-4928-9d29-29a5a13b104e)
![image](https://github.com/user-attachments/assets/89607bea-93c7-4711-9646-a91813c83c87)
![image](https://github.com/user-attachments/assets/d554cef0-5cf9-4c1d-bc85-7ebc232d2bb2)

### 4. What is your TiDB version? (Required)

master

PR #56709, opened 2024-10-17
*: maintain the active users for privilege data

tiancaiamao:

Issue Number: ref #55563

Problem Summary:

Add an `activeUsers` field to the `privilege.Handler`.
This is a step towrads supporting 2.5M users, later we'll only keep the active users' data in-memory.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/privilege/privileges/cache.go
```diff
@@ -578,36 +761,41 @@ func (p *MySQLPrivilege) LoadDefaultRoles(ctx sessionctx.Context) error {
...
+func addUserFilterCondition(sql string, userList []string) string {
+	if len(userList) == 0 {
+		return sql
+	}
+	var b strings.Builder
+	b.WriteString(sql)
+	b.WriteString(" where ")
+	for i, user := range userList {
+		if i > 0 {
+			b.WriteString(" Or ")
```
lance6716, line 773: the letter case style is different with line 770

pkg/privilege/privileges/cache.go
```diff
@@ -578,36 +761,41 @@ func (p *MySQLPrivilege) LoadDefaultRoles(ctx sessionctx.Context) error {
...
+		}
+		fmt.Fprintf(&b, "User = '%s'", user)
```
lance6716, line 775: need to escape the `'` in username

pkg/privilege/privileges/cache.go
```diff
@@ -1633,12 +1815,48 @@ func (p *MySQLPrivilege) getAllRoles(user, host string) []*auth.RoleIdentity {
...
+	err := data.loadSomeUsers(h.sctx, user)
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	old := h.Get()
+	h.priv.Store(old.merge(&data))
+
+	h.mu.Lock()
+	h.mu.activeUsers[user] = struct{}{}
```
lance6716, line 1856: maybe the use case is more suitable for sync.Map, because the user key write once and read for multiple times, and user key accessing are disjoint.

pkg/privilege/privileges/privileges.go
```diff
@@ -905,13 +911,17 @@ func (p *UserPrivileges) ShowGrants(ctx sessionctx.Context, user *auth.UserIdent
...
 		return nil, ErrNonexistingGrant.GenWithStackByArgs("root", "%")
 	}
-	mysqlPrivilege := p.Handle.Get()
 	u := user.Username
 	h := user.Hostname
 	if len(user.AuthUsername) > 0 && len(user.AuthHostname) > 0 {
 		u = user.AuthUsername
 		h = user.AuthHostname
 	}
+	if err := p.Handle.ensureActiveUser(u); err != nil {
```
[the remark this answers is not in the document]
tiancaiamao: Currently there is no evict operation, evict make thing more complex.

pkg/privilege/privileges/cache.go
```diff
@@ -1633,12 +1815,48 @@ func (p *MySQLPrivilege) getAllRoles(user, host string) []*auth.RoleIdentity {
...
+	var priv MySQLPrivilege
+	ret := &Handle{}
+	ret.sctx = sctx
+	ret.priv.Store(&priv)
+	ret.mu.activeUsers = make(map[string]struct{})
+	return ret
+}
+
+// ensureActiveUser ensure that the specific user data is loaded in-memory.
+func (h *Handle) ensureActiveUser(user string) error {
```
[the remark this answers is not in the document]
tiancaiamao: During the data loading process, there is no special handling for case insensitive.
If there are some requirement for case insensitive, it's the usage of those data. @lance6716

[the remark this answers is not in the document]
tiancaiamao: `create user 'Someone';`

can we login as 'someone'?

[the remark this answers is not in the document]
lance6716: I'll check and create a separate issue for username case sensitive problem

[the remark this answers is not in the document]
lance6716: username is case sensetive. In MySQL

```
mysql> select @@lower_case_table_names;
+--------------------------+
| @@lower_case_table_names |
+--------------------------+
|                        2 |
+--------------------------+
1 row in set (0.00 sec)

mysql> CREATE USER 'AAA'@'%' IDENTIFIED BY '123456';
```

```
$ mysql --comments --host 127.0.0.1 --port 3306 -u aaa -p123456
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'aaa'@'localhost' (using password: YES)

# lance6716 @ lances-MacBook-Air in ~ [15:09:59] C:1
$ mysql --comments --host 127.0.0.1 --port 3306 -u AAA -p123456
mysql: [Warning] Using a password on the command line interface can be insecure.
Welcome to the MySQL monitor.  Commands end with ; or \g.
Your MySQL connection id is 10
Server version: 8.0.27 MySQL Community Server - GPL
```

So I think current behaviour is OK

pkg/privilege/privileges/cache.go
```diff
@@ -400,6 +410,121 @@ func (p *MySQLPrivilege) LoadAll(ctx sessionctx.Context) error {
...
+
+// merge construct a new MySQLPrivilege by merging the data of the two objects;.
+func (p *MySQLPrivilege) merge(diff *immutable) *MySQLPrivilege {
+	var ret MySQLPrivilege
+	ret.User = make([]UserRecord, 0, len(p.User)+len(diff.User))
+	ret.User = append(ret.User, p.User...)
+	ret.User = append(ret.User, diff.User...)
+
+	// sort and dedup
+	slices.SortStableFunc(ret.User, compareUserRecord)
```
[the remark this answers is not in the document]
tiancaiamao: 1. stable sort is required here, when we (append old + diff) and dedup, there would be duplicated items, we need to keep the newer one (diff), if it's not a stable sort, old one might be kept.
2. diff are not sorted. If it's sorted, we can use a merge-sort alike to sort + dedup at the same time. But I'm lazy and did not do that.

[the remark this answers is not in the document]
lance6716: What's the bad effect if old one is kept?

pkg/privilege/privileges/cache.go
```diff
@@ -400,6 +410,121 @@ func (p *MySQLPrivilege) LoadAll(ctx sessionctx.Context) error {
...
+	ret.GlobalPriv = make([]globalPrivRecord, 0, len(p.GlobalPriv)+len(diff.GlobalPriv))
+	ret.GlobalPriv = append(ret.GlobalPriv, p.GlobalPriv...)
+	ret.GlobalPriv = append(ret.GlobalPriv, diff.GlobalPriv...)
+	slices.SortStableFunc(ret.GlobalPriv, compareGlobalPrivRecord)
+	ret.GlobalPriv = dedupSorted(ret.GlobalPriv, func(x, y globalPrivRecord) bool {
+		return x.Host == y.Host && x.User == y.User
+	})
+	ret.buildGlobalMap()
+
+	ret.RoleGraph = diff.RoleGraph
```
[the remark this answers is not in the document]
tiancaiamao: RoleGraph is a full load, because I found that the table definitions differ from the others.
Other other tables has the User+Host column, while this table does not have it.

pkg/privilege/privileges/cache.go
tiancaiamao: done

pkg/privilege/privileges/cache.go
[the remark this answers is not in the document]
tiancaiamao: For example, we'll use `loadSomeUsers` for updating password operation later.
If (old + diff) without stable sort, the old password is kept. Then the user cannot login user the new password.

pkg/privilege/privileges/tidb_auth_token_test.go
```diff
@@ -417,3 +417,31 @@ func TestJWKSImpl(t *testing.T) {
 	_, err = jwksImpl.checkSigWithRetry(signedTokenString, 0)
 	require.Error(t, err)
 }
+
+func (i *immutable) User() []UserRecord {
```
lcwangchao, line 421: You can rename `i` to `p` to avoid the warning `Receiver names are different`

pkg/privilege/privileges/cache.go
```diff
@@ -1647,9 +1857,9 @@ func (h *Handle) Get() *MySQLPrivilege {
 }
 
 // Update loads all the privilege info from kv storage.
-func (h *Handle) Update(ctx sessionctx.Context) error {
+func (h *Handle) Update() error {
 	var priv MySQLPrivilege
-	err := priv.LoadAll(ctx)
+	err := priv.LoadAll(h.sctx)
```
lcwangchao, line 1862: So this PR still loads all users into memory and partial load will be implemented in other PR?

pkg/privilege/privileges/cache.go
```diff
@@ -260,6 +262,26 @@ func (g roleGraphEdgesTable) Find(user, host string) bool {
...
 }
 
+type immutable struct {
+	user         []UserRecord
+	db           []dbRecord
+	tablesPriv   []tablesPrivRecord
+	columnsPriv  []columnsPrivRecord
+	defaultRoles []defaultRoleRecord
+
+	globalPriv  []globalPrivRecord
```
lcwangchao, line 272: Why need to keep a immutable `globalPriv` and a mutable `Global`? Any problem to just keep one `Global` field?

pkg/privilege/privilege.go
```diff
@@ -88,7 +88,7 @@ type Manager interface {
 	GetAuthWithoutVerification(user, host string) bool
 
 	// MatchIdentity matches an identity
-	MatchIdentity(user, host string, skipNameResolve bool) (string, string, bool)
+	MatchIdentity(sctx sessionctx.Context, user, host string, skipNameResolve bool) (string, string, bool)
```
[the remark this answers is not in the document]
tiancaiamao: OK... removed.
This is used during the refactoring, and later I store sctx in Handle, so it's useless now.

pkg/privilege/privileges/tidb_auth_token_test.go
tiancaiamao: done

pkg/privilege/privileges/cache.go
```diff
@@ -1633,12 +1815,48 @@ func (p *MySQLPrivilege) getAllRoles(user, host string) []*auth.RoleIdentity {
 
 // Handle wraps MySQLPrivilege providing thread safe access.
 type Handle struct {
+	sctx sessionctx.Context
```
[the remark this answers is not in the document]
tiancaiamao: `sqlexec.RestrictedSQLExecutor` is not enough here.
We need something like a sqlexec.RestrictedSQLExecutor pool to avoid the single instance been used by multiple threads. @lcwangchao

pkg/privilege/privileges/cache.go
tiancaiamao: Yes, this is the first step and the following PR will handle the remains.

pkg/privilege/privileges/cache.go
tiancaiamao: Make the apply diff logic simpler if we're using one single slice representation... 
we can copy and paste a lot code and just handle them the same way.

It's possible to keep different representation for different field, or even using Map for almost all field, but we need to add more code.

pkg/privilege/privileges/cache.go
```diff
@@ -400,6 +411,121 @@ func (p *MySQLPrivilege) LoadAll(ctx sessionctx.Context) error {
...
+	p.roleGraph = make(map[string]roleGraphEdgesTable)
+	err = p.loadTable(ctx, sqlLoadRoleGraph, p.decodeRoleEdgesTable)
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	return nil
+}
+
+func dedupSorted[S ~[]E, E any](s S, eq func(a, b E) bool) S {
```
lance6716, line 459: maybe rename to dedupKeepLast

pkg/privilege/privileges/cache.go
```diff
@@ -1633,12 +1815,40 @@ func (p *MySQLPrivilege) getAllRoles(user, host string) []*auth.RoleIdentity {
...
+	}
+
+	var data immutable
+	err := data.loadSomeUsers(h.sctx, user)
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	old := h.Get()
+	h.priv.Store(old.merge(&data))
```
lance6716, line 1848: seems `ensureActiveUser` is concurrently used? For 2 concurrent call with different `user` and this timeline

1. user 1 is not active, then `loadSomeUsers` for it, then `h.Get()`
2. user 2 is not active, then `loadSomeUsers` for it, then `h.Get()`
3. `h.priv.Store` saved the merge result with user 1, and mark user 1 active
4. `h.priv.Store` saved the merge result with user 2 but data from user 1 is lost, and mark user 2 active

Now both users are active, but the the meaning of "ensure" is not achieved

pkg/privilege/privileges/cache.go
tiancaiamao: only works on sorted, so it's `dedupSortedKeepLast`

pkg/privilege/privileges/cache.go
tiancaiamao: Then we need either `lock` or `cas + version` semantic.

lance6716: Can we design a per-user `priv` instead of current all-users`priv` object?

pkg/privilege/privileges/cache.go
tiancaiamao: done

pkg/privilege/privileges/cache.go
tiancaiamao: Each session handle their own login user's privileges?

tiancaiamao: > Then we need either lock or cas + version semantic.

Now changed to CAS @lance6716

lcwangchao (approved):

pkg/privilege/privileges/cache.go
```diff
@@ -459,6 +586,40 @@ func compareUserRecord(x, y UserRecord) int {
...
+	return compareBaseRecord(&x.baseRecord, &y.baseRecord)
+}
+
+func compareColumnsPrivRecord(x, y columnsPrivRecord) int {
+	cmp := compareBaseRecord(&x.baseRecord, &y.baseRecord)
+	if cmp != 0 {
+		return cmp
+	}
+	switch {
+	case x.DB > y.DB:
```
lance6716, line 603: nit: can use `strings.Compare`

pkg/privilege/privileges/cache.go
```diff
@@ -976,9 +1159,9 @@ func patternMatch(str string, patChars, patTypes []byte) bool {
 
 // matchIdentity finds an identity to match a user + host
 // using the correct rules according to MySQL.
-func (p *MySQLPrivilege) matchIdentity(user, host string, skipNameResolve bool) *UserRecord {
-	for i := 0; i < len(p.User); i++ {
-		record := &p.User[i]
+func (p *MySQLPrivilege) matchIdentity(sctx sqlexec.RestrictedSQLExecutor, user, host string, skipNameResolve bool) *UserRecord {
+	for i := 0; i < len(p.user); i++ {
```
lance6716, line 1163: nit: `for i := range p.User`

pkg/privilege/privileges/cache.go
```diff
@@ -1011,8 +1194,8 @@ func (p *MySQLPrivilege) matchIdentity(user, host string, skipNameResolve bool)
 
 // matchResoureGroup finds an identity to match resource group.
 func (p *MySQLPrivilege) matchResoureGroup(resourceGroupName string) *UserRecord {
-	for i := 0; i < len(p.User); i++ {
-		record := &p.User[i]
+	for i := 0; i < len(p.user); i++ {
```
lance6716, line 1197: ditto

pkg/privilege/privileges/cache.go
```diff
@@ -1090,8 +1273,8 @@ func (p *MySQLPrivilege) matchTables(user, host, db, table string) *tablesPrivRe
 }
 
 func (p *MySQLPrivilege) matchColumns(user, host, db, table, column string) *columnsPrivRecord {
-	for i := 0; i < len(p.ColumnsPriv); i++ {
-		record := &p.ColumnsPriv[i]
+	for i := 0; i < len(p.columnsPriv); i++ {
```
lance6716, line 1276: ditto

lance6716 (approved):

lance6716:
It's OK to fix my comments in next Pr or ignore them @tiancaiamao

tiancaiamao:
This commit could make the login process slower, because now we need to load the users data.
And also some scenario like show grants for some other users.

The general DDL/DML cases should not be affected, this just affects a minority of privileges related.

PR #56712, opened 2024-10-18
br: restore checksum shouldn't rely on backup checksum

Tristan1900:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/restore/snap_client/client.go
+++ b/br/pkg/restore/snap_client/client.go
@@ -467,8 +467,8 @@ func (rc *SnapClient) needLoadSchemas(backupMeta *backuppb.BackupMeta) bool {
 	return !(backupMeta.IsRawKv || backupMeta.IsTxnKv)
 }
 
-// InitBackupMeta loads schemas from BackupMeta to initialize RestoreClient.
-func (rc *SnapClient) InitBackupMeta(
+// LoadSchemaIfNeededAndInitClient loads schemas from BackupMeta to initialize RestoreClient.
+func (rc *SnapClient) LoadSchemaIfNeededAndInitClient(
 	c context.Context,
 	backupMeta *backuppb.BackupMeta,
 	backend *backuppb.StorageBackend,
@@ -989,7 +989,7 @@ func (rc *SnapClient) setSpeedLimit(ctx context.Context, rateLimit uint64) error
 	return nil
 }
 
-func (rc *SnapClient) execChecksum(
+func (rc *SnapClient) execAndValidateChecksum(
 	ctx context.Context,
 	tbl *CreatedTable,
 	kvClient kv.Client,
@@ -1000,13 +1000,8 @@ func (rc *SnapClient) execChecksum(
 		zap.String("table", tbl.OldTable.Info.Name.O),
 	)
 
-	if tbl.OldTable.NoChecksum() {
-		logger.Warn("table has no checksum, skipping checksum")
-		return nil
-	}
-
 	if span := opentracing.SpanFromContext(ctx); span != nil && span.Tracer() != nil {
-		span1 := span.Tracer().StartSpan("Client.execChecksum", opentracing.ChildOf(span.Context()))
+		span1 := span.Tracer().StartSpan("Client.execAndValidateChecksum", opentracing.ChildOf(span.Context()))
 		defer span1.Finish()
 		ctx = opentracing.ContextWithSpan(ctx, span1)
 	}
@@ -1046,16 +1041,16 @@ func (rc *SnapClient) execChecksum(
 			}
 		}
 	}
-	table := tbl.OldTable
-	if item.Crc64xor != table.Crc64Xor ||
-		item.TotalKvs != table.TotalKvs ||
-		item.TotalBytes != table.TotalBytes {
+	expectedChecksumStats := metautil.CalculateChecksumStatsOnAllFiles(tbl.OldTable.Files)
+	if item.Crc64xor != expectedChecksumStats.Crc64Xor ||
+		item.TotalKvs != expectedChecksumStats.TotalKvs ||
+		item.TotalBytes != expectedChecksumStats.TotalBytes {
 		logger.Error("failed in validate checksum",
-			zap.Uint64("origin tidb crc64", table.Crc64Xor),
+			zap.Uint64("origin tidb crc64", expectedChecksumStats.Crc64Xor),
 			zap.Uint64("calculated crc64", item.Crc64xor),
-			zap.Uint64("origin tidb total kvs", table.TotalKvs),
+			zap.Uint64("origin tidb total kvs", expectedChecksumStats.TotalKvs),
 			zap.Uint64("calculated total kvs", item.TotalKvs),
-			zap.Uint64("origin tidb total bytes", table.TotalBytes),
+			zap.Uint64("origin tidb total bytes", expectedChecksumStats.TotalBytes),
 			zap.Uint64("calculated total bytes", item.TotalBytes),
 		)
 		return errors.Annotate(berrors.ErrRestoreChecksumMismatch, "failed to validate checksum")
--- a/br/pkg/task/restore.go
+++ b/br/pkg/task/restore.go
@@ -801,7 +801,7 @@ func runSnapshotRestore(c context.Context, mgr *conn.Mgr, g glue.Glue, cmdName s
 	}
 
 	reader := metautil.NewMetaReader(backupMeta, s, &cfg.CipherInfo)
-	if err = client.InitBackupMeta(c, backupMeta, u, reader, cfg.LoadStats); err != nil {
+	if err = client.LoadSchemaIfNeededAndInitClient(c, backupMeta, u, reader, cfg.LoadStats); err != nil {
 		return errors.Trace(err)
 	}
 
@@ -822,7 +822,7 @@ func runSnapshotRestore(c context.Context, mgr *conn.Mgr, g glue.Glue, cmdName s
 		}
 	}
 
-	archiveSize := reader.ArchiveSize(ctx, files)
+	archiveSize := metautil.ArchiveSize(files)
 	g.Record(summary.RestoreDataSize, archiveSize)
 	//restore from tidb will fetch a general Size issue https://github.com/pingcap/tidb/issues/27247
 	g.Record("Size", archiveSize)
@@ -1124,7 +1124,7 @@ func runSnapshotRestore(c context.Context, mgr *conn.Mgr, g glue.Glue, cmdName s
 
 	finish := dropToBlackhole(ctx, postHandleCh, errCh)
 
-	// Reset speed limit. ResetSpeedLimit must be called after client.InitBackupMeta has been called.
+	// Reset speed limit. ResetSpeedLimit must be called after client.LoadSchemaIfNeededAndInitClient has been called.
 	defer func() {
 		var resetErr error
 		// In future we may need a mechanism to set speed limit in ttl. like what we do in switchmode. TODO
--- a/br/pkg/backup/schema.go
+++ b/br/pkg/backup/schema.go
@@ -106,7 +106,7 @@ func (ss *Schemas) BackupSchemas(
 		}
 
 		var checksum *checkpoint.ChecksumItem
-		var exists bool = false
+		var exists = false
 		if ss.checkpointChecksum != nil && schema.tableInfo != nil {
 			checksum, exists = ss.checkpointChecksum[schema.tableInfo.ID]
 		}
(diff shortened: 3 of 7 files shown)


br/tests/br_file_corruption/run.sh
```diff
@@ -22,33 +22,62 @@ CUR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
...
+# Replace the single file manipulation with a loop over all .sst files
+for filename in $(find $TEST_DIR/$DB -name "*.sst"); do
+    filename_temp="${filename}_temp"
+    filename_bak="${filename}_bak"
+    echo "corruption" > "$filename_temp"
+    cat "$filename" >> "$filename_temp"
+    mv "$filename" "$filename_bak"
+done
+
+# need to drop db otherwise restore will fail because of cluster not fresh but not the expected issue
```
Tristan1900, line 36: previous br_file_corruption is not testing what it should be testing, restore fail because of cluster not fresh but not because of corruption.

br/tests/br_file_corruption/run.sh
```diff
@@ -22,33 +22,62 @@ CUR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
...
-run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB"
+run_br --pd $PD_ADDR backup full -s "local://$TEST_DIR/$DB" --checksum=false
 
-filename=$(find $TEST_DIR/$DB -regex ".*.sst" | head -n 1)
-filename_temp=$filename"_temp"
-filename_bak=$filename"_bak"
-echo "corruption" > $filename_temp
-cat $filename >> $filename_temp
+# Replace the single file manipulation with a loop over all .sst files
+for filename in $(find $TEST_DIR/$DB -name "*.sst"); do
```
Tristan1900, line 28: need to corrupt every sst file cuz during restore not all sst are going to be used, since backup backs up a bunch of tables but some of them are going to be filtered out during restore

br/tests/br_file_corruption/run.sh
```diff
@@ -22,33 +22,62 @@ CUR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
...
     exit 1
 fi
+run_sql "DROP DATABASE IF EXISTS $DB;"
 
 # file corruption
-mv $filename_temp $filename
-truncate --size=-11 $filename
+for filename in $(find $TEST_DIR/$DB -name "*.sst_temp"); do
+    mv "$filename" "${filename%_temp}"
+    truncate -s 11 "${filename%_temp}"
```
Tristan1900, line 53: --size=-11 doesn't work on macOS

br/pkg/restore/snap_client/client.go
```diff
@@ -1000,13 +1000,14 @@ func (rc *SnapClient) execChecksum(
 		zap.String("table", tbl.OldTable.Info.Name.O),
 	)
 
-	if tbl.OldTable.NoChecksum() {
-		logger.Warn("table has no checksum, skipping checksum")
+	expectedChecksumStats := metautil.CalculateChecksumStatsOnFiles(tbl.OldTable.Files)
+	if !expectedChecksumStats.ChecksumExists() {
```
Tristan1900, line 1004: not sure if we need to do the check here, if an empty table is being backed up, should we check after restore it's still empty?

br/tests/br_file_corruption/run.sh
Leavrth:
```suggestion
    truncate -s -11 "${filename%_temp}"
```

br/pkg/restore/snap_client/client.go
Leavrth: Yes, the error log is misleading.

br/pkg/task/backup.go
```diff
@@ -800,6 +800,13 @@ func DefaultBackupConfig() BackupConfig {
 	if err != nil {
 		log.Panic("infallible operation failed.", zap.Error(err))
 	}
+
+	// Check if the checksum flag was set by the user
+	if !fs.Changed("checksum") {
```
Leavrth, line 805: use `flagChecksum` instead

br/pkg/task/restore.go
```diff
@@ -1108,8 +1108,9 @@ func runSnapshotRestore(c context.Context, mgr *conn.Mgr, g glue.Glue, cmdName s
 	errCh := make(chan error, 32)
 	postHandleCh := afterTableRestoredCh(ctx, createdTables)
 
-	// pipeline checksum
-	if cfg.Checksum {
+	// pipeline checksum only when enabled and is not incremental snapshot repair mode cuz incremental doesn't have
+	// enough information in backup meta to validate checksum
```
Leavrth, line 1112: we don't use the collection of `file-level` checksums to match the `table-level` checksum in the incremental backup. However, we use the `table-level` checksum in the incremental backup to match the `table-level` checksum in the incremental restore.

Leavrth (commented):
rest LGTM

br/tests/br_file_corruption/run.sh
Tristan1900: hmm, -11 doesn't work on my laptop

br/pkg/task/backup.go
Tristan1900: oh right... my mistake

br/pkg/task/restore.go
Tristan1900: could you point me to the code, I thought table level checksum for incremental is not calculated during backup 
```
skipChecksum := !cfg.Checksum || isIncrementalBackup
```

Leavrth: Get it.

br/pkg/task/backup_test.go
```diff
@@ -222,3 +223,27 @@ func TestBackupConfigHash(t *testing.T) {
...
+	// Check some default values
+	require.Equal(t, uint32(4), cfg.Concurrency)
+	require.Equal(t, uint32(2), cfg.ChecksumConcurrency)
+	require.False(t, cfg.SendCreds)
+	require.False(t, cfg.Checksum)
+
+	// Test with checksum flag set
+	os.Args = []string{"cmd", "--checksum=true"}
+	cfg = DefaultBackupConfig()
+	require.True(t, cfg.Checksum)
```
YuJuncen, line 240: I don't think the `DefaultBackupConfig` will parse command line from `os.Args`... TBH I'm even not sure how this case passes...

br/tests/br_file_corruption/run.sh
```diff
@@ -22,33 +22,62 @@ CUR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
...
+for filename in $(find $TEST_DIR/$DB -name "*.sst_bak"); do
+    mv "$filename" "${filename%_bak}"
+done
+
+export GO_FAILPOINTS="github.com/pingcap/tidb/br/pkg/restore/snap_client/full-restore-validate-checksum=return(true)"
+restore_fail=0
+run_br --pd $PD_ADDR restore full -s "local://$TEST_DIR/$DB" --checksum=true || restore_fail=1
+export GO_FAILPOINTS=""
+if [ $restore_fail -ne 1 ]; then
+    echo 'expect restore to fail on checksum mismatch but succeed'
```
YuJuncen, line 76: nit: I guess if we have a corrupted SST file, the restoration fails though with `--checksum=false`... Perhaps a better (also harder) way is to delete an `file` entry in the backupmeta.

YuJuncen (commented):
rest lgtm

br/pkg/restore/snap_client/client.go
```diff
@@ -1000,13 +1000,14 @@ func (rc *SnapClient) execChecksum(
...
+		logger.Error("table has no checksum, skipping checksum")
```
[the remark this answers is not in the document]
Tristan1900: right, I think this line above adds the db and table as fields for all subsequent logging
```
	logger := log.L().With(
		zap.String("db", tbl.OldTable.DB.Name.O),
		zap.String("table", tbl.OldTable.Info.Name.O),
	)
```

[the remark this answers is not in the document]
Tristan1900: I'm a bit skeptical about this logic, should we just remove it? If there is an empty table restored(would there be a case?), we should still check the checksum make sure it's empty. What do you think?

br/pkg/restore/snap_client/client.go
```diff
@@ -1046,21 +1047,24 @@ func (rc *SnapClient) execChecksum(
...
+			zap.Uint64("expected tidb total kvs", expectedChecksumStats.TotalKvs),
 			zap.Uint64("calculated total kvs", item.TotalKvs),
-			zap.Uint64("origin tidb total bytes", table.TotalBytes),
+			zap.Uint64("expected tidb total bytes", expectedChecksumStats.TotalBytes),
 			zap.Uint64("calculated total bytes", item.TotalBytes),
 		)
 		return errors.Annotate(berrors.ErrRestoreChecksumMismatch, "failed to validate checksum")
 	}
-	logger.Info("success in validate checksum")
+	logger.Info("success in validating checksum")
```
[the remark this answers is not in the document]
Tristan1900: same as above

br/pkg/task/backup_test.go
Tristan1900: yeah you are absolutely right, should revisit before making PR read for review. It did pass locally and on CI which is weird.

br/tests/br_file_corruption/run.sh
Tristan1900: oh it's not testing corrupted SST file in this test, the valid files are moved back after the previous corruption test, This test is just to verify the checksum process is running even backup disables the checksum, by injecting the fail point into the checksum routine. Right after this test there is a sanity test that can restore successfully, verifying all files are valid.

br/pkg/task/backup.go
Tristan1900: I feel like having override in `DefineBackupFlags` is a bit weird since it's technically not `define`, might be a bit confusing

br/pkg/task/backup_test.go
Leavrth: That's because tests in `br/pkg/task` is skipped on CI.
https://github.com/pingcap/tidb/blob/c550aae6a125e20c2aaa8d4f11a2f090bc8b2b50/Makefile#L627

br/pkg/task/backup.go
```diff
@@ -800,6 +800,10 @@ func DefaultBackupConfig() BackupConfig {
 	if err != nil {
 		log.Panic("infallible operation failed.", zap.Error(err))
 	}
+
+	// disable checksum for backup by default
+	cfg.Checksum = false
```
Leavrth, line 805: It seems that the function is only called by BRIE via SQL, but it will be replaced by another config.
https://github.com/pingcap/tidb/blob/c550aae6a125e20c2aaa8d4f11a2f090bc8b2b50/pkg/executor/brie.go#L388

pkg/executor/brie.go
```diff
@@ -283,7 +283,15 @@ func (b *executorBuilder) buildBRIE(s *ast.BRIEStmt, schema *expression.Schema)
 		Key:  tidbCfg.Security.ClusterSSLKey,
 	}
 	pds := strings.Split(tidbCfg.Path, ",")
+
+	// build common config and override for specific task if needed
 	cfg := task.DefaultConfig()
+	switch s.Kind {
+	case ast.BRIEKindBackup:
+		cfg.OverrideDefaultForBackup()
```
Leavrth, line 291: Will this make `cfg.Checksum` be always false for backup?

Leavrth (commented):
rest LGTM

pkg/executor/brie.go
Tristan1900: good question, the actual user param will be parsed later in this method so overriding the default if needed. 

```
	for _, opt := range s.Options {
		switch opt.Tp {
		case ast.BRIEOptionRateLimit:
			cfg.RateLimit = opt.UintValue
		case ast.BRIEOptionConcurrency:
			cfg.Concurrency = uint32(opt.UintValue)
		case ast.BRIEOptionChecksum:
			cfg.Checksum = opt.UintValue != 0
		case ast.BRIEOptionSendCreds:
			cfg.SendCreds = opt.UintValue != 0
		case ast.BRIEOptionChecksumConcurrency:
			cfg.ChecksumConcurrency = uint(opt.UintValue)
		case ast.BRIEOptionEncryptionKeyFile:
			cfg.CipherInfo.CipherKey, err = task.GetCipherKeyContent("", opt.StrValue)
			if err != nil {
				b.err = err
				return nil
			}
```

Leavrth (approved):

YuJuncen (approved):

PR #56769, opened 2024-10-22
planner: fix nulleq condition to calcalate index range

### LINKED ISSUE #56745: Error 1105 (HY000): interface conversion: expression.Expression is *expression.Constant, not *expression.ScalarFunction
## Bug Report

```
create table lrr( `COL1` varchar(10) NOT NULL,`COL2` char(10) NOT NULL,PRIMARY KEY (`COL1`(5),`COL2`) /*T![clustered_index] CLUSTERED */);
insert into lrr values("","a");
insert into lrr values("test","b");
prepare stmt from 'SELECT * FROM lrr t1 JOIN lrr t2 ON t1.col1 <=> t2.col1 WHERE t1.col1 <=> NULL AND t2.col1 = ?;';
set @a=NULL;
execute stmt using @a;
```

execute success

`ERROR 1105 (HY000): interface conversion: expression.Expression is *expression.Constant, not *expression.ScalarFunction
`

mysql execute success

Release Version: v8.4.0
Edition: Community
Git Commit Hash: 2205f332ffcf3553ffc49438cfe1087a0f08dd18
Git Branch: HEAD
UTC Build Time: 2024-10-17 07:36:44
GoVersion: go1.23.2
Race Enabled: false
Store: tikv

hawkingrei:

Issue Number: close #56745

Problem Summary:

treat col<=>null as range scan instead of point get to avoid incorrect results. so we find it in the ```getPotentialEqOrInColOffset``` and should return ```-1```. But this condition is wrong.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


winoros (approved):

AilinKid (approved):

AilinKid:
in points2EqOrInCond, when we convert points to acess condition, we will build isnull func again from a single point range with null value, when we find the ref col is with not null flag, it will output zero constant which breaks the function check outside.  That's why we abandon the nulleq range detecting, treat it as non-eq-in condition for later range build.

hawkingrei:
> better add a commen: in points2EqOrInCond, when we convert points to acess condition reversely, we will build isnull func again from a single point range with null value, when we find the ref col is with not null flag, it will output zero constant which breaks the function check outside. That's why we abandon the nulleq range detecting, treat it as non-eq-in condition for later range build.

Added.

PR #56771, opened 2024-10-22
lightning: unlimited grpc MaxCallRecvMsgSize for TikvImporter

### LINKED ISSUE #56114: lightning report "received message larger than max (18552234 vs. 4194304)" during duplicate resolution
## Bug Report

currently, we use default max-message-size which is 4M, see

https://github.com/pingcap/tidb/blob/bda79ca6f4a6a0371332a9fd8585a4ecce1916ab/pkg/lightning/backend/local/local.go#L144

no error
err
8.1.0

fishiu:

Issue Number: close #56114

Problem Summary: Currently, we comply with the default max-message-size=4MiB, but TiKVImporter may receive max 4k KV on each response, and each KV can be about 4KiB+, which exceeds the default limit easily.

Set grpc options `MaxCallRecvMsgSize` to math.MaxInt32 (2GiB), i.e., unlimited in most scenarios. This number also aligns with the default setting for MaxCall**Send**MsgSize in gRPC.

MaxCallRecvMsgSize is only for resource control and there will not be any extra memory consumption. Please refer to the discussions below.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


lance6716:
Can you explain why choose 32 MB, is it enough?

Or can we choose the value after we get the error message like DM?

https://github.com/pingcap/tiflow/blob/76ebb371bb629cc8ca64c04694f985f927634acf/dm/ctl/common/util.go#L131-L139

fishiu:
> Can you explain why choose 32 MB, is it enough?
> 
> Or can we choose the value after we get the error message like DM?
> 
> https://github.com/pingcap/tiflow/blob/76ebb371bb629cc8ca64c04694f985f927634acf/dm/ctl/common/util.go#L131-L139

32 MB is not an precise size, I chose it only refering to the size mentioned in the issue. Or does @D3Hunter have any idea about the upper bounds in real scenarios?

The way in DM is better. Bust I first need to locate places that use this client connection and wrap the recv process with adjust-and-retry logics.

fishiu:
> Can you explain why choose 32 MB, is it enough?
> 
> Or can we choose the value after we get the error message like DM?
> 
> https://github.com/pingcap/tiflow/blob/76ebb371bb629cc8ca64c04694f985f927634acf/dm/ctl/common/util.go#L131-L139

I inspected the grpc source code regarding MaxCallRecvMsgSize and concluded that this option is only for limiting (memory/bandwidth) resource usage, i.e. setting MaxCallRecvMsgSize to large number will not directly consume large amount of memory.

Therefore, the strategy to capture the error message and then retry with adjusted options seems to be equivalent with directly setting it to intMax initially.

Additionally, the MaxCallRecvMsgSize is still an experimental option (for many years though) in grpc-go, so currently I do not think it is necessary to add it into lightning's config. Initializing it as maxInt would be simple and clear.

D3Hunter:
> > Can you explain why choose 32 MB, is it enough?
> > Or can we choose the value after we get the error message like DM?
> > https://github.com/pingcap/tiflow/blob/76ebb371bb629cc8ca64c04694f985f927634acf/dm/ctl/common/util.go#L131-L139
> 
> I inspected the grpc source code regarding [MaxCallRecvMsgSize](https://github.com/grpc/grpc-go/blob/d66fc3a1efa1dfb33dfedf9760528f1ac2b923b6/rpc_util.go#L670C1-L677C36) and concluded that this option is only a numeric check to limit (memory/bandwidth) resource usage, i.e. setting MaxCallRecvMsgSize to large number will not directly consume large amount of memory. Actually in contrast, MaxCallSendMsgSize is by default maxInt32 in grpc.
> 
> Therefore, the strategy to capture the error message and then retry with adjusted options seems to be equivalent with directly setting it to intMax initially.
> 
> Additionally, the MaxCallRecvMsgSize is still an experimental option (for many years though) in grpc-go, so currently I do not think it is necessary to add it into lightning's config. Initializing it as maxInt would be simple and clear.

nice insight

Benjamin2037 (approved):
LGTM

pkg/lightning/backend/local/local.go
```diff
@@ -165,6 +168,7 @@ func (f *importClientFactoryImpl) makeConn(ctx context.Context, storeID uint64)
 		addr = store.GetAddress()
 	}
 	opts = append(opts,
+		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(unlimitedRPCRecvMsgSize)),
```
D3Hunter, line 171: have you test it manually?

fishiu: I ran and passed the lightning_duplicate_detection integration test in local, which covered the modified code.

D3Hunter: I mean to reproduce a scenario in the issue that have many duplicates and TiKV send a grpc message larger that some size, say 32MiB

fishiu: No. But I have tried to reduce MaxCallRecvMsgSize e.g. 200KB which did trigger the expected error.

I will try to compose a new integration test that will send 4MiB+ KV back.

D3Hunter: o, that should be enough, please also update this in the PR description

D3Hunter (approved):

PR #56786, opened 2024-10-22
ddl: Corrected index management during REORGANIZE PARTITION

### LINKED ISSUE #56634: Panic after inject an error in onReorganizePartition in `StateDeleteReorganization` state
## Bug Report

 Ref: https://github.com/pingcap/tidb/compare/master...fzzf678:tidb:test_partition?expand=1

1. inject a error an `StateDeleteReorganization` state, mock error for `updateVersionAndTableInfo`
2. run sql like: alter table t partition by range(a) (partition p1 values less than (20))
no panic
panic
```
[2024/10/14 16:03:13.188 +08:00] [ERROR] [misc.go:115] ["panic in the recoverable goroutine"] [label=domain] [funcInfo=loadSchemaInLoop] [r="runtime error: index out of range [0] with length 0"] [stack="github.com/pingcap/tidb/pkg/util.Recover\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/util/misc.go:119\nruntime.gopanic\n\t/Users/fanzhou/.gvm/pkgsets/go1.22.4/global/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.0.darwin-arm64/src/runtime/panic.go:785\nruntime.goPanicIndex\n\t/Users/fanzhou/.gvm/pkgsets/go1.22.4/global/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.0.darwin-arm64/src/runtime/panic.go:115\ngithub.com/pingcap/tidb/pkg/table/tables.getRangeLocateExprs\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/table/tables/partition.go:754\ngithub.com/pingcap/tidb/pkg/table/tables.generateRangePartitionExpr\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/table/tables/partition.go:719\ngithub.com/pingcap/tidb/pkg/table/tables.newPartitionExpr\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/table/tables/partition.go:280\ngithub.com/pingcap/tidb/pkg/table/tables.newPartitionedTable\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/table/tables/partition.go:158\ngithub.com/pingcap/tidb/pkg/table/tables.TableFromMeta\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/table/tables/tables.go:226\ngithub.com/pingcap/tidb/pkg/infoschema.tableFromMeta\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/infoschema/builder.go:916\ngithub.com/pingcap/tidb/pkg/infoschema.(*Builder).createSchemaTablesForDB\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/infoschema/builder.go:945\ngithub.com/pingcap/tidb/pkg/infoschema.(*Builder).InitWithDBInfos\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/infoschema/builder.go:896\ngithub.com/pingcap/tidb/pkg/domain.(*Domain).loadInfoSchema\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/domain/domain.go:386\ngithub.com/pingcap/tidb/pkg/domain.(*Domain).Reload\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/domain/domain.go:766\ngithub.com/pingcap/tidb/pkg/domain.(*Domain).loadSchemaInLoop\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/domain/domain.go:1104\ngithub.com/pingcap/tidb/pkg/domain.(*Domain).Start.func2\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/domain/domain.go:1495\ngithub.com/pingcap/tidb/pkg/util.(*WaitGroupEnhancedWrapper).Run.func1\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/util/wait_group_wrapper.go:99"] [stack="github.com/pingcap/tidb/pkg/util.Recover\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/util/misc.go:115\nruntime.gopanic\n\t/Users/fanzhou/.gvm/pkgsets/go1.22.4/global/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.0.darwin-arm64/src/runtime/panic.go:785\nruntime.goPanicIndex\n\t/Users/fanzhou/.gvm/pkgsets/go1.22.4/global/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.0.darwin-arm64/src/runtime/panic.go:115\ngithub.com/pingcap/tidb/pkg/table/tables.getRangeLocateExprs\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/table/tables/partition.go:754\ngithub.com/pingcap/tidb/pkg/table/tables.generateRangePartitionExpr\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/table/tables/partition.go:719\ngithub.com/pingcap/tidb/pkg/table/tables.newPartitionExpr\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/table/tables/partition.go:280\ngithub.com/pingcap/tidb/pkg/table/tables.newPartitionedTable\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/table/tables/partition.go:158\ngithub.com/pingcap/tidb/pkg/table/tables.TableFromMeta\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/table/tables/tables.go:226\ngithub.com/pingcap/tidb/pkg/infoschema.tableFromMeta\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/infoschema/builder.go:916\ngithub.com/pingcap/tidb/pkg/infoschema.(*Builder).createSchemaTablesForDB\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/infoschema/builder.go:945\ngithub.com/pingcap/tidb/pkg/infoschema.(*Builder).InitWithDBInfos\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/infoschema/builder.go:896\ngithub.com/pingcap/tidb/pkg/domain.(*Domain).loadInfoSchema\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/domain/domain.go:386\ngithub.com/pingcap/tidb/pkg/domain.(*Domain).Reload\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/domain/domain.go:766\ngithub.com/pingcap/tidb/pkg/domain.(*Domain).loadSchemaInLoop\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/domain/domain.go:1104\ngithub.com/pingcap/tidb/pkg/domain.(*Domain).Start.func2\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/domain/domain.go:1495\ngithub.com/pingcap/tidb/pkg/util.(*WaitGroupEnhancedWrapper).Run.func1\n\t/Users/fanzhou/Documents/GitHub/tidb/pkg/util/wait_group_wrapper.go:99"]
```
master

mjonss:

Issue Number: close #56634

Problem Summary:
issue triggered by injected error, resulting in rollback and panic

Uncovered issues:
- Since Global Indexes needs to be replaced (copied with new references to the new partition ids, since we cannot reference two different partitions in the same unique index entry) both indexes were used during StateWriteOnly, StateWriteReorganization and StateDeleteReorganization, which would cause duplicate entry errors when trying to insert new values during those states for Global Indexes.

- Avoiding to use StateDeleteReorganization as index state, since during that state, the global indexes may still be read or and should also be kept up-to-date in case of rollback.
- Remove indexes that are not connected to the partition, like for new partitions it should not update indexes that will be removed, those indexes will be kept up-to-date by the old partitions. So double writing, including indexes, will be on partition level, where the data matching:
  - the non-touched partitions should update:
    - all original indexes
    - new indexes
    - So it needs to include all indexes!
  - the old/reorganized partitions should update:
    - all original indexes
    - So it needs to skip new indexes!
  - the new partitions should update:
    - all non-changed indexes
    - new indexes
    - So it needs to skip old indexes that will be replaced with new indexes (All old Global Indexes, and all indexes that will be converted to Global indexes).

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fixed an issue where REORGANIZE PARTITION block inserts during some states, as well as it could fail rollback, in case of internal errors.
```


pkg/ddl/tests/partition/multi_domain_test.go
```diff
@@ -249,16 +249,81 @@ func TestMultiSchemaDropListColumnsDefaultPartition(t *testing.T) {
...
+	createSQL := `create table t (a int primary key, b varchar(255)) partition by range (a) (partition p1 values less than (200))`
+	initFn := func(tkO *testkit.TestKit) {
+		tkO.MustExec(`insert into t values (1,1),(2,2),(101,101),(102,102)`)
+	}
+	alterSQL := `alter table t reorganize partition p1 into (partition p0 values less than (100), partition p1 values less than (200))`
+	loopFn := func(tkO, tkNO *testkit.TestKit) {
+		res := tkO.MustQuery(`select schema_state from information_schema.DDL_JOBS where table_name = 't' order by job_id desc limit 1`)
+		schemaState := res.Rows()[0][0].(string)
+		switch schemaState {
+		case "write only":
```
Defined2014, line 262: Use `StateWriteOnly.String()` maybe better.

Defined2014 (commented):
> resulting in they would not continue to be updated, so in case of a rollback there might be missing entries.

Can you clarify which sql in the newly added test case covers this problem? And what's the result before this PR?

mjonss:
> > resulting in they would not continue to be updated, so in case of a rollback there might be missing entries.
> 
> Can you clarify which sql in the newly added test case covers this problem? And what's the result before this PR?

Since the indexes was in StateDeleteReorganization, during `job.SchemaState == model.StateDeleteReorganization`, the indexes would not been written, see [`IsIndexWritable()`](https://github.com/pingcap/tidb/blob/48c57ce4996ba04ad08f9edafab0706d2c04f05f/pkg/table/tables/index.go#L679-L686). Which would mean that during normal DDL that is OK, but if an error would happen during StateDeleteReorganization and a rollback would happen, than changes to the indexes to be dropped would miss records written during that state.

Defined2014:
> > > resulting in they would not continue to be updated, so in case of a rollback there might be missing entries.
> > 
> > 
> > Can you clarify which sql in the newly added test case covers this problem? And what's the result before this PR?
> 
> Since the indexes was in StateDeleteReorganization, during `job.SchemaState == model.StateDeleteReorganization`, the indexes would not been written, see [`IsIndexWritable()`](https://github.com/pingcap/tidb/blob/48c57ce4996ba04ad08f9edafab0706d2c04f05f/pkg/table/tables/index.go#L679-L686). Which would mean that during normal DDL that is OK, but if an error would happen during StateDeleteReorganization and a rollback would happen, than changes to the indexes to be dropped would miss records written during that state.

Thx. I have an another does it mean we could cancel reorg partition ddl job in any state? I can't found any information in `IsRollbackable()` function.

mjonss:
> > > > resulting in they would not continue to be updated, so in case of a rollback there might be missing entries.
> > > 
> > > 
> > > Can you clarify which sql in the newly added test case covers this problem? And what's the result before this PR?
> > 
> > 
> > Since the indexes was in StateDeleteReorganization, during `job.SchemaState == model.StateDeleteReorganization`, the indexes would not been written, see [`IsIndexWritable()`](https://github.com/pingcap/tidb/blob/48c57ce4996ba04ad08f9edafab0706d2c04f05f/pkg/table/tables/index.go#L679-L686). Which would mean that during normal DDL that is OK, but if an error would happen during StateDeleteReorganization and a rollback would happen, than changes to the indexes to be dropped would miss records written during that state.
> 
> Thanks your comments. I have an another question, does it mean we could cancel reorg partition ddl job in any state? I can't found any information in `IsRollbackable()` function.

Since we will continue to double write until all processing is done and the old partitions cannot be accessed any longer, it is possible to cancel the DDL Job in any state and the rollback should work. I have not tested the Cancel part, though, do you have any example of tests that I can replicate @Defined2014 ?

pkg/ddl/tests/partition/multi_domain_test.go
```diff
@@ -246,19 +256,304 @@ func TestMultiSchemaDropListColumnsDefaultPartition(t *testing.T) {
...
+				if idx.ID == locIdx {
+					continue LocalLoop
+				}
+			}
+			// local index removed
+			for _, part := range tbl.Meta().Partition.Definitions {
+				require.False(t, HaveEntriesForTableIndex(t, tkO, part.ID, locIdx), "Local index id %d for partition id %d has still entries!", locIdx, tableID)
+			}
+		}
+		// TODO: Fix cleanup issues, most likely it needs one more SchemaState in onReorganizePartition
```
mjonss, line 430: Also disabled due to https://github.com/pingcap/tidb/issues/56822#issuecomment-2436343806.

pkg/ddl/partition.go
```diff
@@ -3552,14 +3554,18 @@ func (w *worker) onReorganizePartition(jobCtx *jobContext, job *model.Job) (ver
...
+			if indexInfo.Unique && indexInfo.State == model.StateWriteOnly {
 				// Drop the old unique (possible global) index, see onDropIndex
 				indexInfo.State = model.StateNone
 				DropIndexColumnFlag(tblInfo, indexInfo)
 				RemoveDependentHiddenColumns(tblInfo, indexInfo)
 				dropIndices = append(dropIndices, indexInfo)
 			}
 		}
+		// TODO: verify that the indexes are dropped,
+		// and that StateDeleteOnly+StateDeleteReorganization is not needed.
```
mjonss, line 3566: Tests shows that at least the old partitions can be written (double write in case we need to rollback), so we need to add a new state to make sure all clients only use (read and write) the new partitions and new indexes, this should not be fixed in this PR, but a new PR for fixing #56819.

pkg/table/tables/partition.go
```diff
@@ -128,9 +128,50 @@ func newPartitionedTable(tbl *TableCommon, tblInfo *model.TableInfo) (table.Part
 	if err := initTableIndices(&ret.TableCommon); err != nil {
 		return nil, errors.Trace(err)
 	}
+	origIndices := ret.meta.Indices
+	DroppingDefinitionIndices := make([]*model.IndexInfo, 0, len(origIndices))
+	AddingDefinitionIndices := make([]*model.IndexInfo, 0, len(origIndices))
+	for _, idx := range origIndices {
+		newIdx, ok := pi.DDLChangedIndex[idx.ID]
```
Defined2014, line 135: Will the `pi.DDLChangedIndex` is nil here?

pkg/ddl/partition.go
```diff
@@ -2186,8 +2186,8 @@ func (w *worker) rollbackLikeDropPartition(jobCtx *jobContext, job *model.Job) (
 	var dropIndices []*model.IndexInfo
 	for _, indexInfo := range tblInfo.Indices {
 		if indexInfo.Unique &&
```
Defined2014, line 2188: What's the special for unique index here?

mjonss: Currently the Global Index can only be unique, so is just filters out indexes that are not affected.

pkg/ddl/partition.go
```diff
@@ -3487,26 +3497,18 @@ func (w *worker) onReorganizePartition(jobCtx *jobContext, job *model.Job) (ver
 			if !index.Unique {
```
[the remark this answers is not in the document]
mjonss: Same here, since global indexes currently must be global, we filter away non-unique indexes here. Should be removed when implementing non-unique global indexes.

pkg/ddl/partition.go
Defined2014: This will affect non-global unique index or not?

pkg/table/tables/partition.go
mjonss: if `pi.DDLChangedIndex` is nil, then there will not be any matches during reading. A nil map will panic on write, but reading is the same as an empty map.

Defined2014: Thx, I don't know it before.

pkg/ddl/partition.go
mjonss: So this path also handles `ALTER TABLE t PARTITION BY ... [UPDATE INDEXES (...)]` so yes, it also handles non-global unique indexes that might be changed from/to global indexes.

Defined2014: Will it handle the case of `unique index` -> `non global unique index`? Maybe it's no need to process.

pkg/ddl/partition.go
```diff
@@ -3279,6 +3285,8 @@ func (w *worker) onReorganizePartition(jobCtx *jobContext, job *model.Job) (ver
 				newIndex := index.Clone()
 				newIndex.State = model.StateDeleteOnly
 				newIndex.ID = AllocateIndexID(tblInfo)
+				tblInfo.Partition.DDLChangedIndex[index.ID] = false
+				tblInfo.Partition.DDLChangedIndex[newIndex.ID] = true
 				newIndex.Global = true
 				tblInfo.Indices = append(tblInfo.Indices, newIndex)
 			} else {
```
Defined2014, line 3292: I forget remove L3293 in https://github.com/pingcap/tidb/pull/56868, maybe we could remove it in this PR.

pkg/ddl/partition.go
Defined2014: After I debuged, maybe `indexInfo.State == model.StateWriteOnly` is enough. Am I right?

mjonss: Yes that should be enough, or even use the new DDLChangedIndex map. It should be safe for ADD PARTITION as well, since it also using this function.

pkg/ddl/partition.go
mjonss: Done.

pkg/ddl/partition.go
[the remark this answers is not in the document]
mjonss: Let's keep the rest as-is, and remove them all when starting to support non-unique global indexes.

Defined2014 (approved):

pkg/ddl/partition.go
```diff
@@ -2185,9 +2185,7 @@ func (w *worker) rollbackLikeDropPartition(jobCtx *jobContext, job *model.Job) (
 
 	var dropIndices []*model.IndexInfo
 	for _, indexInfo := range tblInfo.Indices {
-		if indexInfo.Unique &&
-			indexInfo.State == model.StateDeleteReorganization &&
-			tblInfo.Partition.DDLState == model.StateDeleteReorganization {
+		if indexInfo.State == model.StateWriteOnly {
```
tangenta, line 2188: Why only drop the write-only indexes? What about other states? For example, what will happen when we rollback from the "none -> delete-only" step?

There may be `delete-only` indexes before converting to rolling back job:
```
newIndex := index.Clone()
newIndex.State = model.StateDeleteOnly
newIndex.ID = AllocateIndexID(tblInfo)
```

tangenta:
It seems that rolling back only takes 1 schema version to complete. I am trying to figure out how it works...

mjonss:
> It seems that rolling back only takes 1 schema version to complete. I am trying to figure out how it works...

@tangenta it should take two, one when converting the failed DDL to rollback, and one as a second step in `rollingbackLikeDrop` there are extra changes in the follow up patch for error handling and rollback in #57114.

pkg/ddl/partition.go
mjonss: There should only be either `StatePublic` or `StateWriteOnly` so only write-only indexes that should be removed. Also in the follow ups #57114 and #56974 (based on this PR) it will block the rollback [here](https://github.com/pingcap/tidb/pull/56974/files#diff-443bbe2f8987cd40a2d87394088d85420358c25634111278944472b5b840590eL345) and [here](https://github.com/pingcap/tidb/pull/57114/files#diff-c51ad6b394f865592721a3e18d58574176394570e3efcff7703e8bd0c3c5c2efL744) for the last state. One can see this PR a part 1 of 3, where the optimistic path works, then the failure/rollback/cleanup is fixed in part 2 and 3.

tangenta (approved):

pkg/table/tables/partition.go
```diff
@@ -128,9 +128,50 @@ func newPartitionedTable(tbl *TableCommon, tblInfo *model.TableInfo) (table.Part
...
+		if !ok {
+			// Untouched index
+			DroppingDefinitionIndices = append(DroppingDefinitionIndices, idx)
+			if pi.DDLState != model.StateDeleteReorganization {
+				// If pi.DDLState == DeleteReorg, then keep the StatePublic.
```
tangenta, line 140: Why is the state of the index public?

mjonss: To not skip it when inserting into the index, `StateDeleteReorganization` would work similar to `StateDeleteOnly` so that is why keeping the index state as `StatePublic` instead of `StateDeleteReorganization`, to keep it double written during the delete reorganization phase.

PR #56795, opened 2024-10-23
ddl: add the ddl notifier as a listener to the stats owner

Rustin170506:

Issue Number: ref https://github.com/pingcap/tidb/issues/55722

Problem Summary:

We should only start the ddl notifier on the stats owner node. So in this PR, I added the ddl notifier as a listener to the stats owner.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


Rustin170506 (commented):

Rustin170506 (commented):

pkg/ddl/notifier/subscribe.go
```diff
@@ -255,7 +272,31 @@ func (n *DDLNotifier) processEventForHandler(
 	return nil
 }
 
+// Stop stops the background loop.
+func (n *DDLNotifier) Stop() {
+	// If the notifier is not started, the cancel function is nil.
+	if n.cancel != nil {
```
lance6716, line 278: If caller will call `Close` in a case that hasn't `Start`, it's possible that `Start` races with `Close`? And the execution is complicated.

The only special case which calls `Close` is `Domain.Close`. I think we can delete it. Let `Domain`'s `ctx` control the owner to exit, and then `OnRetireOwner` will be called. It's more clear.

pkg/domain/domain.go
```diff
@@ -2350,6 +2347,9 @@ func (do *Domain) UpdateTableStatsLoop(ctx, initStatsCtx sessionctx.Context) err
 	variable.EnableStatsOwner = do.enableStatsOwner
 	variable.DisableStatsOwner = do.disableStatsOwner
 	do.statsOwner = do.newOwnerManager(handle.StatsPrompt, handle.StatsOwnerKey)
+	if intest.InTest {
+		do.statsOwner.SetListener(do.ddlNotifier)
```
lance6716, line 2351: will statsOwner have more logic to be put in Listener? If so, we'd better create a separate struct for Listener and let it invokes DDLNotifier

Rustin170506: Currently no, but we can always change it later.

pkg/ddl/notifier/subscribe.go
Rustin170506: We can't close the session context in OnRetireOwner, because the node can be the owner again. But we only init the ddl notifier once with the sctx, so if the sctx is closed, there is no way to restart the ddl notifier again.

Rustin170506 (commented):

pkg/ddl/notifier/subscribe.go
```diff
@@ -91,18 +106,17 @@ type DDLNotifier struct {
 }
 
 // NewDDLNotifier initializes the global DDLNotifier. It should be called only
```
lance6716, line 108: I think we don't need to mention "called only once and before any RegisterHandler call.", because 
- it creates a new instance and call multiple times will create multiple instance.
- it's the receiver of RegisterHandler so it must be created first.

pkg/ddl/notifier/subscribe.go
```diff
@@ -255,7 +278,27 @@ func (n *DDLNotifier) processEventForHandler(
...
 }
 
-// Close releases the resources.
-func (n *DDLNotifier) Close() {
-	n.ownedSess.Close()
+// Stop stops the background loop.
+// Exposed for testing.
+// Do not call this function directly. Use owner.Listener interface instead.
+func (n *DDLNotifier) Stop() {
+	// If the notifier is not started, the cancel function is nil.
```
lance6716, line 285: maybe mention that in fact cancel function will not be nil. Just to improve robust.

Also maybe 

```
if n.cancel == nil { return }
n.cancel()
n.wg.Wait()
```

to reduce indent. Seems linter is broken.

pkg/ddl/notifier/subscribe.go
```diff
@@ -255,7 +278,27 @@ func (n *DDLNotifier) processEventForHandler(
...
+		n.cancel()
+		n.wg.Wait()
+	}
+}
+
+// OnBecomeOwner implements the owner.Listener interface.
+// We need to make sure only one DDLNotifier is running at any time.
+func (n *DDLNotifier) OnBecomeOwner() {
+	n.wg.RunWithRecover(n.Start, func(r any) {
+		logutil.BgLogger().Error("panic in ddl notifier", zap.Any("recover", r), zap.Stack("stack"))
```
lance6716, line 296: maybe use `intest` to let it really panic in our testing environment

catching the panic is friendly to other components in the same process to let them keep running, but it also mutes the error and make the bug hard to reveal.

lance6716 (approved):
feel free to unhold

pkg/ddl/notifier/subscribe.go
Rustin170506: > maybe mention that in fact cancel function will not be nil. Just to improve robust.

In the test, it is possible.

Rustin170506: > to reduce indent. Seems linter is broken.

Updated.

pkg/ddl/notifier/subscribe.go
Rustin170506: Updated.

pkg/ddl/notifier/subscribe.go
Rustin170506: Updated.

PR #56821, opened 2024-10-24
unistore: fix the fair locking mechanism in unistore and make foreign key work

### LINKED ISSUE #56663: Unistore TiDB cannot handle foreign key in transaction
## Bug Report

Prepare:

```sql
create table parent (id int primary key);
create table child(id int primary key, ref int, foreign key (ref) references parent(id));

insert into parent values (1);
```

Execute:

| Transaction 1    | Transaction 2 |
| -------- | ------- |
| BEGIN;  | BEGIN;    |
| INSERT INTO child VALUES (1, 1); |     |
|     | UPDATE parent SET id = 2 WHERE id = 1    |
| COMMIT; | |
| | COMMIT; |

`UPDATE` in transaction 2 failed.

`UPDATE` in transaction 2 success.

master fa723c, with unistore.

This issue cannot be reproduced in TiKV store.

YangKeao:

Issue Number: close #56663 

Problem Summary:

When there is an update for the same row in other transactions, it's necessary to update the `for_update_ts` to use the latest data to verify the constraint. Meanwhile, TiDB/TiKV has a mechanism called fair locking to lock the key with conflict. 

Therefore, when a key is locked, the storage layer (unistore or TiKV) should verify whether other transactions have written this key. If so, it should return the conflict info to TiDB, and TiDB can retry and update the `for_update_ts`.

However, the implementation of unistore has a little bug:

For TiKV, it returns the latest TS of both WRITE and LOCK, because they share the similar key (but with different value):

```rust
pub fn commit<S: Snapshot>(
    txn: &mut MvccTxn,
    reader: &mut SnapshotReader<S>,
    key: Key,
    commit_ts: TimeStamp,
) -> MvccResult<Option<ReleasedLock>> {
...
    let mut write = Write::new(
        WriteType::from_lock_type(lock.lock_type).unwrap(),
        reader.start_ts,
        lock.short_value.take(),
    )
    .set_last_change(lock.last_change.clone())
    .set_txn_source(lock.txn_source);
...
    txn.put_write(key.clone(), commit_ts, write.as_ref().to_bytes());
    Ok(txn.unlock_key(key, lock.is_pessimistic_txn(), commit_ts))
...
}
```

And in the `acquire_pessimistic_lock.rs`:

```rust
pub fn acquire_pessimistic_lock<S: Snapshot>(...) {
...
    if let Some((commit_ts, write)) = reader.seek_write(&key, TimeStamp::max())? {
...
    }
...
}
```

For unistore, the logic of handling pessimistic lock is similar, but unistore stores the LOCK info in a different key:

```go
// EncodeExtraTxnStatusKey encodes a extra transaction status key.
// It is only used for Rollback and Op_Lock.
func EncodeExtraTxnStatusKey(key []byte, startTS uint64) []byte {
	b := append([]byte{}, key...)
	ret := codec.EncodeUintDesc(b, startTS)
	ret[0]++
	return ret
}

func (wb *writeBatch) Commit(key []byte, lock *mvcc.Lock) {
...
		opLockKey := y.KeyWithTs(mvcc.EncodeExtraTxnStatusKey(key, wb.startTS), wb.startTS)
		wb.dbBatch.set(opLockKey, nil, userMeta)
...
}
```

We need to also iterate (backward) through the extra txn status keys to get the latest startTS/commitTS.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/test/fktest/BUILD.bazel
+++ b/pkg/executor/test/fktest/BUILD.bazel
@@ -8,7 +8,7 @@ go_test(
         "main_test.go",
     ],
     flaky = True,
-    shard_count = 25,
+    shard_count = 26,
     deps = [
         "//pkg/config",
         "//pkg/executor",
--- a/pkg/executor/test/fktest/foreign_key_test.go
+++ b/pkg/executor/test/fktest/foreign_key_test.go
@@ -2494,3 +2494,36 @@ func TestFKBuild(t *testing.T) {
 	tk.MustExec("delete from test.t3")
 	tk.MustQuery("select * from test.t2").Check(testkit.Rows())
 }
+
+func TestLockKeysInDML(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("create table t1 (id int primary key);")
+	tk.MustExec("create table t2 (id int primary key, foreign key fk (id) references t1(id));")
+
+	tk.MustExec("insert into t1 values (1)")
+	tk.MustExec("BEGIN")
+	tk.MustExec("INSERT INTO t2 VALUES (1)")
+	var wg sync.WaitGroup
+	var tk2CommitTime time.Time
+	tk2StartTime := time.Now()
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		tk2 := testkit.NewTestKit(t, store)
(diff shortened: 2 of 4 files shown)


you06 (approved):
@MyonKeminta PTAL

MyonKeminta (approved):

Issue #47400, opened 2023-10-05
planner: zero estimation result if all values are in TopN


qw4990:
## Enhancement
```
create table t (a int, key(a));
insert into t values (1), (3);
analyze table t with 2 topn;
insert into t values (2);

explain select * from t where a=2;
+------------------------+---------+-----------+---------------------+-------------------------------+
| id                     | estRows | task      | access object       | operator info                 |
+------------------------+---------+-----------+---------------------+-------------------------------+
| IndexReader_6          | 0.00    | root      |                     | index:IndexRangeScan_5        |
| └─IndexRangeScan_5     | 0.00    | cop[tikv] | table:t, index:a(a) | range:[2,2], keep order:false |
+------------------------+---------+-----------+---------------------+-------------------------------+
2 rows in set (0.00 sec)
```

After v6.5, in the case above, the estimation result for `a=2` is zero, which seems dangerous.

If no `TopN`, it'll be `1`:
```
create table t (a int, key(a));
insert into t values (1), (3);
analyze table t with 0 topn;
insert into t values (2);

mysql> explain select * from t where a=2;
+------------------------+---------+-----------+---------------------+-------------------------------+
| id                     | estRows | task      | access object       | operator info                 |
+------------------------+---------+-----------+---------------------+-------------------------------+
| IndexReader_6          | 1.00    | root      |                     | index:IndexRangeScan_5        |
| └─IndexRangeScan_5     | 1.00    | cop[tikv] | table:t, index:a(a) | range:[2,2], keep order:false |
+------------------------+---------+-----------+---------------------+-------------------------------+
```

Should we avoid using `0` in this case (this value is not in TopN and there is no Histogram)?

AilinKid:
the related code about this is as follows
```
	// 3. use uniform distribution assumption for the rest (even when this value is not covered by the range of stats)
	histNDV := float64(c.Histogram.NDV - int64(c.TopN.Num()))
	if histNDV <= 0 {
		return 0, nil
	}
	return c.Histogram.notNullCount() / histNDV, nil
```

in first case, since topN is collected, and it is overlapped with the NDV indication in Histogram, so the computation here is meaningless, leading `histNDV := float64(c.Histogram.NDV - int64(c.TopN.Num()))` = `2-2` = 0

[...]

PR #56848, opened 2024-10-25
planner: Use realtimeRowCount when all topN collected

terry1purcell:

Issue Number: close #47400

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/cardinality/row_count_column.go
```diff
@@ -177,7 +177,7 @@ func equalRowCountOnColumn(sctx planctx.PlanContext, c *statistics.Column, val t
 		if modifyCount == 0 {
 			return 0, nil
 		}
-		return 1, nil
+		return max(1, float64(modifyCount)/float64(c.Histogram.NDV)), nil
```
princejha95, line 180: Shouldn't we be writing any test case for this ?

pkg/planner/cardinality/row_count_index.go
```diff
@@ -419,7 +419,7 @@ func equalRowCountOnIndex(sctx planctx.PlanContext, idx *statistics.Index, b []b
 		if modifyCount == 0 {
 			return 0
 		}
-		return 1
+		return max(1, float64(modifyCount)/float64(idx.Histogram.NDV))
```
princejha95, line 422: same comment as above

pkg/planner/cardinality/row_count_column.go
terry1purcell: A test using real modifyCount will unfortunately be flaky - because the updating of modifyCount after modifications to a table is a timing issue.

pkg/planner/cardinality/row_count_index.go
terry1purcell: same answer as above

qw4990 (commented):
Can we add the case in https://github.com/pingcap/tidb/issues/47400 as a UT here?

pkg/planner/cardinality/row_count_column.go
```diff
@@ -177,7 +179,15 @@ func equalRowCountOnColumn(sctx planctx.PlanContext, c *statistics.Column, val t
 		if modifyCount == 0 {
 			return 0, nil
 		}
-		return 1, nil
+		// Use sqrt to create an NDV if one doesn't already exist, or reset to the original NDV
+		if c.Histogram.NDV <= 0 {
```
qw4990, line 183: Please add some comments here to explain when this can happen "c.Hist.NDV <= 0".

terry1purcell: I don't believe that NDV can be < 0 - but this check is here just to be safe. NDV == 0 seems valid if you add a column to a table, with default NULL - Distinct Count in stats_histograms will be 0. If you add the column with default 0 (rather than NULL) - NDV == 1. This is with auto-analyze disabled.

qw4990 (approved):
LGTM, but it's my first time to see using "Sqrt" in cardinality estimation, is there any source of this formula, or any intuition behind this formula? @terry1purcell

pkg/planner/cardinality/row_count_column.go
```diff
@@ -15,6 +15,8 @@
 package cardinality
```
hawkingrei, line 15: Can we add some test cases for it?

winoros (approved):

pkg/planner/cardinality/row_count_column.go
```diff
@@ -177,7 +179,15 @@ func equalRowCountOnColumn(sctx planctx.PlanContext, c *statistics.Column, val t
...
-		return 1, nil
+		// Reset to the original NDV, or if no NDV - derive an NDV using sqrt
+		if c.Histogram.NDV > 0 {
+			histNDV = float64(c.Histogram.NDV)
+		} else {
+			histNDV = math.Sqrt(max(c.Histogram.NotNullCount(), float64(modifyCount)))
+		}
+		// Return the min of the original notNullCount or the modifyCount/NDV. This is to reduce the
+		// risk of too large or too small an estimate if modifyCount is large or NotNullCount is small
+		return max(1, min(c.Histogram.NotNullCount(), float64(modifyCount)/histNDV)), nil
```
Rustin170506, line 190: I think you should use `c.NotNullCount` here.

Rustin170506: We should use the TopN count + Hist Count here. The c.Histogram.NotNullCount() just for hist.

terry1purcell: I have corrected this in latest revision

pkg/planner/cardinality/row_count_column.go
terry1purcell: Yes - will do this before final review.

hawkingrei (approved):

Rustin170506 (approved):

PR #56928, opened 2024-10-29
planner: Use/force to apply prefer range scan

### LINKED ISSUE #55632: Planner avoid table scans for pseudo stats if index scan exists
## Enhancement
Pseudo stats are unreliable. Preference a matching index scan over table scan for pseudo stats.

terry1purcell:

Issue Number: close #55632 

Problem Summary:

MySQL will increase the cost of a table scan if FORCE INDEX is used in the query. TiDB does not differentiate between USE and FORCE in the query. This enhancement will add a cost penalty to a full table scan if USE or FORCE is used in the query. It will also apply to ANY full table scan in the query, and not only to the table that the USE/FORCE is applied to. This is in response to customer issues where USE/FORCE is specified, but they still see table scans chosen within their query.

If a USE/FORCE is specified against the PRIMARY index, and there are no other predicates (such that the scan is a full table scan of that primary) - this enhancement will increase that table scan cost.

The cost penalty does not apply to a table range scan.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


tests/integrationtest/r/planner/core/plan_cost_ver2.result
```diff
@@ -78,91 +78,91 @@ IndexLookUp_7	100.00	1024158.30	root
...
-  └─Sort_13	100.00	69195.64	root		planner__core__plan_cost_ver2.t.b
-    └─TableReader_12	100.00	35402.80	root		data:TableFullScan_11
-      └─TableFullScan_11	100.00	22730.80	cop[tikv]	table:t	keep order:false
+StreamAgg_8	100.00	311523.51	root		group by:Column#10, funcs:sum(Column#7)->Column#4, funcs:sum(Column#8)->Column#5, funcs:sum(Column#9)->Column#6
+└─Projection_17	100.00	311523.51	root		cast(planner__core__plan_cost_ver2.t.a, decimal(10,0) BINARY)->Column#7, cast(plus(planner__core__plan_cost_ver2.t.a, 2), decimal(20,0) BINARY)->Column#8, cast(plus(planner__core__plan_cost_ver2.t.a, 4), decimal(20,0) BINARY)->Column#9, planner__core__plan_cost_ver2.t.b->Column#10
+  └─Sort_13	100.00	296503.61	root		planner__core__plan_cost_ver2.t.b
+    └─TableReader_12	100.00	262710.77	root		data:TableFullScan_11
+      └─TableFullScan_11	100.00	250038.77	cop[tikv]	table:t	keep order:false
```
AilinKid, line 165: is there any specific test case this pull request is aimed at? by now it's old tests modification, which is appreciated and help me to understand the changes in pkg/planner/core/find_best_task.go

AilinKid:
by the way, if one your pull request refers to a issue and didn't want to close it when merging, because it's not complete or something, you can use: <ref + issue link> instead of <close + issue link>

pkg/planner/core/stats.go
```diff
@@ -404,21 +408,24 @@ func detachCondAndBuildRangeForPath(
...
-			minSelectivity = min(minSelectivity, path.CountAfterAccess/totalRowCount)
-			continue
+		if totalRowCount > 0 {
+			if path.IsTablePath() || path.PartialIndexPaths != nil {
+				minSelectivity = min(minSelectivity, path.CountAfterAccess/totalRowCount)
+			} else {
+				minSelectivity = min(minSelectivity, path.CountAfterIndex/totalRowCount)
+			}
+		}
+		if !hasForce && path.Forced {
```
qw4990, line 424: User can use "use index" to force table scan like: 
```
create table t (a int, b int, primary key(a), key(b));
select /*+ use_index(t, primary) */ * from t;
```
Should we exclude this case like `!path.IsTablePath()`?

qw4990: Or our assumption here is that if the user is using `use_index` hint, they prefer `range scan` no matter it's table-range-scan or index-range-scan?

pkg/planner/core/plan_cost_ver2.go
```diff
@@ -184,7 +184,7 @@ func (p *PhysicalTableScan) GetPlanCostVer2(taskType property.TaskType, option *
 		}
 		hasFullRangeScan := ranger.HasFullRange(p.Ranges, unsignedIntHandle)
 
-		shouldApplyPenalty := hasFullRangeScan && preferRangeScanCondition
+		shouldApplyPenalty := hasFullRangeScan && (sessionVars.StmtCtx.GetHasForce() || preferRangeScanCondition)
```
qw4990, line 187: So our assumption is that if the user is using `use_index` hint, we assume they want `range-scan`?

qw4990: Better to add some comments if I understand this correctly.

pkg/sessionctx/stmtctx/stmtctx.go
```diff
@@ -732,6 +734,11 @@ func (sc *StatementContext) GetPlanHint() (string, bool) {
 	return sc.planHint, sc.planHintSet
 }
 
+// GetHasForce gets the HasForce boolean generated from the plan.
+func (sc *StatementContext) GetHasForce() bool {
```
qw4990, line 738: `Force` is too common and might bring confusion under the statement context, I suggest we can use a more concrete name like `IndexForce`.

elsa0520:
Could you please supply the PR comment to explain what this PR changed and how it works ?

pkg/sessionctx/stmtctx/stmtctx.go
terry1purcell: Done

pkg/planner/core/plan_cost_ver2.go
terry1purcell: Done

qw4990 (approved):

tests/integrationtest/r/planner/core/plan_cost_ver2.result
terry1purcell: I added new testcase TestTableScanCostWithForce

qw4990 (approved):

pkg/planner/core/casetest/dag/testdata/plan_suite_out.json
```diff
@@ -48,7 +48,7 @@
       },
       {
         "SQL": "select c from t order by t.a limit 1",
-        "Best": "IndexReader(Index(t.c_d_e)[[NULL,+inf]]->TopN([test.t.a],0,1))->TopN([test.t.a],0,1)->Projection"
+        "Best": "TableReader(Table(t)->Limit)->Limit->Projection"
```
AilinKid, line 51: this sql doesn't have a use index/force index, it shouldn't be changed?

pkg/planner/core/find_best_task.go
```diff
@@ -1172,13 +1172,15 @@ func skylinePruning(ds *logicalop.DataSource, prop *property.PhysicalProperty) [
...
-	preferRange := ds.SCtx().GetSessionVars().GetAllowPreferRangeScan() && (ds.TableStats.HistColl.Pseudo || ds.TableStats.RowCount < 1)
 	// If we've forced an index merge - we want to keep these plans
 	preferMerge := len(ds.IndexMergeHints) > 0 || fixcontrol.GetBoolWithDefault(
 		ds.SCtx().GetSessionVars().GetOptimizerFixControlMap(),
 		fixcontrol.Fix52869,
 		false,
 	)
+	// tidb_opt_prefer_range_scan is the master switch to control index preferencing
+	preferRange := ds.SCtx().GetSessionVars().GetAllowPreferRangeScan() &&
+		(preferMerge || (ds.TableStats.HistColl.Pseudo || ds.TableStats.RowCount < 1))
```
AilinKid, line 1183: i don't get why should we put preferMerge into this equation,seems irrelevant? it seems change the orignal logic, may be this is reason why irrelevant sql plan result is changed, or you are trying to fix something

AilinKid (commented):
Rest LGTM

pkg/planner/core/find_best_task.go
qw4990: It seems like we can treat `IndexMerge` as a special kind of `IndexRange`, so `preferMerge` also means the user prefers some `IndexRange`, which allow us to set `preferRange=true` here.

AilinKid (approved):

PR #57042, opened 2024-10-31
*: avoid notify privilege update for all users

tiancaiamao:

Issue Number: ref #55563

Problem Summary:

In the previous commit, I have maintained the active user lists, this commit intend to fix the notify part.
When privilege change, just notify the changed users and update data for them, instead of all the users.

The code changes including:

- notify privilege update will encoding the changed user list into the etcd message
- the domain privilege loop will decode the user list and update only the active users among them
- fix bug on the privilege handle merge operation, the old  "append diff + sort + dedup" can not handle the reovke privilege operation
- the roles of a user is also consider as the privilege data and ensureActiveUse() need to load them
- ensureActiveUser() should load the privilege data of the user, the roles of the data, **plus the data of roles recursively** 
- ensureActiveUser() is called more widely now to make the CI pass (we'd better optimize here later)

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/domain/domain.go
```diff
@@ -2801,13 +2822,23 @@ const (
...
 // NotifyUpdatePrivilege updates privilege key in etcd, TiDB client that watches
 // the key will get notification.
-func (do *Domain) NotifyUpdatePrivilege() error {
+func (do *Domain) NotifyUpdatePrivilege(userList []string) error {
 	// No matter skip-grant-table is configured or not, sending an etcd message is required.
 	// Because we need to tell other TiDB instances to update privilege data, say, we're changing the
 	// password using a special TiDB instance and want the new password to take effect.
 	if do.etcdClient != nil {
+		var notifyList strings.Builder
+		for _, user := range userList {
```
lance6716, line 2831: this is `strings.Join`

pkg/domain/domain.go
```diff
@@ -2801,13 +2822,23 @@ const (
...
+				notifyList.WriteString(user)
+			} else {
+				notifyList.WriteString(",")
+				notifyList.WriteString(user)
+			}
+		}
+
 		row := do.etcdClient.KV
-		_, err := row.Put(context.Background(), privilegeKey, "")
+		_, err := row.Put(context.Background(), privilegeKey, notifyList.String())
```
lance6716, line 2841: Domain has a meaningful context `ctx`, please use it so tidb-server shutdown can be more timely.

pkg/executor/simple.go
```diff
@@ -2431,24 +2452,37 @@ func userExists(ctx context.Context, sctx sessionctx.Context, name string, host
...
 	}
+
+	var authPlugin string
+	colIdx := -1
+	for i, f := range recordSet.Fields() {
+		if f.ColumnAsName.L == "plugin" {
+			colIdx = i
+		}
+	}
+	if rows == 1 {
```
lance6716, line 2476: can you add more comments to explain when will `rows` have different values?

tiancaiamao: It can be 0 or 1... 
0 when the user + host does not exist
1 when the user + host exist (this is a primary key)

pkg/domain/domain.go
tiancaiamao: done

pkg/domain/domain.go
```diff
@@ -1861,6 +1861,22 @@ func (do *Domain) GetPDHTTPClient() pdhttp.Client {
 	return nil
 }
 
+func decodePrivilegeEvent(resp clientv3.WatchResponse) []string {
+	var userList []string
+	if len(resp.Events) > 0 {
```
lance6716, line 1866: this `if` check is not necessary, because `for` is no-op when `Events` has zero length

pkg/domain/domain.go
```diff
@@ -1861,6 +1861,22 @@ func (do *Domain) GetPDHTTPClient() pdhttp.Client {
...
+		for _, event := range resp.Events {
+			if event.Kv != nil {
+				val := event.Kv.Value
+				if len(val) > 0 {
+					users := strings.Split(string(val), ",")
```
lance6716, line 1871: what if username contains `,`? And also it's hard to know the format of data rest at etcd. Maybe add comments or create a separate type for it

pkg/domain/domain.go
```diff
@@ -1887,11 +1903,16 @@ func (do *Domain) LoadPrivilegeLoop(sctx sessionctx.Context) error {
...
 			select {
 			case <-do.exit:
 				return
-			case _, ok = <-watchCh:
+			case resp, ok = <-watchCh:
+				if ok {
+					userList = decodePrivilegeEvent(resp)
+				}
 			case <-time.After(duration):
 			}
```
lance6716, line 1917: line 1920: when domain closes, the watch goroutine will leak. So I prefer to use domain's context in `Watch`. And if we do this, re-watch should checking both `ok` and `do.ctx.Err()`

pkg/domain/domain.go
```diff
@@ -2801,13 +2822,14 @@ const (
 
 // NotifyUpdatePrivilege updates privilege key in etcd, TiDB client that watches
 // the key will get notification.
-func (do *Domain) NotifyUpdatePrivilege() error {
+func (do *Domain) NotifyUpdatePrivilege(userList []string) error {
 	// No matter skip-grant-table is configured or not, sending an etcd message is required.
 	// Because we need to tell other TiDB instances to update privilege data, say, we're changing the
 	// password using a special TiDB instance and want the new password to take effect.
 	if do.etcdClient != nil {
+		notifyList := strings.Join(userList, ", ")
```
lance6716, line 2830: you use `,` above but here `, `. Maybe write a separate type for "user list" and its associated member functions for (un)marshal to find bugs easier 😂

pkg/domain/domain.go
tiancaiamao: done

pkg/domain/domain.go
```diff
@@ -1834,16 +1848,21 @@ func (do *Domain) LoadPrivilegeLoop(sctx sessionctx.Context) error {
...
 		for {
+			var userList []string
+			var resp clientv3.WatchResponse
 			ok := true
 			select {
 			case <-do.exit:
 				return
-			case _, ok = <-watchCh:
+			case resp, ok = <-watchCh:
+				if ok {
```
lcwangchao, line 1858: Though it is not introduced in this PR, do we need to handle the `ok == false` here? If the watchCh closed unexpectedly for some reason, this loop would run frequent.

lcwangchao: Just like this issue #49273

pkg/privilege/privileges/cache.go
```diff
@@ -1863,13 +1945,32 @@ func (h *Handle) Get() *MySQLPrivilege {
 }
 
 // Update loads all the privilege info from kv storage.
-func (h *Handle) Update() error {
-	var priv MySQLPrivilege
-	err := priv.LoadAll(h.sctx)
+func (h *Handle) Update(userList []string) error {
```
D3Hunter, line 1948: please split into `UpdateAll` and `Update`, a nil `userList []string` is not a clear parameter to mean we want to update all

pkg/domain/domain.go
```diff
@@ -2751,15 +2770,37 @@ const (
...
+func decodeUserList(userList []string, val string) []string {
+	users := strings.Split(val, ",")
+	for _, userRaw := range users {
+		user, err := base64.StdEncoding.DecodeString(userRaw)
+		if err == nil {
+			userList = append(userList, string(user))
+		}
+	}
+	return userList
+}
```
D3Hunter, line 2792: please use a json as the value of the etcd key, easy to en/decode and extend

also what's the size of the list, etcd is not fit to store large data, max 1.5m by default

br/pkg/restore/snap_client/systable_restore.go
```diff
@@ -200,7 +200,7 @@ func (rc *SnapClient) afterSystemTablesReplaced(ctx context.Context, db string,
 	var err error
 	for _, table := range tables {
 		if table == "user" {
-			if serr := rc.dom.NotifyUpdatePrivilege(); serr != nil {
+			if serr := rc.dom.NotifyUpdatePrivilege(nil); serr != nil {
```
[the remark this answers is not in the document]
tiancaiamao: reload the all active users
`flush privileges` would pass nil
In br I'm not sure how it's used, but this could not cause anything wrong.

pkg/domain/domain.go
tiancaiamao: line 1863 will recreate the watch channel...  if it's not domain close.

br/pkg/restore/snap_client/systable_restore.go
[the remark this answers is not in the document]
lcwangchao: I think a new function `NotifyUpdateAllPrivilege` is better, a `nil` argument may be confusing

pkg/privilege/privileges/cache.go
```diff
@@ -1863,13 +1945,32 @@ func (h *Handle) Get() *MySQLPrivilege {
...
+	if userList == nil {
```
lcwangchao, line 1949: code here: https://github.com/pingcap/tidb/pull/57042/files#diff-61cf704aa1127eb5fc43afc049c6fd07ff429420ff68e3555a26f294df235a5fR1812-R1814 seems always return a non-nil value.

br/pkg/restore/snap_client/systable_restore.go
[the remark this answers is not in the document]
tiancaiamao: done

pkg/domain/domain.go
tiancaiamao: There is a sleep 1s ....

pkg/privilege/privileges/cache.go
tiancaiamao: Done

pkg/domain/domain.go
```diff
@@ -1808,6 +1810,25 @@ func (do *Domain) GetPDHTTPClient() pdhttp.Client {
...
+				err := json.Unmarshal(val, &event)
+				if err == nil {
+					break
+				}
+			}
+		}
+	}
+	if len(event.UserList) == 0 {
+		event.All = true
+	}
```
D3Hunter, line 1828: set the `All` to true at event producer side, we only need to consume the event

tiancaiamao: I do it here on purpose.
One consideration is the old tidb compatibility.
The old tidb will not set any user list, so the json message is null.
And the default behavior before is reload all ...

So this can keep the compatibility with the old tidb.... old tidb notify to new tidb cause a reload all behaviour.

@D3Hunter

pkg/domain/domain.go
```diff
@@ -1808,6 +1810,25 @@ func (do *Domain) GetPDHTTPClient() pdhttp.Client {
...
 }
 
+func decodePrivilegeEvent(resp clientv3.WatchResponse) PrivilegeEvent {
+	var msg PrivilegeEvent
+	for _, event := range resp.Events {
+		if event.Kv != nil {
+			val := event.Kv.Value
+			if len(val) > 0 {
+				err := json.Unmarshal(val, &msg)
+				if err == nil {
```
lcwangchao, line 1820: better to log an error message here.

pkg/domain/domain.go
```diff
@@ -1808,6 +1810,25 @@ func (do *Domain) GetPDHTTPClient() pdhttp.Client {
...
+					break
+				}
+			}
+		}
+	}
+	if len(msg.UserList) == 0 {
```
lcwangchao, line 1826: Any case that `len(UserList) == 0 and All == false` and we should still update all privileges? If not, I think it's better to remove this logic.

pkg/domain/domain.go
D3Hunter: can you put this in the comment, else we don't know the context for this branch

D3Hunter: i see there is a `if len(val) > 0 {` branch above, maybe set it there?

pkg/domain/domain.go
tiancaiamao: When something is wrong, `len(msg.UserList) == 0`
Then we fall back to the old reload all way.

pkg/domain/domain.go
tiancaiamao: Done

pkg/domain/domain.go
tiancaiamao: Done

pkg/privilege/privileges/cache.go
```diff
@@ -711,7 +766,40 @@ func (p *MySQLPrivilege) LoadDBTable(ctx sqlexec.RestrictedSQLExecutor) error {
...
+		return ret
+	}
+
+	switch {
+	case x.DB > y.DB:
+		return 1
+	case x.DB < y.DB:
+		return -1
+	}
+	return 0
```
D3Hunter, line 780:
```suggestion
	return strings.Compare(x.DB, y.DB)
```

pkg/privilege/privileges/cache.go
```diff
@@ -711,7 +766,40 @@ func (p *MySQLPrivilege) LoadDBTable(ctx sqlexec.RestrictedSQLExecutor) error {
...
+	case x.DB < y.DB:
+		return -1
+	}
+
+	switch {
+	case x.TableName > y.TableName:
+		return 1
+	case x.TableName < y.TableName:
+		return -1
+	}
```
D3Hunter, line 801: doitto

D3Hunter (commented):
rest lgtm

pkg/domain/domain.go
```diff
@@ -1808,6 +1810,28 @@ func (do *Domain) GetPDHTTPClient() pdhttp.Client {
...
+				}
+				logutil.BgLogger().Warn("decodePrivilegeEvent unmarshal fail", zap.Error(err))
+			}
+		}
+	}
+	// In case something is wrong, for example, old version tidb mixed with newer, the unmarshal would fail.
+	// Then we fallback to the old way: reload all the users.
+	if len(msg.UserList) == 0 {
+		msg.All = true
+	}
```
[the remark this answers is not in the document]
tiancaiamao: This differs from the old logic.   @D3Hunter 
What happen if something is wrong? if val does have data, but the data is not what we expected?
Neither the UserList contains data nor the ALL flag is set.
Our expected case is reload all on error to tolerance the cases, but the suggested change here is ignore on error.

pkg/domain/domain.go
```diff
@@ -1822,7 +1846,7 @@ func (do *Domain) LoadPrivilegeLoop(sctx sessionctx.Context) error {
 	var watchCh clientv3.WatchChan
 	duration := 5 * time.Minute
 	if do.etcdClient != nil {
-		watchCh = do.etcdClient.Watch(context.Background(), privilegeKey)
+		watchCh = do.etcdClient.Watch(do.ctx, privilegeKey)
```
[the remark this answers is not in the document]
tiancaiamao: Non-critical.
We don't guarantee the notification always works, there is also a 5min ticker here to reload the data.

pkg/domain/domain.go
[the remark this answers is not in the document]
D3Hunter: > What happen if something is wrong? if val does have data, but the data is not what we expected?

if no third-part involved, older version tidb write a `empty value`, new version write a `valid json`, so when will the data be invalid? If it's caused by bug, then let's fix it, else I think the event consumer should do as what the producer told it to do. and maybe you can add a `intest.Assert` to check that either `All=true` or the `len(userList) > 0`, the producer side should make sure no such invalid data are sent to ETCD

and as you have mentioned, the event is not that critical, i think it's acceptable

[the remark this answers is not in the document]
lcwangchao: > > What happen if something is wrong? if val does have data, but the data is not what we expected?
> 
> if no third-part involved, older version tidb write a `empty value`, new version write a `valid json`, so when will the data be invalid? If it's caused by bug, then let's fix it, else I think the event consumer should do as what the producer told it to do. and maybe you can add a `intest.Assert` to check that either `All=true` or the `len(userList) > 0`, the producer side should make sure no such invalid data are sent to ETCD
> 
> and as you have mentioned, the event is not that critical, i think it's acceptable

I agree that we should ignore the message and skip to update privilege cache when failed to decode the json and only handle the case `val == ""`. Because we also have a loop to guarantee the cache should be updated in a certain time, it is not a big deal.

We can also generate some warnings logs when meet a invalid json.

lcwangchao (approved):
LGTM except the disussion: https://github.com/pingcap/tidb/pull/57042/files#r1853232658

D3Hunter (approved):

BornChanger (approved):
br part lgtm

PR #57044, opened 2024-10-31
session: make `TxnInfo()` return even if process info is empty

### LINKED ISSUE #57043: DDL internal txn reports "GC life time is shorter than transaction duration"
## Bug Report

1. Prepare a large dataset.
2. Create an index for a wide column.
3. It may take more than 10 min to finish the partial import.

Adding index complete without errors.

The state of adding index becomes "rollback done". 
```
GC life time is shorter than transaction duration
```

b314a9c6037b

tangenta:

Issue Number: close #57043

Problem Summary:

https://github.com/pingcap/tidb/blob/c240fdf085ce24c6110a6764e8b9657a3788abfe/pkg/session/session.go#L4026-L4030

`TxnInfo()` could be nil if process info is not set(see the implementation of `TxnInfo()`). Such sessions are commonly used in DDL/DXF:

https://github.com/pingcap/tidb/blob/c240fdf085ce24c6110a6764e8b9657a3788abfe/pkg/ddl/backfilling_operators.go#L533-L539

And they only invoke `session.Begin()` before using:

https://github.com/pingcap/tidb/blob/c240fdf085ce24c6110a6764e8b9657a3788abfe/pkg/ddl/index_cop.go#L43-L48

As a result, the `startTS` from internal DDL sessions will not be reported to etcd, and GC safepoint will be advanced unexpectedly.

Make TxnInfo() return even if process info is empty.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/session/session.go
```diff
@@ -4027,6 +4027,10 @@ func GetStartTSFromSession(se any) (startTS, processInfoID uint64) {
 	if txnInfo != nil {
 		startTS = txnInfo.StartTS
 		processInfoID = txnInfo.ConnectionID
+	} else {
+		tmp.sessionVars.TxnCtxMu.Lock()
```
YangKeao, line 4031: `sessionVars.TxnCtx` will not be cleaned before executing another statement, so using the `StartTS` from `TxnCtx` will block the GC for very long time if some internal sessions are not used (e.g. stored in a pool).

It's possible to clean the `sessionVars.TxnCtx`. But IMO, it'd be simpler to use:

```go
txn := tmp.Txn(false)
if txn != nil {
    startTS = txn.StartTS()
}
```

It should have the same value with `txnInfo.StartTS` (ref `changePendingToValid`, it sets the `txnInfo.StartTS` with `t.StartTS()`), and it doesn't depend on the `ProcessInfo`.

pkg/testkit/mocksessionmanager.go
```diff
@@ -158,8 +158,8 @@ func (msm *MockSessionManager) GetInternalSessionStartTSList() []uint64 {
 			txn := se.TxnInfo()
```
YangKeao, line 158: As it also used `TxnInfo`, should we also modify it to align with the logic in `GetStartTSFromSession`?

pkg/session/session.go
tangenta: https://github.com/pingcap/tidb/pull/57044/commits/2cfe63653e9a67e221d560b7345a3cd1657617f9

tangenta: Maybe we still need `TxnInfo()` because `processInfoID` will be used to filter analyze SQL.

pkg/session/session.go
```diff
@@ -467,6 +467,7 @@ func (s *session) FieldList(tableName string) ([]*resolve.ResultField, error) {
 }
 
 // TxnInfo returns a pointer to a *copy* of the internal TxnInfo, thus is *read only*
+// ProcessInfo fields may not initialize if this is a session that used internally.
```
tangenta, line 470:
```suggestion
// Process field may not initialize if this is a session used internally.
```

YangKeao (approved):
LGTM

lcwangchao (approved):

D3Hunter (approved):

PR #57049, opened 2024-11-01
planner: address collation ambiguity in scalar function construction during predicate simplification.

### LINKED ISSUE #56479: unexpected panic during PredicateSimplification
## Bug Report

``` SQL
CREATE TABLE `tl6e913fb9` (
  `col_36` varchar(175) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT 'asMF',
  KEY `idx_35_5` (`col_36`(1)),
  PRIMARY KEY (`col_36`) /*T![clustered_index] NONCLUSTERED */,
  KEY `idx_65` (`col_36`(5))
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;

with cte_192 ( col_1101,col_1102,col_1103,col_1104 ) AS ( select  /*+ use_index_merge( tl6e913fb9 ) */   replace( tl6e913fb9.col_36 , tl6e913fb9.col_36 , tl6e913fb9.col_36 ) as r0 , space( 0 ) as r1 , min( distinct  tl6e913fb9.col_36 ) as r2 , count( distinct  tl6e913fb9.col_36 ) as r3 from tl6e913fb9 where tl6e913fb9.col_36 between 'n92ok$B%W#UU%O' and '()c=KVQ=T%-vzGJ' and tl6e913fb9.col_36 in ( 'T+kf' ,'Lvluod2H' ,'3#Omx@pC^fFkeH' ,'=b$z' ) group by tl6e913fb9.col_36  having tl6e913fb9.col_36 = 'xjV@' or IsNull( tl6e913fb9.col_36 ) ) ( select 1,col_1101,col_1102,col_1103,col_1104 from cte_192 where not( IsNull( cte_192.col_1102 ) ) order by 1,2,3,4,5 limit 72850972 );
```

No error
ERROR 1105 (HY000): runtime error: invalid memory address or nil pointer dereference
master

dash12653:

Issue Number: close #56479 

Problem Summary:

We can simplify the sql as follows:

DROP TABLE t1;

CREATE TABLE `t1` (
  `c1` VARCHAR(175) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT 'asMF'
);

SELECT * 
FROM t1 
WHERE c1 BETWEEN 'string1' AND 'string2' AND (c1 = 'string3' OR IsNull(c1));

When rewriting the BETWEEN and AND clause, the collations of string1 and string2 will be set to c1's collation—utf8mb4_unicode_ci rather than collation_connection.

During predicate simplification, a scalar function 'ge' will be constructed using string1 and string3 as parameters. However, since string1 has a collation of utf8mb4_unicode_ci and string3 has a collation of utf8mb4_general_ci (collation_connection), and both of them have a coercibility of 4, there is ambiguity regarding which collation to use. This leads to an failure to construct a new scalar function, which will lead a panic.

There's an additional concern: if we replace "BETWEEN and AND" with "<= and >=", then both string1 and string3 have a collation of utf8mb4_general_ci(collation_connection), then during predicate simplification, string1 and string3 would be compared using the utf8mb4_general_ci collation. This might lead to potential incorrect results.

Maybe we can:

During predicate simplification, we could reset the collation of the constants (string1, string2, string3) to match the collation of the column c1 to address the problems.
### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/rule_predicate_simplification.go
```diff
@@ -221,6 +221,19 @@ func unsatisfiable(ctx base.PlanContext, p1, p2 expression.Expression) bool {
...
+	// See details at https://github.com/pingcap/tidb/issues/56479.
+	evalCtx := ctx.GetExprCtx().GetEvalCtx()
+	if equalValue.GetArgs()[1].GetType(evalCtx).EvalType() == types.ETString && otherValue.GetArgs()[1].GetType(evalCtx).EvalType() == types.ETString {
+		_, equalValueCollation := equalValue.CharsetAndCollation()
+		_, otherValueCollation := otherValue.CharsetAndCollation()
+		// If these two scalar functions have different collations, we don't want to compare them.
+		if equalValueCollation != otherValueCollation {
+			return false
+		}
+		equalValue.GetArgs()[1].GetType(evalCtx).SetCollate(equalValueCollation)
```
AilinKid, line 234: why should we set it back? it has not changed?

AilinKid:
@dash12653 hi, thanks for your contribution, do you mind have some updates for this pull request recently, we are lanching planner-related issue resolution campaign

pkg/planner/core/rule_predicate_simplification.go
dash12653: In the WHERE clause : WHERE c1 BETWEEN 'string1' AND 'string2' AND (c1 = 'string3' OR IsNull(c1)), 'string1' and 'string2' are set to the column's collation (utf8mb4_unicode_ci) when rewriting BETWEEN ... AND, but 'string3' is set to the connection-level collation (utf8mb4_general_ci). Before comparing "string1" and "string3" here, to avoid such collation mismatches, I explicitly reset the collations for the string constants.

dash12653:
> @dash12653 hi, thanks for your contribution, do you mind having some updates for this pull request recently, we are lanching planner-related issue resolution campaign

Thanks for letting me know! I’ll update the PR accordingly. Let me know if you need anything else.

dash12653:
> Considering the problem in the case is actually an error from `NewFunction` but ignored, I would suggest changing the existing `NewFunctionInternal` to `NewFunction` and correctly handling the error.

Thanks for the feedback! Just to clarify, are you suggesting that I should only update the function use and handle the error, and that my previous changes can be discarded?

pkg/planner/core/rule_predicate_simplification.go
AilinKid: we have a `deriveCollation(ctx, funcName, args, retType, retType)` inside, which serves collation mismatch, while it's embedded in cast function, so i guess the better way here is to build a wrapper cast function as L237's new child but not for sure, you can have a try

dash12653: Thanks for the clarification! I’ll check out `deriveCollation` and try building a wrapper cast function like `BuildCastFunction` as you suggested.

AilinKid: sorry, after discussion with @time-and-fate , seems we could use his suggested way, which will be more clear.

time-and-fate:
> > Considering the problem in the case is actually an error from `NewFunction` but ignored, I would suggest changing the existing `NewFunctionInternal` to `NewFunction` and correctly handling the error.
> 
> Thanks for the feedback! Just to clarify, are you suggesting that I should only update the function use and handle the error, and that my previous changes can be discarded?

Yes.
It's not quite an appropriate place to handle collation mismatch here. Maybe we can be smarter in handling collation mismatch, but if we focus on the bug itself, the direct reason is that: the error inside `NewFunctionInternal` is ignored and the returning `nil` remains unhandled, then goes into the next step, causing panic. Besides, `NewFunctionInternal` is also marked deprecated.
So, for this bug, I think it's better to switch to using `NewFunction` and handle the error.

dash12653:
> > > Considering the problem in the case is actually an error from `NewFunction` but ignored, I would suggest changing the existing `NewFunctionInternal` to `NewFunction` and correctly handling the error.
> > 
> > 
> > Thanks for the feedback! Just to clarify, are you suggesting that I should only update the function use and handle the error, and that my previous changes can be discarded?
> 
> Yes. Probably we can be smarter in handling collation mismatch, but it's not quite an appropriate place to add new logic to handle collation mismatch here. If we focus on the bug itself, the direct reason is that: the error inside `NewFunctionInternal` is ignored and the returning `nil` remains unhandled, then goes into the next step, causing panic. Besides, `NewFunctionInternal` is also marked deprecated. So, for this bug, I think it's better to switch to using `NewFunction` and handle the error.

Got it. I’ll try `NewFunction` and handling the error.

time-and-fate (approved):

AilinKid (approved):

PR #57086, opened 2024-11-03
planner: fix cost adjustment for high risk tablescan

### LINKED ISSUE #57085: Planner: Test TestCostModelVer2ScanRowSize tablescan row size is incorrect
## Bug Report

Running test TestCostModelVer2ScanRowSize - you can see that the table scan row size is (scan(1000*logrowsize(80)*tikv_scan_factor(40.7) - when it should be "scan(10000".

(scan(10000*logrowsize(80)*tikv_scan_factor(40.7) <--- 10,000

(scan(1000*logrowsize(80)*tikv_scan_factor(40.7) <--- 1000

+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| tidb_version()                                                                                                                                                                                                                                                           |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Release Version: v8.5.0-alpha-31-g241c10294b
Edition: Community
Git Commit Hash: 241c10294b9c0305db590576700c00de306be573
Git Branch: fixscan
UTC Build Time: 2024-11-03 20:42:45
GoVersion: go1.23.0
Race Enabled: false
Store: unistore |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+

terry1purcell:

Issue Number: close #57085 

Problem Summary:

Code to provide a minimum cost to table scan - should use "max" rather than "min" to ensure that the minimum is actually applied. Current code will produce a lower cost for tablescans greater than 1000 rows that fit the other criteria for requiring a minimum cost to table scans.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/cardinality/selectivity_test.go
+++ b/pkg/planner/cardinality/selectivity_test.go
@@ -1205,7 +1205,7 @@ func TestIgnoreRealtimeStats(t *testing.T) {
 	testKit := testkit.NewTestKit(t, store)
 	testKit.MustExec("use test")
 	testKit.MustExec("drop table if exists t")
-	testKit.MustExec("create table t(a int, b int, index ib(b))")
+	testKit.MustExec("create table t(a int, b int)")
 	h := dom.StatsHandle()
 	require.NoError(t, h.HandleDDLEvent(<-h.DDLEventCh()))
 
--- a/pkg/planner/core/plan_cost_ver2.go
+++ b/pkg/planner/core/plan_cost_ver2.go
@@ -173,7 +173,7 @@ func (p *PhysicalTableScan) GetPlanCostVer2(taskType property.TaskType, option *
 		// hasHighModifyCount tracks the high risk of a tablescan where auto-analyze had not yet updated the table row count
 		hasHighModifyCount := tblColHists.ModifyCount > tblColHists.RealtimeCount
 		// hasLowEstimate is a check to capture a unique customer case where modifyCount is used for tablescan estimate (but it not adequately understood why)
-		hasLowEstimate := rows > 1 && int64(rows) < tblColHists.RealtimeCount && int64(rows) <= tblColHists.ModifyCount
+		hasLowEstimate := rows > 1 && tblColHists.ModifyCount < tblColHists.RealtimeCount && int64(rows) <= tblColHists.ModifyCount
 		var unsignedIntHandle bool
 		if p.Table.PKIsHandle {
 			if pkColInfo := p.Table.GetPkColInfo(); pkColInfo != nil {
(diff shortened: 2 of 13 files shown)


pkg/planner/cardinality/selectivity_test.go
```diff
@@ -1205,7 +1205,7 @@ func TestIgnoreRealtimeStats(t *testing.T) {
 	testKit := testkit.NewTestKit(t, store)
 	testKit.MustExec("use test")
 	testKit.MustExec("drop table if exists t")
-	testKit.MustExec("create table t(a int, b int, index ib(b))")
```
Rustin170506, line 1208: May I ask why we need to remove this index?

pkg/planner/core/plan_cost_ver2.go
```diff
@@ -162,29 +162,31 @@ func (p *PhysicalTableScan) GetPlanCostVer2(taskType property.TaskType, option *
...
 
-		// preferRangeScan check here is same as in skylinePruning
-		preferRangeScanCondition := allowPreferRangeScan && (tblColHists.Pseudo || tblColHists.RealtimeCount < 1)
+		// hasUnreliableStats is a check for pseudo or zero stats
+		hasUnreliableStats := tblColHists.Pseudo || tblColHists.RealtimeCount < 1
 		// hasHighModifyCount tracks the high risk of a tablescan where auto-analyze had not yet updated the table row count
 		hasHighModifyCount := tblColHists.ModifyCount > tblColHists.RealtimeCount
 		// hasLowEstimate is a check to capture a unique customer case where modifyCount is used for tablescan estimate (but it not adequately understood why)
-		hasLowEstimate := rows > 1 && int64(rows) < tblColHists.RealtimeCount && int64(rows) <= tblColHists.ModifyCount
+		hasLowEstimate := rows > 1 && tblColHists.ModifyCount < tblColHists.RealtimeCount && int64(rows) <= tblColHists.ModifyCount
```
Rustin170506, line 176: Why do we need to change this?

Rustin170506 (commented):
Thanks!

pkg/planner/core/plan_cost_ver2.go
terry1purcell: The goal of the "hasLowEstimate" check is based upon a customer scenario where it was observed that the tablescan estimate used the ModifyCount - which was lower than the RealtimeCount. It was observed in explain, but there was no plan replayer so we could not reproduce the issue. In recent testing - I discovered that the ModifyCount can equal RealtimeCount. That meant that this boolean could be incorrectly set. Therefore - I changed the check to use ModifyCount < RealtimeCount to avoid that situation.

pkg/planner/cardinality/selectivity_test.go
terry1purcell: These tests are to test the realtime stats values. The presence of the index can result in the plan flipping to use an index for some of the tests - since stats are pseudo. Since the test is to demonstrate the realtime stats value and not to test the plan choice, I removed the index so that the plans do not change. Otherwise we need an "ignore index" on the queries to ensure that we keep a tablescan so that the actual realtime stats values can be observed for the test.

Rustin170506 (approved):
Looks good to me. Thanks! 👍

qw4990 (approved):

PR #57103, opened 2024-11-04
planner: add memo stack scheduler and pool management

AilinKid:

Problem Summary:

we did several things in this pull request.
* we have introduced `MemoExpression` in https://github.com/pingcap/tidb/pull/56961, which serves as currency across entire memo optimizing phase.
   * when memo init, we receive logicalPlan tree and convert it into memoExpression tree internally.
   * when XForm is done, we receive a memoExpression as a returned value, leverage unified `CopyIn` to allocate group/insert group for those raw logical plan, while for those group expression already, we just keep as it was.
* remove `*GroupExprHolder` which is previously used as temporary holder for storing binded and matched group expression sub tree. It will be replaced as unified MemoExpression.
* Introduce 3 kind of task: 
   * task_apply_rule
   * task_opt_group
   * task_opt_group_expression
   * the workflow is like below:
```
// Currently, here is how we work.
//
// Singleton Task Stack
//     ┌────┬────┬────┐
//     │    │    │    │
// ┌───┼────┼────┼────┼────────────────────────────────────────
// │ ┌─┼─┐┌─▼─┐┌─▼─┐┌─▼─┐┌───┐┌───┐┌───┐
// │ │ A ││ B ││ B ││ B ││ C ││ C ││ A │      Open Task Stack...
// │ └───┘└───┘└───┘└─┼─┘└─▲─┘└─▲─┘└─▲─┘
// └──────────────────┼────┼────┼────┼─────────────────────────
//                    │    │    │    │
//                    └────┴────┴────┘
// Symbol means:
// A represent OptGroupTask
// A represent OptGroupExpressionTask
// A represent ApplyRuleTask
//
// When memo init is done, the only targeted task is OptGroupTask, say we got
// 3 group expression inside this group, it will trigger and push additional
// 3 OptGroupExpressionTask into the stack. Then task A is wiped out from the
// stack. With the FILO rule, the stack-top B will be popped out and run, from
// which it will find valid rules for its member group expression and encapsulate
// ApplyRuleTask for each of those valid rules. Say we got two valid rules here,
// so it will push another two task with type C into the stack, note, since current
// B's child group hasn't been optimized yet, so the cascades task A will be triggered
// and pushed into the stack as well, and they are queued after rule tasks. then
// the toppest B is wiped out from the stack.
//
// At last, when the stack is out of task calling internally, or forcible mechanism
// is called from the outside, this stack running will be stopped.
//
// State Flow:
//                                                ┌── Opt 4 New Group Expression ──┐
//                                                │                                │
//     ┌────────────────┐            ┌────────────▼───────────┐            ┌───────┴───────┐
//     │ optGroupTask   │  ───────►  │ optGroupExpressionTask │  ───────►  │ ApplyRuleTask │
//     └──────▲─────────┘            └────────────┬───────────┘            └───────────────┘
//            │                                   │
//            └───── Child Opt Group Trigger ─────┘
//
```
* Introduce pool management for memo and stack. 

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/cascades/memo/memo.go
```diff
@@ -16,20 +16,36 @@ package memo
...
+func (mm *Memo) Destroy() {
+	// when a TaskStack itself is useless, we can destroy itself actively.
+	mm.groupIDGen.id = 0
+	mm.rootGroup = nil
+	mm.groups.Init()
+	clear(mm.groupID2Group)
+	clear(mm.hash2GroupExpr)
+	mm.hasher.Reset()
+	MemoPool.Put(mm)
+}
```
winoros, line 43: Is necessary to create a pool for the `Memo` whose lifetime is not very short?
/cc @hawkingrei

pkg/planner/cascades/rule/binder.go
```diff
@@ -143,11 +136,11 @@ type Binder struct {
 	// value, which is used to tell iterator where to start the next iteration.
 	stackInfo []*list.Element
 
-	// expr is the current matched expression dynamically decided during the binder process.
-	holder *GroupExprHolder
+	// holder is the current matched expression dynamically decided during the binder process, lp is unused now.
```
winoros, line 139: What does lp here mean?

pkg/planner/cascades/rule/rule.go
```diff
@@ -54,22 +60,38 @@ func NewBaseRule(tp ruleType, pattern *pattern.Pattern) *BaseRule {
...
+// ID returns the basic ordinal this rule enum type.
+func (r *BaseRule) ID() uint {
+	// impl it in your own rule.
+	return 0
+}
+
+// Promise value return the default promising value of 1.
+func (r *BaseRule) Promise() int {
+	return 1
+}
```
winoros, line 72: What's the usage of this function? And it seems that this function is not registered at the interface.

pkg/planner/cascades/rule/rule.go
```diff
@@ -54,22 +60,38 @@ func NewBaseRule(tp ruleType, pattern *pattern.Pattern) *BaseRule {
 	}
 }
 
+// ID returns the basic ordinal this rule enum type.
+func (r *BaseRule) ID() uint {
+	// impl it in your own rule.
+	return 0
+}
```
winoros, line 67: Or we can use a const itoa list to ensure no two rules use the same ID?

pkg/planner/cascades/rule/rule_type.go
```diff
@@ -14,13 +14,18 @@
...
 
+import "github.com/pingcap/tidb/pkg/planner/cascades/memo"
+
 type ruleType int
 
 const (
-	// DefaultNone indicates this is none rule.
-	DefaultNone ruleType = iota
+	// DEFAULT_NONE indicates this is none rule.
+	DEFAULT_NONE ruleType = iota
```
winoros, line 23: Oh, we just use this `DEFAULT_NONE` as the ID of the BaseRule? or make BaseRule's as `intest.Assert(false, "should not be called")`

pkg/planner/cascades/task/task_opt_group_expression.go
```diff
@@ -0,0 +1,76 @@
...
+	// currently for each opt expression, it should be explore-type.
+}
+
+// NewOptGroupExpressionTask return a targeting optimizing group expression task.
+func NewOptGroupExpressionTask(mctx *memo.MemoContext, ge *memo.GroupExpression) *OptGroupExpressionTask {
+	return &OptGroupExpressionTask{
+		BaseTask:        BaseTask{mctx: mctx},
+		groupExpression: ge,
+	}
+}
```
winoros, line 41: I guess that this object's lifetime is short and may have great impacts on GC?

pkg/planner/cascades/rule/rule_type.go
AilinKid: yes, make sense

pkg/planner/cascades/task/task_opt_group_expression.go
AilinKid: Do you mean the different-type tasks, make sense, currently we only reuse the stack container, maybe the element inside which is as you mentioned here, we should manage a pool of reusing Task obj.

do you think we should manage this Task pool across different sessions (global), or just limited in a session which is only save the reallocation for the same type task which may be just popped before from the stack.

pkg/planner/cascades/rule/rule.go
AilinKid: currently no that necessary, ideally, the physical rule should have a promise value of 2, and the logical rule should have a default 1, this could make the valid rules be sorted with promise, leading to physical-impl be scheduled ahead of logical rules, then derive a base-cost-line for pruning.
but for now, we only think about logical rules, so the promise value is always 1 as returned. just not classify them in the interface.

pkg/planner/cascades/memo/memo.go
hawkingrei: Early optimization is a taboo in software design.

pkg/planner/cascades/rule/binder.go
AilinKid: refactored

Issue #55906, opened 2024-09-06
Implement In-Memory Priority Queue for Auto-Analyze


Rustin170506:
## Overview
Implement a new priority queue that maintains a heap in memory to improve performance and reduce the overhead of rebuilding the queue for auto-analyze operations.

## Goals
- [x] Avoid scanning the table information every time
- [x] Reflect DDL changes in the priority queue
- [x] Ensure DDL events are received with correct reliability and order

## Tasks
1. [x] Implement basic heap structure with thread-safe operations
2. [x] Implement DML change handling
   - [x] Incremental updates from stats cache
   - [x] Score recalculation based on `modify_count` and `count` changes
3. [x] Implement DDL change handling
   - [x] Create `mysql.ddl_notifier` table
   - [x] Implement `DDLNotifier` for fetching and delivering DDL events
   - [x] Implement subscription mechanism for DDL events
4. [x] Implement time-based updates for analysis intervals
5. [x] Implement failover handling
   - [x] Priority queue rebuild after node restart
   - [x] Priority queue rebuild after owner switch
6. [x] Implement test suite
   - [x] Functional tests
   - [x] Scenario tests
   - [ ] Benchmark tests

## Performance Considerations
- [x] Optimize memory usage of in-memory priority queue
- [x] Benchmark performance with large number of tables (1 million+)
- [x] Measure performance of handling large number of DDL events
- [x] Optimize priority queue rebuild process

## Documentation
- [ ] Update developer documentation with new priority queue implementation

## Related PRs
- Design Document PR: #55903

## Unresolved Questions
- [x] Determine most efficient method for looking up table IDs with updates (query `mysql.stats_meta` vs. querying stats cache)

PR #57113, opened 2024-11-04
statistics: Handle the adding index event properly while the job is running

Rustin170506:

Issue Number: ref https://github.com/pingcap/tidb/issues/55906

Problem Summary:

We need to continue tracking the retryable table while the table analysis is in progress. This will prevent a situation where, after the new index is added, there’s no way to analyze it.

I changed the failed jobs to retryable jobs so we can track them. Each time we attempt to push a currently running job, we will mark it as retryable, as we may need to analyze the table again after it has been processed.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/analyze.go
+++ b/pkg/executor/analyze.go
@@ -22,7 +22,6 @@ import (
 	"net"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/failpoint"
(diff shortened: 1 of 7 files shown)


Rustin170506 (commented):

pkg/executor/analyze.go
```diff
@@ -168,6 +169,10 @@ TASKLOOP:
 		}
 	}
 
+	failpoint.Inject("mockStuckAnalyze", func() {
+		time.Sleep(time.Second * 3)
+	})
```
time-and-fate, line 174: Looks like the stability of the test case relies on the timing.
After some thinking, I think we may be able to explicitly control how long it waits for in this way:
```go
if intest.InTest {
	for {
		stop := true
		failpoint.Inject("mockStuckAnalyze", func() {
			stop = false
		})
		if stop {
			break
		}
	}
}
```
Then it will keep waiting after we `Enable` the failpoint and before we `Disable` this failpoint.

time-and-fate (approved):
/hold
Others LGTM.

pkg/executor/analyze.go
Rustin170506: Updated.

Rustin170506: Thanks!

elsa0520 (approved):
LGTM

Issue #56829, opened 2024-10-25
The index data and table data of the TTL table are inconsistent


zeminzhou:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

1. startup a TiDB cluster;
2. create table `cache`;
```
CREATE TABLE `cache` (
  `cache_key` varchar(512) NOT NULL,
  `cache_value` json NOT NULL,
  `created_at` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
  `updated_at` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  `expires` int(11) DEFAULT '-1' COMMENT 'cache will expire after n seconds',
  `expired_at` datetime GENERATED ALWAYS AS (if(`expires` > 0, date_add(`updated_at`, interval `expires` second), date_add(`updated_at`, interval 99 year))) VIRTUAL,
  PRIMARY KEY (`cache_key`) /*T![clustered_index] CLUSTERED */,
  KEY `idx_c_on_created_at` (`created_at`),
  KEY `idx_c_on_expired_at` (`expired_at`),
  KEY `idx_expired_at` (`expired_at`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin /*T![ttl] TTL=`expired_at` + INTERVAL 0 DAY */ /*T![ttl] TTL_ENABLE='ON' */ /*T![ttl] TTL_JOB_INTERVAL='15m' */
```
3. run the following command
```
func main() {
	// Open a connection to the database.
	// Replace the parameters with your own values for the database.
	db, err := sql.Open("mysql", "root@tcp(localhost:4000)/test")
	if err != nil {
		fmt.Println("Error connecting to the database:", err)
		return
	}
	defer db.Close()

	for i := 0; i < 1000000; i++ {
		dateStr := time.Now().Format("2006-01-02 15:04:05")
		sqlStr := fmt.Sprintf("INSERT INTO `cache` (`cache_key`, `cache_value`, `created_at`, `updated_at`, `expires`) VALUES ('%s', '\"%s\"', '%s', '%s', %d) ON DUPLICATE KEY UPDATE expires = expires + 1", dateStr, dateStr, dateStr, dateStr, 60)
		_, err = db.Exec(sqlStr)
		if err != nil {
			panic(err)
		}
		if i%10 == 0 {
[...]

PR #57123, opened 2024-11-05
executor: change the evaluation order of columns in `Update` and `Insert` statements

joechenrh:

Issue Number: ref #56829

Problem Summary:

In the previous logic, when we use `UPDATE` or `INSERT ON DUPLICATE`, the new row will be generated in the following order:

- Fill all the explicitly set columns.
- Evaluate all auto-generated columns. For `UPDATE` and `INSERT`, they are evaluated in `composeGeneratedColumns` and `doDupRowUpdate` respectively.
- Update on-update-now columns if necessary.

However, auto-generated columns may rely on the on-update-now columns to generate value.  For example in https://github.com/pingcap/tidb/issues/56829#issuecomment-2456356173

```sql
CREATE TABLE cache (
  cache_key varchar(512) NOT NULL,
  updated_at datetime NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  expired_at datetime GENERATED ALWAYS AS (if(expires > 0, date_add(updated_at, interval expires second), date_add(updated_at, interval 99 year))) VIRTUAL,
  expires int(11),
  PRIMARY KEY (cache_key) /*T![clustered_index] CLUSTERED */,
  KEY idx_c_on_expired_at (expired_at)
);
```

`expired_at` is generated based on the latest timestamp value from `updated_at`. So we will get wrong `expired_at` value. Even worse, `expired_at` is the part of the index `idx_c_on_expired_at`. So querying data `expired_at` using index scan and full table scan will get different result, since in full table scan, `expired_at` is calculated in real-time.

This also explains https://github.com/pingcap/tidb/issues/56829#issuecomment-2441059720 why changing `VIRTUAL` to `STORED` will not yield such error, although this value itself is incorrect.

To address this problem, this PR refactor the logic of `INSERT ON DUPLICATE` and `UPDATE`. More specifically:
- Move the evalation of auto-generated columns in `updateRecord`.
- ForeignKey Check is also moved into `updateRecord`.
- Extract `errorHandler` function for `UPDATE` and `INSERT` to handle error/warning in `updateRecord`.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/insert_test.go
```diff
@@ -239,6 +239,24 @@ func testInsertOnDuplicateKey(t *testing.T, tk *testkit.TestKit) {
...
+			PRIMARY KEY (cache_key) /*T![clustered_index] CLUSTERED */,
+			KEY idx_c_on_expired_at (expired_at)
+		)`)
+	tk.MustExec("INSERT INTO cache(cache_key, expires) VALUES ('2001-01-01 11:11:11', 60) ON DUPLICATE KEY UPDATE expires = expires + 1")
+	tk.MustExec("select sleep(1)")
+	tk.MustExec("INSERT INTO cache(cache_key, expires) VALUES ('2001-01-01 11:11:11', 60) ON DUPLICATE KEY UPDATE expires = expires + 1")
+	tk.MustExec("admin check table cache")
+	rs1 := tk.MustQuery("select /*+ force_index(test.cache, idx_c_on_expired_at) */ cache_key, expired_at from cache order by cache_key")
+	rs2 := tk.MustQuery("select /*+ ignore_index(test.cache, idx_c_on_expired_at) */ cache_key, expired_at from cache order by cache_key")
+	require.True(t, rs1.Equal(rs2.Rows()))
```
YangKeao, line 258: Should we also assert whether the value is the same?

pkg/executor/insert.go
```diff
@@ -430,8 +443,15 @@ func (e *InsertExec) initEvalBuffer4Dup() {
 }
 
 // doDupRowUpdate updates the duplicate row.
-func (e *InsertExec) doDupRowUpdate(ctx context.Context, handle kv.Handle, oldRow []types.Datum, newRow []types.Datum,
-	extraCols []types.Datum, cols []*expression.Assignment, idxInBatch int, dupKeyMode table.DupKeyCheckMode, autoColIdx int) error {
+func (e *InsertExec) doDupRowUpdate(
```
YangKeao, line 446: It only fixes the `INSERT .. ON DUPLICATE UPDATE .. ` case. As I have tested, the `UPDATE` path also has the same bug:

```
CREATE TABLE cache (
  cache_key varchar(512) NOT NULL,
  updated_at datetime NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  expired_at datetime GENERATED ALWAYS AS (if(expires > 0, date_add(updated_at, interval expires second), date_add(updated_at, interval 99 year))) VIRTUAL,
  expires int(11),
  PRIMARY KEY (cache_key) /*T![clustered_index] CLUSTERED */,
  KEY idx_c_on_expired_at (expired_at)
);

INSERT INTO cache(cache_key, expires) VALUES ('2001-01-01 11:11:11', 60) ON DUPLICATE KEY UPDATE expires = expires + 1;
update cache set expires = expires + 1 where cache_key = '2001-01-01 11:11:11';
```

Then the following two queries will have different result:

```
select /*+ force_index(test.cache, idx_c_on_expired_at) */ cache_key, expired_at from cache order by cache_key;
select /*+ ignore_index(test.cache, idx_c_on_expired_at) */ cache_key, expired_at from cache order by cache_key;
```

pkg/executor/insert.go
```diff
@@ -489,9 +564,12 @@ func (e *InsertExec) doDupRowUpdate(ctx context.Context, handle kv.Handle, oldRo
 			return nil
 		}
 	}
-	_, err := updateRecord(ctx, e.Ctx(), handle, oldRow, newData, assignFlag, e.Table, true, e.memTracker, e.fkChecks, e.fkCascades, dupKeyMode, e.ignoreErr)
-	if err != nil {
-		return err
+
+	if _, err := updateRecord(
```
YangKeao, line 568: Is it possible to move the logic of handling generated column into the `updateRecord`? It seems that the function `updateRecord` also handles the `ON UPDATE` columns (and as we specified all `ON UPDATE` column in this functioin, these logic are meaningless).

Another possible solution is to remove the codes related to `ON UPDATE` columns in `updateRecord` function. However, as the same logic will be used multiple times (in `INSERT .. ON DUPLICATE UPDATE` and a normal `UPDATE` statement), I prefer to write the codes related to generated column in `updateRecord` to avoid repeating the codes.

pkg/executor/insert.go
```diff
@@ -193,7 +197,16 @@ func (e *InsertValues) prefetchDataCache(ctx context.Context, txn kv.Transaction
 }
 
 // updateDupRow updates a duplicate row to a new row.
-func (e *InsertExec) updateDupRow(ctx context.Context, idxInBatch int, txn kv.Transaction, row toBeCheckedRow, handle kv.Handle, _ []*expression.Assignment, dupKeyCheck table.DupKeyCheckMode, autoColIdx int) error {
+func (e *InsertExec) updateDupRow(
```
YangKeao, line 200: BTW, there are a lot of the format changes (e.g. function signature, error handling). Are they reported by linter, or just formated by your IDE/editors?

If it'll not break the CI, I'd suggest not to change them. It'll make this patch cleaner.

joechenrh: Not by linter actually. I used to change format of related code in the PR, to make the function signature clearer.

YangKeao (approved):
LGTM

wjhuang2016 (approved):

joechenrh:
/hold 
schrddl found another problem

joechenrh:
/unhold 
The problem is related to plan builder, not executor. I will fix it in another PR.

Issue #57147, opened 2024-11-05
Merge in Workload Repository from 


wddevries:
The workload repository will persist workload metrics for use by the planner and statistics systems.  These data will be used by:
 * the Statistics Advisor, which will help determine statistics loading priorities,
 * the Index Advisor, which will make recommendations about indexes,
 * the SQL tuning and SQL binding advisor,
 
Select tables will be sampled by two different methods at regular intervals, and persist the results into the "workload_schema" database.  Sampled tables are node-specific and each TiDB instance will sample and write these data independently.

Some tables will be sampled on an interval between 1 and 600 seconds. 
The data from each of these tables will be sampled and recorded in the corresponding tables in 'workload_schema' along with the sampling timestamp and instance id. These tables include:
 * PROCESSLIST
 * DATA_LOCK_WAITS 
 * TIDB_TRX MEMORY_USAGE 
 * DEADLOCKS CLUSTER_LOAD
 * TIDB_HOT_REGIONS 
 * TIKV_STORE_STATUS
 
Other tables will be sampled as part of a snapshot process, which will run every 15 minutes to 2 hours.  When a snapshot is initialized, one row will be written into HIST_SNAPSHOTS, which will contain the snapshot id, the start and end timestamps, and server version information.  The data from each of the following tables will be sampled and recorded in the corresponding tables in 'workload_schema' along with the sampling timestamp, instance id and snapshot id: 

 * CLIENT_ERRORS_SUMMARY_BY_HOST 
 * CLIENT_ERRORS_SUMMARY_BY_USER
 * CLIENT_ERRORS_SUMMARY_GLOBAL 
 * TIKV_REGION_STATUS
 * STATEMENT_SUMMARY

Other tables will probably be added to these lists.

PR #57155, opened 2024-11-06
stmtsummary: implement tidb_statements_stats, a cumulative version of statements_summary

henrybw:

Issue Number: ref #57147 

Problem Summary: The statement summary tables are the current mechanism we have for collecting cumulative metrics about the execution and performance of SQL queries. However, these metrics aren't truly cumulative: they are refreshed at a regular interval, meaning the cumulative metrics collected for queries are reset, and their previous values are made accessible via a separate history table. For the workload repository, we want to be able to snapshot cumulative metrics aggregated over the lifetime of a TiDB instance, which the statement summary tables do not easily provide.

* The statistics captured by the statement summaries for each interval have been factored out into a new struct, `stmtSummaryStats`, that each statement summary interval (`stmtSummaryByDigestElement`) now uses instead. The struct is embedded so existing field accesses continue to work as-is.
* In the statement summary struct itself (`stmtSummaryByDigest`), add a new "cumulative" set of statistics (i.e. another `stmtSummaryStats` struct).
* Create a new memory table, `information_schema.tidb_statements_stats`, that uses the same `stmtSummaryRetriever` as the statement summary tables.
* Modify the column factories for the statement summary rows reader to accept an additional `stmtSummaryStats` parameter. For the existing statement summary tables, this parameter is the `stmtSummaryStats` struct embedded in the `stmtSummaryByDigestElement` (i.e. the same values as before). For the new cumulative statement stats table, this parameter is the cumulative `stmtSummaryStats` struct in the `stmtSummaryByDigest`. The column factories then read from this `stmtSummaryStats` parameter, rather than the `stmtSummaryByDigestElement` struct directly, thus making them polymorphic and able to work for both the cumulative and statements summary tables.
* Add a cluster-wide version of the statement stats table: `cluster_tidb_statements_stats`.

(The diff for this PR is pretty large, so I'd recommend reading each individual commit one at a time, since they build on each other incrementally.)

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code
```
- Started tidb-server.
- Ran one query, e.g. `select 1;`.
- Verified that a row appeared for this query.
- Re-ran the query several more times.
- Verified that the cumulative stats in that row updated accordingly.
- Verified that the corresponding row in cluster_tidb_statements_stats updated accordingly.
```

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Add a new system table `INFORMATION_SCHEMA.TIDB_STATEMENTS_STATS` to provide cumulative metrics about the execution and performance of SQL queries on TiDB nodes.
```


djshow832 (approved):

xhebox (approved):

D3Hunter:
/approve

infoschema part lgtm

pkg/util/stmtsummary/reader.go
```diff
@@ -64,6 +64,27 @@ func NewStmtSummaryReader(user *auth.UserIdentity, hasProcessPriv bool, cols []*
 	return reader
 }
 
+// GetStmtSummaryCumulativeRows gets statement summary rows with cumulative metrics.
+func (ssr *stmtSummaryReader) GetStmtSummaryCumulativeRows() [][]types.Datum {
```
Benjamin2037, line 68: Can we add some input parameters to only needed stats back？like only return records after some ts

henrybw: Hmm, that is a good idea, but I'm not sure if it makes sense in this context. For in-memory tables like this, I think the kind of filtering you're referring to would be done at a higher level, by the executor.

Perhaps we could implement a "pushdown" mechanism (like predicate pushdown to TiKV) for in-memory tables like this. While that could be a useful feature, I think building something like that would be a pretty large project, and is thus out of scope for this PR.

Rustin170506 (approved):
Thanks! :shipit:

hawkingrei:
@henrybw This PR introduces a data race. https://github.com/pingcap/tidb/issues/58403

PR #57305, opened 2024-11-12
session: track LastCommitTS in SessionVars and check timestamps of later txns are larger

### LINKED ISSUE #57165: server: enhance the ts check considering commit timestamp
## Enhancement
After a write transaction is committed, TiDB, as the transaction coordinator, knows the `commit_ts` used for the write transaction. This timestamp can be used as a reference to check the validity of subsequent read requests’ `read_ts`. In theory, all read requests occurring after the write transaction should use a `read_ts` greater than the previous known transaction’s `commit_ts`. 

This approach could help to expose issues like https://github.com/pingcap/tidb/issues/56809 more quickly.

Besides, this timestamp could be considered for updating the [`tidb_low_resolution_ts`](https://docs.pingcap.com/tidb/stable/system-variables#tidb_low_resolution_tso) with the known valid timestamps encountered.

b6g:

Issue Number: close #57165

Problem Summary:

This PR checks the invariant that timestamps of transactions in a session should increase monotonically.

It saves the timestamp of the last transaction in `SessionVars`, after the transaction is committed. The saved timestamp is compared with the `start_ts` and the `commit_ts` of the next transaction.

This PR depends on the TiKV client-go PR, https://github.com/tikv/client-go/pull/1489, which exports `commit_ts` of the transaction.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/kv/kv.go
+++ b/pkg/kv/kv.go
@@ -251,7 +251,7 @@ type Transaction interface {
 	IsReadOnly() bool
 	// StartTS returns the transaction start timestamp.
 	StartTS() uint64
-	// CommitTS returns the transaction commit timestamp.
+	// CommitTS returns the transaction commit timestamp if it is committed; otherwise it returns 0.
 	CommitTS() uint64
 	// Valid returns if the transaction is valid.
 	// A transaction become invalid after commit or rollback.
--- a/pkg/kv/kv.go
+++ b/pkg/kv/kv.go
@@ -251,6 +251,8 @@ type Transaction interface {
 	IsReadOnly() bool
 	// StartTS returns the transaction start timestamp.
 	StartTS() uint64
+	// CommitTS returns the transaction commit timestamp.
+	CommitTS() uint64
 	// Valid returns if the transaction is valid.
 	// A transaction become invalid after commit or rollback.
 	Valid() bool
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -936,9 +936,8 @@ func (s *session) CommitTxn(ctx context.Context) error {
 				zap.String("sql", redact.String(s.sessionVars.EnableRedactLog, s.sessionVars.StmtCtx.OriginalSQL)),
 			)
 			return kv.ErrAssertionFailed
-		} else {
-			s.sessionVars.LastCommitTS = s.txn.lastCommitTS
 		}
+		s.sessionVars.LastCommitTS = s.txn.lastCommitTS
 	}
 
 	// record the TTLInsertRows in the metric
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -930,7 +930,7 @@ func (s *session) CommitTxn(ctx context.Context) error {
 	if err == nil && s.txn.lastCommitTS > 0 {
 		// lastCommitTS could be the same, e.g. when the txn is considered readonly
 		if s.txn.lastCommitTS < s.sessionVars.LastCommitTS {
-			logutil.BgLogger().Fatal("check lastCommitTS failed",
+			logutil.BgLogger().Panic("check lastCommitTS failed",
 				zap.Uint64("sessionLastCommitTS", s.sessionVars.LastCommitTS),
 				zap.Uint64("txnLastCommitTS", s.txn.lastCommitTS),
 				zap.String("sql", s.sessionVars.StmtCtx.OriginalSQL),
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -927,6 +927,19 @@ func (s *session) CommitTxn(ctx context.Context) error {
 		s.sessionVars.StmtCtx.MergeExecDetails(nil, commitDetail)
 	}
 
+	if err == nil && s.txn.lastCommitTS > 0 {
+		// lastCommitTS could be the same, e.g. when the txn is considered readonly
+		if s.txn.lastCommitTS < s.sessionVars.LastCommitTS {
+			logutil.BgLogger().Fatal("check lastCommitTS failed",
+				zap.Uint64("sessionLastCommitTS", s.sessionVars.LastCommitTS),
+				zap.Uint64("txnLastCommitTS", s.txn.lastCommitTS),
+				zap.String("sql", s.sessionVars.StmtCtx.OriginalSQL),
+			)
+		} else {
+			s.sessionVars.LastCommitTS = s.txn.lastCommitTS
+		}
+	}
+
 	// record the TTLInsertRows in the metric
 	metrics.TTLInsertRowsCount.Add(float64(s.sessionVars.TxnCtx.InsertTTLRowsCount))
 
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -933,7 +933,7 @@ func (s *session) CommitTxn(ctx context.Context) error {
 			logutil.BgLogger().Panic("check lastCommitTS failed",
 				zap.Uint64("sessionLastCommitTS", s.sessionVars.LastCommitTS),
 				zap.Uint64("txnLastCommitTS", s.txn.lastCommitTS),
-				zap.String("sql", s.sessionVars.StmtCtx.OriginalSQL),
+				zap.String("sql", redact.String(s.sessionVars.EnableRedactLog, s.sessionVars.StmtCtx.OriginalSQL)),
 			)
 		} else {
 			s.sessionVars.LastCommitTS = s.txn.lastCommitTS
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -930,11 +930,12 @@ func (s *session) CommitTxn(ctx context.Context) error {
 	if err == nil && s.txn.lastCommitTS > 0 {
 		// lastCommitTS could be the same, e.g. when the txn is considered readonly
 		if s.txn.lastCommitTS < s.sessionVars.LastCommitTS {
-			logutil.BgLogger().Panic("check lastCommitTS failed",
+			logutil.BgLogger().Error("check lastCommitTS failed",
 				zap.Uint64("sessionLastCommitTS", s.sessionVars.LastCommitTS),
 				zap.Uint64("txnLastCommitTS", s.txn.lastCommitTS),
 				zap.String("sql", redact.String(s.sessionVars.EnableRedactLog, s.sessionVars.StmtCtx.OriginalSQL)),
 			)
+			return kv.ErrAssertionFailed
 		} else {
 			s.sessionVars.LastCommitTS = s.txn.lastCommitTS
 		}
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -935,7 +935,8 @@ func (s *session) CommitTxn(ctx context.Context) error {
 				zap.Uint64("txnLastCommitTS", s.txn.lastCommitTS),
 				zap.String("sql", redact.String(s.sessionVars.EnableRedactLog, s.sessionVars.StmtCtx.OriginalSQL)),
 			)
-			return kv.ErrAssertionFailed
+			return fmt.Errorf("txn commit_ts:%d is before session last_commit_ts:%d",
+				s.txn.lastCommitTS, s.sessionVars.LastCommitTS)
 		}
 		s.sessionVars.LastCommitTS = s.txn.lastCommitTS
 	}
--- a/pkg/session/txn.go
+++ b/pkg/session/txn.go
@@ -70,6 +70,9 @@ type LazyTxn struct {
 
 	// mark the txn enables lazy uniqueness check in pessimistic transactions.
 	lazyUniquenessCheckEnabled bool
+
+	// commit ts of the last successful transaction, to ensure ordering of TS
+	lastCommitTS uint64
 }
 
 // GetTableInfo returns the cached index name.
@@ -431,7 +434,11 @@ func (txn *LazyTxn) Commit(ctx context.Context) error {
 		}
 	})
 
-	return txn.Transaction.Commit(ctx)
+	err := txn.Transaction.Commit(ctx)
+	if err == nil {
+		txn.lastCommitTS = txn.Transaction.CommitTS()
+	}
+	return err
 }
 
 // Rollback overrides the Transaction interface.
--- a/pkg/session/txn.go
+++ b/pkg/session/txn.go
@@ -437,6 +437,11 @@ func (txn *LazyTxn) Commit(ctx context.Context) error {
 	err := txn.Transaction.Commit(ctx)
 	if err == nil {
 		txn.lastCommitTS = txn.Transaction.CommitTS()
+		failpoint.Inject("mockFutureCommitTS", func(val failpoint.Value) {
+			if ts, ok := val.(int); ok {
+				txn.lastCommitTS = uint64(ts)
+			}
+		})
 	}
 	return err
 }
--- a/pkg/sessiontxn/isolation/base.go
+++ b/pkg/sessiontxn/isolation/base.go
@@ -273,6 +273,8 @@ func (p *baseTxnContextProvider) getTxnStartTS() (uint64, error) {
 	return txn.StartTS(), nil
 }
 
+// TODO: replace usePresetStartTS with a new method StartTSFromPD to make it clear that
+// the timestamp is not allocated by TSO.
 func (p *baseTxnContextProvider) usePresetStartTS() bool {
 	return p.constStartTS != 0 || p.sctx.GetSessionVars().SnapshotTS != 0
 }
--- a/pkg/sessiontxn/isolation/base.go
+++ b/pkg/sessiontxn/isolation/base.go
@@ -36,6 +36,7 @@ import (
 	"github.com/pingcap/tidb/pkg/table/temptable"
 	"github.com/pingcap/tidb/pkg/tablecodec"
 	"github.com/pingcap/tidb/pkg/util/logutil"
+	"github.com/pingcap/tidb/pkg/util/redact"
 	"github.com/pingcap/tidb/pkg/util/tableutil"
 	"github.com/pingcap/tidb/pkg/util/tracing"
 	tikvstore "github.com/tikv/client-go/v2/kv"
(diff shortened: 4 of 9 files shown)


pkg/session/session.go
```diff
@@ -929,6 +929,21 @@ func (s *session) CommitTxn(ctx context.Context) error {
 		s.sessionVars.StmtCtx.MergeExecDetails(nil, commitDetail)
 	}
 
+	if err == nil {
+		// save CommitTS in sessionVars for invariant check
+		// TODO: enable LastCommitTS with a session variable
```
b6g, line 934: I will add a session variable.

pkg/sessiontxn/isolation/base.go
```diff
@@ -304,6 +306,14 @@ func (p *baseTxnContextProvider) ActivateTxn() (kv.Transaction, error) {
 		sessVars.SetInTxn(true)
 	}
 
+	// verify start_ts is later than any previous commit_ts in the session
+	if sessVars.LastCommitTS > 0 && sessVars.LastCommitTS > sessVars.TxnCtx.StartTS {
```
b6g, line 310: I will add unittests to cover the two checks.

pkg/session/session.go
b6g: Added in a separate PR https://github.com/pingcap/tidb/pull/57313, to make review easier.

pkg/sessiontxn/isolation/base.go
```diff
@@ -304,6 +306,14 @@ func (p *baseTxnContextProvider) ActivateTxn() (kv.Transaction, error) {
 		sessVars.SetInTxn(true)
 	}
 
+	// verify start_ts is later than any previous commit_ts in the session
```
dveeden, line 309: Is this comment correct? Isnt it checking the LastCommitTS is later than the StartTS instead of the other way around?

go.mod
```diff
@@ -323,6 +323,8 @@ replace (
 	github.com/go-ldap/ldap/v3 => github.com/YangKeao/ldap/v3 v3.4.5-0.20230421065457-369a3bab1117
 	github.com/pingcap/tidb/pkg/parser => ./pkg/parser
 
+	github.com/tikv/client-go/v2 => ../client-go
```
dveeden, line 326: I think this should be removed. I assume this is to do testing together with https://github.com/tikv/client-go/pull/1489 before it is merged?

pkg/sessiontxn/isolation/base.go
b6g: I think it is correct. It logs fatal if LastCommitTS is later than StartTS.

go.mod
b6g: Yes, you're right. I shouldn't have included this local change.

b6g: Is it the right process?
1. get https://github.com/tikv/client-go/pull/1489 merged
2. update client-go in go.mod to include the change

dveeden: Yes

pkg/session/session.go
b6g: Per discussion in https://github.com/pingcap/tidb/pull/57313#discussion_r1856131594, we don't need a session variable.

b6g:
The client-go PR is merged and go.mod is updated. The PR is ready for review. PTAL Thanks!

go.mod
```diff
@@ -109,7 +109,7 @@ require (
 	github.com/tdakkota/asciicheck v0.2.0
```
hawkingrei, line 109: Please run the ```make bazel_preapre``` and upload the dependencies.

b6g:
Some tests failed with
```Error in download_and_extract: java.io.IOException: Error downloading [http://bazel-cache.pingcap.net:8080/gomod/github.com/tikv/client-go/v2/com_github_tikv_client_go_v2-v2.0.8-0.20241127071526-89643b0e8c9e.zip, http://ats.apps.svc/gomod/github.com/tikv/client-go/v2/com_github_tikv_client_go_v2-v2.0.8-0.20241127071526-89643b0e8c9e.zip, https://cache.hawkingrei.com/gomod/github.com/tikv/client-go/v2/com_github_tikv_client_go_v2-v2.0.8-0.20241127071526-89643b0e8c9e.zip, https://storage.googleapis.com/pingcapmirror/gomod/github.com/tikv/client-go/v2/com_github_tikv_client_go_v2-v2.0.8-0.20241127071526-89643b0e8c9e.zip] to /home/jenkins/.tidb/tmp/c81bba9845037d26d5cccd5f3ccb189e/external/com_github_tikv_client_go_v2/temp2943251639532646796/com_github_tikv_client_go_v2-v2.0.8-0.20241127071526-89643b0e8c9e.zip: GET returned 404 Not Found
```

Was it because the client-go PR was just merged and it takes time to create `http://bazel-cache.pingcap.net:8080/gomod/github.com/tikv/client-go/v2/com_github_tikv_client_go_v2-v2.0.8-0.20241127071526-89643b0e8c9e.zip`?

b6g:
> Some tests failed with
> 
> Was it because the client-go PR was just merged and it takes time to create `http://bazel-cache.pingcap.net:8080/gomod/github.com/tikv/client-go/v2/com_github_tikv_client_go_v2-v2.0.8-0.20241127071526-89643b0e8c9e.zip`?

The error was,
```
Error in download_and_extract: java.io.IOException: Error downloading [http://bazel-cache.pingcap.net:8080/gomod/github.com/tikv/client-go/v2/com_github_tikv_client_go_v2-v2.0.8-0.20241127071526-89643b0e8c9e.zip, http://ats.apps.svc/gomod/github.com/tikv/client-go/v2/com_github_tikv_client_go_v2-v2.0.8-0.20241127071526-89643b0e8c9e.zip, https://cache.hawkingrei.com/gomod/github.com/tikv/client-go/v2/com_github_tikv_client_go_v2-v2.0.8-0.20241127071526-89643b0e8c9e.zip, https://storage.googleapis.com/pingcapmirror/gomod/github.com/tikv/client-go/v2/com_github_tikv_client_go_v2-v2.0.8-0.20241127071526-89643b0e8c9e.zip] to /root/.cache/bazel/_bazel_root/37ff7b0eddea115f60ab9f853788d7f5/external/com_github_tikv_client_go_v2/temp12921244511850196136/com_github_tikv_client_go_v2-v2.0.8-0.20241127071526-89643b0e8c9e.zip: GET returned 404 Not Found
```

b6g:
> Some tests failed with
> 
> Was it because the client-go PR was just merged and it takes time to create `http://bazel-cache.pingcap.net:8080/gomod/github.com/tikv/client-go/v2/com_github_tikv_client_go_v2-v2.0.8-0.20241127071526-89643b0e8c9e.zip`?

How can I update the client-go version? I tried,
```
go get github.com/tikv/client-go/v2@89643b0e8c9ee8cbd0fce4f264a8765c5665db17
```
But these tests failed.

hawkingrei:
> > Some tests failed with
> > Was it because the client-go PR was just merged and it takes time to create `http://bazel-cache.pingcap.net:8080/gomod/github.com/tikv/client-go/v2/com_github_tikv_client_go_v2-v2.0.8-0.20241127071526-89643b0e8c9e.zip`?
> 
> How can I update the client-go version? I tried,
> 
> ```
> go get github.com/tikv/client-go/v2@89643b0e8c9ee8cbd0fce4f264a8765c5665db17
> ```
> 
> But these tests failed.

I has uploaded it. you can try the unit test again.

hawkingrei:
BTW, Can you sync your code with master ?

b6g:
> BTW, Can you sync your code with master ?

Will do. I am fixing the broken unittests.

b6g:
The unittests are fixed. PTAL

pkg/session/txn.go
```diff
@@ -70,6 +70,9 @@ type LazyTxn struct {
 
 	// mark the txn enables lazy uniqueness check in pessimistic transactions.
 	lazyUniquenessCheckEnabled bool
+
+	// commit ts of the last successful transaction, to ensure ordering of TS
+	lastCommitTS uint64
```
cfzjywxk, line 75: This field seems to overlap with the newly added `CommitTS` in the KV interface. Additionally, `LazyTxn` is a structure that is no longer needed once a transaction is completed. Or just retain only one of them?

b6g: This is because `kv.Transaction` will be set to nil in `changeToInvalid` in `LazyTxn.Commit`.

br/pkg/backup/BUILD.bazel
```diff
@@ -3,6 +3,7 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 go_library(
     name = "backup",
     srcs = [
+        "binding__failpoint_binding__.go",
```
b6g, line 6: This is added by `make bazel_prepare`. The bazel version I am using is,
```
$ bazel version
Bazelisk version: 1.22.0
Build label: 7.4.0
Build target: @@//src/main/java/com/google/devtools/build/lib/bazel:BazelServer
Build time: Tue Oct 22 17:25:21 2024 (1729617921)
Build timestamp: 1729617921
Build timestamp as int: 1729617921
```

b6g:
Merge conflicts are resolved. PTAL

pkg/session/txn.go
cfzjywxk: How about returning the `commitTS` of before `changeToInvalid`, like
```
- func (s *session) doCommit(ctx context.Context) (error) {
+ func (s *session) doCommit(ctx context.Context) (uint64, error) {
	if !s.txn.Valid() {
		return nil
	}

	defer func() {
        // Return the commitTS if error is nil
		s.txn.changeToInvalid()
		s.sessionVars.SetInTxn(false)
		s.sessionVars.ClearDiskFullOpt()
	}()
   ...
}
```

pkg/session/session.go
```diff
@@ -925,6 +925,19 @@ func (s *session) CommitTxn(ctx context.Context) error {
 		s.sessionVars.StmtCtx.MergeExecDetails(nil, commitDetail)
 	}
 
+	if err == nil && s.txn.lastCommitTS > 0 {
+		// lastCommitTS could be the same, e.g. when the txn is considered readonly
+		if s.txn.lastCommitTS < s.sessionVars.LastCommitTS {
+			logutil.BgLogger().Panic("check lastCommitTS failed",
+				zap.Uint64("sessionLastCommitTS", s.sessionVars.LastCommitTS),
+				zap.Uint64("txnLastCommitTS", s.txn.lastCommitTS),
+				zap.String("sql", s.sessionVars.StmtCtx.OriginalSQL),
```
ekexium, line 934:
```suggestion
				zap.String("sql", redact.String(s.sessionVars.EnableRedactLog, s.sessionVars.StmtCtx.OriginalSQL)),
```

pkg/kv/kv.go
```diff
@@ -251,6 +251,8 @@ type Transaction interface {
 	IsReadOnly() bool
 	// StartTS returns the transaction start timestamp.
 	StartTS() uint64
+	// CommitTS returns the transaction commit timestamp.
```
ekexium, line 254:
```suggestion
	// CommitTS returns the commit timestamp of the already committed transaction, or zero if it's not committed yet.
```

pkg/sessiontxn/isolation/base.go
```diff
@@ -268,6 +270,10 @@ func (p *baseTxnContextProvider) getTxnStartTS() (uint64, error) {
 	return txn.StartTS(), nil
 }
 
+func (p *baseTxnContextProvider) usePresetStartTS() bool {
+	return p.constStartTS != 0 || p.sctx.GetSessionVars().SnapshotTS != 0
```
ekexium, line 274: I'm not sure about this: `SnapshotTS != 0` should imply a staleness txn ctx provider? It won't hurt though

pkg/session/session.go
ekexium: Also for the other log

pkg/session/txn.go
b6g: Actually the reason is `CommitTS` has to be saved until the next transaction. If we change `doCommit` to return `CommitTS`, we have to,
- save `CommitTS` in `session`
- check if the transaction is committed successfully in `doCommit`
Since we have to save to save `CommitTS`, it would be simpler to save `CommitTS` inside `doCommit`.

pkg/sessiontxn/isolation/base.go
b6g: I am not sure either. That's why I named it `Preset`. The point is we don't check StartTS if it is preset.

pkg/session/txn.go
you06: Can we store commit ts in `doCommit`?

```diff
+func (s *session) doCommit(ctx context.Context) (err error) {
	if !s.txn.Valid() {
		return nil
	}

	defer func() {
+		if err != nil {
+			s.sessionVars.LastCommitTS = s.txn.CommitTS()
+		}
		s.txn.changeToInvalid()
		s.sessionVars.SetInTxn(false)
		s.sessionVars.ClearDiskFullOpt()
	}()
	...
}
```

b6g:
ping

ekexium (approved):

ekexium:
The PR is good to merge as is. We could improve the `lastCommitTs` design to avoid the confusion.

b6g:
I think I need another approval

cfzjywxk:
@Benjamin2037 @yudongusa
The session variable file is modified, PTAL for the approvement.

pkg/sessiontxn/isolation/base.go
cfzjywxk: @ekexium @b6g 

Here, considering various historical features (e.g., `tidb_snapshot`, stale read, etc.), careful attention is required to ensure accurate and comprehensive condition checks. Any oversight could lead to unexpected tidb-server panics.  

An alternative approach is to reverse the logic—confirming that the current transaction is activated by a **PD-allocated timestamp** before proceeding with the check. This seems to be a more reliable method.

like
```
- if !usePresetStartTS
+ if activiatedByPDAllocatedTS
```

b6g: How do we know a txn is activated by a PD allocated timstamp?

cfzjywxk: It seems there is no direct interface at the moment. One approach is to add a state flag to `baseTxnContextProvider` or `txnFuture`, though this comes with some additional overhead.  

Or, we just use the current implementation and address any missed cases with subsequent fixes if they arise.

@ekexium What do you think?

pkg/session/session.go
```diff
@@ -925,6 +925,19 @@ func (s *session) CommitTxn(ctx context.Context) error {
 		s.sessionVars.StmtCtx.MergeExecDetails(nil, commitDetail)
 	}
 
+	if err == nil && s.txn.lastCommitTS > 0 {
+		// lastCommitTS could be the same, e.g. when the txn is considered readonly
+		if s.txn.lastCommitTS < s.sessionVars.LastCommitTS {
+			logutil.BgLogger().Panic("check lastCommitTS failed",
```
crazycs520, line 931: I don't think panic is a good idea here, as it will leave some cleanup work undone, but at the outermost level it will be recovered, leading to possible problems with subsequent requests.

I prefer printing an ERROR level log and then assigning a value to err.   @cfzjywxk @ekexium

cfzjywxk: Make sense, @b6g PTAL

pkg/sessiontxn/isolation/base.go
```diff
@@ -304,6 +311,15 @@ func (p *baseTxnContextProvider) ActivateTxn() (kv.Transaction, error) {
 		sessVars.SetInTxn(true)
 	}
 
+	// verify start_ts is later than any previous commit_ts in the session
+	if !p.usePresetStartTS() && sessVars.LastCommitTS > 0 && sessVars.LastCommitTS > sessVars.TxnCtx.StartTS {
+		logutil.BgLogger().Panic("check session lastCommitTS failed",
```
crazycs520, line 316: ditto.

pkg/session/session.go
b6g: I think this is a violation of an important invariant, and we'd better abort the transaction to avoid unexpected behavior, that is hard to debug.

I can add a global system variable so that we can disable this panic if needed. WDYT?

cfzjywxk: The TiDB upper layer will catch panics. As a result, a panic may result in some cleanup work not being executed. Throwing an error appears to be a more robust approach, and the transaction encountering the error will not proceed with the commit and would abort.

b6g: ok. let me make the change.

b6g: Just to call out. The transaction is already committed with the bad commit_ts even if we return an error here. And more transactions like this would be committed.

pkg/session/txn.go
b6g: I think `s.txn.CommitTS` is already reset in `LazyTxn.Commit` before it reaches the defer of doCommit.

b6g:
replace panic with `ErrAssertionFailed`. PTAL

b6g:
PTAL cc @yudongusa

cfzjywxk (approved):

pkg/sessiontxn/isolation/base.go
you06: `usePresetStartTS` looks fine for me. But better leave a comment that we may have a flag which indicate the `start_ts` is from PD.

I took a look at the code and feel like we can add StartTSFromPD [here](https://github.com/pingcap/tidb/blob/43f2fb91d740aaf45774d99b71554eead12009cc/pkg/sessionctx/variable/session.go#L207), and set it to true if the `start_ts` is from a ts future [here](https://github.com/pingcap/tidb/blob/43f2fb91d740aaf45774d99b71554eead12009cc/pkg/session/txn.go#L295).

pkg/session/session.go
```diff
@@ -927,6 +927,19 @@ func (s *session) CommitTxn(ctx context.Context) error {
...
 
+	if err == nil && s.txn.lastCommitTS > 0 {
+		// lastCommitTS could be the same, e.g. when the txn is considered readonly
+		if s.txn.lastCommitTS < s.sessionVars.LastCommitTS {
+			logutil.BgLogger().Error("check lastCommitTS failed",
+				zap.Uint64("sessionLastCommitTS", s.sessionVars.LastCommitTS),
+				zap.Uint64("txnLastCommitTS", s.txn.lastCommitTS),
+				zap.String("sql", redact.String(s.sessionVars.EnableRedactLog, s.sessionVars.StmtCtx.OriginalSQL)),
+			)
+			return kv.ErrAssertionFailed
```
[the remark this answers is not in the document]
cfzjywxk: @b6g PTAL

b6g:
Updated. PTAL. Need another approval to merge. Thanks!

you06 (approved):

yudongusa (approved):

PR #57320, opened 2024-11-12
*: move the 'user' field to the first one of primary key for mysql tables | tidb-test=pr/2425

tiancaiamao:

Issue Number: ref #55563 

Problem Summary:

To make the modify user or privilege efficient, I have a PR to just reload the changes of one specific user.

The reload SQL query is something like `select * from mysql.user where user = 'xxx'` (There are many other MySQL tables involved too)

But this SQL is slow, because `where user = 'xxx'` has no index.

To make those queries efficient, I want to adjust the primary key layout of some mysql tables.
Move the 'user' field to the first one.

Only do the change for the new created cluster.
Do not adjust the table schema for old cluster upgrading. 
(It means the code still work, but less efficient. This should not be a problem, because in the past we reload all the mysql tables data, and the current implementation is not slower than before.)

~~**Please merge https://github.com/pingcap/tidb/pull/57042 first before review this one, so the changes are minimal.**~~

### Check List

Tests 

- [ ] Unit test
- [X] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


D3Hunter (commented):
it doesn't make sense to break the table structure

just add another index for `user`

D3Hunter:
> The reload SQL query is something like select * from mysql.user where user = 'xxx' (There are many other MySQL tables involved too)

mysql takes `host` as part of the credential, why you cannot know the `host` when reload?

tiancaiamao:
> it doesn't make sense to break the table structure
> 
> just add another index for `user`

If you comapre the mysql tables with MySQL one by one, you'll found our table struct does not match MySQL exactly ...
So keeping the table structure is meanlingless

Add another `user` index is other option, but maybe adjust the primary key is better.
The performance is better using primary key range scan then the double read for index and data then if we have the `user` index. @D3Hunter

tiancaiamao:
> > The reload SQL query is something like select * from mysql.user where user = 'xxx' (There are many other MySQL tables involved too)
> 
> mysql takes `host` as part of the credential, why you cannot know the `host` when reload?

OKay, that's talk more about the background.

In the past, we load all the data in-memory for `myser.user` table.
MySQL use its special matching rule for the user login process.

User + Host determine one specific login user, but there maybe multiple matches.
For example, the mysql/user table contains record:

```
User   Host                               PWD
joe  @ %                                  aaa
joe  @  192.168.%                   bbb
joe  @ 192.168.3.%                 ccc
```

When joe login at host 193.168.2,  use password 'aaa', should the Auth process success?
The match should be this record:

```
joe  @  192.168.%                   bbb
```

rather than

```
joe  @ %                                  aaa
```

How the matching rules works?    we load all the records of  user `joe` , then sort them using some special rule.
The sort rule is so special that it's hard to using simply SQL:

`select from mysql.user  where  user = joe order by XXX`


We don't know the `host` in such scenario, so we need to load all the records of a `user` into memory.

tiancaiamao:
closed it on behalf of https://github.com/pingcap/tidb/pull/57470

Issue #57229, opened 2024-11-08
Support dynamic change to DDL job configuration


tangenta:
## Feature Request

**Is your feature request related to a problem? Please describe:**

TiDB v8.3.0 [removes the dynamic change feature](https://github.com/pingcap/tidb/pull/55334) for `tidb_ddl_reorg_worker_cnt` and `tidb_ddl_reorg_batch_size`, but this feature is useful in some scenarios:
The foreseeable business peak arrives during the DDL operation, and it is necessary to reduce DDL's impact on the cluster.
- DDL execution reaches a certain stage, and resource usage suddenly increases. The resources available to DDL must be reduced (for example, ADD INDEX entering the ingest stage usually incurs a large CPU memory overhead).
- During the DDL operation, the user observed that the data backfilling speed was slow, and it is necessary to increase the backfilling concurrency to complete the DDL within the expected time.

Although the above scenario can be bypassed by canceling and re-executing DDL, it reduces usability and affects user experience.

**Describe the feature you'd like:**

Add a new syntax that supports dynamically changing relevant config for DDL jobs.

```
ADMIN ALTER DDL JOBS *job_id* THREAD = 8;   -- tidb_ddl_reorg_worker_cnt
ADMIN ALTER DDL JOBS *job_id* BATCH_SIZE = 256;  -- tidb_ddl_reorg_batch_size
ADMIN ALTER DDL JOBS *job_id* MAX_WRITE_SPEED = '200Mi'; -- tidb_ddl_reorg_max_write_speed
```

This syntax refers to the ADMIN clause currently supported by TiDB.

```
ADMIN SHOW DDL JOBS ...
ADMIN PAUSE DDL JOBS ...
ADMIN RESUME DDL JOBS ...
ADMIN CANCEL DDL JOBS ...
```

`THREAD` and `BATCH_SIZE` only work for DDLs that need to reorganize data (reorg) types, currently including `ADD INDEX`, `MODIFY COLUMN`, and `REORGANIZE PARTITION`. Changing job config is not supported for non-reorg DDLs.

On the other hand, add a column `COMMENTS` to the result set of `ADMIN SHOW DDL JOBS`, which displays the current DDL job config:

```
ADMIN SHOW DDL JOBS *job_id*;

... |                 COMMENTS |
... | THREAD=4, BATCH_SIZE=256 |
```

**Describe alternatives you've considered:**

- Revert https://github.com/pingcap/tidb/pull/55334

[...]

PR #57336, opened 2024-11-13
ddl: support modify the related reorg config by SQL

fzzf678:

Issue Number: ref #57229

Problem Summary:

For the purpose of dynamically adjusting reorg configs, we need to persist the config value to disk, so in this pr, we support modify reorg config in system table by execute sql `ADMIN ALTER DDL JOBS ...`. Only `ADD INDEX`, `MODIFY COLUMN`, and `REORGANIZE PARTITION` three types ddl is allowed to change config. And for config we support `THREAD`, `BATCH_SIZE` and `MAX_WRITE_SPEED` now.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/operate_ddl_jobs.go
```diff
@@ -85,3 +93,126 @@ type PauseDDLJobsExec struct {
...
+	jobID int64,
+) (*model.Job, error) {
+	sql := fmt.Sprintf("select job_meta from mysql.%s where job_id = %s",
+		ddl.JobTable, strconv.FormatInt(jobID, 10))
+	rows, err := se.Execute(ctx, sql, "get_job_by_id")
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	if len(rows) == 0 {
+		return nil, fmt.Errorf("ddl job %d not found", jobID)
```
tangenta, line 127:
```suggestion
		return nil, fmt.Errorf("ddl job %d is not running", jobID)
```

fzzf678: Yes, I see it. But I have an unrelated naive question: If we get a job_meta from mysql.ddl_job_table, Is the job state always running or may be some other state like pending、canceling ...

fzzf678: Actually I want to ask if we should check the job state

tangenta: We don't need to check the state, because there is no harm in changing job_meta even if they are pending or canceling.

tangenta (approved):

CbcWestwolf (approved):

winoros (approved):

PR #57388, opened 2024-11-14
*: avoid some unnecessary call of ensureActiveUser()

tiancaiamao:

Issue Number: ref #55563

Problem Summary:

In https://github.com/pingcap/tidb/pull/57042, I add too much calls of `ensureActiveUser()` than necessary. That's introduced to make the CI pass.
That PR had already done a lot to make the CI pass .... So I don't like to introduce extra complexity to it.

After merge that, this commit will be used to reduce some unnecessary `ensureActiveUser()`.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

For example

Without optimizing the unnecessary `ensureActiveUser()`, If I use `set password for user%d` to change password for 2M users, here is the QPS

image

After the optimization:

image

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


tests/integrationtest/t/executor/simple.test
```diff
@@ -165,6 +165,7 @@ disconnect conn1;
 # TestFlushPrivileges
 drop user if exists 'testflush'@'localhost';
 CREATE USER 'testflush'@'localhost' IDENTIFIED BY '';
+SHOW GRANTS FOR 'testflush'@'localhost';
 UPDATE mysql.User SET Select_priv='Y' WHERE User="testflush" and Host="localhost";
 connect (conn1, localhost, testflush,,);
 --error 1142
```
tiancaiamao, line 171: This is a behavior change.
In the past, all the user data are in-memory, and login will not trigger reload.
Now login will trigger the reload if the user is not loaded,  login make the user get the latest privileges.

To test the old behavior, `SHOW GRANTS FOR 'testflush'@'localhost';` this line is added.
This cause the user data been loaded
Then login find the user and does not trigger reload.

pkg/executor/simple.go
```diff
@@ -2522,14 +2518,13 @@ func (e *SimpleExec) executeSetPwd(ctx context.Context, s *ast.SetPwdStmt) error
 		h = s.User.Hostname
 
 		checker := privilege.GetPrivilegeManager(e.Ctx())
-		checker.MatchIdentity(ctx, u, h, false)
```
lcwangchao, line 2525: Just curious about what this line is used for  before this PR

pkg/privilege/privileges/privileges.go
```diff
@@ -930,9 +914,6 @@ func (p *UserPrivileges) ShowGrants(ctx context.Context, sctx sessionctx.Context
 		u = user.AuthUsername
 		h = user.AuthHostname
 	}
-	if err := p.Handle.ensureActiveUser(ctx, u); err != nil {
```
lcwangchao, line 933: Will `ShowGrants` returns `ErrNonexistingGrant` if an user is not loaded to memory after deleting this line?

pkg/executor/simple.go
tiancaiamao: In the past, `ensureActiveUser()` is called for both 'show grants' and 'show grants for user xxx'
This change make `ensureActiveUser` only called for the later case.

pkg/privilege/privileges/privileges.go
tiancaiamao: show grants for current user should never fail, because the current user must login to execute commands.
show grants for someone else, the caller ensures `ensureActiveUser()` is called before calling this `ShowGrants` function.

lcwangchao (approved):

pkg/executor/show.go
```diff
@@ -1845,7 +1847,6 @@ func (e *ShowExec) fetchShowCreateUser(ctx context.Context) error {
 		require = privValue.RequireStr()
 	}
 
-	authData := checker.GetEncodedPassword(ctx, e.User.Username, e.User.Hostname)
```
bb7133, line 1848: I don't quite get why we don't need `ensureActiveUser()` in `fetchShowCreateUser` and other functions. Could you please explain it a little bit, and how can we be sure the correctness can be verified?

pkg/executor/simple.go
```diff
@@ -1795,10 +1795,6 @@ func (e *SimpleExec) executeAlterUser(ctx context.Context, s *ast.AlterUserStmt)
 			failedUsers = append(failedUsers, user)
 			continue
 		}
-		currentAuthPlugin, err := privilege.GetPrivilegeManager(e.Ctx()).GetAuthPlugin(ctx, spec.User.Username, spec.User.Hostname)
```
bb7133, line 1798: ditto

pkg/executor/show.go
tiancaiamao: `show create user` is not `show grants for user`
We just need the command for creating that user, we do not need to assign any privileges to the user. @bb7133

pkg/executor/simple.go
tiancaiamao: The authPlugin and authentication data is obtained by the ExecInternalSQL , read from the mysql.user.

bb7133 (approved):
LGTM

PR #56660, opened 2024-10-15
*: make `default_authentication_plugin` more compatible with mysql

CbcWestwolf:
### What problem does this PR solve?

Issue Number: close #54138

Problem Summary:

See the issue decription

### What changed and how does it work?

Get default auth plugin when creating, altering and granting user.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
make `default_authentication_plugin` control default auth plugin when creating user
```

Issue #54138, opened 2024-06-20
`default_authentication_plugin` is not compatable with MySQL's


CbcWestwolf:
Nowadays, [`default_authentication_plugin` in tidb](https://docs.pingcap.com/tidb/stable/system-variables#default_authentication_plugin) only affects the connection between servers and clients. It cannot control the default auth plugin [when creating users without explicit auth plugin as in MySQL](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_default_authentication_plugin). Since `caching_sha2_password` is safer than the default `mysql_native_password`, it is better for tidb to support this compatibility with MySQL.

> It determines which authentication plugin the server assigns to new accounts created by [CREATE USER](https://dev.mysql.com/doc/refman/8.0/en/create-user.html) statements that do not explicitly specify an authentication plugin.

However, since currently using the `caching_sha2_password` in TiDB [requires a SSL connection](https://github.com/pingcap/tidb/blob/4cf86e499364c358fc9420463cccd4ff66240ef6/pkg/server/conn.go#L686), it is not suitable for TiDB to make it default authentication plugin after bootstrapping a new TiDB cluster. It is better to make it default after implementing complete `caching_sha2_password` on TiDB.

ei-sugimoto:
Hi, @CbcWestwolf
I have questions. 
See the code below.
```go
// authSha implements the caching_sha2_password specific part of the protocol.
func (cc *clientConn) authSha(ctx context.Context, resp handshake.Response41) ([]byte, error) {
	const (
		shaCommand       = 1
		requestRsaPubKey = 2 // Not supported yet, only TLS is supported as secure channel.
		fastAuthOk       = 3
		fastAuthFail     = 4
	)

	// If no password is specified, we don't send the FastAuthFail to do the full authentication
	// as that doesn't make sense without a password and confuses the client.
	// https://github.com/pingcap/tidb/issues/40831
	if len(resp.Auth) == 0 {
		return []byte{}, nil
	}

	// Currently we always send a "FastAuthFail" as the cached part of the protocol isn't implemented yet.
	// This triggers the client to send the full response.
	err := cc.writePacket([]byte{0, 0, 0, 0, shaCommand, fastAuthFail})
.....
``` 

Thus, if `caching_sha2_password` is selected, special processing is performed.
Also, this process causes `resp.Auth` to return nil and the test fails.
What should I do?

dveeden:
[...]

PR #57391, opened 2024-11-14
*: fix a bug for default_authentication_plugin

CbcWestwolf:

Issue Number: ref #54138

Problem Summary:

In [`func (n *UserSpec) EncodedPassword()`](https://github.com/pingcap/tidb/blob/e79d22f156ca8305de126297d1a85ff900cba586/pkg/parser/ast/misc.go#L1528), if the [`opt.AuthPlugin`](https://github.com/pingcap/tidb/blob/e79d22f156ca8305de126297d1a85ff900cba586/pkg/parser/ast/misc.go#L1535) is empty, it will always encode the password by the `mysql_native_password` way. However, after introducing `default_authentication_plugin` in https://github.com/pingcap/tidb/pull/56660, here should instead consider the 'default' way.

1. In `encodedPassword`, if the `opt.AuthPlugin` is empty, use the default auth plugin to encode the password. 
2. Remove some unreachable code.
3. Add more tests

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/parser/ast/misc.go
+++ b/pkg/parser/ast/misc.go
@@ -1493,9 +1493,6 @@ type UserSpec struct {
 	User    *auth.UserIdentity
 	AuthOpt *AuthOption
 	IsRole  bool
-
-	// DefaultAuthPlugin is used to store the values of default_authentication_plugin
-	DefaultAuthPlugin string
 }
 
 // Restore implements Node interface.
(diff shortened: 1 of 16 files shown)


pkg/parser/ast/misc.go
```diff
@@ -1493,6 +1493,9 @@ type UserSpec struct {
 	User    *auth.UserIdentity
 	AuthOpt *AuthOption
 	IsRole  bool
+
+	// DefaultAuthPlugin is used to store the values of default_authentication_plugin
+	DefaultAuthPlugin string
```
lcwangchao, line 1498: How about to remove `DefaultAuthPlugin` in `UserSpec` and move `EncodedPassword` to executor? Seems `EncodedPassword` is only used by executor and it's better to keep `UserSpec` to serve for parser only.

pkg/privilege/privileges/cache.go
```diff
@@ -925,11 +925,7 @@ func (p *immutable) decodeUserTableRow(row chunk.Row, fs []*resolve.ResultField)
 				value.AccountLocked = true
 			}
 		case f.ColumnAsName.L == "plugin":
-			if row.GetString(i) != "" {
-				value.AuthPlugin = row.GetString(i)
-			} else {
-				value.AuthPlugin = mysql.AuthNativePassword
-			}
+			value.AuthPlugin = row.GetString(i)
```
lcwangchao, line 928: When `ow.GetString(i) != ""` happen, can we avoid it after this PR?

pkg/parser/ast/misc.go
CbcWestwolf: Update in https://github.com/pingcap/tidb/pull/57391/commits/50a049aea2aa808370a184d7afb0d11cb88dbffa

pkg/privilege/privileges/cache.go
CbcWestwolf: To avoid unexpected cases, I add the barrier back https://github.com/pingcap/tidb/pull/57391/commits/7747b9d868b6d1c48d2634f66c54bdaaaa884174

lcwangchao (approved):

tangenta (approved):

pkg/privilege/privileges/cache.go
```diff
@@ -916,6 +918,16 @@ func (record *baseRecord) assignUserOrHost(row chunk.Row, i int, f *resolve.Resu
 
 func (p *immutable) decodeUserTableRow(row chunk.Row, fs []*resolve.ResultField) error {
 	var value UserRecord
+	defaultAuthPlugin := ""
```
tiancaiamao, line 921: What happen if a user create some users with default auth plugin = x
and then the set the default auth variable = y?
@CbcWestwolf

CbcWestwolf: The 'plugin' field in `mysql.user` table would be filled with 'x'. So in most cases, `defaultAuthPlugin` would not attend to decode a row, unless the `plugin` field is manully cleaned by tidb user, or some bugs.

PR #57418, opened 2024-11-15
table: fix wrong handle comparation in index double write

### LINKED ISSUE #57414: `INSERT ... ON DUPLICATE` during index merge phase will cause data inconsistency
## Bug Report

Run the following test:

``` patch
From 1625d2e39b71cb06d4e4724ec28c5a65f662b873 Mon Sep 17 00:00:00 2001
From: joechenrh <joechenrh@gmail.com>
Date: Fri, 15 Nov 2024 08:15:19 +0000
Subject: [PATCH] Add test

---
 pkg/ddl/index.go             |  1 +
 pkg/ddl/index_modify_test.go | 22 ++++++++++++++++++++++
 2 files changed, 23 insertions(+)

diff --git a/pkg/ddl/index.go b/pkg/ddl/index.go
index 0a73371562..6f853f6545 100644
--- a/pkg/ddl/index.go
+++ b/pkg/ddl/index.go
@@ -1352,6 +1352,7 @@ func doReorgWorkForCreateIndex(
 				MockDMLExecutionStateBeforeMerge()
 			}
 		})
+		failpoint.InjectCall("BeforeBackfillMerge", job)
 		logutil.DDLLogger().Info("index backfill state ready to merge",
 			zap.Int64("job ID", job.ID),
 			zap.String("table", tbl.Meta().Name.O),
diff --git a/pkg/ddl/index_modify_test.go b/pkg/ddl/index_modify_test.go
index b38d05fbff..7999368ea1 100644
--- a/pkg/ddl/index_modify_test.go
+++ b/pkg/ddl/index_modify_test.go
@@ -1422,3 +1422,25 @@ func TestAddVectorIndexRollback(t *testing.T) {
 
 	testfailpoint.Disable(t, "github.com/pingcap/tidb/pkg/ddl/MockCheckVectorIndexProcess")
 }
+
+func TestAddIndex(t *testing.T) {
+	tkVar := testkit.NewTestKit(t, store)
+	tkVar.MustExec("set @@global.tidb_ddl_enable_fast_reorg = 1")
+	tkVar.MustExec("set @@global.tidb_enable_dist_task=0")
+
+	tk.MustExec("use test")
+	tk.MustExec("create table t (col1 int, col2 int, unique index i1(col2) /*T![global_index] GLOBAL */) PARTITION BY HASH (col1) PARTITIONS 2")
+
+	tk2.MustExec("use test")
+
+	testfailpoint.EnableCall(t, "github.com/pingcap/tidb/pkg/ddl/BeforeBackfillMerge", func(job *model.Job) {
+		tk2.MustExec("insert ignore into t values (1, 2), (1, 2) on duplicate key update col1 = 0, col2 = 0")
+	})
+	defer testfailpoint.Disable(t, "github.com/pingcap/tidb/pkg/ddl/BeforeBackfillMerge")
+
+	tk.MustExec("alter table t add unique index i(col1, col2)")
+	tk.MustExec("admin check table t")
+}
-- 
2.43.0
```

Test passed.

`admin check table` failed.

joechenrh:

Issue Number: close #57414

Problem Summary:

Unlike other indices, global index encodes the value of its key-value pairs with a partition ID:

https://github.com/pingcap/tidb/blob/2ca497e23e1488a1675dfe5793296b8b367c261b/pkg/tablecodec/tablecodec.go#L1630-L1649

So during the double write process, we may do comparison between partition handles and non-partition handles.

https://github.com/pingcap/tidb/blob/2ca497e23e1488a1675dfe5793296b8b367c261b/pkg/table/tables/index.go#L419-L432

For example, in #57414, `h` is the handle of global index `i1` whose type is `PartitionHandle` while `oh` is `IntHandle` of the current index. Thus, `!h.Equal(oh)` will return false.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/index.go
+++ b/pkg/ddl/index.go
@@ -2025,14 +2025,6 @@ func (w *addIndexTxnWorker) checkHandleExists(idxInfo *model.IndexInfo, key kv.K
 	if hasBeenBackFilled {
 		return nil
 	}
-	if idxInfo.Global {
-		// 'handle' comes from reading directly from a partition, without partition id,
-		// so we can only compare the handle part of the key.
-		if ph, ok := h.(kv.PartitionHandle); ok && ph.Handle.Equal(handle) {
-			// table row has been back-filled already, OK to add the index entry
-			return nil
-		}
-	}
 	return ddlutil.GenKeyExistsErr(key, value, idxInfo, tblInfo)
 }
 
--- a/pkg/ddl/index.go
+++ b/pkg/ddl/index.go
@@ -1352,7 +1352,7 @@ func doReorgWorkForCreateIndex(
 				MockDMLExecutionStateBeforeMerge()
 			}
 		})
-		failpoint.InjectCall("BeforeBackfillMerge", job)
+		failpoint.InjectCall("BeforeBackfillMerge")
 		logutil.DDLLogger().Info("index backfill state ready to merge",
 			zap.Int64("job ID", job.ID),
 			zap.String("table", tbl.Meta().Name.O),
--- a/pkg/ddl/index.go
+++ b/pkg/ddl/index.go
@@ -1352,6 +1352,7 @@ func doReorgWorkForCreateIndex(
 				MockDMLExecutionStateBeforeMerge()
 			}
 		})
+		failpoint.InjectCall("BeforeBackfillMerge", job)
 		logutil.DDLLogger().Info("index backfill state ready to merge",
 			zap.Int64("job ID", job.ID),
 			zap.String("table", tbl.Meta().Name.O),
--- a/pkg/table/tables/index.go
+++ b/pkg/table/tables/index.go
@@ -386,13 +386,6 @@ func needPresumeKeyNotExistsFlag(ctx context.Context, txn kv.Transaction, key, t
 	return false, nil
 }
 
-func handleEqual(dupHandle, prevHandle kv.Handle) bool {
-	if dupPartition, ok := dupHandle.(kv.PartitionHandle); ok {
-		return dupPartition.Handle.Equal(prevHandle)
-	}
-	return dupHandle.Equal(prevHandle)
-}
-
 // Delete removes the entry for handle h and indexedValues from KV index.
 func (c *index) Delete(ctx table.MutateContext, txn kv.Transaction, indexedValue []types.Datum, h kv.Handle) error {
 	indexedValues := c.getIndexedValue(indexedValue)
@@ -436,7 +429,7 @@ func (c *index) Delete(ctx table.MutateContext, txn kv.Transaction, indexedValue
 						}
 						// The handle passed in may be a `PartitionHandle`,
 						// so we can't directly do comparation with them.
-						if !handleEqual(h, oh) {
+						if !h.Equal(oh) {
 							okToDelete = false
 						}
 					}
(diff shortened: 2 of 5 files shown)


Defined2014 (approved):

pkg/table/tables/index.go
```diff
@@ -427,7 +434,9 @@ func (c *index) Delete(ctx table.MutateContext, txn kv.Transaction, indexedValue
 						if err != nil {
 							return err
 						}
-						if !h.Equal(oh) {
+						// The handle passed in may be a `PartitionHandle`,
+						// so we can't directly do comparation with them.
+						if !handleEqual(h, oh) {
```
Defined2014, line 439: Although when `c.Global` = true h should be a partition handle, it is better to check `c.Global` here, we should compare pid when it's true.

wjhuang2016, line 439: If the passed handle is PartaitionHandle, why oh isn't partitionHandle here?

pkg/table/tables/index.go
```diff
@@ -386,6 +386,13 @@ func needPresumeKeyNotExistsFlag(ctx context.Context, txn kv.Transaction, key, t
 	return false, nil
 }
 
+func handleEqual(dupHandle, prevHandle kv.Handle) bool {
```
Defined2014, line 389: Maybe we could change `func (ph PartitionHandle) Equal(h Handle) bool` function. Only both handles are partition handle, we need compare pid.

pkg/table/tables/index.go
joechenrh: Take the case in #57414 as example, the handle `h` is retrieved from `FetchDuplicatedHandle`. Since global index is first checked and duplication is found, we will get `PartitionHandle`.
https://github.com/pingcap/tidb/blob/043f833c4985ee489948bfb75cd1b3256cc7bbd6/pkg/executor/insert.go#L272-L273
But the newly added index is not global index, so `oh` is just `IntHandle`.

In addition, there may also be opposite situations that `oh` is `PartitionHandle` and `h` is `IntHandle`. But `IntHandle` can be compared with `PartitionHandle` correctly.

pkg/table/tables/index.go
joechenrh: > Maybe we could change func (ph PartitionHandle) Equal(h Handle) bool function.

I'm not sure if it will affect other code so I add a new function. Let me have a try.

pkg/ddl/index.go
```diff
@@ -2024,14 +2025,6 @@ func (w *addIndexTxnWorker) checkHandleExists(idxInfo *model.IndexInfo, key kv.K
...
 		return nil
 	}
-	if idxInfo.Global {
-		// 'handle' comes from reading directly from a partition, without partition id,
-		// so we can only compare the handle part of the key.
-		if ph, ok := h.(kv.PartitionHandle); ok && ph.Handle.Equal(handle) {
-			// table row has been back-filled already, OK to add the index entry
-			return nil
-		}
-	}
```
joechenrh, line 2034: Just remove some unnecessary logic for comparation.

wjhuang2016 (approved):

PR #57428, opened 2024-11-15
planner: Fix the issue where any(nil) is not considered equal to nil

### LINKED ISSUE #57326: any(nil) is not equal to nil, should cast it as its original type
## Bug Report

```
func (col *Column) Equals(other any) bool {
	if other == nil {. // false
		return false
	}
        col2, ok := other.(*Column)
        // ok will be true
        // then col2 == nil is true 
```
when we pass `var s *Column = nil` into as other with type any here, we found it doesn't equal to nil anymore, unless we downcast it with `*Column` type before equaling with nil

as mentioned 
suggestion way
```
func (col *Column) Equals(other any) bool {
    col2, ok := other.(*Column)
    if !ok {
        return false
    }
    if col == nil {
        return col2==nil
    }
    if col2 == nil {
        return false
    }
    // normal equals logic appended
```

as mentioned 

master

King-Dylan:

Issue Number: close #57326 

Problem Summary:
Fix the issue where any(nil) is not considered equal to nil.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/expression/column.go
```diff
@@ -511,9 +511,13 @@ func (col *Column) Hash64(h base.Hasher) {
 
 // Equals implements HashEquals.<1st> interface.
 func (col *Column) Equals(other any) bool {
-	if other == nil {
-		return false
+	// If col itself is nil, check whether other is also nil
+	if col == nil {
```
AilinKid, line 515: better use this format
```
	op2, ok := other.(*DataSource)
	if !ok {
		return false
	}
	if op == nil {
		return op2 == nil
	}
	if op2 == nil {
		return false
	}
```

and there many other places implements this Equals interface with the same bad way, we should change them too

AilinKid: image

you can do it by batch with 2-3, or finish them all, it's up to you, do remember the test case

pkg/expression/column.go
```diff
@@ -511,9 +511,13 @@ func (col *Column) Hash64(h base.Hasher) {
...
+		// Checking if other is of type *Column and nil
+		col2, ok := other.(*Column)
+		return ok && col2 == nil
 	}
+	// Attempt to convert other to *Column
 	var col2 *Column
 	switch x := other.(type) {
```
AilinKid, line 522: this kind of way can be eliminated, we can make it only receive *pointer as its argument here. Let the caller to pass &XX if it's a structure.

AilinKid (approved):
Well done

windtalker (approved):
lgtm

PR #57431, opened 2024-11-16
planner: remove modifyCount from cardinality estimation v1

terry1purcell:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/cardinality/row_count_index.go
+++ b/pkg/planner/cardinality/row_count_index.go
@@ -264,7 +264,7 @@ func getIndexRowCountForStatsV2(sctx planctx.PlanContext, idx *statistics.Index,
 					}
 					continue
 				}
-				count = equalRowCountOnIndex(sctx, idx, lb, realtimeRowCount, modifyCount)
+				count = equalRowCountOnIndex(sctx, idx, lb, realtimeRowCount)
 				// If the current table row count has changed, we should scale the row count accordingly.
 				count *= idx.GetIncreaseFactor(realtimeRowCount)
 				if debugTrace {
@@ -374,7 +374,7 @@ func getIndexRowCountForStatsV2(sctx planctx.PlanContext, idx *statistics.Index,
 
 var nullKeyBytes, _ = codec.EncodeKey(time.UTC, nil, types.NewDatum(nil))
 
-func equalRowCountOnIndex(sctx planctx.PlanContext, idx *statistics.Index, b []byte, realtimeRowCount, modifyCount int64) (result float64) {
+func equalRowCountOnIndex(sctx planctx.PlanContext, idx *statistics.Index, b []byte, realtimeRowCount int64) (result float64) {
 	if sctx.GetSessionVars().StmtCtx.EnableOptimizerDebugTrace {
 		debugtrace.EnterContextCommon(sctx)
 		debugtrace.RecordAnyValuesWithNames(sctx, "Encoded Value", b)
@@ -419,8 +419,11 @@ func equalRowCountOnIndex(sctx planctx.PlanContext, idx *statistics.Index, b []b
 		// idx.TotalRowCount rather than idx.Histogram.NotNullCount() since the histograms are empty.
 		//
 		// If the table hasn't been modified, it's safe to return 0.
-		if modifyCount == 0 {
+		insertRows := float64(realtimeRowCount) - idx.TotalRowCount()
+		if insertRows == 0 {
 			return 0
+		} else if insertRows < 0 {
+			insertRows = float64(realtimeRowCount)
 		}
 		// ELSE calculate an approximate estimate based upon newly inserted rows.
 		//
@@ -432,7 +435,7 @@ func equalRowCountOnIndex(sctx planctx.PlanContext, idx *statistics.Index, b []b
 		}
 		// As a conservative estimate - take the smaller of the orignal totalRows or the additions.
 		// "realtimeRowCount - original count" is a better measure of inserts than modifyCount
-		totalRowCount := min(idx.TotalRowCount(), float64(realtimeRowCount)-idx.TotalRowCount())
+		totalRowCount := min(idx.TotalRowCount(), insertRows)
 		return max(1, totalRowCount/histNDV)
 	}
 	// return the average histogram rows (which excludes topN) and NDV that excluded topN


pkg/planner/cardinality/row_count_column.go
```diff
@@ -182,6 +182,10 @@ func equalRowCountOnColumn(sctx planctx.PlanContext, c *statistics.Column, val t
 		if modifyCount == 0 {
 			return 0, nil
 		}
+		newRows := float64(realtimeRowCount) - c.TotalRowCount()
+		if newRows < 0 {
```
Rustin170506, line 186: May I ask what the purpose of this change is?

Rustin170506: Also, why do we need to use the total row count instead of the null count here?

pkg/statistics/histogram.go
```diff
@@ -921,15 +921,15 @@ func (hg *Histogram) OutOfRange(val types.Datum) bool {
...
-	modifyCount, histNDV int64, increaseFactor float64,
+	realtimeRowCount, modifyCount, histNDV int64,
 ) (result float64) {
 	debugTrace := sctx.GetSessionVars().StmtCtx.EnableOptimizerDebugTrace
 	if debugTrace {
 		debugtrace.EnterContextCommon(sctx)
 		debugtrace.RecordAnyValuesWithNames(sctx,
 			"lDatum", lDatum.String(),
 			"rDatum", rDatum.String(),
-			"modifyCount", modifyCount,
```
Rustin170506, line 932: I guess we can still keep this?

Rustin170506 (commented):
Thanks!

pkg/planner/cardinality/row_count_column.go
terry1purcell: The change is simply to avoid a negative number. The reason is that if there have been modifications - then the original stats are now unreliable. If realtimeRowCount has increased - then it is possible that the value we are searching for is in those added rows. If decreased - it could be that they are ONLY deletes - which means that the value we are searching for isn't in the data - but it also could be a combination of inserts/updates/deletes - we simply do not know.

I use TotalRowCount rather than notNullCount because - if there are NULLs - then notNullCount may always be lower than realtimeRowCount. Because realtimeRowCount counts all rows (I believe). So using notNullCount would incorrectly have us counting those "NULL rows" here.

pkg/statistics/histogram.go
terry1purcell: You are correct - will add it back

pkg/statistics/histogram.go
```diff
@@ -1049,22 +1073,26 @@ func (hg *Histogram) OutOfRangeRowCount(
...
-
 	if !allowUseModifyCount {
 		// In OptObjectiveDeterminate mode, we can't rely on the modify count anymore.
 		// An upper bound is necessary to make the estimation make sense for predicates with bound on only one end, like a > 1.
 		// We use 1/NDV here (only the Histogram part is considered) and it seems reasonable and good enough for now.
 		return min(rowCount, upperBound)
 	}
 
-	// If the modifyCount is large (compared to original table rows), then any out of range estimate is unreliable.
+	// If the realtimeRowCount is larger than the original table rows, then any out of range estimate is unreliable.
```
Rustin170506, line 1083: Can you give an example of where this change would help? I can understand the changes, but I don't understand the impact of this change. Forgive me for asking too many questions. I just want to make sure I understand the motivation behind this change.

pkg/statistics/histogram.go
```diff
@@ -1049,22 +1073,26 @@ func (hg *Histogram) OutOfRangeRowCount(
...
-	} else if rowCount < upperBound {
-		// Adjust by increaseFactor if our estimate is low
-		rowCount *= increaseFactor
+	addedRows := float64(realtimeRowCount) - hg.TotalRowCount()
+	if addedRows > 1 {
+		// Conservatively - use the larger of the left or right percent - since we are working with
+		// changes to the table since last Analyze - any out of range estimate is unreliable.
+		// if the histogram range is invalid (too small/large - histInvalid) - totalPercent is zero
+		// and we will set rowCount to min of upperbound and added rows
+		totalPercent = min(0.5, max(leftPercent, rightPercent))
```
Rustin170506, line 1091: May I ask why 0.5?

Rustin170506 (commented):
Thanks!

pkg/statistics/histogram.go
qw4990, line 1083: `If the realtimeRowCount is larger than the original table rows, then any out of range estimate is unreliable.` Is this assumption too strict? As long as the data size grow, this assumption will be satisfied.

pkg/statistics/histogram.go
qw4990: I guess it's just an arbitrary number to make the final estimated results not too aggressive.

pkg/statistics/histogram.go
terry1purcell: Simple example - if I have a table with 100 rows, and values 1 to 100.
Select count(*) from t where a > 100; <-- will return count == 0.

If I insert 10 rows, values 101 to 110 - and do NOT run ANALYZE - realtimeRowCount == 110.

Select count(*) from t where a > 100; <-- now the result is count == 10 (because I added 10 values after 100.

Now I run ANALYZE (so rowCount == 110), then insert another 10 rows - this time with values 1-10 - realtimeRowCount == 120.
select count(*) from t where a > 110; <-- realtimeRowCount == 120 - and I don't know that the 10 new rows aren't greater than 110.

terry1purcell: @qw4990 - you are correct that any changes which result in an increase in realtimeRowCount will cause a change from this code. However - for this code to be entered - you must be searching for values outside of the range of the current histograms. If the predicates do not search outside this range - then there is no impact from this change.

Using an example such as 20% would miss those situations where a small number of rows were added - similar to the example that I shared above.

pkg/statistics/histogram.go
terry1purcell: I used 0.5 because I did not want to always assume that any new rows added, or updates to the existing rows were all changing the high end 100% or the low end 100%. If you started with 100 rows, and you added 10 rows - you could've added 5 rows outside the low bound, and 5 rows above the high bound. Or all above or below - or all within the range. Assuming all rows added are 100% in one direction cannot be verified - therefore - 0.5 is used.

qw4990 (approved):

Rustin170506 (approved):
Thanks! :shipit:

PR #57468, opened 2024-11-18
ddl: dynamically adjusting the concurrency and batch size of reorganization job

fzzf678:

Issue Number: ref #57229 #57526

Problem Summary:

Dynamically adjust the worker count and batch size for the reorganization DDL job. A background goroutine checks and updates the worker count every 2 seconds. The batch size is evaluated and adjusted as needed before processing each chunk of data.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/ddl/backfilling.go
```diff
@@ -431,8 +431,14 @@ func (w *backfillWorker) run(d *ddlCtx, bf backfiller, job *model.Job) {
...
 
 		// Change the batch size dynamically.
-		newBatchCnt := job.ReorgMeta.GetBatchSizeOrDefault(int(variable.GetDDLReorgBatchSize()))
-		w.GetCtx().batchCnt = newBatchCnt
+		currentBatchCnt := w.GetCtx().batchCnt
+		targetBatchSize := job.ReorgMeta.GetBatchSize()
+		if targetBatchSize != currentBatchCnt {
+			w.GetCtx().batchCnt = targetBatchSize
+			logger.Info("adjust ddl job config success",
+				zap.Int64("job ID", job.ID),
```
tangenta, line 439:
```suggestion
				zap.Int64("jobID", job.ID),
```

We may need to file another PR to unify the logs from `job ID` to `jobID` for better log searching

pkg/ddl/backfilling.go
```diff
@@ -787,11 +793,58 @@ func (dc *ddlCtx) runAddIndexInLocalIngestMode(
...
 	if err != nil {
 		return err
 	}
+
+	// Adjust worker pool size dynamically.
+	if job != nil && avgRowSize != 0 {
+		go func() {
+			adjustWorkerPoolSize(ctx, pipe, job, avgRowSize)
+		}()
+	}
```
tangenta, line 846:
```suggestion
	// Adjust worker pool size dynamically.
	if job != nil {
		go func() {
			adjustWorkerPoolSize(ctx, pipe, job, avgRowSize)
		}()
	}
```

`expectedIngestWorkerCnt` can handle `avgRowSize == 0` correctly. BTW, is it possible that `job` is nil?

fzzf678: we only pass the non nil `job` from `runAddIndexInLocalIngestMode()`, and if the job is nil, the error will be reported in the previous of `runAddIndexInLocalIngestMode()`.

pkg/ddl/backfilling.go
fzzf678: I will check the previous pr and if necessary modify it in this pr

pkg/ddl/backfilling.go
fzzf678: Here we judge whether the job is nil to decide if it's a local ingest job, if not, we don't start the goroutine.

pkg/ddl/backfilling.go
fzzf678: Check completed, no `job ID` string output to log in PR #57336

pkg/meta/model/reorg.go
```diff
@@ -71,26 +72,37 @@ type DDLReorgMeta struct {
...
 }
 
-// GetBatchSizeOrDefault gets the batch size from DDLReorgMeta or returns the default value.
-func (dm *DDLReorgMeta) GetBatchSizeOrDefault(defaultVal int) int {
-	if dm == nil || dm.BatchSize == 0 {
-		return defaultVal
+// GetBatchSize gets the batch size from DDLReorgMeta.
+func (dm *DDLReorgMeta) GetBatchSize() int {
+	return int(atomic.LoadInt64(&dm.BatchSize))
+}
```
tangenta, line 98: We should handle the `0` case because it is possible when the reorg meta comes from old version TiDB cluster.

pkg/meta/model/reorg.go
```diff
@@ -71,26 +72,37 @@ type DDLReorgMeta struct {
...
 }
 
-// GetConcurrencyOrDefault gets the concurrency from DDLReorgMeta or returns the default value.
-func (dm *DDLReorgMeta) GetConcurrencyOrDefault(defaultVal int) int {
-	if dm == nil || dm.Concurrency == 0 {
-		return defaultVal
+// GetConcurrency gets the concurrency from DDLReorgMeta.
+func (dm *DDLReorgMeta) GetConcurrency() int {
+	return int(atomic.LoadInt64(&dm.Concurrency))
+}
```
tangenta, line 85: Ditto

pkg/meta/model/reorg.go
fzzf678: update in https://github.com/pingcap/tidb/pull/57468/commits/14e8a4c2c2077cff4050d7a1390377defd45e6a8, PTAL again

pkg/meta/model/reorg.go
```diff
@@ -79,8 +79,12 @@ type DDLReorgMeta struct {
...
 
-// GetConcurrency gets the concurrency from DDLReorgMeta.
-func (dm *DDLReorgMeta) GetConcurrency() int {
+// GetConcurrency gets the concurrency from DDLReorgMeta,
+// pass the default value in case of the reorg meta coming from old cluster and Concurrency is 0.
+func (dm *DDLReorgMeta) GetConcurrency(defaultVal int) int {
+	if dm == nil || atomic.LoadInt64(&dm.Concurrency) == 0 {
+		return defaultVal
+	}
 	return int(atomic.LoadInt64(&dm.Concurrency))
```
tangenta: Can we run `atomic.LoadInt64()` once?

pkg/meta/model/reorg.go
```diff
@@ -71,26 +72,44 @@ type DDLReorgMeta struct {
...
 		return defaultVal
 	}
-	return dm.Concurrency
+	return int(atomic.LoadInt64(&dm.Concurrency))
 }
 
-// GetBatchSizeOrDefault gets the batch size from DDLReorgMeta or returns the default value.
+// SetConcurrency sets the concurrency in DDLReorgMeta.
+func (dm *DDLReorgMeta) SetConcurrency(concurrency int) {
+	currentValue := atomic.LoadInt64(&dm.Concurrency)
```
CbcWestwolf: Maybe we can directly store the new value.

> // Note: Don't get or set these two variables directly, use the functions instead.
Can we make these two fields private?

pkg/meta/model/reorg.go
fzzf678: update in https://github.com/pingcap/tidb/pull/57468/commits/7993dcc732854e2e221d796f53a14e9f0de933c5

fzzf678: /cc @CbcWestwolf

pkg/meta/model/reorg.go
```diff
@@ -71,26 +72,40 @@ type DDLReorgMeta struct {
...
 	Version           int64                            `json:"version"`
 	TargetScope       string                           `json:"target_scope"`
-	// These two variables are set when corresponding session variables are set explicitly. When they are set,
-	// user cannot change it by setting the global one. Otherwise, they can be adjusted dynamically through global var.
-	Concurrency int `json:"concurrency"`
-	BatchSize   int `json:"batch_size"`
+	// These two variables are used to control the concurrency and batch size of the reorganization process.
+	// They can be adjusted dynamically through `admin alter ddl jobs` command.
+	// Note: Don't get or set these two variables directly, use the functions instead.
+	Concurrency int64 `json:"concurrency"`
```
CbcWestwolf, line 78: Can we make these two fields private?

CbcWestwolf (approved):

pkg/meta/model/reorg.go
fzzf678: Of course these fields should be private, maybe later we can refactor this part and turn it private.

tangenta (approved):

pkg/executor/show_ddl_jobs.go
fzzf678, line 1: @tangenta PTAL again, reomve the != 0 condition

tangenta: lgtm

Issue #56494, opened 2024-10-09
[FATAL] [lib.rs:478] ["range end index 4 out of range for slice of length 3"]


sayJason:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```sql
CREATE TABLE t1 (c1 bit(24));
INSERT IGNORE t1 (c1) VALUES (b'100011111100001110010100001000000101000011011100000011010001110');
SELECT c1 FROM t1 WHERE FALSE NOT LIKE CONVERT(c1, CHAR);
```
### 2. What did you expect to see? (Required)
Execute without crash
### 3. What did you see instead (Required)
Crash down the server
### 4. What is your TiDB version? (Required)

Release Version: v8.3.0
Edition: Community
Git Commit Hash: 1a0c3ac3292fff7742faa0c00a662ccb66ba40db
Git Branch: HEAD
UTC Build Time: 2024-08-20 10:13:01
GoVersion: go1.21.10
Race Enabled: false
Check Table Before Drop: false
Store: tikv

windtalker:
The root cause is the collation of `like` function is derived to be `utf8mb4_general_ci`, which is not expected. Since the collation of cast(bit as char) should be binary, and the collation of `like` should be binary.

lcwangchao:
> The root cause is the collation of `like` function is derived to be `utf8mb4_general_ci`, which is not expected. Since the collation of cast(bit as char) should be binary, and the collation of `like` should be binary.

I think the collation of `cast(bit as char)` should still be  `utf8mb4_general_ci` wich is the connection collation. However, the cast should return a NULL value if the bit is not a legal utf8 format. That is what I tested in MySQL:

```
mysql> create table t1(a bit(24));
Query OK, 0 rows affected (0.02 sec)

mysql> insert into t1 values(0xffff);
Query OK, 1 row affected (0.01 sec)

mysql> select cast(a as char) from t1;
Field   1:  `cast(a as char)`
Catalog:    `def`
Database:   ``
Table:      ``
Org_table:  ``
Type:       VAR_STRING
Collation:  utf8mb4_general_ci (45)
Length:     96
Max_length: 0
Decimals:   31
Flags:


+-----------------+
| cast(a as char) |
+-----------------+
| NULL            |
+-----------------+
1 row in set, 1 warning (0.01 sec)
```

Issue #51877, opened 2024-03-19
Tikv conv function inconsistent results with tidb


yibin87:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
set tidb_opt_projection_push_down = 'on';
create table bug15583(b BIT(8), n INT);
insert into bug15583 values(128, 128);
insert into bug15583 values(null, null);
insert into bug15583 values(0, 0);
insert into bug15583 values(255, 255);
select conv(b, 10, 2), conv(b + 0, 10, 2) from bug15583;

### 2. What did you expect to see? (Required)
conv(b, 10, 2)  conv(b + 0, 10, 2)
10000000       10000000
NULL    NULL
0       0
11111111       11111111

### 3. What did you see instead (Required)
conv(b, 10, 2)  conv(b + 0, 10, 2)
0       10000000
NULL    NULL
0       0
0       11111111

### 4. What is your TiDB version? (Required)

| Release Version: v7.4.0-alpha-2012-gfe01d11df6-dirty
Edition: Community
Git Commit Hash: https://github.com/pingcap/tidb/commit/fe01d11df67ca60573b028ed4c9bc52db9a17a7e
Git Branch: div_increase_prec
UTC Build Time: 2024-03-15 08:10:07

yibin87:
Checked that tidb is consistent with mysql.

yibin87:
In latest tidb version, the predicate pushdown logic seems changed, and a little strange:
```
mysql> explain select conv(b, 10, 2) from bug15583;
+-------------------------+---------+-----------+----------------+-------------------------------------------------------------+
| id                      | estRows | task      | access object  | operator info                                               |
+-------------------------+---------+-----------+----------------+-------------------------------------------------------------+
| TableReader_8           | 4.00    | root      |                | data:Projection_4                                           |
| └─Projection_4          | 4.00    | cop[tikv] |                | conv(cast(test.bug15583.b, var_string(1)), 10, 2)->Column#4 |
|   └─TableFullScan_7     | 4.00    | cop[tikv] | table:bug15583 | keep order:false, stats:pseudo                              |
+-------------------------+---------+-----------+----------------+-------------------------------------------------------------+
3 rows in set (0.00 sec)

mysql> explain select conv(b + 0, 10, 2) from bug15583;
+-------------------------+---------+-----------+----------------+-----------------------------------------------------------------------+
| id                      | estRows | task      | access object  | operator info                                                         |
[...]

PR #57484, opened 2024-11-19
expression: fix tikv crash when `bool like cast(bit as char)`

lcwangchao:

Issue Number: close #56494

When casting the bit type to a non binary string, this PR first convert bits to binary string first. This can ensure `HandleBinaryLiteral` can be used in the following code and convert binary to the specified collation with right behavior.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
fix tikv crash when `bool like cast(bit as char)`
```

### DIFF
--- a/pkg/expression/builtin_cast.go
+++ b/pkg/expression/builtin_cast.go
@@ -300,10 +300,20 @@ func (c *castAsStringFunctionClass) getFunction(ctx BuildContext, args []Express
 	if err != nil {
 		return nil, err
 	}
-	if args[0].GetType(ctx.GetEvalCtx()).Hybrid() {
-		sig = &builtinCastStringAsStringSig{bf}
-		sig.setPbCode(tipb.ScalarFuncSig_CastStringAsString)
-		return sig, nil
+	if ft := args[0].GetType(ctx.GetEvalCtx()); ft.Hybrid() {
+		castBitAsUnBinary := ft.GetType() == mysql.TypeBit && c.tp.GetCharset() != charset.CharsetBin
+		if !castBitAsUnBinary {
+			sig = &builtinCastStringAsStringSig{bf}
+			sig.setPbCode(tipb.ScalarFuncSig_CastStringAsString)
+			return sig, nil
+		}
+		// for type BIT, it maybe an invalid value for the specified charset, we need to convert it to binary first,
+		// and then convert it to the specified charset with `HandleBinaryLiteral` in the following code.
+		tp := types.NewFieldType(mysql.TypeString)
+		tp.SetCharset(charset.CharsetBin)
+		tp.SetCollate(charset.CollationBin)
+		tp.AddFlag(mysql.BinaryFlag)
+		args[0] = BuildCastFunction(ctx, args[0], tp)
 	}
 	argTp := args[0].GetType(ctx.GetEvalCtx()).EvalType()
 	switch argTp {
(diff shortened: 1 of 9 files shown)


pkg/expression/builtin_convert_charset.go
```diff
@@ -173,6 +173,7 @@ type builtinInternalFromBinarySig struct {
 func (b *builtinInternalFromBinarySig) Clone() builtinFunc {
 	newSig := &builtinInternalFromBinarySig{}
 	newSig.cloneFrom(&b.baseBuiltinFunc)
+	newSig.cannotConvertStringAsWarning = b.cannotConvertStringAsWarning
```
lcwangchao, line 176: I think it is lost in previous PR ? @YangKeao

Defined2014 (approved):

pkg/expression/builtin_cast.go
```diff
@@ -300,10 +300,20 @@ func (c *castAsStringFunctionClass) getFunction(ctx BuildContext, args []Express
 	if err != nil {
 		return nil, err
 	}
-	if args[0].GetType(ctx.GetEvalCtx()).Hybrid() {
-		sig = &builtinCastStringAsStringSig{bf}
-		sig.setPbCode(tipb.ScalarFuncSig_CastStringAsString)
-		return sig, nil
+	if ft := args[0].GetType(ctx.GetEvalCtx()); ft.Hybrid() {
+		castBitAsUnBinary := ft.GetType() == mysql.TypeBit && c.tp.GetCharset() != charset.CharsetBin
```
windtalker, line 304: For other type that store binary string like `Blob`, will TiDB meet the same problem?

qw4990 (approved):
LGTM for the optimizer part.

pkg/expression/expr_to_pb_test.go
```diff
@@ -1590,7 +1591,11 @@ func TestExprPushDownToTiKV(t *testing.T) {
 	require.Len(t, pushed, len(exprs))
 	require.Len(t, remained, 0)
 	exprs = exprs[:0]
-	castByteAsStringFunc, err := NewFunction(mock.NewContext(), ast.Cast, types.NewFieldType(mysql.TypeString), byteColumn)
+	// when conv a column with type BIT, a cast function will be used to cast bit to a binary string
+	castTp := types.NewFieldType(mysql.TypeString)
+	castTp.SetCharset(charset.CharsetBin)
+	castTp.SetCollate(charset.CollationBin)
+	castByteAsStringFunc, err := NewFunction(mock.NewContext(), ast.Cast, castTp, byteColumn)
```
windtalker, line 1598: This function can not be push down to TiKV because `from_binary` can not be push down to TiKV?

YangKeao (approved):
LGTM

pkg/expression/expr_to_pb_test.go
lcwangchao: no, I think it's the issue: https://github.com/pingcap/tidb/issues/51877 The tidb have some special logic for function `conv`. When converting a BIT type, it uses `conv(cast(bit as char))` to cast bit first, but `conv` should also handles the result string as a number in binary format instead of string... 

See: https://github.com/pingcap/tidb/blob/9530fdc2388d025af86eb6fce5e617972c361000/pkg/expression/builtin_math.go#L1281-L1290

lcwangchao: I added a comment in this test

windtalker (approved):
lgtm

pkg/expression/expr_to_pb_test.go
lcwangchao: This test is not related with this issue, but we should update it to set to casted type as bin collations to make the test pass.

Issue #56649, opened 2024-10-15
Optimize memory allocation for index lookup


zyguan:
## Enhancement

After profiling some index lookup workloads, we found that the go runtime overhead are remarkable. By optimizing memory allocation on some critical path, performance can be improved by more than 10%. So we plan to merge those optimizations, and here is the task breakdown.
- Reuse goroutines (in copr, disksql, etc.) to avoid the overhead of new/grow stack.
  - related issues: 
    - https://github.com/pingcap/tidb/issues/52887
    - https://github.com/pingcap/tidb/issues/51852
- Reduce memory allocations of ExecDetails and RuntimeStats.
  - related path: distsql fetch resp, unmarshal resp in recv loop, TableHandlesToKVRanges, etc.
- Reduce memory allocations of BuildCopIterator.
  - related path: BuildRPCContext, SplitKeyRangesByLocations, BatchLocateKeyRanges, batchStoreTaskBuilder, etc.

cfzjywxk:
Close for current stage is finished on master and release-8.5.
/cc @zyguan

PR #57522, opened 2024-11-19
store/copr: optimize copIterator by avoid start new goroutine

crazycs520:
<!--

Thank you for contributing to TiDB!

PR Title Format:
1. pkg [, pkg2, pkg3]: what's changed
2. *: what's changed

-->

<!--

There MUST be one line starting with "Issue Number:  " and
linking the relevant issues via the "close" or "ref".

For more info, check https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/contribute-code.html#referring-to-an-issue.

-->

Issue Number: ref #56649

Problem Summary: optimize cop iterator by avoid start new goroutine

Before this PR, `copIterator` always start at least 1 [copIteratorWorker goroutine](https://github.com/pingcap/tidb/blob/f347b9d334495c72d27c53532dfef522213dbaac/pkg/store/copr/coprocessor.go#L884) and 1 [copIteratorTaskSender goroutine](https://github.com/pingcap/tidb/blob/f347b9d334495c72d27c53532dfef522213dbaac/pkg/store/copr/coprocessor.go#L902). That's too wasteful for small queries.

This PR avoid start new goroutine for small queries, for better performance, following is the sysbench [oltp_index_scan](https://github.com/crazycs520/sysbench_lua) test, and the range size is 2:

```sh
sysbench --config-file=sysbench.conf oltp_index_scan --tables=16 --table-size=1000000 --threads=100 --range-size=2 run
```

The example query is like this:

```sql
SELECT
  k
FROM
  sbtest1
WHERE
  k BETWEEN ?
  AND ? [arguments: (503925, 503926)]
```

Which Plan is:

```plan

| id                 | estRows | estCost | actRows | task      | access object               | execution info                                                                                                                                                                                                                                                                                         | operator info                           | memory    | disk  |
| IndexReader_6      | 2.07    | 31.20   | 2       | root      |                             | time:1.31ms, loops:2, cop_task: {num: 1, max: 1.15ms, proc_keys: 2, tot_proc: 81.2µs, tot_wait: 575.8µs, copr_cache: disabled, build_task_duration: 12.2µs, max_distsql_concurrency: 1}, rpc_info:{Cop:{num_rpc:1, total_time:1.12ms}}                                                                 | index:IndexRangeScan_5                  | 278 Bytes | N/A   |
| └─IndexRangeScan_5 | 2.07    | 336.86  | 2       | cop[tikv] | table:sbtest1, index:k_1(k) | tikv_task:{time:0s, loops:1}, scan_detail: {total_process_keys: 2, total_process_keys_size: 92, total_keys: 3, get_snapshot_time: 539.6µs, rocksdb: {key_skipped_count: 2, block: {cache_hit_count: 6}}}, time_detail: {total_process_time: 81.2µs, total_wait_time: 575.8µs, tikv_wall_time: 751.3µs} | range:[503925,503926], keep order:false | N/A       | N/A   |
```

version | workload | thread | QPS | QPS Increase
-- | -- | -- | -- | --
master | oltp_index_scan | 10 | 9224 |  
This PR | oltp_index_scan | 10 | 9471 | 2.6%
master | oltp_index_scan | 50 | 36072 |  
This PR | oltp_index_scan | 50 | 37392 | 3.6%
master | oltp_index_scan | 100 | 41977 |  
This PR | oltp_index_scan | 100 | 43554 | 3.7%
master | oltp_index_scan | 150 | 42785 |  
This PR | oltp_index_scan | 150 | 46430 | 8.5%



### Check List

Tests <!-- At least one of them must be included. -->

- [ ] Unit test
- [x] Integration test
- [ ] No need to test

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

<!-- compatibility change, improvement, bugfix, and new feature need a release note -->

```release-note
None
```


pkg/store/copr/coprocessor.go
```diff
@@ -1113,20 +1145,62 @@ func (it *copIterator) Next(ctx context.Context) (kv.ResultSubset, error) {
...
+		curTask.respChan = respCh
+		bo := chooseBackoffer(ctx, backoffermap, curTask, worker)
+		tasks, err := worker.handleTaskOnce(bo, curTask, respCh)
+		if err != nil {
+			resp = &copResponse{err: errors.Trace(err)}
+			worker.checkRespOOM(resp, true)
+			return resp
+		}
+		if len(tasks) > 0 {
+			it.tasks = append(tasks, it.tasks[1:]...)
```
zyguan, line 1174: Is it possible that there is a remaining task whose batchTaskList is not empty (then multiple copResponses might be pushed into the respCh)?

pkg/store/copr/coprocessor.go
```diff
@@ -1113,20 +1145,62 @@ func (it *copIterator) Next(ctx context.Context) (kv.ResultSubset, error) {
 	return resp, nil
 }
 
+func (it *copIterator) liteSendReq() (resp *copResponse) {
+	defer func() {
+		r := recover()
+		if r != nil {
+			logutil.BgLogger().Error("copIteratorWork meet panic",
```
cfzjywxk, line 1152: The original `ctx` could be used for the logger..

pkg/store/copr/coprocessor.go
```diff
@@ -1113,20 +1145,62 @@ func (it *copIterator) Next(ctx context.Context) (kv.ResultSubset, error) {
...
+				zap.Any("r", r),
+				zap.Stack("stack trace"))
+			resp = &copResponse{err: util2.GetRecoverError(r)}
+		}
+	}()
+
```
cfzjywxk, line 1158: Better to check the presumed constraints for the `it.tasks` before the `for` loop.

pkg/store/copr/coprocessor.go
```diff
@@ -1858,19 +1916,32 @@ func (worker *copIteratorWorker) collectCopRuntimeStats(copStats *CopRuntimeStat
...
-	if worker.kvclient.Stats == nil {
-		return nil
+func (worker *copIteratorWorker) collectKVClientRuntimeStats(copStats *CopRuntimeStats, bo *Backoffer, rpcCtx *tikv.RPCContext) {
+	defer func() {
+		worker.kvclient.Stats = nil
+	}()
+	copStats.ReqStats = worker.kvclient.Stats
+	backoffTimes := bo.GetBackoffTimes()
+	copStats.BackoffTime = time.Duration(bo.GetTotalSleep()) * time.Millisecond
+	copStats.BackoffSleep = make(map[string]time.Duration, len(backoffTimes))
```
cfzjywxk, line 1926: Why not checking the `worker.kvclient.Stats` first so the map creations could be avoided in some cases?

pkg/store/copr/coprocessor.go
crazycs520: Nice catch, but the `it.liteWorker` only work when there is only 1 task and doesn't have `batchTaskList`, so this case won't happen. https://github.com/pingcap/tidb/pull/57522/files#diff-f341da742ab4e8413960a536dfe0038a2bb92ba2dd6e9f336b5708e19a8b3e60R861

pkg/store/copr/coprocessor.go
crazycs520: https://github.com/pingcap/tidb/pull/57522/files#diff-f341da742ab4e8413960a536dfe0038a2bb92ba2dd6e9f336b5708e19a8b3e60R861 Already checked `it.tasks`.

pkg/store/copr/coprocessor.go
zyguan: But what if the first task meets region error and newly tasks do have `batchTaskList`?

pkg/store/copr/coprocessor.go
crazycs520: It has been checked outside before call this function.

pkg/store/copr/coprocessor.go
crazycs520: great, done.

pkg/store/copr/coprocessor.go
```diff
@@ -1064,7 +1086,17 @@ func (it *copIterator) Next(ctx context.Context) (kv.ResultSubset, error) {
 	})
 	// If data order matters, response should be returned in the same order as copTask slice.
 	// Otherwise all responses are returned from a single channel.
-	if it.respChan != nil {
+
+	failpoint.InjectCall("CtxCancelBeforeReceive", ctx)
+	if it.liteWorker != nil {
+		resp = it.liteSendReq(ctx)
```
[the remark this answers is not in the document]
crazycs520: Nice catch, addressed, look like only one cop-reader can use cop-lite-worker.

pkg/store/copr/coprocessor.go
```diff
@@ -850,9 +857,16 @@ func (worker *copIteratorWorker) run(ctx context.Context) {
 }
 
 // open starts workers and sender goroutines.
-func (it *copIterator) open(ctx context.Context, enabledRateLimitAction, enableCollectExecutionInfo bool) {
+func (it *copIterator) open(ctx context.Context, tryCopLiteWorker *uint32) {
+	if len(it.tasks) == 1 && tryCopLiteWorker != nil && atomic.LoadUint32(tryCopLiteWorker) == 0 {
+		atomic.StoreUint32(tryCopLiteWorker, 1)
```
cfzjywxk, line 862: This is not atomic operation, the `CAS` should used instead.

But why it's designed to be read-write concurrently? Should the `open` be a single-threaded operation?

pkg/store/copr/coprocessor.go
cfzjywxk: Sharing internal states across multiple functions and assuming that some states are manipulated in other functions is not a robust programming practice. 

Functions should minimize mutual state dependencies as much as possible, and dependencies should be passed explicitly through parameters.

Or at least, the cross-functions assumptions should be stated clearly in the comments.

pkg/store/copr/coprocessor.go
cfzjywxk: Ditto.

pkg/store/copr/coprocessor.go
crazycs520: In some cases (such as index lookup reader), there are maybe some `TableReader` execute `open` concurrently.

pkg/store/copr/coprocessor.go
crazycs520: Nice catch, already fix it.

pkg/store/copr/coprocessor.go
crazycs520: After modifications, there are no assumptions for `it.tasks` here.

pkg/store/copr/coprocessor.go
crazycs520: ok, already add the check here.

pkg/store/copr/coprocessor.go
```diff
@@ -850,9 +858,15 @@ func (worker *copIteratorWorker) run(ctx context.Context) {
 }
 
 // open starts workers and sender goroutines.
-func (it *copIterator) open(ctx context.Context, enabledRateLimitAction, enableCollectExecutionInfo bool) {
+func (it *copIterator) open(ctx context.Context, tryCopLiteWorker *uint32) {
+	if len(it.tasks) == 1 && tryCopLiteWorker != nil && atomic.CompareAndSwapUint32(tryCopLiteWorker, 0, 1) {
```
cfzjywxk, line 862: How about using the `boolean` type for `tryCopLiteWorker` if its value is just on or off instead of checking `pointer == nil`?

pkg/store/copr/coprocessor.go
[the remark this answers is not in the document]
cfzjywxk: Better to add comments about the to be used scenarios above the `cop-lite-worker`.

pkg/store/copr/coprocessor.go
```diff
@@ -1113,20 +1146,65 @@ func (it *copIterator) Next(ctx context.Context) (kv.ResultSubset, error) {
...
+		for len(it.respChan) > 0 {
+			select {
+			case resp = <-it.respChan:
+				return resp
+			default:
+			}
+		}
+		if len(it.tasks) == 0 {
+			return nil
+		}
```
cfzjywxk, line 1172: How about moving the check to L 1159 instead of doing it in the `for` loop?

pkg/store/copr/coprocessor.go
cfzjywxk: Is there a test case to cover this situation?

pkg/store/copr/coprocessor.go
```diff
@@ -1113,20 +1146,65 @@ func (it *copIterator) Next(ctx context.Context) (kv.ResultSubset, error) {
...
+		curTask := it.tasks[0]
+		if cap(it.respChan) < (len(curTask.batchTaskList) + 2) {
+			it.respChan = make(chan *copResponse, 2+len(curTask.batchTaskList))
```
zyguan, line 1175: Here we assume that all copResponses are fill to the respChan after calling handleTaskOnce, how about just let handleTaskOnce return a slice of copResponses (instead of using a channel)?

```go
		for len(it.respChan) > 0 {
			select {
			case resp = <-it.respChan:
				return resp
			default:
			}
		}
```

The above code also assume that no copResponse will be pushed into the respChan asynchronously, but actually current code cannot guarantee that (eg. someone may change the impl of handleTaskOnce and let it write to respCh asynchronously).

pkg/store/copr/coprocessor.go
you06: The workers under the same session is individual, I think the lite worker can be used in different `TableReader`s under one index lookup executor. However, the `DistSQLContext` is shared in session level.

By using `if atomic.CompareAndSwapUint32(tryCopLiteWorker, 0, 1) {...}`, only one cop iterator can benefit from the lite optimization.

pkg/store/copr/coprocessor.go
crazycs520: I don't understand, CAS operations need to be done here, `boolean`  can't implement?

pkg/store/copr/coprocessor.go
crazycs520: Nice catch, I'm working on this.

pkg/distsql/context/context.go
```diff
@@ -85,6 +85,9 @@ type DistSQLContext struct {
 	SessionAlias                string
 
 	ExecDetails *execdetails.SyncExecDetails
+
+	// Only one cop-reader can use lite worker. Using lite-worker in multiple readers will affect the concurrent execution of readers.
+	TryCopLiteWorker uint32
```
you06, line 90: I'm inclined that the `TryCopLiteWorker` is a [request](https://github.com/pingcap/tidb/blob/a3d60bcfc6c1f18c67eeb319fc70cf26befb8687/pkg/kv/kv.go#L529-L530) level option.

Like an index lookup with 1 row result set. If `TryCopLiteWorker` is a session level option, it's only enabled once in index scan, and in table lookup, the `atomic.CompareAndSwapUint32(tryCopLiteWorker, 0, 1)` will fail and execute with multi-goroutine model.

crazycs520: I'm afraid that make `TryCopLiteWorker` in `request` level option may have similar problem in this [comment](https://github.com/pingcap/tidb/pull/57522#discussion_r1875515863), there is a known case and I have added [test](https://github.com/pingcap/tidb/pull/57522/commits/a3c7e530163c2572dc988dbe908b509edae45bc3#diff-68cc9873623c1d811b385e0f6786521e0a05ff90b96678508881406c6a4d5e8dR321) for it. And I'm not sure if there are other cases like this, so I prefer keep `TryCopLiteWorker ` to be statement level.

As index lookup with 1 row result set, you are right, but because of the above problems, I don't have a good way to fix this.

pkg/store/copr/coprocessor.go
cfzjywxk: @crazycs520 
Is it possible to make the lite worker to be used for all the table workers?

pkg/store/copr/coprocessor.go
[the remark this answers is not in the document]
cfzjywxk: @you06 @crazycs520 
Is it possible to trigger the kv request sending in the `Open` function call, so the result could be prepared doing Next?

zyguan (approved):

pkg/store/copr/coprocessor.go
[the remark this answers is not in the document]
you06: > Is it possible to trigger the kv request sending in the `Open` function call

It's possible that the batch kv client returns a future, but it needs big changes to the kv client's interface.

you06 (commented):
LGTM, once we have async kv interface, we can remove `TryCopLiteWorker` and make more coprocessor tasks benefit from this opimization.
Wait for @cfzjywxk's approve.

cfzjywxk (approved):

PR #57553, opened 2024-11-20
ddl: support pre-split index regions before creating index

### LINKED ISSUE #57551: Cluster suffering write hotspot after creating index
## Bug Report

A write workload that has high concentration of writes to a small range of index columns.

No write hotspot.

One of the TiKV instances is overloaded due to the index write hotspot.

master

### LINKED ISSUE #57552: Support pre-split index regions before creating index
## Feature Request

```sql
CREATE TABLE `test` (
    `a` bigint NOT NULL,
    `b` bigint NOT NULL,
    `c` bigint DEFAULT NULL,
    PRIMARY KEY (`a`, `b`)
) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4 COLLATE = utf8mb4_bin;

```

```sql
CREATE INDEX `idx` ON `test` (`a`, `c`, `b`);
```

Right after the "CREATE INDEX" statement was sent, the cluster latency increased significantly, because one TiKV instance is overloaded due to the index write hotspot. The latency was reduced when PD split and scheduled the hotspot regions to other TiKV instances.

The cause of write hotspot could be one of follows:

- Sequential Inserts: When new rows are inserted with sequential or monotonically increasing values for the indexed column, such as timestamps or auto-incrementing primary keys.
- Skewed Data Distribution: When the data distribution is heavily skewed, causing a disproportionate number of writes to a specific range of index keys.
- High Write Frequency: When there is a high frequency of write operations (inserts, updates, deletes) targeting the same index keys or a small range of keys.

Split temp index regions(or index regions in "txn" mode) before the index state becomes "delete-only". Because TiDB doesn't have information about the workload, we have to guess the upper bound and lower bound index values of incoming traffic and split several regions.

A possible extension is introducing the pre-split "index_option" to let users provide more information about the workload.

```sql
-- pre-split into 4 regions. The range is calculated automatically.
ALTER TABLE t ADD INDEX idx(col1, col2) PRE_SPLIT_REGIONS=4;
CREATE INDEX idx on t (col1, col2) PRE_SPLIT_REGIONS=4;

-- pre-split into 4 regions and specify lower and upper bound.
ALTER TABLE t ADD INDEX idx(col1, col2) PRE_SPLIT_REGIONS = (BETWEEN ('a', 10) AND ('z', 100) REGIONS 4);

-- pre-split on specified index keys.
ALTER TABLE t ADD INDEX idx(col1, col2) PRE_SPLIT_REGIONS = (BY ('a', 10), ('b', 20), ('c', 30));
```

`PRE_SPLIT_REGION...BETWEEN` analogs the behavior of BETWEEN AND clause:
```
SPLIT TABLE ... INDEX ... BETWEEN (...) AND (...) REGIONS ...;
```

And `PRE_SPLIT_REGION...BY` analogs the behavior of SPLIT BY clause:
```
SPLIT TABLE ... INDEX ... BY (...), (...), ...
```

TiDB has already supported `PRE_SPLIT_REGIONS` as a table attribute for `CREATE TABLE `statements, but there is no attribute similar to `PRE_SPLIT_REGIONS...BETWEEN`. This is because the timing for users to switch write traffic is under control. Before switching, users can choose to perform `SPLIT TABLE`. However, this is not the same use case as adding an index.

- Use pd-ctl / TiDB HTTP API, but they don't support splitting a region.
- Use `SPLIT TABLE`, but it doesn't support splitting a non-exists region. We have to block DDL so that the index state can remain at "delete-only".

tangenta:

Issue Number: close #57552, close #57551

Problem Summary:

See #57552.

- Support specifying the split option in the index option at parser. 
- Split the target index region by keys or ranges before data reorganization when adding an index.
- Extract common code and tests from `pkg/executor/split.go` to `pkg/util/split.go`.
- Add two functional tests.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/parser/parser.y
```diff
@@ -6643,6 +6645,20 @@ IndexOption:
 			Global: false,
 		}
 	}
+|   "PRE_SPLIT_REGIONS" "=" '(' SplitOption ')'
```
lance6716, line 6648: in line 6598 we uses `EqOpt` rather than literal `"="`

pkg/parser/ast/ddl.go
```diff
@@ -735,6 +735,7 @@ type IndexOption struct {
 	Visibility   IndexVisibility
 	PrimaryKeyTp model.PrimaryKeyType
 	Global       bool
+	SplitOpt     *SplitOption `json:"-"`
```
lance6716, line 738: can you explain in comment why need to skip JSON encoding for this field?

pkg/parser/ast/ddl.go
```diff
@@ -830,6 +850,32 @@ func (n *IndexOption) Accept(v Visitor) (Node, bool) {
 		return v.Leave(newNode)
 	}
 	n = newNode.(*IndexOption)
+	if n.SplitOpt != nil {
+		for i, val := range n.SplitOpt.Lower {
```
lance6716, line 854: it better to let `SplitOption` implement `Accept`, so here and `SplitRegionStmt` won't repeat

pkg/meta/model/job_args.go
```diff
@@ -1322,6 +1322,16 @@ type IndexArg struct {
 	IndexID  int64 `json:"index_id,omitempty"`
 	IfExist  bool  `json:"if_exist,omitempty"`
 	IsGlobal bool  `json:"is_global,omitempty"`
+
+	SplitOpt *IndexArgSplitOpt `json:"split_opt,omitempty"`
+}
+
+// IndexArgSplitOpt is a field of IndexArg used by index presplit.
+type IndexArgSplitOpt struct {
```
lance6716, line 1330: also omitempty?

pkg/parser/parser_test.go
```diff
@@ -3179,6 +3179,17 @@ func TestDDL(t *testing.T) {
 		{"ALTER TABLE t ADD INDEX (a) USING BTREE COMMENT 'a'", true, "ALTER TABLE `t` ADD INDEX(`a`) USING BTREE COMMENT 'a'"},
 		{"ALTER TABLE t ADD INDEX IF NOT EXISTS (a) USING BTREE COMMENT 'a'", true, "ALTER TABLE `t` ADD INDEX IF NOT EXISTS(`a`) USING BTREE COMMENT 'a'"},
 		{"ALTER TABLE t ADD INDEX (a) USING RTREE COMMENT 'a'", true, "ALTER TABLE `t` ADD INDEX(`a`) USING RTREE COMMENT 'a'"},
+		{"ALTER TABLE t ADD INDEX (a) PRE_SPLIT_REGIONS = 4", true, "ALTER TABLE `t` ADD INDEX(`a`) PRE_SPLIT_REGIONS = 4"},
```
D3Hunter, line 3182:
```suggestion
		{"ALTER TABLE t ADD INDEX (a) PRE_SPLIT_REGIONS 4", true, "ALTER TABLE `t` ADD INDEX(`a`) PRE_SPLIT_REGIONS "},
```
add a case for optional `=`

pkg/ddl/index_presplit.go
```diff
@@ -0,0 +1,417 @@
...
+	idxInfo *model.IndexInfo,
+	splitIdxKeys [][]byte,
+) error {
+	s, ok := store.(kv.SplittableStore)
+	if !ok {
+		return nil
+	}
+	start := time.Now()
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, sctx.GetSessionVars().GetSplitRegionTimeout())
+	defer cancel()
```
D3Hunter, line 262: we need at least make sure split region is done, to avoid hot write, and we need retry on split failure. scatter region can wait at best, so maybe move this timeout for scatter only.

pkg/parser/parser_test.go
tangenta: See the next line

pkg/ddl/index_presplit.go
tangenta: If `SplitTable` is timed out, it will return an error and retry at caller side. It will not proceed to the next state.

D3Hunter: currently, this function doesn't return error

tangenta: d64f5840da596f

pkg/ddl/index_presplit.go
```diff
@@ -0,0 +1,417 @@
...
+		betweenLower: lowerValues,
+		betweenUpper: upperValues,
+	}
+
+	maxSplitRegionNum := int64(config.GetGlobalConfig().SplitRegionMaxNum)
+	if opt.Num > maxSplitRegionNum {
+		return nil, errors.Errorf("Split index region num exceeded the limit %v", maxSplitRegionNum)
+	} else if opt.Num < 1 {
+		return nil, errors.Errorf("Split index region num should more than 0")
+	}
```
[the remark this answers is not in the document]
tangenta: https://github.com/pingcap/tidb/pull/57553/commits/c96e2e3e5961c81c4ab6f8f45b7935dc3e6eb4fc

pkg/meta/model/job_args.go
```diff
@@ -1322,6 +1322,16 @@ type IndexArg struct {
 	IndexID  int64 `json:"index_id,omitempty"`
 	IfExist  bool  `json:"if_exist,omitempty"`
 	IsGlobal bool  `json:"is_global,omitempty"`
+
+	SplitOpt *IndexArgSplitOpt `json:"split_opt,omitempty"`
```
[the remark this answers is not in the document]
tangenta: https://github.com/pingcap/tidb/pull/57553/commits/c96e2e3e5961c81c4ab6f8f45b7935dc3e6eb4fc

D3Hunter (approved):

pkg/ddl/index_presplit.go
```diff
@@ -0,0 +1,413 @@
...
+	"github.com/pingcap/tidb/pkg/util/chunk"
+	contextutil "github.com/pingcap/tidb/pkg/util/context"
+	"github.com/pingcap/tidb/pkg/util/dbterror/exeerrors"
+	"github.com/pingcap/tidb/pkg/util/dbterror/plannererrors"
+	"go.uber.org/zap"
+)
+
+// TODO(tangenta): support global index.
+// Wrap the job.Query to with special comments.
+func presplitIndexRegions(
```
bb7133, line 45: preSplitIndexRegions?

pkg/executor/split_test.go
```diff
@@ -38,52 +37,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestLongestCommonPrefixLen(t *testing.T) {
```
Benjamin2037, line 41: Why delete these two tests?

tangenta: Moved to `split_test.go`

Benjamin2037 (approved):
LGTM

Frank945946 (approved):

benmeadowcroft (approved):

yudongusa (approved):

PR #57560, opened 2024-11-20
ddl: Fixed partitioning a non-partitioned table with placement rules

### LINKED ISSUE #55705: Fail to partition table after changing placement rule
## Bug Report

```sql
create or replace placement policy p2 primary_region="region3", regions="region1,region2, region3";

CREATE TABLE t0(c5 FLOAT ZEROFILL PRIMARY KEY );
ALTER TABLE t0 PLACEMENT POLICY=p2;

ALTER TABLE t0 PARTITION BY HASH(c5) PARTITIONS 7;
```

Statement executed successfully.

```sql
mysql> ALTER TABLE t0 PARTITION BY HASH(c5) PARTITIONS 7;
ERROR 8243 (HY000): "[PD:placement:ErrBuildRuleList]build rule list failed, multiple leader replicas for range {7480000000000000FF9B00000000000000F8, 7480000000000000FF9C00000000000000F8}"
```

```sql
 Release Version: v8.4.0-alpha-21-g162cb15402
Edition: Community
Git Commit Hash: 162cb15402de471152a1e59e2ed5d8d1ffcbb4b0
Git Branch: master
UTC Build Time: 2024-08-22 08:12:28
GoVersion: go1.22.5
Race Enabled: false
Store: tikv
```
tikv was set up with partitioned raft kv
labels are as follows:

```sql
mysql> SHOW PLACEMENT LABELS;
+--------+-----------------------------------+
| Key    | Values                            |
+--------+-----------------------------------+
| host   | ["host1", "host2", "host3"]       |
| region | ["region1", "region2", "region3"] |
| zone   | ["zone1", "zone2", "zone3"]       |
+--------+-----------------------------------+
3 rows in set (0.00 sec)
```

mjonss:

Issue Number: close #55705

Problem Summary:
When updating the placement rule bundle, the table ID was included both as the table and the partition id.

Not adding it as partition id for non-partitioned table (which uses the original table as a single partition, hence the same ID).

Also simplified the bundle handling during schema update / ApplyDiff, by removing logic that:
- first went through all partitions and marked the partition to update the bundle, then went through all partition updates and updated its table

to

- mark the table directly that it should update the bundle.

For REORGANIZE PARTITION, there the bundles will be updated twice, first with the intermediate set of partitions (both old and new), so that data copying and index creations on the new partitions will follow the correct policies directly, and then with the final partition set, avoiding having to move the data again when the DDL is done.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
fixed an issue with partitioning a non-partitioned table with placement rules.
```


pkg/domain/infosync/placement_manager.go
```diff
@@ -107,11 +109,38 @@ func (m *mockPlacementManager) PutRuleBundles(_ context.Context, bundles []*plac
...
+		}
+	}
+
+	// Check that no bundles are overlapping
+	type keyRange struct {
+		start string
+		end   string
+	}
+	keys := make([]keyRange, 0, rules)
+	for k := range m.bundles {
```
lcwangchao, line 128: I think we only need to check the rules inside one bundle just now. If `GroupBundle.Override` is set to `true`, it is allowed to have multiple leader rules. 

And in one `RuleGroup` when a rule has `Override==true`, it should also be skipped.

pkg/ddl/placement_policy_test.go
```diff
@@ -2401,3 +2401,15 @@ func TestRecoverTableWithPlacementPolicy(t *testing.T) {
...
+func TestPartitionByWithLabels(t *testing.T) {
+	store, _ := testkit.CreateMockStoreAndDomain(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+
+	tk.MustExec("create placement policy pp1 primary_region='r1' regions='r1,r2'")
+
+	tk.MustExec(`CREATE TABLE t1 (id INT)`)
+	tk.MustExec(`ALTER TABLE t1 placement policy pp1`)
+	tk.MustExec(`ALTER TABLE t1 PARTITION BY HASH (id) PARTITIONS 3`)
```
lcwangchao, line 2414: You can add `checkExistTableBundlesInPD(t, do, "test", "t1")` here to check the rules in PD is correct.

pkg/ddl/partition.go
```diff
@@ -276,8 +276,13 @@ func alterTablePartitionBundles(t *meta.Mutator, tblInfo *model.TableInfo, addin
 	// tblInfo do not include added partitions, so we should add them first
 	tblInfo = tblInfo.Clone()
 	p := *tblInfo.Partition
-	p.Definitions = append([]model.PartitionDefinition{}, p.Definitions...)
-	p.Definitions = append(tblInfo.Partition.Definitions, addingDefinitions...)
+	if p.DDLAction == model.ActionAlterTablePartitioning && p.Type == pmodel.PartitionTypeNone {
+		// skip adding the original table as partition
+		p.Definitions = []model.PartitionDefinition{}
+	} else {
+		p.Definitions = append([]model.PartitionDefinition{}, p.Definitions...)
```
Defined2014, line 283: Seems this branch is useless.

pkg/ddl/placement_policy_test.go
Defined2014, line 2414: Not sure, but it looks like this sql will not fail in a unistore environment

pkg/ddl/partition.go
mjonss: It is the same as the original code, most likely only to replace the original slice struct. I will check if I can remove it.

pkg/ddl/placement_policy_test.go
mjonss: If removing the change in `partition.go` it will fail in the check in `(* mockPlacementManager) PutRuleBundles()`

Defined2014 (approved):

pkg/infoschema/builder.go
```diff
@@ -119,16 +119,16 @@ func applyTruncateTableOrPartition(b *Builder, m meta.Reader, diff *model.Schema
...
+	b.markTableBundleShouldUpdate(diff.TableID)
+	// TODO: check that all partitions are updated to the cache!
 
 	for _, opt := range diff.AffectedOpts {
 		if diff.Type == model.ActionTruncateTablePartition {
 			// Reduce the impact on DML when executing partition DDL. eg.
 			// While session 1 performs the DML operation associated with partition 1,
 			// the TRUNCATE operation of session 2 on partition 2 does not cause the operation of session 1 to fail.
 			tblIDs = append(tblIDs, opt.OldTableID)
-			b.markPartitionBundleShouldUpdate(opt.TableID)
```
Defined2014, line 131: Does it will miss update partitions?

Defined2014: Oh, I found `updateTableBundles` function will also update partition bundles.

Defined2014 (approved):

pkg/domain/infosync/placement_manager.go
lcwangchao: We still need to check the rules in one bundle even `bundle[k].Override is true`, but the current implementation skips all rules in such bundle?

pkg/ddl/partition.go
```diff
@@ -270,20 +265,36 @@ func alterTableLabelRule(schemaName string, meta *model.TableInfo, ids []int64)
...
+func alterTablePartitionBundles(t *meta.Mutator, tblInfo *model.TableInfo, addingDefinitions []model.PartitionDefinition) (bool, error) {
 	var bundles []*placement.Bundle
 
-	// tblInfo do not include added partitions, so we should add them first
+	// We want to achieve:
+	// - before we do any reorganization/write to new partitions/global indexes that the placement rules are in-place
+	// - not removing any placement rules for removed partitions
+	// - not leaving anything in case of failure/rollback!
+	// So we will:
+	// 1) First write the new bundles including both new and old partitions,
```
lcwangchao, line 276: Why the new bundles should include old partitions? The old bundles have already defined these rules and still take effects here.

pkg/ddl/placement_policy_test.go
```diff
@@ -2401,3 +2412,21 @@ func TestRecoverTableWithPlacementPolicy(t *testing.T) {
 		" PARTITION `p2` VALUES LESS THAN (10000))"))
 	checkExistTableBundlesInPD(t, dom, "test", "tp3")
 }
+
+func TestPartitionByWithLabels(t *testing.T) {
```
lcwangchao, line 2416: `TestAlterPartitioningWithPlacementPolicy` should be a better name?

pkg/ddl/placement_policy_test.go
```diff
@@ -2401,3 +2412,21 @@ func TestRecoverTableWithPlacementPolicy(t *testing.T) {
...
+
+	tk.MustExec(`CREATE TABLE t1 (id INT)`)
+	tk.MustExec(`ALTER TABLE t1 placement policy pp1`)
+	tk.MustExec(`ALTER TABLE t1 PARTITION BY HASH (id) PARTITIONS 3`)
+	tk.MustQuery("show create table t1").Check(testkit.Rows("" +
+		"t1 CREATE TABLE `t1` (\n" +
+		"  `id` int(11) DEFAULT NULL\n" +
+		") ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin /*T![placement] PLACEMENT POLICY=`pp1` */\n" +
+		"PARTITION BY HASH (`id`) PARTITIONS 3"))
+	checkExistTableBundlesInPD(t, do, "test", "t1")
```
lcwangchao, line 2431: We can add more test cases:

1. altering a partitioned table, for example:

```
CREATE TABLE tp(id int) placement policy p1
PARTITION BY RANGE(id) (
    PARTITION p1 VALUES LESS THAN (10) placement policy p2,
    PARTITION p2 VALUES LESS THAN MAXVALUE
);

alter table tp
PARTITION BY RANGE(id) (
    PARTITION p1 VALUES LESS THAN (1),
    PARTITION p2 VALUES LESS THAN (2) placement policy p2,
    PARTITION p3 VALUES LESS THAN MAXVALUE
);
```

2. check the old rule still exists before GC
3. check the old rules are deleted after GC
4. the rollback workflow should delete bundles

pkg/ddl/partition.go
```diff
@@ -3307,31 +3322,21 @@ func (w *worker) onReorganizePartition(jobCtx *jobContext, job *model.Job) (ver
...
-			if err = infosync.PutRuleBundlesWithDefaultRetry(context.TODO(), bundles); err != nil {
-				if !changesMade {
-					job.State = model.JobStateCancelled
-					return ver, errors.Wrapf(err, "failed to notify PD the placement rules")
-				}
-				return rollbackReorganizePartitionWithErr(jobCtx, job, err)
-			}
-			changesMade = true
-		}
+		changesMade = changesMade || changed
```
lcwangchao, line 3333: Seems this method is using `rollbackLikeDropPartition` to rollback a DDL. `rollbackLikeDropPartition` does not rollback the group with ID `TiDB_DDL_{newTableID}`

lcwangchao:
@xhebox PTAL

pkg/ddl/partition.go
mjonss: Oh, so if the Bundle is overwritten, the older bundle rules will still be in effect?
Like if the first bundle had rules for table id 111, part id 112 and 113, and the new bundle would have table id 111, partition id 113 and 114, would part id 112 still be placed as the old bundle rules or would it be placed without any rules?

pkg/ddl/partition.go
mjonss: newTableID will be added to the `TablePartitionArgs.OldPhysicalTableIDs` along with the other new rolled backed partitions in `rollbackLikeDropPartition`, so that it will later on be picked up by GC to both clean up old rows as well as the Table and Partition Bundles.

pkg/ddl/placement_policy_test.go
mjonss: I have added several tests, both for normal flow as well as rollback cases with error injection. Please review and see if there are any more test you would like to see? :)

pkg/ddl/partition.go
xhebox: > or would it be placed without any rules?

This, I believe.

```
// it will filter the rules depends on the interval index override in the same group or the
// group-index override between different groups
// For example, given rules:
// ruleA: group_id: 4, id: 2, override: true
// ruleB: group_id: 4, id: 1, override: true
// ruleC: group_id: 3
// ruleD: group_id: 2
// RuleGroupA: id:4, override: false
// RuleGroupB: id:3, override: true
// RuleGroupC: id:2, override: false
// Finally only ruleA and ruleC will be selected.
```

pkg/domain/infosync/placement_manager.go
xhebox: > We still need to check the rules in one bundle even `bundle[k].Override is true`

true.

pkg/ddl/partition.go
```diff
@@ -270,33 +265,48 @@ func alterTableLabelRule(schemaName string, meta *model.TableInfo, ids []int64)
...
-	// tblInfo do not include added partitions, so we should add them first
 	tblInfo = tblInfo.Clone()
-	p := *tblInfo.Partition
-	p.Definitions = append([]model.PartitionDefinition{}, p.Definitions...)
-	p.Definitions = append(tblInfo.Partition.Definitions, addingDefinitions...)
-	tblInfo.Partition = &p
+	p := tblInfo.Partition
+	if p != nil {
+		// skip the original table as partition if partitioning a non-partitioned table
+		if p.Definitions[0].ID != tblInfo.ID {
```
lcwangchao, line 281: Could you add more comments that when `p.Definitions[0].ID == tblInfo.ID` ? It is a little hack and need more explainations

pkg/ddl/partition.go
```diff
@@ -337,19 +347,15 @@ func updateAddingPartitionInfo(partitionInfo *model.PartitionInfo, tblInfo *mode
 }
 
 // rollbackAddingPartitionInfo remove the `addingDefinitions` in the tableInfo.
-func rollbackAddingPartitionInfo(tblInfo *model.TableInfo) ([]int64, []string, []*placement.Bundle) {
+func rollbackAddingPartitionInfo(tblInfo *model.TableInfo) ([]int64, []string) {
```
lcwangchao, line 350: I think this function should be renamed to `removePartitionAddingDefinitionsFromTableInfo` now ?

pkg/ddl/partition.go
```diff
@@ -2147,20 +2153,18 @@ func dropLabelRules(ctx context.Context, schemaName, tableName string, partNames
...
 		return ver, errors.Trace(err)
 	}
 	tblInfo.Partition.DroppingDefinitions = nil
-	physicalTableIDs, pNames, rollbackBundles := rollbackAddingPartitionInfo(tblInfo)
-	err = infosync.PutRuleBundlesWithDefaultRetry(context.TODO(), rollbackBundles)
-	if err != nil {
-		job.State = model.JobStateCancelled
-		return ver, errors.Wrapf(err, "failed to notify PD the placement rules")
-	}
+	physicalTableIDs, pNames := rollbackAddingPartitionInfo(tblInfo)
```
lcwangchao, line 2167: Also, could you add some comments that `physicalTableIDs` will be assigned to `args. OldPhysicalTblIDs` and it will be used in GC to delete the related bundles. And explain why we do not delete it right now when rolling up.

pkg/ddl/partition.go
mjonss: A better example:
```
CREATE TABLE t (a int) partition by range (a) (partition p0 values less than (1000000), partition p1M values less than (2000000)) placement policy pp1;
ALTER TABLE t REORGANIZE PARTITION p1M INTO (PARTITION p1M values less than (2000000), partition p2M values less than (3000000));
```
During the DDL we would still want the original p1M to follow the table level rule, so we keep it in the bundle.
When the final bundles will be created it will not be included.
Since the table keeps its table id (for REORGANIZE PARTITION) there are no other bundles that covers the inherited table's placement rules for the original partition p1M, which means we need to cover it during the time of the DDL, since it can still be accessed (and double written).

Another alternative would be to create partition level bundles for the old replaced partitions, and let them be removed by GC later, but I think that would qualify for a separate PR.

pkg/domain/infosync/placement_manager.go
mjonss: @lcwangchao @xhebox please review the new duplicate leader range check.

pkg/domain/infosync/placement_manager.go
```diff
@@ -107,11 +109,50 @@ func (m *mockPlacementManager) PutRuleBundles(_ context.Context, bundles []*plac
...
+		if keys[i].start == keys[j].start {
+			if keys[i].id == keys[j].id {
+				return keys[i].end < keys[j].end
+			}
+			return keys[i].id < keys[j].id
+		}
+		return keys[i].start < keys[j].start
+	})
+	for i := 1; i < len(keys); i++ {
+		if keys[i].start < keys[i-1].end && !keys[i].override {
```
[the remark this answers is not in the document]
lcwangchao: And for below case, I think it should still report an error but seems not

```
{
  group_id: 1,
  rules: [
    {override: false, start: 0x01, end: 0x0F},
    {override: true, start: 0x02, end: 0x03},
    {override: false, start: 0x04, end: 0x05}
  ]
}
```

pkg/ddl/partition.go
mjonss: I think the rule bundle check improvements would be better addressed in a separate PR, I will update the check and add a unit test here, but when looking at the code in PD, it looks like the example in the comment for [`prepareRulesForApply()`](https://github.com/tikv/pd/blob/52b42b25ba8f8f75943200b76d48565b8aace2b8/pkg/schedule/placement/rule.go#L182-L190) is wrong, if I'm reading the code correctly, I would expect it to return 'ruleC and ruleD`, since ruleB would override ruleA, and ruleC would override all previous rules, due to its group has override set, and finally ruleD would just be added.

And it does not check any key ranges, neither in this function or in `checkApplyRules()`, which only checks that there are at least one leader or voter, and max one leader.

pkg/domain/infosync/placement_manager.go
mjonss: I have changed the logic to only check bundles one by one, and tried to mimic the logic of PDs `prepareRulesForApply()` and `checkApplyRules()`, for any other check I created https://github.com/pingcap/tidb/issues/57693 as a follow up.

pkg/ddl/partition.go
mjonss: I have changed the logic to only check bundles one by one, and tried to mimic the logic of PDs `prepareRulesForApply()` and `checkApplyRules()`, for any other check I created https://github.com/pingcap/tidb/issues/57693 as a follow up.

pkg/domain/infosync/placement_manager.go
```diff
@@ -99,6 +101,94 @@ func (m *mockPlacementManager) GetAllRuleBundles(_ context.Context) ([]*placemen
...
+	// Could use pd's placement.sortRules() instead.
+	sort.Slice(keys, func(i, j int) bool {
+		if keys[i].start == keys[j].start {
+			if keys[i].id == keys[j].id {
+				return keys[i].end < keys[j].end
+			}
+			return keys[i].id < keys[j].id
+		}
+		return keys[i].start < keys[j].start
+	})
```
lcwangchao, line 172: Maybe we can make above logic more simple by:

```
	for _, rule := range bundle.Rules {
		if rule.GroupID != bundle.ID {
			return errors.New(....)
		}
		if rule.Role == pd.Leader && rule.Override {
		        keys = append(keys, keyRange{
			        start:    rule.StartKeyHex,
			        end:      rule.EndKeyHex,
		        })
		}
	}
	if len(keys) == 0 {
		return nil
	}
	// Sort on Start, id, end
	// Could use pd's placement.sortRules() instead.
	sort.Slice(keys, func(i, j int) bool {
		if keys[i].start == keys[j].start {
			return keys[i].end < keys[j].end
		}
		return keys[i].start < keys[j].start
	})
```

Seems [prepareRulesForApply](https://github.com/tikv/pd/blob/master/pkg/schedule/placement/rule_list.go#L102) is handling rules in the same range which requires building actual rules for each range. We do not need to do it for simple. We can just check there is no overlap in non-override leader rules

lcwangchao (approved):

wjhuang2016 (approved):

Issue #56523, opened 2024-10-10
Implement the "put-and-verify" lock for external storage.


YuJuncen:
### Background 

Sometimes, we need to control concurrency access to the same backup archive, like:

- When compacting / restoring, we want to block other clients from migrating the backup storage to a new version. 
- When [migrating](https://github.com/pingcap/tidb/issues/55661) the backup storage to a new version, we want to forbid reading.
- When truncating the backup storage, we don't want another truncating operation join.
- When backing up, we don't want another backup join.

But external storage locking isn't trivial. Simply putting a lock file isn't safe enough: because after checking there isn't such a lock file, another one may write it immediately. Object locks provide stronger consistency, but also require extra configuration and permissions.

We can use a "put-and-verify" lock to lock the storage, in detail, a locker need to implement this interface:

```go
type PutAndVerify interface {
    Prefix() string
    Content() io.Reader
    Verify(s ExternalStorage) error
}
```

Then we lock the storage by:

- Put a intention file with a random suffix to the `Prefix()`.
- Check if there is another file in the `Prefix()`.
- If there is, remove our intention file and back off or report error to caller.
- If there isn't, and `Verify()` returns no error, put the `Content()` to the `Prefix()`, without suffix, then remove the intention file.

D3Hunter:
I'm not that familiar with BR, can you please clearify some description to help me understand

> When compacting / restoring, we want to block migrating to a new version.

migrating WHO to a new version? you mean converting a backup stored in S3 to format of another version?

> When migrating to a new version, we want to forbid reading.

you mean read/write S3 backup at same time?

> When truncating the storage, we don't want another trancating operation join.

the storage is the external storage, such as s3 right?

YuJuncen:
@D3Hunter The "storage" should be "backup storage", i.e. the set of files generated by log backup. Let me clarify this. "Migration" is an atomic operation that applies to the "stabilized" backup storage (a brief reference [here](https://github.com/pingcap/tidb/issues/55661), after applying a "migration", the "stabilized" backup storage goes to a new version.

D3Hunter:
redefine `migration`, confusing😓

PR #56597, opened 2024-10-12
external_storage: implement locking

YuJuncen:
### What problem does this PR solve?

Issue Number: close #56523

Problem Summary:

See the issue. We need a safer method to lock the external storage.

### What changed and how does it work?

This PR implemented the "put and verify" txn, also a document was provided.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #57591, opened 2024-11-21
br: fix flaky test `TestConcurrentLock`

YuJuncen:

Issue Number: ref https://github.com/pingcap/tidb/issues/56523

close https://github.com/pingcap/tidb/issues/57755

Problem Summary:
The test case `TestConcurrentLock` isn't stable enough.

Make it more stable by stronger syncing.
```console
# go test -count 2000 -run '^TestConcurrentLock$' github.com/pingcap/tidb/br/pkg/storage
ok      github.com/pingcap/tidb/br/pkg/storage  9.397s
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > It is a test.

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/storage/local.go
+++ b/br/pkg/storage/local.go
@@ -262,7 +262,9 @@ func (l *LocalStorage) Rename(_ context.Context, oldFileName, newFileName string
 }
 
 // Close implements ExternalStorage interface.
-func (*LocalStorage) Close() {}
+func (l *LocalStorage) Close() {
+	_ = l.baseFD.Close()
+}
 
 func pathExists(_path string) (bool, error) {
 	_, err := os.Stat(_path)
--- a/br/pkg/storage/local.go
+++ b/br/pkg/storage/local.go
@@ -124,17 +124,6 @@ func (l *LocalStorage) WalkDir(_ context.Context, opt *WalkOption, fn func(strin
 	}
 	base := filepath.Join(l.base, opt.SubDir)
 	return filepath.WalkDir(base, func(path string, f os.DirEntry, err error) error {
-		if os.IsNotExist(err) {
-			if !opt.IncludeTombstone {
-				return nil
-			}
-			// We should return to the caller relative path.
-			path, _ = filepath.Rel(l.base, path)
-			if !strings.HasPrefix(path, opt.ObjPrefix) {
-				return nil
-			}
-			return fn(path, TombstoneSize)
-		}
 		if err != nil {
 			return errors.Trace(err)
 		}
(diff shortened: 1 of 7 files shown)


hawkingrei (approved):

Leavrth (approved):

br/pkg/storage/local.go
```diff
@@ -95,6 +96,11 @@ func (l *LocalStorage) WriteFile(_ context.Context, name string, data []byte) er
 	if err := os.Rename(tmpPath, filepath.Join(l.base, name)); err != nil {
 		return errors.Trace(err)
```
lance6716, line 97: comment for `WriteFile`: maybe check the `name` does not contain path separator. Because it will cause the file to be created outside of the folder, and sync the folder inode is not enough

lance6716 (approved):
/hold

free to unhold

lance6716:
BTW I doubt what's the root cause of the test. Because unit test is run in 1 process, even if we open the same path twice for 2 FD, they still share the same underlying inode? No need to sync it to disk.

br/pkg/storage/local.go
YuJuncen: Perhaps we can sync the dirpath. 🤔

YuJuncen:
> BTW I doubt what's the root cause of the test. Because unit test is run in 1 process, even if we open the same path twice for 2 FD, they still share the same underlying inode? No need to sync it to disk.

In fact the main reason of the failure is that we removed the failpoint too fast, which may cause the two participants both failed to commit.

Syncing the inode is needed because not sure why, we may read steal dir content after creating a file in it. Perhaps just opening th e basedir will be enough, but as you commented, that doesn't work with subdirs.

lance6716:
/hold

(I'll post a long comment)

lance6716 (changes requested):
I think the flaky test reflects a design problem, where as I pointed out in https://github.com/pingcap/tidb/pull/56597#discussion_r1797584077

In design doc, you can also see an example of two clients successfully write their intent, and both failed to commit because see the other one afterward.

Search for "An example of the txn aborting, when there are two conflicting txns, the name of intent files are simplified for reading" in https://github.com/pingcap/tidb/blob/master/docs/design/2024-10-11-put-and-verify-transactions-for-external-storages.md

The error message in #57755 shows one error is "there is conflict file test.lock.INTENT.b310aa1ae9a9453392771608458f4ecc during checking whether there are other intentions" And I guess the other error is also like this. This proves the existence of above root cause. And please reconsider your design about scalability.

Add fsync just hide the design problem rather than solve it. Please delete it.

YuJuncen:
> The error message in https://github.com/pingcap/tidb/issues/57755 shows one error is "there is conflict file test.lock.INTENT.b310aa1ae9a9453392771608458f4ecc during checking whether there are other intentions"

@lance6716 The `sync` call isn't used for resolve this. The race condition you described is solved by syncing with cannels. (see the modifications in the `locking_test.go`)

The problem is, without the `sync` call, the two callers may **both** successfully get the lock, one of them may not see the intention file written by another. This happen in my Linux(`5.15.0-89-generic`) with ext4 file system.

lance6716 (approved):
please fix CI

YuJuncen:
```
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGBUS (0x7) at pc=0x00007f436710f191, pid=14, tid=57
```

It seems Jenkins encounters some problems...

Leavrth (approved):

PR #57629, opened 2024-11-22
planner: fix that vector index output empty result when pk is non-int type

### LINKED ISSUE #57627: vector index on clutstered non-int pk table might output empty result
## Bug Report

```sql
CREATE TABLE `chunks_1_backup` (
  `id` char(32) NOT NULL,
  `hash` varchar(64) NOT NULL,
  `text` text DEFAULT NULL,
  `meta` json DEFAULT NULL,
  `embedding` vector(1536) DEFAULT NULL COMMENT 'hnsw(distance=cosine)',
  `document_id` int DEFAULT NULL,
  `relations` json DEFAULT NULL,
  `source_uri` varchar(512) DEFAULT NULL,
  `index_status` enum('NOT_STARTED','PENDING','RUNNING','COMPLETED','FAILED') NOT NULL,
  `index_result` text DEFAULT NULL,
  `created_at` datetime DEFAULT CURRENT_TIMESTAMP,
  `updated_at` datetime DEFAULT CURRENT_TIMESTAMP,
  PRIMARY KEY (`id`) /*T![clustered_index] CLUSTERED */,
  KEY `fk_1` (`document_id`),
  KEY `ix_chunks_1_id` (`id`),
  VECTOR INDEX `vec_idx_embedding`((VEC_COSINE_DISTANCE(`embedding`)))
```

Insert some data.

Run SQL like `select ... from chunks_1_backup order by VEC_COSINE_DISTANCE(embedding, const) limit 10`

Should output valid result.

empty result.

current master

winoros:

Issue Number: close #57627 

Problem Summary:

When pk is clustered and non-int, the full range should be `[MinNotNull, MaxValue]`. But it's `[MinInt64, MaxInt64]` currently. So it will output an empty result.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/find_best_task.go
```diff
@@ -761,6 +761,19 @@ func compareCandidates(sctx base.PlanContext, prop *property.PhysicalProperty, l
...
+		}
+		if ds.TableInfo.Columns[path.Index.Columns[0].Offset].ID != prop.VectorProp.Column.ID {
+			return false
+		}
+
+		if model.IndexableFnNameToDistanceMetric[prop.VectorProp.DistanceFnName.L] != path.Index.VectorInfo.DistanceMetric {
+			return false
+		}
+		return true
+	}
```
winoros, line 776: I changed it to the highest priority.
Because when for the vector index path, the `IsIntHandlePath` can also be true. We need to use a new way to check the real pk path.

pkg/planner/core/planbuilder.go
```diff
@@ -1191,15 +1191,19 @@ func getPossibleAccessPaths(ctx base.PlanContext, tableHints *hint.PlanHints, in
...
 				}
 			}
-			path := &util.AccessPath{Index: index}
 			if index.VectorInfo != nil {
 				// Because the value of `TiFlashReplica.Available` changes as the user modify replica, it is not ideal if the state of index changes accordingly.
 				// So the current way to use the vector indexes is to require the TiFlash Replica to be available.
 				if !tblInfo.TiFlashReplica.Available {
 					continue
 				}
+				path := genTiFlashPath(tblInfo)
```
zimulala, line 1200: I feel that the original logic is different here set up "tablePath. IsCommonHandlePath = true". What does this setting do?  Why is the original logic not needed？

winoros: A normal index will create the index's ranges.
But the vector index still uses the table range. So we need to correctly set the pk type. So it can generate the correct table range.
For int pk, it will use `[MinInt64, MaxInt64`] as full range. But for other cases, `[MinNotNull, MaxValue]` will be full range.

AilinKid: image
i think if we don't set it here, we will step into here right? which will derive a int range for common handle tiflash path

pkg/planner/core/logical_plan_builder.go
```diff
@@ -5502,7 +5502,10 @@ func (b *PlanBuilder) buildUpdate(ctx context.Context, update *ast.UpdateStmt) (
 		if dbName == "" {
 			dbName = b.ctx.GetSessionVars().CurrentDB
 		}
-		b.visitInfo = appendVisitInfo(b.visitInfo, mysql.SelectPriv, dbName, t.Name.L, "", nil)
+		// Avoid adding CTE table to the SELECT privilege list, maybe we have better way to do this?
+		if _, ok := b.nameMapCTE[t.Name.L]; !ok {
+			b.visitInfo = appendVisitInfo(b.visitInfo, mysql.SelectPriv, dbName, t.Name.L, "", nil)
+		}
```
AilinKid, line 5508: why should we need this for? any direct effect from the issue?

AilinKid (approved):

breezewish (approved):

PR #57712, opened 2024-11-26
statistics: rightly deal with timout when to send sync load 

### LINKED ISSUE #57710: blocking sync load take too much time.
## Bug Report

In order to reproduce the problem easily locally, it is recommended to reduce the configuration file parameter stats-load-queue-size and the system variable tidb_stats_load_sync_wait. It is easier to reproduce the problem if you adjust them to 10 and 1 respectively.
You only need to trigger a large number of fast stats sync load tasks to trigger this problem. I have 2,000 tables locally, each with 20 columns and 20 indexes. Using a script to quickly access all columns of all tables can trigger the problem. The SQL example is as follows.

```
SELECT * FROM %s WHERE c1 = 1 AND c2 = 2 AND c3 = 3 AND c4 = 4 AND c5 = 5 AND c6 = 6 AND c7 = 7 AND c8 = 8 AND c9 = 9 AND c10 = 10 AND c11 = 1 AND c12 = 2 AND c13 = 3 AND c14 = 4 AND c15 = 5 AND c16 = 6 AND c17 = 7 AND c18 = 8 AND c19 = 9 AND c20 = 10
```

no any time

Image

v8.1.0

hawkingrei:

Issue Number: close #57710 

Problem Summary:

in the origin code, we can not check the timeout after sending the sync load. so it will block the sending sync load and take too much time.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
rightly deal with timout when to send sync load

正确处理发送 sync load 的超时
```

### DIFF
--- a/pkg/statistics/handle/syncload/stats_syncload.go
+++ b/pkg/statistics/handle/syncload/stats_syncload.go
@@ -44,7 +44,7 @@ import (
 )
 
 // RetryCount is the max retry count for a sync load task.
-const RetryCount = 3
+const RetryCount = 2
 
 // GetSyncLoadConcurrencyByCPU returns the concurrency of sync load by CPU.
 func GetSyncLoadConcurrencyByCPU() int {
@@ -105,7 +105,7 @@ func (s *statsSyncLoad) SendLoadRequests(sc *stmtctx.StatementContext, neededHis
 		localItem := item
 		resultCh := globalStatsSyncLoadSingleFlight.DoChan(localItem.Key(), func() (any, error) {
 			timer := time.NewTimer(timeout)
-			defer timer.Stop()
+			finalTimer := time.NewTimer(timeout * (RetryCount + 1))
 			task := &statstypes.NeededItemTask{
 				Item:      localItem,
 				ToTimeout: time.Now().Local().Add(timeout),
@@ -115,7 +115,7 @@ func (s *statsSyncLoad) SendLoadRequests(sc *stmtctx.StatementContext, neededHis
 			case s.StatsLoad.NeededItemsCh <- task:
 				metrics.SyncLoadDedupCounter.Inc()
 				select {
-				case <-timer.C:
+				case <-finalTimer.C:
 					return nil, errors.New("sync load took too long to return")
 				case result, ok := <-task.ResultCh:
 					intest.Assert(ok, "task.ResultCh cannot be closed")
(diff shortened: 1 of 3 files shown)


winoros (approved):

pkg/statistics/handle/syncload/stats_syncload.go
```diff
@@ -114,9 +114,13 @@ func (s *statsSyncLoad) SendLoadRequests(sc *stmtctx.StatementContext, neededHis
 			select {
 			case s.StatsLoad.NeededItemsCh <- task:
 				metrics.SyncLoadDedupCounter.Inc()
-				result, ok := <-task.ResultCh
-				intest.Assert(ok, "task.ResultCh cannot be closed")
-				return result, nil
+				select {
+				case <-timer.C:
```
qw4990, line 118: It seems weird that both L118 and L124 are listening to `timer.C`.

hawkingrei: Another part is that this is within singleflight, so it should exit promptly. Otherwise, it will keep accumulating, and others cannot trigger a sync load.

pkg/statistics/handle/syncload/stats_syncload.go
```diff
@@ -105,7 +105,7 @@ func (s *statsSyncLoad) SendLoadRequests(sc *stmtctx.StatementContext, neededHis
 		localItem := item
 		resultCh := globalStatsSyncLoadSingleFlight.DoChan(localItem.Key(), func() (any, error) {
 			timer := time.NewTimer(timeout)
-			defer timer.Stop()
```
hawkingrei, line 108: Unstopped timers and tickers that are no longer referenced can be garbage collected. Before Go 1.23, unstopped timers could not be garbage collected until the timer went off, and unstopped tickers could never be garbage collected. The Go 1.23 implementation avoids resource leaks in programs that don’t use t.Stop.

pkg/statistics/handle/syncload/stats_syncload.go
hawkingrei: I gave it some thought and decided to add a timer to determine whether the task is genuinely stuck. Additionally, I reduced the number of retries since sync load tasks are often triggered by queries, making excessive retries less meaningful.

pkg/statistics/handle/syncload/stats_syncload.go
```diff
@@ -105,7 +105,11 @@ func (s *statsSyncLoad) SendLoadRequests(sc *stmtctx.StatementContext, neededHis
...
```
qw4990, line 108: Better to use "time.After"?

qw4990 (approved):

pkg/statistics/handle/syncload/stats_syncload.go
hawkingrei: We can use it late. but this PR need to cherrypick into other version. It’s best to keep the changes minimal.

PR #56406, opened 2024-09-29
ddl: skip adding index reorg work for empty tables

tangenta:
### What problem does this PR solve?

Issue Number: close #54230, close #51421

Problem Summary:

We can reduce the overhead of data reorganization or distributed framework scheduling if the target table is empty.

### What changed and how does it work?

See code changes.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)
  Before this PR (empty tables):
  ```sql
  mysql> create index idx_listid on tt1(id,listid);
  Query OK, 0 rows affected (2.52 sec)
  
  mysql> create index idx_listid on tt2(listid);
  Query OK, 0 rows affected (1.52 sec)
  ```
  After this PR (empty tables):
  ```sql
  mysql> create index idx_listid on tt1(id,listid);
  Query OK, 0 rows affected (0.34 sec)
  
  mysql> create index idx_listid on tt2(listid);
  Query OK, 0 rows affected (0.29 sec)
  ```
  Before this PR (non-empty tables):
  ```sql
  mysql> create index idx_listid on tt1(id,listid);
  Query OK, 0 rows affected (2.49 sec)
  
  mysql> create index idx_listid on tt2(listid);
  Query OK, 0 rows affected (2.16 sec)
  ```
  After this PR (non-empty tables):
  ```sql
  mysql> create index idx_listid on tt1(id,listid);
  Query OK, 0 rows affected (2.57 sec)
  
  mysql> create index idx_listid on tt2(listid);
  Query OK, 0 rows affected (2.20 sec)
  ```
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #57778, opened 2024-11-28
ddl: fix ExistsTableRow and add tests for skip reorg checks

### LINKED ISSUE #57769: server: using of timestamps not allocated by PD for reads can break the cluster's linearizability constraints
## Bug Report

```
tiup playground nightly --tiflash 0 --kv 1
sysbench oltp_insert --config-file=sysbench.conf --tables=1 --table-size=10000 --threads=8 --auto-inc=false prepare
sysbench oltp_insert --config-file=sysbench.conf --tables=1 --table-size=10000 --threads=8 --auto-inc=false run
```

The async commit and 1pc protocol should work as expected.

Many tidb logs like 
```
[2024/11/27 17:39:11.292 +08:00] [WARN] [prewrite.go:382] ["1pc failed and fallbacks to normal commit procedure"] [conn=1189085216] [session_alias=] [startTS=454217000877293798]  
```
tikv logs like
```
[2024/11/27 17:51:03.217 +08:00] [WARN] [[prewrite.rs:833](http://prewrite.rs:833/)] ["commit_ts is too large, fallback to normal 2PC"] [lock="Lock { lock_type: Put, primary_key: 7480000000000000715F6980000000000000010380000000000013B6037FFFFFFF9C28B6B1, start_ts: TimeStamp(454217187510452356), ttl: 3000, short_value: 30, for_update_ts: TimeStamp(0), txn_size: 2, min_commit_ts: TimeStamp(454217187510452357), use_async_commit: true, secondaries: [7480000000000000715F727FFFFFFF9C28B6B1], rollback_ts: [], last_change: Unknown, txn_source: 0, is_locked_with_conflict: false, generation: 0 }"] [max_commit_ts=454217188034740356] [min_commit_ts=9223372036854775808] [start_ts=454217187510452356] [key=7480000000000000FF715F698000000000FF0000010380000000FF000013B6037FFFFFFFFF9C28B6B1000000FC] [thread_id=47]
```

```
min_commit_ts=9223372036854775808
```
and the `max_ts` in TiKV is pushed to an **invalid** value.

The cause is the PR(https://github.com/pingcap/tidb/pull/56406/files#diff-45ad039a8b87835b475a52615355c444538d3a1f31a84d7b3ac94e15b1ce8721R1159), which uses the user specified timestamp `math.MaxInt64` to read directly with DAG interfaces.

**Using non-transactional interfaces along with self-constructed timestamps for reads and writes is a very dangerous thing, it could break the linearzibility of the whole TiDB cluster.**
Please ensure that data reads are performed using transaction interfaces or internal SQL statements to avoid manually constructing requests and specifying timestamps.

For tidb and tikv, it's necessary to enhance checks for illegal timestamps. For cases where the timestamp might exceed the current maximum PD-allocated timestamp, checks and error reporting should be enhanced.

release-8.5 branch with `8e37df4224c1990a2c8e87646c75da01ea643b4b`

tangenta:

Issue Number: close #57769

Problem Summary:

> The cause is the PR(https://github.com/pingcap/tidb/pull/56406/files#diff-45ad039a8b87835b475a52615355c444538d3a1f31a84d7b3ac94e15b1ce8721R1159), which uses the user specified timestamp math.MaxInt64 to read directly with DAG interfaces.

Use the transaction key-value interface instead to check if a table is empty.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/reorg.go
+++ b/pkg/ddl/reorg.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
-	"math"
 	"math/rand"
 	"strconv"
 	"strings"
@@ -770,9 +769,9 @@ func GetTableMaxHandle(ctx *ReorgContext, store kv.Storage, startTS uint64, tbl 
 
 // existsTableRow checks if there is at least one row in the specified table.
 // In case of an error during the operation, it returns false along with the error.
-func existsTableRow(ctx *ReorgContext, store kv.Storage, tbl table.PhysicalTable) (bool, error) {
+func existsTableRow(ctx *ReorgContext, store kv.Storage, tbl table.PhysicalTable, startTS uint64) (bool, error) {
 	found := false
-	err := iterateSnapshotKeys(ctx, store, kv.PriorityLow, tbl.RecordPrefix(), math.MaxUint64, nil, nil,
+	err := iterateSnapshotKeys(ctx, store, kv.PriorityLow, tbl.RecordPrefix(), startTS, nil, nil,
 		func(_ kv.Handle, _ kv.Key, _ []byte) (bool, error) {
 			found = true
 			return false, nil
--- a/pkg/ddl/reorg.go
+++ b/pkg/ddl/reorg.go
@@ -773,7 +773,7 @@ func GetTableMaxHandle(ctx *ReorgContext, store kv.Storage, startTS uint64, tbl 
 func existsTableRow(ctx *ReorgContext, store kv.Storage, tbl table.PhysicalTable) (bool, error) {
 	found := false
 	err := iterateSnapshotKeys(ctx, store, kv.PriorityLow, tbl.RecordPrefix(), math.MaxUint64, nil, nil,
-		func(_ kv.Handle, rowKey kv.Key, _ []byte) (bool, error) {
+		func(_ kv.Handle, _ kv.Key, _ []byte) (bool, error) {
 			found = true
 			return false, nil
 		})
--- a/pkg/ddl/reorg.go
+++ b/pkg/ddl/reorg.go
@@ -770,19 +770,16 @@ func GetTableMaxHandle(ctx *ReorgContext, store kv.Storage, startTS uint64, tbl 
 // ExistsTableRow checks if there is at least one row in the specified table.
 // In case of an error during the operation, it returns false along with the error.
 func ExistsTableRow(ctx *ReorgContext, store kv.Storage, startTS uint64, tbl table.PhysicalTable) (bool, error) {
-	handleCols := buildHandleCols(tbl)
-	result, err := buildOneRowTableScan(ctx, store, startTS, tbl, handleCols, 1, false)
+	found := false
+	err := iterateSnapshotKeys(ctx, store, kv.PriorityLow, tbl.RecordPrefix(), startTS, nil, nil,
+		func(_ kv.Handle, rowKey kv.Key, _ []byte) (bool, error) {
+			found = true
+			return false, nil
+		})
 	if err != nil {
 		return false, errors.Trace(err)
 	}
-	defer terror.Call(result.Close)
-
-	chk := chunk.New(getColumnsTypes(handleCols), 1, 1)
-	err = result.Next(ctx.ddlJobCtx, chk)
-	if err != nil {
-		return false, errors.Trace(err)
-	}
-	return chk.NumRows() != 0, nil
+	return found, nil
 }
 
 func buildHandleCols(tbl table.PhysicalTable) []*model.ColumnInfo {
--- a/pkg/ddl/reorg.go
+++ b/pkg/ddl/reorg.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"math"
 	"math/rand"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


pkg/ddl/reorg.go
```diff
@@ -767,22 +768,19 @@ func GetTableMaxHandle(ctx *ReorgContext, store kv.Storage, startTS uint64, tbl
...
-	if err != nil {
-		return false, errors.Trace(err)
-	}
-	defer terror.Call(result.Close)
-
-	chk := chunk.New(getColumnsTypes(handleCols), 1, 1)
-	err = result.Next(ctx.ddlJobCtx, chk)
+func existsTableRow(ctx *ReorgContext, store kv.Storage, tbl table.PhysicalTable) (bool, error) {
+	found := false
+	err := iterateSnapshotKeys(ctx, store, kv.PriorityLow, tbl.RecordPrefix(), math.MaxUint64, nil, nil,
```
cfzjywxk, line 775: A safer way is to fetch and use a ts from pd like `pdCli.GetTS(ctx)`. `math.MaxUint64` cannot provide snapshot semantics.

pkg/ddl/tests/indexmerge/merge_test.go
```diff
@@ -857,3 +857,81 @@ func TestAddUniqueIndexFalsePositiveDuplicate(t *testing.T) {
...
+	tk.MustQuery("select * from t;").Check(testkit.Rows("1", "2"))
+	tk.MustExec("admin check table t;")
+}
+
+func TestAddIndexInsertAfterReorgSkipCheck(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("create table t (a int);")
+	err := failpoint.EnableCall("github.com/pingcap/tidb/pkg/ddl/afterCheckTableReorgCanSkip", func() {
```
D3Hunter, line 913: can use `testfailpoint` pkg

D3Hunter (commented):
rest lgtm except existing comments

crazycs520 (approved):

pkg/ddl/reorg.go
tangenta: I think we don't need snapshot semantics here. This function determines if we can use the fast path for data reorganization:

- if it returns false, but actually it is true, we won't enter the fast path
- if it returns true, but actually it is false, we enter the fast path, and the following insertions will be handled by dual-write mechanism (txn mode) or temp-index-merge mechanism (ingest mode). This is covered by test https://github.com/pingcap/tidb/pull/57778/files#diff-2998ac87ef4adb20bfc6bf8ce58ace7015800e57f330bacad08cba4b90350300R908

cfzjywxk: Another layer of semantics for safety is that, in the transaction layer and TiKV, the use of `math.MaxUint64` is only expected for primary key point-get autocommit reads. Other scenarios should not use `math.MaxUint64` for reads. It's better to stick to PD-allocated timestamps instead.

MyonKeminta: > I think we don't need snapshot semantics here.

Did you mean that you don't need a consistent snapshot, so you try to use a *weaker* or *looser* way for the reading?

I think reading with MaxUint64 seems to be much more special than you, as well as many other developers, might think. At least it is after implementing async commit. Maybe it's better if you only consider whether the overhead of getting a new ts from PD is so unacceptable to you.

D3Hunter: > reading with MaxUint64 seems to be much more special

can you give more detail? or some doc link?

lance6716: > the use of `math.MaxUint64` is only expected for primary key point-get autocommit reads. Other scenarios should not use `math.MaxUint64` for reads

In my understand, snapshot reading is not a pure "reading" at TiKV. It has side effect to async commit. https://github.com/pingcap/tidb/issues/57769

D3Hunter: > In my understand, snapshot reading is not a pure "reading" at TiKV. It has side effect to async commit. https://github.com/pingcap/tidb/issues/57769

that's too bad, a read operation corrupt the entire cluster

MyonKeminta (approved):

pkg/ddl/reorg.go
MyonKeminta: > can you give more detail? or some doc link?

There's not any document dedicated to talking about this as far as I know. There's many special handling for reading with MaxUint64. For example, TiKV skipping updating the max_ts using it; special lock-ignoring logic when checking read-write conflict; etc. It also used to cause difficulty for us in developent or optimization work.

> snapshot reading is not a pure "reading" at TiKV.

The max_ts in async commit / 1PC is an approach to let writing transactions to avoid affecting other reading operations. It tells the latest timestamp allocated by PD that has been used on this TiKV node. It's quite doubtful to me whether this can be the reason to say that the reading is not "pure".

> that's too bad

The design is based on the fact that the reads and writes are strongly ordered by PD's globally unique and monotonic TSO allocation, and writing cannot change the snapshot has been read, otherwise the snapshot becomes not repeat-readable, then it's no longer a snapshot. You may think that this makes async commit transactions committed to an incorrect future version and cause its result not readable, and things would be all better if async commit not exists. But that's the choice that async commit transactions made to avoid breaking your snapshot. If the transaction is committed in 2PC mode which is not related to the max_ts mechanism, there's still another form of corruption, that is, your snapshot is changed and becomes not repeat-readable. If you believe there's something bad, I think the bad part is that the strong ordering that's supposed to be provided by TSO is broken.

D3Hunter: for `bad`, i mean a `read only operation` has side effects, and it corrupt the entire cluster

`read only operation` shouldn't have such effects as a common design practice AFAIK

D3Hunter (approved):

Issue #57871, opened 2024-12-02
Use a DDL subscriber to handle updates to the stats meta


Rustin170506:
## Enhancement

We introduced a DDL notifier when developing the new priority queue to deliver DDL events in a highly durable manner.

However, the current stats meta handling still relies on an unreliable in-memory channel. Therefore, it would be better to use the DDL notifier to handle stats meta updates as well.

I tried this before, as seen in [this PR](https://github.com/pingcap/tidb/pull/56723). However, many of our test cases rely on the behavior where the unit tests do not handle DDL events by default, which caused a lot of test failures.

This time, I will keep the in-memory channel for testing purposes only. This will help reduce the effort required to fix all the broken tests.

PR #57872, opened 2024-12-02
*: use DDL subscriber updating stats meta

Rustin170506:

Issue Number: close https://github.com/pingcap/tidb/issues/57871

Problem Summary:

In this PR, we began using the DDL notifier to handle all DDL events and update the stats meta accordingly.

To maintain the old behavior, we do not register it when the stats lease is less than 1. This means that we won't handle the DDL event for unit tests at all unless you explicitly handle it manually.

In the future, we might want to handle this in unit tests by default. However, doing so could disrupt many existing test cases. For now, I prefer to maintain the old behavior. I plan to migrate to the new approach eventually, but I currently don't have enough time to update all the tests that would break.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [x] [Manual test](https://github.com/pingcap/tidb/pull/57872#issuecomment-2535350098)
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


winoros (approved):

Rustin170506:
There is a data race here. I will fix it later.

pkg/ddl/ddl.go
```diff
@@ -588,18 +588,23 @@ func asyncNotifyEvent(jobCtx *jobContext, e *notifier.SchemaChangeEvent, job *mo
...
-			default:
-				time.Sleep(time.Microsecond * 10)
+	// In test environments, we use a channel-based approach to handle DDL events.
+	// This maintains compatibility with existing test cases that expect events to be delivered through channels.
+	// In production, DDL events are handled by the notifier system instead.
+	if intest.InTest {
+		ch := jobCtx.oldDDLCtx.ddlEventCh
+		if ch != nil {
+		forLoop:
+			for i := 0; i < 10; i++ {
```
lance6716, line 598: if it's only a test, we don't need to retry? So the duration of unit test can be reduced

lance6716 (commented):
33 / 57 files viewed

pkg/statistics/handle/ddl/ddl.go
```diff
@@ -46,158 +42,18 @@ func NewDDLHandler(
 		ddlEventCh:   make(chan *notifier.SchemaChangeEvent, 1000),
 		statsWriter:  statsWriter,
 		statsHandler: statsHandler,
+		sub:          NewSubscriber(statsHandler),
```
lance6716, line 45:
```suggestion
		sub:          newSubscriber(statsHandler),
```

lance6716 (approved):
/hold

free to unhold

Issue #57815, opened 2024-11-29
implement some Oracle functions based on TiDB


joccau:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #57890, opened 2024-12-02
PingKai: implement Oracle `instr` function

joechenrh:

Issue Number: ref https://github.com/pingcap/tidb/issues/57815

Problem Summary:

Add two optional parameters, position and occurrence, to `instr` to align this function with Oracle.

- **position**: A _nonzero_ integer indicating the position where the search begins. If the value is negative, it counts backward from the end of the string and then searches backward from that position, i.e. from `-position` to `0`.
- **occurrence**: A _positive_ integer indicating which occurrence of the substring to search. The search will continue until the specified number of matches is found.

The biggest difference between current implementation and Oracle is that:
- In Oracle, if the third and forth arguments can't be converted to `number` type, it will throw an error, like `instr("aaaa", "aa", 1.1, "2.2f")`.
- However, in TiDB, truncate error will be ignored by default in select statement. So the above statement will still return the result.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/expression/builtin_string.go
```diff
@@ -3786,7 +3875,36 @@ func (b *builtinInstrUTF8Sig) evalInt(ctx EvalContext, row chunk.Row) (int64, bo
...
+			return 0, true, errors.Errorf("Occurrence should not be negative, got %d", occurrences)
+		}
+	}
+
+	var idx int
+	if position > 0 {
+		idx = findNthOccurrenceUTF8(str, substr, int(position-1), int(occurrences))
+	} else {
+		idx = findLastOccurrenceUTF8(str, substr, int(-position), int(occurrences))
+	}
```
River2000i, line 3906: why we need to search `substr` using different way? Seems if `position` is negative will only search from `str[-position]`🤔

joechenrh: Because when `position` is negative, search is start from `-position` and end at `0`, not from `-position` to `len(str) - 1`.

pkg/expression/builtin_string.go
```diff
@@ -3770,8 +3774,93 @@ func (b *builtinInstrSig) Clone() builtinFunc {
 	return newSig
 }
 
-// evalInt evals INSTR(str,substr).
+// Find nth occurrence of substring in given string.
+// Search starts from the `position` character to the last character.
+func findNthOccurrence(str, sub string, position, n int) int {
+	currentPos := position
+	for i := 0; i < n; i++ {
+		currentPos = strings.Index(str[currentPos:], sub)
```
River2000i, line 3782: seems we can add some test when `position` exceed the length of `str`?

joechenrh: Added in https://github.com/pingcap/tidb/pull/57890/commits/5d8751f9b18be67d398a5dcd2c96ce6fd2716f02

River2000i (approved):
lgtm

joccau (approved):
LGTM

PR #57919, opened 2024-12-03
ttl: fix the issue that one task losing heartbeat will block other tasks

### LINKED ISSUE #57915: One TTL task losing heartbeat will block other tasks from getting heartbeat
See the following codes:

```go
// updateHeartBeat updates the heartbeat for all tasks with current instance as owner
func (m *taskManager) updateHeartBeat(ctx context.Context, se session.Session, now time.Time) error {
	for _, task := range m.runningTasks {
		state := &cache.TTLTaskState{
			TotalRows:   task.statistics.TotalRows.Load(),
			SuccessRows: task.statistics.SuccessRows.Load(),
			ErrorRows:   task.statistics.ErrorRows.Load(),
		}
		if task.result != nil && task.result.err != nil {
			state.ScanTaskErr = task.result.err.Error()
		}

		intest.Assert(se.GetSessionVars().Location().String() == now.Location().String())
		sql, args, err := updateTTLTaskHeartBeatSQL(task.JobID, task.ScanID, now, state, m.id)
		if err != nil {
			return err
		}
		_, err = se.ExecuteSQL(ctx, sql, args...)
		if err != nil {
			return errors.Wrapf(err, "execute sql: %s", sql)
		}

		if se.GetSessionVars().StmtCtx.AffectedRows() != 1 {
			return errors.Errorf("fail to update task status, maybe the owner is not myself (%s), affected rows: %d",
				m.id, se.GetSessionVars().StmtCtx.AffectedRows())
		}
	}
	return nil
}
```

it'll return error when one of the tasks fail to update. However, we should log the error and continue to try the next task.

YangKeao:

Issue Number: close #57915

Problem Summary:

If one task failed to update heartbeat, the other tasks will not be tried. It's not expected.

Print an error and continue to try other tasks.

This PR also adapts the same change for job.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ttl/ttlworker/job_manager.go
+++ b/pkg/ttl/ttlworker/job_manager.go
@@ -226,10 +226,7 @@ func (m *JobManager) jobLoop() error {
 		// Job Schedule loop:
 		case <-updateJobHeartBeatTicker:
 			updateHeartBeatCtx, cancel := context.WithTimeout(m.ctx, ttlInternalSQLTimeout)
-			err = m.updateHeartBeat(updateHeartBeatCtx, se, now)
-			if err != nil {
-				logutil.Logger(m.ctx).Warn("fail to update job heart beat", zap.Error(err))
-			}
+			m.updateHeartBeat(updateHeartBeatCtx, se, now)
 			cancel()
 		case <-jobCheckTicker:
 			m.checkFinishedJob(se)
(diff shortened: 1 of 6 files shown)


lcwangchao (approved):

zimulala (approved):
LGTM

PR #57954, opened 2024-12-03
statistics: limit the batch size when updating the stats cache

### LINKED ISSUE #57953: Update the stats cache batch by batch instead of updating once in periodic statistics update worker
## Enhancement
Image
In our big test cluster, many tables may be updated in one single round updating.
Image
Currently, we only call the `UpdateStatsCache` only once when we get all the tables finished.

It's not very efficient and may cause inconsistency issues since the interval between getting a table from the cache and pushing the table back to the cache may be quite long.

winoros:

Issue Number: close #57953 

Problem Summary:

Add a middle layer to submit the update cache job more quickly.
Also memory is saved.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/statistics/handle/cache/statscache.go
```diff
@@ -65,6 +65,47 @@ func NewStatsCacheImplForTest() (types.StatsCache, error) {
...
+
+func (t *tablesToUpdateOrDelete) addToDelete(statsCache *StatsCacheImpl, tableID int64) {
+	if len(t.toDelete) == 10 {
+		statsCache.UpdateStatsCache(types.CacheUpdate{
+			Updated: t.toUpdate,
+			Deleted: t.toDelete,
+			Options: t.option,
+		})
+		t.toUpdate = t.toUpdate[:0]
+		t.toDelete = t.toDelete[:0]
```
hawkingrei, line 95: clear(t.ToDelete)

winoros: The clear seems to be reset to zero instead of reset the length.
```
// The clear built-in function clears maps and slices.
// For maps, clear deletes all entries, resulting in an empty map.
// For slices, clear sets all elements up to the length of the slice
// to the zero value of the respective element type. If the argument
// type is a type parameter, the type parameter's type set must
// contain only map or slice types, and clear performs the operation
// implied by the type argument.
```

Rustin170506 (commented):
Thansk!

Rustin170506 (approved):
Thanks! :shipit:

I think adding tests for it is still valuable. But I think it shouldn't block the PR.

hawkingrei (approved):

Issue #11329, opened 2019-07-19
DATETIME's behaviour is different from mysql.


wshwsh12:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.

```
select DATE_ADD('1000-01-01 00:00:00',INTERVAL -2 HOUR);
select DATE_ADD('1000-01-01 00:00:00',INTERVAL -200 HOUR);
select DATE_ADD('0001-01-01 00:00:00',INTERVAL -2 HOUR);
select DATE_ADD('0001-01-01 00:00:00',INTERVAL -25 HOUR);
select DATE_ADD('0001-01-01 00:00:00',INTERVAL -8784 HOUR);
select DATE_ADD('0001-01-01 00:00:00',INTERVAL -8785 HOUR);
```

2. What did you expect to see?

```
mysql> select DATE_ADD('1000-01-01 00:00:00',INTERVAL -2 HOUR);
+--------------------------------------------------+
| DATE_ADD('1000-01-01 00:00:00',INTERVAL -2 HOUR) |
+--------------------------------------------------+
| 0999-12-31 22:00:00                              |
+--------------------------------------------------+
1 row in set (0.00 sec)

mysql> select DATE_ADD('1000-01-01 00:00:00',INTERVAL -200 HOUR);
+----------------------------------------------------+
| DATE_ADD('1000-01-01 00:00:00',INTERVAL -200 HOUR) |
+----------------------------------------------------+
| 0999-12-23 16:00:00                                |
+----------------------------------------------------+
1 row in set (0.00 sec)

mysql> select DATE_ADD('0001-01-01 00:00:00',INTERVAL -2 HOUR);
+--------------------------------------------------+
| DATE_ADD('0001-01-01 00:00:00',INTERVAL -2 HOUR) |
+--------------------------------------------------+
| 0000-00-00 22:00:00                              |
+--------------------------------------------------+
1 row in set (0.00 sec)

mysql> select DATE_ADD('0001-01-01 00:00:00',INTERVAL -25 HOUR);
+---------------------------------------------------+
| DATE_ADD('0001-01-01 00:00:00',INTERVAL -25 HOUR) |
+---------------------------------------------------+
| 0000-00-00 23:00:00                               |
+---------------------------------------------------+
1 row in set (0.00 sec)

mysql> select DATE_ADD('0001-01-01 00:00:00',INTERVAL -8784 HOUR);
[...]

PR #58003, opened 2024-12-05
Oracle-compatibility: add_month and next_day function

OliverS929:

Issue Number: ref https://github.com/pingcap/tidb/issues/57815

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/expression/builtin_time.go
```diff
@@ -809,6 +811,166 @@ func (b *builtinLastMonthSig) evalString(ctx EvalContext, row chunk.Row) (string
...
+func (b *builtinNextDaySig) Clone() builtinFunc {
+	newSig := &builtinNextDaySig{}
+	newSig.cloneFrom(&b.baseBuiltinFunc)
+	return newSig
+}
+
+// evalTime evals a builtinNextDaySig.
+func (b *builtinNextDaySig) evalTime(ctx EvalContext, row chunk.Row) (types.Time, bool, error) {
+	// need a map from string to actual weekdays
+	dayMap := map[string]time.Weekday{
```
joechenrh, line 918: It can be declared as a global variable

pkg/expression/builtin_time.go
```diff
@@ -809,6 +811,166 @@ func (b *builtinLastMonthSig) evalString(ctx EvalContext, row chunk.Row) (string
...
+	resultTime, err := types.AddDate(0, 0, daysToAdd, orgTime)
+	if err != nil {
+		return types.ZeroTime, true, handleInvalidTimeError(ctx, types.ErrDatetimeFunctionOverflow.GenWithStackByArgs("datetime"))
+	}
+
+	// Ensure time is 00:00:00
+	resultTime.Truncate(24 * time.Hour)
+
+	// in case of datetime overflow
+	if resultTime.Year() == 0 {
```
joechenrh, line 962: If the input date is valid, then the check of `resultTime.Year()` is redundant?

OliverS929: the `addData` called above checked for upper bound of `DateTime`, but not the lowerbound. Here is to check and see if it is still above minimum value so that issue like https://github.com/pingcap/tidb/issues/11329 does not happen.

pkg/expression/builtin_time.go
OliverS929: I think `dayMap` would be better as a local variable here. Currently, I don’t see any other usage for this weekday mapping. If it’s needed elsewhere in the future, we could create a more generalized version and potentially include it in coreTime. For now, `dayMap` also serves as an input check, as the `next_day` function only allows these two types of input, which may not necessarily apply to other functions.

joechenrh (approved):

joccau (approved):
LGTM

PR #58029, opened 2024-12-05
planner: fix index range intersection for in-list and other predicates 

### LINKED ISSUE #57694: further improvement of pagination query
customer has a query similar to https://github.com/pingcap/tidb/issues/56005, but more complicated.  In #56005, the query is:

```sql
explain SELECT
  dt.*
FROM
  it 
  LEFT JOIN dt ON it.pk = dt.pk
WHERE
  it.a = "a"
  AND (
    (
      it.a > "a"
    )
    OR (
      it.a = "a" AND it.pk > 1
    )
  )
ORDER BY
  it.pk
LIMIT
  240;
```
Now the query is 
```sql
explain SELECT
  dt.*
FROM
  it 
  LEFT JOIN dt ON it.pk = dt.pk
WHERE
  it.a in ("a","b")
  AND (
    (
      it.a > "b"
    )
    OR (
      it.a = "b" AND it.pk > 1
    )
  )
ORDER BY
  it.pk
LIMIT
  240;
```
The optimizer should see that there is no need to scan index for it.a = "a".

ghazalfamilyusa:

Issue Number: close #57694

### Problem Summary:
This is a limitation for optimal multi-column index range derivation. It is similar to  https://github.com/pingcap/tidb/pull/54166 but different where point ranges are derived from IN list on the prefix column. A simple example is below where the range is based on "t1.a1 IN ( 44, 70, 76)" and not the full predicate  "t1.a1 IN ( 44, 70, 76) AND (t1.a1 > 70 OR (t1.a1 = 70 AND t1.b1 > 41))"

```
CREATE TABLE t1 (
  a1 int,
  b1 int,
  c1 varbinary(767) DEFAULT NULL,
  KEY twoColIndex (a1,b1)
)

EXPLAIN  format = brief
SELECT 1
     FROM t1 FORCE INDEX (twoColIndex)
     WHERE
       t1.a1 IN ( 44, 70, 76)
       AND (t1.a1 > 70 OR (t1.a1 = 70 AND t1.b1 > 41))
--------------

Projection      10.07   root            1->Column#5
└─IndexReader   10.10   root            index:Selection
  └─Selection   10.10   cop[tikv]               or(gt(test.t1.a1, 70), and(eq(test.t1.a1, 70), gt(test.t1.b1, 41)))
    └─IndexRangeScan    30.00   cop[tikv]       table:t1, index:twoColIndex(a1, b1)     range:[44,44], [70,70], [76,76]
```

The optimal solution is below where the range intersection of all predicates is: (70 41,70 +inf], [76,76]
```
Projection      43.33   root            1->Column#5
└─IndexReader   54.17   root            index:IndexRangeScan
  └─IndexRangeScan      54.17   cop[tikv]       table:t1, index:twoColIndex(a1, b1)     range:(70 41,70 +inf], [76,76]
```

We already have the logic of intersection through prior 54166 PR and it is computed for this case as well. The problem is that  we do not choose the intersection for this case. 

The fix is to recognize this case by comparing the conjunctive normal form, or CNF result (applying all predicates) and IN list ranges and pick the most selective. The CNF result is more selective if the covers a superset of filters. In the above example, the CNF result access conditions is "t1.a1 IN ( 44, 45) AND (t1.a1 > 70 OR (t1.a1 = 70 AND t1.b1 > 41))" which is a superset of the IN list access conditions "

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/util/ranger/detacher.go
+++ b/pkg/util/ranger/detacher.go
@@ -362,6 +362,30 @@ func unionColumnValues(lhs, rhs []*valueInfo) []*valueInfo {
 	return lhs
 }
 
+// Check which detach result is more selective. This function is called to choose between point ranges and the best CNF ranges.
+// This is needed because sometimes the best CNF has full intersection and is more selective,
+// and other times it is not when the intersection is not applied.
+func chooseBetweenRangeAndPoint(sctx *rangerctx.RangerContext, r1 *DetachRangeResult, r2 *cnfItemRangeResult) {
+	if fixcontrol.GetBoolWithDefault(sctx.OptimizerFixControl, fixcontrol.Fix54337, false) {
+		if r1 != nil && len(r1.Ranges) > 0 && r2 != nil && r2.rangeResult != nil {
+			r1Minusr2 := removeConditions(sctx.ExprCtx.GetEvalCtx(), r1.AccessConds, r2.rangeResult.AccessConds)
+			r2Minusr1 := removeConditions(sctx.ExprCtx.GetEvalCtx(), r2.rangeResult.AccessConds, r1.AccessConds)
+			// r2 is considered more selective (and more useful) than r1 if its AccessConds are a superset of r1's AccessConds.
+			// This means that r1.AccessConds minus r2.AccessConds should result in an empty set.
+			// The function `removeConditions` is used to perform this subtraction.
+			// For example, if A = {t1.a1 IN (44, 70, 76)} and B = {t1.a1 IN (44, 70, 76), (t1.a1 > 70 OR (t1.a1 = 70 AND t1.b1 > 41))},
+			// then A-B is empty and therefore B is a superset of A.
+			// Avoid the case when both r1 and r2 have the same AccessConds (r2Minusr1 is not empty).
+			if len(r1Minusr2) == 0 && len(r2Minusr1) > 0 {
+				// Update final result and just update: Ranges, AccessConds and RemainedConds
+				r1.RemainedConds = removeConditions(sctx.ExprCtx.GetEvalCtx(), r1.RemainedConds, r2.rangeResult.AccessConds)
+				r1.Ranges = r2.rangeResult.Ranges
+				r1.AccessConds = r2.rangeResult.AccessConds
+			}
+		}
+	}
+}
+
 // detachCNFCondAndBuildRangeForIndex will detach the index filters from table filters. These conditions are connected with `and`
 // It will first find the point query column and then extract the range query column.
 // considerDNF is true means it will try to extract access conditions from the DNF expressions.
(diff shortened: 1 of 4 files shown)


pkg/util/ranger/detacher.go
```diff
@@ -362,6 +362,26 @@ func unionColumnValues(lhs, rhs []*valueInfo) []*valueInfo {
...
+			// r2 covers more access conditions therefore it is more selective than r1. Make sure r1 has access condition to start with.
+			if len(r1.AccessConds) > 0 && len(r1Minusr2) == 0 && len(r2Minusr1) > 0 {
+				// Update final result and just update: Ranges, AccessConds and RemainedConds
+				r1.RemainedConds = removeConditions(sctx.ExprCtx.GetEvalCtx(), r1.RemainedConds, r2.rangeResult.AccessConds)
+				r1.Ranges = r2.rangeResult.Ranges
+				r1.AccessConds = r2.rangeResult.AccessConds
+			}
+		}
+	}
+	return r1
```
qw4990, line 382: Just a comment from the engineering perspective, it seems that we can remove the returned `*DetachRangeResult`, because we update `r1` directly in place.

pkg/util/ranger/detacher.go
```diff
@@ -362,6 +362,26 @@ func unionColumnValues(lhs, rhs []*valueInfo) []*valueInfo {
...
 
+// Check which detach result is more selective. This function is called to choose between point ranges and the best CNF ranges.
+// This is needed because sometimes the best CNF has full intersection and is more selective,
+// and other times it is not when the intersection is not applied.
+func chooseBetweenRangeAndPoint(sctx *rangerctx.RangerContext, r1 *DetachRangeResult, r2 *cnfItemRangeResult) *DetachRangeResult {
+	if fixcontrol.GetBoolWithDefault(sctx.OptimizerFixControl, fixcontrol.Fix54337, false) {
+		if r1 != nil && len(r1.Ranges) > 0 && r2 != nil && r2.rangeResult != nil {
+			r1Minusr2 := removeConditions(sctx.ExprCtx.GetEvalCtx(), r1.AccessConds, r2.rangeResult.AccessConds)
+			r2Minusr1 := removeConditions(sctx.ExprCtx.GetEvalCtx(), r2.rangeResult.AccessConds, r1.AccessConds)
+			// r2 covers more access conditions therefore it is more selective than r1. Make sure r1 has access condition to start with.
```
qw4990, line 373: Why `removeConditions` can work for this purpose? Could you add an example as comments here?

ghazalfamilyusa: The code comment says that "r2 covers more access conditions therefore it is more selective than r1. Make sure r1 has access condition to start with." and added that r1-r2 is empty implies that r2 is a superset of r1 with an example.

pkg/util/ranger/detacher.go
ghazalfamilyusa: Done

qw4990 (approved):

winoros (approved):

PR #58033, opened 2024-12-06
executor: run idxlookup workers in a pool

zyguan:

Issue Number: ref #56649

Problem Summary: IndexLookupExecutor always starts 1+indexLookupConcurrency goroutines which is expensive for common queries.

Use a goroutine pool for workers and spawn new goroutine on demand.

Let prepare sysbench dataset with `--secondary` and run point select workload.

Previously, tidb spawns 6 goroutines for each query and only two of them are used (non-concurrently).

image

With this PR, only 1 goroutine are created for both index task and table task.

Here are benchmark results.

|     workload     | threads |   pool   |   base   |  diff  |
| ---------------- | ------- | -------: | -------: | -----: |
| idxlookup_1      | 15      | 16560.51 | 16320.65 |  1.47% |
| idxlookup_1      | 30      | 26729.51 | 26146.49 |  2.23% |
| idxlookup_1      | 60      | 29680.83 | 28253.65 |  5.05% |

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


crazycs520 (approved):

pkg/executor/utils.go
```diff
@@ -119,3 +121,70 @@ func encodePassword(u *ast.UserSpec, authPlugin *extension.AuthPlugin) (string,
...
+}
+
+type gotask struct {
+	f    func()
+	next *gotask
+}
+
+type gopool struct {
+	lock sync.Mutex
+	head *gotask
```
cfzjywxk, line 136: How about using a lock free queue to store the pending task pointers, so the `Mutex` could be avoided?

pkg/executor/utils.go
```diff
@@ -119,3 +121,70 @@ func encodePassword(u *ast.UserSpec, authPlugin *extension.AuthPlugin) (string,
...
+var taskPool = sync.Pool{
+	New: func() any { return &gotask{} },
+}
+
+type gotask struct {
+	f    func()
+	next *gotask
+}
+
+type gopool struct {
```
cfzjywxk, line 134: There is some `GoroutinePool` like structures in other packages, how about renaming it to `IndexLookupWorkerPool` or `CopWorkerPool` it we do not want it to be a generic used type?

pkg/executor/utils.go
zyguan: Here gopool is mainly for concurrency control, each IndexLookupExecutor has its own pool, that is, there are only about indexLookupConcurrency goroutines access the pool at most 2xNumOfTableTasks times. I think it's acceptable, no need to introduce additional complexity.

pkg/executor/utils.go
zyguan: Actually I prefer to keep it simple and private currently. [gp](https://github.com/tiancaiamao/gp) and [pool](https://github.com/pingcap/tidb/tree/master/pkg/resourcemanager/pool) have already been used in tidb, but neither of them meet my requirement. So I just implement a tiny one here and maybe we can extract uniform GoroutinePool interface in the future.

cfzjywxk (approved):

PR #58062, opened 2024-12-06
domain,infoschema: make infoschema activity block GC safepoint advancing

### LINKED ISSUE #57952: snapshot infoschema should block GC
## Enhancement

Currently, the meta package use old transaction API, unlike normal internal transaction, those APIs can not block the GC safepoint advancing (not updating min start ts).

This may cause "GC life time is shorter than transaction duration" error for InfoSchema V2.

Maybe a possible fix is collect the min start ts of all snapshot infoschemas.

tiancaiamao:

Issue Number: close #57952

Problem Summary:

Currently, when we get a infoschema v2 instance, it's lifetime is in range [ts, ts + 10min)
If the caller hold a infoschema, and use it some time later, it might get error "GC lifetime is shorter than transaction duration" error. That's because infoschema v2 internally use meta package API, while the meta API does not block the GC safepoint advancing.

In this commit, I add a `keepAlive()`  call to the infoschema API,
that `keepAlive()` function will keep the minimal start ts of active infoschema API calls.

The infosync.Syncer call `ReportMinStartTS()` periodically, amd now it will take the active infoschema ts into consideration, which is updated by the `keepAlive()` function.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/infoschema/cache.go
+++ b/pkg/infoschema/cache.go
@@ -17,6 +17,7 @@ package infoschema
 import (
 	"sort"
 	"sync"
+	"sync/atomic"
 
 	infoschema_metrics "github.com/pingcap/tidb/pkg/infoschema/metrics"
 	"github.com/pingcap/tidb/pkg/meta/autoid"
@@ -57,6 +58,13 @@ func NewCache(r autoid.Requirement, capacity int) *InfoCache {
 		r:                   r,
 		Data:                infoData,
 	}
+}
+
+// GetAndResetRecentInfoSchemaTS provides the min start ts for infosync.InfoSyncer.
+func (h *InfoCache) GetAndResetRecentInfoSchemaTS(now uint64) uint64 {
+	ret := atomic.LoadUint64(&h.Data.recentMinTS)
+	atomic.StoreUint64(&h.Data.recentMinTS, now)
+	return ret
 }
 
 // ReSize re-size the cache.
--- a/pkg/infoschema/cache.go
+++ b/pkg/infoschema/cache.go
@@ -60,6 +60,18 @@ func NewCache(r autoid.Requirement, capacity int) *InfoCache {
 }
 
 // GetAndResetRecentInfoSchemaTS provides the min start ts for infosync.InfoSyncer.
+// It works like this:
+//
+//	There is a background infosync worker calling ReportMinStartTS() function periodically.
+//	At the beginning of each round, the Data.recentMinTS here is reset to current TS.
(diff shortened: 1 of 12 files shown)


wjhuang2016 (approved):

tiancaiamao:
One possible case that this fix can not cover is that, the caller take an infoschema V2 instance but never use it. 
Then `keepAlive()` is not called.
10min later, infoschema v2 API is called  on this `is` for the first time...  


This case is rare and if we found it we can adjust at the caller side, so it should not be a real issue.

lcwangchao (approved):

pkg/infoschema/cache.go
```diff
@@ -59,6 +60,13 @@ func NewCache(r autoid.Requirement, capacity int) *InfoCache {
 	}
 }
 
+// GetAndResetRecentInfoSchemaTS provides the min start ts for infosync.InfoSyncer.
+func (h *InfoCache) GetAndResetRecentInfoSchemaTS(now uint64) uint64 {
+	ret := atomic.LoadUint64(&h.Data.recentMinTS)
+	atomic.StoreUint64(&h.Data.recentMinTS, now)
```
MyonKeminta, line 66: Can we use `atomic.Uint64` instead?

tiancaiamao:
@MyonKeminta we need some review from the transaction team, PTAL

pkg/infoschema/cache.go
MyonKeminta: Is it possible that an `infoschemaV2` is acccessed in a lower frequence than `ReportMinStartTS`, so that a tick of `ReportMinStartTS` may miss this `infoschemaV2` instance, which allows GC to proceed, the `infoschemaV2` instance becomes unavailable again?

tiancaiamao: In theory, that is possible, so there is still potential risk. I have discussed that with @lcwangchao 
But in practice I think this fix is good enough for the most common cases. @MyonKeminta

MyonKeminta: Ok. But I suggest this details can be noted in the comments in the code.
Rest LGTM.

tiancaiamao: Done

MyonKeminta (approved):

PR #58110, opened 2024-12-09
ttl: add integration test with random fault session for TTL

### LINKED ISSUE #58107: TTL `getSession` function may return unexpected result
## Bug Report

The function `getSession` returns a session and an error. It's expected to return either `session` or an `error`. However, it's possible that both of them are `nil`, because the `Close()` function may overwrite the `err` return value.

At least one of `session` and `err` are not `nil`.

Both of them are `nil` if the session is error.

### LINKED ISSUE #58109: If a TTL job is cancelled, the `last_job_finish_time` is incorrect.
The first assert `intest.Assert(se.GetSessionVars().Location().String() == now.Location().String())` in `job.finish` is never true when `job.finish` is called from `rescheduleJobs`  (but our tests didn't cover it yet). It's because in `rescheduleJobs` the `now` is in `GlobalTimeZone` (`Asia/Shanghai` in my case), and the `se.GetSessionVars().Location().String()` is always `UTC`.

It'll not be a big issue as the job is scheduled according to the start time, but the wrong finish time is still confusing for users.

### LINKED ISSUE #58411: add fuzzy test for concurrent running TTL job manager
## Enhancement

The E2E test for TTL runs really slow. By modifying some internal variables (through failpoint), we can make it much faster. Meanwhile, we can also simulate the fault situation for TTL job managers.

Let me add such kind of tests 😃 .

YangKeao:

Issue Number: close #58411

1. Add two tests related to TTL. One of them is the test with randomly faulty session, which cannot execute the SQL.
2. Fix the bug that the result of `getSession` is unexpected in some situation.
3. Fix the bug that the `last_job_finish_time` is not in correct TZ when the job is cancelled.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ttl/ttlworker/job_manager.go
+++ b/pkg/ttl/ttlworker/job_manager.go
@@ -66,7 +66,7 @@ const taskGCTemplate = `DELETE task FROM
 const ttlJobHistoryGCTemplate = `DELETE FROM mysql.tidb_ttl_job_history WHERE create_time < CURDATE() - INTERVAL 90 DAY`
 const ttlTableStatusGCWithoutIDTemplate = `DELETE FROM mysql.tidb_ttl_table_status WHERE (current_job_status IS NULL OR current_job_owner_hb_time < %?)`
 
-const timeFormat = time.DateTime
+var timeFormat = time.DateTime
 
 func insertNewTableIntoStatusSQL(tableID int64, parentTableID int64) (string, []any) {
 	return insertNewTableIntoStatusTemplate, []any{tableID, parentTableID}
@@ -86,7 +86,7 @@ func gcTTLTableStatusGCSQL(existIDs []int64, now time.Time) (string, []any) {
 		existIDStrs = append(existIDStrs, strconv.Itoa(int(id)))
 	}
 
-	hbExpireTime := now.Add(-jobManagerLoopTickerInterval * 2)
+	hbExpireTime := now.Add(-getHeartbeatInterval() * 2)
 	args := []any{hbExpireTime.Format(timeFormat)}
 	if len(existIDStrs) > 0 {
 		return ttlTableStatusGCWithoutIDTemplate + fmt.Sprintf(` AND table_id NOT IN (%s)`, strings.Join(existIDStrs, ",")), args
@@ -137,6 +137,10 @@ func NewJobManager(id string, sessPool util.SessionPool, store kv.Storage, etcdC
 
 	manager.init(manager.jobLoop)
 	manager.ctx = logutil.WithKeyValue(manager.ctx, "ttl-worker", "job-manager")
+	if intest.InTest {
+		// in test environment, in the same log there will be multiple ttl managers, so we need to distinguish them
+		manager.ctx = logutil.WithKeyValue(manager.ctx, "ttl-worker", id)
+	}
 
 	manager.infoSchemaCache = cache.NewInfoSchemaCache(getUpdateInfoSchemaCacheInterval())
 	manager.tableStatusCache = cache.NewTableStatusCache(getUpdateTTLTableStatusCacheInterval())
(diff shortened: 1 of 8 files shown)


YangKeao:
/hold

It's not stable yet.

pkg/ttl/ttlworker/job_manager_integration_test.go
```diff
@@ -1614,3 +1621,225 @@ func TestJobHeartBeatFailNotBlockOthers(t *testing.T) {
...
+
+func TestJobManagerWithFault(t *testing.T) {
+	boostJobScheduleForTest(t)
+
+	store, dom := testkit.CreateMockStoreAndDomain(t)
+	waitAndStopTTLManager(t, dom)
+	tk := testkit.NewTestKit(t, store)
+
+	managerCount := 5
+	testDuration := 10 * time.Second
```
YangKeao, line 1743: TODO: don't run it in CI as it'll need a long time. In my laptop it'll need 500ms for each round. It seems that it's quite unstable (but I still don't know why, investigating) in CI.

Run it nightly, and for at least 10 minutes.

pkg/ttl/ttlworker/job_manager.go
```diff
@@ -572,12 +572,14 @@ j:
...
-func (m *JobManager) rescheduleJobs(se session.Session, now time.Time) {
+func (m *JobManager) rescheduleJobs(se session.Session, nowInSessionTz time.Time) {
+	var now time.Time
 	tz, err := se.GlobalTimeZone(m.ctx)
 	if err != nil {
 		terror.Log(err)
+		now = nowInSessionTz
 	} else {
-		now = now.In(tz)
+		now = nowInSessionTz.In(tz)
```
lcwangchao, line 582: Is `now` only used to determine the time window? If so, I think we can change `now` => `nowInGlobalTz`.

For `readyForLockHBTimeoutJobTables` and `lockHBTimeoutJob`, seems it should use the now with session timezone?

pkg/ttl/ttlworker/job_manager_integration_test.go
lcwangchao: We can just add a `t.Skip` here

pkg/ttl/ttlworker/session.go
```diff
@@ -87,7 +87,7 @@ func getSession(pool util.SessionPool) (session.Session, error) {
 	originalIsolationReadEngines, restoreIsolationReadEngines := "", false
 
 	se := session.NewSession(sctx, exec, func(se session.Session) {
-		_, err = se.ExecuteSQL(context.Background(), fmt.Sprintf("set tidb_retry_limit=%d", originalRetryLimit))
+		_, err := se.ExecuteSQL(context.Background(), fmt.Sprintf("set tidb_retry_limit=%d", originalRetryLimit))
```
lcwangchao, line 90: It's better to schedule a new PR to fix it. Though it rarely happens, but I think we should cherry-pick it to the previous versions to avoid potential panics here.

YangKeao: Yeah! I'll submit a new PR to fix the two bugs (and cherry-pick them back to release version), and leave this PR only related to test.

YangKeao:
ref https://github.com/pingcap/tidb/pull/58166

pkg/ttl/ttlworker/session_integration_test.go
```diff
@@ -97,19 +118,12 @@ func (s *sessionWithFault) ExecuteInternal(ctx context.Context, sql string, args
 }
 
 func (s *sessionWithFault) shouldFault(sql string) bool {
-	if s.fault.Load() == nil {
-		return false
-	}
-
-	// as a fault implementation may have side-effect, we should always call it before checking the SQL.
-	shouldFault := (*s.fault.Load()).shouldFault()
```
YangKeao, line 105: When `s.fault.Load() != nil`, the `*s.fault` can still be a nil interface.

pkg/ttl/ttlworker/job_manager_integration_test.go
```diff
@@ -1619,3 +1621,184 @@ func TestJobHeartBeatFailNotBlockOthers(t *testing.T) {
...
+	}
+
+	type managerWithPool struct {
+		m    *ttlworker.JobManager
+		pool util.SessionPool
+	}
+	managers := make([]managerWithPool, 0, managerCount)
+	for i := 0; i < managerCount; i++ {
+		pool := wrapPoolForTest(dom.SysSessionPool())
+		// defer pool.AssertNoSessionInUse(t)
```
lcwangchao, line 1687: useless comment?

pkg/ttl/ttlworker/session_integration_test.go
lcwangchao: Can we modify setFault?

```
func (f *faultSessionPool) setFault(ft fault) {
	if ft == nil {
		f.fault.Store(nil)
	}
	f.fault.Store(&ft)
}
```

lcwangchao (approved):

YangKeao:
/hold

wait for https://github.com/pingcap/tidb/issues/58421

wjhuang2016 (approved):

PR #58112, opened 2024-12-09
br: separate nested logic of table rewrite and building id map

### LINKED ISSUE #57859: br: separate build id map and rewrite logic
## Enhancement

Right now these two functions are interleaved together to reuse some part of the code. The build id map logic can actually be simplified to its own file and the overall code complexity should reduce.

Tristan1900:

Issue Number: close #57859

Problem Summary:

separate the nested logic of table rewriting and building id map.
For building id map we only need to look at the default cf and only for db entry and table entry, rewrite does much more than that, so separating will make code looks cleaner and easier to do table filtering later.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


tests/_utils/run_services
```diff
@@ -261,7 +261,7 @@ start_tiflash() {
     i=0
     while ! run_curl "https://$TIFLASH_HTTP/tiflash/store-status" 1>/dev/null 2>&1; do
         i=$((i+1))
-        if [ "$i" -gt 20 ]; then
+        if [ "$i" -gt 1 ]; then
```
Tristan1900, line 264: will revert later

br/pkg/stream/rewrite_meta_rawkv.go
```diff
@@ -260,22 +145,10 @@ func (sr *SchemasReplace) rewriteDBInfo(value []byte) ([]byte, error) {
...
-			dr.Name = dbInfo.Name.O
-		}
-		return nil, nil
-	}
-
 	dbMap, exist := sr.DbMap[dbInfo.ID]
 	if !exist {
-		return nil, errors.Annotatef(berrors.ErrInvalidArgument, "failed to find id:%v in maps", dbInfo.ID)
+		// db filtered out
+		return nil, nil
```
Tristan1900, line 151: here should only filtered out table/db will not be able to find the mapping. Right now filtered out tables are only some system table. 
Might need some safer mechanism to check if indeed they are filtered out tables, not something we missed due to error in the building id map phase .

br/pkg/restore/log_client/client.go
```diff
@@ -1123,31 +1108,69 @@ func (rc *LogClient) RestoreMetaKVFiles(
...
+	return nil
+}
+
+// buildAndSaveIDMap build id mapping and save it.
+func (rc *LogClient) buildAndSaveIDMap(
+	ctx context.Context,
+	fsInDefaultCF []*backuppb.DataFileInfo,
+	tableMappingManager *stream.TableMappingManager,
+) error {
+	if err := rc.iterAndBuildIDMap(ctx, fsInDefaultCF, tableMappingManager); err != nil {
```
Leavrth, line 1160: Do we need to make sure that the `tableInfo-kv` does not have short value in write cf value?

Leavrth (commented):
rest LGTM

br/pkg/restore/log_client/client.go
Tristan1900: you are right, missed that, adding now!

Leavrth (approved):

br/pkg/stream/rewrite_meta_rawkv.go
```diff
@@ -439,28 +257,13 @@ func (sr *SchemasReplace) rewriteTableInfo(value []byte, dbID int64) ([]byte, er
...
-
-					newID, err = sr.genGenGlobalID(context.Background())
-					if err != nil {
-						return nil, errors.Trace(err)
-					}
-					sr.globalTableIdMap[tbl.ID] = newID
-				}
-				tableReplace.PartitionMap[tbl.ID] = newID
+				// partition filtered out
+				return nil, nil
```
BornChanger, line 261: continue or return here?

Tristan1900: thanks for the review. good call! should continue, but since we don't have partition filter, it should not be triggered at all. Let me add a log at here

BornChanger (approved):

Tristan1900:
bunch of unit tests from other component become flaky it seems

Issue #57497, opened 2024-11-19
DXF: support modify task concurrency at runtime


D3Hunter:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently, after a task is submitted to DXF, we cannot change the max resource it can be used at runtime. But we need this feature in some cases, such as when submitting alter table xxx add index, current value of [tidb_ddl_reorg_worker_cnt](https://docs.pingcap.com/tidb/stable/system-variables#tidb_ddl_reorg_worker_cnt) of which default is 4, is used as the task concurrency, later users might want to accelerate it by increasing the concurrency, which is a supported feature when adding index in standalone mode. And import-into might need it too.

**Describe the feature you'd like:**

support modify task concurrency at runtime, we want to make it easier to extend as we might need to modify other task config at runtime later.

**Describe alternatives you've considered:**

this is new feature
**Teachability, Documentation, Adoption, Migration Strategy:**

DXF is the underlying framework behind upper business, including add-index and import-into, so to support change concurrency for upper business, DXF need to support it first, then upper layer can integrate the ability and expose some API to user

PR #58144, opened 2024-12-11
dxf: support modify concurrency/task-specific-param for later subtask at task executor part

D3Hunter:

Issue Number: ref #57497

Problem Summary:

- support modify concurrency/task-specific-param for later subtask at task executor part, for existing running subtasks, will handle it in later pr.
- `alloc` of `slotManager` might fail now due to the new API `exchange` which is used to exchange slots on concurrency modification
- add `TaskMetaModified` to `StepExecutor` to notify that some business specific param are modified
- use `m.GetResource()` to determine resource instead of `subtask.Concurrency`
- remove `TaskTypeOption` which is useless now
### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


D3Hunter:
@lance6716

pkg/disttask/framework/proto/task.go
```diff
@@ -146,16 +148,19 @@ func (t *TaskBase) CompareTask(other *Task) int {
...
-		return 1
+	if r := t.CreateTime.Compare(other.CreateTime); r != 0 {
+		return r
 	}
-	return int(t.ID - other.ID)
+	return cmp.Compare(t.ID, other.ID)
+}
+
+// String implements fmt.Stringer interface.
+func (t *TaskBase) String() string {
```
lance6716, line 161: if we use JSON, we don't need to adjust the code if we add more fields in future.

lance6716 (commented):
15 / 32 files viewed

pkg/disttask/framework/proto/task.go
D3Hunter: just found I ever write in this way here 😓
https://github.com/pingcap/tidb/blob/4407ef5c2253ea56b838021b954def4bd83c4dca/pkg/executor/importer/job.go#L94-L98
both ways lgtm, currently, I slightly prefer avoiding error here, including unhandled

D3Hunter:
@lance6716 ptal 😁

pkg/disttask/framework/taskexecutor/execute/interface.go
```diff
@@ -111,3 +116,19 @@ func SetFrameworkInfo(exec StepExecutor, step proto.Step, resource *proto.StepRe
 		info.Set(reflect.ValueOf(toInject))
 	}
 }
+
+// ModifyResource modifies the resource of the StepExecutor.
+func ModifyResource(exec StepExecutor, resource *proto.StepResource) {
+	if exec == nil {
+		return
+	}
+	e := reflect.ValueOf(exec)
```
lance6716, line 125: We dont' need to use reflection 😂 Can we add a private `setResource` method to `StepExecFrameworkInfo`. And use it by `exec.setResource`?

`SetFrameworkInfo` is different, because the `StepExecFrameworkInfo` is nil and it can only be initialized by `SetFrameworkInfo` using reflection

pkg/disttask/framework/taskexecutor/manager.go
```diff
@@ -294,39 +317,55 @@ func (m *Manager) cancelTaskExecutors(tasks []*proto.TaskBase) {
...
+	}
+	if !m.slotManager.alloc(&task.TaskBase) {
+		m.logger.Info("alloc slots failed, maybe other task executor alloc more slots at runtime",
+			zap.Int64("task-id", taskBase.ID), zap.Int("concurrency", taskBase.Concurrency),
+			zap.Int("remaining-slots", m.slotManager.availableSlots()))
+		return false
 	}
-	// runCtx only used in executor.Run, cancel in m.fetchAndFastCancelTasks.
+	defer func() {
+		// free the slot if executor not started.
```
lance6716, line 342: otherwise line 373 will free it? Can we unify the free in this `defer`?

lance6716 (commented):
21 / 32 files viewed

pkg/disttask/framework/taskexecutor/task_executor.go
```diff
@@ -223,19 +241,56 @@ func (e *BaseTaskExecutor) Run(resource *proto.StepResource) {
...
 		}
-		task := e.task.Load()
-		if task.State != proto.TaskStateRunning {
+
+		if !bytes.Equal(oldTask.Meta, newTask.Meta) {
+			e.logger.Info("task meta modified",
+				zap.String("oldStep", proto.Step2Str(oldTask.Type, oldTask.Step)),
+				zap.String("newStep", proto.Step2Str(newTask.Type, newTask.Step)))
+			// when task switch to next step, task meta might change too, but in
+			// this case step executor will be recreated, so we only notify it
```
lance6716, line 260: just double check (I didn't check the code) if the new executor is created, it will write task meta with new task step? Will the new concurrency be overwritten in this case?

lance6716 (commented):
rest lgtm

pkg/disttask/framework/taskexecutor/manager.go
D3Hunter: yes, but we cannot unify them. this line is for handling TaskExecutor creation errors, else slots should be freed after TaskExecutor exits, like the `Close` if `Init` failed pattern.

pkg/disttask/framework/taskexecutor/task_executor.go
D3Hunter: yes, the new created executor will use new task info

lance6716: I mean if there's some data race of updating concurrency and updating step.

lance6716 (approved):

pkg/disttask/framework/taskexecutor/task_executor.go
D3Hunter: at this loop, before `runSubtask`, step executor is only read/write by this routine.

pkg/disttask/framework/scheduler/interface.go
```diff
@@ -26,7 +26,7 @@ import (
 // TaskManager defines the interface to access task table.
 type TaskManager interface {
 	// GetTopUnfinishedTasks returns unfinished tasks, limited by MaxConcurrentTask*2,
-	// to make sure lower rank tasks can be scheduled if resource is enough.
+	// to make sure low ranking tasks can be scheduled if resource is enough.
```
tangenta, line 29: I prefer using "high priority" in the description, "ranking" is unclear to readers.

D3Hunter: `priority` is only part of how we order tasks, initially all task have the same priority, so i use rank to clearify it

GMHDBJD (approved):
LGTM

PR #58229, opened 2024-12-13
ddl: Fix issue with concurrent update getting reverted by BackfillData

### LINKED ISSUE #58226: Concurrent updates during Reorganize partition backfill can be overwritten/reverted
## Bug Report

Use current master, a3fa29e58c3bd368982116942f8cff916cb0eb67, apply this diff [backfill-concurrent-dml-reorg-part.diff.txt](https://github.com/user-attachments/files/18118347/backfill-concurrent-dml-reorg-part.diff.txt)
`make failpoint-enable` and run test TestMultiSchemaReorganizeNoPKBackfillDML.

Test passing

The rows updated during the Backfill was overwritten by the just read versions.
error log:
```
    result.go:49: 
        	Error Trace:	/Users/mattias/repos/tidb/pkg/testkit/result.go:49
        	            				/Users/mattias/repos/tidb/pkg/ddl/tests/partition/multi_domain_test.go:1179
        	            				/Users/mattias/repos/tidb/pkg/ddl/tests/partition/multi_domain_test.go:1001
        	            				/Users/mattias/repos/tidb/pkg/ddl/tests/partition/multi_domain_test.go:1195
        	Error:      	Not equal: 
        	            	expected: "[1 1 init O updated]\n[10 30004 delete reorganization NO Original]\n[11 7 public O Original]\n[12 30005 public NO Original]\n[13 8 none O Original]\n[14 30006 none NO Original]\n[2 2 init O updated]\n[3 3 delete only O updated]\n[4 30001 delete only NO updated]\n[5 4 write only O updated]\n[6 30002 write only NO updated]\n[7 5 write reorganization O Original]\n[8 30003 write reorganization NO Original]\n[9 6 delete reorganization O Original]\n"
        	            	actual  : "[1 60001 init O Original]\n[10 30004 delete reorganization NO Original]\n[11 7 public O Original]\n[12 30005 public NO Original]\n[13 8 none O Original]\n[14 30006 none NO Original]\n[2 60002 init O Original]\n[3 60003 delete only O Original]\n[4 60004 delete only NO Original]\n[5 4 write only O Original]\n[6 60005 write only NO Original]\n[7 5 write reorganization O Original]\n[8 30003 write reorganization NO Original]\n[9 6 delete reorganization O Original]\n"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,2 @@
        	            	-[1 1 init O updated]
        	            	+[1 60001 init O Original]
        	            	 [10 30004 delete reorganization NO Original]
        	            	@@ -6,7 +6,7 @@
        	            	 [14 30006 none NO Original]
        	            	-[2 2 init O updated]
        	            	-[3 3 delete only O updated]
        	            	-[4 30001 delete only NO updated]
        	            	-[5 4 write only O updated]
        	            	-[6 30002 write only NO updated]
        	            	+[2 60002 init O Original]
        	            	+[3 60003 delete only O Original]
        	            	+[4 60004 delete only NO Original]
        	            	+[5 4 write only O Original]
        	            	+[6 60005 write only NO Original]
        	            	 [7 5 write reorganization O Original]
        	Test:       	TestMultiSchemaReorganizeNoPKBackfillDML
        	Messages:   	sql:select c1,_tidb_rowid,c2,c3 from t, args:[]
--- FAIL: TestMultiSchemaReorganizeNoPKBackfillDML (0.33s)
```

Current master:
```
tidb_version(): Release Version: v8.5.0-alpha-330-ga3fa29e58c-dirty
Edition: Community
Git Commit Hash: a3fa29e58c3bd368982116942f8cff916cb0eb67
Git Branch: master
UTC Build Time: 2024-12-12 22:57:58
GoVersion: go1.23.4
Race Enabled: false
Store: unistore
```

mjonss:

Issue Number: close #58226, close https://github.com/pingcap/tidb/issues/58692

Problem Summary:
A concurrency test showed that when REORGANIZE PARTITION are copying non-clustered table rows in batches, if an update happens during such batch for the same rows included in the batch, then the batch will overwrite the updates with what the batch originally read.

Reverted the use of table.AddRecord() for non-clustered tables and added the old row into the batch transaction, so the transaction would fail if the old row have been touched (if it already has been copied/double written and has not been touched, it will be skipped from being copied).

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
For non-clustered tables during REORGANIZE PARTITION data copying/backfill, if a row is updated at the same time the reorg is copying that row in a batch, it could overwrite with the state before the update.
```


pkg/ddl/partition.go
```diff
@@ -3869,54 +3837,65 @@ func (w *reorgPartitionWorker) BackfillData(handleRange reorgBackfillTask) (task
...
-					}
-					taskCtx.addedCount++
+				// Due to EXCHANGE PARTITION, the existing _tidb_rowid may collide between partitions!
+				// Generate new _tidb_rowid.
+				recordID, err := tables.AllocHandle(w.ctx, w.tblCtx, w.reorgedTbl)
+				if err != nil {
+					return errors.Trace(err)
 				}
+
+				// tablecodec.prefixLen is not exported, but is just TableSplitKeyLen + 2
```
Defined2014, line 3895: Not get the comments point. Why `+2` here?

Defined2014 (commented):
Is it true that the cluster index will not be duplicated during reorg? This is because the primary key check has ensured this.

pkg/ddl/partition.go
```diff
@@ -3869,54 +3837,65 @@ func (w *reorgPartitionWorker) BackfillData(handleRange reorgBackfillTask) (task
...
+					return errors.Trace(err)
+				}
+				// Ensure the key does not exist!
+				// It actually works without this assertion, but it may help troubleshooting.
+				err = txn.SetAssertion(w.oldKeys[i], kv.SetAssertNotExist)
+				if err != nil {
+					return errors.Trace(err)
+				}
+				// Also don't actually write it :)
+				err = txn.Delete(w.oldKeys[i])
```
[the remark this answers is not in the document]
mjonss: I'm not sure how to only force the transaction to fail if a key already exists, regardless of settings like `tidb_txn_assertion_level`, so by involving the key in the transaction, it will fail if another concurrent transaction has modified it (like insert it due to UPDATE). Just having the `SetAssertion()` is not enough, we could have the `txn.Delete()` or `txn.Set()` only, not sure if Set+Assert+Delete is better though, what is your thought?

pkg/ddl/partition.go
mjonss: TableSplitKeyLen is `t_<encoded tableID>` only, and we want to include the `r_` as well, so that is where the +2 comes from.

Defined2014: Use `len(recordPrefixSep)` instead of 2?

pkg/ddl/partition.go
```diff
@@ -3869,54 +3837,65 @@ func (w *reorgPartitionWorker) BackfillData(handleRange reorgBackfillTask) (task
...
+				if _, ok := found[string(w.oldKeys[i])]; ok {
+					// Already filled, i.e. double written earlier by concurrent DML
+					continue
+				}
+
+				// Check if we can write the old key, since there can still be concurrent update
+				// happening on the rows from fetchRowColVals(), we can involve them in this
+				// transaction and fail when committing, so we will retry with a new snapshot.
+				// TODO: Test a concurrent pessimistic transaction that reads/writes concurrently here,
+				// and commits after after this transaction.
```
tangenta, line 3871:
```suggestion
				// and commits after this transaction.
```

pkg/ddl/partition.go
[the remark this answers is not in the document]
Defined2014: I don't quite understand this code. But why does `SetAssertNotExist` succeed after `txn.Set` it first?

pkg/ddl/partition.go
mjonss: Hmm, all names that start with lower case is not exported :(
I can change those names, but then the PR grows a bit with unrelated changes...
Is it OK if I create a follow-up issue+PR for exporting RecordPrefixSepLength later?

pkg/ddl/partition.go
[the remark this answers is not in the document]
mjonss: I would expect that only lock calls are directly checked/forwarded to the KV store, while SetAssertion/Set/Delete is only applied during Commit.

pkg/ddl/partition.go
```diff
@@ -3869,54 +3837,66 @@ func (w *reorgPartitionWorker) BackfillData(handleRange reorgBackfillTask) (task
...
-			// and we should skip it.
-			var found map[string][]byte
+		failpoint.InjectCall("PartitionBackfillData", len(w.rowRecords) > 0)
+		// For non-clustered tables, we need to replace the _tidb_rowid handles since
+		// there may be duplicates across different partitions, due to EXCHANGE PARTITION.
+		// Meaning we need to check here if a record was double written to the new partition,
+		// i.e. concurrently written by StateWriteOnly or StateWriteReorganization.
+		// and if so, skip it.
+		var found map[string][]byte
+		if len(w.oldKeys) > 0 {
```
Defined2014, line 3847: Does it mean when `len(w.oldKeys) > 0` it shoudn't be a clustered table?

pkg/ddl/partition.go
```diff
@@ -3869,54 +3837,66 @@ func (w *reorgPartitionWorker) BackfillData(handleRange reorgBackfillTask) (task
...
+			key := prr.key
+
 			if len(w.oldKeys) > 0 {
-				// we must check if old IDs already been written,
-				// i.e. double written by StateWriteOnly or StateWriteReorganization.
-				// The good thing is that we can then also skip the index generation for that row and we don't need to
-				// check if duplicate index entries was already copied either!
-				// TODO: while waiting for BatchGet to check for duplicate, do another round of reads in parallel?
-				found, err = txn.BatchGet(ctx, w.oldKeys)
+				if _, ok := found[string(w.oldKeys[i])]; ok {
```
Defined2014, line 3862: Does `len(w.oldKeys) > 0` mean it's a non clustered table? Maybe we could create a function for this if stmt and named it related with `non Clustered`. It's easy to understand.

mjonss: I will clarify with a comment, no need for a single use function I think.

pkg/ddl/partition.go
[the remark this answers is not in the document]
cfzjywxk: If the goal is simply to prevent the row_key from being modified, using txn.LockKeys to apply a pessimistic lock on the corresponding row is the appropriate approach.

The purpose of Assertion is fundamentally different:
- It is designed to validate invariants or constraints to ensure correctness is not violated.
- It is not intended for concurrency control or preventing concurrent modifications.

Using LockKeys in a pessimistic transaction explicitly handles concurrency by preventing conflicting writes, while assertions act as safeguards to check assumptions about the state after operations are performed.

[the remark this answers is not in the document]
cfzjywxk: For the use of internal transactions during the execution of DDL tasks and their concurrency control with DML, it is recommended to consult DDL-related colleagues to confirm whether the logic complies with DDL constraint requirements.

There should be similar code references for regular DDL backfill.
/cc @wjhuang2016 @tangenta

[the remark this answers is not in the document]
Defined2014: The difference with regular DDL it uses the version/lock on a key to control another unrelated key. Looks too strange here.

Defined2014 (approved):

bb7133 (approved):
LGTM

pkg/ddl/partition.go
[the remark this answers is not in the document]
mjonss: After changing to `txn.LockKeys()` it is now same as what `(*addIndexTxnWorker) BackfillData()` does [here](https://github.com/pingcap/tidb/blob/0be1983389d04348f72de36c4d922c333389c1e4/pkg/ddl/index.go#L2327).

tangenta (approved):

PR #58261, opened 2024-12-13
planner: extend predicate simplification for subquery and nested expressions | tidb-test=pr/2455

### LINKED ISSUE #58171: TiDB doesn't apply short-circuit evaluation when there is nested sub-query in where condition
## Bug Report

```
SELECT
  tab.id AS id
FROM
  t1 AS tab
WHERE 1 = 1 OR
  (EXISTS(SELECT 1 FROM t2 WHERE table_id = tab.id))


+---------------------------------+------------+---------------+-----------+----------------------------------------------------------------------------------------------------+------------------------------------------------------------------------------------------------------------+
| id                              | estRows    | estCost       | task      | access object                                                                                      | operator info                                                                                              |
+---------------------------------+------------+---------------+-----------+----------------------------------------------------------------------------------------------------+------------------------------------------------------------------------------------------------------------+
| Projection_8                    | 6071792.00 | 868583556.41  | root      |                                                                                                    | test.t1.id                                                                                |
| └─Selection_9                   | 6071792.00 | 867977591.57  | root      |                                                                                                    | or(1, Column#68)                                                                                           |
|   └─HashJoin_10                 | 7589740.00 | 489249565.57  | root      |                                                                                                    | left outer semi join, equal:[eq(test.t1.id, test.t2.tableid)] |
|     ├─IndexReader_26(Build)     | 1385922.00 | 33841325.91   | root      |                                                                                                    | index:IndexFullScan_25                                                                                     |
|     │ └─IndexFullScan_25        | 1385922.00 | 392860557.71  | cop[tikv] | table:t2, index:PRIMARY(accountid, metastoreid, tableid, principal, privilege) | keep order:false                                                                                           |
|     └─IndexReader_16(Probe)     | 7589740.00 | 159804590.85  | root      |                                                                                                    | index:IndexFullScan_15                                                                                     |
|       └─IndexFullScan_15        | 7589740.00 | 1768611067.74 | cop[tikv] | table:tab, index:securable_id_deleted_at_index(purgestate, deletedat, id)                        | keep order:false                                                                                           |
+---------------------------------+------------+---------------+-----------+----------------------------------------------------------------------------------------------------+------------------------------------------------------------------------------------------------------------+
```

short-circuit should be applied, table t2 no need to show up

short-circuit was not applied, table t2 showed up

current master.

ghazalfamilyusa:

Issue Number: close #58171

### Problem Summary:
The main problem can be explained by the example below. The query has a condition "True OR subquery" which can be simplified to True and therefore avoid the subquery execution.

 ```
SELECT 1 
FROM t1 AS tab 
WHERE 1 = 1 OR (EXISTS(SELECT  /*+ NO_DECORRELATE() */ 1 FROM t2 WHERE a2 = a1 ))

Projection      8000.00 root            1->Column#9
└─Selection     8000.00 root            or(1, Column#8)
  └─Apply       10000.00        root            CARTESIAN left outer semi join, left side:TableReader
    ├─TableReader(Build)        10000.00        root            data:TableFullScan
    │ └─TableFullScan   10000.00        cop[tikv]       table:tab       keep order:false, stats:pseudo
    └─TableReader(Probe)        100000.00       root            data:Selection
      └─Selection       100000.00       cop[tikv]               eq(test.t2.a2, test.t1.a1)
        └─TableFullScan 100000000.00    cop[tikv]       table:t2        keep order:false, stats:pseudo

```

The correct behaviour is this 

```
Projection      10000.00        root            1->Column#9
└─TableReader   10000.00        root            data:TableFullScan
  └─TableFullScan       10000.00        cop[tikv]       table:tab       keep order:false, stats:pseudo

```

There are two problems that prevents the above optimizations:
- Predicate simplification in TiDB optimizer does not cover all cases of True/False simplifications. 
- The subquery expression is rewritten to Apply operator early on during the AST to logical plan building and it is  not possible for predicate simplification to undo that logic.
 

Changes are made to address both problems above by:
- Extending Predicate Simplification rewrite to cover True/False logical simplifications for all expressions.
- Remove subquery plan when the predicate is True by undoing the Apply operator. This logic is added to the decorrelation rewrite logic since it always process Apply and convert it to Join if it decides to decorrelate it. 

### Check List

Tests 

- [X] Unit test
- [] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/rule_predicate_simplification.go
```diff
@@ -311,6 +347,83 @@ func pruneEmptyORBranches(sctx base.PlanContext, predicates []expression.Express
...
+	_, conditionType := FindPredicateType(sctx, condition)
+
+	if conditionType == orPredicate {
+		condition, applied = shortCircuitANDORLogicalConstants(sctx, condition, true)
+	} else if conditionType == andPredicate {
+		condition, applied = shortCircuitANDORLogicalConstants(sctx, condition, false)
+	}
+
+	if applied {
+		sctx.GetSessionVars().StmtCtx.SetSkipPlanCache("True/False predicate simplification is triggered")
```
AilinKid, line 400: we can in the out most caller part set them only once, otherwise, like the embedded case will call it >1 times?

pkg/planner/core/rule_decorrelate.go
```diff
@@ -133,8 +133,65 @@ func (*DecorrelateSolver) aggDefaultValueMap(agg *logicalop.LogicalAggregation)
...
+	logicalSelection, ok := p.(*logicalop.LogicalSelection)
+	if !ok {
+		return nil, false
+	}
+
+	// Retrieve the child of LogicalSelection
+	selectSource := logicalSelection.Children()[0]
+
+	// Check if the child is a LogicalApply
+	apply, ok := selectSource.(*logicalop.LogicalApply)
```
AilinKid, line 153: I'm afriad that this pattern is more tricky
```
tidb> explain select * from (select * from t where EXISTS(SELECT /*+ NO_DECORRELATE() */ 1  FROM t t2 WHERE t2.a = t.a)) s where 1 ;
+-----------------------------+--------------+-----------+---------------+-----------------------------------------------+
| id                          | estRows      | task      | access object | operator info                                 |
+-----------------------------+--------------+-----------+---------------+-----------------------------------------------+
| Apply_12                    | 10000.00     | root      |               | CARTESIAN semi join, left side:TableReader_14 |
| ├─TableReader_14(Build)     | 10000.00     | root      |               | data:TableFullScan_13                         |
| │ └─TableFullScan_13        | 10000.00     | cop[tikv] | table:t       | keep order:false, stats:pseudo                |
| └─TableReader_17(Probe)     | 100000.00    | root      |               | data:Selection_16                             |
|   └─Selection_16            | 100000.00    | cop[tikv] |               | eq(test.t.a, test.t.a)                        |
|     └─TableFullScan_15      | 100000000.00 | cop[tikv] | table:t2      | keep order:false, stats:pseudo                |
+-----------------------------+--------------+-----------+---------------+-----------------------------------------------+
6 rows in set (0.01 sec)
```
in this case:
selection<1> which is finally removed, has a apply as its child, while we couldn't just let apply inner child to directly be the root here? cause the apply source is not comes from outer query block's where clause's scalar Exists, at which we thought Exists is same level with Where 1 here.
/cc @winoros

AilinKid (commented):
for predicate simplification, it's a definite awesome to me

tests/integrationtest/r/planner/core/casetest/physicalplantest/physical_plan.result
```diff
@@ -3835,32 +3835,36 @@ insert into tbl_43 values("BCmuENPHzSOIMJLPB"),("LDOdXZYpOR"),("R"),("TloTqcHhdg
...
-    ├─Selection(Build)	0.00	cop[tikv]		1
-    │ └─TableRangeScan	0.00	cop[tikv]	table:tbl_43	range:["LUBGzGMA","LUBGzGMA"], keep order:true, stats:pseudo
-    ├─IndexRangeScan(Build)	0.42	cop[tikv]	table:tbl_43, index:idx_261(col_304)	range:[-inf,"YEpfYfPVvhMlHGHSMKm"), keep order:true, stats:pseudo
-    ├─IndexRangeScan(Build)	0.42	cop[tikv]	table:tbl_43, index:idx_262(col_304)	range:("PE",+inf], keep order:true, stats:pseudo
-    ├─TableRangeScan(Build)	0.42	cop[tikv]	table:tbl_43	range:[-inf,"MFWmuOsoyDv"), keep order:true, stats:pseudo
-    ├─TableRangeScan(Build)	0.42	cop[tikv]	table:tbl_43	range:("TSeMYpDXnFIyp",+inf], keep order:true, stats:pseudo
-    └─Selection(Probe)	1.00	cop[tikv]		or(or(lt(planner__core__casetest__physicalplantest__physical_plan.tbl_43.col_304, "YEpfYfPVvhMlHGHSMKm"), gt(planner__core__casetest__physicalplantest__physical_plan.tbl_43.col_304, "PE")), or(and(eq(planner__core__casetest__physicalplantest__physical_plan.tbl_43.col_304, "LUBGzGMA"), 1), or(lt(planner__core__casetest__physicalplantest__physical_plan.tbl_43.col_304, "MFWmuOsoyDv"), gt(planner__core__casetest__physicalplantest__physical_plan.tbl_43.col_304, "TSeMYpDXnFIyp"))))
```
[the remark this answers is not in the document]
ghazalfamilyusa: Before new code, the predicate is below. and(eq(planner__core__casetest__physicalplantest__physical_plan.tbl_43.col_304, "LUBGzGMA"), 1) gets simplified to eq(planner__core__casetest__physicalplantest__physical_plan.tbl_43.col_304, "LUBGzGMA") which allows the detacher to consume all the predicates. 

----------------------------------------------------------------
or(
   or(
       lt(planner__core__casetest__physicalplantest__physical_plan.tbl_43.col_304, "YEpfYfPVvhMlHGHSMKm"),
       gt(planner__core__casetest__physicalplantest__physical_plan.tbl_43.col_304, "PE")
   ),
   or(
      and(eq(planner__core__casetest__physicalplantest__physical_plan.tbl_43.col_304, "LUBGzGMA"), 1),
      or(
          lt(planner__core__casetest__physicalplantest__physical_plan.tbl_43.col_304, "MFWmuOsoyDv"),
          gt(planner__core__casetest__physicalplantest__physical_plan.tbl_43.col_304, "TSeMYpDXnFIyp")
      )
   )
)

pkg/planner/core/rule_decorrelate.go
```diff
@@ -133,8 +133,71 @@ func (*DecorrelateSolver) aggDefaultValueMap(agg *logicalop.LogicalAggregation)
...
+	if !ok {
+		return nil, false
+	}
+
+	// Ensure the Apply operator is of a suitable join type to match the required pattern.
+	// Only LeftOuterJoin or LeftOuterSemiJoin are considered valid here.
+	if apply.JoinType != logicalop.LeftOuterJoin && apply.JoinType != logicalop.LeftOuterSemiJoin {
```
AilinKid, line 160: can we add one more line check here?
LeftOuter apply's schema.columns[len-1] (whichi is the scalar result output column) is in the selection's condition, which will make sure that the apply's scalar output is used just in the parent selection.

AilinKid (commented):
rest lgtm

pkg/planner/core/rule_decorrelate.go
winoros: I think we can add a test to check it.
It should the case like `select ((1 = 1) or (exists(...)) from t`. So there'll be a final projection to project the result because the extra column of the semi-join only indicates that the join result is matched or not.
Now that the semi-join's execution result is useless, the extra column should also be useless.

ghazalfamilyusa: The outer join should be sufficient since it only applies if you have subquery predicate within a complex condition. @winoros : did not get your test suggestion. Is the subquery in the select list or where cluase? If where clause then it is already covered. Also, as far as the extra projection, prune columns is applied again at the end of the rules which removes the subquery result column.

AilinKid (approved):

winoros (approved):

Issue #58287, opened 2024-12-16
DATA RACE at the infoschema.*Sieve


hawkingrei:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 3. What did you see instead (Required)

```
==================
WARNING: DATA RACE
Write at 0x00c00c00df50 by goroutine 22527:
  github.com/pingcap/tidb/pkg/infoschema.(*Sieve[go.shape.struct { github.com/pingcap/tidb/pkg/infoschema.tableID int64; github.com/pingcap/tidb/pkg/infoschema.schemaVersion int64 },go.shape.3a1a9f922c98916eb710c4393a2ff31cbdfc97a4f6c06c922620e553dbdcab66]).Set()
      pkg/infoschema/sieve.go:135 +0x1da
  github.com/pingcap/tidb/pkg/infoschema.(*infoschemaV2).TableByName()
      pkg/infoschema/infoschema_v2.go:808 +0xa4b
  github.com/pingcap/tidb/pkg/infoschema.(*SessionExtendedInfoSchema).TableByName()
      pkg/infoschema/infoschema.go:818 +0x2b7
  github.com/pingcap/tidb/pkg/planner/core.(*preprocessor).tableByName()
      pkg/planner/core/preprocess.go:507 +0x371
  github.com/pingcap/tidb/pkg/planner/core.(*preprocessor).handleTableName()
      pkg/planner/core/preprocess.go:1616 +0x73d
  github.com/pingcap/tidb/pkg/planner/core.(*preprocessor).Leave()
      pkg/planner/core/preprocess.go:614 +0x15d2
  github.com/pingcap/tidb/pkg/parser/ast.(*TableName).Accept()
      pkg/parser/ast/dml.go:454 +0x2c1
  github.com/pingcap/tidb/pkg/parser/ast.(*TableSource).Accept()
      pkg/parser/ast/dml.go:606 +0xc4
  github.com/pingcap/tidb/pkg/parser/ast.(*Join).Accept()
      pkg/parser/ast/dml.go:244 +0xeb
  github.com/pingcap/tidb/pkg/parser/ast.(*TableRefsClause).Accept()
      pkg/parser/ast/dml.go:829 +0xc4
  github.com/pingcap/tidb/pkg/parser/ast.(*SelectStmt).Accept()
      pkg/parser/ast/dml.go:1501 +0x6a5
  github.com/pingcap/tidb/pkg/planner/core.Preprocess()
[...]

PR #58292, opened 2024-12-16
Revert "infoschema: change sieve Get() function Mutex to RWMutex"

tiancaiamao:
Reverts pingcap/tidb#58243

Issue Number: close https://github.com/pingcap/tidb/issues/58287

Problem Summary:

Found DATA RACE  in CI

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/infoschema/sieve.go
+++ b/pkg/infoschema/sieve.go
@@ -18,7 +18,6 @@ import (
 	"container/list"
 	"context"
 	"sync"
-	"sync/atomic"
 
 	"github.com/pingcap/failpoint"
 	"github.com/pingcap/tidb/pkg/infoschema/internal"
@@ -28,7 +27,7 @@ import (
 type entry[K comparable, V any] struct {
 	key     K
 	value   V
-	visited atomic.Bool
+	visited bool
 	element *list.Element
 	size    uint64
 }
@@ -49,7 +48,7 @@ func (t *entry[K, V]) Size() uint64 {
 type Sieve[K comparable, V any] struct {
 	ctx    context.Context
 	cancel context.CancelFunc
-	mu     sync.RWMutex
+	mu     sync.Mutex
 	size   uint64
 	// capacity can be set to zero for disabling infoschema v2
 	capacity uint64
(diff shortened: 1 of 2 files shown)


hawkingrei (approved):

lance6716 (approved):

pkg/infoschema/sieve.go
```diff
@@ -157,13 +156,10 @@ func (s *Sieve[K, V]) Get(key K) (value V, ok bool) {
...
 	})
-
-	s.mu.RLock()
-	e, ok := s.items[key]
-	s.mu.RUnlock()
-
-	if ok {
-		e.visited.Store(true)
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
lance6716, line 160: /hold

Do you have time to locate the root cause? Maybe we only need to increase the lock scope a bit

wjhuang2016 (approved):

hawkingrei:
@lance6716 @tiancaiamao  We can investigate thie problem late. now CI have been blocked by the problem.

/unhold

pkg/infoschema/sieve.go
tiancaiamao: Inside write lock protected code, it's

```
	if e, ok := s.items[key]; ok {
		e.value = value
		e.visited.Store(true)
		return
	}
```

rather than

```
	if e, ok := s.items[key]; ok {
		s.items[key] = &entry{
                         key: key,
                         value: value,
                }
	}
```

It's not an atomic  update ...

So when the reader only partly read locked, it still data race.

```
mu.RLock()
e := s.items[key]
mu.RUnlock()

// read e.value
```

tiancaiamao: If we change to 

```
	if e, ok := s.items[key]; ok {
		s.items[key] = &entry{
                         key: key,
                         value: value,
                }
	}
```

It's allocating under write lock, potentially slower.

Overall, changing from Mutex to  RWMutex is not a carefully thought.
If we do need better performance here, we need better design and more time to thought.
Revert it first.

Issue #58321, opened 2024-12-17
Reduce the infoschema v2 btree memory usage


tiancaiamao:
## Enhancement

Currently the infoschema v2 is a two level data struct:

```
Btree:   table name -> table id
Cache:   table id -> table
```

The size of cache is under control by @@tidb_schema_cache_size

The btree is not, it store the information of the table name to id index.

I tested the master branch, currently creating 100K tables the btree takes 40.69MB
creating 200K tables the btree takes 80.95MB
...
1M tables, the btree itself could take 400MB 
After restart (multiple versions excluded), 1M tables still takes 283MB

![Image](https://github.com/user-attachments/assets/5ce92425-3f0b-4e0b-ba5a-45ea4eb4d859)

So if we have 3M tables above, the memory usage by the btree might goes up to more than 1GB
Too many small objects also empose a huge burden on Go runtime GC.

tiancaiamao:
After PR 58324 and 58390, I can see no further improvement with a low cost.
Close it for now.

jebter:
Releted to issue: https://github.com/pingcap/tidb/issues/50959

PR #58324, opened 2024-12-17
infoschema: change infoschem v2 btree element from tableItem to *tableItem

tiancaiamao:

Issue Number: ref #58321

Problem Summary:

`byName` and `byID` both hold the `tableItem`, if we change it to `*tableItem`, the can share the object to reduce memory usage.

Before:

Image

After:

image

1M table the btree memory usage reduce from 283M to 110MB

### Check List

Tests 

- [X] Unit test

No new code added, it should be covered by the existing tests

- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


lance6716:
Please also check GC pressure after the change

tiancaiamao:
> Please also check GC pressure after the change because there are more pointers need to be tracked

Before the change:

image

After the change:

image

The statistics part uses much more objects than this change, if we want to optimize object count for GC, here is not the primary concern.  @lance6716

lance6716 (approved):

wjhuang2016 (approved):

PR #58346, opened 2024-12-17
expression,parser: Remove PASSWORD() function | tidb-test=pr/2459

### LINKED ISSUE #58341: > This function is deprecated in MySQL 5.7 and is removed in MySQL 8.0.
> This function is deprecated in MySQL 5.7 and is removed in MySQL 8.0.

Should we rename it to `password` or just remove it?

I found it included by https://github.com/pingcap/tidb/pull/2680, @XuHuaiyu  PTAL

_Originally posted by @Defined2014 in https://github.com/pingcap/tidb/issues/58312#issuecomment-2547313732_

dveeden:

Issue Number: close #58341

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [x] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
The deprecated `PASSWORD()` function was removed.
```


dveeden:
Now this happens, not sure why:
```
mysql-8.0.11-TiDB-v8.5.0-alpha-359-g8a80a41b8d-dirty> select password("foobar");
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your TiDB version for the right syntax to use line 1 column 16 near "("foobar")" 
```

pkg/expression/distsql_builtin.go
```diff
@@ -581,8 +581,6 @@ func getSignatureByPB(ctx BuildContext, sigCode tipb.ScalarFuncSig, tp *tipb.Fie
 		f = &builtinCompressSig{base}
 	case tipb.ScalarFuncSig_MD5:
 		f = &builtinMD5Sig{base}
-	case tipb.ScalarFuncSig_Password:
-		f = &builtinPasswordSig{base}
```
mjonss, line 585: Does this mean we should clean up the protobuf/tikv/tiflash repos as well? Maybe just create deprecation issues in their repositories?

tests/integrationtest/t/expression/builtin.test
```diff
@@ -1384,9 +1384,6 @@ set tidb_enable_vectorized_expression=default;
 drop table if exists t;
 create table t(a char(41), b char(41), c char(41));
 insert into t values(NULL, '', 'abc');
-select password(a) from t;
```
mjonss, line 1387: Maybe keep one test line, to show that it is not supported any longer?

mjonss (approved):
Are most users still using `mysql_native_password` with TiDB? Maybe we should keep the backwards compatible style `PASSWORD()` function in that case. Otherwise I think it is better to remove it, in-line with 8.0.

dveeden:
> Are most users still using `mysql_native_password` with TiDB? Maybe we should keep the backwards compatible style `PASSWORD()` function in that case. Otherwise I think it is better to remove it, in-line with 8.0.

I think we still have `mysql_native_password` as default for all TiDB versions. So I would expect many users. Moving to `caching_sha2_password` requires TLS (ignoring rsa keypairs that are both a usability and security nightmare) and not everyone is willing to switch to TLS.

MySQL 8.0 still supports `mysql_native_password` in both server and client and many users configured this to be the default. With recent innovation release Oracle has removed `mysql_native_password` from the server and on the client moved it to a plugin.

I think people by now should use these statements instead of the `PASSWORD()` function:
```sql
CREATE USER ... IDENTIFIED BY 'secret';
ALTER USER ... IDENTIFIED BY 'secret';
CREATE USER ... IDENTIFIED WITH mysql_native_password BY 'secret';
```
```

pkg/expression/distsql_builtin.go
dveeden: I would expect this to be listed on one of these pages:
- https://docs.pingcap.com/tidb/stable/expressions-pushed-down
- https://docs.pingcap.com/tidb/stable/tiflash-supported-pushdown-calculations

But I don't see it there. 

Looking at the sourcecode I see this:
- https://github.com/tikv/tikv/blob/7ef4aa9549f136cabdac7e5177f39fe16631b514/components/tidb_query_expr/src/impl_encryption.rs#L141-L161
- https://github.com/pingcap/tiflash/blob/ea3620547107437a9ead0ed7b7075bcf7a08d474/dbms/src/Flash/Coprocessor/DAGUtils.cpp#L407

So it looks cleanup is only needed for TiKV.
https://github.com/tikv/tikv/issues/18031

dveeden: We *could* cleanup https://github.com/pingcap/tipb/blob/8408c760e822f7cd7564ce280842af8a417f0a47/proto/expression.proto#L405 but I think it might be better to leave this.

lcwangchao (approved):

dveeden:
@Defined2014 I didn't add the "compatibility-breaker" label as the description "Violation of forwards/backwards compatibility in a design-time piece." doesn't seem to match as this isn't design-time. Also not sure if the label should be on the PR or on the issue. Also this is removing something that has been deprecated for way more than the required two LTS versions, so this is not breaking any rules.

Let's keep the "compatibility-breaker" on this PR for now as it does have a high impact on compatibility. But maybe we need to clarify the description of the label etc.

dveeden:
@Defined2014 @XuHuaiyu @hawkingrei PTAL

hawkingrei (approved):

windtalker:
I will involve a pm to take a look at this change, before that I suggest we hold this pr? @dveeden

dveeden:
> I will involve a pm to take a look at this change, before that I suggest we hold this pr? @dveeden

Ok. I'll add a hold label for now.

windtalker:
Opinions from @joey-yez 
1. If we wants to deprecate something, there is some internal flow to follow, which involves creating FRM and FD
2. Since there is no clear customer demand and the maintenance cost is not high, he prefers to maintain the status quo

dveeden:
@windtalker I'm ok with that. just a few notes

1. This is already deprecated in TiDB for a long time. So this is just removing something that was already deprecated instead of starting the deprecation process.
2. The are now multiple ways of setting a password. And using `PASSWORD()` only works for one of the multiple authentication methods. I think it would be best to only keep the method that works for all authentication methods to make this less confusing to users.
3. This would follow the changes that MySQL made, keeping TiDB more similar to MySQL.

PR #58352, opened 2024-12-17
sessionctx: Remove Query Cache

### LINKED ISSUE #9156: MySQL 8.0: Remove Query Cache
## Feature Request

MySQL 8.0 removes support for the query cache.

This feature was never implemented in TiDB, but the variables and commands that support it were added.  Similar to MySQL, it is important to keep hints like `SQL_NO_CACHE` since mysqldump/mydumper clients may specify this.  But we should be able to [remove the following](https://dev.mysql.com/doc/refman/8.0/en/mysql-nutshell.html#mysql-nutshell-removals):

* `FLUSH QUERY CACHE` and `RESET QUERY CACHE` (should not apply to TiDB)
* These system variables: query_cache_limit, query_cache_min_res_unit, query_cache_size, query_cache_type, query_cache_wlock_invalidate.
* These status variables: Qcache_free_blocks, Qcache_free_memory, Qcache_hits, Qcache_inserts, Qcache_lowmem_prunes, Qcache_not_cached, Qcache_queries_in_cache, Qcache_total_blocks (should not apply to TiDB)
* These thread states: checking privileges on cached query, checking query cache for query, invalidating query cache entries, sending cached result to client, storing result in query cache, Waiting for query cache lock (should not apply to TiDB)
* The `SQL_CACHE SELECT` modifier (will apply to TiDB)

This should not require support for a compatibility mode.  We could optionally not remove the `SQL_CACHE SELECT` modifier to ensure closer compatibility, but I think this could mislead a user to believe there is a cache.

Improves MySQL 8.0 compatibility.

dveeden:

Issue Number: close #9156

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [x] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
The Query Cache variables have been removed.
```


pkg/executor/set_test.go
```diff
@@ -1351,14 +1351,6 @@ func TestValidateSetVar(t *testing.T) {
...
 	require.True(t, terror.ErrorEqual(err, variable.ErrWrongValueForVar), fmt.Sprintf("err %v", err))
 
-	tk.MustExec("set @@query_cache_type=0")
-	result = tk.MustQuery("select @@query_cache_type;")
-	result.Check(testkit.Rows("OFF"))
-
-	tk.MustExec("set @@query_cache_type=2")
-	result = tk.MustQuery("select @@query_cache_type;")
-	result.Check(testkit.Rows("DEMAND"))
-
```
mjonss, line 1361: Maybe update this to `tk.MustContainErrMsg()` or so, instead of removing it?

tests/integrationtest/t/executor/set.test
```diff
@@ -12,26 +12,26 @@ select @@global.tidb_max_delta_schema_count;
...
 
 # TestGetSetNoopVars
-SELECT @@query_cache_type;
-SHOW VARIABLES LIKE 'query_cache_type';
-SET query_cache_type=2;
-SELECT @@query_cache_type;
+SELECT @@innodb_buffer_pool_size;
+SHOW VARIABLES LIKE 'innodb_buffer_pool_size';
+SET GLOBAL innodb_buffer_pool_size=134217728;
+SELECT @@innodb_buffer_pool_size;
```
mjonss, line 18: Just curious why adding a test for `innodb_buffer_pool_size` here? :)

mjonss (approved):

tests/integrationtest/t/executor/set.test
dveeden: This is not adding a test, this test is to test get/set of noop vars and it was using `query_cache_type`, which is removed, so changed it so it now does the same test but uses `innodb_buffer_pool_size` instead.

pkg/executor/set_test.go
dveeden: done in 4274325b70f9d31f37ff3192a8171ac99b409882

lcwangchao (approved):

yudongusa (approved):

PR #58362, opened 2024-12-18
planner,util/ranger: reduce mem allocs on rebuildRange

zyguan:

Issue Number: ref #56649

Problem Summary: A lot of small objects will be allocated when `rebuildRange`, which leads to notable runtime overhead. See the following cpu profile of a index-scan workload (nearly half of rebuildRange time is spent on mallocgc).

image

Introduce a planctx level allocator for allocating/reusing range releated objects, includes:
- internal `point`
- small `[]*point` (len <= 4)
- common `Range` converted from points (with single low/high value)

After this PR, the cpu profile of the same benchmark can be:

image

The heap alloc metrics are shown as the following:

image

Here are benchmark results:

| workload | threads | baseline | this PR  | diff |
| -------- | ------: | -------: | -------: | ---: |
| idxscan  |      30 | 44223.97 | 44990.54 | 1.7% |
| idxscan  |      60 | 49583.53 | 50980.52 | 2.8% |
| idxscan  |     120 | 50929.30 | 52141.18 | 2.4% |

> The idxscan workload executes `select id from sbtest1 where k between {N} and {N+10}` on sysbench oltp dataset.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


winoros:
The plan cache is going to be instance level. I think this is not resonable for instance level plan cache?

zyguan:
> The plan cache is going to be instance level. I think this is not resonable for instance level plan cache?

I'm not sure, but the instance-level plan cache also need to `adjustCachedPlan -> RebuildPlan4CachedPlan -> rebuildRange`, right?

This is a draft PR to show the issue of util/ranger, a possible solution and the improvement can be archived by it. After discussing with @cfzjywxk and @qw4990 , we think the risk of this PR outweighs the benefits, so no plan to move forward.

PR #58376, opened 2024-12-18
ddl notifier: use pagination for SELECT to reduce memory usage

### LINKED ISSUE #58368: TiDB OOM after created about 6.5M tables
## Bug Report

TiDB is 16c32g(about 27.3G usable to the pod), we are creating 500K DB with 9M tables total with `split-table` disabled, after created about 6.5M table, ddl owner OOM
Image

keep running
OOM
master aec0fc58bde1b6f5bdff46dabcf229ebad9ab1de

lance6716:

Issue Number: close #58368

Problem Summary:

as title. refine the `Store.List`.

```golang
// Store is the (de)serialization and persistent layer.
type Store interface {
...
	// List will start a transaction of given session and read all schema changes
	// through a ListResult. The ownership of session is occupied Store until CloseFn
	// is called.
	List(ctx context.Context, se *sess.Session) ListResult
}

// ListResult is the result stream of a List operation.
type ListResult interface {
	// Read tries to decode at most `len(changes)` SchemaChange into given slices. It
	// returns the number of schemaChanges decoded, 0 means no more schemaChanges.
	//
	// Note that the previous SchemaChange in the slice will be overwritten when call
	// Read.
	Read(changes []*SchemaChange) (int, error)
}
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/ddl/notifier/store.go
```diff
@@ -106,34 +122,75 @@ func (t *tableStore) DeleteAndCommit(
...
+	se                  *sess.Session
+	sqlTemplate         string
+	maxReturnedDDLJobID int64
+	maxReturnedSubJobID int64
+}
+
+// Read implements ListResult interface.
+func (r *listResult) Read(changes []*SchemaChange) (int, error) {
+	if r.maxReturnedDDLJobID == 0 && r.maxReturnedSubJobID == 0 {
+		err := r.se.Begin(r.ctx)
```
tiancaiamao, line 158: Is there a commit or rollback?

tiancaiamao: This API design looks scary to me.
Maybe you can use some like 

````
type CloseFn func()
List(ctx context.Context, se *sess.Session) (ListResult, CloseFn)
```

And wrap the session rollback to `CloseFn` 

If you just use this it doesn't help: 

```
res, _ = List()
```

but if you use this and forget to handle the defer, then lint check may remind you variable `done` is not used 

```
res, done := List() 
// defer done()
```

I'm not sure this work, but maybe you can take a try.

pkg/ddl/notifier/subscribe.go
```diff
@@ -166,61 +166,82 @@ func (n *DDLNotifier) start() {
...
-				if !goerr.Is(err2, ErrNotReadyRetryLater) {
-					logutil.Logger(ctx).Error("Error processing change",
-						zap.Int64("ddlJobID", change.ddlJobID),
-						zap.Int64("multiSchemaChangeSeq", change.multiSchemaChangeSeq),
-						zap.Stringer("handler", handlerID),
-						zap.Error(err2))
+	changes := make([]*SchemaChange, ProcessEventsBatchSize)
+
+	for {
+		count, err2 := result.Read(changes)
```
tiancaiamao, line 197: Why not let read return `[]*SchemaChange` directly?
Return `count` and use `changes[:count]` is just a unnecessary detour

lance6716: It's like an `io.Reader` calling convention. Caller can reuse the previous results. Because we are in 1M tables project, I'm afraid that returning 1M `[]*SchemaChange` (it also has internal pointers like `TableInfo`) will add pressure to GC. 

Though it's less common, I think this package / function is the only caller of `ListResult.Read`. Other developers will not modify this package and will not be confused by it.

However I notice that I use `json.Unmarshal` on an non-empty object, not sure the leftover of the object will be overwritten or causes some problems. I'll check it tomorrow.

Rustin170506: What was your test result?

tiancaiamao: >　pass-in slices so caller can reuse the previous results

Yes, I mean the same thing, just that the code can be write as:

```
func f(changes []*SchemaChange) []*SchemaChange {
     ret := changes[:0]
     ret = append(ret, xxx)
     return ret
}
```

Anyway, this is just some personal taste, so you can choose to accept it or not.

tiancaiamao (approved):

pkg/ddl/notifier/subscribe.go
lance6716: > What was your test result?

I didn't test it yet 😂 I hope other colleague can help check it. Theoretically there's no big query now.

> func f(changes []*SchemaChange) []*SchemaChange {

In this new signature, the batch size of one `Read` is not obvious when input is `nil`, or it seems like an `append` where the slice will be grown and batch size is not be limited. Your suggestion is like `EncodeBytes` but here we want to control the batch size as well as reuse buffer.

pkg/ddl/notifier/store.go
```diff
@@ -106,34 +126,85 @@ func (t *tableStore) DeleteAndCommit(
...
+	for i, row := range rows {
+		if changes[i] == nil {
+			changes[i] = new(SchemaChange)
+		}
+		if changes[i].event == nil {
+			changes[i].event = new(SchemaChangeEvent)
+		}
+		if changes[i].event.inner == nil {
+			changes[i].event.inner = new(jsonSchemaChangeEvent)
+		}
```
Rustin170506, line 192: How about:
```go
// Add constructor function
func NewSchemaChange() *SchemaChange {
    return &SchemaChange{
        event: &SchemaChangeEvent{
            inner: &jsonSchemaChangeEvent{},
        },
    }
}

// Replace existing code
if changes[i] == nil {
    changes[i] = NewSchemaChange()
}
```

Rustin170506 (approved):
Better to verify if this can fix the memory consumption issue. 

Thanks! :shipit:

/hold

pkg/ddl/notifier/store.go
Rustin170506: Or do you mean there is a case that `changes[i]` is not nil, but the event or the inner is nil?

lance6716: just to improve robustness, I don't want to assume caller uses some kind of `SchemaChange`.

I'll test the OOM problem soon

lance6716:
/unhold

due to test resource limit, I can't create a test cluster before the coming version release. So I plan to test the effect later, and don't cherry-pick to v8.5.1 before we get the test result

PR #58401, opened 2024-12-19
planner: add more generated columns into update list

### LINKED ISSUE #58400: `UPDATE` and `INSERT ON DUPLICATE KEY` doesn't update index containing generated column
## Bug Report

Case1:

```mysql

CREATE TABLE t3 (
    id int,
    time1 datetime DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
    time2 datetime GENERATED ALWAYS AS (date_add(time1, interval 1 day)) VIRTUAL,
    KEY idx (id, time2)
);

insert into t3 set id = 1;

update t3 set id = 2;

admin check table t3;
```

Case2:

```sql
drop table tmv;
CREATE TABLE tmv (
  J1 json, J2 json GENERATED ALWAYS AS (j1) VIRTUAL,
  UNIQUE KEY i1 ((cast(j1 as signed array))),
  KEY i2 ((cast(j2 as signed array)))
);

insert into tmv set j1 = '[1]';

insert ignore into tmv set j1 = '[1]' on duplicate key update j1 = '[2]';

admin check table tmv;
```

Both `admin check table` passed.

```
ERROR 8223 (HY000): data inconsistency in table: t3, index: idx, handle: 1, index-values:"handle: 1, values: [KindInt64 2 KindMysqlTime 2024-12-20 10:33:33]" != record-values:"handle: 1, values: [KindInt64 2 KindMysqlTime 2024-12-20 10:33:38]"
                                    
ERROR 8134 (HY000): data inconsistency in table: tmv, index: i2, col: _V$_i2_0, handle: "1",
 index-values:"KindInt64 1" != record-values:"KindMysqlJSON [2]", compare err:<nil> 
```

joechenrh:

Issue Number: close #58400

Problem Summary:

When building `UpdateLists` for `update` statements, the criteria for determining whether to update generated columns is that their dependent columns are changed.

But here we wrongly ignore the on-update now columns since they may be implicitly updated. So some generated columns may not be correctly updated.

When checking the dependent columns, we also check whether it has OnUpdateNow flag.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/logical_plan_builder.go
```diff
@@ -5823,20 +5823,23 @@ func (b *PlanBuilder) buildUpdateLists(ctx context.Context, tableList []*ast.Tab
...
 			dependentColumns := expression.ExtractDependentColumns(newExpr)
-			var isModified bool
+			var mayModified bool
 			for _, col := range dependentColumns {
-				if dependentColumnsModified[col.UniqueID] {
-					isModified = true
+				colTp := col.GetType(b.ctx.GetExprCtx().GetEvalCtx()).GetFlag()
+				// If any of the dependent column has on-update-now flag,
+				// this virtual generated column may be modified too.
+				if mysql.HasOnUpdateNowFlag(colTp) || dependentColumnsModified[col.UniqueID] {
```
AilinKid, line 5833: nice catch!

pkg/planner/core/planbuilder.go
```diff
@@ -3881,10 +3886,14 @@ func (b *PlanBuilder) resolveGeneratedColumns(ctx context.Context, columns []*ta
...
 			continue
 		}
+		// There may be chain dependencies between columns,
+		// so we need to add new columns into onDups.
 		for dep := range column.Dependences {
 			if _, ok := onDups[dep]; ok {
 				assign := &expression.Assignment{Col: colExpr, ColName: column.Name, Expr: expr}
 				igc.OnDuplicates = append(igc.OnDuplicates, assign)
+				// onDups use lower column name, see Insert.resolveOnDuplicate
+				onDups[column.Name.L] = struct{}{}
```
AilinKid, line 3896: Have we set the order of columns following the asc dependency?
`for _, column := range columns`
cuz we only set the onDups for the later coming columns, those prev ones couldn't see it any more.

joechenrh: > cuz we only set the onDups for the later coming columns, those prev ones couldn't see it any more.

Yes, you are right. After testing, I found that this situation may be possible to happen.

For `CREATE TABLE`, a generated column can't depend on columns after it.

But for `ALTER TABLE ADD COLUMN`, we can use `AFTER` keyword to force the new generated column inserted before its dependent column and get a table like:

```sql
mysql> show create table t1;
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                              |
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| t1    | CREATE TABLE `t1` (
  `id` int DEFAULT NULL,
  `col2` int GENERATED ALWAYS AS (`col1` + 1) VIRTUAL,
  `col1` int DEFAULT NULL
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

Let me think about how to modify it.

joechenrh: There are something wrong in the previous comment.

In `resolveGeneratedColumns`, we only process generated columns. Non-generated columns are already processed and stored in `onDups`:

https://github.com/pingcap/tidb/blob/43f2fb91d740aaf45774d99b71554eead12009cc/pkg/planner/core/planbuilder.go#L3859

Since generated columns can only depends on columns prior to it, we don't need to handle the order of columns.

```sql
mysql> alter table t1 add column col3 int generated always as (col2 + 1) virtual after id;                                                                                                    
ERROR 3107 (HY000): Generated column can refer only to generated columns defined prior to it.
```

AilinKid: got
```
// verifyColumnGeneration is for CREATE TABLE, because we need verify all columns in the table.
func verifyColumnGeneration(colName2Generation map[string]columnGenerationInDDL, colName string) error {
	attribute := colName2Generation[colName]
	if attribute.generated {
		for depCol := range attribute.dependences {
			attr, ok := colName2Generation[depCol]
			if !ok {
				err := dbterror.ErrBadField.GenWithStackByArgs(depCol, "generated column function")
				return errors.Trace(err)
			}
			if attr.generated && attribute.position <= attr.position {
				// A generated column definition can refer to other
				// generated columns occurring earlier in the table.
				err := dbterror.ErrGeneratedColumnNonPrior.GenWithStackByArgs()
				return errors.Trace(err)
			}
		}
	}
	return nil
}
```
seems create table has banned the case, when a gen col ref a later coming gen col, so they are asc ordered already from table schema columns

AilinKid (approved):

Issue #58361, opened 2024-12-17
Unify OR/Union type IndexMerge code paths to expand the applicable scenarios of existing capabilities


time-and-fate:
## Background & Problem

During the development of several previous features, the logic of generating OR/Union type IndexMerge paths has diverged into several different code paths:
1. The code path evolved from the original IndexMerge code
    - It only handles non-MV indexes.
    - When there are multiple indexes to choose from, it can consider the required Sort property and choose the correct index to satisfy that.
    - Now the entry is `generateIndexMerge4NormalIndex()`.
2. The code path introduced to support the MV index
    - It's specially for handling the MV index.
    - It can generate IndexMerge paths that involve only one MV index.
    - For filters like `... AND (... OR ...) AND ...`, it can consider the filters connected by `AND` so the user doesn't need to expand the expression manually to match the index.
    - Now the entry is `generateIndexMerge4MVIndex()`
3. The code path later introduced to support more complex patterns involving the MV index.
    - It's mostly similar to 2, but it can generate IndexMerge paths involving multiple MV and non-MV indexes.
    - Now the entry is `generateIndexMerge4ComposedIndex()`.

<details>
  <summary>Links to some previous features</summary>
  <ul>
  <li>Initial support for the MV index (the code path 2): #40191</li>
  <li>Support using different MV indexes and non-MV indexes in an IndexMerge path (the code path 3): #47759</li>
  <li>Support satisfying Sort property for non-MV indexes (enhancement on code path 1): #48359</li>
  <li>Support collecting the top level AND list for filters like `... AND (... OR ...) ...` for paths involving MV indexes (enhancement on code path 2 and 3): #51778</li>
  </ul>
</details>

Apparently, it causes code repetitions. If you go over the code, you'll find there are many similar code snippets and there are also many structural similarities.

Besides, for some later enhancements, we introduced them to only 1 or 2 code paths. They can't benefit the other code paths.

For example, for the non-MV index, sometimes the user needs to expand the `... AND (... OR ...)` filter manually to match indexes so that the optimizer can choose an IndexMerge plan.

```sql
create table t(a int, b int, c int, d int,
index iab(a,b),
index iac(a,c),
index iad(a,d));
explain select /*+ use_index_merge(t) */ * from t where a = 1 and (b = 2 or c = 3 or d = 4);
show warnings;
explain select /*+ use_index_merge(t) */ * from t where (a = 1 and b = 2) or (a = 1 and c = 3) or (a = 1 and d = 4);
[...]

PR #58448, opened 2024-12-22
planner: switch to `AccessPath.PartialAlternativeIndexPaths` for IndexMerge OR path with MV index

time-and-fate:

Issue Number: ref https://github.com/pingcap/tidb/issues/58361

As the title says, the code path that generates IndexMerge OR path for MV indexes now generates `AccessPath.PartialAlternativeIndexPaths` instead of `AccessPath.PartialIndexPaths`.

Specifically:
- In `pkg/planner/core/indexmerge_unfinished_path.go`:
  - Now we store all alternatives instead of deciding on the best one.
  - Similarly, the `needSelection` checking related information is stored in `AccessPath.KeepIndexMergeORSourceFilter` now instead of being decided here.
  - Besides, a unified util function for estimation is extracted.
- In `pkg/planner/core/find_best_task.go`:
  - There's already logic that chooses the best one from the alternatives so we don't need implement it again.
  - The MV index should not be affected by the "avoid choosing the same index for partial paths" logic, so some extra code is still needed.
  - Use the `AccessPath.KeepIndexMergeORSourceFilter` information from the last step.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/indexmerge_unfinished_path.go
```diff
@@ -368,39 +366,41 @@ func buildIntoAccessPath(
...
-			alternativesForORBranch, func(a, b alternative) int {
-				return cmpAlternativesByRowCount(a.paths, b.paths)
-			})
-		if len(bestAlternative.paths) == 0 {
-			return nil
-		}
-		// Succeeded to get valid path(s) for this partial path.
-		partialPaths = append(partialPaths, bestAlternative.paths...)
-		needSelectionGlobal = needSelectionGlobal || bestAlternative.needSelection
+	// 2. TODO
```
Rustin170506, line 385: It seems you have already handled the table filters below. So what is this TODO for?

Rustin170506 (commented):
Thanks!

pkg/planner/core/indexmerge_unfinished_path.go
time-and-fate: Oh, I should add this in the next PR.
Some logic of other code paths will be moved here.

Rustin170506 (approved):
Thanks! :shipit:

pkg/planner/core/indexmerge_unfinished_path.go
```diff
@@ -368,39 +366,41 @@ func buildIntoAccessPath(
...
-	if !needSelectionGlobal {
-		tableFilter = slices.Delete(tableFilter, orListIdxInAllConds, orListIdxInAllConds+1)
+	// For estimation, we need the decided partial paths. So we use a simple heuristic to choose the partial paths by
+	// comparing the row count just for estimation here.
+	pathsForEstimate := make([]*util.AccessPath, 0, len(allAlternativePaths))
+	for _, oneORBranch := range allAlternativePaths {
+		pathsWithMinRowCount := slices.MinFunc(oneORBranch, cmpAlternativesByRowCount)
+		pathsForEstimate = append(pathsForEstimate, pathsWithMinRowCount...)
 	}
+	possiblePath.CountAfterAccess = estimateCountAfterAccessForIndexMergeOR(ds, pathsForEstimate)
```
AilinKid, line 401: from the theoretical, we don't need to compute the count after acess at this point cuz it's still undetermined. but its seems to be used to update ds‘s stats introduced by @jiangtao
image
by now, it's fine, but not that elegant

AilinKid (approved):
LGTM

pkg/planner/core/indexmerge_unfinished_path.go
time-and-fate: Yes. Besides this, the calculation of `DataSource.AccessPathMinSelectivity` also needs this field to be correctly set.

AilinKid: make sense

PR #58474, opened 2024-12-23
executor: fix null value update for not null virtual generated columns

### LINKED ISSUE #58472: insert `null` into not-null virtual generated column leads to wrong index record
## Bug Report

The following problem only exists in **virtual generated** columns.

```sql
drop table t3;
create table t3(
    id int primary key,
    c1 varchar(16) default null,
    c2 varchar(16) GENERATED ALWAYS AS (concat(c1, c1)) VIRTUAL NOT NULL,
    KEY idx (c2)
);

insert into t3(id, c1) values(1, "aaaa");
insert ignore into t3 set id = 1, c1 = "bbbb" on duplicate key update id = 2, c1 = null;
insert into t3 set id = 2, c1 = "cccc" on duplicate key update c1 = "dddd";
```

No error, or get the error `ERROR 1048 (23000): Column 'c2' cannot be null`.

If we follow the MySQL implement, the result should be the latter one.

We get none of the expected results.

```
ERROR 8141 (HY000): assertion failed: key: 7480000000000000715f69800000000000000100038000000000000002, assertion: Exist, start_ts: 454804733357457409, existing start ts: 0, existing commit ts: 0
```

joechenrh:

Issue Number: close #58472

Problem Summary:

This PR solves a corner case for **virtual generated column**.

When we insert a `null` value into not-null column, `HandleBadNull` will do the job to convert `kindNULL` datum to zero value for given column type.

But when we insert a duplicated record, the datum of **not-null virtual generated column **may still be `kindNULL`. If we try to update this record, we will get assertion failed.

As mentioned in the description, one solution is to make the generated column and it's index updated correctly.

Another soution is to align the behavior with MySQL, that is:
- `ON DUPLICATE KEY` can insert `NULL` into not null generated column, while directly insert `NULL` is not allowed.
- These rows can't be updated and return error like `ERROR 1048 (23000): Column 'c2' cannot be null`, we can only delete these rows.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/insert_test.go
```diff
@@ -757,3 +757,22 @@ func TestInsertDuplicateToGeneratedColumns(t *testing.T) {
...
+	tk.MustExec(`create table t3(
+		id int primary key,
+		c1 varchar(16) default null,
+		c2 varchar(16) GENERATED ALWAYS AS (concat(c1, c1)) VIRTUAL NOT NULL,
+		KEY idx (c2)
+	)`)
+	tk.ExecToErr(`insert into t3 set id = 2, c1 = null`)
+	tk.MustExec(`insert into t3(id, c1) values(1, "aaaa")`)
+	tk.MustExec(`insert ignore into t3 set id = 1, c1 = "bbbb" on duplicate key update id = 2, c1 = null`)
+	tk.ExecToErr(`insert into t3 set id = 2, c1 = "cccc" on duplicate key update c1 = "dddd"`)
```
YangKeao, line 778: This behavior is really strange at first glance (though understandable if considering the implementation). As long as it's compatible with MySQL, it's LGTM.

image

joechenrh: Yeah, it's indeed strange. So as I said in the description, we can fix it in another way, to make the insert stmt succeed. I just no sure which is better 🤔

pkg/executor/write.go
```diff
@@ -235,6 +236,15 @@ func updateRecord(
 		}
 	}
 
+	// Null value can be inserted into not nullable generated column by ON DUPLICATE KEY UPDATE.
+	// But we don't allow to update this record later. So we return an error for null value.
+	for i, col := range t.Cols() {
+		if col.IsGenerated() && oldData[i].IsNull() &&
```
wjhuang2016: Do we need to consider virtual or stored generated column?

joechenrh: No, STORED column has different behavior, same for MySQL

```SQL
mysql> insert into t3(id, c1) values(1, "aaaa");

mysql> insert ignore into t3 set id = 1, c1 = "bbbb" on duplicate key update id = 2, c1 = null;
Query OK, 2 rows affected, 1 warning (0.00 sec)

mysql> show warnings;
+---------+------+----------------------------+
| Level   | Code | Message                    |
+---------+------+----------------------------+
| Warning | 1048 | Column 'c2' cannot be null |
+---------+------+----------------------------+
1 row in set (0.00 sec)

mysql> insert into t3 set id = 2, c1 = "cccc" on duplicate key update c1 = "dddd";
Query OK, 2 rows affected (0.01 sec)
```

You can see we will get warning after first `on duplicate key update`, and the second update can succeed.

~~I guess MySQL postpones the check of vritual generated column to update, not insert.~~

Overall, this behavior is indeed very strange.

wjhuang2016 (approved):

PR #58539, opened 2024-12-25
ttl: set the job history status to `cancelled` if it's removed in GC and it's still running

### LINKED ISSUE #58510: `timer` will keep waiting for job if it exists in history but not exists in status and the owner stopped.
## Bug Report

1. Create a TTL table.
2. After the entry in both `tidb_ttl_table_status` and `tidb_ttl_job_history` are created, drop the table.
3. Make the owner fault

After a while, the owner will remove the job (as the logic in `(*JobManager).checkNotOwnJob()`), and nobody will update the history anymore.

Then the timer will always find the job because the entry in `tidb_ttl_job_history` still exists, and wait for it to finish.

Don't spawn new goroutine to wait for these not exist tables.

Keeps waiting for the TTL job for non-exist table to finish.

YangKeao:

Issue Number: close #58510

Problem Summary:

The job history is not updated after removing the table status. Therefore, the timer adapter will still believe there is a job running for it.

Update the status to `cancelled` after removing the table status.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ttl/ttlworker/job_manager.go
+++ b/pkg/ttl/ttlworker/job_manager.go
@@ -66,6 +66,11 @@ const taskGCTemplate = `DELETE task FROM
 const ttlJobHistoryGCTemplate = `DELETE FROM mysql.tidb_ttl_job_history WHERE create_time < CURDATE() - INTERVAL 90 DAY`
 const ttlTableStatusGCWithoutIDTemplate = `DELETE FROM mysql.tidb_ttl_table_status WHERE (current_job_status IS NULL OR current_job_owner_hb_time < %?)`
 
+// don't remove the rows for non-exist tables directly. Instead, set them to cancelled. In some special situations, the TTL job may still be able
+// to finish correctly. If that happen, the status will be updated from 'cancelled' to 'finished' in `(*ttlJob).finish`
+const ttlJobHistoryGCNonExistTableTemplate = `UPDATE mysql.tidb_ttl_job_history SET status = 'cancelled'
+	WHERE table_id NOT IN (SELECT table_id FROM mysql.tidb_ttl_table_status) AND status = 'running'`
+
 var timeFormat = time.DateTime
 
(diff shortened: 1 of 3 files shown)


pkg/ttl/ttlworker/job_manager.go
```diff
@@ -66,6 +66,11 @@ const taskGCTemplate = `DELETE task FROM
 const ttlJobHistoryGCTemplate = `DELETE FROM mysql.tidb_ttl_job_history WHERE create_time < CURDATE() - INTERVAL 90 DAY`
 const ttlTableStatusGCWithoutIDTemplate = `DELETE FROM mysql.tidb_ttl_table_status WHERE (current_job_status IS NULL OR current_job_owner_hb_time < %?)`
 
+// don't remove the rows for non-exist tables directly. Instead, set them to cancelled. In some special situations, the TTL job may still be able
+// to finish correctly. If that happen, the status will be updated from 'cancelled' to 'finished' in `(*ttlJob).finish`
+const ttlJobHistoryGCNonExistTableTemplate = `UPDATE mysql.tidb_ttl_job_history SET status = 'cancelled'
```
YangKeao, line 71: Maybe it'd be better to not remove the `tidb_ttl_table_status` if the current state is running? And let the job owner to cancel and remove the status / update the history.

I'm not sure. This solution is simpler, but a little ugly :facepalm: .

lcwangchao: Yes, I think always takeover unfinished job is better. We can optimize it in future PRs.

lcwangchao (approved):

pkg/ttl/ttlworker/job_manager.go
YangKeao: It's good. I'll do this change later in another PR.

Defined2014 (approved):

PR #58572, opened 2024-12-27
import-into: support running with 1 thread when using global sort

joccau:

Issue Number: close https://github.com/pingcap/tidb/issues/58680

Problem Summary:

- The feature of `import into` can be run with the option of thread=1
- Simplify the code.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
The feature of `import into` with the options `global-sort` can be run at the server that only consumes 1 core resource.
```


D3Hunter:
>  Manual test (add detailed scripts or steps below)

so how you test it manually?

lance6716:
> only consume 1 core resource.

If the goal is during the whole IMPORT INTO only ~100% CPU is used, I think your internal test report can't achieve it.

pkg/disttask/importinto/proto.go
```diff
@@ -158,19 +158,6 @@ func (t *importStepMinimalTask) String() string {
 	return fmt.Sprintf("chunk:%s:%d", t.Chunk.Path, t.Chunk.Offset)
 }
 
-// Chunk records the chunk information.
-type Chunk struct {
```
D3Hunter, line 162: why move it, lightning doesn't share this

pkg/disttask/framework/scheduler/scheduler.go
```diff
@@ -504,9 +504,10 @@ func (s *BaseScheduler) scheduleSubTask(
 	adjustedEligibleNodes := s.slotMgr.adjustEligibleNodes(eligibleNodes, task.Concurrency)
 	var size uint64
 	subTasks := make([]*proto.Subtask, 0, len(metas))
+	randPos := rand.Intn(len(adjustedEligibleNodes))
 	for i, meta := range metas {
-		// we assign the subtask to the instance in a round-robin way.
-		pos := i % len(adjustedEligibleNodes)
+		// we assign the subtask to the instance in a round-robin way at the random start position.
+		pos := (i + randPos) % len(adjustedEligibleNodes)
```
D3Hunter, line 510: we design it this way to pack subtasks onto as less number of nodes as possible, in this way we can make sure later task of higher concurrency can be run, to maximize resource usage

suppose we have 2 nodes, and task A/B of concurrency 4 each with one subtask, and task C of concurrency 8
currently, we will scheduler A/B on the first node, and C can run on the second
but with random node, A/B might scheduled on the same node, C cannot be run

D3Hunter: see https://github.com/pingcap/tidb/pull/59033

pkg/disttask/importinto/proto.go
joccau: move it to resolve cycle import problem

pkg/disttask/framework/scheduler/scheduler.go
joccau: Get it~, I will revert these part of code.

D3Hunter:
/hold

this PR and related test hasn't reproduced the OOM we meet before, so there is no optimization for memory

pkg/lightning/checkpoints/checkpoints.go
```diff
@@ -270,6 +270,19 @@ func (key *ChunkCheckpointKey) less(other *ChunkCheckpointKey) bool {
 	}
 }
 
+// Chunk records the chunk information.
+type Chunk struct {
```
D3Hunter, line 274: please move to `importer` pkg. this `checkpoints` pkg is mostly specific to lightning

D3Hunter (commented):
rest lgtm

pkg/lightning/checkpoints/checkpoints.go
joccau: Modified it.

D3Hunter (approved):

PR #58588, opened 2024-12-27
expression: Fixing Expression Rewrite Bug in Predicate Push Down for NULLIF with Type Mismatch

### LINKED ISSUE #57647: Scalar function causes database crash.
## Bug Report

1. schema.

```sql
create table `t0`
(
    `vkey` integer,
    `pkey` integer,
    `c0`   integer
);

insert into `t0`
values (1, 2, 3);
```

2. sql statement.

```sql
select *
from `t0`
where (nullif(
    3 ^ 10 & (abs(-50)) ,
    round(case when (((`t0`.`c0`) >= 1) or null) then 91 else 86 end)
    )) in (select `vkey` from `t0` where false);
```

MySQL and TIDB have different execution results in the above case.

The normal result in MySQL 8:

```SQL
mysql> select *
    -> from `t0`
    -> where (nullif(
    ->     3 ^ 10 & (abs(-50)) ,
    ->     round(case when (((`t0`.`c0`) >= 1) or null) then 91 else 86 end)
    ->     )) in (select `vkey` from `t0` where false);
Empty set (0.00 sec)
```

TiDB should also return empty result, but failed.

In TiDB v8.1.1:

```sql
MySQL [test]> select *
    -> from `t0`
    -> where (nullif(
    ->     3 ^ 10 & (abs(-50)) ,
    ->     round(case when (((`t0`.`c0`) >= 1) or null) then 91 else 86 end)
    ->     )) in (select `vkey` from `t0` where false);
ERROR 1105 (HY000): interface conversion: expression.Expression is *expression.ScalarFunction, not *expression.Column
```

TiDB crash when execute the same SQL statement.

tidb v8.1.1

dash12653:

Issue Number: close #57647

Problem Summary:

The sql could be simplified as:
```sql
select *
from `t0`
where (
  nullif(3^1, case when (`c0`) then 91 end)
) in (select 1 from `t0`);
```

When rewriting the expression `nullif(3^1, case when (c0) then 91 end)`, it is transformed into:
`if(equals(3^1, case when (c0) then 91 end), null, 3^1)`.

When building this function, the` RetType` of `null` is inferred based on the type of `3^1`, and a new `RetType` is set.
https://github.com/pingcap/tidb/blob/a82a3b41e7dee9b110dbac2ec18e7222f85ddefb/pkg/expression/builtin_control.go#L713-L720
Then, the SQL:

`where (nullif(3^1, case when (c0) then 91 end)) in (select 1 from t0);`
is rewritten to an inner join with otherConditions as:

`nullif(3^1, case when (c0) then 91 end) = 1.`

During `predicate push down`, it will try moving `otherConditions` into `equal conditions`: 
https://github.com/pingcap/tidb/blob/b11cce72848142c279842061285ed457846340c0/pkg/planner/core/operator/logicalop/logical_join.go#L1564-L1572

But when rewriting the `equal condition`, the `RetType of null` has already been reset based on the earlier inference. So `re-inferring` of the `RetType` of null and 3^1 will not match previous result, and the type finally becomes `decimal`.
https://github.com/pingcap/tidb/blob/a82a3b41e7dee9b110dbac2ec18e7222f85ddefb/pkg/expression/builtin_control.go#L279-L289

The constant 1 will be wrapped with a `(cast as decimal)`, making it no longer a column, which ultimately leads to the error.

I modified the `ifInferType4ControlFuncs()` method to determine whether a value is null. If a constant's `value.`k is null, it will also be treated as null.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


XuHuaiyu (approved):

pkg/expression/builtin_control.go
```diff
@@ -255,7 +255,8 @@ func InferType4ControlFuncs(ctx BuildContext, funcName string, args ...Expressio
 	nullFields := make([]*types.FieldType, 0, argsNum)
 	notNullFields := make([]*types.FieldType, 0, argsNum)
 	for i := range args {
-		if args[i].GetType(ctx.GetEvalCtx()).GetType() == mysql.TypeNull {
+		ci, ok := args[i].(*Constant)
+		if args[i].GetType(ctx.GetEvalCtx()).GetType() == mysql.TypeNull || (ok && ci.Value.Kind() == types.KindNull) {
```
windtalker, line 259: I've a little concern about this change, sometimes the null constant do have its own type, for example, user can use `cast(null as xxx)` to create a null constant with specific type, but in this pr, it ignores the null constant type, which I think will cause some inconsistencies with MySQL behavior.
For example, consider the following sql
`select case when 0 then cast(null as char(100)) else 1 end;`
The return type is `char` before this pr, and it will return `bigint` after this pr, and in mysql, the return type is `char`.

windtalker: And I will hold this pr in case it would be merged accidentally.

dash12653:
> @dash12653 I have some step-by-step debug on this issue, and I think the `null` constant value constructed when rewrite `nullif` to `if` has the right type(typeNull), and inside the first `InferType4ControlFuncs`, the `ret` type is inferred as `unsigned int`, so in `newBaseBuiltinFuncWithFieldTypes`, the first argument(the constant `null`) is expected to be converted to `unsigned longlong` type, however, `WrapWithCastAsInt` does not respect the unsigned flag, the `null` constant is converted to `longlong` type instead. In the second `InferType4ControlFuncs`, since the common type of `longlong` and `unsigned longlong` is `decimal`, the `ret` type is inferred as `decimal`. I think the root cause of this issue is that `WrapWithCastAsInt` does not respect unsigned flag, we should fix this issue by let `WrapWithCastAsInt` be aware of the unsigned flag of the target type.

Thank you for your detailed analysis, I really appreciate your time and guidance!

Based on your analysis, I’m considering whether it might be better to modify `WrapWithCastAsInt` to accept `argTps[i]` as an additional parameter, so it can handle the `unsigned flag` directly. Do you think this would be a reasonable solution?

Also, I did a check for `null values` at the end of each iteration, as shown below:

Do you think this approach would be appropriate?

Thanks again for your help.
image

windtalker:
> eck for `null values` at the end of each iter

I'm ok with `modify WrapWithCastAsInt to accept argTps[i] as an additional parameter`. But I'm a little confused that why you need to add special check for null?

dash12653:
> > eck for `null values` at the end of each iter
> 
> I'm ok with `modify WrapWithCastAsInt to accept argTps[i] as an additional parameter`. But I'm a little confused that why you need to add special check for null?

Never mind, I realize I misunderstood the situation. The null value check was unnecessary and was a mistake. I could try modify `WrapWithCastAsInt` to handle the unsigned flag.

dash12653:
Hi @windtalker, after looking into it, I found that many places are doing a unified cast for different argument types(as shown below), and modifying `WrapWithCastAsInt` would require changing the signatures of other methods like `WrapWithCastAsReal`, `WrapWithCastAsString`, etc., even though they don’t actually use the additional parameter.
https://github.com/pingcap/tidb/blob/6180fb7358d216af3407a262cabb1b5ce322bd9f/pkg/expression/aggregation/base_func.go#L439-L466
I was thinking, instead of modifying the existing functions, could we write another `wrapCastAsInt` function who takes the target type as a parameter, and then call it in `newBaseBuiltinFuncWithTp`? This would allow us to handle the target type without introducing unnecessary changes. 
image

Thanks again for your time!

windtalker:
> Hi @windtalker, after looking into it, I found that many places are doing a unified cast for different argument types(as shown below), and modifying `WrapWithCastAsInt` would require changing the signatures of other methods like `WrapWithCastAsReal`, `WrapWithCastAsString`, etc., even though they don’t actually use the additional parameter.
> 
> https://github.com/pingcap/tidb/blob/6180fb7358d216af3407a262cabb1b5ce322bd9f/pkg/expression/aggregation/base_func.go#L439-L466
> 
> 
> I was thinking, instead of modifying the existing functions, could we write another `wrapCastAsInt` function who takes the target type as a parameter, and then call it in `newBaseBuiltinFuncWithTp`? This would allow us to handle the target type without introducing unnecessary changes.
> image
> Thanks again for your time!

Looks like this is one of the technical debt in TiDB :( 
I think all the caller of `WrapWithCastAsInt` should provide extra information of whether the target type should be signed or unsigned. In fact, `WrapWithCastAsTime` has the same problem so `WrapWithCastAsTime` has the third parameter to indicate the target type is `Datetime` or `Timestamp`.  I suggest to add an extra `tp *types.FieldTyp` parameter to `WrapWithCastAsInt`, if the caller does not have this information, then just pass nil, and inside `WrapWithCastAsInt`, we can handle the unsigned flag if input `tp` is not nil, and if the input `tp` is nil, then just keep current code logic. 

For the case you metioned that `many places are doing a unified cast for different argument types(as shown below), and modifying WrapWithCastAsInt would require changing the signatures of other methods like WrapWithCastAsReal, WrapWithCastAsString, etc., even though they don’t actually use the additional parameter`, I think you can refer to the current usage of `WrapWithCastAsTime`.

dash12653:
> > Hi @windtalker, after looking into it, I found that many places are doing a unified cast for different argument types(as shown below), and modifying `WrapWithCastAsInt` would require changing the signatures of other methods like `WrapWithCastAsReal`, `WrapWithCastAsString`, etc., even though they don’t actually use the additional parameter.
> > https://github.com/pingcap/tidb/blob/6180fb7358d216af3407a262cabb1b5ce322bd9f/pkg/expression/aggregation/base_func.go#L439-L466
> > 
> > I was thinking, instead of modifying the existing functions, could we write another `wrapCastAsInt` function who takes the target type as a parameter, and then call it in `newBaseBuiltinFuncWithTp`? This would allow us to handle the target type without introducing unnecessary changes.
> > image
> > Thanks again for your time!
> 
> Looks like this is one of the technical debt in TiDB :( I think all the caller of `WrapWithCastAsInt` should provide extra information of whether the target type should be signed or unsigned. In fact, `WrapWithCastAsTime` has the same problem so `WrapWithCastAsTime` has the third parameter to indicate the target type is `Datetime` or `Timestamp`. I suggest to add an extra `tp *types.FieldTyp` parameter to `WrapWithCastAsInt`, if the caller does not have this information, then just pass nil, and inside `WrapWithCastAsInt`, we can handle the unsigned flag if input `tp` is not nil, and if the input `tp` is nil, then just keep current code logic.
> 
> For the case you metioned that `many places are doing a unified cast for different argument types(as shown below), and modifying WrapWithCastAsInt would require changing the signatures of other methods like WrapWithCastAsReal, WrapWithCastAsString, etc., even though they don’t actually use the additional parameter`, I think you can refer to the current usage of `WrapWithCastAsTime`.

Thanks for your input. It's really detailed. I’ll implement adding the parameter to `WrapWithCastAsInt` and see how it goes.

pkg/expression/builtin_cast.go
```diff
@@ -2550,6 +2550,9 @@ func WrapWithCastAsInt(ctx BuildContext, expr Expression) Expression {
 	tp.SetDecimal(0)
 	types.SetBinChsClnFlag(tp)
 	tp.AddFlag(expr.GetType(ctx.GetEvalCtx()).GetFlag() & (mysql.UnsignedFlag | mysql.NotNullFlag))
+	if targetType != nil {
+		tp.AddFlag(targetType.GetFlag() & mysql.UnsignedFlag)
+	}
```
windtalker, line 2555:
```suggestion
    // inherit NotNullFlag from source type
    tp.AddFlag(expr.GetType(ctx.GetEvalCtx()).GetFlag() & mysql.NotNullFlag)
    if targetType == nil {
        // inherit UnsignedFlag from source type if targetType is nil
        tp.AddFlag(expr.GetType(ctx.GetEvalCtx()).GetFlag() & mysql.UnsignedFlag)
    } else {
        // otherwise set UnsignedFlag based on targetType
		tp.AddFlag(targetType.GetFlag() & mysql.UnsignedFlag)
	}
```

pkg/expression/builtin_cast.go
```diff
@@ -2759,7 +2762,9 @@ func TryPushCastIntoControlFunctionForHybridType(ctx BuildContext, expr Expressi
 	var wrapCastFunc func(ctx BuildContext, expr Expression) Expression
 	switch tp.EvalType() {
 	case types.ETInt:
-		wrapCastFunc = WrapWithCastAsInt
+		wrapCastFunc = func(ctx BuildContext, expr Expression) Expression {
+			return WrapWithCastAsInt(ctx, expr, nil)
```
windtalker, line 2766: why not use `WrapWithCastAsInt(ctx, expr, tp)`

pkg/expression/builtin_cast_test.go
```diff
@@ -1242,7 +1242,7 @@ func TestWrapWithCastAsTypesClasses(t *testing.T) {
 	}
 	for i, c := range cases {
 		// Test wrapping with CastAsInt.
-		intExpr := WrapWithCastAsInt(ctx, c.expr)
```
windtalker, line 1245: please add a test that tp is not nil in `WrapWithCastAsInt`

windtalker (approved):
lgtm

windtalker:
@hawkingrei please take a look

hawkingrei (approved):

Issue #57996, opened 2024-12-04
`mysql.analyze_jobs` missed indexes


winoros:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

https://github.com/winoros/tidb/blob/debugging/pkg/statistics/handle/autoanalyze/priorityqueue/interval.go

Some internal SQLs're running on the table `mysql.analyze_jobs`.

But, this table only has one index.
```
CREATE TABLE `analyze_jobs` (
  `id` bigint unsigned NOT NULL AUTO_INCREMENT,
  `update_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  `table_schema` char(64) NOT NULL DEFAULT '',
  `table_name` char(64) NOT NULL DEFAULT '',
  `partition_name` char(64) NOT NULL DEFAULT '',
  `job_info` text NOT NULL,
  `processed_rows` bigint unsigned NOT NULL DEFAULT '0',
  `start_time` timestamp NULL DEFAULT NULL,
  `end_time` timestamp NULL DEFAULT NULL,
  `state` enum('pending','running','finished','failed') NOT NULL,
  `fail_reason` text DEFAULT NULL,
  `instance` varchar(512) NOT NULL COMMENT 'address of the TiDB instance executing the analyze job',
  `process_id` bigint unsigned DEFAULT NULL COMMENT 'ID of the process executing the analyze job',
  PRIMARY KEY (`id`) /*T![clustered_index] CLUSTERED */,
  KEY `update_time` (`update_time`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin AUTO_INCREMENT=901003903
```

As a result, we'll see a lot of full scan in a big cluster testing.
![Image](https://github.com/user-attachments/assets/97997ecb-94bb-4382-9d46-940f7e260006)

### 4. What is your TiDB version? (Required)

master

PR #58134, opened 2024-12-10
session: add indexes for `mysql.analyze_jobs`

Rustin170506:
### What problem does this PR solve?

Issue Number: close https://github.com/pingcap/tidb/issues/57996

Problem Summary:

### What changed and how does it work?

This PR added two indexes for the `mysql.analyze_jobs` to speed up queries.

I also bumped the bootstrap version from 239 to 140.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [x] [Manual test1](https://github.com/pingcap/tidb/pull/58134#issuecomment-2545067361)
- [x] [Manual test2](https://github.com/pingcap/tidb/pull/58134#issuecomment-2545334778)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
为 `mysql.analyze_jobs` 添加索引加速查询
Add some indexes to `mysql.analyze_jobs` to speed up queries
```

PR #58608, opened 2024-12-30
session: skip creating indexes on the analyze_jobs table for older clusters

Rustin170506:

Issue Number: close https://github.com/pingcap/tidb/issues/57996

Problem Summary:

I test https://github.com/pingcap/tidb/pull/58134 again locally to evaluate the performance of creating indexes. For 100k rows, it takes 16 seconds to create the indexes, although it is not that slow, but it still takes some time. So I decided to undo part of this change. We will only create the new indexes for the new cluster. And we do not create the index for the old clusters during the upgrade process.

Normally, for the smaller cluster, this should not be a problem. But for some huge clusters, we can ask users to manually create it instead of blocking the upgrade process.

```sql
> ALTER TABLE mysql.analyze_jobs ADD INDEX `idx_schema_table_state` (`table_schema`, `table_name`, `state`)
[2024-12-30 14:17:40] completed in 5 s 755 ms
> ALTER TABLE mysql.analyze_jobs ADD INDEX `idx_schema_table_partition_state` (`table_schema`, `table_name`, `partition_name`, `state`)
[2024-12-30 14:17:52] completed in 11 s 860 ms
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/session/bootstrap.go
+++ b/pkg/session/bootstrap.go
@@ -1236,17 +1236,13 @@ const (
 	// add modify_params to tidb_global_task and tidb_global_task_history.
 	version239 = 239
 
-	// version 240
-	// Add indexes to mysql.analyze_jobs to speed up the query.
-	version240 = 240
-
 	// Add index on user field for some mysql tables.
-	version241 = 241
+	version240 = 240
 )
 
 // currentBootstrapVersion is defined as a variable, so we can modify its value for testing.
 // please make sure this is the largest version
-var currentBootstrapVersion int64 = version241
+var currentBootstrapVersion int64 = version240
 
 // DDL owner key's expired time is ManagerSessionTTL seconds, we should wait the time and give more time to have a chance to finish it.
 var internalSQLTimeout = owner.ManagerSessionTTL + 15
@@ -1422,7 +1418,6 @@ var (
 		upgradeToVer218,
 		upgradeToVer239,
 		upgradeToVer240,
-		upgradeToVer241,
 	}
 )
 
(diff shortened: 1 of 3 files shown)


Rustin170506 (commented):

winoros (approved):

time-and-fate (approved):

yudongusa (approved):

pkg/session/bootstrap.go
```diff
@@ -3334,28 +3329,10 @@ func upgradeToVer239(s sessiontypes.Session, ver int64) {
...
-	// addAnalyzeJobsSchemaTablePartitionStateIndex adds an index on (table_schema, table_name, partition_name, state) to mysql.analyze_jobs
-	addAnalyzeJobsSchemaTablePartitionStateIndex = "ALTER TABLE mysql.analyze_jobs ADD INDEX idx_schema_table_partition_state (table_schema, table_name, partition_name, state)"
-)
-
 func upgradeToVer240(s sessiontypes.Session, ver int64) {
 	if ver >= version240 {
 		return
 	}
-	doReentrantDDL(s, addAnalyzeJobsSchemaTableStateIndex, dbterror.ErrDupKeyName)
-	doReentrantDDL(s, addAnalyzeJobsSchemaTablePartitionStateIndex, dbterror.ErrDupKeyName)
```
D3Hunter, line 3352: personally, i don't like the idea that for two clusters of same version, but with different schema. It introduce more maintenance burden, if I don't know this PR, and meet a issue and see this difference, I will take it as a bug at first glance.

for a production cluster with so many tables, say 1M, the upgrade duration is quite long in most cases, not just the process of TiDB upgrade, also for rolling upgrade of other components, it's even longer when there are many online traffic, takes hours or even days. I think it's acceptable for this `add-index` to be slower, and it's the tradeoff we have to make

If we can reduce the size of this table, and the index create faster, that would be better, certainly.

Leavrth (approved):

pkg/session/bootstrap.go
Rustin170506: > personally, i don't like the idea that for two clusters of same version, but with different schema. It introduce more maintenance burden, if I don't know this PR, and meet a issue and see this difference, I will take it as a bug at first glance.

Yes. I agree that having different schemas is annoying. But for most users the full table scan is OK. So introducing the prototail risk to slow down the upgrade is not worth it. We don't want to introduce that risk for users that 99% of them don't have the problem.

Rustin170506: I also asked @tangenta, he suggested that it is better not to do this kind of operation for a volumetric table.

D3Hunter: > 99% of them don't have the problem.

99% users don't have 1M tables, so upgrade is fast even with the index. this PR is for the 1%

Rustin170506: /hold

I am not in a hurry with this PR. So let's discuss it further.

Rustin170506: > 
> 99% users don't have 1M tables, so upgrade is fast even with the index. this PR is for the 1%

My point here is that 99% of users do not have 1M tables, so there is no need to add this index for them to bring the potential risk for them.

D3Hunter: IMU, `99% of users do not have 1M tables` -> `99% users don't have 100K rows` -> `add-index very fast` -> `no such risk`

winoros: We may also need to add new indexes on tables like `mysql.stats_histograms`.
This table is related to column count and index count. It's more likely to be a big table.
So the problem will still exist.

I think we can add the related operation to our upgrading guide.

D3Hunter: > I think we can add the related operation to our upgrading guide.

you mean ask users to manually create index on system table? system table should better be managed by TiDB itself IMO, and most production cluster have very strict permission control, asking DBA or others with root permission to manage what TiDB itself should done, not sure how much they will buy this idea.

Rustin170506:
/hold

I didn't mean it. I commented on the wrong PR.

pkg/session/bootstrap.go
Rustin170506: We asked one of the largest users about their workload. They only have 1000 auto-analyses per week. So I think it is ok to add it during the upragde. Also @tangenta told me that it is not a liner increase case. So in most cases it only takes a few seconds.

Issue #58141, opened 2024-12-11
The max QPS of table creation by Lightning is slow in millions tables scenario


lilinghai:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

500k schemas, 1 million tables
lighting import
![Image](https://github.com/user-attachments/assets/51f69483-4d06-4300-8b4d-244ec96b474c)

### 4. What is your TiDB version? (Required)

[2024/12/10 02:20:07.001 +00:00] [INFO] [info.go:53] ["Welcome to TiDB-Lightning"] [release-version=v8.5.0-alpha-306-gd6b313f268] [git-hash=d6b313f2680f96f775730c555e7ba1d9881d68d5] [git-branch=HEAD] [go-version=go1.23.4] [utc-build-time="2024-12-09 17:11:32"] [race-enabled=false]

D3Hunter:
there are 2 problems here
- max create-table QPS is too low, as lightning uses max 8 threads to do it
- create-table getting slower, it's causes by that querying system table and reloading schema became slower, 

the second is a separate issue, and right now we cannot reproduce it.

this issue will only focus the first

D3Hunter:
for the second problem, we track in https://github.com/pingcap/tidb/issues/58363

D3Hunter:
we also track https://github.com/pingcap/tidb/issues/58649#issuecomment-2567330681 here

jebter:
Releted to issue: https://github.com/pingcap/tidb/issues/60283

PR #58636, opened 2024-12-31
lightning: replace information_schema query when importing table schema with separate show create table

D3Hunter:

Issue Number: ref #58141

Problem Summary:

info schema V2 only store one copy of schema object in memory, so read/write need to lock, if we read too much and takes too long, it affects write, i.e. schema reloading during DDL execution, we can mitigate this by using finer-grained lock, but we cannot avoid it completely with current strategy. so in lightning, we remove related queries, and avoid create table twice by `IF NOT EXIST`, and check table existence on error by `SHOW CREATE TABLE`, to make schema import part faster

we also found that the query to `information_schema.tables` is getting slower over time, and cause lightning dispatch jobs slower, we can fix it too by replace the query.
### Check List

Tests 

- [x] Unit test
- [x] Integration test

import 100K DB and total 200K tables, time reduced from 55m27 -> 19m20

before
image
after
image

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/lightning/mydump/schema_import.go
+++ b/pkg/lightning/mydump/schema_import.go
@@ -317,6 +317,16 @@ func (si *SchemaImporter) getExistingDatabases(ctx context.Context) (set.StringS
 
 // isTableExist checks whether the table exists in the downstream database, it
 // works for view too.
+// info schema V2 only store one copy of schema object in memory, so read/write
+// need to lock, if we read too much and takes too long, it affects write, i.e.
+// schema reloading during DDL execution, we can mitigate this by using
+// finer-grained lock, but we cannot avoid it completely with current strategy.
+// that's why we don't check table existence in batch by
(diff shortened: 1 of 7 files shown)


pkg/lightning/mydump/schema_import.go
```diff
@@ -20,8 +20,10 @@ import (
 	"fmt"
 	"strings"
 
+	dmysql "github.com/go-sql-driver/mysql"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/tidb/br/pkg/storage"
+	tmysql "github.com/pingcap/tidb/pkg/errno"
```
lance6716, line 26: the package name errno is enough

lance6716 (approved):

GMHDBJD (approved):
LGTM

PR #58724, opened 2025-01-06
br: enable parallel restore

### LINKED ISSUE #58725: br: use dedicated checkpoint table for each br task
## Enhancement

separate out the checkpoint table so can enable multiple restore task execution. Right now there is only one set of checkpoint tables per cluster, it doesn't work if more than one restore happens at the same time

Tristan1900:

Issue Number: close #58725

Problem Summary:

1. To allow parallel restore tasks, we introduced a restore id for each task, the id is allocated by an internal TiDB registry table and it's monotonic increasing, the table is in mysql db called `tide_restore_registry`
2. We add this restore id to the suffix of checkpoint DB and storage, so each task can have their own checkpoint and not messing with others
3. For each new task, it registers itself to the registry, and make sure tables it's going to restore is not overlapping with restore task that has a smaller restore id. If there is conflict it will abort the restore and show the reason.
4. After finishing the restore, it will unregister itself from the registry
5. In case of failure, for a graceful one, it will change the task status in registry from running to pause, so next time user can retry and reusing the checkpoint. If disaster happens like OOM that BR didn't get a chance to pause, user needs to manually clean it up. We have a heartbeat mechanism that update the restore task timestamp, it can provide some hints whether this task is active or orphaned.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


br/pkg/task/restore.go
```diff
@@ -822,38 +835,92 @@ func RunRestore(c context.Context, g glue.Glue, cmdName string, cfg *RestoreConf
...
 	if restoreError != nil {
+		// if err happens at register phase no restoreID will be generated and default is 0.
+		if cfg.RestoreID == 0 {
+			return errors.Trace(restoreError)
+		}
+		// if checkpoint is not persisted, let's just unregister the task since we don't need it
+		if hasCheckpointPersisted(c, cfg) {
+			log.Info("pausing restore task from registry",
+				zap.Uint64("restoreId", cfg.RestoreID), zap.Error(restoreError))
+			if err := restoreRegistry.PauseTask(c, cfg.RestoreID); err != nil {
```
Leavrth, line 846: What if BR is killed because OOM? BR might not have chance to pause the task.

br/pkg/registry/registration.go
```diff
@@ -0,0 +1,523 @@
+// Copyright 2025 PingCAP, Inc.
```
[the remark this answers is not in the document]
Tristan1900: let me see what I can do, it's calling sql so not sure unit test can easily cover that

br/pkg/registry/registration.go
```diff
@@ -0,0 +1,523 @@
...
+		if foundUUID == operationUUID {
+			log.Info("successfully resumed existing registration by this process",
+				zap.Uint64("restore_id", taskID),
+				zap.String("uuid", operationUUID),
+				zap.Strings("filters", info.FilterStrings))
+			return taskID, nil
+		}
+		// task exists but was either created by another process or has been running
+		// check if it's in running state
+		runningRows, _, err := execCtx.ExecRestrictedSQL(
```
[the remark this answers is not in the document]
Tristan1900: ah yes you are right

br/pkg/registry/registration.go
```diff
@@ -0,0 +1,523 @@
...
+	log.Info("ensured restore registration table exists",
+		zap.String("db", RegistrationDBName),
+		zap.String("table", RegistrationTableName))
+
+	return nil
+}
+
+// ResumeOrCreateRegistration first looks for an existing registration with the given parameters.
+// If found and paused, it tries to resume it. Otherwise, it creates a new registration.
+func (r *Registry) ResumeOrCreateRegistration(ctx context.Context, info RegistrationInfo) (uint64, error) {
```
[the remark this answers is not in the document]
Tristan1900: That was my first try, maybe we can discuss offline cuz based on my observation `ExecRestrictedSQL` doesn't work with nested transaction, it appears to be a single transaction for admin purpose. But maybe I'm wrong.

br/pkg/task/restore.go
Tristan1900: good call, let me think about that

br/pkg/registry/registration.go
```diff
@@ -0,0 +1,558 @@
...
+	"github.com/pingcap/tidb/br/pkg/utils"
+	"github.com/pingcap/tidb/pkg/domain"
+	"github.com/pingcap/tidb/pkg/kv"
+	"github.com/pingcap/tidb/pkg/util/sqlexec"
+	filter "github.com/pingcap/tidb/pkg/util/table-filter"
+	"go.uber.org/zap"
+)
+
+const (
+	RegistrationDBName    string = utils.TemporaryDBNamePrefix + "Restore_Registration_DB"
```
BornChanger, line 37: why not keep it in mysql db since the records might last for long?

br/pkg/task/common.go
```diff
@@ -663,11 +663,14 @@ func (cfg *Config) ParseFromFlags(flags *pflag.FlagSet) error {
 				Schema: db,
 				Name:   tbl,
 			})
+			cfg.FilterStr = []string{fmt.Sprintf("`%s`.`%s`", db, tbl)}
 		} else {
 			cfg.TableFilter = filter.NewSchemasFilter(db)
+			cfg.FilterStr = []string{fmt.Sprintf("`%s`.*", db)}
```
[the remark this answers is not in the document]
Tristan1900: that should be escaped and filter can identify an escaped * and a wildcard *.

br/pkg/registry/registration.go
Tristan1900: user can always manually clean up, there is nothing to prevent them deleting it. And I have a heartbeat mechanism to publish heartbeat to each task, suggesting the liveness of those tasks, so user can have confidence when deleting it. 


Update: actually it's not possible to make race condition free when doing concurrent create and delete. The idea of making a long term table is better. Thanks for the suggestion!

pkg/session/bootstrap.go
```diff
@@ -730,6 +730,29 @@ const (
...
+		id BIGINT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+		filter_strings TEXT NOT NULL,
+		filter_hash VARCHAR(64) NOT NULL,
+		start_ts BIGINT UNSIGNED NOT NULL,
+		restored_ts BIGINT UNSIGNED NOT NULL,
+		upstream_cluster_id BIGINT UNSIGNED,
+		with_sys_table BOOLEAN NOT NULL DEFAULT TRUE,
+		status VARCHAR(20) NOT NULL DEFAULT 'running',
+		cmd TEXT,
+		start_timestamp BIGINT UNSIGNED NOT NULL,
```
benmeadowcroft: From a user perspective I find the column naming confusing.

First, it is not clear how this is different from `start_ts` referenced above. Can you clarifying the naming here? If they are used for different things, is the a better name that can be used to make it more obvious? Or is this just intended to offer different levels of precision (if so why the same data type)?

Second, the suffix of the column name `_timestamp` implies this is a`TIMESTAMP` while the type is specified as `BIGINT`. This is confusing and inconsistent.

pkg/session/bootstrap.go
```diff
@@ -730,6 +730,29 @@ const (
 		update_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
 		PRIMARY KEY (restore_id, restored_ts, upstream_cluster_id, segment_id));`
 
+	// CreateRestoreRegistryTable is a table that tracks active restore tasks to prevent conflicts.
+	CreateRestoreRegistryTable = `CREATE TABLE IF NOT EXISTS mysql.tidb_restore_registry (
```
benmeadowcroft, line 734: Please use consistent suffixes for `_time` and `_ts` and consistent data types for the desired level of precision.

benmeadowcroft (changes requested):

pkg/session/bootstrap.go
Tristan1900: thanks for taking a look!
yes you are absolutely right the name is confusing, let me rename the latter one.

windtalker (approved):
expression/executor part lgtm

benmeadowcroft (approved):

Leavrth (approved):

winoros (approved):

winoros:
for planner part

Issue #58743, opened 2025-01-07
unexpected internal error when executing a CTE query


wjhuang2016:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
``` SQL

CREATE TABLE `tlf5d55361` (
  `col_9` time NOT NULL,
  `col_10` float NOT NULL,
  `col_11` json NOT NULL,
  `col_12` date NOT NULL,
  `col_13` json NOT NULL,
  `col_14` tinyint NOT NULL,
  `col_15` date DEFAULT NULL,
  `col_16` tinyblob NOT NULL,
  `col_17` time DEFAULT '03:51:26',
  PRIMARY KEY (`col_14`,`col_9`,`col_10`) /*T![clustered_index] CLUSTERED */
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci
PARTITION BY RANGE (`col_14`)
(PARTITION `p0` VALUES LESS THAN (-128),
 PARTITION `p1` VALUES LESS THAN (-84),
 PARTITION `p2` VALUES LESS THAN (-41),
 PARTITION `p3` VALUES LESS THAN (-30));

CREATE TABLE `td8d55878` (
  `col_26` datetime DEFAULT NULL,
  `col_27` time DEFAULT NULL,
  `col_28` json DEFAULT NULL,
  `col_29` char(186) CHARACTER SET utf8 COLLATE utf8_unicode_ci DEFAULT '4-BJKi',
  `col_30` date NOT NULL DEFAULT '1998-07-28',
  `col_31` datetime NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=gbk COLLATE=gbk_chinese_ci;

[...]

PR #58771, opened 2025-01-07
planner: fix an unexpected result of the inlineprojection of TopN

EricZequan:

Issue Number: ref #58743 

Problem Summary:

Modified the inlineprojection order of TopN, and modify the pruning parameter when the columns required by the upper-level plan are empty.
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


hawkingrei:
Fixing a bug is not a good title for a PR.

pkg/planner/core/operator/logicalop/logical_top_n.go
```diff
@@ -92,6 +93,15 @@ func (lt *LogicalTopN) PruneColumns(parentUsedCols []*expression.Column, opt *op
...
 		return nil, err
 	}
+	// If the length of parentUsedCols is 0, it means that the parent plan does not need this plan to output related
+	// results, such as: select count(*) from t
+	// So we set the schema of topN to 0. After inlineprojection, the schema of topN will be set to the shortest column
+	// in its child plan, and this column will not be used later.
+	if len(snapParentUsedCols) == 0 {
+		lt.SetSchema(nil)
+	}
+	lt.InlineProjection(snapParentUsedCols, opt)
```
AilinKid, line 103: i think the comment inside InlineProjection is changed right?

```
// It should be always true.
if len(used) > 0 {
	used[chosenPos] = true
}
```
one is try to keep the schema not empty, while this one will set it to nil actively. Can you conclude the situation of both and make some code clean if any, like InlineProjection is still necessary?

EricZequan: The comment in `inlineprojection` has not changed. It just ensures that the plan needs to return at least one column (even if this column is not needed). This needs to be consistent with the behavior of the executor.

The schema of TopN is usually correct, but when `parentUsedCols` is empty, because the plan needs to return at least one column, when we use `parentUsedCols` to `inlineproject` it, TopN retains the column with the smallest length (including generated-columns), which may not exist in the child plan. Therefore, we need to use the child's schema to `inlineproject` TopN. This change has no effect on the result.

When `parentUsedCols` is empty, it does not matter whether the schema is nil or not. The only difference is whether we use the child's schema or TopN's own schema. In this case, using the child's schema ensures that TopN can find the columns needed by the lower layer during `ResolveIndices`.

pkg/executor/test/executor/executor_test.go
```diff
@@ -3088,3 +3088,66 @@ func TestQueryWithKill(t *testing.T) {
...
+			col_26 datetime DEFAULT NULL,
+			col_27 time DEFAULT NULL,
+			col_28 json DEFAULT NULL,
+			col_29 char(186) CHARACTER SET utf8 COLLATE utf8_unicode_ci DEFAULT '4-BJKi',
+			col_30 date NOT NULL DEFAULT '1998-07-28',
+			col_31 datetime NOT NULL
+		) ENGINE=InnoDB DEFAULT CHARSET=gbk COLLATE=gbk_chinese_ci;
+	`)
+
+	tk.MustExec(`
```
hawkingrei, line 3130:
```suggestion
	tk.MustQuery(`
```

add result checking.

pkg/executor/test/executor/executor_test.go
```diff
@@ -3088,3 +3088,66 @@ func TestQueryWithKill(t *testing.T) {
 	}()
 	wg.Wait()
 }
+
+func TestIssue58743(t *testing.T) {
```
hawkingrei, line 3092: Move this test into ```pkg/planner/core/operator/logicalop```

AilinKid (approved):

pkg/planner/core/operator/logicalop/logical_top_n.go
```diff
@@ -83,7 +83,8 @@ func (lt *LogicalTopN) PruneColumns(parentUsedCols []*expression.Column, opt *op
 	child := lt.Children()[0]
 	var cols []*expression.Column
 
-	lt.InlineProjection(parentUsedCols, opt)
+	snapParentUsedCols := make([]*expression.Column, 0)
```
AilinKid, line 86: alloc cap？

hawkingrei (approved):

PR #58095, opened 2024-12-09
domain: reduce the chance of "Information schema is out of date" error

tiancaiamao:
### What problem does this PR solve?

Issue Number: close #58079

Problem Summary:

### What changed and how does it work?

See the code comments

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [X] Manual test (add detailed scripts or steps below)

```
git diff pkg/sessionctx/variable/session.go
diff --git a/pkg/sessionctx/variable/session.go b/pkg/sessionctx/variable/session.go
index 0551e77718..67bc82199e 100644
--- a/pkg/sessionctx/variable/session.go
+++ b/pkg/sessionctx/variable/session.go
@@ -732,10 +732,16 @@ func (s *UserVars) Clone() UserVarsReader {
        return cloned
 }
 
+var LALALA bool
+
 // SetUserVarVal set user defined variables' value
 func (s *UserVars) SetUserVarVal(name string, dt types.Datum) {
        s.lock.Lock()
        defer s.lock.Unlock()
+       if name == "lalala" {
+               fmt.Println("set lalala = true ===========")
+               LALALA = true
+       }
        s.values[name] = dt
 }
 
@@ -746,6 +752,10 @@ func (s *UserVars) UnsetUserVar(varName string) {
        defer s.lock.Unlock()
        delete(s.values, varName)
        delete(s.types, varName)
+
+       if varName == "lalala" {
+               LALALA = false
+       }
 }
```

And in domain.go, add `time.Sleep(35 * time.Second)` if `variable.LALALA = true`, and mock full reload:

```
@@ -621,6 +638,10 @@ func (do *Domain) tryLoadSchemaDiffs(useV2 bool, m meta.Reader, usedVersion, new
                }
        })
 
+       if variable.LALALA {
+               return nil, nil, nil, errors.New("mock error lalala")
+       }
```

Open one mysql client and use 

```
mysql> set @lalala = true;
Query OK, 0 rows affected (0.00 sec)
```

Check tidb logs to see full reload and everything works well, 

```
...
[2024/12/09 21:31:15.337 +08:00] [INFO] [domain.go:421] ["full load InfoSchema success"] [isV2=true] [currentSchemaVersion=995818] [neededSchemaVersion=995818] ["elapsed time"=45.57806096s]
[2024/12/09 21:31:15.337 +08:00] [INFO] [domain.go:831] ["full load and reset schema validator"]
[...]

PR #58818, opened 2025-01-09
infoschema: remove domain init reload when reload time greater than half of schemaLease

GMHDBJD:

Issue Number: ref https://github.com/pingcap/tidb/issues/58821 

Problem Summary:

- we have check schema version in domain reload after https://github.com/pingcap/tidb/issues/58095, so no need to reload twice when domain init
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
restart tidb with millions of tables
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/domain/domain.go
+++ b/pkg/domain/domain.go
@@ -1482,24 +1482,8 @@ func (do *Domain) Init(
 		return err
 	}
 
-	startReloadTime := time.Now()
 	// step 3: domain reload the infoSchema.
-	err = do.Reload()
-	if err != nil {
-		return err
-	}
-
-	sub := time.Since(startReloadTime)
-	// The reload(in step 2) operation takes more than ddlLease and a new reload operation was not performed,
-	// the next query will respond by ErrInfoSchemaExpired error. So we do a new reload to update schemaValidator.latestSchemaExpire.
-	if sub > (do.schemaLease / 2) {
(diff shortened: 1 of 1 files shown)


zimulala (approved):
LGTM

D3Hunter (approved):

pkg/domain/domain.go
```diff
@@ -1482,24 +1482,8 @@ func (do *Domain) Init(
...
-	// the next query will respond by ErrInfoSchemaExpired error. So we do a new reload to update schemaValidator.latestSchemaExpire.
-	if sub > (do.schemaLease / 2) {
-		logutil.BgLogger().Warn("loading schema and starting ddl take a long time, we do a new reload", zap.Duration("take time", sub))
-		err = do.Reload()
-		if err != nil {
-			return err
-		}
-	}
-	return nil
+	return do.Reload()
```
Benjamin2037, line 1486: What if the reload failed with err? here I saw that still need domain.Reload().

GMHDBJD: Previously, when the reload exceeded half of the lease, we needed to reload again, but now we don’t need to. 
If the reload fails here, we will retry at the upper layer.

GMHDBJD: Previously, when the reload exceeded half of the lease, we needed to reload again, but now we don’t need to. 
If the reload fails here, we will retry at the upper layer.

PR #58899, opened 2025-01-13
json: add json fuzz test for `json_extract` function and fix several bugs found by it | tidb-test=pr/2467

### LINKED ISSUE #37806: json float format should have extra `.0` if it's an integer
## Enhancement

Printing json `1.0` will get `1`, which is not expected. It should print `1.0`.

### LINKED ISSUE #58888: sometimes TiDB will return json float number without float point
## Bug Report

Run the following SQL:

```
select json_extract("0.0", "$");
```

```
mysql> select json_extract("0.0", "$");
+--------------------------+
| json_extract("0.0", "$") |
+--------------------------+
| 0.0                      |
+--------------------------+
1 row in set (0.00 sec)
```

```
mysql> select json_extract("0.0", "$");
+--------------------------+
| json_extract("0.0", "$") |
+--------------------------+
| 0                        |
+--------------------------+
1 row in set (0.00 sec)
```

```
c199ddfcdf9e1f88a243c2b7847d8f3224446004
```

### LINKED ISSUE #58894: json float in scientific format is different with MySQL
## Bug Report

```
SELECT JSON_EXTRACT('[1E27]', '$')
```

```
[1e27]
```

```
[1e+27]
```

In `util.AppendFormatFloat`, TiDB provides a function to format a float in MySQL  compatible way. However, the json value doesn't use this function (because they actually have a little logic difference). We'd better to align the behavior between them.

### LINKED ISSUE #58895: TiDB accepts special character as json path leg, but MySQL doesn't
## Bug Report

```
SELECT json_extract("[1E27]", "$.ѿ")
```

```
ERROR 3143 (42000): Invalid JSON path expression. The error is around character position 4.
```

```
mysql> select json_extract("[1E27]", "$.ѿ");
+--------------------------------+
| json_extract("[1E27]", "$.ѿ")  |
+--------------------------------+
| NULL                           |
+--------------------------------+
1 row in set (0.03 sec)
```

In MySQL document, it says the key name must be `ESIdentifier | doubleQuotedString`. The `doubleQuotedString` accepts special character, but `ESIdentifier` not.

### LINKED ISSUE #58896: slice bounds out of range for special json path
## Bug Report

```
select json_extract("{}", "$.\"\\0\\");
```

```
mysql> select json_extract("{}", "$.\"\\0\\");
ERROR 3143 (42000): Invalid JSON path expression. The error is around character position 6.
```

```
mysql> select json_extract("{}", "$.\"\\0\\");
ERROR 1105 (HY000): runtime error: slice bounds out of range [:7] with length 6
```

### LINKED ISSUE #58897: json key format with `\b`, `\f` is not compatible with MySQL
## Bug Report

```
select json_extract("{\"\\b\":\"\"}", "$");
```

```
mysql> select json_extract("{\"\\b\":\"\"}", "$");
+-------------------------------------+
| json_extract("{\"\\b\":\"\"}", "$") |
+-------------------------------------+
| {"\b": ""}                          |
+-------------------------------------+
1 row in set (0.00 sec)
```

```
mysql> select json_extract("{\"\\b\":\"\"}", "$");
+-------------------------------------+
| json_extract("{\"\\b\":\"\"}", "$") |
+-------------------------------------+
| {"\u0008": ""}                      |
+-------------------------------------+
1 row in set (0.02 sec)
```

YangKeao:

Issue Number: close #58897, close #58896, close #58895, close #58894, close #58888, close #37806

Problem Summary:

This PR adds a fuzz test for `json_extract` function and fix several bugs found by it.

1. Add a fuzz test for `json_extract`.
2. Modify the behavior of `jsonMarshalStringTo` to specially handle the `\b` and `\f`.
3. Fix the json path parser by using `[]rune` to make sure we have correct unicode behavior.
4. Fix the scientific notation of float value.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/types/json_binary.go
+++ b/pkg/types/json_binary.go
@@ -330,6 +330,9 @@ func (bj BinaryJSON) valEntryGet(valEntryOff int) BinaryJSON {
 
 func (bj BinaryJSON) marshalFloat64To(buf []byte) ([]byte, error) {
 	// NOTE: copied from Go standard library.
+	// TODO: this function is very similar to `util.AppendFormatFloat`, it'd be better to unify them.
+	// Now the `util.AppendFormatFloat` is an internal function of `server` package, and the `marshalFloat64To`
+	// handles the tailing `.0` specially, so they are not merged yet.
 	f := bj.GetFloat64()
 	if math.IsInf(f, 0) || math.IsNaN(f) {
 		return buf, &json.UnsupportedValueError{Str: strconv.FormatFloat(f, 'g', -1, 64)}
@@ -344,17 +347,36 @@ func (bj BinaryJSON) marshalFloat64To(buf []byte) ([]byte, error) {
 	ffmt := byte('f')
 	// Note: Must use float32 comparisons for underlying float32 value to get precise cutoffs right.
 	if abs != 0 {
-		if abs < 1e-6 || abs >= 1e21 {
+		// The scientific notation range for MySQL is different from Go JSON. Ref `util.AppendFormatFloat`
+		if abs < 1e-15 || abs >= 1e15 {
 			ffmt = 'e'
 		}
 	}
+
+	floatPos := len(buf)
 	buf = strconv.AppendFloat(buf, f, ffmt, -1, 64)
+	floatBuf := buf[floatPos:]
+
 	if ffmt == 'e' {
 		// clean up e-09 to e-9
-		n := len(buf)
+		n := len(floatBuf)
 		if n >= 4 && buf[n-4] == 'e' && buf[n-3] == '-' && buf[n-2] == '0' {
 			buf[n-2] = buf[n-1]
 			buf = buf[:n-1]
+		}
+
+		// remove the leading '+' in the exponent
+		plusPos := bytes.IndexRune(floatBuf, '+')
+		if plusPos > 0 {
+			buf = append(buf[:floatPos+plusPos], buf[floatPos+plusPos+1:]...)
+		}
+	} else {
+		// keeps at least one digit even if `f` is an integer
+		// assuming that this `floatBuf` will not be too long, it's fine to scan it
+		// to find the dot
+		if !bytes.ContainsRune(floatBuf, '.') {
+			buf = append(buf, '.')
+			buf = append(buf, '0')
 		}
 	}
 	return buf, nil
@@ -417,8 +439,12 @@ func jsonMarshalStringTo(buf, s []byte) []byte {
 				buf = append(buf, '\\', 'r')
 			case '\t':
 				buf = append(buf, '\\', 't')
+			case '\b':
+				buf = append(buf, '\\', 'b')
+			case '\f':
+				buf = append(buf, '\\', 'f')
 			default:
-				// This encodes bytes < 0x20 except for \t, \n and \r.
+				// This encodes bytes < 0x20 except for \t, \n, \r, \b, \f.
 				// If escapeHTML is set, it also escapes <, >, and &
 				// because they can lead to security holes when
 				// user-controlled strings are rendered into JSON
--- a/pkg/types/json_path_expr_test.go
+++ b/pkg/types/json_path_expr_test.go
@@ -73,6 +73,12 @@ func TestValidatePathExpr(t *testing.T) {
 		{`$[`, false, 0},
 		{`$a.***[3]`, false, 0},
 		{`$1a`, false, 0},
+		{`$.ѿ`, false, 0},
+		{`$."ѿ"`, true, 1},
+		{"$.\"\\0\\", false, 0},
+		{`$.Ѡ`, false, 0}, // This test case is special, because Ѡ is 0xD1 0xA0 in UTF-8, and 0xA0 is a space character.
+		{`$."Ѡ"`, true, 1},
+		{`$.µ`, true, 1},
 	}
 
 	for _, test := range tests {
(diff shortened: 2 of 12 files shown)


YangKeao:
TODO: also test these cases for TiKV.

Defined2014:
> TODO: also test these cases with TiKV.

Integration test also run on TiKV env.

pkg/types/json_binary.go
```diff
@@ -330,6 +330,9 @@ func (bj BinaryJSON) valEntryGet(valEntryOff int) BinaryJSON {
 
 func (bj BinaryJSON) marshalFloat64To(buf []byte) ([]byte, error) {
 	// NOTE: copied from Go standard library.
+	// TODO: this function is vary similar to `util.AppendFormatFloat`, it'd be better to unify them.
```
Defined2014, line 333:
```suggestion
	// TODO: this function is very similar to `util.AppendFormatFloat`, it'd be better to unify them.
```

pkg/types/json_path_expr_test.go
```diff
@@ -17,6 +17,7 @@ package types
 import (
 	"testing"
 
+	_ "github.com/go-sql-driver/mysql"
```
Defined2014, line 20: Do we need import this library here?

pkg/types/json_binary.go
YangKeao: Fixed.

YangKeao:
> > TODO: also test these cases with TiKV.
> 
> Integration test also run on TiKV env.

@Defined2014  For the added case, they are not pushed down to TiKV. It's possible to re-write these SQLs to push down, but I think it's better to manually check the TiKV side.

After all, TiKV side is not related to this PR. No matter whether it has these bugs or not, there will be another PR for it (in `tikv/tikv`).

pkg/types/json_path_expr_test.go
YangKeao: Yes. It asserts that the result between `MySQL` and tidb internal functions are the same. Without it, it cannot tell whether a result is correct.

Defined2014: Should it added in `json_binary_test.go` file?

tests/integrationtest/r/expression/builtin.result
```diff
@@ -713,7 +713,7 @@ create table tb5(a json, b bigint unsigned);
 insert into tb5 (a, b) values ('184467440737095516160', 18446744073709551615);
 select * from tb5 where cast(a as unsigned int)=b;
 a	b
-184467440737095500000	18446744073709551615
+1.844674407370955e20	18446744073709551615
```
YangKeao, line 716: It's not a behavior change, but I found it's not compatible with MySQL :thinking:. Record in https://github.com/pingcap/tidb/issues/58901.

pkg/types/json_path_expr_test.go
YangKeao: You are right :facepalm: . It's a mistake. I moved the tests from `json_path_expr_test.go` to `json_binary_test.go` and forgot to modify the `import`.

It's fixed now.

Defined2014 (approved):

tests/integrationtest/r/expression/issues.result
```diff
@@ -1347,7 +1347,7 @@ select col, count(1) c from tx2 group by col order by c desc;
 col	c
 [922337203685477581]	5
 ["3"]	4
-[3]	3
+[3.0]	3
```
Defined2014, line 1350: Seems it's not same on TiKV env, so `check_dev2` failed. :(

YangKeao: It's indeed caused by aggregation, and I think it's somehow expected. The `[3.0]` and `[3]` are aggregated into the same group, so it's expected that the `firstrow` may pick any one of them. The MySQL itself has a non-strict order:

```sql
mysql> create table tx2(j json);

mysql> insert into tx2 values ('[3.0]'), ('[3]');
Query OK, 2 rows affected (0.01 sec)
Records: 2  Duplicates: 0  Warnings: 0

mysql> select j from tx2 group by j;
+-------+
| j     |
+-------+
| [3.0] |
+-------+
1 row in set (0.00 sec)

mysql> delete from tx2;
Query OK, 2 rows affected (0.00 sec)

mysql> insert into tx2 values ('[3]'), ('[3.0]');
Query OK, 2 rows affected (0.01 sec)
Records: 2  Duplicates: 0  Warnings: 0

mysql> select j from tx2 group by j;
+------+
| j    |
+------+
| [3]  |
+------+
1 row in set (0.00 sec)
```

Previously, we also have the same issue, but the TiDB will print `3.0` as `3`, so the test didn't expose it :thinking:. Let me think whether (and how) to rewrite this test.

tests/integrationtest/t/expression/issues.test
```diff
@@ -856,7 +856,7 @@ insert into tx2 values (json_array(922337203685477581)),(json_array(922337203685
 #
 # ordering by a JSON col is not supported in MySQL, and the order is a bit questionable in TiDB.
 # sort by count for test result stability.
-select col, count(1) c from tx2 group by col order by c desc;
+select max(cast(col as char)) j, count(1) c from tx2 group by col order by c desc;
```
YangKeao, line 859: @Defined2014 PTAL. Now it should have a stable result.

tests/integrationtest/r/expression/issues.result
YangKeao: > Maybe use `--replace_regex` comment to rewrite it.

It sounds better. I'll do it.

tests/integrationtest/t/expression/issues.test
YangKeao: Revert

xhebox (approved):

PR #58943, opened 2025-01-15
ttl: rollback the scan trasnaction even if the `BEGIN` failed.

### LINKED ISSUE #58900: TTL scan worker may put a not-committed session to the session pool
This issue is found by flaky test `TestParallelLockNewTask`. The log is attached.

[test.log](https://github.com/user-attachments/files/18396628/test.log)

YangKeao:

Issue Number: close #58900

Problem Summary:

Previously, the `RunInTxn` will only rollback the transaction after the `BEGIN` runs successfully. However, if the `BEGIN` is killed after setting the transaction status, it can still leave a valid transaction. We'll need to also `ROLLBACK` it.

The effect is that when we are cancelling a TTL task (scaling the workers, stepping out of the TTL window, stopping the TTL function...), a session with valid transaction may pollute the session pool.

Move the `defer ... ROLLBACK` logic to the top of the `RunInTxn` function.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
Fix the issue that canceling a TTL task may return a polluted session to global session pool.
```

### DIFF
--- a/pkg/ttl/session/session.go
+++ b/pkg/ttl/session/session.go
@@ -111,6 +111,20 @@ func (s *session) ExecuteSQL(ctx context.Context, sql string, args ...any) ([]ch
 
 // RunInTxn executes the specified function in a txn
 func (s *session) RunInTxn(ctx context.Context, fn func() error, txnMode TxnMode) (err error) {
+	success := false
+	defer func() {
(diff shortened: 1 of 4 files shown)


lcwangchao (approved):

YangKeao:
Wow! It's merged so fast. Amazing.

PR #58945, opened 2025-01-15
privilege/privileges:  refactor the data struct for user privilege data

### LINKED ISSUE #55563: Modify user or privilege will consume a significant amount of time and resources in million users scenario

tiancaiamao:

Issue Number: close #55563

Problem Summary:

The old data struct is not fit for a lot of users, it use copy-on-write and, when there are 2M users, the cost of copy 2M item for every change is not acceptable.

I have optimize the code to avoid holding all user's privilege data in-memory, and instead maintain just the active users.
However, it does not help the worst case when all users are active.

So this commit is for the worst case:  refactor the data struct using a btree, make the update cost small enough.

The main difference is the `merge` function.
Before this change, we merge two (huge) slices s1, s2:  create a new big slice s3,  copy the data from the old one s1, remove duplicated, and than append the data from s2.

Now it works by s1.Clone(), and updates on this clone by adding the data of s2. 
The underlying btree implement is a lazy clone,  the clone will share the same node with the old btree, and copy-on-write after the modifications.  This way the cost is much smaller.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


CbcWestwolf (approved):
LGTM.
Is there any test to show how much memory btree can save compared with slice's?

pkg/privilege/privileges/cache.go
```diff
@@ -506,78 +572,87 @@ func (p *immutable) loadSomeUsers(ctx sqlexec.RestrictedSQLExecutor, userList ..
...
-		if !slices.ContainsFunc(userList, func(usr string) bool { return usr == v.User }) {
-			ret.user = append(ret.user, v)
-		}
-	}
-	ret.user = append(ret.user, diff.user...)
-	slices.SortFunc(ret.user, compareUserRecord)
-	ret.buildUserMap()
+func (p *MySQLPrivilege) merge(diff *MySQLPrivilege, userList []string) *MySQLPrivilege {
+	ret := newMySQLPrivilege()
+	user := p.user.Clone()
```
lcwangchao, line 577: Do we need some mutex for `Clone` ? The BTreeG is documented as:

> // Clone clones the btree, lazily.  Clone should not be called concurrently,
// but the original tree (t) and the new tree (t2) can be used concurrently
// once the Clone call completes.

lcwangchao (approved):

pkg/privilege/privileges/cache.go
tiancaiamao: Done

lance6716 (commented):
/approve

for domain part

GMHDBJD (approved):
LGTM

PR #58962, opened 2025-01-15
planner: remove redundant branches in the OR list | tidb-test=pr/2470

### LINKED ISSUE #58998: Remove redundant branches in the OR list in filters
## Enhancement
Though users don't deliberately write redundant expressions in the SQL, there are still redundant expressions, such as those from generated SQLs.

In our current implementation, we have `RemoveDupExprs()` and `PropagateConstant()` for AND lists. However, the correspondence for the OR list is lacking. So, there could be redundant filters in the OR list in the final plan, which can sometimes cause an inefficient plan.

For example:
```sql
create table t(a int, b int, c int, index ia(a), index ib(b), index ic(c));
explain select * from t where a = 1 or a = 1 or b = 2 or b = 2 or c = 3 or c = 3;
```
```
> explain select * from t where a = 1 or a = 1 or b = 2 or b = 2 or c = 3 or c = 3;
+--------------------------------+---------+-----------+----------------------+---------------------------------------------+
| id                             | estRows | task      | access object        | operator info                               |
+--------------------------------+---------+-----------+----------------------+---------------------------------------------+
| IndexMerge_15                  | 29.97   | root      |                      | type: union                                 |
| ├─IndexRangeScan_8(Build)      | 10.00   | cop[tikv] | table:t, index:ia(a) | range:[1,1], keep order:false, stats:pseudo |
| ├─IndexRangeScan_9(Build)      | 10.00   | cop[tikv] | table:t, index:ia(a) | range:[1,1], keep order:false, stats:pseudo |
| ├─IndexRangeScan_10(Build)     | 10.00   | cop[tikv] | table:t, index:ib(b) | range:[2,2], keep order:false, stats:pseudo |
| ├─IndexRangeScan_11(Build)     | 10.00   | cop[tikv] | table:t, index:ib(b) | range:[2,2], keep order:false, stats:pseudo |
| ├─IndexRangeScan_12(Build)     | 10.00   | cop[tikv] | table:t, index:ic(c) | range:[3,3], keep order:false, stats:pseudo |
| ├─IndexRangeScan_13(Build)     | 10.00   | cop[tikv] | table:t, index:ic(c) | range:[3,3], keep order:false, stats:pseudo |
| └─TableRowIDScan_14(Probe)     | 29.97   | cop[tikv] | table:t              | keep order:false, stats:pseudo              |
+--------------------------------+---------+-----------+----------------------+---------------------------------------------+
```

time-and-fate:

Issue Number: close #58998

In the rule `predicate_simplification`, add a new step to remove redundant (duplicate) expressions from the OR list. It's more like the existing `RemoveDupExprs()` than `PropagateConstant()` for AND lists.

Specifically, it uses `HashCode()` to check the redundancy, like we did in `RemoveDupExprs()` and `generateANDIndexMerge4NormalIndex()`. And it recursively does this for nested AND lists.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


hawkingrei:
Please link the related issue.

pkg/planner/core/rule_predicate_simplification.go
```diff
@@ -424,6 +425,51 @@ func shortCircuitLogicalConstants(sctx base.PlanContext, predicates []expression
...
+			andFunc := orItem.(*expression.ScalarFunction)
+			andList := expression.SplitCNFItems(andFunc)
+			removeRedundantORBranch(sctx, andList)
+			newORList = append(newORList, expression.ComposeCNFCondition(sctx.GetExprCtx(), andList...))
+		} else {
+			// 2. Otherwise, we check if it's a duplicate predicate by checking HashCode().
+			hashCode := string(orItem.HashCode())
+			// 2-1. If it's not a duplicate, we need to keep this predicate.
+			if _, ok := dedupMap[hashCode]; !ok {
+				dedupMap[hashCode] = struct{}{}
```
ghazalfamilyusa, line 461: Minor: the value of the map can be anything and may be just a simple constant like True.

pkg/planner/core/rule_predicate_simplification.go
```diff
@@ -424,6 +425,51 @@ func shortCircuitLogicalConstants(sctx base.PlanContext, predicates []expression
...
+// It modifies the input slice in place.
+func removeRedundantORBranch(sctx base.PlanContext, predicates []expression.Expression) {
+	for i, predicate := range predicates {
+		predicates[i] = recursiveRemoveRedundantORBranch(sctx, predicate)
+	}
+}
+
+func recursiveRemoveRedundantORBranch(sctx base.PlanContext, predicate expression.Expression) expression.Expression {
+	_, tp := FindPredicateType(sctx, predicate)
+	if tp != orPredicate {
```
ghazalfamilyusa, line 439: This is a recursive function (indirectly through removeRedundantORBranch). If you intend to cover the And case (like the code in the function) then you just need a general logic for OR AND lists.

tests/integrationtest/t/planner/core/casetest/predicate_simplification.test
```diff
@@ -282,3 +282,28 @@ SELECT count(*) FROM t1 WHERE (1 = 0 AND a1 = 1) AND (b1 = 2);
...
+
+drop table if exists t3;
+create table t3 (a varchar(10) collate utf8mb4_general_ci, b varchar(10) collate utf8mb4_bin, index ia(a), index ib(b));
+explain format=brief select * from t3 where a = 'a' or a = 'a' or a = 'A';
+explain format=brief select * from t3 where a = 'a' or a = 'a' or a = 'A' or b = _utf8mb4'b' or b = _latin1'b' or b = 'B';
+explain format=brief select * from t3 where a = _utf8mb4'a' collate utf8mb4_unicode_ci or a = _utf8mb4'a' collate utf8mb4_0900_ai_ci or a = 'A' or b = 'b' or b = 'b' or b = 'B';
+
+drop table if exists t4;
+create table t4(a int, b int, c int, d int, index iab(a,b), index iac(a,c), index iad(a,d));
+explain format=brief select /*+ use_index_merge(t4) */ * from t4 where a = 1 and (b = 2 or c = 4 or b = 12 or c = 5 or d = 6 or c = 4 or c = 5 or d = 6);
```
ghazalfamilyusa, line 309: Can you add more nested AND/OR cases since the code does that?

pkg/planner/core/rule_predicate_simplification.go
```diff
@@ -186,6 +186,7 @@ func splitCNF(conditions []expression.Expression) []expression.Expression {
 func applyPredicateSimplification(sctx base.PlanContext, predicates []expression.Expression) []expression.Expression {
 	simplifiedPredicate := shortCircuitLogicalConstants(sctx, predicates)
 	simplifiedPredicate = mergeInAndNotEQLists(sctx, simplifiedPredicate)
+	removeRedundantORBranch(sctx, simplifiedPredicate)
```
ghazalfamilyusa, line 189: Any reason of the order of the sub-rule?

ghazalfamilyusa (changes requested):

pkg/planner/core/rule_predicate_simplification.go
time-and-fate: Technically, yes. But I think using `map[T]struct{}` as a set has become a common practice in Golang.
https://stackoverflow.com/a/47544821

pkg/planner/core/rule_predicate_simplification.go
time-and-fate: I didn't intend to cover the AND case (since that's already covered by other existing logic), I want to handle OR lists nested in another AND list.

tests/integrationtest/t/planner/core/casetest/predicate_simplification.test
time-and-fate: Updated.
I added several more cases in `t1` since the logic of building index ranges has some extra functionality to simplify the ranges. So I think printing the expression in `Selection` can test the effect of this new rule more straightforwardly.
The `t4` case is a simplified version of the original case from the user.

pkg/planner/core/rule_predicate_simplification.go
time-and-fate: There are no strong reasons. Actually, after I put this new sub-rule in a different order, the new test cases don't have any changes.
I have several minor considerations though:
1. I try not to change the order of existing sub-rules.
2. `mergeInAndNotEQLists()` simplifies the predicates by merging some of them, i.e., constructing some new expressions. Probably there will be new redundant expressions after that sub-rule, so put the new sub-rule after it might be a good idea.
3. `pruneEmptyORBranches()` just removes useless OR branches, and should not produce new redundant expressions. So probably it's useless to put the new sub-rule after it.

ghazalfamilyusa (approved):

elsa0520 (approved):
LGTM

PR #59009, opened 2025-01-17
ddl: Implement TableMode feature

### LINKED ISSUE #59008: Protect data consistency when importing or restoring data
## Enhancement

Introduce TableMode to protect data consistency.

fishiu:

Issue Number: ref #59008

Problem Summary: Users may read/write data when other users are importing/restoring tables which leads to data inconsistency, so we need tablemode to protect users from reading/writing in such scenarios.

We add a new field called `TableMode` to `TableInfo` which has three possible values: `TableModeNormal`(default), `TableModeImport`, `TableModeRestore`.
- This field can be read by quering the `TIDB_TABLE_MODE` column of InfoSchema table.
- This filed can only be wrote by a new ddl job `ActionAlterTableMode` or when calling the executor interface `[Batch]CreateTableWithInfo()`. Additionally, these two APIs are only called by the BR or IMPORTINTO internally, meanning it is almost invisible to normal users.

Implementation: The new ddl job `ActionAlterTableMode` is based on the online schema change framework like other ddl job, but with only 1-step state transition.

RULE1: When a table is in `TableModeImport` or `TableModeRestore`, any read or write to table data itself is rejected by the planner and the user will get `ErrProtectedTableMode`.

RULE2: For internal usage, invalid conversion such as conversion from `Import` to `Restore` or from `Restore` to `Import` will end up with `ErrInvalidTableModeSet`.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/meta/model/job_args.go
```diff
@@ -1083,6 +1083,25 @@ func GetLockTablesArgs(job *Job) (*LockTablesArgs, error) {
 	return getOrDecodeArgs[*LockTablesArgs](&LockTablesArgs{}, job)
 }
 
+type AlterTableModeArgs struct {
+	TableMode TableModeState
+	SchemaID  int64
+	TableID   int64
+}
+
+func (a *AlterTableModeArgs) getArgsV1(*Job) []any {
```
D3Hunter, line 1092: this DDL can always use V2

pkg/ddl/executor.go
```diff
@@ -5646,6 +5657,37 @@ func (e *executor) UnlockTables(ctx sessionctx.Context, unlockTables []model.Tab
...
+		return infoschema.ErrDatabaseNotExists.GenWithStackByArgs()
+	}
+
+	t, ok := is.TableByID(e.ctx, args.TableID)
+	if !ok {
+		return infoschema.ErrTableNotExists.GenWithStackByArgs()
+	}
+
+	job := &model.Job{
+		Version:        model.GetJobVerInUse(),
```
D3Hunter, line 5674: use v2 directly for new DDL

pkg/meta/model/job_args.go
fishiu: Can I still keep getArgsV1 and decodeV1? I found that there will be syntax error if these 2 func are removed, seems that AlterTableModeArgs must implement these 2 functions.

pkg/ddl/executor.go
fishiu: already changed to model.JobVersion2

pkg/meta/model/job_args.go
D3Hunter: lgtm

pkg/ddl/executor.go
```diff
@@ -5646,6 +5657,38 @@ func (e *executor) UnlockTables(ctx sessionctx.Context, unlockTables []model.Tab
...
+	}
+
+	job := &model.Job{
+		Version:        model.JobVersion2,
+		SchemaID:       args.SchemaID,
+		TableID:        args.TableID,
+		Type:           model.ActionAlterTableMode,
+		BinlogInfo:     &model.HistoryInfo{},
+		CDCWriteSource: ctx.GetSessionVars().CDCWriteSource,
+		// TODO(xiaoyuan): when is this information needed?
```
D3Hunter, line 5680: add TODO for something you need or better do it later, not questions

fishiu: I forgot to read the definition of InvolvingSchemaInfo and now I understand it. This initialization is exactly the default one, thus can be deleted.

pkg/ddl/executor.go
```diff
@@ -1202,6 +1212,7 @@ func (e *executor) CreateTableWithInfo(
 	if err != nil {
 		// table exists, but if_not_exists flags is true, so we ignore this error.
 		if c.OnExist == OnExistIgnore && infoschema.ErrTableExists.Equal(err) {
+			// TODO(xiaoyuan): I think this branch will never be reached, please see e.createTableWithInfoJob
```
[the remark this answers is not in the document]
fishiu: 1. Sorry, actually I was planning to delete those TODOs when I remove the wip label. But it is not a good habit, and I will keep these kinds of reminders in my notebook rather than inappropriate TODOs that may be committed by mistake.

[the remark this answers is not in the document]
fishiu: 2. Acually I mean that in function `createTableWithInfoJob` this situation (tableExist) is already handled so this `if branch` will never be reached.

pkg/ddl/executor.go
```diff
@@ -1059,6 +1060,15 @@ func (e *executor) createTableWithInfoJob(
 		switch cfg.OnExist {
 		case OnExistIgnore:
 			ctx.GetSessionVars().StmtCtx.AppendNote(err)
+			// If table exists, changing from Normal/Import to Restore is not allowed.
```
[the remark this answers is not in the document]
fishiu: My bad I should not use the word conversion here.
When creating table with ModeRestore/ModeImport(currently only triggered by BR) and under `ignore exist` option , we need to ensure the existing table is already in ModeRestore, otherwise not allowed.

I updated the comments here and please check if it makes sense now.

pkg/ddl/executor.go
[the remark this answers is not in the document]
D3Hunter: > 2. Acually I mean that in function `createTableWithInfoJob` this situation (tableExist) is already handled, so this `if branch` will never be reached.

there is some case to reach here as I have commented before

pkg/ddl/executor.go
```diff
@@ -118,6 +118,7 @@ type Executor interface {
 	RenameTable(ctx sessionctx.Context, stmt *ast.RenameTableStmt) error
 	LockTables(ctx sessionctx.Context, stmt *ast.LockTablesStmt) error
 	UnlockTables(ctx sessionctx.Context, lockedTables []model.TableLockTpInfo) error
+	AlterTableMode(ctx sessionctx.Context, args *model.AlterTableModeArgs) error
```
Tristan1900, line 121: maybe I'm missing something, is there a good way to alter table mode in batches? similar to op like `BatchCreateTableWithInfo `. right now we can create tables with table mode in batches but at the end of the restore we need to set tables back to normal and without batching it's probably going to be slow. thanks!

Tristan1900 (commented):
thanks for the change! just one question from my side

BornChanger:
@fishiu @River2000i please fill the description of the pr, including the problem description and what's change.

pkg/ddl/executor.go
River2000i: > is there a good way to alter table mode in batches

for sure, but we think support modify table mode for one single table as first pr is better. We will support alter table mode in batches in next pr(not difficult implement).

fishiu: Seems @Tristan1900 's current PR for BR requires this batch feature. I will try adding the batch DDL job soon (hopefully this Thursday).

pkg/ddl/executor.go
```diff
@@ -5650,6 +5661,31 @@ func (e *executor) UnlockTables(ctx sessionctx.Context, unlockTables []model.Tab
 	return errors.Trace(err)
 }
 
+func (e *executor) AlterTableMode(ctx sessionctx.Context, args *model.AlterTableModeArgs) error {
+	is := e.infoCache.GetLatest()
+
+	schema, ok := is.SchemaByID(args.SchemaID)
```
Tristan1900, line 5667: here is a request my side, I should also update in the design doc or talk to someone in the team. My understanding is that when altering table mode, ddl will first check from cache if that table exists and if not it will bail early. The problem is that during log restore we will likely create a bunch of new tables, for example exchange partition, that a previous partition was exchanged out to be a new regular table. In this case we cannot create restore mode tables upfront since the global id will conflicts with the partition id.

I am wondering when altering table mode from restore back to normal, can we skip the check and directly go to TiKV to fetch the value, if there is no value on TiKV then we error out? This behavior is kind of crucial to us since without it we cannot load all the restored tables from TiKV.  please let me know your thoughts. thanks!

pkg/ddl/executor.go
fishiu: @D3Hunter I had some difficulty when implementing batchAlterTableModes. In the single table situation, it is easy to update table infomation with the function `updateVersionAndTableInfo` in ddl/table.go which handle one table at a time by calling `UpdateTable`. But I think here it is not good to call UpdateTable in a forloop and instead I may need something like a `batchUpdateTable` for atomicity. Could you please me give me some advice on this?

pkg/ddl/executor.go
D3Hunter: > create a bunch of new tables directly on TiKV without going through DDL, for example exchange partition

if it's a must to bypass DDL, why not set the table mode directly when write the meta-kv?

pkg/ddl/table_mode_test.go
```diff
@@ -0,0 +1,272 @@
...
+	// For testing insert is not allowed when table is in ModeImport
+	tk.MustGetErrCode("insert into t1_restore values(1)", errno.ErrProtectedTableMode)
+
+	// For testing accessing table metadata is allowed when table is in ModeRestore
+	tk.MustExec("show create table t1_restore")
+	tk.MustExec("describe t1_restore")
+
+	// For testing AlterTable ModeRestore -> ModeImport is not allowed
+	err = setTableModeTest(ctx, t, store, de.(ddl.ExecutorForTest), dbInfo, tblInfo, model.TableModeImport)
+	checkErrorCode(t, err, errno.ErrInvalidTableModeSet)
```
D3Hunter, line 145: can use `require.ErrorContains`

pkg/ddl/table_mode_test.go
```diff
@@ -0,0 +1,272 @@
...
+	tblInfo := getClonedTableInfoFromDomain(t, "test", "t1", domain)
+
+	// For testing create table as ModeRestore
+	tblInfo.Name = ast.NewCIStr("t1_restore")
+	tblInfo.TableMode = model.TableModeRestore
+	err := de.CreateTableWithInfo(tk.Session(), ast.NewCIStr("test"), tblInfo, nil, ddl.WithOnExist(ddl.OnExistIgnore))
+	require.NoError(t, err)
+	dbInfo, ok := domain.InfoSchema().SchemaByName(ast.NewCIStr("test"))
+	require.True(t, ok)
+	checkTableModeTest(t, store, dbInfo, tblInfo, model.TableModeRestore)
```
D3Hunter, line 131: add a case to test that if that table already exist with mode `import`, then BR try to create the same table with mode `restore` and `ddl.OnExistIgnore`, an error should be reported

pkg/ddl/table_mode_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+	// For testing select is not allowed when table is in ModeImport
+	tk.MustGetErrCode("select * from t1_restore", errno.ErrProtectedTableMode)
+
+	// For testing insert is not allowed when table is in ModeImport
+	tk.MustGetErrCode("insert into t1_restore values(1)", errno.ErrProtectedTableMode)
```
D3Hunter, line 137: add for `update/delete`, and all modifying DDLs, such as modify column/add column/add index, drop table, truncate table, etc.

pkg/ddl/table_mode_test.go
```diff
@@ -0,0 +1,272 @@
...
+		BinlogInfo: &model.HistoryInfo{},
+		InvolvingSchemaInfo: []model.InvolvingSchemaInfo{
+			{
+				Database: dbInfo.Name.L,
+				Table:    tblInfo.Name.L,
+			},
+		},
+	}
+	ctx.SetValue(sessionctx.QueryString, "skip")
+	err := de.DoDDLJobWrapper(ctx, ddl.NewJobWrapperWithArgs(job, args, true))
```
D3Hunter, line 75: please use `AlterTableMode`, not calling DoDDLJobWrapper directly

tests/integrationtest/r/infoschema/infoschema.result
```diff
@@ -54,11 +54,11 @@ count(*)
 3
 desc format='brief' SELECT count(*) FROM information_schema.TABLES WHERE (TABLE_SCHEMA= 'mysql' or TABLE_SCHEMA = 'test') and (TABLE_NAME = 't1' or TABLE_NAME = 't2');
 id	estRows	task	access object	operator info
-HashAgg	1.00	root		funcs:count(1)->Column#26
+HashAgg	1.00	root		funcs:count(1)->Column#27
```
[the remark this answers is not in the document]
fishiu: Because I added a new column to information_schema so the number of columns increased.

pkg/ddl/executor.go
Leavrth: When another create table SQL `CREATE TABLE t1` from user side, will DDL check the `t1` meta-kv with restore mode despite it isn't loaded in infoschema?

Tristan1900: @D3Hunter it is set directly when writing meta kv, the question is after restore we need info schema to get updated by releasing restore mode, with that safety check, the alter table mode to normal will be rejected since infoshcmea cache does not have the entry, that's my understanding.

D3Hunter: > When another create table SQL CREATE TABLE t1 from user side, will DDL check the t1 meta-kv with restore mode despite it isn't loaded in infoschema?

for CREATE-TABLE, DDL checks table existence by info schema only now, see https://github.com/pingcap/tidb/pull/53882 for more info, if it's not reloaded due to BR updates meta kv directly, it might be overridden. that's why we discourage bypassing DDL to modify meta-kv

> so we need a way to directly fetch things from TiKV and only error out if not found on TiKV

can we fix `In this case we cannot create restore mode tables upfront since the global id will conflicts with the partition id.`, so we can go through DDL?

Tristan1900: I am not sure how to go through DDL in this case. To be more clear, the problem is if we have a partitioned tableID=1, partID=2, and an individual tableID=3, and the exchange partition happen and after restore it becomes tableID=1, partID=3, and individual tableID=2. 
If we go through DDL, we need two tables with restore mode created, tableID=1 and tableID=3, but after restore, tableID=3 becomes the partition, and partID=2 becomes the table, so it won't work to alter the table mode. If we create tableID=1, and tableID=2 instead at the very beginning and ignore tableID=3, the id will conflicts since there is a partition and a table using same ID. I couldn't think of a good way to solve it or maybe I'm misunderstanding the meaning of going through DDL. I'm happy to chat more online or offline if you can get a chance. Thanks

pkg/meta/meta_test.go
```diff
@@ -738,7 +738,7 @@ func TestIsTableInfoMustLoadSubStringsOrder(t *testing.T) {
...
+	expect := `{"id":0,"name":{"O":"","L":""},"charset":"","collate":"","cols":null,"index_info":null,"constraint_info":null,"fk_info":null,"state":0,"pk_is_handle":false,"is_common_handle":false,"common_handle_version":0,"comment":"","auto_inc_id":0,"auto_id_cache":0,"auto_rand_id":0,"max_col_id":0,"max_idx_id":0,"max_fk_id":0,"max_cst_id":0,"update_timestamp":0,"ShardRowIDBits":0,"max_shard_row_id_bits":0,"auto_random_bits":0,"auto_random_range_bits":0,"pre_split_regions":0,"partition":null,"compression":"","view":null,"sequence":null,"Lock":null,"version":0,"tiflash_replica":null,"is_columnar":false,"temp_table_type":0,"cache_table_status":0,"policy_ref_info":null,"stats_options":null,"exchange_partition_info":null,"ttl_info":null,"revision":0,"table_mode":0}`
```
River2000i, line 741: tablemode is not a required field and does not need to be checked

pkg/ddl/executor.go
D3Hunter: thanks for your detailed explain, I guess we need a way to block user DDLs by table name(s), not by ID like current `alter table mode` does, but we don't have a way to bind info to table name now, maybe later we can design a data dictionary table for this purpose, cc @Benjamin2037

pkg/planner/core/optimizer.go
```diff
@@ -248,6 +248,29 @@ func CheckTableLock(ctx tablelock.TableLockReadContext, is infoschema.InfoSchema
...
+// CheckTableMode checks if the table is accessible by table mode, only TableModeNormal can be accessed.
+func CheckTableMode(node *resolve.NodeW) error {
+	// First make exceptions for stmt that only visit table meta;
+	// For example, `describe <table_name>` and `show create table <table_name>`;
+	// These exceptions can be simply categorized as `ast.ShowStmt`;
+	if _, ok := node.Node.(*ast.ShowStmt); ok {
+		return nil
+	}
+
+	tableList := ExtractTableList(node, false)
```
D3Hunter, line 260: is this needed, i guess we already visited the node before calling this

pkg/ddl/table_mode.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	tbInfo.Mode = args.TableMode
+	return nil
+}
+
+func checkTableMode(origin, target model.TableMode) bool {
+	if target == model.TableModeImport {
+		// only transition from ModeNormal to ModeImport is allowed
+		if origin != model.TableModeNormal {
```
D3Hunter, line 77: I forget whether we have discuss this, but why `import` -> `import` report error?

pkg/ddl/executor.go
```diff
@@ -5679,6 +5689,38 @@ func (e *executor) UnlockTables(ctx sessionctx.Context, unlockTables []model.Tab
...
+	if !ok {
+		return infoschema.ErrDatabaseNotExists.GenWithStackByArgs(fmt.Sprintf("SchemaID: %v", args.SchemaID))
+	}
+
+	table, ok := is.TableByID(e.ctx, args.TableID)
+	if !ok {
+		return infoschema.ErrTableNotExists.GenWithStackByArgs(schema.Name, args.TableID)
+	}
+
+	ok = checkTableMode(table.Meta().Mode, args.TableMode)
```
D3Hunter, line 5705: we can return directly if `normal` -> `normal`

pkg/planner/core/optimizer.go
River2000i: let me double check later.

pkg/ddl/table_mode.go
River2000i: `import` and `restore` only allow convert to `normal`. Just like `import` is not allow convert to `restore`. I think `import into` will not create table twice, if the task failed will convert to `cancel` and DXF will alter table mode to `normal`🤔

pkg/ddl/executor.go
River2000i: get it, i will address later and also few UT still need to fix.(another thing blocking me...)

pkg/ddl/executor.go
D3Hunter: I think I misunderstand the key issue here

reuse the case you described, suppose
- we have partition table `p` with id=1 which have a partition with id=2, we also have non-partition table `np` with id=3
- upstream did a exchange partition, we need to sync to downstream with BR log restore, the log should contains below meta KV operation
  - update `meta table id 1 -> {partition id 3}`
  - delete `meta table id 3 -> {name: np,....}`
  - insert `meta table id 2 -> {name: np,....}`
- since BR might affects table `p`/`np`, it sets them to `restore` mode
- BR write above meta changes directly to TiKV bypass DDL, it cause `meta kv` inconsistent with loaded info schema in TiDB, i.e. TiDB thinks id of `np` is 3, but in `meta kv`, it's 2
  - during the restore, BR also set `meta table id 2 -> {name: np,....}` to `restore` mode
- now BR want to restore the mode back, but TiDB doesn't recognize `id=2`. <<<--- this's why you say `partID=2 becomes the table, so it won't work to alter the table mode.`

the key here is BR need another step before restore the mode back, to make sure info schema in TiDB consistent with `meta KV`, either to trigger a full load, or a special DDL which describes what BR have done to trigger diff load

> If we create tableID=1, and tableID=2 instead at the very beginning

this is not possible, we can only set the mode on table, not partition.

Tristan1900: > the key here is BR need another step before restore the mode back, to make sure info schema in TiDB consistent with `meta KV`, either to trigger a full load, or a special DDL which describes what BR have done to trigger diff load
yes exactly, that's the issue I was referring to. However it's not possible to do full reload, the entire feature is mainly for online system, if we do full reload it's going to cause heavy stall for every user, (we noticed the full reload took more than 1 min for a million table size cluster). 
yes and I thought the alter table mode was something like that since it's only used by BR, or at least `restoreMode` is only used by BR. When altering  `restoreMode` back to normal, I was hoping it does something like the `special DDL` you mentioned, because BR knows what's on the TiKV, it just need a DDL to fetch them to info schema without doing a full reload. 
Do you think alter restore mode can do that? Like what we discussed yesterday, that it directly go to TiKV and fetch the tableInfo and do some consistency check with the current info schema, if there is a conflict, it can bail out with an error message saying `table name conficlts` etc. thanks

pkg/planner/core/optimizer.go
River2000i: since `node.GetResolveContext().GetTableName` need the address of `tableName` to get the table info, i go through the context and didn't find which part visited the `tableName` before. Feel free to comment if there is good idea.

D3Hunter: preprocess have done it, you don't need the `tableName`, the `TableInfo` has already stored in the resolve context

pkg/ddl/table_mode.go
D3Hunter: > import into will not create table twice

import-into is only for existing table

pkg/ddl/executor.go
```diff
@@ -1062,6 +1063,15 @@ func (e *executor) createTableWithInfoJob(
 		switch cfg.OnExist {
 		case OnExistIgnore:
 			ctx.GetSessionVars().StmtCtx.AppendNote(err)
+			// if target TableMode is ModeRestore, we check if the existing mode is consistent the new one
+			if tbInfo.Mode == model.TableModeRestore {
+				oldTableMode := oldTable.Meta().Mode
+				if oldTableMode != model.TableModeRestore {
+					// Indeed this is not a conversion problem but an inconsistency problem.
```
D3Hunter, line 1070:
```suggestion
```

pkg/ddl/executor.go
```diff
@@ -4143,6 +4153,10 @@ func (e *executor) dropTableObject(
 		} else if err != nil {
 			return err
 		}
+		// Only table mode is TableModeNormal can be dropped table
+		if tableInfo.Meta().Mode != model.TableModeNormal {
+			return infoschema.ErrProtectedTableMode.GenWithStackByArgs(tableInfo.Meta().Name, tableInfo.Meta().Mode)
+		}
```
D3Hunter, line 4159: abstract into a function like `checkTableMode(tbl *TableInfo) error`

D3Hunter: also comment why they are handled specially

pkg/planner/core/optimizer.go
River2000i: since i want to use `GetTableName()`, which need the `tableName` to access `TableInfo`. Maybe i can add a function for resolve context to get all `TableInfo`?

D3Hunter: yes, add a method to return the inner map is ok

River2000i: address in https://github.com/pingcap/tidb/pull/59009/files#diff-168c33d16a6ad85e67cafc2db1a37a3b5e8f10dfb5333ff98712954117069020R261-R265

pkg/ddl/table_mode.go
River2000i: address in https://github.com/pingcap/tidb/pull/59009/files#diff-bd1c6049cf8b74ff0c3051c788691b679d001b57a938d8e1c348add9a47c80beR74-R86

now only block `import` -> `restore`, `restore` -> `import`. Reset modify is allow

pkg/ddl/executor.go
River2000i: address in https://github.com/pingcap/tidb/pull/59009/files#diff-2f76112c3fadcfa799a40188b2962bc214b08218c0cd9b393a8a2be796f97bbdR7041

pkg/ddl/executor.go
```diff
@@ -6974,3 +7030,17 @@ func NewDDLReorgMeta(ctx sessionctx.Context) *model.DDLReorgMeta {
 		Version:           model.CurrentReorgMetaVersion,
 	}
 }
+
+// checkTableMode checks the table mode is TableModeNormal or not.
+// This is a special validation for tablemode during the DDL phase. Originally, reads and writes to non-normal
+// tables were prohibited during the optimize phase of execution plan generation.
+// However, the current approach relies on the `tableNameW` recorded in the ResolveContext
+// during the preprocessing phase to store the tables involved in the statement,
+// and then checks whether the table mode is normal. However, for some statements, `tableNameW` is not recorded
```
D3Hunter, line 7039:
```suggestion
// and then checks whether the table mode is normal. However, for some statements, `tableNameW` is not recorded as those table might not exists, such as for rename table DDL
```

pkg/ddl/executor.go
```diff
@@ -6974,3 +7030,17 @@ func NewDDLReorgMeta(ctx sessionctx.Context) *model.DDLReorgMeta {
 		Version:           model.CurrentReorgMetaVersion,
 	}
 }
+
+// checkTableMode checks the table mode is TableModeNormal or not.
```
D3Hunter, line 7034: please format the comment a bit to be around 80 chars per line

pkg/ddl/executor.go
```diff
@@ -5679,6 +5700,41 @@ func (e *executor) UnlockTables(ctx sessionctx.Context, unlockTables []model.Tab
...
+	table, ok := is.TableByID(e.ctx, args.TableID)
+	if !ok {
+		return infoschema.ErrTableNotExists.GenWithStackByArgs(schema.Name, args.TableID)
+	}
+
+	ok = validateTableMode(table.Meta().Mode, args.TableMode)
+	if !ok {
+		return infoschema.ErrInvalidTableModeSet.GenWithStackByArgs(table.Meta().Mode, args.TableMode, table.Meta().Name.O)
+	}
+	if table.Meta().Mode == model.TableModeNormal && args.TableMode == model.TableModeNormal {
```
D3Hunter, line 5720:
```suggestion
	if table.Meta().Mode == args.TableMode {
```

pkg/ddl/executor.go
```diff
@@ -4143,6 +4152,10 @@ func (e *executor) dropTableObject(
 		} else if err != nil {
 			return err
 		}
+		// Only table mode is TableModeNormal can be dropped table
```
D3Hunter, line 4155:
```suggestion
```
i think the comment in `checkTableMode` is enough, please change other callings too

pkg/meta/model/table.go
```diff
@@ -660,6 +662,32 @@ func (t TableLockState) String() string {
 	}
 }
 
+// TableMode is the state for table mode.
```
D3Hunter, line 665: please explain more about the purpose and behavior of the mode

pkg/ddl/table_mode.go
```diff
@@ -0,0 +1,86 @@
...
+	metaMut := jobCtx.metaMut
+	tbInfo, err = GetTableInfoAndCancelFaultJob(metaMut, job, job.SchemaID)
+	if err != nil {
+		return ver, err
+	}
+
+	switch tbInfo.Mode {
+	case model.TableModeNormal, model.TableModeImport, model.TableModeRestore:
+		// skip updateVersionAndTableInfo when changing table mode from TableModeNormal to TableModeNormal
+		if tbInfo.Mode == model.TableModeNormal && args.TableMode == model.TableModeNormal {
```
D3Hunter, line 39:
```suggestion
		if tbInfo.Mode == args.TableMode {
```

pkg/planner/core/optimizer.go
```diff
@@ -248,6 +248,25 @@ func CheckTableLock(ctx tablelock.TableLockReadContext, is infoschema.InfoSchema
...
+	// First make exceptions for stmt that only visit table meta;
+	// For example, `describe <table_name>` and `show create table <table_name>`;
+	switch node.Node.(type) {
+	case *ast.ShowStmt, *ast.ExplainStmt:
+		return nil
+	default:
+	}
+
+	for _, tblNameW := range node.GetResolveContext().GetTableNames() {
+		if tblNameW.TableInfo.Mode == model.TableModeImport || tblNameW.TableInfo.Mode == model.TableModeRestore {
```
D3Hunter, line 262:
```suggestion
		if tblNameW.TableInfo.Mode != model.TableModeNormal {
```

pkg/ddl/executor.go
River2000i: fixed in https://github.com/pingcap/tidb/pull/59009/files#diff-2f76112c3fadcfa799a40188b2962bc214b08218c0cd9b393a8a2be796f97bbdR7040

pkg/ddl/executor.go
River2000i: fixed

pkg/meta/model/table.go
River2000i: add comment at https://github.com/pingcap/tidb/pull/59009/files#diff-1730439e8c1b7d1a7019391006a2c0cb2f455b9de48769c9b6097c9c0a2b995eR665-R671

River2000i:
@hawkingrei @winoros PTAL~

wjhuang2016 (approved):

hawkingrei (approved):

PR #59046, opened 2025-01-20
ddl: part 1 of add-index integrating dxf dynamic modify param

D3Hunter:

Issue Number: ref #57497, ref #57229

Problem Summary:

**This is the first part, will add integrate test later**
- add a routine in `executeDistTask` to detect whether there are param mismatch between DDL job and DXF task, and we call `ModifyTaskByID` to let DXF handle the rest param modify work if detected
  - there is a way to achieve this without start a routine, but requires refactor transitOneJobStep to start txn on need first

```
// Note: we can achieve the same effect by calling ModifyTaskByID directly inside
// the process of 'ADMIN ALTER DDL JOB xxx', so we can eliminate the goroutine,
// but if the task hasn't been created we need to make sure the task is created
// with config after ALTER DDL JOB is executed. A possible solution is to make
// the DXF task submission and 'ADMIN ALTER DDL JOB xxx' txn conflict with each
// other when they overlap in time, by modify the job at the same time when submit
// task, as we are using optimistic txn. But this will cause WRITE CONFLICT with
// outer txn in transitOneJobStep.
```

- remove some useless code such as `skipWriteBinlog`
- some refactor such as create static errors using `errors.New` of golang to avoid stack overhead of `pingcap/errors` and replace error equal check with `goerrors.Is`
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


D3Hunter:
@lance6716 @tangenta

lance6716 (commented):
7 / 19 files viewed

D3Hunter:
@lance6716 ptal

pkg/ddl/index.go
```diff
@@ -2616,10 +2643,125 @@ func (w *worker) executeDistTask(stepCtx context.Context, t table.Table, reorgIn
...
+		modifyTaskParamLoop(ctx, jobCtx.sysTblMgr, taskManager, done,
+			reorgInfo.Job.ID, taskID, lastConcurrency, lastBatchSize, lastMaxWriteSpeed)
+		return nil
+	})
+
 	err = g.Wait()
 	return err
 }
 
+// Note: we can achieve the same effect by calling ModifyTaskByID directly inside
```
lance6716, line 2657: I didn't review most of DXF codes so I'm not sure what's the meaning od this function comment. Just to check: 

it says we only allow some functions to write to the system table of DXF, especially, we don't allow ADMIN ALTER DDL JOB to write because it's heavier to rollback on conflict?

Can you explain in this design, which functions are allowed to write the system table? I want to check if they still can conflict with each other.

pkg/ddl/index.go
```diff
@@ -2616,10 +2643,125 @@ func (w *worker) executeDistTask(stepCtx context.Context, t table.Table, reorgIn
...
+			logger.Error("get task failed, will retry later", zap.Error(err))
+			continue
+		}
+		if !currTask.State.CanMoveToModifying() {
+			// user might modify param again while another modify is ongoing.
+			logger.Info("task state is not suitable for modifying, will retry later",
+				zap.String("state", currTask.State.String()))
+			continue
+		}
+		if err = taskManager.ModifyTaskByID(ctx, taskID, &proto.ModifyParam{
```
lance6716, line 2738: when checking the implementation of `ModifyTaskByID` I see the "old values" are not added to the UPDATE .. WHERE clause. So double check that it is expected for two concurrent call of `ModifyTaskByID`, the final task arguments can be updated as any of the call?

pkg/ddl/index.go
```diff
@@ -2616,10 +2643,125 @@ func (w *worker) executeDistTask(stepCtx context.Context, t table.Table, reorgIn
...
+		lastMaxWriteSpeed = maxWriteSpeed
+	}
+}
+
+func adjustConcurrency(ctx context.Context, taskMgr storage.Manager, workerCnt int) (int, error) {
+	cpuCount, err := taskMgr.GetCPUCountOfNode(ctx)
+	if err != nil {
+		return 0, err
+	}
+	return min(workerCnt, cpuCount), nil
```
lance6716, line 2762: Can you explain the reason why we limit the concurrency? If the work is not a pure compute workload, increase the concurrency to more that CPU count can make use of IO time.

lance6716 (commented):
rest lgtm

pkg/ddl/index.go
D3Hunter: ModifyTaskByID make transit the task into `modifying` state if it's in the prev state, so only 1 will success for concurrent calls

pkg/ddl/index.go
D3Hunter: only abstract logic in here https://github.com/pingcap/tidb/pull/59046/files#diff-45ad039a8b87835b475a52615355c444538d3a1f31a84d7b3ac94e15b1ce8721L2550-L2554

valid range of `tidb_ddl_reorg_worker_cnt` is [1, 256], but for a DXF task, valid range of task concurrency or slots is [1, cpuCount]

D3Hunter:
ping @lance6716

lance6716 (approved):

pkg/ddl/index.go
```diff
@@ -2616,10 +2643,125 @@ func (w *worker) executeDistTask(stepCtx context.Context, t table.Table, reorgIn
...
+// outer txn in transitOneJobStep.
+func modifyTaskParamLoop(
+	ctx context.Context,
+	sysTblMgr systable.Manager,
+	taskManager storage.Manager,
+	done chan struct{},
+	jobID, taskID int64,
+	lastConcurrency, lastBatchSize, lastMaxWriteSpeed int,
+) {
+	logger := logutil.DDLLogger().With(zap.Int64("jobId", jobID), zap.Int64("taskId", taskID))
```
tangenta, line 2673:
```suggestion
	logger := logutil.DDLLogger().With(zap.Int64("jobID", jobID), zap.Int64("taskID", taskID))
```

pkg/ddl/index.go
```diff
@@ -2616,10 +2643,125 @@ func (w *worker) executeDistTask(stepCtx context.Context, t table.Table, reorgIn
...
+// other when they overlap in time, by modify the job at the same time when submit
+// task, as we are using optimistic txn. But this will cause WRITE CONFLICT with
+// outer txn in transitOneJobStep.
+func modifyTaskParamLoop(
+	ctx context.Context,
+	sysTblMgr systable.Manager,
+	taskManager storage.Manager,
+	done chan struct{},
+	jobID, taskID int64,
+	lastConcurrency, lastBatchSize, lastMaxWriteSpeed int,
```
tangenta, line 2671: Maybe use "old" / "new" to replace "last" / "latest"?

tangenta (approved):

pkg/ddl/index.go
D3Hunter: see https://github.com/pingcap/tidb/pull/59343

PR #59110, opened 2025-01-22
planner: notify skyline pruning of index statistics availability

### LINKED ISSUE #57948: Update table statistics before publishing new schema version after DDL
## Feature Request

After DDL on large 800GB table, statistics were cleared, leading to degraded performance cause by bad query plans (eg: full table scans).

> The root cause is that after the Modify-Col DDL, the old statistics related to this modified column will no longer be used, and below is the 2 reason of this behavior:
>
>The old statistics is collected based on the column's prior type so the prior stats-data is not compatible with the new column type, for example, after changing the column type from INT to VARCHAR(32), its prior stats-data based on INT can't be used to estimate for predicates like "col = 'xxx'", so we have to collect the stats again;
>
>When modifying the column type, TiDB will create a new column (or an index) with a new colID (or idxID) to replace the old one, and TiDB uses the colID (or idxID) to access its statistics, so after the DDL, TiDB will not be able to find its prior statistics.

TiDB will collect new statistics just before it publishes the final schema version, so that after new schema goes live, stats are already collected and up to date. Something like this could work:

1. DDL is started and goes through normal stages
2. While DDL is in progress, application still sees original schema version and uses original statistics
3. DDL is finished and new schema version is ready to be published 
4. Before publishing, collect statistics against the new version
5. Now publish the new schema version, discard old stats and use new stats collected in 4.
6. This means that there would be no bad query plans as we don’t need to wait for new stats, they were preemptively collected.

It was suggested to use plan binding feature to prevent bad queries. Such workaround is very management heavy and not really feasible to find every query hitting the table and try to make a Plan Binding preemptively. 

Currently large DDLs with MODIFY COLUMN are somewhat of a trap, as they have potential to cause cluster performance degradation.
This feature would prevent bad query plans with outdated stats being executed after large DDLs. Currently, such bad plans are be used until ANALYZE TABLE is run manually or it is picked up by auto analyze.

terry1purcell:
Extend support for skyline pruning recognition of index without statistics to existing rules. 

By passing the "pseudo" result of the index from the CompareCandidates function for the "successful" index - it allows the later checking in SkyLinePruning to keep that index to prioritize above a table scan if tidb_opt_prefer_range_scan global variable is set.

This change is a minor addition to address prior customer concerns that a newly created index that does NOT yet have statistics may be selected by the existing CompareCandidates code - and the index WITH statistics is pruned. But the cost of the index that used defaults (since no statistics) may have a higher cost than tablescan.

Issue Number: close #57948

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/find_best_task.go
+++ b/pkg/planner/core/find_best_task.go
@@ -789,10 +789,10 @@ func compareCandidates(sctx base.PlanContext, statsTbl *statistics.Table, tableI
 		threshold := float64(fixcontrol.GetIntWithDefault(sctx.GetSessionVars().OptimizerFixControl, fixcontrol.Fix45132, 1000))
 		if threshold > 0 { // set it to 0 to disable this rule
 			if lhs.path.CountAfterAccess/rhs.path.CountAfterAccess > threshold {
-				return -1, false
+				return -1, rhsPseudo
 			}
 			if rhs.path.CountAfterAccess/lhs.path.CountAfterAccess > threshold {
-				return 1, false
+				return 1, lhsPseudo
 			}
 		}
 	}
@@ -811,10 +811,10 @@ func compareCandidates(sctx base.PlanContext, statsTbl *statistics.Table, tableI
 		return 0, false
 	}
 	if accessResult >= 0 && scanResult >= 0 && matchResult >= 0 && globalResult >= 0 && sum > 0 {
-		return 1, false
+		return 1, lhsPseudo
 	}
 	if accessResult <= 0 && scanResult <= 0 && matchResult <= 0 && globalResult <= 0 && sum < 0 {
-		return -1, false
+		return -1, rhsPseudo
 	}
 	return 0, false
 }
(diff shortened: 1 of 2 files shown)


hawkingrei (approved):

pkg/planner/core/find_best_task.go
```diff
@@ -811,10 +811,10 @@ func compareCandidates(sctx base.PlanContext, statsTbl *statistics.Table, tableI
 		return 0, false
 	}
 	if accessResult >= 0 && scanResult >= 0 && matchResult >= 0 && globalResult >= 0 && sum > 0 {
-		return 1, false
+		return 1, lhsPseudo
```
winoros, line 814: should we add some comments for this two if block?

terry1purcell: Done

winoros (approved):

PR #59122, opened 2025-01-22
server: add initial status server to provide debug functionalities before starting server

### LINKED ISSUE #59087: Enable the profiling, goroutines, metrics endpoints and signal handler as soon as possible
## Enhancement

For now, during the TiDB boot process, the observability infrastructures are not started, which makes it pretty hard for us to investigate any issue before starting the server. I think the following ones are especially important and helpful for debug:

1. `/debug/pprof`, especially `/debug/pprof/profile` and `/debug/pprof/goroutine`. If the TiDB is blocked in the boot stage, get the backtrace of all goroutines and help us to understand where it is blocked.
2. `/metrics`. Without it, the prometheus cannot scrape the metrics of the booting TiDB server.
3. `SIGUSR1` signal handler. After starting the server, the TiDB server will handle `SIGUSR1` signal by printing the backtrace of all goroutines. Without it, all other signals which will print the backtrace like `SIGQUIT` will kill the process.

## Current behavior

All of them are started after creating `storage`, `dom`, `server`, etc.

```go
storage, dom := createStoreDDLOwnerMgrAndDomain(keyspaceName)
repository.SetupRepository(dom)
svr := createServer(storage, dom)

exited := make(chan struct{})
signal.SetupSignalHandler(func() {
	svr.Close()
	resourcemanager.InstanceResourceManager.Stop()
	cleanup(svr, storage, dom)
	cpuprofile.StopCPUProfiler()
	executor.Stop()
	close(exited)
})
topsql.SetupTopSQL(svr)
terror.MustNil(svr.Run(dom))
```

The signal handler is setup in `signal.SetupSignalHandler`. The status server is setup in `svr.Run(dom)`.

## Change

To overcome this inconvenience, I propose to make the following changes:

1. Create a temporary status server including metrics and `/debug/pprof` before `svr.Run(dom)`, and stop it before creating the new fully functional status server.
2. Setup the signal handler of `SIGUSR1` earlier.

Both of them shouldn't depend on other functions of TiDB, so it's fine to start them in the very beginning of `main()` function.

## Note

As they shouldn't depend on other tidb packages, the result can have tiny difference between the real one:

1. The `/debug/pprof/profile` may include more labels for a goroutine, because the `topsql` function will include the sql/plan signature in the label, and a customized endpoint will clean these labels for profiling. However, the initial one will not depend on `topsql` related things and will not clean these labels.

YangKeao:

Issue Number: close #59087

Problem Summary:

The `/debug/pprof/xxx` endpoints in status server and the signal handler of `SIGUSR1` are very helpful for the developers to debug running tidb-server. However, they are only available after starting the server. If the TiDB server is somehow blocked before starting successfully (like loading `InfoSchema` for a very long time), it'll be hard for us to investigate the issue.

1. Set the signal handler before most of the bootstrap logic.
2. Start an initial status server which only handles several `/debug/pprof/xxx` interface and `/metrics` interface.
3. Stop the initial status server before starting the real fully functional status server.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/server/http_status.go
```diff
@@ -637,3 +630,85 @@ func (s *Server) newStatsPriorityQueueHandler() *optimizor.StatsPriorityQueueHan
...
+// StopInitialStatusServer tries to stop the initial status server. It should be called before starting the fully functional status server
+func StopInitialStatusServer() {
+	if initialStatusServer == nil {
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), stopInitialStatusServerTimeout)
+	defer cancel()
+
+	err := initialStatusServer.Shutdown(ctx)
```
YangKeao, line 705: The only concern is that the requests to `/debug/pprof/profile` will block the `Shutdown` for a long time (so I gave it a `5s` timeout).

It's also fine to `Close` directly IMO, but I'm not sure which one is better.

tiancaiamao (approved):
LGTM

PR #59131, opened 2025-01-22
statistics: BinarySearchRemoveVal should use decoded val instead of encoded

### LINKED ISSUE #59130: wrong stats of TestIssues24349
## Bug Report

Image

Looking into column `b`'s stats:
- partition `p0`: TopN: (3, 3). Hist: one bucket  [2, 2] with count = 1
- partition `p1`: TopN: (2, 3), Hist: empty
- partition `p2`: TopN: (1, 2), Hist: two buckets [1, 3] with count = 2 and repeat = 1, [4, 4] with count = 1 and repeat = 1

After fixing the global TopN, it is now 2 with 4 as its occurrence (local TopN of `p1` and one single bucket from `p0`).

Remained buckets(TopN also treat as bucket)
- [3, 3] count = 4, repeat = 4 (local TopN from `p0` plus 1 from `p2`'s bucket [1, 3]'s repeat)
- [1, 1] count = 2, repeat = 2
- [1, 3] count = 1, repeat = 0
- [4, 4] count = 1, repeat = 1

After merging them to get the global histogram, we should get one bucket [1, 4] with count = 8 and repeat = 1.
But in the current test result, we have

Image
The count is ***10***.

master.

winoros:

Issue Number: close #59130 

Problem Summary:

The function used the wrong val as its parameter.
As a result, when the column is not string type. The function `BinaryRemoveVal` actually removes nothing.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
修复合并全局统计信息时，直方图构造结果可能变大的问题
Fix the issue that the global histogram might be larger than expected when merging partition table's global statistics.
```

### DIFF
--- a/pkg/statistics/histogram_test.go
+++ b/pkg/statistics/histogram_test.go
@@ -712,21 +712,3 @@ func generateData(t *testing.T) *Histogram {
 	}
 	return genHist4Test(t, data, 0)
 }
-
-func TestVerifyHistsBinarySearchRemoveValAndRemoveVals(t *testing.T) {
-	data1 := generateData(t)
-	data2 := generateData(t)
-
-	require.Equal(t, data1, data2)
-	ctx := mock.NewContext()
-	sc := ctx.GetSessionVars().StmtCtx
-	b, err := codec.EncodeKey(sc.TimeZone(), nil, types.NewIntDatum(150))
-	require.NoError(t, err)
-	tmp := TopNMeta{
-		Encoded: b,
-		Count:   2,
-	}
-	data1.RemoveVals([]TopNMeta{tmp})
-	data2.BinarySearchRemoveVal(tmp)
-	require.Equal(t, data1, data2)
-}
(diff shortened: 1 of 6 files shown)


pkg/statistics/histogram_test.go
```diff
@@ -712,21 +712,3 @@ func generateData(t *testing.T) *Histogram {
...
-	require.Equal(t, data1, data2)
-	ctx := mock.NewContext()
-	sc := ctx.GetSessionVars().StmtCtx
-	b, err := codec.EncodeKey(sc.TimeZone(), nil, types.NewIntDatum(150))
-	require.NoError(t, err)
-	tmp := TopNMeta{
-		Encoded: b,
-		Count:   2,
-	}
-	data1.RemoveVals([]TopNMeta{tmp})
```
winoros, line 729: The `RemoveVals` is actually the wrong one. 
But, now that it is only used in stats v1.
I decide not to change it.

hawkingrei (approved):

AilinKid (approved):

pkg/statistics/histogram_test.go
AilinKid: decide to remove it?

PR #58980, opened 2025-01-16
statistics: do not ignore DDL event handing errors

Rustin170506:
### What problem does this PR solve?

Issue Number: close https://github.com/pingcap/tidb/issues/58545

Problem Summary:

### What changed and how does it work?

I checked all the ddl event handlers, the only one that can meet duplicate key error is insert the stats meta. So in this PR I used `insert ignore into` to avoid the error.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #59157, opened 2025-01-23
ddl notifier: use pessimistic txn and fix updating memory state too early

### LINKED ISSUE #59055: DDL notifier does not guarantee delivery when internal SQL failed to COMMIT
## Bug Report

https://github.com/pingcap/tidb/blob/c34a6b69f66ed080bfd4938ae51e134fc70b917d/pkg/ddl/notifier/subscribe.go#L275-L296

If `handler` runs without error, DDL notifier marks the in-memory structure `change` as processed in line 295. However the `handler` may still fail in line 277 when commits. The `change` uses the same SQL transaction so the wrong processed state is not persisted, but the memory state of `change` is used afterward and causes problems

https://github.com/pingcap/tidb/blob/c34a6b69f66ed080bfd4938ae51e134fc70b917d/pkg/ddl/notifier/subscribe.go#L233-L244

Will add UT in fix PR

lance6716:

Issue Number: close #59055

Problem Summary:

in pessimistic txn, errors can be found earlier, such as when execute the SQL. So handler can see the error and process it. Otherwise the error is reported when COMMIT, and in current handler interface it can't be proccessed

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/ddl/notifier/subscribe.go
```diff
@@ -268,15 +268,20 @@ func (n *DDLNotifier) processEventForHandler(
 	if (change.processedByFlag & (1 << handlerID)) != 0 {
 		return nil
 	}
+	newFlag := change.processedByFlag | (1 << handlerID)
 
-	if err = session.Begin(ctx); err != nil {
+	if err = session.BeginPessimistic(ctx); err != nil {
```
D3Hunter, line 273: what's the target issue of using pessimistic txn ?

lance6716: in pessimistic txn, errors can be found earlier, such as when execute the SQL. So handler can see the error and process it. Otherwise the error is reported when COMMIT, and in current handler interface it can't be proccessed

D3Hunter: > in pessimistic txn, errors can be found earlier

if there is no check about `affected rows`, there is no error actually in pessimistic txn

lance6716: The original problem is handler uses the txn to insert some duplicate rows. In optimistic txn, there's no error after execute the INSERT. Using pessimistic txn handler can see "duplicate entry" error. https://github.com/pingcap/tidb/pull/58980 is related

D3Hunter (approved):

pkg/ddl/notifier/subscribe.go
lance6716: I also checked the affected rows in https://github.com/pingcap/tidb/pull/59157/commits/bef13bf134d125a8db59f82a0f91e629052d7eda . It's revealed by old UT after switching to pessimistic txn.

pkg/ddl/notifier/store.go
```diff
@@ -87,17 +88,28 @@ func (t *tableStore) UpdateProcessed(
 	se *sess.Session,
 	ddlJobID int64,
 	multiSchemaChangeID int64,
-	processedBy uint64,
+	oldProcessedBy uint64,
+	newProcessedBy uint64,
 ) error {
 	sql := fmt.Sprintf(`
 		UPDATE %s.%s
```
Rustin170506, line 95: I’m not sure under the repeatable read isolation (snapshot isolation) whether TiDB guarantees that we can get the current processed_by_flag instead of the snapshot value.

Rustin170506 (approved):
I will help test this patch along with [my changes](https://github.com/pingcap/tidb/pull/58980) in our test infrastructure.

So let’s hold off on this PR until I finish the testing.

/hold

pkg/ddl/notifier/store.go
lance6716: > get the current processed_by_flag 

I'm not clear about your meaning. `oldProcessedBy` is read from another SQL session (`sess4List`), and here the processing SQL session uses WHERE processed_by_flag = `oldProcessedBy` and affected rows to check if the value has been changed

Rustin170506: Oh, I see, they are in a different session.

Rustin170506: Thank you!

Rustin170506:
Tested in the test infra:

We don't meet the duplicate key error anymore:
image

Also no records in the system table:

```sql
mysql> select * from mysql.tidb_ddl_notifier;
Empty set (0.05 sec)

```

Rustin170506:
/unhold 

Tested in https://github.com/pingcap/tidb/pull/59157#issuecomment-2614262450

Rustin170506:
/unhold

The reasons behind these commits are:
1. [`c7a8985` (#59157)](https://github.com/pingcap/tidb/pull/59157/commits/c7a8985e19aead514bfc9f0e8d508112491dd46b) If we handle the DDL events correctly, the analyze rate will be determined by the stats' row count instead of the default value. However, since DDL handling is asynchronous, it may sometimes succeed before the analysis and sometimes after.
2. [`e2cd37f` (#59157)](https://github.com/pingcap/tidb/pull/59157/commits/e2cd37f60c2845bc289beadb58218dfe52cfd256) If we don't handle the DDL event, the analyze process will not analyze the `val` column because it is neither a predicate column nor an index column, causing the status to be missing(nil). However, if we do handle the DDL event, we will know that this column hasn't been analyzed yet, meaning it is uninitialized. There is still a race condition between the analyze process and DDL event handling.

Issue #29026, opened 2021-10-21
De-duplicate error code definitions


dveeden:
## Problem description

Error codes are now defined in both `errno` and `parser/mysql`. Sometimes with conflicting definitions.
```
$ git grep ErrAccessDeniedNoPassword | head -4
errno/errcode.go:	ErrAccessDeniedNoPassword                                = 1698
errno/errname.go:	ErrAccessDeniedNoPassword:                                mysql.Message("Access denied for user '%-.48s'@'%-.255s'", nil),
parser/mysql/errcode.go:	ErrAccessDeniedNoPassword                                = 1698
parser/mysql/errname.go:	ErrAccessDeniedNoPassword:                                Message("Access denied for user '%-.48s'@'%-.64s'", nil),
```

From `parser/mysql/errcode.go` ( b01eedeba1 ):
```
b01eedeba1 (tiancaiamao 2020-03-11 16:04:04 +0800 970)  // Stop adding error code here!
b01eedeba1 (tiancaiamao 2020-03-11 16:04:04 +0800 971)  // They are moved to github.com/pingcap/tidb/errno
```

```
$ git grep 'errno.Err' | grep -v '_test.go' | wc -l
61
$ git grep 'mysql.Err' | grep -v '_test.go' | wc -l
534
```

PR #59237, opened 2025-02-01
errno: Add missing error codes to pkg/errno from pkg/parser/mysql.

wddevries:

Issue Number: ref #29026

Problem Summary:

There are two copies of most error codes and error messages.
 
This pull request:
* adds missing error codes from pkg/parser/mysql/errcode.go to pkg/errno/errcode.go
* adds missing error messages from pkg/parser/mysql/errname.go to pkg/errno/errname.go

This pull request is the first step in removing the mysql error codes and error messages from the parser.
 
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


lance6716:
Hi, after checking the linked issue I see most of the missing error codes are features that TiDB does not implemented. And the package `pkg/parser/mysql` can be served as library to other MySQL toolkit but `pkg/errno` is only a TiDB package. So we don't bother to add the missing error codes? Can you explain your plan a bit more?

bb7133 (approved):
LGTM

mjonss (commented):
When removing the error codes from the parser, do you also plan to update the [quickstart](https://github.com/pingcap/tidb/blob/master/pkg/parser/docs/quickstart.md) document in case it needs new dependencies?

dveeden:
Note that `pkg/errno` doesn't have a `state.go`, so it can't be used in cases where the sqlstate is needed.

In MySQL "slave" was renamed to "replica" and other terms like "master", "blacklist", etc were also renamed. On the MySQL side this is mostly done for places that are user visible, but is still ongoing in other places. Not sure if we need to change the names of the errors, but we might have to change the message to match MySQL where applicable. And for new errors we should try to stick with the new terms.

Another thing we probably shouldn't do, but _could_ consider: Taking one set of error definitions and generating (`go generate`) the other set or sets based on it.

PR #59342, opened 2025-02-08
s3: add RecreateWhenSlow for S3 GetObject API

### LINKED ISSUE #59341: overcome the S3 GetObject slow connections
## Enhancement

This is found in an EC2 test. When we opened ~1000 S3 GetObject connections but is slow to read from them, the connection will be stuck for minutes some times. We want to detect the stuck and recreate the connection as a workaround

lance6716:

Issue Number: close #59341

Problem Summary:

We found a problem that in AWS Golang SDK when we don't read the HTTP connection of GetObject for a long time, the connection may be stuck. After inspecting packets from `tcpdump` and consulting AWS supports, we choose to reopen the HTTP connection for the slow connection as a temporary workaround. A better and future solution could be let multiple HTTP requests using the 1 TCP connection but it may need to manually enable http2 or choose AWS CRT client which is not availble in Golang.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [x] Manual test (testing in the use case that revealed the problem) 
- [ ] No need to test
  > 

image

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


br/pkg/storage/s3.go
```diff
@@ -954,29 +969,82 @@ func (r *s3ObjectReader) Read(p []byte) (n int, err error) {
...
+				zap.Duration("expectedDuration", expectedDuration))
+			// wait above goroutine exit
+			r.readerCancel()
+			select {
+			case <-r.parentCtx.Done():
+				return 0, r.parentCtx.Err()
+			case <-done:
+			}
+
+			if err = r.recreateConn(); err != nil {
```
D3Hunter, line 1016: it might keep recreating if new connection is also slow

lance6716: It's expected. If new connection is slow, we still can't make progress, so still recreating to see if the network is getting better. Also I will double `expectedDuration` to avoid the duration is set too small.

lance6716:
I have tested that without this fix, using Amazon Linux as OS (mimic EKS behaviour) the merge step will still stuck. This stuck is caused by S3 server deliberately choose not to send data after a TCP connection has send too many zero window packet (as the AWS support said, I haven't investigated the root cause. I only verified this workaround is OK)

PR #59348, opened 2025-02-08
ttl: fix the issue that the TTL jobs are skipped or handled multiple times in one iteration

### LINKED ISSUE #59347: TTL job iteration will iterate the last item for multiple times
The code of TTL includes this pattern:

```go
for _, job := range m.runningJobs {
    ...
    m.removeJob(job)
    ...
}
```

It will have two bad effects:

1. Some jobs will be skipped during one iteration. If the 4th job is removed, the 5th job will not be iterated.
2. The last job will be iterated for multiple times.

The best practice (IMO) to solve this issue is to reverse interating:

```go
for i := len(m.runningJobs); i >= 0; i-- {
    ...
    m.removeJob(job)
    ...
}
```

YangKeao:

Issue Number: close #59347 
Problem Summary:

Some of the TTL jobs are skipped for handled multiple times in one iteration. It caused the following effect:

1. It'll take more time for `CheckNotOwnedJob`, `checkFinishedJob`, `rescheduleJobs` to handle all existing jobs.
2. The last job may be handled for multiple times, like finished multiple times.

I think it's a minor issue as both of them don't block the TTL.

Iterate the jobs in the reverse order.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ttl/client/notification.go
+++ b/pkg/ttl/client/notification.go
@@ -75,7 +75,7 @@ loop:
 			return ctx.Err()
 		case ch <- clientv3.WatchResponse{}:
 		default:
-			unsent = make([]chan clientv3.WatchResponse, len(watchers), 0)
+			unsent = make([]chan clientv3.WatchResponse, len(watchers))
 			copy(unsent, watchers[i:])
 			break loop
 		}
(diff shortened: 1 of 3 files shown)


YangKeao:
> I think it's a minor issue as both of them don't block the TTL.

NO. It'll also flush all statistic data of a TTL job if a job is finished multiple times 🤦 .

lcwangchao (approved):

Defined2014 (approved):

PR #59354, opened 2025-02-10
ast: Allow SELECT statements to set user-defined variables in read-only mode.

### LINKED ISSUE #55281: Rewriting user-defined variable in select stmt isn't treated as read only
## Bug Report

```
mysql> set global tidb_super_read_only=1;

mysql> select @i:=@i+1;
ERROR 1836 (HY000): Running in read-only mode

mysql> set @i:=@i+1;
```

The behavior is allowed like mysql:
```
mysql> select @@super_read_only;
+-------------------+
| @@super_read_only |
+-------------------+
|                 1 |
+-------------------+
1 row in set (0.00 sec)

mysql> select @i:=@i+1;
+----------+
| @i:=@i+1 |
+----------+
|     NULL |
+----------+
1 row in set, 1 warning (0.00 sec)

mysql> show warnings;
+---------+------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Level   | Code | Message                                                                                                                                                                                              |
+---------+------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Warning | 1287 | Setting user variables within expressions is deprecated and will be removed in a future release. Consider alternatives: 'SET variable=expression, ...', or 'SELECT expression(s) INTO variables(s)'. |
+---------+------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> select @i+1 into @i;

mysql> set @i:=@i+1;
```

ERROR 1836 (HY000): Running in read-only mode

v6.5.10

wddevries:

Issue Number: close #55281

Problem Summary:
In read-only mode user-defined variables cannot be set with select statements.  You can reproduce this by running:

`set global tidb_super_read_only=1; select @i:=@i+1;`

This is allowed on MySQL.

After reviewing the various code paths that can reach ast.IsReadOnly, I could find no reason this function would need to return false for read-only queries that set user-defined variables. This patch removes the part of ast.IsReadOnly that checks for writes to user-defined variables, which fixes the bug.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/parser/ast/util.go
+++ b/pkg/parser/ast/util.go
@@ -32,12 +32,7 @@ func IsReadOnly(node Node) bool {
 			}
 		}
 
-		checker := readOnlyChecker{
-			readOnly: true,
-		}
-
-		node.Accept(&checker)
-		return checker.readOnly
+		return true
 	case *ExplainStmt:
 		return !st.Analyze || IsReadOnly(st.Stmt)
 	case *DoStmt, *ShowStmt:
@@ -68,29 +63,4 @@ func IsReadOnly(node Node) bool {
 	default:
 		return false
 	}
-}
-
-// readOnlyChecker checks whether a query's ast is readonly, if it satisfied
-// 1. selectstmt;
-// 2. need not to set var;
-// it is readonly statement.
-type readOnlyChecker struct {
-	readOnly bool
-}
-
-// Enter implements Visitor interface.
-func (checker *readOnlyChecker) Enter(in Node) (out Node, skipChildren bool) {
-	if node, ok := in.(*VariableExpr); ok {
-		// like func rewriteVariable(), this stands for SetVar.
-		if !node.IsSystem && node.Value != nil {
-			checker.readOnly = false
-			return in, true
-		}
-	}
-	return in, false
-}
-
-// Leave implements Visitor interface.
-func (checker *readOnlyChecker) Leave(in Node) (out Node, ok bool) {
-	return in, checker.readOnly
 }
--- a/pkg/parser/ast/util.go
+++ b/pkg/parser/ast/util.go
@@ -20,8 +20,7 @@ const (
 	UnspecifiedSize = math.MaxUint64
 )
 
-// IsReadOnly checks whether the input ast is readOnly.
-func IsReadOnly(node Node) bool {
+func IsReadOnly(node Node, checkVariables bool) bool {
 	switch st := node.(type) {
 	case *SelectStmt:
 		if st.LockInfo != nil {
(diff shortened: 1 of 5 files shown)


songrijie:
LGTM. There's no reason to make it incompatible with MySQL. 

The impact of the backward-compatibility is minor.  Such logic should not exist in running applications as it errors out.

bb7133 (approved):
LGTM

dveeden (approved):

pkg/parser/ast/util.go
```diff
@@ -69,28 +64,3 @@ func IsReadOnly(node Node) bool {
...
-// Enter implements Visitor interface.
-func (checker *readOnlyChecker) Enter(in Node) (out Node, skipChildren bool) {
-	if node, ok := in.(*VariableExpr); ok {
-		// like func rewriteVariable(), this stands for SetVar.
-		if !node.IsSystem && node.Value != nil {
-			checker.readOnly = false
-			return in, true
-		}
-	}
-	return in, false
```
tiancaiamao, line 90:
```
// Enter implements Visitor interface.
func (checker *readOnlyChecker) Enter(in Node) (out Node, skipChildren bool) {
	if node, ok := in.(*VariableExpr); ok {
		// like func rewriteVariable(), this stands for SetVar.
		// By default, in select statement, readOnly is set to true
		if node.Value != nil {
			// But in 'select var to val', readOnly should be false
			checker.readOnly = false
			if !node.IsSystem {
				// One exception is 'select @val to val', change user variable is readOnly
				checker.readOnly = true
				return in, false
			}
			return in, true
		}
	}
	return in, false
}
```

I think the original author's intention is correct, just that he used wrong logic of `&&` to combine two differect branch and introduce the bug.
I try to express it in the original intention, and this should work.


Your changes still work but less robust, it removes the checker totally, 
the correctness depends on the parser not adding some other `*VariableExpr` in the future,
and once the assumption break, bug may sneak in.

@wddevries

tiancaiamao:
By the way, I find set operation is allowed in `tidb_super_read_only=1`, is that expected?   @wddevries 

`set global tidb_super_read_only=1;  set @@tidb_enable_paging = 0`

pkg/parser/ast/util.go
wddevries: You may be correct.  For the record, your function simplifies to:
```
// Enter implements Visitor interface.
func (checker *readOnlyChecker) Enter(in Node) (out Node, skipChildren bool) {
    if node, ok := in.(*VariableExpr); ok {
        // like func rewriteVariable(), this stands for SetVar.
        if node.IsSystem && node.Value != nil {
            checker.readOnly = false
            return in, true
        }
    }
    return in, false
}
```
Instead of marking queries that update user-variables are non-read-only, we would mark queries that update global variables as non-read-only.  Since updates to global variables are written to the database, this may be reasonable.

Generally ast.IsReadOnly is used to:
1. determine if a user can run a query in read-only mode.
2. determine compile/planning and commit options.

I am not sure you can actually create queries have global variable updates in the ast, but maybe we should still handle it correctly.

I will update the code.

wddevries:
> By the way, I find set operation is allowed in `tidb_super_read_only=1`, is that expected? @wddevries
> 
> `set global tidb_super_read_only=1; set @@tidb_enable_paging = 0`

Yes.  The SET command work correctly already.

tiancaiamao (approved):

tiancaiamao:
PTAL @fixdb

pkg/planner/optimize.go
```diff
@@ -416,7 +421,7 @@ func allowInReadOnlyMode(sctx planctx.PlanContext, node ast.Node) (bool, error)
 	}
 
 	vars := sctx.GetSessionVars()
-	return IsReadOnly(node, vars), nil
+	return isReadOnlyInternal(node, vars, false), nil
```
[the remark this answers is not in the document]
wddevries: I added a comment.

PR #59359, opened 2025-02-10
br: use region label rule to remove schedulers in db/table restore

### LINKED ISSUE #59515: br: use use region label rule for fine grained scheduler control
## Feature Request

currently, restore execution pauses pd operators globally including region scheduling. It's ok for cluster scope restore, but not for database/table level restore.

For database/table level restore, it's better to only stop pd operators for regions in the restore scope.

RidRisR:

Issue Number: close #59515 

Problem Summary:

currently, restore execution pauses pd operators globally including region scheduling. It's ok for cluster scope restore, but not for database/table level restore.

For database/table level restore, it's better to only stop pd operators for regions in the restore scope.

Br will automatically detect the region range affected in a db/table restore, and only close the scheduler in the necessary regions.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

截屏2025-02-13 18 26 35

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Br will automatically detect the region range affected in a db/table restore, and only close the scheduler in the necessary regions.
```


br/pkg/pdutil/pd.go
```diff
@@ -286,6 +289,7 @@ func (p *PdController) doPauseSchedulers(
 	return removedSchedulers, nil
 }
 
+// TODO: (ris) core func
```
BornChanger, line 292: what does it mean?

RidRisR: I will clean them.

br/pkg/restore/import_mode_switcher.go
```diff
@@ -200,6 +204,88 @@ func RestorePreWork(
...
+	}
+	return keyRanges
+}
+
+func calSortedTableIds(ids []int64) [][]int64 {
+	if len(ids) == 0 {
+		return [][]int64{}
+	}
+
+	sort.Slice(ids, func(i, j int) bool {
```
YuJuncen, line 266: sort.Slice can be modernized using `slices.Sort` (modernize)

br/pkg/restore/import_mode_switcher.go
```diff
@@ -200,6 +204,88 @@ func RestorePreWork(
...
+	return undo, &originCfg, errors.Trace(err)
+}
+
+func calSortedKeyRanges(ids []int64) [][]kv.Key {
+	idRanges := calSortedTableIds(ids)
+	if len(idRanges) == 0 {
+		return [][]kv.Key{}
+	}
+	var keyRanges [][]kv.Key
+	for i := 0; i < len(idRanges); i++ {
```
YuJuncen, line 251:
```suggestion
	for i := range idRanges {
```

br/pkg/restore/import_mode_switcher.go
```diff
@@ -200,6 +204,88 @@ func RestorePreWork(
...
+		return ids[i] < ids[j]
+	})
+
+	var idRanges [][]int64
+
+	start := ids[0]
+	end := start + 1
+
+	for i := 1; i < len(ids); i++ {
+		if ids[i] == ids[i-1]+1 {
```
YuJuncen, line 276: Generally table ID won't be continuous as there are DDLs occuiping table ID. Not sure whether merging this is really useful...

tests/realtikvtest/brietest/operator_test.go
```diff
@@ -63,26 +63,6 @@ type GcSafePoints struct {
 	} `json:"service_gc_safe_points"`
 }
 
-func verifyGCStopped(t *require.Assertions, cfg operator.PauseGcConfig) {
```
YuJuncen, line 66: Why remove this?

YuJuncen (commented):
rest lgtm

br/tests/br_region_rule/run.sh
```diff
@@ -0,0 +1,177 @@
...
+    jq 'any(.[]; .labels[]? | (.key=="schedule" and .value=="deny"))'
+  )
+  [ "$exists" = "true" ]
+}
+
+#######################################
+# Exits with 0 if 'schedule=deny' rule is found,
+# otherwise returns 1 (for use in if-statements).
+#######################################
+ensure_region_label_rule_exists() {
```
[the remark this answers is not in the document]
RidRisR: It is just a wrapper, you could see a `ensure_region_label_rule_absent` below

tests/realtikvtest/brietest/operator_test.go
RidRisR: Oh, this is a follow up of https://github.com/pingcap/tidb/pull/59605/files
I have to delete these unused functions to make ci happy.

YuJuncen (approved):

br/tests/br_region_rule/run.sh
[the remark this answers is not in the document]
RidRisR: Yes, you are right.

br/pkg/task/restore.go
```diff
@@ -1237,6 +1187,99 @@ func runSnapshotRestore(c context.Context, mgr *conn.Mgr, g glue.Glue, cmdName s
...
+		if cfg.piTRTaskInfo != nil {
+			logRestoredTS = cfg.piTRTaskInfo.RestoreTS
+		}
+		sets, restoreSchedulersConfigFromCheckpoint, err := client.InitCheckpoint(
+			ctx, cfg.snapshotCheckpointMetaManager, schedulersConfig, logRestoredTS, checkpointFirstRun)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		if restoreSchedulersConfigFromCheckpoint != nil {
+			restoreSchedulersFunc = mgr.MakeUndoFunctionByConfig(*restoreSchedulersConfigFromCheckpoint)
```
Leavrth, line 1241: the ruleID should be saved in the checkpoint too, but schedulersConfig now is `originCfg` in `FineGrainedRestorePreWork` without  field `ruleID`.

Leavrth (commented):
rest LGTM

br/pkg/task/restore.go
```diff
@@ -1237,6 +1187,99 @@ func runSnapshotRestore(c context.Context, mgr *conn.Mgr, g glue.Glue, cmdName s
 		}
 	}
 
+	importModeSwitcher := restore.NewImportModeSwitcher(mgr.GetPDClient(), cfg.Config.SwitchModeInterval, mgr.GetTLSConfig())
+	var restoreSchedulersFunc pdutil.UndoFunc
+	var schedulersConfig *pdutil.ClusterConfig
+	if isFullRestore(cmdName) || client.IsIncremental() {
```
Tristan1900, line 1193: also need to consider no explicit filter from user, cfg.ExplicitFilter

br/pkg/task/restore.go
```diff
@@ -1237,6 +1187,99 @@ func runSnapshotRestore(c context.Context, mgr *conn.Mgr, g glue.Glue, cmdName s
...
+		func() {
+			if !ok {
+				return
+			}
+			f, osErr := os.OpenFile(fileName, os.O_CREATE|os.O_WRONLY, os.ModePerm)
+			if osErr != nil {
+				log.Warn("failed to create file", zap.Error(osErr))
+				return
+			}
+			msg := []byte("schedulers removed\n")
```
Tristan1900, line 1264: I actually don't get the logic of this fault injector. looks like it's just printing out some message to a file, it's not checking scheduler status, how do we know at this point the scheduler is removed? maybe I'm missing something

Tristan1900 (changes requested):
thanks for the change!

br/pkg/task/restore.go
RidRisR: What exactly we need consider, I think `createdTables` are the table passed   filter already

br/pkg/task/restore.go
RidRisR: You are right

br/pkg/pdutil/pd.go
```diff
@@ -344,7 +347,33 @@ func (p *PdController) ResumeSchedulers(ctx context.Context, schedulers []string
 	return errors.Trace(p.resumeSchedulerWith(ctx, schedulers))
 }
 
+func (p *PdController) ResumeRegionLabelRule(ctx context.Context, ruleID string) error {
+	if ruleID == "" {
+		return nil
+	}
+	ruleRet, err := p.pdHTTPCli.GetRegionLabelRulesByIDs(ctx, []string{ruleID})
+	if err != nil {
```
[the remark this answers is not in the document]
RidRisR: ok

br/tests/br_db_skip/run.sh
```diff
@@ -56,7 +56,7 @@ run_sql "CREATE TABLE $DB.usertable1 ( \
 
 echo "restore start must succeed"
 fail=false
-run_br restore db --db $DB -s "local://$TEST_DIR/$DB" --pd $PD_ADDR --no-schema=true || fail=true
+run_br restore db --db $DB -s "local://$TEST_DIR/$DB" --pd $PD_ADDR --no-schema=true --check-requirements=false || fail=true
```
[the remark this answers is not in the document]
RidRisR: Not a part of this pr actually. It is the same as https://github.com/pingcap/tidb/pull/59695/files.

br/pkg/pdutil/pd.go
```diff
@@ -590,6 +632,63 @@ func (p *PdController) RemoveSchedulersWithConfigGenerator(
 	return originCfg, removedCfg, nil
 }
 
+func (p *PdController) RemoveSchedulersConfig(
+	ctx context.Context,
+) (origin ClusterConfig, modified ClusterConfig, err error) {
```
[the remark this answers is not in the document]
RidRisR: Yes, I should delete it.

br/pkg/pdutil/pd.go
```diff
@@ -703,16 +809,13 @@ func pauseSchedulerByKeyRangeWithTTL(
 		RuleType: "key-range",
 		// Data should be a list of KeyRangeRule when rule type is key-range.
 		// See https://github.com/tikv/pd/blob/783d060861cef37c38cbdcab9777fe95c17907fe/server/schedule/labeler/rules.go#L169.
-		Data: []KeyRangeRule{{
-			StartKeyHex: hex.EncodeToString(startKey),
-			EndKeyHex:   hex.EncodeToString(endKey),
-		}},
+		Data: encodedKeyRangeRule,
```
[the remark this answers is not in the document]
RidRisR: It is ok, I could add a test about it

br/pkg/task/restore.go
```diff
@@ -1237,6 +1187,99 @@ func runSnapshotRestore(c context.Context, mgr *conn.Mgr, g glue.Glue, cmdName s
...
+
+		defer func() {
+			// need to flush the whole checkpoint data so that br can quickly jump to
+			// the log kv restore step when the next retry.
+			log.Info("wait for flush checkpoint...")
+			client.WaitForFinishCheckpoint(ctx, len(cfg.FullBackupStorage) > 0 || !canRestoreSchedulers)
+		}()
+	}
+
+	failpoint.Inject("sleep_for_check_scheduler_status", func(val failpoint.Value) {
```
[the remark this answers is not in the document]
RidRisR: I prefer not to bring the fault injection into the routine code structure.
This function would not be used elsewhere after all.

br/pkg/pdutil/pd.go
```diff
@@ -690,9 +789,16 @@ func PauseSchedulersByKeyRange(
 func pauseSchedulerByKeyRangeWithTTL(
 	ctx context.Context,
 	pdHTTPCli pdhttp.Client,
-	startKey, endKey []byte,
+	keyRange [][2]kv.Key,
 	ttl time.Duration,
-) (<-chan struct{}, error) {
+) (<-chan struct{}, string, error) {
+	var encodedKeyRangeRule = []KeyRangeRule{}
+	for _, keyPair := range keyRange {
```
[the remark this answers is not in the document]
RidRisR: Good question,
I may need to check the performance when the table count is large.

br/pkg/task/restore.go
Tristan1900: right the idea is we do fine-grained work if not full cluster restore, with filter it should be a fine-grained even the command is `FullRestoreCmd` since it could only restore a table or a db.

br/pkg/task/restore.go
Tristan1900: I see, it's doing some pause and check thing. Why don't we directly check here when fail point enabled? We do have access to pd in the code I think?

br/pkg/task/restore.go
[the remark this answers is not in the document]
Tristan1900: it's mainly for readability because when we read the code we don't want to pay attention to large chunk of implementation details in a main routine. That's my personal preference though

br/pkg/task/restore.go
RidRisR: We could, but we still need a method to pass the output to the env.
If we check the pause status in the failpoint, then how to interact with the integration test?

br/pkg/pdutil/pd.go
```diff
@@ -590,6 +632,63 @@ func (p *PdController) RemoveSchedulersWithConfigGenerator(
...
+
+// To resume the schedulers, call the cancel function.
+// wait until done is finished to ensure schedulers all resumed
+func (p *PdController) RemoveSchedulersOnRegion(ctx context.Context, keyRange [][2]kv.Key) (string, func(), error) {
+	done, ruleID, err := pauseSchedulerByKeyRangeWithTTL(ctx, p.pdHTTPCli, keyRange, pauseTimeout)
+	// Wait for the rule to take effect because the PD operator is processed asynchronously.
+	// To synchronize this, checking the operator status may not be enough. For details, see
+	// https://github.com/pingcap/tidb/issues/49477.
+	// Let's use two times default value of `patrol-region-interval` from PD configuration.
+	<-time.After(20 * time.Millisecond)
```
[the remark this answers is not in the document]
YuJuncen: (Also I think this can be simplified with `time.Sleep(20 * time.Millisecond)`...)

br/pkg/task/restore.go
RidRisR: We need this feature because we need to keep the cluster running while doing db/table restore.
For full restore, there won't be other transactions, we could safely pause all scheduler with no side effect.

br/pkg/task/restore_txn.go
```diff
@@ -96,6 +96,7 @@ func RunRestoreTxn(c context.Context, g glue.Glue, cmdName string, cfg *Config)
 	}
 
 	importModeSwitcher := restore.NewImportModeSwitcher(mgr.GetPDClient(), cfg.SwitchModeInterval, mgr.GetTLSConfig())
+	//TODO: handle scheduler here
```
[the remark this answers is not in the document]
RidRisR: removed

br/pkg/pdutil/pd.go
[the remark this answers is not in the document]
RidRisR: Now this keyRange will allow a max one ID length gap between two IDs to contain the ddl job ID. This will make the restore range merged together.

br/pkg/pdutil/pd.go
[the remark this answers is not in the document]
RidRisR: It is twice the `patrol-region-interval` from PD configuration as the comment describes. 
I've tested a 300 rule scenario, and the wait time is enough.

br/pkg/task/restore.go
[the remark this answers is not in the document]
RidRisR: I prefer to leave it be. One good method fold the func if that bother you 
:)

br/pkg/pdutil/pd.go
```diff
@@ -590,6 +632,63 @@ func (p *PdController) RemoveSchedulersWithConfigGenerator(
...
+
+	resumeScheduler := func() {
```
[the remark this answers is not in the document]
Tristan1900: @RidRisR from my side it's still unchanged, maybe you forget to push the commit?

br/pkg/task/restore.go
Tristan1900: even full restore can restore only part of db, doing `restore db foo` and `restore full --filter "foo.*"` is exact same operation.

br/pkg/task/restore.go
Tristan1900: I don't see any interaction in the integration test that needs the pause, maybe I'm missing something, looks like every time when you enable the fail-point it's checking if label rule exist and fail if needed, and that can be done in the fail point itself. Can you point me to the place that need the pause? thanks

br/pkg/restore/import_mode_switcher.go
```diff
@@ -200,6 +204,87 @@ func RestorePreWork(
...
+
+	slices.Sort(ids)
+
+	var idRanges [][]int64
+
+	start := ids[0]
+	end := start + 1
+
+	for i := 1; i < len(ids); i++ {
+		if ids[i] <= ids[i-1]+2 {
```
Tristan1900, line 275: I would say a safer/effecient approach would be allocate all the ids at once and assign those ids to tables before creation, that way we only need to set rule for that one range. 
But that involves more code change as right now we allocate only up to max upstream id, to achieve that we need to allocate more for those tables that cannot reuse table id. 
@Leavrth let me know what you think.

Tristan1900:
left a few more comments, thanks!

br/pkg/task/restore.go
RidRisR: Ok, Let me clarify the reasoning. 

The key idea is: by injecting a breakpoint, we could reuse code and get a better observability.

We need to validate two distinct phases:

1. During restore: Ensure the region label rule exists (which handled in the failpoint).
2. After restore: Confirm the rule expires (_which must be validated in the integration test_).

The breakpoint allows us to reuse the same validation logic for both phases. 

Also, The breakpoint provides a window to inspect rule status details during development. It also makes the test more flexible—for example, if we later want to test how user actions (like creating tables) interact with the restore process, we could inject new checks _without modifying core logic_.

br/pkg/pdutil/pd.go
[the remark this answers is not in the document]
RidRisR: My fault, now it should changed.

br/pkg/task/restore.go
Tristan1900: do you mean `NeedCheckFreshCluster`? it has a parameter `explcitiFilter` that if true it will skip the check

RidRisR: Technically they can, but I don't believe using the full restore as database/table restore by adding filter flags is good practice. We should not encourage users to adopt this approach either.

Tristan1900: but that's the meaning of having regex filter as a feature and that's what customers are doing right now, for example you can do `restore full --filter !mysql.*` which means restore everything but not mysql database, you cannot do that using `restore db` or `restore table` because it's impossible to list all the db in the cluster.

RidRisR: I fully aware this point. My concern is not about how to build the filter but if **"all ranges could be scheduled by br"**.
In my opinion, user should not do a full restore on an online cluster even if they only restore one row.

Tristan1900: ok we can discuss more offline cuz I'm still not clear of the reason

br/pkg/pdutil/pd.go
Tristan1900, line 355: we can merge here and below

br/pkg/pdutil/pd.go
```diff
@@ -344,7 +347,33 @@ func (p *PdController) ResumeSchedulers(ctx context.Context, schedulers []string
 	return errors.Trace(p.resumeSchedulerWith(ctx, schedulers))
 }
 
+func (p *PdController) ResumeRegionLabelRule(ctx context.Context, ruleID string) error {
```
Tristan1900, line 350: it seems we suppress all the errors in this method, the error will always be nil, is that expected? If so we can just remove the return type

br/pkg/task/restore.go
```diff
@@ -1242,6 +1199,100 @@ func runSnapshotRestore(c context.Context, mgr *conn.Mgr, g glue.Glue, cmdName s
...
+		var tableIDs []int64
+		for _, table := range createdTables {
+			tableIDs = append(tableIDs, table.Table.ID)
+			if table.Table.Partition != nil {
+				for _, p := range table.Table.Partition.Definitions {
+					tableIDs = append(tableIDs, p.ID)
+				}
+			}
+		}
+		keyRange := SortKeyRanges(tableIDs, preAllocRange)
```
Tristan1900, line 1222: I might misunderstand something, I remember we make one range allocation for all the tables, whether reuse id or not. So we can just tell PD to operate on this single large range instead a bunch of small ranges. But seems in code we are still creating small ranges to update PD. Is there a reason why we still need to do that?

br/pkg/pdutil/pd.go
RidRisR: You are right, there used to be errors, but now all return would be nil

br/pkg/pdutil/pd.go
RidRisR: ok

br/pkg/task/restore.go
RidRisR: Because the pre-alloc don't have the memery of the last restore.  It means that we might restore to a table that has been created in the last restore.
We may need to add the rule into checkpoint to fix this problem, but we can't make it. Now the restore will always apply the checkpoint as long as the previous restore is not finished, even when they have different range. Which means the rule in the checkpoint may not applicable in the latter restore.

br/pkg/task/restore.go
```diff
@@ -1242,6 +1199,100 @@ func runSnapshotRestore(c context.Context, mgr *conn.Mgr, g glue.Glue, cmdName s
...
+			logRestoredTS = cfg.piTRTaskInfo.RestoreTS
+		}
+		sets, restoreSchedulersConfigFromCheckpoint, err := client.InitCheckpoint(
+			ctx, cfg.snapshotCheckpointMetaManager, schedulersConfig, logRestoredTS, checkpointFirstRun)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		if restoreSchedulersConfigFromCheckpoint != nil {
+			// The last range rule will be dropped when the last restore quits.
+			restoreSchedulersConfigFromCheckpoint.RuleID = schedulersConfig.RuleID
```
Leavrth, line 1259: If `restoreSchedulersConfigFromCheckpoint.RuleID` is replaced by the `schedulersConfig.RuleID` from this execution, the `restoreSchedulersConfigFromCheckpoint.RuleID` from last execution won't be removed by BR.

An easy way is that skip pausing pd by ranges if `checkpointFirstRun` is false. (which indicates tables are already created so that IDs are fixed).

br/pkg/task/restore.go
Tristan1900: oh I see your concern. yeah we might need to store that in the checkpoint for enahcnemant, that will makes things much easier. 
if we could find out what's the min and max table id created from last restore, i.e. tables already in the cluster, for example, 10 and 20, we could have just two ranges sent to PD, 10->20 and newly alloc range. 
Right now I'm worried that having many small ranges will cause problem to the PD and beat our purpose of making one large range. happy to discuss more!

br/pkg/restore/snap_client/client.go
```diff
@@ -312,9 +317,25 @@ func (rc *SnapClient) AllocTableIDs(ctx context.Context, tables []*metautil.Tabl
...
+func (rc *SnapClient) GetPreAllocedTableIDRange() ([2]int64, error) {
+	if rc.preallocedIDs == nil {
+		return [2]int64{}, errors.Errorf("No preAlloced IDs")
+	}
+
+	start, end := rc.preallocedIDs.GetIDRange()
+
+	if start >= end {
+		log.Warn("PreAlloced IDs range is empty, no table to restore")
+		return [2]int64{start, end}, nil
```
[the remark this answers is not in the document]
RidRisR: Seems restore empty table is legal in restore. The restore should not be error.

br/pkg/task/restore.go
RidRisR: There wouldn't be "many small ranges" unless the restore range keep change and keep fail, I believe it should be very rare case.

br/pkg/task/restore.go
RidRisR: No, if the restore gracefully quit, the scheduler will always be clean, even if the restore failed.
The scheduler will be clean when background context canceled.

br/pkg/task/restore.go
Tristan1900: right, reading from the code, the way we are doing in `calSortedTableIds` is assuming tables ids are continuous, that might be true for newly allocated tables since we assign continuous ids to them, but most likely not true for tables that reuses their original table id. if thousands of tables are resuing their id and not continuous it could lead to many small ranges. 
an example if we have 5 tables to create and their original id are  1,3,5,7,9, global id = 4, based on previous PR algorithm we allocate from 5 to 11, and actual id used are 5, 7, 9, 10, 11, where 5,7,9 are reused and 1,3 needs rewrite to 10, 11. If during restore we failed after creating 5,7,9, next retry based on `calSortedTableIds` and `mergeIntervals ` we will need to create small range like `5-6`, `7-8` etc. We can expand that to large number of reused table ids and I feel like it might be a potential issue. 
please let me know if I'm understanding correctly!

RidRisR: oh, good point.
I may need add the table range into the checkpoint. But we can't make sure how many failed restore in the previous restore. If these occupied ids come from different restore, there would still be many small ranges.
Seems the only correct method to fix these kind of bugs is not "how to record the key range", but to "check if the checkpoint meta applicable before restore".
Once we could make sure the restore range wouldn't change between restores, we could always reuse the key range in last restore.

Leavrth: Yes, we can `reuse` the preallocated IDs (although upstream ID is different from downstream ID), if we record the preallocated IDs into the checkpoint before `(*snapclient.SnapClient).CreateTables`.
An easy way is directly saved as ID map, like `pitr_id_map`, which records mappings from upstream ID to downstream ID that is preallocated in the first execution.
An space-saving way is just records the preallocated start ID and end ID, and use a deterministic algorithm to preallocated IDs. For example, if we have upstream IDs that are 1,3,5,7,9 and global ID is 4, the upstream IDs can be preallocated to (5->5, 7->7, 9->9, 1->10, 3->11). The reused part is from 5 to 9, and the not reused part is from 10 to 11, to which the upstream IDs are mapped in order of size.

Leavrth: In this PR, perhaps for the reuse IDs, we can merge them in a range if upstream ID equals downstream ID.
For example, if the upstream IDs are 3, 100, 1002, 4325, 40000 and global ID is 1000. We preallocate IDs with range [1001, 40003), and map IDs that are (1002->1002, 4325->4325, 40000->40000, 3->40001, 100-> 40002). In the next execution, if upstream ID is the same as downstream ID that are 1002,  4325 and 40000, we can merge them into a range [1002, 40001).

Leavrth (approved):

Tristan1900 (approved):

PR #59382, opened 2025-02-11
lignthing/importinto: parallel reading files infos from data store

### LINKED ISSUE #56104: lightning might stuck for hours when importing parquet files from cloud storage
## Bug Report

after https://github.com/pingcap/tidb/pull/46984, including `>= v7.5.x`, `7.1.3+`, we will sample parquet all the time in serial, it's very slow and might takes hours if user have large mount of parquet files before lightning start doing import, and the time takes to sample the files might even longer than real import work.

we only need this size for displaying progress more accurately and use it as a reference when splitting engine, but slowing import this much is un-acceptable.

start import fast
it might takes hours before start doing any work

### LINKED ISSUE #60224: lightning might stuck for hours when importing parquet files from cloud storage (round 2)
This is exactly same issue as [#56104](https://github.com/pingcap/tidb/issues/56104)

#56104 is closed by [#56205](https://github.com/pingcap/tidb/pull/56205), but problem is not fixed.

This new issue is to track the new PR [#59382](https://github.com/pingcap/tidb/pull/59382)

joechenrh:

Issue Number: close #56104, close #60224

Problem Summary:

Constructing file infos in parallel for both lightning and `IMPORT INTO`.

- For lightning, the concurrency is `RegionConcurrency * 2`.
- For `IMPORT INTO`, the concurrency is the task's `ThreadCnt * 2`.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

The following results are tested by lightning with 8 core (so concurrency is 16). Data is stored on S3. `IMPORT INTO` should have the similar result.

- Importing **one table with 1000 * 10M parquet files**, the time is reduced from 6.8min to 16.7s
- Importing **one table with 1000 * 7M compressed CSV files**, the time is reduced from 5min to 10s.

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/importer/import.go
```diff
@@ -1106,28 +1106,38 @@ func (e *LoadDataController) InitDataFiles(ctx context.Context) error {
...
-				compressTp := mydump.ParseCompressionOnFileExtension(remotePath)
+				allFiles = append(allFiles, mydump.RawFile{Path: remotePath, Size: size})
+				totalSize += size
+				return nil
+			}); err != nil {
+			return exeerrors.ErrLoadDataCantRead.GenWithStackByArgs(GetMsgFromBRError(err), "failed to walk dir")
+		}
+
+		var err error
+		if dataFiles, err = mydump.ParallelProcess(ctx, allFiles, e.ThreadCnt,
```
D3Hunter, line 1127:
```suggestion
		if dataFiles, err = mydump.ParallelProcess(ctx, allFiles, e.ThreadCnt*2,
```
just like lightning does, we can double it

D3Hunter (commented):
7/9

pkg/lightning/mydump/loader.go
```diff
@@ -371,6 +380,51 @@ type ExtendColumnData struct {
...
+	workerCount = min(workerCount, len(batchCounts))
+
+	starts := make([]int, len(batchCounts))
+	ends := make([]int, len(batchCounts))
+	for i, batch := range batchCounts {
+		if i > 0 {
+			starts[i] = ends[i-1]
+		}
+		ends[i] = starts[i] + batch
+	}
```
kennytm, line 402: The `ends` array seems redundant, you could just store the `starts`

```go
starts := make([]int, len(batchCounts)+1)
for i, batch := range batchCounts {
    starts[i+1] = starts[i] + batch
}
```

and in the loop below use

```go
for j := starts[id]; j < starts[id+1]; j++ {
    ...
}
```

but do we actually *need* to do batching ourselves at all? You could simply [use `SetLimit` to limit the concurrency](https://pkg.go.dev/golang.org/x/sync@v0.11.0/errgroup#Group.SetLimit):

```go
res := make([]T, len(files))
eg, ctx := errgroup.WithContext(ctx)
eg.SetLimit(max(workerCount), 1))
for j, file := range files {
    eg.Go(func() error {
        v, err := hdl(ctx, file)
        if err != nil {
            return err
        }
        res[j] = v
    })
}
if err := eg.Wait(); err != nil {
    return nil, errors.Trace(err)
}
```

pkg/lightning/mydump/loader.go
```diff
@@ -397,12 +451,54 @@ func (s *mdLoaderSetup) setup(ctx context.Context) error {
...
+		return nil
+	}); err != nil {
 		if !s.setupCfg.ReturnPartialResultOnError {
 			return common.ErrStorageUnknown.Wrap(err).GenWithStack("list file failed")
 		}
 		gerr = err
 	}
+
+	// Parallel process all files
+	allInfos, err := ParallelProcess(ctx, allFiles, runtime.NumCPU()*2, s.constructFileInfo)
```
kennytm, line 468: it should be `region-concurrency` * 2 here rather than `runtime.NumCPU() * 2`? in particular in a container the NumCPU may reported may be larger than the actual quota. 

alternatively add yet another config

pkg/lightning/mydump/loader.go
joechenrh: I'm not sure if it will bring some overhead when we have millions of files. So I only created `batchCounts` goroutines.😵‍💫

pkg/lightning/mydump/loader.go
joechenrh: https://github.com/pingcap/tidb/pull/59382/files#diff-22fb9124c24e87e690dc01d994ccaf68812d7f5b05182b1ca14ace1d6d77bddaR541

pkg/lightning/mydump/loader.go
joechenrh: > but do we actually need to do batching ourselves at all? You could simply [use SetLimit to limit the concurrency](https://pkg.go.dev/golang.org/x/sync@v0.11.0/errgroup#Group.SetLimit):

I found that we have already done this in other function, like
https://github.com/pingcap/tidb/blob/c08679bccfad5ee21551dc513bcad43cf1107b22/pkg/lightning/mydump/region.go#L223-L226

Fixed in https://github.com/pingcap/tidb/pull/59382/commits/9666920474121ca38fb624dd7690a60f0179a0bb#diff-5cfbb6c376bc253150edace1783bda923e054c3ade35038f1ad5673b42b63936R405-R408

pkg/lightning/mydump/loader.go
```diff
@@ -371,6 +391,44 @@ type ExtendColumnData struct {
...
+
+	for i, file := range files {
+		eg.Go(func() error {
+			select {
+			case <-egCtx.Done():
+				return nil
+			default:
+			}
+
+			v, err := hdl(ctx, file)
```
kennytm, line 417:
```suggestion
			v, err := hdl(egCtx, file)
```

pkg/lightning/mydump/loader.go
```diff
@@ -371,6 +391,44 @@ type ExtendColumnData struct {
...
+	res := make([]T, len(files))
+
+	eg, egCtx := errgroup.WithContext(ctx)
+	eg.SetLimit(workerCount)
+
+	for i, file := range files {
+		eg.Go(func() error {
+			select {
+			case <-egCtx.Done():
+				return nil
```
kennytm, line 413: 🤔 why return `nil` rather than `context.Cause(egCtx)` (or `egCtx.Err()`)

kennytm (commented):
rest lgtm

pkg/lightning/mydump/loader.go
```diff
@@ -142,6 +144,10 @@ type MDLoaderSetupConfig struct {
 	// MaxScanFiles specifies the maximum number of files to scan.
 	// If the value is <= 0, it means the number of data source files will be scanned as many as possible.
 	MaxScanFiles int
+
+	// ScanFileConcurrency specifes the concurrency of scaning source files.
+	ScanFileConcurrency int
```
D3Hunter, line 149: maybe just move it to LoaderConfig, it's always needed

D3Hunter (commented):
6/11

pkg/lightning/mydump/loader.go
joechenrh: Since it's used for mdLoaderSetup, so I only add it in `MDLoaderSetupConfig` and modify it using `WithScanFileConcurrency`, just like `WithMaxScanFiles`.

pkg/lightning/mydump/loader.go
```diff
@@ -845,46 +959,3 @@ func SampleFileCompressRatio(ctx context.Context, fileMeta SourceFileMeta, store
 	}
 	return float64(tot) / float64(pos), nil
 }
-
-// SampleParquetRowSize samples row size of the parquet file.
-func SampleParquetRowSize(ctx context.Context, fileMeta SourceFileMeta, store storage.ExternalStorage) (float64, error) {
```
joechenrh, line 850: Moved to `pkg/lightning/mydump/parquet_parser.go`

pkg/executor/importer/import.go
```diff
@@ -1136,19 +1146,6 @@ func (e *LoadDataController) InitDataFiles(ctx context.Context) error {
 	return nil
 }
 
-func (e *LoadDataController) getFileRealSize(ctx context.Context,
```
joechenrh, line 1139: Duplicate function with `mydump.EstimateRealSizeForFile`

pkg/executor/importer/import_test.go
```diff
@@ -230,22 +228,6 @@ func TestASTArgsFromStmt(t *testing.T) {
 	require.Equal(t, astArgs.ColumnsAndUserVars, importIntoStmt.ColumnsAndUserVars)
 }
 
-func TestGetFileRealSize(t *testing.T) {
```
joechenrh, line 233: `getFileRealSize` is removed

pkg/lightning/mydump/loader.go
```diff
@@ -496,79 +596,74 @@ func (iter *allFileIterator) IterateFiles(ctx context.Context, hdl FileHandler)
...
+
+		// Only sample once for each table
+		_, loaded := s.sampledParquetRowSizes.LoadOrStore(tableName, 0)
+		var (
+			totalRowCount int64
+			rowSize       float64
+		)
+
+		if !loaded {
+			totalRowCount, rowSize, err = ReadParquetRowCountAndSizeByFile(ctx, s.loader.GetStore(), info.FileMeta)
```
joechenrh, line 635: https://github.com/pingcap/tidb/blob/046e83e7fc1419d67ca117b48e0c00fefcddc27f/pkg/lightning/mydump/loader.go#L542-L552
Previously we will read the same file twice for the first file of each table, to sample row size and read row count, here I merge them into one function `ReadParquetRowCountAndSizeByFile`.

D3Hunter (approved):
rest lgtm

GMHDBJD (approved):

Issue #43885, opened 2023-05-16
optimize the long nested recursive expression performance 


tiancaiamao:
## Enhancement

There are a lot of  recursive function used in our code base. Especially the expression related.
Depending on the user's sql, it may sometimes cause performance issues.

For example, this one https://github.com/pingcap/tidb/issues/42788

Here I upload a flamegraph for this kind of sql:

```
select * from t where 1=0  or (t.a = '0' and t.c = 0) or (t.a = '1' and t.c = 1) or (t.a = '2' and t.c = 2) or (t.a = '3' and t.c = 3)
 or (t.a = '4' and t.c = 4) or (t.a = '5' and t.c = 5) or (t.a = '6' and t.c = 6) or (t.a = '7' and t.c = 7) or (t.a = '8' and t.c = 8
) or (t.a = '9' and t.c = 9) or (t.a = '10' and t.c = 10) or (t.a = '11' and t.c = 11) or (t.a = '12' and t.c = 12) or (t.a = '13' and
 t.c = 13) or (t.a = '14' and t.c = 14) or (t.a = '15' and t.c = 15) or (t.a = '16' and t.c = 16) or (t.a = '17' and t.c = 17) or (t.a
 = '18' and t.c = 18) or (t.a = '19' and t.c = 19) or (t.a = '20' and t.c = 20) or (t.a = '21' and t.c = 21) or (t.a = '22' and t.c =
22) or (t.a = '23' and t.c = 23) or (t.a = '24' and t.c = 24) or (t.a = '25' and t.c = 25) or (t.a = '26' and t.c = 26) or (t.a = '27'
 and t.c = 27) or (t.a = '28' and t.c = 28) or (t.a = '29' and t.c = 29) or (t.a = '30' and t.c = 0) or (t.a = '31' and t.c = 1) or (t
.a = '32' and t.c = 2) or (t.a = '33' and t.c = 3) or (t.a = '34' and t.c = 4) or (t.a = '35' and t.c = 5) or (t.a = '36' and t.c = 6)
 or (t.a = '37' and t.c = 7) or (t.a = '38' and t.c = 8) or (t.a = '39' and t.c = 9) or (t.a = '40' and t.c = 10) or (t.a = '41' and t
.c = 11) or (t.a = '42' and t.c = 12) or (t.a = '43' and t.c = 13) or (t.a = '44' and t.c = 14) or (t.a = '45' and t.c = 15) or (t.a =
[...]

PR #59487, opened 2025-02-12
privilege, domain: reduce the memory jitter of privilege reload activity for 2M users

### LINKED ISSUE #59403: Memory leak of LoadPrivilegeLoop?
## Bug Report

Like https://github.com/pingcap/tidb/issues/59400

> I create 2M users, and the keep 10% of them active, then run
> 
> "set password for test = 'xxx'" for all the 2M users
> 
> then run
> 
> "alter user test%d failed_login_attempts 10" for all the 2M users
> 
> ...

This time analyze the memory usage after the workload is gone.

I suspect there is a leak, after the workload gone, the memory usage is still there:

Image

As you can see the CPU is 0%, but at time 15:45, the memory usage is high.
And after a privilege full reload (the spike), the memory usage finally decreased.

Image

Here is the mem profile before and after the full reload:

[after.pb.gz](https://github.com/user-attachments/files/18746934/after.pb.gz)
[before.pb.gz](https://github.com/user-attachments/files/18746935/before.pb.gz)

The memory usage should reduce after the workload is gone like after.pb.gz

Image

See before.pb.gz
Image

master?
```
commit 27365b47b27820152faf79395f6cf1b96eb3031f (HEAD -> master, origin/master, origin/HEAD)
Author: xzhangxian1008 <xzhangxian@foxmail.com>
Date:   Tue Feb 11 15:35:39 2025 +0800

    executor: fix the incorrect return when hash join encounters error (#59381)
    
    close pingcap/tidb#59377
```

tiancaiamao:

Issue Number: close #59403, ref #55563

Problem Summary:

I create 2M users, and for example, make 10% or 50% of the users active (in-memory).

Then I observe that even when the workload is gone, the tidb-server memory usage jitter periodically.
For example, this one:

image

There are several changes.

1.  Before this PR, `loadAll()` is used when the active user count > 1024 ... that's the direct root cause of the jitter.

That's because `loadSomeUsers()` does not support tooooo many filter condition.
The SQL "select * from user where user = 'a' or user = 'b' or user = 'c'  or ..." works poorly when there are too many or conditions. This is a known issue https://github.com/pingcap/tidb/issues/43885 that we write the code the recursive way and cause stackoverflow.

So the first change is to enhance `loadSomeUsers()` to support unlimited user count.

It works like this:
- if user count > 1024, use the 'or user = xx' filter condition to construct the SQL
- otherwise, use load all SQL but do the 'user = xx' filter condition in the user space.

2.  Use this `SQLExecutor.ExecuteInternal()` streaming API to replace `RestrictedSQLExec.ExecRestrictedSQL()` API

The problem of `ExecRestrictedSQL` is that the API design not fit here.
Its `drainRecordSet` return  `[]chunk.Row` as result and here it can be 2M huge array.
What we need is a streaming API, doing the filter condition at the same time rather than take the whole data set and filter out later.

3.  Use deep copy in the loadTable decode function.

I suspect there is a leak like #59403,  the decode function may using a shallow copy and it references the chunk data.
So the whole chunk cannot be freed.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

The memory usage now:

image

The privilege reload activity is every 10min and you can see that the max memory usage is much less than before:

image

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


tiancaiamao:
I'll comment the changes to make review easier.

pkg/privilege/privileges/cache.go
```diff
@@ -521,147 +527,143 @@ func findUserAndAllRoles(all map[string]struct{}, roleGraph map[string]roleGraph
 			break
 		}
 	}
+	return all
 }
 
-func (p *MySQLPrivilege) loadSomeUsers(ctx sqlexec.RestrictedSQLExecutor, userList ...string) ([]string, error) {
-	if len(userList) > 512 {
-		logutil.BgLogger().Warn("loadSomeUsers called with a long user list", zap.Int("len", len(userList)))
```
tiancaiamao, line 528: This limitation is eliminated, by load all data and do the filter condition in the user space.

pkg/privilege/privileges/cache.go
```diff
@@ -521,147 +527,143 @@ func findUserAndAllRoles(all map[string]struct{}, roleGraph map[string]roleGraph
...
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
 
-	// Including the user list and also their roles
-	extendedUserList := make(map[string]struct{}, len(userList))
-	for _, user := range userList {
-		extendedUserList[user] = struct{}{}
-	}
-	findUserAndAllRoles(extendedUserList, p.roleGraph)
```
tiancaiamao, line 542: It's just moving code, no change here.

pkg/privilege/privileges/cache.go
```diff
@@ -521,147 +527,143 @@ func findUserAndAllRoles(all map[string]struct{}, roleGraph map[string]roleGraph
...
 	return userList, nil
 }
 
-// merge construct a new MySQLPrivilege by merging the data of the two objects;.
-func (p *MySQLPrivilege) merge(diff *MySQLPrivilege, userList []string) *MySQLPrivilege {
+// merge construct a new MySQLPrivilege by merging the data of the two objects.
+func (p *MySQLPrivilege) merge(diff *MySQLPrivilege, userList map[string]struct{}) *MySQLPrivilege {
 	ret := newMySQLPrivilege()
 	user := p.user.Clone()
-	for _, u := range userList {
```
tiancaiamao, line 591: The old way: merge = old - userList (changes) + diff (data of userList)

pkg/privilege/privileges/cache.go
```diff
@@ -521,147 +527,143 @@ func findUserAndAllRoles(all map[string]struct{}, roleGraph map[string]roleGraph
...
-		user.Delete(itemUser{username: u})
+	for u := range userList {
```
tiancaiamao, line 586: The new way: merge = old +- userList data (changes) 
and userList data = apply userList to diff (it can be data of userList or data of all users)

The new way is more general and O(n) to userList instead of O(n) to diff... the diff can be get by `LoadAll`

pkg/privilege/privileges/cache.go
```diff
@@ -866,56 +868,75 @@ func compareTablesPrivRecord(x, y tablesPrivRecord) int {
...
 	}
-	for _, row := range rows {
-		// NOTE: decodeTableRow decodes data from a chunk Row, that is a shallow copy.
-		// The result will reference memory in the chunk, so the chunk must not be reused
-		// here, otherwise some werid bug will happen!
-		err = decodeTableRow(row, fs)
+	defer terror.Call(rs.Close)
+	fs := rs.Fields()
+	req := rs.NewChunk(nil)
+	for {
```
tiancaiamao, line 921: Yes, now the streaming API instead a single 2M rows array.

pkg/privilege/privileges/cache.go
```diff
@@ -953,229 +974,269 @@ func parseHostIPNet(s string) *net.IPNet {
 func (record *baseRecord) assignUserOrHost(row chunk.Row, i int, f *resolve.ResultField) {
 	switch f.ColumnAsName.L {
 	case "user":
-		record.User = row.GetString(i)
+		record.User = strings.Clone(row.GetString(i))
```
tiancaiamao, line 977: Deep copy to avoid referencing chunk data and cause memory leak.

pkg/privilege/privileges/cache.go
```diff
@@ -953,229 +974,269 @@ func parseHostIPNet(s string) *net.IPNet {
...
-		case f.ColumnAsName.L == "user_attributes":
-			if row.IsNull(i) {
-				continue
-			}
-			bj := row.GetJSON(i)
-			pathExpr, err := types.ParseJSONPathExpr("$.metadata.email")
-			if err != nil {
-				return err
+func (p *MySQLPrivilege) decodeUserTableRow(userList map[string]struct{}) func(chunk.Row, []*resolve.ResultField) error {
+	return func(row chunk.Row, fs []*resolve.ResultField) error {
```
tiancaiamao, line 986: The decoder function is changed from 

```
func (p *MySQLPrivilege) decodeUserTableRow(row chunk.Row, fs []*resolve.ResultField) error
```

to 

```
func (p *MySQLPrivilege) decodeUserTableRow(userList map[string]struct{}) func(chunk.Row, []*resolve.ResultField) error
```

Because now we need the user = 'xxx' filter condition during decoding...
The code seems to be changed a lot but it's really just function signature changes.

pkg/privilege/privileges/cache.go
```diff
@@ -1203,56 +1264,72 @@ func (p *MySQLPrivilege) decodeRoleEdgesTable(row chunk.Row, fs []*resolve.Resul
...
-	var value defaultRoleRecord
-	for i, f := range fs {
-		switch f.ColumnAsName.L {
-		case "default_role_host":
-			value.DefaultRoleHost = row.GetString(i)
-		case "default_role_user":
-			value.DefaultRoleUser = row.GetString(i)
-		default:
-			value.assignUserOrHost(row, i, f)
+func (p *MySQLPrivilege) decodeDefaultRoleTableRow(userList map[string]struct{}) func(chunk.Row, []*resolve.ResultField) error {
```
tiancaiamao, line 1267: OK, this is the simplest on among all the decoder function changes.
Read the changes here and you skip all the others (because other the changes are same).

pkg/privilege/privileges/cache.go
```diff
@@ -2090,30 +2181,30 @@ func (h *Handle) UpdateAll() error {
...
 func (h *Handle) UpdateAllActive() error {
-	h.fullData.Store(false)
 	userList := make([]string, 0, 20)
 	h.activeUsers.Range(func(key, _ any) bool {
 		userList = append(userList, key.(string))
 		return true
 	})
-	if len(userList) > 1024 {
-		logutil.BgLogger().Warn("active user count > 1024, revert to update all", zap.Int("len", len(userList)))
-		return h.UpdateAll()
```
tiancaiamao, line 2101: The root cause of the memory usage jitter, now fixed.

pkg/privilege/privileges/cache.go
```diff
@@ -497,7 +497,13 @@ func (p *MySQLPrivilege) LoadAll(ctx sqlexec.RestrictedSQLExecutor) error {
 	return nil
 }
 
-func findUserAndAllRoles(all map[string]struct{}, roleGraph map[string]roleGraphEdgesTable) {
+func findUserAndAllRoles(userList []string, roleGraph map[string]roleGraphEdgesTable) map[string]struct{} {
```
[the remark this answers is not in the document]
lance6716: And I think at least we can iterate `roleGraph` once to build a map of `userOrRoleName` -> `roleList`, so when we found new roles we don't need to iterate  whole `roleGraph` again.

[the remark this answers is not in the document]
tiancaiamao: > I'm checking this function. In line 551 `strings.Cut(userHost, "@")` we should handle the case that username / host containing `@`.

Good point, I find that there some code related to role need to handle the case `@`
That's use another PR to fix the bug...

This one focus on the optimize of reducing privilege relaod memory.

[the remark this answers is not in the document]
tiancaiamao: I create an issue for it https://github.com/pingcap/tidb/issues/59552

pkg/privilege/privileges/cache.go
```diff
@@ -866,58 +869,76 @@ func compareTablesPrivRecord(x, y tablesPrivRecord) int {
...
+			first = false
 		}
 		fmt.Fprintf(&b, "USER = '%s'", sqlescape.EscapeString(user))
 	}
 	return b.String()
 }
 
-func (p *MySQLPrivilege) loadTable(sctx sqlexec.RestrictedSQLExecutor, sql string,
-	decodeTableRow func(chunk.Row, []*resolve.ResultField) error, userList ...string) error {
+func (p *MySQLPrivilege) loadTable(exec sqlexec.SQLExecutor, sql string,
```
CbcWestwolf, line 909:
```suggestion
func (*MySQLPrivilege) loadTable(exec sqlexec.SQLExecutor, sql string,
```

pkg/privilege/privileges/cache.go
```diff
@@ -2079,7 +2162,14 @@ func (h *Handle) Get() *MySQLPrivilege {
 func (h *Handle) UpdateAll() error {
 	logutil.BgLogger().Warn("update all called")
```
CbcWestwolf, line 2163: Since this is a WARN log and `DefTiDBAccelerateUserCreationUpdate` is false, TiDB would always print this log even if UpdateAll only loads tens or hundreds of items in MySQLPrivilege.
Maybe we can only print this WARN if the number of items exceeds a threshold like `1024` after `priv.LoadAll()`

pkg/privilege/privileges/cache.go
```diff
@@ -497,171 +497,174 @@ func (p *MySQLPrivilege) LoadAll(ctx sqlexec.RestrictedSQLExecutor) error {
...
-		}
 	}
+	return all
 }
 
-func (p *MySQLPrivilege) loadSomeUsers(ctx sqlexec.RestrictedSQLExecutor, userList ...string) ([]string, error) {
-	if len(userList) > 512 {
-		logutil.BgLogger().Warn("loadSomeUsers called with a long user list", zap.Int("len", len(userList)))
-	}
+func (p *MySQLPrivilege) loadSomeUsers(ctx sqlexec.SQLExecutor, users ...string) (map[string]struct{}, error) {
```
CbcWestwolf, line 534:
```suggestion
func (p *MySQLPrivilege) loadSomeUsers(ctx sqlexec.SQLExecutor, map[string]struct{}) error {
```

Since the return value of `loadSomeUsers` is indeed the same as `findUserAndAllRoles`, we can call `findUserAndAllRoles` outsides and pass the map in as a parameter.

pkg/privilege/privileges/cache.go
```diff
@@ -2128,12 +2220,19 @@ func (h *Handle) Update(userList []string) error {
 		return nil
 	}
 
+	res, err := h.sctx.Get()
```
CbcWestwolf, line 2223: Actually I do not quite understand why we can skip reload if the user is not 'active'. Should we also Update the user that is not 'active'?

CbcWestwolf (approved):
LGTM
I abstract some common code of `loadSomeUsers` to a new method `updateUsers`. 
Apply it if you prefer this change.

```diff
diff --git a/pkg/privilege/privileges/cache.go b/pkg/privilege/privileges/cache.go
index 2f0b189a20..f475fa124a 100644
--- a/pkg/privilege/privileges/cache.go
+++ b/pkg/privilege/privileges/cache.go
@@ -531,53 +531,43 @@ func findUserAndAllRoles(userList []string, roleGraph map[string]roleGraphEdgesT
 	return all
 }
 
-func (p *MySQLPrivilege) loadSomeUsers(ctx sqlexec.SQLExecutor, users ...string) (map[string]struct{}, error) {
-	// Load the full role edge table first.
-	p.roleGraph = make(map[string]roleGraphEdgesTable)
-	err := p.loadTable(ctx, sqlLoadRoleGraph, p.decodeRoleEdgesTable)
-	if err != nil {
-		return nil, errors.Trace(err)
-	}
-
-	// Including the user and also their roles
-	userList := findUserAndAllRoles(users, p.roleGraph)
-
-	err = p.loadTable(ctx, addUserFilterCondition(sqlLoadUserTable, userList), p.decodeUserTableRow(userList))
+func (p *MySQLPrivilege) loadSomeUsers(ctx sqlexec.SQLExecutor, userAndRoles map[string]struct{}) error {
+	err := p.loadTable(ctx, addUserFilterCondition(sqlLoadUserTable, userAndRoles), p.decodeUserTableRow(userAndRoles))
 	if err != nil {
-		return nil, errors.Trace(err)
+		return errors.Trace(err)
 	}
 
-	err = p.loadTable(ctx, addUserFilterCondition(sqlLoadGlobalPrivTable, userList), p.decodeGlobalPrivTableRow(userList))
+	err = p.loadTable(ctx, addUserFilterCondition(sqlLoadGlobalPrivTable, userAndRoles), p.decodeGlobalPrivTableRow(userAndRoles))
 	if err != nil {
-		return nil, errors.Trace(err)
+		return errors.Trace(err)
 	}
 
-	err = p.loadTable(ctx, addUserFilterCondition(sqlLoadGlobalGrantsTable, userList), p.decodeGlobalGrantsTableRow(userList))
+	err = p.loadTable(ctx, addUserFilterCondition(sqlLoadGlobalGrantsTable, userAndRoles), p.decodeGlobalGrantsTableRow(userAndRoles))
 	if err != nil {
-		return nil, errors.Trace(err)
+		return errors.Trace(err)
 	}
 
-	err = p.loadTable(ctx, addUserFilterCondition(sqlLoadDBTable, userList), p.decodeDBTableRow(userList))
+	err = p.loadTable(ctx, addUserFilterCondition(sqlLoadDBTable, userAndRoles), p.decodeDBTableRow(userAndRoles))
 	if err != nil {
-		return nil, errors.Trace(err)
+		return errors.Trace(err)
 	}
 
-	err = p.loadTable(ctx, addUserFilterCondition(sqlLoadTablePrivTable, userList), p.decodeTablesPrivTableRow(userList))
+	err = p.loadTable(ctx, addUserFilterCondition(sqlLoadTablePrivTable, userAndRoles), p.decodeTablesPrivTableRow(userAndRoles))
 	if err != nil {
-		return nil, errors.Trace(err)
+		return errors.Trace(err)
 	}
 
-	err = p.loadTable(ctx, addUserFilterCondition(sqlLoadDefaultRoles, userList), p.decodeDefaultRoleTableRow(userList))
+	err = p.loadTable(ctx, addUserFilterCondition(sqlLoadDefaultRoles, userAndRoles), p.decodeDefaultRoleTableRow(userAndRoles))
 	if err != nil {
-		return nil, errors.Trace(err)
+		return errors.Trace(err)
 	}
 
-	err = p.loadTable(ctx, addUserFilterCondition(sqlLoadColumnsPrivTable, userList), p.decodeColumnsPrivTableRow(userList))
+	err = p.loadTable(ctx, addUserFilterCondition(sqlLoadColumnsPrivTable, userAndRoles), p.decodeColumnsPrivTableRow(userAndRoles))
 	if err != nil {
-		return nil, errors.Trace(err)
+		return errors.Trace(err)
 	}
 
-	return userList, nil
+	return nil
 }
 
 // merge construct a new MySQLPrivilege by merging the data of the two objects.
@@ -2123,21 +2113,8 @@ func (h *Handle) ensureActiveUser(ctx context.Context, user string) error {
 	if exist {
 		return nil
 	}
-	data := newMySQLPrivilege()
-	data.globalVars = h.globalVars
-	res, err := h.sctx.Get()
-	if err != nil {
-		return errors.Trace(err)
-	}
-	defer h.sctx.Put(res)
-	exec := res.(sqlexec.SQLExecutor)
 
-	userList, err := data.loadSomeUsers(exec, user)
-	if err != nil {
-		return errors.Trace(err)
-	}
-	h.merge(data, userList)
-	return nil
+	return h.updateUsers([]string{user})
 }
 
 func (h *Handle) merge(data *MySQLPrivilege, userList map[string]struct{}) {
@@ -2186,21 +2163,7 @@ func (h *Handle) UpdateAllActive() error {
 		userList = append(userList, key.(string))
 		return true
 	})
-	res, err := h.sctx.Get()
-	if err != nil {
-		return errors.Trace(err)
-	}
-	defer h.sctx.Put(res)
-	exec := res.(sqlexec.SQLExecutor)
-
-	priv := newMySQLPrivilege()
-	priv.globalVars = h.globalVars
-	userList1, err := priv.loadSomeUsers(exec, userList...)
-	if err != nil {
-		return errors.Trace(err)
-	}
-	h.merge(priv, userList1)
-	return nil
+	return h.updateUsers(userList)
 }
 
 // Update loads the privilege info from kv storage for the list of users.
@@ -2220,6 +2183,10 @@ func (h *Handle) Update(userList []string) error {
 		return nil
 	}
 
+	return h.updateUsers(userList)
+}
+
+func (h *Handle) updateUsers(userList []string) error {
 	res, err := h.sctx.Get()
 	if err != nil {
 		return errors.Trace(err)
@@ -2227,12 +2194,20 @@ func (h *Handle) Update(userList []string) error {
 	defer h.sctx.Put(res)
 	exec := res.(sqlexec.SQLExecutor)
 
-	priv := newMySQLPrivilege()
-	priv.globalVars = h.globalVars
-	userList1, err := priv.loadSomeUsers(exec, userList...)
+	p := newMySQLPrivilege()
+	p.globalVars = h.globalVars
+	// Load the full role edge table first.
+	p.roleGraph = make(map[string]roleGraphEdgesTable)
+	err = p.loadTable(exec, sqlLoadRoleGraph, p.decodeRoleEdgesTable)
 	if err != nil {
-		return err
+		return errors.Trace(err)
+	}
+	// Including the user and also their roles
+	userAndRoles := findUserAndAllRoles(userList, p.roleGraph)
+	err = p.loadSomeUsers(exec, userAndRoles)
+	if err != nil {
+		return errors.Trace(err)
 	}
-	h.merge(priv, userList1)
+	h.merge(p, userAndRoles)
 	return nil
 }
```

lcwangchao (approved):

lance6716 (commented):
/approve

for domain part

PR #59496, opened 2025-02-13
prefetch: the prefetch buffer should be fully filled

### LINKED ISSUE #59495: performance is not expected for prefetch reader
In current code we don't fully make use of the prefetch buffer. There's a background goroutine (spawned in line 48) that uses the prefetch buffer to read data from reader (line 57)

https://github.com/pingcap/tidb/blob/b7e97690b9feb019ab0bec9ea6814af432ae948d/pkg/util/prefetch/reader.go#L48-L63

However if the reader only return partial data (like it's a socket and OS chooses to return only few bytes, naming `N`) this reading action is finished and the loop is waiting on sending to channel (line 62). If the caller of "prefetch reader" doesn't consume the channel quickly, we are wasting a large proportion of the prefetch buffer which is `buf[N:]` because they are not filled. We should let the background goroutine use `io.ReadFull` to fully use the prefetch buffer.

lance6716:

Issue Number: close #59495

Problem Summary:

use `io.ReadFull` to fully fill the buffer

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

performance increased in a AWS manual test

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/util/prefetch/reader.go
+++ b/pkg/util/prefetch/reader.go
@@ -16,6 +16,7 @@ package prefetch
 
 import (
 	"bytes"
+	"errors"
 	"io"
 	"sync"
 )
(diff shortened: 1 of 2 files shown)


D3Hunter (approved):

Benjamin2037 (approved):
LGTM

pkg/util/prefetch/reader.go
```diff
@@ -54,14 +55,20 @@ func (r *Reader) run() {
 	for {
 		r.bufIdx = (r.bufIdx + 1) % 2
 		buf := r.buf[r.bufIdx]
-		n, err := r.r.Read(buf)
+		n, err := io.ReadFull(r.r, buf)
```
Benjamin2037, line 58: Do we need add comment to explain why read  will not return the err?

PR #59522, opened 2025-02-13
statistics: release session met error to avoid possible leak

winoros:

Issue Number: ref #54022 

Problem Summary:

that pull request doesn't solve the root problem

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that the internal session used by statistics may not be released and cause possible memory leak.
修复统计信息是使用的内部会话在遇到错误时可能没有被释放的问题，该问题可能导致内存溢出。
```

### DIFF
--- a/pkg/statistics/handle/util/util.go
+++ b/pkg/statistics/handle/util/util.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/failpoint"
+	"github.com/pingcap/tidb/pkg/domain/infosync"
 	"github.com/pingcap/tidb/pkg/kv"
 	"github.com/pingcap/tidb/pkg/meta/model"
 	"github.com/pingcap/tidb/pkg/metrics"
@@ -85,6 +86,8 @@ func CallWithSCtx(pool util.SessionPool, f func(sctx sessionctx.Context) error, 
 	defer func() {
 		if err == nil { // only recycle when no error
 			pool.Put(se)
+		} else {
+			infosync.DeleteInternalSession(se)
 		}
 	}()
 	sctx := se.(sessionctx.Context)
--- a/pkg/domain/infosync/info.go
+++ b/pkg/domain/infosync/info.go
@@ -1339,6 +1339,19 @@ func DeleteInternalSession(se any) {
 	sm.DeleteInternalSession(se)
 }
 
+// HasInternalSessionForTest is the entry function for check whether an internal session is stored in SessionManager.
+func HasInternalSessionForTest(se any) bool {
+	is, err := getGlobalInfoSyncer()
+	if err != nil {
+		return false
+	}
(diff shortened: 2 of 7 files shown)


winoros:
/hold
feel free to unhold after 1 lgtm of sql-infra team

winoros:
/unhold
oh, the approver from the domain package is enough.
no need to add the manually checking

pkg/statistics/handle/util/util.go
```diff
@@ -85,6 +86,8 @@ func CallWithSCtx(pool util.SessionPool, f func(sctx sessionctx.Context) error,
 	defer func() {
 		if err == nil { // only recycle when no error
 			pool.Put(se)
+		} else {
+			infosync.DeleteInternalSession(se)
```
tangenta, line 90: I think it is better to put the session in the pool after cleaning up the dirty state..

winoros: but how to clean it(

lance6716 (commented):
I see that if we return the session back to pool, the pool should be able to call `DeleteInternalSession`

https://github.com/pingcap/tidb/blob/f6bf8e8bec58d9470a64e05603a77295dcaa5133/pkg/domain/domain.go#L1332-L1346

Can you reproduce this problem in our environment first? So we can use goref or other ways to know why the reference is kept

Rustin170506:
> I see that if we return the session back to pool, the pool should be able to call `DeleteInternalSession`

It is challenging to determine whether an error stems from the business logic or from the internal session. This complicates error handling when issues arise. As a result, we decided not to return the session to the pool, but this decision has led to memory leak problems. If we continue failing and create numerous sessions without properly releasing them, it exacerbates the issue.

I propose that we introduce a new API for the session pool that allows the caller to destroy a session instead of directly returning it.

lance6716:
> > I see that if we return the session back to pool, the pool should be able to call `DeleteInternalSession`
> 
> It is challenging to determine whether an error stems from the business logic or from the internal session. This complicates error handling when issues arise. As a result, we decided not to return the session to the pool, but this decision has led to memory leak problems. If we continue failing and create numerous sessions without properly releasing them, it exacerbates the issue.
> 
> I propose that we introduce a new API for the session pool that allows the caller to destroy a session instead of directly returning it.

This is not related to check the source of error? Line 88 of `pkg/statistics/handle/util/util.go` should also call `DeleteInternalSession`

Rustin170506:
> This is not related to check the source of error? Line 88 of `pkg/statistics/handle/util/util.go` should also call `DeleteInternalSession`

If we do not check the source of the error, we cannot put it back in the pool directly. This is also why we call `DeleteInternalSession` from the caller side.

Issue #49008, opened 2023-11-29
resource control inside dist framework


D3Hunter:
## Feature Request

**Is your feature request related to a problem? Please describe:**

- Currently, the dispatcher manager supports up to 4 tasks, and the Scheduler manager can process up to 4 tasks.
- The dispatcher/Scheduler manager does not consider the concurrent configuration of tasks when processing tasks, and resources may be overused, resulting in OOM.
- The above manager does not consider the order of tasks. After task A is processed halfway, it may start processing other tasks due to failover and other reasons, causing task A to keep waiting.

**Describe the feature you'd like:**

This feature will add control over cpu/mem/disk within the nodes managed by framework(see `tidb_service_scope`), improves resource usage, avoids overall resource usage exceeding limits, and ensures task execution order. When executing multiple tasks in parallel, ensure that each task can run normally without affecting each other.
in first version, we consider dist framework can use all resources that it runs on.

### resource control
- [x] #49019
- [x] #49100
- [x] #49135
- [x] https://github.com/pingcap/tidb/pull/49403
- [x] https://github.com/pingcap/tidb/pull/49623
- [x] https://github.com/pingcap/tidb/pull/49758
- [x] https://github.com/pingcap/tidb/pull/49713
- [x] https://github.com/pingcap/tidb/pull/49875
- [x] https://github.com/pingcap/tidb/pull/50091
- [x] https://github.com/pingcap/tidb/pull/50266
- [x] https://github.com/pingcap/tidb/pull/50475
- [ ] add-index restrict resource usage
- [ ] import-into restrict resource usage
- [x] https://github.com/pingcap/tidb/pull/50650
- [x] https://github.com/pingcap/tidb/pull/50540
- [x] https://github.com/pingcap/tidb/pull/50724
- [x] https://github.com/pingcap/tidb/pull/50801
- [x] https://github.com/pingcap/tidb/pull/50962
- [x] https://github.com/pingcap/tidb/pull/51136
- [x] https://github.com/pingcap/tidb/pull/51182
- [ ] https://github.com/pingcap/tidb/pull/51231
- [ ] https://github.com/pingcap/tidb/pull/51022
- [ ] https://github.com/pingcap/tidb/pull/51003
- [ ] https://github.com/pingcap/tidb/pull/50987

### others
- [x] https://github.com/pingcap/tidb/pull/49025
- [x] #49125
- [x] https://github.com/pingcap/tidb/issues/49617
[...]

PR #59607, opened 2025-02-18
disttask: add local disk space to node resource managed by DXF

tangenta:

Issue Number: ref #49008

Problem Summary:

We didn't consider disk space for each subtask when implementing DXF. In the following enhancement, we may utilize the local storage for global sort, it requires managing the disk usage resource.

- Add disk space to `NodeResource`.
- Move `NodeResource` from package `taskexecutor` to `proto`.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/disttask/framework/proto/node.go
```diff
@@ -23,3 +23,44 @@ type ManagedNode struct {
...
+func (nr *NodeResource) GetStepResource(concurrency int) *StepResource {
+	return &StepResource{
+		CPU: NewAllocatable(int64(concurrency)),
+		// same proportion as CPU
+		Mem: NewAllocatable(int64(float64(concurrency) / float64(nr.TotalCPU) * float64(nr.TotalMem))),
+	}
+}
+
+// GetTaskDiskResource gets available disk for a task.
+func (nr *NodeResource) GetTaskDiskResource(concurrency int, quotaHint uint64) uint64 {
```
D3Hunter, line 63: is `quotaHint` used to allocate less resource for some task? why do we need it if we cannot allow some task to alloc more resource?

and it should be part of `StepResource`, no need a special API

tangenta: `StepResource` is only used by the step executor. We need to know the resource in the scheduler before splitting subtasks.

tangenta: > is quotaHint used to allocate less resource for some task?

Yes.

> why do we need it if we cannot allow some task to alloc more resource?

Because other factors may affect the resource allocation, for example, `@@tidb_ddl_disk_quota` can change dynamically.

D3Hunter: if the resource per slot for a subtask is always <= `the mount of disk per slot` = `D`, seems ok to just use `D`, no need to adjust

tangenta: We can't determine which one is larger, `tidb_ddl_disk_quota` or actual disk capacity. For example, a user has a 2TB node, but only wants to use 500GB for DXF, then the user can set `tidb_ddl_disk_quota` to 500GB

D3Hunter: no need to consider `tidb_ddl_disk_quota`, we can always use the `total_disk` we got by inspecting disk stat.

pkg/disttask/framework/testutil/context.go
```diff
@@ -166,13 +167,14 @@ func (c *TestDXFContext) ScaleOut(nodeNum int) {
 // ScaleOutBy scales out a tidb node by id, and set it as owner if required.
 func (c *TestDXFContext) ScaleOutBy(id string, owner bool) {
 	c.T.Logf("scale out node of id = %s, owner = %t", id, owner)
-	exeMgr, err := taskexecutor.NewManager(c.Ctx, id, c.TaskMgr)
+	nodeRes := proto.NewNodeResource(16, 32*units.GiB)
+	exeMgr, err := taskexecutor.NewManager(c.Ctx, id, c.TaskMgr, nodeRes)
```
D3Hunter, line 171: in `TestDXFContext`, the cpu count is a input param that's passed to `github.com/pingcap/tidb/pkg/util/cpu/mockNumCpu",`, the new node need also conform to this

see newTestDXFContext

D3Hunter (commented):
rest lgtm

pkg/ddl/ingest/env.go
```diff
@@ -93,13 +93,20 @@ func InitGlobalLightningEnv(path string) (ok bool) {
 	return true
 }
 
-// GenIngestTempDataDir generates a path for DDL ingest.
+// GetIngestTempDataDir gets the path for DDL ingest.
 // Format: ${temp-dir}/tmp_ddl-{port}
-func GenIngestTempDataDir() (string, error) {
+func GetIngestTempDataDir() string {
```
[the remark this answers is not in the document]
tangenta: This will be used in the next PR which utilize local storage for global sort.

pkg/ddl/backfilling_dist_scheduler.go
```diff
@@ -125,6 +130,9 @@ func (sch *LitBackfillScheduler) OnNextSubtasksBatch(
 		if tblInfo.Partition != nil {
 			return generatePartitionPlan(ctx, storeWithPD, tblInfo)
 		}
+		// TODO(tangenta): use available disk during adding index.
```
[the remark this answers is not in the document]
tangenta: In scheduler we can't get the `StepResource` which is only used in subtask executors.

pkg/disttask/framework/taskexecutor/manager.go
```diff
@@ -75,51 +71,24 @@ type Manager struct {
 	logger      *zap.Logger
 	slotManager *slotManager
 
-	totalCPU int
-	totalMem int64
+	totalCPU     int
+	totalMem     int64
```
D3Hunter, line 75: can be removed now

pkg/disttask/framework/testutil/context.go
```diff
@@ -354,7 +357,7 @@ func (c *TestDXFContext) electIfNeeded() {
 	newOwnerIdx := int(rand.Int31n(int32(len(c.mu.nodes))))
 	ownerNode := c.mu.nodes[newOwnerIdx]
 	c.mu.ownerIndices[ownerNode.id] = newOwnerIdx
-	ownerNode.schMgr = scheduler.NewManager(c.Ctx, c.TaskMgr, ownerNode.id)
+	ownerNode.schMgr = scheduler.NewManager(c.Ctx, c.TaskMgr, ownerNode.id, proto.NodeResourceForTest)
```
D3Hunter, line 360: also need change to `proto.NewNodeResource(c.mockCPUNum, 32*units.GB, 100*units.GB)`

D3Hunter (approved):

CbcWestwolf (approved):
LGTM

PR #59663, opened 2025-02-20
br/stream: implement pause v2

### LINKED ISSUE #59660: br/stream: save more information to pause (pause v2)
### Background

Now, when a task encounters a fatal error, the following keys are written:

```
$prefix/last_error/$task_name → kvproto::brpb::StreamBackupError
$prefix/paused/$task_name     → <nothing>
```

So far, we have suffered some pains due to this like: 
- When someone pauses the task, it is impossible to track the reason for pausing. 
  - This happens when a task is paused by a new feature, like auto pausing tasks with huge lags.
- When a user mistakenly pauses a task, we cannot figure out when and how he paused it.
- Putting pause and error isn't atomic. This may cause a fatal error to be treated as retryable and the user may recover an unrecoveriable task.

### Design

Here I propose a new format of pausing:

```
$prefix/paused/$task_name → PauseReason
Where `PauseReason` looks like:
{
    "severity": "ERROR" | "REGULAR_OPERATION",
    "operation_hostname": string,
    "operation_pid": int,
    "operation_time": datetime,
    "payload_type": "application/x-protobuf;messageType=brpb.StreamBackupError" | "text/plain;charset=UTF-8",
    "payload": bytes,
}
```

Notes:
- Though `payload_type` uses the same format as [MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/MIME_types), for safety and simplicity, only the message `brpb.StreamBackupError` and UTF-8 strings are accepted by `log status`. The BR CLI will refuse to parse the payload if the type isn't one of them.
- I think `PauseReason` can be JSON instead of Protocol Buffer Message, because in almost all cases, the information it provides is for humans. JSON is more flexible and more friendly to humans.

### Compatibility / Feasibility

For now, the pause key was used by both TiKV and the Coordinator. But neither of them reads the value of this key. So modifications to values are safe.

The error key was written by TiKV and only reads by the BR CLI (log status). 

A possible update route:

1. Update TiDB codebase, allowing BR CLI to read information from Pause V2.
    - We just read error messages from both V1 and V2.
    - In this step, the base library for Pause V2 can be added.2. Allowing the coordinator and BR CLI to pause a task by V2.
2. Maybe allow the user to attach a custom message to this pause.
3. Allowing the TiKV to upload errors by V2.

YuJuncen:

Issue Number: close #59660 

Problem Summary:
See the issue. Breifly say, we need more information of pausing for troubleshooting.

When pausing, more information was added.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


br/pkg/streamhelper/client.go
```diff
@@ -19,6 +24,139 @@ import (
...
+// PauseV2 is extra information attached to a paused task.
+type PauseV2 struct {
+	// Severity is the severity of this pause.
+	// `SeverityError`: The task encounters some fatal errors and have to be paused.
+	// `SeverityRegularOperation`: The task was paused by a normal operation.
+	Severity string `json:"severity"`
+	// OperatorHostName is the hostname that pauses the task.
+	OperatorHostName string `json:"operation_hostname"`
+	// OperatorPID is the pid of the operator process.
+	OperatorPID int `json:"operation_pid"`
```
RidRisR, line 84: Why do we need to record the pid?

RidRisR (approved):

br/pkg/streamhelper/client.go
```diff
@@ -19,6 +24,139 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	SeverityError            = "ERROR"
```
Tristan1900, line 28: what do you think of replacing `Severity` to be `Reason` and `REGULAR_OPERATION` to be `MANUAL`

br/pkg/stream/stream_status.go
```diff
@@ -46,13 +46,22 @@ type TaskStatus struct {
...
+	// PauseV2 is the extra information attached to the pause key.
+	PauseV2 *PauseV2
 }
 
 type TaskPrinter interface {
 	AddTask(t TaskStatus)
 	PrintTasks()
 }
 
+func TeeTaskPrinter(p TaskPrinter, output *[]TaskStatus) TaskPrinter {
```
Tristan1900, line 58: what does `Tee` mean here?

Tristan1900 (commented):
I suppose we need to do the same thing on TiKV side as it can pause a task as well?

br/pkg/streamhelper/client.go
YuJuncen: LGTM, and the TiKV side will be implemented later.

br/pkg/streamhelper/client.go
YuJuncen: Just to distunglish two commands from the same host.

BornChanger (approved):

Issue #59588, opened 2025-02-17
Unit tests of stale-read should use real TiKV


ekexium:
## Enhancement
Unit tests of stale-read should use real TiKV to expose potential problems when interacting with PD or PD client, especially when it's related to txn scope or read replica scope.

#59402 could have been detected earlier by unit tests if we've done so before.

PR #59715, opened 2025-02-24
test: TestStaleReadAllCombinations

ekexium:

Issue Number: ref #59588

Problem Summary:

Add a test to cover the combinations of {stale read methods} x {replica read settings} x {label settings}

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


tests/realtikvtest/txntest/stale_read_test.go
```diff
@@ -1506,3 +1506,210 @@ func TestStaleReadNoBackoff(t *testing.T) {
...
+				_, _ = tk.Exec(fmt.Sprintf("set @@global.tidb_external_ts=%d", externalTS))
+			},
+			query: "select * from t",
+			clean: func() {
+				tk.MustExec("set @@tidb_enable_external_ts_read=0")
+				// cannot set it
+				// tk.MustExec("set @@global.tidb_external_ts=0")
+			},
+			expect: []string{"1 10"},
+		},
```
cfzjywxk, line 1606: The `set @@tidb_snapshot='2016-10-08 16:45:26'` could also be covered.

tests/realtikvtest/txntest/stale_read_test.go
```diff
@@ -1506,3 +1506,210 @@ func TestStaleReadNoBackoff(t *testing.T) {
...
+		{
+			name:   "no labels",
+			labels: map[string]string{},
+		},
+		{
+			name: "with DC label",
+			labels: map[string]string{
+				placement.DCLabelKey: "bj",
+			},
+		},
```
cfzjywxk, line 1622: The `zone`, `dc`, `rack` could also be covered though some may not take effect on closest.

cfzjywxk (approved):

tests/realtikvtest/txntest/stale_read_test.go
```diff
@@ -140,16 +140,14 @@ func TestSelectAsOf(t *testing.T) {
 	store := realtikvtest.CreateMockStoreAndSetup(t)
 	tk := testkit.NewTestKit(t, store)
 	tk.MustExec("use test")
-	if !*realtikvtest.WithRealTiKV {
-		// For mocktikv, safe point is not initialized, we manually insert it for snapshot to use.
-		safePointName := "tikv_gc_safe_point"
-		safePointValue := "20160102-15:04:05 -0700"
-		safePointComment := "All versions after safe point can be accessed. (DO NOT EDIT)"
-		updateSafePoint := fmt.Sprintf(`INSERT INTO mysql.tidb VALUES ('%[1]s', '%[2]s', '%[3]s')
+	// For mocktikv, safe point is not initialized, we manually insert it for snapshot to use.
```
MyonKeminta, line 143: Is this still mocktikv-only? Or is it needed for real tikv?

tests/realtikvtest/txntest/stale_read_test.go
```diff
@@ -1506,3 +1493,245 @@ func TestStaleReadNoBackoff(t *testing.T) {
...
+		{
+			name:   "AS OF TIMESTAMP sees both rows (using secondTime)",
+			setup:  func() {},
+			query:  fmt.Sprintf("SELECT * FROM t AS OF TIMESTAMP '%s'", secondTime.Format("2006-1-2 15:04:05.000")),
+			clean:  func() {},
+			expect: []string{"1 10", "2 20"},
+		},
+		/*
+			We cannot test TIDB_BOUNDED_STALENESS here because it relies
+			on the update of safe ts, which cannot be controlled by the test
```
MyonKeminta, line 1565: 🤔 Do we have any other way to test this case?

tests/realtikvtest/txntest/stale_read_test.go
```diff
@@ -1506,3 +1493,245 @@ func TestStaleReadNoBackoff(t *testing.T) {
...
+			query:  "select * from t",
+			clean:  func() {},
+			expect: []string{"1 10"},
+		},
+		{
+			name: "external ts read",
+			setup: func() {
+				tk.MustExec("set @@tidb_enable_external_ts_read=1")
+				// ignore the failure when setting external_ts, as it reports error if not set to a larger value
+				_, _ = tk.Exec(fmt.Sprintf("set @@global.tidb_external_ts=%d", externalTS))
```
MyonKeminta, line 1588: I didn't understand it. Does it mean that it returns error while it's actually successful to se it?

tests/realtikvtest/txntest/stale_read_test.go
ekexium: It's needed for real TiKV as well. The comments are deleted

tests/realtikvtest/txntest/stale_read_test.go
ekexium: In the combination test the setting is executed multiple times. When setting to a same value(namely not a larger one) the SQL returns error, which we do not care about.

tests/realtikvtest/txntest/stale_read_test.go
ekexium: I haven't found an effective way to implement this in the combinatorial test.

tests/realtikvtest/txntest/stale_read_test.go
MyonKeminta: I'm afraid it doesn't look like a good idea to do this on real tikv. But let's accept it for now as it would be too complex to do it elegantly.

tests/realtikvtest/txntest/stale_read_test.go
MyonKeminta: I'm afraid it would make us unable to identify the problem when there's some other problem causing the statement failed. Can we consider using a increasing timestamp in this case, as it seems we have a 500ms gap of timestamps free to use in this test, so that the error can be avoided?

ekexium: That makes sense. Modified.

MyonKeminta (approved):

Issue #58068, opened 2024-12-08
Using modifyCount in the optimizer can inflate estimation compared to realtimeRowCount


terry1purcell:
Title says it all

PR #59724, opened 2025-02-25
planner: Adjust out of range for added rows

terry1purcell:

Issue Number: ref #58068

Problem Summary:

Further fine tuning of the out of range estimation based upon customer feedback. This change focuses on adjusting the rowCount only if the added rows since last ANALYZE exceeds the out-of-range estimation based upon the predicate search range.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/statistics/histogram.go
```diff
@@ -1078,19 +1078,18 @@ func (hg *Histogram) OutOfRangeRowCount(
...
-		totalPercent = min(0.5, max(leftPercent, rightPercent))
-		rowCount += totalPercent * addedRows
-		if rowCount < upperBound {
-			rowCount = min(upperBound, addedRows)
+		if histInvalid {
+			totalPercent = min(addedPct, 0.5)
 		}
+		// Attempt to account for the added rows - but not more than the totalPercent
+		outOfRangeAdded := addedRows * totalPercent
+		rowCount = max(rowCount, max(outOfRangeAdded, upperBound))
```
qw4990, line 1092: Should be `rowCount = max(rowCount, min(outOfRangeAdded, upperBound))`?

qw4990: Using an max function on `upperBound` seems unusual.

qw4990, line 1092: So the main change if I understand it correctly is that we change the formula here from `rowCount = rowCount + outOfRangeAdded` to `rowCount = max(rowCount, outOfRangeAdded)`, which seems more conservative than before.

terry1purcell: Yes - that is correct.

terry1purcell: Since this value is used in the code as an "upper bound" in one scenario, and a "lower bound" in another - I renamed the variable to "oneValue" to avoid such confusion.

qw4990 (approved):

hawkingrei (approved):

Issue #50451, opened 2024-01-16
import into failed when inject global sort dir network delay 50ms with error “read xxx bytes from external storage, exceed max limit 1073741824”


Lily2025:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
1、tidb_enable_dist_task='on' and 
2、enable global sort and tidb_cloud_storage_uri is set to minio dir
3、run import into
4、inject minio network delay 50ms

### 2. What did you expect to see? (Required)
import into can success

### 3. What did you see instead (Required)
import failed
						
the status of import job is not finished or running (now: 2024-01-16 02:48:14, jobId: 150003, step: importing, status: failed)
operatorLogs:
[2024-01-16 02:28:44] ###### start import into
[2024-01-16 02:28:44] ###### wait for import job to finish
[2024-01-16 02:48:14] ###### wait for import job to finish failed
select id, step, status from mysql.tidb_import_jobs where start_time >= '2024-01-16 02:28:44'
jobId: 150003, step: importing, status: failed

tidb logs：
[2024/01/16 02:48:06.486 +08:00] [ERROR] [reader.go:49] ["read all data failed"] [takeTime=2m27.255433123s] [error="failed to read file 150003/90036/data/b0dba35c-6041-47e1-af29-cc98ec29d765/40: read 316276121654947278 bytes from external storage, exceed max limit 1073741824"]

[tidb-0.tar.gz](https://github.com/pingcap/tidb/files/13945076/tidb-0.tar.gz)
[tidb-1.tar.gz](https://github.com/pingcap/tidb/files/13945078/tidb-1.tar.gz)

### 4. What is your TiDB version? (Required)
./tidb-server -V
 Release Version: v7.6.0
Edition: Community
Git Commit Hash: 42960c3352d165810c79852b85ada448a68efa4c
Git Branch: heads/refs/tags/v7.6.0
UTC Build Time: 2024-01-15 11:33:46
GoVersion: go1.21.5
Race Enabled: false
Check Table Before Drop: false
Store: unistore
2024-01-16T02:28:41.835+0800

Lily2025:
another error for this scenario
[...]

PR #59752, opened 2025-02-25
prefetch: fix underlying unexpected-eof taken as EOF

D3Hunter:

Issue Number: ref #59495

Problem Summary:

the underlying reader might also return unexpected-eof, we need different between them, introduced in https://github.com/pingcap/tidb/pull/59496
- add `rangeSize` to `prefetch.Reader` to determine where `UnexpectedEOF` comes from
- check error equals using `goerrors.Is` instead of `==`. we might add stack to the error to help debug
  - Note: for the implementation of `io.Reader`, we still follow the API definition to return the `io.EOF` directly without stack, our `byteReader` doesn't follow the API of `io.Reader`, so it's ok to add stack to the error
  - And to unify code, we always use `goerrors.Is` to check `io.EOF` even for errors from `io.Reader`
- avoid retry when context cancel
### Check List

Tests 

- [x] Unit test
- [ ] Integration test

same test in https://github.com/pingcap/tidb/issues/50451, before it will fail with unexpected-eof, now success

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/storage/s3.go
+++ b/br/pkg/storage/s3.go
@@ -386,7 +386,7 @@ func NewS3Storage(ctx context.Context, backend *backuppb.S3, opts *ExternalStora
 		}
 	}
 
-	s3CliConfigs := []*aws.Config{}
+	s3CliConfigs := []*aws.Config{aws.NewConfig().WithLogLevel(aws.LogDebugWithRequestRetries)}
 	// if role ARN and external ID are provided, try to get the credential using this way
 	if len(qs.RoleArn) > 0 {
 		creds := stscreds.NewCredentials(ses, qs.RoleArn, func(p *stscreds.AssumeRoleProvider) {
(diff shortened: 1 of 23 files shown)


pkg/lightning/backend/external/concurrent_reader.go
```diff
@@ -87,13 +86,7 @@ func (r *concurrentFileReader) read(bufs [][]byte) ([][]byte, error) {
 				buf,
 			)
 			if err != nil {
-				log.FromContext(r.ctx).Error(
```
D3Hunter, line 90: it prints too many logs when context cancel, it's not needed, there is no further usefully info in it.

br/pkg/storage/s3.go
```diff
@@ -977,7 +983,7 @@ func (r *s3ObjectReader) Read(p []byte) (n int, err error) {
 	n, err = r.reader.Read(p[:maxCnt])
 	// TODO: maybe we should use !errors.Is(err, io.EOF) here to avoid error lint, but currently, pingcap/errors
 	// doesn't implement this method yet.
-	for err != nil && errors.Cause(err) != io.EOF && retryCnt < maxErrorRetries { //nolint:errorlint
+	for err != nil && errors.Cause(err) != io.EOF && r.ctx.Err() == nil && retryCnt < maxErrorRetries { //nolint:errorlint
```
D3Hunter, line 986: no need to retry on context cancel

lance6716 (approved):

tangenta (approved):

PR #59934, opened 2025-03-06
domain: batch handle the privilege reload event

### LINKED ISSUE #59400: The load privilege event is laggy after the workload gone
## Bug Report

I create 2M users, and the keep 10% of them active, then run 

"set password for test = 'xxx'" for all the 2M users

then run

"alter user test%d failed_login_attempts 10" for all the 2M users

... 

After the workload gone, I expect tidb server to be idle.

The CPU is 100%, using up one core handling privilege reload event.

Image

Image

Image

As we can see, tidb is still receiving privilege reload event from etcd, for a long time after the workload finish ...
And it's using 1 cpu core to handle the privilege reload loop.

master?  I modify some code for testing but the changes are unrelated to this issue.

```
commit 27365b47b27820152faf79395f6cf1b96eb3031f (HEAD -> master, origin/master, origin/HEAD)
Author: xzhangxian1008 <xzhangxian@foxmail.com>
Date:   Tue Feb 11 15:35:39 2025 +0800

    executor: fix the incorrect return when hash join encounters error (#59381)
    
    close pingcap/tidb#59377
```

tiancaiamao:

Issue Number: close #59400

Problem Summary:

After a `NotifyPrivilegeUpdate()`, the notification from etcd is laggy.

When we're in a batch of privilege modification operations, the laggy event cause that even when the workload is done,
tidb still have a long period of time handling the laggy events. 
This use 100% cpu (1 core) even there are no other events.

Wait for a while to get a batch of the laggy events, handle them together.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

Before:
image

After:
image

As you can see, after the workload finish, the CPU drop to 0 immediately with this commit.

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/domain/domain.go
+++ b/pkg/domain/domain.go
@@ -1944,6 +1944,42 @@ func decodePrivilegeEvent(resp clientv3.WatchResponse) PrivilegeEvent {
 	return msg
 }
 
+func ignoreBatchData(ch clientv3.WatchChan) {
+	ticker := time.NewTicker(5 * time.Millisecond)
+	defer ticker.Stop()
+	for i := 0; i < 1024; i++ {
+		select {
+		case <-ch:
+		case <-ticker.C:
+			return
+		}
+	}
+}
+
+func batchReadMoreData(ch clientv3.WatchChan, event PrivilegeEvent) PrivilegeEvent {
+	ticker := time.NewTicker(5 * time.Millisecond)
+	defer ticker.Stop()
+	for i := 0; i < 128; i++ {
+		select {
+		case resp, ok := <-ch:
+			if !ok {
+				return event
+			}
+			tmp := decodePrivilegeEvent(resp)
+			if tmp.All {
+				event.All = true
+			} else {
+				if !event.All {
+					event.UserList = append(event.UserList, tmp.UserList...)
+				}
+			}
+		case <-ticker.C:
+			return event
+		}
+	}
+	return event
+}
+
 // LoadPrivilegeLoop create a goroutine loads privilege tables in a loop, it
 // should be called only once in BootstrapSession.
 func (do *Domain) LoadPrivilegeLoop(sctx sessionctx.Context) error {
@@ -1978,6 +2014,7 @@ func (do *Domain) LoadPrivilegeLoop(sctx sessionctx.Context) error {
 				if ok {
 					count = 0
 					event = decodePrivilegeEvent(resp)
+					event = batchReadMoreData(watchCh, event)
 				} else {
 					if do.ctx.Err() == nil {
 						logutil.BgLogger().Error("load privilege loop watch channel closed")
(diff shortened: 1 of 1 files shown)


pkg/domain/domain.go
```diff
@@ -1944,6 +1944,42 @@ func decodePrivilegeEvent(resp clientv3.WatchResponse) PrivilegeEvent {
 	return msg
 }
 
+func ignoreBatchData(ch clientv3.WatchChan) {
+	ticker := time.NewTicker(5 * time.Millisecond)
+	defer ticker.Stop()
+	for i := 0; i < 1024; i++ {
```
Defined2014, line 1950: Does `1024` here means maxBatchSize? Maybe use a const is more reasonable.

pkg/domain/domain.go
```diff
@@ -1944,6 +1944,42 @@ func decodePrivilegeEvent(resp clientv3.WatchResponse) PrivilegeEvent {
...
+		case <-ticker.C:
+			return
+		}
+	}
+}
+
+func batchReadMoreData(ch clientv3.WatchChan, event PrivilegeEvent) PrivilegeEvent {
+	ticker := time.NewTicker(5 * time.Millisecond)
+	defer ticker.Stop()
+	for i := 0; i < 128; i++ {
```
Defined2014, line 1962: Why the batch size is different with previous one? Is there any reason for this?

tiancaiamao: When loadAll happen, we can ignore all the loadSomeUsers event, because after loadAll, load one specifiy user is a redundant operation.

But I still limit a 1024 because imagine this case: one privilege event come, 4.99ms later, another privilege event come ... then 4.99ms later, yet another privilege event come ...
If there is no limit, in this extreme case, the `loadAll` will never be called, because we're batching forever.

128 make the event more real-time, while 1024 discard as much events as possible but avoid the extreme cases.

Defined2014 (approved):

pkg/domain/domain.go
```diff
@@ -1944,6 +1944,35 @@ func decodePrivilegeEvent(resp clientv3.WatchResponse) PrivilegeEvent {
...
 
+func batchReadMoreData(ch clientv3.WatchChan, event PrivilegeEvent) PrivilegeEvent {
+	timer := time.NewTimer(5 * time.Millisecond)
+	defer timer.Stop()
+	const maxBatchSize = 128
+	for i := 0; i < maxBatchSize; i++ {
+		select {
+		case resp, ok := <-ch:
+			if !ok {
+				return event
```
wjhuang2016, line 1955: Is it ok to rerun the event here？
The later use may be wrong？err := privReloadEvent(do.privHandle, &event)

tiancaiamao: This function just focus on the batching functionality, and the first input event  is a correct one passed by the caller.
The close of the channel will be handled by the reloadPrivilegeLoad instaed of this function.

wjhuang2016 (approved):

PR #58629, opened 2024-12-31
sessionctx/stmtctx: avoid unlock of unlocked mutex panic on Statement…

chibiegg:
### What problem does this PR solve?

Issue Number: close #58600

Problem Summary:

### What changed and how does it work?

Similar to #39368 .

In `infoschema_reader.go` , I modified the code so that `RefCountOfStmtCtx` is referenced in places where StatementContext was being used.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)
Ref https://github.com/pingcap/tidb/issues/58600#issuecomment-2572344438, no panic anymore.
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

Issue #58600, opened 2024-12-29
fatal error: sync: unlock of unlocked mutex (v8.5.0 stmtctx.go:803)


chibiegg:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

I'm sorry, but I'm not sure how to reproduce the issue. It occurs sporadically on multiple running TiDB processes.

### 3. What did you see instead (Required)

fatal error on https://github.com/pingcap/tidb/blob/d13e52ed6e22cc5789bed7c64c861578cd2ed55b/pkg/sessionctx/stmtctx/stmtctx.go#L800-L804 .

These errors always occur on the same line.

```
fatal error: sync: unlock of unlocked mutex

goroutine 322590843 [running]:
sync.fatal({0x67ac522?, 0x0?})
	/usr/local/go/src/runtime/panic.go:1031 +0x18
sync.(*Mutex).unlockSlow(0xc0414d09c8, 0xffffffff)
	/usr/local/go/src/sync/mutex.go:231 +0x35
sync.(*Mutex).Unlock(...)
	/usr/local/go/src/sync/mutex.go:225
github.com/pingcap/tidb/pkg/sessionctx/stmtctx.(*StatementContext).AffectedRows(0xa6f0580?)
	/workspace/source/tidb/pkg/sessionctx/stmtctx/stmtctx.go:803 +0x9e
github.com/pingcap/tidb/pkg/util.(*ProcessInfo).ToRow(0xc040ae1860, 0xc003726ee0)
	/workspace/source/tidb/pkg/util/processinfo.go:152 +0x87
github.com/pingcap/tidb/pkg/executor.(*memtableRetriever).setDataForProcessList(0xc084d3a8c0, {0x7114e40, 0xc06e015208})
	/workspace/source/tidb/pkg/executor/infoschema_reader.go:1843 +0x25a
github.com/pingcap/tidb/pkg/executor.(*memtableRetriever).setDataForClusterProcessList(0xc084d3a8c0, {0x7114e40, 0xc06e015208})
	/workspace/source/tidb/pkg/executor/infoschema_reader.go:1816 +0x26
github.com/pingcap/tidb/pkg/executor.(*memtableRetriever).retrieve(0xc084d3a8c0, {0x707dc08, 0xc07adb7c20}, {0x7114e40, 0xc06e015208})
	/workspace/source/tidb/pkg/executor/infoschema_reader.go:179 +0xef3
[...]

PR #60004, opened 2025-03-11
utifl: increase ref count of `StmtCtx` before use it

Defined2014:

Issue Number: close #58629

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
Follow https://github.com/pingcap/tidb/issues/58600, test it on multiple running TiDB processes, no panic again.
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/util/processinfo.go
+++ b/pkg/util/processinfo.go
@@ -137,19 +137,17 @@ func (pi *ProcessInfo) txnStartTs(tz *time.Location) (txnStart string) {
 func (pi *ProcessInfo) ToRow(tz *time.Location) []any {
 	bytesConsumed := int64(0)
 	diskConsumed := int64(0)
-	if pi.StmtCtx != nil {
-		if pi.MemTracker != nil {
-			bytesConsumed = pi.MemTracker.BytesConsumed()
-		}
-		if pi.DiskTracker != nil {
-			diskConsumed = pi.DiskTracker.BytesConsumed()
-		}
-	}
-
 	var affectedRows any
 	var cpuUsages ppcpuusage.CPUUsages
-	if pi.StmtCtx != nil {
+	if pi.StmtCtx != nil && pi.RefCountOfStmtCtx != nil && pi.RefCountOfStmtCtx.TryIncrease() {
 		affectedRows = pi.StmtCtx.AffectedRows()
+		pi.RefCountOfStmtCtx.Decrease()
+	}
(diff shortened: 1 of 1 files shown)


pkg/util/processinfo.go
```diff
@@ -137,19 +137,17 @@ func (pi *ProcessInfo) txnStartTs(tz *time.Location) (txnStart string) {
...
-		}
-		if pi.DiskTracker != nil {
-			diskConsumed = pi.DiskTracker.BytesConsumed()
-		}
-	}
-
 	var affectedRows any
 	var cpuUsages ppcpuusage.CPUUsages
-	if pi.StmtCtx != nil {
+	if pi.StmtCtx != nil && pi.RefCountOfStmtCtx != nil && pi.RefCountOfStmtCtx.TryIncrease() {
```
Defined2014, line 142: I think the better way to do this is get the locks on 'sc.Reset()' function. But it's hard to achieve it, so I choose this way to fix the problem.

pkg/sessionctx/stmtctx/stmtctx.go
```diff
@@ -468,7 +511,26 @@ func NewStmtCtxWithTimeZone(tz *time.Location) *StatementContext {
 }
 
 // Reset resets a statement context
-func (sc *StatementContext) Reset() {
+func (sc *StatementContext) Reset() bool {
+	// make sure no other goroutines still get locks.
+	if sc.mu != nil {
+		if !sc.mu.TryLock() {
+			return false
```
tiancaiamao, line 518: In this case, it always means a bug: someone is still referencing this statement context, but we try to reset.

If we could solve why that happen, that's much better.  But it's too difficult to figure out why.
So this is a workaround to get rid of the fatal error ... panic.

tiancaiamao (approved):

pkg/sessionctx/stmtctx/stmtctx.go
Defined2014: We can know that this situation has occurred in a certain place, but it is difficult to avoid all situations. `TryLock` here could avoid panic again in any other situations.

xhebox (approved):

yudongusa (approved):

PR #60022, opened 2025-03-11
ddl: DROP PARTITION sees a row from a dropped partition in StateDeleteOnly with Global Index

### LINKED ISSUE #60021: During StateDeleteOnly in DROP PARTITION, old row in Global Index could still be seen
## Bug Report

See this [test line](https://github.com/pingcap/tidb/blob/21f9bb7b7a7478d825c4dee4fbf2ce913e3e7cf1/pkg/ddl/tests/partition/multi_domain_test.go#L260), tkNO should not see this entry!

Test fail, since the row should not be visible (see the [select above](https://github.com/pingcap/tidb/blob/21f9bb7b7a7478d825c4dee4fbf2ce913e3e7cf1/pkg/ddl/tests/partition/multi_domain_test.go#L253)).

Test passes

current master 21f9bb7b7a7478d825c4dee4fbf2ce913e3e7cf1

mjonss:

Issue Number: close #60021 

Problem Summary:
The dropped partition was not filtered out when reading from the global index during StateWriteOnly, resulting in a rows from the dropped partition could be seen if the Global Index was used, but not if other access paths was used during that DDL State.

Filtering out the dropped partition also for StateWriteOnly

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
During DROP PARTITION StateWriteOnly one could still access rows from the dropped partition through the Global Index.
```

### DIFF
--- a/pkg/meta/model/table.go
+++ b/pkg/meta/model/table.go
@@ -1065,7 +1065,7 @@ func (pi *PartitionInfo) IDsInDDLToIgnore() []int64 {
 			return ids
 		}
 	case ActionDropTablePartition:
-		if len(pi.DroppingDefinitions) > 0 && (pi.DDLState == StateDeleteOnly || pi.DDLState == StateWriteOnly) {
+		if len(pi.DroppingDefinitions) > 0 {
 			ids := make([]int64, 0, len(pi.DroppingDefinitions))
 			for _, def := range pi.DroppingDefinitions {
 				ids = append(ids, def.ID)
--- a/pkg/meta/model/table.go
+++ b/pkg/meta/model/table.go
@@ -1065,7 +1065,7 @@ func (pi *PartitionInfo) IDsInDDLToIgnore() []int64 {
 			return ids
 		}
 	case ActionDropTablePartition:
-		if len(pi.DroppingDefinitions) > 0 && pi.DDLState == StateDeleteOnly {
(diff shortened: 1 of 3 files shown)


Defined2014 (approved):

pkg/meta/model/table.go
```diff
@@ -1065,7 +1065,7 @@ func (pi *PartitionInfo) IDsInDDLToIgnore() []int64 {
 			return ids
 		}
 	case ActionDropTablePartition:
-		if len(pi.DroppingDefinitions) > 0 && pi.DDLState == StateDeleteOnly {
+		if len(pi.DroppingDefinitions) > 0 && (pi.DDLState == StateDeleteOnly || pi.DDLState == StateWriteOnly) {
```
lance6716, line 1068: Do you think it's more clear to change the condition to `pi.DDLState != StatePublic`, or even remove this state filter? IMO Any partition state that started dropping should be ignored.

mjonss: For `pi.DDLAction == ActionDropTablePartition` the only `pi.DDLState`'s that are used are `StateWriteOnly`, `StateDeleteOnly` and `StateDeleteReorganization`, and for `StateDeleteReorganization`, `pi.DroppingDefinitions` is nil, so you are right, we can remove the check for `pi.DDLState`. Thanks!

lance6716 (approved):

PR #60048, opened 2025-03-13
table: use writable columns to avoid columns not match when add column DDL exec parallel with update DML

### LINKED ISSUE #60047: executor: handleAssertionFailure failed when add column DDL exec parallel with update DML
## Bug Report

```golang

func TestAddColumnAfter(t *testing.T) {
	tk.MustExec("use test")
	tk.MustExec("drop table if exists t")
	tk.MustExec(`CREATE TABLE t (
		a INT,
		b INT,
		c VARCHAR(10),
		unique key idx(a, c)
	) partition by range columns(c) (
	partition p0 values less than ('30'),
	partition p1 values less than ('60'),
	partition p2 values less than ('90'));`)

	// initialize the data.
	for i := 0; i < 90; i++ {
		tk.MustExec("insert into t values (?, ?, ?)", i, i, i)
	}

	// parallel execute `insert ... on duplicate key update` and `alter table ... add column after ...`
	exitCh := make(chan struct{})
	var err error
	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		defer wg.Done()

		exitLoop := false
		ticker := time.After(10 * time.Second)
		for {
			tk1 := testkit.NewTestKit(t, store)
			tk1.MustExec("use test")
			val1 := 30 + rand.Intn(60)
			insertSQL := fmt.Sprintf("insert into t(a, b, c) values(%v, %v, %v) on duplicate key update a=values(a), b=values(b), c=values(c)",
				val1, rand.Intn(90), strconv.FormatInt(int64(val1), 10))
			err = tk1.ExecToErr(insertSQL)
			if err != nil {
				close(exitCh)
				break
			}
			select {
			case <-ticker:
				exitLoop = true
			default:
			}
			if exitLoop {
				break
			}
		}
	}()

	colNamesCh := make(chan string, 16)
	wg.Add(1)
	go func() {
		defer wg.Done()

		tk2.MustExec("use test")
		exitLoop := false
		ticker := time.After(10 * time.Second)
		for {
			select {
			case <-ticker:
				exitLoop = true
			case <-exitCh:
				exitLoop = true
			default:
				colName := randomString(32)
				ddlSQL := fmt.Sprintf("alter table t add column `%s` decimal(20,4) not null default '0' after a", colName)
				tk2.MustExec(ddlSQL)
				colNamesCh <- colName
			}
			if exitLoop {
				break
			}
		}
	}()

	wg.Add(1)
	go func() {
		defer wg.Done()

		tk3 := testkit.NewTestKit(t, store)
		tk3.MustExec("use test")
		exitLoop := false
		ticker := time.After(10 * time.Second)
		for {
			select {
			case <-ticker:
				exitLoop = true
			case <-exitCh:
				exitLoop = true
			default:
				colName := <-colNamesCh
				ddlSQL := fmt.Sprintf("alter table t drop column `%s`", colName)
				tk3.MustExec(ddlSQL)
			}
			if exitLoop {
				break
			}
		}
	}()
	wg.Wait()

	require.NoError(t, err)
}
```

test run success and pass.

Received unexpected error:
                                [tikv:8141]assertion failed: key: 74800000000000006f5f72800000000000003e, assertion: Exist, start_ts: 456614814054875137, existing start ts: 0, existing commit ts: 0

| Release Version: v9.0.0-alpha-206-g430fa8fc6a-dirty
Edition: Community
Git Commit Hash: 430fa8fc6a34e442cf84c0f3ca9c52fb61ff2bef
Git Branch: master
UTC Build Time: 2025-03-13 06:25:05
GoVersion: go1.23.6
Race Enabled: false
Store: unistore |

L-maple:

Issue Number: close #60047

Problem Summary:

When execute `evalBuffer.SetDatums(r...) ` in `locateRangeColumnPartition`，a non-public column may overlap with handle column, which causes locating to a wrong partition.

* How to judge?
1）On the Update Path: partitionedTableUpdateRecord -> locatePartition -> locateRangeColumnPartition, currData and newData both contain writable Columns。
That is to say，if `add column` DDL reach StateWriteOnly phase, updated currData and newData both contain the new added column。
2）While in `locateRangeColumnPartition` method，we get MutRow from t.evalBufferPool, whose Columns are all public, and actually extraHandle column is added at the end(see `initEvalBufferType`)。

* Give an example?
A table like the following:
```
CREATE TABLE t (
		a INT, b INT, c VARCHAR(10),
		unique key idx(a, c)
) partition by range columns(c) (
	partition p0 values less than ('30'),
	partition p1 values less than ('60'),
	partition p2 values less than ('90'));
```
When DDL `add column d decimal(20,4) not null default '0'` reaches StateWriteOnly phase：
* DML `insert...on duplicate key update...` updates a row from (20, 20, '20') to (20, 15, '20'). Then we can get currData as (20, 20, '20', 0), and newData as (20, 15, '20', 0). Both of them contains the non-public column, at the same time columns in evalBuffer are still (a, b, c)。
`evalBuffer.SetDatum` in locateRangeColumnPartition method set columns (a, b, c, handle) as (20, 20, '20', 0)。
Notice, handle is int64 type，but the written column d is decimal type. So they have different types and SetDatum sets the wrong value:
```
func (mr MutRow) SetDatum(colIdx int, d types.Datum) {
	switch d.Kind() {
    case types.KindMysqlDecimal:
    	*(*types.MyDecimal)(unsafe.Pointer(&col.data[0])) = *d.GetMysqlDecimal()
  }
}
```
* When debugging，I found t.c column's value is overlapped by zero:
```
// before handle column is set
*{length: 1,
  nullBitmap: []uint8 len: 1, cap: 1, [1],
  offsets: []int64 len: 2, cap: 2, [0,2],
  data: []uint8 len: 2, cap: 3, [51,49],
  elemBuf: []uint8 len: 0, cap: 0, nil,},
// after handle column is set
*{length: 1,
  nullBitmap: []uint8 len: 1, cap: 1, [0],
  offsets: []int64 len: 2, cap: 2, [0,0],
  data: []uint8 len: 2, cap: 3, [0,0],
  elemBuf: []uint8 len: 0, cap: 0, nil,},
```
t.c column's nullBitmap becomes 0，so EvalString get isNull=true. So the first partition is visited.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/db_test.go
+++ b/pkg/ddl/db_test.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"math/rand"
 	"slices"
 	"strconv"
 	"strings"
@@ -502,6 +503,48 @@ func TestShowCountWarningsOrErrors(t *testing.T) {
 	// Error: Table exist
 	_, _ = tk.Exec("create table show_errors (a int)")
 	tk.MustQuery("show count(*) errors").Check(tk.MustQuery("select @@session.error_count").Rows())
+}
+
+func TestIssue60047(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("drop table if exists t")
+	tk.MustExec(`CREATE TABLE t (
+		a INT,
+		b INT,
+		c VARCHAR(10),
+		unique key idx(a, c)
+	) partition by range columns(c) (
+	partition p0 values less than ('30'),
+	partition p1 values less than ('60'),
+	partition p2 values less than ('90'));`)
+
+	// initialize the data.
+	for i := 0; i < 90; i++ {
+		tk.MustExec("insert into t values (?, ?, ?)", i, i, i)
+	}
+
+	// parallel execute `insert ... on duplicate key update` and `alter table ... add column after ...`
+	var err error
+	hookFunc := func(job *model.Job) {
+		if job.SchemaState == model.StateWriteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+			val := 30 + rand.Intn(60)
+			insertSQL := fmt.Sprintf("insert into t(a, b, c) values(%v, %v, %v) on duplicate key update a=values(a), b=values(b), c=values(c)",
+				val, rand.Intn(90), strconv.FormatInt(int64(val), 10))
+			err = tk1.ExecToErr(insertSQL)
+		}
+	}
+	testfailpoint.EnableCall(t, "github.com/pingcap/tidb/pkg/ddl/beforeRunOneJobStep", hookFunc)
+
+	tk2 := testkit.NewTestKit(t, store)
+	tk2.MustExec("use test")
+	ddlSQL := "alter table t add column `d` decimal(20,4) not null default '0'"
+	tk2.MustExec(ddlSQL)
+
+	require.NoError(t, err)
 }
 
 // Close issue #24172.
--- a/pkg/table/tables/partition.go
+++ b/pkg/table/tables/partition.go
@@ -367,13 +367,13 @@ func (kp *ForKeyPruning) LocateKeyPartition(numParts uint64, r []types.Datum) (i
 
 func initEvalBufferType(t *partitionedTable) {
 	hasExtraHandle := false
-	numCols := len(t.Cols())
+	numCols := len(t.WritableCols())
 	if !t.Meta().PKIsHandle {
 		hasExtraHandle = true
 		numCols++
 	}
 	t.evalBufferTypes = make([]*types.FieldType, numCols)
-	for i, col := range t.Cols() {
+	for i, col := range t.WritableCols() {
 		t.evalBufferTypes[i] = &col.FieldType
 	}
 


L-maple:
@Defined2014 I find a critical bug，could you help me review the PR please?

pkg/ddl/db_test.go
```diff
@@ -504,6 +505,127 @@ func TestShowCountWarningsOrErrors(t *testing.T) {
...
+	go func() {
+		defer wg.Done()
+
+		exitLoop := false
+		ticker := time.After(10 * time.Second)
+		for {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+			val1 := 30 + rand.Intn(60)
+			insertSQL := fmt.Sprintf("insert into t(a, b, c) values(%v, %v, %v) on duplicate key update a=values(a), b=values(b), c=values(c)",
```
Defined2014, line 542: Could use failpoint `beforeRunOneJobStep` and `afterRunOneJobStep` to simplify this test case.

L-maple: OK

pkg/ddl/db_test.go
```diff
@@ -504,6 +505,63 @@ func TestShowCountWarningsOrErrors(t *testing.T) {
...
+
+	// initialize the data.
+	for i := 0; i < 90; i++ {
+		tk.MustExec("insert into t values (?, ?, ?)", i, i, i)
+	}
+
+	// parallel execute `insert ... on duplicate key update` and `alter table ... add column after ...`
+	var err error
+	hookFunc := func(job *model.Job) {
+		if job.Type == model.ActionAddColumn && job.State == model.JobStateRunning && job.SchemaState == model.StateWriteOnly {
```
Defined2014, line 531: `job.SchemaState == model.StateWriteOnly` is enough

pkg/ddl/db_test.go
```diff
@@ -504,6 +505,63 @@ func TestShowCountWarningsOrErrors(t *testing.T) {
...
+			err = tk1.ExecToErr(insertSQL)
+		}
+	}
+	testfailpoint.EnableCall(t, "github.com/pingcap/tidb/pkg/ddl/beforeRunOneJobStep", hookFunc)
+
+	tk2 := testkit.NewTestKit(t, store)
+	tk2.MustExec("use test")
+	colName := randomString(32)
+	ddlSQL := fmt.Sprintf("alter table t add column `%s` decimal(20,4) not null default '0'", colName)
+	if rand.Intn(2) > 0 {
```
Defined2014, line 546: If the rand test fails, ci will ignore it. Because ci will retry three times, it will only report an error if all tests fail.

pkg/ddl/db_test.go
L-maple: OK

pkg/ddl/db_test.go
L-maple: OK, this rand is not important, so I remove it.

pkg/ddl/db_test.go
```diff
@@ -504,6 +505,60 @@ func TestShowCountWarningsOrErrors(t *testing.T) {
...
+		tk.MustExec("insert into t values (?, ?, ?)", i, i, i)
+	}
+
+	// parallel execute `insert ... on duplicate key update` and `alter table ... add column after ...`
+	var err error
+	hookFunc := func(job *model.Job) {
+		if job.SchemaState == model.StateWriteOnly {
+			tk1 := testkit.NewTestKit(t, store)
+			tk1.MustExec("use test")
+			val := 30 + rand.Intn(60)
```
Defined2014, line 534: Does this test guarantee that the bug is triggered? `writeOnly` should only be triggered once, so there will be no dup rows.

L-maple: @Defined2014 Data have been initialized above, so I think dup rows always exist?

Defined2014 (approved):

pkg/ddl/db_test.go
```diff
@@ -504,6 +505,60 @@ func TestShowCountWarningsOrErrors(t *testing.T) {
...
+	colName := randomString(32)
+	ddlSQL := fmt.Sprintf("alter table t add column `%s` decimal(20,4) not null default '0'", colName)
+	tk2.MustExec(ddlSQL)
+
+	require.NoError(t, err)
+}
+
+func randomString(length int) string {
+	letters := []rune("abcdefghijklmnopqrstuvwxyz")
+	rand.Seed(time.Now().UnixNano())
```
Defined2014, line 553: It's no need after golang 1.20
image

Defined2014: > Seed uses the provided seed value to initialize the default Source to a deterministic state. Seed values that have the same remainder when divided by 2³¹-1 generate the same pseudo-random sequence. Seed, unlike the [Rand.Seed] method, is safe for concurrent use.

> If Seed is not called, the generator is seeded randomly at program startup.

> Prior to Go 1.20, the generator was seeded like Seed(1) at program startup. To force the old behavior, call Seed(1) at program startup. Alternately, set GODEBUG=randautoseed=0 in the environment before making any calls to functions in this package.

> Deprecated: As of Go 1.20 there is no reason to call Seed with a random value. Programs that call Seed with a known value to get a specific sequence of results should use New(NewSource(seed)) to obtain a local random generator.
> As of Go 1.24 [Seed] is a no-op. To restore the previous behavior set GODEBUG=randseednop=0.

L-maple: ok，I will remove it.

L-maple: resolved

L-maple:
@mjonss Hi teacher，I find a major bug. Could you help me review the PR please?

lance6716:
@L-maple not sure if CI failure is caused by your change. Let rerun CI and check again

mjonss (approved):
LGTM, thank you for your help finding and fixing this!

tangenta (approved):

Issue #55043, opened 2024-07-30
Statistics Tech Debt


hi-rustin:
## Tech Debt

In this issue, I will record all tech debts I found in the TiDB statistics module.

## Code duplication

The worst and most serious technical debt in the statistics module is code redundancy, and the following implementation pattern is used almost everywhere there is concurrent processing:
1. Keep single-threaded code untouched.
2. Directly implement another concurrent version of the exact same code.

This pattern creates a lot of problems, we have a few issues that need to be fixed twice, and code redundancy is serious. Theoretically, we just need to think of single-threaded as a special case of a multi-threaded implementation, and we shouldn't be copying and pasting code.

Here are the relevant modules that have this problem:
1. Concurrency analysis of partitioned tables
2. Result Stored Procedures for Partitioned Tables
3. Init Stats 
4. Merge Global Stats
and so on...

hi-rustin:
# NCRMTA1: Surprise analyze-partition-concurrency-quota
> Abbreviation: NCRMTA -> Nobody can really master TiDB analyze

The whole document is about how we can improve in the future. So please do not take it as criticism, even though I call it that, nobody can really master TiDB analyze. 

## Background 
In this PR, https://github.com/pingcap/tidb/pull/55046, I aimed to delete some duplicated code related to saving stats results. 
However, we were surprised to discover a new configuration called `analyze-partition-concurrency-quota`. There is no documentation or explanation for this configuration, but it does impact how we store stats results. In this document, I will explain why this configuration exists, how it works, and how to address it.

## Concurrently saving analysis results
When analyzing partitioned tables, it is important to concurrently analyze their partitions instead of analyzing them one by one.
The analysis process dataflow looks like this:
![image](https://github.com/user-attachments/assets/c2133fc6-a86d-4fce-8a1a-ab91a52376e5)

As you can see, we will spawn some workers to analyze these partitions concurrently,  then we will collect all analysis results in the result handler.
In the result handler, we also need to save the statistics concurrently; otherwise, all calculated statistics will heap in memory and queue for writing.

![image](https://github.com/user-attachments/assets/b37c8ec7-813e-4285-ad49-1d6e70007059)

As you can see, we spawn some saving workers to save the result table by table.  This is where the `analyze-partition-concurrency-quota` comes into play.

[...]

PR #60090, opened 2025-03-14
statistics: enhance concurrency handling in stats initialization to reduce duplication

Rustin170506:

Issue Number: ref https://github.com/pingcap/tidb/issues/55043

Problem Summary:

We shouldn’t simply duplicate the same code for both concurrent and non-concurrent init stats.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] [Manual test](https://github.com/pingcap/tidb/pull/60090#issuecomment-2731649241)
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/handle/bootstrap.go
+++ b/pkg/statistics/handle/bootstrap.go
@@ -600,12 +600,12 @@ func (h *Handle) initStatsBuckets(cache statstypes.StatsCache, totalMemory uint6
 		return nil
 	}
 	if config.GetGlobalConfig().Performance.ConcurrentlyInitStats {
-		err := h.initStatsBucketsConcurrenctly(cache, totalMemory, initstats.GetConcurrency())
+		err := h.initStatsBucketsConcurrently(cache, totalMemory, initstats.GetConcurrency())
 		if err != nil {
 			return errors.Trace(err)
 		}
 	} else {
-		err := h.initStatsBucketsConcurrenctly(cache, totalMemory, 1)
+		err := h.initStatsBucketsConcurrently(cache, totalMemory, 1)
 		if err != nil {
 			return errors.Trace(err)
 		}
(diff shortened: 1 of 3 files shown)


Rustin170506 (commented):

Rustin170506:
> Currently, concurrent init stats are for specific users. This feature can be removed in the future, especially since sync load is enough for users.

We should definitely keep this feature, as BR will soon rely on this behavior. Unless we completely remove non-lite init stats, we still need to maintain this functionality, right? If we eventually decide to remove non-lite init stats, then all these codes would be removed anyway. So in theory, removing the duplicate code here shouldn’t cause any issues?

Rustin170506:
Tested locally(100k tables)

Before change:
```log
2025-03-18 12:44:09 (UTC+08:00)TiDB tc-tidb-1.tc-tidb-peer.forge85-scenario-tps-7748182-1-995.svc:4000[domain.go:2494] ["init stats info time"] [lite=false] ["take time"=3m7.798889455s]
```

After change:

```log
2025-03-18 12:58:09 (UTC+08:00)TiDB tc-tidb-0.tc-tidb-peer.forge85-scenario-tps-7748182-1-995.svc:4000[domain.go:2644] ["init stats info time"] [lite=false] ["take time"=3m22.618695301s]
```

No huge difference here. So I think it is safe to make this change.

Rustin170506:
/unhold

I understand that regardless of whether we eventually decide to remove concurrently-init-stats or non-lite init stats, this PR shouldn’t have much impact.

In theory, we wouldn’t remove concurrently-init-stats alone, because without it, non-lite init stats would be almost unusable in scenarios with a large number of tables.

If we decide to remove non-lite init stats, then concurrently-init-stats would also be removed, meaning all related code would be deleted.

Removing a feature or parameter typically takes at least two years. I feel that simplifying redundant code during this period to reduce maintenance complexity would still be beneficial.

time-and-fate (approved):

Issue #60099, opened 2025-03-17
import into: base64 encoded NULL values can't be imported into barbinary(1) column


fzzf678:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
```shell
(base) ➜  ~ cat base64null.csv
XE4=
```


```sql
create table t(a varbinary(1));

import into t from "/Users/fanzhou/base64null.csv" with fields_encoded_by='base64', fields_enclosed_by='', fields_escaped_by='', character_set='binary';

mysql> show import jobs where job_id = 3\G;
*************************** 1. row ***************************
          Job_ID: 3
     Data_Source: /Users/fanzhou/base64null.csv
    Target_Table: `test`.`t`
        Table_ID: 110
           Phase: importing
          Status: awaiting-resolution
Source_File_Size: 5B
   Imported_Rows: 0
  Result_Message: [types:1406]Data Too Long, field len 1, data len 2
     Create_Time: 2025-03-17 09:45:55.113214
      Start_Time: 2025-03-17 09:45:55.624219
        End_Time: NULL
      Created_By: root@%
1 row in set (0.01 sec)
```

<!-- a step by step guide for reproducing the bug. -->

### 2. What did you expect to see? (Required)
Import into success
### 3. What did you see instead (Required)
Import into stuck
### 4. What is your TiDB version? (Required)
**Not master.**
https://github.com/pingcap/tidb/tree/feature/release-8.1-gsort-test
```
mysql> select tidb_version();
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| tidb_version()                                                                                                                                                                                                                                                                         |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Release Version: v8.1.2-42-g14ea5f3db54
Edition: Community
Git Commit Hash: 14ea5f3db54b197393b5efbb104ba5779ad21c96
Git Branch: feature/release-8.1-gsort-test
UTC Build Time: 2025-03-15 10:41:12
GoVersion: go1.23.6
Race Enabled: false
Check Table Before Drop: false
Store: tikv |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```
<!-- Paste the output of SELECT tidb_version() -->

PR #60102, opened 2025-03-17
importinto: fix null value for base64

GMHDBJD:

Issue Number: close https://github.com/pingcap/tidb/issues/60099

Problem Summary:

- fix null value for base64 by check `\N`
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/lightning/mydump/csv_parser.go
+++ b/pkg/lightning/mydump/csv_parser.go
@@ -215,6 +215,12 @@ func (parser *CSVParser) unescapeString(input field) (unescaped string, isNull b
 			return
 		}
 		unescaped = string(decoded)
+		// we set escapedBy to empty when use base64 encoding
+		// also parser.escFlavor is not set when escapedBy is empty
+		// so directly check the unescaped is \N or not
+		if unescaped == `\N` {
+			return unescaped, true, nil
+		}
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/lightning/mydump/csv_parser.go
```diff
@@ -215,6 +215,12 @@ func (parser *CSVParser) unescapeString(input field) (unescaped string, isNull b
 			return
 		}
 		unescaped = string(decoded)
+		// we set escapedBy to empty when use base64 encoding
+		// also parser.escFlavor is not set when escapedBy is empty
+		// so directly check the unescaped is \N or not
+		if unescaped == `\N` {
```
lance6716, line 221: it's better to check `isNull` as the normal code path

```
isNull = !parser.cfg.NotNull &&
			slices.Contains(parser.cfg.Null, unescaped)
```

GMHDBJD: since fields are already decoded and generally not quoted, I don't think similar quoting‐related logic should be apply.

lance6716: it's not about quoted. User may want to have a `\N` liternal value to be inserted into table (so need to check `cfg.NotNull`), and may choose another value to represent NULL (so need to check cfg.Null)

D3Hunter:
empty string is taken as null now see https://github.com/pingcap/tidb/pull/60409, we can close this one

Issue #60115, opened 2025-03-17
Introduce a enhancement session pool to manage the system session


lcwangchao:
## Enhancement

RFC: #60120

lance6716:
I found that the internal session's `User` field may be modified, and seems there's no resetting of it. Like https://github.com/pingcap/tidb/blob/43da6d1457dc0550f2d0f97a7feabdec7045d2d2/pkg/executor/grant.go#L138-L139

Not sure if it will cause problems

Issue #56934, opened 2024-10-29
some memory leaks in TTL


lcwangchao:
## Bug Report

![image](https://github.com/user-attachments/assets/c0d95d27-53b1-4476-96db-a18dcc4940ef)

PR #60120, opened 2025-03-17
docs: RFC: Enhancement of Internal System Session Management

lcwangchao:

Issue Number: ref #60115

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


hawkingrei:
I don't know if the new version of the session pool will still use the implementation from github.com/ngaut/pools. However, I think this implementation is quite problematic and even less practical than the official pool. It's just that historically, there was a time when the Go language did not have   sync.Pool  , and implementing it ourselves had some advantages.

hawkingrei:
According to the convention of ```sync.Pool```, if your object is problematic, you just don't put it back into the Pool. It will naturally be cleaned up after the next GC.

docs/design/2025-03-17-sys-session-manage-enhancement.md
```diff
@@ -0,0 +1,167 @@
...
+            se.Destroy()    
+            panic(r)
+        }
+        p.put(se)
+    }()
+    return f(se)
+}
+```
+
+We can see that the `WithSession`n method always calls `p.put(se)` within a defer function to ensure the session is returned to the pool. One exception to this occurs when a panic happens. In such cases, the session is destroyed instead of being returned to the pool, as it may be in an undefined state, making it safer to discard rather than reuse.
```
djshow832, line 68:
```suggestion
We can see that the `WithSession` method always calls `p.put(se)` within a defer function to ensure the session is returned to the pool. One exception to this occurs when a panic happens. In such cases, the session is destroyed instead of being returned to the pool, as it may be in an undefined state, making it safer to discard rather than reuse.
```

lcwangchao:
> I don't know if the new version of the session pool will still use the implementation from github.com/ngaut/pools. However, I think this implementation is quite problematic and even less practical than the official pool. It's just that historically, there was a time when the Go language did not have sync.Pool , and implementing it ourselves had some advantages.

Thanks for the advise. I'll try to use the `sync.Pool` in the next pull request. This demo only demonstrates other features and use the original pool implementation.

docs/design/2025-03-17-sys-session-manage-enhancement.md
```diff
@@ -0,0 +1,167 @@
...
+    s.mu.Lock()
+    defer s.mu.Unlock()
+    if s.owner == nil || s.owner != owner {
+        return errors.New("invalid owner")
+    }
+    return nil
+}
+
+func (s *Session) ExecuteInternal(ctx context.Context, sql string, args ...any) (sqlexec.RecordSet, error) {
+    if err := s.internal.checkOwner(s); err != nil {
```
lance6716, line 146: instead of `checkOwner`, I think we can add a atomic bool `closed` field to `*Session`. When `*Session` is put back to the pool, store `true` to `closed`.

In other words, lifecycle management belongs to `Session`, not `session`

docs/design/2025-03-17-sys-session-manage-enhancement.md
```diff
@@ -0,0 +1,167 @@
...
+        return err
+    }
+    
+    defer func() {
+        if r := recover(); r != nil {
+            // Always destroy the session when panic to avoid undetermined state.
+            se.Destroy()    
+            panic(r)
+        }
+        p.put(se)
```
Rustin170506, line 62: I think if any error occurs, we should destroy it as well.

docs/design/2025-03-17-sys-session-manage-enhancement.md
```diff
@@ -0,0 +1,167 @@
...
+// session is private and can only be accessed by the `Pool` and `Session`.
+type session struct {
+    mu sync.Mutex
+    owner any
+    sctx sessionctx.Context
+    ...
+}
+```
+
+The `mu` protects the internal session from concurrent access, and the field `sctx` is the real session context that holds the session's state.
```
Rustin170506, line 129: I think we should discourage users from using the same session across multiple threads, as I’m concerned it could create a lot of issues. In the current session design, our expectation is that it shouldn’t be used this way.

Rustin170506: Or am I misunderstanding the purpose of this mutex?

docs/design/2025-03-17-sys-session-manage-enhancement.md
lcwangchao: I'm not sure whether it is too aggressive to regard all errors as undetermined state. For example, `insert into ...` may return an dup key error, sometimes it is a normal case for certain scenes...

docs/design/2025-03-17-sys-session-manage-enhancement.md
lcwangchao: The `mu` protects `owner` or other fields in `session` except `sctx`.

For example, if `Session.Destory` will set the `session.owner = nil`. If `Destory` is called in multiple threads, only one thread take effects and another should exit silently without any error.

docs/design/2025-03-17-sys-session-manage-enhancement.md
lcwangchao: I think it is equivalent to the current design. When `session.owner != *Session` indicates the Session is closed, and when `session.owner == nil` means the session has been destroyed. One extra benefit for the current design is that the meta fields are more "private" and it's hard to construct a "illegal" `Session`. For example:

```
s1, err := pool.Get()
if err != nil {
    // handle error ...
}

s2 := *s1

// ...
```

For the current design, `s2.ExecuteInternal(...)` will produce an error because the inner session's owner is not s2. However, if `Session` keeps a `closed` field in it, this field will be cloned and both two session can be used at the same time

lance6716: The difference I want to stress is `session` should not care about upper layer's management like ownership checking. The upper layer's `Session` should done this only in its fields, without modifying `session` to add extra complexity.

For the new case of copying `Session`, I think it can be avoided by some nocopy mechanism.

lcwangchao: It is reasonable as long as the current design's semantics are self-consistent, I think. I've tried the above two practices and found each of them has the same complexity (both of them need to handle the meta fields, lock for thread-safe, etc...)

In this design, I think the semantics is clear that: 1. The pool cached `session` instead of `sessionctx.Context` to have more information about the internal context. 2. Each `session` has one and only one `owner` that has the privilege to operate on it. 3.  `Session` is only a simple wrapper to proxy the access and expose only necessary methods to callers.

lance6716: > same complexity

The difference is, your proposal is a two-direction relationship. `Session` need to manage `session`, and `session` also need to check `Session`. I'm not sure if it's really necessary to let `session` access `Session` in checking owner or other actions.

lcwangchao: You can think of `session` as the real `Session` in your design but we used `Session` in current design to provide methods to outside coders.

* `session` is used by `Pool` and `Session`. The public methods of `session` can be used by them.
* `Session` is used by the outside caller.

docs/design/2025-03-17-sys-session-manage-enhancement.md
Rustin170506: Got it. Thank you!

docs/design/2025-03-17-sys-session-manage-enhancement.md
Rustin170506: @qw4990 What do you think?

docs/design/2025-03-17-sys-session-manage-enhancement.md
```diff
@@ -0,0 +1,167 @@
...
+}
+```
+
+## Tests
+
+We should do the below tests to make sure the new introduced session pool works as expected:
+
+- Unit tests.
+- Integration tests for specified scenarios:
+  - Test 1000+ TTL tables with small `tidb_ttl_job_internal` settings to test the frequent session creation and destruction in the TTL scene.
```
Rustin170506, line 159:
```suggestion
  - Test 1000+ TTL tables with small `tidb_ttl_job_internal` settings to test the frequent session creation and destruction in the TTL scene.
  - Test async load with corrupted statistics to ensure there are no memory leaks.
```

Rustin170506 (approved):
Thanks! :shipit:

Rustin170506 (approved):

docs/design/2025-03-17-sys-session-manage-enhancement.md
```diff
@@ -0,0 +1,173 @@
...
+func (p *Pool) put(s *Session) {
+    ...
+}
+
+func (p *Pool) WithSession(func(*Session) error) error {
+    ...
+}
+```
+
+The most important method of `Pool` is `WithSession` which is introduced to replace the `Get` and `Put` methods of the current session pool. The `WithSession` method accepts a function as an argument, which will be executed with a session retrieved from the pool. After the function completes, the session will be automatically returned to the pool. This design ensures that the session is always returned to the pool after use, preventing memory leaks.
```
lance6716, line 45: > This design ensures that the session is always returned to the pool after use

This can't be strictly ensured, because the function may store the provided session to elsewhere. We still need to use comment of `WithSession` to tell developer not to do it.

docs/design/2025-03-17-sys-session-manage-enhancement.md
```diff
@@ -0,0 +1,173 @@
...
+
+Let's see how `WithSession` works:
+
+```go
+func (p *Pool) WithSession(fn func(*Session) error) error {
+    se, err := p.Get()
```
lance6716, line 51:
```suggestion
    se, err := p.get()
```

docs/design/2025-03-17-sys-session-manage-enhancement.md
```diff
@@ -0,0 +1,173 @@
...
+    if err != nil {
+        return err
+    }
+
+    success := false
+    defer func() {
+        if success {
+            p.Put(se)
```
lance6716, line 59:
```suggestion
            p.put(se)
```

docs/design/2025-03-17-sys-session-manage-enhancement.md
```diff
@@ -0,0 +1,173 @@
...
+}
+```
+
+It provides a set of methods that proxy access to the internal session. When the caller invokes `WithSession`, a new `Session` instance is created and passed to the function. Once the function completes, the `Session` instance is destroyed, and only the internal session is returned to the pool.
+
+We have several reasons for proposing the above design:
+
+- The wrapped methods allow for additional checks to prevent misuse of the internal session. For example, if a caller attempts to invoke methods after the session has been returned to the pool, we can detect this and return an error.
+- The `Session` type can be extended with new methods, providing more convenient functionality for internal use.
+- The original session contains numerous methods, many of which may not be necessary for the caller. By filtering out unnecessary methods, we can expose only the required ones, simplifying the interface and improving usability.
```
lance6716, line 120: > may not be necessary for the caller

to achieve this goal, we can check and change if `session`'s methods are exposed. So it's not very accurate that it's done by `Session`

docs/design/2025-03-17-sys-session-manage-enhancement.md
```diff
@@ -0,0 +1,173 @@
...
+    mu sync.Mutex
+    owner any
+    sctx sessionctx.Context
+    ...
+}
+```
+
+The `mu` protects the internal session from concurrent access, and the field `sctx` is the real session context that holds the session's state. 
+
+Let's provide further clarification on the `owner` field. The "owner" refers to the entity that holds the internal session, and only the owner is permitted to use it. When an internal session is in the pool, the pool itself is considered the owner. When a caller retrieves the session from the pool, the wrapped `Session` becomes the new owner, which is then passed to the caller.
```
lance6716, line 136: So the type of `owner` is any because it may be the pool or `Session`? 

Will it be `nil` `owner`? If not, I think we can change the type to `*Session` and let nilness representing it's owned by pool

docs/design/2025-03-17-sys-session-manage-enhancement.md
lcwangchao: `WithSession` is designed to avoid forgetting to call `Close` of a session which will never be used again like issue: #56934 . Refering a `Session` in  `WithSession ` elsewhere will indeed cause some memory leak but I think it is not the main problem because:

1. As the `session` returned to the pool, the next `Get` will reuse it without creating a new one.
2. Usually, if `Session` is refered  somewhere by mistake, the amount should not be too much because most of them should be GC by go.

docs/design/2025-03-17-sys-session-manage-enhancement.md
lcwangchao: `session` is also used by `Pool`, and some methods are provided to `Pool` only. One way is to make these pool-only methods private, but I don't want to do it. I think the current design is more clear:
- `session` is provided to both `Session` and  `Pool`, it wraps the internal `sessionctx.Context` and provide a safe access. 
- `Pool` and `Session` should only used the public method of `session` to make a clear semantic.
- `Session` is only used to proxies the caller's access to `session`, expose necessary methods only.

lance6716 (approved):

Issue #60148, opened 2025-03-18
planner: SQL Plan Management - Quick Regression Fix


qw4990:
## Enhancement

### Feature Overview

1. A new sys-table to help identify regressed queries;
2. A new SQL statement `show plans for <SQL>` to show all historical plans of a specified query;

Below is an example of using these 2 new capabilities to fix regressions:
```
mysql> select * from regressed_query\G
*************************** 1. row ***************************
                   digest: 3f87b465219c7d0a8e13564b927d0a3f652c198e7d0a652f9c138e4b7d652c3f
                    query: SELECT * FROM t WHERE b in (...) AND c = ?
     current_plan_latency: 3131.2727
  prior_good_plan_latency: 791.8161
             current_plan: TableReader_7      root     0.03   data:Selection_6                       0      time:1.15ms, open:16.7µs, close:1.79µs, loops:1, cop_task: {num: 1, max: 1.11ms, proc_keys: 0, tot_proc: 56.7µs, tot_wait: 31.8µs, copr_cache_hit_ratio: 0.00, build_task_duration: 3.54µs, max_distsql_concurrency: 1}, rpc_info:{Cop:{num_rpc:1, total_time:1.1ms}}234 BytesN/A
                           └─Selection_6      cop[tikv]0.03   eq(test.t.c, 2), in(test.t.b, 1, 2, 3) 0      tikv_task:{time:0s, loops:1}, scan_detail: {total_keys: 1, get_snapshot_time: 19.8µs, rocksdb: {block: {cache_hit_count: 1}}}, time_detail: {total_process_time: 56.7µs, total_wait_time: 31.8µs, tikv_wall_time: 938.3µs}                                            N/A      N/A
                             └─TableFullScan_5cop[tikv]10000  table:t, keep order:false, stats:pseudo0      tikv_task:{time:0s, loops:1}
          prior_good_plan: IndexLookUp_12           root     0.03                                                                   0      time:460µs, open:6.96µs, close:1.63µs, loops:1                                                                                                                                                                                                                                                                                                                                                                                                                                              243 BytesN/A
[...]

PR #60238, opened 2025-03-24
planner: implement "SHOW PLAN FOR <SQL>" for SPM

qw4990:

Issue Number: ref #60148

Problem Summary: planner: implement "SHOW PLAN FOR <SQL>" for SPM

planner: implement "SHOW PLAN FOR <SQL>" for SPM

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/bindinfo/binding_auto.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+}
+
+// ShowPlansForSQL shows historical plans for a specific SQL.
+func (ba *bindingAuto) ShowPlansForSQL(currentDB, sqlOrDigest, charset, collation string) ([]*BindingPlanInfo, error) {
+	// parse and normalize sqlOrDigest
+	p := parser.New()
+	var normalizedSQL, whereCond string
+	stmtNode, err := p.ParseOneStmt(sqlOrDigest, charset, collation)
+	if err == nil {
```
time-and-fate, line 71: What if there are real errors? And also we need comments.

pkg/bindinfo/binding_auto.go
```diff
@@ -0,0 +1,150 @@
...
+		plan.ProcessedKeys += row.GetInt64(2)
+		plan.TotalTime += row.GetInt64(3)
+		plan.Plan = row.GetString(4)
+	}
+	return plan, nil
+}
+
+// planExecInfo represents the plan info from information_schema.tidb_statements_stats table.
+type planExecInfo struct {
+	// exec info
```
time-and-fate, line 144: The comments seem useless.

pkg/bindinfo/binding_auto.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// read plan info from information_schema.tidb_statements_stats
+	bindingPlans := make([]*BindingPlanInfo, 0, len(bindings))
+	for _, binding := range bindings {
+		pInfo, err := ba.getPlanExecInfo(binding.PlanDigest)
+		if err != nil {
+			logutil.BgLogger().Error("getStmtStatsByDigestInCluster", zap.String("plan_digest", binding.PlanDigest), zap.Error(err))
```
time-and-fate, line 92: Why don't we use `bindingLogger`?

pkg/bindinfo/binding_auto.go
```diff
@@ -0,0 +1,150 @@
...
+	var normalizedSQL, whereCond string
+	stmtNode, err := p.ParseOneStmt(sqlOrDigest, charset, collation)
+	if err == nil {
+		db := utilparser.GetDefaultDB(stmtNode, currentDB)
+		normalizedSQL, _ = NormalizeStmtForBinding(stmtNode, db, false)
+	}
+
+	// read binding data from mysql.bind_info
+	if normalizedSQL != "" {
+		whereCond = fmt.Sprintf("where original_sql='%s'", normalizedSQL)
```
time-and-fate, line 78: Please use the standard method to escape it since there could be arbitrary content in the SQL.

pkg/bindinfo/binding_auto.go
```diff
@@ -0,0 +1,150 @@
...
+		if pInfo != nil && pInfo.ExecCount > 0 {
+			autoBinding.Plan = pInfo.Plan
+			autoBinding.ExecTimes = pInfo.ExecCount
+			autoBinding.AvgLatency = float64(pInfo.TotalTime) / float64(pInfo.ExecCount)
+			autoBinding.AvgScanRows = float64(pInfo.ProcessedKeys) / float64(pInfo.ExecCount)
+			autoBinding.AvgReturnedRows = float64(pInfo.ResultRows) / float64(pInfo.ExecCount)
+			autoBinding.LatencyPerReturnRow = autoBinding.AvgLatency / autoBinding.AvgReturnedRows
+			autoBinding.ScanRowsPerReturnRow = autoBinding.AvgScanRows / autoBinding.AvgReturnedRows
+		}
+		bindingPlans = append(bindingPlans, autoBinding)
```
time-and-fate, line 105: If `pInfo` is `nil`, do we still need to do this?

pkg/bindinfo/binding_auto.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+		autoBinding := &BindingPlanInfo{Binding: binding}
+		if pInfo != nil && pInfo.ExecCount > 0 {
+			autoBinding.Plan = pInfo.Plan
+			autoBinding.ExecTimes = pInfo.ExecCount
+			autoBinding.AvgLatency = float64(pInfo.TotalTime) / float64(pInfo.ExecCount)
+			autoBinding.AvgScanRows = float64(pInfo.ProcessedKeys) / float64(pInfo.ExecCount)
+			autoBinding.AvgReturnedRows = float64(pInfo.ResultRows) / float64(pInfo.ExecCount)
+			autoBinding.LatencyPerReturnRow = autoBinding.AvgLatency / autoBinding.AvgReturnedRows
+			autoBinding.ScanRowsPerReturnRow = autoBinding.AvgScanRows / autoBinding.AvgReturnedRows
```
time-and-fate, line 103: Please check to avoid division by 0.

pkg/bindinfo/binding_auto.go
```diff
@@ -0,0 +1,150 @@
...
+	if err == nil {
+		db := utilparser.GetDefaultDB(stmtNode, currentDB)
+		normalizedSQL, _ = NormalizeStmtForBinding(stmtNode, db, false)
+	}
+
+	// read binding data from mysql.bind_info
+	if normalizedSQL != "" {
+		whereCond = fmt.Sprintf("where original_sql='%s'", normalizedSQL)
+	} else { // treat sqlOrDigest as a digest
+		whereCond = fmt.Sprintf("where sql_digest='%s'", sqlOrDigest)
```
henrybw, line 80: Is it possible to verify that a string is a valid digest? If we can, that would be a bit more robust than the current logic, which assumes that any input that fails to parse is a digest.

pkg/bindinfo/binding_auto.go
```diff
@@ -0,0 +1,150 @@
...
+
+	plan = new(planExecInfo)
+	for _, row := range rows {
+		// Sum them up if there are multiple records.
+		plan.ResultRows += row.GetInt64(0)
+		plan.ExecCount += row.GetInt64(1)
+		plan.ProcessedKeys += row.GetInt64(2)
+		plan.TotalTime += row.GetInt64(3)
+		plan.Plan = row.GetString(4)
+	}
```
henrybw, line 138: Rather than summing up the column values by hand here, maybe it would be better to instead use `SUM()` in the original query?

henrybw (commented):
Mostly looks good so far.

pkg/bindinfo/binding_auto.go
qw4990: I updated the algorithm to recognize whether it's a digest or a SQL, now it will return the error if parse fails.

pkg/bindinfo/binding_auto.go
qw4990: Yes, then in this case, we can still see the binding, but all plan execution info is empty. I added some comments here.

pkg/bindinfo/binding_auto.go
qw4990: Yes, I updated the logic here, now we'll recognize whether the `sqlOrDigest` is a digest or a SQL first.

pkg/bindinfo/binding_auto.go
qw4990: Good idea. I've updated it.

qw4990:
Thanks for your reviewing, all comments have been addressed, PTAL @time-and-fate @henrybw

pkg/bindinfo/binding_auto.go
```diff
@@ -0,0 +1,152 @@
...
+	return bindingPlans, nil
+}
+
+// getPlanExecInfo gets the plan execution info from information_schema.tidb_statements_stats table.
+func (ba *bindingAuto) getPlanExecInfo(planDigest string) (plan *planExecInfo, err error) {
+	if planDigest == "" {
+		return nil, nil
+	}
+	stmtQuery := fmt.Sprintf(`select cast(sum(result_rows) as signed), cast(sum(exec_count) as signed),
+       cast(sum(processed_keys) as signed), cast(sum(total_time) as signed), any_value(plan)
```
time-and-fate, line 121: I see `plan` is a NULLable column. I don't know when it could be `NULL`. But please note that if there is one `NULL`, the aggregation result will be `NULL`.

time-and-fate (approved):

pkg/bindinfo/binding_auto.go
henrybw: Odd. The fact that the column is nullable might be an oversight (that is, it probably should not be nullable). Both `tidb_statements_stats` and `statements_summary` implement this column using `plancodec.DecodePlan()`, but even if this function errors it still returns an empty string.

henrybw (approved):
Thanks! LGTM

tangenta (approved):

PR #60255, opened 2025-03-25
planner: change the semantic of `tidb_opt_ordering_index_selectivity_threshold` from `<` to `<=`

### LINKED ISSUE #60242: Planner: Need to ignore the ordering index path even if the index selectivity is 1.
## Bug Report

In some versions, we still need to set tidb_opt_ordering_index_selectivity_threshold to control whether we want to use the index order property to scan rows. 
However, when the index selectivity is equal to 1 and we set tidb_opt_ordering_index_selectivity_threshold=1, we still can't ignore the ordering index plan. This is because we always require [index_selectivity < selectivity_threshold](https://github.com/pingcap/tidb/blob/release-7.1-20240802-v7.1.5/planner/core/find_best_task.go#L2471)  to eliminate the ordering path. 
Actually, we just need to replace `<` with `<=`.

```
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                                                                                                                                     |
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| tttt1 | CREATE TABLE `tttt1` (
  `id1` int NOT NULL,
  `id2` int NOT NULL,
  `id3` varchar(20) NOT NULL,
  `id4` int NOT NULL,
  `id5` int NOT NULL,
  `id6` int NOT NULL,
  KEY `idx_id1` (`id3`,`id2`),
  KEY `idx_id2` (`id4`,`id2`),
  KEY `idx` (`id1`)
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.03 sec)
mysql> set session tidb_opt_ordering_index_selectivity_threshold=1;
```

```
mysql> explain select * from tttt1 where id6=1 order by id1 limit 10;
+------------------------------+------------+-----------+---------------+----------------------------------------------------+
| id                           | estRows    | task      | access object | operator info                                      |
+------------------------------+------------+-----------+---------------+----------------------------------------------------+
| TopN_8                       | 10.00      | root      |               | test.tttt1.id1, offset:0, count:10                 |
| └─TableReader_16             | 10.00      | root      |               | data:TopN_15                                       |
|   └─TopN_15                  | 10.00      | cop[tikv] |               | test.tttt1.id1, offset:0, count:10                 |
|     └─Selection_14           | 2097.15    | cop[tikv] |               | eq(test.tttt1.id6, 1)                              |
|       └─TableFullScan_13     | 2097152.00 | cop[tikv] | table:tttt1   | keep order:false, stats:partial[id6:unInitialized] |
+------------------------------+------------+-----------+---------------+----------------------------------------------------+
5 rows in set (0.00 sec)
```
```
mysql> explain select * from tttt1 where id6=1 order by id1 limit 10;
+-----------------------------------+----------+-----------+-----------------------------+------------------------------------------------------------------------------------------------+
| id                                | estRows  | task      | access object               | operator info                                                                                  |
+-----------------------------------+----------+-----------+-----------------------------+------------------------------------------------------------------------------------------------+
| Limit_12                          | 10.00    | root      |                             | offset:0, count:10                                                                             |
| └─Projection_21                   | 10.00    | root      |                             | test.tttt1.id1, test.tttt1.id2, test.tttt1.id3, test.tttt1.id4, test.tttt1.id5, test.tttt1.id6 |
|   └─IndexLookUp_20                | 10.00    | root      |                             |                                                                                                |
|     ├─IndexFullScan_17(Build)     | 10000.00 | cop[tikv] | table:tttt1, index:idx(id1) | keep order:true, stats:partial[id6:unInitialized]                                              |
|     └─Selection_19(Probe)         | 10.00    | cop[tikv] |                             | eq(test.tttt1.id6, 1)                                                                          |
|       └─TableRowIDScan_18         | 10000.00 | cop[tikv] | table:tttt1                 | keep order:false, stats:partial[id6:unInitialized]                                             |
+-----------------------------------+----------+-----------+-----------------------------+------------------------------------------------------------------------------------------------+
6 rows in set (0.00 sec)
```

v7.1.6

time-and-fate:

Issue Number: close #60242

Problem Summary:

As said in #42060 and the [doc](https://docs.pingcap.com/tidb/stable/system-variables/#tidb_opt_ordering_index_selectivity_threshold-new-in-v700), this variable was originally introduced to control the choice between a "filter index" that satisfies the filter conditions and an "ordering index" that satisfies the `ORDER BY` clause. The allowed range for this variable is `[0,1]`.
Obviously, it's meaningless to choose a "filter index" when the selectivity is 1, which basically means there are no filters. So we made the semantic of this variable "less than". This provides a benefit: when setting it to 0, the behavior is not changed, so we can safely use 0 as the default value.

However, in a recent ticket, we met another case where there were no indexes satisfying the filters, and the choice was between a tiflash path and an "ordering index" path. In this specific ticket, we want to let the optimizer always prefer the tiflash path.
The tiflash path is a pure "selection + table full scan" plan, which means the selectivity will be 1. Due to the "less than" semantic, we are unable to prefer a path where the selectivity is 1. And we believe it's reasonable to extend this variable slightly to handle this case, i.e., change `<` to `<=`.

There is a tricky case we need to care about when we make this change. When we set it to 0, we want to keep the behavior unchanged. So we need to add an extra check for 0: when it's set to 0, the behavior will be unchanged instead of preferring the "filter index" with 0 selectivity.

Besides, we talked about another solution: change the allowed range from `[0,1]` to `[0,1.1]`. So the user can set it to 1.1 to get the same effect, and we don't need to change the code logic. This also avoids the subtle behavior change. But we think it might be a bit weird to set selectivity to larger than 1. So we decide to use the current solution.

1. Change the semantic of `tidb_opt_ordering_index_selectivity_threshold` from `<` to `<=`.
2. Add a special case to this new semantic: when it's set to 0, it's disabled, i.e. don't ignore the `ExpectedCnt`.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/cardinality/selectivity_test.go
+++ b/pkg/planner/cardinality/selectivity_test.go
@@ -1127,7 +1127,7 @@ func TestOrderingIdxSelectivityThreshold(t *testing.T) {
 
 	testKit.MustExec("use test")
 	testKit.MustExec("drop table if exists t")
-	testKit.MustExec("create table t(a int primary key , b int, c int, index ib(b), index ic(c))")
+	testKit.MustExec("create table t(a int primary key , b int, c int, d int, index ib(b), index ic(c))")
 	err := statstestutil.HandleNextDDLEventWithTxn(h)
 	require.NoError(t, err)
 	is := dom.InfoSchema()
@@ -1138,7 +1138,7 @@ func TestOrderingIdxSelectivityThreshold(t *testing.T) {
 	// Mock the stats:
 	// total row count 100000
 	// column a: PK, from 0 to 100000, NDV 100000
-	// column b, c: from 0 to 10000, each value has 10 rows, NDV 10000
+	// column b, c, d: from 0 to 10000, each value has 10 rows, NDV 10000
 	// indexes are created on (b), (c) respectively
 	mockStatsTbl := mockStatsTable(tblInfo, 100000)
 	pkColValues, err := generateIntDatum(1, 100000)
@@ -1158,7 +1158,7 @@ func TestOrderingIdxSelectivityThreshold(t *testing.T) {
 		idxValues = append(idxValues, types.NewBytesDatum(b))
 	}
 
-	for i := 2; i <= 3; i++ {
+	for i := 2; i <= 4; i++ {
 		mockStatsTbl.SetCol(int64(i), &statistics.Column{
 			Histogram:         *mockStatsHistogram(int64(i), colValues, 10, types.NewFieldType(mysql.TypeLonglong)),
 			Info:              tblInfo.Columns[i-1],
--- a/pkg/planner/cardinality/testdata/cardinality_suite_in.json
+++ b/pkg/planner/cardinality/testdata/cardinality_suite_in.json
@@ -281,6 +281,8 @@
       "explain format = 'brief' select * from t where a < 9999 order by c limit 1",
       "explain format = 'brief' select * from t where b >= 0 and b <= 100 or c >= 0 and c <= 100 order by c limit 1",
       "explain format = 'brief' select * from t where a < 1000 and c < 100 limit 1",
+      "explain format = 'brief' select * from t where d >= 9950 order by c limit 1",
+      "explain format = 'brief' select * from t where d < 9950 order by c limit 1",
       "set @@tidb_opt_ordering_index_selectivity_threshold = 0.1",
       "explain format = 'brief' select * from t where b >= 9950 order by c limit 1",
       "explain format = 'brief' select * from t where b >= 9950 order by c desc limit 1",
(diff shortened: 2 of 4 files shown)


Rustin170506 (approved):
Thanks! :shipit:

winoros (approved):

PR #60265, opened 2025-03-26
executor: adjust the SQL digest of stmtsummary to match that of bindings

henrybw:

Issue Number: ref #60148 

Problem Summary:

SPM will use the execution and runtime stats collected by the statement summary when automatically creating bindings, and thus needs to be able to identify a SQL query consistently across statements summary and bindings. However, currently the normalization used for generating SQL digests differs slightly between the two: for bindings, SQL queries are first normalized to always include a schema name.

For example, these queries will have the same SQL digest for bindings, but will have two different SQL digests in the statement summary:

```sql
SELECT * FROM test.t;
SELECT * FROM t;
```

* Modify the statement summary's SQL digest logic to always include a schema name.
* Use `parser.NormalizeDigestForBinding()` rather than `parser.NormalizeDigest()` to ensure that the digests in the statement summary match those used for bindings.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/adapter.go
+++ b/pkg/executor/adapter.go
@@ -68,6 +68,7 @@ import (
 	"github.com/pingcap/tidb/pkg/util/execdetails"
 	"github.com/pingcap/tidb/pkg/util/hint"
 	"github.com/pingcap/tidb/pkg/util/logutil"
+	utilparser "github.com/pingcap/tidb/pkg/util/parser"
 	"github.com/pingcap/tidb/pkg/util/plancodec"
 	"github.com/pingcap/tidb/pkg/util/redact"
 	"github.com/pingcap/tidb/pkg/util/replayer"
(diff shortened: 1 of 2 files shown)


pkg/executor/adapter.go
```diff
@@ -1957,7 +1958,13 @@ func (a *ExecStmt) SummaryStmt(succ bool) {
 	if stmtCtx.StmtType == "" {
 		stmtCtx.StmtType = ast.GetStmtLabel(a.StmtNode)
 	}
-	normalizedSQL, digest := stmtCtx.SQLDigest()
```
qw4990, line 1960: Can we update the digest algo used by `stmtCtx.SQLDigest()` directly? Then all places should use the same algo then.

henrybw: Nice. I'll have to look into how `stmtCtx.SQLDigest()` is used a bit more deeply, but at the moment I don't see any reason not to try this out.

qw4990: Yes, `stmtCtx.SQLDigest()` might be used widely, so this change might cause some UT failures and we have to update these UTs accordingly.

PR #48946, opened 2023-11-28
br: remove duplicate bind info after snapshot restore

Leavrth:
### What problem does this PR solve?

Issue Number: close #46527

Problem Summary:
After snapshot restore, there are duplicated bind infos.
### What changed and how does it work?
execute SQL to remove duplicate bind info after snapshot restore.
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)

```
+-------------------------------------------+--------------------------------------------------------------+------------+---------+-------------------------+-------------------------+---------+-----------------+---------+------------------------------------------------------------------+-------------+
| original_sql                              | bind_sql                                                     | default_db | status  | create_time             | update_time             | charset | collation       | source  | sql_digest                                                       | plan_digest |
+-------------------------------------------+--------------------------------------------------------------+------------+---------+-------------------------+-------------------------+---------+-----------------+---------+------------------------------------------------------------------+-------------+
| builtin_pseudo_sql_for_bind_lock          | builtin_pseudo_sql_for_bind_lock                             | mysql      | builtin | 0000-00-00 00:00:00.000 | 0000-00-00 00:00:00.000 |         |                 | builtin | NULL                                                             | NULL        |
| select * from `test` . `t1` where `b` = ? | SELECT * FROM `test`.`t1` IGNORE INDEX (`b`) WHERE `b` = 123 | test       | enabled | 2023-11-20 16:22:22.239 | 2023-11-20 16:22:22.239 | utf8    | utf8_general_ci | manual  | 94c28782ed1110d4cf67f3a836f92d110a0ebe69113e8d9e1ff5c86523905a3e |             |
+-------------------------------------------+--------------------------------------------------------------+------------+---------+-------------------------+-------------------------+---------+-----------------+---------+------------------------------------------------------------------+-------------+
```

- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #60296, opened 2025-03-27
planner: add an unique key into mysql.bind_info

qw4990:

Issue Number: ref #60148

Problem Summary: planner: add an unique key into mysql.bind_info

add an unique key into mysql.bind_info, later we'll use this unique index to avoid duplicated data when automatically recording bindings.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


qw4990:
These 3 UTs `TestUpdateDuplicateBindInfo`, `TestUpgradeToVer85`, `TestTiDBBindingInListToVer175` are not standard, too old and conflicting with the current PR (they insert duplicated rows into `mysql.bind_info` via insert-stmt directly), so we delete them.

pkg/session/bootstrap.go
```diff
@@ -3406,6 +3411,23 @@ func upgradeToVer245(s sessiontypes.Session, ver int64) {
...
 
+func upgradeToVer246(s sessiontypes.Session, ver int64) {
+	if ver >= version246 {
+		return
+	}
+
+	// to avoid the failure of adding the unique index, remove duplicated rows on these 2 digest columns first.
+	// in most cases, there should be no duplicated rows, since now we only store one binding for each sql_digest.
+	// compared with upgrading failure, it's OK to set these 2 columns to null, the
+	doReentrantDDL(s, `UPDATE mysql.bind_info SET plan_digest=null, sql_digest=null
```
AilinKid, line 3422: 1. should we have some msg out like log or warnings.
2. this rows with plan_digest=null, sql_digest=null can only be deleted by what? user delete their binding with SQL, in the internal, we may fetch the binding with specific like sql_digest=xxx, this may not works for this kind of "dirty rows" ever?

henrybw: I agree that it would be a good idea to log or otherwise indicate that duplicate bindings were found during the upgrade (and their digests were set to null). It's probably not a big deal if this is ignored, but I would think that a DBA would like to know about these updates.

henrybw (commented):
The changes and tests look good to me, but for the upgrader I think we should probably note (via logging probably) any duplicate rows that were found.

qw4990:
Tested the log locally, it can work well when upgrading:
```
[2025/04/01 17:35:08.272 +08:00] [WARN] [bootstrap.go:3449] ["set the following (plan digest, sql digest) in mysql.bind_info to null for adding new unique index: (y, ), (, y), (x, ), (x, x), (y, y), (y, x), (, ), (x, y), (, x)"]
```

pkg/session/bootstrap.go
qw4990: For the second comment, YES, if we set `sql_digest` to null, we are not able to delete them via SQLs like `delete ... where sql_digest = ?` and then they'll be some kind of `dirty rows` forever.
But compared with upgrading failure, some `dirty rows` seems more acceptable. And in most cases, `sql_digest` should not be duplicated, so this problem is less likely to happen.

henrybw (approved):
LGTM. Thanks!

3pointer (approved):
After restore bind_info, br should do similar work like upgradeToVer246 too.

yudongusa (approved):

Leavrth:
After this PR is merged, we can also revert these PRs:
https://github.com/pingcap/tidb/pull/48806/files
https://github.com/pingcap/tidb/pull/48946/files

Issue #60247, opened 2025-03-25
Infinite loop of asynchronous statistics loading


Rustin170506:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

1. Start the cluster with a mock patch:
```diff
diff --git a/pkg/statistics/handle/syncload/stats_syncload.go b/pkg/statistics/handle/syncload/stats_syncload.go
index 4463d1d03a..85693dd2d4 100644
--- a/pkg/statistics/handle/syncload/stats_syncload.go
+++ b/pkg/statistics/handle/syncload/stats_syncload.go
@@ -437,6 +437,8 @@ var errGetHistMeta = errors.New("fail to get hist meta")
 
 // readStatsForOneItem reads hist for one column/index, TODO load data via kv-get asynchronously
 func (*statsSyncLoad) readStatsForOneItem(sctx sessionctx.Context, item model.TableItemID, w *statsWrapper, isPkIsHandle bool, fullLoad bool) (*statsWrapper, error) {
+       // mock timeout
+       time.Sleep(1 * time.Second)
        failpoint.Inject("mockReadStatsForOnePanic", nil)
        failpoint.Inject("mockReadStatsForOneFail", func(val failpoint.Value) {
                if val.(bool) {
```
2. Create a table and insert some data:
```sql
CREATE TABLE t1 (a INT, b INT, c INT, INDEX idx_b (b));
INSERT INTO t1 VALUES (1,3,0), (2,2,0), (3,2,0);
ANALYZE TABLE t1 ALL COLUMNS;
```
3. Run the query:
```sql
SELECT * FROM t1 where b = 2;
DROP TABLE t1;
```
4. Check the logs

### 2. What did you expect to see? (Required)
Do not try to load the stats for the dropped table.

### 3. What did you see instead (Required)
It keeps trying it and prints a lot of logs:

```log
[2025/03/25 15:45:11.781 +08:00] [INFO] [read.go:798] ["Table statistics item not found, possibly due to table being dropped"] [category=stats] [tableID=80] [indexID=2]
[2025/03/25 15:45:11.781 +08:00] [INFO] [read.go:798] ["Table statistics item not found, possibly due to table being dropped"] [category=stats] [tableID=80] [indexID=3]
[2025/03/25 15:45:11.781 +08:00] [INFO] [read.go:659] ["Table statistics item not found, possibly due to table being dropped"] [category=stats] [tableID=80] [columnID=7]
[...]

PR #60318, opened 2025-03-28
statistics: remove invalid histogram items in loadNeededColumnHistograms and loadNeededIndexHistograms

Rustin170506:

Issue Number: close https://github.com/pingcap/tidb/issues/60247

Problem Summary:

If the stats item or table information cannot be found, we should not attempt to load the statistics again, as this could lead to an infinite loop. Therefore, in this PR, I have removed the invalid items from loadNeededColumnHistograms and loadNeededIndexHistograms.

Additionally, if the index has been dropped, we shouldn't return an error; we can simply delete it from the load list.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [x] Manual test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/asyncload/async_load_test.go
+++ b/pkg/statistics/asyncload/async_load_test.go
@@ -1,0 +1,75 @@
+// Copyright 2025 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package asyncload_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/pingcap/tidb/pkg/parser/ast"
+	"github.com/pingcap/tidb/pkg/statistics/asyncload"
+	"github.com/pingcap/tidb/pkg/testkit"
+	"github.com/pingcap/tidb/tests/realtikvtest"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLoadStatisticsAfterTableDrop(t *testing.T) {
+	// Use real tikv to enable the sync and async load.
+	store, dom := realtikvtest.CreateMockStoreAndDomainAndSetup(t)
+	testKit := testkit.NewTestKit(t, store)
+	// Turn off the sync load.
+	testKit.MustExec("SET @@tidb_stats_load_sync_wait = 0;")
+	testKit.MustExec("use test")
+	testKit.MustExec("DROP TABLE IF EXISTS t1")
+	testKit.MustExec("CREATE TABLE t1 (a INT, b INT, c INT, INDEX idx_b (b));")
+	testKit.MustExec("INSERT INTO t1 VALUES (1,3,0), (2,2,0), (3,2,0);")
+	handle := dom.StatsHandle()
+	require.NoError(t, handle.DumpColStatsUsageToKV())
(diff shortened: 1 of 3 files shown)


Rustin170506 (commented):

Rustin170506 (commented):

qw4990 (approved):

pkg/statistics/asyncload/async_load_test.go
```diff
@@ -0,0 +1,219 @@
...
+	store, dom := realtikvtest.CreateMockStoreAndDomainAndSetup(t)
+	testKit := testkit.NewTestKit(t, store)
+	// Turn off the sync load.
+	testKit.MustExec("SET @@tidb_stats_load_sync_wait = 0;")
+	testKit.MustExec("use test")
+	testKit.MustExec("DROP TABLE IF EXISTS t1")
+	testKit.MustExec("CREATE TABLE t1 (a INT, b INT, c INT);")
+	testKit.MustExec("INSERT INTO t1 VALUES (1,3,0), (2,2,0), (3,2,0);")
+	handle := dom.StatsHandle()
+	require.NoError(t, handle.DumpColStatsUsageToKV())
```
time-and-fate, line 88: I think we can have a `handle.DumpStatsDeltaToKV(true)` here to make it more stable.

pkg/statistics/handle/storage/read.go
```diff
@@ -661,6 +661,7 @@ func loadNeededColumnHistograms(sctx sessionctx.Context, statsHandle statstypes.
 			zap.Int64("tableID", col.TableID),
 			zap.Int64("columnID", col.ID),
 		)
+		asyncload.AsyncLoadHistogramNeededItems.Delete(col)
```
time-and-fate, line 664: After this PR, it looks like the rules we are following are:
1. Successful load: the normal case
    - return `nil`
    - remove it from `AsyncLoadHistogramNeededItems`
2. We don't need to load: not normal, but also expected
    - return `nil`
    - remove it from `AsyncLoadHistogramNeededItems`
    - possibly log some information
3. Unexpected case: unexpected error. And we assume a retry is needed the next time (this behavior may also need discussion, though)
    - return an error
    - don't remove it from `AsyncLoadHistogramNeededItems`
    - don't log it since we will log the error outside

So I'm wondering if we should add some comments to emphasize this rule and simplify the code correspondingly. For example:
```
defer func() {
    if err == nil {
        asyncload.AsyncLoadHistogramNeededItems.Delete(col)
    }
}()

pkg/statistics/asyncload/async_load_test.go
Rustin170506: Good catch!

pkg/statistics/handle/storage/read.go
Rustin170506: Make sense!

Rustin170506: > 3\. Unexpected case: unexpected error. And we assume a retry is needed the next time (this behavior may also need discussion, though)

Yes. I would also like to discuss it.

Rustin170506: > So I'm wondering if we should add some comments to emphasize this rule and simplify the code correspondingly. For example:

https://github.com/pingcap/tidb/blob/a7c4f0c07de435d757e3a0f4c63184c3fb1c8b68/pkg/statistics/handle/storage/read.go#L722

I think the only difference here is that for the hist meta, we always remove it. I’m not sure whether we should change this behavior or just perform a duplicate delete here. We can keep this line unchanged and move the others to the defer function.

time-and-fate: Oh, You are right, I missed this case. Seems like it's the only exceptional case now.
I'm OK with this PR. We can try to simplify and unify the logic in these two functions in the future.

Rustin170506:
/unhold

https://github.com/pingcap/tidb/pull/60318#discussion_r2026409064

/test all

Issue #60106, opened 2025-03-17
index join physicalization should follow the prop pushdown requirement mechanism


AilinKid:
## Enhancement

![Image](https://github.com/user-attachments/assets/a132d727-6a4f-4290-80f1-8edfc84f3ed4)

Currently, logical join physicalization is embedded in the logical interface under the name findBestTask()...  Like, when we do the physical enumeration of logic join, the first option is hashJoin, then we enumerate a physical hash join singleton here(default without any physicalized child). Then, we recursively dig down into the logical child, say we saw the projection here, then we enumerate the physical options for this logical projection according to the pushed down prop required and merged by its parent and ancestors.

While an index join breaks this loop, when we enumerate an physical index join, current tidb will assert some inner child pattern in current logical join's logic, then traverse down the asserted inner sub-tree, and build them up completely with undetermined ranges embedded in the underlying tableScan or indexScan (so-called index usage driven when probing rows). And the returned index-join's the inner side has already done its physicalization stuff, why bother to step into findBestTask interface again where the red lined commented in the picture.

Actually, this special workflow can be built with normal prop requirement mechanism, when an index join is enumerated, we build a rough physical index join as well with special index join runtime prop required inside. When we dig into the logical children, we should only handle this special index join runtime prop within those asserted inner pattern logical operator, while for those not, we can just refuse to build the inner physical task for them.

That's all the ideas, since the index join pattern covers more than thousands of lines of code, it needs a familiarity of all the index join building logic, and migrate all these logic into the prop requirement mechanism. Thank you for reading to the end. Thx again.

PR #60349, opened 2025-04-01
planner: add tableScan and indexScan build portal by feeling indexJoinProp.

AilinKid:

Problem Summary:

this is a split pull request from the complete one: https://github.com/pingcap/tidb/pull/60204

in this pull request, we add 3 functions:
* `getBestIndexJoinPathResultByProp` which is changed from `getBestIndexJoinPathResult`, which we don't give them the operator join and related info, because all elements should be passed down with indexJoinProp in our refactor, so I change the parameter to receive indexJoinProp, besides that, the basic detect whether the index path can be valid by this kind of indexJoinProp is still the same by call indexJoinPathBuild with a wrapped structure as indexJoinPathInfo as before.
* `buildDataSource2TableScanByIndexJoinProp` which is changed from `buildIndexJoinInner2TableScan`, the main change between this two is as below:
```
// buildDataSource2TableScanByIndexJoinProp differs with buildIndexJoinInner2TableScan in that
// the first one is try to build a single table scan as the inner child of an index join then return
// this inner task(raw table scan) bottom-up, which will be attached with other inner parents of an
// index join in attach2Task when bottom-up of enumerating the physical plans;
//
// while the second is try to build a table scan as the inner child of an index join, then build
// entire inner subtree of a index join out as innerTask instantly according those validated and
// zipped inner patterns with calling constructInnerTableScanTask. That's not done yet, it also
// tries to enumerate kinds of index join operators based on the finished innerTask and un-decided
// outer child which will be physical-ed in the future.
```
* `buildDataSource2TableScanByIndexJoinProp` which is changed from `buildIndexJoinInner2IndexScan`, the main change is quite same as buildDataSource2TableScanByIndexJoinProp, we just call getBestIndexJoinPathResultByProp to get the index path or table path choosing information, and build raw table-scan or index-scan copTask, with indexJoinInfo wrapped inside, which will be passed upward and end usage in PhysicalIndexJoin.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/exhaust_physical_plans.go
+++ b/pkg/planner/core/exhaust_physical_plans.go
@@ -53,6 +53,8 @@ func exhaustPhysicalPlans4LogicalUnionScan(lp base.LogicalPlan, prop *property.P
 		return nil, true, nil
 	}
 	childProp := prop.CloneEssentialFields()
+	// here we just pass down the keep order property to the child.
+	// cuz, in union scan exec, it will feel the underlying tableReader or indexReader to get the keepOrder.
 	us := PhysicalUnionScan{
 		Conditions: p.Conditions,
 		HandleCols: p.HandleCols,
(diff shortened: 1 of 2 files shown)


pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -859,9 +861,165 @@ childLoop:
...
+	if tblPath == nil {
+		return base.InvalidTask
+	}
+	keyOff2IdxOff := make([]int, len(prop.IndexJoinProp.InnerJoinKeys))
+	var ranges ranger.MutableRanges = ranger.Ranges{}
+	var innerTask base.Task
+	var indexJoinResult *indexJoinPathResult
+	if ds.TableInfo.IsCommonHandle {
+		// for the leaf datasource, we use old logic to get the indexJoinResult, which contain the chosen path and ranges.
+		indexJoinResult, keyOff2IdxOff = getBestIndexJoinPathResultByProp(ds, prop.IndexJoinProp, func(path *util.AccessPath) bool { return path.IsCommonHandlePath })
```
hawkingrei, line 947:
```suggestion
	var ranges ranger.MutableRanges = ranger.Ranges{}
	var innerTask base.Task
	var indexJoinResult *indexJoinPathResult
	if ds.TableInfo.IsCommonHandle {
		// for the leaf datasource, we use old logic to get the indexJoinResult, which contain the chosen path and ranges.
		indexJoinResult, keyOff2IdxOff := getBestIndexJoinPathResultByProp(ds, prop.IndexJoinProp, func(path *util.AccessPath) bool { return path.IsCommonHandlePath })
```

hawkingrei: ```keyOff2IdxOff```'s ```make``` is useless.

pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -859,9 +861,165 @@ childLoop:
 	return wrapper
 }
 
+// buildDataSource2TableScanByIndexJoinProp builds an IndexScan as the inner child for an
```
qw4990, line 864:
```suggestion
// buildDataSource2IndexScanByIndexJoinProp builds an IndexScan as the inner child for an
```

pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -859,9 +861,165 @@ childLoop:
...
+// IndexJoin based on IndexJoinProp included in prop if possible.
+//
+// buildDataSource2TableScanByIndexJoinProp differs with buildIndexJoinInner2TableScan in that
```
qw4990, line 867:
```suggestion
buildDataSource2IndexScanByIndexJoinProp
```

qw4990 (approved):

hawkingrei (approved):

PR #60366, opened 2025-04-02
planner: recommend optimal plans in SHOW PLAN statement

qw4990:

Issue Number: ref #60148

Problem Summary: planner: recommend optimal plans in SHOW PLAN statement

Implement 3 rules to recommend optimal plans in SHOW PLAN statement.
// Rules:
// 1. If any plan is a simple PointGet or BatchPointGet, recommend it.
// 2. If `ScanRowsPerReturnedRow` of a plan is 50% better than others', recommend it.
// 3. If `Latency`, `ScanRows` and `LatencyPerReturnRow` of a plan are 50% better than others', recommend it.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/bindinfo/binding_auto.go
+++ b/pkg/bindinfo/binding_auto.go
@@ -271,6 +271,7 @@ func IsSimplePointPlan(plan string) bool {
 			continue
 		}
 		operatorName := strings.Split(line, " ")[0]
+		// TODO: these hard-coding lines are a temporary implementation, refactor this part later.
 		if operatorName == "id" || // the first line with column names
 			strings.Contains(operatorName, "Point_Get") ||
 			strings.Contains(operatorName, "Batch_Point_Get") ||
--- a/pkg/bindinfo/binding_auto.go
+++ b/pkg/bindinfo/binding_auto.go
@@ -199,7 +199,7 @@ func FillRecommendation(bindingPlans []*BindingPlanInfo) {
 	hitRule := false
 	// rule 1
 	for _, cur := range bindingPlans {
-		if isSimplePointPlan(cur.Plan) {
+		if IsSimplePointPlan(cur.Plan) {
 			hitRule = true
 			cur.Recommend = "YES (rule-based)"
 			cur.Reason = "Simple PointGet or BatchPointGet is the best plan"
@@ -260,7 +260,9 @@ func FillRecommendationViaLLM([]*BindingPlanInfo) {
 	// TODO
 }
 
-func isSimplePointPlan(plan string) bool {
+// IsSimplePointPlan checks whether the plan is a simple point plan.
+// Expose this function for testing.
+func IsSimplePointPlan(plan string) bool {
 	// if the plan only contains Point_Get, Batch_Point_Get, Selection and Projection, it's a simple point plan.
 	lines := strings.Split(plan, "\n")
 	for _, line := range lines {
--- a/pkg/bindinfo/binding_auto.go
+++ b/pkg/bindinfo/binding_auto.go
@@ -16,6 +16,7 @@ package bindinfo
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/pingcap/errors"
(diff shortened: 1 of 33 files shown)


pkg/bindinfo/binding_auto.go
```diff
@@ -174,3 +177,106 @@ type planExecInfo struct {
...
+		FillRecommendationViaLLM(bindingPlans)
+	}
+}
+
+// FillRecommendationViaLLM fills the recommendation field for each binding plan via LLM.
+func FillRecommendationViaLLM([]*BindingPlanInfo) {
+	// TODO
+}
+
+func isSimplePointPlan(plan string) bool {
```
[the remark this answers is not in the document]
qw4990: Thanks, I've added a new dedicated test for this function now.

pkg/bindinfo/binding_auto.go
```diff
@@ -174,3 +177,108 @@ type planExecInfo struct {
...
+// 1. If any plan is a simple PointGet or BatchPointGet, recommend it.
+// 2. If `ScanRowsPerReturnedRow` of a plan is 50% better than others', recommend it.
+// 3. If `Latency`, `ScanRows` and `LatencyPerReturnRow` of a plan are 50% better than others', recommend it.
+// LLM: TODO
+func FillRecommendation(bindingPlans []*BindingPlanInfo) {
+	for _, p := range bindingPlans {
+		if p.ExecTimes == 0 {
+			// if we can't get plan info for any binding, we can't give any recommendation.
+			FillRecommendationViaLLM(bindingPlans)
+			return
```
AilinKid, line 194:
```suggestion
			return FillRecommendationViaLLM(bindingPlans)
```

AilinKid: since this is in a loop, once a p has a 0 ExecTimes, it may turn to llm fill and return back, is this ok, what about the others in the loop?

pkg/bindinfo/binding_auto.go
```diff
@@ -174,3 +177,108 @@ type planExecInfo struct {
...
+			cur.Reason = "Simple PointGet or BatchPointGet is the best plan"
+		}
+	}
+
+	// rule 2
+	if !hitRule {
+		for i, cur := range bindingPlans {
+			hitRule2 := true
+			for j, other := range bindingPlans {
+				if i == j || cur.ScanRowsPerReturnRow < other.ScanRowsPerReturnRow/2 {
```
henrybw, line 214: This (and rule 3) are both O(N^2), which is not good for scalability. Is there any way we can avoid having multiple nested loops? Perhaps we could create an auxiliary structure that sorts the binding plans by `ScanRowsPerReturnRow`? Or maybe we could have one outer loop over `bindingPlans` and evaluate all rules for each plan, rather than 3 loops over `bindingPlans`?

pkg/bindinfo/binding_auto.go
```diff
@@ -174,3 +177,108 @@ type planExecInfo struct {
...
+		if line == "" {
+			continue
+		}
+		operatorName := strings.Split(line, " ")[0]
+		if operatorName == "id" || // the first line with column names
+			strings.Contains(operatorName, "Point_Get") ||
+			strings.Contains(operatorName, "Batch_Point_Get") ||
+			strings.Contains(operatorName, "Selection") ||
+			strings.Contains(operatorName, "Projection") {
+			continue
```
henrybw, line 279: Is there any way we can avoid hardcoding the operator names here? Are there utility functions in the planner or plancodec that we can use instead? Maybe this suggestion doesn't make sense, but I think it's better if we can avoid hardcoding plan operator names (especially if we want to change them in the future, for example).

pkg/bindinfo/binding_auto.go
qw4990: Yes, it's expected, if we can't get plan info for any binding where binding.ExecTimes == 0, we don't do this rule-based recommendation.

pkg/bindinfo/binding_auto.go
qw4990: Yes, I hate these hardcoding lines also... I've checked with @elsa0520  and she is working on some workload-learning feature which has to parse the Plan more standardly. I left a comment here that this is a temporary implementation and I'll refactor this part after @elsa0520 merges her code.

pkg/bindinfo/binding_auto.go
qw4990: Good suggestion, now I update the implementation here, I sort bindings first, now the complexity here should be O(NlogN) now, which should be acceptable.

pkg/bindinfo/binding_auto.go
AilinKid: Soga LGTM

AilinKid (approved):

henrybw (approved):
LGTM. Thanks!

PR #60383, opened 2025-04-02
planner: add enumeration of simple index join and complete it with indexJoinInfo built bottom-up

AilinKid:

Problem Summary:

we split the original `constructIndexJoin` function into two parts
* `constructIndexJoinStatic`: simple enumerate a index join with inner side a pushed-down index join prop. In the construction of this physic index join, we can only fill something that's static, which is derived from the logical join node.
* `completePhysicalIndexJoin`: which index path does the inner ds choose? What are the ranges? What's used index column length? What's the last col's non-eq range compare function? All these questions can only be answered when the inner side of the plan tree is built from the bottom up. So in `(physicIndexJoin) Attach2task`, the physical index join can get what they want, only at that time, the many fields in physicIndexJoin can be filled compared with the original `constructIndexJoin`.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -523,6 +523,203 @@ func getHashJoin(p *logicalop.LogicalJoin, prop *property.PhysicalProperty, inne
...
+					outerSchema, innerSchema := outerS, innerS
+					if outerSchema.Contains(lhs) && innerSchema.Contains(rhs) {
+						outerHashKeys = append(outerHashKeys, lhs) // nozero
+						innerHashKeys = append(innerHashKeys, rhs) // nozero
+					} else if innerSchema.Contains(lhs) && outerSchema.Contains(rhs) {
+						outerHashKeys = append(outerHashKeys, rhs) // nozero
+						innerHashKeys = append(innerHashKeys, lhs) // nozero
+					}
+					// if not, this EQ function is useless, keep it in new other conditions.
+					newOtherConds = append(newOtherConds[:i], newOtherConds[i+1:]...)
```
hawkingrei, line 695: here can use ```slices.Delete```.

hawkingrei (approved):

pkg/planner/core/exhaust_physical_plans.go
AilinKid: got it

zimulala (approved):

PR #60449, opened 2025-04-09
planner: add enumeration of index join with indexJoinProp pushed down to inner plan.

AilinKid:

Problem Summary:

this is a split pull request from the complete one: https://github.com/pingcap/tidb/pull/60204

* `enumerateIndexJoinByOuterIdx` is quite similar to `getIndexJoinByOuterIdx`, while inside we do not assert the inner plan's pattern and build their physical out, but pushing down the indexJoinProp instead and return a static(temporary) physical index join out directly.

* `tryToEnumerateIndexJoin` logic is quite similar to `tryToGetIndexJoin`, while inside we do not build the inner physical plan out, but pushing down the indexJoinProp instead and return a static(temporary) physical index join out directly.

* `suitLogicalJoinHint` logic is quite similar to `handleForceIndexJoinHints`; the later one will prune other physical plans at the enumeration stage, which is too early because we are not sure the index join can be successfully built later. So the `suitLogicalJoinHint` is try to handle the hint related pruning in `findBestTask`  when see the real physical plans.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/exhaust_physical_plans.go
+++ b/pkg/planner/core/exhaust_physical_plans.go
@@ -16,6 +16,7 @@ package core
 
 import (
 	"fmt"
+	"github.com/pingcap/tidb/pkg/util/dbterror/plannererrors"
 	"math"
 	"slices"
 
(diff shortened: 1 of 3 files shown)


pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -2473,6 +2637,7 @@ func exhaustPhysicalPlans4LogicalJoin(lp base.LogicalPlan, prop *property.Physic
 		}
 		joins = append(joins, mergeJoins...)
 
+		// todo: feel vars.EnhanceIndexJoinBuildV2 and tryToEnumerateIndexJoin(p, prop)
```
AilinKid, line 2640: here is the usage protal, which will be controlled with EnhanceIndexJoinBuildV2 with default ON later

pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -970,6 +996,40 @@ func constructIndexHashJoin(
...
+	if outerIdx == 0 {
+		outerJoinKeys, innerJoinKeys, _, _ = p.GetJoinKeys()
+	} else {
+		innerJoinKeys, outerJoinKeys, _, _ = p.GetJoinKeys()
+	}
+	// computed the avgInnerRowCnt
+	var avgInnerRowCnt float64
+	if outerChild.StatsInfo().RowCount > 0 {
+		avgInnerRowCnt = p.EqualCondOutCnt / outerChild.StatsInfo().RowCount
+	}
```
hawkingrei, line 1021:
```suggestion
	if count := outerChild.StatsInfo().RowCount; count > 0 {
		avgInnerRowCnt = p.EqualCondOutCnt / count
	}
```

hawkingrei (approved):

pkg/planner/core/find_best_task.go
```diff
@@ -249,13 +249,33 @@ func enumeratePhysicalPlans4Task(
...
+		// one among them.
+		// todo: extend suitLogicalJoinHint to be a normal logicalOperator's interface to handle the hint related stuff.
+		if suitLogicalJoinHint(p.Self(), curTask.Plan()) {
+			// curTask is a preferred physic plan, compare cost with previous preferred one and cache the low-cost one.
+			if curIsBetter, err := compareTaskCost(curTask, preferTask, opt); err != nil {
+				return nil, 0, err
+			} else if curIsBetter {
+				preferTask = curTask
+			}
+		}
```
zimulala, line 272: I noticed that `recordIndexJoinHintWarnings` is not being called at the moment. Are we planning to call it here, or somewhere else?

zimulala (approved):
LGTM

PR #60462, opened 2025-04-10
session: syssession checks the concurrency of thread-unsafe calls

lcwangchao:

Issue Number: ref #60115

Problem Summary:

Consider below case:

```go
pool := NewAdvancedSessionPool(256, factory);
s, err := pool.Get()
// handle error ...
go func() {
    rs, err := s.ExecuteInternal(ctx, "select * from t");
    // handle result
}()

go func() {
    rs, err := s.ExecuteInternal(ctx, "select * from t");
    // handle result
}()
```

The above two statements are executed concurrently which is not expected and may cause data race. It's better to deny the second execution and return an error instead.

This PR adds some checks of the concurrency of thread-unsafe calls. When one thread-unsafe operation is ongoing and another operation (also thread-unsafe) begins, the latter one will return an error.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/session/syssession/BUILD.bazel
+++ b/pkg/session/syssession/BUILD.bazel
@@ -35,7 +35,7 @@ go_test(
     ],
     embed = [":syssession"],
     flaky = True,
-    shard_count = 19,
+    shard_count = 21,
     deps = [
         "//pkg/kv",
         "//pkg/parser/ast",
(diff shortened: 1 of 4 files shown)


wjhuang2016 (approved):

Rustin170506 (approved):
Thanks! :shipit:

PR #23666, opened 2021-03-30
planner: fix the issue that planner hints don't work in some batch/point-get plans

qw4990:
### What problem does this PR solve?

Issue Number: close #23570 

Problem Summary: planner: fix the issue that planner hints don't work in some batch/point-get plans

### What is changed and how it works?
Add some logic about checking planner hints in `TryFastPlan`.
Please see the root cause and more details in https://github.com/pingcap/tidb/issues/23570.

### Check List 

Tests 

- Unit test

### Release note 

- planner: fix the issue that planner hints don't work in some batch/point-get plans

PR #60516, opened 2025-04-13
planner: add comments style hints check for the fast path

### LINKED ISSUE #59427: comments style optimizer hints can't work for point get plan sometimes
## Bug Report

```sql
create table t(a int, b int, c int, primary key(a,b), index ib(b));
explain select /*+ use_index(t,ib) */ * from t where a = 1 and b = 1;
explain select * from t use index (ib)  where a = 1 and b = 1;
```

Both hints are effective.

One of the hints is ineffective.

```
> create table t(a int, b int, c int, primary key(a,b), index ib(b));
Query OK, 0 rows affected (0.065 sec)

> explain select /*+ use_index(t,ib) */ * from t where a = 1 and b = 1;
+-------------+---------+------+----------------------------------------+---------------+
| id          | estRows | task | access object                          | operator info |
+-------------+---------+------+----------------------------------------+---------------+
| Point_Get_1 | 1.00    | root | table:t, clustered index:PRIMARY(a, b) |               |
+-------------+---------+------+----------------------------------------+---------------+
1 row in set (0.001 sec)

> explain select * from t use index (ib)  where a = 1 and b = 1;
+-------------------------------+---------+-----------+----------------------+---------------------------------------------+
| id                            | estRows | task      | access object        | operator info                               |
+-------------------------------+---------+-----------+----------------------+---------------------------------------------+
| IndexLookUp_8                 | 1.00    | root      |                      |                                             |
| ├─Selection_7(Build)          | 1.00    | cop[tikv] |                      | eq(test.t.a, 1)                             |
| │ └─IndexRangeScan_5          | 10.00   | cop[tikv] | table:t, index:ib(b) | range:[1,1], keep order:false, stats:pseudo |
| └─TableRowIDScan_6(Probe)     | 1.00    | cop[tikv] | table:t              | keep order:false, stats:pseudo              |
+-------------------------------+---------+-----------+----------------------+---------------------------------------------+
```

v8.5.1

time-and-fate:

Issue Number: close #59427

Problem Summary:

The comment style optimizer hint is not checked in the fast path. It should not cause any actual performance problems, but the user just can't enforce another index.

Previously, we added a check for the `USE/FORCE/IGNORE INDEX (idx_name)` syntax hint in #23666. But checks for the comment style hint are missing.

The code structure in the fast path is very different from the normal path, so it's hard to reuse the code there directly. I try to re-implement a simpler version while keeping the same checking behavior as the normal path. See the comments for details.

Similar to #23666, we still do the checks in `indexIsAvailableByHints()`. However, since the checking logic for the comment style hints is more complicated, we need to pass extra information to this function.
Several new parameters are added to this function: the current DB, the DB name and the table alias name of the accessed table, and the hint itself.
To pass the information along the way from `TryFastPlan()`, some other functions are also changed.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


winoros (approved):

winoros:
setvar needs to be handled separately

pkg/planner/core/point_get_plan.go
```diff
@@ -1527,8 +1566,56 @@ func checkTblIndexForPointPlan(ctx base.PlanContext, tblName *resolve.TableNameW
...
+		case hint.HintUseIndex:
+			hintType = ast.HintUse
+		case hint.HintIgnoreIndex:
+			hintType = ast.HintIgnore
+		case hint.HintForceIndex:
+			hintType = ast.HintForce
+		default:
+			continue
+		}
+		if len(h.Tables) != 1 {
```
Rustin170506, line 1600: May I ask when this is expected to happen?

pkg/planner/core/point_get_plan.go
```diff
@@ -1527,8 +1566,56 @@ func checkTblIndexForPointPlan(ctx base.PlanContext, tblName *resolve.TableNameW
...
+			continue
+		}
+		hintDBName := h.Tables[0].DBName
+		if hintDBName.L == "" {
```
Rustin170506, line 1604: This means that if the user doesn’t specify the database name here, we’ll use the current default database, right?

Rustin170506 (approved):
Thanks! :shipit:

pkg/planner/core/point_get_plan.go
time-and-fate: I just checked the .y definition. It should not happen. I also added an intest assert here.

https://github.com/pingcap/tidb/blob/b2f2faaa95bff642920560c24e3714957bcb4c74/pkg/parser/hintparser.y#L521-L555

pkg/planner/core/point_get_plan.go
time-and-fate: Yes.

PR #60592, opened 2025-04-16
planner: add variable control for index join build v2 and default them on | tidb-test=8b31b9d8bbfa9b3c1f6911ffc733f14fab32372d

AilinKid:

Issue Number: close https://github.com/pingcap/tidb/issues/60106
Problem Summary:

some changes in previous split pull requests: 
* we found some plan cache case inside `planner/core/casetest/rule/rule_join_reorder` is missing some rows, Finally we found that when building the index path, we should use `indexJoinResult.chosenRanges` as mutable ranges to rebuild the range when the parameter marker changes in the plan cache case. while in the original code we use `indexJoinResult.chosenRanges.Ranges` which won't do the rebuild process.

* previously, index join prop will be passed down once asked for datasource index/table range path, which means tableRangeScan copTask will compare the cost with indexRangeScan copTask if any. Actually, some double read cost and attached upper operators cost haven't been summarized here, so we just let index join to enumerate two kinds of index join prop: one is for asking tableRangeScan if any, the other one is for asking indexRangeScan. So finally, in the physicalIndexJoin itself, it will compare the two kinds of subtrees' cost completely.

* index join couldn't prune other physical alternatives: previously, physical plan pruning was more hard-coded in the enumeration phase, while for index join property. We couldn't say that the user specified an index join hint, so we enumerated an index join choice. That's all done, and other physical choices should
be pruned. NO! because index join prop couldn't be promised to succeed in the underlying inner operators, and this kind of prop couldn't even be enforced as well. In such a case, we should have the other physical choices as an output plan.

* warnings handle: warnings should be bound with a specific physical operator. For example, the user hints want to choose index join, while it is impossible, and we generate hash join as the best task at this level. Then, the latter upper operator which chooses this hashJoin(prop & bestTask) as its children. The upper encapsulated task, no matter what type it is, should inherit the warnings from lower cached tasks.

* index join inner side operators: since inner side only uses the avgInnerRows to keep the stats rowcount, which is the meaning for when we say an outer row, how many inner rows will be produced. So the inner tree's original stats info is not suitable for this kind of usage, just as some issues reported that some index join inner operators' row count is GREATER than the underlying range scan. So in this pull request, we simply inherit stats from inner DS for all index join inner high-level operators.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -2197,12 +2233,47 @@ func recordIndexJoinHintWarnings(lp base.LogicalPlan, prop *property.PhysicalPro
...
+	if physic == nil {
+		return false
+	}
+	forceLeftToBuild := ((p.PreferJoinType & h.PreferLeftAsHJBuild) > 0) || ((p.PreferJoinType & h.PreferRightAsHJProbe) > 0)
+	forceRightToBuild := ((p.PreferJoinType & h.PreferRightAsHJBuild) > 0) || ((p.PreferJoinType & h.PreferLeftAsHJProbe) > 0)
+	preferHashJoin := p.PreferJoinType&h.PreferHashJoin > 0 || forceRightToBuild || forceLeftToBuild
+	if _, ok := physic.(*PhysicalHashJoin); ok && preferHashJoin {
+		return true
+	}
+	return false
```
hawkingrei, line 2259:
```suggestion
	_, ok := physic.(*PhysicalHashJoin)
	return ok && preferHashJoin
```

pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -2197,12 +2233,47 @@ func recordIndexJoinHintWarnings(lp base.LogicalPlan, prop *property.PhysicalPro
...
+	if !ok {
+		return false
+	}
+	if physic == nil {
+		return false
+	}
+	if _, ok := physic.(*PhysicalMergeJoin); ok && p.PreferJoinType&h.PreferMergeJoin > 0 {
+		return true
+	}
+	return false
```
hawkingrei, line 2273:
```suggestion
	_, ok := physic.(*PhysicalMergeJoin)
	return ok && p.PreferJoinType&h.PreferMergeJoin > 0 
```

pkg/planner/core/find_best_task.go
```diff
@@ -610,30 +621,46 @@ func findBestTask(lp base.LogicalPlan, prop *property.PhysicalProperty, planCoun
...
-			// work anyway, we give up `plansNeedEnforce` for efficiency,
-			plansNeedEnforce = nil
+		// since index join hint can only be known as worked or not after physic implementation.
+		// it means hintCanWork and hintWorksWithProp is not determined here. if we try to eliminate
+		// some physical alternatives here like clean plansFitsProp or plansNeedEnforce we need to
+		// guarantee that no index join is conclude in the both.
+		//
+		// so we need to check the plansFitsProp or plansNeedEnforce both to find the low-cost and
+		// hint applicable plan.
+		if !enumerationContainIndexJoin(plansFitsProp) && !enumerationContainIndexJoin(plansNeedEnforce) {
```
AilinKid, line 631:
```


mysql> desc format = 'brief' select /*+ INL_JOIN(t2) */ * from t t1, t t2 where t1.a = t2.b order by t2.a;

physical sort (14998331.429508679)
    +--prop(sortItem=nil)---->  projection
                                    +--prop(sortItem=nil)----> join
                                                                +--prop(sortItem=nil)---->   index join1 (t1, t2)   ✅ hint-preferred picked ✅
                                                                                             index join2 (t2, t1)
                                                                                             hash join1 (t1, t2)  
                                                                                             hash join2 (t2, t1)  

norminal sort (5579942.460057036)
    +--prop(sortItem=t2.'a')---->  projection 
                                    +--prop(sortItem=t2.'a')----> join
                                                                +--prop(sortItem=t2.'a')----> index join1 (t2, t1)  ✅ lowest cost picked ✅
                                                                            |                 // hash join can't keep order  
                                                                            |
                                                                            | (enforce)
                                                                            V
                                                                +--prop(sortItem=nil) ------> index join1 (t1, t2)  ✅ hint-preferred alternative ✅
                                                                                              index join2 (t2, t1)  
                                                                                              hash join1 (t1, t2)   
                                                                                              hash join2 (t2, t1)         

                                                                                              
```
the target case, previous hintCanWork and hintWorksWithProp is right, but for index join case, hint can't work is not for sure, if we want to leverage this two variable to do some accerlaration (set plansFitsProp=nil, or set plansNeedEnforce =nil), we should guarantee that there is no index join of this two plans.

pkg/planner/core/find_best_task.go
```diff
@@ -181,14 +181,18 @@ func rebuildChildTasks(p *logicalop.BaseLogicalPlan, childTasks *[]base.Task, pp
 	return nil
 }
 
+// @first: indicates the best task returned.
```
qw4990, line 184: How about also adding variable names to all returned values like `(bestTask base.Task, planCnt int64, hintApplied bool, err error)`?

qw4990 (approved):

yudongusa (approved):
Please also open a document PR, and do more plan regression and performance tests.

wjhuang2016 (approved):

hawkingrei (approved):

Issue #58720, opened 2025-01-06
Refactor GC related PD APIs and GC procedures


MyonKeminta:
## Enhancement

After years of iteration, the GC related code, which originally lacked of proper design, has become a huge historical debt that makes our futher mantainance, development and optimization difficult. Some badly-designed interfaces cause misuses frequently, and these misuses can be dangerous to affect the data correctness. Now, we've planned a refactor to redesign most of these related PD APIs, redefine some terms, and avoid those code that's obscure and easy to misuse. After the refactoring which is expected to make the GC related APIs and concetps more clear and comprehensible, we will then implement a feature named *protected snapshots* over it, which is expected to help PiTR to reduce their cost to maintain ther backup progress.

### Subtasks

- [x] PD API refactorying: https://github.com/tikv/pd/issues/8978
- [ ] TiKV: Adapt to the new PD API
    - [x] Next gen
    - [ ] master
- [x] Adapt TiDB's GCWorker to the new PD API  https://github.com/pingcap/tidb/pull/60632
    - [x] Mock PD of unistore: https://github.com/pingcap/tidb/pull/61529
    - [x] Required adaption in client-go: https://github.com/tikv/client-go/pull/1634
- [ ] TiKV: Add checks to prevent reading/writing to unsafe snapshots

### Fixes

- https://github.com/pingcap/tidb/pull/61925

(...to be more detailed)

### Next step

After finishing the refactoring, we will then implement *protected snapshots* upon it. We will then track that task in another issue.

PR #60632, opened 2025-04-17
gc_worker: Adapt to new GC API and support keyspace level GC

MyonKeminta:

Issue Number: ref #58720

Problem Summary:

This PR updates the implementation of GCWorker to adapt the changes of PD's GC related APIs made in https://github.com/tikv/pd/issues/8978 . This update also makes it able to support keyspace level GC.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/store/gcworker/gc_worker.go
```diff
@@ -751,62 +729,76 @@ func (w *GCWorker) runGCJob(ctx context.Context, safePoint uint64, concurrency g
 	})
 	metrics.GCWorkerCounter.WithLabelValues("run_job").Inc()
 
-	err := w.resolveLocks(ctx, safePoint, concurrency.v)
+	// ----------*--------------------*--------------------> time
```
cfzjywxk, line 741: For the TiDB-side code, `TxnSafePoint` is a newly introduced concept. It's better to briefly explain this concept and how it differes from the original `GcSafepoint`, also we could paste the relevant design document links here.

pkg/store/gcworker/gc_worker.go
```diff
@@ -389,8 +393,11 @@ func (w *GCWorker) leaderTick(ctx context.Context) error {
 	// If `keyspace-name` is set, the TiDB node will only do its own delete range, and will not calculate gc safe point and resolve locks.
 	// Note that when `keyspace-name` is set, `checkLeader` will be done within the key space.
 	// Therefore only one TiDB node in each key space will be responsible to do delete range.
-	if w.store.GetCodec().GetKeyspace() != nil {
-		err = w.runKeyspaceGCJob(ctx, concurrency)
+	// TODO: Use result of AdvanceTxnSafePoint instead, which makes it unnecessary to manually check the config.
+	const cfgGCManagementType = "gc_management_type"
+	const cfgGCManagementTypeKeyspaceLevel = "keyspace_level"
+	if keyspaceMeta := w.store.GetCodec().GetKeyspaceMeta(); keyspaceMeta != nil && keyspaceMeta.Config[cfgGCManagementType] != cfgGCManagementTypeKeyspaceLevel {
```
tiancaiamao, line 399: I think we need the metrics to distinguish whether the tidb is running in keyspace level GC... 
I'll add it later.

pkg/store/gcworker/gc_worker.go
```diff
@@ -675,7 +635,7 @@ func (w *GCWorker) validateGCLifeTime(lifeTime time.Duration) (time.Duration, er
 	return gcMinLifeTime, err
 }
 
-func (w *GCWorker) calcNewSafePoint(ctx context.Context, now time.Time) (*time.Time, uint64, error) {
+func (w *GCWorker) calcNewTxnSafePoint(ctx context.Context, now time.Time) (*time.Time, uint64, error) {
 	lifeTime, err := w.loadDurationWithDefault(gcLifeTimeKey, gcDefaultLifeTime)
```
tiancaiamao, line 639: Why the return value is `*time.Time` rather than `time.Time`, it looks weird.

pkg/store/gcworker/gc_worker.go
tiancaiamao: We need to upload the internal design document to the repo for the link

pkg/store/gcworker/gc_worker.go
```diff
@@ -751,62 +729,76 @@ func (w *GCWorker) runGCJob(ctx context.Context, safePoint uint64, concurrency g
...
 	metrics.GCWorkerCounter.WithLabelValues("run_job").Inc()
 
-	err := w.resolveLocks(ctx, safePoint, concurrency.v)
+	// ----------*--------------------*--------------------> time
+	//                                ^ Txn safe point (not synced)
+	//           ^ GC safe point (last value)
+
+	// The txn safe point should be guaranteed to be synchronized over all necessary components after
+	// `txnSafePointSyncWaitTime`. We wait for that time before performing any operation about the GC.
+	time.Sleep(txnSafePointSyncWaitTime)
```
tiancaiamao, line 747: Let me check if I understand it correctly:
if other tidb instance cannot load the txn safe point within `txnSafePointSyncWaitTime` interval, 
it will reject any txn commit operation. The check commit ts step should fail.

pkg/store/gcworker/gc_worker.go
```diff
@@ -1239,36 +1231,25 @@ func (w *GCWorker) resolveLocks(
...
-			}
-			err = bo.Backoff(tikv.BoPDRPC(), errors.Errorf("failed to upload safe point to PD, err: %v", err))
-			if err != nil {
-				return errors.Trace(err)
-			}
-			continue
-		}
-		break
+func (w *GCWorker) broadcastGCSafePoint(ctx context.Context, gcSafePoint uint64) error {
+	result, err := w.pdGCControllerClient.AdvanceGCSafePoint(ctx, gcSafePoint)
```
tiancaiamao, line 1244: Before we use keyspace level GC, there is only one gc worker leader invoking this kind of API, so the QPS should be low enough.
Now one PD may serve many keyspace, **so concurrency matters**.
In the PD side there is a big mutex to protect all things for this kind of API, 
we should do more test to verify the performance later.

pkg/store/gcworker/gc_worker.go
MyonKeminta: I'll add some brief documents for it, and link to comments from PD repo which is quite rich. Unluckily the design document is not public. Maybe we can public that later.

pkg/store/gcworker/gc_worker.go
MyonKeminta: To make it nil-able. It looks that this is common in Go programming. I dislike it though.

pkg/store/gcworker/gc_worker.go
MyonKeminta: transaction committing is currently not checked, and only read operations are affected.
`txnSafePointSyncWaitTime` is 100s, but the max allowed interval between updating txn safe point is [90s](https://github.com/tikv/client-go/blob/74c0a811505babbb876c1f8e9c6f72b6fea12648/tikv/kv.go#L186C36-L186C60).

Before PR https://github.com/tikv/client-go/pull/1634 , this timeout was 99s.

pkg/store/gcworker/gc_worker.go
MyonKeminta: Yes.
But if we can accept sacrificing the time cost of `SetGlobalGCBarrier` calls, in the future, we can consider make the mutex and gc state revision key fine-grained to each keyspace. And then we need to find some syncing algorithm for `SetGlobalGCBarrier` to guarantee its safety without the global lock.

pkg/store/gcworker/gc_worker.go
cfzjywxk: @MyonKeminta How about pritting some logs here to display which `keyspace` name is configured and information about the mode?

PR #60648, opened 2025-04-18
planner: add skew risk for equals

### LINKED ISSUE #60722: Planner: Selectivity estimate can be too low if not found in TopN
## Enhancement

If TopN is not available, or a value is not found in the TopN - the optimizer can underestimate the number of qualified rows for values that are actually skewed.

terry1purcell:

Issue Number: close #60722

Problem Summary:

This enhancement is controlled by a new global/session variable tidb_opt_risk_eq_skew_ratio - that is also eligible for query level SET_VAR hint.

When an equal predicate value is NOT found in TopN, or as the last value in a histogram bucket - we will use the histogram rowcount (which excludes topN in stats version 2) divided by the remaining NDV. Thus, we assume that every remaining value is evenly distributed throughout the range. In reality, any single value could be skewed - since the topN values are limited to 100 values, and the histograms do not attempt to preference skewed values as the last in the bucket.

This enhancement will estimate the worst case skew for a value - which is (remaining count - remaining NDV) - but not to exceed the smallest value in the existing TopN. And then the tidb_opt_risk_eq_skew_ratio variable controls how much of that skew should be considered in the estimate - with 0 (the default) resulting in an estimate that matches the current count/ndv, and 1 == 100% of the skew.

This allows a user or external process to control the amount of skew should be considered in the optimizer's estimate when statistics are insufficient for estimating the exact value's selectivity.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/sessionctx/vardef/tidb_vars.go
```diff
@@ -323,6 +323,9 @@ const (
 	// TiDBOptCorrelationExpFactor is an exponential factor to control heuristic approach when tidb_opt_correlation_threshold is not satisfied.
 	TiDBOptCorrelationExpFactor = "tidb_opt_correlation_exp_factor"
 
+	// TiDBOptSkewRatio controls the amount of skew is applied to equal predicate estimation when a value is not found in TopN/buckets.
+	TiDBOptSkewRatio = "tidb_opt_skew_ratio"
```
qw4990, line 327: Could you add a simple test for this variable?

terry1purcell: Added new testcase as suggested

pkg/planner/cardinality/row_count_index.go
```diff
@@ -590,3 +605,22 @@ func getOrdinalOfRangeCond(sc *stmtctx.StatementContext, ran *ranger.Range) int
 	}
 	return len(ran.LowVal)
 }
+
+func getTopNMinMax(topN *statistics.TopN) (minN, maxN uint64) {
```
qw4990, line 609: It seems like we can use `topN.MinCount()` directly.

pkg/sessionctx/variable/setvar_affect.go
```diff
@@ -71,6 +71,7 @@ var isHintUpdatableVerified = map[string]struct{}{
 	"tidb_opt_table_rowid_scan_cost_factor":           {},
 	"tidb_opt_table_tiflash_scan_cost_factor":         {},
 	"tidb_opt_topn_cost_factor":                       {},
+	"tidb_opt_skew_ratio":                             {},
```
qw4990, line 74: `skew_ratio` seems too general, could you rename it to a more specified one like `tidb_opt_eq_estimation_skew_ratio` or something like this.

qw4990 (approved):
Left 2 small comments, thanks, the rest LGTM.

pkg/planner/cardinality/row_count_index.go
terry1purcell: Done

pkg/sessionctx/variable/setvar_affect.go
terry1purcell: Changed to tidb_opt_risk_eq_skew_ratio - to be more descriptive (as suggested), but also to align with potential for naming of other "risk" related variables.

hawkingrei (approved):

yudongusa (approved):
Please also create a document PR.

Issue #60649, opened 2025-04-18
Make fast admin check table support multi-value index


joechenrh:
## Enhancement

https://github.com/pingcap/tidb/pull/43694 has support fast table check, but we need to further extend it to support multi-value index.

PR #60728, opened 2025-04-23
expression: introduce new function for JSON array

joechenrh:

Issue Number: ref https://github.com/pingcap/tidb/issues/60649

Problem Summary:

To support fast data consistency check for MV Index, we have to add a new function for JSON array. More detail is shown in https://github.com/pingcap/tidb/issues/60649.

Add a new function `JSON_SUM_CRC32`, which is used to process JSON array, and the output will be **sum of `CRC32` of each element in the array**.

The syntax (as well as the implement) is almost the same as `CAST`, for example, the below case should output `CRC32(4) + CRC32(2) + CRC32(1)`.

``` SQL
mysql> CREATE TABLE t(i INT PRIMARY KEY, j JSON);

mysql> INSERT INTO t VALUES(2, "[4, 2, 1]");

mysql> SELECT JSON_SUM_CRC32(j AS UNSIGNED ARRAY) as sum_crc32 from t;
+------------+
| sum_crc32  |
+------------+
| 6751308028 |
+------------+
1 row in set (0.01 sec)

mysql> select CRC32(1)+CRC32(2)+CRC32(4) as sum_crc32;
+------------+
| sum_crc32  |
+------------+
| 6751308028 |
+------------+
1 row in set (0.01 sec)
```

The reason why we keep the syntax same as `CAST` is that kernel side has no information about the type of array, so we have to manually set the type. If the mismatched type is found during execution, an error will be returned.

```SQL
mysql> CREATE TABLE charTable(i INT, j JSON, INDEX mvi((CAST(j AS CHAR(10) ARRAY))));
mysql> INSERT INTO charTable VALUES (2, '["a"]'), (4, null), (3, '["c"]');
mysql> SELECT JSON_SUM_CRC32(j AS DOUBLE ARRAY) FROM charTable;                    
ERROR 3903 (22018): Invalid JSON value for CAST to type double
```

### Check List

Tests 

- [X] Unit test
- [X] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

Although this PR is for adding new function (as well as new syntax), since this function is intended for internal use, more specifically, admin check table, we don't add it to document for now.

### Release note

```release-note
None
```


pkg/expression/builtin_json.go
```diff
@@ -180,6 +184,112 @@ func (b *builtinJSONTypeSig) evalString(ctx EvalContext, row chunk.Row) (val str
 	return j.Type(), false, nil
 }
 
+type jsonSumFunctionClass struct {
```
YangKeao, line 187: `jsonSumCRC32FunctionClass`

pkg/expression/builtin_json.go
```diff
@@ -180,6 +184,112 @@ func (b *builtinJSONTypeSig) evalString(ctx EvalContext, row chunk.Row) (val str
...
+
+	bf, err := newBaseBuiltinFunc(ctx, c.funcName, args, c.tp)
+	if err != nil {
+		return nil, err
+	}
+	sig = &builtinJSONSumSig{bf}
+	return sig, nil
+}
+
+type builtinJSONSumSig struct {
```
YangKeao, line 229: ditto

YangKeao (commented):
Rest LGTM

pkg/parser/misc.go
```diff
@@ -489,6 +489,7 @@ var tokenMap = map[string]int{
 	"JOB":                      job,
 	"JOBS":                     jobs,
 	"JOIN":                     join,
+	"JSON_SUM_CRC32":           jsonSumCrc32,
```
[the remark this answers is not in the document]
joechenrh: Because this function is not designed to be used externally, and I don't want to add it to the document. So I think name w/w.o prefix are both OK.

hawkingrei (approved):
LGTM for the planner part

YangKeao (approved):

pkg/expression/builtin_json.go
```diff
@@ -180,6 +184,112 @@ func (b *builtinJSONTypeSig) evalString(ctx EvalContext, row chunk.Row) (val str
...
+	tp *types.FieldType
+}
+
+func (c *jsonSumCRC32FunctionClass) verifyArgs(ctx EvalContext, args []Expression) error {
+	if err := c.baseFunctionClass.verifyArgs(args); err != nil {
+		return err
+	}
+
+	if args[0].GetType(ctx).EvalType() != types.ETJson {
+		return ErrInvalidTypeForJSON.GenWithStackByArgs(1, "JSON_SUM_CRC32;")
```
XuHuaiyu, line 199: Is the `;` necessary?

joechenrh: Nope, I missed this error Orz, fixed.

XuHuaiyu (approved):

Frank945946 (approved):

tangenta (approved):

pkg/expression/builtin_json.go
```diff
@@ -180,6 +184,112 @@ func (b *builtinJSONTypeSig) evalString(ctx EvalContext, row chunk.Row) (val str
...
+	return nil
+}
+
+func (c *jsonSumCRC32FunctionClass) getFunction(ctx BuildContext, args []Expression) (sig builtinFunc, err error) {
+	if err := c.verifyArgs(ctx.GetEvalCtx(), args); err != nil {
+		return nil, err
+	}
+	arrayType := c.tp.ArrayType()
+	switch arrayType.GetType() {
+	case mysql.TypeYear, mysql.TypeJSON, mysql.TypeFloat, mysql.TypeNewDecimal:
```
Benjamin2037, line 211: Why this switch only has one case？ if so can we just use 'if' statement to replace?

joechenrh: Wait, there are no function to check if the given type is one of those four types like:
```go
func IsTypeTime(tp byte) bool {
	return tp == mysql.TypeDatetime || tp == mysql.TypeDate || tp == mysql.TypeTimestamp
}
```
Using `if` makes the code a bit ugly IMO, so let's just leave it as it was.

yudongusa (approved):
Please also open document PR

PR #60781, opened 2025-04-23
planner: gopls modernize

### LINKED ISSUE #60780: modernize pkg/planner, gopls modernize
## Enhancement
run `cd pkg/planner ; go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix -test ./...`

mjonss:

Issue Number: close #60780

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > - [x] no logic changed, existing tests should be enough

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/operator/logicalop/logical_apply.go
```diff
@@ -136,9 +139,7 @@ func (la *LogicalApply) DeriveStats(childStats []*property.StatsInfo, selfSchema
 		RowCount: leftProfile.RowCount,
 		ColNDVs:  make(map[int64]float64, selfSchema.Len()),
 	})
-	for id, c := range leftProfile.ColNDVs {
```
hawkingrei, line 139: I have do this linter at https://github.com/pingcap/tidb/pull/60406. I find this changement will create bug. so we can revert it.

mjonss: I have reverted it and added a comment.

pkg/planner/cascades/old/transformation_rules.go
```diff
@@ -1327,6 +1316,7 @@ func (*PushTopNDownProjection) OnTransform(old *memo.ExprIter) (newExprs []*memo
 	for i := len(newTopN.ByItems) - 1; i >= 0; i-- {
 		switch newTopN.ByItems[i].Expr.(type) {
 		case *expression.Constant, *expression.CorrelatedColumn:
+			// TODO: investigate why we cannot use: topN.ByItems = slices.Delete(newTopN.ByItems, i, i+1)
```
Rustin170506, line 1319: Very interesting issue. May I ask what happens if we change it here?

I suspect it might be caused by the behavior shown in this example: https://play.golang.com/p/Tem8vEmY5U3.

mjonss: Oh, I missed that it was not the same slice that was assigned and appended... Good catch. I will see if I can fix it.

mjonss: We should probably use slices.DeleteFunc() instead of looping and deleting in the loop.

mjonss: @Rustin170506 Should `topN.ByItems` also be set? Or is that an unknown bug since before?

pkg/planner/cascades/old/testdata/transformation_rules_suite_out.json
```diff
@@ -496,7 +496,7 @@
           "Group#6 Schema:[test.t.a,Column#25]",
           "    Projection_5 input:[Group#7], test.t.a, test.t.b->Column#25",
           "Group#7 Schema:[test.t.a,test.t.b]",
-          "    TopN_14 input:[Group#8], test.t.b, offset:0, count:1",
+          "    TopN_14 input:[Group#8], , offset:0, count:1",
```
mjonss, line 499: @Rustin170506 is this test result change correct?

pkg/planner/core/task.go
```diff
@@ -2132,7 +2127,7 @@ func RemoveUnnecessaryFirstRow(
 				}
 			}
 			if canOptimize {
-				partialSchema.Columns = append(partialSchema.Columns[:partialCursor], partialSchema.Columns[partialCursor+1:]...)
```
mjonss, line 2135: @Rustin170506 could it be a hidden bug here, if `canOptimize` is true for more than one iterations? I.e. since it loops over a fixed number, and here it will decrease the number of columns by one.

pkg/planner/core/tiflash_selection_late_materialization.go
```diff
@@ -218,6 +218,7 @@ func removeSpecificExprsFromSelection(physicalSelection *PhysicalSelection, expr
 	conditions := physicalSelection.Conditions
 	for i := len(conditions) - 1; i >= 0; i-- {
 		if expression.Contains(physicalSelection.SCtx().GetExprCtx().GetEvalCtx(), exprs, conditions[i]) {
+			// TODO: Investigate why we cannot replace this with slices.Delete(conditions, i, i+1)
```
mjonss, line 221: @mjonss please follow up. Maybe try using `slices.DeleteFunc()` instead?

hawkingrei (approved):

pkg/planner/core/tiflash_selection_late_materialization.go
mjonss: Running `TestFTSSyntax` fails on this. A child plan is using the same slice and will crash, since after delete the length will be the same for the copy, but the end element will be `nil`. Solved by first cloning the slice, and then DeleteFunc and using the result for `physicalSelection.Conditions`.
@Rustin170506 is that common that the same slice is copied in multiple places, so similar bugs may appear in other places?

pkg/planner/cascades/old/testdata/transformation_rules_suite_out.json
Rustin170506: No, I think we shouldn't have any plan changes here.

pkg/planner/core/task.go
Rustin170506: No, I don't think we have this problem here, because we directly pass the partialSchema pointer. So I expect most references to point to the schema itself, rather than to the column slice.
We cloned the schema before passing it to the RemoveUnnecessaryFirstRow function.

0xPoe:
Thanks for working on this!  💚 💙 💜 💛 ❤️

PR #60837, opened 2025-04-25
ddl: support refresh meta ddl

River2000i:

Issue Number: close https://github.com/pingcap/tidb/issues/60924

Problem Summary:
When BR uses Tablemode, which is expected to be set to `RestoreMode` when full restore batch crate tables, including tables created during log backup. However, if an exchange partition occurs during log backup, the table will still be a partition during full restore. In log restoring, the table ID will exchange with partition ID, which exchange by meta kv directly instead of infoschema, so the table mode cannot be modified for the table(since the new table ID is not store in infoschema). In addition to exchange partition, this problem may also occur in operations such as drop table\schema. The root cause is that br will directly restore meta kv, which leads to inconsistency between meta kv and infoschema.

Implement `RefreshMeta()` DDL to update schema version by schema ID and new table ID. So infoschema can be consistent with meta kv when br restore meta kv.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/ddl/table_test.go
```diff
@@ -836,3 +837,32 @@ func TestIssue59238(t *testing.T) {
 	tk.MustExec("alter table t exchange partition p1 with table t1")
 	require.True(t, tk.MustQuery("select distinct create_time from information_schema.partitions where table_name = 't'").Equal(testkit.Rows(rs)))
 }
+
+func TestRefreshMeta(t *testing.T) {
```
Benjamin2037, line 841: What is this test purpose？

River2000i: test `RefreshMeta` basic function, which can update info schema after update table name by txn.

pkg/meta/model/job_args.go
```diff
@@ -1785,3 +1785,22 @@ func GetFinishedModifyColumnArgs(job *Job) (*ModifyColumnArgs, error) {
 	}
 	return getOrDecodeArgsV2[*ModifyColumnArgs](job)
 }
+
+// RefreshMetaArgs is the argument for RefreshMeta.
+type RefreshMetaArgs struct {
+	SchemaID int64 `json:"schema_id,omitempty"`
+	TableID  int64 `json:"table_id,omitempty"`
```
Leavrth, line 1792: Will it be faster if we can specify batch table IDs?

Leavrth (commented):
rest LGTM

pkg/meta/model/job_args.go
Leavrth: Get

Leavrth (approved):

River2000i:
@D3Hunter @wjhuang2016 PTAL

pkg/ddl/table.go
```diff
@@ -1699,3 +1699,25 @@ func onAlterNoCacheTable(jobCtx *jobContext, job *model.Job) (ver int64, err err
 	}
 	return ver, err
 }
+
+func onRefreshMeta(jobCtx *jobContext, job *model.Job) (ver int64, err error) {
```
Tristan1900, line 1703: could we add check here to verify that the info schema doesn't have tableInfo with same table id or name but without restoreMode to be safe that refresh doesn't accidentally override anything unexpected?

Tristan1900 (changes requested):
rest LGTM!

pkg/ddl/table.go
```diff
@@ -1699,3 +1699,25 @@ func onAlterNoCacheTable(jobCtx *jobContext, job *model.Job) (ver int64, err err
...
+	_, err = model.GetRefreshMetaArgs(job)
+	if err != nil {
+		job.State = model.JobStateCancelled
+		return 0, errors.Trace(err)
+	}
```
Tristan1900, line 1708: if underlying table is removed by BR, can we update that to info schema as well. Right now it's assuming underlying TiKV also contain the specified table meta KV.

pkg/ddl/table.go
River2000i: maybe we can skip refresh when tableinfo is exist in info schema and the tablemode isn't restoreMode?

pkg/ddl/table.go
River2000i: > if underlying table is removed by BR

now skip table validate in ddl

wjhuang2016 (approved):

pkg/infoschema/builder.go
```diff
@@ -110,6 +112,64 @@ func (b *Builder) applyCreateTables(m meta.Reader, diff *model.SchemaDiff) ([]in
...
+	// Schema doesn't exist in kv, drop it from infoschema.
+	if dbInfo == nil {
+		schemaDiff := &model.SchemaDiff{
+			Version:  diff.Version,
+			Type:     model.ActionDropSchema,
+			SchemaID: schemaID,
+		}
+		return applyDropSchema(b, schemaDiff), nil
+	}
+	// Schema exists in kv but not in infoschema, drop it from infoschema.
```
GMHDBJD, line 132: drop -> create?

GMHDBJD (approved):

pkg/infoschema/builder.go
River2000i: fix in https://github.com/pingcap/tidb/pull/60837/commits/bfeca3c8cf8b1552438f6295cfe4292342ae68a5

Issue #60792, opened 2025-04-23
modernize pkg/store, gopls modernize


mjonss:
## Enhancement
run `cd pkg/store ; go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix -test ./...`

PR #60922, opened 2025-04-28
store: nogo intrange linter

mjonss:

Issue Number: ref #60792

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > - [x] no logic changed, existing tests should be enough

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/build/nogo_config.json
+++ b/build/nogo_config.json
@@ -368,6 +368,7 @@
       "pkg/planner/cardinality": "pkg/planner/cardinality code",
       "pkg/planner/funcdep": "pkg/planner/funcdep code",
       "pkg/server/": "server code",
+      "pkg/store/": "store code",
       "pkg/statistics": "pkg/statistics code"
     }
   },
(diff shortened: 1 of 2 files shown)


zimulala (approved):

pkg/store/mockstore/unistore/tikv/mock_region.go
```diff
@@ -513,6 +513,7 @@ func (rm *MockRegionManager) splitKeys(keys [][]byte) ([]*regionCtx, error) {
 
 		var i int
 		// Cannot be: i = range keys, probably the scope of i is different then
+		// nolint:intrange
```
lance6716, line 516: why not fix it in this PR 😂

mjonss: The [issue](https://go.dev/play/p/dDQr2Dm04vM) is that `for i = range N`, `i` will end up as N-1, vs `for i = 0; i < N; i++`, where i will end up as N. I will see how to fix it...

mjonss: So "fixing" it would make the code less readable :(
like it would be needed to add something like `fullRange` variable:
```
fullRange := true
for i = range keys {
  if ... {
    fullRange = false
    break
  }
}
if fullRange {
  i = len(keys)
}
splits := keys[:i]
keys = keys[i:]
```

mjonss: Hmm, I just moved the splits and keys into the loop instead, should work :)

lance6716: Oh I just though `for i = range N` is OK without seeing the end value problem 😂 However your really find an elegant way to fix it.

PR #61023, opened 2025-05-08
planner: distinguish the index join probe as indexScan or tableScan from indexJoinProp side.

AilinKid:

Problem Summary:

* previously the tidb will enumerate and build index join‘s inner task as tableRangeScan and indexRangeScan separately, which means two kinds of this index join will compare their total cost from indexJoin itself. 

* obviously, one indexJoinProp couldn't output two kinds of underlying cop tasks, so we just add one more field in the indexJoinProp to guide the underlying datasource to build separate copTask(tableRangeScan or indexRangeScan) out, then leverage the attack2Task to wrap them as TableReader or IndexReader or IndexLookUpReader just like old logic does, consequently, leading to different physical index join and the cost comparing just between them.

* some small fix: 
    * clone function should check whether the inner plan is nil or not. because in the v2, we didn't need this field anymore.
    * innerTask should be checked as nil or not; if yes, we should return an invalid task directly.
    * the hint handling should be able to see the direct built out physical plan, rather then Enforcer operator like sort or exchanger.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/exhaust_physical_plans.go
+++ b/pkg/planner/core/exhaust_physical_plans.go
@@ -1024,14 +1024,26 @@ func enumerateIndexJoinByOuterIdx(p *logicalop.LogicalJoin, prop *property.Physi
 	if count := outerChild.StatsInfo().RowCount; count > 0 {
 		avgInnerRowCnt = p.EqualCondOutCnt / count
 	}
-	indexJoinProp := &property.IndexJoinRuntimeProp{
+	// for pk path
+	indexJoinPropTS := &property.IndexJoinRuntimeProp{
 		OtherConditions: p.OtherConditions,
 		InnerJoinKeys:   innerJoinKeys,
 		OuterJoinKeys:   outerJoinKeys,
 		AvgInnerRowCnt:  avgInnerRowCnt,
+		TableRangeScan:  true,
 	}
-	indexJoins := constructIndexJoinStatic(p, prop, outerIdx, indexJoinProp)
-	indexJoins = append(indexJoins, constructIndexHashJoinStatic(p, prop, outerIdx, indexJoinProp)...)
+	// for normal index path
+	indexJoinPropIS := &property.IndexJoinRuntimeProp{
+		OtherConditions: p.OtherConditions,
+		InnerJoinKeys:   innerJoinKeys,
+		OuterJoinKeys:   outerJoinKeys,
+		AvgInnerRowCnt:  avgInnerRowCnt,
+		TableRangeScan:  false,
+	}
+	indexJoins := constructIndexJoinStatic(p, prop, outerIdx, indexJoinPropTS)
+	indexJoins = append(indexJoins, constructIndexJoinStatic(p, prop, outerIdx, indexJoinPropIS)...)
+	indexJoins = append(indexJoins, constructIndexHashJoinStatic(p, prop, outerIdx, indexJoinPropTS)...)
+	indexJoins = append(indexJoins, constructIndexHashJoinStatic(p, prop, outerIdx, indexJoinPropIS)...)
 	return indexJoins
 }
 
@@ -1190,10 +1202,14 @@ func buildDataSource2IndexScanByIndexJoinProp(
 		}
 	}
 	var innerTask base.Task
-	if isMatchProp(ds, indexJoinResult.chosenPath, prop) {
-		innerTask = constructDS2IndexScanTask(ds, indexJoinResult.chosenPath, indexJoinResult.chosenRanges.Range(), indexJoinResult.chosenRemained, indexJoinResult.idxOff2KeyOff, rangeInfo, true, !prop.IsSortItemEmpty() && prop.SortItems[0].Desc, prop.IndexJoinProp.AvgInnerRowCnt, maxOneRow)
+	if !prop.IsSortItemEmpty() && isMatchProp(ds, indexJoinResult.chosenPath, prop) {
+		innerTask = constructDS2IndexScanTask(ds, indexJoinResult.chosenPath, indexJoinResult.chosenRanges.Range(), indexJoinResult.chosenRemained, indexJoinResult.idxOff2KeyOff, rangeInfo, true, prop.SortItems[0].Desc, prop.IndexJoinProp.AvgInnerRowCnt, maxOneRow)
 	} else {
 		innerTask = constructDS2IndexScanTask(ds, indexJoinResult.chosenPath, indexJoinResult.chosenRanges.Range(), indexJoinResult.chosenRemained, indexJoinResult.idxOff2KeyOff, rangeInfo, false, false, prop.IndexJoinProp.AvgInnerRowCnt, maxOneRow)
+	}
+	// since there is a possibility that inner task can't be built and the returned value is nil, we just return base.InvalidTask.
+	if innerTask == nil {
+		return base.InvalidTask
 	}
 	// prepare the index path chosen information and wrap them as IndexJoinInfo and fill back to CopTask.
 	// here we don't need to construct physical index join here anymore, because we will encapsulate it bottom-up.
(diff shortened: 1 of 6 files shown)


pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -2083,8 +2103,8 @@ func tryToEnumerateIndexJoin(p *logicalop.LogicalJoin, prop *property.PhysicalPr
 		stmtCtx.SetHintWarning("Some INL_MERGE_JOIN and NO_INDEX_MERGE_JOIN hints conflict, NO_INDEX_MERGE_JOIN may be ignored")
 	}
 	// previously we will think about force index join hints here, but we have to wait the inner plans to be a valid
-	// physical one/ones. Because indexJoinProp may not be admitted by its inner patterns, so we innovatively move all
-	// hint related handling to the findBestTask function when we see the entire inner physicalized plan tree. See xxx
+	// physical one/ones. Because indexJoinProp may not be admitted by its inner patterns, so we innovative-ly move all
```
hawkingrei, line 2106:
```suggestion
	// physical one/ones. Because indexJoinProp may not be admitted by its inner patterns, so we innovatively move all
```

hawkingrei (approved):

tests/integrationtest/r/planner/core/casetest/physicalplantest/physical_plan.result
```diff
@@ -1853,11 +1853,11 @@ explain format = 'brief' select /*+ hash_join_probe(t1) */ sum(t1.a in (select a
 id	estRows	task	access object	operator info
 HashAgg	1.00	root		funcs:sum(Column#9)->Column#8
 └─Projection	10000.00	root		cast(Column#7, decimal(3,0) BINARY)->Column#9
-  └─MergeJoin	10000.00	root		left outer semi join, left side:TableReader, left key:planner__core__casetest__physicalplantest__physical_plan.t1.a, right key:planner__core__casetest__physicalplantest__physical_plan.t2.a
+  └─HashJoin	10000.00	root		left outer semi join, left side:TableReader, equal:[eq(planner__core__casetest__physicalplantest__physical_plan.t1.a, planner__core__casetest__physicalplantest__physical_plan.t2.a)]
```
AilinKid, line 1856: This change is reasonable, because the hash join probe hint is specified, while the previous result doesn't respect to that

pkg/planner/core/index_join_path.go
```diff
@@ -634,35 +633,29 @@ func getIndexJoinIntPKPathInfo(ds *logicalop.DataSource, innerJoinKeys, outerJoi
...
 	// reason2: the ranges from TS and IS couldn't be directly used to derive the stats' estimation, it's not real.
 	// reason3: skyline pruning should not prune the possible index path which could feel the runtime EQ access conditions.
-	innerTSCopTask := buildDataSource2TableScanByIndexJoinProp(ds, prop)
-	if !innerTSCopTask.Invalid() {
-		planCounter.Dec(1)
-		if planCounter.Empty() {
-			// planCounter is counted to end, just return this one.
-			return innerTSCopTask, 1, nil
+	//
+	// here we build TS and IS separately according to different index join prop is for we couldn't decide
```
hawkingrei, line 642:
```suggestion
	// here we build TableScan(TS) and IndexScan(IS) separately according to different index join prop is for we couldn't decide
```

hawkingrei (approved):

pkg/planner/core/physical_plans.go
```diff
@@ -1678,7 +1678,11 @@ func (p *PhysicalIndexJoin) Clone(newCtx base.PlanContext) (base.PhysicalPlan, e
 		return nil, err
 	}
 	cloned.basePhysicalJoin = *base
-	cloned.innerPlan, err = p.innerPlan.Clone(newCtx)
+	// after the new refactor of indexJoin, the inner plan field is not a necessary one.
```
time-and-fate, line 1681: It's hard for other people to know what "the new refactor" means, and why it becomes unnecessary.
It's better to specifically explain why it could be `nil`, or just remove the comments.

pkg/planner/core/physical_plans.go
```diff
@@ -1678,7 +1678,11 @@ func (p *PhysicalIndexJoin) Clone(newCtx base.PlanContext) (base.PhysicalPlan, e
...
+	// So we need to check if the inner plan is nil before we do some cloning.
+	if p.innerPlan != nil {
+		cloned.innerPlan, err = p.innerPlan.Clone(newCtx)
```
time-and-fate, line 1684: The error handling should be in the `if` block.

pkg/planner/core/find_best_task.go
```diff
@@ -1456,11 +1461,16 @@ func findBestTask4LogicalDataSource(lp base.LogicalPlan, prop *property.Physical
 			// even enforce hint can not work with this.
 			return base.InvalidTask, 0, nil
 		}
+		// cache the physical for indexJoinProp
+		defer func() {
+			ds.StoreTask(prop, t)
+			err = validateTableSamplePlan(ds, t, err)
```
time-and-fate, line 1467: Do we need this? I think there will never be a `TableSample` in this case.

pkg/planner/core/physical_plans.go
qw4990: +1

pkg/planner/core/index_join_path.go
```diff
@@ -634,35 +633,29 @@ func getIndexJoinIntPKPathInfo(ds *logicalop.DataSource, innerJoinKeys, outerJoi
...
-		return base.InvalidTask, 0, err
-	}
-	if leftIsBetter {
 		return innerTSCopTask, 1, nil
 	}
+	innerISCopTask := buildDataSource2IndexScanByIndexJoinProp(ds, prop)
+	if innerISCopTask.Invalid() {
+		return base.InvalidTask, 0, nil
+	}
+	planCounter.Dec(1)
```
qw4990, line 658: How about removing `planCounter` later? No one is using this, and it'll increase the complexity of our code.

qw4990 (approved):

pkg/planner/core/find_best_task.go
AilinKid: good catch, make sense

pkg/planner/core/index_join_path.go
AilinKid: This comes from a nth plan-related test failure, and it was added. I will remove it systematically after that.

pkg/planner/core/physical_plans.go
AilinKid: sure, make sense

pkg/planner/core/index_join_path.go
```diff
@@ -634,35 +633,29 @@ func getIndexJoinIntPKPathInfo(ds *logicalop.DataSource, innerJoinKeys, outerJoi
...
-		if planCounter.Empty() {
-			// planCounter is counted to end, just return this one.
-			return innerTSCopTask, 1, nil
+	//
+	// here we build TableScan(TS) and IndexScan(IS) separately according to different index join prop is for we couldn't decide
+	// which one as the copTask here is better, some more possible upper attached operator cost should be
+	// considered, besides the row count, double reader cost for index lookup should also be considered as
+	// a whole, so we leave the cost compare for index join itself just like what it was before.
+	if prop.IndexJoinProp.TableRangeScan {
+		innerTSCopTask := buildDataSource2TableScanByIndexJoinProp(ds, prop)
```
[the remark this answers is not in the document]
AilinKid: wo! you are right, catched

PR #61057, opened 2025-05-09
txn: also update the `@@tidb_last_txn_info` for readyonly or rollback txns

### LINKED ISSUE #61056: also update `@@tidb_last_txn_info` for readonly and rollback txn
## Enhancement

Currently, the `@@tidb_last_txn_info` will be updated when txn is committed. However, if a txn does not have any updates, i.e., a readonly txn, it will keep `@@tidb_last_txn_info` unchanged.

Maybe it's better to also update `@@tidb_last_txn_info` for readonly transactions to have a better tracing.

lcwangchao:

Issue Number: close #61056

also update the `@@tidb_last_txn_info` for readyonly or rollback transactions

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
@@tidb_last_txn_info will also be updated when readonly transaction ends or transaction rollback
```

### DIFF
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -919,6 +919,7 @@ func (s *session) CommitTxn(ctx context.Context) error {
 	r, ctx := tracing.StartRegionEx(ctx, "session.CommitTxn")
 	defer r.End()
 
+	s.setLastTxnInfoBeforeTxnEnd()
 	var commitDetail *tikvutil.CommitDetails
 	ctx = context.WithValue(ctx, tikvutil.CommitDetailCtxKey, &commitDetail)
 	err := s.doCommitWithRetry(ctx)
@@ -957,6 +958,7 @@ func (s *session) RollbackTxn(ctx context.Context) {
 	r, ctx := tracing.StartRegionEx(ctx, "session.RollbackTxn")
 	defer r.End()
 
+	s.setLastTxnInfoBeforeTxnEnd()
 	if s.txn.Valid() {
 		terror.Log(s.txn.Rollback())
 	}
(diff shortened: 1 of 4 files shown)


pkg/session/session.go
```diff
@@ -919,6 +919,7 @@ func (s *session) CommitTxn(ctx context.Context) error {
 	r, ctx := tracing.StartRegionEx(ctx, "session.CommitTxn")
 	defer r.End()
 
+	s.setLastTxnInfoBeforeTxnEnd()
```
ekexium, line 922: How about putting it in `onTxnEnd`?

lcwangchao: Seems onTxnEnd，the txn info has been cleared

MyonKeminta: After this line, in `doCommitWithRetry` -> `doCommit`, it seems it still have possibility to finally enter a rolled-back state or encounter error. Perhaps it's better to consider put it in a later place where the commit/rollback result is finally determined.

pkg/session/session.go
```diff
@@ -970,6 +972,24 @@ func (s *session) RollbackTxn(ctx context.Context) {
...
+	txnCtx := s.GetSessionVars().TxnCtx
+	if txnCtx.StartTS == 0 {
+		// If the txn is not active, for example, executing "SELECT 1", skip setting the last txn info.
+		return
+	}
+
+	s.GetSessionVars().LastTxnInfo = fmt.Sprintf(
+		`{"txn_scope": "%s", "start_ts": %d, "for_update_ts": %d}`,
+		txnCtx.TxnScope, txnCtx.StartTS, txnCtx.GetForUpdateTS(),
+	)
```
MyonKeminta, line 990: Consider using json serialization instead of text formatting to prevent potential mistakes. See the original logic for setting last txn info, which is located in client-go: https://github.com/tikv/client-go/blob/2058fbc06214bce4a71986d62a6ada9e89d013ff/txnkv/transaction/txn.go#L1033-L1048

lcwangchao: Seems if a txn comitted or fails to commit, the CommitHook will update it: https://github.com/pingcap/tidb/blob/c71ebaf73ed9620745f327ad147f27fc06d73297/pkg/sessiontxn/isolation/base.go#L506

`setLastTxnInfoBeforeTxnEnd` in `CommitTxn` will be overridden at this time. So `setLastTxnInfoBeforeTxnEnd` is only affects some readonly txns.

pkg/session/session.go
lcwangchao: Seems if a txn comitted or fails to commit, the CommitHook will update it: https://github.com/pingcap/tidb/blob/c71ebaf73ed9620745f327ad147f27fc06d73297/pkg/sessiontxn/isolation/base.go#L506

`setLastTxnInfoBeforeTxnEnd` in `CommitTxn` will be overridden at this time. So `setLastTxnInfoBeforeTxnEnd` is only affects some readonly txns.

pkg/session/session.go
```diff
@@ -970,6 +973,26 @@ func (s *session) RollbackTxn(ctx context.Context) {
 	sessiontxn.GetTxnManager(s).OnTxnEnd()
 }
 
+// setLastTxnInfoBeforeTxnEnd sets the @@last_txn_info variable before commit/rollback the transaction.
+// The `LastTxnInfo` updated with a JSON string that contains start_ts, for_update_ts, etc.
+// It updated `LastTxnInfo` does not contain the `commit_ts` fields because it is unknown
```
cfzjywxk, line 978:
```suggestion
// The `LastTxnInfo` is updated without the `commit_ts` fields because it is unknown
```

pkg/session/session.go
```diff
@@ -970,6 +973,26 @@ func (s *session) RollbackTxn(ctx context.Context) {
...
+// until the commit is done (or do not need to commit for readonly or a rollback transaction).
+// The non-readonly transaction will overwrite the `LastTxnInfo` again after commit to update the `commit_ts` field.
+func (s *session) setLastTxnInfoBeforeTxnEnd() {
```
cfzjywxk, line 981: Should we rename it to `setLastTxnInfoForReadonlyTxn`?

lcwangchao: The on-readonly transaction also calls this method before committing, and this will be overwritten.

cfzjywxk (approved):
LGTM

zyguan (approved):

PR #61119, opened 2025-05-14
planner: a new function to get all relevant optimizer variables related to a certain SQL for SPM

qw4990:

Issue Number: ref #60148

Problem Summary: planner: a new function to get all relevant optimizer variables related to a certain SQL for SPM

Add a new function `relevantOptVars` which can get all relevant optimizer variables like `fix-control`, `cost-factors` for a certain query. 
SPM can use this function to get all variables and generate more new plans via adjusting these variables.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/bindinfo/binding_auto.go
```diff
@@ -33,6 +33,9 @@ import (
 // PlanDigestFunc is used to get the plan digest of this SQL.
 var PlanDigestFunc func(sctx sessionctx.Context, stmt ast.StmtNode) (planDigest string, err error)
 
+// RelevantOptVars is used to get the relevant optimizer variables for this SQL.
```
henrybw, line 36: This is a nitpick, but perhaps we should note in the comment that this function calls the optimizer? (Although, in fairness, the comment for `PlanDigestFunc` above doesn't note this either). To me, "get the relevant optimizer variables" implies a lightweight read operation, similar to what I would expect a `SHOW` query to do. Based on this description, I would be a bit surprised to find out that this runs the optimizer.

That said, feel free to disagree with my opinion here: this is just based on what I (a relative newcomer to TiDB) would expect reading this comment.

pkg/bindinfo/binding_auto.go
```diff
@@ -33,6 +33,9 @@ import (
...
+var RelevantOptVars func(sctx sessionctx.Context, stmt ast.StmtNode) (varNames []string, err error)
```
henrybw, line 37: This is a nitpick as well, but how about naming this something like `RecordRelevantOptVars` instead? Adding the verb, to me, makes it clearer that this is a bit of a heavy operation. To me, `RelevantOptVars` implies a lookup (like how `GetSessionVars()` returns a map that was previously constructed), so it's a bit surprising that calling this function records the variables as well.

But again, feel free to leave this as-is if you disagree with me.

pkg/planner/optimize.go
```diff
@@ -654,6 +655,38 @@ func planDigestFunc(sctx sessionctx.Context, stmt ast.StmtNode) (planDigest stri
...
+
+	_, _, err = Optimize(context.Background(), sctx, nodeW, sctx.GetDomainInfoSchema().(infoschema.InfoSchema))
+	if err != nil {
+		return nil, err
+	}
+
+	for varName := range sctx.GetSessionVars().RelevantOptVars {
+		varNames = append(varNames, varName)
+	}
+	sort.Strings(varNames)
```
henrybw, line 686: Out of curiosity, will SPM expect this to be sorted? (Not a big deal either way, but if this is just to keep the output stable for testability, I was thinking we could move the sorting into the test case instead.)

pkg/planner/optimize.go
```diff
@@ -654,6 +655,38 @@ func planDigestFunc(sctx sessionctx.Context, stmt ast.StmtNode) (planDigest stri
 	return digest.String(), nil
 }
 
+func relevantOptVars(sctx sessionctx.Context, stmt ast.StmtNode) (varNames []string, err error) {
+	sctx.GetSessionVars().RecordRelevantOptVars = true
```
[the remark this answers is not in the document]
qw4990: Oh you are right, I'll update this place to set it to `nil` for safety.

pkg/planner/optimize.go
henrybw: Got it, makes sense!

pkg/bindinfo/binding_auto.go
qw4990: Thank you for this suggestion! I've updated the comments and renamed this function. And I also renamed "PlanDigestFunc" to "CalculatePlanDigest".

pkg/planner/core/exhaust_physical_plans.go
```diff
@@ -1798,6 +1798,7 @@ func constructDS2IndexScanTask(
 	// We can calculate the lower bound of the NDV therefore we can get an upper bound of the row count here.
 	rowCountUpperBound := -1.0
 	fixControlOK := fixcontrol.GetBoolWithDefault(ds.SCtx().GetSessionVars().GetOptimizerFixControlMap(), fixcontrol.Fix44855, false)
+	ds.SCtx().GetSessionVars().RecordRelevantOptVar("Fix44855")
```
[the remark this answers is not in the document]
qw4990: Thanks, I've updated this part of code, which seems safer than before.

terry1purcell (approved):
/lgtm

henrybw (approved):
Thank you for making those changes! LGTM

yudongusa (approved):

Issue #49491, opened 2023-12-15
Refractor the usage of session context in `executor`


YangKeao:
# Current situation

A normal routine of using pkg/executor package is to call `(*ExecStmt).Exec` or `(*ExecStmt).PointGet`. The session context is wrapped inside the `ExecStmt`. The method call `(*ExecStmt).Exec` will return a record set. The rows are dumped by calling `RecordSet.Next`. If everything goes well, the RecordSet will be closed after draining the rows. 
During this process, the session context exists in several places:

1. `ExecStmt.Ctx`
2. `executorBuilder.ctx`
3. `BaseExecutor` and some other related structures inside executors (e.g. `FKCheckExec`...)

It would be suggested to fix them from bottom to top, as the context required by the bottom part is always a subset of the context required by the upper part. The process of refractoring the bottom part is squeezing the usage of session context to the upper part.

# Plan

## Executor

From the time when the context is needed, they can also be split into two parts:

1. In build and open. Though most of the time build should be seen as part of `executorBuilder`.
2. In the `Next()` function. Some of the context currently needed by the `Next()` function can also be migrated to build / open parts and stored inside the executor.

We can use different parameters to build each executor actually, and give them different context needed by them. The context needed by Open can also be set during build, because Open and build are always run in the same "statement", and the context should be the same when calling them. 

However, we should have a unified interface to actually run different kinds of executors, so for the `Next()` function we'll need the unified interface.

## Target

After refractoring, the `BaseExecutor` shouldn't contain the session context (and so does any other fields of an executor). All context needed by the build and open should be passed and stored in the executor in the build stage. `BaseExecutor` can still be used to store some information needed by all executors, like `memTracker`, but it's suggested to keep it tidy and leave more configurations (e.g. timezone) to the field of each executor itself.

We can also have part (one low level) of the expression.Context in some executors.

The `Next()` function should accept an interface:

```go
Next(ctx context.Context, ectx Context, req *chunk.Chunk) error
```

The ectx should contain all information needed by the executor to run. 

```go
type Context interface {
    ... // TODO. 
    // It's also possible that the executor may don't need any context for `Next` function. We can have a conclusion after refractoring / analyzing each executors. It doesn't really matter.
    // Also, we should keep this interface as small as possible.
}
```

For example, to solve the cursor fetch situation, we'll need to store the expression context in executor.

```go
type ProjectionExec struct {
    evalContext expression.Context
    ....
} 
```

[...]

PR #61182, opened 2025-05-19
*: remove `StatementContext` from `CommonHandleCols` to fix a bug caused by shallow clone in plan cache

### LINKED ISSUE #61118: Instance level plan cache may cause "ERROR 8141 assertion failed" in UPDATE stmt under high concurrency
## Bug Report

#### Code changes

To make it reproducible without high concurrency, we need to introduce latency in the code.

https://github.com/pingcap/tidb/blob/7c0c2b16f965d7f5c16065aa28ece2a7718bff75/pkg/executor/select.go#L891-L899

Modify the code here as below:

```diff
	var sc *stmtctx.StatementContext
	if vars.TxnCtx.CouldRetry || vars.HasStatusFlag(mysql.ServerStatusCursorExists) {
		// Must construct new statement context object, the retry history need context for every statement.
		// TODO: Maybe one day we can get rid of transaction retry, then this logic can be deleted.
		sc = stmtctx.NewStmtCtx()
	} else {
		sc = vars.InitStatementContext()
	}
+	if !vars.InRestrictedSQL {
+		time.Sleep(1 * time.Second)
+	}
	sc.SetTimeZone(vars.Location())
```

The patch would make the SQL execution slow, be patient~

#### Reproduction Steps

1. Preparation

```sql
set global tidb_enable_instance_plan_cache = 1;
create table t(a timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP, b int, c int, primary key(a), unique key(b,c));
insert into t(b,c) value (1,1);
insert into t(b,c) value (2,2);
```

2. Trigger

We need prepare two sessions (connections) here.

In session 1:

```sql
prepare stmt from 'update t set a = NOW() where b = ? and c = ?';
set @a = 1;
execute stmt using @a, @a;
execute stmt using @a, @a;
```

In session 2:

```sql
prepare stmt from 'update t set a = NOW() where b = ? and c = ?';
set @a = 1;
```

Then execute the SQL below first in session 2, then immediately in session 1. This needs to be very fast.

```sql
execute stmt using @a, @a;
```

No errors.

```
ERROR 8141 (HY000): assertion failed: key: 7480000000000000705f720419b69d6e07000000, assertion: Exist, start_ts: 458027064388747265, existing start ts: 0, existing commit ts: 0
```

v8.5.0, v9.0.0-beta.1, and the latest nightly

time-and-fate:

Issue Number: close #61118

Problem Summary:

See the [comments](https://github.com/pingcap/tidb/issues/61118#issuecomment-2881297944) in the issue for details.

In brief, the `CommonHandleCols` contains `StatementContext`, and the `Update` operator may contain `CommonHandleCols` in `Update.TblColPosInfos`.
The `CommonHandleCols` and the `StatementContext` inside will be passed to `UpdateExec` and used during execution, which causes the bug when we reuse the plan in another session.

In the issue, we mentioned replacing `StatementContext` in the `CommonHandleCols` with a `types.Context`. However, it turns out that the `warnHandler` in it is still a `StatementContext`, so this doesn't really remove it.

So, it's probably better to remove `StatementContext` from `CommonHandleCols` directly and pass it when it's needed. This rationale is also consistent with [previous refactoring](https://github.com/pingcap/tidb/issues/49491) related to the usage of `Context` in the executor.

Changes:
- `pkg/planner/util/handle_cols.go`
    - Remove `StatementContext` from `CommonHandleCols`.
    - Remove the `StatementContext` parameter for functions that construct `CommonHandlesCols`.
    - Add a `StatementContext` or `types.Context` parameter for functions that use it
- Other changes are for adapting to the new function signatures.
    - For many places in `executor/`, we need to pass a `StatementContext` now.
    - For many places in `planner/`, we don't need to pass `StatementContext` anymore.
- A test case is added in `pkg/planner/core/issuetest/planner_issue_test.go`.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


time-and-fate:
/hold
Waiting for everyone to finish review

pkg/planner/util/handle_cols.go
```diff
@@ -77,11 +77,10 @@ type CommonHandleCols struct {
 	tblInfo *model.TableInfo
 	idxInfo *model.IndexInfo
 	columns []*expression.Column
-	sc      *stmtctx.StatementContext
```
qw4990, line 80: Could you also add some comments here to notify that this structure must be cross-session thread safe?

qw4990 (approved):

pkg/planner/util/handle_cols.go
qw4990: To at least notify others the trick here via comments... and avoid them putting another non-thread-safe structures into it again.

terry1purcell (approved):
/lgtm

pkg/planner/util/handle_cols.go
time-and-fate: updated~

lcwangchao (approved):

PR #61279, opened 2025-05-22
planner: consider leading hints (join order) when exploring plans

qw4990:

Issue Number: ref #60148

Problem Summary: planner: consider leading hints (join order) when exploring plans

Consider leading hints (join order) when exploring plans:
1. extract all table names at the beginning;
2. enumerate all leading-2 pairs;

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/bindinfo/binding_plan_generation.go
```diff
@@ -80,6 +81,15 @@ func (g *planGenerator) Generate(defaultSchema, sql, charset, collation string)
 	return
 }
 
+type tableName struct {
+	schema string
+	name   string
```
hawkingrei, line 86: We can try ```unique.Handle[string]```

https://pkg.go.dev/unique#Handle

qw4990: Emm, since in this scenario "EXPLAIN EXPLORE", performance is not the first priority, then we'd tend to keep our code simpler.

pkg/bindinfo/binding_plan_generation.go
```diff
@@ -114,6 +125,15 @@ type state struct {
 // Encode encodes the state into a string.
 func (s *state) Encode() string {
 	sb := new(strings.Builder)
+	for _, t := range s.leading2 {
```
AilinKid, line 128: encode string will be a burden here, since it's all about string concat.

pkg/bindinfo/binding_plan_generation.go
```diff
@@ -434,3 +509,50 @@ func getStartState(vars []string, fixes []uint64) (*state, error) {
 	}
 	return s, nil
 }
+
+type tableNameExtractor struct {
+	defaultSchema string
+	tableNames    map[string]*tableName
+}
+
```
AilinKid, line 517: export func should have a comment

pkg/bindinfo/binding_plan_generation.go
qw4990: Yes, but it should be OK, since "explain explore" is a lowly-frequency operation here, it can only be triggered by users manually. We want to keep our code as simple as possible at the beginning, and we can optimize this part later when there are real performance problems.

pkg/bindinfo/binding_plan_generation.go
qw4990: Yes... but they all are not exposed... and since these structures are simple and only used internal, this seems OK to re-define them.

AilinKid (approved):

pkg/bindinfo/binding_plan_generation.go
hawkingrei: https://github.com/pingcap/tidb/pull/60230

We actually have one that is being done, but it is really a hard labor

PR #61322, opened 2025-05-26
cgmon: fix the issue that CPU quota is 0 if it's not limited by cgroup

### LINKED ISSUE #61223: TiDB should be able to handle the case when `cgroup.max` doesn't exist and the cgroup config doesn't have cpu controller
## Bug Report

Use `tiup cluster` to deploy a cluster in rockylinux 9.

TiDB doesn't report any error.

TiDB reports error that the `cpu.max` doesn't exist.

YangKeao:

Issue Number: close #61223

Problem Summary:

The cpu quota will always be 0 if it failed to get the cgroup limit. The same issue also exists for memory. 

Change the logic of error handling in cgmon. Make the cpu quota defaults to the count of CPU, and the memory quota defaults to the memory get from `/proc/meminfo`. Then update these values according to the cgroup settings.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

1. Deploy master branch of tidb in rockylinux 9. Open the grafana. The cpu quota is 0.

image

2. Deploy this branch of TiDB in rockylinux 9. Open the grafana. The cpu quota is 16.

image

3. Modify the tiup config of TiDB. Set CPU quota to 800 and memory quota to 8G. The value is correct.

image
image

4. Manually remove the memory controller from related cgroup, so that TiDB will get error when it's reading the memory quota, then the memory quota goes to the memory of the whole machine (16G).

image

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that the cpu quota is always 0 if `cpu.max` doesn't exist
```

### DIFF
--- a/pkg/util/cgmon/cgmon.go
+++ b/pkg/util/cgmon/cgmon.go
@@ -38,9 +38,11 @@ var (
 	ctx             context.Context
 	cancel          context.CancelFunc
 	wg              sync.WaitGroup
-	cfgMaxProcs     int
-	lastMaxProcs    int
+	lastCPU         int
 	lastMemoryLimit uint64
+
+	getCgroupCPUPeriodAndQuota = cgroup.GetCPUPeriodAndQuota
+	getCgroupMemoryLimit       = cgroup.GetMemoryLimit
 )
 
 // StartCgroupMonitor uses to start the cgroup monitoring.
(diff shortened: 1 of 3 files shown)


pkg/util/cgmon/cgmon.go
```diff
@@ -38,7 +38,6 @@ var (
 	ctx             context.Context
 	cancel          context.CancelFunc
 	wg              sync.WaitGroup
-	cfgMaxProcs     int
```
YangKeao, line 41: This variable is not used. I don't know why it was here before.

hawkingrei (approved):

pkg/util/cgmon/cgmon.go
```diff
@@ -115,45 +114,47 @@ func refreshCgroupCPU() error {
...
 	if memLimit != lastMemoryLimit {
 		log.Info("set the memory limit", zap.Uint64("memLimit", memLimit))
 		metrics.MemoryLimit.Set(float64(memLimit))
 		lastMemoryLimit = memLimit
 	}
-	return nil
+	return err
 }
+
+var getCgroupCPUPeriodAndQuota = cgroup.GetCPUPeriodAndQuota
```
hawkingrei, line 159: This kind of variable is best placed at the beginning of the file

Defined2014 (approved):

PR #61340, opened 2025-05-27
statistics: the totalCount/minCount of TopN is calculated only once

### LINKED ISSUE #61339: TotalCount spends too much time
## Enhancement

The issue is as shown in this flame graph.

Image

You can see that calculating the total for topn takes a lot of time, because it needs to traverse topn each time to compute the total count. This is very time-consuming. Moreover, the result of each calculation is the same, so you can cache this result to avoid performance issues.

hawkingrei:

Issue Number: close #61339

Problem Summary:

The issue is as shown in this flame graph.

Image

You can see that calculating the total for topn takes a lot of time, because it needs to traverse topn each time to compute the total count. This is very time-consuming. Moreover, the result of each calculation is the same, so you can cache this result to avoid performance issues.

create an cache for ```totalCount``` and ```minCount``` in the ```TopN```. Use ```sync.Once``` to avoid the situation where a large number of tasks concurrently perform the same calculation during the first computation.
### Check List

Tests 

- [x] Unit test
- [ ] Integration test

Sorry, I cannot public this benchmark code. because it contains the customer information.

before:

```
BenchmarkABC-8   	     206	   6164857 ns/op
```

after:

```
BenchmarkABC-8   	     362	   3142310 ns/op
```

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/builder.go
+++ b/pkg/statistics/builder.go
@@ -466,7 +466,11 @@ func BuildHistAndTopN(
 	}
 
 	topn := &TopN{TopN: topNList}
-	topNTotalCount := topn.Scale(sampleFactor)
+	var topNTotalCount uint64
+	for i := range topn.TopN {
+		topn.TopN[i].Count = uint64(float64(topn.TopN[i].Count) * sampleFactor)
+		topNTotalCount += topn.TopN[i].Count
+	}
 
 	if uint64(count) <= topNTotalCount || int(hg.NDV) <= len(topn.TopN) {
 		// If we've collected everything  - don't create any buckets
--- a/pkg/statistics/builder.go
+++ b/pkg/statistics/builder.go
@@ -466,9 +466,9 @@ func BuildHistAndTopN(
 	}
 
 	topn := &TopN{TopN: topNList}
-	topn.Scale(sampleFactor)
+	topNTotalCount := topn.Scale(sampleFactor)
 
-	if uint64(count) <= topn.TotalCount() || int(hg.NDV) <= len(topn.TopN) {
+	if uint64(count) <= topNTotalCount || int(hg.NDV) <= len(topn.TopN) {
 		// If we've collected everything  - don't create any buckets
 		return hg, topn, nil
 	}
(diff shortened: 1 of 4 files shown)


pkg/statistics/builder.go
```diff
@@ -466,9 +466,9 @@ func BuildHistAndTopN(
 	}
 
 	topn := &TopN{TopN: topNList}
-	topn.Scale(sampleFactor)
+	topNTotalCount := topn.Scale(sampleFactor)
 
-	if uint64(count) <= topn.TotalCount() || int(hg.NDV) <= len(topn.TopN) {
+	if uint64(count) <= topNTotalCount || int(hg.NDV) <= len(topn.TopN) {
```
hawkingrei, line 471: Since ```topn.TotalCount()``` requires traversing the entire dataset, I will obtain the total count during the first ```Scale``` operation, so I don't need to traverse it again.

pkg/statistics/cmsketch.go
```diff
@@ -520,13 +522,23 @@ func (c *CMSketch) CalcDefaultValForAnalyze(ndv uint64) {
...
+	minCount   uint64
+	// minCount and totalCount are initialized only once.
+	once sync.Once
 }
 
-// Scale scales the TopN by the given factor.
-func (c *TopN) Scale(scaleFactor float64) {
+// Scale scales the TopN by the given factor and return the total row count.
+// It is only used when to build topN from sampled data during analyze.
+func (c *TopN) Scale(scaleFactor float64) uint64 {
```
qw4990, line 534: Do we need to call "calculateMinCountAndCount" again after this Scale?

qw4990 (commented):
And should we add any test for this change?

hawkingrei:
> And should we add any test for this change?

I added some assertions for it. During the development process, we actually encountered this kind of issue, just like this

https://tiprow.hawkingrei.com/view/gs/pingcapprow/pr-logs/pull/pingcap_tidb/61340/fast_test_tiprow/1927250012035616768#1:build-log.txt%3A6201

This indicates that the existing tests also cover this part.

pkg/statistics/cmsketch.go
hawkingrei: No, The `calculateMinCountAndCount` in ' scale ' is actually not necessary. The subsequent logic doesn't need `TotalCount` anymore. It just writes directly to TiKV.In fact, `scale` is only used in that one place, so we could actually remove the `scale` function and integrate it directly into the ' analyze ' part.

Issue #61326, opened 2025-05-26
Lightning: import table failed with "context deadline exceeded"


OliverS929:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

The key to reproducing this issue is to trigger a context timeout while Lightning region workers are writing to TiKV.
In the customer scenario where we encountered this error in the Lightning logs, the conditions were:
1. TiKV was under high throughput pressure.
2. `store-write-bwlimit` was set to 40MB to reduce the impact of physical import on the online workload.
3. The Import RPC Duration metric on TiKV peaked at over 20 minutes.

### 2. What did you expect to see? (Required)

Upon a context timeout, we expected to return a common.ErrWriteTooSlow error, which would then trigger the normal retry process.

### 3. What did you see instead (Required)

[...]

PR #61346, opened 2025-05-27
Lightning: Attempt to return writeTooSlow when we experience write timeout.

OliverS929:

Issue Number: ref #61326

Problem Summary: 
Correct the behavior when an error is thrown during a Lightning write timeout to TiKV.
The original intention was to return ErrWriteTooSlow so that the caller could proceed with a retry. However, there was a mistake: ctx.Err() does not carry the timeout cause when it returns. To address this, we now detect the timeout scenario explicitly and manually return the ErrWriteTooSlow error instead.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


CbcWestwolf (approved):

tests/_utils/check_lightning_log_contains
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+#
+# Copyright 2024 PingCAP, Inc.
```
D3Hunter, line 3:
```suggestion
# Copyright 2025 PingCAP, Inc.
```

pkg/lightning/backend/local/region_job.go
```diff
@@ -454,6 +456,29 @@ func (local *Backend) doWrite(ctx context.Context, j *regionJob) (*tikvWriteResu
...
+	// preparation work for the write timeout fault injection, worked only if the following failpoint is enabled
+	originalCtx := ctx             // save the original context with 15 minutes timeout
+	var innerTimeout time.Duration // disabled if 0
+	failpoint.Inject("shortWaitNTimeout", func(val failpoint.Value) {
+		// GO_FAILPOINTS action supplies the duration in ms
+		if ms, ok := val.(int); ok && ms > 0 {
+			innerTimeout = time.Duration(ms) * time.Millisecond
+		} else {
+			innerTimeout = 5 * time.Millisecond // default
+		}
```
D3Hunter, line 468: no need to check in failpoint, simple cast is enough

pkg/lightning/backend/local/region_job.go
```diff
@@ -454,6 +456,29 @@ func (local *Backend) doWrite(ctx context.Context, j *regionJob) (*tikvWriteResu
 		regionMaxSize = j.regionSplitSize * 4 / 3
 	}
 
+	// preparation work for the write timeout fault injection, worked only if the following failpoint is enabled
+	originalCtx := ctx             // save the original context with 15 minutes timeout
+	var innerTimeout time.Duration // disabled if 0
```
D3Hunter, line 461: we can set previous `timeout` directly

lightning/tests/lightning_write_timeout/run.sh
D3Hunter, line 1: maybe consider add to `lightning_write_limit`

pkg/lightning/backend/local/region_job.go
```diff
@@ -464,7 +489,24 @@ func (local *Backend) doWrite(ctx context.Context, j *regionJob) (*tikvWriteResu
 		}
 
 		for i := range clients {
-			if err := writeLimiter.WaitN(ctx, allPeers[i].StoreId, int(size)); err != nil {
+			// original ctx would be used when failpoint is not enabled
+			// that new context would be used when failpoint is enabled
+			err := writeLimiter.WaitN(wctx, allPeers[i].StoreId, int(size))
+			if err != nil {
+				// We expect to encounter two types of errors here:
```
D3Hunter, line 496: i don't think the scenario we meet is caused by rate limiter, it wound return `rate: Wait(n=100) would exceed context deadline` if the left time is not enough inside the context, it will return directly

the scenario we meet should be caused by network issue, and block for a long time with very little or no progress, then context deadline exceed

D3Hunter (commented):
/hold

the fix seems invalid

pkg/lightning/backend/local/region_job.go
OliverS929: Sure. I've updated the code to use a simple cast as recommended.

pkg/lightning/backend/local/region_job.go
OliverS929: Yeah, that’s a valid scenario. How about we open a separate issue to track the investigation for that case?

As noted above, it's quite difficult to deterministically reproduce a context deadline exceeded error. Even basic testing would require a tightly controlled setup and carefully tuned timing to slightly increase the likelihood of the error occurring, and it would take a non-trivial amount of time and effort.

One possible workaround I can think of is to wrap the error returned from each gRPC call, similar to what we did for WaitN. However, I wouldn’t be able to validate those changes for the same reasons mentioned above.

pkg/lightning/backend/local/region_job.go
```diff
@@ -454,6 +456,23 @@ func (local *Backend) doWrite(ctx context.Context, j *regionJob) (*tikvWriteResu
...
+	wctx := ctx
+	wcancel := func() {}
+	failpoint.Inject("shortWaitNTimeout", func(val failpoint.Value) {
+		var innerTimeout time.Duration
+		// GO_FAILPOINTS action supplies the duration in ms
+		if ms := val.(int); ms > 0 {
+			innerTimeout = time.Duration(ms) * time.Millisecond
+		} else {
+			innerTimeout = 5 * time.Millisecond // default
+		}
```
D3Hunter, line 469:
```suggestion
			innerTimeout = time.Duration(ms) * time.Millisecond
```

pkg/lightning/backend/local/region_job.go
OliverS929: Thank you for the suggestion! I’ve followed your advice and used a defer function to handle context deadline exceeded errors with the cause set to common.ErrWriteTooSlow under this specific context.
Please take a look when you have a chance.

pkg/lightning/backend/local/region_job.go
```diff
@@ -464,7 +498,17 @@ func (local *Backend) doWrite(ctx context.Context, j *regionJob) (*tikvWriteResu
...
+			// that new context would be used when failpoint is enabled
+			err := writeLimiter.WaitN(wctx, allPeers[i].StoreId, int(size))
+			if err != nil {
+				// We expect to encounter two types of errors here:
+				// 1. context.DeadlineExceeded — occurs when the calculated delay is less than the remaining time in the context, but the context expires while sleeping.
+				// 2. "rate: Wait(n=%d) would exceed context deadline" — a fast-fail path triggered when the delay already exceeds the remaining time for context before sleeping.
+				//
+				// Unfortunately, we cannot precisely control when the context will expire, so both scenarios are valid and expected.
+				// Fortunately, the "rate: Wait" error is already treated as retryable, so we only need to explicitly handle context.DeadlineExceeded here.
+				// We reply on the defer function at the top of doWrite to handle it for us in general.
```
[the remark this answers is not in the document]
OliverS929: I was hoping that we keep the comment here, so that the situation is well explained and the two error type and their encounters are clearly stated.

pkg/lightning/backend/local/region_job.go
OliverS929: Code updated.

pkg/lightning/backend/local/region_job.go
```diff
@@ -333,9 +333,27 @@ func (local *Backend) doWrite(ctx context.Context, j *regionJob) (*tikvWriteResu
 	})
 
 	var cancel context.CancelFunc
-	ctx, cancel = context.WithTimeoutCause(ctx, 15*time.Minute, common.ErrWriteTooSlow)
+	// set a timeout for the write operation, if it takes too long, we will return with common.ErrWriteTooSlow and let caller retry the whole job instead of being stuck forever.
+	timeout := 15 * time.Minute
+	ctx, cancel = context.WithTimeoutCause(ctx, timeout, common.ErrWriteTooSlow)
 	defer cancel()
 
+	// A defer function to handle all DeadlineExceeded errors that may occur during the write operation using this context with 15 minutes timeout. When the error is "context deadline exceeded", we will check if the cause is common.ErrWriteTooSlow and return thecommon.ErrWriteTooSlow instead so our caller would be able to retry this doWrite operation. By doing this defer we are hoping to handle all DeadlineExceeded error during this write, either from gRPC stream or write limiter WaitN operation.
```
D3Hunter: too long, please split into multiple line within 80 chars

OliverS929: Thanks for pointing out. Will do.

PR #61402, opened 2025-05-29
parser: add SEM SQL commands for all statements

### LINKED ISSUE #61396: Remove the `ChangeStmt` codes in tidb
## Enhancement

The parser has removed support for `ChangeStmt`. But some codes are still left in planner, parser, etc... We should also remove them.

### LINKED ISSUE #61397: Add a method to `StmtNode` to give a more detailed label to a statement to fulfill the needs of SEM
## Enhancement

The `SEM` needs a more detailed label for a statement. The current `GetStmtLabel` function is not well maintained and is also not detailed enough. We should add a new method for `StmtNode`.

However, it's still not clear whether the current definition is good for all cases. Therefore, I'll make it named as `SEMCommand` method. If someone is about to re-use it later, it's fine to just rename it.

YangKeao:

Issue Number: close #61397, close #61396

Problem Summary:

1. The `CHANGE` statement has been removed. It should be removed everywhere in TiDB.
2. Add `SEMCommand` needed by the SEM feature. The SEM feature will use it to filter out the SQL.

For example, the new design of SEM may have the following config to restrict the SQL:

```json
{
"restricted_SQL" : {
    "SQL": [
        "BACKUP",
        "RESTORE",
        "ALTER RESOURCE GROUP"
    ],
    "rule": [
        "select_into_file",
        "import_from_disk"
    ],
}
}
```

This PR implements methods to get "SQL" from a real sql statements. Then we can match them in the sem module.

1. Add a `SEMCommand` method to `StmtNode`, and implement it for all `StmtNode`.
2. Add tests to make sure every subcommand of `ADMIN`, `SHOW`, `BRIE` are included.
3. Remove all codes related to `CHANGE` statement.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


CbcWestwolf (approved):

pkg/parser/ast/sem.go
```diff
@@ -0,0 +1,1315 @@
...
+}
+
+// SEMCommand returns the command string for the statement.
+func (n *SelectStmt) SEMCommand() string {
+	return SelectCommand
+}
+
+// SEMCommand returns the command string for the statement.
+func (n *SetOprStmt) SEMCommand() string {
+	return UnionCommand
```
bb7133: Seems this is wrong?

pkg/parser/ast/sem.go
```diff
@@ -0,0 +1,1315 @@
...
+	UnionCommand = "UNION"
+	// SplitRegionCommand represents SPLIT REGION statement
+	SplitRegionCommand = "SPLIT REGION"
+	// UpdateCommand represents UPDATE statement
+	UpdateCommand = "UPDATE"
+)
+
+// Show Statements
+
+const (
```
bb7133, line 122: I have some concerns about how can we maintain them...

For example, we support `SHOW PROFILE` in syntax only(it does nothing), so should we include it in this list?

pkg/parser/ast/sem.go
YangKeao: Fixed. It has been modified to `SET OPERATION` instead as it may contain `UNION`, `INTERSECT` and `EXCEPT`.

pkg/parser/ast/sem.go
YangKeao: > For example, we support `SHOW PROFILE` in syntax only(it does nothing), so should we include it in this list?

I think we should include every syntax supported by the parser. This list is not exposed outside (to the user), so it's fine to have more commands than what TiDB actually supported.

bb7133 (approved):
LGTM

PR #61405, opened 2025-05-29
*: Set keyspace_name const labels for client-go related metrics

yibin87:

Issue Number: ref #60864 

Problem Summary:

In this PR, init tikv_client and pd_client metrics with keyspace_name const labels. Note, pd client uses init-only-once mechanism to init metrics, thus create a temporary pd client for this usage.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
Hack code to set const labels manually, and verify the metrics got from http API.
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/util/metricsutil/common.go
+++ b/pkg/util/metricsutil/common.go
@@ -57,8 +57,7 @@ func RegisterMetrics() error {
 	}
 
 	if kerneltype.IsNextGen() {
-		registerMetricsForNextGen(cfg.KeyspaceName)
-		return nil
+		metrics.SetConstLabels("keyspace_name", cfg.KeyspaceName)
 	}
 
 	pdAddrs, _, _, err := tikvconfig.ParsePath("tikv://" + cfg.Path)
@@ -77,12 +76,15 @@ func RegisterMetrics() error {
 	}
 	defer pdCli.Close()
 
-	keyspaceMeta, err := getKeyspaceMeta(pdCli, cfg.KeyspaceName)
-	if err != nil {
-		return err
+	if kerneltype.IsNextGen() {
+		registerMetrics(nil) // metrics' const label already set
+	} else {
+		keyspaceMeta, err := getKeyspaceMeta(pdCli, cfg.KeyspaceName)
+		if err != nil {
+			return err
+		}
+		registerMetrics(keyspaceMeta)
 	}
-
-	registerMetrics(keyspaceMeta)
 	return nil
 }
 
@@ -94,8 +96,7 @@ func RegisterMetricsForBR(pdAddrs []string, tls task.TLSConfig, keyspaceName str
 	}
 
 	if kerneltype.IsNextGen() {
-		registerMetricsForNextGen(keyspaceName)
-		return nil
+		metrics.SetConstLabels("keyspace_name", keyspaceName)
 	}
 
 	timeoutSec := 10 * time.Second
@@ -110,12 +111,15 @@ func RegisterMetricsForBR(pdAddrs []string, tls task.TLSConfig, keyspaceName str
 	}
 	defer pdCli.Close()
 
-	keyspaceMeta, err := getKeyspaceMeta(pdCli, keyspaceName)
-	if err != nil {
-		return err
+	if kerneltype.IsNextGen() {
+		registerMetrics(nil) // metrics' const label already set
+	} else {
+		keyspaceMeta, err := getKeyspaceMeta(pdCli, keyspaceName)
+		if err != nil {
+			return err
+		}
+		registerMetrics(keyspaceMeta)
 	}
-
-	registerMetrics(keyspaceMeta)
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/util/metricsutil/common.go
```diff
@@ -67,22 +66,26 @@ func RegisterMetrics() error {
...
-	if err != nil {
-		return err
+	if kerneltype.IsNextGen() {
+		registerMetrics(nil) // metrics' const label already set
+	} else {
+		keyspaceMeta, err := getKeyspaceMeta(pdCli, cfg.KeyspaceName)
+		if err != nil {
+			return err
+		}
+		registerMetrics(keyspaceMeta)
```
zimulala, line 87: Could we move the call to `metrics.SetConstLabels("keyspace_id"...)` earlier in the logic(line59-61)? This way, we can handle the next-gen keyspace info setup in the same place. However, I'm not sure if it's feasible due to the different initMetrics calls.

yibin87: Actually, for next gen, we should set const labels first, then we can use const labels to init pdCli; while for non-next gen, we should get pdCli first, then get keyspaceMeta using pdCli, and finally set const labels...

zimulala: Well. 
Then, should we consider extracting these lines of code into a function, so that the related code only needs to be modified in one place

zimulala: Got it.

zimulala (approved):

pkg/util/metricsutil/common.go
```diff
@@ -67,22 +66,26 @@ func RegisterMetrics() error {
...
 	if err != nil {
 		return err
 	}
 	defer pdCli.Close()
 
-	keyspaceMeta, err := getKeyspaceMeta(pdCli, cfg.KeyspaceName)
-	if err != nil {
-		return err
+	if kerneltype.IsNextGen() {
+		registerMetrics(nil) // metrics' const label already set
```
lance6716, line 81: I see in line 69 you said "for NextGen, pdCli is created to init the metrics' const labels", but here when it's NextGen, `pdCli` does nothing. I guess line 69 are wrong?

pkg/util/metricsutil/common.go
```diff
@@ -67,22 +66,26 @@ func RegisterMetrics() error {
 	}
 
 	timeoutSec := time.Duration(cfg.PDClient.PDServerTimeout) * time.Second
+	// Note: for NextGen, pdCli is created to init the metrics' const labels
 	pdCli, err := pd.NewClient(componentName, pdAddrs, pd.SecurityOption{
```
lance6716, line 70: seems `pdCli` is not used when `kerneltype.IsNextGen()`, we don't need to create it.

pkg/util/metricsutil/common.go
yibin87: For NextGen, we first set const labels, then create pdCli with const labels as parameter which will init pd-client metrics with const labels. And for NextGen, we only need to create pdCli, after that, it has no usage.

pkg/util/metricsutil/common.go
yibin87: Explained in previous comment.

lance6716: I mean

```diff
diff --git i/pkg/util/metricsutil/common.go w/pkg/util/metricsutil/common.go
index fd0d5cc7e9..847b0cf3be 100644
--- i/pkg/util/metricsutil/common.go
+++ w/pkg/util/metricsutil/common.go
@@ -65,21 +65,21 @@ func RegisterMetrics() error {
 		return err
 	}
 
-	timeoutSec := time.Duration(cfg.PDClient.PDServerTimeout) * time.Second
-	// Note: for NextGen, pdCli is created to init the metrics' const labels
-	pdCli, err := pd.NewClient(componentName, pdAddrs, pd.SecurityOption{
-		CAPath:   cfg.Security.ClusterSSLCA,
-		CertPath: cfg.Security.ClusterSSLCert,
-		KeyPath:  cfg.Security.ClusterSSLKey,
-	}, opt.WithCustomTimeoutOption(timeoutSec), opt.WithMetricsLabels(metrics.GetConstLabels()))
-	if err != nil {
-		return err
-	}
-	defer pdCli.Close()
-
 	if kerneltype.IsNextGen() {
 		registerMetrics(nil) // metrics' const label already set
 	} else {
+		timeoutSec := time.Duration(cfg.PDClient.PDServerTimeout) * time.Second
+		// Note: for non-NextGen, pdCli is created to init the metrics' const labels
+		pdCli, err := pd.NewClient(componentName, pdAddrs, pd.SecurityOption{
+			CAPath:   cfg.Security.ClusterSSLCA,
+			CertPath: cfg.Security.ClusterSSLCert,
+			KeyPath:  cfg.Security.ClusterSSLKey,
+		}, opt.WithCustomTimeoutOption(timeoutSec), opt.WithMetricsLabels(metrics.GetConstLabels()))
+		if err != nil {
+			return err
+		}
+		defer pdCli.Close()
+
 		keyspaceMeta, err := getKeyspaceMeta(pdCli, cfg.KeyspaceName)
 		if err != nil {
 			return err
```

yibin87: Well, for NextGen, we need to create pdCli to init pd-client's metrics with const labels

pkg/util/metricsutil/common.go
```diff
@@ -67,22 +66,26 @@ func RegisterMetrics() error {
 	}
 
 	timeoutSec := time.Duration(cfg.PDClient.PDServerTimeout) * time.Second
+	// Note: for NextGen, pdCli is created to init the metrics' const labels
```
lance6716:
```suggestion
	// Note: for NextGen, we need to use the side effect of `NewClient` to init the metrics' builtin const labels
```

PR #61421, opened 2025-05-29
planner: Only count all topn when necessary

terry1purcell:

Issue Number: ref #61339 

Problem Summary:

This PR does the following in addition to a prior PR targeted at this issue:
1) It removes the need to accumulate the count of all topN for the purposes of determining if we have all rows in the TopN.
2) Moves the accumulation of total count to immediately where it is used in the 2nd scenario (after being removed in 1st scenario).
3) Moved the check for "all values in TopN" prior to removing these from the samples - since there is no point to removing the TopN values from the sample if the TopN covers the whole set - since we'll exit anyway and no longer need the samples.
4) When removing the topN values from the samples (to build histograms) - it was incorrectly using the TopN count (from ANALYZE) as the entry point of the IF statement rather than the number in the actual TopN after pruning.
5) Reduce redundant calculation of the lengths of samples/TopN

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/statistics/builder.go
```diff
@@ -416,23 +416,37 @@ func BuildHistAndTopN(
...
 				firstTimeSample types.Datum
 			)
 			for j := range topNList {
 				if bytes.Equal(sampleBytes, topNList[j].Encoded) {
 					// This should never happen, but we met this panic before, so we add this check here.
 					// See: https://github.com/pingcap/tidb/issues/35948
-					if foundTwice {
+					// If we have "foundOnce" and entered this branch again, it means that
+					// the same value appears multiple times in the topN, which is not expected.
+					if foundOnce {
```
0xPoe: It seems that after this change, this check will never be true because the loop breaks in the else branch below.

0xPoe (commented):
Thanks!

pkg/statistics/builder.go
terry1purcell: You are correct. This change was made assuming TopN are sorted - which they are not. I reverted this part of the change in the next commit.

pkg/statistics/builder.go
```diff
@@ -416,13 +416,25 @@ func BuildHistAndTopN(
...
+	if lenTopN > 0 && lenSamples > 0 {
+		for i := int64(0); i < lenSamples; i++ {
 			sampleBytes, err := getComparedBytes(samples[i].Value)
 			if err != nil {
 				return nil, nil, errors.Trace(err)
 			}
+			if i > 0 && bytes.Equal(sampleBytes, samples[i-1].Value.GetBytes()) {
+				// If the sample is the same as the previous one, we can skip it.
+				continue
+			}
```
0xPoe, line 441: Maybe I am misunderstanding it, but I think this would break the original code. Before this change, the index would start from the first value of the list if we found it in the topN, but after this change, it would start from the last value of the sample list.

terry1purcell: This code is to avoid looping through the TopN if we have already validated that this sample key value is NOT found in the TopN. I used the following test data to confirm that debugging through this code we will enter this IF twice given the example below (for the 2nd iteration of value 1, and the 2nd iteration of value 2). If I specify "with 2 topn" in the ANALYZE statement, we only enter this IF once - as expected. Therefore, my testing demonstrates that this check is valid in attempting to avoid looping through the TopN for a NOT FOUND value. I believe your question is related to when a value is FOUND in the TopN, and then removed from the sample.

CREATE TABLE `t2` (
  `a` int DEFAULT NULL,
  KEY `ia` (`a`)
);
insert into t2 values (1);
insert into t2 values (1);
insert into t2 values (2);
insert into t2 values (2);
insert into t2 values (3);
insert into t2 values (3);
insert into t2 values (3);
analyze table t2 with 1 topn;

pkg/statistics/builder.go
```diff
@@ -416,13 +416,25 @@ func BuildHistAndTopN(
...
-	if numTopN != 0 {
-		for i := int64(0); i < int64(len(samples)); i++ {
+	lenSamples := int64(len(samples))
+	if lenTopN > 0 && lenSamples > 0 {
+		for i := int64(0); i < lenSamples; i++ {
 			sampleBytes, err := getComparedBytes(samples[i].Value)
 			if err != nil {
 				return nil, nil, errors.Trace(err)
 			}
+			if i > 0 && bytes.Equal(sampleBytes, samples[i-1].Value.GetBytes()) {
```
0xPoe, line 438:
```suggestion
			// Safety:
			// When a sample value matches an entry in the topN list, it is removed from the samples array,
			// so it can never match the previous sample. Therefore, this equality check will never be true
			// in that case. For values not found in topN, we don't need keep checking the same value.
			if i > 0 && bytes.Equal(sampleBytes, samples[i-1].Value.GetBytes()) {
```

0xPoe (approved):
I understand the point of that line. I thought it was `i+1`, but actually it is `i-1`. I am completely blind.

Thank you for your clarification.

Thanks! :shipit:

0xPoe:
/hold

For https://github.com/pingcap/tidb/pull/61421#discussion_r2123703815

0xPoe:
Feel free to unhold it when you add the comments. Thank you!

0xPoe:
/unhold

I just added those comments.

pkg/statistics/builder.go
```diff
@@ -416,13 +416,29 @@ func BuildHistAndTopN(
 		topNList = pruneTopNItem(topNList, ndv, nullCount, sampleNum, count)
 	}
 
+	topn := &TopN{TopN: topNList}
+	lenTopN := int64(len(topn.TopN))
+	if hg.NDV <= lenTopN {
+		// If we've collected everything  - don't create any buckets
+		return hg, topn, nil
```
AilinKid, line 423: maybe the returned topn here missed * scaleFactor？

AilinKid: oh，we can't generally did it here, the underlying code may still need original topNList[j].Count to skip samples

terry1purcell: @AilinKid - I believe that you are correct that it misses "* scalefactor". And you are correct - that we can only add it here if we are exiting.

Issue #60418, opened 2025-04-07
add-index/import-into on global sort for next generation TiDB


D3Hunter:
## Feature Request
next generation TiDB will store data on cloud storage, such as S3/GCS, and will use cloud infrastructure to make sure high availability, and TiKV will load data as needed.

in this case, for add-index/import-into on global sort, we need to adapt this change, the major change is:
- the KV encoding, global-sorting, pause region scheduler, region split and scatter, part is the same as current code path
- when write KV pair, we write to a servitization component, let's call it `tikv-worker`, which will accept the KV pair and write a SST file to cloud storage, and return the filename back
  - note: we only need write one SST file, not 3
- add-index/import-into will call ingest API on TiKV using the filename as part of the component

currently, we will implememt this into a feature branch, and use a separate binary for NG TiDB.

TODO items
- [x] https://github.com/pingcap/tidb/issues/61198

PR #61427, opened 2025-05-30
dxf: make target scope fixed to dxf_service on nextgen

D3Hunter:

Issue Number: ref #60418 

Problem Summary:

DXF have a compatibility logic for `background` scope, DXF will try schedule tasks with target scope equals to default scope(i.e. empty scope) to nodes with `background` scope if there are any `background` nodes, but if there are no such nodes, DXF will schedule those tasks on nodes with default scope which also runs heavy user workload, that's not expected, as we want to separate user workload and DXF related workload completely on tidb side on nextgen. to avoid this issue, we make target scope of tasks fixed to `dxf_service` on nextgen.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test

run a nextgen cluster with default scope, then start import into job. this job will be blocked as we don't have node with scope `dxf_service`. then we set one tidb scope to `dxf_service`, then the import run and succeed.

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


Benjamin2037 (approved):
LGTM

PR #61533, opened 2025-06-05
types/time: invalid timestamp during DST spring forward fixed to next valid ts instead of zero date.

### LINKED ISSUE #61335: Reporting a tablefullscan When I read the data between invalid data due to DST
## Bug Report

If there is DST happen, the time '2025-03-30 02:00:00'~'2025-03-30 02:59:00' is not exist. When I read the data between invalid data , Resulting almost read all the data which is not expacted.  And if I use a valid data , the explain is normal.

```
mysql> SET GLOBAL time_zone = 'Europe/Stockholm';

Create Table: CREATE TABLE `orders` (
  `id` int DEFAULT NULL,
  `created` timestamp NULL DEFAULT NULL,
  KEY `idx_create` (`created`)
1 row in set (0.00 sec)

mysql> explain select *  from orders  where created between '2025-03-30 02:01:59'   and '2025-03-30 03:00:01' order by 2;
+--------------------------------+--------------+-----------+------------------------------------------------+-------------------------------------------------------------------------+
| id                             | estRows      | task      | access object                                  | operator info                                                           |
+--------------------------------+--------------+-----------+------------------------------------------------+-------------------------------------------------------------------------+
| IndexLookUp_17                 | 103218635.00 | root      |                                                |                                                                         |
| ├─IndexRangeScan_15(Build)     | 103218635.00 | cop[tikv] | table:orders, index:order_created_idx(created) | range:[2025-03-30 02:01:59.000000,2025-03-30 03:00:01], keep order:true |
| └─TableRowIDScan_16(Probe)     | 103218635.00 | cop[tikv] | table:orders                                   | keep order:false                                                        |
+--------------------------------+--------------+-----------+------------------------------------------------+-------------------------------------------------------------------------+
3 rows in set, 26 warnings (0,024 sec)

mysql> explain select * from orders  where created between '2025-03-30 01:01:59'   and '2025-03-30 03:00:01' order by 2;
+--------------------------------+---------+-----------+------------------------------------------------+------------------------------------------------------------------+
| id                             | estRows | task      | access object                                  | operator info                                                    |
+--------------------------------+---------+-----------+------------------------------------------------+------------------------------------------------------------------+
| IndexLookUp_14                 | 2110.60 | root      |                                                |                                                                  |
| ├─IndexRangeScan_12(Build)     | 2110.60 | cop[tikv] | table:orders, index:order_created_idx(created) | range:[2025-03-30 01:01:59,2025-03-30 03:00:01], keep order:true |
| └─TableRowIDScan_13(Probe)     | 2110.60 | cop[tikv] | table:orders                                   | keep order:false                                                 |
+--------------------------------+---------+-----------+------------------------------------------------+------------------------------------------------------------------+
3 rows in set (0,011 sec)
```

Table

v8.5.1

mjonss:

Issue Number: close #61335

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/test/simpletest/simple_test.go
+++ b/pkg/executor/test/simpletest/simple_test.go
@@ -818,3 +818,60 @@ func TestKillStmt(t *testing.T) {
 	tk.MustExecToErr("kill rand()", "Invalid operation. Please use 'KILL TIDB [CONNECTION | QUERY] [connectionID | CONNECTION_ID()]' instead")
 	// remote kill is tested in `tests/globalkilltest`
 }
+
+func TestSelectWhereInvalidDSTTime(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("create table t (id int, ts timestamp, index idx_ts(ts))")
+	tk.MustExec(`set sql_mode = ''`)
+	tk.MustExec(`set time_zone = "Europe/Amsterdam"`)
+	tk.MustExec("insert into t values (1, '2025-03-30 02:00:00')")
+	//tk.MustQuery(`show warnings`).Check(testkit.Rows("Warning 1292 Incorrect timestamp value: '2025-03-30 02:00:00' for column 'ts' at row 1"))
+	// TODO: Why is this ZERO date and not adjusted?!?
+	tk.MustQuery(`select * from t`).Check(testkit.Rows("1 0000-00-00 00:00:00"))
+	tk.MustExec(`set sql_mode = DEFAULT`)
+	tk.MustExec("insert into t values (2, '2025-03-30 01:59:59')")
+	tk.MustExec("insert into t values (3, '2025-03-30 03:00:00')")
+	tk.MustExec("insert into t values (4, '1970-01-01 01:00:01')")
+	tk.MustExec(`set time_zone = "Europe/Amsterdam"`)
+	tk.MustQuery(`select *, unix_timestamp(ts) from t where ts between '2025-03-30 02:30:00' AND '2025-03-30 03:00:00'`).Check(testkit.Rows("3 2025-03-30 03:00:00 1743296400"))
+	tk.MustQuery(`select * from t where ts between '2025-03-30 02:30:00' AND '2025-03-30 03:00:00'`).Check(testkit.Rows("3 2025-03-30 03:00:00"))
+	tk.MustQuery(`select * from t where ts between '2025-03-30 01:00:00' AND '2025-03-30 02:00:00'`).Check(testkit.Rows("2 2025-03-30 01:59:59"))
+	tk.MustQuery(`show warnings`).Check(testkit.Rows("Warning 8179 Timestamp is not valid, since it is in Daylight Saving Time transition '{2025 3 30 2 0 0 0}' for time zone 'Europe/Amsterdam'",
+		"Warning 8179 Timestamp is not valid, since it is in Daylight Saving Time transition '{2025 3 30 2 0 0 0}' for time zone 'Europe/Amsterdam'",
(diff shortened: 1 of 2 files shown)


qw4990 (commented):
A nice fix! Could you add more test cases like :
1. where column <cmp> date_add(<invalid_time>, interval 1 day)
2. where date_add(column, interval 1 day) <cmp> <invalid_time>

pkg/types/time.go
```diff
@@ -2179,6 +2199,7 @@ func checkTimestampType(t CoreTime, tz *gotime.Location) error {
 		convertTime := NewTime(t, mysql.TypeTimestamp, DefaultFsp)
 		err := convertTime.ConvertTimeZone(tz, BoundTimezone)
 		if err != nil {
+			_, err = adjustTimestampErrForDST(tz, t.String(), mysql.TypeTimestamp, Time{t}, err)
```
dveeden:
```suggestion
			_, err2 = adjustTimestampErrForDST(tz, t.String(), mysql.TypeTimestamp, Time{t})
```
And then `if err2 != nil ....`


I find this a bit clearer. Both are fine. This might make the function a bit more generic so it could be exported at some point and use from other packages if needed.

pkg/executor/test/simpletest/simple_test.go
```diff
@@ -818,3 +818,44 @@ func TestKillStmt(t *testing.T) {
 	tk.MustExecToErr("kill rand()", "Invalid operation. Please use 'KILL TIDB [CONNECTION | QUERY] [connectionID | CONNECTION_ID()]' instead")
 	// remote kill is tested in `tests/globalkilltest`
 }
+
+func TestSelectWhereInvalidDSTTime(t *testing.T) {
+	store := testkit.CreateMockStore(t)
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("use test")
+	tk.MustExec("create table t (id int, ts timestamp)")
```
dveeden, line 826: I assume `timestamp` and `timestamp(6)` are expected to behave the same?

dveeden:
```
mysql-8.0.11-TiDB-v9.0.0-alpha-892-gb79cae3387-dirty> show warnings;
+---------+------+--------------------------------------------------------------------------------------------------------------------------------+
| Level   | Code | Message                                                                                                                        |
+---------+------+--------------------------------------------------------------------------------------------------------------------------------+
| Warning | 8179 | Timestamp is not valid, since it is in Daylight Saving Time transition '{2025 3 30 2 30 0 0}' for time zone 'Europe/Amsterdam' |
| Warning | 8179 | Timestamp is not valid, since it is in Daylight Saving Time transition '{2025 3 30 2 30 0 0}' for time zone 'Europe/Amsterdam' |
| Warning | 8179 | Timestamp is not valid, since it is in Daylight Saving Time transition '{2025 3 30 2 30 0 0}' for time zone 'Europe/Amsterdam' |
+---------+------+--------------------------------------------------------------------------------------------------------------------------------+
3 rows in set (0.000 sec)
```

This warning:
* Is TiDB specific
* Is similar to the 1292 "Incorrect datetime value..." warning in MySQL
* Is more specific to what is wrong with the supplied value
* Has a timestamp format that is different than what the user supplied or what is usual for MySQL
* Was not added in this PR

I would prefer to keep things as similar to MySQL as possible. However the added info here is useful.

pkg/executor/test/simpletest/simple_test.go
mjonss: Yes. I just tested manually and it is the same, but would prefer to not include `timestamp(6)` in this test. I am currently working on a more covering tests for timestamp, which also includes timestamp with fractions, which should cover this as well.

pkg/types/time.go
mjonss: Done

PR #61478, opened 2025-06-03
planner: outer join pruning for constants

terry1purcell:
### What problem does this PR solve?

Issue Number: ref #61305

Problem Summary:

### What changed and how does it work?

Issue 61305 exposed limitations in the outer join pruning if constants are in the select list. If a constant is in the select list, then we don't recognize that this column is irrelvant - and thus if you ONLY otherwise have columns from the left side, we don't prune right side tables due to the constant.

Enhanced outer join pruning to cover these cases.

An example is - before:
``` 
EXPLAIN SELECT DISTINCT 1, t3.parent_id as w1
    -> FROM t3
    ->          LEFT OUTER JOIN t1 ON t1.parent_id = t3.parent_id
    ->       WHERE t3.search_field = 'Find me'
    -> LIMIT 101;
+---------------------------------+------------+-----------+--------------------------------------------+---------------------------------------------------------------------------------------------------------------+
| id                              | estRows    | task      | access object                              | operator info                                                                                                 |
+---------------------------------+------------+-----------+--------------------------------------------+---------------------------------------------------------------------------------------------------------------+
| Limit_13                        | 8.00       | root      |                                            | offset:0, count:101                                                                                           |
| └─HashAgg_14                    | 8.00       | root      |                                            | group by:test.t3.parent_id, funcs:firstrow(1)->Column#6, funcs:firstrow(test.t3.parent_id)->test.t3.parent_id |
|   └─HashJoin_24                 | 10.00      | root      |                                            | left outer join, left side:IndexReader_26, equal:[eq(test.t3.parent_id, test.t1.parent_id)]                   |
|     ├─IndexReader_26(Build)     | 8.00       | root      |                                            | index:IndexRangeScan_25                                                                                       |
|     │ └─IndexRangeScan_25       | 8.00       | cop[tikv] | table:t3, index:search_field(search_field) | range:["Find me","Find me"], keep order:false                                                                 |
|     └─TableReader_28(Probe)     | 1116372.00 | root      |                                            | data:TableFullScan_27                                                                                         |
|       └─TableFullScan_27        | 1116372.00 | cop[tikv] | table:t1                                   | keep order:false                                                                                              |
+---------------------------------+------------+-----------+--------------------------------------------+---------------------------------------------------------------------------------------------------------------+
```

Example after this enhancement:
```
EXPLAIN SELECT DISTINCT 1, t3.parent_id as w1
    -> FROM t3
[...]

PR #61597, opened 2025-06-09
Revert "planner: outer join pruning for constants"

hawkingrei:
Reverts pingcap/tidb#61478


hawkingrei:
Sorry, revert wrong commit id.

PR #61598, opened 2025-06-09
Revert "planner: Only count all topn when necessary"

hawkingrei:
Reverts pingcap/tidb#61421


hawkingrei:
@terry1purcell @fixdb PTAL

Issue #61088, opened 2025-05-13
Show more detailed progress in `SHOW IMPORT JOBS`


joechenrh:
## Enhancement

Currently, `SHOW IMPORT JOBS` displays few information about the importing progress, user can only get the ingested rows from the output, like:

```
+--------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------------------+----------+-------+----------+------------------+---------------+----------------+----------------------------+----------------------------+----------------------------+------------+
| Job_ID | Data_Source                                                                                                                                                                                                     | Target_Table     | Table_ID | Phase | Status   | Source_File_Size | Imported_Rows | Result_Message | Create_Time                | Start_Time                 | End_Time                   | Created_By |
+--------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------------------+----------+-------+----------+------------------+---------------+----------------+----------------------------+----------------------------+----------------------------+------------+
|      1 | s3://global-sort/joechenrh/wide-table-normal/part000/test.0.csv?access-key=xxxxxx&endpoint=http%3A%2F%2Fks3-cn-beijing.ksyuncs.com&force-path-style=false&provider=ks&region=Beijing&secret-access-key=xxxxxx | `test`.`sbtest1` |      112 |       | finished | 282.4MiB         |        500000 |                | 2025-05-13 10:54:56.135831 | 2025-05-13 10:54:56.653506 | 2025-05-13 10:57:38.210800 | root@%     |
+--------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------------------+----------+-------+----------+------------------+---------------+----------------+----------------------------+----------------------------+----------------------------+------------+
```

The only way to estimated progress is resorting to log or grafana. So here we want to output more informations in `SHOW IMPORT JOBS`.

PR #61670, opened 2025-06-11
lightning: change the implementation of `ScannedPos`

joechenrh:

Issue Number: ref https://github.com/pingcap/tidb/issues/61088

Problem Summary:

In the encode step, we will track the amount of data read and show it on the panel. This is achieved by calling `Parser.ScannedPos` to get the latest read position. However, the current implement is wrong:

https://github.com/pingcap/tidb/blob/11276faa9dc05ed7047f766a4c1fb45cd2b83109/pkg/lightning/mydump/parquet_parser.go#L377-L379

Because we will open one reader for each column, `Seek` can't reflect the actual bytes we read. Below is a simple illustration.

```
File layout (Dict page and meta are ignored)
--------------------------------------------------------------------------------------------------
|   Column 0 Group 0    |    Column 1 Group 0    |   Column 0 Group 1    |   Column 1 Group 1    | 
--------------------------------------------------------------------------------------------------
                ^                       ^
                |                       |
            Reader 0                 Reader 1
```

Since it's hard to retrieve total scanned bytes for the parser, we used the size of read rows instead which is better than using `Seek`.

Additionally, we use `ScannedPos` to update metrics for all types of data.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [X] No need to test
  > 

As mentioned before, this interface is used to update the metrics, so changes in this PR won't affect the execution of import jobs.

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/lightning/mydump/parser.go
+++ b/pkg/lightning/mydump/parser.go
@@ -156,8 +156,9 @@ type Parser interface {
 	// TODO: replace pos with a new structure to specify position offset and rows offset
 	Pos() (pos int64, rowID int64)
 	SetPos(pos int64, rowID int64) error
-	// ScannedPos always returns the current file reader pointer's location
-	ScannedPos() (int64, error)
+	// ScannedPos always returns the current file reader pointer's location.
+	// It's used to track the progress of the reader.
+	ScannedPos() int64
 	Close() error
 	ReadRow() error
 	LastRow() Row
@@ -218,8 +219,8 @@ func (parser *blockParser) SetPos(pos int64, rowID int64) error {
 
 // ScannedPos gets the read position of current reader.
 // this always returns the position of the underlying file, either compressed or not.
-func (parser *blockParser) ScannedPos() (int64, error) {
-	return parser.reader.Seek(0, io.SeekCurrent)
+func (parser *blockParser) ScannedPos() int64 {
+	return parser.pos
 }
 
 // Pos returns the current file offset.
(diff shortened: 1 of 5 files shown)


pkg/lightning/mydump/parser.go
```diff
@@ -218,8 +219,8 @@ func (parser *blockParser) SetPos(pos int64, rowID int64) error {
 
 // ScannedPos gets the read position of current reader.
 // this always returns the position of the underlying file, either compressed or not.
```
lance6716: This comment says for compression files, `ScannedPos` will return a smaller value compared with accumulated bytes returned.  In `Parser` interface, `Pos` is similar to `ScannedPos` but they have a different implementation before this PR. So I suggest double check the caller's usage and what property caller expects. Maybe also rename the methods to provide a more clear meaning.

joechenrh: > Maybe also rename the methods to provide a more clear meaning.

Good suggestion, I agree with this. These two positions are used for checkpoint and calculating several metrics like read, total read, etc. But the logic is a bit complicated IMO.

Takes restored bytes as example (other metrics may have different logic 😂):
- For compressed/parquet files, we use `ScannedPos`
- For other files, we use `Pos` 



Maybe we should rename `Pos` as `Processed`, and unify the usage of the upper level.

joechenrh:
To some extent, the value get from `ScannedPos` won't affect the progress we calculate too much.

Because for each parser, the `ScannedPos` points to the start position of meta part and never changed after opening (because we will open a new reader for eacn column). As the meta is small, the delta we get is almost the same as the file size. 😂

So just keep the current logic and fix it after switching to new parse library.

PR #61677, opened 2025-06-11
privilege: add `ColumnsPrivMap` for accelerating column-privilege

CbcWestwolf:

Issue Number: close #xxx

Problem Summary:

Adapt the MySQLPrivilege to support column privilege

This PR pick `pkg/privilege/privileges/cache.go` and `pkg/privilege/privileges/cache_test.go` in https://github.com/pingcap/tidb/pull/61638.

### Check List

Tests 

- [x] Unit test (existent cases)
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/privilege/privileges/cache.go
+++ b/pkg/privilege/privileges/cache.go
@@ -276,11 +276,12 @@ type immutable struct {
 }
 
 type extended struct {
-	UserMap       map[string][]UserRecord // Accelerate User searching
-	Global        map[string][]globalPrivRecord
-	Dynamic       map[string][]dynamicPrivRecord
-	DBMap         map[string][]dbRecord         // Accelerate DB searching
-	TablesPrivMap map[string][]tablesPrivRecord // Accelerate TablesPriv searching
+	UserMap        map[string][]UserRecord // Accelerate User searching
+	Global         map[string][]globalPrivRecord
+	Dynamic        map[string][]dynamicPrivRecord
+	DBMap          map[string][]dbRecord          // Accelerate DB searching
+	TablesPrivMap  map[string][]tablesPrivRecord  // Accelerate TablesPriv searching
+	ColumnsPrivMap map[string][]columnsPrivRecord // Accelerate ColumnsPriv searching
 }
 
 // MySQLPrivilege is the in-memory cache of mysql privilege tables.
@@ -499,6 +500,7 @@ func (p *MySQLPrivilege) merge(diff *immutable) *MySQLPrivilege {
 		return x.Host == y.Host && x.User == y.User &&
 			x.DB == y.DB && x.TableName == y.TableName && x.ColumnName == y.ColumnName
 	})
+	ret.buildColumnsPrivMap()
 
 	ret.defaultRoles = make([]defaultRoleRecord, 0, len(p.defaultRoles)+len(diff.defaultRoles))
 	ret.defaultRoles = append(ret.defaultRoles, p.defaultRoles...)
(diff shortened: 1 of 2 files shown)


pkg/privilege/privileges/cache.go
```diff
@@ -625,19 +627,27 @@ func compareColumnsPrivRecord(x, y columnsPrivRecord) int {
 func compareHost(x, y string) int {
```
ystaticy, line 627: How about adding a UT for compareHost?

wjhuang2016, line 627: Why do you change it?

pkg/privilege/privileges/cache.go
```diff
@@ -1364,15 +1391,14 @@ func (p *MySQLPrivilege) RequestVerification(activeRoles []*auth.RoleIdentity, u
 		if tableRecord != nil {
 			tablePriv |= tableRecord.TablePriv
 			if column != "" {
-				columnPriv |= tableRecord.ColumnPriv
+				columnPriv |= tableRecord.TablePriv
```
wjhuang2016, line 1395: Why change it?

pkg/privilege/privileges/cache.go
CbcWestwolf: [`817dfea` (#61677)](https://github.com/pingcap/tidb/pull/61677/commits/817dfea1b67dfdbff86690c3d18c8e4809cd957a)

pkg/privilege/privileges/cache.go
CbcWestwolf: Previous implementation is wrong here.

CbcWestwolf: If a user has table-level `SELECT` priv, it can pass all columns' `SELECT` check in this table.
But in previous implementation, ANY column's `SELECT` priv makes this check pass.

pkg/privilege/privileges/cache.go
CbcWestwolf: See this change of test https://github.com/pingcap/tidb/pull/61677/files#diff-860eeb97f0492c2eeb38ecca6442094a483cb30f9afe7aac9f6ec4ce6b0d9a6bR413-R414

pkg/privilege/privileges/cache.go
```diff
@@ -1145,10 +1166,12 @@ func (record *tablesPrivRecord) match(user, host, db, table string) bool {
 }
 
 func (record *columnsPrivRecord) match(user, host, db, table, col string) bool {
+	// `SELECT COUNT(*) ...` requires a column-level SELECT privilege of any column,
+	// so we add a special case "*" here
 	return record.baseRecord.match(user, host) &&
 		strings.EqualFold(record.DB, db) &&
 		strings.EqualFold(record.TableName, table) &&
-		strings.EqualFold(record.ColumnName, col)
+		(strings.EqualFold(record.ColumnName, col) || col == "*")
```
tiancaiamao: The requirement is to match **any** column.
But here it actually match **the first** column.

For example, we have three columns a, b, c where
a: no privilege
b: write-only privilege
c: read-only privilege

The match function returns record of b (the first match) ... but it's not correct.
Later, columnPriv&select_priv would return 0 and the check fail, right?

tiancaiamao: See https://github.com/pingcap/tidb/pull/61677/files#diff-c3393c2c85014927cf2d80a965fb61f15fdc3f375d9deefeb167ef3057c90f7eR1303

pkg/privilege/privileges/cache_test.go
```diff
@@ -383,16 +384,34 @@ func TestFindAllUserEffectiveRoles(t *testing.T) {
...
 	})
 	p.SortUserTable()
-	result := []privileges.UserRecord{
-		privileges.NewUserRecord("localhost", "root"),
+	result = []privileges.UserRecord{
 		privileges.NewUserRecord("localhost", ""),
+		privileges.NewUserRecord("localhost", "root"),
 		privileges.NewUserRecord(`%`, "jeffrey"),
 		privileges.NewUserRecord(`%`, "root"),
 	}
```
tiancaiamao, line 448: Is there any case cover the comparison of columnsPrivRecord, including the special `*` rule?

pkg/privilege/privileges/cache.go
CbcWestwolf: Nice catch! 
I fix it in [`373492b` (#61677)](https://github.com/pingcap/tidb/pull/61677/commits/373492be112cd49d2d1044d0bdf3153ca3786c8c), by checking that the record should has `SelectPriv`. I will add this as a test case in the phase 2's PR.

pkg/privilege/privileges/cache_test.go
CbcWestwolf: Added in [`6956abf` (#61677)](https://github.com/pingcap/tidb/pull/61677/commits/6956abf5fdf38796ea3028c36d166dc71124aaaf)

ystaticy (approved):

wjhuang2016 (approved):

tiancaiamao (approved):

PR #61679, opened 2025-06-11
*: adapt the show columns, show index and rename for column privilege

CbcWestwolf:

Issue Number: close #xxx

Problem Summary:

Adapt the show columns, show index and rename for column privilege
This PR pick `pkg/executor/show.go`, `pkg/executor/simple.go`, `pkg/parser/auth/auth.go`, `tests/integrationtest/t/executor/show.test` and `tests/integrationtest/r/executor/show.result` in https://github.com/pingcap/tidb/pull/61638

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/show.go
+++ b/pkg/executor/show.go
@@ -694,10 +694,16 @@ func (e *ShowExec) fetchShowColumns(ctx context.Context) error {
 		fieldPatternsLike = e.Extractor.FieldPatternLike()
 	}
 
+	passTblPrivCheck, passColPrivCheck := false, false
 	checker := privilege.GetPrivilegeManager(e.Ctx())
 	activeRoles := e.Ctx().GetSessionVars().ActiveRoles
-	if checker != nil && e.Ctx().GetSessionVars().User != nil && !checker.RequestVerification(activeRoles, e.DBName.O, tb.Meta().Name.O, "", mysql.InsertPriv|mysql.SelectPriv|mysql.UpdatePriv|mysql.ReferencesPriv) {
-		return e.tableAccessDenied("SELECT", tb.Meta().Name.O)
+	priv := mysql.InsertPriv | mysql.SelectPriv | mysql.UpdatePriv | mysql.ReferencesPriv
+	if checker != nil && e.Ctx().GetSessionVars().User != nil {
+		if checker.RequestVerification(activeRoles, e.DBName.O, tb.Meta().Name.O, "", priv) {
+			passTblPrivCheck = true
+		}
+	} else {
+		passTblPrivCheck = true
 	}
 
 	var cols []*table.Column
@@ -716,6 +722,13 @@ func (e *ShowExec) fetchShowColumns(ctx context.Context) error {
 			continue
 		} else if fieldPatternsLike != nil && !fieldPatternsLike.DoMatch(col.Name.L) {
 			continue
+		}
+		if !passTblPrivCheck {
+			if !checker.RequestVerification(activeRoles, e.DBName.O, tb.Meta().Name.O, col.Name.O, priv) {
+				// check privileges in column level
+				continue
+			}
+			passColPrivCheck = true
 		}
 		desc := table.NewColDesc(col)
 		var columnDefault any
@@ -763,6 +776,9 @@ func (e *ShowExec) fetchShowColumns(ctx context.Context) error {
 			})
 		}
 	}
+	if !passTblPrivCheck && !passColPrivCheck {
+		return e.tableAccessDenied("SELECT", tb.Meta().Name.O)
+	}
 	return nil
 }
 
@@ -777,10 +793,20 @@ func (e *ShowExec) fetchShowIndex() error {
 
 	statsTbl := h.GetTableStats(tb.Meta())
 
+	// SHOW INDEX requires some privilege for any column in the table.
 	checker := privilege.GetPrivilegeManager(e.Ctx())
 	activeRoles := e.Ctx().GetSessionVars().ActiveRoles
-	if checker != nil && e.Ctx().GetSessionVars().User != nil && !checker.RequestVerification(activeRoles, e.DBName.O, tb.Meta().Name.O, "", mysql.AllPrivMask) {
-		return e.tableAccessDenied("SELECT", tb.Meta().Name.O)
+	if checker != nil && e.Ctx().GetSessionVars().User != nil {
+		passCheck := false
+		for _, col := range tb.VisibleCols() {
+			if checker.RequestVerification(activeRoles, e.DBName.O, tb.Meta().Name.O, col.Name.O, mysql.AllPrivMask) {
+				passCheck = true
+				break
+			}
+		}
+		if !passCheck {
+			return e.tableAccessDenied("SELECT", tb.Meta().Name.O)
+		}
 	}
 
 	if tb.Meta().PKIsHandle {
@@ -2010,12 +2036,7 @@ func (e *ShowExec) dbAccessDenied() error {
 
 func (e *ShowExec) tableAccessDenied(access string, table string) error {
 	user := e.Ctx().GetSessionVars().User
-	u := user.Username
-	h := user.Hostname
-	if len(user.AuthUsername) > 0 && len(user.AuthHostname) > 0 {
-		u = user.AuthUsername
-		h = user.AuthHostname
-	}
+	u, h := auth.GetUserAndHostName(user)
 	return exeerrors.ErrTableaccessDenied.GenWithStackByArgs(access, u, h, table)
 }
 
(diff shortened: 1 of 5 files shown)


pkg/executor/show.go
```diff
@@ -694,10 +694,16 @@ func (e *ShowExec) fetchShowColumns(ctx context.Context) error {
 		fieldPatternsLike = e.Extractor.FieldPatternLike()
 	}
 
+	passTblPrivCheck, passColPrivCheck := false, false
 	checker := privilege.GetPrivilegeManager(e.Ctx())
 	activeRoles := e.Ctx().GetSessionVars().ActiveRoles
-	if checker != nil && e.Ctx().GetSessionVars().User != nil && !checker.RequestVerification(activeRoles, e.DBName.O, tb.Meta().Name.O, "", mysql.InsertPriv|mysql.SelectPriv|mysql.UpdatePriv|mysql.ReferencesPriv) {
-		return e.tableAccessDenied("SELECT", tb.Meta().Name.O)
+	priv := mysql.InsertPriv | mysql.SelectPriv | mysql.UpdatePriv | mysql.ReferencesPriv
```
ystaticy, line 700: Is this just optimizing the code format?

pkg/executor/show.go
```diff
@@ -717,6 +723,13 @@ func (e *ShowExec) fetchShowColumns(ctx context.Context) error {
 		} else if fieldPatternsLike != nil && !fieldPatternsLike.DoMatch(col.Name.L) {
 			continue
 		}
+		if !passTblPrivCheck {
+			if !checker.RequestVerification(activeRoles, e.DBName.O, tb.Meta().Name.O, col.Name.O, priv) {
```
ystaticy, line 732: How about use
```
if checker.RequestVerification(col) {
      passColPrivCheck = true
}
```
instead of 
```
if !checker.RequestVerification(col) {
    continue
}
```

CbcWestwolf: We need to skip the record if it lacks the column's privilege. So a `continue` is needed here.

ystaticy: @CbcWestwolf Got it. I think adding a comment would be better.

CbcWestwolf: Sure, added in [`4fe5a54` (#61679)](https://github.com/pingcap/tidb/pull/61679/commits/4fe5a540468022828beed9364a3cd1b0ba001435)

pkg/executor/show.go
```diff
@@ -694,10 +694,21 @@ func (e *ShowExec) fetchShowColumns(ctx context.Context) error {
 		fieldPatternsLike = e.Extractor.FieldPatternLike()
 	}
 
+	// SHOW COLUMNS displays information only for those columns for which you have some privilege.
+	// https://dev.mysql.com/doc/refman/8.4/en/show-columns.html
+	// 1. If you have table privilege, all columns can be shown
+	// 2. If you have only some columns privielge, these columns can be shown
+	// 3. If you have neither table or column privilege, an error is returned
+	passTblPrivCheck, passColPrivCheck := false, false
```
tiancaiamao, line 702: What happen to 'show create table'? will we get the full table definition or just some column?

pkg/executor/show.go
```diff
@@ -777,10 +798,21 @@ func (e *ShowExec) fetchShowIndex() error {
 
 	statsTbl := h.GetTableStats(tb.Meta())
 
+	// SHOW INDEX requires some privilege for any column in the table.
+	// https://dev.mysql.com/doc/refman/8.4/en/show-index.html
 	checker := privilege.GetPrivilegeManager(e.Ctx())
 	activeRoles := e.Ctx().GetSessionVars().ActiveRoles
```
tiancaiamao, line 804: will 'show columns' work correctly on view?
And especially when we have some column alias in the view:

```
CREATE VIEW myView AS (
  SELECT 
    id AS friendly_id,
    value AS friendly_value
  FROM table
);
```

tiancaiamao: And we also need to check column privilege of any kind of column alias usage, as long as the SQL gramma permit it.
There should be no prior test cover such scenario to my knowledge.

tests/integrationtest/r/executor/show.result
```diff
@@ -1141,3 +1141,51 @@ t	CREATE TABLE `t` (
...
+Field	Type	Null	Key	Default	Extra
+a	int	YES		NULL	
+b	int	YES		NULL	
+drop database test_show_columns_db;
+drop user test_show_columns_user;
+create database test_show_index_db;
+create user test_show_index_user;
+create table test_show_index_db.t (a int, b int);
+alter table test_show_index_db.t add index i1(b);
+alter table test_show_index_db.t add index i2(b);
```
tiancaiamao, line 1191: both index on b?
How about index on a and multiple value index on (a, b)?

tests/integrationtest/r/executor/show.result
```diff
@@ -1141,3 +1141,51 @@ t	CREATE TABLE `t` (
...
+t	1	i2	1	b	A	0	NULL	NULL	YES	BTREE			YES	NULL	NO	NO
+// with table privilege, indexes are shown.
+revoke update(a) on test_show_index_db.t from test_show_index_user;
+grant references on test_show_index_db.t to test_show_index_user;
+show index from test_show_index_db.t;
+Table	Non_unique	Key_name	Seq_in_index	Column_name	Collation	Cardinality	Sub_part	Packed	Null	Index_type	Comment	Index_comment	Visible	Expression	Clustered	Global
+t	1	i1	1	b	A	0	NULL	NULL	YES	BTREE			YES	NULL	NO	NO
+t	1	i2	1	b	A	0	NULL	NULL	YES	BTREE			YES	NULL	NO	NO
+drop database test_show_index_db;
+drop user test_show_index_user;
```
tiancaiamao, line 1214: Do we also have tests that cover the DDL operations, like add/drop index where we just have privilege for some columns ...

pkg/executor/show.go
CbcWestwolf: After testing in mysql, **any** table-level privilege can get the full table definition.
Column level `SELECT` can not execute `show create table` even if the user has all columns' priv.
TiDB is compatible with mysql

pkg/executor/show.go
CbcWestwolf: Yes, `show columns from view` is the same as from table

tests/integrationtest/r/executor/show.result
CbcWestwolf: Most of the DDL operations require table privileges like `INDEX` for `CREATE INDEX`.
Check `func (b *PlanBuilder) buildDDL` for more details.

tests/integrationtest/r/executor/show.result
CbcWestwolf: All of the indexed columns will be shown. Update test in [`112f4fe` (#61679)](https://github.com/pingcap/tidb/pull/61679/commits/112f4fe4a12c9353bf745fad001d5080748bbc16)

pkg/executor/show.go
CbcWestwolf: I add a case `show columns from view` in [`ef333d5` (#61679)](https://github.com/pingcap/tidb/pull/61679/commits/ef333d53f83a82c4e8b83cd2c6fc19d717d0b71e), which considers column alias. 
We need to test more alias cases in https://github.com/pingcap/tidb/pull/61741

ystaticy (approved):

pkg/executor/simple.go
```diff
@@ -2168,6 +2168,12 @@ func (e *SimpleExec) executeRenameUser(s *ast.RenameUserStmt) error {
 			break
```
tiancaiamao, line 2168: will drop user update the column tables?

tests/integrationtest/r/executor/show.result
tiancaiamao: So is this the same behavior as MySQL, modify index only check `CREATE_INDEX`, no matter whether the index references a non-privilege column?

tiancaiamao (approved):

pkg/executor/simple.go
CbcWestwolf: We can understand the manner in this way: DDL would not affect privilege tables, but DCL will

CbcWestwolf:
> So is this the same behavior as MySQL, modify index only check CREATE_INDEX, no matter whether the index references a non-privilege column?

@tiancaiamao Yes, only `INDEX` privilege is needed.

tiancaiamao (approved):

PR #61714, opened 2025-06-12
planner: optimize bucket collection

### LINKED ISSUE #61738: Reduce overhead for removing TopN from samples
## Enhancement

During stats collection, we will build TopN and then remove those from the samples before building buckets. The current process has a loop within a loop - which can be optimized to improve performance.

ichen032:

Issue Number: close #61738

Problem Summary:

Modifies BuildHistAndTopN function to reduce redundant operations when removing topn from the samples for bucket collection. When iterating through the samples, this change compares the frequency of the current value with the least frequent topn value in order to skip unnecessary iterations of an inner loop. 
Benchmark Test Results:
| | minTopNCount Branch | Master Branch | 
| --- | --- | --- |
| BenchmarkBuildHistAndTopN | 1,249,879,721 ns/op <br>32,016,465 B/op <br>1,997,787 allocs/op | 1,246,620,408 ns/op <br>32,016,448 B/op <br>1,997,787 allocs/op |
| BenchmarkBuildHistAndTopNIncreaseSkew | 153,580,312 ns/op <br>56,433,903 B/op <br>3,525,843 allocs/op | 381,080,775 ns/op <br>37,936,000 B/op <br>2,369,732 allocs/op |
| BenchmarkBuildHistAndTopNLargeNDV | 1,542,063,958 ns/op <br>571,980,995 B/op <br>35,747,528 allocs/op | 3,707,413,654 ns/op <br>383,104,660 B/op <br>23,942,765 allocs/op |

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/builder.go
+++ b/pkg/statistics/builder.go
@@ -420,21 +420,21 @@ func BuildHistAndTopN(
 	lenTopN := int64(len(topn.TopN))
 
 	// Step2: exclude TopN from samples if the NDV is larger than the number of topN items.
-	lenSamples := uint64(len(samples))
+	lenSamples := int64(len(samples))
 	if lenTopN > 0 && lenTopN < hg.NDV && lenSamples > 0 {
-		minTopNCnt := int64(topNList[len(topNList)-1].Count - 1)
-		for i := int64(0); uint64(i) < lenSamples; i++ {
+		minTopNCnt := int64(topNList[len(topNList)-1].Count)
+		for i := int64(0); i < lenSamples; i++ {
 			sampleBytes, err := getComparedBytes(samples[i].Value)
 			if err != nil {
 				return nil, nil, errors.Trace(err)
 			}
 			// If current sample value is less frequent than the least frequent topn value, then it cannot
 			// be a topn value. This check is only helpful if the least frequent topn n value occurs more than once.
-			if minTopNCnt > 0 {
-				if i+minTopNCnt >= int64(lenSamples) {
+			if minTopNCnt > 1 {
+				if i+minTopNCnt-1 >= lenSamples {
 					continue
 				}
-				nextBytes, err := getComparedBytes(samples[i+minTopNCnt].Value)
+				nextBytes, err := getComparedBytes(samples[i+minTopNCnt-1].Value)
 				if err != nil {
 					return nil, nil, errors.Trace(err)
 				}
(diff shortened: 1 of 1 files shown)


pkg/statistics/builder.go
```diff
@@ -468,8 +475,8 @@ func BuildHistAndTopN(
 					firstTimeSample = samples[i].Value
 					// Found the same value in topn: need to skip over this value in samples.
 					copy(samples[i:], samples[uint64(i)+topNList[j].Count:])
```
winoros, line 477: the `copy` here is o(n), which can introduce unnecessary time complexity.
we can also optimize the algorithm to not use `copy` later.

pkg/statistics/builder.go
```diff
@@ -420,20 +420,27 @@ func BuildHistAndTopN(
...
-			// Safety:
-			// When a sample value matches an entry in the topN list, it is removed from the samples array,
-			// so it can never match the previous sample. Therefore, this equality check will never be true
-			// in that case. For values not found in topN, we don't need keep checking the same value.
-			if i > 0 && bytes.Equal(sampleBytes, samples[i-1].Value.GetBytes()) {
-				// If the sample is the same as the previous one, we can skip it.
-				continue
+			// If current sample value is less frequent than the least frequent topn value, then it cannot
+			// be a topn value. This check is only helpful if the least frequent topn n value occurs more than once.
+			if minTopNCnt > 0 {
```
winoros: Now that it only takes effect when `the least frequent topn n value occurs more than once`, we can directly use `if minTopCnt > 1`?

winoros:
And for the memory allocation mentioned by @0xPoe 
Seems that the allocation is coming from
```
		getComparedBytes = func(datum types.Datum) ([]byte, error) {
			encoded, err := codec.EncodeKey(ctx.GetSessionVars().StmtCtx.TimeZone(), nil, datum)
			err = ctx.GetSessionVars().StmtCtx.HandleError(err)
			if memTracker != nil {
				// tmp memory usage
				deltaSize := int64(cap(encoded))
				memTracker.BufferedConsume(&bufferedMemSize, deltaSize)
				memTracker.BufferedRelease(&bufferedReleaseSize, deltaSize)
			}
			return encoded, err
		}
```

So we can use a byte slice buffer to reduce the memory usage. Then I think there will be no regression anymore.

The second param of the `EncodeKey` is just the buffer. I think the diff size is limited.

But if this makes the pr too large. The changes can be split into an individual pr.


The general idea LGTM🥰

Issue #61706, opened 2025-06-12
Implement column level privilege on v8.5


CbcWestwolf:
We need to implement column-level privilege on https://github.com/pingcap/tidb/tree/release-8.5-20250606-v8.5.2. For efficient review, we split the huge https://github.com/pingcap/tidb/pull/61638 into 3 phases

### Phase 1
1. [P0] Adapt the `GRANT` and `REVOKE` for column privilege https://github.com/pingcap/tidb/pull/61678
2. [P1] Adapt the `MySQLPrivilege` to support column privilege https://github.com/pingcap/tidb/pull/61677
3. [P2] Adapt the `show columns`, `show index` and `rename` for column privilege https://github.com/pingcap/tidb/pull/61679

### Phase 2
1. [P0] Adapt the normal plan build process to collect column-level visit info https://github.com/pingcap/tidb/pull/61741

### Phase 3
1. [P0] Adapt the point-get plan build process https://github.com/pingcap/tidb/pull/62024
2. [P2] Introduces column-priv view in `information_schema` https://github.com/pingcap/tidb/pull/61777

### Performance enhancement:
[P0] After the 3 phases, we also need to cherry-pick #58945 for improving performance: https://github.com/pingcap/tidb/pull/61719

### Bug fix & test cases && enhancement:
* https://github.com/pingcap/tidb/pull/62241
* https://github.com/pingcap/tidb/pull/62443
* https://github.com/pingcap/tidb/pull/62693
* https://github.com/pingcap/tidb/pull/63175
* https://github.com/pingcap/tidb/pull/63728

CbcWestwolf:
There is another issue tracking this feature: https://github.com/pingcap/tidb/issues/9766

PR #61741, opened 2025-06-15
planner: adapt the normal plan build process to collect column-level visit info | tidb-test=pr/2532

CbcWestwolf:

Issue Number: ref #61706

Problem Summary:

Utilize the `expressionRewriter.toColumn` to collect visited column during building logical plan

Related test: https://github.com/PingCAP-QE/tidb-test/pull/2532

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/parser/ast/dml.go
+++ b/pkg/parser/ast/dml.go
@@ -714,7 +714,9 @@ type SelectField struct {
 
 	// IsUnfoldFromWildCard indicates whether this field is unfolded from a wildcard, which can be used in checking privilege.
 	// Although we always check SELECT privilege in column-level, a table-level access deny error will be return if the
-	// column is unfolded from a wildcard.
+	// column is unfolded from a wildcard. For example, assume that table t has columns (a,b,c)
+	// Both `SELECT * FROM t` and `SELECT a,b,c FROM t`` require SELECT privilege of a,b and c.
+	// But if lacking privilege, the former reports error code 1142, the latter reports 1143.
 	IsUnfoldFromWildCard bool
 }
 
--- a/pkg/planner/core/expression_rewriter.go
+++ b/pkg/planner/core/expression_rewriter.go
@@ -2422,6 +2422,16 @@ func (er *expressionRewriter) funcCallToExpression(v *ast.FuncCallExpr) {
 	}
 }
 
+// Assume that there are two tables t and t1, which contain a and b columns individually. In MySQL:
+// -- Require `a` or `b` SELECT privilege
+// select count(*) from t;
+// select count(1) from t;
+// -- Require `a` SELECT privilege
+// select count(a) from t;
+// -- Require `b` SELECT privilege
+// select count(b) from t;
+// -- Require `SELCT` privilege of `t.a` and `t1.a`
+// select count(*) from t join t1 on t.a = t1.a;
 func (er *expressionRewriter) collectPrivsForCount() {
 	b := er.planCtx.builder
 	if b == nil {


pkg/planner/core/mock.go
```diff
@@ -414,6 +414,38 @@ func MockView() *model.TableInfo {
 	return table
 }
 
+// MockViewInvoker is only used for plan related tests.
+func MockViewInvoker() *model.TableInfo {
```
tiancaiamao: We can put it to the _test.go file if it is used just there.

pkg/parser/ast/dml.go
```diff
@@ -707,6 +711,11 @@ type SelectField struct {
 	Auxiliary             bool
 	AuxiliaryColInAgg     bool
 	AuxiliaryColInOrderBy bool
+
+	// IsUnfoldFromWildCard indicates whether this field is unfolded from a wildcard, which can be used in checking privilege.
+	// Although we always check SELECT privilege in column-level, a table-level access deny error will be return if the
```
ystaticy, line 716: Is this the logic in the design? Or is it the same as the MySQL logic? Can we add an explanation?

CbcWestwolf: Yes. Assume that table `t`  has columns `(a,b,c)`
Both `SELECT * FROM t` and `SELECT a,b,c FROM t` require `SELECT` privilege of `a`,`b` and `c`.
But if lacking privilege, the former reports error code `1142`, the latter reports `1143`.

pkg/planner/core/expression_rewriter.go
```diff
@@ -1432,8 +1442,12 @@ func (er *expressionRewriter) Leave(originInNode ast.Node) (retNode ast.Node, ok
 	}
 
 	switch v := inNode.(type) {
-	case *ast.AggregateFuncExpr, *ast.ColumnNameExpr, *ast.ParenthesesExpr, *ast.WhenClause,
-		*ast.SubqueryExpr, *ast.ExistsSubqueryExpr, *ast.CompareSubqueryExpr, *ast.ValuesExpr, *ast.WindowFuncExpr, *ast.TableNameExpr:
+	case *ast.AggregateFuncExpr:
+		if v.F == ast.AggFuncCount {
+			er.collectPrivsForCount()
```
CbcWestwolf, line 1448: To collect any columns' `SELECT` privilege of the table(s)

pkg/planner/core/logical_plan_builder.go
```diff
@@ -1334,7 +1349,14 @@ func (b *PlanBuilder) buildProjection(ctx context.Context, p base.LogicalPlan, f
 			newNames = append(newNames, name)
 			continue
 		}
+		originColPriv := b.checkColPriv
+		if field.IsUnfoldFromWildCard {
+			b.checkColPriv = reportTableErrOption
+		} else {
+			b.checkColPriv = reportColumnErrOption
+		}
 		newExpr, np, err := b.rewriteWithPreprocess(ctx, field.Expr, p, mapper, windowMapper, true, nil)
```
CbcWestwolf, line 1382: If the `field.Expr` is a column name `a`, it may be unfolded from a wildcard `select * from t` or from `select a from t`

pkg/planner/core/logical_plan_builder.go
```diff
@@ -670,7 +677,10 @@ func (b *PlanBuilder) buildJoin(ctx context.Context, joinNode *ast.Join) (base.L
 		}
 	} else if joinNode.On != nil {
 		b.curClause = onClause
+		originColPriv := b.checkColPriv
+		b.checkColPriv = reportColumnErrOption
```
CbcWestwolf, line 681: We need to explicitly enable the collection of columns here, for the `rewrite` in line 682.

pkg/planner/core/logical_plan_builder.go
```diff
@@ -4907,8 +4942,12 @@ func (b *PlanBuilder) checkRecursiveView(dbName pmodel.CIStr, tableName pmodel.C
 // qbNameMap4View maps the query block name to the view table lists.
 // viewHints group the view hints based on the view's query block name.
 func (b *PlanBuilder) BuildDataSourceFromView(ctx context.Context, dbName pmodel.CIStr, tableInfo *model.TableInfo, qbNameMap4View map[string][]ast.HintTable, viewHints map[string][]*ast.TableOptimizerHint) (base.LogicalPlan, error) {
-	viewDepth := b.ctx.GetSessionVars().StmtCtx.ViewDepth
-	b.ctx.GetSessionVars().StmtCtx.ViewDepth++
+	stmtCtx := b.ctx.GetSessionVars().StmtCtx
+	viewDepth := len(stmtCtx.SavedViews)
+	stmtCtx.SavedViews = append(stmtCtx.SavedViews, &ast.TableName{Schema: dbName, Name: tableInfo.Name})
```
CbcWestwolf: `SavedViews` is a stack that saves all views when traversing the AST. 
We depend on it to report precise error in `appendColumnsToVisitedInfo`

pkg/planner/core/logical_plans_test.go
```diff
@@ -1128,79 +1162,30 @@ func TestAggPrune(t *testing.T) {
 }
 
 func TestVisitInfo(t *testing.T) {
+	t.Run("test non-column privilege visit info", testNormalVisitInfo)
+	t.Run("test column privilege visit info", testColumnPrivilegeVisitInfo)
```
CbcWestwolf, line 1166: `testColumnPrivilegeVisitInfo` is the core testing function that considers many cases.

pkg/planner/core/logical_plans_test.go
```diff
@@ -1547,32 +1524,632 @@ func TestVisitInfo(t *testing.T) {
...
+			sql: "insert into t (a) values (1)",
+			ans: []visitInfo{
+				{mysql.InsertPriv, "test", "t", "a", nil, false, nil, false},
+			},
+		},
+		{
+			sql: "delete from t order by a",
+			ans: []visitInfo{
+				{mysql.DeletePriv, "test", "t", "", plannererrors.ErrTableaccessDenied.FastGenByArgs("DELETE", "", "", "t"), false, nil, false},
+				{mysql.SelectPriv, "test", "t", "a", plannererrors.ErrColumnaccessDenied.FastGenByArgs("SELECT", "", "", "a", "t"), false, nil, false},
```
CbcWestwolf, line 1552: This `visitInfo` means that:
> we should have a `SELECT` privilege of column `a` in `test.t`, otherwise an error of `plannererrors.ErrColumnaccessDenied.FastGenByArgs("SELECT", "", "", "a", "t")` will be reported

pkg/planner/core/logical_plans_test.go
```diff
@@ -1547,32 +1524,632 @@ func TestVisitInfo(t *testing.T) {
...
+			},
+		},
+		{
+			sql: "delete from a1 using t as a1 inner join t as a2 where a1.a = a2.a",
+			ans: []visitInfo{
+				{mysql.DeletePriv, "test", "t", "", nil, false, nil, false},
```
CbcWestwolf, line 1558: This `visitInfo` means that:
> we should have a `DELETE` privilege of table `test.t`, otherwise an error of will be reported

pkg/planner/core/expression_rewriter.go
ystaticy: @CbcWestwolf Don't other aggregation operations require privilege checks?

pkg/planner/core/expression_rewriter.go
```diff
@@ -332,6 +339,9 @@ type exprRewriterPlanCtx struct {
 	insertPlan *Insert
 
 	rollExpand *logicalop.LogicalExpand
+
+	// columnVisited records the visited columns, which is used for checking columns privilege
```
ystaticy: It's better to note in the comment that columnVisited only precollects all necessary columns in the request for unified permission checks at the xxx location later.

pkg/planner/core/expression_rewriter.go
```diff
@@ -332,6 +339,9 @@ type exprRewriterPlanCtx struct {
...
+	columnVisited []*ast.ColumnName
```
ystaticy: How about optimizing the variable name columnVisited to imply precollection? The current name suggests that columns have already been processed/checked, which might cause confusion.

pkg/planner/core/expression_rewriter.go
```diff
@@ -2640,3 +2695,64 @@ func hasCurrentDatetimeDefault(col *model.ColumnInfo) bool {
 	}
 	return strings.ToLower(x) == ast.CurrentTimestamp
 }
+
+func (b *PlanBuilder) appendColumnsToVisitedInfo(columnVisited []*ast.ColumnName) {
```
ystaticy: How about optimizing the function name to imply pre-adding info into a collection for direct use during xxx checks?

pkg/planner/core/expression_rewriter.go
CbcWestwolf: Other (aggregation) functions will require the column inside them.
But `COUNT` has some special cases. 
Assume that there are two tables `t` and `t1`, which contain `a` and `b` columns individually.
In MySQL:

```sql
-- Require `a` or `b` SELECT privilege
select count(*) from t;
select count(1) from t;

-- Require `a` SELECT privilege
select count(a) from t;
-- Require `b` SELECT privilege
select count(b) from t;

-- Require `SELCT` privilege of `t.a` and `t1.a`
select count(*) from t join t1 on t.a = t1.a;
```

So we have to handle `COUNT` here

pkg/planner/core/expression_rewriter.go
CbcWestwolf: Done in [`2c5bcb0` (#61741)](https://github.com/pingcap/tidb/pull/61741/commits/2c5bcb01fe197e24e4fa337da334c2819b412eed)

pkg/planner/core/expression_rewriter.go
```diff
@@ -2407,6 +2422,38 @@ func (er *expressionRewriter) funcCallToExpression(v *ast.FuncCallExpr) {
 	}
 }
 
+func (er *expressionRewriter) collectPrivsForCount() {
+	// dbName -> tableName
+	tableNames := make(map[string]map[string]any)
```
tiancaiamao: This line can be moved to line 2432 so when er.planCtx.builder == nil, unnecessary object allocation is avoid.

tiancaiamao: And it can be `make(map[string]map[string]struct{})` since any is never used.

pkg/planner/core/expression_rewriter.go
```diff
@@ -2640,3 +2696,64 @@ func hasCurrentDatetimeDefault(col *model.ColumnInfo) bool {
 	}
 	return strings.ToLower(x) == ast.CurrentTimestamp
 }
+
+func (b *PlanBuilder) appendColNamesToVisitInfo(columnVisited []*ast.ColumnName) {
+	user, host := auth.GetUserAndHostName(b.ctx.GetSessionVars().User)
+	views := b.ctx.GetSessionVars().StmtCtx.SavedViews
+	switch {
+	case len(views) > 0:
+		view := views[len(views)-1]
```
tiancaiamao, line 2715: why the error report is always the last view, will there be multiple views?

wjhuang2016 (dismissed):

pkg/planner/core/logical_plan_builder.go
tiancaiamao: Why put it in StmtCtx? can we put it to PlanBuilder or somewhere else?

pkg/planner/core/logical_plan_builder.go
```diff
@@ -4998,36 +5038,37 @@ func (b *PlanBuilder) BuildDataSourceFromView(ctx context.Context, dbName pmodel
 			terror.ErrorNotEqual(err, plannererrors.ErrMixOfGroupFuncAndFields) &&
 			terror.ErrorNotEqual(err, plannererrors.ErrViewNoExplain) &&
 			terror.ErrorNotEqual(err, plannererrors.ErrNotSupportedYet) {
-			err = plannererrors.ErrViewInvalid.GenWithStackByArgs(dbName.O, tableInfo.Name.O)
+			err = errViewInvalid
```
tiancaiamao: The old error comes with the correct stack that may help us to debug ...
And performance is .. er, worse now .. because if there is no error, there is no need to generate the error.
Especally `GenWithStack` is a heavy operation.

pkg/planner/core/planbuilder.go
```diff
@@ -3943,28 +3969,54 @@ func (b *PlanBuilder) buildInsert(ctx context.Context, insert *ast.InsertStmt) (
...
+				authErr = plannererrors.ErrColumnaccessDenied.FastGenByArgs("INSERT", user.AuthUsername, user.AuthHostname, col.Name.L, tableInfo.Name.L)
+			}
+			b.visitInfo = appendVisitInfo(b.visitInfo, mysql.InsertPriv, tnW.DBInfo.Name.L,
+				tableInfo.Name.L, col.Name.L, authErr)
+		}
+	} else {
+		if user != nil {
+			authErr = plannererrors.ErrTableaccessDenied.FastGenByArgs("INSERT", user.AuthUsername, user.AuthHostname, tableInfo.Name.L)
+		}
+		for _, col := range tableInPlan.VisibleCols() {
```
tiancaiamao: If we can extract the for loop into a function, we can save some lines, I see several repetitions of this pattern.

tiancaiamao: `appendVisitColumns( VisibleCols(), priv, tableName, ...)`

pkg/planner/core/expression_rewriter.go
CbcWestwolf: Updated

pkg/planner/core/expression_rewriter.go
CbcWestwolf: Here the `views[len(views)-1]` is the one that defines this "subquery", so it always reports an error for it.
For multiple/embedded views, after traversing one view, the view will be popped from the stack `SavedViews`.
Please refer to the new test case [`TestColumnPrivilege4Views`](https://github.com/pingcap/tidb/pull/61741/commits/31139a6c6aa8fa90f0f64622ab3f45a1a09fc1ef) for more details

pkg/planner/core/planbuilder.go
CbcWestwolf: Updated

ystaticy (approved):

wjhuang2016 (approved):

winoros (approved):
approve for hotfix, we may refine the implementation when merging it to master.

PR #61787, opened 2025-06-17
planner: refactor order ratio code | tidb-test=pr/2545

terry1purcell:

Issue Number: ref #62034 

Problem Summary:

While investigating issue 62034, it became apparent that there was an opportunity to refactor this code to make the estimation more consistent if the variable tidb_opt_ordering_index_selectivity_ratio is disabled or enabled.

Prior to this PR - if disabled, the rows are considered to be distributed throughout the range. If enabled, tidb_opt_ordering_index_selectivity_ratio results in a penalty being applied before the first row is found, but once found - the rows are considered clustered. It is more consistent if tidb_opt_ordering_index_selectivity_ratio only impacts when the first row is found, but not subsequent rows.

NOTE: On testcase result change. There is one large "estimated rows" change in pkg/planner/core/casetest/physicalplantest/testdata/plan_suite_out.json.

To test out that change in TestRuleColumnPruningLogicalApply - I used the below insert to insert 10,000 rows into tables t, t1, t2 & t3 (just change the "into" table each execution). With this PR - the new plan runs approximately 50% faster (58 seconds vs 1 minute 58 seconds) compared to the old plan - you can toggle the following variable to change the plan:

For old plan:
```
 set @@session.tidb_opt_ordering_index_selectivity_ratio=-1;
```
For new plan:
```
 set @@session.tidb_opt_ordering_index_selectivity_ratio=0.01;
```
To insert rows:
```
set @@cte_max_recursion_depth=10000000;
INSERT INTO t (a, b, c)
SELECT a, mod(a, 1000) AS b, a AS c
FROM (
    WITH RECURSIVE x AS (
        SELECT 1 AS a
        UNION ALL
        SELECT a + 1 AS a
        FROM x
        WHERE a < 10000
    )
    SELECT a
    FROM x
) AS subquery;
```
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/cardinality/cross_estimation.go
```diff
@@ -100,23 +100,39 @@ func AdjustRowCountForIndexScanByLimit(sctx planctx.PlanContext,
...
+		rowsToMeetFirst := 0.0
+		if len(path.IndexFilters) > 0 && path.CountAfterIndex > 0 && path.CountAfterAccess > path.CountAfterIndex {
+			// First target - if there is index filtering, we need to estimate the number of
+			// index rows that need to be scanned.
+			rowsToMeetFirst = path.CountAfterAccess - path.CountAfterIndex
+			if path.CountAfterIndex > dsStatsInfo.RowCount {
+				// After index filtering, consider table rows scanned.
+				rowsToMeetFirst += path.CountAfterIndex - dsStatsInfo.RowCount
+			}
+		} else if len(path.TableFilters) > 0 && path.CountAfterAccess > dsStatsInfo.RowCount {
```
qw4990: Should we use 2 `if` instead of `if` and `else if`? For example:
```
if len(path.IndexFilters) > 0 && path.CountAfterIndex > 0 && path.CountAfterAccess > path.CountAfterIndex {}
if len(path.TableFilters) > 0 && path.CountAfterAccess > dsStatsInfo.RowCount {}
```
Since in some cases, the path might have both `IndexFilters` and `TableFilters`:
```
mysql> explain select * from t where a>99 and c>1 and d>1 limit 1;
+-------------------------------+---------+-----------+-----------------------------+-------------------------------------------------+
| id                            | estRows | task      | access object               | operator info                                   |
+-------------------------------+---------+-----------+-----------------------------+-------------------------------------------------+
| Limit_10                      | 1.00    | root      |                             | offset:0, count:1                               |
| └─IndexLookUp_22              | 1.00    | root      |                             |                                                 |
|   ├─Selection_19(Build)       | 2.88    | cop[tikv] |                             | gt(test.t.c, 1)                                 |
|   │ └─IndexRangeScan_17       | 8.63    | cop[tikv] | table:t, index:abc(a, b, c) | range:(99,+inf], keep order:false, stats:pseudo |
|   └─Limit_21(Probe)           | 1.00    | cop[tikv] |                             | offset:0, count:1                               |
|     └─Selection_20            | 1.00    | cop[tikv] |                             | gt(test.t.d, 1)                                 |
|       └─TableRowIDScan_18     | 2.88    | cop[tikv] | table:t                     | keep order:false, stats:pseudo                  |
+-------------------------------+---------+-----------+-----------------------------+-------------------------------------------------+
```

terry1purcell: The first IF check covers both index and table filters (or only index filters if no table filters). The 2nd "else if" is if there are no index filters, and only table filters.

Therefore, there isn't a need for separate IFs.

pkg/planner/cardinality/cross_estimation.go
```diff
@@ -100,23 +100,39 @@ func AdjustRowCountForIndexScanByLimit(sctx planctx.PlanContext,
...
+			// Second target - if there is no index filtering, but there is table filtering,
+			// we need to estimate the number of table rows that need to be scanned.
+			rowsToMeetFirst = (path.CountAfterAccess - dsStatsInfo.RowCount)
 		}
+		rowsToMeetFirst = max(0, (rowsToMeetFirst-rowCount)) * orderRatio
```
time-and-fate: I think we can directly use `rowsToMeetFirst = (path.CountAfterAccess-rowCount) * orderRatio` to make it more consistent with the recently added logic of the table scan estimation.

pkg/planner/cardinality/cross_estimation.go
```diff
@@ -100,23 +100,39 @@ func AdjustRowCountForIndexScanByLimit(sctx planctx.PlanContext,
...
+	if path.CountAfterAccess > rowCount && orderRatio >= 0 {
+		rowsToMeetFirst := 0.0
+		if len(path.IndexFilters) > 0 && path.CountAfterIndex > 0 && path.CountAfterAccess > path.CountAfterIndex {
+			// First target - if there is index filtering, we need to estimate the number of
+			// index rows that need to be scanned.
+			rowsToMeetFirst = path.CountAfterAccess - path.CountAfterIndex
+			if path.CountAfterIndex > dsStatsInfo.RowCount {
+				// After index filtering, consider table rows scanned.
+				rowsToMeetFirst += path.CountAfterIndex - dsStatsInfo.RowCount
+			}
```
qw4990: `path.CounterAfterIndex` should be equal or larger than `dsStatsInfo.RowCount`, then there are 2 scenarios:
1. if `path.CounterAfterIndex > ds.StatsInfo.RowCount`, then `rowsToMeetFirst = (path.CountAfterAccess - path.CountAfterIndex) + (path.CountAfterIndex - dsStatsInfo.RowCount) = path.CountAfterAccess-dsStatsInfo.RowCount`.
2. if they are equal, then `rowsToMeetFirst = path.CountAfterAccess - path.CountAfterIndex = path.CountAfterAccess - dsStatsInfo.RowCount`.

Then for table-filters, the formula is also `path.CountAfterAccess - dsStatsInfo.RowCount`.

So finally it seems that we can simplify all if/else branches into one singel if branch:
```
if path.CountAfterAccess > dsStatsInfo.RowCount {
    // no matter where the filters are (in index side of table side), we assume all qualified rows
    // (dsStatsInfo.RowCount) are clustered at the end of all scanned rows (CountAfterAccess). 
    rowsToMeetFirst = (path.CountAfterAccess - dsStatsInfo.RowCount)
}
```

pkg/planner/cardinality/cross_estimation.go
terry1purcell: Latest commit implements suggested change.

pkg/planner/cardinality/cross_estimation.go
terry1purcell: Latest commit implements suggested change.

pkg/planner/cardinality/cross_estimation.go
```diff
@@ -100,22 +100,30 @@ func AdjustRowCountForIndexScanByLimit(sctx planctx.PlanContext,
...
+	// exists outside of that index. Such plans have high risk since we cannot estimate when rows will be found.
+	orderRatio := sctx.GetSessionVars().OptOrderingIdxSelRatio
+	sctx.GetSessionVars().RecordRelevantOptVar(vardef.TiDBOptOrderingIdxSelRatio)
+	if path.CountAfterAccess > rowCount && orderRatio >= 0 && (len(path.IndexFilters) > 0 || len(path.TableFilters) > 0) {
+		if orderRatio > 0 {
+			rowsToMeetFirst := (path.CountAfterAccess - rowCount) * orderRatio
+			rowCount += rowsToMeetFirst
+		} else if orderRatio == 0 {
+			// If the order ratio is 0, it means that we expect to find the rows immediately.
+			rowCount = min(rowCount, expectedCnt)
```
time-and-fate: Why do we have a special case just when `orderRatio == 0`? This will make `orderRatio == 0` has a very different behavior.
I think the formula above can handle this correctly.

qw4990 (approved):

pkg/planner/cardinality/cross_estimation.go
terry1purcell: There is a difference between 0 and > 0. 0 would assume that we find the rows immediately.

But you are correct in your implication that this is inconsistent with other usages of this variable - since all other use cases only apply if > 0. I have removed this branch (to support == 0) in the latest commit.

Issue #61792, opened 2025-06-18
estimation should not be zero when here is no buckets for index


AilinKid:
## Enhancement

same to https://github.com/pingcap/tidb/issues/59563, the index topN judgement may have the same senario, just like what shown in the pic

![Image](https://github.com/user-attachments/assets/895e2888-cee9-4ae9-bb5a-82f3ddcdd2cc)

the histNDV is greater than 0, is 4 here, and try to seek histogram buckets for computing avgRowEstimate, while according NDV collecting mechanism, the  histogram NDV is collected when scan all the data before sampling, while the topN is collected after sampling. So this is possible that, the former is greater than the later, while there is still no buckets collected.

PR #59564, opened 2025-02-16
planner: fix a estimation case when column.hist.NDV is greater than topN.num.

AilinKid:
<!--

Thank you for contributing to TiDB!

PR Title Format:
1. pkg [, pkg2, pkg3]: what's changed
2. *: what's changed

-->

### What problem does this PR solve?
<!--

Please create an issue first to describe the problem.

There MUST be one line starting with "Issue Number:  " and
linking the relevant issues via the "close" or "ref".

For more info, check https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/contribute-code.html#referring-to-an-issue.

-->

Issue Number: close https://github.com/pingcap/tidb/issues/59563
since the branch1 is targeted for no-hist cases, we could judge from reality: when column is full load and there is no buckets, which is indeed mean that this column should walk into branch1.

Problem Summary:

### What changed and how does it work?

### Check List

Tests <!-- At least one of them must be included. -->

- [ ] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)
```
tidb>  load stats '/Users/arenatlx/Downloads/tidb_stats_by_table_1738707809.json';
ERROR 2013 (HY000): Lost connection to MySQL server during query
No connection. Trying to reconnect...
Connection id:    2097154
Current database: cardcore_issuing

Query OK, 0 rows affected (0.88 sec)

tidb> use cardcore_issuing
Database changed
tidb> EXPLAIN SELECT * FROM   `tbl_cardcore_transaction` `transactio0_` WHERE   `transactio0_`.`period` = '202502'   AND `transactio0_`.`account_number` = '1901040107462200' ORDER BY   `transactio0_`.`transaction_status`,   `transactio0_`.`account_number`,   `transactio0_`.`entry_date` ASC,   `transactio0_`.`id` ASC;
+----------------------------------+---------+-----------+-------------------------------------------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| id                               | estRows | task      | access object                                                                                   | operator info                                                                                                                                                                                                                   |
+----------------------------------+---------+-----------+-------------------------------------------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
[...]

PR #61793, opened 2025-06-18
planner: fix a estimation case when index.hist.NDV is greater than index.topN.num.

AilinKid:

Issue Number: close https://github.com/pingcap/tidb/issues/61792

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/cardinality/row_count_index.go
```diff
@@ -426,7 +426,16 @@ func equalRowCountOnIndex(sctx planctx.PlanContext, idx *statistics.Index, b []b
...
 	histNDV := float64(idx.Histogram.NDV - int64(idx.TopN.Num()))
-	if histNDV <= 0 {
+	// branch1: histDNV <= 0 means that all NDV's are in TopN, and no histograms.
+	// branch2: histDNA > 0 basically means while there is still a case, c.Histogram.NDV >
+	// c.TopN.Num() a little bit, but the histogram is still empty. In this case, we should use the branch1 and for the diff
+	// in NDV, it's mainly comes from the NDV is conducted and calculated ahead of sampling.
+	if histNDV <= 0 || (idx.IsFullLoad() && idx.Histogram.NotNullCount() == 0) {
+		// branch 1: all NDV's are in TopN, and no histograms
+		// special case of c.Histogram.NDV > c.TopN.Num() a little bit, but the histogram is still empty.
+		if histNDV > 0 && modifyCount == 0 {
```
AilinKid, line 436: https://github.com/pingcap/tidb/pull/59564, introduce the same logic from this pr which fix the case on column est.

hawkingrei (approved):

hawkingrei:
@AilinKid Please remember to update the test result sets.

pkg/planner/cardinality/row_count_index.go
```diff
@@ -426,7 +426,16 @@ func equalRowCountOnIndex(sctx planctx.PlanContext, idx *statistics.Index, b []b
 	}
 	// 3. use uniform distribution assumption for the rest (even when this value is not covered by the range of stats)
 	histNDV := float64(idx.Histogram.NDV - int64(idx.TopN.Num()))
-	if histNDV <= 0 {
+	// branch1: histDNV <= 0 means that all NDV's are in TopN, and no histograms.
+	// branch2: histDNA > 0 basically means while there is still a case, c.Histogram.NDV >
+	// c.TopN.Num() a little bit, but the histogram is still empty. In this case, we should use the branch1 and for the diff
+	// in NDV, it's mainly comes from the NDV is conducted and calculated ahead of sampling.
+	if histNDV <= 0 || (idx.IsFullLoad() && idx.Histogram.NotNullCount() == 0) {
```
qw4990, line 433: What if it's not fully loaded?

qw4990 (approved):

pkg/planner/cardinality/row_count_index.go
AilinKid: no way, when topn exists the index here must have the buckets, which means the index is fullLoad here. @0xPoe @time-and-fate has checked the code path here, the idx.IsFullLoad() is definitely true here, keeping it or removing it is both acceptable, feel free to comment further

terry1purcell:
-1

terry1purcell:
Added a negative 1 comment above for the following reasons:
1) Questions were raised by two reviewers about the need for "isFullLoad" check in this PR - but neither comment was answered/resolved in this PR.
2) If the usage of isFullLoad is correct - it implies that the accuracy of notNullCount() is of concern if the statistics are not fully loaded. However, if this IF check fails (because of !isFullLoad), then the next step in the code (after that IF) will go ahead and use notNullCount() to calculate avgRowEstimate. That line should also have the same accuracy risk.
This PR therefore appears to be incomplete.

PR #52283, opened 2024-04-01
domain: avoit to print too many log if the ddl job of runaway table is not finished

CabinfeverB:
…nished

### What problem does this PR solve?

Issue Number: close #52048

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)
> 
before
> tiup playground nightly --tag upgrade3 --db 3
> ctrl + C
> tiup playground nightly --tag upgrade3 --db 3
image

after
> tiup playground v7.1.1 --tag upgrade --db 3
> ctrl + C
> tiup playground nightly --tag upgrade --db 3 --db.binpath=~/github/tidb/bin/tidb-server
image

- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #61795, opened 2025-06-18
fix(runaway): resolve the dead channel in UpdateNewAndDoneWatch

### LINKED ISSUE #61782: dead chanel in UpdateNewAndDoneWatch
## Bug Report

It started here

https://github.com/pingcap/tidb/blob/e601725d7fd0389107e9848a5f3182e05e0b9d98/pkg/domain/runaway.go#L75-L80

will insert record into channels here

https://github.com/pingcap/tidb/blob/6a12f31171dda2ef20ad4f81f6b2cd95bf13feab/pkg/resourcegroup/runaway/manager.go#L424-L426
The receive loop however is started later than the first call of `UpdateNewAndDoneWatch(RunawayRecordFlushLoop)`

https://github.com/pingcap/tidb/blob/6a12f31171dda2ef20ad4f81f6b2cd95bf13feab/pkg/resourcegroup/runaway/manager.go#L211

If the channel exceeds the maximum(1024), it blocks.

introduced by #52283

non blocking.

`runawayStartLoop` blocks forever

v8.1.0 and later

JmPotato:

Issue Number: close #61782.

Problem Summary:

Revert #52283 and implement a simpler method to prevent error logs.

### Check List

Tests 

- [x] Unit test
- [x] Integration test

Mannually modify `maxWatchRecordChannelSize` to 10 to ensure easier triggering.

image

Before:

image

After:

image

### Release note

```release-note
Resolve the issue causing system table read/write operations related to runaways to hang after restarting TiDB instance.
```

### DIFF
--- a/pkg/resourcegroup/runaway/syncer.go
+++ b/pkg/resourcegroup/runaway/syncer.go
@@ -15,6 +15,7 @@
 package runaway
 
 import (
+	"fmt"
 	"strings"
 	"sync"
 	"time"
@@ -24,10 +25,8 @@ import (
 	"github.com/pingcap/tidb/pkg/util/chunk"
 )
 
-const (
-	// watchSyncInterval is the interval to sync the watch record.
-	watchSyncInterval = time.Second
-)
+// watchSyncInterval is the interval to sync the watch record.
+const watchSyncInterval = time.Second
 
 // Syncer is used to sync the runaway records.
 type syncer struct {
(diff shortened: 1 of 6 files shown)


okJiang (approved):

pkg/resourcegroup/tests/resource_group_test.go
```diff
@@ -468,7 +460,7 @@ func TestResourceGroupRunawayFlood(t *testing.T) {
 	defer func() {
 		require.NoError(t, failpoint.Disable("github.com/pingcap/tidb/pkg/resourcegroup/runaway/FastRunawayGC"))
 	}()
-	store, dom := testkit.CreateMockStoreAndDomain(t)
+	store, _ := testkit.CreateMockStoreAndDomain(t)
```
lance6716:
```suggestion
	store := testkit.CreateMockStore(t)
```

okJiang:
Do we need to add some tests for this issue?

pkg/resourcegroup/tests/resource_group_test.go
```diff
@@ -337,7 +337,7 @@ func TestResourceGroupRunaway(t *testing.T) {
...
```
lance6716:
```suggestion
	store := testkit.CreateMockStore(t)
```

pkg/resourcegroup/runaway/syncer.go
```diff
@@ -45,12 +44,30 @@ func newSyncer(sysSessionPool util.SessionPool) *syncer {
...
 
+func (s *syncer) checkWatchTableExist() bool {
+	return s.checkTableExist("tidb_runaway_watch")
+}
+
+func (s *syncer) checkWatchDoneTableExist() bool {
+	return s.checkTableExist("tidb_runaway_watch_done")
+}
+
+func (s *syncer) checkTableExist(tableName string) bool {
```
lance6716: I'm a bit worried that mixing `err != nil` with other condition will cause problem. For example, somehow `sysSessionPool` has a problem and `ExecRCRestrictedSQL` always fail. And caller will just think related table is not initialized, the functionality will not work and no log indicating the root cause.

How about return `(bool, error)` here? We still log something when there's error

xhebox: Then we will have lots of log again as proposed in your issue? maybe adding 

`if count %= runawayLoopLogErrorIntervalCount; count == 0 {`
or
`if count > initRetryCount {`

lance6716: We can use https://github.com/pingcap/tidb/blob/8e6d77ae274fa73225a1efd485af08509e15e8f6/pkg/util/logutil/log.go#L437-L443

And I want to express when return `(false /* found */, nil /* error */)`there should be no log, we can just silently wait the table to be created. For `(..., not-nil)` we can use above sampled logger to print the error

JmPotato:
> Do we need to add some tests for this issue?

I have added the manual test results in the PR description, please take a look.

xhebox (approved):

PR #61823, opened 2025-06-19
metrics/nextgengrafana: display keyspace separately

tiancaiamao:

Issue Number: ref #58720

Problem Summary:

I'm checking the grafana about the GC related metrics on NextGen
Currently if we load the file metrics/nextgengrafana/tidb_with_keyspace_name.json, data from all keyspaces are mixed!

Since each keyspace works just like a tenant's cluster, we should display for each of them.

Here is the operations to get the new json based on the old one.

1. load the old tidb_with_keyspace_name.json file
2. Dashboard Settings -> Variable -> add the keyspace_name variable so it can be used as $keyspace_name
3. Dashboard Settings -> Save Dashboard
4. Shard dashboard or panel -> Export -> Export for sharing externally -> Save to file
5. open the saved json file, pattern match and replace, add "keyspace_name=$keyspace_name" to every expression
6. Reload the new modified json file, check it works correctly.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

load the new json file, there will be a keyspace_name button:
image

switch this button, it displays data for different keyspace:
image

image
image

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/metrics/nextgengrafana/tidb_with_keyspace_name.json
+++ b/pkg/metrics/nextgengrafana/tidb_with_keyspace_name.json
@@ -2,7 +2,7 @@
   "__inputs": [
     {
       "name": "DS_TEST-CLUSTER",
-      "label": "test-cluster",
+      "label": "Test-Cluster",
       "description": "",
       "type": "datasource",
       "pluginId": "prometheus",
@@ -20,7 +20,7 @@
       "type": "grafana",
       "id": "grafana",
       "name": "Grafana",
-      "version": "7.5.11"
+      "version": "7.5.17"
     },
     {
       "type": "panel",
@@ -51,6 +51,12 @@
       "id": "table",
       "name": "Table",
       "version": ""
+    },
+    {
+      "type": "panel",
+      "id": "timeseries",
+      "name": "Time series",
+      "version": ""
     }
   ],
   "annotations": {
@@ -70,7 +76,7 @@
   "gnetId": null,
   "graphTooltip": 1,
   "id": null,
-  "iteration": 1699393280088,
+  "iteration": 1750295573884,
   "links": [],
   "panels": [
     {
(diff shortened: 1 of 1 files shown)


pkg/metrics/nextgengrafana/tidb_with_keyspace_name.json
```diff
@@ -2,7 +2,7 @@
   "__inputs": [
     {
       "name": "DS_TEST-CLUSTER",
-      "label": "test-cluster",
+      "label": "Test-Cluster",
```
MyonKeminta: Is this change necessary?

pkg/metrics/nextgengrafana/tidb_with_keyspace_name.json
```diff
@@ -126,38 +132,47 @@
...
-          "pluginVersion": "7.5.11",
+          "pluginVersion": "7.5.17",
           "pointradius": 5,
           "points": false,
           "renderer": "flot",
+          "scopedVars": {
+            "keyspace_name": {
+              "selected": true,
+              "text": "admin",
+              "value": "admin"
```
MyonKeminta: What's the value "admin"?

MyonKeminta: By the way I noticed that the `scopedVars` field shows up only in several panels but not all. Is this expected?

pkg/metrics/nextgengrafana/tidb_with_keyspace_name.json
```diff
@@ -24396,6 +24740,6 @@
   },
   "timezone": "browser",
   "title": "Test-Cluster-TiDB-KeyspaceName",
-  "uid": "000000011",
-  "version": 1
+  "uid": "000000012",
```
MyonKeminta: Consider removing this change

MyonKeminta (commented):
By the way did you verify if this still work properly by default in an OP (non-keyspaced) deployment?

yibin87:
Better to adjust other _with_keyspace_name.json files also:)

pkg/metrics/nextgengrafana/tidb_with_keyspace_name.json
tiancaiamao: I delete it and find it does not affect anything.

MyonKeminta:
It was not known to me until reviewing this PR that there are these "KeyspaceName" grafana files, in which the contents are mostly the same as non "KeyspaceName" ones except that it distinguishes keyspaces.
I have a question: how can we make sure the contents of the corresponding "KeyspaceName" dashboards and non-"KeyspaceName" dashboards are consistent, or to say, prevent people to modify one of them forgetting the other?
@yibin87

Issue #61702, opened 2025-06-12
DXF service on nextgen kernel


D3Hunter:
## Feature Request

**Is your feature request related to a problem? Please describe:**

In nextgen, a single cluster will be shared by multiple tenants, each user cluster will have its own keyspace, we want all keyspaces share the same resource pool to run resource intensive DXF tasks such as import-into and add-index:
- To make management of resources for DXF tasks easier, scale in/out, etc. No need dedicated DXF nodes for each keyspace.
- To fully utilize resources and reduce cost
  - Different steps of different DXF tasks might have different requirement for TiDB resources, by sharing same resource pool we can better utilize those resources
  - Dedicated DXF nodes have the drawback that if there not too many load in this keyspace, it's wasting, if we use a scale-out-on-load strategy, it will be less responsive for smaller task, as scale out takes time too.
**Describe the feature you'd like:**

**Describe alternatives you've considered:**

Dedicated DXF for each keyspace

**Teachability, Documentation, Adoption, Migration Strategy:**

As DXF is shared by many user keyspaces, it cannot run on any of the user keyspaces, so we run it in a separate SYSTEM keyspace. This SYSTEM keyspace will be used as the keyspace for all shared services among a nextgen cluster, but currently only DXF service use it.

As user keyspaces depends on SYSTEM keyspace now, we need to make sure SYSTEM keyspace exist before user keyspace start serving any user traffic. So for the deployment of nextgen cluster, we need to:
- Deploy PD/TiKV and other components, wait them to be ready to serve TiDB access.
- Deploy SYSTEM keyspace, wait it fully bootstrapped.
- Deploy user keyspaces, they can be deployed concurrently.
During upgrade, we also need to follow above order, i.e. We need to upgrade the SYSTEM keyspace first, then user keyspace.

In DXF service, we are servicing all user keyspaces using the same DXF meta store, so we need some adaption to task meta abstraction:
1. We need add a keyspace column to the task table to identify the source keyspace of the task.
2. The task_key of task also need to add keyspace name as prefix to unique identify a task.

We don't need to run DXF components in user keyspace, only run them in SYSTEM keyspace.

Import-into and add-index need integrate DXF service, refer https://github.com/pingcap/tidb/issues/60418

#### implementing DXF service and cross keyspace access
- [x] https://github.com/pingcap/tidb/pull/61752
- [x] https://github.com/pingcap/tidb/pull/61755
- [x] https://github.com/pingcap/tidb/pull/61794
[...]

PR #61830, opened 2025-06-19
bootstrap: use versioned upgrade function

D3Hunter:

Issue Number: ref #61702 

Problem Summary:

- use `versionedUpgradeFunction` to force every upgrade have correct bootstrap version, avoid simple mistake. and then we can unifiy the `if ver >= versionXXX` check.
- add UT to enforce upgrade version must in ascending order and the function name must follow a defined pattern
- extract `upgrade` part out to `upgrade.go` from `bootstrap.go`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test

upgrade from v6.1.0/6.5.0/7.1.0/7.5.6/8.1.1/8.5.0 to this PR,  works, and no abnormal logs

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/session/bootstrap.go
+++ b/pkg/session/bootstrap.go
@@ -1002,10 +1002,9 @@ func upgrade(s sessiontypes.Session) {
 	// It is only used in test.
 	upgradeFns := addMockBootstrapVersionForTest(s)
 	for _, verFn := range upgradeFns {
-		if ver < verFn.version || verFn.checkRunnableInside {
+		if ver < verFn.version {
 			verFn.fn(s, ver)
 			logutil.BgLogger().Info("upgrade to bootstrap version.",
-				zap.Bool("check-runnable-inside", verFn.checkRunnableInside),
 				zap.Int64("old-start-version", ver),
 				zap.Int64("in-progress-version", verFn.version),
 				zap.Int64("target-version", currentBootstrapVersion))
--- a/pkg/session/upgrade.go
+++ b/pkg/session/upgrade.go
@@ -465,15 +465,11 @@ const (
 
 // versionedUpgradeFunction is a struct that holds the upgrade function related
 // to a specific bootstrap version.
-// normally we will run the upgrade function when the current bootstrapped version
-// is less than the version in this struct, but there are some exceptions where
-// we might always run the function, such as for version29, and let the function
-// itself to decide whether to run or not and what to run.
+// we will run the upgrade function fn when the current bootstrapped version is
+// less than the version in this struct
 type versionedUpgradeFunction struct {
 	version int64
 	fn      func(sessiontypes.Session, int64)
-	// if true, we will always run fn, and let fn decide whether to run or not.
-	checkRunnableInside bool
 }
 
 // currentBootstrapVersion is defined as a variable, so we can modify its value for testing.
@@ -511,7 +507,7 @@ var (
 		{version: version26, fn: upgradeToVer26},
 		{version: version27, fn: upgradeToVer27},
 		{version: version28, fn: upgradeToVer28},
-		{version: version29, fn: upgradeToVer29, checkRunnableInside: true},
+		{version: version29, fn: upgradeToVer29},
 		{version: version30, fn: upgradeToVer30},
 		{version: version31, fn: upgradeToVer31},
 		{version: version32, fn: upgradeToVer32},


pkg/session/mock_bootstrap.go
```diff
@@ -153,19 +153,18 @@ const (
...
 	TestHook.OnBootstrapBefore(s)
-	if MockUpgradeToVerLatestKind == defaultMockUpgradeToVerLatest {
-		// TODO if we run all tests in this pkg, it will append again, and fail the test.
-		bootstrapVersion = append(bootstrapVersion, mockUpgradeToVerLatest)
-	} else {
-		bootstrapVersion = append(bootstrapVersion, mockSimpleUpgradeToVerLatest)
-	}
 	currentBootstrapVersion = mockLatestVer
+	additionalUpgradeFn := versionedUpgradeFunction{version: mockLatestVer, fn: mockUpgradeToVerLatest}
+	if MockUpgradeToVerLatestKind == MockSimpleUpgradeToVerLatest {
```
D3Hunter, line 164: avoid changing defined global var

pkg/session/mock_bootstrap.go
```diff
@@ -153,19 +153,18 @@ const (
...
+		additionalUpgradeFn = versionedUpgradeFunction{version: mockLatestVer, fn: mockSimpleUpgradeToVerLatest}
+	}
+	return append(upgradeToVerFunctions, additionalUpgradeFn)
```
D3Hunter, line 167: we return the slice, to avoid changing defined global var

pkg/session/bootstrap.go
```diff
@@ -1626,15 +1000,16 @@ func upgrade(s sessiontypes.Session) {
...
-		funcName, err := getFunctionName(upgrade)
-		terror.MustNil(err)
-		upgrade(s, ver)
-		logutil.BgLogger().Info("upgrade in progress, a version has just been completed or be skipped.",
-			zap.Int64("old-start-version", ver),
-			zap.String("in-progress-version", funcName),
-			zap.Int64("latest-version", currentBootstrapVersion))
+	upgradeFns := addMockBootstrapVersionForTest(s)
+	for _, verFn := range upgradeFns {
+		if ver < verFn.version || verFn.alwaysRun {
```
D3Hunter: now we can unify the version check, no need to check in every upgrade function

pkg/session/upgrade.go
```diff
@@ -0,0 +1,2024 @@
...
+// to a specific bootstrap version.
+// normally we will run the upgrade function when the current bootstrapped version
+// is less than the version in this struct, but there are some exceptions where
+// we might always run the function, such as for version29, and let the function
+// itself to decide whether to run or not and what to run.
+type versionedUpgradeFunction struct {
+	version   int64
+	fn        func(sessiontypes.Session, int64)
+	alwaysRun bool
+}
```
D3Hunter: new added struct to hold version and upgrade function at same time

pkg/session/upgrade.go
```diff
@@ -0,0 +1,2024 @@
...
+		{version: version249, fn: upgradeToVer249},
+		{version: version250, fn: upgradeToVer250},
+	}
+)
+
+// upgradeToVer2 updates to version 2.
+func upgradeToVer2(s sessiontypes.Session, ver int64) {
+	// Version 2 add two system variable for DistSQL concurrency controlling.
+	// Insert distsql related system variable.
+	distSQLVars := []string{vardef.TiDBDistSQLScanConcurrency}
```
D3Hunter: notice that we have removed the `if ver >= versionXXX` check

pkg/session/bootstrap.go
lance6716: for the special case ver29, it only runs when upgrades from ver28. So it has _less_ chance to run (other function ver N will also run when upgrades from 1 to N-2). we don't need this alwaysRun member?

lance6716 (commented):
rest lgtm

pkg/session/bootstrap.go
D3Hunter: i thought to add a `shouldRunFn` field, but i am not sure whether we need more complex check logic later, so just use a simple `alwaysRun`, simpler, we can change it later

D3Hunter: https://github.com/pingcap/tidb/pull/61830/commits/7585f41cc5893ea1a0fcd3b5d95b431e95af610d renamed to `checkRunnableInside`

lance6716: I think we don't bother to add this special checking branch for future usage. My points are

1. In current code, only ver29 uses `checkRunnableInside`. But in fact ver29 does not need it. With `checkRunnableInside`, ver29 will skip for [ver1, ver27] and [ver29, ...) inside `upgradeToVer29`. Without `checkRunnableInside`, ver29 will skip for [ver1, ver27] inside `upgradeToVer29` and skip for [ver29, ...) at `if ver < verFn.version` in `upgrade`. Also `upgradeToVer29` still need to have `if ver != version28` check inside itself.
2. In future if `checkRunnableInside` is really needed, I think it means this version should still rerun upgrading logic even if `ver >= verFn.version`. Then line 989 ⬇️ should also be changed to allow rerun of a `checkRunnableInside` version

```
	if ver >= currentBootstrapVersion {
		// It is already bootstrapped/upgraded by a higher version TiDB server.
		return
	}
```

D3Hunter: https://github.com/pingcap/tidb/pull/61830/commits/5fa5114e5bf3e00f3d66107d6da7ecca06295e24

lance6716 (approved):

wjhuang2016 (approved):

PR #61839, opened 2025-06-19
fix(runaway): prevent marking quarantine when settings expired

### LINKED ISSUE #61834: Expired WATCH resource group settings should no longer take effect

Let's assume at T1 we have a resource group `rg1` with `QUERY_LIMIT` like:

```sql
ALTER RESOURCE GROUP rg1 QUERY_LIMIT=(EXEC_ELAPSED='60s', ACTION=KILL, WATCH=SIMILAR DURATION='10m');
```

We change this resource group at T2 to:

```sql
ALTER RESOURCE GROUP rg1 QUERY_LIMIT=(RU = 1000, ACTION=SWITCH_GROUP(default), WATCH=SIMILAR DURATION='10m');
```

Then, a SQL query that starts between T1 and T2 reaches its previous `EXEC_ELAPSED` threshold at T3, triggering a watch rule that uses the expired `QUERY_LIMIT` settings to be added.

Once the `QUERY_LIMIT` settings are changed after T2, the old watch rule should not take effect any more.

After T3, both types of rules will be triggered within a maximum of 10 minutes.

https://github.com/pingcap/tidb/commit/56ba961c7076da1677e1f471356e4651dca3051e

JmPotato:

Issue Number: close #61834

Introduced a check to ensure that quarantine marking only proceeds if the current resource group settings match the existing settings.

### Check List

Tests 

- [x] Unit test
- [x] Integration test

Slow SQL will be killed after 10s:

image

Before:

image

After:

image

### Release note

```release-note
Fix the issue where expired `WATCH` rule still take effect after the `QUERY_LIMIT` setting is modified.
```

### DIFF
--- a/pkg/resourcegroup/runaway/checker.go
+++ b/pkg/resourcegroup/runaway/checker.go
@@ -20,6 +20,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/gogo/protobuf/proto"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/failpoint"
 	rmpb "github.com/pingcap/kvproto/pkg/resource_manager"
@@ -260,6 +261,11 @@ func (r *Checker) CheckRuleKillAction() (string, bool) {
 
 func (r *Checker) markQuarantine(now *time.Time, exceedCause string) {
 	if r.settings == nil || r.settings.Watch == nil {
+		return
+	}
+	// If the latest group settings have been changed, do not mark quarantine.
+	group, err := r.manager.ResourceGroupCtl.GetResourceGroup(r.resourceGroupName)
+	if err != nil || group == nil || !proto.Equal(r.settings, group.RunawaySettings) {
 		return
 	}
 	ttl := time.Duration(r.settings.Watch.LastingDurationMs) * time.Millisecond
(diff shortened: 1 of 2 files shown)


xhebox (approved):

pkg/resourcegroup/runaway/checker.go
```diff
@@ -262,6 +263,11 @@ func (r *Checker) markQuarantine(now *time.Time, exceedCause string) {
 	if r.settings == nil || r.settings.Watch == nil {
 		return
 	}
+	// If the latest group settings have been changed, do not mark quarantine.
+	group, err := r.manager.ResourceGroupCtl.GetResourceGroup(r.resourceGroupName)
+	if err != nil || group == nil || !proto.Equal(r.settings, group.RunawaySettings) {
+		return
```
okJiang, line 269: How about printing some logs here?

JmPotato: I'm concerned that adding a log here could become overwhelming if many SQL queries meet this condition during high QPS. A metrics-based approach might be more suitable. However, the current implementation lacks sufficient observability, which we can address as an enhancement after merging this PR.

PR #61856, opened 2025-06-19
gcworker,metrics: add txn and gc safe point metrics

tiancaiamao:

Issue Number: ref #58720

Problem Summary:

Add metrics for gc_safe_point, txn_safe_point

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

load the new tidb.json

image

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/store/gcworker/gc_worker.go
+++ b/pkg/store/gcworker/gc_worker.go
@@ -680,7 +680,8 @@ func (w *GCWorker) advanceTxnSafePoint(ctx context.Context, target uint64) (newT
 	if err != nil {
 		return 0, errors.Trace(err)
 	}
-	metrics.SafePointGauge.WithLabelValues("txn").Set(float64(result.NewTxnSafePoint))
+	gcSafePointInMillisec := oracle.ExtractPhysical(result.NewTxnSafePoint)
+	metrics.SafePointGauge.WithLabelValues("txn").Set(float64(gcSafePointInMillisec))
 
 	if result.NewTxnSafePoint <= result.OldTxnSafePoint {
 		logutil.BgLogger().Info("txn safe point not advanced, GC will be skipped. this may be caused by GC being blocked, or user enlarged GC life time.",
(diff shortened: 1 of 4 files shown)


pkg/store/gcworker/gc_worker.go
```diff
@@ -680,6 +680,7 @@ func (w *GCWorker) advanceTxnSafePoint(ctx context.Context, target uint64) (newT
 	if err != nil {
 		return 0, errors.Trace(err)
 	}
+	metrics.SafePointGauge.WithLabelValues("txn").Set(float64(result.NewTxnSafePoint))
```
MyonKeminta: The same problem as I said in another PR:

> And another potential issue is that, according to my experiment, grafana losts the precision of timestamps, which are 64bit integers. It's even worse that the Set function of the gauge in golang accepts fload64 values instead of integers, making the precision losing happens even before being uploaded to prometheus.
As the precision of timestamps are quite important, I'm afraid it's not a good idea to upload the timestamp itself. It's impreciseness may cause misleading in problem investigations. Instead, please consider extracting the physical part from the ts, and upload only the physical. People who need the precise values should search in the logs.

MyonKeminta:
By the way, why not observing from PD, where all information, including GC barriers and global GC barriers (when supported), are available?  The grafana dashboard can still be added in TiDB & TiKV, just like how panels in TiKV-Details / GC actually includes information from TiDB's metrics.

tiancaiamao:
> By the way, why not observing from PD, where all information, including GC barriers and global GC barriers (when supported), are available? The grafana dashboard can still be added in TiDB & TiKV, just like how panels in TiKV-Details / GC actually includes information from TiDB's metrics.

Because there is no GC dashboard on the PD grafana.
There are GC dashboard on TiDB and TiKV.

tiancaiamao:
PTAL @MyonKeminta @cfzjywxk

MyonKeminta:
> > By the way, why not observing from PD, where all information, including GC barriers and global GC barriers (when supported), are available? The grafana dashboard can still be added in TiDB & TiKV, just like how panels in TiKV-Details / GC actually includes information from TiDB's metrics.
> 
> Because there is no GC dashboard on the PD grafana. There are GC dashboard on TiDB and TiKV.

I mean recording the metrics from PD's code. It's ok that you show the panel in TiDB or TiKV's page. In fact, curretly about half of the panels in TiKV's GC section are actually recorded by TiDB instead of TiKV.

PR #61868, opened 2025-06-20
Replica `Call` with `CallWithRecover` in the close of hash join v1

### LINKED ISSUE #60926: Child executors can not be closed when panic happens in the `Close` function
## Bug Report

- Execute a sql with hash join v1 executor with disk spill feature turned off
- Let the memory usage of this sql exceeds the memory limit

No goroutine is leaked

Some goroutines may be leaked

v8.5.0

xzhangxian1008:

Issue Number: close #60926

Problem Summary:

We introduce a new function `CallWithRecover` to recover the panic in the close of hash join v1

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that the panic in the close of hash join v1 is not recovered
```


pkg/executor/join/hash_join_v1.go
```diff
@@ -138,7 +138,7 @@ func (e *HashJoinV1Exec) Close() error {
 			channel.Clear(e.ProbeWorkers[i].joinChkResourceCh)
 		}
 		e.ProbeSideTupleFetcher.probeChkResourceCh = nil
-		terror.Call(e.RowContainer.Close)
+		terror.CallWithRecover(e.RowContainer.Close)
```
lance6716: I see the linked issue has some resource leaking problem. Can you point out where the resource cleanup happens?

I'm worried that because panic will break the execution, if the resource cleanup is called after the panic lines, cleanup is still skipped and the issue is not fixed.

BTW did you know the root cause of panic?

xzhangxian1008: > I see the linked issue has some resource leaking problem. Can you point out where the resource cleanup happens?
> 
> I'm worried that because panic will break the execution, if the resource cleanup is called after the panic lines, cleanup is still skipped and the issue is not fixed.
> 
> BTW did you know the root cause of panic?

Panic in `Close` leads to the failure of calling `Close` of child executors, which leads to the resource leaking in child executors.

Root cause: memory tracker may be called in `Close`. If it consumes a positive number and the memory consumption in current exceeds the memoy quota, it may panic.

lance6716: I understand that if line 141 `e.RowContainer.Close` panics, line 162 will be skipped and leak happens. And panic happens because memory tracker by design will use panic to break the control flow.

IMO, the problem is panic is really rare and can happen in a deep level of callee functions, we developers are not aware about it. If this by-design panic is handled too early, it does not jump out of execution quickly enough. If panic is handled too later, some cleanup is skipped by it.

I wonder if there's a more systematic way to handle panic from memory tracker. Like we always call the cleanup functions with `defer` so panic will not skip them.

pkg/executor/test/issuetest/executor_issue_test.go
```diff
@@ -772,3 +772,23 @@ func TestIssue55881(t *testing.T) {
...
+
+	tk.MustExec("use test")
+	tk.MustExec("drop table if exists t1")
+	tk.MustExec("drop table if exists t2")
+	tk.MustExec("create table t1 (col0 int, col1 int);")
+	tk.MustExec("create table t2 (col0 int, col1 int);")
+	tk.MustExec("insert into t1 values (0, 10), (1, 10), (2, 10), (3, 10), (4, 10), (5, 10), (6, 10), (7, 10), (8, 10), (9, 10), (10, 10);")
+	tk.MustExec("insert into t2 values (0, 5), (0, 5), (1, 5), (2, 5), (2, 5), (3, 5), (4, 5), (5, 5), (5, 5), (6, 5), (7, 5), (8, 5), (8, 5), (9, 5), (9, 5), (10, 5);")
+
+	require.NoError(t, failpoint.Enable("github.com/pingcap/tidb/pkg/executor/join/issue60923", "panic"))
```
lance6716: You can try `testfailpoint.Enable`. It will use `t.Cleanup` so no need for extra `failpoint.Disable`

pkg/executor/join/hash_join_v1.go
xzhangxian1008: > I understand that if line 141 `e.RowContainer.Close` panics, line 162 will be skipped and leak happens. And panic happens because memory tracker by design will use panic to break the control flow.
> 
> IMO, the problem is panic is really rare and can happen in a deep level of callee functions, we developers are not aware about it. If this by-design panic is handled too early, it does not jump out of execution quickly enough. If panic is handled too later, some cleanup is skipped by it.
> 
> I wonder if there's a more systematic way to handle panic from memory tracker. Like we always call the cleanup functions with `defer` so panic will not skip them.

I think it's ok to catch the panic early in the `Close` as the execution is finished and there is no need to handle panic caused by memory problems.

pkg/executor/test/issuetest/executor_issue_test.go
xzhangxian1008: > testfailpoint.Enable

done

pkg/executor/join/hash_join_v1.go
lance6716: Sorry I didn't get your point. My concern is maybe in future, in above lines `close` channels or `channel.Clear` will also panic. Your fix only handles one function but not prevent unknown panics. `defer` the critical cleanup function is better. Another fixes are, maybe, in `"pkg/executor/internal/exec".Close`'s recover we double check the cleanup.

And if we focus on fixing this issue, I'm not sure what's the proper place to wrap recover, `func (c *hashRowContainer) Close()` or inner `c.rowContainer.Close()` or elsewhere.

PR #61906, opened 2025-06-21
planner: Simplify plan cache code.gi

### LINKED ISSUE #61876: Simplify the plan cache code path.
## Enhancement

Currently, when one of the plan caches is in use, planner.Optimize calls into the plan cache code, which in turn can call back into planner.Optimize.  There are also various other places in executor and plan builder code that call back into planner.Optimize.   We should simplify this code.

wddevries:

Issue Number: close #61876

Problem Summary:

The existing code for plan cache is convoluted.  Currently, when one of the plan caches is in use, planner.Optimize calls into the plan cache code, which in turn can call back into planner.Optimize. There are also various other places in executor and plan builder code that call back into planner.Optimize.

This patch make two changes:
* The code path for the plan cache is simplified, and planner.Optimize() is not called indirectly by plan cache code,
* Point-get plans can now be cached for non-prepared queries,

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/optimize.go
+++ b/pkg/planner/optimize.go
@@ -141,10 +141,6 @@ func Optimize(ctx context.Context, sctx sessionctx.Context, node *resolve.NodeW,
 	defer tracing.StartRegion(ctx, "planner.Optimize").End()
 	sessVars := sctx.GetSessionVars()
 	pctx := sctx.GetPlanCtx()
-	if sessVars.StmtCtx.EnableOptimizerDebugTrace {
-		debugtrace.EnterContextCommon(pctx)
-		defer debugtrace.LeaveContextCommon(pctx)
-	}
 
 	if !sessVars.InRestrictedSQL && (vardef.RestrictedReadOnly.Load() || vardef.VarTiDBSuperReadOnly.Load()) {
 		allowed, err := allowInReadOnlyMode(pctx, node.Node)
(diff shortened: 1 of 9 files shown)


tests/integrationtest/r/planner/core/plan_cache.result
```diff
@@ -936,7 +936,7 @@ select b from t where b = 1;
 b
 select @@last_plan_from_cache;
 @@last_plan_from_cache
-0
+1
```
hawkingrei, line 939: Your PR is just to simplify the code, why does it change the result? Is there a problem?

wddevries: It also allows point-get plans to be cached for non-prepared queries.  (This was simpler than preserving the existing behavior, and they are already cached for prepared queries.)

terry1purcell (approved):
/lgtm

pkg/planner/optimize.go
```diff
@@ -156,26 +152,78 @@ func Optimize(ctx context.Context, sctx sessionctx.Context, node *resolve.NodeW,
...
 
-	// handle the execute statement
+			// Handle SetVars hints for cached plans.
+			for name, val := range sessVars.StmtCtx.StmtHints.SetVars {
+				oldV, err := sessVars.SetSystemVarWithOldStateAsRet(name, val)
+				if err != nil {
+					sessVars.StmtCtx.AppendWarning(err)
+				}
+				sessVars.StmtCtx.AddSetVarHintRestore(name, oldV)
+			}
```
time-and-fate: Since we disable the plan cache when there are `set_var` hints now, is this part useless currently?

wddevries: That is a good catch.  Originally, I was going to remove the restriction on set_var hints, but I removed that from this pull request.  I'll remove this code and move it to another patch.

PR #61914, opened 2025-06-22
planner: Preference index with most matching columns

terry1purcell:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/find_best_task.go
+++ b/pkg/planner/core/find_best_task.go
@@ -1299,13 +1299,14 @@ func skylinePruning(ds *logicalop.DataSource, prop *property.PhysicalProperty) [
 	ds.SCtx().GetSessionVars().RecordRelevantOptFix(fixcontrol.Fix52869)
 	if preferRange {
 		// Override preferRange with the following limitations to scope
-		preferRange = preferMerge || idxMissingStats || ds.TableStats.HistColl.Pseudo || ds.TableStats.RowCount < 1
+		preferRange = (ds.MostMatchingIndex > 1 && ds.AccessPathMinSelectivity < 1) || preferMerge || idxMissingStats || ds.TableStats.HistColl.Pseudo || ds.TableStats.RowCount < 1
 	}
 	if preferRange && len(candidates) > 1 {
 		// If a candidate path is TiFlash-path or forced-path or MV index or global index, we just keep them. For other
 		// candidate paths, if there exists any range scan path, we remove full scan paths and keep range scan paths.
 		preferredPaths := make([]*candidatePath, 0, len(candidates))
-		var hasRangeScanPath bool
+		mostMatchPath := make([]*candidatePath, 0, len(candidates))
+		var hasRangeScanPath, hasMostMatchPath bool
 		for _, c := range candidates {
 			if c.path.Forced || c.path.StoreType == kv.TiFlash || (c.path.Index != nil && (c.path.Index.Global || c.path.Index.MVIndex)) {
 				preferredPaths = append(preferredPaths, c)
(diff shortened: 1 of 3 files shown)


pkg/planner/core/find_best_task.go
```diff
@@ -1297,15 +1297,17 @@ func skylinePruning(ds *logicalop.DataSource, prop *property.PhysicalProperty) [
 		false,
 	)
 	ds.SCtx().GetSessionVars().RecordRelevantOptFix(fixcontrol.Fix52869)
+	minIndexRows := ds.TableStats.RowCount * ds.AccessPathMinSelectivity
 	if preferRange {
 		// Override preferRange with the following limitations to scope
-		preferRange = preferMerge || idxMissingStats || ds.TableStats.HistColl.Pseudo || ds.TableStats.RowCount < 1
+		preferRange = (ds.MostMatchingIndex > 1 && minIndexRows < 2 && ds.AccessPathMinSelectivity < 0.1) || preferMerge || idxMissingStats || ds.TableStats.HistColl.Pseudo || ds.TableStats.RowCount < 1
```
qw4990, line 1481: Just wonder why we need `minIndexRows < 2 && ds.AccessPathMinSelectivity < 0.1` here, to indicate the current `best-index` based on cardinality estimation might be extremely under-estimated due to any reason like `stats-missing` or `incorrect-estimation-assumption`?

pkg/planner/core/find_best_task.go
```diff
@@ -1318,9 +1320,17 @@ func skylinePruning(ds *logicalop.DataSource, prop *property.PhysicalProperty) [
 					preferredPaths = append(preferredPaths, c)
 					hasRangeScanPath = true
 				}
+				// Preference the most matching index path for very selective index plans
+				if c.path.EqOrInCondCount > 1 && c.path.EqOrInCondCount == ds.MostMatchingIndex &&
+					c.path.CountAfterIndex > 0 && minIndexRows < 2 && c.path.CountAfterIndex <= (minIndexRows*1.2) {
```
qw4990: `c.path.CountAfterIndex <= (minIndexRows*1.2)` seems too strict for me, if some index stats is missing, its wrong estimation result could be extremely underestimated, for example, from hundreds to around 0?
Could we be more aggressive here to set it to 50 or even 100, then the upperbound is `minIndexRows*100 < 200` since `minIndexRows < 2`?

pkg/planner/core/find_best_task.go
terry1purcell: The target of this prototype is to preference the index with the most matching predicates if that index is also one of the most selective. This is because otherwise, the costing will potentially bias towards a smaller index if it has similar filtering, since rowSize is one of the inputs to the index cost.

minIndexRows < 2 is simply to limit the scope to very selective indexes - where we may have estimated below 1 row or zero rows, and set a low bound == 1, or some other selective value such as 1.01. And ds.AccessPathMinSelectivity < 0.1 so that we aren't targeting very small tables - such that this "< 2" isn't because it is because it is a very small table.

There may be many reasons why we got to 1 row or less - out of range estimation, all values in TopN, or assumption of independence (including exponential backoff).

pkg/planner/core/find_best_task.go
terry1purcell: What this check "should" do is - once we have determined that at least one index exists such that "minIndexRows < 2" - then for this index to be added to the "mostMatchPath", it's CountAfterIndex must NOT be more than 20% than minIndexRows. That is - if the "best" index qualifies 1 row, then any index that has the "most matching predicates" must qualify <= 1.2 rows - otherwise it's too risky to use the "best matching index" to override the index that wins on cost.

terry1purcell: In the latest commit - I updated the "1.2" to be "1.5". This may partially address your concern. I understand you still may have a concern on the minIndexRows < 2. We need to discuss this to come to an agreement on what we think is the correct compromise.

PR #61952, opened 2025-06-24
fix: using slice Compact to simple the code

yihong0618:
This patch  follow @hawkingrei comment 

> TiDB likely has a lot of code that could be replaced with functions from the official Go slices package. Is there anyone strong in AI who can do a comprehensive sweep of this area?

to use some code agent to try to this part:

steps:

1. using agent to find the code maybe(in the code base I choosed `utils` like code
2. make sure the code is right, if the function has no test need to add test first(which can use LLM for help)
3. add test, change the code, and make sure the benchmark did not decrease so better to benchmark both of them
4. make test

if the steps works we can finish others as the comments.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/store/mockstore/unistore/tikv/util_test.go
```diff
@@ -0,0 +1,432 @@
+// Copyright 2019-present PingCAP, Inc.
```
hawkingrei:
```suggestion
// Copyright 2025 PingCAP, Inc.
```

hawkingrei:
```
pkg/store/mockstore/unistore/tikv/util_test.go:371:7: should omit nil check; len() for nil slices is defined as zero (S1009)
```

Please fix the linter.

and run ```make bazel_prepare``` to update bazel config.

pkg/store/mockstore/unistore/tikv/util_test.go
```diff
@@ -0,0 +1,432 @@
...
+				{Key: []byte("key2")},
+				{Key: []byte("key3")},
+			},
+			expected: func() []uint64 {
+				vals := []uint64{
+					farm.Fingerprint64([]byte("key1")),
+					farm.Fingerprint64([]byte("key2")),
+					farm.Fingerprint64([]byte("key3")),
+				}
+				return sortAndDedupHashVals(vals)
```
lance6716: because these tests calls `sortAndDedupHashVals`, it doesn't check the logic inside `sortAndDedupHashVals`. I think you can delete them. Only `TestSortAndDedupHashVals` is enough

pkg/store/mockstore/unistore/tikv/util_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+// Benchmark tests for performance-critical functions
+func BenchmarkSortAndDedupHashVals(b *testing.B) {
+	sizes := []int{10, 100, 1000, 10000}
+
+	for _, size := range sizes {
+		b.Run(fmt.Sprintf("size_%d", size), func(b *testing.B) {
+			testData := make([]uint64, size)
+			for i := 0; i < size; i++ {
+				testData[i] = uint64(i % (size / 2))
```
lance6716: I'm not sure if we need this benchmark. The input data pattern is a bit simple, and maybe it's better to stop the timer before cloning the input but the `testing.B.N` estimation might go wrong too much. WDYT @hawkingrei 

My suggestion is delete it. It's not a performance-critical function. mockstore/unistore is not used in production.

hawkingrei: I argee with you. /cc @yihong0618

yihong0618: addressed

Issue #59827, opened 2025-02-28
Wrong Ranges for partition pruning


xhebox:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```mysql
DROP TABLE IF EXISTS t;
CREATE TABLE `t` (
  `a` varchar(150) NOT NULL,
  `b` varchar(100) NOT NULL,
  `c` int NOT NULL DEFAULT '0',
  PRIMARY KEY (`a`,`b`) /*T![clustered_index] CLUSTERED */
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci
PARTITION BY LIST COLUMNS(`b`)
(PARTITION `p0` VALUES IN ('0'),
 PARTITION `p1` VALUES IN ('1'),
 PARTITION `p2` VALUES IN ('2'));

explain analyze select * from t where a = 'b' and b = '2';
explain analyze select * from t where a = 'b' and (b = '2');
```

https://github.com/pingcap/tidb/blob/75695928b955564835e1284dcdc2c7785cfe1d67/pkg/executor/point_get.go#L58-L65

Here a safe guard is added by #49161, trying to prune and see if it succeeded.

But when there is a pair of `()`, it will not go through the fast path, falling back to findbesttask.

In the slow path, before partition pruning processor rule, the juding paths is already generated by https://github.com/pingcap/tidb/blob/75695928b955564835e1284dcdc2c7785cfe1d67/pkg/planner/core/stats.go#L385-L417 ，which is using `DetachCondAndBuildRangeForIndex`.

While it should be processed by the pruning processor, and use `DetachCondAndBuildRangeForPartition` like https://github.com/pingcap/tidb/blob/75695928b955564835e1284dcdc2c7785cfe1d67/pkg/planner/core/rule_partition_processor.go#L737-L751.

This eventually leads to the failure of that safety guard, returning `table dual` instead.

### 2. What did you expect to see? (Required)
Both pruning results the same

### 3. What did you see instead (Required)
Different pruning results (with bad result as well, no rows would be found)
```mysql
tidb> explain analyze select * from t where a = 'b' and b = '2';
+-------------+---------+---------+------+------------------------------------------------------+----------------------------------------------------------------------------------------------+---------------+--------+------+
| id          | estRows | actRows | task | access object                                        | execution info                                                                               | operator info | memory | disk |
[...]

PR #59918, opened 2025-03-05
Planner: Run full partition pruning if convertToPointGet converted keys to SortKey

mjonss:
### What problem does this PR solve?

Issue Number: close #59827

Problem Summary:
When PointGetPlan is created by `convertToPointGet`, the `IndexValues` comes from the Range optimizer, and if any of the LIST/RANGE COLUMNS has a non-binary collation, they conditions values are converted to SortKey, which does not work with the LIST/RANGE COLUMNS partitioning pruning code, since they will unconditionally convert it again to SortKey, which may fail finding a matching partition.

NOTE: This only affects PointGet, since BatchPointGet is not supported yet.

### What changed and how does it work?
If the IndexValues are already converted, run the normal partition pruning code on the conditions, to find the matching partition.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
Partition Pruning could fail and choose 'TableDual', i.e. no matching partition, in case the optimizer could convert the query to PointGet (if not using the FastPlan) and any LIST COLUMNS or RANGE COLUMNS had non-binary collation.
```

PR #62002, opened 2025-06-25
planner: fix `PointGetPlan.PrunePartitions` function works with non-binary collate

### LINKED ISSUE #61965: Report `too many partitions matching for PointGetPlan` for range partition
## Bug Report

```
CREATE TABLE `t5f0caf27` (
  `col_95` char(181) COLLATE gbk_bin NOT NULL DEFAULT 'SaMKHTyg+nlID-X3Y',
  PRIMARY KEY (`col_95`) /*T![clustered_index] CLUSTERED */
) ENGINE=InnoDB DEFAULT CHARSET=gbk COLLATE=gbk_bin
PARTITION BY RANGE COLUMNS(`col_95`)
(PARTITION `p0` VALUES LESS THAN ('6)nvX^uj0UGxqX'),
 PARTITION `p1` VALUES LESS THAN ('BHSluf6'),
 PARTITION `p2` VALUES LESS THAN (MAXVALUE));

select t5f0caf27.col_95 as r0 from t5f0caf27 where t5f0caf27.col_95 between 'Dyw=*7nigCMh' and 'Im0*7sZ' or t5f0caf27.col_95 in ( '58y-j)84-&Y*' ,'WNe(rS5uwmvIvFnHw' ,'j9FsMawX5uBro%$p' ,'C(#EQm@J' ) group by t5f0caf27.col_95  having t5f0caf27.col_95 between '%^2' and '38ABfC-' or t5f0caf27.col_95 between 'eKCAE$d2x_hxscj' and 'zcw35^ATEEp1md=L';
```

PointGet plan with access partition p2.

report error

caused by #59918, affects v8.5.2 and later.

Defined2014:

Issue Number: close #61965, close https://github.com/pingcap/tidb/issues/59827

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/point_get_plan.go
+++ b/pkg/planner/core/point_get_plan.go
@@ -392,9 +392,7 @@ func (p *PointGetPlan) PrunePartitions(sctx sessionctx.Context) (bool, error) {
 		// and the fast path does not convert the values to `sortKey`.
 		for _, col := range p.IdxCols {
 			// TODO: We could check whether `col` belongs to the partition columns to avoid unnecessary ranger building.
-			// However, we currently don't have a safe and reliable way to do that.
-			// `col` is of type `expression.Column`, while the IDs from `pt.GetPartitionColumnIDs()` are derived from `t.Cols()`
-			// which type is `table.Column`. It's not always correctly to compare them.
+			// https://github.com/pingcap/tidb/pull/62002#discussion_r2171420731
 			if !collate.IsBinCollation(col.GetType(evalCtx).GetCollate()) {
 				// If a non-binary collation is used, the values in `p.IndexValues` are sort keys and cannot be used for partition pruning.
 				r, err := ranger.DetachCondAndBuildRangeForPartition(sctx.GetRangerCtx(), p.AccessConditions, p.IdxCols, p.IdxColLens, sctx.GetSessionVars().RangeMaxSize)
--- a/pkg/planner/core/point_get_plan.go
+++ b/pkg/planner/core/point_get_plan.go
@@ -387,7 +387,6 @@ func (p *PointGetPlan) PrunePartitions(sctx sessionctx.Context) (bool, error) {
 	row := make([]types.Datum, len(p.TblInfo.Columns))
 	if p.HandleConstant == nil && len(p.IndexValues) > 0 {
 		indexValues := p.IndexValues
-
 		evalCtx := sctx.GetExprCtx().GetEvalCtx()
 		for _, col := range p.IdxCols {
 			if !collate.IsBinCollation(col.GetType(evalCtx).GetCollate()) {
--- a/pkg/planner/core/point_get_plan.go
+++ b/pkg/planner/core/point_get_plan.go
@@ -386,30 +386,30 @@ func (p *PointGetPlan) PrunePartitions(sctx sessionctx.Context) (bool, error) {
 	}
 	row := make([]types.Datum, len(p.TblInfo.Columns))
 	if p.HandleConstant == nil && len(p.IndexValues) > 0 {
-		hasNonBinaryCollate := false
 		indexValues := p.IndexValues
 
 		evalCtx := sctx.GetExprCtx().GetEvalCtx()
 		for _, col := range p.IdxCols {
-			switch col.GetType(evalCtx).GetCollate() {
-			case "utf8mb4_bin", "binary", "ascii_bin", "latin1_bin", "utf8_bin", "utf8mb4_0900_bin":
-			default:
-				hasNonBinaryCollate = true
+			binaryCollations := map[string]struct{}{
+				"utf8mb4_bin":      {},
+				"binary":           {},
+				"ascii_bin":        {},
+				"latin1_bin":       {},
+				"utf8_bin":         {},
+				"utf8mb4_0900_bin": {},
 			}
-			if hasNonBinaryCollate {
+			if _, exists := binaryCollations[col.GetType(evalCtx).GetCollate()]; !exists {
+				// If a non-binary collation is used, the values in `p.IndexValues` are sort keys and cannot be used for partition pruning.
+				r, err := ranger.DetachCondAndBuildRangeForPartition(sctx.GetRangerCtx(), p.AccessConditions, p.IdxCols, p.IdxColLens, sctx.GetSessionVars().RangeMaxSize)
+				if err != nil {
+					return false, err
+				}
+				if len(r.Ranges) != 1 || !r.Ranges[0].IsPoint(sctx.GetRangerCtx()) {
+					return false, errors.Errorf("internal error, build ranger for PointGet failed")
+				}
+				indexValues = r.Ranges[0].LowVal
 				break
 			}
-		}
-		// If a non-binary collation is used, the values in `p.IndexValues` are sort keys and cannot be used for partition pruning.
-		if hasNonBinaryCollate {
-			r, err := ranger.DetachCondAndBuildRangeForPartition(sctx.GetRangerCtx(), p.AccessConditions, p.IdxCols, p.IdxColLens, sctx.GetSessionVars().RangeMaxSize)
-			if err != nil {
-				return false, err
-			}
-			if len(r.Ranges) != 1 || !r.Ranges[0].IsPoint(sctx.GetRangerCtx()) {
-				return false, errors.Errorf("internal error, build ranger for PointGet failed")
-			}
-			indexValues = r.Ranges[0].LowVal
 		}
 		for i := range p.IndexInfo.Columns {
 			// TODO: Skip copying non-partitioning columns?
--- a/pkg/planner/core/point_get_plan.go
+++ b/pkg/planner/core/point_get_plan.go
@@ -388,7 +388,13 @@ func (p *PointGetPlan) PrunePartitions(sctx sessionctx.Context) (bool, error) {
 	if p.HandleConstant == nil && len(p.IndexValues) > 0 {
 		indexValues := p.IndexValues
 		evalCtx := sctx.GetExprCtx().GetEvalCtx()
+		pColIDs := pt.GetPartitionColumnIDs()
+		// If the plan is created via the fast path, `IdxCols` will be nil here,
+		// and the fast path does not convert the values to `sortKey`.
 		for _, col := range p.IdxCols {
+			if !slices.Contains(pColIDs, col.ID) {
+				continue
+			}
 			if !collate.IsBinCollation(col.GetType(evalCtx).GetCollate()) {
 				// If a non-binary collation is used, the values in `p.IndexValues` are sort keys and cannot be used for partition pruning.
 				r, err := ranger.DetachCondAndBuildRangeForPartition(sctx.GetRangerCtx(), p.AccessConditions, p.IdxCols, p.IdxColLens, sctx.GetSessionVars().RangeMaxSize)
--- a/pkg/planner/core/point_get_plan.go
+++ b/pkg/planner/core/point_get_plan.go
@@ -386,8 +386,11 @@ func (p *PointGetPlan) PrunePartitions(sctx sessionctx.Context) (bool, error) {
 	row := make([]types.Datum, len(p.TblInfo.Columns))
 	if p.HandleConstant == nil && len(p.IndexValues) > 0 {
 		hasNonBinaryCollate := false
+		indexValues := p.IndexValues
+
+		evalCtx := sctx.GetExprCtx().GetEvalCtx()
 		for _, col := range p.IdxCols {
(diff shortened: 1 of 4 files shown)


pkg/planner/core/point_get_plan.go
```diff
@@ -341,67 +342,6 @@ func (p *PointGetPlan) LoadTableStats(ctx sessionctx.Context) {
...
 }
 
-// needsPartitionPruning checks if IndexValues can be used by GetPartitionIdxByRow() or if they have already been
-// converted to SortKey and would need GetPartitionIdxByRow() to be refactored to work, since it will unconditionally
-// convert it again.
-// Returns:
-// Matching partition
-// if done Partition pruning (else not needed, can use GetPartitionIdxByRow() instead)
-// error
-// TODO: Also supporting BatchPointGet? Problem is that partition ID must be mapped to handle/IndexValue.
```
hawkingrei, line 351: Has this TODO been completed?

pkg/planner/core/point_get_plan.go
```diff
@@ -446,32 +385,32 @@ func (p *PointGetPlan) PrunePartitions(sctx sessionctx.Context) (bool, error) {
...
 	if p.HandleConstant == nil && len(p.IndexValues) > 0 {
-		partColsNames := pt.Meta().Partition.Columns
-		if len(partColsNames) > 0 {
-			partIdx, done, err := needsPartitionPruning(sctx, p.TblInfo, pt, p.dbName, p.IndexInfo, p.IdxCols, p.IndexValues, p.AccessConditions, p.PartitionNames)
-			if table.ErrNoPartitionForGivenValue.Equal(err) {
-				err = nil
-				partIdx = nil
+		hasNonBinaryCollate := false
+		for _, col := range p.IdxCols {
+			switch col.GetType(sctx.GetExprCtx().GetEvalCtx()).GetCollate() {
```
hawkingrei: We can move this ```sctx.GetExprCtx().GetEvalCtx()```  outside the for-loop.

pkg/planner/core/point_get_plan.go
Defined2014: No, this function is useless and the result is different with the result from planner part. So I remove it.

pkg/planner/core/point_get_plan.go
Defined2014: Done

hawkingrei (approved):

pkg/planner/core/point_get_plan.go
```diff
@@ -419,7 +359,6 @@ func (p *PointGetPlan) PrunePartitions(sctx sessionctx.Context) (bool, error) {
 			if def.Name.L == p.PartitionNames[0].L {
 				idx := i
 				p.PartitionIdx = &idx
-				break
```
AilinKid: why should we remove this break, cuz i think only one partition can be matched?

pkg/planner/core/point_get_plan.go
```diff
@@ -446,32 +385,34 @@ func (p *PointGetPlan) PrunePartitions(sctx sessionctx.Context) (bool, error) {
...
-			if table.ErrNoPartitionForGivenValue.Equal(err) {
-				err = nil
-				partIdx = nil
+		hasNonBinaryCollate := false
+		indexValues := p.IndexValues
+
+		evalCtx := sctx.GetExprCtx().GetEvalCtx()
+		for _, col := range p.IdxCols {
+			switch col.GetType(evalCtx).GetCollate() {
+			case "utf8mb4_bin", "binary", "ascii_bin", "latin1_bin", "utf8_bin", "utf8mb4_0900_bin":
```
AilinKid: we can use the defined string: CollationUTF8MB4 and so on

pkg/planner/core/point_get_plan.go
```diff
@@ -446,32 +385,34 @@ func (p *PointGetPlan) PrunePartitions(sctx sessionctx.Context) (bool, error) {
...
+			default:
+				hasNonBinaryCollate = true
 			}
+			if hasNonBinaryCollate {
+				break
+			}
+		}
+		// If a non-binary collation is used, the values in `p.IndexValues` are sort keys and cannot be used for partition pruning.
+		if hasNonBinaryCollate {
+			r, err := ranger.DetachCondAndBuildRange(sctx.GetRangerCtx(), p.AccessConditions, p.IdxCols, p.IdxColLens, sctx.GetSessionVars().RangeMaxSize, false, true)
```
AilinKid: i saw a similar function named: DetachCondAndBuildRangeForPartition, which the last bool is fixed as false, i think since this func is also amended for partition usage, maybe we can use the same portal and adjust the param?

pkg/planner/core/point_get_plan.go
```diff
@@ -446,32 +385,34 @@ func (p *PointGetPlan) PrunePartitions(sctx sessionctx.Context) (bool, error) {
...
-					return true, nil
-				}
-				return false, errors.Errorf("too many partitions matching for PointGetPlan")
+			if len(r.Ranges) != 1 || !r.Ranges[0].IsPoint(sctx.GetRangerCtx()) {
+				return false, errors.Errorf("internal error, build ranger for PointGet failed")
 			}
+			indexValues = r.Ranges[0].LowVal
 		}
 		for i := range p.IndexInfo.Columns {
 			// TODO: Skip copying non-partitioning columns?
```
AilinKid, line 410: can this todo be solved by the way

pkg/planner/core/point_get_plan.go
AilinKid: with that we don't need to export DetachCondAndBuildRange

AilinKid (commented):
rest LGTM

pkg/planner/core/point_get_plan.go
Defined2014: No

pkg/planner/core/point_get_plan.go
mjonss: Are there not `isBinaryCollation()` or similar function that can be used?

Defined2014: The sortKey for `gbk_bin` and `gb18030_bin` is different with original string

Defined2014: It's the default collation for utf8mb4, maybe it will changed in some day.

pkg/planner/core/point_get_plan.go
Defined2014: It'a a mistake.

pkg/planner/core/point_get_plan.go
```diff
@@ -446,32 +386,34 @@ func (p *PointGetPlan) PrunePartitions(sctx sessionctx.Context) (bool, error) {
...
-				return false, err
-			}
-			if done {
-				if len(partIdx) == 1 {
-					p.PartitionIdx = &partIdx[0]
-					return false, nil
+		indexValues := p.IndexValues
+
+		evalCtx := sctx.GetExprCtx().GetEvalCtx()
+		binaryCollations := map[string]struct{}{
```
mjonss: Would it be possible to use `CompatibleCollate("utf8mb4_bin", col.GetType(evalCtx).GetCollate())` or something else, to avoid having yet another place to maintain the list of binary collations?

mjonss: Ah, there is a `IsBinCollation()`, please use that :)

pkg/planner/core/point_get_plan.go
mjonss: Oh, so binary collations can still use a different sortKey? Then we should have a collation utility function saying if a sort key is needed or not?

pkg/planner/core/point_get_plan.go
Defined2014: Done, PTAL

pkg/planner/core/point_get_plan.go
```diff
@@ -446,32 +386,25 @@ func (p *PointGetPlan) PrunePartitions(sctx sessionctx.Context) (bool, error) {
...
+				// If a non-binary collation is used, the values in `p.IndexValues` are sort keys and cannot be used for partition pruning.
+				r, err := ranger.DetachCondAndBuildRangeForPartition(sctx.GetRangerCtx(), p.AccessConditions, p.IdxCols, p.IdxColLens, sctx.GetSessionVars().RangeMaxSize)
+				if err != nil {
+					return false, err
 				}
-				if len(partIdx) == 0 {
-					idx := -1
-					p.PartitionIdx = &idx
-					return true, nil
+				if len(r.Ranges) != 1 || !r.Ranges[0].IsPoint(sctx.GetRangerCtx()) {
```
mjonss, line 402: Please add a test where `len(r.Ranges) == 0`. Like RANGE partition without MAXVALUE and a condition for value greater than last partition.
I would assume it should `return true, nil` in such case and use table DUAL.

pkg/planner/core/point_get_plan.go
```diff
@@ -446,32 +386,25 @@ func (p *PointGetPlan) PrunePartitions(sctx sessionctx.Context) (bool, error) {
...
-			if done {
-				if len(partIdx) == 1 {
-					p.PartitionIdx = &partIdx[0]
-					return false, nil
+		indexValues := p.IndexValues
+		evalCtx := sctx.GetExprCtx().GetEvalCtx()
+		for _, col := range p.IdxCols {
+			if !collate.IsBinCollation(col.GetType(evalCtx).GetCollate()) {
+				// If a non-binary collation is used, the values in `p.IndexValues` are sort keys and cannot be used for partition pruning.
+				r, err := ranger.DetachCondAndBuildRangeForPartition(sctx.GetRangerCtx(), p.AccessConditions, p.IdxCols, p.IdxColLens, sctx.GetSessionVars().RangeMaxSize)
```
mjonss, line 398: Why are we looping on `p.IdxCols` and doing this for the full conditions, and all columns (`p.IdxCols`)?
Should we also check if the column is part of the partitioning expression and then check if binary collation?

tests/integrationtest/r/planner/core/casetest/partition/partition_pruner.result
```diff
@@ -928,3 +928,19 @@ Point_Get	1.00	root	table:t, partition:p4, index:idx_3(col_4)
 select * from t where col_4 in ( 'u^D92@_4' ,null );
 col_4
 u^D92@_4
+drop table if exists t;
+CREATE TABLE `t` (
+`col_95` char(181) COLLATE gbk_bin NOT NULL DEFAULT 'SaMKHTyg+nlID-X3Y',
+PRIMARY KEY (`col_95`) /*T![clustered_index] CLUSTERED */
+) ENGINE=InnoDB DEFAULT CHARSET=gbk COLLATE=gbk_bin
+PARTITION BY RANGE COLUMNS(`col_95`)
```
mjonss, line 936: Do we need to test other partitioning schemes as well? Like RANGE COLUMNS, LIST [COLUMNS] and KEY?

pkg/planner/core/point_get_plan.go
Defined2014: This function isn't related to partition tables — it only processes the `AccessConditions` and returns the ranger. So even it's not meet any exists partition, the `len(r.Ranges)` also should equals to 1.

The pruning stage is handled separately [here](https://github.com/pingcap/tidb/pull/62002/files#diff-989673423ebb0def5c5df96de2bdf7095411d398cf7ee5a04f89d3921f49e168L485).

Defined2014: The root cause of these problems is that the planner uses a different function from this one, and the results of the two are not always consistent. To ensure correctness, we should use the same function here.

tests/integrationtest/r/planner/core/casetest/partition/partition_pruner.result
Defined2014: I don't think this is related to partition types. And also I think the test cases have already been added in [PR #59918](https://github.com/pingcap/tidb/pull/59918).

pkg/planner/core/point_get_plan.go
Defined2014: Done

Defined2014: I added a comments here, I think it's not easy to do this right now. And this part will not affect the performance very much.

> However, we currently don't have a safe and reliable way to do that. `col` is of type `expression.Column`, while the IDs from `pt.GetPartitionColumnIDs()` are derived from `t.Cols()` which type is `table.Column`. It's not always correctly to compare them.

Defined2014:
/cherry-pick-release-8.5

PR #62007, opened 2025-06-25
domain: abstract cross keyspace manager and sql server interface

D3Hunter:

Issue Number: ref #61702

Problem Summary:
in DXF service:
- we need to access SYSTEM keyspace tables related to DXF inside user keyspace.
- we need to access user keyspace `mysql.tidb_import_jobs`, so we need to create user keyspace store and load its info schema inside SYSTEM keyspace, 

- abstract cross keyspace manager, to manager everything related to cross keyspace access. 
- abstract sql server API and bind to session, to avoid session depends on domain
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/domain/domain.go
```diff
@@ -878,32 +857,52 @@ func (do *Domain) Start(startMode ddl.StartMode) error {
...
+			if targetKS == keyspace.System {
+				store = kvstore.GetSystemStorage()
+			} else {
+				store, err = kvstore.InitStorage(targetKS)
+				if err != nil {
+					return nil, err
+				}
+			}
+			return store, nil
+		},
```
tangenta: Why not put this closure inside `GetOrCreate` since it is the only usage?

pkg/keyspace/keyspace.go
```diff
@@ -91,7 +91,8 @@ func WrapZapcoreWithKeyspace() zap.Option {
 	})
 }
 
-// IsRunningOnUser checks if keyspace of current instance is a user keyspace.
-func IsRunningOnUser() bool {
-	return config.GetGlobalKeyspaceName() != System
+// IsRunningOnUserKS return true if we are on nextgen, and keyspace of current
+// instance is a user keyspace.
+func IsRunningOnUserKS() bool {
```
tangenta: `IsFromUser()` is better? The package name is already "keyspace".

tangenta (approved):

pkg/domain/crossks/cross_ks.go
```diff
@@ -0,0 +1,186 @@
...
+	return sessMgr, ok
+}
+
+// GetOrCreate gets or creates a session manager for the specified keyspace.
+func (m *Manager) GetOrCreate(
+	ks string,
+	ksSessFactoryGetter func(string) pools.Factory,
+) (_ *SessionManager, err error) {
+	if kerneltype.IsClassic() || config.GetGlobalKeyspaceName() == ks {
+		return nil, errors.New("cross keyspace session manager is not available in classic kernel or current keyspace")
```
wjhuang2016, line 75: How about using Intest.assert to check it.

pkg/domain/crossks/cross_ks.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	return kvstore.InitStorage(targetKS)
+}
+
+// SessionManager manages sessions for a specific keyspace.
+type SessionManager struct {
+	store     kv.Storage
+	infoCache *infoschema.InfoCache
+	isLoader  *issyncer.Loader
+	sessPool  util.DestroyableSessionPool
```
wjhuang2016, line 174: DestroyableSessionPool is deprecated, use `advancedSysSessionPool` instead

pkg/domain/crossks/cross_ks.go
D3Hunter: i can add it too, but this struct should only be called for cross keyspace access, i will keep this check in case some code path is not covered

pkg/domain/crossks/cross_ks.go
D3Hunter: advancedSysSessionPool automatically register session to session manager, i.e. tidb server, but for cross keyspace session, we should use another session manager for the target keyspace, which we don't have it now

pkg/domain/crossks/cross_ks.go
D3Hunter: https://github.com/pingcap/tidb/pull/62007/commits/e762ff0b7e046bfe6ad90de07033b2e08e6973c2 add a comment to explain why not use `intest.Assert`

PR #62015, opened 2025-06-26
ddl: do some code cleanup in ddl package

joechenrh:

Issue Number: ref https://github.com/pingcap/tidb/issues/62559

Problem Summary:

This PR does some code refactor in pkg/ddl module:

- Remove function `LoadDDLReorgVars` and related unit test. It was used to change reorg worker number, and now we have a newer approach to do this.
- Set `DiskFullOpt` before get session from pool and clear it before put it back.
- Rename `job.SessionVars` to `job.SystemVars`, so these usage won't be checked by linter.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/meta/model/job.go
+++ b/pkg/meta/model/job.go
@@ -390,7 +390,8 @@ type Job struct {
 	// SQLMode for executing DDL query.
 	SQLMode mysql.SQLMode `json:"sql_mode"`
 
-	// SessionVars store system variables
+	// SessionVars store system variables used in the DDL execution.
+	// To keep the backward compatibility, we still name it SessionVars.
 	SessionVars map[string]string `json:"session_vars,omitempty"`
 }
 


pkg/ddl/delete_range.go
```diff
@@ -549,11 +549,7 @@ func (sdr *sessionDelRangeExecWrapper) AppendParamsList(jobID, elemID int64, sta
 }
 
 func (sdr *sessionDelRangeExecWrapper) ConsumeDeleteRange(ctx context.Context, sql string) error {
-	// set session disk full opt
-	sdr.sctx.GetSessionVars().SetDiskFullOpt(kvrpcpb.DiskFullOpt_AllowedOnAlmostFull)
```
joechenrh, line 553: This session is get from ddl session pool. We will set SetDiskFullOpt before get and clear after put back.

pkg/ddl/tests/fail/fail_db_test.go
```diff
@@ -347,79 +343,6 @@ func TestRunDDLJobPanicEnableFastCreateTable(t *testing.T) {
 	require.EqualError(t, err, "[ddl:8214]Cancelled DDL job")
 }
 
-func testAddIndexWorkerNum(t *testing.T, s *failedSuite, test func(*testkit.TestKit)) {
```
joechenrh, line 350: Seems like previously we support changing worker number by directly set `tidb_ddl_reorg_worker_cnt` in global scope. But it's achieved by `ALTER TABLE` now. So this test is no longer needed (Actually, it's skipped for master branch, and it can't pass without dxf).

pkg/executor/test/ddl/ddl_test.go
```diff
@@ -853,33 +852,6 @@ func TestSetDDLReorgBatchSize(t *testing.T) {
 	tk.MustQuery("select @@global.tidb_ddl_reorg_batch_size").Check(testkit.Rows("1000"))
 }
 
-func TestSetDDLErrorCountLimit(t *testing.T) {
```
joechenrh, line 856: Used to test `LoadDDLVars`, which has been removed in this PR.

pkg/ddl/util/util.go
```diff
@@ -186,47 +184,6 @@ func UpdateDeleteRange(sctx sessionctx.Context, dr DelRangeTask, newStartKey, ol
...
 
-// LoadDDLReorgVars loads ddl reorg variable from mysql.global_variables.
-func LoadDDLReorgVars(ctx context.Context, sctx sessionctx.Context) error {
-	// close issue #21391
-	// variable.TiDBRowFormatVersion is used to encode the new row for column type change.
-	return loadGlobalVars(ctx, sctx, []string{vardef.TiDBDDLReorgWorkerCount, vardef.TiDBDDLReorgBatchSize, vardef.TiDBRowFormatVersion})
-}
-
-// LoadDDLVars loads ddl variable from mysql.global_variables.
-func LoadDDLVars(ctx sessionctx.Context) error {
```
joechenrh, line 197: Take `pkg/ddl/job_worker.go` as example:
`LoadDDLVars` read variables from `mysql.global_variables` and store them in session. But later we don't use this value, instead, we read values from global variable `DDLErrorCountLimit`.
So this function is quite redundant.

zhangjinpeng87:
bugbot run

pkg/ddl/job_worker.go
```diff
@@ -1071,15 +1071,17 @@ func (w *worker) runOneJobStep(
...
+func (w *worker) loadGlobalVars(varName ...string) error {
 	// Get sessionctx from context resource pool.
 	var ctx sessionctx.Context
 	ctx, err := w.sessPool.Get()
 	if err != nil {
 		return errors.Trace(err)
 	}
 	defer w.sessPool.Put(ctx)
-	return util.LoadDDLVars(ctx)
+	return util.LoadGlobalVars(ctx, varName...)
```
joechenrh, line 1084: This function is to read `TiDBDDLErrorCountLimit` from `mysql.global_variable`.
In the current logic, we will read the latest `TiDBDDLErrorCountLimit` before use it.
Since I'm not sure why we do this, just keep it for now.

pkg/executor/test/ddl/ddl_test.go
```diff
@@ -768,113 +768,28 @@ func TestAutoRandomTableOption(t *testing.T) {
 	require.Contains(t, err.Error(), autoid.AutoRandomRebaseNotApplicable)
 }
 
-func TestSetDDLReorgWorkerCnt(t *testing.T) {
```
joechenrh, line 771: This test is related to the old logic of changing `tidb_ddl_reorg_worker_cnt` during ddl execution too

pkg/ddl/backfilling_txn_executor.go
```diff
@@ -261,9 +259,6 @@ func (b *txnBackfillExecutor) adjustWorkerSize() error {
 	reorgInfo := b.reorgInfo
 	job := reorgInfo.Job
 	jc := b.jobCtx
-	if err := loadDDLReorgVars(b.ctx, b.sessPool); err != nil {
-		ddllogutil.DDLLogger().Error("load DDL reorganization variable failed", zap.Error(err))
-	}
```
joechenrh, line 266: We no longer change reorg parameters by reading system variable, see https://github.com/pingcap/tidb/pull/55334

> For session scope variables, the value cannot be changed during DDL job execution.
For global scope variables, the they only provide the default values of session variables.

pkg/ddl/backfilling.go
```diff
@@ -1131,28 +1111,6 @@ func (dc *ddlCtx) writePhysicalTableRecord(
 	return eg.Wait()
 }
 
-func injectCheckBackfillWorkerNum(curWorkerSize int, isMergeWorker bool) error {
```
joechenrh, line 1134: Unused function after removing `loadDDLReorgVars`

tests/realtikvtest/addindextest3/ingest_test.go
```diff
@@ -215,58 +213,6 @@ func TestIngestMVIndexOnPartitionTable(t *testing.T) {
 	}
 }
 
-func TestAddIndexIngestAdjustBackfillWorker(t *testing.T) {
```
joechenrh, line 218: This test can't pass with txn backfill. When using fast reorg, the test can pass, but failpoint is not used.

pkg/meta/model/job.go
```diff
@@ -390,8 +390,8 @@ type Job struct {
 	// SQLMode for executing DDL query.
 	SQLMode mysql.SQLMode `json:"sql_mode"`
 
-	// SessionVars store session variables
-	SessionVars map[string]string `json:"session_vars,omitempty"`
+	// SystemVars store session variables
+	SystemVars map[string]string `json:"system_vars,omitempty"`
```
D3Hunter: for compatibility, don't rename the json tag

and add a comment on why we  the Field  name is different with the json tag

pkg/meta/model/job.go
```diff
@@ -697,13 +697,13 @@ func (job *Job) InFinalState() bool {
 }
 
 // AddSessionVars add a session variable in DDL job.
-func (job *Job) AddSessionVars(name, value string) {
-	job.SessionVars[name] = value
+func (job *Job) AddSessionVars(name string, value string) {
```
CbcWestwolf: Should we also rename it?

D3Hunter:
> Rename job.SessionVars to job.SystemVars, so these usage won't be checked by linter.

maybe refine the linter to only check the method call of SessionVars(), not field access

pkg/meta/model/job.go
joechenrh: Changed back to `session_vars`, and only rename the functions

joechenrh:
> maybe refine the linter to only check the method call of SessionVars(), not field access

The linter part PR has been hold, so I have to change that 🥲

D3Hunter:
> > maybe refine the linter to only check the method call of SessionVars(), not field access
> 
> The linter part PR has been hold, so I have to change that 🥲

can we do lint in another way, maybe do static analyze, and see if we call the target method, i don't quite like the idea that we have to rename fields to adapt to lint

joechenrh:
> i don't quite like the idea that we have to rename fields to adapt to lint

Yeah, I found that we can only match call from `sessionctx.Context.GetSessionVars`, not other interface or struct.

Besides, I think `Get/SetSystemVar` is more proper than `Get/SetSessionVar` because what we actually do is read and store system variables.

joechenrh:
Now we can only get warning only on usage of `Session.GetSessionVars`
/unhold

PR #59851, opened 2025-03-01
objstore: retry on GCS EOF error

D3Hunter:
### What problem does this PR solve?

Issue Number: close #59754

Problem Summary:

### What changed and how does it work?
see https://github.com/googleapis/google-cloud-go/issues/7090 and https://github.com/golang/go/issues/53472, it seems a bug of golang net/http. we workaround it
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #62061, opened 2025-06-27
release-8.5: Fix prefetch buffer and GCS EOF retry (#59496, #59851)

### LINKED ISSUE #59495: performance is not expected for prefetch reader

### LINKED ISSUE #59754: should retry on  GCS error `Get \"https://storage.googleapis.com/storage": EOF`
## Bug Report

when generating subtasks using global sort on GCS, we meet a `Get \"https://storage.googleapis.com/storage/v1/b/global-sort-datasets/o/internal-sort%2F30001%2F755%2Fdata%2F6d55efb2-0766-4699-a8f7-757c1578bf2b_stat%2F0?alt=json&prettyPrint=false&projection=full\": EOF`, in most cases EOF shouldn't be retried, but this case we need to

retry
task fail
master

GMHDBJD:
This PR manually cherry-picks two PRs to `release-8.5` to resolve a dependency.

---

cherry-pick #59496
cherry-pick #59851

Issue Number: close #59495, close #59754

Problem Summary:
This PR addresses two separate but related issues:
1. The prefetch buffer was not always being fully filled, potentially leading to inefficient reads.
2. A bug in Go's `net/http` library can cause `io.EOF` errors when using GCS, which were not being retried, causing operations to fail.

1.  **Prefetch Buffer Fix:** Switched to using `io.ReadFull` to ensure the prefetch buffer is always filled completely.
2.  **GCS EOF Retry:** Added a workaround in the GCS storage layer to specifically retry on `io.EOF` errors, making the connection more resilient to the underlying library bug.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/storage/gcs.go
+++ b/br/pkg/storage/gcs.go
@@ -8,6 +8,7 @@ import (
 	goerrors "errors"
 	"fmt"
 	"io"
+	"net/url"
 	"os"
 	"path"
 	"strings"
@@ -505,6 +506,14 @@ func shouldRetry(err error) bool {
 	if e := (&googleapi.Error{}); goerrors.As(err, &e) {
 		if e.Code == 401 {
 			log.Warn("retrying gcs request due to internal authentication error", zap.Error(err))
+			return true
+		}
+	}
+
+	// workaround for https://github.com/googleapis/google-cloud-go/issues/7090
+	// seems it's a bug of golang net/http: https://github.com/golang/go/issues/53472
+	if e := (&url.Error{}); goerrors.As(err, &e) {
+		if goerrors.Is(e.Err, io.EOF) {
 			return true
 		}
 	}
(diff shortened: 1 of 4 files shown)


Benjamin2037 (approved):
LGTM

D3Hunter (approved):

PR #62085, opened 2025-06-30
store/copr: refresh region cache and retry when no alive tiflash

### LINKED ISSUE #61826: update region cache and rebuild batchCopTask when all tiflash probe failed
## Enhancement
because if all tiflash for one specific region was scaled in, and new tiflash was scaled out. Balance batchCopTask may fail(because it use old region cache to build batchCopTask), and tidb will use original batchCopTask to dispatch MPPTask, which will fail.

Instead we should update region cache actively and rebuild batchCopTask if there is no alive tiflash for batchCopTask. So we should port https://github.com/tikv/client-c/pull/203 to tidb

guo-shaoge:

Issue Number: close #61826

Problem Summary:
1. If there is no alive tiflash, force refresh store cache and retry until backoff fail. (5s)

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/store/copr/BUILD.bazel
+++ b/pkg/store/copr/BUILD.bazel
@@ -86,7 +86,7 @@ go_test(
     embed = [":copr"],
     flaky = True,
     race = "on",
-    shard_count = 32,
+    shard_count = 33,
     deps = [
         "//pkg/kv",
         "//pkg/store/driver/backoff",


pkg/store/copr/batch_coprocessor.go
```diff
@@ -1074,13 +1076,58 @@ func buildBatchCopTasksCore(bo *backoff.Backoffer, store *kvStore, rangesForEach
 				zap.Duration("elapsed", elapsed),
 				zap.Duration("balanceElapsed", balanceElapsed),
 				zap.Int("range len", rangesLen),
-				zap.Int("task len", len(batchTasks)))
+				zap.Int("task len", len(batchTasks)),
+				zap.Int("retry num", retryNum))
```
windtalker, line 1081: Actually, the task number in batch cop does not contain too many information(the max task number is the tiflash store number), I think we should also add region numbers here?

pkg/store/copr/batch_coprocessor.go
```diff
@@ -1074,13 +1076,58 @@ func buildBatchCopTasksCore(bo *backoff.Backoffer, store *kvStore, rangesForEach
...
+			for _, storeID := range allStoresPerRegion {
+				if _, ok := checkAliveStoreTarget[storeID]; ok {
+					storeOk = true
+					break
+				}
+			}
+			if !storeOk {
+				cache.InvalidateCachedRegion(tasks[i].region)
+				needRetry = true
+				// To avoid too many logs.
```
windtalker: better to print all the invalid regions when log level is debug or trace and only print the first 10 invalid regions when log level is higher than debug

guo-shaoge: done

pkg/store/copr/batch_coprocessor.go
guo-shaoge: done

windtalker (approved):
lgtm

wshwsh12 (approved):
LGTM

PR #62094, opened 2025-07-01
planner: Use NDV from the sample for TopN/Buckets

terry1purcell:

Issue Number: ref #61792

Problem Summary:

When building TopN and histogram buckets, we collect a sample of the data - which may be the full set of data if less than (approximately) 100,000 rows, but will be sampled if above this amount. When processing the sample of data, we may have less values in that sample than there are in the full table. This PR will use the "sampleNDV" for calculating the size of the TopN but will continue to use the original NDV when estimating against the full table size for pruning TopN and building buckets. This reduces the risk of hitting the issue in optimizer cardinality estimation where there are zero buckets due to sampling.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/statistics/builder.go
```diff
@@ -351,6 +355,7 @@ func BuildHistAndTopN(
 			curCnt++
 			continue
 		}
+		sampleNDV++
```
elsa0520: Maybe put this line after the comment ```// case 2, meet a different value: counting for the "current" is complete``` is better ~

pkg/statistics/builder.go
```diff
@@ -484,8 +501,8 @@ func BuildHistAndTopN(
 		topNTotalCount += topn.TopN[i].Count
 	}
 
-	if hg.NDV <= lenTopN {
-		// If we've collected everything  - don't create any buckets
+	if haveAllNDV || numBuckets <= 0 {
```
elsa0520, line 506: The code seems that it remove one item from SampleTopN and exclude the TopN values from sample.
But the numBuckets is a input value in function which will be 0 if all sample values can be covered by TopN. So it will return in here.
So what is this PR mainly goal ? Or how does it solve the problem caused by sample? Because it seems that it just makes topn have one less value? I wonder how it indirectly affects the estimation?
（I just don't quite understand the background, not that I have any objection to the purpose of this PR itself.）

terry1purcell: numBuckets <= 0 should never really occur - since numBuckets is based upon the value specified in the ANALYZE statement - such as ANALYZE TABLE T1 WITH 10 BUCKETS. A value of 0 is invalid - it must be a positive value. But this clause is there in case we ever support a value of 0 - which I believe we actually have a customer requirement for this.

But the point of this check - is to exit early if all of the values are in the TopN - since there will be no values remaining in the samples to build buckets. Therefore, we can exit early.

This PR will ONLY reduce the number of values in the TopN by 1 if the NDV (from the full table without sampling) is larger than the NDV within the sample (sampleNDV), and we have all sampleNDV values in the TopN.

If the sampleNDV && NDV are the same - then this does NOT subtract 1 from the TopN.

How this helps estimation is that if we do not find a value in the TopN, or as the last value in a bucket - we will assume that the value is evenly distributed throughout the values stored in the bucket. Therefore, we reduce the risk that sampling results in us assuming zero rows qualify because we don't have any buckets, but the total number of values in the TopN is less than the column/index NDV.

pkg/statistics/builder.go
terry1purcell: Done

0xPoe:
/cc

I will take a look at this PR.

pkg/statistics/builder.go
```diff
@@ -412,16 +417,28 @@ func BuildHistAndTopN(
 		}
 	}
 
+	if allowPruning && sampleNDV > 1 && sampleFactor > 1 && ndv > sampleNDV && len(topNList) >= int(sampleNDV) {
```
qw4990: Agree this new strategy, but should we do it after `pruneTopNItem` in line 431?

qw4990 (approved):

pkg/statistics/builder.go
terry1purcell: That is a reasonable suggestion. I have made this change.

PR #60197, opened 2025-03-20
ddl: modify param at runtime for add-index on DXF global sort

joechenrh:
### What problem does this PR solve?

Issue Number: ref https://github.com/pingcap/tidb/issues/57497, ref https://github.com/pingcap/tidb/issues/57229

Problem Summary:

### What changed and how does it work?

Support dynamic parameters change for add-index on DXF with global sort.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Manual test result TBD.

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #62120, opened 2025-07-01
ddl/dxf: modify param at runtime for add-index with global sort (part 3)

joechenrh:

Issue Number: ref https://github.com/pingcap/tidb/issues/57497, ref https://github.com/pingcap/tidb/issues/57229

This is the third part for supporting dynamic parameters change for add-index with global sort in https://github.com/pingcap/tidb/pull/60197:

Problem Summary:

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/disttask/operator/operator.go
```diff
@@ -127,3 +129,39 @@ func (s *asyncWorker[T, R]) HandleTask(task T, rsFn func(R)) {
 }
 
 func (*asyncWorker[T, R]) Close() {}
+
+// Context is the context used for worker pool
+// TODO(joechenrh): use this context in WorkerPool to simplify
+// the error handling logic of pipelines.
+type Context struct {
```
joechenrh: Will remove this after previous PR merged

CbcWestwolf:
We can remove this line in this PR:
https://github.com/pingcap/tidb/blob/01bfccb1e8e3d3275ebb8821405bb12c27e50a45/pkg/meta/model/job.go#L658

D3Hunter (commented):
7/20 finished

pkg/ddl/ingest/backend.go
```diff
@@ -157,7 +157,7 @@ func (bc *litBackendCtx) CollectRemoteDuplicateRows(indexID int64, tbl table.Tab
 }
 
 func (bc *litBackendCtx) collectRemoteDuplicateRows(indexID int64, tbl table.Table) error {
-	dupeController := bc.backend.GetDupeController(bc.cfg.WorkerConcurrency, nil)
+	dupeController := bc.backend.GetDupeController(bc.cfg.Concurrency(), nil)
```
D3Hunter: backend config also have a region split concurrency, maybe use GetWorkerConcurrency?

pkg/ddl/backfilling_import_cloud.go
```diff
@@ -208,14 +219,39 @@ func (e *cloudImportExecutor) ResetSummary() {
...
-func (*cloudImportExecutor) ResourceModified(context.Context, *proto.StepResource) error {
-	// Will be added in the future PR
+func (e *cloudImportExecutor) ResourceModified(ctx context.Context, newResource *proto.StepResource) error {
+	logutil.Logger(ctx).Info("cloud import executor update resource")
+	newConcurrency := int(newResource.CPU.Capacity())
+	if newConcurrency == e.backend.Concurrency() {
+		return nil
+	}
+
+	eng := e.engine.Load()
```
D3Hunter, line 248: wait engine exist? else if we are creating the engine but hasn't store into this atomic var, the resource will not be set

joechenrh: Yeah, but maybe not wait until engine exist. I think we should align wiht other executor, return error if there is no engine, and let framework retry, like:
https://github.com/pingcap/tidb/blob/c0a5bceae405912f4196ff2382d8497ac3eaa164/pkg/ddl/backfilling_read_index.go#L280-L285

GMHDBJD (approved):
LGTM

pkg/lightning/backend/local/region_job.go
```diff
@@ -973,6 +978,85 @@ type regionJobRetryer struct {
...
+) *dispatcher {
+	return &dispatcher{
+		workerCtx:       workerCtx,
+		jobFromWorkerCh: jobFromWorkerCh,
+		jobWg:           jobWg,
+		retryer:         retryer,
+	}
+}
+
+func (d *dispatcher) run() error {
```
D3Hunter, line 1004: what's done in this loop is more like a post-processing of the job which finished one round of running, we can make this part of logic run directly inside job worker, not sure how much performance gain we get from pipeline execution, this part of logic is quite simple.
but anyway, no need to change in this pr

pkg/lightning/backend/external/engine.go
```diff
@@ -438,10 +445,83 @@ func (e *Engine) loadRangeBatchData(ctx context.Context, jobKeys [][]byte, outCh
...
+		case <-tick.C:
+			e.activeIngestData = filterUnreleased(e.activeIngestData)
+		}
+		if len(e.activeIngestData) == 0 {
+			break
+		}
+	}
+
+	// Now we can safely reset the memory pool.
+	e.Reset()
```
D3Hunter, line 518: the `Reset` never return error, maybe change its signature

pkg/lightning/backend/external/engine.go
```diff
@@ -438,10 +445,83 @@ func (e *Engine) loadRangeBatchData(ctx context.Context, jobKeys [][]byte, outCh
...
+	defer func() {
+		tick.Stop()
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return currBatchSize
+		case <-tick.C:
+			e.activeIngestData = filterUnreleased(e.activeIngestData)
```
D3Hunter: make `filterUnreleased` a method of `Engine` named as `updateActiveIngestDataFlags`

pkg/lightning/backend/external/engine.go
```diff
@@ -133,6 +134,13 @@ type Engine struct {
 
 	memKVsAndBuffers memKVsAndBuffers
 
+	// activeIngestData store flags generated by this engine
+	activeIngestData []*atomic.Bool
```
D3Hunter:
```suggestion
	activeIngestDataFlags []*atomic.Bool
```

pkg/lightning/backend/external/engine.go
```diff
@@ -438,10 +445,83 @@ func (e *Engine) loadRangeBatchData(ctx context.Context, jobKeys [][]byte, outCh
...
+
+	if newBatchSize == currBatchSize {
+		e.activeIngestData = filterUnreleased(e.activeIngestData)
+		return currBatchSize
+	}
+
+	logutil.Logger(ctx).Info("waiting ingest data batch size change",
+		zap.Int("prev batch size", currBatchSize),
+		zap.Int("new batch size", newBatchSize),
+	)
```
D3Hunter: maybe use a new logger with fields, so no need to add those fields every time
```
	logger := logutil.Logger(ctx).With(zap.Int("prev batch size", currBatchSize),
		zap.Int("new batch size", newBatchSize))
```

pkg/lightning/backend/external/engine.go
```diff
@@ -438,10 +445,83 @@ func (e *Engine) loadRangeBatchData(ctx context.Context, jobKeys [][]byte, outCh
...
+		}
+		if len(e.activeIngestData) == 0 {
+			break
+		}
+	}
+
+	// Now we can safely reset the memory pool.
+	e.Reset()
+
+	logutil.Logger(ctx).Info("load ingest data batch size changed",
```
D3Hunter: also log the time used?

pkg/lightning/backend/external/engine.go
```diff
@@ -438,10 +445,83 @@ func (e *Engine) loadRangeBatchData(ctx context.Context, jobKeys [][]byte, outCh
...
+		}
+	}
+	return res
+}
+
+// checkConcurrencyChange checks if concurrency change.
+// If changed, we need to recreate the memory pool. Before this,
+// we should wait for all previous data being consumed. Then we
+// can drop and create new memory pool.
+func (e *Engine) checkConcurrencyChange(ctx context.Context, currBatchSize int) int {
```
D3Hunter:
```suggestion
func (e *Engine) handleConcurrencyChange(ctx context.Context, currBatchSize int) int {
```
maybe called `handleXXX`, except checking, it also wait all active ingest data released and notify that it's done

pkg/lightning/backend/external/engine.go
```diff
@@ -518,12 +601,49 @@ func (e *Engine) buildIngestData(kvs []KVPair, buf []*membuf.Buffer) *MemoryInge
...
+// UpdateResource changes the concurrency and the memory pool size of this engine.
+func (e *Engine) UpdateResource(ctx context.Context, concurrency int, memCapacity int64) error {
+	v := e.worker.Load()
+	if v == nil {
+		return errors.New("region job worker is not initialized, retry later")
+	}
+	worker, _ := v.(operator.TunableOperator)
+
+	e.workerConcurrency.Store(int32(concurrency))
+	e.memLimit = int(float64(memCapacity) / writeStepMemShareCount * 3)
```
D3Hunter: 2 place using the same formula, maybe abstract a method

pkg/lightning/backend/external/engine.go
```diff
@@ -518,12 +601,49 @@ func (e *Engine) buildIngestData(kvs []KVPair, buf []*membuf.Buffer) *MemoryInge
...
+
+	// Wait for current data being consumed and new memory pool created.
+	for {
+		select {
+		case <-ctx.Done():
+			logutil.BgLogger().Info("context done when updating resource, skip update resource")
```
D3Hunter:
```suggestion
			logutil.Logger(ctx).Info("context done when updating resource, skip update resource")
```
others too, so we can print the fields like task id with it

pkg/lightning/backend/external/engine.go
```diff
@@ -518,12 +601,49 @@ func (e *Engine) buildIngestData(kvs []KVPair, buf []*membuf.Buffer) *MemoryInge
...
 }
 
+// SetWorker sets the worker for this engine.
+func (e *Engine) SetWorker(worker operator.TunableOperator) {
+	e.worker.Store(worker)
+}
+
+// UpdateResource changes the concurrency and the memory pool size of this engine.
+func (e *Engine) UpdateResource(ctx context.Context, concurrency int, memCapacity int64) error {
+	v := e.worker.Load()
```
D3Hunter: also check `e.workerConcurrency.Load() == int32(concurrency)`? as the `checkConcurrencyChange` won't send to the ready channel in this case

pkg/lightning/backend/external/engine.go
```diff
@@ -518,12 +601,49 @@ func (e *Engine) buildIngestData(kvs []KVPair, buf []*membuf.Buffer) *MemoryInge
...
+	if v == nil {
+		return errors.New("region job worker is not initialized, retry later")
+	}
+	worker, _ := v.(operator.TunableOperator)
+
+	e.workerConcurrency.Store(int32(concurrency))
+	e.memLimit = int(float64(memCapacity) / writeStepMemShareCount * 3)
+
+	// Wait for current data being consumed and new memory pool created.
+	for {
```
D3Hunter: seems no need a for loop

pkg/lightning/backend/external/engine.go
```diff
@@ -133,6 +134,13 @@ type Engine struct {
 
 	memKVsAndBuffers memKVsAndBuffers
 
+	// activeIngestData store flags generated by this engine
+	activeIngestData []*atomic.Bool
+
+	// region job worker related to this engine.
+	// We store the operator.TunableOperator interface in it to avoid import cycle.
+	worker atomic.Value
```
D3Hunter:
```suggestion
	worker atomic.Pointer[operator.TunableOperator]
```

pkg/lightning/backend/external/engine_test.go
```diff
@@ -364,3 +324,71 @@ func TestEngineOnDup(t *testing.T) {
...
+		})
+	}
+
+	t.Run("reduce concurrency", func(t *testing.T) {
+		resetFn()
+		// Wait part of the data being processed
+		require.Eventually(t, func() bool {
+			return finished.Load() > 2
+		}, time.Second, 10*time.Millisecond)
+		e.UpdateResource(context.Background(), 1, 1024)
```
D3Hunter: check return value, below too

pkg/lightning/backend/external/engine_test.go
```diff
@@ -364,3 +324,71 @@ func TestEngineOnDup(t *testing.T) {
 		}
 	})
 }
+
+func TestChangeEngineConcurrency(t *testing.T) {
```
D3Hunter: after run with coverage, this case test nothing, as `e.worker` is always nil, so nothing happened

pkg/lightning/backend/local/job_worker.go
```diff
@@ -27,26 +27,40 @@ import (
...
+	"github.com/pingcap/tidb/pkg/disttask/operator"
 	"github.com/pingcap/tidb/pkg/ingestor/engineapi"
 	"github.com/pingcap/tidb/pkg/ingestor/errdef"
 	"github.com/pingcap/tidb/pkg/ingestor/ingestcli"
 	"github.com/pingcap/tidb/pkg/lightning/common"
 	"github.com/pingcap/tidb/pkg/lightning/log"
 	"github.com/pingcap/tidb/pkg/lightning/membuf"
 	"github.com/pingcap/tidb/pkg/metrics"
+	"github.com/pingcap/tidb/pkg/resourcemanager/pool/workerpool"
+	rutil "github.com/pingcap/tidb/pkg/resourcemanager/util"
```
D3Hunter:
```suggestion
	rcmgrutil "github.com/pingcap/tidb/pkg/resourcemanager/util"
```

pkg/lightning/backend/local/job_worker.go
```diff
@@ -64,88 +78,104 @@ type regionJobBaseWorker struct {
...
+func (w *regionJobBaseWorker) HandleTask(job *regionJob, _ func(*regionJob)) error {
+	// As we need to call job.done() after panic, we recover here rather than in worker pool.
+	defer putil.Recover("fast_check_table", "handleTableScanTaskWithRecover", func() {
+		w.ctx.OnError(errors.Errorf("region job worker panic"))
+		job.done(w.jobWg)
+	}, false)
+
+	failpoint.Inject("injectPanicForRegionJob", func() {
+		panic("mock panic")
+	})
```
D3Hunter:
```suggestion
	failpoint.Inject("injectPanicForRegionJob")
```
you can use the `panic` action directly, see https://github.com/pingcap/failpoint/blob/b51a646c7c862d49540d5109da93c13f2d8b9264/failpoints_test.go#L228

pkg/lightning/backend/local/job_worker.go
```diff
@@ -403,3 +433,79 @@ func (w *objStoreRegionJobWorker) ingest(ctx context.Context, job *regionJob) er
...
+	)
+
+	op := &jobOperator{
+		ctx:           wctx,
+		AsyncOperator: operator.NewAsyncOperator(wctx, pool),
+		workerGroup:   workGroup,
+	}
+	op.SetSink(operator.NewSimpleDataChannel(jobFromWorkerCh))
+	op.SetSource(operator.NewSimpleDataChannel(sourceChannel))
+	return op
```
D3Hunter: we don't need the operator abstraction, we only need the worker pool to help us to tune resource usage. maybe use the worker pool directly, without introducing the operator.

the initial thinking of switch to use the operator abstraction is to help us to extend and better performance, but it turns out to be a very bad abstraction for DXF task scenario, there is no gains on performance point of view, and i also don't see additional help on extensibility. but it makes the code logic complex and harder to reason

D3Hunter (commented):
please consider split the PR into multiple smaller ones next time, especially if your changes include refactor works

pkg/lightning/backend/external/engine_test.go
joechenrh: Ohh, I added a dummy worker to make concurrency change works.

pkg/lightning/backend/external/engine.go
joechenrh: Yeah, it's not needed, removed.

pkg/lightning/backend/external/engine.go
joechenrh: Introduce a new interface in `workerpool.go` to avoid import cycle here.

pkg/lightning/backend/local/job_worker.go
joechenrh: Revert back to use workerpool directly.

pkg/lightning/backend/external/engine.go
joechenrh: Change the function signature to `func (e *Engine) Reset()`

pkg/lightning/backend/local/region_job.go
joechenrh: It's introduced in https://github.com/pingcap/tidb/pull/42780, maybe the purpose is to make the pipeline of the region jobs clearer.

pkg/lightning/backend/external/engine.go
joechenrh: Abstract it as `getEngineMemoryLimit`

pkg/lightning/backend/external/engine.go
```diff
@@ -82,6 +83,12 @@ import (
 //	|  512M |       256M(2) |       512M(1) |
 const writeStepMemShareCount = 6.5
 
+// getEngineMemoryLimit calculates the memory limit for external engine according to the memory capacity.
```
D3Hunter:
```suggestion
// getEngineMemoryLimit calculates the memory limit for external engine according
// to the memory capacity.
```

pkg/lightning/backend/external/engine.go
```diff
@@ -133,6 +140,13 @@ type Engine struct {
 
 	memKVsAndBuffers memKVsAndBuffers
 
+	// activeIngestDataFlags store flags generated by this engine
+	activeIngestDataFlags []*atomic.Bool
+
+	// region job worker pool related to this engine.
+	// We store the operator.TunableOperator interface here to avoid import cycle.
```
D3Hunter:
```suggestion
```

pkg/lightning/backend/external/engine.go
```diff
@@ -133,6 +140,13 @@ type Engine struct {
...
+	worker atomic.Pointer[workerpool.Tuner]
```
D3Hunter:
```suggestion
	workerPool atomic.Pointer[workerpool.Tuner]
```

pkg/lightning/backend/external/engine.go
```diff
@@ -522,12 +608,52 @@ func (e *Engine) buildIngestData(kvs []KVPair, buf []*membuf.Buffer) *MemoryInge
...
+	if worker == nil {
+		return errors.New("region job worker is not initialized, retry later")
+	}
+
+	if e.workerConcurrency.Load() == int32(concurrency) {
+		return nil
+	}
+
+	e.workerConcurrency.Store(int32(concurrency))
+	e.memLimit = getEngineMemoryLimit(memCapacity)
```
D3Hunter, line 635: data race with `handleConcurrencyChange` which calls `Reset` and read this value.
and since we are triggering `handleConcurrencyChange` by previous worker concurrency change, if the `Reset` runs between those 2 field change, it might still use the old mem limit. so maybe reverse the order, and make it a atomic would be fine too, although not a must

pkg/lightning/backend/external/engine_test.go
```diff
@@ -364,3 +324,86 @@ func TestEngineOnDup(t *testing.T) {
...
+			return len(outCh) >= 4
+		}, 5*time.Second, 10*time.Millisecond)
+		require.NoError(t, e.UpdateResource(context.Background(), 8, 1024))
+		require.NoError(t, eg.Wait())
+	})
+
+	t.Run("increase concurrency after loading all data", func(t *testing.T) {
+		resetFn()
+		close(updatedCh)
+		// Wait all of the data being processed
```
D3Hunter:
```suggestion
		// Wait all the data being processed
```

pkg/lightning/backend/local/job_worker_test.go
```diff
@@ -27,149 +28,234 @@ import (
...
-			},
-			regenerateJobsFn: func(
-				ctx context.Context, data engineapi.IngestData, sortedJobRanges []engineapi.Range,
-				regionSplitSize, regionSplitKeys int64,
-			) ([]*regionJob, error) {
-				return []*regionJob{
-					{}, {}, {},
-				}, nil
-			},
+func newRegionJobOperatorForTest(
```
D3Hunter:
```suggestion
func newRegionJobWorkerPoolForTest(
```

pkg/lightning/backend/local/local.go
```diff
@@ -511,6 +515,16 @@ func (c *BackendConfig) adjust() {
 	c.MaxOpenFiles = max(c.MaxOpenFiles, openFilesLowerThreshold)
 }
 
+// GetWorkerConcurrency gets the current concurrency of the backend
+func (c *BackendConfig) GetWorkerConcurrency() int {
+	return int(c.WorkerConcurrency.Load())
+}
+
+// SetConcurrency sets the current concurrency of the backend
+func (c *BackendConfig) SetConcurrency(concurrency int) {
```
D3Hunter:
```suggestion
func (c *BackendConfig) SetWorkerConcurrency(concurrency int) {
```

pkg/lightning/backend/local/local.go
```diff
@@ -1400,34 +1433,46 @@ func (local *Backend) doImport(
...
-	//
-	// 2. All other components will exit because of the canceled context. No need to
-	// close channels.
+	// 1. Error occurs in worker pool
+	//      1.1 The subcontext will be canceled and cause its goroutine to exit.
+	//      1.2 The error is then broadcasted to the workGroup.
+	//      1.3 All other components exit due to canceled context. No need to close channels.
 	//
-	// 3. the main goroutine can see the error and exit after workGroup.Wait().
+	// 2. Error occuers in other components (e.g. retryer, dispatcher,)
```
D3Hunter:
```suggestion
	// 2. Error occurs in other components (e.g. retryer, dispatcher)
```

pkg/lightning/backend/local/region_job_test.go
```diff
@@ -617,3 +621,164 @@ func TestUpdateAndGetLimiterConcurrencySafety(t *testing.T) {
 	}
 	wg.Wait()
 }
+
+func TestWorkerPoolWithErrors(t *testing.T) {
+	if kerneltype.IsNextGen() {
+		t.Skip("skip this test on next-gen kernel")
```
D3Hunter: please add the reason of skip in the message

joechenrh: Add the test back for next-gen.

Issue #61999, opened 2025-06-25
dumpling: Parallel processing implementation for string-type


takaidohigasi:
## Enhancement

Currently, parallelization supports only numeric types for dumpling. so support string value.

## related implementation

* pickupPossibleField
    https://github.com/pingcap/tidb/blob/master/dumpling/export/sql.go#L1286-L1299
* estimateCount
    https://github.com/pingcap/tidb/blob/master/dumpling/export/sql.go#L1301
    https://github.com/pingcap/tidb/blob/master/dumpling/export/sql.go#L1306
* detectEstimateRows
    https://github.com/pingcap/tidb/blob/master/dumpling/export/sql.go#L1343

## Pull Request

https://github.com/pingcap/tidb/pull/62172

PR #62172, opened 2025-07-03
dumpling: improve string key handling, streaming process with chunking

takaidohigasi:
### Latest state (2026-04-24)

Updates since initial submission:

- **Merged master** (1a87518). Resolved `dumpling/export/dump_test.go` import conflict (`sync` + `sync/atomic`) and semantic merge in `writer_serial_test.go` (switched PR's new tests from the removed `storage.NewBufferWriter()` to master's local `NewBufferWriter()` / `BytesWriter`).
- **Added a large-scale round-trip CI test** (919858e): `dumpling/tests/composite_string_key_large/` — 500-row events table with shared PK prefix and 300-row translations table with a 3-column composite key and unicode. Runs dumpling at `--rows 50`, asserts chunks ≥ 2 and every chunk is a self-contained INSERT ending with `;` (regression guard for the `isLastChunk` discussion), then round-trips via `tidb-lightning` + `sync_diff_inspector` for MySQL-source vs TiDB-target checksum equality.
- **Removed LLM-noise helpers and flattened the chunking API** (629efbb), in response to @lance6716's review. −700 net lines. Details in [issuecomment-4311143074](https://github.com/pingcap/tidb/pull/62172#issuecomment-4311143074). Highlights: dropped `interpolateStringBoundary`, `buildStringWhereClauses`, `Config.IsStringChunking`, the singular backward-compat `concurrentDumpStringField`, the redundant small-table guard, and the duplicate `SHOW INDEX` query. Downgraded per-iteration Info logs to Debug.

Manual verification against the latest head: dumping the same 500-row composite-string-PK dataset with the `master` binary and this branch's binary produces 1 chunk vs 10 chunks respectively, with byte-identical row content after normalization (sort + strip trailing `,`/`;`).

---

Issue Number: ref #61999

Problem Summary:

* dumpling is processed in parallel only for integer PK [ref1](https://github.com/pingcap/tidb/blob/master/dumpling/export/sql.go#L1286-L1299) [ref2](https://github.com/pingcap/tidb/blob/master/dumpling/export/dump.go#L615-L630)
    * takes long time to complete
    * string pk is handled sequentially, regarded as 1 chunk, resulted in un-predictable estimation of remaining process time.

* Implemented to handle primary keys starting with a string streamingly. writing might be in parallel processing
    * chunking and writing in parallel
    * chunking (sequential process)
        * cursor based chunking: use composit key condition(also available for MySQL 5.7 or lower or version which we can't use `ROW_NUMBER()` function)

* note: I don't change the existing integer based implementation.

#### simplified diagram

```
  Timeline: 0----5----10---15---20---25---30---35---40---45---50

  * BEFORE (Sequential Approach):
 
  Chunking: [------C1------]
  Writing:                  [================W1================]
  Total Time: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  
  * AFTER (Pipeline/Parallel Approach):

  Chunking: [C1][C2][C3][C4]
  Writing:      [====W1====]
                    [====W2====]
                        [====W3====]
                            [====W4====]
  Total Time: ~~~~~~~~~~~~~~~~~~~~~~~~~~~

chunking is sequential but writing start just after each chunking completed, and run in parallel.
```

### Check List

Tests 

- [x] Unit test
- [x] Integration test

* step
  1. create records for composit key
  2. confirm the number of record
  3. execute dumpling with old and new version and compare the number of records and processing time.

v8.5.2
```
[2025/07/04 07:13:09.746 +09:00] [INFO] [status.go:37] [progress] [tables="40/41 (97.6%)"] ["finished rows"=21907452704] ["estimate total rows"=16365664263] ["finished size"=1.661TB] ["average speed(MiB/s)"=53.78349999284309] ["recent speed bps"=56396034.34047334] ["chunks progress"="99.97 %"]
[2025/07/04 07:13:49.491 +09:00] [INFO] [collector.go:264] ["backup success summary"] [total-ranges=3005] [ranges-succeed=3005] [ranges-failed=0] [total-take=6h32m39.744718607s] [total-kv-size=1.663TB] [average-speed=70.6MB/s] [total-rows=21943318733]
```

this PR version
```
[2025/07/04 12:13:58.240 +09:00] [INFO] [status.go:37] [progress] [tables="33/41 (80.5%)"] ["finished rows"=21474305967] ["estimate total rows"=16365664263] ["finished size"=1.644TB] ["average speed(MiB/s)"=387.8366052577693] ["recent speed bps"=406675792.1587332] ["chunks progress"="98.02 %"]
[2025/07/04 12:15:20.846 +09:00] [INFO] [collector.go:264] ["backup success summary"] [total-ranges=21071] [ranges-succeed=21071] [ranges-failed=0] [total-take=4h49m22.606312684s] [total-kv-size=1.663TB] [average-speed=95.79MB/s] [total-rows=21943318733]
```
      => number of records(tota-rows) is the same, less total-take. 

I tested for MySQL `5.7.34-log`

I understand version matrix is covered by
https://github.com/pingcap/tidb/blob/master/.github/workflows/integration-test-dumpling.yml#L51-L56

- [ ] No need to test
  > 

Side effects

no,  I don't change dumpling command interface.

Documentation

- [ ] Affects user behaviors

https://docs.pingcap.com/tidb/stable/dumpling-overview/#export-to-sql-files

> The -r option enables the in-table concurrency to speed up the export. The default value is 0, which means disabled. A value greater than 0 means it is enabled, and the value is of INT type. When the source database is TiDB, a -r value greater than 0 indicates that the TiDB region information is used for splitting, and reduces the memory usage. The specific -r value does not affect the split algorithm. When the source database is MySQL and the primary key or the first column of the composite primary key is of the INT type, specifying -r can also enable the in-table concurrency.

=> fixed in
https://github.com/pingcap/docs/pull/21324

- [ ] Contains syntax changes
no
no
no
no

### Release note

```release-note
improve dumpling string key handling, streaming process with chunking is available
```

## Summary by CodeRabbit

* **New Features**
  * Streaming export for composite string primary-key tables (string-key chunking)
  * Best-effort real-time progress percentage for streaming chunked exports

* **Bug Fixes**
  * Improved chunk boundary generation and stable row ordering for composite-key dumps
  * INSERT output now always emits well-formed, terminated statements

* **Tests**
  * Expanded unit/integration tests for string-key chunking, SQL escaping, WHERE/cursor generation, progress, and writer chunking

* **Documentation**
  * Added integration test README and runnable test suites with fixtures and validation scripts


lance6716:
This is a good idea !

/cc @OliverS929

takaidohigasi:
```
taka-h@LCW9RLFMFX ~/git/tidb % bazel build //dumpling/export:export_test
INFO: Reading 'startup' options from /Users/taka-h/git/tidb/.bazelrc: --host_jvm_args=-Xmx4g, --unlimit_coredumps
INFO: Options provided by the client:
  Inherited 'common' options: --isatty=1 --terminal_columns=228
INFO: Reading rc options for 'build' from /Users/taka-h/git/tidb/.bazelrc:
  'build' options: --http_timeout_scaling=0.1 --announce_rc --experimental_guard_against_concurrent_changes --incompatible_strict_action_env --incompatible_enable_cc_toolchain_resolution
INFO: Analyzed target //dumpling/export:export_test (2 packages loaded, 12 targets configured).
INFO: Found 1 target...
Target //dumpling/export:export_test up-to-date:
  bazel-bin/dumpling/export/export_test_/export_test
INFO: Elapsed time: 0.259s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
```

takaidohigasi:
```
nogo: errors found by nogo during build-time code analysis:

dumpling/export/dump_test.go:406:7: unused write to field tctx (unusedwrite)

dumpling/export/dump_test.go:407:7: unused write to field conf (unusedwrite)

dumpling/export/dump_test.go:408:12: unused write to field cancelCtx (unusedwrite)

dumpling/export/dump_test.go:457:7: unused write to field tctx (unusedwrite)

dumpling/export/dump_test.go:458:7: unused write to field conf (unusedwrite)

dumpling/export/dump_test.go:459:12: unused write to field cancelCtx (unusedwrite)

dumpling/export/string_chunking.go:79:26: var-declaration: should drop = 0 from declaration of var totalChunks; it is the zero value (revive)

dumpling/export/dump.go:358:38: unused-parameter: parameter 'task' seems to be unused, consider removing or renaming it as _ (revive)

dumpling/export/sql_util.go:73:5: early-return: if c { ... } else { ... break } can be simplified to if !c { ... break } ... (all_revive)
```

I understand which one should I resolve.

dumpling/export/BUILD.bazel
```diff
@@ -18,6 +19,7 @@ go_library(
         "sql.go",
         "sql_type.go",
         "sql_util.go",
+        "sql_util_test.go",
```
takaidohigasi: I once fixed but why reverted...?
https://github.com/pingcap/tidb/pull/62172/commits/084e8ac8ca3be698ca83e2cbaf67265b885ea2e9

takaidohigasi:
/home/jenkins/.tidb/tmp/63a9840cd0739f2c243bb46478607469/execroot/__main__/bazel-out/k8-fastbuild/testlogs/dumpling/export/export_test/shard_8_of_50/coverage.dat
Test cases: finished with 4257 passing and 1 failing out of 4258 test cases

Executed 1 out of 467 tests: 233 tests pass and 234 were skipped.
make: *** [Makefile:661: bazel_coverage_test] Error 3
root: /home/jenkins/.tidb/tmp/63a9840cd0739f2c243bb46478607469/execroot/__main__/bazel-out/k8-fastbuild/testlogs
complete to collect bazel result.
script returned exit code 2

takaidohigasi:
I will fixed up commits if all tests are passed

takaidohigasi:
https://docs.pingcap.com/tidb/stable/dumpling-overview/

need to fix the focument

takaidohigasi:
```

Caused by: java.util.concurrent.TimeoutException: not ready after 15000 MILLISECONDS

	... 79 more

Retrying in 8s ...

java.util.concurrent.TimeoutException

	at PluginClassLoader for kubernetes-client-api//io.fabric8.kubernetes.client.utils.AsyncUtils.lambda$withTimeout$0(AsyncUtils.java:42)

	at PluginClassLoader for kubernetes-client-api//io.fabric8.kubernetes.client.utils.Utils.lambda$schedule$6(Utils.java:473)

	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)

	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)

	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)

	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)

	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)

	at java.base/java.lang.Thread.run(Unknown Source)

Retrying...

Failed to start websocket connection: io.fabric8.kubernetes.client.KubernetesClientException: An error has occurred.

	at PluginClassLoader for kubernetes-client-api//io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:129)

	at PluginClassLoader for kubernetes-client-api//io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:122)

	at PluginClassLoader for kubernetes-client-api//io.fabric8.kubernetes.client.utils.Utils.waitUntilReadyOrFail(Utils.java:186)

	at PluginClassLoader for kubernetes-client-api//io.fabric8.kubernetes.client.dsl.internal.core.v1.PodOperationsImpl.setupConnectionToPod(PodOperationsImpl.java:387)

	at PluginClassLoader for kubernetes-client-api//io.fabric8.kubernetes.client.dsl.internal.core.v1.PodOperationsImpl.exec(PodOperationsImpl.java:293)

	at PluginClassLoader for kubernetes//org.csanchez.jenkins.plugins.kubernetes.pipeline.ContainerExecDecorator$1.doLaunch(ContainerExecDecorator.java:517)

	at PluginClassLoader for kubernetes//org.csanchez.jenkins.plugins.kubernetes.pipeline.ContainerExecDecorator$1.launch(ContainerExecDecorator.java:376)
```

takaidohigasi:
re running test after rebase (fix-up commits)

takaidohigasi:
all tests were passed!

OliverS929:
@takaidohigasi Thank you for your contribution! I started reviewing this PR this morning and will invite others to take a look as well. 
@D3Hunter @GMHDBJD @River2000i — feel free to join the review when you have time.

lance6716:
I just deleted some comments which are related to bot to improve readability 😂 our bot really comments too much

takaidohigasi:
I will also take a look for comment implementation.

If it is Github Actions one, I also try to improve the comment

takaidohigasi:
it seems it isn't GitHub Actions based comment, so I can't touch it.
https://github.com/pingcap/tidb/tree/master/.github

OliverS929:
My review of this PR will be delayed, as I’m currently tied up with higher-priority tasks. In the meantime, please feel free to refer to the reviewers I tagged above for their input.

takaidohigasi:
thanks.

I also want to add the change like this also to sync-diff inspector if this PR is approved.

lance6716:
Hi @takaidohigasi we have fixed dumpling in other PR. Now you can see some errors in your new tests.

takaidohigasi:
```
+(dumpling/tests/composite_string_key/run.sh:35): diff dumpling/tests/composite_string_key/result/comp_str_case_0.sql.clean /tmp/dumpling_test_result/sql_res.composite_string_key/combined_comp_str_case_0.sql.clean
6c6,7
< ('tenant_b','user_002','Eve Brown','eve@tenant-b.com','2023-01-19 16:30:00'),
---
> ('tenant_b','user_002','Eve Brown','eve@tenant-b.com','2023-01-19 16:30:00');
> INSERT INTO `comp_str_case_0` VALUES
11c12
< ('tenant_d','user_001','Jack Moore','jack@tenant-d.com','2023-01-24 10:15:00');
\ No newline at end of file
---
> ('tenant_d','user_001','Jack Moore','jack@tenant-d.com','2023-01-24 10:15:00');
make: *** [Makefile:588: dumpling_integration_test] Error 1
```

takaidohigasi:
https://tiprow.hawkingrei.com/view/gs/pingcapprow/pr-logs/pull/pingcap_tidb/62172/tidb_parser_test/1943593869308334080

```
$ PWD=/home/prow/go/src/github.com/pingcap/tidb  git fetch https://x-access-token:CENSORED@github.com/pingcap/tidb.git 3c7d6c705ac4205a98181c5ce9eb5e0e7dfcfbed (runtime: 2.920887776s)
From https://github.com/pingcap/tidb
 * branch                3c7d6c705ac4205a98181c5ce9eb5e0e7dfcfbed -> FETCH_HEAD
$ PWD=/home/prow/go/src/github.com/pingcap/tidb GIT_AUTHOR_DATE=1752215980 GIT_COMMITTER_DATE=1752215980 git merge --no-ff 3c7d6c705ac4205a98181c5ce9eb5e0e7dfcfbed (runtime: 6.334254ms)
fatal: refusing to merge unrelated histories
# Error: exit status 128
```

?

takaidohigasi:
will merge remote upstream once

takaidohigasi:
```
/tmp/dumpling_test_result/sql_res.composite_string_key/combined_comp_str_case_1.sql.clean
4c4
< ('electronics','item_002','Laptop "Gaming Edition"',1299.50),
---
> ('electronics','item_002','Laptop \"Gaming Edition\"',1299.50),
6,7c6
< ('home & garden','item_001','Garden Tools Set',89.99);
< INSERT INTO `comp_str_case_1` VALUES
---
> ('home & garden','item_001','Garden Tools Set',89.99),
11c10
< ('toys & games','item_001','Board Game "Strategy Master"',45.00),
---
> ('toys & games','item_001','Board Game \"Strategy Master\"',45.00),
make: *** [Makefile:588: dumpling_integration_test] Error 1
```

behavior of escaping is changed.

takaidohigasi:
I will fix some.

takaidohigasi:
I was confused again and again, but

https://github.com/pingcap/tidb/blob/0434f6de7b70de51d70d6a27f087c1908fbc22dc/dumpling/tests/composite_string_key/data/comp_str_case_3.sql#L19
https://github.com/pingcap/tidb/blob/0434f6de7b70de51d70d6a27f087c1908fbc22dc/dumpling/tests/composite_string_key/result/comp_str_case_3.sql#L10


```
: input
('fr', 'goodbye', 'Merci d''utiliser notre service'),
: out
('fr','goodbye','Merci d\'utiliser notre service'),
```

and so on.

confirmed this is compatible with current master branch behavior. (by escaping mode)

dumpling/tests/composite_string_key/result/comp_str_case_1.sql
```diff
@@ -0,0 +1,13 @@
+/*!40014 SET FOREIGN_KEY_CHECKS=0*/;
+/*!40101 SET NAMES binary*/;
```
takaidohigasi, line 2: `/*!40101 SET NAMES binary*/;` currently fixed one even if we set collation.

takaidohigasi: I will create another branch for this.

takaidohigasi:
```
++(dumpling/tests/rows/run.sh:52): find /tmp/dumpling_test_result/sql_res.rows -maxdepth 1 -iname 'rows.t.*.sql'
++(dumpling/tests/rows/run.sh:52): wc -l
+(dumpling/tests/rows/run.sh:52): file_num=41
+(dumpling/tests/rows/run.sh:53): '[' 41 -ne 50 ']'
+(dumpling/tests/rows/run.sh:54): echo 'obtain file number: 41, but expect: 50'
+(dumpling/tests/rows/run.sh:54): exit 1
```

dumpling/export/string_chunking.go
```diff
@@ -0,0 +1,332 @@
...
+
+// concurrentDumpStringFields handles composite key chunking with multiple columns
+func (d *Dumper) concurrentDumpStringFields(tctx *tcontext.Context, conn *BaseConn, meta TableMeta, taskChan chan<- Task, fields []string, orderByClause string, estimatedCount uint64) error {
+	conf := d.conf
+	db, tbl := meta.DatabaseName(), meta.TableName()
+
+	// Calculate total count and chunk parameters
+	totalCount := int64(estimatedCount)
+	if totalCount <= 0 {
+		totalCount = int64(conf.Rows) * 5 // Conservative fallback
```
lance6716: this logic will not take effect. `estimatedCount` comes from `count := estimateCount(d.tctx, db, tbl, conn, field, conf)` in dump.go:767. And in dump.go:772 there's a check for `if count < conf.Rows {`. if code runs here `totalCount == 0`, it means `count < conf.Rows` is false. Which further means `conf.Rows` is also 0. So `int64(conf.Rows) * 5` is also zero.

dumpling/export/writer_util.go
```diff
@@ -148,6 +148,10 @@ func WriteInsert(
 	tblIR TableDataIR,
 	w storage.ExternalFileWriter,
 	metrics *metrics,
+	chunkIndex int,
+	_ int,
+	isLastChunk bool,
+	isStringChunking bool,
```
lance6716: nit: I see there are lots of changes to pass `isStringChunking` through function. Maybe it's easier to save it in `Config`. There are non-persistent (calculated and fill after code runs) fields like `TableFilter`. `IsStringChunking` can also be handled like it.

dumpling/export/writer_util.go
```diff
@@ -212,20 +216,31 @@ func WriteInsert(
...
-			wrapBackTicks(escapeString(meta.TableName())), selectedField)
-	} else {
-		insertStatementPrefix = fmt.Sprintf("INSERT INTO %s VALUES\n",
-			wrapBackTicks(escapeString(meta.TableName())))
+	// Determine if we should write INSERT prefix based on chunk position
+	// For string chunking, only first chunk gets prefix; for row chunking, each chunk is independent
+	shouldWritePrefix := (chunkIndex == 0) || !isStringChunking
+
+	var insertStatementPrefixLen uint64
+	if shouldWritePrefix {
```
lance6716: just a glance, we still have 1 writer for 1 `WriteInsert` call? So different chunks will write to different files, there's no chance to process a `false` `shouldWritePrefix`

lance6716 (commented):
I noticed that some of the changes in this PR seem over-designed or has no effect, which might be due to using an LLM for code generation. These unrelated modifications make the code change longer and review a bit more difficult.

Can you ask your LLM to try to clean these cases 😄

takaidohigasi:
You are absolutely right and I will refine codes.

dumpling/export/string_chunking.go
takaidohigasi: Removed in
https://github.com/pingcap/tidb/pull/62172/commits/46e88ca283058eb4f9fd1c02e54e6afac545fc3f

dumpling/export/writer_util.go
takaidohigasi: simplified interface by using Confg. thanks so much
https://github.com/pingcap/tidb/pull/62172/commits/b41f6a871c375540667726feca0ffb3da3e6c283

dumpling/export/writer_util.go
```diff
@@ -212,20 +215,31 @@ func WriteInsert(
...
 	selectedField := meta.SelectedField()
 
-	// if has generated column
-	if selectedField != "" && selectedField != "*" {
-		insertStatementPrefix = fmt.Sprintf("INSERT INTO %s (%s) VALUES\n",
-			wrapBackTicks(escapeString(meta.TableName())), selectedField)
-	} else {
-		insertStatementPrefix = fmt.Sprintf("INSERT INTO %s VALUES\n",
-			wrapBackTicks(escapeString(meta.TableName())))
+	// Determine if we should write INSERT prefix based on chunk position
```
lance6716: I think we don't need to change old logic. As `tryToWriteTableData`, there's `chunkIdx` in input arguments. We build a new writer for a given `chunkIdx`. The INSERT INTO prefix is always written.

dumpling/export/writer_util.go
takaidohigasi: Yes, you are right.
https://github.com/pingcap/tidb/pull/62172/commits/ed0afc1721e326d2a6b850080dd105c9acdf56eb

dumpling/export/writer_util.go
takaidohigasi: thanks. right.
https://github.com/pingcap/tidb/pull/62172/commits/ed0afc1721e326d2a6b850080dd105c9acdf56eb

takaidohigasi:
https://tiprow.hawkingrei.com/view/gs/pingcapprow/pr-logs/pull/pingcap_tidb/62172/fast_test_tiprow/1946220607515398144

caching error? (bazel build --config=ci //... --//build:with_nogo_flag=true --//build:with_rbe_flag=true succeeded in my laptop)
```
ERROR: /home/prow/go/src/github.com/pingcap/tidb/pkg/statistics/handle/autoanalyze/priorityqueue/BUILD.bazel:45:8: Running nogo on //pkg/statistics/handle/autoanalyze/priorityqueue:priorityqueue_test failed: (Exit 34): Missing digest: ab994e29c9dcc214ddaf54c160649197e1b2aac909d234951f8738ce0401fe87/1974398 for /root/.cache/bazel/_bazel_root/37ff7b0eddea115f60ab9f853788d7f5/execroot/_main/bazel-out/k8-fastbuild/bin/pkg/statistics/handle/autoanalyze/priorityqueue/ingest.recompile1333.x
ERROR: /home/prow/go/src/github.com/pingcap/tidb/pkg/statistics/handle/autoanalyze/priorityqueue/BUILD.bazel:45:8: GoCompilePkg pkg/statistics/handle/autoanalyze/priorityqueue/ddl.recompile1334.a failed: (Exit 34): Missing digest: ab994e29c9dcc214ddaf54c160649197e1b2aac909d234951f8738ce0401fe87/1974398 for /root/.cache/bazel/_bazel_root/37ff7b0eddea115f60ab9f853788d7f5/execroot/_main/bazel-out/k8-fastbuild/bin/pkg/statistics/handle/autoanalyze/priorityqueue/ingest.recompile1333.x
```

takaidohigasi:
Before explaining this complex implementation, we needed to clarify the prerequisites

current dumpling implementation ignores -r (and -F) option for string key.
```
  -r, --rows uint                         If specified, dumpling will split table into chunks and concurrently dump them to different files to improve efficiency. For TiDB v3.0+, specify this will make dumpling split table with each file one TiDB region(no matter how many rows is).
```

```
[taka-h@tc2-taka-h-test tidb]$ ./bin/dumpling version
Release version: 063fa72687
Git commit hash: 063fa72687159a8c93afd258919f66e0e148895a
Git branch:      master
Build timestamp: 2025-07-21 11:33:28Z
Go version:      go version go1.23.10 linux/amd64

[taka-h@tc2-taka-h-test tidb]$ ./bin/dumpling --tables-list "test.comp_str_case_0" -r 5 -u taka-h -p $MYSQL_PWD -P 3306 -h 10.64.12.24 -o /tmp/dump

[taka-h@tc2-taka-h-test tidb]$ cat /tmp/dump/test.comp_str_case_0.000000000.sql
/*!40014 SET FOREIGN_KEY_CHECKS=0*/;
/*!40101 SET NAMES binary*/;
INSERT INTO `comp_str_case_0` VALUES
('tenant_a','user_001','Alice Johnson','alice@tenant-a.com','2023-01-15 10:30:00'),
('tenant_a','user_002','Bob Smith','bob@tenant-a.com','2023-01-16 11:45:00'),
('tenant_a','user_003','Carol Davis','carol@tenant-a.com','2023-01-17 09:15:00'),
('tenant_b','user_001','Dave Wilson','dave@tenant-b.com','2023-01-18 14:20:00'),
('tenant_b','user_002','Eve Brown','eve@tenant-b.com','2023-01-19 16:30:00'),
('tenant_b','user_003','Frank Miller','frank@tenant-b.com','2023-01-20 08:45:00'),
('tenant_c','user_001','Grace Taylor','grace@tenant-c.com','2023-01-21 13:10:00'),
('tenant_c','user_002','Henry Anderson','henry@tenant-c.com','2023-01-22 12:00:00'),
('tenant_c','user_003','Ivy Thomas','ivy@tenant-c.com','2023-01-23 15:30:00'),
('tenant_d','user_001','Jack Moore','jack@tenant-d.com','2023-01-24 10:15:00');
```

to keep backward compatibility, we need a little complex logic (writer_util.go, writer.go).
But It might be better to align for integer key implementation. how do you think...?

dumpling/export/dump.go
```diff
@@ -349,25 +355,23 @@ func (d *Dumper) startWriters(tctx *tcontext.Context, wg *errgroup.Group, taskCh
...
-		writer.setFinishTableCallBack(func(task Task) {
-			if _, ok := task.(*TaskTableData); ok {
-				IncCounter(d.metrics.finishedTablesCounter)
-				// FIXME: actually finishing the last chunk doesn't means this table is 'finished'.
-				//  We can call this table is 'finished' if all its chunks are finished.
-				//  Comment this log now to avoid ambiguity.
-				// tctx.L().Debug("finished dumping table data",
-				//	zap.String("database", td.Meta.DatabaseName()),
-				//	zap.String("table", td.Meta.TableName()))
-				failpoint.Inject("EnableLogProgress", func() {
```
lance6716, line 381: can you explain why this failpoint is deleted?

dumpling/export/dump.go
```diff
@@ -1707,5 +1765,27 @@ func (d *Dumper) renewSelectTableRegionFuncForLowerTiDB(tctx *tcontext.Context)
...
+
+// extractOrderByColumns extracts column names from ORDER BY clause
+// Input: "ORDER BY `item_id`,`photo_index`"
+// Output: ["`item_id`", "`photo_index`"]
+func extractOrderByColumns(orderByClause string) []string {
+	// Remove "ORDER BY " prefix
+	columnsStr := strings.TrimPrefix(orderByClause, "ORDER BY ")
+
+	// Split by comma and trim spaces
+	columns := strings.Split(columnsStr, ",")
```
lance6716: Should also consider column names contaions `,`. We should respect escaping.

dumpling/export/writer.go
```diff
@@ -225,11 +225,11 @@ func (w *Writer) WriteTableData(meta TableMeta, ir TableDataIR, currentChunk int
 		defer func() {
 			_ = ir.Close()
 		}()
-		return w.tryToWriteTableData(tctx, meta, ir, currentChunk)
+		return w.tryToWriteTableData(tctx, meta, ir, currentChunk, isLastChunk)
 	}, newRebuildConnBackOffer(canRebuildConn(conf.Consistency, conf.TransactionalConsistency)))
 }
 
-func (w *Writer) tryToWriteTableData(tctx *tcontext.Context, meta TableMeta, ir TableDataIR, curChkIdx int) error {
+func (w *Writer) tryToWriteTableData(tctx *tcontext.Context, meta TableMeta, ir TableDataIR, curChkIdx int, isLastChunk bool) error {
```
lance6716: I think we don't need `isLastChunk` argument? As my previous comment, every chunk has its own writer. There's no need to specially handle the last chunk for INSERT prefix in file.

takaidohigasi: it's related to
https://github.com/pingcap/tidb/pull/62172#issuecomment-3100321643 (backward compatibility)

current implementation

  1. When string chunking is enabled (cfg.IsStringChunking):
    - If it's NOT the last chunk: rows end with ,\n to allow concatenation with the next chunk
    - If it IS the last chunk: rows end with ;\n to properly terminate the INSERT statement
  2. When string chunking is disabled:
    - All rows end with ;\n regardless of chunk position

required for `    - If it's NOT the last chunk: rows end with ,\n to allow concatenation with the next chunk
`

I started to think we don't need this backward compatibility, but how do you think?
If you think so, I will fix integration test expectation, and simplify the implementation which do not change some existing function interface and implementation.

takaidohigasi: test file
https://github.com/pingcap/tidb/blob/802bc95db8d530c1f20fdca47ed28e1ea479b46f/dumpling/tests/composite_string_key/run.sh#L24

test input
https://github.com/pingcap/tidb/pull/62172/files#diff-aab75a15391cc6af72dcf00783eaf2b2c019b51bc8719fbe6c96db94bb68005b

expected result
https://github.com/pingcap/tidb/blob/802bc95db8d530c1f20fdca47ed28e1ea479b46f/dumpling/tests/composite_string_key/result/comp_str_case_0.sql

↑current implementation, only 1 INSERT statement got for the same imput.

dumpling/export/dump.go
takaidohigasi: fixed in
https://github.com/pingcap/tidb/pull/62172/commits/1c464d6e3bb53bf503fa6b5088a1454d2138c21e

dumpling/export/dump.go
takaidohigasi: I'll check again. thanks

takaidohigasi:
@lance6716 thanks for helping us

1. documentation is mistakenly merged and it will need to revert. 
https://github.com/pingcap/docs/pull/21489
https://github.com/pingcap/docs/pull/21490

2. would you please tell me your thought on?
https://github.com/pingcap/tidb/pull/62172#discussion_r2244153426

I interrupted work on this PR but resumed again.

takaidohigasi:
https://tiprow.hawkingrei.com/view/gs/pingcapprow/pr-logs/pull/pingcap_tidb/62172/fast_test_tiprow/1953214628251045888

```
$ PWD=/home/prow/go/src/github.com/pingcap/tidb  git fetch https://x-access-token:CENSORED@github.com/pingcap/tidb.git 59322e44814f759d95d3e64e52cceaa322b48ff4 (runtime: 3.984647139s)
From https://github.com/pingcap/tidb
 * branch                59322e44814f759d95d3e64e52cceaa322b48ff4 -> FETCH_HEAD
$ PWD=/home/prow/go/src/github.com/pingcap/tidb GIT_AUTHOR_DATE=1754494555 GIT_COMMITTER_DATE=1754494555 git merge --no-ff 59322e44814f759d95d3e64e52cceaa322b48ff4 (runtime: 8.473327ms)
fatal: refusing to merge unrelated histories
# Error: exit status 128
# Final SHA: 
```

maybe rebase or master merge is required

dumpling/export/writer.go
lance6716: > When string chunking is enabled (cfg.IsStringChunking):
> - If it's NOT the last chunk: rows end with ,\n to allow concatenation with the next chunk
> - If it IS the last chunk: rows end with ;\n to properly terminate the INSERT statement

I think this is not expected. Every chunk is a file contains complete INSERT SQL.

lance6716: I made this simple test (MySQL on 3306)

```
mysql --host 127.0.0.1 -u root test -e 'create table t (c int key)'
for i in {1..100}; do mysql --host 127.0.0.1 -u root -e "insert into test.t values ($i)"; done
tiup dumpling:v8.5.2 -f test.t -r 10 -P 3306
```

You can check every output file. They are a complete INSERT

```
$ head export-2025-08-07T10:50:15+08:00/test.t.*.sql
==> export-2025-08-07T10:50:15+08:00/test.t.000000000.sql <==
/*!40014 SET FOREIGN_KEY_CHECKS=0*/;
/*!40101 SET NAMES binary*/;
INSERT INTO `t` VALUES
(1),
(2),
(3),
(4),
(5),
(6),
(7),

==> export-2025-08-07T10:50:15+08:00/test.t.000000001.sql <==
/*!40014 SET FOREIGN_KEY_CHECKS=0*/;
/*!40101 SET NAMES binary*/;
INSERT INTO `t` VALUES
(11),
(12),
(13),
(14),
(15),
...
```

takaidohigasi:
```
[root@gt-tidb-client-1a work]$ tiup dumpling:v8.5.2 -f test.t -r 10 -P 3306 -h tidb.xfidkq1s0cfw.clusters.tidb-cloud.com -p xxxx -u root

Release version: v8.5.2
Git commit hash: f43a13324440f92209e2a9f04c0bbe9cf763978d
Git branch:      HEAD
Build timestamp: 2025-05-29 03:30:55Z
Go version:      go version go1.23.8 linux/amd64

[2025/09/04 22:35:44.727 +09:00] [INFO] [versions.go:54] ["Welcome to dumpling"] ["Release Version"=v8.5.2] ["Git Commit Hash"=f43a13324440f92209e2a9f04c0bbe9cf763978d] ["Git Branch"=HEAD] ["Build timestamp"="2025-05-29 03:30:55"] ["Go Version"="go version go1.23.8 linux/amd64"]
[2025/09/04 22:35:44.746 +09:00] [INFO] [version.go:457] ["detect server version"] [type=TiDB] [version=8.1.2]
[2025/09/04 22:35:44.847 +09:00] [WARN] [dump.go:1517] ["If the amount of data to dump is large, criteria: (data more than 60GB or dumped time more than 10 minutes)\nyou'd better adjust the tikv_gc_life_time to avoid export failure due to TiDB GC during the dump process.\nBefore dumping: run sql `update mysql.tidb set VARIABLE_VALUE = '720h' where VARIABLE_NAME = 'tikv_gc_life_time';` in tidb.\nAfter dumping: run sql `update mysql.tidb set VARIABLE_VALUE = '10m' where VARIABLE_NAME = 'tikv_gc_life_time';` in tidb.\n"]
[2025/09/04 22:35:44.865 +09:00] [INFO] [dump.go:1585] ["cannot check whether TiDB has TiKV, will apply tidb_snapshot by default. This won't affect dump process"] [error="sql: SELECT COUNT(1) as c FROM MYSQL.TiDB WHERE VARIABLE_NAME='tikv_gc_safe_point': Error 1142 (42000): SELECT command denied to user 'root'@'%' for table 'tidb'"]
[2025/09/04 22:35:44.894 +09:00] [INFO] [dump.go:152] ["begin to run Dump"] [conf="{\"s3\":{\"endpoint\":\"\",\"region\":\"\",\"storage-class\":\"\",\"sse\":\"\",\"sse-kms-key-id\":\"\",\"acl\":\"\",\"access-key\":\"\",\"secret-access-key\":\"\",\"session-token\":\"\",\"provider\":\"\",\"force-path-style\":true,\"use-accelerate-endpoint\":false,\"role-arn\":\"\",\"external-id\":\"\",\"object-lock-enabled\":false},\"gcs\":{\"endpoint\":\"\",\"storage-class\":\"\",\"predefined-acl\":\"\",\"credentials-file\":\"\"},\"azblob\":{\"endpoint\":\"\",\"account-name\":\"\",\"account-key\":\"\",\"access-tier\":\"\",\"sas-token\":\"\",\"encryption-scope\":\"\",\"encryption-key\":\"\"},\"SpecifiedTables\":false,\"AllowCleartextPasswords\":false,\"SortByPk\":true,\"NoViews\":true,\"NoSequences\":true,\"NoHeader\":false,\"NoSchemas\":false,\"NoData\":false,\"CompleteInsert\":false,\"TransactionalConsistency\":true,\"EscapeBackslash\":true,\"DumpEmptyDatabase\":true,\"PosAfterConnect\":false,\"CompressType\":0,\"Host\":\"tidb.xfidkq1s0cfw.clusters.tidb-cloud.com\",\"Port\":3306,\"Threads\":4,\"User\":\"root\",\"Security\":{\"CAPath\":\"\",\"CertPath\":\"\",\"KeyPath\":\"\"},\"LogLevel\":\"info\",\"LogFile\":\"\",\"LogFormat\":\"text\",\"OutputDirPath\":\"./export-2025-09-04T22:35:44+09:00\",\"StatusAddr\":\":8281\",\"Snapshot\":\"460585158530039809\",\"Consistency\":\"snapshot\",\"CsvNullValue\":\"\\\\N\",\"SQL\":\"\",\"CsvSeparator\":\",\",\"CsvDelimiter\":\"\\\"\",\"CsvLineTerminator\":\"\\r\\n\",\"Databases\":[],\"Where\":\"\",\"FileType\":\"sql\",\"ServerInfo\":{\"ServerType\":3,\"ServerVersion\":\"8.1.2\",\"HasTiKV\":true},\"Rows\":10,\"ReadTimeout\":900000000000,\"TiDBMemQuotaQuery\":0,\"FileSize\":0,\"StatementSize\":1000000,\"SessionParams\":{\"tidb_snapshot\":\"460585158530039809\"},\"Tables\":{},\"CollationCompatible\":\"loose\",\"CsvOutputDialect\":0,\"IOTotalBytes\":null,\"Net\":\"\"}"]
[2025/09/04 22:35:45.047 +09:00] [INFO] [writer.go:272] ["no data written in table chunk"] [database=test] [table=t] [chunkIdx=0]
[2025/09/04 22:35:45.048 +09:00] [INFO] [collector.go:264] ["backup success summary"] [total-ranges=4] [ranges-succeed=4] [ranges-failed=0] [total-take=32.866921ms] [total-kv-size=681B] [average-speed=20.72kB/s] [total-rows=100]
[2025/09/04 22:35:45.049 +09:00] [INFO] [main.go:82] ["dump data successfully, dumpling will exit now"]

[root@gt-tidb-client-1a work]$ ls -lat export-2025-09-04T22\:35\:44+09\:00/
total 16
drwxrwxrwx 2 root root 105 Sep  4 22:35 .
-rw-r--r-- 1 root root 146 Sep  4 22:35 metadata
-rw-r--r-- 1 root root 681 Sep  4 22:35 test.t.000000001.sql
-rw-r--r-- 1 root root 225 Sep  4 22:35 test.t-schema.sql
-rw-r--r-- 1 root root 132 Sep  4 22:35 test-schema-create.sql
drwxr-xr-x 4 root root  86 Sep  4 22:35 ..

[root@gt-tidb-client-1a work]$ head -n 15 export-2025-09-04T22\:35\:44+09\:00/test.t.000000001.sql
/*!40014 SET FOREIGN_KEY_CHECKS=0*/;
/*!40101 SET NAMES binary*/;
INSERT INTO `t` VALUES
(1),
(2),
(3),
(4),
(5),
(6),
(7),
(8),
(9),
(10),
(11),
(12),

[root@gt-tidb-client-1a work]$ tail -n 10 export-2025-09-04T22\:35\:44+09\:00/test.t.000000001.sql
(91),
(92),
(93),
(94),
(95),
(96),
(97),
(98),
(99),
(100);
```

I will also confirm for MySQL

takaidohigasi: for MySQL, you were right. sorry, I will fix the expectation.

```
[taka-h@jump2 ~]$ ls -lat export-2025-09-04T22\:59\:27+09\:00/
total 60
drwxrwxrwx   2 taka-h taka-h 4096 Sep  4 22:59 .
-rw-rw-r--   1 taka-h taka-h  180 Sep  4 22:59 metadata
-rw-rw-r--   1 taka-h taka-h  149 Sep  4 22:59 test.t.000000002.sql
-rw-rw-r--   1 taka-h taka-h  149 Sep  4 22:59 test.t.000000006.sql
-rw-rw-r--   1 taka-h taka-h  149 Sep  4 22:59 test.t.000000008.sql
-rw-rw-r--   1 taka-h taka-h  150 Sep  4 22:59 test.t.000000009.sql
-rw-rw-r--   1 taka-h taka-h  149 Sep  4 22:59 test.t.000000004.sql
-rw-rw-r--   1 taka-h taka-h  149 Sep  4 22:59 test.t.000000007.sql
-rw-rw-r--   1 taka-h taka-h  149 Sep  4 22:59 test.t.000000001.sql
-rw-rw-r--   1 taka-h taka-h  149 Sep  4 22:59 test.t.000000005.sql
-rw-rw-r--   1 taka-h taka-h  140 Sep  4 22:59 test.t.000000000.sql
-rw-rw-r--   1 taka-h taka-h  149 Sep  4 22:59 test.t.000000003.sql
-rw-rw-r--   1 taka-h taka-h  193 Sep  4 22:59 test.t-schema.sql
-rw-rw-r--   1 taka-h taka-h  194 Sep  4 22:59 test-schema-create.sql
drwx------. 13 taka-h taka-h 4096 Sep  4 22:59 .
```

takaidohigasi:
back again. 

only integration-test is failing

dumpling/export/dump.go
takaidohigasi: reverted.

lance6716:
> back again.
> 
> only integration-test is failing

Hi, you can remove the "draft" status of this PR to trigger CI

takaidohigasi:
sorry I plan to resume work again!

takaidohigasi:
Thanks for the blunt feedback @lance6716 — you were right. I went back through the diff and removed everything I couldn't trace to a production caller. The cleanup is in 629efbb ("dumpling: drop LLM-noise helpers and flatten chunking API", -700 net lines).

**Removed (unreachable from any dump path):**
- `interpolateStringBoundary` — left over from an earlier interpolation-based design; the streaming cursor path never calls it.
- `buildStringWhereClauses` — superseded by `buildCursorWhereClause` / `buildUpperBoundWhereClause` / `buildBoundedWhereClause`.
- `dumpling/export/composite_string_key_integration_test.go` (205 lines) — every test in it ran against one of the two helpers above.
- `pickupPossibleFieldForStringChunking` / `pickupPossibleFieldsForStringChunking` wrappers.
- `concurrentDumpStringField` (singular) — zero callers.
- `Config.IsStringChunking` — became write-only after the `isLastChunk` removal you requested in the other thread.
- Redundant `totalCount <= chunkSize` guard inside `concurrentDumpStringFields`; the caller already early-returns on `count < conf.Rows`.

**Simplified:**
- `pickupPossibleField` now returns `([]string, bool, error)`. The string-chunking path no longer re-runs `SHOW INDEX FROM` to fetch the full composite key.
- The direct `COUNT(*)` fallback uses `buildWhereCondition` instead of concatenating `conf.Where`.
- Per-iteration boundary-sampling logs downgraded from Info to Debug so large tables don't spam.

Verified locally with an old-vs-new dump on a 500-row composite-string-PK dataset: master emits 1 chunk, this branch emits 10 chunks, row content byte-identical after normalization. There's also a new CI test in 919858e that does the same round-trip through `tidb-lightning` + `sync_diff_inspector`.

takaidohigasi:
once all tests were passed but I need to verify again to pass all. please wait for a while.

takaidohigasi:
once all the CI, and will cope with coderabbit AI reviews.

dumpling/tests/composite_string_key/result/comp_str_case_1.sql
takaidohigasi: Kept out of scope here as noted — the `/*!40101 SET NAMES binary*/;` fixture line is unchanged in this PR. Collation-aware output stays for a follow-up branch.

dumpling/export/string_chunking.go
takaidohigasi: Removed in 46e88ca, and the surrounding `estimatedCount`-consuming helper was later deleted entirely in 629efbb.

dumpling/export/writer_util.go
takaidohigasi: Took the simplification one step further: `Config.IsStringChunking` became write-only after the `isLastChunk` cross-chunk continuation was removed, so in 629efbb the field was deleted entirely. Config is clean again.

dumpling/export/dump.go
takaidohigasi: Reverted in dd6da5670c — the `failpoint.Inject("EnableLogProgress", ...)` in `setFinishTableCallBack` is present in the current tree (line shifted after the master merge, now at dump.go:366).

dumpling/export/dump.go
takaidohigasi: Fixed in 1c464d6 by parsing the ORDER BY column list with a backtick-aware splitter; `TestExtractOrderByColumns` covers comma-in-name cases.

dumpling/export/writer.go
takaidohigasi: You were right. The per-chunk self-contained INSERT is now the behavior — each chunk file has its own `INSERT INTO ... VALUES` prefix and ends with `;\n`, matching the MySQL 8 / `tiup dumpling:v8.5.2` output you demonstrated. Locally verified across MySQL 5.7/8.0/8.4; CI `integration-test` green on current HEAD.

dumpling/export/dump_test.go
```diff
@@ -904,3 +905,179 @@ func TestSetSessionParams(t *testing.T) {
...
+		// Simulate chunk completion after some processing
+		go func() {
+			stats.finished.Add(1)
+		}()
+	}
+
+	// Wait for all chunks to be processed (in real code this would be handled by workers)
+	for stats.finished.Load() < int32(numChunks) {
+		time.Sleep(1 * time.Millisecond)
+	}
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 23e7d56 — replaced with `require.Eventually(t, ..., 5*time.Second, 1*time.Millisecond)` so a stuck goroutine fails the test rather than hanging to the suite timeout.

dumpling/export/dump.go
```diff
@@ -662,7 +672,11 @@ func (d *Dumper) dumpTableData(tctx *tcontext.Context, conn *BaseConn, meta Tabl
 	}
 
 	// Update total rows
-	fieldName, _ := pickupPossibleField(tctx, meta, conn)
+	fields, _, _ := pickupPossibleField(tctx, meta, conn)
+	fieldName := ""
+	if len(fields) > 0 {
+		fieldName = fields[0]
+	}
 	c := estimateCount(tctx, meta.DatabaseName(), meta.TableName(), conn, fieldName, conf)
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 23e7d56 — `dumpTableData` now uses `"*"` for string-leading composite keys so `estimateTotalRowsCounter` matches `concurrentDumpTable`'s estimate, and surfaces the previously-discarded error at Debug.

dumpling/export/dump.go
```diff
@@ -858,6 +908,15 @@ func (d *Dumper) concurrentDumpTable(tctx *tcontext.Context, conn *BaseConn, met
...
 	nullValueCondition := ""
+	chunkStats := newTableChunkStat()
+	d.chunkedTables.Store(meta.ChunkKey(), chunkStats)
+	defer func() {
+		chunkStats.finalized.Store(true)
+		if chunkStats.finished.Load() == chunkStats.sent.Load() {
+			IncCounter(d.metrics.finishedTablesCounter)
+			d.chunkedTables.Delete(meta.ChunkKey())
+		}
+	}()
```
takaidohigasi, line 1031: Fixed in 23e7d56 using `sync.Map.LoadAndDelete` as an atomic claim at all four sites (the `startWriters` finishTask callback, `concurrentDumpTable`, `concurrentDumpTiDBPartitionTables`, and `streamStringChunks`). Only the winner gets `loaded==true` and runs `IncCounter`. The `tableChunkStat` doc comment is rewritten to describe the actual handshake rather than the incorrect ordering argument the previous comment implied. Thanks for catching this.

dumpling/export/ir_impl_test.go
```diff
@@ -128,3 +128,135 @@ func TestChunkRowIter(t *testing.T) {
...
+	mock.ExpectQuery("SELECT tenant_id, user_id, name FROM users WHERE.*").WillReturnRows(expectedRows)
+
+	td := newTableData(compositeKeyQuery, 3, false)
+	require.NotNil(t, td, "Should create table data for composite key")
+
+	// Test that the query supports composite key WHERE clauses
+	require.Contains(t, td.query, "tenant_id", "Query should reference first key column")
+	require.Contains(t, td.query, "user_id", "Query should reference second key column")
+	require.Contains(t, td.query, "OR", "Query should use OR logic for composite keys")
+}
```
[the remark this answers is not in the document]
takaidohigasi: Deleted in 987f030 — the test registered an sqlmock expectation it never consumed and then asserted tautologies over its own query string.

dumpling/export/status.go
```diff
@@ -83,6 +83,14 @@ func (d *Dumper) GetStatus() *DumpStatus {
...
 		}
+	} else {
+		// Debug: Show ongoing chunk progress even when progressReady is false for streaming tables
+		totalChunks := d.metrics.totalChunks.Load()
+		completedChunks := d.metrics.completedChunks.Load()
+		if totalChunks > 0 {
+			progress := float64(completedChunks) / float64(totalChunks)
+			ret.Progress = fmt.Sprintf("%5.2f %% (streaming)", progress*100)
+		}
 	}
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 23e7d56 — streaming branch now clamps progress to `<=1` and the stray `Debug:` prefix is replaced with a proper intent comment.

dumpling/export/writer_serial_test.go
```diff
@@ -369,7 +369,7 @@ func TestSQLDataTypes(t *testing.T) {
 		bf := NewBufferWriter()
 
 		conf := configForWriteSQL(cfg, UnspecifiedSize, UnspecifiedSize)
-		m := newMetrics(conf.PromFactory, conf.Labels)
+			m := newMetrics(conf.PromFactory, conf.Labels)
 		n, err := WriteInsert(tcontext.Background(), conf, tableIR, tableIR, bf, m)
```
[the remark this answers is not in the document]
takaidohigasi: Fixed via `gofmt -w` in e2edf364 — leftover indentation after an earlier `replace_all` deleted the `conf.IsStringChunking = false` lines.

dumpling/tests/composite_string_key_large/run.sh
```diff
@@ -0,0 +1,147 @@
...
+    for e in range(50):
+        eid = f"evt-{e:04d}"
+        # Exercise SQL escaping: single/double quotes and a backslash.
+        # Double single-quotes so the literal is valid MySQL.
+        raw = f"payload for tenant {t} event {e} with 'single' and \"double\" quotes and a backslash \\."
+        payload = raw.replace("'", "''")
+        rows.append(f"('{tenant}','{eid}','{payload}')")
+print("INSERT INTO events (tenant,event_id,payload) VALUES")
+print(",\n".join(rows) + ";")
+PY
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 4e7015b — the python generator now writes to a tempfile and `run_sql_file` loads it, so the test picks up `$DUMPLING_TEST_USER` / port / future auth changes from the harness.

dumpling/tests/composite_string_key_large/run.sh
```diff
@@ -0,0 +1,147 @@
...
+    \( -name "$DB_NAME.events.[0-9]*.sql" -o -name "$DB_NAME.translations.[0-9]*.sql" \)); do
+  if ! grep -q "^INSERT INTO" "$chunk"; then
+    echo "FAIL: $chunk missing INSERT prefix"
+    exit 1
+  fi
+  if ! tail -c 3 "$chunk" | grep -q ";"; then
+    echo "FAIL: $chunk does not end with ';' (continuation-style leaked back in)"
+    exit 1
+  fi
+done
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 4e7015b — switched the chunk-iteration loop to `find ... -print0 | while IFS= read -r -d ''`.

dumpling/tests/composite_string_key_large/run.sh
```diff
@@ -0,0 +1,147 @@
...
+for chunk in $(find "$DUMPLING_OUTPUT_DIR" -maxdepth 1 -type f \
+    \( -name "$DB_NAME.events.[0-9]*.sql" -o -name "$DB_NAME.translations.[0-9]*.sql" \)); do
+  if ! grep -q "^INSERT INTO" "$chunk"; then
+    echo "FAIL: $chunk missing INSERT prefix"
+    exit 1
+  fi
+  if ! tail -c 3 "$chunk" | grep -q ";"; then
+    echo "FAIL: $chunk does not end with ';' (continuation-style leaked back in)"
+    exit 1
+  fi
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 4e7015b — replaced with `last_line=$(tail -n 1 "$chunk"); case "$last_line" in *';') ;; *) FAIL ;; esac` so only a line that actually ends with `;` satisfies the contract.

dumpling/tests/composite_string_key/run.sh
```diff
@@ -0,0 +1,86 @@
...
+# With --rows 5, tables will be split into multiple chunks
+# Each chunk should contain a complete INSERT statement with ~5 rows
+run_dumpling --rows 5
+
+for file_path in "$DUMPLING_BASE_NAME"/data/*; do
+  base_name=$(basename "$file_path")
+  table_name="${base_name%.sql}"
+  
+  # Count the number of chunk files created
+  chunk_count=$(ls -1 "$DUMPLING_OUTPUT_DIR"/composite_string_key.$table_name.*.sql 2>/dev/null | wc -l)
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 4e7015b — `ls ... | wc -l` replaced with `find ... | wc -l`, glob expansions are quoted, and the for-loop over chunks uses `find -print0 | while read -d ''`.

dumpling/tests/composite_string_key/run.sh
```diff
@@ -0,0 +1,86 @@
...
+    # Extract the chunk number from the filename (e.g., composite_string_key.comp_str_case_0.000000000.sql -> comp_str_case_0.000000000.sql)
+    expected_file="${chunk_basename#composite_string_key.}"
+    
+    if [ ! -f "$DUMPLING_BASE_NAME/result/$expected_file" ]; then
+      echo "ERROR: Expected result file $DUMPLING_BASE_NAME/result/$expected_file not found"
+      exit 1
+    fi
+    
+    # Compare the chunk with expected result
+    if ! diff -B -w "$chunk_file" "$DUMPLING_BASE_NAME/result/$expected_file"; then
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 4e7015b — dropped the `-B -w` flags. The committed fixtures were regenerated from fresh dumpling output in 0e2468c0 so plain `diff` passes; whitespace-sensitive regressions now fail loudly.

takaidohigasi:
fixed for the coderabbitai review.

dumpling/export/dump.go
```diff
@@ -858,6 +926,16 @@ func (d *Dumper) concurrentDumpTable(tctx *tcontext.Context, conn *BaseConn, met
...
+	chunkStats := newTableChunkStat()
+	d.chunkedTables.Store(meta.ChunkKey(), chunkStats)
+	defer func() {
+		chunkStats.finalized.Store(true)
+		if chunkStats.finished.Load() == chunkStats.sent.Load() {
+			if _, loaded := d.chunkedTables.LoadAndDelete(meta.ChunkKey()); loaded {
+				IncCounter(d.metrics.finishedTablesCounter)
+			}
+		}
+	}()
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 6f6ec69b — `buildConcatTask` now drops the leaked `chunkedTables` entry after the inner `concurrentDumpTable` returns, so the concat (or fallback) task hits the no-tracking branch in `startWriters` and `finishedTablesCounter` increments exactly once per table. The `handleSubTask`-bumps-`finished` alternative was rejected because it races with the inner defer's `LoadAndDelete` and would double-count.

dumpling/export/dump.go
```diff
@@ -1014,6 +1126,17 @@ func (d *Dumper) concurrentDumpTiDBPartitionTables(tctx *tcontext.Context, conn
...
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 6f6ec69b — extracted a `beginChunkTracking(meta)` helper and applied `defer d.beginChunkTracking(meta)()` to `concurrentDumpTiDBTables`, `concurrentDumpTiDBPartitionTablesWithTableSample`, and `concurrentDumpTiDBPartitionTables` (the inline copy in the latter was replaced with the helper). All three TiDB producer paths now register a single chunkedTables entry per table, so `finishedTablesCounter` increments exactly once.

dumpling/export/dump_test.go
```diff
@@ -904,3 +905,179 @@ func TestSetSessionParams(t *testing.T) {
...
+	stats.finalized.Store(true)
+
+	// Verify cleanup condition
+	if stats.finished.Load() == stats.sent.Load() {
+		d.chunkedTables.Delete(chunkKey)
+	}
+
+	_, exists := d.chunkedTables.Load(chunkKey)
+	require.False(t, exists, "Chunk statistics should be cleaned up after completion")
+}
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 6c0c4951 — both `TestDumperChunkedTablesManagement` and `TestStreamingChunkProgressIntegration` (and the now-unused `mockTableMeta` helper) were removed from `dump_test.go`. They simulated the cleanup condition rather than driving the real producer/consumer callbacks, so a regression in the `LoadAndDelete` claim or the `finalized`/`finished`/`sent` ordering would not have been caught. The real invariants are exercised by the `composite_string_key{,_large}` integration tests.

dumpling/export/dump.go
```diff
@@ -2037,5 +2160,49 @@ func (d *Dumper) renewSelectTableRegionFuncForLowerTiDB(tctx *tcontext.Context)
...
+		}
+	}
+
+	// Add the last column
+	if col := strings.TrimSpace(currentColumn.String()); col != "" {
+		columns = append(columns, col)
+	}
+
+	return columns
 }
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 6c0c4951 — `extractOrderByColumns` now returns `nil` (instead of `[]string{""}`) when the trimmed clause is empty, and the table-driven test row in `string_chunking_test.go` was updated. Also added a comment noting the prefix `"ORDER BY "` is matched case-sensitively with a single space so callers know to normalize.

dumpling/export/ir_impl_test.go
```diff
@@ -128,3 +128,108 @@ func TestChunkRowIter(t *testing.T) {
...
+	chunkIndex := false // Streaming chunking doesn't use traditional chunk index
+
+	td := newTableData(query, selectLen, chunkIndex)
+	require.NotNil(t, td, "Should create table data successfully")
+
+	// Verify the query is stored correctly for streaming chunking
+	require.Contains(t, td.query, "WHERE", "Query should contain WHERE clause for chunking")
+	require.Contains(t, td.query, ">=", "Query should contain lower bound")
+	require.Contains(t, td.query, "<", "Query should contain upper bound")
+}
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 6c0c4951 — both `TestNewTableDataWithImprovedChunking` and `TestTableDataQueryValidation` were removed. They only round-tripped a constructed query string through `newTableData` and asserted substrings of the same string, and `expectValid` was never exercised with `false`. `TestRowIterWithStringKeyProgress` is kept since it does real iterator work.

dumpling/export/string_chunking.go
```diff
@@ -0,0 +1,294 @@
+package export
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 506810284 — added `// Copyright 2025 PingCAP, Inc. Licensed under Apache-2.0.` at the top of `string_chunking.go` (using 2025 since the file was created last year).

dumpling/tests/composite_string_key_large/run.sh
```diff
@@ -0,0 +1,161 @@
...
+        # Exercise SQL escaping: single/double quotes and a backslash.
+        # Double single-quotes so the literal is valid MySQL.
+        raw = f"payload for tenant {t} event {e} with 'single' and \"double\" quotes and a backslash \\."
+        payload = raw.replace("'", "''")
+        rows.append(f"('{tenant}','{eid}','{payload}')")
+print("INSERT INTO events (tenant,event_id,payload) VALUES")
+print(",\n".join(rows) + ";")
+PY
+run_sql_file "$events_sql"
+rm -f "$events_sql"
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 506810284 — the Python here-doc now escapes backslashes before the existing single-quote handling, so the literal `\` survives MySQL's default `sql_mode` and the test actually exercises backslash preservation through the dump/import path.

dumpling/tests/composite_string_key/run.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 506810284 — switched the shebang on both `composite_string_key/run.sh` and `composite_string_key_large/run.sh` to `#!/usr/bin/env bash`. They use `read -r -d ''`, which is a bash extension; under dash the script would silently misparse the loop. Also added a comment near the shebang noting the requirement.

dumpling/export/string_chunking_test.go
```diff
@@ -0,0 +1,160 @@
+package export
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 3cf73d29 — added `// Copyright 2025 PingCAP, Inc. Licensed under Apache-2.0.` at the top of `string_chunking_test.go` to match the rest of the package.

dumpling/export/string_chunking.go
```diff
@@ -0,0 +1,296 @@
...
+						currentBoundary[j] = strconv.FormatFloat(v, 'f', -1, 64)
+					case float32:
+						currentBoundary[j] = strconv.FormatFloat(float64(v), 'f', -1, 32)
+					default:
+						currentBoundary[j] = fmt.Sprintf("%v", v)
+					}
+				}
+			}
+			return nil
+		}, func() {}, sampleQuery)
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 3cf73d29 — the `conn.QuerySQL` reset callback now does `currentBoundary = nil` so a retried boundary-sample query cannot inherit a partial row from a previous attempt. The loop's `len(currentBoundary) == 0` check after the call now reflects the *retried* attempt rather than stale state.

dumpling/export/string_chunking.go
```diff
@@ -0,0 +1,296 @@
...
+			whereClause := buildBoundedWhereClause(orderByColumns, previousBoundary, currentBoundary)
+			fullWhere := buildWhereCondition(conf, whereClause)
+			query := buildSelectQuery(db, tbl, selectField, "", fullWhere, orderByClause)
+			newTask = d.newTaskTableData(meta, newTableData(query, selectLen, false), int(totalChunks), -1)
+		}
+
+		// Send previous buffered chunk (now we know it's not the last)
+		if err := sendBufferedChunk(false); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 3cf73d29 — `TaskTableData.Brief()` in `task.go` now renders `TotalChunks < 0` as `?`, so streaming intermediate chunks log as `(idx/?)` instead of `(idx/-1)`. The writer's `ChunkIndex+1 == TotalChunks` finality check at `writer.go:119` is unaffected (still relies on the integer comparison).

dumpling/export/string_chunking.go
```diff
@@ -0,0 +1,302 @@
...
+						currentBoundary[j] = strconv.Itoa(v)
+					case float64:
+						currentBoundary[j] = strconv.FormatFloat(v, 'f', -1, 64)
+					case float32:
+						currentBoundary[j] = strconv.FormatFloat(float64(v), 'f', -1, 32)
+					default:
+						currentBoundary[j] = fmt.Sprintf("%v", v)
+					}
+				}
+			}
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 8e5eb75c — the row-callback default case now returns `fmt.Errorf("unsupported boundary column type %T for %s in table ...", ...)` instead of silently coercing via `fmt.Sprintf("%v", v)`. Added a comment listing the supported driver-returned types and noting that `parseTime` is intentionally not enabled in `GetDriverConfig`, so a future caller that widens chunking beyond strings (or flips `parseTime`) will trip the error and the boundary loop's existing `return err` (also added in this commit) will surface it.

dumpling/export/string_chunking.go
```diff
@@ -0,0 +1,302 @@
...
+			break
+		}
+
+		if len(currentBoundary) == 0 {
+			tctx.L().Debug("boundary sampling returned no results - reached end of data",
+				zap.String("database", db),
+				zap.String("table", tbl),
+				zap.Int64("chunkIndex", i))
+			break
+		}
```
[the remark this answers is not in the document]
takaidohigasi: Addressed in 8e5eb75c — replaced `break` with `return err`. You're right that swallowing the error fed two silent failure modes: a first-iteration failure dumped the entire table as one un-chunked task via the `else if totalChunks == 0` branch, and a later failure emitted an oversized tail chunk from the last good boundary to +∞ via the `if len(previousBoundary) > 0 && totalChunks > 0` branch. Now the dump fails loudly and the job-level retry/backoff applies. Added a comment explaining the trap so a future reader doesn't undo it.

dumpling/export/sql_util.go
```diff
@@ -0,0 +1,277 @@
...
+
+		if i > 0 {
+			condition.WriteString(")")
+		}
+
+		conditions = append(conditions, condition.String())
+	}
+
+	return strings.Join(conditions, " OR ")
+}
```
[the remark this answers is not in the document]
takaidohigasi: Declining this one. The only callers (`buildCursorWhereClause` and `buildUpperBoundWhereClause` in `string_chunking.go`) build `currentBoundary` / `previousBoundary` from the same `orderByColumns` slice (string_chunking.go:174 and the boundary loop), so the lengths are equal by construction. AGENTS.md asks us not to add validation for scenarios that can't happen at internal boundaries ("Don't add error handling, fallbacks, or validation for scenarios that can't happen. Trust internal code and framework guarantees. Only validate at system boundaries"), so I'd rather not add a defensive guard here. Happy to revisit if a new caller is introduced.

takaidohigasi:
thanks. ready again.

takaidohigasi:
it was ready, but conflicted now, so will resolve it.

takaidohigasi:
will resolve the conflict

PR #62271, opened 2025-07-08
ddl: handle PD error from getting timestamp

### LINKED ISSUE #58871: PD error is not checked in `waitVersionSyncedWithoutMDL`
## Bug Report

here error is not checked https://github.com/pingcap/tidb/blob/b74eb0f906a73215548c1cf2352c41a802ea37f9/pkg/ddl/schema_version.go#L408

If PD goes wrong here, the error is ignored and continue with `ver=0`. Then It uses `0` to create a snapshot and `latestSchemaVersion` will be `0` too and `updateGlobalVersionAndWaitSynced` will not work.

master

joechenrh:

Issue Number: close #58871

Problem Summary:

When we failed to get timestamp from PD, the returned ver is set to zero. This makes `WaitVersionSynced` pass unexpectedly.

https://github.com/tikv/client-go/blob/557a4986e4c43f7809e0305dde9b30d5e433055a/tikv/kv.go#L527-L530

We should check the error returned from client to make caller retry syncing version.

I have checked the repo, and now all places that call this function check the error.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/db_test.go
+++ b/pkg/ddl/db_test.go
@@ -59,7 +59,6 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/tikv/client-go/v2/oracle"
 	"github.com/tikv/client-go/v2/tikv"
-	"golang.org/x/sync/errgroup"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/ddl/db_test.go
```diff
@@ -1352,3 +1353,28 @@ func TestGetAllTableInfos(t *testing.T) {
...
+	eg.Go(func() error {
+		tk.MustExec("alter table t add column b int")
+		return nil
+	})
+
+	eg.Go(func() error {
+		time.Sleep(2 * time.Second)
+		testfailpoint.Disable(t, "github.com/pingcap/tidb/pkg/ddl/mockGetCurrentVersionFailed")
+		return nil
+	})
```
tangenta: It is better not to use sleep in UT, maybe change the failpoint term from `return(true)` to `2*return(true)`

PR #62283, opened 2025-07-08
disttask, importer: show more detailed progress in `SHOW IMPORT JOBS`

joechenrh:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/disttask/importinto/planner.go
+++ b/pkg/disttask/importinto/planner.go
@@ -55,6 +55,8 @@ type LogicalPlan struct {
 	Stmt              string
 	EligibleInstances []*infosync.ServerInfo
 	ChunkMap          map[int32][]importer.Chunk
+
+	summary importer.StepSummary
 }
 
 // GetTaskExtraParams implements the planner.LogicalPlan interface.
@@ -331,6 +333,10 @@ func generateImportSpecs(pCtx planner.PlanCtx, p *LogicalPlan) ([]planner.Pipeli
 			},
 			Plan: p.Plan,
 		}
+		for _, chunk := range chunks {
+			p.summary.RowCnt = max(p.summary.RowCnt, chunk.RowIDMax)
+			p.summary.Bytes += chunk.FileSize
+		}
 		importSpecs = append(importSpecs, importSpec)
 	}
 	return importSpecs, nil
@@ -366,6 +372,10 @@ func generateMergeSortSpecs(planCtx planner.PlanCtx, p *LogicalPlan) ([]planner.
 				zap.Int64("task-id", planCtx.TaskID),
 				zap.String("kv-group", kvGroup))
 			continue
+		}
+		p.summary.Bytes += int64(kvMeta.TotalKVSize)
+		if kvGroup == dataKVGroup {
+			p.summary.RowCnt += int64(kvMeta.TotalKVCnt)
 		}
 		dataFiles := kvMeta.GetDataFiles()
 		nodeCnt := max(1, planCtx.ExecuteNodesCnt)
@@ -421,6 +431,10 @@ func generateWriteIngestSpecs(planCtx planner.PlanCtx, p *LogicalPlan) ([]planne
 
 	specs := make([]planner.PipelineSpec, 0, 16)
 	for kvGroup, kvMeta := range kvMetas {
+		p.summary.Bytes += int64(kvMeta.TotalKVSize)
+		if kvGroup == dataKVGroup {
+			p.summary.RowCnt += int64(kvMeta.TotalKVCnt)
+		}
 		specsForOneSubtask, err3 := splitForOneSubtask(ctx, store, kvGroup, kvMeta, ver.Ver)
 		if err3 != nil {
 			return nil, err3
--- a/pkg/disttask/importinto/scheduler.go
+++ b/pkg/disttask/importinto/scheduler.go
@@ -306,6 +306,10 @@ func (sch *importScheduler) OnNextSubtasksBatch(
 			return nil, err
 		}
 		previousSubtaskMetas[proto.ImportStepEncodeAndSort] = sortAndEncodeMeta
+		// Update timestamp
+		if err = sch.job2Step(ctx, logger, taskMeta, importer.JobStepGlobalSorting); err != nil {
+			return nil, err
+		}
 	case proto.ImportStepWriteAndIngest:
 		failpoint.Inject("failWhenDispatchWriteIngestSubtask", func() {
 			failpoint.Return(nil, errors.New("injected error"))
@@ -336,17 +340,17 @@ func (sch *importScheduler) OnNextSubtasksBatch(
 		failpoint.Inject("failWhenDispatchPostProcessSubtask", func() {
 			failpoint.Return(nil, errors.New("injected error after ImportStepImport"))
 		})
-		// we need get metas where checksum is stored.
-		if err := updateResult(taskHandle, task, taskMeta, sch.GlobalSort); err != nil {
-			return nil, err
-		}
 		step := getStepOfEncode(sch.GlobalSort)
 		metas, err := taskHandle.GetPreviousSubtaskMetas(task.ID, step)
 		if err != nil {
 			return nil, err
 		}
 		previousSubtaskMetas[step] = metas
-		logger.Info("move to post-process step ", zap.Any("result", taskMeta.Result))
+		var importResult importer.Summary
+		if err := json.Unmarshal(taskMeta.TaskResult, &importResult); err != nil {
+			return nil, errors.Trace(err)
+		}
+		logger.Info("move to post-process step ", zap.Any("result", importResult))
 	case proto.StepDone:
 		return nil, nil
 	default:
@@ -372,6 +376,10 @@ func (sch *importScheduler) OnNextSubtasksBatch(
 	}
 	metaBytes, err := physicalPlan.ToSubtaskMetas(planCtx, nextStep)
 	if err != nil {
+		return nil, err
+	}
+
+	if err := updateTaskSummary(taskHandle, task, taskMeta, task.Step, nextStep, logicalPlan); err != nil {
 		return nil, err
 	}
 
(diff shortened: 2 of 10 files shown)


pkg/disttask/importinto/job.go
```diff
@@ -154,50 +158,158 @@ type RuntimeInfo struct {
...
+func formatSecondToTime(sec int64) string {
+	days := sec / (24 * 3600)
+	hours := (sec % (24 * 3600)) / 3600
+	minutes := (sec % 3600) / 60
+	seconds := sec % 60
+	if days > 0 {
+		return fmt.Sprintf("%d d %02d:%02d:%02d", days, hours, minutes, seconds)
+	}
+	return fmt.Sprintf("%02d:%02d:%02d", hours, minutes, seconds)
+}
```
D3Hunter: maybe use this one
```
func TestTimePrint(t *testing.T) {
	fmt.Printf("%s\n", time.Duration(100000*time.Second))
}
```
```
=== RUN   TestTimePrint
27h46m40s
--- PASS: TestTimePrint (0.00s)
```

pkg/disttask/importinto/job.go
```diff
@@ -154,50 +158,155 @@ type RuntimeInfo struct {
...
+	UpdateTime types.Time
+	Processed  int64
+	Total      int64
+}
+
+// Percent returns the progress percentage of the current step.
+func (ri *RuntimeInfo) Percent() string {
+	// Currently, we can't track the progress of post process
+	if ri.Step == proto.ImportStepPostProcess || ri.Step == proto.StepInit {
+		return "N/A"
```
D3Hunter: maybe declare a constant, it appears twice

pkg/disttask/importinto/job.go
```diff
@@ -154,50 +158,155 @@ type RuntimeInfo struct {
 	Status     proto.TaskState
 	ImportRows int64
 	ErrorMsg   string
+
+	Step       proto.Step
+	StartTime  types.Time
+	UpdateTime types.Time
```
D3Hunter, line 164: maybe use `time.Time`, we can use `info.StartTime.GoTime(time.Local)` to get it

pkg/disttask/importinto/job.go
joechenrh: > Durations over 24 hours use D d HH:MM:SS.

According to the design, I think we need `1 d 03:46:40`

D3Hunter: but we still can simplify the code as
```go
func formatSecondToTime(sec int64) string {
	dur := time.Duration(sec) * time.Second
	const timeDay = 24 * time.Hour
	if dur <= timeDay {
		return dur.String()
	}
	days := dur / timeDay
	return fmt.Sprintf("%d d %s", days, dur%timeDay)
}
```

pkg/disttask/importinto/job.go
```diff
@@ -154,50 +158,159 @@ type RuntimeInfo struct {
...
+
+	result := types.NewTime(types.FromGoTime(tr), mysql.TypeDatetime, 0)
+	err = result.ConvertTimeZone(t.Location(), loc)
+	return result, err
 }
 
 // GetRuntimeInfoForJob get the corresponding DXF task runtime info for the job.
-func GetRuntimeInfoForJob(ctx context.Context, jobID int64) (*RuntimeInfo, error) {
+func GetRuntimeInfoForJob(
+	ctx context.Context, sctx sessionctx.Context,
```
D3Hunter: maybe pass the location, we don't depends on the whole session here

pkg/disttask/importinto/job_testkit_test.go
```diff
@@ -205,7 +236,123 @@ func TestGetTaskImportedRows(t *testing.T) {
...
 
 	switchTaskStep(ctx, t, manager, taskID, proto.ImportStepWriteAndIngest)
-	runInfo, err = importinto.GetRuntimeInfoForJob(ctx, 222)
+	runInfo, err = importinto.GetRuntimeInfoForJob(ctx, tk.Session(), 222)
+	require.NoError(t, err)
+	require.EqualValues(t, 300, runInfo.ImportRows)
+	require.Equal(t, "30", runInfo.Percent())
+}
+
+func TestShowImportProgress(t *testing.T) {
```
D3Hunter, line 249: better enable failpoint `MockDisableDistTask` to avoid DXF interfere with the manually switch step in the test

as the task meta is not valid, DXF actually panic on scheduling, so the test can still pass

and please change previous `TestGetTaskImportedRows` too, i forget this part

pkg/executor/importer/import.go
```diff
@@ -307,6 +307,33 @@ type ASTArgs struct {
 	LinesInfo          *ast.LinesClause
 }
 
+// StepSummary records the number of data involved in each step.
+type StepSummary struct {
+	Bytes  int64 `json:"input-bytes,omitempty"`
+	RowCnt int64 `json:"input-rows,omitempty"`
```
D3Hunter, line 314: seems only used in test, and this row count is not accurate during encode step, maybe add a comment

pkg/disttask/importinto/planner.go
```diff
@@ -331,6 +334,10 @@ func generateImportSpecs(pCtx planner.PlanCtx, p *LogicalPlan) ([]planner.Pipeli
 			},
 			Plan: p.Plan,
 		}
+		for _, chunk := range chunks {
+			p.summary.RowCnt = max(p.summary.RowCnt, chunk.RowIDMax)
+			p.summary.Bytes += chunk.FileSize
```
D3Hunter: if we specify `split_file`, the chunk will be part of the file, this size will be larger than expected

use `p.Plan.TotalFileSize` directly

pkg/disttask/importinto/scheduler.go
```diff
@@ -504,16 +512,47 @@ func getStepOfEncode(globalSort bool) proto.Step {
...
+	// Process row count and data size
+	switch nextStep {
+	case proto.ImportStepEncodeAndSort, proto.ImportStepImport:
+		importSummary.EncodeSummary = p.summary
+	case proto.ImportStepMergeSort:
+		importSummary.MergeSummary = p.summary
+	case proto.ImportStepWriteAndIngest:
+		importSummary.IngestSummary = p.summary
+	case proto.ImportStepPostProcess:
+		subtaskSummaries, err := handle.GetPreviousSubtaskSummary(task.ID, currStep)
```
D3Hunter: maybe use `getStepOfEncode`, we might have conflict resolution later, at that time the currStep will be `conflict-resolution`

pkg/disttask/importinto/scheduler.go
```diff
@@ -569,13 +608,22 @@ func (sch *importScheduler) finishJob(ctx context.Context, logger *zap.Logger,
...
+
+	summary := &importer.Summary{}
+	if len(taskMeta.TaskResult) > 0 {
+		if err := json.Unmarshal(taskMeta.TaskResult, summary); err != nil {
+			return errors.Trace(err)
+		}
+	}
+
+	// we have already switch import-mode when switch to post-process step.
+	sch.unregisterTask(ctx, task)
```
D3Hunter: why call it again

pkg/executor/importer/import.go
```diff
@@ -307,6 +307,33 @@ type ASTArgs struct {
...
+
+	RowCnt int64 `json:"row-count,omitempty"`
+}
+
+// MockSummary creates a mock summary for testing.
+func MockSummary(rowCnt int64) *Summary {
+	return &Summary{
+		RowCnt: rowCnt,
+	}
+}
```
D3Hunter: we can move it to test files, all fields are exported

pkg/executor/importer/import.go
```diff
@@ -307,6 +307,33 @@ type ASTArgs struct {
...
+type Summary struct {
+	// EncodeSummary stores the bytes and rows needed to be processed in encode step.
+	// Same for other summaries.
+	EncodeSummary StepSummary `json:"encode-summary,omitempty"`
+
+	MergeSummary StepSummary `json:"merge-summary,omitempty"`
+
+	IngestSummary StepSummary `json:"ingest-summary,omitempty"`
+
+	RowCnt int64 `json:"row-count,omitempty"`
```
D3Hunter:
```suggestion
	// ImportedRows is the number of rows imported into TiKV.
	ImportedRows uint64 `json:"imported-rows,omitempty"`
```
maybe keep original name, so existing job can also show correct result

pkg/executor/show.go
```diff
@@ -2431,52 +2432,96 @@ func (e *ShowExec) fetchShowSessionStates(ctx context.Context) error {
...
-	result.AppendString(2, fullTableName)
-	result.AppendInt64(3, info.TableID)
-	result.AppendString(4, info.Step)
-	result.AppendString(5, info.Status)
-	result.AppendString(6, units.BytesSize(float64(info.SourceFileSize)))
-	if info.Summary != nil {
-		result.AppendUint64(7, info.Summary.ImportedRows)
-	} else if importedRowCount >= 0 {
-		result.AppendUint64(7, uint64(importedRowCount))
+	result.AppendNull(1) // group key, which will be filled in next PR
```
D3Hunter, line 2438: this will make the result in-compatible with previous version, maybe append to tail?

tests/realtikvtest/importintotest4/import_summary_test.go
```diff
@@ -0,0 +1,138 @@
...
+	s.tk.MustExec("create table t (a bigint primary key, b varchar(100), key(b), key(a, b), key(b, a))")
+	// 1 subtask, encoding 10 files using 4 threads.
+	sortStorageURI := fmt.Sprintf("gs://sorted/gs_multi_files?endpoint=%s", gcsEndpoint)
+	importSQL := fmt.Sprintf(`import into t FROM 'gs://global-sort-files/t.*.csv?endpoint=%s'
+		with __force_merge_step, cloud_storage_uri='%s'`, gcsEndpoint, sortStorageURI)
+	rs := s.tk.MustQuery(importSQL).Rows()
+	jobID, err := strconv.Atoi(rs[0][0].(string))
+	require.NoError(s.T(), err)
+
+	time.Sleep(time.Second)
```
D3Hunter: why sleep

pkg/disttask/importinto/job.go
joechenrh: As we still need to convert to `types.Time` when writing to chunk, I think store `types.Time` directly is better.

pkg/disttask/importinto/job_testkit_test.go
joechenrh: `MockDisableDistTask` will make `CreateTask` fail.
Since we are manually inserting rows into subtask tables and query them, I think we don't need to consider DXF scheduling.

pkg/disttask/importinto/job.go
joechenrh: Keep the previous format but use `time.Duration` to simplify the code in https://github.com/pingcap/tidb/pull/62283/commits/ed5c50411fbca277e0cb85bb4b0730746cd2ee8b

pkg/disttask/importinto/job_testkit_test.go
D3Hunter: > MockDisableDistTask will make CreateTask fail.

you can insert a node meta, like this https://github.com/pingcap/tidb/blob/dc876e5c51c63948ca6d47f1b7ffe3c692fa220c/pkg/disttask/framework/taskexecutor/task_executor_testkit_test.go#L103

joechenrh: Cool, it works!

PR #62294, opened 2025-07-08
*: bump client-go to enable async-batch-get

### LINKED ISSUE #62293: Let batch-get use async kv API
## Enhancement

https://github.com/tikv/client-go/issues/1586 introduces the async kv API, we can add a config item to the [`performance`](https://github.com/pingcap/tidb/blob/5c4977b21420dd4c276e6c4fcb368e1625606dc5/pkg/config/config.go#L198) section and enable async-batch-get by default.

zyguan:

Issue Number: close #62293

Problem Summary: ref #62293 & https://github.com/tikv/client-go/issues/1586

Add a config item to the [performance](https://github.com/pingcap/tidb/blob/5c4977b21420dd4c276e6c4fcb368e1625606dc5/pkg/config/config.go#L198) section and enable async-batch-get by default.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/config/config.go
```diff
@@ -748,6 +749,9 @@ type Performance struct {
 
 	// Deprecated: this config will not have any effect
 	ProjectionPushDown bool `toml:"projection-push-down" json:"projection-push-down"`
+
+	// EnableAsyncBatchGet indicates whether to use async API when sending batch-get requests.
+	EnableAsyncBatchGet bool `toml:"enable-async-batch-get" json:"enable-async-batch-get"`
```
cfzjywxk, line 754: Do we need to make it extensible by defining like
```
EnableAsyncKVReq string
```
if the coprcoessor path is likely to be supported in the future?

zyguan: I prefer adding another config item (eg. `EnableAsyncCoprocessor`) for coprocessor later, since their code paths are independent of each other.

cfzjywxk: Is it possible or necessary to support the write path as well in the future? The write requests are handled by the `BatchExecutor` which spawns a goroutine for each request if I remember right.

zyguan: Yes, the batch executor can also use the async API. However the benefits of doing so may not be significant since IMO tikv-side processing is the main contributor to write latency. We need to perform further profiling on write-heavy workloads to evaluate the benefits.

cfzjywxk (approved):

you06:
@yudongusa @BenMeadowcroft this PR introduces a new config and needs your approve, PTAL thanks.

zyguan:
@yudongusa @benmeadowcroft PTAL, thanks.

benmeadowcroft (approved):

PR #62308, opened 2025-07-09
*: use virtual-hosted-style S3 URI for AWS S3

### LINKED ISSUE #62309: Use virtual hosted style access rather than path style access
## Enhancement
For the [force-path-style](https://docs.pingcap.com/tidb/stable/external-storage-uri/#amazon-s3-uri-format), using virtual-hosted–style access is more recommended, as it enables support for several key enhancements offered by cloud service providers. For example: [AWS](https://aws.amazon.com/cn/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/)

fzzf678:

Issue Number: close #62309

Problem Summary:

As recommended, we use virtual-hosted-style for `force-path-style` as default value. But this value still can be modified in the URI as before if necessary

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


br/pkg/storage/parse_test.go
```diff
@@ -416,3 +416,15 @@ func TestParseBackend(t *testing.T) {
 		}, *retBackend3.AzureBlobStorage)
 	}
 }
+
+func TestS3DefaultForceStylePath(t *testing.T) {
+	s, err := ParseBackend(`s3://bucket3/prefix/path`, nil)
+	require.NoError(t, err)
+	require.Equal(t, false, s.GetS3().ForcePathStyle)
```
D3Hunter:
```suggestion
	require.False(t, s.GetS3().ForcePathStyle)
```

br/pkg/storage/parse_test.go
```diff
@@ -416,3 +416,15 @@ func TestParseBackend(t *testing.T) {
...
+	s, err = ParseBackend(`s3://bucket3/prefix/path?force_path_style=false`, nil)
+	require.NoError(t, err)
+	require.Equal(t, false, s.GetS3().ForcePathStyle)
+	s, err = ParseBackend(`s3://bucket3/prefix/path?force_path_style=true`, nil)
+	require.NoError(t, err)
+	require.Equal(t, true, s.GetS3().ForcePathStyle)
```
D3Hunter: do it too

fzzf678: Fixed in https://github.com/pingcap/tidb/pull/62308/commits/2dd57b0e5711afa681ff7175a86de1cb2258527d, PTAL  again

D3Hunter:
lgtm

lance6716:
I remember there's https://github.com/pingcap/tidb/pull/62227 . After we change default value, the providers ("alibaba", ...) changed in that PR can be deleted?

br/pkg/storage/parse.go
```diff
@@ -86,7 +86,7 @@ func parseBackend(u *url.URL, rawURL string, options *BackendOptions) (*backuppb
 		}
 		prefix := strings.Trim(u.Path, "/")
 		s3 := &backuppb.S3{Bucket: u.Host, Prefix: prefix}
-		var s3Options S3BackendOptions = S3BackendOptions{ForcePathStyle: true}
+		var s3Options S3BackendOptions = S3BackendOptions{ForcePathStyle: false}
```
3pointer: If we only consider AWS S3, changing the default value is not an issue.
However, this code is also used with a wide range of S3-compatible services, including third-party providers and self-hosted solutions like MinIO.
Therefore, adjusting the default behavior should be done with caution to avoid breaking compatibility in other environments.

If the change is necessary, I suggest at least providing an additional configuration option to help users of non-virtual-host-style S3 services migrate smoothly.

fzzf678: Do you mean we add a config to control the default value for the non-virtual-host-style users? For these users, when they migrate, they will need to either update this new config option or adjust the existing `force-path-style` setting, both ways need to modify a config.

lance6716: Can you change for known & specific providers? like

https://github.com/pingcap/tidb/blob/a7c824dfe0a2ed9eea89605e50fff4ed95068273/br/pkg/storage/s3.go#L200-L205

For S3, maybe this provider field is empty. We can check domain or other field to know it's AWS S3

lance6716 (approved):

br/pkg/storage/s3.go
```diff
@@ -186,7 +186,7 @@ type S3BackendOptions struct {
 }
 
 // Apply apply s3 options on backuppb.S3.
-func (options *S3BackendOptions) Apply(s3 *backuppb.S3) error {
+func (options *S3BackendOptions) Apply(s3 *backuppb.S3, rawURL string) error {
```
YuJuncen: I will prefer add an extra field say `ForcePathStyle bool` in `S3BackendOptions` and fill it before calling `Apply`. Passing a rawURL somehow breaks the abstraction boundary. Also you have to modify all call sites of `Apply`.

br/pkg/storage/s3.go
```diff
@@ -234,6 +234,24 @@ func (options *S3BackendOptions) Apply(s3 *backuppb.S3) error {
...
+func useVirtualHostStyleForAWSS3(opts *S3BackendOptions, rawURL string) bool {
+	// User has explicitly specified ForcePathStyle, use specified value
+	if rawURL == "" || strings.Contains(rawURL, "force-path-style") || strings.Contains(rawURL, "force_path_style") {
+		return false
+	}
+
+	if opts.Provider == "aws" {
+		return true
+	}
+	if strings.Contains(opts.Endpoint, "amazonaws.com") {
```
YuJuncen: If `opts.Endpoint ` is empty, I think the SDK will set the endpoint to `amazonaws.com`. So you may return `true` when it is empty IMO. Also would you extract this string literal to a constant?

br/pkg/storage/s3.go
```diff
@@ -234,6 +234,24 @@ func (options *S3BackendOptions) Apply(s3 *backuppb.S3) error {
 	return nil
 }
 
+func useVirtualHostStyleForAWSS3(opts *S3BackendOptions, rawURL string) bool {
+	// User has explicitly specified ForcePathStyle, use specified value
+	if rawURL == "" || strings.Contains(rawURL, "force-path-style") || strings.Contains(rawURL, "force_path_style") {
```
YuJuncen: Perhaps you need to return `true` when `force-path-style=true`?  Not sure how can `TestS3DefaultForceStylePath` passes... Anyway it is better to not directly handle the raw URL but an extra configuration in the `S3BackendOptions`.

YuJuncen (commented):
rest lgtm

br/pkg/storage/s3.go
YuJuncen: Oh, I see this modification was made for back-compatibility. Perhaps you may add a new field like `UserForcedPathStyle`(well...), I think that can still be a little bit better than directly passing the `rawURl`...

br/pkg/storage/s3.go
fzzf678: If user has explicitly specified ForcePathStyle, `ExtractQueryParameters(u, &s3Options)` will set the specified value, we don't need hand it., so we return false.

br/pkg/storage/s3.go
fzzf678: If user has explicitly specified ForcePathStyle, `ExtractQueryParameters(u, &s3Options)` will set the specified value, we don't need hand it., so we can return false here.

br/pkg/storage/s3.go
fzzf678: I have implement this check in another way, could you please help take a look again https://github.com/pingcap/tidb/pull/62308/commits/32a2b9c90818a46962435486eac1b96f349953be

YuJuncen (approved):

Issue #62069, opened 2025-06-29
Deadlock encountered during statistics collection


0xPoe:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

Start the cluster and run ANALYZE STATISTICS commands immediately after the cluster has started.

### 2. What did you expect to see? (Required)
No deadlock encountered during statistics collection.

### 3. What did you see instead (Required)
```log
2025-06-26 18:15:01 (UTC+02:00)TiKV master-tikv-0.master-tikv-peer.schrddl-sql-tps-7893364-1-595.svc.cluster.local:20160[future.rs:148] ["starting working thread"] [worker=deadlock-detector] [thread_id=1]
2025-06-26 18:15:01 (UTC+02:00)TiKV master-tikv-0.master-tikv-peer.schrddl-sql-tps-7893364-1-595.svc.cluster.local:20160[deadlock.rs:761] ["became the leader of deadlock detector!"] [self_id=1] [thread_id=131]
[...]

PR #62327, opened 2025-07-09
statistics: prevent deadlock in SaveAnalyzeResultToStorage by adding a fake table ID for concurrent updates

0xPoe:

Issue Number: close https://github.com/pingcap/tidb/issues/62069

ref https://github.com/pingcap/tidb/issues/62565

Problem Summary:

We encountered a stable deadlock issue during the analyze process, where the transaction ultimately failed and the error was propagated to the top layer. From the deadlock records in information_schema.deadlocks, it’s confirmed that this is a non-retryable deadlock. The root cause is that the point get operation involves two separate lock phases: one to lock the index and another to lock the row. If another transaction locks both index and row in a single batch point get, it can lead to a deadlock that is not retryable because the lock keys are not within the same lockKeys call.

The sequence looks like this:

```log
txn1: lockKeys on point get (index)

txn2: lockKeys on batch point get (index & row) — waits for txn1

txn1: lockKeys on point get (row) — deadlock occurs here and it's not retryable
```

This fix is to force the point get to lock both index and row in a single batch get operation, by adding a dummy key (like -1988) to ensure a one-round lock acquisition.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] Manual test
	- [x] I tested it in TCMS. 
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
修复收集统计信息时可能遇到的死锁问题
Fix a potential deadlock issue that may occur when collecting statistics
```

### DIFF
--- a/pkg/statistics/handle/storage/save.go
+++ b/pkg/statistics/handle/storage/save.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/pingcap/errors"
@@ -148,8 +149,8 @@ func SaveAnalyzeResultToStorage(sctx sessionctx.Context,
 	// txn1: lockKeys on point get (index lock)
 	// txn2: lockKeys on batch point get (row lock) — waits for txn1 for index lock
 	// txn1: lockKeys on point get (row lock) — deadlock occurs here and it's not retryable
-	fakeTableID := -1988
-	tableIDStrs := []string{fmt.Sprintf("%d", fakeTableID), fmt.Sprintf("%d", tableID)}
+	fakeID := int64(-1988)
+	tableIDStrs := []string{strconv.FormatInt(fakeID, 10), strconv.FormatInt(tableID, 10)}
 	rs, err = util.Exec(sctx, "select snapshot, count, modify_count from mysql.stats_meta where table_id in (%?) for update", tableIDStrs)
 	if err != nil {
 		return 0, err


0xPoe (commented):

- [x] Unit tests added

- [x] All tests pass

- [x] Bazel files updated

- [x] PR size is reasonable


/cc @terry1purcell @time-and-fate

terry1purcell (approved):
/lgtm

pkg/statistics/handle/storage/save.go
```diff
@@ -129,7 +129,28 @@ func SaveAnalyzeResultToStorage(sctx sessionctx.Context,
...
+	// Meanwhile, another transaction doing a batch point get (e.g. during stats delta dump) acquires both locks
+	// together but in the opposite order (row then index). This creates a circular wait condition where:
+	// 1. Transaction 1 holds index lock and waits for row lock
+	// 2. Transaction 2 holds row lock and waits for index lock
+	// Since the locks are acquired in separate lockKeys calls, TiDB cannot detect and retry this deadlock.
+	//
+	// Deadlock Sequence:
+	// txn1: lockKeys on point get (index lock)
+	// txn2: lockKeys on batch point get (row lock) — waits for txn1 for index lock
+	// txn1: lockKeys on point get (row lock) — deadlock occurs here and it's not retryable
```
0xPoe, line 151: @you06 Could you please help check the accuracy of the comment here? Thank you!

you06: LGTM

0xPoe:
/unhold
 
I have tested this PR on TCMS a couple of times. It worked well. I think it is ready to go.

Issue #62093, opened 2025-06-30
planner: underestimate selectivity for ranges within a bucket


ichen032:
## Enhancement
When we are querying for a range within a bucket, the optimizer estimates based on an assumption of even distribution, so may underestimate the number of rows that qualify if we have skewed data.

PR #62363, opened 2025-07-10
planner: use addedRows for out of range and add skew risk ratio | tidb-test=pr/2566

### LINKED ISSUE #55199: Planner: Range predicate estimates at beginning/end of range negatively impacted by sampling
Sampling results in not all rows being seen - which can create gaps in data for ANALYZE.

Following example shows 1 million row table, with values from 1 to 1million - for all columns. Estimate is zero when searching beyond histogram range:

```
tidb> explain select * from t1 where a > 999989 and b > 999974;
+-------------------------------+---------+-----------+---------------------------+---------------------------------------+
| id                            | estRows | task      | access object             | operator info                         |
+-------------------------------+---------+-----------+---------------------------+---------------------------------------+
| IndexLookUp_11                | 0.00    | root      |                           |                                       |
| ├─Selection_10(Build)         | 0.00    | cop[tikv] |                           | gt(test.t1.b, 999974)                 |
| │ └─IndexRangeScan_8          | 0.00    | cop[tikv] | table:t1, index:iab(a, b) | range:(999989,+inf], keep order:false |
| └─TableRowIDScan_9(Probe)     | 0.00    | cop[tikv] | table:t1                  | keep order:false                      |
+-------------------------------+---------+-----------+---------------------------+---------------------------------------+
```

```
tidb> select count(*) from t1 where a > 999989 and b > 999974;
+----------+
| count(*) |
+----------+
|       11 |
+----------+
```

To recreate problem:
```
CREATE TABLE `t1` (
  `a` int DEFAULT NULL,
  `b` int DEFAULT NULL,
  `c` int DEFAULT NULL,  
  KEY `ia` (`a`),
  UNIQUE KEY `iab` (`a`, `b`)
);

Insert into t1 values (1, 1, 1);
Insert into t1 select a + 1, b + 1, c + 1 from t1;
Insert into t1 select a + 2, b + 2, c + 2 from t1;
Insert into t1 select a + 4, b + 4, c + 4 from t1;
Insert into t1 select a + 2, b + 2, c + 2 from t1 where a > 6;

Insert into t1 select a + 10, b + 10, c + 10 from t1;
Insert into t1 select a + 20, b + 20, c + 20 from t1;
Insert into t1 select a + 40, b + 40, c + 40 from t1;
Insert into t1 select a + 20, b + 20, c + 20 from t1 where a > 60;

Insert into t1 select a + 100, b + 100, c + 100 from t1;
Insert into t1 select a + 200, b + 200, c + 200 from t1;
Insert into t1 select a + 400, b + 400, c + 400 from t1;
Insert into t1 select a + 200, b + 200, c + 200 from t1 where a > 600;

insert into t1 select a + 1000, b + 1000, c + 1000 from t1;
insert into t1 select a + 2000, b + 2000, c + 2000 from t1;
insert into t1 select a + 4000, b + 4000, c + 4000 from t1;
insert into t1 select a + 2000, b + 2000, c + 2000 from t1 where a > 6000;

insert into t1 select a + 10000, b + 10000, c + 10000 from t1;
insert into t1 select a + 20000, b + 20000, c + 20000 from t1;
insert into t1 select a + 40000, b + 40000, c + 40000 from t1;
insert into t1 select a + 20000, b + 20000, c + 20000 from t1 where a > 60000;

insert into t1 select a + 100000, b + 100000, c + 100000 from t1;
insert into t1 select a + 200000, b + 200000, c + 200000 from t1;
insert into t1 select a + 400000, b + 400000, c + 400000 from t1;
insert into t1 select a + 200000, b + 200000, c + 200000 from t1 where a > 600000;
```

ichen032:

Issue Number: ref #62093 close #55199 

Problem Summary:

Range predicates outside of the histogram bucket range face a similar risk to ranges within a bucket, because we do not have collected stats we can rely on for the cardinality estimation. So, this change uses the tidb_opt_risk_range_skew_ratio variable to allow users or external processes to adjust how much skew is considered in out of range estimations. This also adjusts and simplifies the out of range estimation function to rely more on addedRows (the difference between realTimeRowCount and the histogram count) than histogram row count itself.

Previously, when estimating a predicate outside of our histogram bucket range, we would calculate a percentage, TotalPercent, and multiply it with histogram row count. In the case that addedRows was significant, we would multiply TotalPercent and addedRows and take the larger of the two estimations. Finally, we would ensure our estimate didn't go over ModifyCount. 
Part of the concern with using histogram row count is that one-sided ranges, such as a > 0, could easily have a TotalPercent = 50%, leading to large out of range estimations. It seems more reasonable for out of range cardinality to be correlated with addedRows rather than histogram row count, no matter the scenario. So, to simplify the logic, the estimation is changed to be TotalPercent * addedRows * 0.5 by default (the 0.5 accounts for the assumption that only half of newly added rows are added outside our histogram range). 

To account for underestimation due to skew, since it is possible that all of our newly added rows are within our queried range, we can use the tidb_opt_risk_range_skew_ratio variable. Based on the value of this variable, we account for more or less skew in our estimate. For a default value of 0, there is no change to the original estimate, and for a value of 1, our out of range estimation = worst case skew = addedRows.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/cardinality/selectivity_test.go
+++ b/pkg/planner/cardinality/selectivity_test.go
@@ -1042,14 +1042,12 @@ func TestIssue39593(t *testing.T) {
 	count, _, err := cardinality.GetRowCountByIndexRanges(sctx.GetPlanCtx(), &statsTbl.HistColl, idxID, getRanges(vals, vals))
 	require.NoError(t, err)
 	// estimated row count without any changes, use range to reduce test flakiness
-	require.Less(t, count, float64(463))
-	require.Greater(t, count, float64(462))
+	require.InDelta(t, float64(462.6), count, float64(1))
 	statsTbl.RealtimeCount *= 10
 	count, _, err = cardinality.GetRowCountByIndexRanges(sctx.GetPlanCtx(), &statsTbl.HistColl, idxID, getRanges(vals, vals))
 	require.NoError(t, err)
 	// estimated row count after mock modify on the table, use range to reduce test flakiness
-	require.Less(t, count, float64(3703))
-	require.Greater(t, count, float64(3702))
+	require.InDelta(t, float64(3702.6), count, float64(1))
 }
 
 func TestIndexJoinInnerRowCountUpperBound(t *testing.T) {
--- a/pkg/statistics/histogram.go
+++ b/pkg/statistics/histogram.go
@@ -49,7 +49,7 @@ import (
 	"go.uber.org/zap"
 )
 
-var (
+const (
 	outOfRangeBetweenRate float64 = 100
 )
 
@@ -966,7 +966,7 @@ func (hg *Histogram) OutOfRangeRowCount(
 		return 0
 	}
 
-	// oneValue assumes "one value qualifes", and is used as either an Upper & lower bound.
+	// oneValue assumes "one value qualifes", and is used as a lower bound.
 	oneValue := float64(0)
 	if histNDV > 0 {
 		oneValue = hg.NotNullCount() / max(float64(histNDV), outOfRangeBetweenRate) // avoid inaccurate selectivity caused by small NDV


terry1purcell (approved):
/lgtm

pkg/planner/cardinality/selectivity_test.go
```diff
@@ -526,13 +527,13 @@ func TestSelectivity(t *testing.T) {
...
 			exprs:                    "b > 1",
-			selectivity:              0.9629629629629629,
+			selectivity:              0.9729629629629629,
 			selectivityAfterIncrease: 1,
 		},
 		{
 			exprs:                    "a > 1 and b < 2 and c > 3 and d < 4 and e > 5",
 			selectivity:              0.001851851851851852,
-			selectivityAfterIncrease: 0.21916666666666668,
+			selectivityAfterIncrease: 0.125425,
```
mjonss, line 536: Just curious, what would the actual selectivity, based on current state/rows be for this test? Feel free to ignore :)

pkg/planner/cardinality/selectivity_test.go
```diff
@@ -1040,13 +1041,15 @@ func TestIssue39593(t *testing.T) {
 	vals := []int64{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20}
 	count, _, err := cardinality.GetRowCountByIndexRanges(sctx.GetPlanCtx(), &statsTbl.HistColl, idxID, getRanges(vals, vals))
 	require.NoError(t, err)
-	// estimated row count without any changes
-	require.Equal(t, float64(360), count)
+	// estimated row count without any changes, use range to reduce test flakiness
```
mjonss, line 1044: You could also use [require.InDelta](https://pkg.go.dev/github.com/stretchr/testify/require#InDelta)

pkg/planner/cardinality/testdata/cardinality_suite_out.json
```diff
@@ -198,40 +198,40 @@
...
+          "└─Selection 407.20 cop[tikv]  gt(test.t.a, 800), lt(test.t.a, 1000)",
           "  └─TableFullScan 2000.00 cop[tikv] table:t keep order:false"
         ]
       },
       {
         "SQL": "explain format = 'brief' select * from t where a > 900 and a < 1000",
         "Result": [
-          "TableReader 458.19 root  data:Selection",
-          "└─Selection 458.19 cop[tikv]  gt(test.t.a, 900), lt(test.t.a, 1000)",
+          "TableReader 76.37 root  data:Selection",
```
mjonss, line 209: I was a bit curious here and checked with EXPLAIN ANALYZE and the actual number of rows was 0, so for this specific test the PR goes in the right direction :)

pkg/statistics/histogram.go
```diff
@@ -962,12 +966,23 @@ func (hg *Histogram) OutOfRangeRowCount(
 		return 0
 	}
 
-	// If there are no modifications to the table, return 0 - since all of this logic is
-	// redundant if we get to the end and return the min - which includes zero,
-	// TODO: The execution here is if we are out of range due to sampling of the histograms - which
-	// may miss the lowest/highest values - and we are out of range without any modifications.
+	// oneValue assumes "one value qualifes", and is used as either an Upper & lower bound.
```
mjonss: "either an Upper **or** lower value"? Should it be 'Either .. or ..' ? Or 'Both ... and ...' ?

pkg/statistics/histogram.go
```diff
@@ -962,12 +966,23 @@ func (hg *Histogram) OutOfRangeRowCount(
...
+	oneValue := float64(0)
+	if histNDV > 0 {
+		oneValue = hg.NotNullCount() / max(float64(histNDV), outOfRangeBetweenRate) // avoid inaccurate selectivity caused by small NDV
+	}
+
+	// We may have missed the true lowest/highest values due to sampling - and we are out of
+	// range without any modifications. So return oneValue to avoid underestimation.
 	if modifyCount == 0 {
```
mjonss, line 977: Just curious, there is a small time frame where rows has been committed (by other TiDB nodes) and modifyCount is still 0 right? (feel free to ignore)

tests/integrationtest/r/imdbload.result
```diff
@@ -304,12 +304,12 @@ CE_trace
 
 explain select * from movie_companies where company_type_id > 2;
 id	estRows	task	access object	operator info
-IndexLookUp_11	1.00	root		
-├─IndexRangeScan_9(Build)	1.00	cop[tikv]	table:movie_companies, index:movie_companies_idx_ctypeid(company_type_id)	range:(2,+inf], keep order:false
-└─TableRowIDScan_10(Probe)	1.00	cop[tikv]	table:movie_companies	keep order:false
+IndexLookUp_11	49582.96	root		
+├─IndexRangeScan_9(Build)	49582.96	cop[tikv]	table:movie_companies, index:movie_companies_idx_ctypeid(company_type_id)	range:(2,+inf], keep order:false
```
mjonss, line 308: This was a big change in result! And the test says "# Actual row count: 0". Is this still acceptable? Is the test correct?

tests/integrationtest/r/planner/core/plan_cost_ver2.result
```diff
@@ -27,24 +27,24 @@ set @@tidb_index_lookup_concurrency=1;
 set @@tidb_cost_model_version=2;
 explain format='verbose' select /*+ use_index(t, primary) */ a from t where a<1;
 id	estRows	estCost	task	access object	operator info
-TableReader_7	1.00	290.67	root		data:TableRangeScan_6
-└─TableRangeScan_6	1.00	227.31	cop[tikv]	table:t	range:[-inf,1), keep order:false
+TableReader_7	2.00	581.34	root		data:TableRangeScan_6
+└─TableRangeScan_6	2.00	454.62	cop[tikv]	table:t	range:[-inf,1), keep order:false
```
mjonss, line 31: Here is the ActRows 1, so this PR goes from correct estimation of (1) to less good (2), is that OK?

tests/integrationtest/r/planner/core/plan_cost_ver2.result
```diff
@@ -27,24 +27,24 @@ set @@tidb_index_lookup_concurrency=1;
...
-IndexReader_7	1.00	226.16	root		index:IndexRangeScan_6
-└─IndexRangeScan_6	1.00	162.80	cop[tikv]	table:t, index:b(b)	range:[-inf,1), keep order:false
+IndexReader_7	2.00	452.32	root		index:IndexRangeScan_6
+└─IndexRangeScan_6	2.00	325.60	cop[tikv]	table:t, index:b(b)	range:[-inf,1), keep order:false
 explain format='verbose' select /*+ use_index(t, b) */ b from t where b<10;
 id	estRows	estCost	task	access object	operator info
-IndexReader_7	10.00	2261.60	root		index:IndexRangeScan_6
-└─IndexRangeScan_6	10.00	1628.00	cop[tikv]	table:t, index:b(b)	range:[-inf,10), keep order:false
+IndexReader_7	11.00	2487.76	root		index:IndexRangeScan_6
+└─IndexRangeScan_6	11.00	1790.80	cop[tikv]	table:t, index:b(b)	range:[-inf,10), keep order:false
```
mjonss, line 47: Same here, the estRows is changed from 10 to 11, while the actRows is 10. Can you explain why this PR changes this, and if it is inline with what the PR should do? (I'm mostly curious)

mjonss (commented):
In general it looks good, but I do have some curious questions about some of the impact of the PR

pkg/planner/cardinality/testdata/cardinality_suite_out.json
mjonss: Hmm, this is probably based only on statistics, so no real underlying data...

pkg/statistics/histogram.go
```diff
@@ -49,6 +49,10 @@ import (
 	"go.uber.org/zap"
 )
 
+var (
```
mjonss: Should this be const?
```suggestion
const (
```

pkg/planner/cardinality/selectivity_test.go
ichen032: Unfortunately for this test, there is no actual selectivity to compare our results to because they simulated the result of rows being inserted/modified by manually increasing realTimeRowCount and modifyCount, but didn't make any actual changes to the table. So, depending on how these new rows would've been distributed and what types of modifications would've been made, the actual selectivity for this situation could vary.

pkg/planner/cardinality/selectivity_test.go
ichen032: got it!

pkg/statistics/histogram.go
ichen032: This was copied from the previous code and I missed updating it, it is now only used as a lower bound. Thanks for catching this!

tests/integrationtest/r/imdbload.result
ichen032: I believe this is still acceptable, this situation arose because the new changes try to always return row count for one value qualifying out of range to reduce the risk from sampling (there may be higher or lower values in our data that we missed due to sampling). I've added an upper bound so this oneValue estimate never goes over 1% of the entire histogram row count. This example has a table with NDV of 2 and row count of 5 million, so that's why the row count still increased by quite a bit.

tests/integrationtest/r/planner/core/plan_cost_ver2.result
ichen032: This is expected and originates from the same situation in the comment above, where due to the outOfRange now always returning at least one value qualifying, it assumes one more row qualifies than before.

tests/integrationtest/r/planner/core/plan_cost_ver2.result
ichen032: This is inline with what the PR should do because even if modifyCount = 0, the new OutOfRange function still returns row count with the assumption that one value qualifies (total row count / NDV), in order to account for the risk that we didn't see the true bounds due to sampling. So, very likely outOfRange returns 1 row, causing the cardinality estimation to be 1 greater than the actual count.

pkg/statistics/histogram.go
ichen032: Yes, I will change that.

pkg/statistics/histogram.go
terry1purcell: Yes - there is a time delay to when the update to stats_meta is triggered such that modifyCount and realtimeRowCount are updated. The larger the number of tables in the cluster, this can also extend the time between updates.

mjonss (approved):
LGTM, and my curious questions has all been answered, thanks for helping me understand!

PR #62387, opened 2025-07-14
tables: force to lock the touched index in DML when DDL merging temp index

### LINKED ISSUE #62337: txn: support secondary index key write conflict check for pessimistic transactions considering DDL fast add index
## Enhancement
In the original transaction model design, there was an implicit assumption that any modification to a secondary index key would always be accompanied by a corresponding modification to the row key/row value. This assumption holds true for user transactions. However, with the introduction of *fast add index*, a new scenario has emerged in internal DDL transactions: during one of the phases of the *add index merge* state, the DDL process needs to perform internal transactions that write **only secondary index** keys based on results from a temporary index.

By default, user pessimistic transactions do not perform write conflict checks on secondary index keys during the prewrite phase. As a result, when such DDL internal transactions and user DML transactions run concurrently and conflict, the write conflict cannot be properly detected and ordered. This may lead to inconsistencies between data and index after the *add index* operation completes—posing a correctness issue in certain scenarios.

Possible solutions include:

1. **Redesign the pessimistic transaction model** so that all secondary index keys undergo write conflict checks during the prewrite phase. However, this would heavily impact the existing pessimistic transaction model and potentially introduce performance regressions.

2. **Introduce write conflict checks on secondary index keys only during the `reorg-merge` phase of DDL add index**, for user pessimistic transactions involving the corresponding table. While this approach couples internal DDL state with transaction behavior, it minimizes performance impact and contains the change scope.

lcwangchao:

Issue Number: close #62337

Problem Summary:

see descriptions in issue. We choose the second way to fix it.

When the index is in state `BackfillStateReadyToMerge` or `BackfillStateMerging`, we also lock the touched non-unique indices to ensure the correctness, even if the DDL does not lock the row.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/table/tables/index.go
```diff
@@ -499,6 +514,17 @@ func (c *index) Delete(ctx table.MutateContext, txn kv.Transaction, indexedValue
...
 
+// mayDDLMergingTempIndex checks whether the DDL worker may be merging the temporary index to the normal index.
+// In most times, if an index is not unique, its primary record is assumed to be mutated and locked.
+// The only exception is when the DDL worker is merging the temporary index in fast reorging,
+// the DDL txn will not lock the primary record to reduce unnecessary conflicts.
+// At this time, the index record should be locked in force
+// to make sure the serialization between the DDL and DML transactions.
+func (c *index) mayDDLMergingTempIndex() bool {
+	return c.idxInfo.BackfillState == model.BackfillStateReadyToMerge ||
+		c.idxInfo.BackfillState == model.BackfillStateMerging
```
lcwangchao, line 525: @tangenta Please to review this method. It means when ddl worker enters the state `BackfillStateReadyToMerge` , all the DML should have seen the `BackfillStateReadyToMerge` or the previous state

tangenta (approved):

pkg/table/tables/index.go
tangenta: lgtm

pkg/session/test/tidb_test.go
```diff
@@ -71,5 +72,12 @@ func TestKeysNeedLock(t *testing.T) {
 		need = session.KeyNeedToLock(test.key, test.val, flag)
 		require.True(t, flag.HasPresumeKeyNotExists())
 		require.True(t, need)
+
+		if bytes.Equal(test.key, nonUniqueIndexKey) && bytes.Equal(test.val, nonUniqueVal) {
+			flag = kv.ApplyFlagsOps(kv.KeyFlags(0), kv.SetNeedLocked)
+			require.True(t, flag.HasNeedLocked())
+			need = session.KeyNeedToLock(test.key, test.val, flag)
+			require.True(t, need)
```
D3Hunter:
```suggestion
			require.True(t, session.KeyNeedToLock(test.key, test.val, flag))
```

pkg/table/tables/index.go
cfzjywxk: @lcwangchao Better to add comments about the constraint ensured by DDL or the MDL mechanism

PR #62406, opened 2025-07-14
fix: add mutex to prevent concurrent btree.Clone() calls in infoschema v2

### LINKED ISSUE #62391: btree.BTreeG.clone() maybe called concurrently
## Bug Report

In the comments of btree.BTree.Clone(), it is stated that btree.BTree.Clone() method cannot be called concurrently.

https://github.com/google/btree/blob/aeba20f7a1e1315badec4eca4fdc9f754f5f880a/btree_generic.go#L613

But I noticed that in the tidb code, using btree.BTree.clone() does not add a lock.
I am concerned that btree.BTree.Clone() may be called concurrently and result in some strange memory leaks.

https://github.com/pingcap/tidb/blob/3580f5196ef1c12d5d7b2b1c2bb11f8112c33b0f/pkg/infoschema/infoschema_v2.go#L91

Clone should not be called concurrently

code did not add a lock when using btree.BTree.Clone()。

https://asktug.com/t/topic/1043747

Image

v8.5.1

abrarr21:

Issue Number: close #62391 

Problem Summary:
`btree.BTree.Clone()` is not safe to call concurrently, as mentioned in the upstream BTree library’s documentation. However, in `btreeSet()` (used in infoschema v2), `Clone()` was being called without synchronization, which could lead to race conditions or memory corruption under concurrent writes.

A `sync.Mutex` named `cloneMu` has been introduced to wrap the `Clone()` call within `btreeSet()`. This ensures that `Clone()` is only executed by one goroutine at a time, as required by the BTree implementation.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/infoschema/infoschema_v2.go
+++ b/pkg/infoschema/infoschema_v2.go
@@ -82,24 +82,32 @@ func (si *schemaItem) Name() string {
 	return si.dbInfo.Name.L
 }
 
-// btree.Clone() is not safe to call from multiple goroutines at once,
-// as stated in the BTree source code. We use a mutex to ensure safety.
-var cloneMu sync.Mutex
-
-// btreeSet updates the btree.
+// btreeSet safely updates the btree by cloning and inserting the new item.
+// It avoids concurrent Clone() calls using atomic.Swap and retries with backoff.
 // Concurrent write is supported, but should be avoided as much as possible.
 func btreeSet[T any](ptr *atomic.Pointer[btree.BTreeG[T]], item T) {
-	succ := false
-	for !succ {
-		var t = ptr.Load()
-		cloneMu.Lock()
-		t2 := t.Clone()
-		cloneMu.Unlock()
-		t2.ReplaceOrInsert(item)
-		succ = ptr.CompareAndSwap(t, t2)
-		if !succ {
+	backoff := time.Microsecond
+
+	for {
+		// Atomically take ownership of the tree for cloning
+		old := ptr.Swap(nil)
+		if old == nil {
+			// Another goroutine is likely cloning; back off and retry
 			logutil.BgLogger().Info("infoschema v2 btree concurrently multiple writes detected, this should be rare")
+			time.Sleep(backoff)
+			if backoff < time.Millisecond {
+				backoff *= 2 // exponential backoff to reduce contention
+			}
+			continue
 		}
+
+		// Clone and modify the tree
+		newTree := old.Clone()
+		newTree.ReplaceOrInsert(item)
+
+		// Publish the updated tree atomically
+		ptr.Store(newTree)
+		return
 	}
 }
 
--- a/pkg/infoschema/infoschema_v2.go
+++ b/pkg/infoschema/infoschema_v2.go
@@ -82,13 +82,19 @@ func (si *schemaItem) Name() string {
 	return si.dbInfo.Name.L
 }
 
+// btree.Clone() is not safe to call from multiple goroutines at once,
+// as stated in the BTree source code. We use a mutex to ensure safety.
+var cloneMu sync.Mutex
+
 // btreeSet updates the btree.
 // Concurrent write is supported, but should be avoided as much as possible.
 func btreeSet[T any](ptr *atomic.Pointer[btree.BTreeG[T]], item T) {
 	succ := false
 	for !succ {
 		var t = ptr.Load()
+		cloneMu.Lock()
 		t2 := t.Clone()
+		cloneMu.Unlock()
 		t2.ReplaceOrInsert(item)
 		succ = ptr.CompareAndSwap(t, t2)
 		if !succ {
--- a/pkg/infoschema/issyncer/syncer.go
+++ b/pkg/infoschema/issyncer/syncer.go
@@ -61,6 +61,7 @@ func init() {
 // Syncer is the main structure for syncing the info schema.
 type Syncer struct {
 	m           syncutil.Mutex
+	btreeMu     sync.Mutex // Added to guard infoSchemaV2 btree mutations.
 	store       kv.Storage
 	schemaLease time.Duration
 	// Note: If you no longer need the session, you must call Destroy to release it.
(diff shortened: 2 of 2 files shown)


pkg/infoschema/infoschema_v2.go
```diff
@@ -82,13 +82,19 @@ func (si *schemaItem) Name() string {
 	return si.dbInfo.Name.L
 }
 
+// btree.Clone() is not safe to call from multiple goroutines at once,
+// as stated in the BTree source code. We use a mutex to ensure safety.
+var cloneMu sync.Mutex
```
lance6716: seems this global lock will hurt performance

pkg/infoschema/infoschema_v2.go
```diff
@@ -82,13 +82,19 @@ func (si *schemaItem) Name() string {
...
+
 // btreeSet updates the btree.
 // Concurrent write is supported, but should be avoided as much as possible.
 func btreeSet[T any](ptr *atomic.Pointer[btree.BTreeG[T]], item T) {
 	succ := false
 	for !succ {
 		var t = ptr.Load()
```
lance6716: I think first we should verify the issue really exists. if so, we can use `Swap` or other atomic actions to limit only one goroutine can clone it.

lance6716 (commented):
/cc @tiancaiamao

pkg/infoschema/infoschema_v2.go
yao-2110: https://github.com/pingcap/tidb/issues/62391#issuecomment-3078934678

Your concerns about performance are very reasonable, but the risk of data race does exist here.

yao-2110: https://github.com/pingcap/tidb/issues/62391#issuecomment-3079002273

Your concerns about performance are very reasonable, but the risk of data race does exist here.

abrarr21: Thanks for the feedback!

I'm currently working on updating the fix to remove the global mutex and instead use an atomic-based approach as suggested — specifically using `atomic.Swap(nil)` to ensure only one goroutine performs `Clone()` at a time. I'm also reviewing edge cases and adding a small backoff strategy to reduce contention under load.

Let me know if this direction sounds good — I’ll update the PR shortly with the improved version.

lance6716: Thanks for your update. I'll take a look later.

lance6716:
Hi @abrarr21 CI found there's a panic

```
[2025/07/17 03:38:58.412 +00:00] [ERROR] [wait_group_wrapper.go:178] ["panic in the wait group"] [recover="runtime error: invalid memory address or nil pointer dereference"] [stack="github.com/pingcap/tidb/pkg/util.(*WaitGroupWrapper).RunWithLog.func1.1\n\tpkg/util/wait_group_wrapper.go:178\nruntime.gopanic\n\tGOROOT/src/runtime/panic.go:791\ngithub.com/pingcap/tidb/pkg/executor.(*Compiler).Compile.func1\n\tpkg/executor/compiler.go:59\nruntime.gopanic\n\tGOROOT/src/runtime/panic.go:791\nruntime.panicmem\n\tGOROOT/src/runtime/panic.go:262\nruntime.sigpanic\n\tGOROOT/src/runtime/signal_unix.go:917\ngithub.com/google/btree.(*BTreeG[...]).DescendLessOrEqual\n\texternal/com_github_google_btree/btree_generic.go:794\ngithub.com/pingcap/tidb/pkg/infoschema.(*infoschemaV2).TableByName\n\tpkg/infoschema/infoschema_v2.go:1037\ngithub.com/pingcap/tidb/pkg/infoschema.(*SessionExtendedInfoSchema).TableByName\n\tpkg/infoschema/infoschema.go:843\ngithub.com/pingcap/tidb/pkg/planner/core.(*preprocessor).tableByName\n\tpkg/planner/core/preprocess.go:540\ngithub.com/pingcap/tidb/pkg/planner/core.(*preprocessor).handleTableName\n\tpkg/planner/core/preprocess.go:1759\ngithub.com/pingcap/tidb/pkg/planner/core.(*preprocessor).Leave\n\tpkg/planner/core/preprocess.go:647\ngithub.com/pingcap/tidb/pkg/parser/ast.(*TableName).Accept\n\t...
```

pkg/infoschema/infoschema_v2.go
```diff
@@ -82,18 +82,32 @@ func (si *schemaItem) Name() string {
...
-		var t = ptr.Load()
-		t2 := t.Clone()
-		t2.ReplaceOrInsert(item)
-		succ = ptr.CompareAndSwap(t, t2)
-		if !succ {
+	backoff := time.Microsecond
+
+	for {
+		// Atomically take ownership of the tree for cloning
+		old := ptr.Swap(nil)
```
lance6716, line 93: I think the panic is caused by concurrent read access after swapping it to nil.

BTW, please check the original issue really exists. Our using pattern may not call this function concurrently.

pkg/infoschema/infoschema_v2.go
```diff
@@ -82,18 +82,32 @@ func (si *schemaItem) Name() string {
...
+		if old == nil {
+			// Another goroutine is likely cloning; back off and retry
 			logutil.BgLogger().Info("infoschema v2 btree concurrently multiple writes detected, this should be rare")
```
tiancaiamao, line 96: There should be only one writer and multiple readers, and this is the perfect use pattern for that btree library.
The only writer to infoschema v2's related data struct is the domain schema reloader (now renamed to pkg/infoschema/issyncer) ... 

If we do find this kind of logs, we should fix on the caller side, here is the callee side.
Although we can do some defensive programming, it would hide the real problem.

abrarr21:
I looked into the panic and added a mutex (s.btreeMu) inside the `Reload` method of Syncer to ensure serialized access. This ensures btreeSet doesn't get invoked concurrently, avoiding access after `ptr.Swap(nil)`.

I understand that the expected pattern is one writer (the reloader) and multiple readers — which should ideally be safe — but added this guard to be cautious, especially since this is the only writer path and a single log line (multiple writes detected) could indicate rare overlaps. So rather than masking the issue in the callee, this change confines the fix to the caller (where the write happens) without interfering with the reader path.

PR #62443, opened 2025-07-16
domain: skip loading privilege if the notification is from current tidb  | tidb-test=release-8.5-20250606-v8.5.2

CbcWestwolf:

Issue Number: ref #61706

Problem Summary:

If a DCL like `GRANT` or `REVOKE` is called on tidb-0, tidb-0 will load the whole privilege cache twice:
* one by `Update()` in `NotifyUpdatePrivilege()`
* another by `Update()` in `LoadPrivilegeLoop()`

If the size of the privilege cache is quite large, the cost of `Update()` can not be ignored

* When `NotifyUpdatePrivilege`, write the current tidb server id into the value of `privilegeKey`
* When watching an event in `LoadPrivilegeLoop`, skip `Update()` if the event is from the tidb itself

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

#### test 1

1. Add a log in the entry at `func (h *Handle) Update()`
2. Bootstrap 2 tidb clusters with multiple tidb separately
3. Do DCL like `create user`, `grant` or `revoke`, see the number of log for `func (h *Handle) Update()`
Result:
* With the PR, the number of invocations of `func (h *Handle) Update()` among all tidb-server are the same
* Without this PR, the tidb-server receiving DCL invokes `func (h *Handle) Update()` twice as many times as other tidb

#### test 2 

1. bootstrap a cluster with one tidb, initialize 130k column privilege records
5. execute `GRANT` per second, observe the CPU framegraph and metric
**Before**
image
image

**After**
image
image

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/domain/domain.go
+++ b/pkg/domain/domain.go
@@ -1868,7 +1868,7 @@ func (do *Domain) LoadPrivilegeLoop(sctx sessionctx.Context) error {
 			resp         clientv3.WatchResponse
 		)
 		for {
-			ok := true
+			var ok bool
 			select {
 			case <-do.exit:
 				return


ystaticy:
Should we add more UT?

pkg/domain/domain.go
```diff
@@ -1881,6 +1884,21 @@ func (do *Domain) LoadPrivilegeLoop(sctx sessionctx.Context) error {
...
+					val := string(event.Kv.Value)
+					if do.serverID != 0 && val != strconv.FormatUint(do.serverID, 10) {
+						canSkip = false
+						break
+					}
+				}
+			}
+			if canSkip {
+				continue
+			}
```
tiancaiamao: I think it's better to make `canSkip` default to false and check resp to make it to true.
That's more fault-tolerant when event.Kv is not set or the values unmarshal get error. @CbcWestwolf

pkg/domain/domain.go
```diff
@@ -2834,7 +2852,7 @@ func (do *Domain) NotifyUpdatePrivilege() error {
 	// password using a special TiDB instance and want the new password to take effect.
 	if do.etcdClient != nil {
 		row := do.etcdClient.KV
-		_, err := row.Put(context.Background(), privilegeKey, "")
+		_, err := row.Put(context.Background(), privilegeKey, strconv.FormatUint(do.serverID, 10))
```
tiancaiamao: Is this the same encoding way as the master tidb?
We need to make sure this change do not conflict with master branch.

CbcWestwolf:
> Should we add more UT?

It is hard to add UT for such case. Let me check again.

pkg/domain/domain.go
CbcWestwolf: [`9b85a79` (#62443)](https://github.com/pingcap/tidb/pull/62443/commits/9b85a794f6cad7fdd05b127237de8fd4e1745598)

pkg/domain/domain.go
CbcWestwolf: [`9b85a79` (#62443)](https://github.com/pingcap/tidb/pull/62443/commits/9b85a794f6cad7fdd05b127237de8fd4e1745598)

CbcWestwolf:
@ystaticy It is hard to add such a UT. I just added another manual test. Please see the detail in the description.

pkg/domain/domain.go
```diff
@@ -1862,25 +1863,50 @@ func (do *Domain) LoadPrivilegeLoop(sctx sessionctx.Context) error {
...
+			resp  clientv3.WatchResponse
+		)
 		for {
 			ok := true
 			select {
 			case <-do.exit:
 				return
-			case _, ok = <-watchCh:
+			case resp, ok = <-watchCh:
 			case <-time.After(duration):
```
wjhuang2016: Seems we don't use duration anymore

CbcWestwolf: The duration still takes effect if no DCL occurs within 10 minutes, so we cannot remove this line.

wjhuang2016 (approved):

PR #62449, opened 2025-07-16
ddl: move auto id handling from submitter to executor

### LINKED ISSUE #60804: Rebase auto increment not executed after second restore
## Bug Report
In the restore process, if the first restore fails but leaves partial table structures in the cluster, and the second restore uses OnExistIgnore, those pre-existing tables will be skipped.

As a result, row_id rebase is not executed for these tables, leading to incorrect auto-increment behavior.

The root cause is that batchCreateTables is not atomic.
https://github.com/pingcap/tidb/blob/287f27ea0e5b7be6d75bd4c6fc9982f9d80e4cf7/pkg/ddl/table.go#L231
Some table could become public while the whole `Create Tables` job still on going.

1. doing a restore with batch tables.
2. random kill restore process, try to make createTables partial succeed.

All tables in the createTables batch should either be created successfully together or fail entirely — there should be no partial success.
After the restore process exited, some tables from the batch were still present, indicating a partial creation.

v7.5.5

joechenrh:

Issue Number: close #60804

Problem Summary:

Previously, the procedure of (batch) creating table is as follow:
- submitter submit the job and wait it finished
- ddl owner execute the job and mark table as public
- submitter pre-scatter the table and update the auto id.

But if the submitter quit before waiting job finished, the post process will be skipped. So the creation itself is **not atomic**.

Now ddl owner takes the responsibility to update auto id and submitter only needs to pre-scatter the table.

https://github.com/pingcap/tidb/blob/1f24032d4e641c38fbd0e72b036bf5f679b809b0/pkg/ddl/job_submitter.go#L224-L227

Because merged ddl job only holds the session variables from the first job for now, which means all jobs share the same variables, we don't apply pre-scatter in ddl owner.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/create_table.go
+++ b/pkg/ddl/create_table.go
@@ -145,9 +145,9 @@ func createTable(jobCtx *jobContext, job *model.Job, r autoid.Requirement, args 
 			return tbInfo, errors.Wrapf(err, "failed to notify PD the placement rules")
 		}
 
-		// Updating auto id metakv is done in a separate txn.
-		// It's ok as these data are binded with table ID, and we won't use see these table IDs
-		// until info schema version is updated.
+		// Updating auto id meta kv is done in a separate txn.
+		// It's ok as these data are bind with table ID, and we won't use these
+		// table IDs until info schema version is updated.
 		if err := handleAutoIncID(r, job, tbInfo); err != nil {
 			return tbInfo, errors.Trace(err)
 		}
@@ -291,7 +291,6 @@ func (w *worker) createTableWithForeignKeys(jobCtx *jobContext, job *model.Job, 
 }
 
 func (w *worker) onCreateTables(jobCtx *jobContext, job *model.Job) (int64, error) {
-	failpoint.InjectCall("beforeCreateTables")
 	var ver int64
 
 	args, err := model.GetBatchCreateTableArgs(job)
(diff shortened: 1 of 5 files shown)


pkg/ddl/create_table.go
```diff
@@ -145,12 +145,59 @@ func createTable(jobCtx *jobContext, job *model.Job, args *model.CreateTableArgs
 			return tbInfo, errors.Wrapf(err, "failed to notify PD the placement rules")
 		}
 
+		// Updating auto id metakv is done in a separate txn.
+		// It's ok as these data are binded with table ID, and we won't use see these table IDs
+		// until info schema version is updated.
```
D3Hunter:
```suggestion
		// Updating auto id meta kv is done in a separate txn.
		// It's ok as these data are bind with table ID, and we won't use these
		// table IDs until info schema version is updated.
```

pkg/ddl/create_table.go
```diff
@@ -238,6 +291,7 @@ func (w *worker) createTableWithForeignKeys(jobCtx *jobContext, job *model.Job,
 }
 
 func (w *worker) onCreateTables(jobCtx *jobContext, job *model.Job) (int64, error) {
+	failpoint.InjectCall("beforeCreateTables")
```
D3Hunter: not used?

joechenrh: forgot to delete

PR #62490, opened 2025-07-18
fix: close issue #62461 return empty array instead of null for json

### LINKED ISSUE #62461: `CHECK INDEX` statement meets runtime error
## Bug Report

`CHECK INDEX` statement meets runtime error. It can be reproduced on the docker image `pingcap/tidb:nightly`.

```sql
CREATE TABLE t (c1 INT PRIMARY KEY, c2 JSON, INDEX i ((CAST(c2->'$.array' AS UNSIGNED ARRAY))));
INSERT INTO t (c1, c2) VALUES (1, '{"name": "John", "age": 30}');
ADMIN CHECK INDEX t i (500, 600);
```

No runtime error.

```sql
mysql> ADMIN CHECK INDEX t i (500, 600);
ERROR 1105 (HY000): runtime error: index out of range [3] with length 3
```

Logs:
```
[2025/07/16 14:12:00.800 +00:00] [INFO] [conn.go:1190] ["command dispatched failed"] [conn=2097160] [session_alias=] [connInfo="id:2097160, addr:172.17.0.1:36354 status:10, collation:utf8mb4_0900_ai_ci, user:root"] [command=Query] [status="inTxn:0, autocommit:1"] 
[sql="ADMIN CHECK INDEX t i (500, 600)"] [txn_mode=PESSIMISTIC] [timestamp=459453266866733056]
runtime error: index out of range [3] with length 3
github.com/pingcap/errors.Trace
        /root/go/pkg/mod/github.com/pingcap/errors@v0.11.5-0.20250523034308-74f78ae071ee/juju_adaptor.go:15
github.com/pingcap/tidb/pkg/util.GetRecoverError
        /workspace/source/tidb/pkg/util/util.go:312
github.com/pingcap/tidb/pkg/executor/internal/exec.Next.func1
        /workspace/source/tidb/pkg/executor/internal/exec/executor.go:444
runtime.gopanic
        /root/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.11.linux-amd64/src/runtime/panic.go:791
runtime.goPanicIndex
        /root/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.11.linux-amd64/src/runtime/panic.go:115
encoding/binary.littleEndian.Uint32
        /root/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.11.linux-amd64/src/encoding/binary/binary.go:87
github.com/pingcap/tidb/pkg/util/chunk.(*Codec).decodeColumn
        /workspace/source/tidb/pkg/util/chunk/codec.go:111
github.com/pingcap/tidb/pkg/util/chunk.(*Codec).DecodeToChunk
        /workspace/source/tidb/pkg/util/chunk/codec.go:98
github.com/pingcap/tidb/pkg/util/chunk.(*Decoder).Reset
        /workspace/source/tidb/pkg/util/chunk/codec.go:262
github.com/pingcap/tidb/pkg/distsql.(*selectResult).readFromChunk
        /workspace/source/tidb/pkg/distsql/select_result.go:493
github.com/pingcap/tidb/pkg/distsql.(*selectResult).Next
        /workspace/source/tidb/pkg/distsql/select_result.go:436
github.com/pingcap/tidb/pkg/executor.(*CheckIndexRangeExec).Next
        /workspace/source/tidb/pkg/executor/admin.go:72
github.com/pingcap/tidb/pkg/executor/internal/exec.Next
        /workspace/source/tidb/pkg/executor/internal/exec/executor.go:460
github.com/pingcap/tidb/pkg/executor.(*ExecStmt).next
        /workspace/source/tidb/pkg/executor/adapter.go:1322
github.com/pingcap/tidb/pkg/executor.(*recordSet).Next
        /workspace/source/tidb/pkg/executor/adapter.go:173
github.com/pingcap/tidb/pkg/server/internal/resultset.(*tidbResultSet).Next
        /workspace/source/tidb/pkg/server/internal/resultset/resultset.go:72
github.com/pingcap/tidb/pkg/server.(*clientConn).writeChunks
        /workspace/source/tidb/pkg/server/conn.go:2339
github.com/pingcap/tidb/pkg/server.(*clientConn).writeResultSet
        /workspace/source/tidb/pkg/server/conn.go:2282
github.com/pingcap/tidb/pkg/server.(*clientConn).handleStmt
        /workspace/source/tidb/pkg/server/conn.go:2085
github.com/pingcap/tidb/pkg/server.(*clientConn).handleQuery
        /workspace/source/tidb/pkg/server/conn.go:1794
github.com/pingcap/tidb/pkg/server.(*clientConn).dispatch
        /workspace/source/tidb/pkg/server/conn.go:1384
github.com/pingcap/tidb/pkg/server.(*clientConn).Run
        /workspace/source/tidb/pkg/server/conn.go:1153
github.com/pingcap/tidb/pkg/server.(*Server).onConn
        /workspace/source/tidb/pkg/server/server.go:740
runtime.goexit
        /root/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.11.linux-amd64/src/runtime/asm_amd64.s:1700
```

```
Release Version: v9.0.0-beta.2.pre-118-g6ec4810
Edition: Community
Git Commit Hash: 6ec48101cae75529682a0f9fa8c16cbc42f246d3
Git Branch: HEAD
UTC Build Time: 2025-07-16 11:00:02
GoVersion: go1.23.11
Race Enabled: false
Store: unistore
Kernel Type: Classic
```

yihong0618:
This patch close issue #62461 

the root cause is not from  runtime error: index out of range [3] with length 3

it is because json array here should not be null 
in

```go
func (c *Codec) decodeColumn(buffer []byte, col *Column, ordinal int) (remained []byte) {
	// Todo(Shenghui Wu): Optimize all data is null.
	// decode length.
	col.length = int(binary.LittleEndian.Uint32(buffer))
	buffer = buffer[4:]

	// decode nullCount.
	nullCount := int(binary.LittleEndian.Uint32(buffer))
	buffer = buffer[4:]
```

binary.LittleEndian.Uint32(buffer)  change to the wrong type buffer only 3 bytes not 4 bytes

so this patch correct the no json patch as empty array here correct it.

and maybe need to add guard for 	`buffer = buffer[4:]` but seems need to disscuss
so this patch do not contain it.

```go
func (c *Codec) decodeColumn(buffer []byte, col *Column, ordinal int) (remained []byte) {
	// Todo(Shenghui Wu): Optimize all data is null.
	// decode length.
	col.length = int(binary.LittleEndian.Uint32(buffer))
	buffer = buffer[4:]

	// decode nullCount.
	nullCount := int(binary.LittleEndian.Uint32(buffer))
	buffer = buffer[4:]

	// decode nullBitmap.
	if nullCount > 0 {
		numNullBitmapBytes := (col.length + 7) / 8
		col.nullBitmap = buffer[:numNullBitmapBytes:numNullBitmapBytes]
		buffer = buffer[numNullBitmapBytes:]
	} else {
		c.setAllNotNull(col)
	}

	// decode offsets.
	numFixedBytes := getFixedLen(c.colTypes[ordinal])
	numDataBytes := int64(numFixedBytes * col.length)
	if numFixedBytes == -1 {
		numOffsetBytes := (col.length + 1) * 8
		col.offsets = bytesToI64Slice(buffer[:numOffsetBytes:numOffsetBytes])
		buffer = buffer[numOffsetBytes:]
		numDataBytes = col.offsets[col.length]
	} else if cap(col.elemBuf) < numFixedBytes {
		col.elemBuf = make([]byte, numFixedBytes)
	}

	// decode data.
	col.data = buffer[:numDataBytes:numDataBytes]
	// The column reference the data of the grpc response, the memory of the grpc message cannot be GCed if we reuse
	// this column. Thus, we set `avoidReusing` to true.
	col.avoidReusing = true
	return buffer[numDataBytes:]
}
```

cc @lance6716 @hawkingrei

Issue Number: close #62461

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/expression/builtin_cast.go
+++ b/pkg/expression/builtin_cast.go
@@ -664,8 +664,15 @@ func newFakeSctx() *stmtctx.StatementContext {
 
 func (b *castJSONAsArrayFunctionSig) evalJSON(ctx EvalContext, row chunk.Row) (res types.BinaryJSON, isNull bool, err error) {
 	val, isNull, err := b.args[0].EvalJSON(ctx, row)
-	if isNull || err != nil {
-		return res, isNull, err
+	if err != nil {
+		return res, false, err
+	}
+
+	// When JSON path does not exist, return empty array instead of null
+	// More to check issue: #62461
+	if isNull {
+		// Return empty array when input is null (e.g., JSON path not found)
+		return types.CreateBinaryJSON([]any{}), false, nil
 	}
 
 	if val.TypeCode == types.JSONTypeCodeObject {
--- a/pkg/expression/builtin_cast_test.go
+++ b/pkg/expression/builtin_cast_test.go
@@ -1762,6 +1762,80 @@ func TestCastArrayFunc(t *testing.T) {
 	}
 }
 
+func TestCastJSONAsArrayWithNullInput(t *testing.T) {
+	ctx := createContext(t)
+
+	// Test case for issue #62461: When JSON path does not exist (returns null),
+	// CAST should return empty array instead of null
+	tp := types.NewFieldTypeBuilder().SetType(mysql.TypeLonglong).SetCharset(charset.CharsetBin).SetCollate(charset.CollationBin).SetArray(true).BuildP()
+
+	// Test 1: null input should return empty array
+	nullConstant := &Constant{
+		Value:   types.NewDatum(nil),
+		RetType: types.NewFieldType(mysql.TypeJSON),
+	}
+
+	baseFunc, err := newBaseBuiltinFunc(ctx, "", []Expression{nullConstant}, tp)
+	require.NoError(t, err)
+
+	castFunc := &castJSONAsArrayFunctionSig{
+		baseBuiltinFunc: baseFunc,
+	}
+	castFunc.tp = tp
+
+	result, isNull, err := castFunc.evalJSON(ctx, chunk.Row{})
+	require.NoError(t, err)
+	require.False(t, isNull, "Result should not be null")
+	require.Equal(t, types.JSONTypeCodeArray, result.TypeCode)
+	require.Equal(t, 0, result.GetElemCount(), "Should return empty array")
+
+	expectedEmptyArray := types.CreateBinaryJSON([]any{})
+	cmp := types.CompareBinaryJSON(expectedEmptyArray, result)
+	require.Equal(t, 0, cmp, "Result should be an empty array")
+
+	// Test 2: non-null JSON input should work normally
+	jsonInput := types.CreateBinaryJSON([]any{int64(1), int64(2), int64(3)})
+	nonNullConstant := &Constant{
+		Value:   types.NewDatum(jsonInput),
+		RetType: types.NewFieldType(mysql.TypeJSON),
+	}
+
+	baseFunc2, err := newBaseBuiltinFunc(ctx, "", []Expression{nonNullConstant}, tp)
+	require.NoError(t, err)
+
+	castFunc2 := &castJSONAsArrayFunctionSig{
+		baseBuiltinFunc: baseFunc2,
+	}
+	castFunc2.tp = tp
+
+	result2, isNull2, err := castFunc2.evalJSON(ctx, chunk.Row{})
+	require.NoError(t, err)
+	require.False(t, isNull2, "Result should not be null for valid input")
+	require.Equal(t, types.JSONTypeCodeArray, result2.TypeCode)
+	require.Equal(t, 3, result2.GetElemCount(), "Should return array with 3 elements")
+
+	// Test 3: single non-array JSON value should be wrapped in array
+	jsonSingle := types.CreateBinaryJSON(int64(42))
+	singleConstant := &Constant{
+		Value:   types.NewDatum(jsonSingle),
+		RetType: types.NewFieldType(mysql.TypeJSON),
+	}
+
+	baseFunc3, err := newBaseBuiltinFunc(ctx, "", []Expression{singleConstant}, tp)
+	require.NoError(t, err)
+
+	castFunc3 := &castJSONAsArrayFunctionSig{
+		baseBuiltinFunc: baseFunc3,
+	}
+	castFunc3.tp = tp
+
+	result3, isNull3, err := castFunc3.evalJSON(ctx, chunk.Row{})
+	require.NoError(t, err)
+	require.False(t, isNull3, "Result should not be null for valid input")
+	require.Equal(t, types.JSONTypeCodeArray, result3.TypeCode)
+	require.Equal(t, 1, result3.GetElemCount(), "Should return array with 1 element")
+}
+
 func TestCastAsCharFieldType(t *testing.T) {
 	type testCase struct {
 		input      *Constant


yihong0618:
old test is different will check

joechenrh:
TiDB document doesn't describe the behavior when the path does not exist
https://docs.pingcap.com/tidb/stable/json-functions-search/#json_extract
But MySQL does
https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html#function_json-extract
> Returns NULL if any argument is NULL or no paths locate a value in the document

Not sure if we should change this behavior

yihong0618:
> TiDB document doesn't describe the behavior when the path does not exist https://docs.pingcap.com/tidb/stable/json-functions-search/#json_extract But MySQL does https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html#function_json-extract
> 
> > Returns NULL if any argument is NULL or no paths locate a value in the document
> 
> Not sure if we should change this behavior

will try to figure out if there's other way to fix this

yihong0618:
> TiDB document doesn't describe the behavior when the path does not exist https://docs.pingcap.com/tidb/stable/json-functions-search/#json_extract But MySQL does https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html#function_json-extract
> 
> > Returns NULL if any argument is NULL or no paths locate a value in the document
> 
> Not sure if we should change this behavior

new commit maybe the real root cause, also add guard  code

joechenrh:
The problem seems not related to null
```SQL
mysql> CREATE TABLE t (c1 INT PRIMARY KEY, c2 JSON, INDEX i ((CAST(c2->'$.array' AS UNSIGNED ARRAY))));

mysql> INSERT INTO t (c1, c2) VALUES (11, '{"name": "John", "array": 30}');

mysql> select c2->'$.array' from t;
+---------------+
| c2->'$.array' |
+---------------+
| 30            |
+---------------+
1 row in set (0.01 sec)

mysql> ADMIN CHECK INDEX t i (0, 22);
ERROR 1105 (HY000): runtime error: index out of range [3] with length 3
```

yihong0618:
> The problem seems not related to null
> 
> ```sql
> mysql> CREATE TABLE t (c1 INT PRIMARY KEY, c2 JSON, INDEX i ((CAST(c2->'$.array' AS UNSIGNED ARRAY))));
> Query OK, 0 rows affected (0.06 sec)
> 
> mysql> INSERT INTO t (c1, c2) VALUES (11, '{"name": "John", "array": 30}');
> Query OK, 1 row affected (0.01 sec)
> 
> mysql> select c2->'$.array' from t;
> +---------------+
> | c2->'$.array' |
> +---------------+
> | 30            |
> +---------------+
> 1 row in set (0.01 sec)
> 
> mysql> ADMIN CHECK INDEX t i (0, 22);
> ERROR 1105 (HY000): runtime error: index out of range [3] with length 3
> ```

cool will continue

joechenrh:
```patch
diff --git a/pkg/executor/admin.go b/pkg/executor/admin.go
index e67dd9a37c..61ecc9feaa 100644
--- a/pkg/executor/admin.go
+++ b/pkg/executor/admin.go
@@ -27,6 +27,7 @@ import (
 	"github.com/pingcap/tidb/pkg/kv"
 	"github.com/pingcap/tidb/pkg/meta/model"
 	"github.com/pingcap/tidb/pkg/parser/ast"
+	"github.com/pingcap/tidb/pkg/parser/mysql"
 	"github.com/pingcap/tidb/pkg/parser/terror"
 	plannercore "github.com/pingcap/tidb/pkg/planner/util"
 	"github.com/pingcap/tidb/pkg/table"
@@ -96,6 +97,29 @@ func (e *CheckIndexRangeExec) Next(ctx context.Context, req *chunk.Chunk) error
 	}
 }
 
+func buildHandleCols(tblInfo *model.TableInfo) []*model.ColumnInfo {
+	var handleCols []*model.ColumnInfo
+	var pkIdx *model.IndexInfo
+	switch {
+	case tblInfo.PKIsHandle:
+		for _, col := range tblInfo.Columns {
+			if mysql.HasPriKeyFlag(col.GetFlag()) {
+				handleCols = []*model.ColumnInfo{col}
+				break
+			}
+		}
+	case tblInfo.IsCommonHandle:
+		pkIdx = tables.FindPrimaryIndex(tblInfo)
+		cols := tblInfo.Cols()
+		for _, idxCol := range pkIdx.Columns {
+			handleCols = append(handleCols, cols[idxCol.Offset])
+		}
+	default:
+		handleCols = []*model.ColumnInfo{model.NewExtraHandleColInfo()}
+	}
+	return handleCols
+}
+
 // Open implements the Executor Open interface.
 func (e *CheckIndexRangeExec) Open(ctx context.Context) error {
 	tCols := e.table.Cols()
@@ -104,12 +128,14 @@ func (e *CheckIndexRangeExec) Open(ctx context.Context) error {
 		e.cols = append(e.cols, col)
 	}
 
-	colTypeForHandle := e.GetSchema().Columns[len(e.cols)].RetType
-	e.cols = append(e.cols, &model.ColumnInfo{
-		ID:        model.ExtraHandleID,
-		Name:      model.ExtraHandleName,
-		FieldType: *colTypeForHandle,
-	})
+	e.cols = append(e.cols, buildHandleCols(e.table)...)
+
+	tps := make([]*types.FieldType, 0, len(e.cols))
+	for _, tp := range e.RetFieldTypes() {
+		t := tp.Clone()
+		t.SetArray(false)
+		tps = append(tps, t)
+	}
 
 	e.srcChunk = exec.TryNewCacheChunk(e)
 	dagPB, err := e.buildDAGPB()
@@ -133,7 +159,7 @@ func (e *CheckIndexRangeExec) Open(ctx context.Context) error {
 		return err
 	}
 
-	e.result, err = distsql.Select(ctx, e.Ctx().GetDistSQLCtx(), kvReq, e.RetFieldTypes())
+	e.result, err = distsql.Select(ctx, e.Ctx().GetDistSQLCtx(), kvReq, tps)
 	if err != nil {
 		return err
 	}

```


This patch can make distsql work, but this is not the final solution, as the logic of `CheckIndexRangeExec.Next` is still wrong. The problem is that
- we assume that the table has implicit primary key (by appending column with `ExtraHandleID`)
- the type passed into the distsql is wrong

Maybe we shoud add more tests in this PR, like:
```SQL
CREATE TABLE t (c1 INT PRIMARY KEY, c2 JSON, INDEX i (c1, (CAST(c2->'$.array' AS UNSIGNED ARRAY))));

CREATE TABLE t (c1 INT PRIMARY KEY, c2 JSON, INDEX i (c1, (c1, CAST(c2->'$.array' AS UNSIGNED ARRAY))));

CREATE TABLE t (c1 INT, c2 JSON, INDEX i (c1, (CAST(c2->'$.array' AS UNSIGNED ARRAY))));

CREATE TABLE t (c1 INT, c2 JSON, C3 INT, PRIMARY KEY (c1, c3), INDEX i (c1, (CAST(c2->'$.array' AS UNSIGNED ARRAY))));
```

yihong0618:
> ```diff
> diff --git a/pkg/executor/admin.go b/pkg/executor/admin.go
> index e67dd9a37c..61ecc9feaa 100644
> --- a/pkg/executor/admin.go
> +++ b/pkg/executor/admin.go
> @@ -27,6 +27,7 @@ import (
>  	"github.com/pingcap/tidb/pkg/kv"
>  	"github.com/pingcap/tidb/pkg/meta/model"
>  	"github.com/pingcap/tidb/pkg/parser/ast"
> +	"github.com/pingcap/tidb/pkg/parser/mysql"
>  	"github.com/pingcap/tidb/pkg/parser/terror"
>  	plannercore "github.com/pingcap/tidb/pkg/planner/util"
>  	"github.com/pingcap/tidb/pkg/table"
> @@ -96,6 +97,29 @@ func (e *CheckIndexRangeExec) Next(ctx context.Context, req *chunk.Chunk) error
>  	}
>  }
>  
> +func buildHandleCols(tblInfo *model.TableInfo) []*model.ColumnInfo {
> +	var handleCols []*model.ColumnInfo
> +	var pkIdx *model.IndexInfo
> +	switch {
> +	case tblInfo.PKIsHandle:
> +		for _, col := range tblInfo.Columns {
> +			if mysql.HasPriKeyFlag(col.GetFlag()) {
> +				handleCols = []*model.ColumnInfo{col}
> +				break
> +			}
> +		}
> +	case tblInfo.IsCommonHandle:
> +		pkIdx = tables.FindPrimaryIndex(tblInfo)
> +		cols := tblInfo.Cols()
> +		for _, idxCol := range pkIdx.Columns {
> +			handleCols = append(handleCols, cols[idxCol.Offset])
> +		}
> +	default:
> +		handleCols = []*model.ColumnInfo{model.NewExtraHandleColInfo()}
> +	}
> +	return handleCols
> +}
> +
>  // Open implements the Executor Open interface.
>  func (e *CheckIndexRangeExec) Open(ctx context.Context) error {
>  	tCols := e.table.Cols()
> @@ -104,12 +128,14 @@ func (e *CheckIndexRangeExec) Open(ctx context.Context) error {
>  		e.cols = append(e.cols, col)
>  	}
>  
> -	colTypeForHandle := e.GetSchema().Columns[len(e.cols)].RetType
> -	e.cols = append(e.cols, &model.ColumnInfo{
> -		ID:        model.ExtraHandleID,
> -		Name:      model.ExtraHandleName,
> -		FieldType: *colTypeForHandle,
> -	})
> +	e.cols = append(e.cols, buildHandleCols(e.table)...)
> +
> +	tps := make([]*types.FieldType, 0, len(e.cols))
> +	for _, tp := range e.RetFieldTypes() {
> +		t := tp.Clone()
> +		t.SetArray(false)
> +		tps = append(tps, t)
> +	}
>  
>  	e.srcChunk = exec.TryNewCacheChunk(e)
>  	dagPB, err := e.buildDAGPB()
> @@ -133,7 +159,7 @@ func (e *CheckIndexRangeExec) Open(ctx context.Context) error {
>  		return err
>  	}
>  
> -	e.result, err = distsql.Select(ctx, e.Ctx().GetDistSQLCtx(), kvReq, e.RetFieldTypes())
> +	e.result, err = distsql.Select(ctx, e.Ctx().GetDistSQLCtx(), kvReq, tps)
>  	if err != nil {
>  		return err
>  	}
> ```
> 
> This patch can make distsql work, but this is not the final solution, as the logic of `CheckIndexRangeExec.Next` is still wrong. The problem is that
> 
> * we assume that the table has implicit primary key
> * the type passed into the distsql is wrong

trying to learn it thank you very much, learn a lot from it


try this patch seems do not handle

```sql
mysql> CREATE TABLE t (c1 INT PRIMARY KEY, c2 JSON, INDEX i ((CAST(c2->'$.array' AS UNSIGNED ARRAY))));

mysql> INSERT INTO t (c1, c2) VALUES (11, '{"name": "John", "array": 30}');

mysql> select c2->'$.array' from t;
+---------------+
| c2->'$.array' |
+---------------+
| 30            |
+---------------+
1 row in set (0.01 sec)

mysql> ADMIN CHECK INDEX t i (0, 22);
ERROR 1105 (HY000): runtime error: index out of range [3] with length 3
```

joechenrh:
> try this patch seems do not handle

Sorry for the late reply. Yeah, this patch doesn't fix the whole problem, it just make the index scan part works (Line72), although the data stored inside is not totally correct.
https://github.com/pingcap/tidb/blob/d0ac8e61518286b3f1477c322a834385ee1e0227/pkg/executor/admin.go#L68-L75

> BTW, the patch itself may also have other problems(for example, for non-clustered index), it just make the case in issue pass.

The rest part of the code is still wrong and need to be fixed. Below is the result from my test:
```SQL
mysql> CREATE TABLE t (c1 INT PRIMARY KEY, c2 JSON, INDEX i ((CAST(c2->'$.array' AS UNSIGNED ARRAY))));

mysql> INSERT INTO t (c1, c2) VALUES (11, '{"name": "John", "array": 30}');

mysql> select c2->'$.array' from t;
+---------------+
| c2->'$.array' |
+---------------+
| 30            |
+---------------+
1 row in set (0.00 sec)

mysql> ADMIN CHECK INDEX t i (0, 22);
ERROR 1105 (HY000): runtime error: index out of range [1] with length 1
```

yihong0618:
> tidb/pkg/executor/admin.go

yes mine is too...

joechenrh:
The problem is caused by wrong type stored in the executor. 

Multi-valued index is a special index, it will be bind to a hidden column with its type is array. It's to indicate that this column is corresponding to the MV-Index.

image

But when we scan this index, we need to remove this flag because the values stored are not array, just int. It will affect the  logic related to chunk processing. 

You can try to fix the issue in this way.

PR #62527, opened 2025-07-21
importinto: reduce def engine size to 100G and always use max node count to adjust subtask count

D3Hunter:

Issue Number: ref #61702

Problem Summary:

as title，100g is the same as the size we used to split `ingest` step subtasks, smaller size makes the slow down due to failover as less as possible, and as resources is scaled out on demand, smaller subtask makes the long tail issue, if exist, less obvious
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/disttask/importinto/scheduler.go
```diff
@@ -166,6 +167,7 @@ func NewImportScheduler(
 func NewImportSchedulerForTest(globalSort bool) scheduler.Scheduler {
 	return &importScheduler{
 		GlobalSort: globalSort,
+		taskKS:     tidb.GetGlobalKeyspaceName(),
```
joechenrh, line 170: `taskKS` is not used?

D3Hunter: here 
https://github.com/pingcap/tidb/blob/e2a9f04ee7a9ad2e66681bda0cb41e8332fc5524/pkg/disttask/importinto/scheduler.go#L682-L681

joechenrh: Got it

Issue #62551, opened 2025-07-22
rollup will output more rows than expected


AilinKid:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

<!-- a step by step guide for reproducing the bug. -->
```
create table `testpay`(
`bill_code` varchar(100),
`org_id` decimal(19,0),
`amt` decimal(15,2),
`pay_date` varchar(10),
PRIMARY KEY (`bill_code`));

create table `testreturn`(
`bill_code` varchar(100),
`org_id` decimal(19,0),
`amt` decimal(15,2),
`ret_date` varchar(10),
PRIMARY KEY (`bill_code`));

create table `testorg`(
`org_id` decimal(19,0),
`org_code` varchar(100),
`org_name` varchar(100),
`org_type` varchar(100),
PRIMARY KEY (`org_id`));
insert into testorg (org_id,org_code,org_name,org_type) values(1,'ORG0001','部门1','DEPT');
insert into testorg (org_id,org_code,org_name,org_type) values(2,'ORG0002','部门2','DEPT');
insert into testorg (org_id,org_code,org_name,org_type) values(3,'ORG0003','部门3','DEPT');
insert into testorg (org_id,org_code,org_name,org_type) values(4,'ORG0004','部门4','DEPT');
insert into testorg (org_id,org_code,org_name,org_type) values(5,'ORG0005','公司1','ORG');
insert into testorg (org_id,org_code,org_name,org_type) values(6,'ORG0006','公司2','ORG');
insert into testorg (org_id,org_code,org_name,org_type) values(7,'ORG0007','公司3','ORG');
insert into testpay (bill_code,org_id,amt,pay_date) values('PAY0001',1,100,'2024-06-01');
insert into testpay (bill_code,org_id,amt,pay_date) values('PAY0002',2,200,'2024-06-02');
insert into testpay (bill_code,org_id,amt,pay_date) values('PAY0003',3,300,'2024-06-03');
insert into testpay (bill_code,org_id,amt,pay_date) values('PAY0004',4,400,'2024-07-01');
insert into testpay (bill_code,org_id,amt,pay_date) values('PAY0005',5,500,'2024-07-02');
insert into testpay (bill_code,org_id,amt,pay_date) values('PAY0006',6,600,'2024-07-03');
insert into testreturn (bill_code,org_id,amt,ret_date) values('RET0001',1,100,'2024-06-01');
[...]

PR #62558, opened 2025-07-22
planner: fix expand operator shouldn't keep child keys && fix grouping function forget to encode their func meta

AilinKid:

Issue Number: close https://github.com/pingcap/tidb/issues/62551

Problem Summary:

* previously, tidb use buildKeyInfo to derive unique key and pk for each logical operator. While for logical expand, if we use baseLogicalPlan's buildKeyInfo logic, it will derive and keep child unique key info. Since expand op will duplicate rows, we just keep it sample as setting it nil.
* the second problem is that: expression hashcode will simply encode func name and return type for normal scalar function, while for grouping function, its function meta should also be taken into consideration. Otherwise, two different grouping function will be seen as a same one in predicateSimplication logic.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/operator/logicalop/logical_expand.go
+++ b/pkg/planner/core/operator/logicalop/logical_expand.go
@@ -401,4 +401,5 @@ func (p *LogicalExpand) GenerateGroupingIDIncrementModeNumericSet(oneSetOffset i
 func (*LogicalExpand) BuildKeyInfo(selfSchema *expression.Schema, _ []*expression.Schema) {
 	// since LogicalExpand is a logical operator which will split the rows out, duplicated rows may exist in the output.
 	selfSchema.SetKeys(nil)
+	selfSchema.SetUniqueKeys(nil)
 }


pkg/planner/core/operator/logicalop/logicalop_test/logical_operator_test.go
```diff
@@ -147,3 +147,59 @@ LIMIT 0,
...
+		require.Equal(t, len(res.Rows()), 1)
+		res = tk.MustQuery("SELECT SUM(IFNULL(pay.payamt, 0)) AS payamt,     SUM(IFNULL(ret.retamt, 0)) AS retamt,     GROUPING(org.org_type) AS grouptype,     org.org_type,     GROUPING(org.org_id) AS groupid,     org.org_id,     org.org_name   FROM testorg org   LEFT JOIN (     SELECT       SUM(IFNULL(amt, 0)) AS payamt,       org_id     FROM testpay tp     WHERE tp.pay_date BETWEEN '2024-06-01' AND '2024-07-31'     GROUP BY org_id   ) pay ON pay.org_id = org.org_id   LEFT JOIN (     SELECT       SUM(IFNULL(amt, 0)) AS retamt,       org_id     FROM testreturn tr     WHERE tr.ret_date BETWEEN '2024-06-01' AND '2024-07-31'     GROUP BY org_id   ) ret ON ret.org_id = org.org_id   GROUP BY org.org_type, org.org_id WITH ROLLUP having  groupid = 1 AND grouptype = 1;")
+		require.Equal(t, len(res.Rows()), 1)
```
hawkingrei, line 204: Can you add test case to save the explain?

hawkingrei: I want to see the new ```groupingID``` in the explain。

windtalker (approved):
expression part lgtm

pkg/expression/scalar_function.go
```diff
@@ -741,6 +741,17 @@ func ReHashCode(sf *ScalarFunction) {
...
 	}
+	if sf.FuncName.L == ast.Grouping {
+		sf.hashcode = codec.EncodeInt(sf.hashcode, int64(sf.Function.(*BuiltinGroupingImplSig).GetGroupingMode()))
+		marks := sf.Function.(*BuiltinGroupingImplSig).GetMetaGroupingMarks()
+		sf.hashcode = codec.EncodeInt(sf.hashcode, int64(len(marks)))
+		for _, mark := range marks {
+			sf.hashcode = codec.EncodeInt(sf.hashcode, int64(len(mark)))
+			for k := range mark {
+				sf.hashcode = codec.EncodeInt(sf.hashcode, int64(k))
+			}
```
time-and-fate: In golang, the order of iterating a map with `for ... range` is nondeterministic, so one expression will produce different results here.
I think we need to sort it before encoding.

pkg/planner/core/operator/logicalop/logical_expand.go
```diff
@@ -396,3 +396,9 @@ func (p *LogicalExpand) GenerateGroupingIDIncrementModeNumericSet(oneSetOffset i
 	// {0,1,2}. This grouping id set is returned back as this grouping function's specified meta when rewriting the grouping function,
 	// and the evaluating logic is quite simple as IN compare.
 }
+
+// BuildKeyInfo implements base.LogicalPlan interface.
+func (*LogicalExpand) BuildKeyInfo(selfSchema *expression.Schema, _ []*expression.Schema) {
+	// since LogicalExpand is a logical operator which will split the rows out, duplicated rows may exist in the output.
+	selfSchema.SetKeys(nil)
```
time-and-fate, line 403: Looks like we should also call `selfSchema.SetUniqueKeys(nil)`?

pkg/expression/scalar_function.go
AilinKid: make sense

pkg/planner/core/operator/logicalop/logicalop_test/testdata/cascades_suite_out.json
```diff
@@ -0,0 +1,57 @@
...
+        "Plan": [
+          "Projection 6400.00 root  Column#18->Column#22, Column#19->Column#23, grouping(gid)->Column#24, Column#15->Column#25, grouping(gid)->Column#26, Column#16->Column#27, test.testorg.org_name",
+          "└─HashAgg 6400.00 root  group by:Column#45, Column#46, Column#47, funcs:sum(Column#42)->Column#18, funcs:sum(Column#43)->Column#19, funcs:firstrow(Column#44)->test.testorg.org_name, funcs:firstrow(Column#45)->Column#15, funcs:firstrow(Column#46)->Column#16, funcs:firstrow(Column#47)->gid",
+          "  └─Projection 8000.00 root  ifnull(Column#9, 0.00)->Column#42, ifnull(Column#14, 0.00)->Column#43, test.testorg.org_name->Column#44, Column#15->Column#45, Column#16->Column#46, gid->Column#47",
+          "    └─Selection 8000.00 root  eq(grouping(gid), 1), eq(grouping(gid), 1)",
```
hawkingrei, line 11:
```suggestion
          "    └─Selection 8000.00 root  eq(grouping(gid), 1), eq(grouping(gid), 1)",
```

Interesting, it has the same condition in the selection.

hawkingrei: Some meta cannot be printed here.

PR #62563, opened 2025-07-22
infoschema, server: add per connection TLS status

### LINKED ISSUE #64223: TLS status per connection
## Enhancement

Create a table in `information_schema` or `performance_schema` that allows one to query the per connection TLS version, TLS cipher and possibly other TLS related information.

# Related
- https://github.com/pingcap/tidb/issues/27903

dveeden:

Issue Number: close #64223

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
The `performance_schema.status_by_connection` table was added to show per-connection TLS status.
```


dveeden:
Current status:

Start a server with AutoTLS enabled
```
./bin/tidb-server -config <(echo -en "[security]\nauto-tls=true\n")
```

Connect  with different TLS config
```
mysql --host 127.0.0.1 --port 4000 -u root test
mysql -h 127.0.0.1 -u root -P 4000 --tls-version TLSv1.2
```

Result:
```
mysql-8.0.11-TiDB-v9.0.0-beta.2.pre-152-g7237eb9cbb-dirty> select * from status_by_connection;
+---------------+---------------+---------------------------------------+
| CONNECTION_ID | VARIABLE_NAME | VARIABLE_VALUE                        |
+---------------+---------------+---------------------------------------+
|       2097156 | Tls_cipher    | TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 |
|       2097156 | Tls_version   | TLS 1.2                               |
|       2097154 | Tls_cipher    | TLS_AES_128_GCM_SHA256                |
|       2097154 | Tls_version   | TLS 1.3                               |
+---------------+---------------+---------------------------------------+
4 rows in set (0.001 sec)
```

Open issues with this:
- Missing tests
- The order is not stable
- This is per server, not cluster wide
- This is not fully the same as MySQL as this is per connection instead of per thread. We don't have PS threads in TiDB, at least not yet.
- This is only for TLS status variables instead of a generic way to get per-session info for all sessions.
- I named these `Tls_version`, while it probably should be `Ssl_version` to match session variables and MySQL. (However it actually is TLS, not the older SSL)
- The protocol names and cipher names might not match the ones of MySQL as these are Go specific. (e.g. `TLSv1.3` vs `TLS 1.3`). I think we adjust some of this in the status variable code.

dveeden:
The `join_outer` test failures from mysql-test don't seem related to this PR and don't happen when I run the test locally.

dveeden:
> Open issues with this:
> 
>     * Missing tests
A test has been added

>     * The order is not stable
I think this is fine. If one needs a table order just add an `ORDER BY`.

>     * This is per server, not cluster wide
This is still the case. 

>     * This is not fully the same as MySQL as this is per connection instead of per thread. We don't have PS threads in TiDB, at least not yet.
I don't think this is for this PR to solve

>     * This is only for TLS status variables instead of a generic way to get per-session info for all sessions.
This is still the case.

>     * I named these `Tls_version`, while it probably should be `Ssl_version` to match session variables and MySQL. (However it actually is TLS, not the older SSL)
I've renamed them to fully match the session variables 

>     * The protocol names and cipher names might not match the ones of MySQL as these are Go specific. (e.g. `TLSv1.3` vs `TLS 1.3`). I think we adjust some of this in the status variable code.
This has been changed to fully match what is now used for session variables.

## Demo

```
mysql> TABLE performance_schema.status_by_connection;
+---------------+---------------+------------------------+
| CONNECTION_ID | VARIABLE_NAME | VARIABLE_VALUE         |
+---------------+---------------+------------------------+
|       2097154 | Ssl_cipher    | TLS_AES_128_GCM_SHA256 |
|       2097154 | Ssl_version   | TLSv1.3                |
+---------------+---------------+------------------------+
2 rows in set (0.001 sec)

mysql> SHOW STATUS LIKE 'Ssl\_cipher';
+---------------+------------------------+
| Variable_name | Value                  |
+---------------+------------------------+
| Ssl_cipher    | TLS_AES_128_GCM_SHA256 |
+---------------+------------------------+
1 row in set (0.003 sec)

mysql> SHOW STATUS LIKE 'Ssl\_version';
+---------------+---------+
| Variable_name | Value   |
+---------------+---------+
| Ssl_version   | TLSv1.3 |
+---------------+---------+
1 row in set (0.002 sec)
```

dveeden:
I just added this:

```
$ curl -s http://127.0.0.1:10080/metrics | grep -E '^tidb_server_tls'
tidb_server_tls_cipher{cipher="TLS_AES_128_GCM_SHA256"} 39
tidb_server_tls_cipher{cipher="TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA"} 1
tidb_server_tls_cipher{cipher="TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"} 3
tidb_server_tls_cipher{cipher="TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA"} 1
tidb_server_tls_cipher{cipher="TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384"} 1
tidb_server_tls_cipher{cipher="TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256"} 1
tidb_server_tls_version{version="TLSv1.2"} 7
tidb_server_tls_version{version="TLSv1.3"} 39
```

For testing something like this might be useful:
```sh
for c in $(openssl ciphers -v | awk '{ print $1}'); do mysql -h 127.0.0.1 -u root -P 4000 --tls-version TLSv1.2 --ssl-cipher $c -e 'DO 1'; echo "$c $?"; done
```

pkg/metrics/server.go
```diff
@@ -422,6 +426,20 @@ func InitServerMetrics() {
...
+			Name:      "tls_version",
+			Help:      "Counter per TLS Version.",
+		}, []string{LblVersion})
+	TLSCipher = metricscommon.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "tidb",
+			Subsystem: "server",
+			Name:      "tls_cipher",
+			Help:      "Counter per TLS Cipher.",
+		}, []string{LblCipher})
```
yibin87, line 442: Are LblCipher and LblVersion enumerable? I ask this question because metrics vec with high distinct values will cost much more resource.

dveeden: For `LblVersion`: there are two possible values: `TLSv1.2` and `TLSv1.3`. See `versionString` in `pkg/util/tls/tls.go`.

For `LblCipher`: there are 25 values. See `tlsCipherString` in `pkg/util/tls/tls.go`. However in a production setup I would expect something like 5 distinct values.

disksing (approved):

pkg/session/sessmgr/processinfo.go
```diff
@@ -258,4 +258,5 @@ type Manager interface {
 	GetInternalSessionStartTSList() []uint64
 	// GetConAttrs gets the connection attributes
 	GetConAttrs(user *auth.UserIdentity) map[uint64]map[string]string
+	GetStatusVars() map[uint64]map[string]string
```
D3Hunter, line 265: please add comment on the meaning of return values

pkg/infoschema/tables.go
```diff
@@ -2892,3 +2892,24 @@ func FilterClusterServerInfo(serversInfo []ServerInfo, nodeTypes, addresses set.
...
 }
+
+// GetDataFromStatusByConn is getting the per-connection status for `performance_schema.status_by_connection`
+func GetDataFromStatusByConn(sctx sessionctx.Context) ([][]types.Datum, error) {
+	sm := sctx.GetSessionManager()
+	if sm == nil {
+		return nil, nil
+	}
+	statusVars := sm.GetStatusVars()
+	rows := make([][]types.Datum, 0, 1000)
```
D3Hunter:
```suggestion
	rows := make([][]types.Datum, 0, 2 * len(statusVars))
```

D3Hunter (commented):
infoschema part lgtm

PR #62616, opened 2025-07-24
planner: TTL scan can trigger sync/async load/generateRuntimeFilter

### LINKED ISSUE #61822: TTL scan cannot trigger sync/async load/generateRuntimeFilter
## Bug Report

We previously thought that internal SQL was for system tables,and since system tables do not have statistics,we could skip the sync load.However,TTL is different because it generates SQL execution based on user tables.Therefore,we need to align its behavior with regular tables.

hawkingrei:

Issue Number: close #61822

Problem Summary:

The TTL scan is different from our other internal SQL, as it reads user tables. So, our previous approach of determining whether to perform sync load based on whether it is internal SQL has issues.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/optimizer.go
+++ b/pkg/planner/core/optimizer.go
@@ -481,8 +481,7 @@ func postOptimize(ctx context.Context, sctx base.PlanContext, plan base.Physical
 }
 
 func generateRuntimeFilter(sctx base.PlanContext, plan base.PhysicalPlan) {
-	if !sctx.GetSessionVars().IsRuntimeFilterEnabled() ||
-		(sctx.GetSessionVars().InRestrictedSQL && !sctx.GetSessionVars().InternalSQLScanUserTable) {
+	if !sctx.GetSessionVars().IsRuntimeFilterEnabled() || sctx.GetSessionVars().InRestrictedSQL {
 		return
 	}
 	logutil.BgLogger().Debug("Start runtime filter generator")
--- a/pkg/planner/core/optimizer.go
+++ b/pkg/planner/core/optimizer.go
@@ -375,7 +375,8 @@ func adjustOptimizationFlags(flag uint64, logic base.LogicalPlan) uint64 {
 		// When we use the straight Join Order hint, we should disable the join reorder optimization.
 		flag &= ^rule.FlagJoinReOrder
 	}
-	if !logic.SCtx().GetSessionVars().InRestrictedSQL {
+	// InternalSQLUseUserTable is for ttl scan.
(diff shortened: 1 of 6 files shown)


pkg/sessionctx/variable/session.go
```diff
@@ -1732,6 +1732,9 @@ type SessionVars struct {
 
 	// BulkDMLEnabled indicates whether to enable bulk DML in pipelined mode.
 	BulkDMLEnabled bool
+
+	// InternalSQLUseUserTable indicates whether to use user table for internal SQL. it will be used by TTL scan
+	InternalSQLUseUserTable bool
```
elsa0520: Is it better to add this variable here? Because it seems to be an internal variable, is there any other place to add it? Because the session variable seems to be visible to the client.

hawkingrei: The session variable needs to be hooked onto the session in order to be used externally. Since we haven't implemented this hook, it is only for internal use and there won't be any issues in this regard. `se.GetSessionVars()` is a mature and universal solution.

hawkingrei: We cannot allow this switch to be turned on but appear in places where it should not be.

AilinKid:
```suggestion
	InternalSQLScanUserTable bool
```

AilinKid (commented):
rest lgtm

pkg/sessionctx/variable/session.go
hawkingrei: Updated

AilinKid (approved):

qw4990 (approved):

pkg/planner/core/optimizer.go
```diff
@@ -480,7 +481,8 @@ func postOptimize(ctx context.Context, sctx base.PlanContext, plan base.Physical
 }
 
 func generateRuntimeFilter(sctx base.PlanContext, plan base.PhysicalPlan) {
-	if !sctx.GetSessionVars().IsRuntimeFilterEnabled() || sctx.GetSessionVars().InRestrictedSQL {
+	if !sctx.GetSessionVars().IsRuntimeFilterEnabled() ||
```
elsa0520: Will ttl scan using the Join?

hawkingrei: No, I think we can remove it.

Issue #62200, opened 2025-07-03
involvedSchemaInfo is invalid for AlterTableMode and RefreshMeta


D3Hunter:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

there are no explicit set or schema/table name set
https://github.com/pingcap/tidb/blob/339f07ae8f2ba406abc7f4757ffcce918cc86a56/pkg/ddl/executor.go#L7076-L7084
https://github.com/pingcap/tidb/blob/339f07ae8f2ba406abc7f4757ffcce918cc86a56/pkg/ddl/executor.go#L5769-L5777
```
[2025/07/03 16:44:33.813 +00:00] [ERROR] [wait_group_wrapper.go:178] ["panic in the wait group"] [recover="assert failed, InvolvingSchemaInfo should be like an enumerate type: model.InvolvingSchemaInfo{Database:\"\", Table:\"\", Policy:\"\", ResourceGroup:\"\", Mode:0}"] [stack="github.com/pingcap/tidb/pkg/util.(*WaitGroupWrapper).RunWithLog.func1.1
  pkg/util/wait_group_wrapper.go:178
runtime.gopanic
  GOROOT/src/runtime/panic.go:791
github.com/pingcap/tidb/pkg/util/intest.doPanic
  pkg/util/intest/assert_common.go:58
github.com/pingcap/tidb/pkg/util/intest.doAssert
  pkg/util/intest/assert_common.go:30
github.com/pingcap/tidb/pkg/util/intest.Assert
  pkg/util/intest/assert.go:25
github.com/pingcap/tidb/pkg/ddl.(*runningJobs).checkRunnable
  pkg/ddl/ddl_running_jobs.go:166
github.com/pingcap/tidb/pkg/ddl.(*jobScheduler).loadAndDeliverJobs
  pkg/ddl/job_scheduler.go:423
github.com/pingcap/tidb/pkg/ddl.(*jobScheduler).schedule
  pkg/ddl/job_scheduler.go:333
github.com/pingcap/tidb/pkg/ddl.(*jobScheduler).scheduleLoop
  pkg/ddl/job_scheduler.go:268
github.com/pingcap/tidb/pkg/util.(*WaitGroupWrapper).RunWithLog.func1
  pkg/util/wait_group_wrapper.go:181"] [stack="github.com/pingcap/tidb/pkg/util.(*WaitGroupWrapper).RunWithLog.func1.1
[...]

PR #62618, opened 2025-07-24
ddl: check DDL args set InvolvedSchemaInfo

River2000i:

Issue Number: close https://github.com/pingcap/tidb/issues/62200

Problem Summary:
Not verify InvolvedSchemaInfo set or not before submit DDL job.

check `GetInvolvingSchemaInfo()` get `""` before submit DDL job.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/executor.go
+++ b/pkg/ddl/executor.go
@@ -6780,6 +6780,12 @@ func (e *executor) DoDDLJobWrapper(ctx sessionctx.Context, jobW *JobWrapper) (re
 		// Instead, we merge all the jobs into one pending job.
 		return appendToSubJobs(mci, jobW)
 	}
+	// // all DDLs should set InvolvingSchemaInfo to make sure DDL execution in order.
+	// for _, info := range job.GetInvolvingSchemaInfo() {
+	// 	if info.Database == model.InvolvingNone || info.Table == model.InvolvingNone {
+	// 		// return dbterror.ErrInvalidDDLJob.GenWithStack("invalid ddl job args: InvolvingSchemaInfo")
+	// 	}
+	// }
 	// Get a global job ID and put the DDL job in the queue.
 	setDDLJobQuery(ctx, job)
 	e.deliverJobTask(jobW)
--- a/pkg/ddl/executor.go
+++ b/pkg/ddl/executor.go
@@ -7087,14 +7087,10 @@ func (e *executor) RefreshMeta(sctx sessionctx.Context, args *model.RefreshMetaA
 		BinlogInfo:     &model.HistoryInfo{},
 		CDCWriteSource: sctx.GetSessionVars().CDCWriteSource,
 		SQLMode:        sctx.GetSessionVars().SQLMode,
-		InvolvingSchemaInfo: []model.InvolvingSchemaInfo{
-			{
-				// For RefreshMeta, the table/schema might not exist anymore.
-				// Since we can't determine the exact target, use model.InvolvingAll to block concurrent DDLs as a safe default.
-				Database: model.InvolvingAll,
-				Table:    model.InvolvingAll,
-			},
-		},
+		InvolvingSchemaInfo: []model.InvolvingSchemaInfo{{
+			Database: args.SchemaName,
+			Table:    args.TableName,
+		}},
 	}
 	sctx.SetValue(sessionctx.QueryString, "skip")
 	err := e.doDDLJob2(sctx, job, args)
--- a/pkg/ddl/executor.go
+++ b/pkg/ddl/executor.go
@@ -5770,6 +5770,8 @@ func (e *executor) AlterTableMode(sctx sessionctx.Context, args *model.AlterTabl
 		Version:        model.JobVersion2,
 		SchemaID:       args.SchemaID,
 		TableID:        args.TableID,
+		SchemaName:     schema.Name.O,
+		TableName:      table.Meta().Name.O,
 		Type:           model.ActionAlterTableMode,
 		BinlogInfo:     &model.HistoryInfo{},
 		CDCWriteSource: sctx.GetSessionVars().CDCWriteSource,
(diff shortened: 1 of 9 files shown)


pkg/ddl/executor.go
```diff
@@ -5770,6 +5770,8 @@ func (e *executor) AlterTableMode(sctx sessionctx.Context, args *model.AlterTabl
 		Version:        model.JobVersion2,
 		SchemaID:       args.SchemaID,
 		TableID:        args.TableID,
+		SchemaName:     schema.Name.O,
+		TableName:      table.Meta().Name.O,
```
D3Hunter: `InvolvingSchemaInfo` is set explicitly, set only one of them is enough

pkg/ddl/executor.go
```diff
@@ -6778,6 +6780,12 @@ func (e *executor) DoDDLJobWrapper(ctx sessionctx.Context, jobW *JobWrapper) (re
 		// Instead, we merge all the jobs into one pending job.
 		return appendToSubJobs(mci, jobW)
 	}
+	// all DDLs should set InvolvingSchemaInfo to make sure DDL execution in order.
+	for _, info := range job.GetInvolvingSchemaInfo() {
+		if info.Database == model.InvolvingNone || info.Table == model.InvolvingNone {
+			return dbterror.ErrInvalidDDLJob.GenWithStack("invalid ddl job args: InvolvingSchemaInfo")
+		}
+	}
```
D3Hunter: please wrap it like `checkHistoryJobInTest`, only run when `intest.EnableInternalCheck` is `true`, this check is mostly to avoid human error when we add new DDL

pkg/ddl/executor.go
```diff
@@ -7083,16 +7091,16 @@ func (e *executor) RefreshMeta(sctx sessionctx.Context, args *model.RefreshMetaA
 		Version:        model.JobVersion2,
 		SchemaID:       args.SchemaID,
 		TableID:        args.TableID,
+		SchemaName:     args.SchemaName,
+		TableName:      args.TableName,
```
D3Hunter: don't set it, `TableName` is for real table, InvolveAll is not a valid value for this field

pkg/meta/model/job_args.go
```diff
@@ -1788,8 +1788,10 @@ func GetFinishedModifyColumnArgs(job *Job) (*ModifyColumnArgs, error) {
 
 // RefreshMetaArgs is the argument for RefreshMeta.
 type RefreshMetaArgs struct {
-	SchemaID int64 `json:"schema_id,omitempty"`
-	TableID  int64 `json:"table_id,omitempty"`
+	SchemaID   int64  `json:"schema_id,omitempty"`
+	TableID    int64  `json:"table_id,omitempty"`
+	SchemaName string `json:"schema_name,omitempty"`
+	TableName  string `json:"table_name,omitempty"`
```
D3Hunter: prefer to name them as  `InvolvedDB`/`InvolvedTable`

JQWong7:
@D3Hunter https://github.com/pingcap/tidb/pull/62618/commits/5d1f4ced1dc15cbb07eb41f21da80a287a6e5fec  address comment, PTAL~

Tristan1900:
@River2000i will it break anything if names are empty strings passed down to DDL?

pkg/meta/model/job_args.go
```diff
@@ -1788,8 +1788,10 @@ func GetFinishedModifyColumnArgs(job *Job) (*ModifyColumnArgs, error) {
 
 // RefreshMetaArgs is the argument for RefreshMeta.
 type RefreshMetaArgs struct {
-	SchemaID int64 `json:"schema_id,omitempty"`
-	TableID  int64 `json:"table_id,omitempty"`
+	InvolvedDBID      int64  `json:"involved_db_id,omitempty"`
+	InvolvedTableID   int64  `json:"involved_table_id,omitempty"`
+	InvolvedDBName    string `json:"involved_db_name,omitempty"`
+	InvolvedTableName string `json:"involved_table_name,omitempty"`
```
tangenta: Will this change cause an upgrade compatibility issue?

JQWong7:
> @River2000i will it break anything if names are empty strings passed down to DDL?

The `InvolvingSchemaInfo` indicate which table/schema will the DDL affect. It may break the DDL execution order. In log restore, it will call `RefreshMeta` for every table/schema in order. So i think it will not affect log restore for now.

pkg/meta/model/job_args.go
JQWong7: ~~Do you means if there is a DDL job still running but cluster upgrade?~~ ~~The DDL job only call by BR log restore, seems it will not support online upgrade with running a BR log restore job. But if BR version is lower than TiDB cluster may occur error🤔.~~ ~~Maybe we should upgrade BR version after upgrade TiDB?~~

reverted json tag modify

pkg/meta/model/job_args.go
```diff
@@ -1788,8 +1788,10 @@ func GetFinishedModifyColumnArgs(job *Job) (*ModifyColumnArgs, error) {
 
 // RefreshMetaArgs is the argument for RefreshMeta.
 type RefreshMetaArgs struct {
-	SchemaID int64 `json:"schema_id,omitempty"`
-	TableID  int64 `json:"table_id,omitempty"`
+	InvolvedDBID      int64  `json:"involved_db_id,omitempty"`
+	InvolvedTableID   int64  `json:"involved_table_id,omitempty"`
```
D3Hunter: don't change the json tag, it breaks compatibility

and I think previous field name is fine

JQWong7: https://github.com/pingcap/tidb/pull/62618/commits/b658da0a89518eae39f12b8366e44a53bdaeed31

tangenta (approved):

PR #62638, opened 2025-07-25
disttask: refine the speed calculation of import jobs

joechenrh:

Issue Number: ref https://github.com/pingcap/tidb/issues/61088

Problem Summary:

Previously, the speed was calculated by `ProcessedBytes / RunningTimes`, where `RunningTimes` is the duration from now to `update_time` stored in `tidb_import_jobs`. However, this value can't reflect actual speed when step is amost complete, as there may have only a small number of subtasks.

To better track the progress, we store more data points in the summary rather than only the latest data. These points are used to calcuate the speed for each subtask in the short time window. And the speed of the import job is calculated as **the sum of the speeds of all the subtasks**. For example, some recently finished tasks will also contribute to the speed calculation.

```
       finished task
|----|----|----|----|----|
                                 running task
                                 |----|----|
               |<------------------------->|
                        time range
```

By this approach, we can
- Better reflect the actual speed of running tasks
- Make the calculated speed smoother when some tasks finish/start.

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/disttask/importinto/scheduler.go
```diff
@@ -315,10 +315,6 @@ func (sch *importScheduler) OnNextSubtasksBatch(
 			return nil, err
 		}
 		previousSubtaskMetas[proto.ImportStepEncodeAndSort] = sortAndEncodeMeta
-		// Update update_time in tidb_import_jobs
```
joechenrh, line 318: Previously, we use `tidb_import_jobs .update_time` to calculate average speed, so we need to update it after switching steps.

Now it's not needed, so just revert the change.

pkg/executor/show.go
```diff
@@ -2455,9 +2454,8 @@ func FillOneImportJobInfo(result *chunk.Chunk, info *importer.JobInfo, runInfo *
 	result.AppendString(16, runInfo.ProcessedSize())
 	result.AppendString(17, runInfo.TotalSize())
 	result.AppendString(18, runInfo.Percent())
-	speed, eta := runInfo.SpeedAndETA()
-	result.AppendString(19, speed)
-	result.AppendString(20, eta)
+	result.AppendString(19, fmt.Sprintf("%s/s", units.HumanSize(float64(runInfo.Speed))))
```
D3Hunter:
```suggestion
	result.AppendString(19, fmt.Sprintf("%s/s", units.ByteSize(float64(runInfo.Speed))))
```

pkg/disttask/framework/taskexecutor/execute/interface.go
```diff
@@ -64,18 +64,105 @@ type StepExecutor interface {
...
+const (
+	// UpdateSubtaskSummaryInterval is the interval for updating the subtask summary to
+	// subtask table.
+	UpdateSubtaskSummaryInterval = 3 * time.Second
+
+	// maxProgressInSummary is the number of progress stored in subtask summary
+	maxProgressInSummary = 5
+
+	// SubtaskSpeedDuration is the duration for calculating the speed of a subtask.
+	SubtaskSpeedDuration = UpdateSubtaskSummaryInterval * maxProgressInSummary / 2
```
D3Hunter:
```suggestion
	SubtaskSpeedUpdateInterval = UpdateSubtaskSummaryInterval * maxProgressInSummary / 2
```

pkg/disttask/framework/taskexecutor/execute/interface.go
```diff
@@ -64,18 +64,105 @@ type StepExecutor interface {
...
+)
+
+// Progress represents the progress of a subtask.
+type Progress struct {
+	RowCnt     int64     `json:"row_count,omitempty"`
```
joechenrh: Although only `UpdateTime` and `Bytes` are used for speed calculation, I still store `RowCnt` here

pkg/disttask/framework/taskexecutor/execute/interface.go
```diff
@@ -64,18 +64,105 @@ type StepExecutor interface {
...
+		UpdateTime: time.Now(),
+	})
+
+	if len(s.Progresses) > maxProgressInSummary {
+		s.Progresses = s.Progresses[len(s.Progresses)-maxProgressInSummary:]
+	}
+}
+
+// GetSpeedInTimeRange returns the speed in the specified time range.
+func (s *SubtaskSummary) GetSpeedInTimeRange(endTime time.Time, duration time.Duration) int64 {
```
D3Hunter, line 117: the algorithm is a little complex, we can simply use 
`(Progresses[last].bytes -  Progresses[first].bytes) / (Progresses[last].updateTime -  Progresses[first].updateTime)` 
we don't need that much accurate

if the progress is not updated for some time, say 30s, we can return 0 directly

pkg/disttask/framework/taskexecutor/execute/interface.go
D3Hunter: maybe put in the comment, so later we know why it's there without going back to this PR

pkg/disttask/framework/taskexecutor/execute/interface.go
joechenrh: > we can simply use
(Progresses[last].bytes -  Progresses[first].bytes) / (Progresses[last].updateTime -  Progresses[first].updateTime)

Previously I want to do so, and return 0 for finished tasks. And now I want to make the value smoother. So if we want to display it in grafana later, this can avoid some sudden speed changes when one task finished and another task starts.

D3Hunter: that's what exactly `if the progress is not updated for some time, say 30s, we can return 0 directly` does, for a finished subtask, we are still accounting it's speed, even after a subtask finishes, we are still accounting its old speed for a while, it's the same as what you does, just without the proportion part. and the proportion calculation part is actually not that useful, as in those recent `Progress`, the speed is similar, so take a proportion or not doesn't matter, but the code is more complex.

and we can change the `30s` to `UpdateSubtaskSummaryInterval * maxProgressInSummary`,

joechenrh: >  and the proportion calculation part is actually not that useful, as in those recent Progress, the speed is similar

The time window of the caluclation is based on the current time, not the latest updated time in summary.

For example, if the progress stored in a finished subtask is `(0B, 00:00:40), (100B, 00:00:50), (200B, 00:01:00)` and the length of time window is 10s, the speed we get at `00:01:02`, `00:01:05`, `00:01:10` is `80B/10s = 8B/s`, `50B/10s = 5B/s`, `0B/s` respectively.

It is indeed a little bit complicated, just a small tradeoff

D3Hunter: the `0` mostly happens at subtask start, and even with this case, the speed is `200/20 = 10B/s`, not too much difference

D3Hunter: but anyway, it's ok to use a complex but more accurate one

pkg/disttask/framework/taskexecutor/execute/interface.go
```diff
@@ -64,18 +64,109 @@ type StepExecutor interface {
...
+const (
+	// UpdateSubtaskSummaryInterval is the interval for updating the subtask summary to
+	// subtask table.
+	UpdateSubtaskSummaryInterval = 3 * time.Second
+
+	// maxProgressInSummary is the number of progress stored in subtask summary
+	maxProgressInSummary = 5
+
+	// SubtaskSpeedUpdateInterval is the interval for updating the subtasks' speed.
+	SubtaskSpeedUpdateInterval = UpdateSubtaskSummaryInterval * maxProgressInSummary / 2
```
D3Hunter: maybe remove the `/ 2` part, then we can average over all `Progresses`, `3 * 5 = 15 seconds` is ok for the time window

pkg/disttask/framework/taskexecutor/execute/interface_test.go
```diff
@@ -0,0 +1,137 @@
...
+			name: "no overlap with data range",
+			setup: func(s *SubtaskSummary) {
+				baseTime := time.Unix(1000, 0)
+				s.Progresses = []Progress{
+					{Bytes: 0, UpdateTime: baseTime},
+					{Bytes: 100, UpdateTime: baseTime.Add(1 * time.Second)},
+				}
+			},
+			endTime:     time.Unix(1010, 0),
+			duration:    time.Second, // [9, 10]
```
D3Hunter, line 55:
```suggestion
			duration:    time.Second, // [1009, 1010]
```
please update below too

pkg/disttask/framework/taskexecutor/execute/interface_test.go
```diff
@@ -0,0 +1,137 @@
...
+			expected:    0,
+			description: "should return 0 when time range doesn't overlap with data",
+		},
+		{
+			name: "partial time range overlap",
```
D3Hunter, line 60: add a case that they are equal at the border, both start and end

D3Hunter (commented):
rest lgtm

D3Hunter (approved):

GMHDBJD (approved):

PR #62694, opened 2025-07-29
planner, executor: support access path `keep order` with `IN` conditions using merge sort

### LINKED ISSUE #34882: `Limit` operator can not be pushed down to tikv when using `IN` clause
## Bug Report

- create a table

```sql
CREATE TABLE `t` (
  `id` bigint(22) NOT NULL AUTO_INCREMENT,
  `a` bigint(20) NOT NULL DEFAULT '0',
  `b` bigint(20) NOT NULL DEFAULT '0',
  `c` bigint(20) NOT NULL DEFAULT '0',
  `d` bigint(20) NOT NULL DEFAULT '0',
  UNIQUE KEY `uk_id` (`id`),
  KEY `a_b_c_id` (`a`,`b`,`c`,`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin AUTO_INCREMENT=1;
```

- perform a query

```sql
explain
SELECT id, a, b, c, d
FROM t
WHERE a = 0 AND b = 1 AND c IN (2, 3) AND d IN (4, 5)
ORDER BY id ASC
LIMIT 10;
```

The `Limit` operator be pushed down to `tikv`'s each range, then `tidb` perform a final `TopN`, something like

```sql
mysql> explain SELECT id, a, b, c, d FROM t WHERE a = 0 AND b = 1 AND c IN (2, 3) ORDER BY id ASC LIMIT 10;
+----------------------------------+---------+-----------+--------------------------------------+--------------------------------------------------------------------+
| id                               | estRows | task      | access object                        | operator info                                                      |
+----------------------------------+---------+-----------+--------------------------------------+--------------------------------------------------------------------+
| TopN_9                           | 0.00    | root      |                                      | test.t.id, offset:0, count:10                                      |
| └─IndexLookUp_20                 | 0.00    | root      |                                      |                                                                    |
|   ├─Limit_19(Build)              | 0.00    | cop[tikv] |                                      | test.t.id, offset:0, count:10                                      |
|   │ └─IndexRangeScan_17          | 0.00    | cop[tikv] | table:t, index:a_b_c_id(a, b, c, id) | range:[0 1 2,0 1 2], [0 1 3,0 1 3], keep order:true, stats:pseudo |
|   └─TableRowIDScan_18(Probe)     | 0.00    | cop[tikv] | table:t                              | keep order:false, stats:pseudo                                     |
+----------------------------------+---------+-----------+--------------------------------------+--------------------------------------------------------------------+
5 rows in set (0.00 sec)
```

The execution plan choose push down `TopN` rather than `Limit`, which results in an unnecessary `Scan + Sort`.

```sql
mysql> explain SELECT id, a, b, c, d FROM t WHERE a = 0 AND b = 1 AND c IN (2, 3) ORDER BY id ASC LIMIT 10;
+----------------------------------+---------+-----------+--------------------------------------+--------------------------------------------------------------------+
| id                               | estRows | task      | access object                        | operator info                                                      |
+----------------------------------+---------+-----------+--------------------------------------+--------------------------------------------------------------------+
| TopN_9                           | 0.00    | root      |                                      | test.t.id, offset:0, count:10                                      |
| └─IndexLookUp_20                 | 0.00    | root      |                                      |                                                                    |
|   ├─TopN_19(Build)               | 0.00    | cop[tikv] |                                      | test.t.id, offset:0, count:10                                      |
|   │ └─IndexRangeScan_17          | 0.00    | cop[tikv] | table:t, index:a_b_c_id(a, b, c, id) | range:[0 1 2,0 1 2], [0 1 3,0 1 3], keep order:false, stats:pseudo |
|   └─TableRowIDScan_18(Probe)     | 0.00    | cop[tikv] | table:t                              | keep order:false, stats:pseudo                                     |
+----------------------------------+---------+-----------+--------------------------------------+--------------------------------------------------------------------+
5 rows in set (0.00 sec)
```

```
Release Version: v6.0.0
Edition: Community
Git Commit Hash: 36a9810441ca0e496cbd22064af274b3be771081
Git Branch: heads/refs/tags/v6.0.0
UTC Build Time: 2022-03-31 10:27:25
GoVersion: go1.18
Race Enabled: false
TiKV Min Version: v3.0.0-60965b006877ca7234adaced7890d7b029ed1306
```

### LINKED ISSUE #63449: Support Access Path `keep order` with `IN` Conditions using Merge Sort
## The problem

Given an index `idx(a, b, c)`, for a query like `WHERE a IN (1, 2, 3) ORDER BY b, c`, the database cannot guarantee a globally sorted order by simply scanning the index. The scan for `a=1` is ordered by `(b, c)`, as is the scan for `a=2 and a=3`, but there is no guarantee that the results from these scans will be globally ordered across all values of `a`.
Consequently, TiDB must abandon the `keep order` optimization. The executor has to scan all rows matching the `IN` condition, and then perform an explicit sort (`TopN` operator) on the result set, which is significantly less performant.

```
SELECT * FROM t
WHERE a IN ('aaa', 'bbb', 'ccc')
  AND b < 1000
ORDER BY b ASC LIMIT 30;

+----------------------------------+---------+---------+-----------+-------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| id                               | estRows | actRows | task      | access object                                                                                   | operator info                                                                                                                                                                                                                                                                                                                                                                          |
+----------------------------------+---------+---------+-----------+-------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| TopN_9                           | 30.00   | 30      | root      |                                                                                                 | test.t.b, offset:0, count:30                                                                                                                                                                                                                                                                                                              |
| └─IndexLookUp_16                | 30.00   | 30      | root      |                                                                                                 |                                                                                                                                                                                                                                                                                                                                                                                        |
|   ├─TopN_15(Build)              | 30.00   | 30      | cop[tikv] |                                                                                                 | test.t.b, offset:0, count:30                                                                                                                                                                                                                                                                                                              |
|   │ └─IndexRangeScan_13        | 1317.05 | 1265    | cop[tikv] | table:t, index:idx_2(a, b, c, d) | range:["aaa" -inf,"aaa" 1000), ["bbb" -inf,"bbb" 1000), ["ccc" -inf,"ccc" 1000), keep order:false |
|   └─TableRowIDScan_14(Probe)    | 30.00   | 30      | cop[tikv] | table:t                                                                        | keep order:false                                                                                                                                                                                                                                                                                                                                                                       |
+----------------------------------+---------+---------+-----------+-------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```
Currently, the only workaround is for users to manually rewrite the query using UNION ALL to simulate the desired execution pattern, which is cumbersome and impractical for long IN lists:
```sql
(SELECT * FROM t WHERE a = 1 ORDER BY b, c LIMIT 30)
UNION ALL
(SELECT * FROM t WHERE a = 2 ORDER BY b, c LIMIT 30)
UNION ALL
(SELECT * FROM t WHERE a = 3 ORDER BY b, c LIMIT 30)
ORDER BY b, c LIMIT 30;
```
This project aims to automate this optimization within TiDB, making it transparent to the user.

## Proposal: Merge Sort Optimization

The core idea is to recognize that while a query with an `IN` condition is not globally ordered, it can be decomposed into a set of sub-problems, each of which is ordered. For `WHERE a IN (1, 2, 3) ORDER BY b, c`, we can think of this as three independent, ordered streams of data: one for `a=1`, one for `a=2`, and one for `a=3`.
We propose to enhance the optimizer to perform this decomposition automatically. The execution plan will be modified to:
1. Split the `IndexRangeScan` based on the values in the IN list.
2. For each value, perform a scan that leverages the `keep order:true` optimization.
3. Use a "merge sort" operation that consumes these multiple, individually sorted streams and merges them into a single, globally sorted result set.

There is good news that we have an existing infrastructure that can be leveraged: when a query accesses multiple partitions of a partitioned table, if the data needs to be sorted and the index matches this order (each partition's index is independent, meaning the data within each partition is ordered, but there is no guarantee of order between partitions), it can use a merge sort to ensure that all data across partitions conforms to the order.
For example, in the following case, we can use keep order to terminate early.
```sql
create table t(a int, b int, c int, d int, index iabc(a,b,c)
) partition by range(d) (partition p0 values less than (100), partition p1 values less than(200));

explain select * from t where a<1 order by a,b,c limit 10;
```
```
+----------------------------------+---------+-----------+------------------------------+-----------------------------------------------+
| id                               | estRows | task      | access object                | operator info                                 |
+----------------------------------+---------+-----------+------------------------------+-----------------------------------------------+
| Projection_33                    | 10.00   | root      |                              | test.t.a, test.t.b, test.t.c, test.t.d        |
| └─IndexLookUp_32                | 10.00   | root      | partition:all                | limit embedded(offset:0, count:10)            |
|   ├─Limit_31(Build)             | 10.00   | cop[tikv] |                              | offset:0, count:10                            |
|   │ └─IndexRangeScan_29        | 10.00   | cop[tikv] | table:t, index:iabc(a, b, c) | range:[-inf,1), keep order:true, stats:pseudo |
|   └─TableRowIDScan_30(Probe)    | 10.00   | cop[tikv] | table:t                      | keep order:false, stats:pseudo                |
+----------------------------------+---------+-----------+------------------------------+-----------------------------------------------+
```
This is a very similar requirement to the existing functionality for partitioned tables. Moreover, the infrastructure used for partitioned tables (`sortedSelectResults`) can directly merge multiple `distsql` layer result sets (`SelectResult`) into a globally ordered result set using a built-in merge sort feature. We can leverage this same infrastructure to achieve the desired functionality.

## Detailed design

### Extending `isMatchProp()`
The `isMatchProp()` function will be extended to support this new optimization.
- When matching index columns against the `ORDER BY` property, if an index prefix column is not present in the `ORDER BY` list, the function will perform a new check: for this column, is it an equality condition within each range? (i.e., `lowerBound == upperBound`).
- If this condition holds for all ranges, it means we can apply the merge-sort strategy. The function's return value will be changed from a simple boolean to an enum to signify this new state, e.g., `KeepOrderByMergeSort`.
- This process will generate two crucial artifacts, which will be passed to and stored in the `AccessPath`, the physical operator and the executor. These `struct`s will add the two new fields.
  - `GroupedRanges [][]*ranger.Range`: A two-dimensional slice that contains the original ranges, now grouped by the values of the equality-condition columns.
  - `GroupByColIdxs []int`: A slice of integers storing the column indices (within the range) that were used for grouping. This is essential for rebuilding the groups from the ranges. We'll mention it later.

### Executor Part Implementation
The executors responsible for data access (`IndexReaderExecutor`, `IndexLookUpReaderExecutor`, `TableReaderExecutor`) will be modified to handle the new `GroupedRanges` property.
The code complexity in this area is high, requiring much effort during development. In the past, several different features were introduced by different people, leading to a situation where the implementation of this similar logic varies greatly among the different executors, i.e., `IndexReader`, `IndexLookUpReader`, and `TableReader`. This leads to a need for extensive time to clarify and implement them separately, and the code is also difficult to reuse.
So the specific logic for this part will not be detailed here but will be explained in the PR and code comments. The core logic is to treat each group in `GroupedRanges` as a separate task that produces a sorted result set. These multiple result sets are then fed into a merge sort operation.
This design must also correctly handle cases where this new `IN`-based splitting occurs simultaneously with partition-based splitting. For example, consider the following query on a partitioned table:
```sql
CREATE TABLE tp (
  a INT, b INT, c INT, d INT,
  PRIMARY KEY (a, b, d)
) PARTITION BY RANGE(d) (
  PARTITION p0 VALUES LESS THAN (100),
  PARTITION p1 VALUES LESS THAN (200)
);

SELECT d FROM tp WHERE a = 1 AND b IN (1, 3, 5) ORDER BY d;
```
Here, the ranges must be split by both the partition key (`d`) and the `IN` condition on `b`. The executor will generate six sorted streams to merge:
- Partition `p0`: `a = 1 and b = 1`
- Partition `p0`: `a = 1 and b = 3`
- Partition `p0`: `a = 1 and b = 5`
- Partition `p1`: `a = 1 and b = 1`
- Partition `p1`: `a = 1 and b = 3`
- Partition `p1`: `a = 1 and b = 5`

The merge-sort logic will be generalized to handle this combination of streams seamlessly.


### Handling Plan Cache and Correlated Subqueries
Essentially, the `GroupedRanges` field is a derivative of the `ranges`. When `ranges` is updated, the `GroupedRanges` should also be updated accordingly.
A key challenge is ensuring this optimization works with the plan cache and correlated subqueries. In both scenarios, the ranges are rebuilt with new parameter values for each execution. So the `GroupedRanges` must also be rebuilt to reflect the new `ranges`.
In the current plan cache implementation, if complex range merging occurs (e.g., multiple interval ranges merging into a single point range), it will not be cached. Therefore, this scenario is safe for the plan cache. For other cases, while the structure of the new ranges may change with different parameter values, it will not become unable to keep order. In other words, if it can keep order originally, then it can still keep order after rebuilding the range with new constants.
Also, we do not need to re-run the full `matchProperty` logic. We only need to regroup based on the same grouping columns.

Therefore, the solution is to use the `GroupByColIdxs` mentioned earlier. When we rebuild the ranges, we can use these stored column indices to efficiently re-group the newly generated ranges into a new `GroupedRanges` structure. This provides a fast and correct way to update the plan for new executions.

time-and-fate:

Issue Number: close #63449 close #34882

Problem Summary:

- planner
  - As mentioned in the issue, new fields `GroupedRanges` and `GroupByColIdxs` are added to `AccessPath`, `PhysicalIndexScan`, and `PhysicalTableScan`. (`pkg/planner/core/operator/physicalop/*`)
  - `matchProperty()` (`pkg/planner/core/find_best_task.go`)
    - Add the new logic to match the sort property as described in the issue.
  - Due to changes in `matchProperty()`, the names and data types of several other fields and the function return value are also modified correspondingly, like `candidatePath.isMatchProp`, `matchPropForIndexMergeAlternatives()`. The checking logic is also updated correspondingly. (`pkg/planner/core/find_best_task.go`)
  - When it's `PropMatchedNeedMergeSort`, set up `ByItems`, `GroupedRanges`, and `GroupByColIdxs` for `PhysicalIndexScan` and `PhysicalTableScan`. (`convertToIndexScan()`, `convertToTableScan()`)
  - New function `GroupRangesByCols()` that builds/rebuilds the `GroupedRanges` from the `GroupByColIdxs` and `ranges`. (`pkg/planner/core/find_best_task.go`)
  - Enum `PhysicalPropMatchResult` (`pkg/planner/property/physical_property.go`)
    - `PropMatchedNeedMergeSort`: means the newly introduced enhancement.
  - When rebuilding ranges for the plan cache, also rebuild the `GroupedRanges`. (`pkg/planner/core/plan_cache_rebuild.go`)
- executor
  - In `Open()` of the 3 executors, when there are correlated columns, also use `GroupRangesByCols` to rebuild grouped ranges after rebuilding ranges.
  - `IndexReaderExecutor`
    - New field `groupedRanges [][]*ranger.Range`, directly comes from `PhysicalIndexScan`.
    - Update `Open()` to consider the new case when building kv ranges.
  - `IndexLookUpExecutor`
    - New field `groupedRanges [][]*ranger.Range`, directly comes from `PhysicalIndexScan`.
    - New field `groupedKVRanges []*kvRangesWithPhysicalTblID` as a unified place to store the kv ranges. It replaces the previous `partitionKVRanges` and `kvRanges` field.
    - Update `buildTableKeyRanges()` to consider the new case when building kv ranges.
  - `TableReaderExecutor`
    - New field `groupedRanges` and `groupByColIdxs`, directly come from `PhysicalTableScan`.
    - Update `Open()` to consider the new case when building kv ranges.
    - `buildResp()` is updated and renamed to `buildRespForGroupedRanges()` as a unified method to handle ranges and grouped ranges.
  - `mem*Reader`
    - Correctly set up `usedIndex` for the extra sorting (mem reader doesn't support merge sort). (`pkg/executor/mem_reader.go`)
    - When `usedIndex` is set, it means we need extra sorting. (`pkg/executor/mem_reader.go`)
    - `memIndexLookUpReader`
      - New field `groupedKVRanges []*kvRangesWithPhysicalTblID` as a unified place to store the kv ranges. It replaces the previous `partitionKVRanges`, `partitionTables`, and `partitionMode` fields. It corresponds to the changes in `IndexLookUpExecutor`.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/builder.go
```diff
@@ -1506,6 +1506,12 @@ func (b *executorBuilder) buildUnionScanFromReader(reader exec.Executor, v *phys
 	case *TableReaderExecutor:
 		us.desc = x.desc
 		us.keepOrder = x.keepOrder
+		colIdxes, err := collectColIdxFromByItems(x.byItems, x.columns)
+		if err != nil {
+			b.err = err
+			return nil
+		}
+		us.usedIndex = colIdxes
```
wshwsh12, line 1514: Why TableReader need set us.usedIndex?

time-and-fate: The core problem is that the merge sort operation is only supported in `Table/Index/IndexLookUpReaderExecutor`, but is not supported in `UnionScan`.
In the current implementation, the solution is to use an extra sorting in `UnionScan` to satisfy it, which is introduced in https://github.com/pingcap/tidb/pull/45140.

When the extra sorting is needed, the `usedIndex` should be set. But if it's not set, it will implicitly use the `Handle` as the sort key (see `(compareExec).compare()`), so it's not needed for `TableReader` previously. But in this PR, we may require order that is not the same as the `Handle`, so this becomes necessary.

time-and-fate: I added a new flag to make it more explicit and add some comments to try to make it more clear.

pkg/planner/core/find_best_task.go
```diff
@@ -1245,39 +1249,122 @@ func matchProperty(ds *logicalop.DataSource, path *util.AccessPath, prop *proper
...
+			path.GroupedRanges = groups
+			path.GroupByColIdxs = groupByColIdxs
+			return property.PropMatchedNeedMergeSort
+		}
+	}
 	return matchResult
 }
 
+// GroupRangesByCols groups the ranges by the values of the columns specified by groupByColIdxs.
+func GroupRangesByCols(ranges []*ranger.Range, groupByColIdxs []int) ([][]*ranger.Range, error) {
```
winoros, line 1253: The given range should be sorted.
I think we may not need to call the `codec.EncodeValue` to do the grouping.

PR #62695, opened 2025-07-29
planner: handle histogram last bucket end value underrepresented

### LINKED ISSUE #62665: Equal predicate at end of range or out of range does not recognize modifyCount
## Bug Report

The following is a (significantly) simplified version of an issue seen at a customer. The customer's example was a date column, but this example is simplified to use INT.

Statistics were collected early in the day, and during the day - new rows were inserted with today's date (100,000s of rows). But since ANALYZE had not been executed again - the estimated number of rows was lower than the actual number of rows.

```
-- disable auto analyze (to ensure that analyze does not automatically correct the issue)
set @@global.tidb_enable_auto_analyze=OFF;

CREATE TABLE `t1` (
  `a` int,
  `b` int,
  `c` int,
  KEY `ia` (`a`)
);
set @@cte_max_recursion_depth=100000;
INSERT INTO t1 (a, b, c)
SELECT mod(a, 150) as a, mod(a, 1000) AS b, mod(a, 10) AS c
FROM (
    WITH RECURSIVE x AS (
        SELECT 1 AS a
        UNION ALL
        SELECT a + 1 AS a
        FROM x
        WHERE a < 10000
    )
    SELECT a
    FROM x
) AS subquery;
-- check the max values for each column
 select max(a), max(b), max(c) from t1;
+--------+--------+--------+
| max(a) | max(b) | max(c) |
+--------+--------+--------+
|    149 |    999 |      9 |
+--------+--------+--------+
-- insert a single new row as the last row in the table
Insert into t1 values (150, 1000, 10);
-- collect statistics
Analyze table t1;
-- insert 600+ new rows with the same maximum values as the single row
Insert into t1 select 150, 100, 10 from t1 where a < 10;

show stats_meta;
+---------+------------+----------------+---------------------+--------------+-----------+---------------------+
| Db_name | Table_name | Partition_name | Update_time         | Modify_count | Row_count | Last_analyze_time   |
+---------+------------+----------------+---------------------+--------------+-----------+---------------------+
| test    | t1         |                | 2025-07-28 14:57:53 |          669 |     10670 | 2025-07-28 14:53:23 |
+---------+------------+----------------+---------------------+--------------+-----------+---------------------+
```

Explain of the query should recognize that there have been rows inserted. Given that the count for this value (using reproduce steps above) shows 669 rows.
```
 select count(*) from t1 where a = 150;
+----------+
| count(*) |
+----------+
|      669 |
+----------+
```

The explain output is similar to before the rows were inserted
```
tidb> explain select * from t1 where a = 150;
+--------------------------------+---------+-----------+-----------------------+-----------------------------------+
| id                             | estRows | task      | access object         | operator info                     |
+--------------------------------+---------+-----------+-----------------------+-----------------------------------+
| IndexLookUp_11                 | 1.07    | root      |                       |                                   |
| ├─IndexRangeScan_9(Build)      | 1.07    | cop[tikv] | table:t1, index:ia(a) | range:[150,150], keep order:false |
| └─TableRowIDScan_10(Probe)     | 1.07    | cop[tikv] | table:t1              | keep order:false                  |
+--------------------------------+---------+-----------+-----------------------+-----------------------------------+
3 rows in set (0.00 sec)
```

Current release 8.5

Tristan1900:

Issue Number: close #62665

Problem Summary:

Like what the linked issue describes, the end value in the last histogram bucket could be underrepresented due to stale stats. This PR tries to detect such case precisely by looking at two conditions:
1.  The value should be in last histogram bucket and it should be the upper bound
2. The last bucket end count is much less than the average count of other NDV, indicating this might be a tail write traffic

After detecting the underrepresented value, instead of passing it back using stale stats from histogram, we reuse the current logic of handling other value by taking the average histogram value count.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
better handle histogram last bucket end value underrepresented issue, tail writes on equal predicate on the latest value  should have better performance even with stale stats 
```

### DIFF
--- a/pkg/planner/cardinality/row_count_column.go
+++ b/pkg/planner/cardinality/row_count_column.go
@@ -168,7 +168,14 @@ func equalRowCountOnColumn(sctx planctx.PlanContext, c *statistics.Column, val t
 	// 2. try to find this value in bucket.Repeat(the last value in every bucket)
 	histCnt, matched := c.Histogram.EqualRowCount(sctx, val, true)
 	if matched {
-		return histCnt, nil
+		// Check if this last bucket end value should fall through to uniform distribution
+		if IsLastBucketEndValueUnderrepresented(&c.Histogram, val, histCnt, realtimeRowCount, modifyCount, sctx) {
+			matched = false
+		}
+
+		if matched {
+			return histCnt, nil
+		}
 	}
 	// 3. use uniform distribution assumption for the rest (even when this value is not covered by the range of stats)
 	// branch1: histDNV <= 0 means that all NDV's are in TopN, and no histograms.
--- a/pkg/planner/cardinality/row_count_column.go
+++ b/pkg/planner/cardinality/row_count_column.go
@@ -167,9 +167,11 @@ func equalRowCountOnColumn(sctx planctx.PlanContext, c *statistics.Column, val t
 	}
 	// 2. try to find this value in bucket.Repeat(the last value in every bucket)
 	histCnt, matched := c.Histogram.EqualRowCount(sctx, val, true)
+	// Calculate histNDV here as it's needed for both the underrepresented check and later calculations
+	histNDV := float64(c.Histogram.NDV - int64(c.TopN.Num()))
 	// also check if this last bucket end value is underrepresented
-	if matched && !IsLastBucketEndValueUnderrepresented(
-		&c.Histogram, val, histCnt, realtimeRowCount, modifyCount, sctx) {
+	if matched && !IsLastBucketEndValueUnderrepresented(sctx,
+		&c.Histogram, val, histCnt, histNDV, realtimeRowCount, modifyCount) {
 		return histCnt, nil
 	}
 	// 3. use uniform distribution assumption for the rest (even when this value is not covered by the range of stats)
@@ -177,7 +179,6 @@ func equalRowCountOnColumn(sctx planctx.PlanContext, c *statistics.Column, val t
 	// branch2: histDNA > 0 basically means while there is still a case, c.Histogram.NDV >
 	// c.TopN.Num() a little bit, but the histogram is still empty. In this case, we should use the branch1 and for the diff
 	// in NDV, it's mainly comes from the NDV is conducted and calculated ahead of sampling.
-	histNDV := float64(c.Histogram.NDV - int64(c.TopN.Num()))
 	if histNDV <= 0 || (c.IsFullLoad() && c.Histogram.NotNullCount() == 0) {
 		// branch 1: all NDV's are in TopN, and no histograms
 		// If histNDV is zero - we have all NDV's in TopN - and no histograms. This function uses
--- a/pkg/planner/cardinality/row_count_column.go
+++ b/pkg/planner/cardinality/row_count_column.go
@@ -167,15 +167,10 @@ func equalRowCountOnColumn(sctx planctx.PlanContext, c *statistics.Column, val t
 	}
 	// 2. try to find this value in bucket.Repeat(the last value in every bucket)
 	histCnt, matched := c.Histogram.EqualRowCount(sctx, val, true)
-	if matched {
-		// Check if this last bucket end value should fall through to uniform distribution
-		if IsLastBucketEndValueUnderrepresented(&c.Histogram, val, histCnt, realtimeRowCount, modifyCount, sctx) {
-			matched = false
-		}
-
-		if matched {
-			return histCnt, nil
-		}
+	// also check if this last bucket end value is underrepresented
+	if matched && !IsLastBucketEndValueUnderrepresented(
+		&c.Histogram, val, histCnt, realtimeRowCount, modifyCount, sctx) {
+		return histCnt, nil
 	}
 	// 3. use uniform distribution assumption for the rest (even when this value is not covered by the range of stats)
 	// branch1: histDNV <= 0 means that all NDV's are in TopN, and no histograms.
--- a/pkg/planner/cardinality/row_count_index.go
+++ b/pkg/planner/cardinality/row_count_index.go
@@ -421,15 +421,10 @@ func equalRowCountOnIndex(sctx planctx.PlanContext, idx *statistics.Index, b []b
 	}
 	// 2. try to find this value in bucket.Repeat(the last value in every bucket)
 	histCnt, matched := idx.Histogram.EqualRowCount(sctx, val, true)
-	if matched {
-		// Check if this last bucket end value should fall through to uniform distribution
-		if IsLastBucketEndValueUnderrepresented(&idx.Histogram, val, histCnt, realtimeRowCount, modifyCount, sctx) {
-			matched = false
-		}
-
-		if matched {
-			return histCnt
-		}
+	// also check if this last bucket end value is underrepresented
+	if matched && !IsLastBucketEndValueUnderrepresented(
+		&idx.Histogram, val, histCnt, realtimeRowCount, modifyCount, sctx) {
+		return histCnt
 	}
 	// 3. use uniform distribution assumption for the rest (even when this value is not covered by the range of stats)
 	histNDV := float64(idx.Histogram.NDV - int64(idx.TopN.Num()))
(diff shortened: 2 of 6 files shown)


Tristan1900:
Manual test using steps in linked issue description verified fixed
```
tidb> Show stats_meta;
+---------+------------+----------------+---------------------+--------------+-----------+---------------------+
| Db_name | Table_name | Partition_name | Update_time         | Modify_count | Row_count | Last_analyze_time   |
+---------+------------+----------------+---------------------+--------------+-----------+---------------------+
| test    | t1         |                | 2025-07-30 11:28:56 |        10670 |     20671 | 2025-07-30 11:28:09 |
+---------+------------+----------------+---------------------+--------------+-----------+---------------------+
1 row in set (0.01 sec)

tidb> Explain select * from t1 where a = 150;
+--------------------------------+---------+-----------+-----------------------+-----------------------------------+
| id                             | estRows | task      | access object         | operator info                     |
+--------------------------------+---------+-----------+-----------------------+-----------------------------------+
| IndexLookUp_11                 | 133.78  | root      |                       |                                   |
| ├─IndexRangeScan_9(Build)      | 133.78  | cop[tikv] | table:t1, index:ia(a) | range:[150,150], keep order:false |
| └─TableRowIDScan_10(Probe)     | 133.78  | cop[tikv] | table:t1              | keep order:false                  |
+--------------------------------+---------+-----------+-----------------------+-----------------------------------+
3 rows in set (0.00 sec)
tidb> Explain select * from t1 where a = 149;
+--------------------------------+---------+-----------+-----------------------+-----------------------------------+
| id                             | estRows | task      | access object         | operator info                     |
+--------------------------------+---------+-----------+-----------------------+-----------------------------------+
| IndexLookUp_11                 | 136.41  | root      |                       |                                   |
| ├─IndexRangeScan_9(Build)      | 136.41  | cop[tikv] | table:t1, index:ia(a) | range:[149,149], keep order:false |
| └─TableRowIDScan_10(Probe)     | 136.41  | cop[tikv] | table:t1              | keep order:false                  |
+--------------------------------+---------+-----------+-----------------------+-----------------------------------+
3 rows in set (0.00 sec)
```

pkg/planner/cardinality/row_count_index.go
```diff
@@ -422,7 +422,14 @@ func equalRowCountOnIndex(sctx planctx.PlanContext, idx *statistics.Index, b []b
 	// 2. try to find this value in bucket.Repeat(the last value in every bucket)
 	histCnt, matched := idx.Histogram.EqualRowCount(sctx, val, true)
 	if matched {
-		return histCnt
+		// Check if this last bucket end value should fall through to uniform distribution
+		if IsLastBucketEndValueUnderrepresented(&idx.Histogram, val, histCnt, realtimeRowCount, modifyCount, sctx) {
```
terry1purcell: sctx is usually first in the list of variables

pkg/planner/cardinality/row_count_column.go
```diff
@@ -168,7 +168,14 @@ func equalRowCountOnColumn(sctx planctx.PlanContext, c *statistics.Column, val t
 	// 2. try to find this value in bucket.Repeat(the last value in every bucket)
 	histCnt, matched := c.Histogram.EqualRowCount(sctx, val, true)
 	if matched {
-		return histCnt, nil
+		// Check if this last bucket end value should fall through to uniform distribution
+		if IsLastBucketEndValueUnderrepresented(&c.Histogram, val, histCnt, realtimeRowCount, modifyCount, sctx) {
```
terry1purcell: sctx is usually first in list of variables

pkg/planner/cardinality/selectivity.go
```diff
@@ -592,6 +592,67 @@ func compareType(l, r int) int {
...
+	originalNonNullCount := hg.NotNullCount()
+
+	// Only apply heuristic to tables with sufficient size to avoid false positives
+	// on very small tables where a few new rows create misleadingly high growth ratios
+	if originalNonNullCount < MinRowCountForStaleHeuristic {
+		return false
+	}
+
+	// This represents new rows added since ANALYZE
+	newRowsAdded := realtimeRowCount - int64(hg.TotalRowCount())
```
terry1purcell: outOfRangeRowCount uses absolute function to calculate added rows such as:
addedRows := math.Abs(float64(realtimeRowCount) - hg.TotalRowCount())
Should we be consistent with that function?

pkg/planner/cardinality/selectivity.go
```diff
@@ -592,6 +592,67 @@ func compareType(l, r int) int {
...
+	if newRowsAdded <= 0 {
+		return false
+	}
+
+	// Only apply heuristic when new rows are significant relative to average bucket size
+	// Use non-null count for bucket size since buckets contain non-null distributions
+	avgBucketSize := originalNonNullCount / float64(len(hg.Buckets))
```
terry1purcell: This should use histNDV rather than len(hg.Buckets) - histNDV is calculated in row_count_index.go (immediately after this function is called) as:
histNDV := float64(idx.Histogram.NDV - int64(idx.TopN.Num()))
This should calculate the avgRows := orginalNonNullCount / histNDV, meaning histNDV would need to be passed in.

pkg/planner/cardinality/selectivity.go
```diff
@@ -592,6 +592,67 @@ func compareType(l, r int) int {
...
+
+	// Check if this is the last bucket's upper bound value (end value)
+	isLastBucketEndValue := (bucketIdx == len(hg.Buckets)-1) && inBucket && matchLastValue
+	if !isLastBucketEndValue {
+		return false
+	}
+
+	// Check if the count is suspiciously low compared to other buckets
+	// Use same non-null count basis as bucket size calculation for consistency
+	avgBucketCount := originalNonNullCount / float64(len(hg.Buckets))
```
qw4990: `avgBucketCount` and `avgBucketSize` seem the same, why don't we just use `avgBucketSize` here?

pkg/planner/cardinality/row_count_column.go
```diff
@@ -168,7 +168,14 @@ func equalRowCountOnColumn(sctx planctx.PlanContext, c *statistics.Column, val t
...
+			matched = false
+		}
+
+		if matched {
+			return histCnt, nil
+		}
```
qw4990: How about `if matched && IsLastBucketEndValueUnderrepresented(&c.Histogram, val, histCnt, realtimeRowCount, modifyCount, sctx) {return histCnt, nil}`?

pkg/planner/cardinality/selectivity.go
```diff
@@ -592,6 +592,67 @@ func compareType(l, r int) int {
...
+const bucketAwareRowAddedThreshold = 0.5
+
+// MinRowCountForStaleHeuristic is the minimum original row count required to apply
+// the stale bucket heuristic. This prevents the heuristic from triggering on very
+// small tables where a few new rows would create misleadingly high growth ratios.
+const MinRowCountForStaleHeuristic = 100
+
+// IsLastBucketEndValueUnderrepresented detects when the last value (upper bound) of the last bucket
+// has a suspiciously low count that may be stale due to concentrated writes after ANALYZE.
+func IsLastBucketEndValueUnderrepresented(hg *statistics.Histogram, val types.Datum,
```
qw4990: Nice work, very clear and reasonable!

Tristan1900: thanks for the quick review! I have addressed comments, can you take a quick look again when you get a chance

pkg/planner/cardinality/selectivity.go
Tristan1900: sounds good, I made that a method so can used by both places now

pkg/planner/cardinality/selectivity.go
Tristan1900: good call!

terry1purcell (approved):
/lgtm

PR #62714, opened 2025-07-30
parser: Add hint `INDEX_LOOKUP_PUSH_DOWN`

### LINKED ISSUE #62713: Add hint `INDEX_LOOKUP_PUSH_DOWN` to parser

lcwangchao:

Issue Number: close #62713

- Add a new hint `INDEX_LOOKUP_PUSH_DOWN`
- Because the IndexLookUp push-down feature is still in development, we should not expose this hint to the user. This PR also disables it in production.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/util/hint/hint.go
```diff
@@ -833,6 +841,11 @@ func ParsePlanHints(hints []*ast.TableOptimizerHint,
 				hintType = ast.HintOrderIndex
 			case HintNoOrderIndex:
 				hintType = ast.HintNoOrderIndex
+			case HintIndexLookUpPushDown:
+				if !EnableIndexLookUpPushDownForTest.Load() {
+					warnHandler.SetHintWarningFromError(parser.ErrWarnOptimizerHintUnsupportedHint.FastGenByArgs(hint.HintName.O))
```
lcwangchao, line 856: Before the feature is finished to develop, `EnableIndexLookUpPushDownForTest` should always be kept `false` to make sure the production build cannot see this hint. We have a test to cover this case: https://github.com/pingcap/tidb/pull/62714/files#diff-c749a37e35e3e853f8a574e6c1b415a95a6e30ca01ae96c2d5a699585c1e4c88R244

pkg/planner/core/operator/physicalop/physical_indexlookup.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func resetPlanIDRecursively(ctx base.PlanContext, p base.PhysicalPlan) {
+	p.SetID(int(ctx.GetSessionVars().PlanID.Add(1)))
+	for _, child := range p.Children() {
+		resetPlanIDRecursively(ctx, child)
+	}
+}
+
+func buildPushDownIndexLookUpPlan(
```
cfzjywxk, line 47:
```suggestion
func tryBuildPushDownIndexLookUpPlan(
```
Looks consistent with the caller function.

pkg/planner/core/operator/physicalop/physical_indexlookup.go
```diff
@@ -0,0 +1,172 @@
...
+	ctx base.PlanContext, indexPlan base.PhysicalPlan, tablePlan base.PhysicalPlan, keepOrder bool,
+) (indexLookUpPlan base.PhysicalPlan, err error) {
+	if keepOrder {
+		return nil, errors.NewNoStackError("keep order is not supported")
+	}
+
+	tablePlan, err = tablePlan.Clone(ctx)
```
cfzjywxk, line 50: Could we `Clone` after the pre-check has finished successfully?

pkg/planner/core/operator/physicalop/physical_indexlookup.go
lcwangchao: I've moved the most hint checks to `getPossibleAccessPaths` see https://github.com/pingcap/tidb/pull/62714/commits/a2b69abf06f4dbbcbab1f3b4dc929186ecdc9d82

So the `buildPushDownIndexLookUpPlan` should always be successful. I think the name is appropriate now

pkg/planner/core/operator/physicalop/physical_indexlookup.go
lcwangchao: The pre-checks have been moved out of this function.

pkg/planner/core/planbuilder.go
```diff
@@ -1309,6 +1338,12 @@ func getPossibleAccessPaths(ctx base.PlanContext, tableHints *hint.PlanHints, in
 			if path := getTablePath(publicPaths); path != nil {
 				hasUseOrForce = true
 				path.Forced = true
+				if i >= indexHintsLen {
```
cfzjywxk: Better to have some checks or test case to ensure the implicit constaint `path.IsIndexLookUpPushDown = indexLookUpPushDown[i]` is hold, and add some comments here too.

lcwangchao: I've added some comments here. But the tests may be added in another PR because in the current PR the hint is disabled and cannot be tested in integration MySQL tests.

The tests may like this: https://github.com/lcwangchao/tidb/blob/executor_indexlookup_pushdown/tests/integrationtest/t/executor/index_lookup_pushdown.test

pkg/planner/core/operator/physicalop/physical_utils.go
```diff
@@ -51,26 +51,74 @@ func ClonePhysicalPlan(sctx base.PlanContext, plans []base.PhysicalPlan) ([]base
...
+		plans = flattenPlanWithPreorderTraversal(child, plans)
 	}
 	return plans
 }
 
-// FlattenPushDownPlan converts a plan tree to a list, whose head is the leaf node like table scan.
-func FlattenPushDownPlan(p base.PhysicalPlan) []base.PhysicalPlan {
+// FlattenListOrTiFlashPushDownPlan converts a plan tree to a list, whose head is the leaf node like table scan.
+// It is used to flatten the plan tree for TiFlash or TiKV plan true that all parents have only one child.
+func FlattenListOrTiFlashPushDownPlan(p base.PhysicalPlan) []base.PhysicalPlan {
```
cfzjywxk: Why adding `TiFLash` to the function name?

pkg/planner/core/operator/physicalop/physical_indexlookup.go
```diff
@@ -0,0 +1,160 @@
...
+func buildPushDownIndexLookUpPlan(
+	ctx base.PlanContext, indexPlan base.PhysicalPlan, tablePlan base.PhysicalPlan,
+) (indexLookUpPlan base.PhysicalPlan, err error) {
+	tablePlan, err = tablePlan.Clone(ctx)
+	if err != nil {
+		return nil, err
+	}
+	resetPlanIDRecursively(ctx, tablePlan)
+
+	tableScanPlan, tableScanParent, err := detachRootTableScanPlan(tablePlan)
```
cfzjywxk:
```suggestion
	tableScanPlan, parentOfTableScan, err := detachRootTableScanPlan(tablePlan)
```
is more understandable.

pkg/planner/core/operator/physicalop/physical_indexlookup_reader.go
```diff
@@ -205,14 +222,51 @@ func (p *PhysicalIndexLookUpReader) ExplainInfo() string {
...
+	}
+	p.TablePlans = FlattenListOrTiFlashPushDownPlan(p.TablePlan)
+	p.IndexPlans, p.IndexPlansUnNatureOrders = FlattenTreePushDownPlan(p.IndexPlan)
 	return &p
 }
 
+// tryPushDownLookUp tries to push down the index lookup to TiKV.
+func (p *PhysicalIndexLookUpReader) tryPushDownLookUp(ctx base.PlanContext) {
+	intest.Assert(!p.IndexLookUpPushDown)
+	if p.KeepOrder {
```
cfzjywxk, line 240: Could it be included in `checkIndexLookUpPushDownSupported`?

pkg/planner/core/operator/physicalop/physical_utils.go
lcwangchao: Because this function may also used by some tiflash cases:

https://github.com/pingcap/tidb/blob/8f24f95670be0d8180647fad5b000394b88304be/pkg/planner/core/runtime_filter_generator.go#L71-L73

pkg/planner/core/operator/physicalop/physical_indexlookup_reader.go
lcwangchao: I think it's hard to do this because keepOrder cannot be determined in `checkIndexLookUpPushDownSupported`. And some logic plan may result in multiple plans, with some keeping order and some not.  In the future, if we support the keep order in the push down case, this checks can be removed.

pkg/util/hint/hint.go
```diff
@@ -201,6 +205,10 @@ const (
 	PreferTiFlash
 )
 
+// EnableIndexLookUpPushDownForTest is only used for testing to enable index lookup push down.
+// TODO: remove it when index lookup push down finished to dev
+var EnableIndexLookUpPushDownForTest atomic.Bool
```
winoros, line 210: Why do we need it now that it can only be triggered by the hint?

lcwangchao: Because the execution is not implemented in this PR, we need it to make sure the hint is still disabled in relase build. After the execution is implemented, this flag should be removed.

pkg/planner/core/operator/physicalop/physical_utils.go
winoros: Name it `List` is enough.
TiFlash MPP plan is not a list either.

pkg/planner/core/task.go
```diff
@@ -641,9 +641,8 @@ func buildIndexLookUpTask(ctx base.PlanContext, t *CopTask) *RootTask {
 		CommonHandleCols: t.commonHandleCols,
 		ExpectedCnt:      t.expectCnt,
 		KeepOrder:        t.keepOrder,
-	}.Init(ctx, t.tablePlan.QueryBlockOffset())
-	p.PlanPartInfo = t.physPlanPartInfo
-	p.SetStats(t.tablePlan.StatsInfo())
```
winoros, line 646: The stats is missed here.

pkg/planner/core/operator/physicalop/physical_indexlookup.go
```diff
@@ -0,0 +1,163 @@
...
+	// runtimeFilterList has tag `plan-cache-clone:"must-nil"`
+	if len(tableScanPlan.runtimeFilterList) == 0 {
+		tableScanPlan.runtimeFilterList = nil
+	}
+
+	// to make sure tableScanPlan can be cloned,
+	// UsedColumnarIndexes has tag `plan-cache-clone:"must-nil"`
+	if len(tableScanPlan.UsedColumnarIndexes) == 0 {
+		tableScanPlan.UsedColumnarIndexes = nil
+	}
```
winoros: These are attributes of tiflash scan(or the future tici scan).
And actually, you will see that the Clone method will throw FALSE when it's a tiflash scan.
So you can remove the check.

pkg/planner/core/operator/physicalop/physical_indexlookup.go
```diff
@@ -0,0 +1,163 @@
...
+				)
+			}
+			rootParent = currentParent
+			break
+		} else if l > 1 {
+			return nil, nil, errors.Errorf(
+				"the tablePlan is expected to have one child for each node, but %T has multiple children: %T, %T",
+				p, children[0], children[1],
+			)
+		}
```
winoros: These error checks are guaranteed.
You can change it to an intest.

pkg/planner/core/operator/physicalop/physical_indexlookup_reader.go
```diff
@@ -205,14 +222,52 @@ func (p *PhysicalIndexLookUpReader) ExplainInfo() string {
...
+	// So we just use the row count as 0 of tablePlan in TiDB side which displays all lookup
+	// can be performed in the TiKV side.
+	scalePlanStatusInfoAsZeroRecursively(ctx.GetSessionVars(), p.TablePlan)
+	// The status info of IndexLookupReader should be the same as indexPlan in the push-down mode if
+	// all lookup can be performed in the TiKV side.
+	p.SetStats(p.IndexPlan.StatsInfo())
+	p.IndexLookUpPushDown = true
+}
+
+func scalePlanStatusInfoAsZeroRecursively(vars *variable.SessionVars, p base.PhysicalPlan) {
```
winoros:
```suggestion
func resetRowCountAsZeroRecursively(vars *variable.SessionVars, p base.PhysicalPlan) {
```
Though we may not need this function.
You can change its name first. `Stats` is short for `statistics`, not `status`.

pkg/planner/core/operator/physicalop/physical_indexlookup.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/pingcap/tidb/pkg/util/plancodec"
+	"github.com/pingcap/tidb/pkg/util/size"
+	"github.com/pingcap/tipb/go-tipb"
+)
+
+// PhysicalIndexLookUp represents a index lookup
+// It contains two parts: indexPlan and tablePlan
+// IndexPlan maybe hierarchical, e.g. IndexScan -> Selection -> Limit
+// TablePlan contains only TableScan executor
+type PhysicalIndexLookUp struct {
```
winoros:
```suggestion
type PhysicaLocallIndexLookUp struct {
```
should be better.

pkg/planner/core/operator/physicalop/physical_indexlookup_reader.go
```diff
@@ -205,14 +222,52 @@ func (p *PhysicalIndexLookUpReader) ExplainInfo() string {
 }
 
 // Init initializes PhysicalIndexLookUpReader.
-func (p PhysicalIndexLookUpReader) Init(ctx base.PlanContext, offset int) *PhysicalIndexLookUpReader {
+func (p PhysicalIndexLookUpReader) Init(ctx base.PlanContext, offset int, tryPushDownIndexLookUp bool) *PhysicalIndexLookUpReader {
 	p.BasePhysicalPlan = NewBasePhysicalPlan(ctx, plancodec.TypeIndexLookUp, &p, offset)
```
winoros, line 226:
```suggestion
	p.BasePhysicalPlan = NewBasePhysicalPlan(ctx, plancodec.TypeLocalIndexLookUp, &p, offset)
```
And use a new plan name.

pkg/planner/core/task.go
lcwangchao: The `SetStats` has been moved to `Init` : https://github.com/pingcap/tidb/pull/62714/files#diff-c96dcc05159b94e62e0d8953f2fc054c3fdfe63820320764b15888ece2364516R228

pkg/planner/core/operator/physicalop/physical_indexlookup.go
lcwangchao: I added this check because when PhysicalScan is cloned, `runtimeFilterList` will be forced to set to a non-nil value (even if the original value is nil):

https://github.com/pingcap/tidb/blob/71caddf56d9c1887650568101590417558149e3f/pkg/planner/core/operator/physicalop/physical_table_scan.go#L282-L285

But in `CloneForPlanCache`, it will just check `op.runtimeFilterList != nil` 

https://github.com/pingcap/tidb/blob/71caddf56d9c1887650568101590417558149e3f/pkg/planner/core/operator/physicalop/plan_clone_generated.go#L117-L119

It makes a cloned plan can not be cloned again.

I don't know why we should set the `runtimeFilterList` a non-nil value in the first step. Could we just keep nil if it is not set?

pkg/planner/core/operator/physicalop/physical_indexlookup_reader.go
cfzjywxk: Better to add comments about it.

winoros (approved):

cfzjywxk (approved):

bb7133 (approved):
LGTM

PR #62725, opened 2025-07-30
planner: improve index pseudo-estimation with column stats

ichen032:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/statistics/handle/handletest/handle_test.go
```diff
@@ -1429,7 +1429,7 @@ func TestInitStatsLite(t *testing.T) {
 	require.True(t, colBStats1.IsFullLoad())
 	idxBStats1 := statsTbl2.GetIdx(idxBID)
 	require.True(t, idxBStats1.IsFullLoad())
-	require.True(t, colCStats.IsAllEvicted())
+	require.True(t, colCStats.IsFullLoad())
```
ichen032, line 1432: With the new pseudo estimation, if there are no index stats we check if there are column stats and attempt to use them if they exist, loading them in the process. Originally, only index C stats were being loaded, but not the column stats, but now both are being loaded.

pkg/planner/cardinality/pseudo.go
```diff
@@ -243,3 +245,52 @@ func getPseudoRowCountByColumnRanges(tc types.Context, tableRowCount float64, co
...
+			colID = idxCols[i].UniqueID
+			if statistics.ColumnStatsIsInvalid(coll.GetCol(colID), sctx, coll, colID) {
+				count, err = getPseudoRowCountByColumnRanges(sctx.GetSessionVars().StmtCtx.TypeCtx(), tableRowCount, tmpRan, 0)
+			} else {
+				count, err = GetRowCountByColumnRanges(sctx, coll, colID, tmpRan)
+			}
+			if err != nil {
+				return 0, 0, errors.Trace(err)
+			}
+			selectivity *= count / tableRowCount
```
Tristan1900: nit:  count / tableRowCount calculated twice, can extract to be a variable.

pkg/planner/cardinality/pseudo.go
```diff
@@ -243,3 +245,52 @@ func getPseudoRowCountByColumnRanges(tc types.Context, tableRowCount float64, co
 	}
 	return rowCount, nil
 }
+
+// getPseudoRowCountWithPartialStats calculates the row count if there are no statistics on the index, but there are column stats available.
+func getPseudoRowCountWithPartialStats(sctx planctx.PlanContext, coll *statistics.HistColl, indexRanges []*ranger.Range,
```
Tristan1900: nit: feels like this method can be a in better place than pseudo.go. This file seems to contain methods with only complete guess, however getPseudoRowCountWithPartialStats is potentially using GetRowCountByColumnRanges, maybe it can be in row_count_column.go? Just a thought.

pkg/planner/cardinality/row_count_index.go
```diff
@@ -61,16 +63,26 @@ func GetRowCountByIndexRanges(sctx planctx.PlanContext, coll *statistics.HistCol
...
-		colsLen := -1
-		if idx != nil && idx.Info.Unique {
-			colsLen = len(idx.Info.Columns)
-		}
-		result, err = getPseudoRowCountByIndexRanges(sc.TypeCtx(), indexRanges, float64(coll.RealtimeCount), colsLen)
-		if err == nil && sc.EnableOptimizerCETrace && idx != nil {
-			ceTraceRange(sctx, coll.PhysicalID, colNames, indexRanges, "Index Stats-Pseudo", uint64(result))
+		if hasColumnStats(sctx, coll, idxCols) {
+			// If it is a single column index and there are column stats, use column estimation instead.
+			if len(idxCols) == 1 {
```
Tristan1900: feels like the len(idxCols) == 1 case can also be handled in getPseudoRowCountWithPartialStats, do we need to special handle it here?

Tristan1900 (commented):
reviewing this PR to ramp up myself, just a few comments, feel free to ignore if it doesn't make sense

pkg/planner/cardinality/row_count_index.go
```diff
@@ -627,3 +633,13 @@ func getOrdinalOfRangeCond(sc *stmtctx.StatementContext, ran *ranger.Range) int
 	}
 	return len(ran.LowVal)
 }
+
+// hasColumnStats checks if we have collected stats on any of the given columns.
+func hasColumnStats(sctx planctx.PlanContext, coll *statistics.HistColl, idxCols []*expression.Column) bool {
+	for i := range idxCols {
```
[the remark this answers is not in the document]
ichen032: hasColumnStats is only called if index statistics are invalid, and in this scenario idxCols should not be nil.

pkg/planner/cardinality/row_count_column.go
```diff
@@ -408,3 +409,58 @@ func ColumnEqualRowCount(sctx planctx.PlanContext, t *statistics.Table, value ty
...
+	)
+	totalCount = float64(0)
+	corrCount = float64(0)
+	for _, indexRange := range indexRanges {
+		selectivity := float64(1.0)
+		corrSelectivity := float64(1.0)
+		for i := range indexRange.LowVal {
+			tmpRan[0].LowVal[0] = indexRange.LowVal[i]
+			tmpRan[0].HighVal[0] = indexRange.HighVal[i]
+			tmpRan[0].Collators[0] = indexRange.Collators[0]
```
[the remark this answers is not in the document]
terry1purcell: I don't see anywhere else in our optimizer code where we need to check that collators has at least one value. I do see it in sampling code, but not optimizer.

pkg/planner/cardinality/row_count_column.go
```diff
@@ -408,3 +409,58 @@ func ColumnEqualRowCount(sctx planctx.PlanContext, t *statistics.Table, value ty
...
+			if i == len(indexRange.LowVal)-1 {
+				tmpRan[0].LowExclude = indexRange.LowExclude
+				tmpRan[0].HighExclude = indexRange.HighExclude
+			}
+			colID = idxCols[i].UniqueID
+			if statistics.ColumnStatsIsInvalid(coll.GetCol(colID), sctx, coll, colID) {
+				count, err = getPseudoRowCountByColumnRanges(sctx.GetSessionVars().StmtCtx.TypeCtx(), tableRowCount, tmpRan, 0)
+			} else {
+				count, err = GetRowCountByColumnRanges(sctx, coll, colID, tmpRan)
+			}
```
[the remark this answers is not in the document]
terry1purcell: This seems like a valid recommendation

pkg/planner/cardinality/row_count_index.go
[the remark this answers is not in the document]
terry1purcell: First suggestion seems reasonable - to avoid the caller passing nil. Not sure about the 2nd.

terry1purcell (approved):
/lgtm

pkg/planner/cardinality/row_count_index.go
```diff
@@ -627,3 +633,16 @@ func getOrdinalOfRangeCond(sc *stmtctx.StatementContext, ran *ranger.Range) int
...
 	return len(ran.LowVal)
 }
+
+// hasColumnStats checks if we have collected stats on any of the given columns.
+func hasColumnStats(sctx planctx.PlanContext, coll *statistics.HistColl, idxCols []*expression.Column) bool {
+	if idxCols == nil {
+		return false
+	}
+	for i := range idxCols {
+		if !statistics.ColumnStatsIsInvalid(coll.GetCol(idxCols[i].UniqueID), sctx, coll, idxCols[i].UniqueID) {
```
AilinKid, line 643: !ColumnStatsIsInvalid => return true means 
once we got any index column' stats, the partial check is passed? not all index columns?

ichen032: Yes, as long as any have stats it passes. The assumption was that an estimation partially based on collected stats is still better than a fully pseudo estimation. When only some of the columns have stats, we will use stats where they are available, and then fall back to pseudo-estimation for those without stats.

pkg/planner/cardinality/selectivity_test.go
```diff
@@ -390,6 +390,28 @@ func TestIssue57948(t *testing.T) {
...
+func TestNewIndexWithColumnStats(t *testing.T) {
+	// Test two identical tables where one has no statistics at all, and the other only has column statistics but no index statistics
+	// Test that we supplement index estimation with column stats if there are no index stats available
+	store, _ := testkit.CreateMockStoreAndDomain(t)
+	testKit := testkit.NewTestKit(t, store)
+	testKit.MustExec("use test")
+	testKit.MustExec("drop table if exists t")
+	testKit.MustExec("create table t(a int)")
+	testKit.MustExec("create table t2(a int, index idxa(a))")
+	testKit.MustExec("set @@tidb_analyze_version=2")
```
0xPoe: I guess this is unnecessary.

pkg/planner/cardinality/selectivity_test.go
```diff
@@ -390,6 +390,28 @@ func TestIssue57948(t *testing.T) {
...
+	testKit.MustExec("set @@global.tidb_enable_auto_analyze='OFF'")
+	testKit.MustExec("insert into t select mod(a,250) from (with recursive x as (select 1 as a union all select a + 1 AS a from x where a < 500) select a from x) as subquery")
+	testKit.MustExec("insert into t2 select mod(a,250) from (with recursive x as (select 1 as a union all select a + 1 AS a from x where a < 500) select a from x) as subquery")
+	testKit.MustExec("analyze table t")
```
0xPoe: If I remember correctly, the predicate column feature is enabled by default, which means this analysis command doesn't collect the stats for column A at all.

0xPoe: You can verify it by adding the following test steps:
```go
func TestNewIndexWithColumnStats(t *testing.T) {
	// Test two identical tables where one has no statistics at all, and the other only has column statistics but no index statistics
	// Test that we supplement index estimation with column stats if there are no index stats available
	store, dom := testkit.CreateMockStoreAndDomain(t)
	testKit := testkit.NewTestKit(t, store)
	testKit.MustExec("use test")
	testKit.MustExec("drop table if exists t")
	testKit.MustExec("create table t(a int)")
	testKit.MustExec("create table t2(a int, index idxa(a))")
	testKit.MustExec("set @@tidb_analyze_version=2")
	testKit.MustExec("set @@global.tidb_enable_auto_analyze='OFF'")
	testKit.MustExec("insert into t select mod(a,250) from (with recursive x as (select 1 as a union all select a + 1 AS a from x where a < 500) select a from x) as subquery")
	testKit.MustExec("insert into t2 select mod(a,250) from (with recursive x as (select 1 as a union all select a + 1 AS a from x where a < 500) select a from x) as subquery")
	testKit.MustExec("analyze table t")
	is := dom.InfoSchema()
	tblInfo, err := is.TableByName(context.Background(), ast.NewCIStr("test"), ast.NewCIStr("t"))
	require.NoError(t, err)
	h := dom.StatsHandle()
	statsTbl := h.GetTableStats(tblInfo.Meta())
	colStats := statsTbl.GetCol(1)
	require.NotNil(t, colStats)
	require.True(t, colStats.IsAnalyzed())
	testKit.MustExec("create index idxa on t(a)")
...
}
```

0xPoe: I believe `analyze t all columns` should solve this issue.

pkg/planner/core/casetest/planstats/testdata/plan_stats_suite_out.json
```diff
@@ -128,7 +128,7 @@
           "└─IndexJoin 1.00 root  inner join, inner:IndexLookUp, outer key:test.tp.c, inner key:test.t.b, equal cond:eq(test.tp.c, test.t.b)",
           "  ├─TableReader(Build) 1.00 root partition:p1 data:Selection",
           "  │ └─Selection 1.00 cop[tikv]  eq(test.tp.a, 10), not(isnull(test.tp.c))",
-          "  │   └─TableFullScan 6.00 cop[tikv] table:tp keep order:false, stats:partial[c:unInitialized]",
+          "  │   └─TableFullScan 6.00 cop[tikv] table:tp keep order:false",
```
0xPoe, line 131: Could you please explain how this change would impact this test case?

AilinKid (approved):
/hold unhold this when Poe's comment are reviewed

pkg/planner/core/casetest/planstats/testdata/plan_stats_suite_out.json
ichen032: For this test case, there were column stats, but tidb_stats_load_sync_wait is set to 0 so previously they wouldn't be loaded in yet. For my code because I check for if there are valid column stats when index stats are invalid, they are loaded in in the process. So after my change, the column c is no longer unInitialized. Is this an important behavior to preserve?

pkg/planner/cardinality/selectivity_test.go
ichen032: Makes sense, I have removed this line.

pkg/planner/cardinality/selectivity_test.go
ichen032: Thank you for catching this! I have updated so that the two tables work correctly now. Additionally, before it was passing because there was some pseudo-estimation logic I did not account for, I have refined the final check so that now it only passes with my changes.

pkg/planner/core/casetest/planstats/testdata/plan_stats_suite_out.json
0xPoe: No, I think it is OK. Thank you for your reply.

0xPoe:
/unhold 

cc: @ichen032  https://github.com/pingcap/tidb/pull/62725#pullrequestreview-3101286926

PR #62759, opened 2025-08-01
ddl,parser: modify schema to store the partial condition

### LINKED ISSUE #62758: Implement this feature in parser and DDL. Do proper validation and store/load the metadata.

YangKeao:

Issue Number: close #62758

https://github.com/pingcap/tidb/pull/63448
-> https://github.com/pingcap/tidb/pull/62759
https://github.com/pingcap/tidb/pull/62762

1. Modify the parser to add support for partial index. The partial index is also part of the `IndexOption`.
2. Modify the schema to store the partial condition. The DDL will validate and convert the partial condition to a string and store it in the schema.
3. Add a new column `predicate` to the `tidb_indexes` table to show the index condition.
4. Modify the `ShowExec` to show the index condition.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/ddl/index.go
```diff
@@ -3438,3 +3446,171 @@ func renameHiddenColumns(tblInfo *model.TableInfo, from, to ast.CIStr) {
...
+	restoreFlags := format.RestoreStringSingleQuotes | format.RestoreKeyWordLowercase | format.RestoreNameBackQuotes |
+		format.RestoreSpacesAroundBinaryOperation | format.RestoreWithoutSchemaName | format.RestoreWithoutTableName
+	restoreCtx := format.NewRestoreCtx(restoreFlags, &sb)
+	sb.Reset()
+	err = indexConditionExpr.Restore(restoreCtx)
+	if err != nil {
+		return "", errors.Trace(err)
+	}
+
+	return sb.String(), nil
```
wjhuang2016, line 3634: We need to check the length of the condition.

pkg/ddl/index.go
```diff
@@ -3438,3 +3446,171 @@ func renameHiddenColumns(tblInfo *model.TableInfo, from, to ast.CIStr) {
...
+		// Reference `types.DefaultTypeForValue`, they are all possible types for literal values.
+		// However, this switch-case still includes more types than the ones we have in that function
+		// to avoid breaking in the future.
+		//
+		// Accept tiny type conversion as the type of the literal value is too limited. We shouldn't
+		// force the user to use such a limited range of types.
+		//
+		// It'll allow precision / length difference in most of the cases.
+		switch constantExpr.GetType().GetType() {
+		case mysql.TypeTiny, mysql.TypeShort, mysql.TypeLong, mysql.TypeLonglong,
```
wjhuang2016, line 3719: The type of an expression can't be tinyint, short, ...

pkg/ddl/index.go
```diff
@@ -3438,3 +3446,171 @@ func renameHiddenColumns(tblInfo *model.TableInfo, from, to ast.CIStr) {
...
+				columnInfo.GetType() != mysql.TypeNewDecimal {
+				return dbterror.ErrUnsupportedAddPartialIndex.GenWithStackByArgs(
+					fmt.Sprintf("the type %s of the column `%s` in partial index condition is not compatible with the literal value type %s",
+						columnInfo.FieldType.String(), columnName.Name.Name.L, constantExpr.GetType().String()))
+			}
+			return nil
+		case mysql.TypeVarchar, mysql.TypeVarString, mysql.TypeString,
+			mysql.TypeTinyBlob, mysql.TypeMediumBlob, mysql.TypeLongBlob, mysql.TypeBlob:
+			// the target column must be a string type
+			if columnInfo.GetType() != mysql.TypeVarchar &&
```
wjhuang2016: IsString()

pkg/ddl/index.go
YangKeao: Is there any special limitation for the length of the condition?

pkg/ddl/index.go
YangKeao: Yes. I've added the explaination in the comments. The types of a constant / literal is always defined in `types.DefaultTypeForValue` (for integer, it'll always be `TypeLonglong`).

I added more conditions here to avoid the case that someone modified the logic of parsing literals (or adding new literals) in the future, and forget to modify this function. It's fine to cover all kinds of integer types because it'll also not cause type conversion.

pkg/ddl/index.go
YangKeao: Done.

tests/realtikvtest/addindextest4/partial_index_test.go
```diff
@@ -0,0 +1,134 @@
...
+	idx := findIndex(t, dom, tableName, indexName)
+	require.NotNil(t, idx)
+	require.NotEmpty(t, idx.ConditionExprString)
+}
+
+func TestPartialIndexDDL(t *testing.T) {
+	store, dom := realtikvtest.CreateMockStoreAndDomainAndSetup(t)
+	tk := testkit.NewTestKit(t, store)
+
+	t.Run("TestCreatePartialIndex", func(t *testing.T) {
```
YangKeao, line 47: These tests have to be added in `realtikvtest`, because creating index is only allowed for fast-reorg. Creating index with other methods will be avoided in https://github.com/pingcap/tidb/pull/62762.

pkg/ddl/index.go
```diff
@@ -3543,3 +3496,174 @@ func renameHiddenColumns(tblInfo *model.TableInfo, from, to ast.CIStr) {
...
+		//
+		// Accept tiny type conversion as the type of the literal value is too limited. We shouldn't
+		// force the user to use such a limited range of types.
+		//
+		// It'll allow precision / length difference in most of the cases.
+		switch constantExpr.GetType().GetType() {
+		case mysql.TypeTiny, mysql.TypeShort, mysql.TypeLong, mysql.TypeLonglong,
+			mysql.TypeInt24, mysql.TypeBit:
+			// the target column must be an integer type
+			if columnInfo.GetType() != mysql.TypeTiny &&
```
wjhuang2016, line 3722: How about year(4)?

pkg/ddl/index.go
```diff
@@ -3543,3 +3496,174 @@ func renameHiddenColumns(tblInfo *model.TableInfo, from, to ast.CIStr) {
...
+			mysql.TypeTinyBlob, mysql.TypeMediumBlob, mysql.TypeLongBlob, mysql.TypeBlob:
+			// the target column must be a string type
+			if !types.IsString(columnInfo.GetType()) {
+				return dbterror.ErrUnsupportedAddPartialIndex.GenWithStackByArgs(
+					fmt.Sprintf("the type %s of the column `%s` in partial index condition is not compatible with the literal value type %s",
+						columnInfo.FieldType.String(), columnName.Name.Name.L, constantExpr.GetType().String()))
+			}
+
+			// also check the collation of the column and the literal value
+			if columnInfo.FieldType.GetCollate() != constantExpr.GetType().GetCollate() {
```
wjhuang2016: I think it's ok to allow different collation. Otherwise, it may be difficult to use partial index.

YangKeao: Agree. I'll still check the charset, to avoid the issue in converting between binary and other charsets.

pkg/ddl/index.go
YangKeao: They are added in the following branch:

```
		case mysql.TypeTimestamp, mysql.TypeDate, mysql.TypeDuration, mysql.TypeNewDate,
			mysql.TypeYear, mysql.TypeDatetime, mysql.TypeJSON, mysql.TypeEnum, mysql.TypeSet:
			// The `DATE '2025-07-28'` is actually a `cast` function, so they are also not supported yet.
			intest.Assert(false, "should never generate literal values of these types")

			return dbterror.ErrUnsupportedAddPartialIndex.GenWithStackByArgs(
				fmt.Sprintf("the type %s of the literal value in partial index condition is not supported",
					constantExpr.GetType().String()))
```

pkg/ddl/index.go
YangKeao: Also added a test to cover it.

pkg/ddl/index.go
YangKeao: Oh! You are right, `YEAR` is also acceptable as it doesn't have special conversion logic between YEAR and other integers. I've added it and also added a test for it.

wjhuang2016 (approved):

winoros (approved):

Issue #62685, opened 2025-07-29
show stats_meta is too slow on v8.5.1


0xPoe:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

1. Create one million tables in v8.5.1 cluster.
2. Run `show stats_meta`.

### 2. What did you expect to see? (Required)
It returns quickly.

### 3. What did you see instead (Required)
It took 10 minutes.
```sql
1000000 rows in set (10 min 37.25 sec)
```

### 4. What is your TiDB version? (Required)

v8.5.1, master

0xPoe:
This issue also occurs with all other commands related to show statistics.

Benjamin2037:
We may need to sort out more detail information and then see how the customer will expect to use these commands. Let's have   a further research and discussion.

0xPoe:
Fixed by https://github.com/pingcap/tidb/pull/62811.

PR #62790, opened 2025-08-02
*: replace GetTableStats with GetPhysicalTableStats

0xPoe:

Issue Number: ref https://github.com/pingcap/tidb/issues/62685

Problem Summary:

We should avoid always taking table information if we don't need a pseudo object.

1. Change the function name from `GetTableStats` to `GetPhysicalTableStats`. This new function will directly accept the physical table ID as its input and also requires the table information to generate pseudo stats when necessary.
2. Change the function name from `GetTableStatsForAutoAnalyze` to `GetNonPseudoPhysicalTableStats`. This new function is similar to GetPhysicalTableStats but does not require the table information and does not generate pseudo stats, so it may return nil as its return value.
3. Remove the APIs `GetPartitionStats` and `GetPartitionStatsForAutoAnalyze`, as these have been deprecated.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


0xPoe (commented):

- [x] Unit tests updated

- [x] All tests pass

- [x] Bazel files updated

- [x] PR size is reasonable


/cc @time-and-fate @AilinKid

tangenta (approved):

pkg/statistics/handle/handle.go
```diff
@@ -171,50 +171,47 @@ func NewHandle(
...
+			tbl := statistics.PseudoTable(tblInfo, false, false)
+			tbl.PhysicalID = physicalTableID
+			return tbl, true
+		}
+		return nil, false
 	}
-	tbl, ok := h.Get(pid)
+
+	tbl, ok := h.Get(physicalTableID)
 	if !ok {
```
time-and-fate: I think `if ok {return tbl, true}` will make it cleaner by reducing a nested block.

time-and-fate (approved):

pkg/statistics/handle/handle.go
0xPoe: Cool! I will update it.

PR #62804, opened 2025-08-04
ddl: support read only ddl

fzzf678:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/executor.go
+++ b/pkg/ddl/executor.go
@@ -90,6 +90,8 @@ const (
 	// Once tiflashCheckPendingTablesLimit is reached, we trigger a limiter detection.
 	tiflashCheckPendingTablesLimit = 100
 	tiflashCheckPendingTablesRetry = 7
+	// "1" means the database is read-only, "0" or "default" means read-write.
+	databaseReadOnly = "1"
 )
 
 var errCheckConstraintIsOff = errors.NewNoStackError(variable.TiDBEnableCheckConstraint + " is off")
@@ -413,6 +415,40 @@ func (e *executor) ModifySchemaDefaultPlacement(ctx sessionctx.Context, stmt *as
 	return errors.Trace(err)
 }
 
+func (e *executor) ModifySchemaReadOnlyState(ctx sessionctx.Context, stmt *ast.AlterDatabaseStmt, readOnly bool) (err error) {
+	dbName := stmt.Name
+	if metadef.IsMemOrSysDB(dbName.L) {
+		return dbterror.ErrAccessSystemDBRejected.GenWithStackByArgs(dbName.L)
+	}
+	is := e.infoCache.GetLatest()
+	dbInfo, ok := is.SchemaByName(dbName)
+	if !ok {
+		return infoschema.ErrDatabaseNotExists.GenWithStackByArgs(dbName.O)
+	}
+	if dbInfo.ReadOnly == readOnly {
+		return nil
+	}
+	// Do the DDL job.
+	job := &model.Job{
+		Version:        model.GetJobVerInUse(),
+		SchemaID:       dbInfo.ID,
+		SchemaName:     dbInfo.Name.L,
+		Type:           model.ActionModifySchemaReadOnly,
+		BinlogInfo:     &model.HistoryInfo{},
+		CDCWriteSource: ctx.GetSessionVars().CDCWriteSource,
+		InvolvingSchemaInfo: []model.InvolvingSchemaInfo{{
+			Database: dbInfo.Name.L,
+			Table:    model.InvolvingAll,
+		}},
+		SQLMode: ctx.GetSessionVars().SQLMode,
+	}
+	args := &model.ModifySchemaArgs{
+		ReadOnly: readOnly,
+	}
+	err = e.doDDLJob2(ctx, job, args)
+	return errors.Trace(err)
+}
+
 // getPendingTiFlashTableCount counts unavailable TiFlash replica by iterating all tables in infoCache.
 func (e *executor) getPendingTiFlashTableCount(originVersion int64, pendingCount uint32) (int64, uint32) {
 	is := e.infoCache.GetLatest()
@@ -680,10 +716,11 @@ func checkMultiSchemaSpecs(_ sessionctx.Context, specs []*ast.DatabaseOption) er
 func (e *executor) AlterSchema(sctx sessionctx.Context, stmt *ast.AlterDatabaseStmt) (err error) {
 	// Resolve target charset and collation from options.
 	var (
-		toCharset, toCollate     string
-		isAlterCharsetAndCollate bool
-		placementPolicyRef       *model.PolicyRefInfo
-		tiflashReplica           *ast.TiFlashReplicaSpec
+		toCharset, toCollate           string
+		isAlterCharsetAndCollate       bool
+		placementPolicyRef             *model.PolicyRefInfo
+		tiflashReplica                 *ast.TiFlashReplicaSpec
+		modifyReadOnlyOption, readOnly bool
 	)
 
 	err = checkMultiSchemaSpecs(sctx, stmt.Options)
@@ -716,6 +753,11 @@ func (e *executor) AlterSchema(sctx sessionctx.Context, stmt *ast.AlterDatabaseS
 			placementPolicyRef = &model.PolicyRefInfo{Name: pmodel.NewCIStr(val.Value)}
 		case ast.DatabaseSetTiFlashReplica:
 			tiflashReplica = val.TiFlashReplica
+		case ast.DatabaseOptionReadOnly:
+			modifyReadOnlyOption = true
+			if val.Value == databaseReadOnly {
+				readOnly = true
+			}
 		}
 	}
 
@@ -731,6 +773,11 @@ func (e *executor) AlterSchema(sctx sessionctx.Context, stmt *ast.AlterDatabaseS
 	}
 	if tiflashReplica != nil {
 		if err = e.ModifySchemaSetTiFlashReplica(sctx, stmt, tiflashReplica); err != nil {
+			return err
+		}
+	}
+	if modifyReadOnlyOption {
+		if err = e.ModifySchemaReadOnlyState(sctx, stmt, readOnly); err != nil {
 			return err
 		}
 	}
--- a/pkg/ddl/executor.go
+++ b/pkg/ddl/executor.go
@@ -417,7 +417,7 @@ func (e *executor) ModifySchemaDefaultPlacement(ctx sessionctx.Context, stmt *as
 
 func (e *executor) ModifySchemaReadOnlyState(ctx sessionctx.Context, stmt *ast.AlterDatabaseStmt, readOnly bool) (err error) {
 	dbName := stmt.Name
-	if metadef.IsMemOrSysDB(dbName.L) {
+	if util.IsSysDB(dbName.L) || util.IsSystemView(dbName.L) {
 		return dbterror.ErrAccessSystemDBRejected.GenWithStackByArgs(dbName.L)
 	}
 	is := e.infoCache.GetLatest()
--- a/pkg/ddl/schema.go
+++ b/pkg/ddl/schema.go
@@ -17,15 +17,22 @@ package ddl
 import (
 	"context"
 	"fmt"
+	"strconv"
+	"strings"
+	"time"
 
 	"github.com/pingcap/errors"
+	"github.com/pingcap/failpoint"
 	"github.com/pingcap/tidb/pkg/ddl/label"
 	"github.com/pingcap/tidb/pkg/ddl/notifier"
+	sess "github.com/pingcap/tidb/pkg/ddl/session"
 	"github.com/pingcap/tidb/pkg/domain/infosync"
 	"github.com/pingcap/tidb/pkg/infoschema"
 	"github.com/pingcap/tidb/pkg/kv"
 	"github.com/pingcap/tidb/pkg/meta"
 	"github.com/pingcap/tidb/pkg/meta/model"
+	"github.com/pingcap/tidb/pkg/util/logutil"
+	"go.uber.org/zap"
 )
 
 func onCreateSchema(jobCtx *jobContext, job *model.Job) (ver int64, _ error) {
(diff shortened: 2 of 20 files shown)


pkg/meta/model/job.go
```diff
@@ -111,6 +111,7 @@ const (
 	ActionAlterTablePartitioning ActionType = 71
 	ActionRemovePartitioning     ActionType = 72
 	ActionAddVectorIndex         ActionType = 73
+	ActionModifySchemaReadOnly   ActionType = 77
```
fzzf678, line 114: [74, 76] are occupied ar master branch

D3Hunter: better file a PR to master to reserve `77` if this PR will be cherry-pick back at a very late time

pkg/ddl/schema.go
```diff
@@ -154,6 +161,144 @@ func onModifySchemaDefaultPlacement(jobCtx *jobContext, job *model.Job) (ver int
...
+		if err = jobCtx.metaMut.UpdateDatabase(dbInfo); err != nil {
+			return ver, errors.Trace(err)
+		}
+		if ver, err = updateSchemaVersion(jobCtx, job); err != nil {
+			return ver, errors.Trace(err)
+		}
+		job.FinishDBJob(model.JobStateDone, model.StatePublic, ver, dbInfo)
+		return ver, nil
+	}
+	trxTableName := "CLUSTER_TIDB_TRX"
```
joechenrh:
```suggestion
	trxTableName := infoschema.ClusterTableTiDBTrx
```

pkg/ddl/schema.go
```diff
@@ -154,6 +161,144 @@ func onModifySchemaDefaultPlacement(jobCtx *jobContext, job *model.Job) (ver int
...
+			}
+		})
+		if err != nil {
+			return ver, errors.Trace(err)
+		}
+
+		// check and wait all uncommitted txn
+		for len(uncommittedTxn) > 0 {
+			for txnID := range uncommittedTxn {
+				sql := fmt.Sprintf("SELECT ID FROM INFORMATION_SCHEMA.%s WHERE ID = %d", trxTableName, txnID)
```
joechenrh: Can we use `IN` to reduce the time of query? For now, we need to query at least `len(uncommittedTxn)` times.

pkg/ddl/schema_test.go
```diff
@@ -493,3 +501,221 @@ func TestRenameTableAutoIDs(t *testing.T) {
...
+// enableReadOnlyDDLFp enables the failpoint to mock read-only DDLs.
+func enableReadOnlyDDLFp() {
+	failpoint.Enable("github.com/pingcap/tidb/pkg/ddl/mockModifySchemaReadOnlyDDL", "return")
+	failpoint.Enable("github.com/pingcap/tidb/pkg/executor/mockModifySchemaReadOnlyDDL", "return")
+}
+
+func disableReadOnlyDDLFp() {
+	failpoint.Disable("github.com/pingcap/tidb/pkg/ddl/mockModifySchemaReadOnlyDDL")
+	failpoint.Disable("github.com/pingcap/tidb/pkg/executor/mockModifySchemaReadOnlyDDL")
+}
```
joechenrh: Why don't use `testfailpoint`, seems we only disable failpoint after the test finished

fzzf678: update in https://github.com/pingcap/tidb/pull/62804/commits/3f5bc0b682f8ef4dea1cdaa3f695ccb2f8493c82, PTAL again

pkg/ddl/schema.go
fzzf678: Yes, use `IN` indeed can reduce the time of query, but I think we'd better provide some info about uncommitted transactions here to check and take action on long-term stuck situations. If we use in subquery,  we can only provide a batch of IDs by log, not the uncommitted one.

pkg/ddl/schema.go
```diff
@@ -154,6 +161,144 @@ func onModifySchemaDefaultPlacement(jobCtx *jobContext, job *model.Job) (ver int
...
+		}
+		if ver, err = updateSchemaVersion(jobCtx, job); err != nil {
+			return ver, errors.Trace(err)
+		}
+		job.FinishDBJob(model.JobStateDone, model.StatePublic, ver, dbInfo)
+		return ver, nil
+	}
+	trxTableName := infoschema.ClusterTableTiDBTrx
+	failpoint.Inject("mockModifySchemaReadOnlyDDL", func() {
+		trxTableName = "TIDB_TRX"
```
CbcWestwolf:
```suggestion
		trxTableName = infoschema.TableTiDBTrx
```

pkg/ddl/schema.go
```diff
@@ -154,6 +161,144 @@ func onModifySchemaDefaultPlacement(jobCtx *jobContext, job *model.Job) (ver int
...
+			return ver, errors.Trace(err)
+		}
+		if ver, err = updateSchemaVersion(jobCtx, job); err != nil {
+			return ver, errors.Trace(err)
+		}
+		job.FinishDBJob(model.JobStateDone, model.StatePublic, ver, dbInfo)
+		return ver, nil
+	}
+	trxTableName := infoschema.ClusterTableTiDBTrx
+	failpoint.Inject("mockModifySchemaReadOnlyDDL", func() {
```
CbcWestwolf, line 192: Maybe use `if intest.InTest` instead

pkg/infoschema/infoschema_v2.go
```diff
@@ -1477,6 +1484,25 @@ func (b *Builder) applyModifySchemaDefaultPlacementV2(m meta.Reader, diff *model
...
+		// This should never happen.
+		return ErrDatabaseNotExists.GenWithStackByArgs(
+			fmt.Sprintf("(Schema ID %d)", diff.SchemaID),
+		)
+	}
+	oldDBInfo, _ := b.infoschemaV2.SchemaByID(diff.SchemaID)
+	newDBInfo := oldDBInfo.Clone()
+	newDBInfo.ReadOnly = di.ReadOnly
+	b.infoschemaV2.deleteDB(di, diff.Version)
+	b.infoschemaV2.addDB(diff.Version, di)
```
JQWong7: should we add `newDBInfo`?

fzzf678: Yes, fixed in https://github.com/pingcap/tidb/pull/62804/commits/4b108e0c654805c6c56e02421accf2c7516d5cd9

pkg/ddl/schema.go
fzzf678: `mockModifySchemaReadOnlyDDL` controls two failpoint, for another failpoint, we don't need to always keep it open, just open it in read only ut. Use `if intest.InTest` can't totally replace `failpoint.Inject`, so I 'd like to maintain the original implementation.

CbcWestwolf (approved):

pkg/ddl/executor.go
```diff
@@ -413,6 +415,40 @@ func (e *executor) ModifySchemaDefaultPlacement(ctx sessionctx.Context, stmt *as
...
+		return dbterror.ErrAccessSystemDBRejected.GenWithStackByArgs(dbName.L)
+	}
+	is := e.infoCache.GetLatest()
+	dbInfo, ok := is.SchemaByName(dbName)
+	if !ok {
+		return infoschema.ErrDatabaseNotExists.GenWithStackByArgs(dbName.O)
+	}
+	if dbInfo.ReadOnly == readOnly {
+		return nil
+	}
```
tangenta: Maybe add a note to notify user nothing is changed?

```
ctx.GetSessionVars().StmtCtx.AppendNote(...)
```

pkg/ddl/schema.go
```diff
@@ -154,6 +161,144 @@ func onModifySchemaDefaultPlacement(jobCtx *jobContext, job *model.Job) (ver int
...
+		})
+		if err != nil {
+			return ver, errors.Trace(err)
+		}
+
+		// check and wait all uncommitted txn
+		for len(uncommittedTxn) > 0 {
+			for txnID := range uncommittedTxn {
+				sql := fmt.Sprintf("SELECT ID FROM INFORMATION_SCHEMA.%s WHERE ID = %d", trxTableName, txnID)
+				r, err := w.sess.Execute(jobCtx.stepCtx, sql, "check if txn committed")
```
tangenta: It is better to use a new session from session pool instead of `w.sess` because `w.sess` use an explicit transaction (see `w.prepareTxn(job)`), even if `ClusterTableTiDBTrx` always return the latest info?

pkg/ddl/schema.go
```diff
@@ -154,6 +161,144 @@ func onModifySchemaDefaultPlacement(jobCtx *jobContext, job *model.Job) (ver int
...
+	failpoint.Inject("mockModifySchemaReadOnlyDDL", func() {
+		trxTableName = infoschema.TableTiDBTrx
+	})
+
+	switch job.SchemaState {
+	case model.StateNone:
+		dbInfo.ReadOnly = args.ReadOnly
+		if err = jobCtx.metaMut.UpdateDatabase(dbInfo); err != nil {
+			return ver, errors.Trace(err)
+		}
```
tangenta:
```suggestion
		if err = jobCtx.metaMut.UpdateDatabase(dbInfo); err != nil {
		    job.State = model.JobStateCancelled
			return ver, errors.Trace(err)
		}
```

pkg/ddl/schema.go
```diff
@@ -154,6 +161,144 @@ func onModifySchemaDefaultPlacement(jobCtx *jobContext, job *model.Job) (ver int
...
+				sql := fmt.Sprintf("SELECT ID FROM INFORMATION_SCHEMA.%s WHERE ID = %d", trxTableName, txnID)
+				r, err := w.sess.Execute(jobCtx.stepCtx, sql, "check if txn committed")
+				if err != nil {
+					return ver, errors.Trace(err)
+				}
+				if len(r) == 0 {
+					delete(uncommittedTxn, txnID)
+					continue
+				}
+				logutil.BgLogger().Info("uncommitted txn block read only ddl", zap.Int64("txn ID", txnID))
```
tangenta: Maybe use `SampleLogger` instead to prevent too many logs. Also add the job ID to zap fields.

pkg/ddl/schema.go
```diff
@@ -154,6 +161,144 @@ func onModifySchemaDefaultPlacement(jobCtx *jobContext, job *model.Job) (ver int
...
+	}
+	return ver, nil
+}
+
+func getUncommittedTxnIDs(jobCtx *jobContext, sess *sess.Session, ddlQuery string, targetDBID int64, trxTableName string) (map[int64]struct{}, error) {
+	var currTS = uint64(0)
+	err := kv.RunInNewTxn(jobCtx.ctx, jobCtx.store, true, func(_ context.Context, txn kv.Transaction) error {
+		currTS = txn.StartTS()
+		return nil
+	})
```
tangenta, line 269: Just confirm, after we get the `StartTS`, is it guaranteed that no new entries will occur in table `trxTableName ` that satisfied `ID < StartTS`?

pkg/ddl/schema.go
```diff
@@ -154,6 +161,144 @@ func onModifySchemaDefaultPlacement(jobCtx *jobContext, job *model.Job) (ver int
...
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	is := jobCtx.infoCache.GetLatest()
+	sql := fmt.Sprintf(
+		"SELECT RELATED_TABLE_IDS, CURRENT_SQL_DIGEST, ID "+
+			"FROM INFORMATION_SCHEMA.%s "+
+			"WHERE (CURRENT_SQL_DIGEST IS NULL OR CURRENT_SQL_DIGEST != TIDB_ENCODE_SQL_DIGEST('%s')) "+ // exclude ddl
```
tangenta: This assumption seems to be too strong. Maybe find another way to exclude DDL, or add comments to the place where `CURRENT_SQL_DIGEST` is set.

pkg/ddl/schema.go
```diff
@@ -154,6 +161,144 @@ func onModifySchemaDefaultPlacement(jobCtx *jobContext, job *model.Job) (ver int
...
+	)
+	t := time.Now()
+	r, err := sess.Execute(jobCtx.stepCtx, sql, "get uncommitted txn id")
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	queryDur := time.Since(t)
+	ids := make(map[int64]struct{})
+	for _, row := range r {
+		txnID := row.GetInt64(2)
```
tangenta:
```suggestion
		txnID := row.GetInt64(1)
```

After the above comment addressed, `CURRENT_SQL_DIGEST` could be removed in the projection fields (`SELECT RELATED_TABLE_IDS, ID`).

pkg/ddl/schema_test.go
```diff
@@ -493,3 +500,206 @@ func TestRenameTableAutoIDs(t *testing.T) {
...
+	r := tk1.MustQuery("select @@tidb_current_ts").Rows()
+	txnID, err := strconv.ParseInt(r[0][0].(string), 10, 64)
+	require.NoError(t, err)
+	var txnIDs map[int64]struct{}
+	go func() {
+		require.Eventually(t, func() bool {
+			return len(txnIDs) == 1 && txnIDs[txnID] == struct{}{}
+		}, 5*time.Second, 100*time.Millisecond)
+		tk1.MustExec("commit")
+	}()
```
tangenta: Use a wait group and `wg.Wait()`, otherwise the test may complete quickly.

pkg/ddl/schema_test.go
```diff
@@ -493,3 +500,206 @@ func TestRenameTableAutoIDs(t *testing.T) {
...
+	store := testkit.CreateMockStore(t)
+	tk1 := testkit.NewTestKit(t, store)
+	tk2 := testkit.NewTestKit(t, store)
+
+	tk1.MustExec("create database test_db")
+	tk1.MustExec("create table test_db.t(a int)")
+	tk1.MustExec("begin; use test_db;")
+	tk2.MustExec("alter database test read only = 1")
+	tk2.MustQuery("show create database test").Check(testkit.Rows("test CREATE DATABASE `test` /*!40100 DEFAULT CHARACTER SET utf8mb4 */ /* READ ONLY = 1 */"))
+	tk1.MustGetErrMsg("select * from test_db", "[schema:1146]Table 'test_db.test_db' doesn't exist")
```
tangenta: What is the purpose to select a non-exist table?

pkg/ddl/schema.go
```diff
@@ -154,6 +161,144 @@ func onModifySchemaDefaultPlacement(jobCtx *jobContext, job *model.Job) (ver int
...
+					return ver, errors.Trace(err)
+				}
+				if len(r) == 0 {
+					delete(uncommittedTxn, txnID)
+					continue
+				}
+				logutil.BgLogger().Info("uncommitted txn block read only ddl", zap.Int64("txn ID", txnID))
+				break
+			}
+			time.Sleep(300 * time.Millisecond)
```
tangenta: Use a `time.Ticker` instead of sleep.

pkg/ddl/schema.go
fzzf678: How about exclude current ddl by `SESSION_ID `,  this seems more accurate. We can pass this variable by ddl job args
```
	args := &model.ModifySchemaArgs{
		ReadOnly:        readOnly,
		DDLConnID: ctx.GetSessionVars().ConnectionID,
	}
```

pkg/ddl/schema.go
fzzf678: I tried the `SampleLogger` and found the different message will not be output, the `id=2` log is gone.  Is there something wrong with my usage?
```
func TestLog(t *testing.T) {
	sampleLogger := logutil2.SampleLoggerFactory(
		time.Minute, 3, zap.String(logutil2.LogFieldCategory, "ddl"),
	)

	sampleLogger().Info("test log", zap.Int("id", 1))
	sampleLogger().Info("test log", zap.Int("id", 1))
	sampleLogger().Info("test log", zap.Int("id", 1))
	sampleLogger().Info("test log", zap.Int("id", 2))
	sampleLogger().Info("test log", zap.Int("id", 2))
}
```

pkg/ddl/schema.go
fzzf678: Yes, checked with @lcwangchao, after we get the `StartTS`, all new transactions 's ID will greater than `StartTS`

pkg/ddl/schema_test.go
fzzf678: It's a mistake, the original purpose is to test the behavior of accessing the db's table in a transaction after the DDL done, corrcted in https://github.com/pingcap/tidb/pull/62804/commits/3ca980016ce8c7087c974afa084fe639f4810b7f

pkg/ddl/schema.go
fzzf678: Besides this comment, all fixed in https://github.com/pingcap/tidb/pull/62804/commits/3ca980016ce8c7087c974afa084fe639f4810b7f, PTAL again

pkg/ddl/schema.go
```diff
@@ -256,12 +272,11 @@ func getUncommittedTxnIDs(jobCtx *jobContext, sess *sess.Session, ddlQuery strin
 	}
 	is := jobCtx.infoCache.GetLatest()
 	sql := fmt.Sprintf(
-		"SELECT RELATED_TABLE_IDS, CURRENT_SQL_DIGEST, ID "+
-			"FROM INFORMATION_SCHEMA.%s "+
-			"WHERE (CURRENT_SQL_DIGEST IS NULL OR CURRENT_SQL_DIGEST != TIDB_ENCODE_SQL_DIGEST('%s')) "+ // exclude ddl
+		"SELECT RELATED_TABLE_IDS, ID FROM INFORMATION_SCHEMA.%s "+
+			"WHERE SESSION_ID != %d "+ // exclude current ddl
 			"AND ID < %d", // exclude new transactions
```
tangenta: Can we get the `ID` of current DDL directly here?

fzzf678: update in https://github.com/pingcap/tidb/pull/62804/commits/b28fcb0565ba072c9e00baa89afd0e902c22a96c, PTAL

PR #62811, opened 2025-08-04
statistics: use GetNonPseudoPhysicalTableStats to get stats in the show commands

0xPoe:

Issue Number: ref https://github.com/pingcap/tidb/issues/62685

Problem Summary:

based on https://github.com/pingcap/tidb/pull/62790

In this PR, we utilize SchemaSimpleTableInfos for the commands `show stats_meta` and `show stats_healthy`, which helps to accelerate the process of retrieving table information.

Additionally, I utilized `ListTablesWithSpecialAttribute` to retrieve information about the partitioned tables, as we still need the relationship between the global table and its partitions. 

After this change, the duration reduced from `5 min 44.55 sec` to `2.3 sec`.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [x] [Manual test](https://github.com/pingcap/tidb/pull/62811#issuecomment-3154778715)
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
修复 `show stats_meta` 和 `show stats_helathy` 在表非常多的情况下过慢的问题
Fix the issue where `show stats_meta` and `show stats_healthy` are too slow when there are very many tables
```

### DIFF
--- a/pkg/executor/show_stats.go
+++ b/pkg/executor/show_stats.go
@@ -124,10 +124,10 @@ func (e *ShowExec) fetchShowStatsMeta(ctx context.Context) error {
 		fieldPatternsLike = e.Extractor.FieldPatternLike()
 	}
 	tableInfoResult := do.InfoSchema().ListTablesWithSpecialAttribute(infoschemacontext.PartitionAttribute)
-	paritionedTables := make(map[int64]*model.TableInfo)
+	partitionedTables := make(map[int64]*model.TableInfo)
 	for _, result := range tableInfoResult {
 		for _, tbl := range result.TableInfos {
-			paritionedTables[tbl.ID] = tbl
+			partitionedTables[tbl.ID] = tbl
 		}
 	}
 	for _, db := range dbs {
@@ -140,19 +140,19 @@ func (e *ShowExec) fetchShowStatsMeta(ctx context.Context) error {
 		terror.Log(err)
 		for _, nameInfo := range tableNames {
 			tblID := nameInfo.ID
-			paritionedTable, ok := paritionedTables[tblID]
+			partitionedTable, ok := partitionedTables[tblID]
 			// Partitioned table:
 			if ok {
 				// For dynamic partitioned table, we need to display the global table as well.
 				if e.Ctx().GetSessionVars().IsDynamicPartitionPruneEnabled() {
-					if stats, found := h.GetNonPseudoPhysicalTableStats(paritionedTable.ID); found {
-						e.appendTableForStatsMeta(db.O, paritionedTable.Name.O, "global", stats)
+					if stats, found := h.GetNonPseudoPhysicalTableStats(partitionedTable.ID); found {
+						e.appendTableForStatsMeta(db.O, partitionedTable.Name.O, "global", stats)
 					}
 				}
-				pi := paritionedTable.GetPartitionInfo()
+				pi := partitionedTable.GetPartitionInfo()
 				for _, def := range pi.Definitions {
 					if stats, found := h.GetNonPseudoPhysicalTableStats(def.ID); found {
-						e.appendTableForStatsMeta(db.O, paritionedTable.Name.O, def.Name.O, stats)
+						e.appendTableForStatsMeta(db.O, partitionedTable.Name.O, def.Name.O, stats)
 					}
 				}
 			} else {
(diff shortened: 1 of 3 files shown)


0xPoe:
Tested locally:
Before Change:
```sql
1000000 rows in set (5 min 44.55 sec)
```
After Change:
```sql
1000000 rows in set (2.3 sec)
```

pkg/executor/show_stats.go
```diff
@@ -114,36 +118,53 @@ func (e *ShowExec) fetchShowStatsMeta(ctx context.Context) error {
...
-				e.appendTableForStatsMeta(db.O, tbl.Name.O, partitionName, h.GetTableStats(tbl))
-				if pi != nil {
-					for _, def := range pi.Definitions {
-						e.appendTableForStatsMeta(db.O, tbl.Name.O, def.Name.O, h.GetPartitionStats(tbl, def.ID))
-					}
+		tableNames, err := do.InfoSchema().SchemaSimpleTableInfos(ctx, db)
+		terror.Log(err)
+		for _, nameInfo := range tableNames {
+			tblID := nameInfo.ID
+			paritionedTable, ok := paritionedTables[tblID]
```
tiancaiamao: I find that,  in theory,  the speical attribute tables in infoschema V2 can provide TableInfoByID ability.
And if we do provide such API, we can avoid calling `ListTablesWithSpecialAttribute` to making the temporary hash table here. 

Never mind, since we did provoid such API, and its unrelated to this PR, just go ahead.

pkg/statistics/handle/handle.go
```diff
@@ -171,40 +171,36 @@ func NewHandle(
 	return handle, nil
 }
 
-// GetTableStats retrieves the statistics table from cache, and the cache will be updated by a goroutine.
-// TODO: remove GetTableStats later on.
-func (h *Handle) GetTableStats(tblInfo *model.TableInfo) *statistics.Table {
```
tiancaiamao: What's the problem with the old API?
Since it changes so many lines and and it make me confused that, is it an unrelated refactor?

0xPoe: image

These changes are in a separate PR.

The reason we need to change it is that we shouldn't always ask the caller to pass the table info to use this function.

pkg/executor/show_stats.go
0xPoe: > I find that, in theory, the speical attribute tables in infoschema V2 can provide TableInfoByID ability.

Do you mean only for the partitioned tables or all tables? Or are you implying that we can use TableInfoByID instead of the special API here?

tiancaiamao: For partition tables only.
I mean if we provide a TableInfoByID just for partition tables ... that kind of special API could help here.

pkg/executor/show_stats.go
```diff
@@ -143,7 +167,7 @@ func (e *ShowExec) fetchShowStatsMeta(ctx context.Context) error {
 }
 
 func (e *ShowExec) appendTableForStatsMeta(dbName, tblName, partitionName string, statsTbl *statistics.Table) {
-	if statsTbl.Pseudo {
+	if statsTbl == nil || statsTbl.Pseudo {
```
0xPoe, line 170: This is defensive programming (

pkg/executor/show_stats.go
```diff
@@ -114,27 +115,50 @@ func (e *ShowExec) fetchShowStatsMeta(ctx context.Context) error {
...
+	var (
+		fieldPatternsLike collate.WildcardPattern
+		fieldFilter       string
+	)
+	if e.Extractor != nil {
+		fieldFilter = e.Extractor.Field()
+		fieldPatternsLike = e.Extractor.FieldPatternLike()
+	}
+	tableInfoResult := do.InfoSchema().ListTablesWithSpecialAttribute(infoschemacontext.PartitionAttribute)
+	paritionedTables := make(map[int64]*model.TableInfo)
```
terry1purcell: spelling error - it's partitioned. Not paritioned.

0xPoe: Updated! Thank you!

0xPoe:
All suggestions are pointless, so I resolved them.

tiancaiamao (approved):

pkg/executor/show_stats.go
```diff
@@ -513,30 +538,42 @@ func (e *ShowExec) fetchShowStatsHealthy(ctx context.Context) {
 		fieldFilter = e.Extractor.Field()
 		fieldPatternsLike = e.Extractor.FieldPatternLike()
 	}
+	tableInfoResult := do.InfoSchema().ListTablesWithSpecialAttribute(infoschemacontext.PartitionAttribute)
```
AilinKid, line 541: since each show will build a slices out 
image

i just wonder, why not just keep this inside schema itself, then when create/delete table, we just maintain this partition info map just like dbinfo/tableinfo/placementRule and so on.

AilinKid (commented):
rest LGTM

pkg/executor/show_stats.go
0xPoe: You may look at the old implementation. You should check the v2 implementation, and I think we maintained it in the data structure.

Issue #61759, opened 2025-06-16
Import into support Full-text index using TiCI


OliverS929:
## Enhancement
As title.

PR #62820, opened 2025-08-05
tici: integrate with TiCI import into interfaces | tidb-test=ef726505f262e22c6ec9c8951273bb5ee20abefb

JaySon-Huang:

Issue Number: ref https://github.com/pingcap/tidb/issues/61759

Problem Summary:

1. If the `IMPORT INTO` is failed, the user may run the `IMPORT INTO` again. So `table_id` + `index_id` can not consider as a unique_id. Add task_id to the request so TiCI can use `tidb_task_id` + `table_id` + `index_id` as a unique_id.
  - TiCI need this unique id to represent a "Job" in the TiCI side. TiCI rely on the TiCI job status to decide whether to cleanup the cloud storage path, update the compaction strategy in the TiCI side, and get the `IMPORT INTO` job status from tidb-server
2. Update `tici.proto`
  - Use enum class ErrorCode for the status field in response.
  - Rename `MarkPartitionUploadFinished` -> `FinishImportPartitionUpload`. Add some fields (`tidb_task_id` and so on) in the request so that TiCI can retrieve the TiCI job.
  - Rename `MarkTableUploadFinished` -> `FinishImportIndexUpload`. Add `tidb_task_id` so that TiCI can retrieve the TiCI job.
  - Rename `GetCloudStoragePath` -> `GetImportStoragePrefix`
    - This API now support represent an `IMPORT INTO` job with multiple IndexIDs
    - This API now return the remote storage prefix instead of a remote storage filename. The remote storage prefix is kept in the `DataWriter`
    - For each partition, TiDB generate a uuid with the remote storage prefix and upload the file to the remote storage. The logic is written in `DataWriter::InitTICIFileWriter`
  - Generate using kvproto scripts: https://github.com/pingcap/kvproto/pull/1335. But now keep a copy of tici.pb.go
 inside tidb-server code. Will move the proto into kvproto repo later.
3. Refactor the codes on `ticiWriteGroup`
  - `ticiWriteGroup` is shared for multiple region_job so it is not correct to init TICIFileWriters inside `ticiWriteGroup`.
  - Move the `TiCIFileWriter` as a local variable in region_job
  - Only write the table data once when the `IMPORT INTO` job is running on a table with multiple fulltext index
4. Fix a typo `MetaServiceEelectionKey` -> `MetaServiceElectionKey`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
 As https://github.com/pingcap-inc/tici/issues/239#issuecomment-3206740111 describe
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/lightning/backend/local/region_job.go
+++ b/pkg/lightning/backend/local/region_job.go
@@ -482,7 +482,7 @@ func (local *Backend) doWrite(ctx context.Context, j *regionJob) (ret *tikvWrite
 	if local.ticiWriteGroup != nil {
 		// Write headers for all tici file writers.
 		if err = local.ticiWriteGroup.WriteHeader(ctx, ticiFileWriter, dataCommitTS); err != nil {
-			return nil, errors.Annotate(err, "ticiWriteGroup.WriteHeader failed")
+			return nil, errors.Annotate(err, "failed to write header to tici file writer")
 		}
 	}
 
@@ -558,7 +558,7 @@ func (local *Backend) doWrite(ctx context.Context, j *regionJob) (ret *tikvWrite
 		// If TiCI is enabled, write the batch to all TiCI writers.
 		if local.ticiWriteGroup != nil {
 			if err := local.ticiWriteGroup.WritePairs(ctx, ticiFileWriter, pairs, count); err != nil {
-				return errors.Annotate(err, "ticiWriteGroup.WritePairs failed")
+				return errors.Annotate(err, "failed to write pairs to tici file writer")
 			}
 		}
 
@@ -648,10 +648,10 @@ func (local *Backend) doWrite(ctx context.Context, j *regionJob) (ret *tikvWrite
 
 	if local.ticiWriteGroup != nil {
 		if err := local.ticiWriteGroup.CloseFileWriters(ctx, ticiFileWriter); err != nil {
-			return nil, errors.Annotate(err, "ticiWriteGroup.CloseFileWriters failed")
+			return nil, errors.Annotate(err, "failed to close tici file writer")
 		}
 		if err := local.ticiWriteGroup.FinishPartitionUpload(ctx, ticiFileWriter, firstKey, lastKey); err != nil {
-			return nil, errors.Annotate(err, "ticiWriteGroup.FinishPartitionUpload failed")
+			return nil, errors.Annotate(err, "failed to finish upload for tici file writer")
 		}
 	}
 
--- a/pkg/tici/tici_write.go
+++ b/pkg/tici/tici_write.go
@@ -104,12 +104,12 @@ func getEtcdClient() (cli *clientv3.Client, err error) {
 	if err != nil {
 		return nil, err
 	}
-	ectdEndpoints, err := util.ParseHostPortAddr(tidbCfg.Path)
+	etcdEndpoints, err := util.ParseHostPortAddr(tidbCfg.Path)
 	if err != nil {
 		return nil, err
 	}
 	return clientv3.New(clientv3.Config{
-		Endpoints:        ectdEndpoints,
+		Endpoints:        etcdEndpoints,
 		DialTimeout:      etcdDialTimeout,
 		TLS:              tls,
 		AutoSyncInterval: 30 * time.Second,


pkg/executor/importer/table_import.go
```diff
@@ -202,7 +202,7 @@ func NewTableImporter(
 		return nil, err
 	}
 
-	if err := localBackend.InitTiCIWriterGroup(ctx, getEtcdClient, e.Table.Meta(), e.DBName); err != nil {
+	if err := localBackend.InitTiCIWriterGroup(ctx, getEtcdClient, e.Table.Meta(), e.DBName, id); err != nil {
```
JaySon-Huang: @OliverS929 Can we suppose this `id` is a unique identifier for one `IMPORT INTO` job (or a `CREATE INDEX` backfilling job that the generate index of existing data)?

pkg/lightning/backend/local/local.go
```diff
@@ -1477,7 +1477,7 @@ func (local *Backend) doImport(
 
 	if local.ticiWriteGroup != nil {
 		// If the import is done, we can close the write group.
-		return local.ticiWriteGroup.MarkTableUploadFinished(ctx)
+		return local.ticiWriteGroup.FinishIndexUpload(ctx)
 	}
```
JaySon-Huang, line 1481: @OliverS929 If there are multiple DXF instances to run the `IMPORT INTO` task, will this API called only once in the whole `IMPORT INTO` task? Or the API will be called by each DXF instance once?

pkg/tici/tici_write.go
```diff
@@ -60,9 +58,11 @@ func GetPrimaryIndex(tbl *model.TableInfo) *model.IndexInfo {
 
 // DataWriter handles S3 path management and upload notifications via TiCI Meta Service.
 type DataWriter struct {
+	tidbTaskID     string // TiDB task ID for this writer
```
OliverS929: This PR introduces a major refactor compared to the original design. It will be on hold until the internal discussion concludes and the necessary modifications are made.

pkg/tici/tici.proto
```diff
@@ -454,3 +531,89 @@ message GetIndexProgressResponse {
...
+
+message FinishImportPartitionUploadRequest {
+  // TiDB unique task ID for this Import Index job.
+  string tidb_task_id = 1;
+  int64 table_id = 2;
+  int64 index_id = 3;
+  // Key range for the partition uploaded.
+  KeyRange key_range = 4;
+  string storage_path = 5;
+}
```
JaySon-Huang, line 587:
```suggestion
message FinishImportPartitionUploadRequest {
  // TiDB unique task ID for this Import Index job.
  string tidb_task_id = 1;
  // Key range for the partition uploaded.
  KeyRange key_range = 2;
  string storage_uri = 3;
}
```

pkg/tici/tici.proto
```diff
@@ -454,3 +531,89 @@ message GetIndexProgressResponse {
...
+
+message FinishImportIndexUploadRequest {
+  // TiDB unique task ID for this Import Index job.
+  string tidb_task_id = 1;
+  // Table ID of the target table.
+  int64 table_id = 2;
+  // Index ID of the target index.
+  int64 index_id = 3;
+}
```
JaySon-Huang, line 617:
```suggestion
message FinishImportIndexUploadRequest {
  // TiDB unique task ID for this Import Into/Create Index job.
  string tidb_task_id = 1; 
  // If status is SUCCESS, the job is finished successfully.
  // If statis is not SUCCESS, the job is aborted or cancelled in TiDB side,
  // and TiCI should clean up all the related data.
  ErrorCode status = 2;
  // The error_message if the status is not SUCCESS.
  string error_message = 3;
}
```

JaySon-Huang:
I've do the following changes in this PR. Because I found after refactor the codes on `ticiWriteGroup` to move `TiCIFileWriter` as a local variable in region_job. It is better to also only write table data once when the `IMPORT INTO` job is running on a table with multiple fulltext index, otherwise there are unnecessary works for fixing the unit tests in `tici_write_test.go`.

3. Refactor the codes on `ticiWriteGroup`
  - `ticiWriteGroup` is shared for multiple region_job so it is not correct to init TICIFileWriters inside `ticiWriteGroup`.
  - Move the `TiCIFileWriter` as a local variable in region_job
  - Only write the table data once when the `IMPORT INTO` job is running on a table with multiple fulltext index

JaySon-Huang:
@OliverS929 PTAL again

pkg/tici/tici.proto
```diff
@@ -454,3 +529,99 @@ message GetIndexProgressResponse {
   // S3 path where the index is stored
   string s3_path = 7;
 }
+
+// -----------------------------------------------------------------------------
+// TiCI Full-text index data import workflow during Import Into (with DXF):
+// -----------------------------------------------------------------------------
+// 1. TiDB requests a cloud storage object prefix from the TiCI Meta Service.
+//    It provides a unique tidb_task_id that specify a job in the TiDB side. A
+//    job means one Import Into or Create Index operation.
```
OliverS929: Instead of "Create Index", I think "Index refill" might be more accurate here, as "Create Index" does not necessarily lead to a data import workload.

pkg/tici/tici.proto
```diff
@@ -454,3 +529,99 @@ message GetIndexProgressResponse {
...
+  // TiDB unique task ID for this Import Into/Create Index job.
+  string tidb_task_id = 1;
+  // Key range for the partition uploaded.
+  KeyRange key_range = 2;
+  // The cloud storage path where the partition data is written.
+  // The format is "s3://bucket/prefix/filename".
+  string storage_uri = 3;
+}
+
+// - If the index is built successfully, the `status` field should be 0,
```
OliverS929, line 589: I don’t see this being used anywhere in the PR. Is it intended for TiDB–TiCI interaction?

pkg/tici/tici.proto
```diff
@@ -454,3 +529,99 @@ message GetIndexProgressResponse {
...
+
+message FinishImportIndexUploadRequest {
+  // TiDB unique task ID for this Import Into/Create Index job.
+  string tidb_task_id = 1;
+  // If status is SUCCESS, the job is finished successfully.
+  // If status is not SUCCESS, the job is aborted or cancelled in TiDB side,
+  // and TiCI should clean up all the related data.
+  ErrorCode status = 2;
+  // The error_message if the status is not SUCCESS.
+  string error_message = 3;
```
OliverS929, line 616: I don’t think status and error_message are relevant here. TiDB would not send out this gRPC if there were any exceptions on its side.

pkg/tici/tici_manager_client.go
```diff
@@ -228,153 +229,162 @@ func (t *ManagerCtx) DropFullTextIndex(ctx context.Context, tableID, indexID int
...
 	}
-	logutil.BgLogger().Info("MarkPartitionUploadFinished success", zap.String("s3Path", s3Path))
 	return nil
 }
 
-// MarkTableUploadFinished notifies TiCI Meta Service that the whole table/index upload is finished.
-func (t *ManagerCtx) MarkTableUploadFinished(
+// AbortIndexUpload notifies TiCI that the job has failed and the related TiCI job should be aborted.
+// TiCI will clean up all the related data belonging to the tidbTaskID.
+func (t *ManagerCtx) AbortIndexUpload(
```
OliverS929, line 368: I don’t see AbortIndexUpload being used in this PR, and I also don’t see where it would be necessary in the current implementations.

pkg/lightning/backend/local/local.go
```diff
@@ -1777,9 +1778,10 @@ func GetRegionSplitSizeKeys(ctx context.Context, cli pd.Client, tls *common.TLS)
 	return 0, 0, errors.New("get region split size and keys failed")
 }
 
-// InitTiCIWriterGroup initializes the ticiWriteGroup field for the Backend using the given table info and schema.
-func (local *Backend) InitTiCIWriterGroup(ctx context.Context, getEtcdClient func() (*etcd.Client, error), tblInfo *model.TableInfo, schema string) error {
-	ticiWriteGroup, err := tici.NewTiCIDataWriterGroup(ctx, getEtcdClient, tblInfo, schema)
+// InitTiCIWriterGroup initializes the ticiWriteGroup field for the Backend using the given table info and schema
+// in the TiDB instance level. The `taskID` is a unique identifier for this Job.
+func (local *Backend) InitTiCIWriterGroup(ctx context.Context, getEtcdClient func() (*etcd.Client, error), tblInfo *model.TableInfo, schema string, taskID string) error {
```
OliverS929: By the way, do we need to pass getEtcdClient all the way down from the call in local.go? Could we instead use it directly in tici_write.go?

pkg/tici/tici_write_test.go
```diff
@@ -26,27 +25,9 @@ import (
 	"github.com/pingcap/tidb/pkg/parser/ast"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
-	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
 )
 
-func newTestTiCIDataWriter(t *testing.T) *DataWriter {
```
OliverS929, line 33: Maybe we should think twice before removing newTestTiCIDataWriter and the related test case here. Later, when we optimize to write once for multiple indexes, there may no longer be a DataWriterGroup.

pkg/executor/importer/table_import.go
OliverS929: Yes, we already discussed this offline. We can use TaskID when running IMPORT INTO from DXF. I will go ahead and add an error prompt later if an index is built from an IMPORT FROM SELECT.

pkg/tici/BUILD.bazel
```diff
@@ -24,6 +23,8 @@ go_library(
         "//pkg/util/intest",
         "//pkg/util/logutil",
         "@com_github_docker_go_units//:go-units",
+        "@com_github_gogo_protobuf//gogoproto",
```
OliverS929, line 27: Could you explain briefly why are we switching back to gogo/protobuf toolchain instead of the official v2 toolchain here? I thought gogo/protobuf has already gone deprecated and new implementation would rather prefer actively maintained v2 toolchain.

pkg/tici/tici.proto
JaySon-Huang: No, it is used for tici meta and tici worker interaction for building the index from the tidb row-format file.

pkg/tici/tici_write_test.go
JaySon-Huang: I've refactor the code because of this https://github.com/pingcap/tidb/pull/62820#issuecomment-3251541117.
So there is no longer a "Group" for DataWriter. I just keep the name `DataWriterGroup` not changed.

JaySon-Huang: If you think we need to add some test about the `DataWriteGroup` or `TiCIFileWriter` functions, you can point that out.

pkg/lightning/backend/local/local.go
JaySon-Huang: I can recheck it after resolve the conflict

pkg/tici/tici.proto
JaySon-Huang: OK, I will update the comment

D3Hunter:
> As https://github.com/pingcap-inc/tici/issues/239#issuecomment-3206740111 describe

can you paste your result here, i don't have privilege to view it

pkg/tici/tici.proto
```diff
@@ -294,6 +370,12 @@ message ShardManifestHeader {
   bytes start_key = 2;
   bytes end_key = 3;
   uint64 epoch = 4;
+  uint64 seq = 5;
+}
+
+message S3Location {
```
D3Hunter, line 378: is the S3 prefix needed,  if we need to support other cloud, like GCP/azure

pkg/tici/tici_file_writer.go
```diff
@@ -90,6 +91,11 @@ func NewTICIFileWriter(ctx context.Context, store storage.ExternalStorage, dataF
 	}, nil
 }
 
+// URI returns the URI of the key stored in external storage.
+func (w *FileWriter) URI() string {
+	return w.store.URI() + w.dataFile
```
D3Hunter: better use `path.join`

pkg/tici/tici_file_writer.go
```diff
@@ -79,6 +79,7 @@ func NewTICIFileWriter(ctx context.Context, store storage.ExternalStorage, dataF
 	if err != nil {
 		return nil, err
 	}
+	logger.Info("NewTiCIFileWriter", zap.String("dataFile", dataFile))
```
D3Hunter: please write the log message like `create TiCI file writer`, to follow current convention

pkg/tici/tici_write.go
```diff
@@ -395,88 +199,142 @@ func SetTiCIDataWriterGroupWritable(
...
+	if err != nil {
+		return nil, err
+	}
+	store, err := storage.NewWithDefaultOpt(ctx, storeBackend)
+	if err != nil {
+		return nil, err
+	}
+	if logger == nil {
+		logger = logutil.Logger(ctx)
+	}
```
D3Hunter: already called at L208

pkg/lightning/backend/local/region_job.go
```diff
@@ -448,14 +450,11 @@ func (local *Backend) doWrite(ctx context.Context, j *regionJob) (ret *tikvWrite
...
 	if local.ticiWriteGroup != nil {
-		// Call FetchCloudStoragePath for all tici writers in the group.
-		if err = local.ticiWriteGroup.FetchCloudStoragePath(ctx, firstKey, lastKey); err != nil {
-			return nil, errors.Annotate(err, "ticiWriteGroup.FetchCloudStoragePath failed")
-		}
 		// Initialize TICI file writers for each full-text index in the group.
-		if err = local.ticiWriteGroup.InitTICIFileWriters(ctx); err != nil {
-			return nil, errors.Annotate(err, "ticiWriteGroup.InitTICIFileWriters failed")
+		if ticiFileWriter, err = local.ticiWriteGroup.CreateFileWriter(ctx); err != nil {
+			return nil, errors.Annotate(err, fmt.Sprintf("ticiWriteGroup.CreateFileWriters failed, startKey=%s endKey=%s", hex.EncodeToString(firstKey), hex.EncodeToString(lastKey)))
```
D3Hunter: use `Annotatef`

pkg/tici/tici.proto
JaySon-Huang: The logic of interacting between TiCI and TiDB DXF does not use this struct.
This is an internal struct using inside TiCI and we will refactor it later for supporting multiple cloud.

pkg/lightning/backend/local/region_job.go
JaySon-Huang: Addressed

pkg/tici/tici_write.go
JaySon-Huang: Addressed

pkg/tici/tici_file_writer.go
JaySon-Huang: addressed

pkg/tici/tici_file_writer.go
JaySon-Huang: addressed

pkg/lightning/backend/local/local.go
JaySon-Huang: resolved the conflict and this is done

pkg/tici/tici.proto
JaySon-Huang: resolved

JaySon-Huang:
@D3Hunter here is the comment for running IMPORT INTO test. However, the script of transforming the JSON to CSV input is in the tici repo

```
# download the JSON input if not exists in your env
wget -q https://quickwit-datasets-public.s3.amazonaws.com/hdfs-logs-multitenants.json.gz
gzip -d hdfs-logs-multitenants.json.gz

# Transform the JSON to CSV input. The csv file size is about 3.6GB
ASSET_DIR=./ python3 ci/read_hdfs.py --out hdfs-logs-multitenants.csv hdfs-log
# Upload the CSV files to cloud storage. Here use minio for example
mc cp ./hdfs-logs-multitenants.csv k79/jayson/import-src/hdfs-logs-multitenants.csv

# Create the table and fulltext index
DROP TABLE IF EXISTS test.hdfs_log;
CREATE TABLE IF NOT EXISTS test.hdfs_log (id BIGINT AUTO_INCREMENT,timestamp BIGINT,severity_text VARCHAR(50),body TEXT,tenant_id INT,PRIMARY KEY (tenant_id, id));
ALTER TABLE test.hdfs_log ADD FULLTEXT INDEX ft_idx(severity_text, body);

# Run import into
IMPORT INTO test.hdfs_log (`timestamp`, severity_text, body, tenant_id)
  FROM 's3://jayson/import-src/hdfs-logs-multitenants.csv?access-key=minioadmin&secret-access-key=minioadmin&endpoint=http://10.2.12.79:9000'
  WITH CLOUD_STORAGE_URI='s3://jayson/import-sort?access-key=minioadmin&secret-access-key=minioadmin&endpoint=http://10.2.12.79:9000'
```

pkg/tici/BUILD.bazel
JaySon-Huang: I try to generate the tici.pb.go as how kvproto generate go codes  https://github.com/pingcap/kvproto/pull/1335 .

All other files in kvproto use gogoproto toolchain. For example, the tikvpb.proto https://github.com/pingcap/kvproto/blob/d6cbb9eb86389c67fcd15a085360494365bcdb13/proto/tikvpb.proto#L13-L19

I just do the same in tici.proto
https://github.com/pingcap/kvproto/pull/1335/files#diff-bec3e06e0dde369c8ba1e09fe9a7b59921243e2a4e6e60920328baf7a8172f5cR22-R28

I think this will be resolve after we move the tici.proto to kvproto later.

pkg/tici/tici_manager_client.go
JaySon-Huang: I would like to keep this interface for later that we can find a way to actively send the "abort" to tici.

OliverS929: Sure, that’s fine with me if you want to keep it. However, I don’t think we are sending aborts to TiCI during the import process under the current design.

pkg/tici/tici_write.go
```diff
@@ -314,12 +104,12 @@ func getEtcdClient() (cli *clientv3.Client, err error) {
 	if err != nil {
 		return nil, err
 	}
-	ectdEndpoints, err := util.ParseHostPortAddr(tidbCfg.Path)
```
OliverS929, line 317: Moving getEtcdClient() out of the current file and into another file (e.g., tici_manager_client.go) might be a good idea, since it is not directly related to writing index data to TiCI. cc @Lloyd-Pottiger

JaySon-Huang: Leave it to be resolve later

pkg/tici/tici.proto
JaySon-Huang: The same as https://github.com/pingcap/tidb/pull/62820#discussion_r2370786441
I would like to keep this interface for later that we can find a way to actively send the "abort" to tici.

PR #62837, opened 2025-08-05
txn: tag non-transcational DML's metrics with `NTDml`

### LINKED ISSUE #62799: Non-transactional DELETE DML metrics should be seperated
## Enhancement

The Non-transactional DML metrics is usually shown as `DELETE` in the metrics. However, itself is not recorded in the slowlog (though the internal statement can be recorded). It'll make the user confused.

I propose to move the non-transational DML metrics to `other`, or a standalone tag.

you06:

Issue Number: close #62799

Problem Summary:

Tag the Non-transactional DML with `NTDml` because it's usually very large DML and mixing it with general DML statements is confusing.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

image
image
image
image

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Tag non-transcational DML's metrics with `NTDml`
```


YangKeao (approved):
LGTM

pkg/executor/adapter.go
```diff
@@ -1546,6 +1546,8 @@ func (a *ExecStmt) recordAffectedRows2Metrics() {
 			metrics.AffectedRowsCounterDelete.Add(float64(affectedRows))
 		case "Update":
 			metrics.AffectedRowsCounterUpdate.Add(float64(affectedRows))
+		case "NTDml":
```
cfzjywxk: How about making it `NTDML-Update`, `NTDML-Delete` and `NTDML-Insert`?

tangenta (approved):

pkg/session/nontransactional.go
```diff
@@ -448,16 +460,43 @@ func buildShardJobs(ctx context.Context, stmt *ast.NonTransactionalDMLStmt, se s
 	originalSelectLimit := se.GetSessionVars().SelectLimit
 	se.GetSessionVars().SelectLimit = math.MaxUint64
 	// NT-DML is a write operation, and should not be affected by read_staleness that is supposed to affect only SELECT.
-	rss, err := se.Execute(ctx, selectSQL)
+	stmts, err := se.Parse(ctx, selectSQL)
```
cfzjywxk: Is this an extra parse? Could we reuse the result from somwhere?

pkg/session/nontransactional.go
```diff
@@ -275,6 +279,14 @@ func runJobs(ctx context.Context, jobs []job, stmt *ast.NonTransactionalDMLStmt,
 	if shardColumnRefer == nil && stmt.ShardColumn.Name.L != model.ExtraHandleName.L {
 		return nil, errors.New("Non-transactional DML, shard column not found")
 	}
+	switch s := stmt.DMLStmt.(type) {
+	case *ast.DeleteStmt:
+		s.IsNontransactionalDML = true
+	case *ast.UpdateStmt:
+		s.IsNontransactionalDML = true
+	case *ast.InsertStmt:
+		s.IsNontransactionalDML = true
```
zimulala:
```suggestion
	case *ast.DeleteStmt, *ast.UpdateStmt, *ast.InsertStmt:
		s.IsNontransactionalDML = true
```

pkg/session/nontransactional.go
you06: That's not an extra parse I think, this function is used to shard tasks, and the input is a select SQL. We just extract the `se.Execute` for injecting nt-dml tags here.

https://github.com/pingcap/tidb/blob/8acfef88c43fbe50df830363ff00d52346096838/pkg/session/session.go#L1599-L1621

zimulala: Or would we consider putting the logic for the `NontransactionalDMLType` into the `preprocessor.Enter`?

pkg/session/nontransactional.go
you06: There will be build error if we merge them together, because they only implement the common interface `ShardableDMLStmt`.

```bash
pkg/session/nontransactional.go:284:5: s.IsNontransactionalDML undefined (type "github.com/pingcap/tidb/pkg/parser/ast".ShardableDMLStmt has no field or method IsNontransactionalDML)
make: *** [Makefile:231: server] Error 1
```

https://github.com/pingcap/tidb/blob/8acfef88c43fbe50df830363ff00d52346096838/pkg/parser/ast/dml.go#L2707-L2717

pkg/session/nontransactional.go
you06: Sorry I don't get how to handling `NontransactionalDMLType` in `preprocessor.Enter`, seems it's a optimizor's visiter of ast.

The main reason of embedding it into ast is `ast.GetStmtLabel` is called in many places, handle the stmt type out of this function may lead to mistake in the future.

pkg/session/nontransactional.go
zimulala: Got it

pkg/session/nontransactional.go
zimulala: I'm asking if we can incorporate the logic here into "preprocessor.Enter" and set "NontransactionalDMLType" based on the conditions of "SelectStmt" and "IsNontransactionalDML". 
Of course, if possible, we should also include the content of "ast.GetStmtLabel" in "preprocessor.Enter" (this might be difficult to handle).

zimulala (approved):

cfzjywxk (approved):

hawkingrei (approved):

PR #62852, opened 2025-08-06
txn: follower read only affect read-only statements

### LINKED ISSUE #62856: Replica-read only works for read-only statements.
## Enhancement

Now, follower-read can take effect in both read-only statements and the read phase of write statements.

This enhancement restricts the replica-read only for read-only statements for safety. The main purpose is to reduce feature overlap.

you06:

Issue Number: close #62856

Problem Summary:

For the follower read function, DML is neither an optimization target nor a benefit one. Letting the reading phase of DML be affected by follower read only increases risk. This PR makes follower read no longer apply to DML, thereby reducing the risk of functionality overlap.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Run range update with `tidb_replica_read="closest-replicas"`.

image

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Replica read only works for read-only statements.
```

### DIFF
--- a/pkg/sessionctx/variable/session.go
+++ b/pkg/sessionctx/variable/session.go
@@ -2430,7 +2430,7 @@ func (s *SessionVars) SetEnablePseudoForOutdatedStats(val bool) {
 
 // GetReplicaRead get ReplicaRead from sql hints and SessionVars.replicaRead.
 func (s *SessionVars) GetReplicaRead() kv.ReplicaReadType {
-	// Replica only works for read-only statements.
+	// Replica read only works for read-only statements.
 	if !s.StmtCtx.IsReadOnly {
 		if s.StmtCtx.HasReplicaReadHint {
 			s.StmtCtx.AppendWarning(errors.New("Ignore replica read hint for non-read-only statement"))
--- a/pkg/sessionctx/variable/session.go
+++ b/pkg/sessionctx/variable/session.go
@@ -2437,6 +2437,9 @@ func (s *SessionVars) GetReplicaRead() kv.ReplicaReadType {
 		}
 		return kv.ReplicaReadLeader
 	}
+	if s.StmtCtx.RCCheckTS || s.RcWriteCheckTS {
+		return kv.ReplicaReadLeader
+	}
 	if s.StmtCtx.HasReplicaReadHint {
 		return kv.ReplicaReadType(s.StmtCtx.ReplicaRead)
 	}
--- a/pkg/sessionctx/variable/session.go
+++ b/pkg/sessionctx/variable/session.go
@@ -30,6 +30,7 @@ import (
 	"time"
 
 	"github.com/pingcap/errors"
+	"github.com/pingcap/failpoint"
 	"github.com/pingcap/kvproto/pkg/kvrpcpb"
 	"github.com/pingcap/tidb/pkg/config"
 	"github.com/pingcap/tidb/pkg/executor/join/joinversion"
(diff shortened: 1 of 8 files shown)


pkg/sessionctx/variable/session.go
```diff
@@ -2440,6 +2450,12 @@ func (s *SessionVars) GetReplicaRead() kv.ReplicaReadType {
 	return s.replicaRead
 }
 
+// GetReplicaReadUnadjusted returns the unadjusted replica read type.
+// In most cases, you should use GetReplicaRead instead of this method.
+func (s *SessionVars) GetReplicaReadUnadjusted() kv.ReplicaReadType {
```
cfzjywxk: It is confusing to leave an `GetReplicaReadUnadjusted` method, when should it be used? Or should we merge the selection logic into one?

you06: `GetReplicaReadUnadjusted` is generally used for testing the variables setting statements, we can comment it with "test-only" or inject a failpoint to return the unadjusted replica-read option.

cfzjywxk: Better not to expose it publicly to avoid mistakes.

tests/realtikvtest/txntest/replica_read_test.go
```diff
@@ -0,0 +1,130 @@
...
+		{
+			"delete can't use replica read",
+			"DELETE FROM t WHERE %s",
+			false,
+		},
+		{
+			"insert can't use replica read",
+			"INSERT INTO t SELECT id + 100, v + 100 FROM t WHERE %s",
+			false,
+		},
```
cfzjywxk, line 77: `Insert into on duplicate upcate` could be covered too.

cfzjywxk (approved):

you06:
I don't know that's wrong with checking out the branch, but those CIs are not required, we ignore them.

image

pkg/sessionctx/variable/session.go
```diff
@@ -2428,8 +2429,22 @@ func (s *SessionVars) SetEnablePseudoForOutdatedStats(val bool) {
...
+// GetReplicaRead get ReplicaRead from sql hints and SessionVars.replicaRead with adjusted.
 func (s *SessionVars) GetReplicaRead() kv.ReplicaReadType {
+	// For test purpose, you can enable this failpoint to get the unadjusted replica read.
+	failpoint.Inject("GetReplicaReadUnadjusted", func(_ failpoint.Value) {
+		failpoint.Return(s.replicaRead)
+	})
+	// Replica read only works for read-only statements.
+	if !s.StmtCtx.IsReadOnly {
+		if s.StmtCtx.HasReplicaReadHint {
+			s.StmtCtx.AppendWarning(errors.New("Ignore replica read hint for non-read-only statement"))
```
lcwangchao: Seems `GetReplicaRead` has some side effects to append a warning. Could we estimate it only to append warning if necessary?

lcwangchao (commented):
Reset LGTM

pkg/sessionctx/variable/session.go
you06: It appends warning when user specify a hint which will be ignored.

`select /*+ READ_CONSISTENT_REPLICA() */ * from t where id = 1 for update;`

There is a problem that this function can be call multi times in a query, we don't want to send repeated warnings.

you06: Check the exist warnings before appending new one. The implementation is not such good but I don’t want to introduce too many changes for this usecase which is not a good practice. PTAL again.

PR #62867, opened 2025-08-07
ddl: prohibit the execution of some ddl when database is in read only status

fzzf678:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/preprocess.go
```diff
@@ -252,6 +252,24 @@ type preprocessor struct {
...
 }
 
+func (p *preprocessor) schemaReadOnly(dbName pmodel.CIStr) {
+	if p.err != nil {
+		return
+	}
+
+	if dbName.L == "" {
+		currentDB := p.sctx.GetSessionVars().CurrentDB
+		if currentDB == "" {
```
CbcWestwolf: Better to add a comment here to say at what time `currentDB == ""`, and why we do NOT need to handle it here.

pkg/planner/core/preprocess.go
```diff
@@ -252,6 +252,24 @@ type preprocessor struct {
...
+		}
+		dbName = pmodel.NewCIStr(currentDB)
+	}
+	dbInfo, exists := p.ensureInfoSchema().SchemaByName(dbName)
+	if exists && dbInfo.ReadOnly {
+		p.err = errors.Trace(infoschema.ErrSchemaInReadOnlyMode.GenWithStackByArgs(dbName.O))
+	}
+}
+
 func (p *preprocessor) Enter(in ast.Node) (out ast.Node, skipChildren bool) {
```
joechenrh, line 319: Can we move the check into `Leave`? I think the schema name has been resolved at that time.

pkg/planner/core/preprocess.go
```diff
@@ -252,6 +252,24 @@ type preprocessor struct {
 	resolveCtx *resolve.Context
 }
 
+func (p *preprocessor) schemaReadOnly(dbName pmodel.CIStr) {
```
D3Hunter:
```suggestion
func (p *preprocessor) checkSchemaReadOnly(dbName pmodel.CIStr) {
```
methods better have a verb as the first word, unless it's a property get/setter

pkg/planner/core/preprocess.go
```diff
@@ -283,35 +301,57 @@ func (p *preprocessor) Enter(in ast.Node) (out ast.Node, skipChildren bool) {
...
 	case *ast.DropTableStmt:
 		p.flag |= inCreateOrDropTable
 		p.stmtTp = TypeDrop
 		p.checkDropTableGrammar(node)
+		for _, tbl := range node.Tables {
+			table, err := p.tableByName(tbl)
+			if err == nil && table.Meta().TempTableType == model.TempTableNone {
+				p.schemaReadOnly(tbl.Schema)
+			}
+		}
```
D3Hunter: can you unify the check location, refer `CheckTableMode` to get the schema from resolve context

tests/integrationtest/r/ddl/db_read_only.result
```diff
@@ -0,0 +1,111 @@
+drop database if exists db_read_only;
+create database db_read_only;
+drop database if exists db_read_only_1;
+create database db_read_only_1;
+use db_read_only;
+create table s(a int primary key);
+create table t(id int primary key auto_increment, a int, b int, c varchar(1), index b(b), foreign key fk_a(b) references s(a));
```
JQWong7, line 7:
```
mysql> show create table fk1.child;
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                                                                               |
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| child | CREATE TABLE `child` (
  `id` int DEFAULT NULL,
  `pid` int DEFAULT NULL,
  KEY `idx_pid` (`pid`),
  CONSTRAINT `fk_1` FOREIGN KEY (`pid`) REFERENCES `test`.`parent` (`id`) ON DELETE CASCADE
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```
We can add cross schema foreign key constraint case here

pkg/planner/core/preprocess.go
fzzf678: The db name field in different ast node has different ways to get, so need an extractor to help get the db name, then we can check it at the fixed place, update in https://github.com/pingcap/tidb/pull/62867/commits/e0f2b7fb3e295cffd2f5efe927bae8b57038905e, PTAL again

pkg/planner/core/preprocess.go
fzzf678: update in https://github.com/pingcap/tidb/pull/62867/commits/e0f2b7fb3e295cffd2f5efe927bae8b57038905e, PTAL again

joechenrh: Could `dbName` be empty after https://github.com/pingcap/tidb/commit/e0f2b7fb3e295cffd2f5efe927bae8b57038905e?
https://github.com/pingcap/tidb/blob/e0f2b7fb3e295cffd2f5efe927bae8b57038905e/pkg/planner/core/preprocess.go#L309-L314
I tested `create table/view`, and the `dbName` has been filled, not test other stmt yet

fzzf678: I tried the 'AlterTableStmt' and saw the same result as you, but i think here can be used as a redundancy check and prevention, if here are any changes in the future, it can ensure our safety. 🤔

tests/integrationtest/r/ddl/db_read_only.result
fzzf678: supply this case at https://github.com/pingcap/tidb/pull/62867/commits/6efc5c844e19b727b47dbf401cea26a26e738537, PTAL again

pkg/planner/core/preprocess.go
```diff
@@ -252,6 +252,78 @@ type preprocessor struct {
...
+		dbNames = append(dbNames, node.Table.Schema)
+	case *ast.DropIndexStmt:
+		dbNames = append(dbNames, node.Table.Schema)
+	case *ast.LoadDataStmt:
+		dbNames = append(dbNames, node.Table.Schema)
+	}
+	return dbNames
+}
+
+func (p *preprocessor) checkSchemaReadOnly(in ast.Node) {
```
D3Hunter: maybe rename it to `checkSchemaReadOnlyInStmt` and below one as `checkSchemaReadOnly`

pkg/planner/core/preprocess.go
D3Hunter: and also when will `dbName.L == ""` happen

tests/integrationtest/r/ddl/db_read_only.result
```diff
@@ -0,0 +1,114 @@
...
+Error 3809 (HY000): Schema 'db_read_only' is in read only mode.
+alter table t modify column c varchar(10);
+Error 3809 (HY000): Schema 'db_read_only' is in read only mode.
+alter table t rename index b to b_1;
+Error 3809 (HY000): Schema 'db_read_only' is in read only mode.
+alter table t drop foreign key fk_a;
+Error 3809 (HY000): Schema 'db_read_only' is in read only mode.
+alter table t add foreign key fk_a1(b) references s(a);
+Error 3809 (HY000): Schema 'db_read_only' is in read only mode.
+alter table t add foreign key fk_a2(b) references db_read_only_1.s(a);
```
D3Hunter, line 67: test this case after we make it `read write`

tests/integrationtest/r/ddl/db_read_only.result
```diff
@@ -0,0 +1,114 @@
...
+partition p2010 values less than (2020),
+partition p2020 values less than (2030),
+partition pMax values less than (maxvalue));
+Error 3809 (HY000): Schema 'db_read_only' is in read only mode.
+rename table t to t_old;
+Error 3809 (HY000): Schema 'db_read_only' is in read only mode.
+rename table db_read_only.t to db_read_only_1.t;
+Error 3809 (HY000): Schema 'db_read_only' is in read only mode.
+alter schema db_read_only read only 0;
+alter schema db_read_only_1 read only 1;
```
D3Hunter, line 89: the grammer is a little odd, the `0` is redundant, and the `read only 1` can be `read write` to be more readable

pkg/planner/core/preprocess.go
joechenrh: I think ut itself should be a form of protection. If one day somebody make some changes in the planner and break the assumption that `dbName` must be set after processing the children nodes, ut should fail.

fzzf678: ok, update in https://github.com/pingcap/tidb/pull/62867/commits/66bacebf482e0e4659e8fecc30bc881112d63fbd, PTAL again

pkg/planner/core/preprocess.go
fzzf678: Previously, the read only check was done in the preprocessor's `Enter` method, we should handle the empty name, after we move the check logical in the `Leave`, the schema name won't be empty. So I remove these redundant code in  https://github.com/pingcap/tidb/pull/62867/commits/66bacebf482e0e4659e8fecc30bc881112d63fbd

tests/integrationtest/r/ddl/db_read_only.result
fzzf678: Can't agree more, but to keep the syntax consistent with [mysql](https://dev.mysql.com/doc/refman/8.4/en/alter-database.html#alter-database-read-only), we implement in the same syntax. Maybe in the future, we can achieve `read only` and `read write` on this basis.

tests/integrationtest/r/ddl/db_read_only.result
fzzf678: update in https://github.com/pingcap/tidb/pull/62867/commits/66bacebf482e0e4659e8fecc30bc881112d63fbd, PTAL again

D3Hunter (approved):

fzzf678:
https://github.com/pingcap/tidb/pull/63120

PR #62881, opened 2025-08-07
txn, planner: Fix the problem that with pessimistic-auto-commit enabled, some internal transactions may try to commit transaction with MaxUint64

### LINKED ISSUE #62775: gc_worker keeps reporting "try to commit with invalid txnStartTS: 18446744073709551615" in next-gen deployment
## Bug Report

### Minimal reproduce step (Required)

Create and run a next-gen cluster

### What did you expect to see? (Required)

Everything goes well

### What did you see instead (Required)

TiDB keeps printing logs similar to this:

```
[2025/08/01 07:26:32.487 +00:00] [ERROR] [terror.go:317] ["function call errored"] [keyspaceName=ks3] [error="try to commit with invalid txnStartTS: 18446744073709551615"] [stack="github.com/pingcap/tidb/pkg/parser/terror.Call
	/home/_/tidb/pkg/parser/terror/terror.go:317
github.com/pingcap/tidb/pkg/store/gcworker.(*GCWorker).loadValueFromSysTable
	/home/_/tidb/pkg/store/gcworker/gc_worker.go:1540
github.com/pingcap/tidb/pkg/store/gcworker.(*GCWorker).loadTime
	/home/_/tidb/pkg/store/gcworker/gc_worker.go:1466
github.com/pingcap/tidb/pkg/store/gcworker.(*GCWorker).checkLeader
	/home/_/tidb/pkg/store/gcworker/gc_worker.go:1425
github.com/pingcap/tidb/pkg/store/gcworker.(*GCWorker).tick
	/home/_/tidb/pkg/store/gcworker/gc_worker.go:277
github.com/pingcap/tidb/pkg/store/gcworker.(*GCWorker).start
	/home/_/tidb/pkg/store/gcworker/gc_worker.go:220"]
```

### What is your TiDB version? (Required)

[4f02c1800cb7a94464b35aa59394eda7ddbdd3ee](https://github.com/pingcap/tidb/tree/4f02c1800cb7a94464b35aa59394eda7ddbdd3ee)
`NEXT_GETN=1`

### Analysis

Note that the statement reporting error is a `select ... for update` statement but in a autocommit session.

* It seems some recent change makes system tables becomes clustered, and `mysql.tidb` then has clutered index on its `variable_name` column. Thus `loadValueFromSysTable` in GC worker becomes a point get. It's not the core reason of the problem, but might be the reason that it begin to be affected by this problem.
* [`planner.core.IsPointGetWithPKOrUniqueKeyByAutoCommit`](https://github.com/pingcap/tidb/blob/22580a3cc83bee083224806debbddcc2d6b3cfd4/pkg/planner/core/common_plans.go#L1572-L1585) returns true for `select ... for update` statements in autocommit mode (causes [this code](https://github.com/pingcap/tidb/blob/4734c9c2f4a240818bb2d7aac84a4c169318b6e9/pkg/sessiontxn/isolation/optimistic.go#L153) to set the startTS to MaxUint64 later in the execution).
  * However, [`planner.IsReadOnly`](https://github.com/pingcap/tidb/blob/c7003d8d65fdf0dff2a196cc8fc18230bbc7bb1f/pkg/planner/optimize.go#L57) returns false to the same statement.
* The internal statement [doesn't trigger the fast plan](https://github.com/pingcap/tidb/blob/8d4b52a70a0707c34f52d031be9c57c42e19161c/pkg/planner/core/point_get_plan.go#L954), but it works for user session created by mysql client.
* The internal autocommit `select for update` statement may try to write locks, which is unexpected (but as it finally reports error as shown above, so it doesn't cause severe affect to the data in TiKV).
* Because it tries to write lock, the transaction loses the readonly flag and cannot skip committing [here](https://github.com/pingcap/tidb/blob/102ffeccba45bd62f7de67ebd146426e79278be6/pkg/session/session.go#L532-L534).
    * It works for user transactions (with pessimistic autocommit disabled), but it's quite doubtful whether it's the correct place to skip the transaction, as it has already checked whther the statement is readonly multiple times before, and gets negative result.

It seems the core problem is that:

* Why PointGet fast plan can't be used for the internal statement
* Why the internal statement tries to write locks.

By the way, there's another problem found during the investigation:

* With pessimstic autocommit enabled, but the MaxUint64 point get optimization won't work, while the fast plan can be generated.

MyonKeminta:

Issue Number: close #62775

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the problem that with pessimistic-auto-commit enabled, some internal transactions may try to commit transaction with MaxUint64
```


pkg/sessionctx/variable/session.go
```diff
@@ -3651,13 +3651,13 @@ func (s *SessionVars) UseLowResolutionTSO() bool {
 
 // PessimisticLockEligible indicates whether pessimistic lock should not be ignored for the current
 // statement execution. There are cases the `for update` clause should not take effect, like autocommit
-// statements with “pessimistic-auto-commit disabled.
+// statements with `pessimistic-auto-commit` disabled.
```
cfzjywxk: Is `pessimistic-auto-commit` still affecet the inner-session or sessions inner SQL would be executed?

MyonKeminta: No, I forgot to change this comment..

pkg/sessionctx/variable/session.go
```diff
@@ -3650,14 +3650,14 @@ func (s *SessionVars) UseLowResolutionTSO() bool {
...
+// statement execution. There are cases the `for update` clause should not take effect, like being
+// executed in an autocommit session.
 func (s *SessionVars) PessimisticLockEligible() bool {
 	if s.StmtCtx.ForShareLockEnabledByNoop {
 		return false
 	}
-	if !s.IsAutocommit() || s.InTxn() || (config.GetGlobalConfig().
-		PessimisticTxn.PessimisticAutoCommit.Load() && !s.BulkDMLEnabled) {
+	if !s.IsAutocommit() || s.InTxn() ||
+		s.StmtCtx.InInsertStmt || s.StmtCtx.InUpdateStmt || s.StmtCtx.InDeleteStmt {
```
AilinKid, line 3661: equivalent as IsDML(), better add a comment

AilinKid (commented):
rest LGTM

ekexium:
This should fix the next-gen test `TestInTxnPSProtoPointGet`

pkg/sessionctx/variable/session.go
```diff
@@ -3650,14 +3650,15 @@ func (s *SessionVars) UseLowResolutionTSO() bool {
 }
 
 // PessimisticLockEligible indicates whether pessimistic lock should not be ignored for the current
-// statement execution. There are cases the `for update` clause should not take effect, like autocommit
-// statements with “pessimistic-auto-commit disabled.
+// statement execution. There are cases the `for update` clause should not take effect, like being
+// executed in an autocommit session.
 func (s *SessionVars) PessimisticLockEligible() bool {
```
ekexium, line 3655: I'm thinking how we define the relationship between `PessimisticLockEligible` and `PrepareTxnCtx`(when it decides txn mode).
Is there duplicate code/logic/knowledge shared between this function and `PrepareTxnCtx`? Should we consider somehow attain single source of truth by merging them?

pkg/sessionctx/variable/session.go
ekexium: Is `LOAD DATA` counted as DML?

MyonKeminta: > Is `LOAD DATA` counted as DML?

Not sure if it should, but as I see the existing checks like `isDML` didn't include it either. Neither does it need pessimistic locks if I understand it correctly.

AilinKid (approved):

yudongusa (approved):

PR #62941, opened 2025-08-12
planner: Moved the code for PhysicalCTETable to the physicalop package.

### LINKED ISSUE #52714: planner core pkg de-coupling and make it well organized
### Refactoring
The current `plan/core` pkg is quite huge not as slim as we expected. The hybrid placement of logicalOp, physicalOp, property, task, logical-rewrite, build-phase, binder, cost, exhaustion, etc makes the hierarchy complicated. The boundary of them is also not as clear as we desired. I concluded that there is some reason for this phenomenon.

* One is if Golang's structure wants to implement member functions with itself as a receiver, including the interface implementation, which can only be defined in the same pkg where the structure is defined. (here clearly the pkg is `core`) [stackoverflow ref](https://stackoverflow.com/questions/41109708/cant-define-receiver-from-another-package-in-go) 
    + eg: `LogicalSort` is defined in core pkg somewhere. If you want to implement (p *LogicalSort) buildKeyInfo for the sake of interface implementation or just member function of them, this implementation **can only be done in the same `core` pkg**, because Golang only allows local type as a receiver. Given buildKeyInfo is logical-rewrite-related code, it should be put into another part/component/pkg of core logic for clarity, that calls for the interface simplification or elimination of self-receiver's function.

* Numerous utility and facility functions related to the core logic are defined and used in the `core` package. These include trace, hashcode, cost mode, selectivity, task, enumeration, and more. Moving or adding a new feature outside the `core` package may result in many internal structures or functions being exported out as well.

* The biggest block is the import cycle problem. Say we create a new pkg called core2 and we want to add a new feature inside. The new feature couldn't work well without the current core context support, so by some means (exported original `core` structure or interface, using function pointer or something), we got a dependency from `core2` -> `core`. However, this feature must also be called from `core`, cause `core` is a current unified portal for the entire optimization phase, so we got a dependency from `core` -> `core`. As a result, hops, golang's import cycle is formed. The final solution to the problem is **Back** to integrate them all. Eventually, the core pkg becomes larger and larger.

current core pkg should be arranged as:
```
core/
    operator/
        base/
        logical/

```

ghoshh:

Issue Number: close #52714 

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/operator/physicalop/physical_cte_table.go
+++ b/pkg/planner/core/operator/physicalop/physical_cte_table.go
@@ -1,0 +1,39 @@
+package physicalop
+
+import (
+	"strconv"
+
+	"github.com/pingcap/tidb/pkg/planner/core/base"
+	"github.com/pingcap/tidb/pkg/planner/core/operator/baseimpl"
+	"github.com/pingcap/tidb/pkg/planner/property"
+	"github.com/pingcap/tidb/pkg/util/plancodec"
+	"github.com/pingcap/tidb/pkg/util/size"
+)
+
+// PhysicalCTETable is for CTE table.
+type PhysicalCTETable struct {
+	PhysicalSchemaProducer
+
+	IDForStorage int
+}
+
+// Init only assigns type and context.
+func (p PhysicalCTETable) Init(ctx base.PlanContext, stats *property.StatsInfo) *PhysicalCTETable {
+	p.Plan = baseimpl.NewBasePlan(ctx, plancodec.TypeCTETable, 0)
+	p.SetStats(stats)
+	return &p
+}
+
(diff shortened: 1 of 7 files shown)


pkg/planner/core/operator/physicalop/physical_cte_table.go
```diff
@@ -0,0 +1,39 @@
+package physicalop
```
hawkingrei, line 15: Please add the license header

hawkingrei (approved):

PR #62946, opened 2025-08-12
executor, conn: wait for the goroutine in preprocessLoadDataLocal to finish before step to next query

### LINKED ISSUE #62673: `LOAD DATA LOCAL INFILE` will cause race condition in buffered connection, and cause panic
## Bug Report

1. Start a `LOAD DATA LOCAL INFILE`, and use a lot of memory.
2. The query is killed. Then there are two goroutines reading the same connection: one for the session, one for reading the infile (which is not killed after the query is killed).
3. The internal connection buffer cannot manage the memory well, so it may panic. The error will be like: `panic: runtime error: slice bounds out of range`

TiDB doesn't panic.

TiDB paniced and restarted.

8.1. But I think it affects all existing versions until now (312226f869d56922135e1c177cac6c90319c3134)

YangKeao:

Issue Number: close #62673 

Problem Summary:

The current implementation of `LOAD DATA ... INFILE ...` doesn't wait for the background goroutine (which is used to continuously read file data from client) to stop. Therefore, if the `LOAD DATA` statement meets any error, it's possible that the connection has two reading goroutines:

1. Expect to read the file data.
2. Expect to read the next command.

Then there will be a data race in the internal buffer, and cause deadlock or panic. The user reported a panic but in my testcase it'd be easier to reproduce deadlock (but theoritically all possible, maybe related to the error position).

Also create a `WaitGroup` to wait for this goroutine to finish before stepping to the next statement.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that if the `LOAD DATA ... INFILE ...` meets error, it may cause panic or deadlock.
```

### DIFF
--- a/pkg/server/tests/commontest/tidb_test.go
+++ b/pkg/server/tests/commontest/tidb_test.go
@@ -3815,3 +3815,36 @@ func TestAuditPluginRetrying(t *testing.T) {
 		runExplicitTransactionRetry(db, true)
 	})
 }
+
+func TestIssue62673(t *testing.T) {
+	ts := servertestkit.CreateTidbTestSuite(t)
+	ts.RunTests(t, func(c *mysql.Config) {
+		c.AllowAllFiles = true
+	}, func(dbt *testkit.DBTestKit) {
+		ctx := context.Background()
+
+		// Generate a big local file
+		filePath := filepath.Join(t.TempDir(), "big_file.txt")
+		mysql.RegisterLocalFile(filePath)
+		for i := range 1000 {
+			err := os.WriteFile(filePath, fmt.Appendf(nil, "%d\n", i), os.ModePerm)
+			require.NoError(t, err)
+		}
+
+		conn, err := dbt.GetDB().Conn(ctx)
+		require.NoError(t, err)
+
+		testserverclient.MustExec(ctx, t, conn, "create table t (a int)")
+
+		testfailpoint.Enable(t, "github.com/pingcap/tidb/pkg/executor/CommitWorkError", "return(1)")
+
+		testStart := time.Now()
+		testDuration := time.Second
+		for time.Since(testStart) < testDuration {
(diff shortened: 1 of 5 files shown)


pkg/server/tests/commontest/tidb_test.go
```diff
@@ -3534,3 +3534,38 @@ func TestCloseConnForUndeterminedError(t *testing.T) {
...
+			err := os.WriteFile(filePath, fmt.Appendf(nil, "%d\n", i), os.ModePerm)
+			require.NoError(t, err)
+		}
+
+		conn, err := dbt.GetDB().Conn(ctx)
+		require.NoError(t, err)
+
+		testserverclient.MustExec(ctx, t, conn, "create table t (a int)")
+
+		err = failpoint.Enable("github.com/pingcap/tidb/pkg/executor/InCommitWork", "panic(\"mock panic\")")
```
lance6716: 1. we have `testfailpoint` package to simplify these lines
2. Why this is panic? The effect of panic is not clear, maybe `sleep` is better

YangKeao: > we have testfailpoint package to simplify these lines

Thanks! Modified.

> Why this is panic? The effect of panic is not clear, maybe sleep is better

Because this issue will occur when `(w *commitWorker) commitWork` returns an error. 

I agree that the `panic` is not clear. I've modified it into:

```go
failpoint.Inject("CommitWorkError", func(_ failpoint.Value) {
	failpoint.Return(errors.New("mock commit work error"))
})
```

lance6716:
```
pkg/executor/test/loaddatatest/load_data_test.go:46:54: cannot use (func(_ string) (r io.ReadCloser, err error) literal) (value of type func(_ string) (r io.ReadCloser, err error)) as executor.LoadDataReaderBuilder value in variable declaration
```

Issue #59333, opened 2025-02-07
Optimizer should recognize risk in cardinality estimation


terry1purcell:
## Enhancement

Combining multiple predicates may exploit single column statistics (NDVs, topN, buckets) and index NDVs - with the combination of selectivities using either an exponential backoff or independence assumption. Neither of the backoff or independence are supported by statistics. It would be beneficial for the optimizer to acknowledge when an estimate was made with such assumptions.

Using the following DDL and insert for test purposes:

CREATE TABLE t2 (
a INT PRIMARY KEY,
b int,
c int,
d int,
e int,
Key (b, c, a),
key (e, d, a)
);

set @@cte_max_recursion_depth=10000000;
INSERT INTO t2 (a, b, c, d, e)
SELECT a, mod(a, 1000) AS b, mod(a, 1000) AS c, mod(a, 10000) as d, mod(a, 2) as e
FROM (
WITH RECURSIVE x AS (
SELECT 1 AS a
UNION ALL
SELECT a + 1 AS a
FROM x
WHERE a < 1000000
)
SELECT a
FROM x
) AS subquery;
ANALYZE TABLE t2;

Query examples

tidb> explain select * from t2 where b = 0 and c = 0 and d = 0 and e = 0;
+-------------------------------+---------+-----------+----------------------------+------------------------------------+
| id | estRows | task | access object | operator info |
+-------------------------------+---------+-----------+----------------------------+------------------------------------+
| IndexLookUp_11 | 0.00 | root | | |
| ├─IndexRangeScan_8(Build) | 31.45 | cop[tikv] | table:t2, index:b(b, c, a) | range:[0 0,0 0], keep order:false |
| └─Selection_10(Probe) | 0.00 | cop[tikv] | | eq(test.t2.d, 0), eq(test.t2.e, 0) |
| └─TableRowIDScan_9 | 31.45 | cop[tikv] | table:t2 | keep order:false |
+-------------------------------+---------+-----------+----------------------------+------------------------------------+
4 rows in set (0.01 sec)

tidb> select count() from t2 where b = 0 and c = 0;
+----------+
| count() |
+----------+
| 1000 |
+----------+
1 row in set (0.01 sec)

tidb> select count() from t2 where d = 0 and e = 0;
+----------+
| count() |
+----------+
| 100 |
+----------+
tidb> explain select * from t2 use index (e) where b = 0 and c = 0 and d = 0 and e = 0;
[...]

PR #62952, opened 2025-08-12
planner: account for risk of using skew ratio vars

ichen032:

Issue Number: ref #59333

Problem Summary:

Simplify the usage (and accumulation) of min and max estimated row counts. This creates an infrastructure to allow the max and min to be compared against the base row estimates for identifying the estimation risk of each index choice. While this is limited to indexes today, it can be expanded to more plan steps in future.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


terry1purcell (approved):
/lgtm

pkg/planner/cardinality/cross_estimation.go
```diff
@@ -179,8 +179,10 @@ func crossEstimateRowCount(sctx planctx.PlanContext,
 		return path.CountAfterAccess, true, 0
 	}
 	var rangeCount float64
+	var tempResult statistics.RowEstimate
```
Tristan1900: tempResult declared here but not used anywhere, perhaps can move it into the if clause

pkg/planner/cardinality/cross_estimation.go
```diff
@@ -199,14 +201,16 @@ func crossEstimateRowCount(sctx planctx.PlanContext,
 func getColumnRangeCounts(sctx planctx.PlanContext, colID int64, ranges []*ranger.Range, histColl *statistics.HistColl, idxID int64) (rangeCounts []float64, minCount float64, maxCount float64, ok bool) {
 	var err error
 	var count float64
+	var tempResult statistics.RowEstimate
```
Tristan1900: same here

pkg/planner/cardinality/row_count_index.go
```diff
@@ -87,14 +90,15 @@ func GetRowCountByIndexRanges(sctx planctx.PlanContext, coll *statistics.HistCol
 		)
 	}
 	if idx.CMSketch != nil && idx.StatsVer == statistics.Version1 {
-		result, err = getIndexRowCountForStatsV1(sctx, coll, idxID, indexRanges)
+		count, err = getIndexRowCountForStatsV1(sctx, coll, idxID, indexRanges)
+		result = statistics.RowEstimate{Est: count, MinEst: 0, MaxEst: 0}
```
Tristan1900: curious what's the reason for min and max to be 0 for  StatsV1, I see other places they are equal to the count

pkg/planner/cardinality/row_count_index.go
```diff
@@ -178,6 +183,7 @@ func getIndexRowCountForStatsV1(sctx planctx.PlanContext, coll *statistics.HistC
 			}
 			var count float64
 			var err error
+			var tempResult statistics.RowEstimate
```
Tristan1900: same here

pkg/planner/core/find_best_task.go
```diff
@@ -1158,12 +1158,12 @@ func compareCandidates(sctx base.PlanContext, statsTbl *statistics.Table, tableI
 		// has the same or higher number of equal/IN predicates.
 		if !lhsPseudo && globalResult >= 0 && sum >= 0 &&
 			lhs.path.EqOrInCondCount > 0 && lhs.path.EqOrInCondCount >= rhs.path.EqOrInCondCount &&
-			(rhs.path.MaxCountAfterAccess <= 0 || lhs.path.CountAfterAccess < rhs.path.MaxCountAfterAccess) {
+			(rhs.path.MaxCountAfterAccess <= rhs.path.CountAfterAccess || lhs.path.CountAfterAccess < rhs.path.MaxCountAfterAccess) {
```
Tristan1900, line 1162: Just for me, I am not very clear of what it means to compare MaxCountAfterAccess with either 0 or CountAfterAccess, a comment or make a method with good name will be very helpful

Tristan1900 (commented):
Just a few minor questions! Thanks for the change!

pkg/planner/cardinality/row_count_index.go
terry1purcell: I am OK with either setting 0 for stats V1, or setting it to the same as V2. We are really trying NOT to enhance stats V1, because we want customers to be motivated to move to stats V2. In future - it'd be great to deprecate stats V1 and clean up the redundant code.

pkg/planner/core/find_best_task.go
terry1purcell: Initially - MaxCountAfterAccess was only set in some situations where there was risk (but not all). Therefore, a setting of zero meant that there was either no risk, or the type of risk wasn't tracked. Now this will always be set, and if there is no risk, it will be set to the same value as CountAfterAccess.

henrybw (approved):

PR #63002, opened 2025-08-14
unistore: Add mock keyspace manager to mock PD

MyonKeminta:
### What problem does this PR solve?

Issue Number: ref #58720

Problem Summary:

### What changed and how does it work?

This PR adds a simple mock implementation of keyspace manager to mock PD of unistore.

A controllable keyspace meta source is needed for writing unit tests for GC worker. Instead of implementing a mock locally only for GC worker, let's directly make it into the mock PD to make it reuseable.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #63003, opened 2025-08-14
gc_worker: Fix resolve locks which doesn't work properly for null keyspace when there is other keyspaces and adapt some old tests

MyonKeminta:

Issue Number: ref #58720

Problem Summary:

The current implementation of GCWorker is not completely correct. When it works in the null keyspace, and the cluster contains other keyspaces, it doesn't handle the ranges properly. We expect it to:

* Resolve locks for the null keyspace
* Resolve locks for keyspaces that was configured running unified GC
* Do not resolve locks for keyspaces that was configured running keyspace level GC
* Do not resolve locks for raw kv (v2) ranges.

It's not that important though, as currently the next gen kernel won't use the null keyspace and classic kernel does not use keysapces. But on the other hands, the current GC procedure is designed to work no matter how keyspaces are used, so that it can then also work properly on serverless without too much change, and minimizes the problem we might encounter when the usage of keyspaces is changed in the future. We need to fix it.

This PR also fixes some related existing tests.

Requires
* https://github.com/pingcap/tidb/pull/63002
* https://github.com/tikv/client-go/pull/1720

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/store/gcworker/gc_worker.go
```diff
@@ -175,6 +176,8 @@ const (
 	tidbGCLeaderLease = "tidb_gc_leader_lease"
 	tidbGCLeaderUUID  = "tidb_gc_leader_uuid"
 	tidbGCSafePoint   = "tidb_gc_safe_point"
+
+	loadAllKeyspacesForUnifiedGCTestsBatchSize = 50
```
ekexium: What does the "tests" mean in the name?

pkg/store/gcworker/gc_worker.go
```diff
@@ -1236,21 +1232,119 @@ func (w *GCWorker) resolveLocks(
...
+		metrics.GCHistogram.WithLabelValues("resolve_locks").Observe(time.Since(startTime).Seconds())
+		return nil
 	}
 
-	logutil.Logger(ctx).Info("finish resolve locks", zap.String("category", "gc worker"),
-		zap.String("uuid", w.uuid),
-		zap.Uint64("safePoint", safePoint),
-		zap.Int("regions", runner.CompletedRegions()))
-	metrics.GCHistogram.WithLabelValues("resolve_locks").Observe(time.Since(startTime).Seconds())
+	// Otherwise, the null keyspace, which is the current keyspace, has the responsibility to resolve locks forr
```
ekexium:
```suggestion
	// Otherwise, the null keyspace, which is the current keyspace, has the responsibility to resolve locks for
```

pkg/store/gcworker/gc_worker.go
```diff
@@ -1236,21 +1232,119 @@ func (w *GCWorker) resolveLocks(
...
+		if !(len(nullKeyspaceExcludePrefixes) > 0 && len(nextStartKey) == 0) {
+			nullKeyspaceKeyRanges = append(nullKeyspaceKeyRanges, tikvstore.KeyRange{
+				StartKey: nextStartKey,
+				EndKey:   []byte(""),
+			})
+		}
+
+		for _, r := range nullKeyspaceKeyRanges {
+			err := runner.RunOnRange(ctx, r.StartKey, r.EndKey)
+			if err != nil {
```
ekexium, line 1303: We can consider adding metrics for such errors later

pkg/store/gcworker/gc_worker.go
```diff
@@ -1212,13 +1208,13 @@ func (w *GCWorker) checkUseDistributedGC() bool {
 
 func (w *GCWorker) resolveLocks(
```
ekexium, line 1209: Nitpick: the function is long and I would personally prefer splitting it into multiple smaller ones.

pkg/store/gcworker/gc_worker.go
```diff
@@ -1236,21 +1232,119 @@ func (w *GCWorker) resolveLocks(
...
+		// If there are prefixes that has been processed but `nextStartKey` is set to empty, it means that there
+		// exist a prefix full of 0xff, causing the PrefixNextKey founds the global end. So no remaining range
+		// is needed to be added.
+		// Currently, there's no such prefix in use, but we handle it in code for strictness.
+		if !(len(nullKeyspaceExcludePrefixes) > 0 && len(nextStartKey) == 0) {
+			nullKeyspaceKeyRanges = append(nullKeyspaceKeyRanges, tikvstore.KeyRange{
+				StartKey: nextStartKey,
+				EndKey:   []byte(""),
+			})
+		}
```
ekexium: Seems to me this paragraph is to add a the final range of [succ(prefix), nil). But the comments are not obvious enough to me... At first sight it makes me think this `if` is an edge case that won't be triggered in current implementation.
Am I understanding it correctly?

ekexium (approved):
Generally LGTM. Nice comments and tests.
Please update client-go and resolve conflicts.

pkg/store/gcworker/gc_worker.go
MyonKeminta: typo...

pkg/store/gcworker/gc_worker.go
MyonKeminta: > this paragraph is to add a the final range of [succ(prefix), nil)

Yes

>  this if is an edge case that won't be triggered in current implementation

Yes

PR #63004, opened 2025-08-14
stats: optimize stats table copy to save memory

### LINKED ISSUE #63076: stats: optimize stats table copy usage
## Enhancement

To update statsCache, code uses table.copy() to update and then do cache.put(). The copy is a deep copy and memory intensive, we noticed high memory usage in TiDB memory profile. 

We should look at those copy case by case and see if they are really needed

Tristan1900:

Issue Number: close #63076

Problem Summary:

1. remove copy in InitStatsLite as it's single thread update, so can update in place
2. postpone deep copy in TableStatsFromStorage and use shallow copy if applicable
3. add copyForCol and copyForIndex method to only copy relevant components instead of everything and apply to updateCachedItem and TableStatsFromStorage
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/handle/bootstrap.go
+++ b/pkg/statistics/handle/bootstrap.go
@@ -139,7 +139,6 @@ func (*Handle) initStatsHistograms4ChunkLite(cache statstypes.StatsCache, iter *
 			if !ok {
 				continue
 			}
-			table = table.Copy()
 		}
 		isIndex := row.GetInt64(1)
 		id := row.GetInt64(2)
--- a/pkg/statistics/handle/bootstrap.go
+++ b/pkg/statistics/handle/bootstrap.go
@@ -188,7 +188,6 @@ func (h *Handle) initStatsHistograms4Chunk(is infoschema.InfoSchema, cache stats
 			if !ok {
 				continue
 			}
-			table = table.Copy()
 		}
 		// All the objects in the table share the same stats version.
 		if statsVer != statistics.Version0 {
@@ -395,7 +394,6 @@ func (*Handle) initStatsTopN4Chunk(cache statstypes.StatsCache, iter *chunk.Iter
 			if !ok {
 				continue
 			}
-			table = table.Copy()
 		}
 		idx := table.GetIdx(row.GetInt64(1))
 		if idx == nil || (idx.CMSketch == nil && idx.StatsVer <= statistics.Version1) {
@@ -568,7 +566,6 @@ func (*Handle) initStatsBuckets4Chunk(cache statstypes.StatsCache, iter *chunk.I
 			if !ok {
 				continue
 			}
-			table = table.Copy()
 		}
 		var lower, upper types.Datum
 		var hist *statistics.Histogram
--- a/pkg/statistics/handle/bootstrap.go
+++ b/pkg/statistics/handle/bootstrap.go
@@ -139,6 +139,7 @@ func (*Handle) initStatsHistograms4ChunkLite(cache statstypes.StatsCache, iter *
 			if !ok {
 				continue
 			}
+			table = table.ShallowCopy(statistics.CopyWithBothMaps)
 		}
 		isIndex := row.GetInt64(1)
 		id := row.GetInt64(2)
@@ -188,6 +189,7 @@ func (h *Handle) initStatsHistograms4Chunk(is infoschema.InfoSchema, cache stats
 			if !ok {
 				continue
 			}
+			table = table.ShallowCopy(statistics.CopyWithBothMaps)
 		}
 		// All the objects in the table share the same stats version.
 		if statsVer != statistics.Version0 {
@@ -394,6 +396,9 @@ func (*Handle) initStatsTopN4Chunk(cache statstypes.StatsCache, iter *chunk.Iter
 			if !ok {
 				continue
 			}
+			// existing idx is modified so if using shallow copy the modified idx is shared and will mess up the cache
+			// cost calculation
+			table = table.Copy()
 		}
 		idx := table.GetIdx(row.GetInt64(1))
 		if idx == nil || (idx.CMSketch == nil && idx.StatsVer <= statistics.Version1) {
@@ -566,6 +571,9 @@ func (*Handle) initStatsBuckets4Chunk(cache statstypes.StatsCache, iter *chunk.I
 			if !ok {
 				continue
 			}
+			// existing idx is modified so if using shallow copy the modified idx is shared and will mess up the cache
+			// cost calculation
+			table = table.Copy()
 		}
 		var lower, upper types.Datum
 		var hist *statistics.Histogram
--- a/pkg/statistics/handle/bootstrap.go
+++ b/pkg/statistics/handle/bootstrap.go
@@ -139,7 +139,11 @@ func (*Handle) initStatsHistograms4ChunkLite(cache statstypes.StatsCache, iter *
 			if !ok {
 				continue
 			}
-			table = table.CopyAs(statistics.MetaOnly)
+			// optimization: doesn't need to copy and can do in place changes since
+			// 1. initStatsHistograms4ChunkLite is a single thread populating entries in a local cache, and later the
+			// cache will be called h.Replace(cache) to be the global cache
+			// 2. following logic only modifies ColAndIdxExistenceMap, so it won't impact the cache memory cost
+			// calculation
 		}
 		isIndex := row.GetInt64(1)
 		id := row.GetInt64(2)
@@ -396,8 +400,7 @@ func (*Handle) initStatsTopN4Chunk(cache statstypes.StatsCache, iter *chunk.Iter
 			if !ok {
 				continue
 			}
-			// existing idx is modified so if using shallow copy the modified idx is shared and will mess up the cache
-			// cost calculation
+			// existing idx histogram is modified have to use deep copy
 			table = table.CopyAs(statistics.AllDataWritable)
 		}
 		idx := table.GetIdx(row.GetInt64(1))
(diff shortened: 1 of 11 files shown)


pkg/statistics/handle/storage/read.go
```diff
@@ -763,7 +783,7 @@ func loadNeededColumnHistograms(sctx sessionctx.Context, statsHandle statstypes.
 		)
 		return nil
 	}
-	statsTbl = statsTbl.Copy()
+	statsTbl = statsTbl.CopyForColumnUpdate()
```
hawkingrei: It can be extremely dangerous. In Golang, if you perform concurrent read and write operations on a map, it may cause a panic.

hawkingrei: Additionally, there is actually a possibility that by using SetFinalizer to automatically clean up when no one is using it, we might achieve higher memory reuse.

hawkingrei: https://pkg.go.dev/runtime#SetFinalizer

hawkingrei: https://pkg.go.dev/runtime#AddCleanup

Tristan1900: thanks for the review. 
forgive my ignorance, which map are we referring to? My thought is only copy what's needed for the update, in this call only col related map are updated so I thought only copy for that. Do I miss any map that needs to be copied in this case?

Tristan1900:
almost all the copies are gone

before
image

after

image

pkg/statistics/table.go
```diff
@@ -662,6 +663,85 @@ func (t *Table) ShallowCopy() *Table {
...
+		LastStatsHistVersion:  t.LastStatsHistVersion,
+	}
+	return nt
+}
+
+// CopyForIndexUpdate creates a copy optimized for index updates.
+func (t *Table) CopyForIndexUpdate() *Table {
+	newIndices := make(map[int64]*Index, len(t.indices))
+	for id, idx := range t.indices {
+		newIndices[id] = idx
```
Tristan1900: I noticed in deep copy, we copy the idx object as well, such as 
`newHistColl.indices[id] = idx.Copy()`

And I wonder if that's necessary for all our use cases since it seems our updates are adding/removing new col instead of modifying the existing col content, I wonder if for the other copy cases in the repo can we avoid the idx/col.copy()?

pkg/statistics/handle/storage/read.go
Tristan1900: discussed offline. The map should be safe since it's copied before the update. Should be no concurrent access.

AddCleanup is potentially a good tool but out of the scope of this PR, will explore to use it in another PR

pkg/statistics/handle/storage/read.go
```diff
@@ -505,6 +505,10 @@ func TableStatsFromStorage(sctx sessionctx.Context, snapshot uint64, tableInfo *
 	tracker := memory.NewTracker(memory.LabelForAnalyzeMemory, -1)
 	tracker.AttachTo(sctx.GetSessionVars().MemTracker)
 	defer tracker.Detach()
+
+	var needsCopy bool
+	var originalTable *statistics.Table
```
time-and-fate: I think we actually don't need this new variable. `table` should be enough.

pkg/statistics/table.go
```diff
@@ -662,6 +663,85 @@ func (t *Table) ShallowCopy() *Table {
 	return nt
 }
 
+// CopyForColumnUpdate creates a copy optimized for column updates.
+func (t *Table) CopyForColumnUpdate() *Table {
+	newColumns := make(map[int64]*Column, len(t.columns))
+	for id, col := range t.columns {
+		newColumns[id] = col
+	}
```
time-and-fate: `maps.Clone()` is cleaner.

pkg/statistics/table.go
```diff
@@ -662,6 +663,85 @@ func (t *Table) ShallowCopy() *Table {
...
+	return nt
+}
+
+// CopyForMapUpdate creates a copy optimized for map-only updates like DelCol/DelIdx.
+func (t *Table) CopyForMapUpdate() *Table {
+	newHistColl := HistColl{
+		PhysicalID:    t.PhysicalID,
+		RealtimeCount: t.RealtimeCount,
+		columns:       t.columns,
+		indices:       t.indices,
```
time-and-fate: These two fields are also modified in `DelCol` and `DelIdx`. Don't we need to clone them?

pkg/statistics/handle/storage/read.go
Tristan1900: Tracked by https://github.com/pingcap/tidb/issues/63184

pkg/statistics/table.go
```diff
@@ -662,6 +663,75 @@ func (t *Table) ShallowCopy() *Table {
...
+		ExtendedStats:         t.ExtendedStats,
+		ColAndIdxExistenceMap: t.ColAndIdxExistenceMap.Clone(),
+		LastAnalyzeVersion:    t.LastAnalyzeVersion,
+		LastStatsHistVersion:  t.LastStatsHistVersion,
+	}
+	return nt
+}
+
+// CopyForMapsUpdate creates a copy for both col and index map updates.
+func (t *Table) CopyForMapsUpdate() *Table {
```
0xPoe: What is the difference between this function and copy? I think they are the same, right?

0xPoe (commented):
The changes look very good. Could you please add some tests to cover the different copy behaviors? Otherwise, I think it would be very easy to accidentally break or bypass the current design.

pkg/statistics/table.go
0xPoe: Oh, I see. This actually is a shallow clone for the key and value.

pkg/statistics/table.go
```diff
@@ -662,6 +663,75 @@ func (t *Table) ShallowCopy() *Table {
...
+		TblInfoUpdateTS:       t.TblInfoUpdateTS,
+		ExtendedStats:         t.ExtendedStats,
+		ColAndIdxExistenceMap: t.ColAndIdxExistenceMap.Clone(),
+		LastAnalyzeVersion:    t.LastAnalyzeVersion,
+		LastStatsHistVersion:  t.LastStatsHistVersion,
+	}
+	return nt
+}
+
+// CopyForMapsUpdate creates a copy for both col and index map updates.
```
0xPoe: It would be great if we could specify when to use these functions accordingly in the comments. Thank you!

pkg/statistics/handle/bootstrap.go
```diff
@@ -139,7 +139,6 @@ func (*Handle) initStatsHistograms4ChunkLite(cache statstypes.StatsCache, iter *
 			if !ok {
 				continue
 			}
-			table = table.Copy()
```
0xPoe, line 142: Can we also do this for non-lite init stats?

pkg/statistics/table.go
time-and-fate: I think the 4 similar functions (3 new functions + `ShallowCopy()`) can be simplified to one function: `Table.ShallowCopy(forColumnMapUpdate, forIndexMapUpdate bool)`, and the behavior is that:
- `forColumnMapUpdate` => `maps.Clone(t.columns)`
- `forIndexMapUpdate` => `maps.Clone(t.indices)`
- `forColumnMapUpdate || forIndexMapUpdate` => `t.ColAndIdxExistenceMap.Clone()`
- `!forColumnMapUpdate && !forIndexMapUpdate` => current `ShallowCopy()`

This makes `ShallowCopy()` a bit more complicated, but overall they are easier to understand.

What do you think of it? @Tristan1900 @0xPoe

Tristan1900: yeah that's a great idea. Will do

Tristan1900:
/hold

the cost model for LFU cache kind of breaks since the underlying value is not copied, which results in cost of the cache being 0

pkg/statistics/table.go
0xPoe: Good idea. I think we can further merge the Copy and ShallowCopy together. Then, every time somebody wants to use this API, they need to consider which update policy applies to this caller. This would make it difficult to misuse the Copy and ShallowCopy. Every caller has to know what the requirements are here.

time-and-fate (approved):

pkg/statistics/table.go
```diff
@@ -599,53 +622,70 @@ func (t *Table) MemoryUsage() *TableMemoryUsage {
...
+		existenceMap = t.ColAndIdxExistenceMap
+	case ColumnMapWritable:
+		columns = maps.Clone(t.columns)
+		indices = t.indices
+		if t.ColAndIdxExistenceMap != nil {
+			existenceMap = t.ColAndIdxExistenceMap.Clone()
+		}
+	case IndexMapWritable:
+		columns = t.columns
+		indices = maps.Clone(t.indices)
```
hawkingrei, line 655: Can we create a pool to reuse the map here?

Tristan1900: curious how do we reuse map? will follow up offline

Tristan1900: discussed offline, created issue to track
https://github.com/pingcap/tidb/issues/63239

0xPoe:
@Tristan1900 Do we need to unhold it?

Tristan1900:
> @Tristan1900 Do we need to unhold it?

yes, upholding now, just want to take a final look before merging. thanks

/unhold

PR #62742, opened 2025-07-31
test: use SYSTEM keyspace to create store and domain in nextgen

D3Hunter:
### What problem does this PR solve?

Issue Number: ref #61702

Problem Summary:

### What changed and how does it work?
in nextgen, all stores must have a keyspace meta set. to simplify the test, we set the default keyspace meta to `SYSTEM` keyspace.
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)

**ONLY CHANGES TESTS**

run unit test in `session` and other pkg which are creating store and domain in nextgen, pass

- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #57865, opened 2024-12-02
keyspace: support API V2 mock store

ystaticy:
### What problem does this PR solve?

Issue Number: ref https://github.com/pingcap/tidb/issues/57969

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
support keyspace mock store, and add UT case.
```

PR #63043, opened 2025-08-18
test: fix standalone TiDB of next-gen report 'invalid key'

D3Hunter:

Issue Number: ref #61702

Problem Summary:

https://github.com/pingcap/tidb/pull/57865 add the `RemoveKeyspacePrefix` for uni-store, but all integration tests in standalone mode are not using keyspace, it means there are **NO** tests covers the logic in the caller `DecodeRowKey`'s check about keys, i.e at that time there is no key with keyspace.
after https://github.com/pingcap/tidb/pull/62742, we are creating the uni-store with SYSTEM keyspace, so keys will have keyspace prefix, but not removed by `RemoveKeyspacePrefix`, we fix it by introducing a variable `kv.StandAloneTiDB` which is `true` only when run in standalone TiDB, and will remove the keyspace prefix.

we also make the keyspace-meta related test only run in NextGen, Classic kernel doesn't support keyspace.
### Check List

Tests 

- [x] Unit test
- [ ] Integration test

before this PR, run standalone TiDB of next-gen always report `[tikv:8221]invalid key - \"x\\x00\\x00\\x01t\\x80\\x00\\xff\\xff\\xff\\xff\\xff\` during bootstrap

after this PR, standalone TiDB bootstraps normally, run some queries, and it works without error.

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/cmd/tidb-server/main.go
+++ b/cmd/tidb-server/main.go
@@ -474,6 +474,9 @@ func registerStores() error {
 }
 
 func createStoreDDLOwnerMgrAndDomain(keyspaceName string) (kv.Storage, *domain.Domain, error) {
+	if config.GetGlobalConfig().Store == config.StoreTypeUniStore {
+		kv.StandAloneTiDB = true
+	}
 	storage := kvstore.MustInitStorage(keyspaceName)
 	if tikvStore, ok := storage.(kv.StorageWithPD); ok {
 		pdhttpCli := tikvStore.GetPDHTTPClient()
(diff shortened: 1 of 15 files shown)


joechenrh (approved):

tiancaiamao (approved):
Keyspace will only be used with the nextgen build, in classic mode it will not be used.
You can leave some comment if you have any concern.

/cc @ystaticy

PR #63047, opened 2025-08-18
planner: Show Execution Plans for Scalar Subqueries | tidb-test=pr/2585

### LINKED ISSUE #63045: Show Execution Plans for Scalar Subqueries
### Background  
Currently in TiDB, **non-correlated scalar subqueries** are executed early during the planner phase.  
This behavior introduces two issues:  
1. In `EXPLAIN`, the execution plan for these subqueries is not displayed.  
2. In actual executed statements (such as slow query logs, statement summary, or `EXPLAIN ANALYZE`), the execution plan for these subqueries is also missing.  

To address the first issue, a system variable was previously introduced by Yiding:  
- `tidb_opt_enable_non_eval_scalar_subquery`  
- Default: OFF  
- When enabled, it prevents the planner from pre-evaluating non-correlated scalar subqueries, ensuring that `EXPLAIN` always shows the complete execution plan.  

However, the second issue still remains unresolved.  

### Problem  
- For **actually executed SQL statements**, the execution plans of subqueries that were pre-executed are not recorded or displayed.  
- As a result, in slow query logs, statement summary, or `EXPLAIN ANALYZE`, these parts of the plan are missing.  
- This makes troubleshooting and performance tuning more difficult.  

### Proposal / Improvement  
- Build upon the existing `tidb_opt_enable_non_eval_scalar_subquery` logic to extend the execution plan collection mechanism.  
- Ensure that for actual executed statements (slow query, statement summary, `EXPLAIN ANALYZE`), the execution plan of pre-executed subqueries is also collected and displayed.  
- The solution should remain transparent to end users and avoid introducing significant overhead.  

### Reference  
- Implementation of `tidb_opt_enable_non_eval_scalar_subquery`.  

### Expected Benefits  
- **Consistency**: Execution plans are complete in both `EXPLAIN` and actual execution.  
- **Observability**: DBAs and developers can more easily analyze and troubleshoot issues.  
- **Usability**: Prevents users from being misled by missing parts of the plan.

King-Dylan:

Issue Number: close #63045

Problem Summary:
For actually executed SQL statements, the execution plans of subqueries that were pre-executed are not recorded or displayed. As a result, in slow query logs, statement summary, or EXPLAIN ANALYZE, these parts of the plan are missing.
This makes troubleshooting and performance tuning more difficult.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/domain/domain_sysvars.go
+++ b/pkg/domain/domain_sysvars.go
@@ -86,7 +86,6 @@ func (do *Domain) setPDClientDynamicOption(name, sVal string) error {
 		if err != nil {
 			return err
 		}
-
 		vardef.EnablePDFollowerHandleRegion.Store(val)
 	case vardef.TiDBTSOClientRPCMode:
 		var concurrency int


King-Dylan:
https://github.com/pingcap/tipb/pull/377

pkg/planner/core/scalar_subq_expression.go
```diff
@@ -324,14 +329,21 @@ func (*ScalarSubqueryEvalCtx) Schema() *expression.Schema {
...
+	if len(s.outputColIDs) > 0 {
+		builder.WriteString("Output: ")
+		for i, id := range s.outputColIDs {
+			fmt.Fprintf(builder, "ScalarQueryCol#%d", id)
+			if i+1 != len(s.outputColIDs) {
+				builder.WriteString(", ")
+			}
 		}
+	} else {
+		builder.WriteString("Output: N/A")
```
time-and-fate: When would this happen?

pkg/expression/explain.go
```diff
@@ -191,8 +200,16 @@ func ExplainExpressionList(ctx EvalContext, exprs []Expression, schema *Schema,
 				builder.WriteString(schema.Columns[i].StringWithCtx(ctx, redactMode))
 			}
 		case *Constant:
-			v := expr.StringWithCtx(ctx, errors.RedactLogDisable)
-			redact.WriteRedact(builder, v, redactMode)
+			// For Projection operators, show only the subquery reference without the constant value
```
time-and-fate: Why?

pkg/planner/core/common_plans.go
```diff
@@ -1189,6 +1189,20 @@ func getRuntimeInfo(ctx base.PlanContext, p base.Plan, runtimeStatsColl *execdet
...
+			rootStats = runtimeStatsColl.GetRootStats(subqueryID)
+		}
+		if runtimeStatsColl != nil && runtimeStatsColl.ExistsCopStats(subqueryID) {
+			copStats = runtimeStatsColl.GetCopStats(subqueryID)
+		}
+		memTracker = ctx.GetSessionVars().StmtCtx.MemTracker.SearchTrackerWithoutLock(subqueryID)
+		diskTracker = ctx.GetSessionVars().StmtCtx.DiskTracker.SearchTrackerWithoutLock(subqueryID)
+		return rootStats, copStats, memTracker, diskTracker
+	}
+
```
time-and-fate: Looks like the runtime info of the actual plan tree of the subquery will be printed completely without this, so I think `ScalarSubqueryEvalCtx` itself doesn't need runtime info.

pkg/planner/core/expression_rewriter.go
```diff
@@ -1360,6 +1376,22 @@ func (er *expressionRewriter) handleScalarSubquery(ctx context.Context, planCtx
...
+			// If preserving subquery plan, add reference information to the constant
+			var matchedSubQ *ScalarSubqueryEvalCtx
+			for _, scalarSubQ := range planCtx.builder.ctx.GetSessionVars().MapScalarSubQ {
+				if subQ, ok := scalarSubQ.(*ScalarSubqueryEvalCtx); ok {
+					if subQ.ScalarSubQuery().ID() == physicalPlan.ID() {
+						matchedSubQ = subQ
+						break
+					}
+				}
+			}
```
time-and-fate: I think we don't need this since we can use `newColIDs` directly.

pkg/planner/core/expression_rewriter.go
```diff
@@ -1374,6 +1406,22 @@ func (er *expressionRewriter) handleScalarSubquery(ctx context.Context, planCtx
...
+		// If preserving subquery plan, add reference information to the constant
+		var matchedSubQ *ScalarSubqueryEvalCtx
+		for _, scalarSubQ := range planCtx.builder.ctx.GetSessionVars().MapScalarSubQ {
+			if subQ, ok := scalarSubQ.(*ScalarSubqueryEvalCtx); ok {
+				if subQ.ScalarSubQuery().ID() == physicalPlan.ID() {
+					matchedSubQ = subQ
+					break
+				}
+			}
+		}
```
time-and-fate: ditto.

pkg/expression/explain.go
King-Dylan: Since I think showing the column mapping is sufficient. For example: (ScalarQueryCol#14) -> Column#15 vs (ScalarQueryCol#14) -> 55 -> Column#15.

pkg/planner/core/common_plans.go
King-Dylan: You are right, they are using the same plan id allocator.

pkg/planner/core/expression_rewriter.go
King-Dylan: Ok got it

pkg/planner/core/expression_rewriter.go
King-Dylan: Done

pkg/planner/core/scalar_subq_expression.go
```diff
@@ -72,6 +72,11 @@ func (ssctx *ScalarSubqueryEvalCtx) selfEval() error {
 	return nil
 }
 
+// ScalarSubQuery returns the physical plan of the subquery
+func (ssctx *ScalarSubqueryEvalCtx) ScalarSubQuery() base.PhysicalPlan {
+	return ssctx.scalarSubQuery
+}
+
```
time-and-fate: It's useless now.

pkg/expression/explain.go
```diff
@@ -191,8 +200,16 @@ func ExplainExpressionList(ctx EvalContext, exprs []Expression, schema *Schema,
...
+			if expr.SubqueryRefID > 0 {
+				refStr := fmt.Sprintf("ScalarQueryCol#%d", expr.SubqueryRefID)
+				builder.WriteString("(")
+				builder.WriteString(refStr)
+				builder.WriteString(")")
```
time-and-fate: I think the parentheses are inconsistent with other places.

Image

pkg/planner/core/casetest/scalarsubquery/cases_test.go
```diff
@@ -91,3 +91,82 @@ func TestExplainNonEvaledSubquery(t *testing.T) {
...
+			input []struct {
+				SQL              string
+				IsExplainAnalyze bool
+				HasErr           bool
+			}
+			output []struct {
+				SQL   string
+				Plan  []string
+				Error string
+			}
```
time-and-fate: `HasErr` and `Error` seem useless.

pkg/expression/constant_fold.go
```diff
@@ -248,7 +248,15 @@ func foldConstant(ctx BuildContext, expr Expression) (Expression, bool) {
 		if isDeferredConst {
 			return &Constant{Value: value, RetType: retType, DeferredExpr: x}, true
 		}
-		return &Constant{Value: value, RetType: retType}, false
+		// Extract SubqueryRefID from function arguments
+		var subqRef int64
+		for _, arg := range x.GetArgs() {
+			if c, ok := arg.(*Constant); ok && c.SubqueryRefID > 0 {
+				subqRef = c.SubqueryRefID
+				break
```
windtalker, line 256: what happens when there are more than 1 constant arguments with valid SubqueryRefID ?

pkg/expression/explain.go
windtalker: Before your pr, it will show `55`?

pkg/expression/constant_fold.go
King-Dylan: Will output like: 
```
ScalarSubQuery	N/A	root		Output: ScalarQueryCol#14, ScalarQueryCol#15, ScalarQueryCol#16
```

pkg/expression/explain.go
King-Dylan: Yes, what I’m thinking is that, rather than the exact numbers, what we really need to know is which subquery influenced that operator, or in other words, their causal relationship. 
Before:
```
55 -> Column#15
```
After this pr:
```
ScalarQueryCol#14 -> Column#15
```

windtalker: Maybe append the constant value after ScalarQueryCol#14? like 
`ScalarQueryCol#14(55) -> Column#15`

pkg/expression/constant_fold.go
windtalker: What I mean is during constant folding, if a function has more than 1 arguments which are coming from scalar subquery, then that is that looks like? In your code logic, it seems only display the first argument's source?

King-Dylan: Ok, the in list case actually won’t reach this logic because the earlier call will exit in advance. For eq, le, and similar cases, only one  SubqueryRefID > 0, so that’s fine. For + - * /, (select a from t where b = sub1+sub2-sub3) If we display all the subqueries that contribute to it, it would look really messy, so we just take the first one.

pkg/expression/explain.go
```diff
@@ -141,20 +141,32 @@ func (col *Column) ExplainNormalizedInfo4InList() string {
...
 	if err != nil {
 		return "not recognized const value"
 	}
+
+	valueStr := expr.format(dt)
+
+	// Add subquery reference if available (same logic as StringWithCtx)
+	if expr.SubqueryRefID > 0 {
+		return fmt.Sprintf("ScalarQueryCol#%d(%s)", expr.SubqueryRefID, valueStr)
+	}
```
time-and-fate: This behavior here is different from `(*Constant).StringWithCtx()` when it's `RedactLogMarker`. `‹›` is not added here.

PR #63077, opened 2025-08-19
Planner: Add min/max for out of range | tidb-test=pr/2615

### LINKED ISSUE #63724: planner: avoid selecting a high-risk plan, when there is a gap between modifycount and the actual value
## Enhancement

Because TiDB is a distributed system, its updates to **modifycount** and rowcount are **asynchronous**.
This inevitably creates a **time window** between: 
**data is written  <-> modifycount is updated to its true value.**

Because the estimation relies heavily on the modifycount value, TiDB may select an incorrect plan within this time window.

## For example

1. create table t1

```
CREATE TABLE `t1` (
  `a` int,
  `username` int,
  `createtime`bigint,
  `sitecode` int,
  `e` int,
  KEY good_idx (username, e, sitecode, createtime),
  KEY bad_idx (sitecode, createtime)
);
```

2. Insert some data which columns( username, sidecode) ndv is fixed value, createtime is automatically increases according to the insertion time.
```
INSERT INTO t1 (a, username, createtime, sitecode, e)
SELECT a, mod(a,1000) AS username, now()+a AS createtime, mod(a, 10) as sitecode, a as e
FROM (
    WITH RECURSIVE x AS (
        SELECT 1 AS a
        UNION ALL
        SELECT a + 1 AS a
        FROM x
        WHERE a < 1000000
    )
    SELECT a
    FROM x
) AS subquery;
```
Data characteristics of each column
```
mysql> select count(distinct username) from t1;
+--------------------------+
| count(distinct username) |
+--------------------------+
|                     1000 |
+--------------------------+
1 row in set (0.24 sec)

mysql> select count(distinct createtime) from t1;
+----------------------------+
| count(distinct createtime) |
+----------------------------+
|                    1000000 |
+----------------------------+
1 row in set (0.24 sec)

mysql> select count(distinct sitecode) from t1;
+--------------------------+
| count(distinct sitecode) |
+--------------------------+
|                       10 |
+--------------------------+
1 row in set (0.22 sec)
```

3. Analyze table 
```
analyze table t1;
```

4. Stop the analyze. Stop to update the stats_meta (modifycount, rowcount)
**keep the modifycount value to 0**
```
mysql> show stats_meta;
+---------+------------+----------------+---------------------+--------------+-----------+---------------------+
| Db_name | Table_name | Partition_name | Update_time         | Modify_count | Row_count | Last_analyze_time   |
+---------+------------+----------------+---------------------+--------------+-----------+---------------------+
| test    | t1         |                | 2025-09-24 16:42:23 |            0 |   1000000 | 2025-09-23 16:42:23 |
+---------+------------+----------------+---------------------+--------------+-----------+---------------------+
1 row in set (0.02 sec)
```

5. Insert data again and keep the username and sitecode ndv with the same value in step2 but createtime is automatically increases according to the insertion time. 
```
INSERT INTO t1 (a, username, createtime, sitecode, e)
SELECT a, mod(a,1000) AS username, now()+a AS createtime, mod(a, 10) as sitecode, a as e
FROM (
    WITH RECURSIVE x AS (
        SELECT 1000000 AS a
        UNION ALL
        SELECT a + 1 AS a
        FROM x
        WHERE a < 2000000
    )
    SELECT a
    FROM x
) AS subquery;
```
Data characteristics of each column
```
mysql> select count(distinct username) from t1;
+--------------------------+
| count(distinct username) |
+--------------------------+
|                     1000 |
+--------------------------+
1 row in set (0.42 sec)

mysql> select count(distinct createtime) from t1;
+----------------------------+
| count(distinct createtime) |
+----------------------------+
|                    2000001 |
+----------------------------+
1 row in set (0.53 sec)

mysql> select count(distinct sitecode) from t1;
+--------------------------+
| count(distinct sitecode) |
+--------------------------+
|                       10 |
+--------------------------+
1 row in set (0.43 sec)
```

6. Keep the stats_meta with the same value in  step4

```
mysql> show stats_meta;
+---------+------------+----------------+---------------------+--------------+-----------+---------------------+
| Db_name | Table_name | Partition_name | Update_time         | Modify_count | Row_count | Last_analyze_time   |
+---------+------------+----------------+---------------------+--------------+-----------+---------------------+
| test    | t1         |                | 2025-09-24 16:42:23 |            0 |   1000000 | 2025-09-23 16:42:23 |
+---------+------------+----------------+---------------------+--------------+-----------+---------------------+
1 row in set (0.02 sec)

```

7. Query the data which is within the time window: createtime > 20250925163638 . 20250925163638 is the max value after step2.

```
select a, username, createtime, sitecode, e
from t1
where 
username = 1
and createtime >= 20250925163638
and sidecode in (1)
```

8. The planner choose the bad_index(sitecode, createtime) instead of the good_index(username, e, sitecode, createtime)

```
| id                             | estRows | actRows | task      | access object                                     |                                                       |
| Projection_49                  | 0.00    | 1       | root      |                                                   |                                                       |
| └─Projection_55                | 0.00    | 1       | root      |                                                   |                                                       |
|   └─IndexLookUp_54             | 0.00    | 1       | root      |                                                   |                                                       |
|     ├─IndexRangeScan_51(Build) | 1.00    | 1000001 | cop[tikv] | table:t1, index:bad_index(site_code, create_time) |                                                       |
|     └─Selection_53(Probe)      | 0.00    | 1000    | cop[tikv] |                                                   | eq(test.t1.username, 1)                               |
|       └─TableRowIDScan_52      | 1.00    | 1000001 | cop[tikv] | table:t1                                          |                                                       |    
```

# Investigate

Reasons for selecting the wrong index:
1. The user inserted a large amount of data within the time window —— step 5
2. The modify count failed to update **timely** —— modifycount = 0
3. The query condition covered the time window —— createtime > the max value of last bucket

The three conditions above caused the optimizer to inaccurately estimate the number of rows for the createtime condition.
Because the createtime condition triggered an out-of-bounds estimate, and modifycount was 0,
the optimizer assumed there was virtually no out-of-bounds value and estimated a value of 1.
In reality, modifycount was not updated in time, and 1,000,001 rows had already been inserted into the out-of-bounds section, resulting in a true value of 1,000,001.
```
| id                             | estRows | actRows | task      | access object                                     |                                                       |
|     ├─IndexRangeScan_51(Build) | 1.00    | 1000001 | cop[tikv] | table:t1, index:bad_index(site_code, create_time) |                                                       

```

# Enhance
In real-world situations, this scenario **often occurs**:
1. The user's traffic suddenly increases, 
2. The modifycount function hasn't been updated yet
3. User wants to query the newly inserted data.

The optimizer will try to **avoid selecting indexes with higher risks (larger errors) in the above scenarios**.

### Better plan
The username predicate is a better predicate which will almost no, or very slow, change in selectivity when new traffic is written. —— username = 1

The createtime is a very bad predicate because it changes with the traffic. —— createtime > xxx

### Optimization ideas
By using predicate types and modifycount update methods, we can identify higher-risk indexes and avoid selecting such plans **under specific conditions**. 
This improves optimizer stability.

### LINKED ISSUE #63801: planner: cardinality estimation should be aware of incorrect ModifyCnt/RealtimeRowCount in out-of-range estimation
## Enhancement
In out-of-range estimation, the optimizer uses `ModifyCnt` as the upper bound of the result, but in some cases the `ModifyCnt` might be incorrect zero. For example, 1) if there are too many tables like 10,000+, the sync of `ModifyCnt` could be delayed, 2) if the user forcely restarts TiDB after inserting a batch of data, the `ModifyCnt` info might be lost.
The estimation would better be aware of these cases and use some heuristic strategies to mitigate them to make the estimates more accurate.

terry1purcell:

Issue Number: close #63724, close #63801

Problem Summary:

There are 3 main aspects to this PR:
1) When out-of-range faces a modifyCount == 0 - we have found that there is a latency to updating stats meta. And it could be as much as 1% of the table being modified before an update is triggered. Therefore, in outOfRangeEstimation - we will allow the "max estimate" to utilize 1% of the table size as it's worst case estimate. The standard estimate is not changed.
2) Track the min & max estimate through column and index predicate estimation - so that it can be stored with the index choice.
3) Update the skylinePruning compareCandidates logic to consider the risk (from min and max) in it's logic to help determine the winning index with the lowest risk.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/casetest/planstats/testdata/plan_stats_suite_out.json
+++ b/pkg/planner/core/casetest/planstats/testdata/plan_stats_suite_out.json
@@ -189,9 +189,9 @@
       {
         "Query": "select * from t use index(idx_c) where c > 1",
         "Result": [
-          "IndexLookUp 48.50 root  ",
-          "├─IndexRangeScan(Build) 48.50 cop[tikv] table:t, index:idx_c(c) range:(1,+inf], keep order:false",
-          "└─TableRowIDScan(Probe) 48.50 cop[tikv] table:t keep order:false"
+          "IndexLookUp 49.00 root  ",
+          "├─IndexRangeScan(Build) 49.00 cop[tikv] table:t, index:idx_c(c) range:(1,+inf], keep order:false",
+          "└─TableRowIDScan(Probe) 49.00 cop[tikv] table:t keep order:false"
         ]
       },
       {
@@ -212,8 +212,8 @@
       {
         "Query": "select * from t use index(idx_bc) where b=1 and c <2",
         "Result": [
-          "IndexReader 2.00 root  index:Selection",
-          "└─Selection 2.00 cop[tikv]  eq(cast(test.t.b, double BINARY), 1), lt(test.t.c, 2)",
+          "IndexReader 2.40 root  index:Selection",
+          "└─Selection 2.40 cop[tikv]  eq(cast(test.t.b, double BINARY), 1), lt(test.t.c, 2)",
           "  └─IndexFullScan 50.00 cop[tikv] table:t, index:idx_bc(b, c) keep order:false"
         ]
       },
--- a/pkg/planner/core/casetest/planstats/testdata/plan_stats_suite_xut.json
+++ b/pkg/planner/core/casetest/planstats/testdata/plan_stats_suite_xut.json
@@ -189,9 +189,9 @@
       {
         "Query": "select * from t use index(idx_c) where c > 1",
         "Result": [
-          "IndexLookUp 48.50 root  ",
-          "├─IndexRangeScan(Build) 48.50 cop[tikv] table:t, index:idx_c(c) range:(1,+inf], keep order:false",
-          "└─TableRowIDScan(Probe) 48.50 cop[tikv] table:t keep order:false"
+          "IndexLookUp 49.00 root  ",
+          "├─IndexRangeScan(Build) 49.00 cop[tikv] table:t, index:idx_c(c) range:(1,+inf], keep order:false",
+          "└─TableRowIDScan(Probe) 49.00 cop[tikv] table:t keep order:false"
         ]
       },
       {
@@ -212,8 +212,8 @@
       {
         "Query": "select * from t use index(idx_bc) where b=1 and c <2",
         "Result": [
-          "IndexReader 2.00 root  index:Selection",
-          "└─Selection 2.00 cop[tikv]  eq(cast(test.t.b, double BINARY), 1), lt(test.t.c, 2)",
+          "IndexReader 2.40 root  index:Selection",
+          "└─Selection 2.40 cop[tikv]  eq(cast(test.t.b, double BINARY), 1), lt(test.t.c, 2)",
           "  └─IndexFullScan 50.00 cop[tikv] table:t, index:idx_bc(b, c) keep order:false"
         ]
       },


pkg/statistics/statistics_test.go
```diff
@@ -300,22 +334,48 @@ func SubTestIntColumnRanges() func(*testing.T) {
...
+		countEst, err = GetRowCountByIntColumnRanges(ctx, &tbl.HistColl, 0, ran)
+		if err != nil {
+			return
+		}
+		count = countEst.Est
+		if err != nil {
+			return
+		}
+		count = countEst.Est
 		require.NoError(t, err)
```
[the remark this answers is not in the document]
terry1purcell: I don't see it

hawkingrei:
Actually, TiDB already supports [Constraints](https://docs.pingcap.com/tidb/stable/constraints/#check), so can we trim the estimates based on the prompts of Constraints?

pkg/statistics/histogram.go
```diff
@@ -1195,24 +1204,44 @@ func (hg *Histogram) OutOfRangeRowCount(
...
 	// half are distributed out of range according to the diagram in the function description.
-	avgRowCount = (addedRows * 0.5) * totalPercent
+	avgRowCount = (addedRows * addedOutOfRangePct) * totalPercent
+
+	// We may have missed the true lowest/highest values due to sampling OR there could be a delay in
+	// updates to modifyCount (meaning modifyCount is incorrectly set to 0). So ensure we always
+	// account for at least 1% of the total row count as a worst case for "addedRows".
+	// We inflate this here so ONLY to impact the MaxEst value.
+	if modifyCount == 0 || addedRows == 0 {
+		addedRows = max(addedRows, float64(realtimeRowCount)/outOfRangeBetweenRate)
```
qw4990, line 1229: If `modifyCount` is incorrect zero, could `realtimeRowCount` also be incorrect zero at the same time, since they are updated at the same transaction in the same function `UpdateStatsMeta`? If Yes, is it safer to use `hg.TotalRowCount()` here? Please help confirm @0xPoe

qw4990, line 1229: 1% of the current table size as the newly added rows seems a little bit aggressive to me, since many TiDB tables are very large compared with tables in a standalone database. And below is table sizes of one of our customers, then even 1% might be more than 1,000,000 rows. Should we give this `addedRows` an upper bound in this case, like `10,000` or `1,000`?
image

pkg/planner/core/find_best_task.go
```diff
@@ -1021,17 +1021,33 @@ func compareRiskRatio(lhs, rhs *candidatePath) (int, float64) {
 	// MaxCountAfterAccess tracks the worst case "CountAfterAccess", accounting for scenarios that could
 	// increase our row estimation, thus lhs/rhsRiskRatio represents the "risk" of the CountAfterAccess value.
 	// Lower value means less risk that the actual row count is higher than the estimated one.
-	if lhs.path.MaxCountAfterAccess > 0 && rhs.path.MaxCountAfterAccess > 0 {
+	if lhs.path.MaxCountAfterAccess > lhs.path.CountAfterAccess {
```
qw4990: Should we keep `lhs.path.CountAfterAccess > 0` in case of error of dividing zero?

pkg/planner/core/find_best_task.go
```diff
@@ -1117,13 +1136,26 @@ func compareCandidates(sctx base.PlanContext, statsTbl *statistics.Table, tableI
 		}
 	}
 
+	leftIsBetter := predicateResult >= 0 && scanResult >= 0 && matchResult >= 0 && globalResult >= 0
```
qw4990: What if all results here are zero? It seems like left would win unexpectedly in this case.

pkg/planner/cardinality/cross_estimation.go
0xPoe, line 193: Since we handle the error in the else branch, it would be better to move this error handling into the if branch.

0xPoe, line 227: ditto

pkg/planner/cardinality/row_count_index.go
```diff
@@ -196,7 +195,11 @@ func getIndexRowCountForStatsV1(sctx planctx.PlanContext, coll *statistics.HistC
...
 			} else {
-				count, err = GetRowCountByColumnRanges(sctx, coll, colUniqueID, []*ranger.Range{&rang})
+				countEst, err := GetRowCountByColumnRanges(sctx, coll, colUniqueID, []*ranger.Range{&rang})
+				if err != nil {
+					return 0, errors.Trace(err)
+				}
+				count = countEst.Est
 			}
 			if err != nil {
 				return 0, errors.Trace(err)
```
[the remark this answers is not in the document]
0xPoe: This is not dead code; it still applies to the if block, similar to the cases I mentioned above.

pkg/statistics/histogram.go
0xPoe: I think it is possible, but it’s almost impossible to lose the full real-time row count since it’s based on the analyze row count plus the real-time count. So I believe it’s safe to use the real-time row count here.

0xPoe: 1000 sounds too small. Maybe we can set it to a value that caused issue actually as the upper bound?
The issues I’ve encountered recently have all been in the range of 100,000 to 200,000.

0xPoe (commented):
Thanks!

pkg/planner/core/find_best_task.go
terry1purcell: I will rename these variables - because this is really intended to determine that "left did not lose".

pkg/planner/core/find_best_task.go
terry1purcell: Done

pkg/statistics/histogram.go
terry1purcell: Done

pkg/planner/core/find_best_task.go
terry1purcell: Done

pkg/planner/cardinality/cross_estimation.go
terry1purcell: Corrected

pkg/planner/cardinality/row_count_index.go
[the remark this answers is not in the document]
terry1purcell: I have resolved this confusion

terry1purcell:
> Actually, TiDB already supports [Constraints](https://docs.pingcap.com/tidb/stable/constraints/#check), so can we trim the estimates based on the prompts of Constraints?

Thank you for suggesting this - but I don't see how we could use constraints here. Because we can't stop the customer from inserting rows beyond the current high value of the histogram buckets. Since this is merely their application behavior.

If I misunderstood - please let me know.

pkg/planner/core/find_best_task.go
```diff
@@ -1021,17 +1021,33 @@ func compareRiskRatio(lhs, rhs *candidatePath) (int, float64) {
 	// MaxCountAfterAccess tracks the worst case "CountAfterAccess", accounting for scenarios that could
 	// increase our row estimation, thus lhs/rhsRiskRatio represents the "risk" of the CountAfterAccess value.
 	// Lower value means less risk that the actual row count is higher than the estimated one.
-	if lhs.path.MaxCountAfterAccess > 0 && rhs.path.MaxCountAfterAccess > 0 {
+	if lhs.path.MaxCountAfterAccess > lhs.path.CountAfterAccess {
 		lhsRiskRatio = lhs.path.MaxCountAfterAccess / lhs.path.CountAfterAccess
+	}
+	if rhs.path.MaxCountAfterAccess > rhs.path.CountAfterAccess {
 		rhsRiskRatio = rhs.path.MaxCountAfterAccess / rhs.path.CountAfterAccess
 	}
```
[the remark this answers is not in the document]
terry1purcell: Yep - it's a nitpick (as you said).

hawkingrei:
> > Actually, TiDB already supports [Constraints](https://docs.pingcap.com/tidb/stable/constraints/#check), so can we trim the estimates based on the prompts of Constraints?
> 
> Thank you for suggesting this - but I don't see how we could use constraints here. Because we can't stop the customer from inserting rows beyond the current high value of the histogram buckets. Since this is merely their application behavior.
> 
> If I misunderstood - please let me know.

OK, the main issue is that the feature isn't enabled by default yet. We can continue to address it in the future.

The corresponding documentation is here.

https://dev.mysql.com/doc/refman/8.4/en/create-table-check-constraints.html
https://docs.pingcap.com/tidb/stable/constraints/

pkg/planner/cardinality/selectivity.go
```diff
@@ -136,20 +136,23 @@ func Selectivity(
 				return 0, nil, errors.Trace(err)
 			}
 			nodes = append(nodes, &StatsNode{Tp: ColType, ID: id, mask: maskCovered, Ranges: ranges, numCols: 1})
+			var cnt float64
+			var cntEst statistics.RowEstimate
```
hawkingrei, line 140:
```suggestion
			var cntEst *statistics.RowEstimate
```

I suggest creating a pointer here to avoid allocating memory once here and then again inside GetRowCountByColumnRanges.

pkg/planner/core/find_best_task.go
```diff
@@ -1118,13 +1137,28 @@ func compareCandidates(sctx base.PlanContext, statsTbl *statistics.Table, tableI
 		}
 	}
 
+	leftDidNotLose := predicateResult >= 0 && scanResult >= 0 && matchResult >= 0 && globalResult >= 0
+	rightDidNotLose := predicateResult <= 0 && scanResult <= 0 && matchResult <= 0 && globalResult <= 0
 	if !comparable1 && !comparable2 {
```
qw4990, line 949: Should be `!comparable1 || !comparable2` here? If anyone of them is not comparable, we should apply the rule here?

hawkingrei:
/hold 

Waiting for others' approval.

pkg/planner/core/find_best_task.go
0xPoe: Copied from the Lark Group(Terry): Thanks for highlighting this. The reason I have tried to relax this is that comparable1 requires that both indexes are accessing the same columns. So if you have WHERE C1 = ? AND C2 > ?. Then if you have indexes on idx1 (C1, C2) and idx2 (C2) - these ARE comparable. But if you had indexes on idx1 (C1) and idx2(C2) - these are NOT comparable. So if you have a predicate on C2 that carries high risk (due to out of range - as seen in customer1 & customer2) - then this logic can't be used to preference one index over another. That means - if you underestimate the selectivity on C2 - depending on whether the indexes you are competing are comparable or not?

pkg/planner/cardinality/selectivity.go
terry1purcell: Discussed offline

pkg/planner/core/find_best_task.go
terry1purcell: Updated the above to remove any customer names.

PR #63085, opened 2025-08-20
planner: Include DNF in compareCandidates

terry1purcell:

Issue Number: ref #59333 

Problem Summary:

CompareCandidates function will attempt to apply rules to prioritize one plan over another at the datasource level. This PR enhances the current "in or equal" predicate check to also account for predicates for DNF conditions. Additional enhancements include:
- removing global index check for pseudo stats - since this needs more analysis.
- removing checks for CountAfterAccess vs MaxCountAfterAccess - since these checks are irrelevant if one object has pseudo stats (because any estimate is totally unreliable).

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/find_best_task.go
```diff
@@ -1133,20 +1133,19 @@ func compareCandidates(sctx base.PlanContext, statsTbl *statistics.Table, tableI
...
-		if !lhsPseudo && globalResult >= 0 && sum >= 0 &&
-			lhs.path.EqOrInCondCount > 0 && lhs.path.EqOrInCondCount >= rhs.path.EqOrInCondCount &&
+		if !lhsPseudo && globalResult >= 0 && sum >= 0 && eqOrInResult > 0 &&
 			// if rhs's maxCount hasn't been adjusted or its adjusted max risk is greater than than lhs's regular count, then lhs can win
 			(rhs.path.MaxCountAfterAccess <= rhs.path.CountAfterAccess || lhs.path.CountAfterAccess < rhs.path.MaxCountAfterAccess) {
 			return 1, lhsPseudo // left wins and has statistics (lhsPseudo==false)
 		}
-		if !rhsPseudo && globalResult <= 0 && sum <= 0 &&
-			rhs.path.EqOrInCondCount > 0 && rhs.path.EqOrInCondCount >= lhs.path.EqOrInCondCount &&
+		if !rhsPseudo && globalResult <= 0 && sum <= 0 && eqOrInResult < 0 &&
```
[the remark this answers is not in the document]
terry1purcell: This is existing code.

pkg/planner/core/find_best_task.go
```diff
@@ -1129,38 +1129,50 @@ func compareCandidates(sctx base.PlanContext, statsTbl *statistics.Table, tableI
...
 	scanResult, comparable2 := compareIndexBack(lhs, rhs)
+	// riskResult: comparison result of risk factor (1=LHS better, -1=RHS better, 0=equal)
 	riskResult, _ := compareRiskRatio(lhs, rhs)
-	sum := accessResult + scanResult + matchResult + globalResult
+	// eqOrInResult: comparison result of equal/IN predicate coverage (1=LHS better, -1=RHS better, 0=equal)
+	eqOrInResult, lhsEqOrInCount, rhsEqOrInCount := compareEqOrIn(lhs, rhs)
+
+	// sum is the aggregate score of various comparison metrics between lhs and rhs. riskResult is excluded
+	// because more work is required.
+	// TO-DO - extend riskResult such that risk factors can be integrated into the aggregate score. Risk should
```
hawkingrei:
```suggestion
	// TODO: - extend riskResult such that risk factors can be integrated into the aggregate score. Risk should
```

We usually write it as `TODO:`, which allows some TODO management tools to automatically match, highlight, and remind developers.

image

pkg/planner/core/find_best_task.go
```diff
@@ -1228,7 +1241,29 @@ func isCandidatesPseudo(lhs, rhs *candidatePath, lhsFullScan, rhsFullScan bool,
 	return lhsPseudo, rhsPseudo
 }
 
+func compareEqOrIn(lhs, rhs *candidatePath) (predCompare, lhsEqOrInCount, rhsEqOrInCount int) {
```
qw4990, line 1250: A nice new concept of score for skyline pruning, but could you add a concrete example in the comment here? For example, `where a=1 and b=1 and c=1 and d=1` for `idx(a, b, e)` and `idx(d, c, b)`.

pkg/planner/core/find_best_task.go
```diff
@@ -1129,38 +1129,50 @@ func compareCandidates(sctx base.PlanContext, statsTbl *statistics.Table, tableI
...
+	// TO-DO: Consider a separate set of rules for global indexes.
 	if (lhsPseudo || rhsPseudo) && !tablePseudo && !lhsFullScan && !rhsFullScan { // At least one index doesn't have statistics
 		// If one index has statistics and the other does not, choose the index with statistics if it
 		// has the same or higher number of equal/IN predicates.
-		if !lhsPseudo && globalResult >= 0 && sum >= 0 &&
-			lhs.path.EqOrInCondCount > 0 && lhs.path.EqOrInCondCount >= rhs.path.EqOrInCondCount &&
-			// if rhs's maxCount hasn't been adjusted or its adjusted max risk is greater than than lhs's regular count, then lhs can win
-			(rhs.path.MaxCountAfterAccess <= rhs.path.CountAfterAccess || lhs.path.CountAfterAccess < rhs.path.MaxCountAfterAccess) {
+		if !lhsPseudo && sum >= 0 &&
+			(eqOrInResult > 0 || (lhs.path.EqOrInCondCount > 0 && lhsEqOrInCount >= rhsEqOrInCount)) {
```
qw4990, line 1159: It seems like `(rhs.path.MaxCountAfterAccess <= rhs.path.CountAfterAccess || lhs.path.CountAfterAccess < rhs.path.MaxCountAfterAccess)` is removed, is this expected?

terry1purcell: Yes. That is expected. It was added recently by another team member. At the time I didn't realize the issue - but my concern with that prior code is that in this check - we are comparing an index that has statistics vs one that does not have statistics. So the CountAfterAccess and MaxCountAfterAccess are going to be unreliable for the object without statistics - thus it makes no sense to have this as part of the IF check there.

pkg/planner/core/find_best_task.go
terry1purcell: Done

pkg/planner/core/find_best_task.go
```diff
@@ -1129,38 +1129,50 @@ func compareCandidates(sctx base.PlanContext, statsTbl *statistics.Table, tableI
...
-	sum := accessResult + scanResult + matchResult + globalResult
+	// eqOrInResult: comparison result of equal/IN predicate coverage (1=LHS better, -1=RHS better, 0=equal)
+	eqOrInResult, lhsEqOrInCount, rhsEqOrInCount := compareEqOrIn(lhs, rhs)
+
+	// sum is the aggregate score of various comparison metrics between lhs and rhs. riskResult is excluded
+	// because more work is required.
+	// TODO: - extend riskResult such that risk factors can be integrated into the aggregate score. Risk should
+	// consider what "type" of risk is being evaluated (eg. out of range, implied independence, data skew, whether a
+	// bound was applied, etc.)
+	sum := accessResult + scanResult + matchResult + globalResult + eqOrInResult
```
elsa0520, line 1151: A small suggestion: Regarding the risk comparisons from lines 1132 to 1151, it would be better to encapsulate them into a single function. It seems we will make further changes to this section in the future.

```
function compareRisk() {
   	matchResult, globalResult := compareBool(lhs.isMatchProp, rhs.isMatchProp), compareGlobalIndex(lhs, rhs)
	accessResult, comparable1 := util.CompareCol2Len(lhs.accessCondsColMap, rhs.accessCondsColMap)
	scanResult, comparable2 := compareIndexBack(lhs, rhs)
	riskResult, _ := compareRiskRatio(lhs, rhs)
    xxxx
}
```

elsa0520, line 1151: Another thinking, whether sum is a good way for all risk ? Could we have some other combine risk formula ? Or just sum them ? 
Are there priorities or weights among these risks?

terry1purcell: I think compareCandidates is already the correct place. I believe instead that more of logic above this (within compareCandidates) could be factored out.

terry1purcell: The current list is limited - so it is difficult to determine how to prioritize or weigh them until we have made all enhancements necessary.

elsa0520 (approved):
LGTM

Issue #61185, opened 2025-05-19
Support fulltext search execution


winoros:
## Enhancement

- [ ] Support the simplest form
- [ ] Support simple row count calculation based on a constant factor
- [ ] Support returning the distance/score and use it in the projection and order by clause
- [ ] Support index merge

PR #63128, opened 2025-08-21
planner: don't read table kv when only needs pk from tici | tidb-test=ef726505f262e22c6ec9c8951273bb5ee20abefb

winoros:

Issue Number: ref https://github.com/pingcap/tidb/issues/61185

Problem Summary:

TiCI lib contains the original value of pk.
So when we only need to read the column from pk. We can directly use the values from TiCI instead of reading them from TiKV's table kv.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/casetest/tici/testdata/tici_index_suite_out.json
```diff
@@ -6,26 +6,24 @@
         "SQL": "explain format='brief' select * from t1 where fts_match_word('hello', title)",
         "Plan": [
           "IndexLookUp 1000.00 root  ",
-          "├─IndexRangeScan(Build) 1000.00 batchCop[tiflash] table:t1, index:title(title) range:[NULL,+inf], search func:fts_match_word(\"hello\", test.t1.title), keep order:false, stats:pseudo",
+          "├─IndexRangeScan(Build) 1000.00 cop[tici] table:t1, index:title(title) range:[NULL,+inf], search func:fts_match_word(\"hello\", test.t1.title), keep order:false, stats:pseudo",
```
kolafish, line 9: is cop[tici] means that there is only one region(shard) in this request ?  if there are two shards on the same tiflash, would there be two cop requests to tiflash?

pkg/planner/core/find_best_task.go
```diff
@@ -2420,6 +2418,20 @@ func indexCoveringColumn(ds *logicalop.DataSource, column *expression.Column, in
...
+func pkCoveringColumn(ds *logicalop.DataSource, column *expression.Column, ignoreLen bool) (bool, bool) {
+	if ds.TableInfo.PKIsHandle && mysql.HasPriKeyFlag(column.RetType.GetFlag()) {
+		return true, false
+	}
+	if column.ID == model.ExtraHandleID || column.ID == model.ExtraPhysTblID {
+		return true, false
+	}
+	evalCtx := ds.SCtx().GetExprCtx().GetEvalCtx()
+	coveredByClusteredIndex := isIndexColsCoveringCol(evalCtx, column, ds.CommonHandleCols, ds.CommonHandleLens, ignoreLen)
+	return coveredByClusteredIndex, true
```
time-and-fate: I think these and the second return value of this function are useless. Only `indexCoveringColumn()` uses it.

pkg/planner/core/task_base.go
```diff
@@ -498,7 +498,12 @@ func (t *CopTask) convertToRootTaskImpl(ctx base.PlanContext) (rt *RootTask) {
 	if t.indexPlan != nil && t.tablePlan != nil {
 		newTask = buildIndexLookUpTask(ctx, t)
 	} else if t.indexPlan != nil {
-		p := PhysicalIndexReader{indexPlan: t.indexPlan}.Init(ctx, t.indexPlan.QueryBlockOffset())
+		ip := t.indexPlan
+		for len(ip.Children()) > 0 {
+			ip = ip.Children()[0]
+		}
+		is := ip.(*PhysicalIndexScan)
+		p := PhysicalIndexReader{indexPlan: t.indexPlan, StoreType: is.StoreType}.Init(ctx, t.indexPlan.QueryBlockOffset())
```
time-and-fate: You can use `p.IndexPlans[0].(*PhysicalIndexScan)` directly after `Init()`.

pkg/domain/domain.go
```diff
@@ -2012,7 +2012,7 @@ func (do *Domain) loadStatsWorker() {
 		case <-loadTicker.C:
 			err = statsHandle.Update(ctx, do.InfoSchema())
 			if err != nil {
-				logutil.BgLogger().Warn("update stats info failed", zap.Error(err))
```
hawkingrei: Has it problems?

pkg/planner/core/casetest/tici/testdata/tici_index_suite_out.json
winoros: No, this adjustment is necessary because we need to explicitly define what can be pushed down to the tici. Thus, a new label is introduced.
The explain result can be adjusted again according to our needs.

pkg/planner/core/find_best_task.go
```diff
@@ -2399,27 +2399,49 @@ func isIndexColsCoveringCol(sctx expression.EvalContext, col *expression.Column,
...
 	}
 	return true
 }
 
+type pkCoverState uint8
+
+const (
+	stateNotCoveredByPK pkCoverState = iota
+	stateCoveredByIntPK
+	stateCoveredByClusterPK
```
time-and-fate:
```suggestion
	stateNotCoveredByHandle handleCoverState = iota
	stateCoveredByIntHandle
	stateCoveredByCommonHandle
```

pkg/planner/core/find_best_task.go
```diff
@@ -2399,27 +2399,49 @@ func isIndexColsCoveringCol(sctx expression.EvalContext, col *expression.Column,
...
+)
+
+// pkCoveringColumn checks if the column is covered by the primary key or extra handle columns.
+func pkCoveringColumn(ds *logicalop.DataSource, column *expression.Column, ignoreLen bool) pkCoverState {
```
time-and-fate:
```suggestion
func handleCoveringColumn(ds *logicalop.DataSource, column *expression.Column, ignoreLen bool) pkCoverState {
```

hawkingrei (approved):

pkg/planner/core/task.go
```diff
@@ -755,7 +758,7 @@ func attach2Task4PhysicalLimit(pp base.PhysicalPlan, tasks ...base.Task) base.Ta
 		if len(cop.idxMergePartPlans) == 0 {
 			// For double read which requires order being kept, the limit cannot be pushed down to the table side,
 			// because handles would be reordered before being sent to table scan.
-			if (!cop.keepOrder || !cop.indexPlanFinished || cop.indexPlan == nil) && len(cop.rootTaskConds) == 0 {
+			if (!cop.keepOrder || !cop.indexPlanFinished || cop.indexPlan == nil) && len(cop.rootTaskConds) == 0 && cop.getStoreType() != kv.TiCI {
```
time-and-fate, line 761: I think it's not needed. Because we already called `finishIndexPlan()` in `convertToIndexScan()`, so it's impossible to push down `Limit` to the index side.

time-and-fate (approved):

pkg/planner/core/task.go
winoros: We'll soon try to support the limit pushdown case. Let's keep it for a while.

Issue #63131, opened 2025-08-21
Parse and match slow log trigger rules for multi-dimensional triggering


zimulala:
### Rule Parsing Logic
Parse Slow Log Trigger Rules:
- Conditional level logic (AND relationship) 
  - In a single rule, multiple conditions are separated by , , indicating a logical AND relationship. All conditions must be met for the rule to be considered a match. Example: 
```
"rule1_threshold1 = param1, rule1_threshold2 = param2, rule1_threshold3 = param3"
 
                               -- >
                               
rule1_threshold1 = param1 and rule1_threshold2 = param2 and rule1_threshold3 = param3
```  
  - After parsing, store a single Key-Value into SlowLogCondition structure, and store a single rule into SlowLogRule. 
- Rule-level logic (OR relationship)
  - Multiple rules are separated by;, indicating a logical OR relationship between them. If any rule is satisfied, the rules are considered to match. Example:
```
"rule1_threshold1 = param1; 
 rule2_threshold1 = param2, rule2_threshold2 = param3;
 rule3_threshold1 = param2, rule3_threshold3 = param3;"
 
                               -- >
                               
 rule1_threshold1 = param1 or 
 (rule2_threshold1 = param2 and rule2_threshold2 = param3) or
 (rule3_threshold1 = param2 and rule3_threshold3 = param3)
```
  - After parsing, multiple rules are stored in  SlowLogRules.Rules  structure. 

### tidb_slow_log_trigger_rules Variable Usage Constraints
- Rule Capacity:
  - Session and Global scopes each support up to 10 rules. 
- Overlay Mode: 
  - Each time SET is executed, the new value will overwrite the old rule instead of appending to it.
    - If set @@tidb_slow_log_rules =  "" ;  is equivalent to  clearing the rules . 
  - Specific examples of behavior are as follows: 
```
step1：set  @@tidb_slow_log_trigger_rules = "rule1_threshold1 = param1; rule2_threshold2 = param2, rule2_threshold3 = param3";
step2：set  @@tidb_slow_log_trigger_rules = "rule1_threshold1 = param2";

result：
    tidb_slow_log_trigger_rules = "rule1_threshold1 = param2";
```

PR #63132, opened 2025-08-21
*: parse and match slow log trigger rules for multi-dimensional triggering

zimulala:

Issue Number: close https://github.com/pingcap/tidb/issues/63131

Problem Summary:
Ref: https://github.com/pingcap/tidb/issues/63131

```
// SlowLogFieldAccessor defines how to get or set a specific field in SlowQueryLogItems.
// - Parse converts a user-provided threshold string into the proper type for comparison.
// - Setter is optional and pre-fills the field before matching if it needs explicit preparation.
// - Match evaluates whether the field in SlowQueryLogItems meets a specific threshold.
//   - threshold is the value to compare against when determining a match.
type SlowLogFieldAccessor struct {
	Parse  func(string) (any, error),
	Setter func(ctx context.Context, seVars *SessionVars, items *SlowQueryLogItems)
	Match  func(seVars *SessionVars, items *SlowQueryLogItems, threshold any) bool
}
 
var SlowLogRuleFieldAccessors = map[string]SlowLogFieldAccessor{
    variable.SlowLogConnIDStr: {
       Parse: parseUint64,
       Match: func(seCtx sessionctx.Context, items *variable.SlowQueryLogItems, threshold any) bool {
           ...
        },
    },
    variable.SlowLogExecRetryCount: {
        Parse  func(string) (any, error),
        Setter: func(ctx context.Context, seCtx sessionctx.Context, items *variable.SlowQueryLogItems) {
            ...
        },
        Match: func(seCtx sessionctx.Context, items *variable.SlowQueryLogItems, threshold any) bool {
            ...
    },
...
}
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/adapter_slow_log.go
```diff
@@ -0,0 +1,452 @@
...
+	"github.com/pingcap/tidb/pkg/util/execdetails"
+	"github.com/tikv/client-go/v2/util"
+)
+
+// SlowLogFieldAccessor defines how to get or set a specific field in SlowQueryLogItems.
+// - Setter is optional and pre-fills the field before matching if it needs explicit preparation.
+// - Match evaluates whether the field in SlowQueryLogItems meets a specific threshold.
+//   - threshold is the value to compare against when determining a match.
+type SlowLogFieldAccessor struct {
+	Setter func(ctx context.Context, seCtx sessionctx.Context, items *variable.SlowQueryLogItems)
```
zimulala: `SlowLogFieldAccessor` is defined here to avoid circular imports, because `seCtx` is defined in `pkg/sessionctx/variable/slow_log.go`.

pkg/executor/adapter_slow_log.go
```diff
@@ -0,0 +1,452 @@
...
+	Match  func(seCtx sessionctx.Context, items *variable.SlowQueryLogItems, threshold any) bool
+}
+
```
XuHuaiyu:
```suggestion
type ordered interface {
	~int | ~int64 | ~uint64 | ~float64  // | time.Duration 
}

func matchEqual[T comparable](threshold any, v T) bool {
	tv, ok := threshold.(T)
	if !ok {
		return false
	}
	return v == tv
}

func matchGE[T ordered](threshold any, v T) bool {
	tv, ok := threshold.(T)
	if !ok {
		return false
	}
	return v >= tv
}
```

pkg/executor/adapter_slow_log.go
```diff
@@ -0,0 +1,452 @@
...
+	return true
+}
+
+func matchDurationVal(threshold any, v time.Duration) bool {
+	tv, ok := threshold.(float64)
+	if !ok || v.Seconds() < tv {
+		return false
+	}
+	return true
+}
```
XuHuaiyu:
```suggestion

```

pkg/executor/adapter_slow_log.go
```diff
@@ -0,0 +1,452 @@
...
+	variable.SlowLogConnIDStr: {
+		Match: func(seCtx sessionctx.Context, _ *variable.SlowQueryLogItems, threshold any) bool {
+			return matchUint64Val(threshold, seCtx.GetSessionVars().ConnectionID)
+		},
+	},
+	variable.SlowLogSessAliasStr: {
+		Match: func(seCtx sessionctx.Context, _ *variable.SlowQueryLogItems, threshold any) bool {
+			return matchStringVal(threshold, seCtx.GetSessionVars().SessionAlias)
+		},
+	},
```
XuHuaiyu:
```suggestion
	variable.SlowLogConnIDStr: {
		Match: func(seCtx sessionctx.Context, _ *variable.SlowQueryLogItems, threshold any) bool {
			return matchGE(threshold, seCtx.GetSessionVars().ConnectionID)
		},
	},
	variable.SlowLogSessAliasStr: {
		Match: func(seCtx sessionctx.Context, _ *variable.SlowQueryLogItems, threshold any) bool {
			return matchEqual(threshold, seCtx.GetSessionVars().SessionAlias)
		},
	},
```

pkg/sessionctx/stmtctx/stmtctx.go
```diff
@@ -333,7 +333,9 @@ type StatementContext struct {
 	}
 	// BindSQL used to construct the key for plan cache. It records the binding used by the stmt.
 	// If the binding is not used by the stmt, the value is empty
-	BindSQL string
+	BindSQL        string
+	ExecRetryCount uint64
```
yibin87, line 337: Comments needed for ExecRetryCount and ExecSuccess

pkg/executor/adapter_slow_log.go
```diff
@@ -0,0 +1,181 @@
...
+	"time"
+
+	"github.com/pingcap/tidb/pkg/keyspace"
+	"github.com/pingcap/tidb/pkg/parser/ast"
+	"github.com/pingcap/tidb/pkg/sessionctx/variable"
+	"github.com/tikv/client-go/v2/util"
+)
+
+func init() {
+	variable.SlowLogRuleFieldAccessors[variable.SlowLogPlanDigest] = variable.SlowLogFieldAccessor{
```
yibin87: Why put it here alone? No parser here

pkg/sessionctx/variable/slow_log.go
```diff
@@ -514,3 +521,406 @@ func (s *SessionVars) SlowLogFormat(logItems *SlowQueryLogItems) string {
...
+
+// SlowLogRules represents all slow log rules defined for the current scope (e.g., session/global).
+// The rules are evaluated using logical OR between them: if any rule matches, it triggers the slow log.
+type SlowLogRules struct {
+	AllConditionFields map[string]struct{} // Set of all unique field names used in all conditions.
+	Rules              []SlowLogRule       // List of rules combined with logical OR.
+}
+
+// SlowLogFieldAccessor defines how to get or set a specific field in SlowQueryLogItems.
+// - Parse converts a string value (from user input) into the proper type for comparison.
```
yibin87: Maybe we can make it more clear that Parse is used for parsing threshold value.

pkg/sessionctx/variable/slow_log.go
```diff
@@ -514,3 +521,406 @@ func (s *SessionVars) SlowLogFormat(logItems *SlowQueryLogItems) string {
...
+			})
+			slowLogRules.AllConditionFields[fieldName] = struct{}{}
+		}
+		slowLogRules.Rules = append(slowLogRules.Rules, slowLogRule)
+	}
+	return slowLogRules, nil
+}
+
+// ParseSlowLogFieldValue is exporting for testing.
+func ParseSlowLogFieldValue(fieldName string, value string) (any, error) {
```
yibin87, line 928: Maybe we can change to something like ParseSlowLogFieldThresholdValue

pkg/sessionctx/variable/slow_log.go
```diff
@@ -514,3 +521,406 @@ func (s *SessionVars) SlowLogFormat(logItems *SlowQueryLogItems) string {
...
+	slowLogRules := &SlowLogRules{
+		AllConditionFields: make(map[string]struct{}),
+		Rules:              make([]SlowLogRule, 0, len(rules))}
+	for _, rule := range rules {
+		rule = strings.TrimSpace(rule)
+		if rule == "" {
+			continue
+		}
+
+		fields := strings.Split(rule, ",")
```
yibin87: Since, the parsing work doesn't happen frequently, thus performance is not a big deal. Maybe we can try regexp to simplify this whole process.

pkg/sessionctx/variable/slow_log.go
```diff
@@ -514,3 +521,406 @@ func (s *SessionVars) SlowLogFormat(logItems *SlowQueryLogItems) string {
...
+	if rawRules == "" {
+		return nil, nil
+	}
+
+	rules := strings.Split(rawRules, ";")
+	if len(rules) == 0 {
+		return nil, nil
+	}
+	if len(rules) > 10 {
+		return nil, errors.Errorf("invalid slow log rules count:%d", len(rules))
```
yibin87: Show the max allowed rule count in error message also?

yibin87:
Do we need to consider collations when do string comparison?

pkg/sessionctx/variable/slow_log.go
```diff
@@ -514,3 +521,406 @@ func (s *SessionVars) SlowLogFormat(logItems *SlowQueryLogItems) string {
...
+
+		fields := strings.Split(rule, ",")
+		fieldMap := make(map[string]any, len(fields))
+		for _, field := range fields {
+			kv := strings.Split(field, ":")
+			if len(kv) != 2 {
+				return nil, errors.Errorf("invalid slow log field format:%s", field)
+			}
+
+			fieldName := strings.TrimSpace(kv[0])
```
yibin87: Do we consider upper/lower cases for fieldName and string type threshold?

pkg/sessionctx/stmtctx/stmtctx.go
zimulala: These field names should be easy to understand, so they are not added here.

pkg/executor/adapter_slow_log.go
zimulala: Yes. Because the "GetPlanDigest" function is implemented in this package, putting it in the variable package would result in a circular dependency. Also, "GetPlanDigest" depends on other functions, which can be hard to move from the current package to another.
Good catch. I forgot the parser here. Added it and some tests.

pkg/sessionctx/variable/slow_log.go
zimulala: I don't think the name is particularly ideal, so I won't change it for now. If there is more suitable, you can recommend it again, and I will update the name

pkg/sessionctx/stmtctx/stmtctx.go
yibin87: I'm not sure, from my understanding fields which begin with Upper case, needs a comment :)

zimulala: There are a lot of capitalized variables that don't have comments. Mainly, I think these two fields are really easy to understand

zimulala:
> Do we need to consider collations when do string comparison?

It should not be necessary.
1. After the user sets the rules, we will internally convert them to utf8-mb4 charset (when storing in mysql.global_variables, the collection information will not be used).
2. The data generated internally is already in utf8-mb4 charset. 

When comparing these two sets of data, it seems that the collection factor does not need to be considered.

zimulala:
PTAL @yibin87 @XuHuaiyu

pkg/sessionctx/variable/slow_log.go
yibin87: OK.

pkg/sessionctx/variable/slow_log.go
```diff
@@ -514,3 +522,414 @@ func (s *SessionVars) SlowLogFormat(logItems *SlowQueryLogItems) string {
...
+		if rule == "" {
+			continue
+		}
+
+		matches := slowLogFieldRe.FindAllStringSubmatch(rule, -1)
+		if len(matches) == 0 {
+			return nil, fmt.Errorf("invalid slow log rule format:%s", rule)
+		}
+		fieldMap := make(map[string]any, len(matches))
+		for _, matche := range matches {
```
yibin87:
```suggestion
		for _, match := range matches {
```

pkg/sessionctx/variable/slow_log.go
```diff
@@ -514,3 +522,414 @@ func (s *SessionVars) SlowLogFormat(logItems *SlowQueryLogItems) string {
...
+		}),
+	strings.ToLower(execdetails.PrewriteRegionStr): makeExecDetailAccessor(
+		parseUint64,
+		func(d *execdetails.ExecDetails, threshold any) bool {
+			return matchGE(threshold, int64(atomic.LoadInt32(&d.CommitDetail.PrewriteRegionNum)))
+		}),
+}
+
+// slowLogFieldRe is uses to compile field:value
+var slowLogFieldRe = regexp.MustCompile(`\s*([^:,]+)\s*:\s*("[^"]*"|'[^']*'|[^,]+)\s*`)
```
yibin87:
```suggestion
var slowLogFieldRe = regexp.MustCompile(`\s*(\w+)\s*:\s*("[^"]*"|'[^']*'|[^,]+)\s*`)
```

yibin87: BTW, maybe we can match the separator "," to get the end index of value, assuming no ","/":"/";" included in field or value. If any, it would lead to a incorrect Field name, just report error.

zimulala: ~It seems that the format of "resource-group-1" might occur.~

pkg/sessionctx/variable/slow_log.go
```diff
@@ -865,7 +865,7 @@ var SlowLogRuleFieldAccessors = map[string]SlowLogFieldAccessor{
 }
 
 // slowLogFieldRe is uses to compile field:value
-var slowLogFieldRe = regexp.MustCompile(`\s*([^:,]+)\s*:\s*("[^"]*"|'[^']*'|[^,]+)\s*`)
+var slowLogFieldRe = regexp.MustCompile(`\s*(\w+)\s*:\s*([^,]+)\s*`)
```
yibin87, line 868: Use "regexp" to replace "github.com/grafana/regexp"?

pkg/sessionctx/stmtctx/stmtctx.go
```diff
@@ -333,7 +333,9 @@ type StatementContext struct {
...
+	ExecSuccess    bool
```
yibin87, line 338: Is this var set? I don't find the place where it is set a value.

zimulala: The next PR will be invoked and there will be corresponding tests (since these two PRs are related, this PR will not be updated for now).

yibin87: Got it.

pkg/sessionctx/variable/slow_log_test.go
```diff
@@ -0,0 +1,360 @@
...
+
+	// uint64 fields
+	v, err = variable.ParseSlowLogFieldValue(variable.SlowLogConnIDStr, "456")
+	require.NoError(t, err)
+	require.Equal(t, uint64(456), v)
+
+	_, err = variable.ParseSlowLogFieldValue(variable.SlowLogConnIDStr, "-1")
+	require.Error(t, err)
+
+	// float64 fields
```
yibin87, line 188: Do we support scientific notation for float64, like 1.3e4

zimulala: Yes, but the user manual does not specify support for this format.

zimulala:
PTAL @XuHuaiyu

pkg/executor/adapter_slow_log.go
```diff
@@ -0,0 +1,183 @@
...
+	"time"
+
+	"github.com/pingcap/tidb/pkg/keyspace"
+	"github.com/pingcap/tidb/pkg/parser/ast"
+	"github.com/pingcap/tidb/pkg/sessionctx/variable"
+	"github.com/tikv/client-go/v2/util"
+)
+
+func init() {
+	variable.SlowLogRuleFieldAccessors[strings.ToLower(variable.SlowLogPlanDigest)] = variable.SlowLogFieldAccessor{
```
XuHuaiyu, line 30: why put the makeAccessor of plan digest here？

pkg/executor/adapter_slow_log.go
```diff
@@ -0,0 +1,183 @@
...
+		},
+		Match: func(_ *variable.SessionVars, items *variable.SlowQueryLogItems, threshold any) bool {
+			return variable.MatchEqual(threshold, strings.ToLower(items.PlanDigest))
+		},
+	}
+}
+
+// PrepareSlowLogItemsForRules builds a SlowQueryLogItems containing only the fields referenced by current session's SlowLogRules.
+// These pre-collected fields are later used for matching SQL execution details against the rules.
+func PrepareSlowLogItemsForRules(ctx context.Context, seVars *variable.SessionVars) *variable.SlowQueryLogItems {
```
XuHuaiyu, line 44: Will this function be called for each SQL statement? If yes, could it result in excessive small object allocations?

pkg/executor/adapter_slow_log.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+	}
+
+	return false
+}
+
+// CompleteSlowLogItemsForRules fills in the remaining SlowQueryLogItems fields
+// that are relevant to triggering the current session's SlowLogRules.
+// It's exporting for testing.
+func CompleteSlowLogItemsForRules(ctx context.Context, seVars *variable.SessionVars, items *variable.SlowQueryLogItems) {
```
XuHuaiyu, line 89: What's the difference between CompleteSlowLogItemsForRules and PrepareSlowLogItemsForRules

pkg/executor/adapter_slow_log.go
```diff
@@ -0,0 +1,183 @@
...
+			}
+			idxMap[idx] = struct{}{}
+			if buf.Len() > 1 {
+				buf.WriteByte(',')
+			}
+			buf.WriteString(idx)
+		}
+		buf.WriteByte(']')
+		indexNames = buf.String()
+	}
```
XuHuaiyu, line 130: I see this code block is duplicated in adapter.go. Can we extract it into a separate function?

pkg/executor/adapter_slow_log.go
zimulala: Here, the `GetPlanDigest` function needs to be called. This function is within the current package and cannot be moved to other packages (as it relies on many other packages).

pkg/executor/adapter_slow_log.go
zimulala: PrepareSlowLogItemsForRules: The obtained object only contains the contents of the fields present in the rule; fields that are not present in the rule will not be processed. 
CompleteSlowLogItemsForRules will also obtain the fields whose values were not retrieved in PrepareSlowLogItemsForRules.

pkg/executor/adapter_slow_log.go
zimulala: Yes, this function has not been called yet. 
In the next PR, the code in adapter.go will be directly replaced.

pkg/executor/adapter_slow_log.go
zimulala: Yes, if there is a slow log trigger rule.
Additionally,  only certain conditions within the rule will be retrieved; those that are not present will not be considered.

PR #63143, opened 2025-08-22
planner: create index with embedded analyze

AilinKid:

Issue Number: close https://github.com/pingcap/tidb/issues/63326

Problem Summary:

- Add `analyzed` field for modify column & adding index event, and skip the event in subscriber if the embedded analyze is executed.
- Add the analyze routine to DDL worker before the index changes are public.
- Add a new field `job.AnalyzeState` to "modify column" and "add index" jobs: 
  - `model.AnalyzeStateNone`: the analyze routine is not started.
  - `model.AnalyzeStateRunning`: the analyze routine is running. This happens after data reorganization is complete.
  - `model.AnalyzeStateDone`: the analyze routine is complete.
  - `model.AnalyzeStateSkipped`: the analyze routine is skipped. Possible reasons include `tidb_enable_ddl_analyze == off`, `tidb_analyze_version != 2` or changing indexes are empty.
- Pass `tidb_enable_ddl_analyze` and `tidb_analyze_version` from DDL front-end to DDL worker through job.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/statistics/handle/autoanalyze/autoanalyze.go
```diff
@@ -474,6 +474,9 @@ func RandomPickOneTableAndTryAutoAnalyze(
 				}
 
 				sql := "analyze table %n.%n"
+				if tblInfo.Name.L == "t" {
+					continue
```
AilinKid: block auto analyze

pkg/ddl/notifier/events.go
```diff
@@ -177,12 +177,14 @@ func (s *SchemaChangeEvent) GetAddColumnInfo() (
...
 	tableInfo *model.TableInfo,
 	modifiedColumns []*model.ColumnInfo,
+	analyzed bool,
 ) *SchemaChangeEvent {
 	return &SchemaChangeEvent{
 		inner: &jsonSchemaChangeEvent{
 			Tp:        model.ActionModifyColumn,
 			TableInfo: tableInfo,
 			Columns:   modifiedColumns,
+			Analyzed:  analyzed,
```
AilinKid, line 187: previously, for modify column, the handling of this kind of ddl event, it will init (insert a row into stats_hist) the stats related system table, while for analyze embedded case, the complete process of the new row is done by ddl, we don't need analyze additional steps here, otherwise, we will find some duplicated-key error which errored from here.

pkg/ddl/ddl.go
```diff
@@ -483,6 +483,27 @@ func (dc *ddlCtx) getResourceGroupTaggerForTopSQL(jobID int64) *kv.ResourceGroup
...
+		dc.jobCtx.jobCtxMap[jobID] = ctx
+	}
+	ctx.analyzeDone = ch
+}
+
+func (dc *ddlCtx) getAnalyzeDoneCh(jobID int64) chan struct{} {
+	dc.jobCtx.RLock()
+	defer dc.jobCtx.RUnlock()
+	ctx, exists := dc.jobCtx.jobCtxMap[jobID]
+	if !exists {
```
0xPoe, line 501: When will this occur? I suppose we can add an assertion to ensure it never happens in the test.

pkg/ddl/executor.go
```diff
@@ -6958,3 +6962,15 @@ func getScatterScopeFromSessionctx(sctx sessionctx.Context) string {
 	}
 	return scatterScope
 }
+
+func getEnableDDLAnalyze(sctx sessionctx.Context) string {
+	var enableDDLAnalyze string
+	val, ok := sctx.GetSessionVars().GetSystemVar(vardef.TiDBEnableDDLAnalyze)
+	if !ok {
```
0xPoe: May I ask when this will happen?

pkg/ddl/modify_column.go
```diff
@@ -618,6 +641,57 @@ func (w *worker) doModifyColumnTypeWithData(
...
+		})
+		w.ddlCtx.setAnalyzeDoneCh(job.ID, doneCh)
+	}
+	select {
+	case <-doneCh:
+		logutil.DDLLogger().Info("analyze table after create index done", zap.Int64("jobID", job.ID))
+		return true
+	case <-w.ctx.Done():
+		logutil.DDLLogger().Info("analyze table after create index context done", zap.Int64("jobID", job.ID))
+		return true
```
0xPoe, line 737: What happens if the analysis fails? It seems the DDL will still complete without any warnings.

pkg/ddl/notifier/events.go
0xPoe: We also need to skip the create index event for the priority queue. There is another DDL event handler associated with the priority queue.

0xPoe:
Some tests continue to fail. cc: @AilinKid

pkg/ddl/executor.go
AilinKid: in ddl frontend, it's user session will wrap the ddl as a job stored in system, here, we need to pass the user session's var into the job and be stored, otherwise, the owner who run this job in another tidb node can not feel the variable.

pkg/ddl/ddl.go
AilinKid: it means the first time to get to here.
image
you can tell the from caller, if call result is nil, it will start to pull a session and do the analyze job. but if is not nil, which means we just poll once this time with timeout as 10 sec

pkg/ddl/index.go
```diff
@@ -1450,18 +1479,14 @@ func loadCloudStorageURI(w *worker, job *model.Job) {
...
-		if done {
-			job.MarkNonRevertible()
-			if err == nil {
-				ver, err = updateVersionAndTableInfo(jobCtx, job, tbl.Meta(), true)
-			}
+	done, ver, err = doReorgWorkForCreateIndex(w, jobCtx, job, tbl, allIndexInfos)
+	if done {
+		// looks like we still need update version here, which couldn't be merged with the outer
+		if err == nil {
+			ver, err = updateVersionAndTableInfo(jobCtx, job, tbl.Meta(), true)
```
AilinKid: @tangenta for index side, i find we still need updateVersion here for done is true. Otherwise, the index's BackfillStateInapplicable state will be lost

pkg/ddl/modify_column.go
AilinKid: there will be a warn log. Ref: L677 above, do we need ddl warnings?

pkg/ddl/notifier/events.go
AilinKid: nice catch, you are right, addressed

pkg/ddl/index.go
```diff
@@ -1162,72 +1162,103 @@ SwitchIndexState:
...
-		// Set column index flag.
-		for _, indexInfo := range allIndexInfos {
-			AddIndexColumnFlag(tblInfo, indexInfo)
-			if isPK {
-				if err = UpdateColsNull2NotNull(tblInfo, indexInfo); err != nil {
-					return ver, errors.Trace(err)
+			if done {
+				// when done is true, the index's BackfillStateInapplicable need to be stored back.
+				if err == nil {
+					ver, err = updateVersionAndTableInfo(jobCtx, job, tbl.Meta(), true)
```
AilinKid: @tangenta  for index side, i find we still need updateVersion here for done is true. Otherwise, the index's BackfillStateInapplicable state will be lost

pkg/sessionctx/variable/sysvar.go
```diff
@@ -2796,6 +2796,11 @@ var defaultSysVars = []*SysVar{
 		GenerateBinaryPlan.Store(TiDBOptOn(val))
 		return nil
 	}},
+	{Scope: vardef.ScopeGlobal | vardef.ScopeSession, Name: vardef.TiDBEnableDDLAnalyze, Value: BoolToOnOff(vardef.DefTiDBEnableDDLAnalyze), Type: vardef.TypeBool,
```
0xPoe, line 2799: I think we should check if this variable is enabled in Analyze Version 1. With Analyze Version 1, it may be better to throw an error early.

AilinKid: make sense

AilinKid: addressed:

mysql> set @@tidb_analyze_version=1;
Query OK, 0 rows affected (0.006 sec)

mysql> set @@tidb_enable_ddl_analyze=1;
ERROR 1105 (HY000): tidb_enable_ddl_analyze can only be enable with tidb_analyze_version 2

AilinKid: and double checked in the real place when we try to do analyze with ddl

pkg/ddl/modify_column.go
```diff
@@ -618,19 +647,60 @@ func (w *worker) doModifyColumnTypeWithData(
...
+			if !ok {
+				return errors.Errorf("not restricted SQL executor: %T", sessCtx)
+			}
+			// internal sql may not init the analysis related variable correctly.
+			err = statsutil.UpdateSCtxVarsForStats(sessCtx)
+			if err != nil {
+				return err
+			}
+			// double-check the inner session analyze version is 2.
+			// nolint:forbidigo
```
0xPoe: May I ask which API will trigger this lint warning?

pkg/ddl/modify_column.go
```diff
@@ -618,19 +647,60 @@ func (w *worker) doModifyColumnTypeWithData(
...
+			if sessCtx.GetSessionVars().AnalyzeVersion == 1 {
+				return errors.New("tidb_enable_ddl_analyze can only be enabled with tidb_analyze_version 2")
+			}
+			_, _, err = exec.ExecRestrictedSQL(w.ctx, []sqlexec.OptionFuncAlias{sqlexec.ExecOptionUseCurSession, sqlexec.ExecOptionEnableDDLAnalyze}, "ANALYZE TABLE "+dbTable+";", "ddl analyze table")
+			if err != nil {
+				logutil.DDLLogger().Warn("analyze table failed",
```
0xPoe, line 718: Will this logger print the error stack? It would be easier to locate the bug if we can get the analyze error stack.

pkg/ddl/modify_column.go
0xPoe: So the failed job won’t block the DDL operation. Is this intentional by design?

pkg/ddl/executor.go
0xPoe: Could you please add some comments for it? Thank you!

pkg/ddl/modify_column.go
```diff
@@ -590,7 +615,11 @@ func (w *worker) doModifyColumnTypeWithData(
 			}
 		case model.StateDeleteOnly:
 			removedIdxIDs := removeOldObjects(tblInfo, oldCol, oldIdxInfos)
-			modifyColumnEvent := notifier.NewModifyColumnEvent(tblInfo, []*model.ColumnInfo{changingCol})
+			analyzed := false
+			if val, ok := job.GetSystemVars(vardef.TiDBEnableDDLAnalyze); ok && variable.TiDBOptOn(val) && tblInfo.Partition == nil {
```
0xPoe:
```suggestion
			if val, ok := job.GetSystemVars(vardef.TiDBEnableDDLAnalyze); ok && variable.TiDBOptOn(val) && tblInfo.GetPartitionInfo() == nil {
```

pkg/planner/core/planbuilder.go
```diff
@@ -2076,7 +2076,10 @@ func (b *PlanBuilder) getMustAnalyzedColumns(tbl *resolve.TableNameW, cols *calc
 		err := b.addColumnsWithVirtualExprs(tbl, cols, func(columns []*expression.Column) []expression.Expression {
 			virtualExprs := make([]expression.Expression, 0, len(tblInfo.Columns))
 			for _, idx := range tblInfo.Indices {
-				if idx.State != model.StatePublic || idx.MVIndex || idx.IsColumnarIndex() {
+				indexStateAnalyzable := idx.State == model.StatePublic ||
+					(idx.State == model.StateWriteReorganization && b.ctx.GetSessionVars().EnableDDLAnalyze)
```
0xPoe: Could you please add some comments for it? Thank you!

pkg/planner/core/planbuilder.go
```diff
@@ -2355,7 +2358,13 @@ func getModifiedIndexesInfoForAnalyze(
 	independentIdxsInfo = make([]*model.IndexInfo, 0)
 	specialGlobalIdxsInfo = make([]*model.IndexInfo, 0)
 	for _, originIdx := range tblInfo.Indices {
-		if originIdx.State != model.StatePublic {
+		// create index with analyze:
+		// * we analyze the reorg state index.
+		// change column with covered index analyze:
+		// * we analyze the public state index.
```
0xPoe: Could you please explain this comment a bit more? What do you mean by ‘covered index analyze’ here? I thought that for lossy DDL, it should behave the same as creating a new index.

pkg/statistics/handle/autoanalyze/priorityqueue/queue_ddl_handler.go
```diff
@@ -169,7 +169,11 @@ func (pq *AnalysisPriorityQueue) handleAddIndexEvent(
 	sctx sessionctx.Context,
 	event *notifier.SchemaChangeEvent,
 ) error {
-	tableInfo, idxes := event.GetAddIndexInfo()
+	tableInfo, idxes, analyzed := event.GetAddIndexInfo()
```
0xPoe, line 172: Could you please add a unit test case for this? Thank you!

pkg/statistics/handle/ddl/subscriber.go
```diff
@@ -112,7 +112,11 @@ func (h subscriber) handle(
 			}
 		}
 	case model.ActionModifyColumn:
-		newTableInfo, modifiedColumnInfo := change.GetModifyColumnInfo()
+		newTableInfo, modifiedColumnInfo, analyzed := change.GetModifyColumnInfo()
+		if analyzed {
+			return nil
+		}
+		// since tidb_enable_ddl_analyze will do analyze in ddl, skip col init here.
```
0xPoe: It seems this comment was placed in the wrong place.

0xPoe: Could you please add a unit test to cover this change? Thank you!

0xPoe (commented):
rest LGTM 

Thank you!

pkg/ddl/modify_column.go
AilinKid: /tidb/build/linter/forbidigo/analyzer.go

pkg/ddl/modify_column.go
AilinKid: make sense, sure

pkg/planner/core/planbuilder.go
AilinKid: you are right, it's old comment, fixed.

AilinKid:
reply to
> So the failed job won’t block the DDL operation. Is this intentional by design?

currently yes, ddl block is a bigger problem then a inside analyze err, if we encountered a error in analyze, we just stop it for a quick ddl finish of DDL, and print a warning to notify user who create about the stats analyze result.

pkg/ddl/index.go
tangenta: lgtm

pkg/ddl/modify_column.go
```diff
@@ -618,19 +648,64 @@ func (w *worker) doModifyColumnTypeWithData(
...
+			// double-check the inner session analyze version is 2.
+			// nolint:forbidigo
+			if sessCtx.GetSessionVars().AnalyzeVersion == 1 {
+				return errors.New("tidb_enable_ddl_analyze can only be enabled with tidb_analyze_version 2")
+			}
+			_, _, err = exec.ExecRestrictedSQL(w.ctx, []sqlexec.OptionFuncAlias{sqlexec.ExecOptionUseCurSession, sqlexec.ExecOptionEnableDDLAnalyze}, "ANALYZE TABLE "+dbTable+";", "ddl analyze table")
+			if err != nil {
+				buf := make([]byte, 4096)
+				stackSize := runtime.Stack(buf, false)
+				buf = buf[:stackSize]
```
tangenta: Maybe use `zap.Stack()` instead?

pkg/ddl/modify_column.go
```diff
@@ -546,37 +551,58 @@ func (w *worker) doModifyColumnTypeWithData(
...
-		moveChangingColumnInfoToDest(tblInfo, oldCol, changingCol, pos)
-		moveOldColumnInfo(tblInfo, oldCol)
-		moveIndexInfoToDest(tblInfo, changingCol, oldIdxInfos, changingIdxInfos)
-		updateModifyingCols(oldCol, changingCol)
+			if !done {
+				return ver, err
+			}
+			job.AnalyzeState = model.AnalyzeStateRunning
+		case model.AnalyzeStateRunning:
+			if val, ok := job.GetSystemVars(vardef.TiDBEnableDDLAnalyze); ok && variable.TiDBOptOn(val) && tbl.GetPartitionedTable() == nil {
```
tangenta: How about partition tables?

pkg/ddl/modify_column.go
0xPoe: Yes, it would be better to use `zap.Stack()` or the verbose error logger here.

0xPoe:
/hold 

For https://github.com/pingcap/tidb/pull/63143#pullrequestreview-3252587857

pkg/ddl/modify_column.go
AilinKid: addressed

AilinKid:
> /hold
> 
> For [#63143 (review)](https://github.com/pingcap/tidb/pull/63143#pullrequestreview-3252587857)

addressed

pkg/ddl/modify_column.go
tangenta: Get it. For the first phase, we don't support partitioned table because it is slow and could be problematic.

pkg/ddl/modify_column.go
```diff
@@ -546,37 +550,58 @@ func (w *worker) doModifyColumnTypeWithData(
...
```
tangenta: How about the `changingIdxs` slice is empty?

pkg/ddl/modify_column.go
```diff
@@ -546,37 +550,58 @@ func (w *worker) doModifyColumnTypeWithData(
...
+				job.AnalyzeState = model.AnalyzeStateDone
+			}
+			// previously, when reorg is done, and before we set the state to public, we need all other sub-task to
+			// be ready as well which is via setting this job as NornRevertible, then we can continue skip current
+			// sub and process the others.
+			// And when all sub-task are ready, which means each of them could be public in one more ddl round. And
+			// in onMultiSchemaChange, we just give all sub-jobs each one more round to public the schema, and only
+			// use the schema version generated once.
+			if job.MultiSchemaInfo != nil && job.MultiSchemaInfo.Revertible && job.AnalyzeState == model.AnalyzeStateDone {
+				// We need another round to wait for all the others sub-jobs to finish.
```
tangenta:
```suggestion
				// This is the final revertible state before public.
```

pkg/ddl/modify_column.go
AilinKid: how comes the empty slices

pkg/ddl/modify_column.go
```diff
@@ -590,7 +615,11 @@ func (w *worker) doModifyColumnTypeWithData(
 			}
 		case model.StateDeleteOnly:
 			removedIdxIDs := removeOldObjects(tblInfo, oldCol, oldIdxInfos)
-			modifyColumnEvent := notifier.NewModifyColumnEvent(tblInfo, []*model.ColumnInfo{changingCol})
+			analyzed := false
+			if val, ok := job.GetSystemVars(vardef.TiDBEnableDDLAnalyze); ok && variable.TiDBOptOn(val) && tblInfo.GetPartitionInfo() == nil {
+				analyzed = true
+			}
```
tangenta: Maybe wrap this into a function that returns a bool value so that we can reuse it in `modify column`

pkg/ddl/modify_column.go
```diff
@@ -546,37 +550,58 @@ func (w *worker) doModifyColumnTypeWithData(
...
-		updateObjectState(changingCol, changingIdxInfos, model.StatePublic)
-		markOldObjectRemoving(oldCol, changingCol, oldIdxInfos, changingIdxInfos, colName)
-		moveChangingColumnInfoToDest(tblInfo, oldCol, changingCol, pos)
-		moveOldColumnInfo(tblInfo, oldCol)
-		moveIndexInfoToDest(tblInfo, changingCol, oldIdxInfos, changingIdxInfos)
-		updateModifyingCols(oldCol, changingCol)
+			if !done {
+				return ver, err
+			}
+			job.AnalyzeState = model.AnalyzeStateRunning
```
tangenta: If analyze is not executed, `job.AnalyzeState` can still be "running"? How about adding a state `model.AnalyzeStateSkipped` to indicate this point?

pkg/ddl/modify_column.go
```diff
@@ -618,19 +647,61 @@ func (w *worker) doModifyColumnTypeWithData(
...
+			// internal sql may not init the analysis related variable correctly.
+			err = statsutil.UpdateSCtxVarsForStats(sessCtx)
+			if err != nil {
+				return err
+			}
+			// double-check the inner session analyze version is 2.
+			// nolint:forbidigo
+			if sessCtx.GetSessionVars().AnalyzeVersion == 1 {
+				return errors.New("tidb_enable_ddl_analyze can only be enabled with tidb_analyze_version 2")
+			}
```
tangenta: I suggest checking this before analyzing instead of returning an error.

pkg/statistics/handle/autoanalyze/priorityqueue/queue_ddl_handler_test.go
```diff
@@ -1044,3 +1044,63 @@ func TestAddIndexTriggerAutoAnalyzeWithStatsVersion1AndStaticPartition(t *testin
...
+	testKit.MustExec("alter table t add index idx(c1, c2)")
+
+	// Find the create index event.
+	addIndexEvent := findEvent(h.DDLEventCh(), model.ActionAddIndex)
+	tblInfo, idxInfo, analyzed := addIndexEvent.GetAddIndexInfo()
+	require.Equal(t, tableInfo.ID, tblInfo.ID)
+	require.Equal(t, analyzed, true)
+	require.Equal(t, idxInfo[0].Name.L, "idx")
+
+	// Handle the truncate table event.
```
tangenta:
```suggestion
	// Handle the create index event.
```

pkg/statistics/handle/autoanalyze/priorityqueue/queue_ddl_handler_test.go
```diff
@@ -1044,3 +1044,63 @@ func TestAddIndexTriggerAutoAnalyzeWithStatsVersion1AndStaticPartition(t *testin
...
+	err = statstestutil.HandleDDLEventWithTxn(h, addIndexEvent)
+	require.NoError(t, err)
+
+	sctx := testKit.Session().(sessionctx.Context)
+	// Handle the truncate table event in priority queue.
```
tangenta:
```suggestion
	// Handle the create index event in priority queue.
```

pkg/statistics/handle/autoanalyze/priorityqueue/queue_ddl_handler_test.go
```diff
@@ -1044,3 +1044,63 @@ func TestAddIndexTriggerAutoAnalyzeWithStatsVersion1AndStaticPartition(t *testin
...
+	require.NoError(t, pq.HandleDDLEvent(ctx, sctx, addIndexEvent))
+
+	// The table is truncated, the job should be removed from the priority queue.
+	isEmpty, err = pq.IsEmpty()
+	require.NoError(t, err)
+	require.True(t, isEmpty)
+
+	// create index.
```
tangenta:
```suggestion
	// modify column.
```

pkg/statistics/handle/autoanalyze/priorityqueue/queue_ddl_handler_test.go
```diff
@@ -1044,3 +1044,63 @@ func TestAddIndexTriggerAutoAnalyzeWithStatsVersion1AndStaticPartition(t *testin
...
+	testKit.MustExec("alter table t modify c1 varchar(10)")
+
+	// Find the create index event.
```
tangenta:
```suggestion
	// Find the modify column event.
```

pkg/ddl/modify_column.go
```diff
@@ -590,7 +621,12 @@ func (w *worker) doModifyColumnTypeWithData(
 			}
 		case model.StateDeleteOnly:
 			removedIdxIDs := removeOldObjects(tblInfo, oldCol, oldIdxInfos)
-			modifyColumnEvent := notifier.NewModifyColumnEvent(tblInfo, []*model.ColumnInfo{changingCol})
+			analyzed := false
+			// nolint:forbidigo
+			if checkAnalyzeNecessary(w.sess.GetSessionVars().AnalyzeVersion, job, changingIdxs, tblInfo) {
+				analyzed = true
+			}
+			modifyColumnEvent := notifier.NewModifyColumnEvent(tblInfo, []*model.ColumnInfo{changingCol}, analyzed)
```
tangenta:
```suggestion
			modifyColumnEvent := notifier.NewModifyColumnEvent(tblInfo, []*model.ColumnInfo{changingCol}, job.AnalyzeState == model.AnalyzeStateDone)
```

pkg/ddl/modify_column.go
```diff
@@ -618,19 +654,91 @@ func (w *worker) doModifyColumnTypeWithData(
...
+	if val, ok := job.GetSystemVars(vardef.TiDBEnableDDLAnalyze); ok && variable.TiDBOptOn(val) && tbl.GetPartitionInfo() == nil &&
+		// make sure that this reorg has affected the existed indexes, otherwise, re-analyze is not necessary.
+		len(changingIdxes) > 0 {
+		// double-check the inner ddl session analyze version is 2.
+		if analyzeVer == 2 {
+			return true
 		}
-		// We need another round to run the reorg process.
-		return false, ver, err
+		logutil.DDLLogger().Warn("tidb_enable_ddl_analyze can only be enabled with tidb_analyze_version 2", zap.Int("tidb_analyze_version", analyzeVer))
```
tangenta: Always print log when returning false, and change the log level to "info"?

0xPoe (commented):
I am curious whether we handle multiple ADD INDEX statements correctly.

It seems we may not handle the following case properly:

```sql
TiDB root@127.0.0.1:(none)> use test;
TiDB root@127.0.0.1:test> create table t1( id int, a int, b int, index idx(id, a));
Query OK, 0 rows affected
Time: 0.022s
TiDB root@127.0.0.1:test> insert into t1 values (1, 1, 1), (2, 2, 2), (3, 3, 3), (4, 4, 4), (5, 5, 5);
Query OK, 5 rows affected
Time: 0.002s
TiDB root@127.0.0.1:test> analyze table t1 all columns with 1 topn, 10 buckets;
Query OK, 0 rows affected
Time: 0.027s
TiDB root@127.0.0.1:test> select * from t1 where a = 1;
1 row in set
Time: 0.005s
TiDB root@127.0.0.1:test> show stats_histograms;
2 rows in set
Time: 0.006s
TiDB root@127.0.0.1:test> explain select * from t1 where a = 1;
3 rows in set
Time: 0.002s
TiDB root@127.0.0.1:test> show stats_histograms;
2 rows in set
Time: 0.006s
TiDB root@127.0.0.1:test>   ALTER TABLE t1
                       ->       ADD INDEX idx_a(a),
                       ->       ADD INDEX idx_b(b);
You're about to run a destructive command.
Do you want to proceed? (y/n): y
Your call!
Query OK, 0 rows affected
Time: 1.945s
TiDB root@127.0.0.1:test> explain select * from t1 where a = 1;
3 rows in set
Time: 0.002s
+---------------------------+---------+-----------+--------------------------+----------------------------------------------------------------------------+
| id                        | estRows | task      | access object            | operator info                                                              |
+---------------------------+---------+-----------+--------------------------+----------------------------------------------------------------------------+
| IndexLookUp_8             | 2.00    | root      |                          |                                                                            |
| ├─IndexRangeScan_6(Build) | 2.00    | cop[tikv] | table:t1, index:idx_a(a) | range:[1,1], keep order:false, stats:partial[idx_a:missing, idx_b:missing] |
| └─TableRowIDScan_7(Probe) | 2.00    | cop[tikv] | table:t1                 | keep order:false, stats:partial[idx_a:missing, idx_b:missing]              |
+---------------------------+---------+-----------+--------------------------+----------------------------------------------------------------------------+

```

AilinKid:
> I am curious whether we handle multiple ADD INDEX statements correctly.
> 
> It seems we may not handle the following case properly:
> 
> ```sql
> TiDB root@127.0.0.1:(none)> use test;
> TiDB root@127.0.0.1:test> create table t1( id int, a int, b int, index idx(id, a));
> Query OK, 0 rows affected
> Time: 0.022s
> TiDB root@127.0.0.1:test> insert into t1 values (1, 1, 1), (2, 2, 2), (3, 3, 3), (4, 4, 4), (5, 5, 5);
> Query OK, 5 rows affected
> Time: 0.002s
> TiDB root@127.0.0.1:test> analyze table t1 all columns with 1 topn, 10 buckets;
> Query OK, 0 rows affected
> Time: 0.027s
> TiDB root@127.0.0.1:test> select * from t1 where a = 1;
> 1 row in set
> Time: 0.005s
> TiDB root@127.0.0.1:test> show stats_histograms;
> 2 rows in set
> Time: 0.006s
> TiDB root@127.0.0.1:test> explain select * from t1 where a = 1;
> 3 rows in set
> Time: 0.002s
> TiDB root@127.0.0.1:test> show stats_histograms;
> 2 rows in set
> Time: 0.006s
> TiDB root@127.0.0.1:test>   ALTER TABLE t1
>                        ->       ADD INDEX idx_a(a),
>                        ->       ADD INDEX idx_b(b);
> You're about to run a destructive command.
> Do you want to proceed? (y/n): y
> Your call!
> Query OK, 0 rows affected
> Time: 1.945s
> TiDB root@127.0.0.1:test> explain select * from t1 where a = 1;
> 3 rows in set
> Time: 0.002s
> +---------------------------+---------+-----------+--------------------------+----------------------------------------------------------------------------+
> | id                        | estRows | task      | access object            | operator info                                                              |
> +---------------------------+---------+-----------+--------------------------+----------------------------------------------------------------------------+
> | IndexLookUp_8             | 2.00    | root      |                          |                                                                            |
> | ├─IndexRangeScan_6(Build) | 2.00    | cop[tikv] | table:t1, index:idx_a(a) | range:[1,1], keep order:false, stats:partial[idx_a:missing, idx_b:missing] |
> | └─TableRowIDScan_7(Probe) | 2.00    | cop[tikv] | table:t1                 | keep order:false, stats:partial[idx_a:missing, idx_b:missing]              |
> +---------------------------+---------+-----------+--------------------------+----------------------------------------------------------------------------+
> ```


img_v3_02qj_3430d884-2b7f-4b36-994c-1dd6a53e3ecg
i found the passed switch control is missing, which always skip the analyze process, i will look into it more

0xPoe:
> i found the passed switch control is missing, which always skip the analyze process, i will look into it more

I guess we lost it when we created the subtasks.

tangenta (approved):

PR #63162, opened 2025-08-22
planner: Convert OR to IN predicate | tidb-test=pr/2588

### LINKED ISSUE #59652: Rewrite exprs like `a = 10 or a = 20 or a = 30 ...` to `a in (10,20,30)` for better expr evaluation performance in the execution engine
## Enhancement

Currently, if conditions are not converted to index ranges, they will be evaluated in the `Selection` operator unchanged. However, for expressions like `a = 10 or a = 20 or a = 30 ...`, it's better to convert it to `a in (10,20,30)`, especially when there are lots of values.
It's because:
1. The structure of the first expression is more complicated. Each `eq` expression is an independent expression, and they are connected by `or` expressions, which can connect two expressions at once. The `in` expression has a more compact structure.
2. We have implemented more optimizations for the `in` expression, like https://github.com/tikv/tikv/pull/6000. According to members of the execution engine team, the performance of `in` is better than `eq` connected by `or` in tidb, tikv, and tiflash.

In one of the cases we met, there are ~1200 varchar values. The total row count is ~200k and the row count after the filter is ~50k. The query execution time with `or` + `eq` is ~5s while the equivalent query with `in` is ~1s.

#### Example:
```sql
create table t(a int, b int, c int);
explain select * from t where a = 1 or a = 2 or a = 3 or a = 4;
explain select * from t where a in (1,2,3,4);
```

Current execution plan:
```
> explain select * from t where a = 1 or a = 2 or a = 3 or a = 4;
+-------------------------+----------+-----------+---------------+--------------------------------------------------------------------------------+
| id                      | estRows  | task      | access object | operator info                                                                  |
+-------------------------+----------+-----------+---------------+--------------------------------------------------------------------------------+
| TableReader_7           | 40.00    | root      |               | data:Selection_6                                                               |
| └─Selection_6           | 40.00    | cop[tikv] |               | or(or(eq(test.t.a, 1), eq(test.t.a, 2)), or(eq(test.t.a, 3), eq(test.t.a, 4))) |
|   └─TableFullScan_5     | 10000.00 | cop[tikv] | table:t       | keep order:false, stats:pseudo                                                 |
+-------------------------+----------+-----------+---------------+--------------------------------------------------------------------------------+
```

Better execution plan:
```
> explain select * from t where a in (1,2,3,4);
+-------------------------+----------+-----------+---------------+--------------------------------+
| id                      | estRows  | task      | access object | operator info                  |
+-------------------------+----------+-----------+---------------+--------------------------------+
| TableReader_7           | 40.00    | root      |               | data:Selection_6               |
| └─Selection_6           | 40.00    | cop[tikv] |               | in(test.t.a, 1, 2, 3, 4)       |
|   └─TableFullScan_5     | 10000.00 | cop[tikv] | table:t       | keep order:false, stats:pseudo |
+-------------------------+----------+-----------+---------------+--------------------------------+
```

ghazalfamilyusa:

Issue Number: close #59652

### Problem Summary:
Initial motivation in https://github.com/pingcap/tidb/issues/59652 is that IN list equivalent to OR list produce better outcomes with index-selection and producing a unified canonical form using IN list produce better plans and make it consistent. 

Predicate simplification is extended to convert OR list to IN list as the last rewrite. This uncovered other limitations in predicate simplification that does not cover IN list and IR list equally. Additional changes were made: (1) flip flop between OR and IN lists to make sure existing prediate simplification is applied fully. (2) Use equivalence classes to support simplification. For example a1=a2 and a2 in (1,2,3) and a1 > 2 should simplify to a1=a2 and a1=3 and a2=3. 

### Followup?

This change may "expose" regressions if the planner produces a worse plan for IN list compared to OR list.
Fundamentally, we need planner to be consistent for both forms with focus on IN list after this PR.
Two examples I can see that should be addressed (not urgent):
1.  index range merge. OR may produce less number of ranges than IN but still same number of rows.
2. Index merge is not applied for IN compared to OR. An example i here tests/integrationtest/r/planner/core/indexmerge_path.result

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


hawkingrei:
I find a test case and it will be panic after this PR.

```
CREATE TABLE `t1` (                                                              
  `c1` timestamp NOT NULL DEFAULT '1974-10-16 00:00:00' ON UPDATE CURRENT_TIMESTAMP 
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;


delete from t1 where t1.c1 in ( '1991-04-12' ,'2019-11-05' ,'1969-12-31' ,'1986-04-19' ,'2025-09-24' ,'2005-07-28' ,'2018-01-09' ,'1999-03-09' ,'2008-03-27' );
```

ghazalfamilyusa:
> I find a test case and it will be panic after this PR.
> 
> ```
> CREATE TABLE `t1` (                                                              
>   `c1` timestamp NOT NULL DEFAULT '1974-10-16 00:00:00' ON UPDATE CURRENT_TIMESTAMP 
> ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
> 
> 
> delete from t1 where t1.c1 in ( '1991-04-12' ,'2019-11-05' ,'1969-12-31' ,'1986-04-19' ,'2025-09-24' ,'2005-07-28' ,'2018-01-09' ,'1999-03-09' ,'2008-03-27' );
> ```

Fixed. Added guard for expression.NewFunctionInternal and bail out if expression.NewFunctionInternal fails.

pkg/planner/core/rule/rule_predicate_simplification.go
```diff
@@ -519,10 +642,183 @@ func recursiveRemoveRedundantORBranch(sctx base.PlanContext, predicate expressio
...
+
+// inListToOrListAll applies inListToOrList on a slice of predicates.
+// It normalizes any "col IN (...)" into explicit OR-form so that
+// all subsequent simplification rules can work uniformly.
+// Example:
+//
+//	[ col1 IN (1,2), col2 = 5 ]
+//
+// → [ (col1 = 1 OR col1 = 2), col2 = 5 ]
+func inListToOrListAll(ctx base.PlanContext, exprs []expression.Expression) []expression.Expression {
```
fixdb, line 849: Instead of converting IN to OR list, can we merge IN list if there are multiple IN operators on the same column?

ghazalfamilyusa: The conversion here is to just facilitate applying rules consistently and not final result. In list intersection is a missing rewrite that we can add. It should be easy now with this infrastructure.

fixdb: I am afraid this will cause additional overhead. If the query is `select * from foo where a in (.....);`, the list has tens of thousands of constant values, we will end up with converting in list to OR list, then back to IN list, but with addditional overhead. And "IN" list only or "OR" list only is the common case, query with 'a in () or a = x' may be a rare case.

ghazalfamilyusa: Good point. I was thinking of limiting the conversion for long in list. May be a 1000 (done)

pkg/planner/core/rule/rule_predicate_simplification.go
```diff
@@ -199,15 +288,27 @@ func applyPredicateSimplification(sctx base.PlanContext, predicates []expression
...
+
+	simplifiedPredicate = inListToOrListAll(sctx, simplifiedPredicate)
+	simplifiedPredicate = shortCircuitLogicalConstants(ctx, simplifiedPredicate)
+	simplifiedPredicate = orList2InListAll(ctx, simplifiedPredicate)
+	simplifiedPredicate = mergeInAndNotEQLists(ctx, simplifiedPredicate)
+	simplifiedPredicate = inListToOrListAll(sctx, simplifiedPredicate)
+	removeRedundantORBranch(ctx, simplifiedPredicate)
+	simplifiedPredicate = pruneEmptyORBranches(ctx, simplifiedPredicate)
 	simplifiedPredicate = constraint.DeleteTrueExprs(exprCtx, sctx.GetSessionVars().StmtCtx, simplifiedPredicate)
+	simplifiedPredicate = orList2InListAll(ctx, simplifiedPredicate)
```
qw4990, line 329: Why do we have to call `inListToOrListAll` twice here? Since our goal is to convert all `OR` to `IN`, can we just call `orList2InListAll` here?

pkg/planner/core/rule/rule_predicate_simplification.go
```diff
@@ -199,15 +288,27 @@ func applyPredicateSimplification(sctx base.PlanContext, predicates []expression
...
-	simplifiedPredicate = mergeInAndNotEQLists(sctx, simplifiedPredicate)
-	removeRedundantORBranch(sctx, simplifiedPredicate)
-	simplifiedPredicate = pruneEmptyORBranches(sctx, simplifiedPredicate)
+
+	eqClasses := buildEquivalenceClasses(simplifiedPredicate)
+	ctx := simplifyContext{
+		planContext:  sctx,
+		ColEqClasses: eqClasses,
+	}
+
```
qw4990, line 320: Better to skip below advanced simplification for Plan Cache for safety:
```
	if expression.MaybeOverOptimized4PlanCache(sctx, predicates...) {
	    return ...	
	}
```

ghazalfamilyusa: If I understand this correctly., this suggests skipping cache unconditionally and this will make plan cache not usable for most queries since all queries go through this code path.

pkg/planner/core/rule/rule_predicate_simplification.go
ghazalfamilyusa: orList2InListAll is used for two purposes here: (1) allow IN list optimizations be applied to OR list equally and avoid implementing rewrites for both form redundancy (2) make final form as IN list.

pkg/planner/core/rule/rule_predicate_simplification.go
fixdb: I still prefer not to expand IN list to OR list. I don't mind keeping as it is if there are ''a in () or a = x'' in the query.

pkg/planner/core/rule/rule_predicate_simplification.go
qw4990: No, only these whose `predicates` contains some parameters.
For example, `select * from t where a in (?, ?, ?)` will skip this optimization, but `select * from t where a in (1, 2, 3)` wont'.

pkg/planner/core/rule/rule_predicate_simplification.go
qw4990: Got it, but why do we have to call them twice here? Because some intermediate function could generate new `or` or `in` predicates?

tests/integrationtest/r/planner/core/casetest/predicate_simplification.result
```diff
@@ -708,3 +706,122 @@ TopN	root		Column#13, offset:0, count:562300838
...
+d boolean
+);
+drop table if exists tt;
+create table tt (
+a int,
+b varchar(20),
+c double,
+d boolean
+);
+explain format=brief select * from t where a = 1 or a = 2 or a = 3;
```
qw4990, line 723: Could you add some new test cases with `or` and `in` together like `where a = 1 or a = 2 or a in (4, 5, 6)`

pkg/planner/core/rule/rule_predicate_simplification.go
ghazalfamilyusa: That is what I meant, all parameterized queries with in will skip the cacche which is a a big scope. I think it  is better to do it only when we have simplification. I know the skipping logic is scattered and need cleanup and centralized.

pkg/planner/core/rule/rule_predicate_simplification.go
ghazalfamilyusa: line 398 does convert in to or to apply AND/OR short circuit. line 300 convert back or to in to apply merging in list and <>. 303 need the form back in OR list and that is why 302 will do that. So, we need to flip flop between OR and IN list to apply all optimizations generically.

tests/integrationtest/r/planner/core/casetest/predicate_simplification.result
ghazalfamilyusa: Good idea. Done. This scenario is not covered (we get same result as  a = 1 or a = 2 or a =4 or a=5 or a = 6) yet and should be easy to add later on.

ghazalfamilyusa: The logic right now check if the full OR list like col = constant and does not consider partial case.

pkg/planner/core/rule/rule_predicate_simplification.go
```diff
@@ -142,12 +169,74 @@ func (*PredicateSimplification) Optimize(_ context.Context, p base.LogicalPlan,
...
+//
+// Otherwise, it returns the original expression.
+func inListToOrList(ctx base.PlanContext, expr expression.Expression) expression.Expression {
+	sf, ok := expr.(*expression.ScalarFunction)
+	if !ok || sf.FuncName.L != ast.In {
+		return expr
+	}
+
+	// First argument should be the column
+	col, ok := sf.GetArgs()[0].(*expression.Column)
```
hawkingrei, line 218: We could first put `sf.GetArgs()` into a variable and then read from this variable later. This way, the code would be more friendly for debugging.

pkg/planner/core/rule/rule_predicate_simplification.go
```diff
@@ -157,7 +246,7 @@ func updateInPredicate(ctx base.PlanContext, inPredicate expression.Expression,
 		return inPredicate, true
 	}
 	newValues := make([]expression.Expression, 0, len(v.GetArgs()))
-	evalCtx := ctx.GetExprCtx().GetEvalCtx()
+	evalCtx := ctx.planContext.GetExprCtx().GetEvalCtx()
```
hawkingrei, line 272: Can we set variable to save the result `ctx.planContext.GetExprCtx()`?

pkg/planner/core/rule/rule_predicate_simplification.go
```diff
@@ -361,36 +474,42 @@ func updateOrPredicate(ctx base.PlanContext, orPredicateList expression.Expressi
...
 			jthPredicate := predicates[j]
-			_, iType := FindPredicateType(sctx, ithPredicate)
-			_, jType := FindPredicateType(sctx, jthPredicate)
+			_, iType := FindPredicateType(ctx.planContext, ithPredicate)
+			_, jType := FindPredicateType(ctx.planContext, jthPredicate)
 			iType = comparisonPred(iType)
 			jType = comparisonPred(jType)
 			maybeOverOptimized4PlanCache := expression.MaybeOverOptimized4PlanCache(
-				sctx.GetExprCtx(),
+				ctx.planContext.GetExprCtx(),
```
hawkingrei, line 513: this can move out of the for-loop and save the result into tmp variable.

pkg/planner/core/rule/rule_predicate_simplification.go
```diff
@@ -480,31 +599,43 @@ func shortCircuitLogicalConstants(sctx base.PlanContext, predicates []expression
...
 	orFunc := predicate.(*expression.ScalarFunction)
 	orList := expression.SplitDNFItems(orFunc)
+	maybeOverOptimized4PlanCache := expression.MaybeOverOptimized4PlanCache(ctx.planContext.GetExprCtx(), predicate)
 
 	dedupMap := make(map[string]struct{}, len(orList))
 	newORList := make([]expression.Expression, 0, len(orList))
 
 	for _, orItem := range orList {
-		_, tp := FindPredicateType(sctx, orItem)
+		// Skip "col = NULL" cases
```
hawkingrei, line 644: So we also need to skip ```isnull(col)``` ?

ghazalfamilyusa: current OR list is checked fior col = constant only. Adding more cases is future.

pkg/planner/core/rule/rule_predicate_simplification.go
ghazalfamilyusa: The check is done for ith and jth predicate and not the whole predicate. Also, this is old code prior to this PR.

pkg/planner/core/rule/rule_predicate_simplification.go
ghazalfamilyusa: Is it for debugging? I do not see much value unless the intermediate variable is used a lot.

pkg/planner/core/rule/rule_predicate_simplification.go
ghazalfamilyusa: could be but debugging can do that as well. Do not see a need to change it.

pkg/planner/core/rule/rule_predicate_simplification.go
hawkingrei: I agree with fixdb's point of view. Another concern I have is that I see in the code there are additional places where some maps are allocated, and these maps will eventually be allocated on the heap, increasing the processing time of Golang GC. This may affect the final performance of TiDB. So, can we reuse these kinds of maps?

hawkingrei: Additionally, it's worth mentioning that Golang's performance is not optimal when it comes to garbage collecting a large number of small objects, which is what I'm concerned about.

pkg/planner/core/rule/rule_predicate_simplification.go
```diff
@@ -30,11 +30,30 @@ import (
...
-// PredicateSimplification consolidates different predcicates on a column and its equivalence classes.  Initial out is for
-//  1. in-list and not equal list intersection.
-//  2. Drop OR predicates if they are empty for this pattern: P AND (P1 OR P2 ... OR Pn)
-//     Pi is removed if P & Pi is false/empty.
-//  3. Simplify predicates with logical constants (True/False).
+// PredicateSimplification performs logical rewrites to consolidate and simplify
+// predicates on columns and their equivalence classes. It applies the following:
+//
+// 1. Equivalence-class analysis
+//   - Build equivalence classes from column equalities (e.g., t.a = tt.a).
```
hawkingrei, line 37: We are currently performing predicate simplification in multiple places, so it might be beneficial to provide a switch for converting 'OR' to 'IN'. This way, the conversion would only be executed within specific rules, rather than at every entry point.

hawkingrei: image
image

hawkingrei: The profiler is here

https://flamegraph.com/share/7e05250e-8366-11f0-94ec-760777e76ccd

ghazalfamilyusa: There is no relationship between the caller and what rules to apply. We may want to optimize the calls to rules as a general optimization where we check if the same rule applied to the same input. This is something useful outside this PR.

pkg/planner/core/rule/rule_predicate_simplification.go
ghazalfamilyusa: The in list to OR list is needed so we can apply rewrites to OR list without duplicating the code. I think we should be OK given the size limit control.

ghazalfamilyusa: > I agree with fixdb's point of view. Another concern I have is that I see in the code there are additional places where some maps are allocated, and these maps will eventually be allocated on the heap, increasing the processing time of Golang GC. This may affect the final performance of TiDB. So, can we reuse these kinds of maps?

Is that a general comment or specific to this PR?

ghazalfamilyusa: As I discussed with Haisheng, the plan is to reduce the size of in list to OR list to 50 to avoid potential regressions in the planner. Also, to re-iteratie the objective of this function:
- Make sure all rules are applied equally to OR and IN lists.
- Avoid a regression caused by OR to IN list in predicate simplification when constant propgation is involved. For example, "select 1 from t1 join t2 on a1=a2 where a1 in (a1 =1 or a1 = 2 or a1 =3 )  and a2 > 2" is rewritten by new rule to select 1 from t1 join t2 on a1=a2 where a1 in (1,2,3)  and a2 > 2 and constant  propagation produces select 1 from t1 join t2 on a1=a2 where a1 in (1,2,3)  and a2 > 2 and a1 > 2 and a2 in (1,2,3); which can not be simplified by pruneEmptyORBranches (which can simplify it if it is in the original OR list form the second round predicate simplification is called).

pkg/planner/core/rule/rule_predicate_simplification.go
qw4990: `a big scope` this should be OK, only a few users of TiDB are using parameterized queries actually. And this is for safety, we've met several bugs caused by query rewrite without handling plan cache properly.
So the current principle is that we just disable these rewrites for Plan Cache by default at first, and then enable them later as we add more test cases by another dedicated PR.

pkg/planner/core/rule/rule_predicate_simplification.go
qw4990: Got it, please add some comments here.

tests/integrationtest/r/planner/core/casetest/predicate_simplification.result
```diff
@@ -708,3 +706,127 @@ TopN	root		Column#13, offset:0, count:562300838
...
+└─TableFullScan	10000.00	cop[tikv]	table:t	keep order:false, stats:pseudo
+explain format=brief select * from t where (a = 1 and b = 2) or a = 3;
+id	estRows	task	access object	operator info
+TableReader	17.99	root		data:Selection
+└─Selection	17.99	cop[tikv]		or(and(eq(planner__core__casetest__predicate_simplification.t.a, 1), eq(cast(planner__core__casetest__predicate_simplification.t.b, double BINARY), 2)), eq(planner__core__casetest__predicate_simplification.t.a, 3))
+  └─TableFullScan	10000.00	cop[tikv]	table:t	keep order:false, stats:pseudo
+explain format=brief select * from t where a = 1 or a = 2 or a = 3 or a in (4,5,6);
+id	estRows	task	access object	operator info
+TableReader	60.00	root		data:Selection
+└─Selection	60.00	cop[tikv]		or(or(eq(planner__core__casetest__predicate_simplification.t.a, 1), eq(planner__core__casetest__predicate_simplification.t.a, 2)), or(eq(planner__core__casetest__predicate_simplification.t.a, 3), in(planner__core__casetest__predicate_simplification.t.a, 4, 5, 6)))
```
qw4990, line 831: Why is it not working for this case? Should the final result be "a in (1, 2, 3, 4, 5, 6)"?

qw4990 (commented):
Almost LGTM, please address my left comments, THX! @ghazalfamilyusa 
For parameterized queries, I still think we can skip them now for safety 🤔

tests/integrationtest/r/planner/core/casetest/predicate_simplification.result
ghazalfamilyusa: @qw4990 : I mentioned this in previous comemnt. The current code handles col = constant or col = const2 ... and skips any other case in a disjunct. That is why this case is not covered which  seems to be uncommon. Extending the code for this case should be straightforward if needed later on.

pkg/planner/core/casetest/partition/testdata/integration_partition_suite_out.json
```diff
@@ -903,16 +903,16 @@
         "SQL": "explain format = 'brief' select * from t where b = 1 OR b = 2",
         "DynamicPlan": [
           "IndexReader 2.00 root partition:all index:IndexRangeScan",
-          "└─IndexRangeScan 2.00 cop[tikv] table:t, index:b(b) range:[1,2], keep order:false"
+          "└─IndexRangeScan 2.00 cop[tikv] table:t, index:b(b) range:[1,1], [2,2], keep order:false"
         ],
```
King-Dylan, line 907: In extreme cases, there is a potential fallback where a range scan of [1,100] may perform better than individual point scans like [1,1], [2,2], …, [100,100] on the TiKV side.
```mysql> explain analyze select * from t where b in(1,2,3,4,5,6,7,8,9,10);
+------------------------+---------+---------+-----------+---------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------+-----------+------+
| id                     | estRows | actRows | task      | access object       | execution info                                                                                                                                                                                                                                                                                                                                                                                                                   | operator info                                                                                                | memory    | disk |
+------------------------+---------+---------+-----------+---------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------+-----------+------+
| IndexReader_7          | 1.25    | 10      | root      | partition:all       | time:559.5µs, open:21.3µs, close:1.88µs, loops:2, RU:1.51, cop_task: {num: 3, max: 498.9µs, min: 368.1µs, avg: 442.8µs, p95: 498.9µs, max_proc_keys: 4, p95_proc_keys: 4, tot_proc: 234.9µs, tot_wait: 341.6µs, copr_cache_hit_ratio: 0.00, build_task_duration: 5.83µs, max_distsql_concurrency: 3}, fetch_resp_duration: 497.6µs, rpc_info:{Cop:{num_rpc:3, total_time:1.31ms}}                                                | index:IndexRangeScan_6                                                                                       | 460 Bytes | N/A  |
| └─IndexRangeScan_6     | 1.25    | 10      | cop[tikv] | table:t, index:b(b) | tikv_task:{proc max:0s, min:0s, avg: 0s, p80:0s, p95:0s, iters:3, tasks:3}, scan_detail: {total_process_keys: 10, total_process_keys_size: 460, total_keys: 40, get_snapshot_time: 284µs, rocksdb: {delete_skipped_count: 10, key_skipped_count: 20, block: {}}}, time_detail: {total_process_time: 234.9µs, total_wait_time: 341.6µs, tikv_grpc_process_time: 25µs, tikv_grpc_wait_time: 40.7µs, tikv_wall_time: 766.2µs}       | range:[1,1], [2,2], [3,3], [4,4], [5,5], [6,6], [7,7], [8,8], [9,9], [10,10], keep order:false, stats:pseudo | N/A       | N/A  |
+------------------------+---------+---------+-----------+---------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------+-----------+------+
2 rows in set (0.00 sec)

mysql> explain analyze select * from t where b =1 or b=2 or b=3 or b=4 or b=5 or b=6 or b=7 or b=8 or b=9 or b=10;
+------------------------+---------+---------+-----------+---------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------+-----------+------+
| id                     | estRows | actRows | task      | access object       | execution info                                                                                                                                                                                                                                                                                                                                                                                                                  | operator info                                | memory    | disk |
+------------------------+---------+---------+-----------+---------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------+-----------+------+
| IndexReader_7          | 1.25    | 10      | root      | partition:all       | time:342.7µs, open:13.2µs, close:1.5µs, loops:2, RU:1.45, cop_task: {num: 3, max: 300.5µs, min: 266.6µs, avg: 284.9µs, p95: 300.5µs, max_proc_keys: 4, p95_proc_keys: 4, tot_proc: 42.8µs, tot_wait: 80.4µs, copr_cache_hit_ratio: 0.00, build_task_duration: 3.92µs, max_distsql_concurrency: 3}, fetch_resp_duration: 317.7µs, rpc_info:{Cop:{num_rpc:3, total_time:844.2µs}}                                                 | index:IndexRangeScan_6                       | 417 Bytes | N/A  |
| └─IndexRangeScan_6     | 1.25    | 10      | cop[tikv] | table:t, index:b(b) | tikv_task:{proc max:0s, min:0s, avg: 0s, p80:0s, p95:0s, iters:3, tasks:3}, scan_detail: {total_process_keys: 10, total_process_keys_size: 460, total_keys: 13, get_snapshot_time: 21µs, rocksdb: {delete_skipped_count: 10, key_skipped_count: 20, block: {}}}, time_detail: {total_process_time: 42.8µs, total_wait_time: 80.4µs, tikv_grpc_process_time: 48.8µs, tikv_grpc_wait_time: 77.8µs, tikv_wall_time: 295.5µs}       | range:[1,10], keep order:false, stats:pseudo | N/A       | N/A  |
+------------------------+---------+---------+-----------+---------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------+-----------+------+
2 rows in set (0.00 sec)
```

ghazalfamilyusa: correct. There is a followup to merge adjacent ranges in the planner.

pkg/planner/core/rule/rule_predicate_simplification.go
```diff
@@ -142,12 +169,97 @@ func (*PredicateSimplification) Optimize(_ context.Context, p base.LogicalPlan,
...
+	// First argument should be the column
+	col, ok := sf.GetArgs()[0].(*expression.Column)
+	if !ok {
+		return expr
+	}
+
+	// Quick size check: expanding a large IN-list into an OR-tree can cause significant overhead.
+	// If the number of values exceeds maxInListToExpand, skip the rewrite.
+	if len(sf.GetArgs())-1 > maxInListToExpand {
+		return expr
```
King-Dylan, line 226: Do we have any plans for this? I think there’s also a potential optimization here: if the values in the inlist are consecutive, converting them into OR conditions could actually yield better performance, since OR can be built into a large range.

King-Dylan: This is largely affected by underlying TiKV factors, such as the number of MVCC versions, so I think we can stick with the current approach for now.

pkg/planner/core/rule/rule_predicate_simplification.go
```diff
@@ -438,32 +580,32 @@ func shortCircuitANDORLogicalConstants(sctx base.PlanContext, predicate expressi
...
 
 // shortCircuitLogicalConstants evaluates a list of predicates, applying short-circuit logic
 // to simplify the list and eliminate redundant or trivially true/false predicates.
-func shortCircuitLogicalConstants(sctx base.PlanContext, predicates []expression.Expression) []expression.Expression {
+func shortCircuitLogicalConstants(ctx simplifyContext, predicates []expression.Expression) []expression.Expression {
 	finalResult := make([]expression.Expression, 0, len(predicates))
 
 	for _, predicate := range predicates {
-		predicate, predicateType := processCondition(sctx, predicate)
+		predicate, predicateType := processCondition(ctx, predicate)
```
King-Dylan, line 609: Declaration of "predicate" shadows declaration

pkg/planner/core/rule/rule_predicate_simplification.go
```diff
@@ -438,32 +580,32 @@ func shortCircuitANDORLogicalConstants(sctx base.PlanContext, predicate expressi
...
 // and returns the potentially simplified condition and its updated type.
-func processCondition(sctx base.PlanContext, condition expression.Expression) (expression.Expression, predicateType) {
+func processCondition(ctx simplifyContext, condition expression.Expression) (expression.Expression, predicateType) {
 	applied := false
-	maybeOverOptimized4PlanCache := expression.MaybeOverOptimized4PlanCache(sctx.GetExprCtx(), condition)
-	_, conditionType := FindPredicateType(sctx, condition)
+	maybeOverOptimized4PlanCache := expression.MaybeOverOptimized4PlanCache(ctx.planContext.GetExprCtx(), condition)
+	_, conditionType := FindPredicateType(ctx.planContext, condition)
 
 	if conditionType == orPredicate {
```
King-Dylan, line 588: Using a switch on conditionType might be better.

tests/integrationtest/r/planner/core/plan_cache.result
```diff
@@ -1058,7 +1058,7 @@ a	b
 2023-01-21	2023-01-05
 select @@last_plan_from_cache;
 @@last_plan_from_cache
-1
+0
```
King-Dylan, line 1061: Is this regression expected?

pkg/planner/core/rule/rule_predicate_simplification.go
ghazalfamilyusa: The inlist to OR is intermediate to allow OR based predicate simplification (same module). The size control to avoid potential performance for very large IN list.

tests/integrationtest/r/planner/core/plan_cache.result
ghazalfamilyusa: Yes. We skipped plan cache due to simplification.

tests/integrationtest/r/planner/core/indexmerge_path.result
```diff
@@ -754,18 +754,9 @@ c = 3
...
-  ├─IndexRangeScan(Build)	cop[tikv]	table:t, index:mvi1(c, cast(json_extract(`j`, _utf8mb4'$.a') as unsigned array), b)	range:[2 6,2 6], keep order:false, stats:pseudo
-  ├─IndexRangeScan(Build)	cop[tikv]	table:t, index:mvi1(c, cast(json_extract(`j`, _utf8mb4'$.a') as unsigned array), b)	range:[3 4,3 4], keep order:false, stats:pseudo
-  ├─IndexRangeScan(Build)	cop[tikv]	table:t, index:mvi1(c, cast(json_extract(`j`, _utf8mb4'$.a') as unsigned array), b)	range:[3 5,3 5], keep order:false, stats:pseudo
-  ├─IndexRangeScan(Build)	cop[tikv]	table:t, index:mvi1(c, cast(json_extract(`j`, _utf8mb4'$.a') as unsigned array), b)	range:[3 6,3 6], keep order:false, stats:pseudo
-  └─Selection(Probe)	cop[tikv]		or(eq(planner__core__indexmerge_path.t.c, 1), or(eq(planner__core__indexmerge_path.t.c, 2), eq(planner__core__indexmerge_path.t.c, 3)))
-    └─TableRowIDScan	cop[tikv]	table:t	keep order:false, stats:pseudo
+└─TableReader	root		data:Selection
+  └─Selection	cop[tikv]		in(planner__core__indexmerge_path.t.c, 1, 2, 3)
+    └─TableFullScan	cop[tikv]	table:t	keep order:false, stats:pseudo
```
King-Dylan, line 759: This also looks like a regression case. If the table contains a large number of rows, the impact should be quite noticeable.

tests/integrationtest/r/planner/core/plan_cache.result
King-Dylan: Got it

hawkingrei (approved):
It seems that further improvements are still needed going forward.

ghazalfamilyusa:
> It seems that further improvements are still needed going forward.

Yes. Further additional improvements are needed as pointed out in the description.

tests/integrationtest/r/planner/core/plan_cache.result
terry1purcell: Why are we OK with a plan cache regression?

ghazalfamilyusa: For those cases, predicate simplifications or other rewrites can cause that. So, there could a tradeoff and in general we assume these rewrites have more value than plan cache.

tests/integrationtest/r/planner/core/indexmerge_path.result
ghazalfamilyusa: This is one of the follow up cases where we need to make sure IN and equivalent OR produce same plan.

pkg/planner/core/rule/rule_predicate_simplification.go
ghazalfamilyusa: I see. In general, there is a tradeoff between the plan cache and rewrites. This change is more generic than this PR. Can we do this as a follow up PR?

ghazalfamilyusa:
> Almost LGTM, please address my left comments, THX! @ghazalfamilyusa For parameterized queries, I still think we can skip them now for safety 🤔

Sorry I missed this and the other thread is hidden and did not see it. I replied to your comment and agree with it (safety of correctness over optimizations). But, the change is more generic for all simplifications. How about we do it as a follow up.

PR #63188, opened 2025-08-26
planner: Add more context to CompareCol2Len

terry1purcell:

Issue Number: ref #59333

Problem Summary:

Update the CompareCol2Len function to return which "map" is the winner when they otherwise are not comparable. This will allow the caller - and specifically compareCandidates func (in skyLinePruning) - to use the information in conjunction with other criteria to improve index pruning.

This issue was exposed while testing PR #63085.

In TestSkylinePruning - the following 2 indexes exist:
KEY c_d_e_str (c_str, d_str, e_str),
KEY e_d_c_str_prefix (e_str, d_str, c_str(10))
Given the following query:
select * from t where c_str = 'a' and d_str = 'b' and e_str = 'c';
Both indexes have 3 equal predicates, but they are NOT comparable by the original logic. Function compareCandidates (within skylinePruning) does not currently recognize that the 3 column match on the first index is better than the 2nd.

After this PR (63188) and 63085 merge, we will be able to enhance compareCandidates to understand the "winner/loser" from CompareCol2Len and compare this to the EqOrInCondCount.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/util/path.go
+++ b/pkg/planner/util/path.go
@@ -386,7 +386,7 @@ func (c1 Col2Len) dominate(c2 Col2Len) bool {
 // CompareCol2Len will compare the two Col2Len maps.
 // The first return value:
 // (1) -1 means that len(c1) is less than len(c2);
-// (2) 0 means that c1 equals to c2;
+// (2) 0 means that len(c1) equals to len(c2);
 // (3) 1 means that c1 is better than c2;
 // The 2nd return value is used to indicate whether they are comparable. If they are NOT comparable, then the caller
 // should use other criteria to determine whether the winner is justified.


pkg/planner/util/path.go
```diff
@@ -383,30 +383,36 @@ func (c1 Col2Len) dominate(c2 Col2Len) bool {
...
 	for colID, colLen2 := range c2 {
 		colLen1, ok := c1[colID]
 		if !ok || colLen1 != colLen2 {
-			return 0, false
+			// If lengths are not equal, return 1 if c1 is larger, or -1 if c2 is larger
+			if colLen1 > colLen2 {
+				return 1, false
+			}
+			return -1, false
 		}
```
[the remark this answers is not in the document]
terry1purcell: Will resolve in next commit.

henrybw (approved):
LGTM

pkg/planner/util/path.go
```diff
@@ -383,31 +383,40 @@ func (c1 Col2Len) dominate(c2 Col2Len) bool {
...
+// The 2nd return value is used to indicate whether they are comparable. If they are NOT comparable, then the caller
+// should use other criteria to determine whether the winner is justified.
 func CompareCol2Len(c1, c2 Col2Len) (int, bool) {
 	l1, l2 := len(c1), len(c2)
 	if l1 > l2 {
 		if c1.dominate(c2) {
 			return 1, true
 		}
-		return 0, false
+		return 1, false
```
lance6716, line 399: Just double check, I searched some caller of this function. Seem when return `..., false` there's no usage of the first return value. And in this PR the first return value is more precise. Do you plan to add usage on it in `..., false` case?

terry1purcell: Yes - it is planned to be used in compareCandidates function that is called within skylinePruning.

pkg/planner/util/path.go
```diff
@@ -383,31 +383,40 @@ func (c1 Col2Len) dominate(c2 Col2Len) bool {
...
 	for colID, colLen2 := range c2 {
 		colLen1, ok := c1[colID]
-		if !ok || colLen1 != colLen2 {
+		if !ok {
 			return 0, false
 		}
+		if colLen1 != colLen2 {
+			// If lengths are not equal, return 1 if c1 is larger, or -1 if c2 is larger
+			if colLen1 > colLen2 {
+				return 1, false
```
qw4990, line 416: How to understand the returned values here? `c1` and `c2` are uncomparable, but `c1` is better than `c2`?

terry1purcell: Yes - let me give you an example. In TestSkylinePruning - the following 2 indexes exist:
KEY c_d_e_str (c_str, d_str, e_str),
KEY e_d_c_str_prefix (e_str, d_str, c_str(10))
Given the following query:
select * from t where c_str = 'a' and d_str = 'b' and e_str = 'c';
Both indexes have 3 equal predicates, but they are NOT comparable by the original logic. Function compareCandidates (within skylinePruning) does not currently recognize that the 3 column match on the first index is better than the 2nd.

I will add this example to the PR description and reference the other PR where this issue was uncovered.

qw4990: Gotcha, better to add this example into our comments.

qw4990 (approved):

pkg/planner/util/path.go
```diff
@@ -383,31 +383,40 @@ func (c1 Col2Len) dominate(c2 Col2Len) bool {
 	return true
 }
 
-// CompareCol2Len will compare the two Col2Len maps. The last return value is used to indicate whether they are comparable.
-// When the second return value is true, the first return value:
+// CompareCol2Len will compare the two Col2Len maps.
+// The first return value:
 // (1) -1 means that c1 is worse than c2;
```
elsa0520:
```suggestion
// (1) -1 means that len(c1) is less than len(c2);
```

pkg/planner/util/path.go
```diff
@@ -383,31 +383,40 @@ func (c1 Col2Len) dominate(c2 Col2Len) bool {
...
 // (2) 0 means that c1 equals to c2;
```
elsa0520:
```suggestion
// (2) 0 means that len(c1) equals to len(c2);
```

pkg/planner/util/path.go
```diff
@@ -383,31 +383,40 @@ func (c1 Col2Len) dominate(c2 Col2Len) bool {
...
 // (3) 1 means that c1 is better than c2;
```
elsa0520, line 390: Same as above

pkg/planner/util/path.go
elsa0520, line 416: Should we return (1, true) in this case? Because the len(c1) is same as the len(c2) but the column c len in C1 is more than column c len in C2.

pkg/planner/util/path.go
elsa0520, line 399: It seems that this is not a complete change. This function's behavior has changed, but the external processing logic hasn't.
And this function change will directly affect the behavior of the master branch.
So, Maybe it might be better to submit all subsequent changes together.

pkg/planner/util/path.go
terry1purcell: We can't return true, because the caller expects them to be equivalent if it is true. A length difference should not result in true.

pkg/planner/util/path.go
terry1purcell: See other PR #63085 - which added a testcase that exposed this issue. Until both of these PRs are merged, I cannot resolve the usage of this new behavior from the calling function.

PR #63189, opened 2025-08-26
planner: fix handling of empty plan digests in `getBindingPlanInfo`

henrybw:

Issue Number: ref #60148

Problem Summary:

When SPM's `EXPLAIN EXPLORE` obtains plan and execution information for existing bindings, it uses the plan digest stored in the binding to obtain execution information. If the plan digest is empty, we run the optimizer to calculate the plan digest, which we then use to obtain execution info. However, we don't update the binding structure with this calculated plan digest, which means the plan displayed to the user has an empty plan digest, resulting in `EXPLAIN EXPLORE` producing the following incorrect output:

```
explain_analyze: EXPLAIN ANALYZE ''                                                                                                                                                  
binding: CREATE GLOBAL BINDING FROM HISTORY USING PLAN DIGEST ''
``` 

* If the binding's stored plan digest is empty, update `binding.PlanDigest` with the newly calculated plan digest.
* If we fail to get the plan digest, we now skip that plan (instead of just logging an error and including the plan).
* Modify the `TestExplainExploreBasic` unit test to ensure that no plan digest is ever empty.
    * Also, change `require.True` to `require.Equal` when checking row counts, so test failures will show what the actual (unexpected) output was, instead of just saying the condition was false.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/bindinfo/binding_auto.go
+++ b/pkg/bindinfo/binding_auto.go
@@ -245,7 +245,9 @@ func (ba *bindingAuto) getBindingPlanInfo(currentDB, sqlOrDigest, charset, colla
 			}); err != nil {
 				bindingLogger().Error("get plan digest failed",
 					zap.String("bind_sql", binding.BindSQL), zap.Error(err))
+				continue
 			}
+			binding.PlanDigest = planDigest
 		}
 
 		pInfo, err := ba.getPlanExecInfo(planDigest)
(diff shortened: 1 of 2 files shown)


hawkingrei (approved):

hawkingrei:
/hold

@qw4990 PTAL

PR #63191, opened 2025-08-26
ddl: optimize the asyncNotifyEvent for test 

hawkingrei:

Issue Number: close #xxx

Problem Summary:

`fail to notify DDL event` has too much in our test case. 

The main issue is that the handling of this channel is all manual. Some tests might have forgotten to handle this part, leading to multiple retries. These retries are unnecessary, so we should reduce the number of retries.

image

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/ddl.go
+++ b/pkg/ddl/ddl.go
@@ -611,12 +611,12 @@ func asyncNotifyEvent(jobCtx *jobContext, e *notifier.SchemaChangeEvent, job *mo
 			// Try sending the event to the channel with a backoff strategy to avoid blocking indefinitely.
 			// Since most unit tests don't consume events, we make a few attempts and then give up rather
 			// than blocking the DDL job forever on a full channel.
-			for range 10 {
+			for range 3 {
 				select {
 				case ch <- e:
 					break forLoop
 				default:
-					time.Sleep(time.Microsecond * 10)
+					time.Sleep(time.Microsecond * 2)
 				}
 			}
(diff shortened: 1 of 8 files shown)


pkg/ddl/ddl.go
```diff
@@ -611,12 +611,12 @@ func asyncNotifyEvent(jobCtx *jobContext, e *notifier.SchemaChangeEvent, job *mo
 			// Try sending the event to the channel with a backoff strategy to avoid blocking indefinitely.
 			// Since most unit tests don't consume events, we make a few attempts and then give up rather
 			// than blocking the DDL job forever on a full channel.
-			for range 10 {
+			for range 3 {
 				select {
 				case ch <- e:
```
lance6716, line 616: maybe it's doomed to fail because in test there's no receiver of channel. @0xPoe can we delete this sending now?

hawkingrei: The handling here is currently done manually.

hawkingrei: With `HandleNextDDLEventWithTxn`.

hawkingrei: Whether to delete it can be discussed in another issue. Let's first alleviate the problem in this PR.

PR #63236, opened 2025-08-27
ddl/placement: fix meta range to exclude RawKV keys from placement rules

### LINKED ISSUE #63133: ALTER RANGE meta generates confusing key range that overlaps RawKV prefix
## Bug Report

When using ALTER RANGE meta in TiDB to assign placement rules to the meta key range, the generated end key is currently set to the table prefix (tablePrefix), as seen [here](https://github.com/pingcap/tidb/blob/d48ac38676f9569112cd881edbc700fedd675b7d/pkg/ddl/placement/bundle.go#L466).

However, there is a RawKV prefix ('r', 0x72) that falls between the meta key range end and the table key range start. This means the generated placement rule technically covers both:

Meta key ranges (intended target) ✅
Raw key ranges ('r' prefix) ❌ (not expected for most TiDB users)

Most TiDB customers never use the RawKV interface, so this overlap usually has no practical effect. Still, the current behavior is confusing and misleading because:

The placement rules suggest RawKV keys are part of the meta range.
It creates unnecessary ambiguity when reasoning about region placement.
For users who do use RawKV, this could lead to unexpected replica placement policies being applied.

* Create a placement policy for meta regions:
`CREATE PLACEMENT POLICY `meta_policy` FOLLOWERS=4;`
* Assign the policy to the meta key range:
`ALTER RANGE META PLACEMENT POLICY = 'meta_policy';`
* Check the generated placement rules in PD 
* Observe that the end key of the meta range is set to the table prefix (t, 0x74).
* Note that the RawKV prefix (r, 0x72) falls between the meta prefix (m, 0x6d) and the table prefix.
The placement rule for the meta range should exclude the RawKV keyspace (r, 0x72) and only cover the actual meta range.
That is, the key range should be:

`[metaPrefix ('m'), rawPrefix ('r'))`
instead of:

`[metaPrefix ('m'), tablePrefix ('t'))`
The placement rule currently sets the end key to the table prefix (t, 0x74), which includes the RawKV keyspace (r, 0x72`) in the range.
This is confusing because it suggests RawKV keys are part of the meta range, even though most TiDB users do not use RawKV.
For users who do use RawKV, it may cause unintended placement of raw keys.
v7.5 & v8.5

HaoW30:

Issue Number: close [#63133 ](https://github.com/pingcap/tidb/issues/63133)

Problem Summary:

When using ALTER RANGE META to assign placement rules to the meta key range, the generated end key was previously set to the table prefix ('t', 0x74). This incorrectly included RawKV keys ('r', 0x72) in the meta range placement rules, as the range was [metaPrefix, tablePrefix) = ['m', 't').

This fix changes the end key to use the raw prefix ('r', 0x72), making the meta range [metaPrefix, rawPrefix) = ['m', 'r'). This ensures:

1. Meta placement rules only apply to actual meta keys
2. RawKV keys are excluded from meta placement policies
3. No unintended placement policy application for users who use RawKV

The key ordering is: meta ('m', 0x6d) < raw ('r', 0x72) < table ('t', 0x74)

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/placement/bundle.go
+++ b/pkg/ddl/placement/bundle.go
@@ -463,7 +463,10 @@ func GetRangeStartAndEndKeyHex(rangeBundleID string) (startKey string, endKey st
 	if rangeBundleID == TiDBBundleRangePrefixForMeta {
 		// Use codec.EncodeBytes to properly encode the meta prefix in table mode
 		startKey = hex.EncodeToString(codec.EncodeBytes(nil, metaPrefix))
-		endKey = hex.EncodeToString(codec.EncodeBytes(nil, tablecodec.GenTablePrefix(0)))
+		// Use rawPrefix as endKey to ensure the meta range [metaPrefix, rawPrefix) only
+		// contains meta keys and excludes RawKV keys. This prevents unintended placement
+		// policy application to raw keys that fall between meta ('m', 0x6d) and table ('t', 0x74) prefixes.
+		endKey = hex.EncodeToString(codec.EncodeBytes(nil, rawPrefix))
 	}
 	return startKey, endKey
 }
--- a/pkg/ddl/placement/bundle.go
+++ b/pkg/ddl/placement/bundle.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 
 	"github.com/pingcap/failpoint"
+	"github.com/pingcap/tidb/pkg/kv"
 	"github.com/pingcap/tidb/pkg/meta/model"
 	"github.com/pingcap/tidb/pkg/tablecodec"
 	"github.com/pingcap/tidb/pkg/util/codec"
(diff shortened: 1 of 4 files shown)


pkg/ddl/placement/bundle.go
```diff
@@ -463,7 +463,10 @@ func GetRangeStartAndEndKeyHex(rangeBundleID string) (startKey string, endKey st
 	if rangeBundleID == TiDBBundleRangePrefixForMeta {
 		// Use codec.EncodeBytes to properly encode the meta prefix in table mode
 		startKey = hex.EncodeToString(codec.EncodeBytes(nil, metaPrefix))
-		endKey = hex.EncodeToString(codec.EncodeBytes(nil, tablecodec.GenTablePrefix(0)))
+		// Use rawPrefix as endKey to ensure the meta range [metaPrefix, rawPrefix) only
+		// contains meta keys and excludes RawKV keys. This prevents unintended placement
+		// policy application to raw keys that fall between meta ('m', 0x6d) and table ('t', 0x74) prefixes.
+		endKey = hex.EncodeToString(codec.EncodeBytes(nil, rawPrefix))
```
nolouch: How about use `endKey = hex.EncodeToString(kv.Key(startKey).PrefixNext())`

lance6716: +1. In case we add another prefix between `m` and `r` in future.

Please also add a unit test for the change.

HaoW30: @nolouch @lance6716 
Updated, please take another look. 
And the change already covered by the unit test TestGetRangeStartAndEndKeyHex
Thanks!

nolouch (approved):

PR #63285, opened 2025-08-29
stats: optimize build topn and histogram

Tristan1900:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/statistics/sort_utils.go
```diff
@@ -0,0 +1,101 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package statistics
+
+import "container/heap"
+
+// TopNHeap implements a min-heap for maintaining top N items efficiently.
```
hawkingrei: This seems like a very common implementation. Will it be better to place this file separately under the util directory so that everyone can conveniently reuse it?

Tristan1900: good idea, actually there is another heap impl in the codebase but coupled with its application logic. 
Need some effort to merge the two.

Tristan1900:
some preliminary result
`go test -benchmem -run=^$ -bench ^BenchmarkBuildHistAndTopN$ github.com/pingcap/tidb/pkg/statistics -count=10`

previous implementation
```
goos: darwin
goarch: arm64
pkg: github.com/pingcap/tidb/pkg/statistics
cpu: Apple M3
                   │  old.txt   │
                   │   sec/op   │
BuildHistAndTopN-8   1.194 ± 1%

                   │   old.txt    │
                   │     B/op     │
BuildHistAndTopN-8   30.54Mi ± 0%

                   │   old.txt   │
                   │  allocs/op  │
BuildHistAndTopN-8   2.000M ± 0%
```
After improvement 
```
goos: darwin
goarch: arm64
pkg: github.com/pingcap/tidb/pkg/statistics
cpu: Apple M3
                   │   new.txt   │
                   │   sec/op    │
BuildHistAndTopN-8   63.99m ± 1%

                   │   new.txt    │
                   │     B/op     │
BuildHistAndTopN-8   15.36Mi ± 0%

                   │   new.txt   │
                   │  allocs/op  │
BuildHistAndTopN-8   1.001M ± 0%
```

Roughly 19x faster and 50% mem saved

Tristan1900:
run bench again after PR ready, confirmed the performance gain

```
goos: darwin
goarch: arm64
pkg: github.com/pingcap/tidb/pkg/statistics
cpu: Apple M3
                   │   new.txt   │
                   │   sec/op    │
BuildHistAndTopN-8   62.44m ± 3%

                   │   new.txt    │
                   │     B/op     │
BuildHistAndTopN-8   15.36Mi ± 0%

                   │   new.txt   │
                   │  allocs/op  │
BuildHistAndTopN-8   1.001M ± 0%
```

pkg/util/generic/bounded_min_heap.go
```diff
@@ -0,0 +1,104 @@
...
+// It keeps the N items with the highest values according to the comparison function.
+// The root of the heap is always the smallest item, making it easy to remove when adding better items.
+type BoundedMinHeap[T any] struct {
+	cmpFunc func(T, T) int
+	items   []T
+	maxSize int
+}
+
+// NewBoundedMinHeap creates a new bounded min-heap with the specified maximum size and comparison function.
+func NewBoundedMinHeap[T any](maxSize int, cmpFunc func(T, T) int) *BoundedMinHeap[T] {
```
0xPoe, line 59: Do we need to check the cmpFunc is not nil here?

pkg/util/generic/bounded_min_heap.go
```diff
@@ -0,0 +1,104 @@
...
+	item := old[n-1]
+	h.items = old[0 : n-1]
+	return item
+}
+
+// Add adds an item to the bounded min-heap. If the heap is full and the new item
+// is better than the worst item, it replaces the worst item.
+func (h *BoundedMinHeap[T]) Add(item T) {
+	// handle zero capacity case
+	if h.maxSize <= 0 {
```
0xPoe: Should we panic early if maxSize is less than or equal to 0?

pkg/statistics/builder.go
```diff
@@ -17,18 +17,95 @@ package statistics
...
 	"github.com/pingcap/tidb/pkg/util/codec"
 	"github.com/pingcap/tidb/pkg/util/collate"
+	"github.com/pingcap/tidb/pkg/util/generic"
 	"github.com/pingcap/tidb/pkg/util/memory"
-	"go.uber.org/zap"
 )
 
+const (
+	// defaultTopNValue is the default value for numTopN parameter
+	defaultTopNValue = 100
```
0xPoe: Thank you for extracting them as constants. However, I believe we placed these magic numbers somewhere else. Could you please try to reuse them?

pkg/statistics/builder.go
```diff
@@ -180,8 +277,16 @@ func buildHist(
 	}()
 
 	var upper = new(types.Datum)
-	// Note: Start from 1 because we have already processed the first sample.
-	for i := int64(1); i < sampleNum; i++ {
+	processedCount := int64(1) // we've processed the first sample
+
+	// Note: Start from firstSampleIdx+1 because we have already processed the first non-skipped sample.
```
0xPoe:
```suggestion
	// Start from firstSampleIdx + 1 since the first non-skipped sample has already been processed when the range checker is not null.
```

pkg/statistics/builder.go
```diff
@@ -143,16 +220,18 @@ func BuildColumnHist(ctx sessionctx.Context, numBuckets, id int64, collector *Sa
...
 	sc *stmtctx.StatementContext,
 	hg *Histogram,
 	samples []*SampleItem,
 	count, ndv, numBuckets int64,
 	memTracker *memory.Tracker,
+	sampleCountExcludeTopN int64,
+	rangeChecker *SequentialRangeChecker, // optional range checker for skipping TopN indices
 ) (corrXYSum float64, err error) {
 	sampleNum := int64(len(samples))
-	// As we use samples to build the histogram, the bucket number and repeat should multiply a factor.
```
0xPoe, line 154: I guess we don't need to delete this comment?

pkg/statistics/builder.go
```diff
@@ -316,22 +422,29 @@ func BuildHistAndTopN(
 
 	sampleNum := int64(len(samples))
 	// As we use samples to build the histogram, the bucket number and repeat should multiply a factor.
-	sampleFactor := float64(count) / float64(len(samples))
-	// If a numTopn value other than 100 is passed in, we assume it's a value that the user wants us to honor
+	sampleFactor := float64(count) / float64(sampleNum)
+	// If a numTopn value other than default is passed in, we assume it's a value that the user wants us to honor
 	allowPruning := true
-	if numTopN != 100 {
+	if numTopN != defaultTopNValue {
```
0xPoe: 👍🏼 Love this const, I hate magic numbers!

pkg/statistics/builder.go
0xPoe:
```go
// TopN reduced from 500 to 100 due to concerns over large number of TopN values collected for customers with many tables.
// 100 is more inline with other databases. 100-256 is also common for NumBuckets with other databases.
var analyzeOptionDefaultV2 = map[ast.AnalyzeOptionType]uint64{
	ast.AnalyzeOptNumBuckets:    256,
	ast.AnalyzeOptNumTopN:       100,
	ast.AnalyzeOptCMSketchWidth: 2048,
	ast.AnalyzeOptCMSketchDepth: 5,
	ast.AnalyzeOptNumSamples:    0,
	ast.AnalyzeOptSampleRate:    math.Float64bits(-1),
}
```

terry1purcell: In a separate PR - we should make them variables such that customers can adjust them if needed.

Tristan1900: Created an issue to track https://github.com/pingcap/tidb/issues/63432

Tristan1900:
/hold

one test fails for legit reason

Tristan1900:
/retest

actually a flaky test that happened before this PR

pkg/statistics/builder.go
0xPoe: It seems this comment hasn’t been restored.

pkg/util/generic/bounded_min_heap.go
```diff
@@ -0,0 +1,111 @@
...
+		return nil
+	}
+
+	// copy items to avoid modifying the original heap
+	result := make([]T, len(h.items))
+	copy(result, h.items)
+
+	sort.Slice(result, func(i, j int) bool {
+		return h.cmpFunc(result[i], result[j]) > 0
+	})
```
time-and-fate: 1. `slices.SortFunc` is a newer replacement for it. In most cases, we should use the `slices` package instead of the `sort` package.
2. I think using `-h.cmpFunc(result[i], result[j])` would be more explicit that the order here is reversed.

pkg/util/generic/bounded_min_heap.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// Pop removes and returns the smallest item from the heap.
+func (h *BoundedMinHeap[T]) Pop() any {
+	old := h.items
+	n := len(old)
+	item := old[n-1]
+	h.items = old[0 : n-1]
+	return item
+}
```
time-and-fate: I think we need to reconsider the API design of this struct.
These methods are not expected to be used externally and should be unexported. I think the heap from the std lib should be an internal data type.

pkg/util/generic/bounded_min_heap.go
```diff
@@ -0,0 +1,111 @@
...
+package generic
+
+import (
+	"container/heap"
+	"sort"
+)
+
+// BoundedMinHeap implements a min-heap for maintaining the best N items efficiently.
+// It keeps the N items with the highest values according to the comparison function.
+// The root of the heap is always the smallest item, making it easy to remove when adding better items.
```
time-and-fate: I think we need to unify the terminology: best/worst or high/low.

pkg/statistics/builder.go
```diff
@@ -316,22 +421,29 @@ func BuildHistAndTopN(
...
 
-	// Step1: collect topn from samples
+	// Step1: collect topn from samples using bounded min-heap and track their index ranges
+	boundedMinHeap := generic.NewBoundedMinHeap(numTopN, func(a, b TopNWithRange) int {
+		if a.Count < b.Count {
+			return -1 // min-heap: smaller counts at root
+		} else if a.Count > b.Count {
+			return 1
+		}
+		return 0
```
time-and-fate: `cmp.Compare(a.Count, b.Count)`

pkg/util/generic/bounded_min_heap.go
Tristan1900: yes that's heap's interface we must comply. https://pkg.go.dev/container/heap
I think that's the standard lib heap, and I looked at other impl in the code base and they all follow the same pattern. Do you mind sharing which std lib you are referring to?

time-and-fate: Yes. I understand that you implemented `BoundedMinHeap` based on the `container/heap` from the std lib.

I mean the `container/heap` here should be an internal data type and should not be accessed externally. For example, you don't want other people to do something like `NewBoundedMinHeap().Push(val)`.

What I have in mind is that the implementation should be something like:
```
type internalHeap[T any] struct {
	....
}

func (h *internalHeap[T]) Len() int {...}

func (h *internalHeap[T]) Less(i, j int) bool {...}

func (h *internalHeap[T]) Swap(i, j int) {...}

func (h *internalHeap[T]) Push(x any) {...}

func (h *internalHeap[T]) Pop() any {...}

type BoundedMinHeap[T any] struct {
	data *internalHeap[T]
	...
}

func (h *BoundedMinHeap[T]) Add(item T) {...}


func (h *BoundedMinHeap[T]) ToSortedSlice() []T {...}

```

Tristan1900: oh I see the idea. Yeah that's a great suggestion. Let me do that

PR #63319, opened 2025-09-02
distsql: Add `SelectResultIter` to read rows from cop-task in the iterator way

### LINKED ISSUE #63318: Make `SelectResult` to read multiple outputs from one dist task request.

lcwangchao:

Issue Number: close #63318

Problem Summary:

After https://github.com/pingcap/tipb/pull/376 is merged, the cop-task response may contain multiple output channels (i.e., the push down IndexLookup executor), and each of these channels may have a different kind of schema.

`SelectResult` should support reading the results from multiple channels. 

1. A new interface `SelectResultIter` is provided.
2. `SelectResultIter` is defined as:

```go
// SelectResultIter is an iterator that is used to iterate the rows from SelectResult.
type SelectResultIter interface {
	// Next returns the next row.
	// If the iterator is drained, the `SelectResultRow.IsEmpty()` returns true.
	Next(ctx context.Context) (SelectResultRow, error)
	// Close closes the iterator.
	Close() error
}
```

The `Next` is used to read the rows from the cop-task. `SelectResultRow` contains an extra field `Channel` to indicate where the row is from:

```
// SelectResultRow indicates the row returned by the SelectResultIter
type SelectResultRow struct {
	// `Channel` indicates the index where this row locates.
	// When Channel < len(IntermediateChannels), it means this row is from intermediate result.
	// Otherwise, if Channel == len(IntermediateChannels), it means this row is from final result.
	Channel int
	// Row is the actual data of this row.
	chunk.Row
}
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/distsql/select_result.go
+++ b/pkg/distsql/select_result.go
@@ -82,10 +82,10 @@ type SelectResult interface {
 
 // SelectResultRow indicates the row returned by the SelectResultIter
 type SelectResultRow struct {
-	// `Channel` indicates the index where this row locates.
-	// When Channel < len(IntermediateChannels), it means this row is from intermediate result.
-	// Otherwise, if Channel == len(IntermediateChannels), it means this row is from final result.
-	Channel int
+	// `ChannelIndex` indicates the index where this row locates.
+	// When ChannelIndex < len(IntermediateChannels), it means this row is from intermediate result.
+	// Otherwise, if ChannelIndex == len(IntermediateChannels), it means this row is from final result.
+	ChannelIndex int
 	// Row is the actual data of this row.
 	chunk.Row
 }
@@ -825,8 +825,8 @@ func (iter *selRespChannelIter) Next() (SelectResultRow, error) {
 	if iter.chk != nil && iter.chkOffset < iter.chk.NumRows() {
 		iter.chkOffset++
 		return SelectResultRow{
-			Channel: iter.channel,
-			Row:     iter.chk.GetRow(iter.chkOffset - 1),
+			ChannelIndex: iter.channel,
+			Row:          iter.chk.GetRow(iter.chkOffset - 1),
 		}, nil
 	}
 
@@ -836,8 +836,8 @@ func (iter *selRespChannelIter) Next() (SelectResultRow, error) {
 
 	iter.chkOffset = 1
 	return SelectResultRow{
-		Channel: iter.channel,
-		Row:     iter.chk.GetRow(0),
+		ChannelIndex: iter.channel,
+		Row:          iter.chk.GetRow(0),
 	}, nil
 }
 
--- a/pkg/distsql/select_result_test.go
+++ b/pkg/distsql/select_result_test.go
@@ -257,7 +257,7 @@ func TestSelRespChannelIterRead(t *testing.T) {
 				require.True(t, row.IsEmpty())
 			} else {
 				require.False(t, row.IsEmpty())
-				require.Equal(t, 1, row.Channel)
+				require.Equal(t, 1, row.ChannelIndex)
 				strVal := row.GetString(0)
 				intVal := row.GetInt64(1)
 				tmVal, err := row.GetTime(2).GoTime(loc)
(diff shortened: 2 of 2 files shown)


pkg/distsql/select_result.go
```diff
@@ -381,6 +415,16 @@ func (r *selectResult) fetchResp(ctx context.Context) error {
 		if err != nil {
 			return errors.Trace(err)
 		}
+
+		if len(r.selectResp.IntermediateOutputs) != len(intermediateOutputTypes) {
+			return errors.Errorf(
```
cfzjywxk, line 424: Does it indicates inconsistency? If so the `inconsistencyReporter` could be used to log necessary information.

lcwangchao: This situation should never happen, but it has no relationship with the data consistency. If it happens, it is more like the TiDB or TiKV code has a bug that makes the wrong assumption of the data channels.

pkg/distsql/select_result.go
```diff
@@ -72,6 +74,27 @@ type SelectResult interface {
 	NextRaw(context.Context) ([]byte, error)
 	// Next reads the data into chunk.
 	Next(context.Context, *chunk.Chunk) error
+	// IntoIter converts the SelectResult into an iterator.
+	IntoIter([][]*types.FieldType) (SelectResultIter, error)
```
ekexium, line 78: The method "consumes" the result, but there is no mechanism to prevent others accessing the original result again. I would suggest add a validity check or something

pkg/distsql/select_result.go
```diff
@@ -683,6 +763,199 @@ func (r *selectResult) Close() error {
...
+	// curChkIdx indicates the index of the current chunk in chkData read currently.
+	curChkIdx int
+	// chk buffers the rows read from the current response
+	chk *chunk.Chunk
+	// offset indicates the read offset in iter.chk
+	chkOffset int
+}
+
+func newSelRespChannelIter(result *selectResult, channel int) (*selRespChannelIter, error) {
+	intest.Assert(result != nil && result.selectResp != nil && len(result.selectResp.IntermediateOutputs) == len(result.intermediateOutputTypes))
```
ekexium, line 785: Should we return an error in production?

pkg/distsql/select_result.go
lcwangchao: There is already a check: https://github.com/pingcap/tidb/pull/63319/files#diff-f45688fe21dcc5e4a40e356dc44d0fe3622a8557ac3f8e58ce6d86d3efc0ed43R474-R478

An error `selectResult is invalid after IntoIter()` will be returned if an iter created but the original methods are called then

pkg/distsql/select_result.go
lcwangchao: This is an internal call, and some conditions have been checked outside: https://github.com/pingcap/tidb/pull/63319/files#diff-f45688fe21dcc5e4a40e356dc44d0fe3622a8557ac3f8e58ce6d86d3efc0ed43R425

Seems it's more simple to just use an assert

pkg/distsql/select_result.go
```diff
@@ -72,6 +74,27 @@ type SelectResult interface {
...
+	// Close closes the iterator.
+	Close() error
+}
+
+// SelectResultRow indicates the row returned by the SelectResultIter
+type SelectResultRow struct {
+	// `Channel` indicates the index where this row locates.
+	// When Channel < len(IntermediateChannels), it means this row is from intermediate result.
+	// Otherwise, if Channel == len(IntermediateChannels), it means this row is from final result.
+	Channel int
```
cfzjywxk:
```suggestion
	ChannelIndex int
```
Seems better.

pkg/distsql/select_result.go
```diff
@@ -409,11 +457,25 @@ func (r *selectResult) fetchResp(ctx context.Context) error {
...
 			break
 		}
+
+		if intermediate := r.selectResp.IntermediateOutputs; len(intermediate) > 0 {
+			for _, output := range intermediate {
+				if len(output.Chunks) != 0 {
+					// some intermediate output contains data,
+					// we should return the response even if the main output is empty.
+					return nil
+				}
```
cfzjywxk, line 467: Is it expected to contain zero lengh chunk in the `IntermediateOutputs` results? If not should we add assertions here too?

lcwangchao: Do you mean the length of `output.Chunks`? Zero length is expected if all rows can be looked up on the TikV side.

cfzjywxk: I mean `len(intermediate) > 0` but all the inner `output chunks` contain zero length, is this expected?

cfzjywxk (approved):

pkg/distsql/select_result.go
lcwangchao: Yes, it is expected. The response always have `len(resp.IntermediateOutputs) == len(req.IntermediateOutputChannels)`

ekexium (approved):

Issue #63336, opened 2025-09-02
planner: a new risk-aware index selection heuristic rule for IndexLookup in skyline pruning


qw4990:
## Enhancement
Many wrong index selections are the following case:
```
create table t (a int, b int, c int, d int, e int, f int, key abcd(a, b, c, d), key dc(d, c));

mysql> explain select * from t use index(dc) where a=1 and d=1 and c>1;
+-------------------------------+---------+-----------+-------------------------+----------------------------------------------------+
| id                            | estRows | task      | access object           | operator info                                      |
+-------------------------------+---------+-----------+-------------------------+----------------------------------------------------+
| IndexLookUp_9                 | 1.00    | root      |                         |                                                    |
| ├─IndexRangeScan_6(Build)     | 5.00    | cop[tikv] | table:t, index:dc(d, c) | range:(1 1,1 +inf], keep order:false, stats:pseudo |
| └─Selection_8(Probe)          | 1.00    | cop[tikv] |                         | eq(test.t.a, 1)                                    |
|   └─TableRowIDScan_7          | 5.00    | cop[tikv] | table:t                 | keep order:false, stats:pseudo                     |
+-------------------------------+---------+-----------+-------------------------+----------------------------------------------------+

mysql> explain select * from t where a=1 and d=1 and c>1;
+-------------------------------+---------+-----------+---------------------------------+---------------------------------------------+
| id                            | estRows | task      | access object                   | operator info                               |
+-------------------------------+---------+-----------+---------------------------------+---------------------------------------------+
| IndexLookUp_9                 | 1.00    | root      |                                 |                                             |
| ├─Selection_8(Build)          | 1.00    | cop[tikv] |                                 | eq(test.t.d, 1), gt(test.t.c, 1)            |
| │ └─IndexRangeScan_6          | 10.00   | cop[tikv] | table:t, index:abcd(a, b, c, d) | range:[1,1], keep order:false, stats:pseudo |
| └─TableRowIDScan_7(Probe)     | 1.00    | cop[tikv] | table:t                         | keep order:false, stats:pseudo              |
+-------------------------------+---------+-----------+---------------------------------+---------------------------------------------+
```

In the case above, the optimizer is prone to select `dc` since its estimated rows is slightly smaller than `abcd`.
[...]

PR #63349, opened 2025-09-02
planner: refactor compareCandidates logic | tidb-test=pr/2594

terry1purcell:

Issue Number: ref #63336 

Problem Summary:

The following changes to the compareCandidates logic:
1. Factor out pseudo index comparison logic to simplify compareCandidates.
2. Only fail logic if both comparable1 && comparable2 are false - since comparable2 indicates that the index covers more predicates.
3. Fix the following bugs:
- Fix for DNF checking in isFullMatch.
- Fix for DNF equal predicate checking to include nested AND under OR.
After these bugs are fixed, it will be easier to focus on the issue referenced above.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > - [ ] Refactor of code with no testcase changes.

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/find_best_task.go
+++ b/pkg/planner/core/find_best_task.go
@@ -1145,10 +1145,14 @@ func compareCandidates(sctx base.PlanContext, statsTbl *statistics.Table, tableI
 	// eqOrInResult: comparison result of equal/IN predicate coverage (1=LHS better, -1=RHS better, 0=equal)
 	eqOrInResult, lhsEqOrInCount, rhsEqOrInCount := compareEqOrIn(lhs, rhs)
 	tableFilterResult := 0
-	if len(lhs.path.TableFilters) < len(rhs.path.TableFilters) {
-		tableFilterResult = 1
-	} else if len(lhs.path.TableFilters) > len(rhs.path.TableFilters) {
-		tableFilterResult = -1
+	// Count table filters - lowest wins. Do not count table filters if this is a table
+	// path, since that will inadvertently discourage a table path.
+	if !lhs.path.IsTablePath() && !rhs.path.IsTablePath() {
+		if len(lhs.path.TableFilters) < len(rhs.path.TableFilters) {
+			tableFilterResult = 1
+		} else if len(lhs.path.TableFilters) > len(rhs.path.TableFilters) {
+			tableFilterResult = -1
+		}
 	}
 
 	// indexSum is the aggregate score of index predicate comparison metrics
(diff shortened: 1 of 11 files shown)


pkg/planner/core/find_best_task.go
```diff
@@ -1142,39 +1144,39 @@ func compareCandidates(sctx base.PlanContext, statsTbl *statistics.Table, tableI
...
+	if !lhs.path.IsTablePath() && !rhs.path.IsTablePath() {
+		if len(lhs.path.TableFilters) < len(rhs.path.TableFilters) {
+			tableFilterResult = 1
+		} else if len(lhs.path.TableFilters) > len(rhs.path.TableFilters) {
+			tableFilterResult = -1
+		}
+	}
+	// Count the total number of filters for each path. We don't want to prioritize plans
+	// that only have 1 predicate. They should compete on cost.
+	lhsTotalIndexFilters := max(len(lhs.path.AccessConds), lhsEqOrInCount) + len(lhs.path.IndexFilters)
```
qw4990: Why consider `lhsEqOrInCount` here? `len(AccessConds)` is not enough? `lhsTotalIndexFilters := len(lhs.path.AccessConds) + len(lhs.path.IndexFilters)`

pkg/planner/core/find_best_task.go
```diff
@@ -1142,39 +1144,39 @@ func compareCandidates(sctx base.PlanContext, statsTbl *statistics.Table, tableI
...
+	rhsTotalIndexFilters := max(len(rhs.path.AccessConds), rhsEqOrInCount) + len(rhs.path.IndexFilters)
```
qw4990: `path with less table filters is the winner` seems a little bit confusing, since the implicit assumption is that these 2 paths have the same number of `index filters + table filters`, could we just compare their total index filters, `path with more index filters is the winner`.
`idnexFilterResult = lhsTotalIndexFilters > rhsTotalIndexFilters `, which should be equal to `tableFitlerResult`.

pkg/planner/core/find_best_task.go
terry1purcell: I have removed this, because I added it incorrectly after finding an obscure result in how predicates were counted for DNF - but that turned out to be a bug in a called function.

pkg/planner/core/find_best_task.go
terry1purcell: I removed this - similar explanation to the index filters topic - these additional checks were added due to results that didn't make sense. That ended up being a bug in a called function for DNF predicates.

pkg/planner/core/find_best_task.go
```diff
@@ -1197,24 +1185,13 @@ func compareCandidates(sctx base.PlanContext, statsTbl *statistics.Table, tableI
...
-	// Below compares the two candidate paths on four dimensions:
-	// (1): the set of columns that occurred in the access condition,
-	// (2): does it require a double scan,
-	// (3): whether or not it matches the physical property,
-	// (4): it's a global index path or not.
-	// (5): whether the path has more equal/IN predicates.
-	// If `x` is not worse than `y` at all factors,
-	// and there exists one factor that `x` is better than `y`, then `x` is better than `y`.
-	if !comparable1 {
+	if !comparable1 && !comparable2 {
```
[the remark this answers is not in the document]
terry1purcell: Theoretically - copilot is correct. But in reality, this is a restriction that reduces our ability to compare the other criteria. Because comparable2 is more beneficial than comparable1. So once we add other criteria such as eqOrInCount - we can identify an index as a stronger candidate that comparable2 can qualify, but comparable1 will disqualify.

pkg/planner/core/find_best_task.go
```diff
@@ -1685,7 +1687,7 @@ func (c *candidatePath) hasOnlyEqualPredicatesInDNF() bool {
 			return false
 		}
 
-		if sf.FuncName.L == ast.LogicOr {
+		if sf.FuncName.L == ast.LogicOr || sf.FuncName.L == ast.LogicAnd {
```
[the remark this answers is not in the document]
terry1purcell: Because this is a recursive call - then the child of an OR may be an AND. I added this check for LogicAnd because it was incorrectly rejecting "AND underneath OR". Thus, this recommendation is incorrect.

hawkingrei (approved):

PR #63409, opened 2025-09-08
bindinfo: add last_used_date to track bindinfo usage frequency

### LINKED ISSUE #63407: Add last_used_time column to mysql.bind_info system table to track usage frequency
## Enhancement

Currently, the mysql.bind_info table stores various bindings created by customers. To better manage resources and optimize system performance, we need to distinguish between frequently used and rarely used bindings. This information is crucial for potential cleanup tasks or understanding usage patterns.

We propose adding a new column, last_used_time, to the mysql.bind_info system table. This column will record the last time a specific binding was accessed or used by a query.

hawkingrei:

Issue Number: close #63407 

Problem Summary:

Currently, some customers have created a large number of bindings. However, it is difficult to determine whether these bindings are still in use. The sheer volume of bindings also puts pressure on TiDB. Therefore, we need a way to identify and mark the bindings that are not in use.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test

start a tidb with master which is started by script. ```Tikv``` and ```PD``` is started by ```tiup playground nightly --mode tikv-slim```.

```
+----------------------------------+----------------------------------+------------+---------+-------------------------+-------------------------+---------+-----------+---------+------------+-------------+
| original_sql                     | bind_sql                         | default_db | status  | create_time             | update_time             | charset | collation | source  | sql_digest | plan_digest |
+----------------------------------+----------------------------------+------------+---------+-------------------------+-------------------------+---------+-----------+---------+------------+-------------+
| builtin_pseudo_sql_for_bind_lock | builtin_pseudo_sql_for_bind_lock | mysql      | builtin | 0000-00-00 00:00:00.000 | 0000-00-00 00:00:00.000 |         |           | builtin | <null>     | <null>      |
+----------------------------------+----------------------------------+------------+---------+-------------------------+-------------------------+---------+-----------+---------+------------+-------------+
```

kill this TiDB and start changed tidb verson.

```
+----------------------------------+----------------------------------+------------+---------+-------------------------+-------------------------+---------+-----------+---------+------------+-------------+----------------+
| original_sql                     | bind_sql                         | default_db | status  | create_time             | update_time             | charset | collation | source  | sql_digest | plan_digest | last_used_time |
+----------------------------------+----------------------------------+------------+---------+-------------------------+-------------------------+---------+-----------+---------+------------+-------------+----------------+
| builtin_pseudo_sql_for_bind_lock | builtin_pseudo_sql_for_bind_lock | mysql      | builtin | 0000-00-00 00:00:00.000 | 0000-00-00 00:00:00.000 |         |           | builtin | <null>     | <null>      | <null>         |
+----------------------------------+----------------------------------+------------+---------+-------------------------+-------------------------+---------+-----------+---------+------------+-------------+----------------+
```

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/bindinfo/binding.go
+++ b/pkg/bindinfo/binding.go
@@ -100,6 +100,7 @@ func (b *Binding) UpdateUsageInfo() {
 	b.UsageInfo.Update()
 }
 
+// ResetUsageInfo is to reset the bindinfo useage info after writing into storage.
 func (b *Binding) ResetUsageInfo() {
 	b.UsageInfo.CreateAt.Store(nil)
 	b.UsageInfo.LastUsedAt.Store(nil)
--- a/pkg/bindinfo/binding.go
+++ b/pkg/bindinfo/binding.go
@@ -80,7 +80,8 @@ type Binding struct {
 	// TableNames records all schema and table names in this binding statement, which are used for cross-db matching.
 	TableNames []*ast.TableName `json:"-"`
 
-	// usageInfo records the binding usage info
+	// UsageInfo is to track the usage information `last_used_time` of this binding
+	// and it will be updated when this binding is used.
 	UsageInfo bindingInfoUsageInfo
 }
 
@@ -100,21 +101,27 @@ func (b *Binding) UpdateUsageInfo() {
 	b.UsageInfo.Update()
 }
 
-// ResetUsageInfo is to reset the bindinfo usage info after writing into storage.
-func (b *Binding) ResetUsageInfo() {
-	b.UsageInfo.CreateAt.Store(nil)
-	b.UsageInfo.LastUsedAt.Store(nil)
+// UpdateSavedAt is to update the last saved time
+func (b *Binding) UpdateSavedAt(ts *time.Time) {
+	b.UsageInfo.LastSavedAt.Store(ts)
 }
 
 type bindingInfoUsageInfo struct {
+	// LastUsedAt records the last time when this binding is used.
+	// It is nil if this binding has never been used or has been reset after writing into storage.
+	// It is updated when this binding is used.
+	// It is used to update the `last_used_time` field in mysql.bind_info table.
 	LastUsedAt atomic.Pointer[time.Time]
-	CreateAt   atomic.Pointer[time.Time]
+	// LastSavedAt records the last time when this binding is saved into storage.
+	LastSavedAt atomic.Pointer[time.Time]
 }
 
 func (b *bindingInfoUsageInfo) Update() {
 	now := time.Now()
-	if b.CreateAt.Load() == nil {
-		b.CreateAt.Store(&now)
+	if b.LastSavedAt.Load() == nil {
+		// If `LastSavedAt`  is equal to `nil`, we can consider it as a record that has not been written.
+		// Thus, from the first read, we take this as its saved time. After that, it will be written after a while.
+		b.LastSavedAt.Store(&now)
 	}
 	b.LastUsedAt.Store(&now)
 }
--- a/pkg/bindinfo/binding.go
+++ b/pkg/bindinfo/binding.go
@@ -96,9 +96,9 @@ func (b *Binding) size() float64 {
 	return float64(res)
 }
 
-// UpdateUsageInfo is to update binding usage info
-func (b *Binding) UpdateUsageInfo() {
-	b.UsageInfo.Update()
+// UpdateLastUsedAt is to update binding usage info when this binding is used.
+func (b *Binding) UpdateLastUsedAt() {
+	b.UsageInfo.UpdateLastUsedAt()
 }
 
 // UpdateSavedAt is to update the last saved time
@@ -116,13 +116,8 @@ type bindingInfoUsageInfo struct {
 	LastSavedAt atomic.Pointer[time.Time]
 }
 
-func (b *bindingInfoUsageInfo) Update() {
+func (b *bindingInfoUsageInfo) UpdateLastUsedAt() {
 	now := time.Now()
-	if b.LastSavedAt.Load() == nil {
-		// If `LastSavedAt`  is equal to `nil`, we can consider it as a record that has not been written.
-		// Thus, from the first read, we take this as its saved time. After that, it will be written after a while.
-		b.LastSavedAt.Store(&now)
-	}
 	b.LastUsedAt.Store(&now)
 }
 
@@ -189,7 +184,7 @@ func matchSQLBinding(sctx sessionctx.Context, stmtNode ast.StmtNode, info *Bindi
 	binding, matched = globalHandle.MatchingBinding(sctx, noDBDigest, tableNames)
 	if matched {
 		// After hitting the cache, update the usage time of the bind.
-		binding.UpdateUsageInfo()
+		binding.UpdateLastUsedAt()
 		return binding, matched, metrics.ScopeGlobal
 	}
 
--- a/pkg/bindinfo/binding.go
+++ b/pkg/bindinfo/binding.go
@@ -100,7 +100,7 @@ func (b *Binding) UpdateUsageInfo() {
 	b.UsageInfo.Update()
 }
 
-// ResetUsageInfo is to reset the bindinfo useage info after writing into storage.
+// ResetUsageInfo is to reset the bindinfo usage info after writing into storage.
 func (b *Binding) ResetUsageInfo() {
 	b.UsageInfo.CreateAt.Store(nil)
 	b.UsageInfo.LastUsedAt.Store(nil)
--- a/pkg/bindinfo/binding.go
+++ b/pkg/bindinfo/binding.go
@@ -98,11 +98,12 @@ func (b *Binding) size() float64 {
 
 // UpdateLastUsedAt is to update binding usage info when this binding is used.
 func (b *Binding) UpdateLastUsedAt() {
-	b.UsageInfo.UpdateLastUsedAt()
+	now := time.Now()
+	b.UsageInfo.LastUsedAt.Store(&now)
 }
 
-// UpdateSavedAt is to update the last saved time
-func (b *Binding) UpdateSavedAt(ts *time.Time) {
+// UpdateLastSavedAt is to update the last saved time
+func (b *Binding) UpdateLastSavedAt(ts *time.Time) {
 	b.UsageInfo.LastSavedAt.Store(ts)
 }
 
@@ -114,11 +115,6 @@ type bindingInfoUsageInfo struct {
 	LastUsedAt atomic.Pointer[time.Time]
 	// LastSavedAt records the last time when this binding is saved into storage.
 	LastSavedAt atomic.Pointer[time.Time]
-}
-
-func (b *bindingInfoUsageInfo) UpdateLastUsedAt() {
-	now := time.Now()
-	b.LastUsedAt.Store(&now)
 }
 
 var (
--- a/pkg/bindinfo/binding.go
+++ b/pkg/bindinfo/binding.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"sync/atomic"
+	"time"
 	"unsafe"
 
 	"github.com/pingcap/tidb/pkg/metrics"
@@ -77,6 +79,9 @@ type Binding struct {
 
 	// TableNames records all schema and table names in this binding statement, which are used for cross-db matching.
 	TableNames []*ast.TableName `json:"-"`
+
+	// usageInfo records the binding usage info
+	UsageInfo bindingInfoUsageInfo
 }
 
 // IsBindingEnabled returns whether the binding is enabled.
@@ -88,6 +93,29 @@ func (b *Binding) IsBindingEnabled() bool {
 func (b *Binding) size() float64 {
 	res := len(b.OriginalSQL) + len(b.Db) + len(b.BindSQL) + len(b.Status) + 2*int(unsafe.Sizeof(b.CreateTime)) + len(b.Charset) + len(b.Collation) + len(b.ID)
 	return float64(res)
+}
+
+// UpdateUsageInfo is to update binding usage info
+func (b *Binding) UpdateUsageInfo() {
+	b.UsageInfo.Update()
+}
+
+func (b *Binding) ResetUsageInfo() {
+	b.UsageInfo.CreateAt.Store(nil)
+	b.UsageInfo.LastUsedAt.Store(nil)
+}
+
+type bindingInfoUsageInfo struct {
+	LastUsedAt atomic.Pointer[time.Time]
+	CreateAt   atomic.Pointer[time.Time]
+}
+
+func (b *bindingInfoUsageInfo) Update() {
+	now := time.Now()
+	if b.CreateAt.Load() == nil {
+		b.CreateAt.Store(&now)
+	}
+	b.LastUsedAt.Store(&now)
 }
 
 var (
@@ -152,6 +180,8 @@ func matchSQLBinding(sctx sessionctx.Context, stmtNode ast.StmtNode, info *Bindi
 	}
 	binding, matched = globalHandle.MatchingBinding(sctx, noDBDigest, tableNames)
 	if matched {
+		// After hitting the cache, update the usage time of the bind.
+		binding.UpdateUsageInfo()
 		return binding, matched, metrics.ScopeGlobal
 	}
 
--- a/pkg/bindinfo/binding.go
+++ b/pkg/bindinfo/binding.go
@@ -109,7 +109,7 @@ func (b *Binding) UpdateLastSavedAt(ts *time.Time) {
 
 type bindingInfoUsageInfo struct {
 	// LastUsedAt records the last time when this binding is used.
-	// It is nil if this binding has never been used or has been reset after writing into storage.
+	// It is nil if this binding has never been used.
 	// It is updated when this binding is used.
 	// It is used to update the `last_used_time` field in mysql.bind_info table.
 	LastUsedAt atomic.Pointer[time.Time]
--- a/pkg/bindinfo/binding_cache.go
+++ b/pkg/bindinfo/binding_cache.go
@@ -38,6 +38,9 @@ type BindingCacheUpdater interface {
 	// LoadFromStorageToCache loads global bindings from storage to the memory cache.
 	LoadFromStorageToCache(fullLoad bool) (err error)
 
+	// UpdateBindingUsageInfoToStorage is to update the binding usage info into storage
+	UpdateBindingUsageInfoToStorage()
+
 	// LastUpdateTime returns the last update time.
 	LastUpdateTime() types.Time
 }
@@ -94,6 +97,15 @@ func (u *bindingCacheUpdater) LoadFromStorageToCache(fullLoad bool) (err error) 
 	metrics.BindingCacheMemLimit.Set(float64(u.GetMemCapacity()))
 	metrics.BindingCacheNumBindings.Set(float64(len(u.GetAllBindings())))
 	return nil
+}
+
+// UpdateBindingUsageInfoToStorage is to update the binding usage info into storage
+func (u *bindingCacheUpdater) UpdateBindingUsageInfoToStorage() {
+	bindings := u.GetAllBindings()
+	err := UpdateBindingUsageInfoToStorage(u.sPool, bindings)
+	if err != nil {
+		bindingLogger().Warn("BindingHandle.UpdateBindingUsageInfoToStorage", zap.Error(err))
+	}
 }
 
 // LastUpdateTime returns the last update time.
--- a/pkg/bindinfo/binding_cache.go
+++ b/pkg/bindinfo/binding_cache.go
@@ -38,7 +38,7 @@ type BindingCacheUpdater interface {
 	// LoadFromStorageToCache loads global bindings from storage to the memory cache.
 	LoadFromStorageToCache(fullLoad bool) (err error)
 
-	// UpdateBindingUsageInfoToStorage is to update the binding usage info into storage
+	// updateBindingUsageInfoToStorage is to update the binding usage info into storage
 	UpdateBindingUsageInfoToStorage()
 
 	// LastUpdateTime returns the last update time.
@@ -99,10 +99,10 @@ func (u *bindingCacheUpdater) LoadFromStorageToCache(fullLoad bool) (err error) 
 	return nil
 }
 
-// UpdateBindingUsageInfoToStorage is to update the binding usage info into storage
+// updateBindingUsageInfoToStorage is to update the binding usage info into storage
 func (u *bindingCacheUpdater) UpdateBindingUsageInfoToStorage() {
 	bindings := u.GetAllBindings()
-	err := UpdateBindingUsageInfoToStorage(u.sPool, bindings)
+	err := updateBindingUsageInfoToStorage(u.sPool, bindings)
 	if err != nil {
 		bindingLogger().Warn("BindingHandle.UpdateBindingUsageInfoToStorage", zap.Error(err))
 	}
--- a/pkg/bindinfo/binding_cache.go
+++ b/pkg/bindinfo/binding_cache.go
@@ -38,8 +38,8 @@ type BindingCacheUpdater interface {
 	// LoadFromStorageToCache loads global bindings from storage to the memory cache.
 	LoadFromStorageToCache(fullLoad bool) (err error)
 
-	// updateBindingUsageInfoToStorage is to update the binding usage info into storage
-	UpdateBindingUsageInfoToStorage()
+	// UpdateBindingUsageInfoToStorage is to update the binding usage info into storage
+	UpdateBindingUsageInfoToStorage() error
 
 	// LastUpdateTime returns the last update time.
 	LastUpdateTime() types.Time
@@ -99,13 +99,15 @@ func (u *bindingCacheUpdater) LoadFromStorageToCache(fullLoad bool) (err error) 
 	return nil
 }
 
-// updateBindingUsageInfoToStorage is to update the binding usage info into storage
-func (u *bindingCacheUpdater) UpdateBindingUsageInfoToStorage() {
+// UpdateBindingUsageInfoToStorage is to update the binding usage info into storage
+func (u *bindingCacheUpdater) UpdateBindingUsageInfoToStorage() error {
+	defer func() {
+		if r := recover(); r != nil {
+			bindingLogger().Warn("panic when update usage info for binding", zap.Any("recover", r))
+		}
+	}()
 	bindings := u.GetAllBindings()
-	err := updateBindingUsageInfoToStorage(u.sPool, bindings)
-	if err != nil {
-		bindingLogger().Warn("BindingHandle.UpdateBindingUsageInfoToStorage", zap.Error(err))
-	}
+	return updateBindingUsageInfoToStorage(u.sPool, bindings)
 }
 
 // LastUpdateTime returns the last update time.
--- a/pkg/bindinfo/tests/bind_test.go
+++ b/pkg/bindinfo/tests/bind_test.go
@@ -44,7 +44,6 @@ func utilCleanBindingEnv(tk *testkit.TestKit) {
 func TestPrepareCacheWithBinding(t *testing.T) {
 	store := testkit.CreateMockStore(t)
 	tk := testkit.NewTestKit(t, store)
-	tk.MustExec(`set tidb_enable_prepared_plan_cache=1`)
 	tk.MustExec("use test")
 	tk.MustExec("drop table if exists t1, t2")
 	tk.MustExec("create table t1(a int, b int, c int, key idx_b(b), key idx_c(c))")
--- a/pkg/bindinfo/tests/bind_usage_info_test.go
+++ b/pkg/bindinfo/tests/bind_usage_info_test.go
@@ -75,11 +75,10 @@ func TestBindUsageInfo(t *testing.T) {
 		tk.MustExec("execute stmt3;")
 		tk.MustExec("select * from t1, t2, t3, t4, t5")
 		time.Sleep(1 * time.Second)
-		checkBindinfoInMemory(t, bindingHandle, checklist, true)
 		// Set all last_used_date to null to simulate that the bindinfo in storage is not updated.
 		tk.MustExec(`update mysql.bind_info set last_used_date = null where original_sql != 'builtin_pseudo_sql_for_bind_lock'`)
 		require.NoError(t, bindingHandle.UpdateBindingUsageInfoToStorage())
-		checkBindinfoInMemory(t, bindingHandle, checklist, false)
+		checkBindinfoInMemory(t, bindingHandle, checklist)
 		tk.MustQuery(`select last_used_date from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock' and last_used_date is null`).Check(testkit.Rows())
 		result := tk.MustQuery(`select sql_digest,last_used_date from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock' order by sql_digest`)
 		t.Log("result:", result.Rows())
@@ -115,13 +114,12 @@ func TestBindUsageInfo(t *testing.T) {
 	tk.MustQuery(`select * from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock' and last_used_date is null`).Check(testkit.Rows())
 }
 
-func checkBindinfoInMemory(t *testing.T, bindingHandle bindinfo.BindingHandle, checklist []string, writeBefore bool) {
+func checkBindinfoInMemory(t *testing.T, bindingHandle bindinfo.BindingHandle, checklist []string) {
 	for _, digest := range checklist {
 		binding := bindingHandle.GetBinding(digest)
 		require.NotNil(t, binding)
-		if writeBefore {
-			require.GreaterOrEqual(t, *binding.UsageInfo.LastUsedAt.Load(), *binding.UsageInfo.LastSavedAt.Load())
-		} else {
+		lastSaved := binding.UsageInfo.LastSavedAt.Load()
+		if lastSaved != nil {
 			require.GreaterOrEqual(t, *binding.UsageInfo.LastSavedAt.Load(), *binding.UsageInfo.LastUsedAt.Load())
 		}
 	}
--- a/pkg/bindinfo/tests/bind_usage_info_test.go
+++ b/pkg/bindinfo/tests/bind_usage_info_test.go
@@ -52,7 +52,7 @@ func TestBindUsageInfo(t *testing.T) {
 	tk.MustExec("create global binding for delete from t1 where b = 1 and c > 1 using delete /*+ use_index(t1,idx_c) */ from t1 where b = 1 and c > 1")
 	tk.MustExec("create global binding for delete t1, t2 from t1 inner join t2 on t1.b = t2.b using delete /*+ inl_join(t1) */ t1, t2 from t1 inner join t2 on t1.b = t2.b")
 	tk.MustExec("create global binding for update t1 set a = 1 where b = 1 and c > 1 using update /*+ use_index(t1,idx_c) */ t1 set a = 1 where b = 1 and c > 1")
-	// corss database binding
+	// cross database binding
 	tk.MustExec(`create global binding using select /*+ leading(t1, t2, t3, t4, t5) */ * from *.t1, *.t2, *.t3, *.t4, *.t5`)
 	tk.MustExec("select * from t1, t2, t3, t4, t5")
 	tk.MustExec("execute stmt1;")
@@ -69,8 +69,7 @@ func TestBindUsageInfo(t *testing.T) {
 	// The last_used_date should be updated.
 	require.True(t, !origin.Equal(result.Rows()))
 	var last *testkit.Result
-	for idx := range 5 {
-		bindinfo.UpdateBindingUsageInfoBatchSize = max(1, idx)
+	for range 5 {
 		tk.MustExec("execute stmt1;")
 		tk.MustExec("execute stmt2;")
 		tk.MustExec("execute stmt3;")
@@ -91,8 +90,7 @@ func TestBindUsageInfo(t *testing.T) {
 	}
 	// Set all last_used_date to null to simulate that the bindinfo in storage is not updated.
 	resetAllLastUsedData(tk)
-	for idx := range 5 {
-		bindinfo.UpdateBindingUsageInfoBatchSize = max(1, idx)
+	for range 5 {
 		time.Sleep(1 * time.Second)
 		// No used, so last_used_date should not be updated.
 		require.NoError(t, bindingHandle.UpdateBindingUsageInfoToStorage())
@@ -101,12 +99,12 @@ func TestBindUsageInfo(t *testing.T) {
 	tk.MustExec("execute stmt1;")
 	tk.MustExec("execute stmt2;")
 	tk.MustExec("execute stmt3;")
-	tk.MustExec("select * from t1, t2, t3, t4, t5")
 	time.Sleep(1 * time.Second)
 	require.NoError(t, bindingHandle.UpdateBindingUsageInfoToStorage())
 	// it has been updated again.
-	tk.MustQuery(
-		fmt.Sprintf(`select * from mysql.bind_info where original_sql != '%s' and last_used_date is null`, bindinfo.BuiltinPseudoSQL4BindLock)).Check(testkit.Rows())
+	rows := tk.MustQuery(
+		fmt.Sprintf(`select * from mysql.bind_info where original_sql != '%s' and last_used_date is not null`, bindinfo.BuiltinPseudoSQL4BindLock)).Rows()
+	require.Len(t, rows, 3)
 }
 
 func resetAllLastUsedData(tk *testkit.TestKit) {
--- a/pkg/bindinfo/tests/bind_usage_info_test.go
+++ b/pkg/bindinfo/tests/bind_usage_info_test.go
@@ -56,7 +56,7 @@ func TestBindUsageInfo(t *testing.T) {
 	tk.MustExec(`create global binding using select /*+ leading(t1, t2, t3, t4, t5) */ * from *.t1, *.t2, *.t3, *.t4, *.t5`)
 	tk.MustExec("select * from t1, t2, t3, t4, t5")
 	tk.MustExec("execute stmt1;")
-	origin := tk.MustQuery(`select sql_digest,last_used_date from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock' order by sql_digest`)
+	origin := tk.MustQuery(fmt.Sprintf(`select sql_digest,last_used_date from mysql.bind_info where original_sql != '%s' order by sql_digest`, bindinfo.BuiltinPseudoSQL4BindLock))
 	origin.Check(testkit.Rows(
 		"5ce1df6eadf8b24222668b1bd2e995b72d4c88e6fe9340d8b13e834703e28c32 <nil>",
 		"5d3975ef2160c1e0517353798dac90a9914095d82c025e7cd97bd55aeb804798 <nil>",
--- a/pkg/bindinfo/tests/bind_usage_info_test.go
+++ b/pkg/bindinfo/tests/bind_usage_info_test.go
@@ -68,7 +68,7 @@ func TestBindUsageInfo(t *testing.T) {
 	t.Log("result:", result.Rows())
 	// The last_used_date should be updated.
 	require.True(t, !origin.Equal(result.Rows()))
-	var last *testkit.Result
+	var first *testkit.Result
 	for range 5 {
 		tk.MustExec("execute stmt1;")
 		tk.MustExec("execute stmt2;")
@@ -82,10 +82,11 @@ func TestBindUsageInfo(t *testing.T) {
 		tk.MustQuery(fmt.Sprintf(`select last_used_date from mysql.bind_info where original_sql != '%s' and last_used_date is null`, bindinfo.BuiltinPseudoSQL4BindLock)).Check(testkit.Rows())
 		result := tk.MustQuery(fmt.Sprintf(`select sql_digest,last_used_date from mysql.bind_info where original_sql != '%s' order by sql_digest`, bindinfo.BuiltinPseudoSQL4BindLock))
 		t.Log("result:", result.Rows())
-		if last == nil {
-			last = result
+		if first == nil {
+			first = result
 		} else {
-			require.True(t, last.Equal(result.Rows()))
+			// in fact, The result of each for-loop should be the same.
+			require.True(t, first.Equal(result.Rows()))
 		}
 	}
 	// Set all last_used_date to null to simulate that the bindinfo in storage is not updated.
--- a/pkg/bindinfo/tests/bind_usage_info_test.go
+++ b/pkg/bindinfo/tests/bind_usage_info_test.go
@@ -16,11 +16,16 @@ package tests
 
 import (
 	"testing"
+	"time"
 
+	"github.com/pingcap/tidb/pkg/bindinfo"
 	"github.com/pingcap/tidb/pkg/testkit"
 )
 
 func TestBindUsageInfo(t *testing.T) {
+	bindinfo.WriteIntervalAfterNoReadBinding = 10 * time.Microsecond
+	bindinfo.MinCheckIntervalForUpdateBindingUsageInfo = 1
+	bindinfo.MaxCheckIntervalForUpdateBindingUsageInfo = 2
 	store := testkit.CreateMockStore(t)
 	tk := testkit.NewTestKit(t, store)
 
@@ -34,4 +39,5 @@ func TestBindUsageInfo(t *testing.T) {
 	tk.MustExec(`create table t (a int, b int, c int, d int, e int, key(a), key(b), key(c), key(d), key(e))`)
 
 	tk.MustExec(`admin reload bindings`)
+
 }
--- a/pkg/bindinfo/tests/bind_usage_info_test.go
+++ b/pkg/bindinfo/tests/bind_usage_info_test.go
@@ -100,13 +100,6 @@ func TestBindUsageInfo(t *testing.T) {
 	tk.MustExec("execute stmt1;")
 	tk.MustExec("execute stmt2;")
 	tk.MustExec("execute stmt3;")
-	tk.MustExec(`delete from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock'`)
-	// Althought we execute the stmts, we set a very long interval, so last_used_date should not be updated.
-	require.NoError(t, bindingHandle.UpdateBindingUsageInfoToStorage())
-	tk.MustQuery(`select * from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock'`).Check(testkit.Rows())
-	tk.MustExec("execute stmt1;")
-	tk.MustExec("execute stmt2;")
-	tk.MustExec("execute stmt3;")
 	tk.MustExec("select * from t1, t2, t3, t4, t5")
 	time.Sleep(1 * time.Second)
 	require.NoError(t, bindingHandle.UpdateBindingUsageInfoToStorage())
--- a/pkg/bindinfo/tests/bind_usage_info_test.go
+++ b/pkg/bindinfo/tests/bind_usage_info_test.go
@@ -1,4 +1,4 @@
-// Copyright 2023 PingCAP, Inc.
+// Copyright 2025 PingCAP, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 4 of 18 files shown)


pkg/session/bootstrap.go
```diff
@@ -296,6 +296,7 @@ const (
 		source VARCHAR(10) NOT NULL DEFAULT 'unknown',
 		sql_digest varchar(64) DEFAULT NULL,
 		plan_digest varchar(64) DEFAULT NULL,
+		last_used_time TIMESTAMP DEFAULT NULL,
```
qw4990: Please also specify the precision, @henrybw just updated the precision to 6 to fix some unstable problems: https://github.com/pingcap/tidb/pull/63524/files

pkg/bindinfo/utils.go
```diff
@@ -167,6 +169,55 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+	// WriteIntervalAfterNoReadBinding indicates the interval at which a write operation needs to be performed after a binding has not been read.
+	WriteIntervalAfterNoReadBinding = 1 * time.Hour
+	// MinCheckIntervalForUpdateBindingUsageInfo indicates the minimum interval to check whether to update binding usage info.
+	MinCheckIntervalForUpdateBindingUsageInfo = 5
+	// MaxCheckIntervalForUpdateBindingUsageInfo indicates the maximum interval to check whether to update binding usage info.
+	MaxCheckIntervalForUpdateBindingUsageInfo = 60
+)
+
+func updateBindingUsageInfoToStorage(sPool util.DestroyableSessionPool, bindings []*Binding) error {
+	err := callWithSCtx(sPool, true, func(sctx sessionctx.Context) error {
```
qw4990: In one of our customer's workload, there are around 100,000 bindings, is this OK to execute 100,000 update statements in one transaction?

qw4990: It's better to limit the number of update statements in this transaction to around 200 or 500 for safety I think. Better to confirm this with our Txn Team members.

pkg/bindinfo/binding.go
```diff
@@ -90,6 +95,30 @@ func (b *Binding) size() float64 {
...
+
+// ResetUsageInfo is to reset the bindinfo usage info after writing into storage.
+func (b *Binding) ResetUsageInfo() {
+	b.UsageInfo.CreateAt.Store(nil)
+	b.UsageInfo.LastUsedAt.Store(nil)
+}
+
+type bindingInfoUsageInfo struct {
+	LastUsedAt atomic.Pointer[time.Time]
+	CreateAt   atomic.Pointer[time.Time]
```
qw4990: Better to rename this field to `FirstUsedAt`.

pkg/bindinfo/utils.go
hawkingrei: OK, I will confirm this with the transaction team.

Additionally, if a cluster really has that many binds, there must also be many TiDB instances. In fact, we need to consider the impact on the entire cluster from multiple instances. So as long as it is sufficiently scattered, with small batches and executed by multiple instances, it should be fine.

hawkingrei: The slow write here actually won't have much impact. DBAs won't delete a bindinfo just because it hasn't been used for a day or two; that would be quite risky.

pkg/domain/domain.go
```diff
@@ -1583,11 +1585,29 @@ func (do *Domain) globalBindHandleWorkerLoop(owner owner.Manager) {
...
+				bindHandle := do.BindingHandle()
+				err := bindHandle.UpdateBindingUsageInfoToStorage()
+				if err != nil {
+					logutil.BgLogger().Warn("BindingHandle.UpdateBindingUsageInfoToStorage", zap.Error(err))
+				}
+				writeBindingUsageTicker.Reset(
+					randomDuration(
+						bindinfo.MinCheckIntervalForUpdateBindingUsageInfo,
+						bindinfo.MaxCheckIntervalForUpdateBindingUsageInfo,
+					),
```
henrybw, line 1621: Picking random durations seems unnecessarily complicated. What's wrong with picking just one check interval?

pkg/domain/domain.go
```diff
@@ -1559,9 +1559,11 @@ func (do *Domain) globalBindHandleWorkerLoop(owner owner.Manager) {
 
 		bindWorkerTicker := time.NewTicker(bindinfo.Lease)
 		gcBindTicker := time.NewTicker(100 * bindinfo.Lease)
+		writeBindingUsageTicker := time.NewTicker(bindinfo.MaxWriteInterval)
```
henrybw: I don't think using `MaxWriteInterval` makes sense here. What happens after restarting a TiDB server that already has existing bindings? This will unconditionally wait 6 hours to check the binding usage, even if the bindings haven't been used since long before the restart.

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,56 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
 
+const updateBindingUsageInfoBatchSize = 10
+
+var (
+	// MaxWriteInterval indicates the interval at which a write operation needs to be performed after a binding has not been read.
+	MaxWriteInterval = 6 * time.Hour
+	// MinCheckIntervalForUpdateBindingUsageInfo indicates the minimum interval to check whether to update binding usage info.
+	MinCheckIntervalForUpdateBindingUsageInfo = 5
+	// MaxCheckIntervalForUpdateBindingUsageInfo indicates the maximum interval to check whether to update binding usage info.
+	MaxCheckIntervalForUpdateBindingUsageInfo = 60
```
henrybw: It's not clear what units these intervals are in.

pkg/bindinfo/tests/bind_usage_info_test.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/pingcap/tidb/pkg/bindinfo"
+	"github.com/pingcap/tidb/pkg/testkit"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBindUsageInfo(t *testing.T) {
```
henrybw, line 27: We should also test accessing the usage info via the `Binding` structure, i.e. accessing it from the binding cache.

henrybw (changes requested):
This doesn't seem to load usage info from storage into the bindings cache? I think we need to also update `readBindingsFromStorage()` to read the `last_used_time` column into the `Binding` structures.

pkg/bindinfo/binding_cache.go
```diff
@@ -38,6 +38,9 @@ type BindingCacheUpdater interface {
 	// LoadFromStorageToCache loads global bindings from storage to the memory cache.
 	LoadFromStorageToCache(fullLoad bool) (err error)
 
+	// updateBindingUsageInfoToStorage is to update the binding usage info into storage
```
time-and-fate:
```suggestion
	// UpdateBindingUsageInfoToStorage is to update the binding usage info into storage
```

pkg/bindinfo/binding_cache.go
```diff
@@ -96,6 +99,12 @@ func (u *bindingCacheUpdater) LoadFromStorageToCache(fullLoad bool) (err error)
 	return nil
 }
 
+// updateBindingUsageInfoToStorage is to update the binding usage info into storage
+func (u *bindingCacheUpdater) UpdateBindingUsageInfoToStorage() error {
+	bindings := u.GetAllBindings()
+	return updateBindingUsageInfoToStorage(u.sPool, bindings)
+}
```
time-and-fate, line 111: I think we need to add `recover()` to prevent possible `panic` from quitting this entire goroutine.

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,56 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+			}
+			intest.Assert(lastUsed.Compare(*lastSaved) >= 0, " lastUsed should be later than or equal to lastSaved")
+			if time.Since(*lastSaved) > MaxWriteInterval {
+				err := saveBindUsage(sctx, binding.SQLDigest, *lastUsed)
+				if err != nil {
+					return err
+				}
+				binding.ResetUsageInfo()
+			}
+			if cnt > updateBindingUsageInfoBatchSize {
```
time-and-fate: `cnt` is never updated.

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,56 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+		return nil
+	})
+	return err
+}
+
+func saveBindUsage(sctx sessionctx.Context, sqldigest string, ts time.Time) error {
+	lastUsedTime := ts.UTC().Format(types.TimeFormat)
+	_, _, err := execRows(
+		sctx,
+		"UPDATE mysql.bind_info SET last_used_time = CONVERT_TZ(%?, '+00:00', @@TIME_ZONE) WHERE sql_digest = %?",
```
time-and-fate: I think this would be a table full scan.

time-and-fate (commented):
In the current implementation, if the tidb server shuts down in 6 hours, the usage info will never get a chance to be recorded.
I think we need to confirm with PM that this is the expected behavior.

hawkingrei:
> This doesn't seem to load usage info from storage into the bindings cache? I think we need to also update `readBindingsFromStorage()` to read the `last_used_time` column into the `Binding` structures.

Yes, it is only for the user to check whether the bind is in use.

pkg/domain/domain.go
hawkingrei: Because the customer's cluster has many TiDB nodes, especially for SaaS clusters, we need to avoid having all TiDB nodes execute batch updates at the same time, which could impact the customer's business. Additionally, updating bindinfo is not urgent, so it's fine to have larger time intervals between updates. Moreover, the code for this is not extensive.

hawkingrei: In the current code, there is also a random interval inside the `deltaUpdateTickerWorker`.

pkg/domain/domain.go
hawkingrei: `MaxWriteInterval` only refers to the maximum write interval for a single record, which is to write the `last used time` after 6 hours. Here, it only controls the timing of the write operation. A restart is a low-probability event; it's unlikely that all nodes in a cluster would be restarting at the same time. Moreover, since real-time performance isn't required, increasing the write interval can help avoid impacting customer operations.

hawkingrei: Since the task will write after 6 hours, there's no need to perform checks within the first 6 hours before startup. After 6 hours, it will simply write according to a random interval.

pkg/bindinfo/utils.go
hawkingrei: Update. I use plan_digest and sql_digest as the condition. then it can use the digest_index to update.

```
+---------------+---------+------+--------------------------------------------------------------+---------------+
| id            | estRows | task | access object                                                | operator info |
+---------------+---------+------+--------------------------------------------------------------+---------------+
| Update_3      | N/A     | root |                                                              | N/A           |
| └─Point_Get_1 | 1.00    | root | table:bind_info, index:digest_index(plan_digest, sql_digest) |               |
+---------------+---------+------+--------------------------------------------------------------+---------------+
```

pkg/bindinfo/binding_cache.go
hawkingrei: OK, I have added.

pkg/bindinfo/utils.go
hawkingrei: Refactor this code and add the new comments.

pkg/bindinfo/binding_cache.go
hawkingrei: Updated.

pkg/bindinfo/utils.go
hawkingrei: I have updated.

pkg/bindinfo/tests/bind_usage_info_test.go
hawkingrei: Great, I have updated.

pkg/session/upgrade.go
```diff
@@ -2027,4 +2028,5 @@ func upgradeToVer251(s sessionapi.Session, _ int64) {
 func upgradeToVer252(s sessionapi.Session, _ int64) {
 	doReentrantDDL(s, "ALTER TABLE mysql.bind_info CHANGE create_time create_time TIMESTAMP(6)")
 	doReentrantDDL(s, "ALTER TABLE mysql.bind_info CHANGE update_time update_time TIMESTAMP(6)")
+	doReentrantDDL(s, "ALTER TABLE mysql.bind_info ADD COLUMN last_used_date DATE DEFAULT NULL AFTER `plan_digest`", infoschema.ErrColumnExists)
```
henrybw, line 2038: Is it safe to modify an existing upgrader? I would expect this to have to be in a new version with its own upgrader. We should confirm this with the DDL team.

pkg/domain/domain.go
henrybw: I think the choice to use `100 * bindinfo.Lease` (like the GCBinding ticker) in the current version of this PR is better. 👍

pkg/domain/domain.go
henrybw: OK.

pkg/bindinfo/binding.go
```diff
@@ -90,6 +95,33 @@ func (b *Binding) size() float64 {
...
+
+// ResetUsageInfo is to reset the bindinfo usage info after writing into storage.
+func (b *Binding) ResetUsageInfo() {
+	ts := time.Now()
+	b.UsageInfo.LastSavedAt.Store(&ts)
+	b.UsageInfo.LastUsedAt.Store(nil)
+}
+
+type bindingInfoUsageInfo struct {
+	LastUsedAt  atomic.Pointer[time.Time]
```
henrybw: Please add a brief comment explaining what `LastUsedAt` is supposed to be: specifically when it is supposed to be nil. It is not obvious until reading the code that this is more like a "last used time since the last save".

pkg/bindinfo/binding.go
```diff
@@ -77,6 +79,9 @@ type Binding struct {
 
 	// TableNames records all schema and table names in this binding statement, which are used for cross-db matching.
 	TableNames []*ast.TableName `json:"-"`
+
+	// usageInfo records the binding usage info
```
henrybw: This comment doesn't add any information. Either remove the comment, or use it to explain what the usage info is for (for example: tracking usage frequency, the `last_used_time` column, etc.).

pkg/session/upgrade.go
hawkingrei: I've already asked. There hasn't been an official release yet, so it's fine to write it this way.

pkg/bindinfo/binding.go
hawkingrei: I've updated the comments here.

pkg/bindinfo/binding.go
hawkingrei: I have added comments for both of these fields.

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,61 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+				// if the last used time is updated for more than 1 hour, we also write it back to storage.
+				time.Since(*lastUsed) > 1*time.Hour {
+				err := saveBindUsage(sctx, binding.SQLDigest, binding.PlanDigest, *lastUsed)
+				if err != nil {
+					return err
+				}
+				binding.ResetUsageInfo()
+				cnt++
+			}
+			if cnt > updateBindingUsageInfoBatchSize {
```
qw4990: This seems really weird, `update 10 random bindings' usage info every 5~100s`.
Can't we simplify this to like `update all bindings' usage info every 2h`?
What do you think of this? @henrybw @time-and-fate

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,61 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+			}
+			lastUsed := binding.UsageInfo.LastUsedAt.Load()
+			if lastUsed == nil {
+				continue
+			}
+			intest.Assert(lastUsed.Compare(*lastSaved) >= 0, " lastUsed should be later than or equal to lastSaved")
+			if time.Since(*lastSaved) > MaxWriteInterval ||
+				// if the last used time is updated for more than 1 hour, we also write it back to storage.
+				time.Since(*lastUsed) > 1*time.Hour {
+				err := saveBindUsage(sctx, binding.SQLDigest, binding.PlanDigest, *lastUsed)
```
qw4990: Please use a explicit transaction to reduce the number of transactions cause by this process, for example:
```
BEGIN;
UPDATE mysql.bind_info USE INDEX(digest_index) ...
UPDATE mysql.bind_info USE INDEX(digest_index) ...
UPDATE mysql.bind_info USE INDEX(digest_index) ...
...
COMMIT;
```

hawkingrei: Updated. saveBindUsage use ```exec``` to execute the sql.

pkg/bindinfo/utils.go
hawkingrei: Now, I update the logic. Update all the binds that meet the conditions in batches of 20 binds

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,109 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+	})
+	if err == nil {
+		for _, binding := range bindings {
+			binding.ResetUsageInfo()
+		}
+	}
+	return err
+}
+
+func addLockForBinds(sctx sessionctx.Context, bindings []*Binding) error {
```
hawkingrei, line 256: @you06 Please take a look.

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,109 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+		planDigest := binding.PlanDigest
+		sql := fmt.Sprintf(" sql_digest = '%s'", sqlDigest)
+		if planDigest == "" {
+			sql += " AND plan_digest IS NULL"
+		} else {
+			sql += fmt.Sprintf(" AND plan_digest = '%s' ", planDigest)
+		}
+		condition = append(condition, sql)
+	}
+	locksql := "select 1 from mysql.bind_info use index(digest_index) where " + strings.Join(condition, " or ") + " for update"
```
you06: Will you consider SQL like this?

```sql
select 1 from mysql.bind_info use index(digest_index) 
where (plan_digest, sql_digest)
in ((...), ..., (...))
for update;
```

hawkingrei: OK, I have updated.

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,110 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+			continue
+		}
+		lastUsed := binding.UsageInfo.LastUsedAt.Load()
+		if lastUsed == nil {
+			continue
+		}
+		intest.Assert(lastUsed.Compare(*lastSaved) >= 0, " lastUsed should be later than or equal to lastSaved")
+		if time.Since(*lastSaved) > MaxWriteInterval ||
+			// if the last used time is updated for more than 1 hour, we also write it back to storage.
+			time.Since(*lastUsed) > 1*time.Hour {
```
qw4990: How about only keeping the first condition here? Updating active bindings every 1h seems a little aggressive?

pkg/bindinfo/binding.go
```diff
@@ -90,6 +96,38 @@ func (b *Binding) size() float64 {
...
+func (b *Binding) UpdateUsageInfo() {
+	b.UsageInfo.Update()
+}
+
+// ResetUsageInfo is to reset the bindinfo usage info after writing into storage.
+func (b *Binding) ResetUsageInfo() {
+	ts := time.Now()
+	b.UsageInfo.LastSavedAt.Store(&ts)
+	b.UsageInfo.LastUsedAt.Store(nil)
+}
```
qw4990: I suggest that we can rename and change this function to like `SetLastSavedAt()`, which 1) make it more clear and simple, 2) it's unnecessary to update `LastUsedAt`?

pkg/domain/domain.go
```diff
@@ -1583,11 +1585,31 @@ func (do *Domain) globalBindHandleWorkerLoop(owner owner.Manager) {
...
+					logutil.BgLogger().Warn("BindingHandle.UpdateBindingUsageInfoToStorage", zap.Error(err))
+				}
+				// randomize the next write interval to avoid thundering herd problem
+				// if there are many tidb servers. The next write interval is [5min, 30min].
+				writeBindingUsageTicker.Reset(
+					randomDuration(
+						5*60,  // 5min
+						30*60, // 30min
+					),
+				)
```
qw4990, line 1622: Since now each time we'll update all bindings, we don't need to trigger it so frequently now? How about every `3h~6h`?

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,110 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
 	return
 }
 
+var (
+	// UpdateBindingUsageInfoBatchSize indicates the batch size when updating binding usage info to storage.
+	UpdateBindingUsageInfoBatchSize = 20
```
qw4990: Updating 20 bindings (rows) each time seems too small from my experience, could we increase this number to like 100 or 200 since they are just simple update statements? Please help me confirm @you06

pkg/bindinfo/utils.go
hawkingrei: It has been removed. and clean this code ahout `1h`.

pkg/session/upgrade.go
henrybw: I still don't think this is a good idea: we should create a new version/upgrader rather than modifying the old one.

Even if this won't impact customers, it's still unnecessarily disruptive to developers because it will break TiDB for anyone who has created or upgraded a cluster to the current version 252. Some developers try to keep their database state around for reproing issues, etc, and this will unnecessarily make their `mysql.bind_info` schema inconsistent with what TiDB expects.

Creating a new version/upgrader is not that much more work, and it avoids causing unnecessary disruption to other developers.

pkg/bindinfo/utils.go
henrybw: > This seems really weird, `update 10 random bindings' usage info every 5~100s`. Can't we simplify this to like `update all bindings' usage info every 2h`? What do you think of this? @henrybw @time-and-fate

@qw4990 I do agree with you that this seems really weird. I'm willing to accept this approach for now since (1) I guess there is precedent in TiDB for this method to avoid the thundering herd problem, and (2) there doesn't seem to be an abstraction for rate-limiting or batching large amounts of updates for internal system operations.

**However**, we really can (and should) do better. Both DDL and analyze/stats processing have their own job/queueing systems for handling this kind of problem, in addition to other one-off solutions in other places (like `deltaUpdateTickerWorker`). It would be better to have an abstraction that solves this problem *once*, rather than repeatedly solving it in different ways in different places.

That said, I understand that such an effort would be outside the scope of this PR. But I really think we should do better than this.

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,109 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+				return errors.Trace(err)
+			}
+		}
+		return nil
+	})
+	if err == nil {
+		ts := time.Now()
+		for _, binding := range bindings {
+			binding.UpdateSavedAt(&ts)
+		}
```
henrybw, line 251: How about calling `binding.UpdateSavedAt(&ts)` in the `callWithSCtx` callback, after we `saveBindUsage()`? That way we don't have to loop over the bindings twice.

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,109 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+func updateBindingUsageInfoToStorageInternal(sPool util.DestroyableSessionPool, bindings []*Binding) error {
+	err := callWithSCtx(sPool, true, func(sctx sessionctx.Context) (err error) {
+		if err = addLockForBinds(sctx, bindings); err != nil {
+			return errors.Trace(err)
+		}
+		for _, binding := range bindings {
+			lastUsed := binding.UsageInfo.LastUsedAt.Load()
+			if lastUsed == nil {
+				continue
+			}
```
henrybw: This shouldn't be possible, since a binding is only included if its `lastSaved` and `lastUsed` are both not nil. I think this should at least be an `intest.Assert` instead.

pkg/bindinfo/tests/bind_usage_info_test.go
```diff
@@ -0,0 +1,114 @@
...
+	tk.MustExec("prepare stmt3 from 'update t1 set a = 1 where b = 1 and c > 1';")
+	tk.MustExec("execute stmt1;")
+	tk.MustExec("create global binding for delete from t1 where b = 1 and c > 1 using delete /*+ use_index(t1,idx_c) */ from t1 where b = 1 and c > 1")
+	tk.MustExec("create global binding for delete t1, t2 from t1 inner join t2 on t1.b = t2.b using delete /*+ inl_join(t1) */ t1, t2 from t1 inner join t2 on t1.b = t2.b")
+	tk.MustExec("create global binding for update t1 set a = 1 where b = 1 and c > 1 using update /*+ use_index(t1,idx_c) */ t1 set a = 1 where b = 1 and c > 1")
+	// corss database binding
+	tk.MustExec(`create global binding using select /*+ leading(t1, t2, t3, t4, t5) */ * from *.t1, *.t2, *.t3, *.t4, *.t5`)
+	tk.MustExec("select * from t1, t2, t3, t4, t5")
+	tk.MustExec("execute stmt1;")
+	origin := tk.MustQuery(`select sql_digest,last_used_date from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock' order by sql_digest`)
```
henrybw: Instead of hardcoding the literal value `'builtin_pseudo_sql_for_bind_lock'` everywhere, I think it would be better to use the `bindinfo.BuiltinPseudoSQL4BindLock` constant.

pkg/bindinfo/tests/bind_usage_info_test.go
```diff
@@ -0,0 +1,114 @@
+// Copyright 2023 PingCAP, Inc.
```
henrybw: 2025?

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,109 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+		sql := fmt.Sprintf("('%s'", sqlDigest)
+		if planDigest == "" {
+			sql += ",NULL)"
+		} else {
+			sql += fmt.Sprintf(",'%s')", planDigest)
+		}
+		condition = append(condition, sql)
+	}
+	locksql := "select 1 from mysql.bind_info use index(digest_index) where (plan_digest, sql_digest) in (" +
+		strings.Join(condition, " , ") + ") for update"
```
henrybw, line 270: I thought `builtin_pseudo_sql_for_bind_lock` was the synchronization mechanism that bindings used for locking? If this uses `SELECT FOR UPDATE`, but the bindings cache expects `builtin_pseudo_sql_for_bind_lock`, I don't think this lock will work as intended.

(cc @qw4990: is the above accurate?)

pkg/domain/domain.go
hawkingrei: Updated

pkg/bindinfo/utils.go
hawkingrei: Inside the `callWithSCtx` function, it will run "begin" and "commit" for us. It's possible that we don't encounter any errors during execution, but an issue occurs when committing. In such cases, the commit is considered a failure, and we should not record the write timestamp.

pkg/bindinfo/utils.go
hawkingrei: You are right.

pkg/bindinfo/tests/bind_usage_info_test.go
hawkingrei: The only thing I'm worried about is that when there are suddenly a large number of nodes, if multiple nodes write simultaneously, we might all compete for a single row lock, which could lead to additional issues.

pkg/bindinfo/tests/bind_usage_info_test.go
hawkingrei: Updated.

pkg/bindinfo/utils.go
qw4990: @henrybw Oh you are right... Nice catch, I almost forget we have such a way to lock binding table before updating. There is an implicit assumption here that each time before updating bindings, we have to run `func lockBindInfoTable(sctx sessionctx.Context)` to lock this table, please use this way instead of `select for update`. @hawkingrei

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,109 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
 }
 
+var (
+	// UpdateBindingUsageInfoBatchSize indicates the batch size when updating binding usage info to storage.
+	UpdateBindingUsageInfoBatchSize = 100
+	// MaxWriteInterval indicates the interval at which a write operation needs to be performed after a binding has not been read.
+	MaxWriteInterval = 6 * time.Hour
+)
+
+func updateBindingUsageInfoToStorage(sPool util.DestroyableSessionPool, bindings []*Binding) error {
```
qw4990, line 180: Do we have any test data that how long this would take for 10,000 and 100,000 bindings?

qw4990: Please add some comments of its performance like `around 1m for 10,000 bindings`. I'm afraid users would ask us this in the future, better to leave some comments for it in case other developers don't know its performance.

pkg/bindinfo/utils.go
you06: You can lock the table via `lockBindInfoTable`, but I still suggest `select for update` which can be kind of prefetch operation. All the updates after `select for update` can be handled in TiDB locally without loading data from TiKV or acquiring any pessimistic locks.

pkg/bindinfo/utils.go
hawkingrei: I create 44679 bindings and remove the limit condition to make tidb always write the binding info.

```
[2025/09/22 19:11:41.210 +08:00] [INFO] [utils.go:185] ["update binding usage info to storage"] [category=sql-bind] [count=44679] [duration=26.241038459s]
```

hawkingrei: it spends 26.241038459s to update 44679 bindings

pkg/bindinfo/utils.go
hawkingrei: OK, so I use both lockBindInfoTable and ```select for update```.

I create a cluster with 44679 bindings to update all bindinfo usage.

it speeds 26.241038459s to update 44679 bindings last_uesd_data.

pkg/session/upgrade.go
hawkingrei: Yes, but since we haven't officially released the version yet, we try to keep the modifications to a table within a single function using one version. If this is cherrypicked to an older version, additional version modifications would be required. The version numbers between 218 and 239 are reserved for subsequent additional table structure updates. In fact, the versions here are also limited, so don't waste them excessively.

image

hawkingrei: OK, I have updated and create a new version for it.

pkg/bindinfo/binding.go
hawkingrei: Update

pkg/session/upgrade.go
henrybw: Thank you.

pkg/bindinfo/tests/bind_usage_info_test.go
henrybw: That's not what I was talking about. I was referring to the use of the literal string `'builtin_pseudo_sql_for_bind_lock'` in the tests, and was suggesting to replace that with the constant `bindinfo.BuiltinPseudoSQL4BindLock` instead.

pkg/bindinfo/utils.go
henrybw: Got it. Makes sense.

pkg/bindinfo/tests/bind_test.go
```diff
@@ -44,7 +44,6 @@ func utilCleanBindingEnv(tk *testkit.TestKit) {
 func TestPrepareCacheWithBinding(t *testing.T) {
 	store := testkit.CreateMockStore(t)
 	tk := testkit.NewTestKit(t, store)
-	tk.MustExec(`set tidb_enable_prepared_plan_cache=1`)
```
henrybw, line 47: Why was this removed?

pkg/bindinfo/binding_cache.go
```diff
@@ -96,6 +99,17 @@ func (u *bindingCacheUpdater) LoadFromStorageToCache(fullLoad bool) (err error)
 	return nil
 }
 
+// updateBindingUsageInfoToStorage is to update the binding usage info into storage
```
henrybw: Nitpick:
```suggestion
// UpdateBindingUsageInfoToStorage is to update the binding usage info into storage
```

pkg/session/upgrade.go
```diff
@@ -466,8 +466,13 @@ const (
 	version251 = 251
 
 	// version 252
-	// Update FSP of mysql.bind_info timestamp columns to microsecond precision.
+	// Update FSP of mysql.bind_info timestamp columns to microsecond precision and
+	//
```
henrybw: Nitpick: this change is unnecessary.
```suggestion
	// Update FSP of mysql.bind_info timestamp columns to microsecond precision.
```

henrybw (commented):
This seems like it should work. I don't have any more major complaints.

pkg/bindinfo/binding.go
```diff
@@ -90,6 +96,36 @@ func (b *Binding) size() float64 {
...
+	LastSavedAt atomic.Pointer[time.Time]
+}
+
+func (b *bindingInfoUsageInfo) Update() {
+	now := time.Now()
+	if b.LastSavedAt.Load() == nil {
+		// If `LastSavedAt`  is equal to `nil`, we can consider it as a record that has not been written.
+		// Thus, from the first read, we take this as its saved time. After that, it will be written after a while.
+		b.LastSavedAt.Store(&now)
+	}
```
henrybw: > Thus, from the first read, we take this as its saved time. After that, it will be written after a while.

This doesn't actually make sense to me. There is no justification for why we need to make `LastSavedAt` have a misleading/incorrect value sometimes, but not others.

Why should we take this "first used" time as its "last saved" time? Why not leave `LastSavedAt` as nil until it is first written?

pkg/bindinfo/binding_cache.go
```diff
@@ -96,6 +99,17 @@ func (u *bindingCacheUpdater) LoadFromStorageToCache(fullLoad bool) (err error)
...
+func (u *bindingCacheUpdater) UpdateBindingUsageInfoToStorage() error {
+	defer func() {
+		if r := recover(); r != nil {
+			bindingLogger().Warn("panic when update usage info for binding", zap.Any("recover", r))
+		}
+	}()
+	bindings := u.GetAllBindings()
+	return updateBindingUsageInfoToStorage(u.sPool, bindings)
```
henrybw, line 110: I'm concerned that using `GetAllBindings()` like this will add unnecessary memory and CPU pressure, especially if we're dealing with 10,000+ or 100,000+ bindings.

This is similar to the problematic pattern we saw in the usage of `ListTables`: get all objects, then loop over every object to determine which is relevant. Those became scalability bottlenecks because we would have to pay the cost of scanning every object, even if only a small subset was relevant.

Similarly, if there are many bindings but only a few of them are used, we still have to pay the cost of looping through every binding in the system.

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,123 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+			continue
+		}
+		intest.AssertFunc(func() bool {
+			lastUsed := binding.UsageInfo.LastUsedAt.Load()
+			if lastUsed == nil {
+				return false
+			}
+			return lastUsed.Compare(*lastSaved) >= 0
+		}, " lastUsed should be later than or equal to lastSaved and lastSaved is not nil")
+		if time.Since(*lastSaved) > MaxWriteInterval {
```
henrybw: Sorry, this logic doesn't actually make sense to me either. If `lastSaved == nil`, that means this is a record that has not been written yet. But here, we skip writing the record if `lastSaved == nil`. So when should a record be written?

It seems that we have to wait until the binding has been used first before we can write its usage info, because then `lastSaved` will be set to its first usage time. That doesn't make sense though, because then the `lastSaved` time is sometimes not actually the time the usage info was last saved?

I think this is really confusing. `lastSaved` shouldn't matter here. The only thing `lastSaved` should affect is the `MaxWriteInterval` check, and if `lastSaved` is nil, then we should just write if the binding has been used.

henrybw (commented):
Sorry, after going over these changes again, I actually do have some more substantial concerns that need to be addressed before we can merge this.

Most importantly, the `LastSavedAt` logic is confusing and overly complicated. I think it needs to be simplified.

pkg/bindinfo/tests/bind_test.go
hawkingrei: We have enable `tidb_enable_prepared_plan_cache` by defaults.

pkg/bindinfo/binding_cache.go
hawkingrei: Reasonable, but here we need to make additional modifications to the binding cache. However, given that the performance tested here still meets the requirements and will not be triggered frequently, we can optimize the performance here later.

pkg/bindinfo/utils.go
hawkingrei: You are right. I did write it a bit confusingly here. I have updated it again and strengthened the testing.

hawkingrei:
> Sorry, after going over these changes again, I actually do have some more substantial concerns that need to be addressed before we can merge this.
> 
> Most importantly, the `LastSavedAt` logic is confusing and overly complicated. I think it needs to be simplified.

Actually, if the last save time is not recorded, it is not possible to determine whether it has been written last time, which can easily lead to additional issues.

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,121 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+	locksql := "select 1 from mysql.bind_info use index(digest_index) where (plan_digest, sql_digest) in (" +
+		strings.Join(condition, " , ") + ") for update"
+	_, err := exec(sctx, locksql)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	return nil
+}
+
+func saveBindUsage(sctx sessionctx.Context, sqldigest, planDigest string, ts time.Time) error {
```
qw4990: Please unify the term, `bidning` not `bind`, better to rename it to `saveBindingUsage`

qw4990:
Please help review it @henrybw ~

pkg/session/upgrade.go
```diff
@@ -2028,3 +2033,7 @@ func upgradeToVer252(s sessionapi.Session, _ int64) {
 	doReentrantDDL(s, "ALTER TABLE mysql.bind_info CHANGE create_time create_time TIMESTAMP(6)")
 	doReentrantDDL(s, "ALTER TABLE mysql.bind_info CHANGE update_time update_time TIMESTAMP(6)")
 }
+
+func upgradeToVer253(s sessionapi.Session, _ int64) {
+	doReentrantDDL(s, "ALTER TABLE mysql.bind_info ADD COLUMN last_used_date DATE DEFAULT NULL AFTER `plan_digest`", infoschema.ErrColumnExists)
```
D3Hunter, line 2038: please test the upgrade works for nextgen

pkg/bindinfo/tests/bind_usage_info_test.go
henrybw: This comment still has not been addressed.

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,121 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+		lastUsedAt := binding.UsageInfo.LastUsedAt.Load()
+		if lastUsedAt == nil {
+			continue
+		}
+		lastSaved := binding.UsageInfo.LastSavedAt.Load()
+		// If a certain amount of time specified by MaxWriteInterval has passed since the last record was written,
+		// and it has been used in between, it will be written.
+		// If it has never been written before, it will be written if lastUsedAt exceeds MaxWriteInterval.
+		if (lastSaved != nil && time.Since(*lastSaved) >= MaxWriteInterval && lastUsedAt.After(*lastSaved)) ||
+			(lastSaved == nil && time.Since(*lastUsedAt) > MaxWriteInterval) {
```
henrybw: I don't think this delay for the first write makes sense. If a binding is used frequently, then this prevents `last_used_date` from ever being updated, so it will always be NULL. But that makes frequently used bindings misleadingly look like they have never been used.

I think this would be confusing from the user/DBA's perspective, because a NULL `last_used_date` could mean that the binding has never been used, or that the binding is frequently being used.

I think we should just always write if `lastSaved` is nil. We don't need to delay based on how recent `lastUsedAt` is. This ticker is only going to run once every 3-6 hours, so there's already going to be a delay for the first write anyway.

```suggestion
			lastSaved == nil {
```

pkg/bindinfo/tests/bind_usage_info_test.go
```diff
@@ -0,0 +1,126 @@
...
+		require.NoError(t, bindingHandle.UpdateBindingUsageInfoToStorage())
+		checkBindinfoInMemory(t, bindingHandle, checklist)
+		tk.MustQuery(`select last_used_date from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock' and last_used_date is null`).Check(testkit.Rows())
+		result := tk.MustQuery(`select sql_digest,last_used_date from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock' order by sql_digest`)
+		t.Log("result:", result.Rows())
+		if last == nil {
+			last = result
+		} else {
+			require.True(t, last.Equal(result.Rows()))
+		}
```
henrybw: What is this trying to test? `last` is never updated after the first iteration.

pkg/bindinfo/tests/bind_usage_info_test.go
```diff
@@ -0,0 +1,126 @@
...
+		bindinfo.UpdateBindingUsageInfoBatchSize = max(1, idx)
+		time.Sleep(1 * time.Second)
+		// No used, so last_used_date should not be updated.
+		require.NoError(t, bindingHandle.UpdateBindingUsageInfoToStorage())
+		tk.MustQuery(`select last_used_date from mysql.bind_info where last_used_date is not null`).Check(testkit.Rows())
+	}
+	tk.MustExec("execute stmt1;")
+	tk.MustExec("execute stmt2;")
+	tk.MustExec("execute stmt3;")
+	tk.MustExec(`delete from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock'`)
```
henrybw: I don't think this `DELETE` statement is correct. It deletes all the bindings created earlier, so the below tests can never fail, because there are no longer any bindings in `mysql.bind_info` to check.

pkg/bindinfo/binding.go
```diff
@@ -90,6 +96,27 @@ func (b *Binding) size() float64 {
...
+}
+
+// UpdateLastSavedAt is to update the last saved time
+func (b *Binding) UpdateLastSavedAt(ts *time.Time) {
+	b.UsageInfo.LastSavedAt.Store(ts)
+}
+
+type bindingInfoUsageInfo struct {
+	// LastUsedAt records the last time when this binding is used.
+	// It is nil if this binding has never been used or has been reset after writing into storage.
```
henrybw: > or has been reset after writing into storage.

I don't think this part is true anymore: `LastUsedAt` should only be nil if the binding has never been used.

```suggestion
	// It is nil if this binding has never been used.
```

pkg/bindinfo/tests/bind_usage_info_test.go
```diff
@@ -0,0 +1,126 @@
...
+	tk.MustExec("execute stmt1;")
+	origin := tk.MustQuery(`select sql_digest,last_used_date from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock' order by sql_digest`)
+	origin.Check(testkit.Rows(
+		"5ce1df6eadf8b24222668b1bd2e995b72d4c88e6fe9340d8b13e834703e28c32 <nil>",
+		"5d3975ef2160c1e0517353798dac90a9914095d82c025e7cd97bd55aeb804798 <nil>",
+		"9d3995845aef70ba086d347f38a4e14c9705e966f7c5793b9fa92194bca2bbef <nil>",
+		"aa3c510b94b9d680f729252ca88415794c8a4f52172c5f9e06c27bee57d08329 <nil>"))
+	time.Sleep(50 * time.Microsecond)
+	require.NoError(t, bindingHandle.UpdateBindingUsageInfoToStorage())
+	result := tk.MustQuery(`select sql_digest,last_used_date from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock' order by sql_digest`)
```
henrybw: Again, I think we should use the `bindinfo.BuiltinPseudoSQL4BindLock` constant, and not use the literal hardcoded string `'builtin_pseudo_sql_for_bind_lock'`:

```suggestion
	result := tk.MustQuery(fmt.Sprintf(`select sql_digest,last_used_date from mysql.bind_info where original_sql != '%s' order by sql_digest`, bindinfo.BuiltinPseudoSQL4BindLock))
```

Please make sure all uses of `'builtin_pseudo_sql_for_bind_lock'` in this file are changed.

Notice that the literal string `"builtin_pseudo_sql_for_bind_lock"` is never used outside of pkg/bindinfo/binding_handle.go. Also, using the constant instead of the literal string is how `TestRemoveDuplicatedPseudoBinding` in pkg/bindinfo/binding_operator_test.go works.

For a magic value/constant like the name `"builtin_pseudo_sql_for_bind_lock"`, it is generally better to refer to its symbolic constant rather than its literal value, because it means the value is only defined in one place. If the value needs to change, it only needs to be updated in one place, rather than in every place where it is used.

pkg/infoschema/test/clustertablestest/cluster_tables_test.go
```diff
@@ -1530,7 +1531,10 @@ func TestSetBindingStatusBySQLDigest(t *testing.T) {
 	sql = "select * from t where t.a = 1"
 	tk.MustExec(sql)
 	tk.MustQuery("select @@last_plan_from_binding").Check(testkit.Rows("1"))
-
+	bindinfo.MaxWriteInterval = 1 * time.Microsecond
+	time.Sleep(1 * time.Second)
+	require.NoError(t, s.dom.BindingHandle().UpdateBindingUsageInfoToStorage())
+	tk.MustQuery(`select last_used_date from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock' and last_used_date is null`).Check(testkit.Rows())
```
henrybw: Please use the `bindinfo.BuiltinPseudoSQL4BindLock` constant instead.

pkg/session/bootstrap_test.go
```diff
@@ -157,6 +157,12 @@ func TestBootstrap(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, 0, req.NumRows())
 	se.Close()
+	r = MustExecToRecodeSet(t, se, "select * from mysql.bind_info where original_sql = 'builtin_pseudo_sql_for_bind_lock'")
```
henrybw: Please use the `bindinfo.BuiltinPseudoSQL4BindLock` constant instead.

pkg/bindinfo/tests/bind_usage_info_test.go
```diff
@@ -0,0 +1,126 @@
...
+	tk.MustExec("create table t5(a int, b int, c int, key idx_b(b), key idx_c(c))")
+
+	tk.MustExec("prepare stmt1 from 'delete from t1 where b = 1 and c > 1';")
+	tk.MustExec("prepare stmt2 from 'delete t1, t2 from t1 inner join t2 on t1.b = t2.b';")
+	tk.MustExec("prepare stmt3 from 'update t1 set a = 1 where b = 1 and c > 1';")
+	tk.MustExec("execute stmt1;")
+	tk.MustExec("create global binding for delete from t1 where b = 1 and c > 1 using delete /*+ use_index(t1,idx_c) */ from t1 where b = 1 and c > 1")
+	tk.MustExec("create global binding for delete t1, t2 from t1 inner join t2 on t1.b = t2.b using delete /*+ inl_join(t1) */ t1, t2 from t1 inner join t2 on t1.b = t2.b")
+	tk.MustExec("create global binding for update t1 set a = 1 where b = 1 and c > 1 using update /*+ use_index(t1,idx_c) */ t1 set a = 1 where b = 1 and c > 1")
+	// corss database binding
```
henrybw:
```suggestion
	// cross database binding
```

pkg/bindinfo/tests/bind_usage_info_test.go
```diff
@@ -0,0 +1,126 @@
...
+		time.Sleep(1 * time.Second)
+		// No used, so last_used_date should not be updated.
+		require.NoError(t, bindingHandle.UpdateBindingUsageInfoToStorage())
+		tk.MustQuery(`select last_used_date from mysql.bind_info where last_used_date is not null`).Check(testkit.Rows())
+	}
+	tk.MustExec("execute stmt1;")
+	tk.MustExec("execute stmt2;")
+	tk.MustExec("execute stmt3;")
+	tk.MustExec(`delete from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock'`)
+	// Alought we execute the stmts, but we set a very long interval, so last_used_date should not be updated.
```
henrybw:
```suggestion
	// Although we execute the stmts, we set a very long interval, so last_used_date should not be updated.
```

pkg/session/bootstrap_test.go
hawkingrei: Has been updated.

pkg/bindinfo/tests/bind_usage_info_test.go
hawkingrei: I have update it.

pkg/bindinfo/tests/bind_usage_info_test.go
hawkingrei: L79 has reset all last date. The content here is updated each time, and remains consistent.

pkg/bindinfo/tests/bind_usage_info_test.go
hawkingrei: Updated

pkg/bindinfo/tests/bind_usage_info_test.go
hawkingrei: Updated.

pkg/bindinfo/tests/bind_usage_info_test.go
hawkingrei: Updated

pkg/bindinfo/tests/bind_usage_info_test.go
hawkingrei: Updated

pkg/infoschema/test/clustertablestest/cluster_tables_test.go
hawkingrei: Updated

pkg/bindinfo/binding.go
hawkingrei: Updated

pkg/bindinfo/utils.go
hawkingrei: Yes, I have updated it.

pkg/bindinfo/tests/bind_usage_info_test.go
henrybw: So this is testing that every result is equal to the first result?

If so, `first` would be a better name. `last` is confusing, since "last" usually means "most recent" (which is how "last used" and "last saved" are both used in this PR).

pkg/bindinfo/utils.go
```diff
@@ -167,6 +170,127 @@ func readBindingsFromStorage(sPool util.DestroyableSessionPool, condition string
...
+		err := updateBindingUsageInfoToStorageInternal(sPool, toWrite)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func shouldUpdateBinding(lastSaved, lastUsed *time.Time) bool {
+	if lastUsed == nil {
```
henrybw: This should be checking `lastSaved`, not `lastUsed`.

pkg/bindinfo/tests/bind_usage_info_test.go
```diff
@@ -0,0 +1,123 @@
...
+	tk.MustExec("prepare stmt3 from 'update t1 set a = 1 where b = 1 and c > 1';")
+	tk.MustExec("execute stmt1;")
+	tk.MustExec("create global binding for delete from t1 where b = 1 and c > 1 using delete /*+ use_index(t1,idx_c) */ from t1 where b = 1 and c > 1")
+	tk.MustExec("create global binding for delete t1, t2 from t1 inner join t2 on t1.b = t2.b using delete /*+ inl_join(t1) */ t1, t2 from t1 inner join t2 on t1.b = t2.b")
+	tk.MustExec("create global binding for update t1 set a = 1 where b = 1 and c > 1 using update /*+ use_index(t1,idx_c) */ t1 set a = 1 where b = 1 and c > 1")
+	// cross database binding
+	tk.MustExec(`create global binding using select /*+ leading(t1, t2, t3, t4, t5) */ * from *.t1, *.t2, *.t3, *.t4, *.t5`)
+	tk.MustExec("select * from t1, t2, t3, t4, t5")
+	tk.MustExec("execute stmt1;")
+	origin := tk.MustQuery(`select sql_digest,last_used_date from mysql.bind_info where original_sql != 'builtin_pseudo_sql_for_bind_lock' order by sql_digest`)
```
henrybw: `bindinfo.BuiltinPseudoSQL4BindLock`

pkg/session/bootstrap_test.go
henrybw: `bindinfo.BuiltinPseudoSQL4BindLock`

pkg/bindinfo/tests/bind_usage_info_test.go
hawkingrei: Updated

pkg/session/bootstrap_test.go
hawkingrei: Updated.

pkg/bindinfo/utils.go
hawkingrei: Right.

pkg/bindinfo/tests/bind_usage_info_test.go
hawkingrei: Great suggestion. I've updated it and added some comments.

yudongusa (approved):

PR #63457, opened 2025-09-11
planner, executor: code preparation for supporting access path `keep order` with `IN` Conditions using merge sort

time-and-fate:

Issue Number: ref #63449

Problem Summary:

This PR doesn't contain actual logic changes. It's for reducing modifications for the next PR.

- planner
  - `isMatchProp()` (`pkg/planner/core/find_best_task.go`)
    - Renamed to `matchProperty()`
    - Return type changed from a `bool` to `PhysicalPropMatchResult`, which is a newly introduced `enum`.
    - Slightly refactor the current code to make it cleaner without modifying the actual logic.
  - New enum `PhysicalPropMatchResult` (`pkg/planner/property/physical_property.go`)
    - `PropNotMatched`: corresponds to previous `isMatchProp() == false`.
    - `PropMatched`: corresponds to previous `isMatchProp() == true`.
- executor
  - Extract function `collectColIdxFromByItems()` for later use.
  - Add assertion that expressions in `ByItems` must be `Column`s.
- tests
  - Added the integration tests so that we can see the plan changes and also make sure there are no correctness issues in the next PR.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/distsql.go
+++ b/pkg/executor/distsql.go
@@ -377,7 +377,7 @@ func (e *IndexReaderExecutor) open(ctx context.Context, kvRanges []kv.KeyRange) 
 	slices.SortFunc(kvRanges, func(i, j kv.KeyRange) int {
 		return bytes.Compare(i.StartKey, j.StartKey)
 	})
-	if !needMergeSort(e.byItems, len(kvRanges)) {
+	if !needMergeSort(e.byItems, len(e.partitions)) {
 		kvReq, err := e.buildKVReq(kvRanges)
 		if err != nil {
 			return err


pkg/executor/builder.go
```diff
@@ -3876,6 +3887,15 @@ func (b *executorBuilder) buildMPPGather(v *physicalop.PhysicalTableReader) exec
 	return gather
 }
 
+// assertByItemsAreColumns asserts that all expressions in ByItems are Column types.
+// This function is used to validate PhysicalIndexScan and PhysicalTableScan ByItems.
+func assertByItemsAreColumns(byItems []*plannerutil.ByItems) {
```
winoros, line 3892: We can use intest.AssertFunc, this can reduce the cost when the intest is not enabled.
golang's optimization will not optimize this function block...

pkg/executor/distsql.go
```diff
@@ -377,8 +377,7 @@ func (e *IndexReaderExecutor) open(ctx context.Context, kvRanges []kv.KeyRange)
 	slices.SortFunc(kvRanges, func(i, j kv.KeyRange) int {
 		return bytes.Compare(i.StartKey, j.StartKey)
 	})
-	// use sortedSelectResults only when byItems pushed down and partition numbers > 1
-	if e.byItems == nil || len(e.partitions) <= 1 {
```
winoros, line 381: This check is changed?

pkg/planner/property/physical_property.go
```diff
@@ -220,6 +220,21 @@ const (
...
 )
 
+// PhysicalPropMatchResult describes the result of matching PhysicalProperty against an access path.
+type PhysicalPropMatchResult int
+
+const (
+	// PropNotMatched means the access path cannot satisfy the required order.
+	PropNotMatched PhysicalPropMatchResult = iota
+	// PropMatched means the access path can satisfy the required property directly.
+	PropMatched
```
winoros, line 230: Will we extend its value so that bool's 0/1 representation is not enough?

pkg/executor/builder.go
time-and-fate: Updated.

pkg/executor/distsql.go
time-and-fate: Good catch. Though this doesn't affect correctness, I reverted it. It would be included in the next PR.

pkg/planner/property/physical_property.go
time-and-fate: Yes. We will add a new `PropMatchedNeedMergeSort`.
You can see the full PR in https://github.com/pingcap/tidb/pull/62694

wshwsh12 (approved):
executor part LGTM

winoros (approved):

Issue #63366, opened 2025-09-04
Lossy DDL optimization


Benjamin2037:
## Enhancement
**Background** 
TiDB has already implemented all DDL  online changes, and this capability enables TiDB users to maintain online business operations while performing DDL changes. This is crucial for TiDB databases for users whose business is developing rapidly and requires continuous changes to the database object structure. The ability to execute these DDL statements online, as well as the ability to execute them at the fastest speed with minimal impact on user business, has become an important indicator for helping customers succeed. 

**Phase 1:**

- Improve the execution efficiency of lossy column changes by first directly supporting 1PC transaction processing on the TiDB side.

- Enhance the scalability of lossy column changes by integrating with Distributed eXecution Framework(DXF), which, building on the 1PC transaction optimization, will improve the utilization and flexibility of computing resources.

- Provide the simplest DDL and user transaction conflict handling logic to ensure that online DDL transactions can balance execution efficiency with minimal business impact.

PR #63465, opened 2025-09-11
ddl: make some `MODIFY COLUMN` skip row reorg

joechenrh:

Issue Number: ref #63595 #63366

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/modify_column.go
+++ b/pkg/ddl/modify_column.go
@@ -143,7 +143,7 @@ func getModifyColumnType(
 	}
 
 	relatedIndexes := getRelatedIndexIDs(tblInfo, oldCol.ID, false)
-	if len(relatedIndexes) == 0 || !needIndexReorg(tblInfo, oldCol, args.Column) {
+	if len(relatedIndexes) == 0 || !needIndexReorg(oldCol, args.Column) {
 		return ModifyTypeNoReorgWithCheck
 	}
 	return ModifyTypeIndexReorg
@@ -774,7 +774,7 @@ func adjustForeignKeyChildTableInfoAfterModifyColumn(infoCache *infoschema.InfoC
 	return infoList, nil
 }
 
-func needIndexReorg(_ *model.TableInfo, oldCol, changingCol *model.ColumnInfo) bool {
+func needIndexReorg(oldCol, changingCol *model.ColumnInfo) bool {
 	if mysql.IsIntegerType(oldCol.GetType()) && mysql.IsIntegerType(changingCol.GetType()) {
 		return mysql.HasUnsignedFlag(oldCol.GetFlag()) != mysql.HasUnsignedFlag(changingCol.GetFlag())
 	}
@@ -790,6 +790,8 @@ func needIndexReorg(_ *model.TableInfo, oldCol, changingCol *model.ColumnInfo) b
 	}
 
 	// Check index value part, ref tablecodec.GenIndexValuePortal
+	// TODO(joechenrh): It's better to check each index here, because not all indexes need
+	// reorg even if the below condition is true.
 	return types.NeedRestoredData(&oldCol.FieldType) != types.NeedRestoredData(&changingCol.FieldType)
 }
 
--- a/pkg/meta/model/column.go
+++ b/pkg/meta/model/column.go
@@ -61,9 +61,10 @@ type ColumnInfo struct {
 	GeneratedStored     bool                `json:"generated_stored"`
 	Dependences         map[string]struct{} `json:"dependences"`
 	FieldType           types.FieldType     `json:"type"`
-	ChangingFieldType   *types.FieldType    `json:"changing_type,omitempty"`
-	State               SchemaState         `json:"state"`
-	Comment             string              `json:"comment"`
+	// ChangingFieldType is used to store the new type of modify column.
+	ChangingFieldType *types.FieldType `json:"changing_type,omitempty"`
+	State             SchemaState      `json:"state"`
+	Comment           string           `json:"comment"`
 	// A hidden column is used internally(expression index) and are not accessible by users.
 	Hidden           bool `json:"hidden"`
 	*ChangeStateInfo `json:"change_state_info"`
--- a/pkg/table/tables/tables.go
+++ b/pkg/table/tables/tables.go
@@ -280,21 +280,27 @@ func initTableIndices(t *TableCommon) error {
 	return nil
 }
 
-// checkDataWithModifyColumn checks if the data can be stored in the column with changingType.
+// checkDataForModifyColumn checks if the data can be stored in the column with changingType.
 // It's used to prevent illegal data being inserted if we want to skip reorg.
-func checkDataWithModifyColumn(data types.Datum, changingType *types.FieldType) error {
-	strictCtx := exprstatic.NewExprContext(
+func checkDataForModifyColumn(row []types.Datum, col *table.Column) error {
+	if col.ChangingFieldType == nil {
+		return nil
+	}
+
+	var strictCtx = exprstatic.NewExprContext(
 		exprstatic.WithEvalCtx(
 			exprstatic.NewEvalContext(
 				exprstatic.WithSQLMode(
 					mysql.ModeStrictAllTables | mysql.ModeStrictTransTables))))
-	value, err := table.CastColumnValueToType(strictCtx, data, changingType, false, false)
+
+	data := row[col.Offset]
+	value, err := table.CastColumnValueToType(strictCtx, data, col.ChangingFieldType, false, false)
 	if err != nil {
 		return err
 	}
 
 	// For the case from VARCHAR -> CHAR
-	if changingType.GetType() == mysql.TypeVarString && value.GetString() != data.GetString() {
+	if col.ChangingFieldType.GetType() == mysql.TypeVarString && value.GetString() != data.GetString() {
 		return errors.New("data truncation error during modify column")
 	}
 	return nil
@@ -462,11 +468,8 @@ func (t *TableCommon) updateRecord(sctx table.MutateContext, txn kv.Transaction,
 	for _, col := range t.Columns {
 		var value types.Datum
 		var err error
-
-		if col.ChangingFieldType != nil {
-			if err := checkDataWithModifyColumn(newData[col.Offset], col.ChangingFieldType); err != nil {
-				return err
-			}
+		if err := checkDataForModifyColumn(newData, col); err != nil {
+			return err
 		}
 
 		if col.State == model.StateDeleteOnly || col.State == model.StateDeleteReorganization {
@@ -794,10 +797,8 @@ func (t *TableCommon) addRecord(sctx table.MutateContext, txn kv.Transaction, r 
 	defer memBuffer.Cleanup(sh)
 
 	for _, col := range t.Columns {
-		if col.ChangingFieldType != nil {
-			if err := checkDataWithModifyColumn(r[col.Offset], col.ChangingFieldType); err != nil {
-				return nil, err
-			}
+		if err := checkDataForModifyColumn(r, col); err != nil {
+			return nil, err
 		}
 
 		var value types.Datum


pkg/ddl/modify_column.go
```diff
@@ -244,80 +419,67 @@ func getModifyColumnInfo(
...
+		delayForAsyncCommit()
+	}
+
+	checked, err := checkModifyColumnData(
+		jobCtx.stepCtx, w,
+		dbInfo.Name, tblInfo.Name,
+		oldCol, newCol, checkValueRange)
+
+	if checked {
 		if err != nil {
```
wjhuang2016: The error may be retryable. We'd better distinguish them

pkg/ddl/modify_column.go
```diff
@@ -73,38 +195,44 @@ func (w *worker) onModifyColumn(jobCtx *jobContext, job *model.Job) (ver int64,
...
-
 	if err = checkModifyTypes(oldCol, args.Column, isColumnWithIndex(oldCol.Name.L, tblInfo.Indices)); err != nil {
 		job.State = model.JobStateCancelled
 		return ver, errors.Trace(err)
 	}
 
-	if !needChangeColumnData(oldCol, args.Column) {
-		err := checkColumnAlreadyExists(tblInfo, args)
-		if err != nil {
+	if args.ModifyColumnType == ModifyTypeNoReorg {
```
wjhuang2016: Please split doModifyColumn() to `doModifyColumnNoCheck()` and `doModifyColumnWithCheck()`

pkg/ddl/modify_column.go
```diff
@@ -430,6 +570,187 @@ func adjustForeignKeyChildTableInfoAfterModifyColumn(infoCache *infoschema.InfoC
...
+
+	if checkValueRange {
+		if mysql.IsIntegerType(oldTp) && mysql.IsIntegerType(changingTp) {
+			// Integer conversion
+			conditions = append(conditions, buildCheckRangeForIntegerTypes(oldCol, changingCol))
+		} else if oldTp == mysql.TypeString && (changingTp == mysql.TypeVarchar || changingTp == mysql.TypeString) {
+			// CHAR to VARCHAR/CHAR
+			conditions = append(conditions, fmt.Sprintf("LENGTH(%s) > %d", checkColName, changingCol.FieldType.GetFlen()))
+		}
+		oldCol.ChangingFieldType = &changingCol.FieldType
```
wjhuang2016: This function modifies the column. The name buildCheckSQLFromModifyColumn() is not accurate, as it seems to be a read-only function.

pkg/ddl/column.go
```diff
@@ -511,7 +514,9 @@ func LocateOffsetToMove(currentOffset int, pos *ast.ColumnPosition, tblInfo *mod
 // BuildElements is exported for testing.
 func BuildElements(changingCol *model.ColumnInfo, changingIdxs []*model.IndexInfo) []*meta.Element {
 	elements := make([]*meta.Element, 0, len(changingIdxs)+1)
-	elements = append(elements, &meta.Element{ID: changingCol.ID, TypeKey: meta.ColumnElementKey})
+	if changingCol != nil {
```
fzzf678: When is the `changingCol` empty?

pkg/ddl/column.go
```diff
@@ -998,7 +1016,9 @@ func renameColumnTo(col *model.ColumnInfo, idxInfos []*model.IndexInfo, newName
 }
 
 func updateObjectState(col *model.ColumnInfo, idxs []*model.IndexInfo, state model.SchemaState) {
-	col.State = state
+	if col != nil {
```
fzzf678, line 965: When is the`col` empty?

pkg/ddl/column.go
joechenrh: Just wanna to reuse the existing function, because most part of the logic are same for two code path, see
https://github.com/pingcap/tidb/pull/63465/files/e847856d8b78f041d34f210a252a0dbedf1b7327#diff-d6c304fd538040182d3a53c3f82abff03e074b59725b87ab711779d00f91222eR1087-R1098

pkg/ddl/column.go
joechenrh: ditto

pkg/ddl/modify_column.go
```diff
@@ -1248,7 +1676,7 @@ func GetModifiableColumnJob(
 }
 
 // needChangeColumnData is a strong check to decide whether we need to change the column data.
-// If it returns true, it means we don't need the reorg no matter what the data is..
+// If it returns true, it means we don't need the reorg no matter what the data is.
```
fzzf678, line 1779: This function return `true` means don't need reorg, it is confused with the function name  `needChangeColumnData`

pkg/ddl/modify_column.go
```diff
@@ -167,27 +242,35 @@ func (w *worker) onModifyColumn(jobCtx *jobContext, job *model.Job) (ver int64,
 
 	// For first time running this job, or the job from older version,
 	// we need to fill the ModifyColumnType.
-	if args.ModifyColumnType == modifyTypeNone ||
+	if args.ModifyColumnType == ModifyTypeNone ||
 		args.ModifyColumnType == mysql.TypeNull {
```
fzzf678, line 271: When will the condition `args.ModifyColumnType == mysql.TypeNull` satisfied, i think the `ModifyColumnType` can only be these values. https://github.com/pingcap/tidb/blob/c7212e00d50e16c10c09ba100fe97e003f840891/pkg/ddl/modify_column.go#L59-L72

joechenrh: https://github.com/pingcap/tidb/blob/cf587d3793d7d147132d90eb1850981d3ec41780/pkg/ddl/modify_column.go#L998-L1004
Previously, it can be set as `mysql.TypeNull` here, although I think it's a redundant setting (we can infer this during execution, and it's not a heavy operation). So I reuse this field to store the type. 
To keep the backward compatibilty, for example, the job is submitted from older TiDB, we need to make sure the process logic is same for both TiDB version.
```go
    if !needChangeColumnData(oldCol, args.Column) {
		// It's not NULL->NOTNULL change
		if !isNullToNotNullChange(oldCol, newCol) {
			return ModifyTypeNoReorg
		}
		return ModifyTypeNoReorgWithCheck
	}
	// For backward compatibility
	if args.ModifyColumnType == mysql.TypeNull {
		return ModifyTypeReorg
	}
```
That's why I add special check here. The first if clause is equivalent to old logic and we skip other check for `mysql.TypeNull`.

pkg/ddl/modify_column.go
```diff
@@ -278,6 +372,71 @@ func rollbackModifyColumnJob(
...
+		return ver, errors.Trace(err)
+	}
+	job.FinishTableJob(model.JobStateRollbackDone, model.StateNone, ver, tblInfo)
+	args.IndexIDs = changingIdxIDs
+	args.PartitionIDs = getPartitionIDs(tblInfo)
+	job.FillFinishedArgs(args)
+	return ver, nil
+}
+
+// rollbackModifyColumnJobWithReorg is used to rollback modify-column job which need to reorg the data.
```
fzzf678: Please update this comment

pkg/ddl/modify_column.go
```diff
@@ -596,6 +742,145 @@ func adjustForeignKeyChildTableInfoAfterModifyColumn(infoCache *infoschema.InfoC
...
+	if changingUnsigned {
+		upperBound := types.IntegerUnsignedUpperBound(changingTp)
+		conditions = append(conditions, fmt.Sprintf("NOT (%s >= 0 AND %s <= %d)", columnName, columnName, upperBound))
+	} else {
+		lowerBound := types.IntegerSignedLowerBound(changingTp)
+		upperBound := types.IntegerSignedUpperBound(changingTp)
+		conditions = append(conditions, fmt.Sprintf("NOT (%s >= %d AND %s <= %d)", columnName, lowerBound, columnName, upperBound))
+	}
+
+	return strings.Join(conditions, " OR ")
```
fzzf678: Seems no need to use `strings.Join`, only one condition here 🤔, I think use string as definition of `conditions`  is enough

joechenrh: There may be up to three conditions here, including not null check, range check, trailing space check.

fzzf678: I see `buildCheckRangeForIntegerTypes` checks the integer range, only one string be appended to the `conditions`, did not see `not null check`, `trailing space check` conditions need to append here.
```
	var conditions []string
	if changingUnsigned {
		upperBound := types.IntegerUnsignedUpperBound(changingTp)
		conditions = append(conditions, fmt.Sprintf("NOT (%s >= 0 AND %s <= %d)", columnName, columnName, upperBound))
	} else {
		lowerBound := types.IntegerSignedLowerBound(changingTp)
		upperBound := types.IntegerSignedUpperBound(changingTp)
		conditions = append(conditions, fmt.Sprintf("NOT (%s >= %d AND %s <= %d)", columnName, lowerBound, columnName, upperBound))
	}
```

joechenrh: Ohhh, you are right, I read another function orz.

pkg/ddl/modify_column_test.go
```diff
@@ -594,16 +585,164 @@ func TestModifyColumnWithIndexesWriteConflict(t *testing.T) {
...
+	tk.MustExec("admin check table t")
+
+	// insert should succeed before adding flag, and this will make modify column fail.
+	testfailpoint.EnableCall(t, "github.com/pingcap/tidb/pkg/ddl/afterModifyColumnCheckValue", func() {
+		tk2 := testkit.NewTestKit(t, store)
+		tk2.MustExec("use test")
+		tk2.MustExec("insert into t values (512, 512)")
+	})
+	tk.MustExecToErr("alter table t modify column b tinyint not null")
+	tk.MustExec("admin check table t")
```
fzzf678: What' this for, will `admin check table check` check the reorg correctness

pkg/ddl/modify_column_test.go
```diff
@@ -594,16 +585,164 @@ func TestModifyColumnWithIndexesWriteConflict(t *testing.T) {
...
+	tk.MustExec("alter table t modify column val0 varchar(8) not null;")
 	tk.MustExec("admin check table t;")
 	tk.MustQuery("select * from t order by id;").Check(testkit.Rows(
 		"2 2 2 b",
 		"3 3 3 c",
 		"4 4 4 d"))
 }
+
+func TestMultiSchemaModifyColumnWithSkipReorg(t *testing.T) {
+	store, _ := testkit.CreateMockStoreAndDomain(t)
```
fzzf678: How about using `CreateMockStore` instead

tests/realtikvtest/addindextest4/integration_test.go
```diff
@@ -221,7 +221,7 @@ func TestAddIndexShowAnalyzeProgress(t *testing.T) {
 		require.Contains(t, show, "analyzing")
 		analyzed = true
 	})
-	tk1.MustExec("alter table t modify column b smallint;")
+	tk1.MustExec("alter table t modify column b char(16);")
```
fzzf678, line 224: I have a question, why change these tests, i see only the column type being changed

joechenrh: Because these tests are broken after this PR due to the logic change. Changing type can make the execution logic same as before. 🥲 Do you have any suggestions?

fzzf678: Understood, No suggestions~

pkg/ddl/modify_column_test.go
joechenrh: Add some indexes in https://github.com/pingcap/tidb/pull/63465/commits/90a91afd65eba05a45ed07cbf3627c9d2c89fc45

pkg/ddl/modify_column.go
```diff
@@ -608,6 +774,137 @@ func adjustForeignKeyChildTableInfoAfterModifyColumn(infoCache *infoschema.InfoC
...
+		return true
+	}
+
+	// Check index key part, ref tablecodec.GenIndexKey
+	if oldCol.GetCollate() != changingCol.GetCollate() {
+		return true
+	}
+
+	// Check index value part, ref tablecodec.GenIndexValuePortal
+	return types.NeedRestoredData(&oldCol.FieldType) != types.NeedRestoredData(&changingCol.FieldType)
```
wjhuang2016, line 795: I think NeedRestoredData here should consider all the indexed columns instead of the changing column.

PR #63479, opened 2025-09-11
stats: show index should load stats when cache miss

### LINKED ISSUE #63215: tidb 8.5.3 autoincrement primary key  Cardinality is 0
## Bug Report

tidb 8.5.3 autoincrement primary key  Cardinality is 0

Tristan1900:

Issue Number: close #63215

Problem Summary:

Instead of solely checking stats cache to get the NDV, this PR reads everything from storage for NDV values if there is a cache miss. This solves the problem of missing stats caused by cache eviction or cache not populated. 

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/executor/show.go
```diff
@@ -904,6 +906,89 @@ func (e *ShowExec) fetchShowIndex() error {
...
+
+const selectNDVForColumnsQuery = "SELECT hist_id, distinct_count FROM mysql.stats_histograms WHERE table_id = %? AND is_index = 0 AND hist_id IN (%?)"
+
+// fetchColumnNDVs loads NDVs for the given column IDs from storage in one round-trip without mutating stats cache
+func fetchColumnNDVs(ctx sessionctx.Context, tableID int64, columnIDs []string) (map[int64]int64, error) {
+	ndvByColID := make(map[int64]int64, len(columnIDs))
+	if len(columnIDs) == 0 {
+		return ndvByColID, nil
+	}
+	rows, _, err := statsutil.ExecRows(ctx, selectNDVForColumnsQuery, tableID, columnIDs)
```
qw4990: It seems unsafe to run another query with the same `sessionctx` of the current query and the current query has not finished yet.
Better to grab a new session from the session pool to run this query.

pkg/executor/show.go
```diff
@@ -904,6 +906,89 @@ func (e *ShowExec) fetchShowIndex() error {
...
+				if _, exists := ndvByColID[k]; !exists {
+					ndvByColID[k] = v
+				}
+			}
+		}
+	}
+	return ndvByColID
+}
+
+const selectNDVForColumnsQuery = "SELECT hist_id, distinct_count FROM mysql.stats_histograms WHERE table_id = %? AND is_index = 0 AND hist_id IN (%?)"
```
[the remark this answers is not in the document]
Tristan1900: %? is the correct usage in TiDB code base

pkg/executor/show.go
Tristan1900: thanks, that makes sense!

pkg/executor/show_test.go
```diff
@@ -162,3 +164,41 @@ func TestShowSessionStates(t *testing.T) {
...
+	require.NotEqual(t, "0", ndvs["PRIMARY:a"]) // pk
+	require.NotEqual(t, "0", ndvs["ib:b"])      // first column of composite index
+	require.NotEqual(t, "0", ndvs["ib:c"])      // second column of composite index
+
+	// Ensure cache remains pseudo (not populated) after SHOW INDEX
+	is := dom.InfoSchema()
+	tbl, err := is.TableByName(context.Background(), ast.NewCIStr("test"), ast.NewCIStr("t"))
+	require.NoError(t, err)
+	statsTbl := dom.StatsHandle().GetPhysicalTableStats(tbl.Meta().ID, tbl.Meta())
+	require.True(t, statsTbl.Pseudo)
```
wddevries: Will the system automatically reanalyze the table in the background?  If it can, could that make this test flakey?

pkg/executor/show.go
```diff
@@ -904,6 +906,90 @@ func (e *ShowExec) fetchShowIndex() error {
...
+			}
+			for _, ic := range idxInfo.Columns {
+				tblCol := tb.Meta().Columns[ic.Offset]
+				if tblCol.Hidden {
+					continue
+				}
+				needed[tblCol.ID] = struct{}{}
+			}
+		}
+	} else {
```
wddevries: This section is a lot like the section below.

pkg/executor/show.go
```diff
@@ -904,6 +906,90 @@ func (e *ShowExec) fetchShowIndex() error {
 	return nil
 }
 
+// prefetchColumnNDVs prepares a map of colID->NDV using cache-first and storage-fallback without mutating the stats cache
+func (e *ShowExec) prefetchColumnNDVs(tb table.Table, statsTbl *statistics.Table, pkCol *table.Column) map[int64]int64 {
+	ndvByColID := make(map[int64]int64)
+	needed := make(map[int64]struct{})
```
wddevries: Wouldn't it be simpler to build 'colIDs' directly instead of creating 'needed' and then looping over it afterwards?

pkg/executor/show.go
```diff
@@ -904,6 +906,90 @@ func (e *ShowExec) fetchShowIndex() error {
...
+	if err != nil {
+		return ndvByColID, err
+	}
+	for _, r := range rows {
+		histID := r.GetInt64(0)
+		distinctCnt := r.GetInt64(1)
+		ndvByColID[histID] = distinctCnt
+	}
+	return ndvByColID, nil
+}
```
wddevries, line 981: Are stats not loaded in some sort of transactional way in TiDB?  There must be a point where stats are fetched normally for queries, where the statistics for this table could be updated.

wddevries (commented):
If the statistics can be fetched directly from storage instead of utilizing the statistics cache, then 'SHOW INDEX' may not accurately reflect the TiDB node's current state.  This could be unexpected behavior from a user's perspective, and it may hinder introspection into the statistics system making the diagnosis of issues more difficult.

If the goal was just to ignore that statistics cache and show the statistics stored on the cluster directly, then the command could now potentially mix stale statistics from the statistics cache with newer statistics from storage.

With this patch, 'SHOW INDEX' will no longer reflect the status of the statistics cache on the node, and neither will it reflect the statistics stored on the cluster.

Tristan1900:
> If the statistics can be fetched directly from storage instead of utilizing the statistics cache, then 'SHOW INDEX' may not accurately reflect the TiDB node's current state. This could be unexpected behavior from a user's perspective, and it may hinder introspection into the statistics system making the diagnosis of issues more difficult.
> 
> If the goal was just to ignore that statistics cache and show the statistics stored on the cluster directly, then the command could now potentially mix stale statistics from the statistics cache with newer statistics from storage.
> 
> With this patch, 'SHOW INDEX' will no longer reflect the status of the statistics cache on the node, and neither will it reflect the statistics stored on the cluster.

Thanks for the detailed review!

Please correct me if I'm wrong, my understanding is that show index is to reflect the states for the TiDB cluster, not the single TiDB node cache. So the goal is to read through cache to have a more consistent view, I'm a bit concerned that just reading from storage for every call can lead to performance issues and that's why I take the current approach.

As for the inconsistency concern, analyze table will update the stats cache once it's done, so the result of show index should be consistent fetching from either cache or storage. Yes, there might be a brief inconsistency when show index is called while analyze table running, but I think that inconsistency can also happen for just fetching from cache (current behavior) since analyze table doesn't do atomic updates to the cache. 

Let me know if that makes sense and happy to discuss more!

pkg/executor/show_test.go
Tristan1900: good call, I will disable auto analyze just to be safe. In practice auto analyze won't trigger for small tables less than 1000 rows such as this test.

pkg/executor/show.go
Tristan1900: yes you are right! I don't remember why I did this...

pkg/executor/show.go
Tristan1900: yes, it's in TableStatsFromStorage, however it's sort of difficult to reuse since that method is fairly convoluted already and I don't want to pass in more parameters to overwhelm it.

0xPoe:
> If the goal was just to ignore that statistics cache and show the statistics stored on the cluster directly, then the command could now potentially mix stale statistics from the statistics cache with newer statistics from storage.
> 
> With this patch, 'SHOW INDEX' will no longer reflect the status of the statistics cache on the node, and neither will it reflect the statistics stored on the cluster.

Very nice observation. I checked the MySQL implementation, and if I understand correctly, it first tries to retrieve the value from the hidden view. If it does not exist or has expired, it will then read it from the system table, store it in the cache, and use that as the result.

Back to this issue, I think the main problem is that for most SHOW STATS commands, we only use the values from the current statistics cache. Since some indexes have not been used in queries, they are not loaded into memory, which causes users to see a cardinality of 0. I believe this behavior was introduced [in this PR](https://github.com/pingcap/tidb/pull/47496). Before this PR, we always loaded the histogram metadata for all indexes, so this issue did not occur.

Your observation also highlights an issue that even within the same cluster, different instances may show different results with the SHOW STATS commands. I think it is acceptable and reasonable for these commands to reflect the cache of the current instance.

However, for the SHOW INDEX command, I am not sure what the expected behavior should be. Most values come directly from index metadata, while only the cardinality is retrieved from the statistics layer.

wddevries:
> > If the statistics can be fetched directly from storage instead of utilizing the statistics cache, then 'SHOW INDEX' may not accurately reflect the TiDB node's current state. This could be unexpected behavior from a user's perspective, and it may hinder introspection into the statistics system making the diagnosis of issues more difficult.
> > If the goal was just to ignore that statistics cache and show the statistics stored on the cluster directly, then the command could now potentially mix stale statistics from the statistics cache with newer statistics from storage.
> > With this patch, 'SHOW INDEX' will no longer reflect the status of the statistics cache on the node, and neither will it reflect the statistics stored on the cluster.
> 
> Thanks for the detailed review!
> 
> Please correct me if I'm wrong, my understanding is that show index is to reflect the states for the TiDB cluster, not the single TiDB node cache. So the goal is to read through cache to have a more consistent view, I'm a bit concerned that just reading from storage for every call can lead to performance issues and that's why I take the current approach.
> 
> As for the inconsistency concern, analyze table will update the stats cache once it's done, so the result of show index should be consistent fetching from either cache or storage. Yes, there might be a brief inconsistency when show index is called while analyze table running, but I think that inconsistency can also happen for just fetching from cache (current behavior) since analyze table doesn't do atomic updates to the cache.
> 
> Let me know if that makes sense and happy to discuss more!

With this pull-request, the data returned by SHOW INDEX may either come from the cache or may be read from storage directly.  You will not be able to use the command to know exactly what data is in the cache, and you will not be able to use the command to know exactly what data is stored on the cluster.   Outdated data from cache could be mixed with current data from storage. 

It would be better to have one command to return the data that is currently in the cache (which SHOW INDEX currently does), and a separate command that returns data only from storage.  The latter could be an option on SHOW INDEX.

Tristan1900:
> > > If the statistics can be fetched directly from storage instead of utilizing the statistics cache, then 'SHOW INDEX' may not accurately reflect the TiDB node's current state. This could be unexpected behavior from a user's perspective, and it may hinder introspection into the statistics system making the diagnosis of issues more difficult.
> > > If the goal was just to ignore that statistics cache and show the statistics stored on the cluster directly, then the command could now potentially mix stale statistics from the statistics cache with newer statistics from storage.
> > > With this patch, 'SHOW INDEX' will no longer reflect the status of the statistics cache on the node, and neither will it reflect the statistics stored on the cluster.
> > 
> > 
> > Thanks for the detailed review!
> > Please correct me if I'm wrong, my understanding is that show index is to reflect the states for the TiDB cluster, not the single TiDB node cache. So the goal is to read through cache to have a more consistent view, I'm a bit concerned that just reading from storage for every call can lead to performance issues and that's why I take the current approach.
> > As for the inconsistency concern, analyze table will update the stats cache once it's done, so the result of show index should be consistent fetching from either cache or storage. Yes, there might be a brief inconsistency when show index is called while analyze table running, but I think that inconsistency can also happen for just fetching from cache (current behavior) since analyze table doesn't do atomic updates to the cache.
> > Let me know if that makes sense and happy to discuss more!
> 
> With this pull-request, the data returned by SHOW INDEX may either come from the cache or may be read from storage directly. You will not be able to use the command to know exactly what data is in the cache, and you will not be able to use the command to know exactly what data is stored on the cluster. Outdated data from cache could be mixed with current data from storage.
> 
> It would be better to have one command to return the data that is currently in the cache (which SHOW INDEX currently does), and a separate command that returns data only from storage. The latter could be an option on SHOW INDEX.

got it, that makes sense, I will see what I can do. Thanks!

mjonss:
@wddevries I change the logic, so that it would use the cardinality (ndv) from the cache when it exists, otherwise it would set it to NULL. Do you agree on that change? It also simplified the code a bit.

mjonss (approved):
LGTM, So it will first use the cache if available, if not, then fetch from TiKV, and last resort just set it to 0.

PR #63519, opened 2025-09-15
stats: reduce updating col predicate usage table

### LINKED ISSUE #63426: The update of table column_stats_usage is too slow in huge TiDB cluster
## Enhancement

The goal of enhancement: 
Resolves the write conflict issue of column_stats_usage in large-scale clusters, which causes the execution efficiency of the update column_stats_usage statement to be low and slow.

## Environment required to reproduce the problem
version: v8.5.2

1. 100 TiDB instance

Image

2. Heavy query traffic

Image

3. Phenomenon

The slow query for insert into mysql.column_stats_usage

Image

Tristan1900:

Issue Number: close #63426 

Problem Summary:

We noticed updating to cols usage query has been taking 10s seconds when scale TIDB nodes to 100. It was mainly due to lock contention that every node trying to update the same set of cols. 

By doing some experiments locally, it is noticed that the batch size might contribute a lot to the lock contention. By tuning down the batch size, it should mitigate the problem.

This PR also tries to not update the cols if it's only timestamp change, we decide to update the col timestamp every 12 hours, that should also reduce the update time and lock contention.

Perf with different batch size:
```
// batch size 1024
DumpColStatsUsage concurrent: goroutines=10 tables=100 cols/table=1000 per-g: avg=7.486120933s min=7.300108708s max=7.718235625s; updated=100/100000; batches=980 per-batch: avg=75.721586ms min=11.163959ms max=927.45675ms

// batch size 2048
DumpColStatsUsage concurrent: goroutines=10 tables=100 cols/table=1000 per-g: avg=6.412777945s min=6.262234292s max=6.551383291s; updated=100/100000; batches=490 per-batch: avg=128.168868ms min=29.5895ms max=538.059583ms

// batch size 4096
DumpColStatsUsage concurrent: goroutines=10 tables=100 cols/table=1000 per-g: avg=6.889603845s min=6.633772208s max=7.037077708s; updated=100/100000; batches=250 per-batch: avg=273.157262ms min=37.115208ms max=1.215064s

// batch size 8192
DumpColStatsUsage concurrent: goroutines=10 tables=100 cols/table=1000 per-g: avg=8.217947175s min=7.566108792s max=8.744043541s; updated=100/100000; batches=130 per-batch: avg=626.931753ms min=28.640833ms max=3.102622292s
```
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/handle/usage/session_stats_collect.go
+++ b/pkg/statistics/handle/usage/session_stats_collect.go
@@ -46,9 +46,17 @@ var (
 	// dumpStatsMaxDuration is the max duration since last update.
 	dumpStatsMaxDuration = 5 * time.Minute
 
+	// colStatsUsageUpdateInterval is the minimum interval to update last_used_at when it already exists.
+	colStatsUsageUpdateInterval = 12 * time.Hour
+
 	// batchInsertSize is the batch size used by internal SQL to insert values to stats usage table.
-	batchInsertSize = 8192
+	batchInsertSize = 2048
 )
+
+// TimeCostRecorder can collect per-batch timings when provided (e.g., in tests).
+type TimeCostRecorder interface {
+	Record(duration time.Duration)
+}
 
 // needDumpStatsDelta checks whether to dump stats delta.
 // 1. If the table doesn't exist or is a mem table or system table, then return false.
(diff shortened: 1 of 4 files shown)


pkg/statistics/handle/usage/session_stats_collect.go
```diff
@@ -46,10 +46,18 @@ var (
 	// dumpStatsMaxDuration is the max duration since last update.
 	dumpStatsMaxDuration = 5 * time.Minute
 
+	// colStatsUsageUpdateInterval is the minimum interval to update last_used_at when it already exists.
+	colStatsUsageUpdateInterval = 12 * time.Hour
```
0xPoe: Maybe we should make the name more explicit to indicate that it is only for updating the last used time.

pkg/statistics/handle/usage/session_stats_collect.go
```diff
@@ -46,10 +46,18 @@ var (
...
+
 	// batchInsertSize is the batch size used by internal SQL to insert values to stats usage table.
-	batchInsertSize = 8192
+	batchInsertSize = 2048
 )
 
+// TimeCostRecorder can collect per-batch timings when provided (e.g., in tests).
+type TimeCostRecorder interface {
```
0xPoe:
```suggestion
type TimeCostRecorderForTest interface {
```

Tristan1900:
> What do you think about using [this random offset mechanism](https://github.com/0xPoe/tidb/blob/master/pkg/domain/domain.go#L2238) to further mitigate this issue?
> 
> Also, does it make sense to lock the records first with SELECT FOR UPDATE to reduce the overhead of retries and query rollbacks?
> 
> Thank you for working on this!

Thanks for the review! 

The jitter makes sense and I have added in the commit.

For the SELECT FOR UPDATE, I feel like the current INSERT ... ON DUPLICATE KEY UPDATE pattern might be better since it doesn't hold lock for very long time, so it might actually reduce contention. Also it's one round trip where SELECT FOR UPDATE needs two round trips, which might add the overall execution time.

I feel like the main issue is that our previous batch is big so it has greater chance to have contention, I am leaning towards staying with the current IODKU pattern and only to change if all other methods fail. Let me know your thoughts and happy to discuss more!

0xPoe:
> I feel like the main issue is that our previous batch is big so it has greater chance to have contention,

I’m not sure what caused the contention issue here. I used a very large number for the dump stats delta, and according to the same cluster metrics, it seems to be working well. I’m also not sure why the dump stats delta process performs well. It could be because we added a random offset for each instance so that they trigger the process at different times, or because the SELECT FOR UPDATE statement reduces retry overhead as it’s just a simple lock operation. In any case, I’m fine with moving this PR forward as it is. Alternatively, we could reach out to the Txn team to confirm which approach is more suitable for this scenario.

Tristan1900:
> > I feel like the main issue is that our previous batch is big so it has greater chance to have contention,
> 
> I’m not sure what caused the contention issue here. I used a very large number for the dump stats delta, and according to the same cluster metrics, it seems to be working well. I’m also not sure why the dump stats delta process performs well. It could be because we added a random offset for each instance so that they trigger the process at different times, or because the SELECT FOR UPDATE statement reduces retry overhead as it’s just a simple lock operation. In any case, I’m fine with moving this PR forward as it is. Alternatively, we could reach out to the Txn team to confirm which approach is more suitable for this scenario.

Got it, thanks for the explantation. Sounds good, let me check with the Txn team

0xPoe (approved):
Thanks! :shipit:

D3Hunter:
/approve

domain part lgtm

henrybw (approved):
LGTM

terry1purcell (approved):

PR #63521, opened 2025-09-15
planner: fix the USE_INDEX hint for tici index

winoros:

Issue Number: ref https://github.com/pingcap/tidb/issues/61185

Problem Summary:

We didn't implement the expected behavior when the index hint meets the tici index.

This pr will
- Using an index hint will not cause SQL to throw an error.
- If the use index hint is used on the tici index without any filter. We'll throw warning and use table scan.
- If the use index hint is used on the normal index with fts_match functions. We'll throwing warning and use the corresponding tici index.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/casetest/tici/testdata/tici_index_suite_out.json
+++ b/pkg/planner/core/casetest/tici/testdata/tici_index_suite_out.json
@@ -173,5 +173,31 @@
         ]
       }
     ]
+  },
+  {
+    "Name": "TestTiCIWithIndexHintCases",
+    "Cases": [
+      {
+        "SQL": "explain format='brief' select * from t1 use index(idx_title)",
+        "Plan": [
+          "TableReader 10000.00 root  data:TableFullScan",
+          "└─TableFullScan 10000.00 cop[tikv] table:t1 keep order:false, stats:pseudo"
+        ],
+        "Warn": [
+          "[planner:3126]Hint USE_INDEX is ignored as conflicting/duplicated."
+        ]
+      },
+      {
+        "SQL": "explain format='brief' select * from t1 use index(idx_field) where fts_match_word('hello', title)",
+        "Plan": [
+          "IndexLookUp 1000.00 root  ",
+          "├─IndexRangeScan(Build) 1000.00 batchCop[tiflash] table:t1, index:idx_title(title) range:[NULL,+inf], search func:fts_match_word(\"hello\", test.t1.title), keep order:false, stats:pseudo",
+          "└─TableRowIDScan(Probe) 1000.00 cop[tikv] table:t1 keep order:false, stats:pseudo"
+        ],
+        "Warn": [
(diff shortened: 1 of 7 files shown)


pkg/planner/core/casetest/tici/testdata/tici_index_suite_out.json
```diff
@@ -263,5 +263,31 @@
...
+    "Name": "TestTiCIWithIndexHintCases",
+    "Cases": [
+      {
+        "SQL": "explain format='brief' select * from t1 use index(idx_title)",
+        "Plan": [
+          "TableReader 10000.00 root  data:TableFullScan",
+          "└─TableFullScan 10000.00 cop[tikv] table:t1 keep order:false, stats:pseudo"
+        ],
+        "Warn": [
+          "[planner:3126]Hint USE_INDEX is ignored as conflicting/duplicated."
```
AilinKid, line 277: maybe the msg is not clear here, user can not notice that they lack the column index filter to use that index.

pkg/planner/core/planbuilder.go
```diff
@@ -1286,10 +1287,23 @@ func getPossibleAccessPaths(ctx base.PlanContext, tableHints *hint.PlanHints, in
...
 		}
 	}
+	var ticiIndexPaths []*util.AccessPath
+	if hasTiCIIndex {
+		// FTS_MATCH_XXX can only be executed in TiCI engine.
+		// So we need to store it here and try to add it later if the USE_INDEX hint delete any of them.
+		// The removal of the unhinted TiCI index paths will be done after we dicide the availability of each index.
+		ticiIndexPaths = make([]*util.AccessPath, 0, len(publicPaths))
+		for _, path := range publicPaths {
+			if path.Index != nil && path.Index.IsFulltextIndexOnTiCI() {
```
AilinKid: why don't we add it in previous loop, which i mean L1290, when we see index is TiCIIndex, is must be a public path, so just collecting it into ticiIndexPaths in a same loop is profitable.

pkg/planner/core/stats.go
```diff
@@ -130,6 +130,9 @@ func deriveStats4DataSource(lp base.LogicalPlan) (*property.StatsInfo, bool, err
 	for i, expr := range ds.PushedDownConds {
 		ds.PushedDownConds[i] = expression.EliminateNoPrecisionLossCast(exprCtx, expr)
 	}
+	// Cleanup the unused TiCI indexes.
+	// They are not suitable for normal read.
+	ds.CleanUnusedTiCIIndexes()
```
AilinKid, line 135: clean unused tici index path seems nothing todo with derived stats, it's more like a rule pruning.

AilinKid: Tight coupling with stats driving func may a little bit complicated

PR #63599, opened 2025-09-18
next-gen: fix tablesample test and null keyspace test

tangenta:

Issue Number: ref https://github.com/pingcap/tidb/issues/61702

Problem Summary:

- Skip null keyspace test in next-gen.
- Set tablesample end key to table end key if the length is zero.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/sample.go
+++ b/pkg/executor/sample.go
@@ -15,6 +15,7 @@
 package executor
 
 import (
+	"bytes"
 	"context"
 	"slices"
 
@@ -309,6 +310,7 @@ func (s *tableRegionSampler) scanFirstKVForEachRange(ranges []kv.KeyRange,
 			kvChan:      make(chan *sampleKV),
 			snapshot:    snap,
 			ranges:      ranges,
+			prefix:      s.table.RecordPrefix(),
 		}
 		go fetchers[i].run()
 	}
@@ -347,6 +349,7 @@ type sampleFetcher struct {
 	err         error
 	snapshot    kv.Snapshot
 	ranges      []kv.KeyRange
+	prefix      []byte
 }
 
 func (s *sampleFetcher) run() {
@@ -362,7 +365,7 @@ func (s *sampleFetcher) run() {
 		}
 		hasValue := false
 		for it.Valid() {
-			if !tablecodec.IsRecordKey(it.Key()) {
+			if !bytes.HasPrefix(it.Key(), s.prefix) {
 				if err = it.Next(); err != nil {
 					s.err = err
 					return
--- a/pkg/executor/sample.go
+++ b/pkg/executor/sample.go
@@ -15,7 +15,6 @@
 package executor
 
 import (
-	"bytes"
 	"context"
 	"slices"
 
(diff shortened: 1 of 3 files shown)


pkg/session/bootstrap_test.go
```diff
@@ -2591,7 +2591,9 @@ func makeStore(t *testing.T, keyspaceMeta *keyspacepb.KeyspaceMeta, isHasPrefix
 			mockstore.WithStoreType(mockstore.EmbedUnistore),
 		)
 	} else {
-		store, err = mockstore.NewMockStore(mockstore.WithStoreType(mockstore.EmbedUnistore))
+		store, err = mockstore.NewMockStore(
+			mockstore.WithStoreType(mockstore.EmbedUnistore),
+			mockstore.WithKeyspaceMeta(nil))
```
D3Hunter: why set it to nil, what about nextgen?

pkg/executor/sample.go
```diff
@@ -362,7 +365,7 @@ func (s *sampleFetcher) run() {
 		}
 		hasValue := false
 		for it.Valid() {
-			if !tablecodec.IsRecordKey(it.Key()) {
+			if !bytes.HasPrefix(it.Key(), s.prefix) {
```
D3Hunter: the key contains keyspace ID part?

pkg/session/bootstrap_test.go
tangenta: The purpose of this test to is test null keyspace. This option is to revert the next-gen's setting, keeping keyspace meta nil.

D3Hunter: maybe skip it for nextgen? nextgen don't have null keyspace

pkg/executor/sample.go
tangenta: No, but it may contain unrelated rows from different tables in the same region..

D3Hunter: sounds like a bug, or only happens in the test?

tangenta: It should be a bug, I found it read system table keys ":4000" under next-gen test.

pkg/session/bootstrap_test.go
tangenta: I think it is fine to keep this test? It only affect tests

D3Hunter: maybe skip it completely, let it run only in classic kernel

then i guess we can revert the change in `pkg/store/mockstore/mockstore.go` too

pkg/session/bootstrap_test.go
```diff
@@ -2578,6 +2578,9 @@ func TestKeyspaceEtcdNamespace(t *testing.T) {
 }
 
 func TestNullKeyspaceEtcdNamespace(t *testing.T) {
+	if kerneltype.IsNextGen() {
+		t.Skip("keyspace is required in next-gen kernel")
```
D3Hunter:
```suggestion
		t.Skip("next-gen kernel doesn't have the NULL keyspace concept")
```

D3Hunter (approved):

fzzf678 (approved):

PR #63615, opened 2025-09-19
variable: tweak set instance for all instance variables 

### LINKED ISSUE #63614: Fix `SET INSTANCE` for both old and new instance variables

xhebox:

Issue Number: close #63614

Problem Summary: Previously, `SET INSTANCE` only works for old instance variables, and will only work for new instance variables if they are set from config file. Now it works for all variables with `ScopeInstance`.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -1412,6 +1412,21 @@ func (s *session) SetGlobalSysVarOnly(ctx context.Context, name string, value st
 	return s.replaceGlobalVariablesTableValue(ctx, sv.Name, value, updateLocal)
 }
 
+// SetInstanceSysVar implements InstanceVarAccessor.SetInstanceSysVar interface.
+func (s *session) SetInstanceSysVar(ctx context.Context, name string, value string) (err error) {
+	sv := variable.GetSysVar(name)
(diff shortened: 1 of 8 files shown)


pkg/session/session.go
```diff
@@ -1386,9 +1385,6 @@ func (s *session) SetGlobalSysVar(ctx context.Context, name string, value string
 	if err = sv.SetGlobalFromHook(ctx, s.sessionVars, value, false); err != nil {
 		return err
 	}
-	if sv.HasInstanceScope() { // skip for INSTANCE scope
```
xhebox, line 1390: This is removed. It is mainly used by `set.go` to bypass some `atomic global variables` that only need to `SetGlobalFromHook` but not setting the global variable table. Since now we have `v.IsInstance`, we can just call `SetInstanceSysVar`.

yudongusa (approved):

PR #54514, opened 2024-07-09
statistics: avoid using infoschema when to init stats

hawkingrei:
### What problem does this PR solve?

Issue Number: close #55851

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #63626, opened 2025-09-19
stats: remove ColAndIdxExistenceMap checked flag

### LINKED ISSUE #60316: Remove the unnecessary checked field from ColAndIdxExistenceMap
## Enhancement
https://github.com/pingcap/tidb/blob/aacd24317bdd67d09c7155474cc02dffd2d19c4b/pkg/statistics/table.go#L89

We introduced this field to verify whether the column and index have been checked when updating the stats item.

We used it in two places:
1. https://github.com/pingcap/tidb/blob/aacd24317bdd67d09c7155474cc02dffd2d19c4b/pkg/statistics/handle/syncload/stats_syncload.go#L589 For synchronous loading, it will be checked here, but this code path is rarely reachable since the stats update always sets this field to checked.
2. https://github.com/pingcap/tidb/blob/aacd24317bdd67d09c7155474cc02dffd2d19c4b/pkg/statistics/table.go#L866 For this usage, we don't care if the index is checked, the only thing we need to know is that this index has been analyze or not.

Tristan1900:

Issue Number: close #60316

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/handle/bootstrap.go
+++ b/pkg/statistics/handle/bootstrap.go
@@ -199,7 +199,6 @@ func (h *Handle) initStatsHistograms4Chunk(is infoschema.InfoSchema, cache stats
 		tblID, statsVer := row.GetInt64(0), row.GetInt64(8)
 		if table == nil || table.PhysicalID != tblID {
 			if table != nil {
-				table.ColAndIdxExistenceMap.SetChecked()
 				cache.Put(table.PhysicalID, table) // put this table in the cache because all statstics of the table have been read.
 			}
 			var ok bool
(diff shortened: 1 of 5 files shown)


0xPoe:
/hold

Please make sure @winoros approves this PR. 🐐

winoros:
The `check` is introduced initially in https://github.com/pingcap/tidb/pull/54514/files#diff-3adb2615f63129b3afd7655cdade72b806af0e0738c6b59571e73951c62d3234L175.
Before that change, we'll remove the deleted columns/indexes by checking the table info.
We had this `check` field to do the same thing lazily.

But this can be done by stats' GC. And it will only bring one more element for the `map[int64]bool` before the stats is GCed. I think it's acceptable, and we can remove this `check` field.

PR #63723, opened 2025-09-24
executor: incremental probing for IndexHashJoin

### LINKED ISSUE #63303: Incremental probing for `IndexJoin` and `IndexHashJoin` to reduce memory usage
## Enhancement

https://github.com/pingcap/tidb/blob/22580a3cc83bee083224806debbddcc2d6b3cfd4/pkg/executor/join/index_lookup_join.go#L731-L748

In the current probing process of `IndexJoin` and `IndexHashJoin`, all matched inner rows for the join keys of an outer chunk are fetched at once. This can lead to excessive memory usage and potentially cause the query to exceed the memory limit.

To address this, the probing process should fetch and consume inner rows incrementally, in a streaming fashion similar to hash join.

ChangRui-Ryan:

Issue Number: close #63303 

Problem Summary:

For `INNER JOIN` and `LEFT JOIN` that do not require order preservation, when looking up data from the inner table using a single batch of outer table data: if more than 4096 rows have been collected, the process will return early to complete the join for these 4096 rows first, then continue iterating to look up data from the inner table for subsequent join operations.
### Simple benchmark
Under the TPCH 1G scale and tested on a MacBook Pro with an M2 Pro CPU, for the following SQL statement:

```sql
select /*+ INL_HASH_JOIN(lineitem) */ * 
from nation 
inner join lineitem 
on n_regionkey = l_linenumber;
```

The memory consumption of the IndexHashJoin operator decreased significantly from 1.49G (before optimization) to only 1.4M (after optimization). Meanwhile, the execution time was reduced from approximately 22 seconds to around 18 seconds.
### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/join/index_lookup_hash_join.go
+++ b/pkg/executor/join/index_lookup_hash_join.go
@@ -29,6 +29,8 @@ import (
 	"github.com/pingcap/failpoint"
 	"github.com/pingcap/tidb/pkg/executor/internal/exec"
 	"github.com/pingcap/tidb/pkg/expression"
+	"github.com/pingcap/tidb/pkg/parser/terror"
+	"github.com/pingcap/tidb/pkg/planner/core/base"
 	"github.com/pingcap/tidb/pkg/types"
 	"github.com/pingcap/tidb/pkg/util"
 	"github.com/pingcap/tidb/pkg/util/channel"
@@ -47,6 +49,8 @@ import (
 //     Otherwise, there will be at most `concurrency` resource chunks throughout
 //     the execution of IndexNestedLoopHashJoin.
 const numResChkHold = 4
+
+const maxRowsPerFetch = 4096
 
 // IndexNestedLoopHashJoin employs one outer worker and N inner workers to
 // execute concurrently. The output order is not promised.
@@ -466,6 +470,14 @@ func (e *IndexNestedLoopHashJoin) newOuterWorker(innerCh chan *indexHashJoinTask
 	return ow
 }
 
+func (e *IndexNestedLoopHashJoin) supportIncrementalLookUp() bool {
+	return !e.KeepOuterOrder &&
+		(JoinerType(e.Joiners[0]) == base.InnerJoin ||
+			JoinerType(e.Joiners[0]) == base.LeftOuterJoin ||
+			JoinerType(e.Joiners[0]) == base.RightOuterJoin ||
+			JoinerType(e.Joiners[0]) == base.AntiSemiJoin)
+}
+
 func (e *IndexNestedLoopHashJoin) newInnerWorker(taskCh chan *indexHashJoinTask, workerID int) *indexHashJoinInnerWorker {
 	// Since multiple inner workers run concurrently, we should copy join's IndexRanges for every worker to avoid data race.
 	copiedRanges := make([]*ranger.Range, 0, len(e.IndexRanges.Range()))
@@ -476,12 +488,19 @@ func (e *IndexNestedLoopHashJoin) newInnerWorker(taskCh chan *indexHashJoinTask,
 	if e.stats != nil {
 		innerStats = &e.stats.innerWorker
 	}
+	var maxRows int
+	// If the joiner supports incremental look up, we can fetch inner results in batches,
+	// retrieving partial results multiple times until all are fetched.
+	// Otherwise, we fetch all inner results in one batch.
+	if e.supportIncrementalLookUp() {
+		maxRows = maxRowsPerFetch
+	}
 	iw := &indexHashJoinInnerWorker{
 		innerWorker: innerWorker{
 			InnerCtx:      e.InnerCtx,
 			outerCtx:      e.OuterCtx,
 			ctx:           e.Ctx(),
-			executorChk:   e.AllocPool.Alloc(e.InnerCtx.RowTypes, e.InitCap(), e.MaxChunkSize()),
+			maxFetchSize:  maxRows,
 			indexRanges:   copiedRanges,
 			keyOff2IdxOff: e.KeyOff2IdxOff,
 			stats:         innerStats,
@@ -650,14 +669,6 @@ func (iw *indexHashJoinInnerWorker) buildHashTableForOuterResult(task *indexHash
 	}
 }
 
-func (iw *indexHashJoinInnerWorker) fetchInnerResults(ctx context.Context, task *lookUpJoinTask) error {
-	lookUpContents, err := iw.constructLookupContent(task)
-	if err != nil {
-		return err
-	}
-	return iw.innerWorker.fetchInnerResults(ctx, task, lookUpContents)
-}
-
 func (iw *indexHashJoinInnerWorker) handleHashJoinInnerWorkerPanic(resultCh chan *indexHashJoinResult, err error) {
 	defer func() {
 		iw.wg.Done()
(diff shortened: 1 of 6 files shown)


pkg/planner/core/base/plan_base.go
```diff
@@ -336,6 +336,11 @@ func (tp JoinType) IsInnerJoin() bool {
 	return tp == InnerJoin
 }
 
+// SupportIncrementalLookUp returns if this joiner supports incremental look up.
+func (tp JoinType) SupportIncrementalLookUp() bool {
```
windtalker: I think it's better to define this function in `IndexHashJoin`

pkg/executor/join/index_lookup_join.go
```diff
@@ -554,7 +555,7 @@ func (iw *innerWorker) handleTask(ctx context.Context, task *lookUpJoinTask) err
 	if err != nil {
 		return err
 	}
-	err = iw.fetchInnerResults(ctx, task, lookUpContents)
+	err = iw.fetchInnerResults(ctx, task, lookUpContents, -1)
```
windtalker: why `indexLookupJoin` is changed?

pkg/executor/join/index_lookup_hash_join.go
```diff
@@ -717,7 +711,16 @@ func (iw *indexHashJoinInnerWorker) handleTask(ctx context.Context, task *indexH
 			iw.handleHashJoinInnerWorkerPanic(resultCh, err)
 		},
 	)
-	err = iw.fetchInnerResults(ctx, task.lookUpJoinTask)
+	lookUpContents, err := iw.constructLookupContent(task.lookUpJoinTask)
+	if err == nil {
+		// If the joiner supports incremental look up, we can fetch inner results in batches,
+		// retrieving partial results multiple times until all are fetched.
+		if JoinerType(iw.joiner).SupportIncrementalLookUp() && !task.keepOuterOrder {
```
windtalker: how about define this `maxRowsPerFetch` in `Open()`, then there is no need to use this if branch here?

pkg/executor/join/index_lookup_hash_join.go
```diff
@@ -731,7 +734,20 @@ func (iw *indexHashJoinInnerWorker) handleTask(ctx context.Context, task *indexH
...
+		if err != nil {
+			return err
+		}
+		// innerExec not finished, we need to fetch inner results and do join again.
+		if task.innerExec != nil {
+			err = iw.innerWorker.fetchInnerResults(ctx, task.lookUpJoinTask, lookUpContents, maxRowsPerFetch)
+			if err != nil {
+				return err
+			}
+			goto JoinNext
```
windtalker: how about use the `for` loop and break if `task.innerExec` is nil instead of using the `goto` statement?

pkg/executor/join/index_lookup_hash_join.go
```diff
@@ -745,6 +761,10 @@ func (iw *indexHashJoinInnerWorker) doJoinUnordered(ctx context.Context, task *i
 			return joinResult.err
 		}
 	}
+	if task.innerExec != nil {
```
windtalker, line 788: it actually a little tricky to use `task.innerExec != nil` to check if all the inner results is fetched, since `task.innerExec` can be set to nil if error happens, how about explicitly maintain a flag in `task` to indicate whether all the inner results are fetched or not

pkg/executor/join/index_lookup_hash_join.go
```diff
@@ -731,7 +734,20 @@ func (iw *indexHashJoinInnerWorker) handleTask(ctx context.Context, task *indexH
 
 	joinStartTime = time.Now()
```
windtalker: `joinStartTime` is not accurate for incremental probing, maybe we should collect the join time inside `doJoinUnordered` and `doJoinInOrder` independently.

pkg/executor/test/jointest/join_test.go
```diff
@@ -950,3 +950,34 @@ func TestIssue11896(t *testing.T) {
...
+		if i > 1 {
+			sql2 += ","
+		}
+		sql2 += fmt.Sprintf("(%d, 'abc')", i/1000)
+	}
+	tk.MustExec(sql2)
+
+	tk.MustQuery("select /*+ inl_hash_join(t1,t2) */ * from t1 inner join t2 on t1.a = t2.b")
+	tk.MustQuery("select /*+ inl_hash_join(t1,t2) */ * from t1 left join t2 on t1.a = t2.b")
+	tk.MustQuery("select /*+ inl_hash_join(t2,t1) */ * from t2 right join t1 on t1.a = t2.b")
```
windtalker, line 988: looks like the test does not check the result?

windtalker: Maybe we can at least add a count(*) query to check the join result.

pkg/executor/join/index_lookup_join.go
ChangRui-Ryan: `IndexNestedLoopHashJoin` also call the `fetchInnerResults` method of `indexLookupJoin`, while golang does not support default parameters.

pkg/executor/join/index_lookup_hash_join.go
ChangRui-Ryan: How about define `maxRowsPerFetch` as member variable in `indexHashJoinInnerWorker` and set it in `newInnerWorker` or `indexHashJoinInnerWorker.run`?  
Only `IndexNestedLoopHashJoin` has `Open`, but `maxRowsPerFetch` only use for `indexHashJoinInnerWorker`.

pkg/executor/join/index_lookup_hash_join.go
ChangRui-Ryan: I think it's reasonable that when error happens and `task.innerExec` is set to nil, we don't do fetch anymore. Because fetch depend on innerExec. In fact, I add a new flag needClose to determine whether set innerExec to nil after each `fetchInnerResults`.

pkg/executor/join/index_lookup_hash_join.go
```diff
@@ -476,12 +486,20 @@ func (e *IndexNestedLoopHashJoin) newInnerWorker(taskCh chan *indexHashJoinTask,
 	if e.stats != nil {
 		innerStats = &e.stats.innerWorker
 	}
+	var maxRows int
+	// If the joiner supports incremental look up, we can fetch inner results in batches,
+	// retrieving partial results multiple times until all are fetched.
+	// Otherwise, we fetch all inner results in one batch.
+	if e.supportIncrementalLookUp() && !e.KeepOuterOrder {
```
windtalker: why not just add `e.KeepOuterOrder`'s check inside `e.supportIncrementalLookup()`?

pkg/executor/join/index_lookup_hash_join.go
```diff
@@ -729,14 +743,35 @@ func (iw *indexHashJoinInnerWorker) handleTask(ctx context.Context, task *indexH
 		return err
 	}
 
-	joinStartTime = time.Now()
 	if !task.keepOuterOrder {
-		return iw.doJoinUnordered(ctx, task, joinResult, h, resultCh)
+		for {
+			err = iw.doJoinUnordered(ctx, task, joinResult, h, resultCh)
+			if err != nil {
```
windtalker, line 750: do we need to close `innerExec` here if `task.innerExec` is not nil ?

pkg/executor/join/index_lookup_join.go
```diff
@@ -717,33 +719,55 @@ func (iw *innerWorker) fetchInnerResults(ctx context.Context, task *lookUpJoinTa
...
 		}
 		if iw.executorChk.NumRows() == 0 {
+			needClose = true
 			break
 		}
 		innerResult.Add(iw.executorChk)
-		iw.executorChk = exec.TryNewCacheChunk(innerExec)
+		// If maxFetchSize is set, we need to break the loop when the fetched rows reach the max.
+		if iw.maxFetchSize > 0 && innerResult.Len() >= iw.maxFetchSize {
+			break
```
gengliqi: If it breaks here, the `iw.executorChk` will be reused in the next round.
Seems like it's ok, and if it's true, the chunk list can be reused to better reduce memory allocation pressure.

gengliqi (approved):

Issue #63713, opened 2025-09-24
Duplicate column names exist in `show grants` if current user set roles


CbcWestwolf:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```go
func TestShowGrantsWithRoles(t *testing.T) {
	store := testkit.CreateMockStore(t)
	tk := testkit.NewTestKit(t, store)
	tk.MustExec("use test")
	tk.MustExec("CREATE USER u")
	tk.MustExec("CREATE USER u@'r1'")
	tk.MustExec("CREATE USER u@'r2'")
	tk.MustExec("CREATE TABLE t (c1 int, c2 int)")

	tk.MustExec("GRANT select(c1, c2) ON test.t TO u@'r1'")
	tk.MustExec("GRANT select(c1, c2) ON test.t TO u@'r2'")
	tk.MustExec("GRANT u@'r1' TO u")
	tk.MustExec("GRANT u@'r2' TO u")
	tk.MustExec("SET DEFAULT ROLE ALL to u")

	usertk := testkit.NewTestKit(t, store)
	require.NoError(t, usertk.Session().Auth(&auth.UserIdentity{Username: "u", Hostname: "%"}, nil, nil, nil))
	usertk.MustQuery("show grants").Check(testkit.Rows(
		"GRANT USAGE ON *.* TO 'u'@'%'",
		"GRANT SELECT(c1, c2) ON `test`.`t` TO 'u'@'%'",
		"GRANT 'u'@'r1', 'u'@'r2' TO 'u'@'%'",
	))
}
```

### 2. What did you expect to see? (Required)

The test succeeded.

### 3. What did you see instead (Required)

The test failed since `show grants` outputs `GRANT SELECT(c1, c2, c1, c2) ON `test`.`t` TO 'u'@'%'`

### 4. What is your TiDB version? (Required)

master

PR #63714, opened 2025-09-24
privilege: fix a bug for `show grants` with roles

CbcWestwolf:
### What problem does this PR solve?

Issue Number: close #63713

Problem Summary:

See the issue description

### What changed and how does it work?

Dedup the column names before output

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #63728, opened 2025-09-24
privilege: fix some bugs for privilege cache and `show grants`   | tidb-test=pr/2616

CbcWestwolf:

Issue Number: ref #61706, #63713

Problem Summary:

This PR will fix some bugs and enhance some compatibility:
1. See #63713
2. When `show grants` for column privileges, we missed escaping the column name for pretty-printing
3. When `revoke` some privilege from a user **without** an explicit db name, it may cause inconsistency between `mysql.columns_priv` and `mysql.tables_priv`
4. If the number of column privilege cache records exceeds 1024*6=6144, memory leak happens and causes inconsistency between `mysql.columns_priv` and `information_schema.column_privileges`

**The first two issues also exists in master, and the latter two are bugs in https://github.com/pingcap/tidb/tree/release-8.5-20250606-v8.5.2**.

1. Dedup column name
2. Add escaping column name for `show grants`
3. Use default db name if `revoke` some privilege from a user **without** an explicit db name
4. Use `strings.Clone()` to prevent memory leak

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/privilege/privileges/cache.go
+++ b/pkg/privilege/privileges/cache.go
@@ -349,9 +349,9 @@ type MySQLPrivilege struct {
 	// In MySQL, a user identity consists of a user + host.
 	// Either portion of user or host can contain wildcards,
 	// requiring the privileges system to use a list-like
+	// structure instead of a hash.
+	// TiDB contains a sensible behavior difference from MySQL,
 	// which is that usernames can not contain wildcards.
-	// This means that DB-records are organized in both a
-	// slice (p.DB) and a Map (p.DBMap).
 
 	user         bTree[itemUser]
 	db           bTree[itemDB]
@@ -839,7 +839,7 @@ func (p *MySQLPrivilege) LoadGlobalPrivTable(exec sqlexec.SQLExecutor) error {
 	return nil
 }
 
-// LoadGlobalGrantsTable loads the mysql.global_priv table from database.
+// LoadGlobalGrantsTable loads the mysql.global_grants table from database.
 func (p *MySQLPrivilege) LoadGlobalGrantsTable(exec sqlexec.SQLExecutor) error {
 	if err := loadTable(exec, sqlLoadGlobalGrantsTable, p.decodeGlobalGrantsTableRow(nil)); err != nil {
 		return errors.Trace(err)
@@ -1461,7 +1461,7 @@ func (p *MySQLPrivilege) matchIdentity(user, host string, skipNameResolve bool) 
 func (p *MySQLPrivilege) connectionVerification(user, host string) *UserRecord {
 	records, exists := p.user.Get(itemUser{username: user})
 	if exists {
-		for i := 0; i < len(records.data); i++ {
+		for i := range records.data {
 			record := &records.data[i]
 			if record.Host == host { // exact match
 				return record
@@ -1477,7 +1477,7 @@ func (p *MySQLPrivilege) matchGlobalPriv(user, host string) *globalPrivRecord {
 		return nil
 	}
 	uGlobal := item.data
-	for i := 0; i < len(uGlobal); i++ {
+	for i := range uGlobal {
 		record := &uGlobal[i]
 		if record.match(user, host) {
 			return record
@@ -1490,7 +1490,7 @@ func (p *MySQLPrivilege) matchUser(user, host string) *UserRecord {
 	item, exists := p.user.Get(itemUser{username: user})
 	if exists {
 		records := item.data
-		for i := 0; i < len(records); i++ {
+		for i := range records {
 			record := &records[i]
 			if record.match(user, host) {
 				return record
@@ -1504,7 +1504,7 @@ func (p *MySQLPrivilege) matchDB(user, host, db string) *dbRecord {
 	item, exists := p.db.Get(itemDB{username: user})
 	if exists {
 		records := item.data
-		for i := 0; i < len(records); i++ {
+		for i := range records {
 			record := &records[i]
 			if record.match(user, host, db) {
 				return record
@@ -1518,7 +1518,7 @@ func (p *MySQLPrivilege) matchTables(user, host, db, table string) *tablesPrivRe
 	item, exists := p.tablesPriv.Get(itemTablesPriv{username: user})
 	if exists {
 		records := item.data
-		for i := 0; i < len(records); i++ {
+		for i := range records {
 			record := &records[i]
 			if record.match(user, host, db, table) {
 				return record
@@ -1532,7 +1532,7 @@ func (p *MySQLPrivilege) matchTables(user, host, db, table string) *tablesPrivRe
 func (p *MySQLPrivilege) MatchColumns(user, host, db, table, column string) *columnsPrivRecord {
 	item, exists := p.columnsPriv.Get(itemColumnsPriv{username: user})
 	if exists {
-		for i := 0; i < len(item.data); i++ {
+		for i := range item.data {
 			record := &item.data[i]
 			if record.match(user, host, db, table, column) {
 				return record


pkg/privilege/privileges/cache.go
```diff
@@ -2062,10 +2062,10 @@ func (p *MySQLPrivilege) fetchSchemaPrivileges(users, hosts []string) (rows [][]
 }
 
 type columnStr = string
-type columnStrs = []columnStr
+type columnStrs = map[columnStr]struct{}
 type privOnColumns = map[mysql.PrivilegeType]columnStrs
 
-func privOnColumnsToString(p privOnColumns) string {
+func privOnColumnsToString(p privOnColumns, sqlMode mysql.SQLMode) string {
```
fzzf678, line 2077: What‘s this `sqlMode` for? Just for pretty-printing?

pkg/privilege/privileges/cache.go
```diff
@@ -2092,7 +2097,7 @@ func privOnColumnsToString(p privOnColumns) string {
 }
 
 func collectColumnGrant(record *columnsPrivRecord, user, host string, columnPrivTable map[string]privOnColumns, sqlMode mysql.SQLMode) bool {
-	if record.baseRecord.fullyMatch(user, host) {
```
fzzf678, line 2113: Why here use `fullyMatch` before, i see the master is also `match`

pkg/executor/revoke.go
```diff
@@ -426,7 +426,7 @@ func (e *RevokeExec) revokeColumnPriv(ctx context.Context, internalSession sessi
 
 	if checkTablePrivTbl {
 		sql := strings.Join([]string{"SELECT * FROM %n.%n WHERE User=%? AND Host=%? AND DB=%? AND Table_name=%? AND Column_priv LIKE '%%", priv.Priv.String(), "%%';"}, "")
-		exists, err := recordExists(internalSession, sql, mysql.SystemDB, mysql.ColumnPrivTable, user, host, e.Level.DBName, e.Level.TableName)
+		exists, err := recordExists(internalSession, sql, mysql.SystemDB, mysql.ColumnPrivTable, user, host, dbName, e.Level.TableName)
```
wjhuang2016, line 429: Why change it？

pkg/privilege/privileges/cache.go
wjhuang2016, line 2077: We need to test the different escape behavior on different SQL mode.

CbcWestwolf: Yes, and it is compatible with mysql

pkg/privilege/privileges/cache.go
CbcWestwolf: ~Change back to `match`~
It should be `fullyMatch`. I will correct master in https://github.com/pingcap/tidb/pull/63714

pkg/executor/revoke.go
CbcWestwolf: If no explicit dbname, `e.Level.DBName` would be empty

pkg/privilege/privileges/cache.go
CbcWestwolf: [`b629d2c` (#63728)](https://github.com/pingcap/tidb/pull/63728/commits/b629d2c88caae88712f87ad4eaf2593777b83c69)

pkg/executor/revoke.go
Benjamin2037: If db empty will it reture other DB same table privs back？

pkg/privilege/privileges/cache.go
```diff
@@ -985,9 +985,9 @@ func parseHostIPNet(s string) *net.IPNet {
 func (record *baseRecord) assignUserOrHost(row chunk.Row, i int, f *resolve.ResultField) {
 	switch f.ColumnAsName.L {
 	case "user":
-		record.User = row.GetString(i)
+		record.User = strings.Clone(row.GetString(i))
```
Benjamin2037, line 997: Did this cause the duplicate column privileges in cache？

pkg/executor/revoke.go
CbcWestwolf: If `e.Level.DBName` is empty,  the `exists` will always be `false`. So the following unexpected `removeTablePriv` would be executed, causing potential inconsistency between `mysql.columns_priv` and `mysql.tables_priv`

pkg/privilege/privileges/cache.go
CbcWestwolf: Yes, since some `record.User` shares the same string, some strange things happen.

pkg/privilege/privileges/cache_test.go
```diff
@@ -576,3 +578,64 @@ func TestDBIsVisible(t *testing.T) {
 	require.True(t, isVisible)
 	tk.MustExec("TRUNCATE TABLE mysql.user")
 }
+
```
Benjamin2037, line 581: Please add some comment to description the test case use for which kind of problem?

CbcWestwolf: [`8bbb99c` (#63728)](https://github.com/pingcap/tidb/pull/63728/commits/8bbb99c2f3b29572e788b89cf5e567b22d9b4ae8)

pkg/privilege/privileges/cache_test.go
```diff
@@ -576,3 +578,68 @@ func TestDBIsVisible(t *testing.T) {
...
+// privilege cache. When loading privilege table into cache, if we forget to 'clone'
+// a string from a chunk row into cache record, inconsistency between privilege
+// table and cache happens.
+func TestColumnPrivilegeCacheCorrectness(t *testing.T) {
+	store := createStoreAndPrepareDB(t)
+
+	tk := testkit.NewTestKit(t, store)
+	tk.MustExec("create database column_priv_test")
+	tk.MustExec("use column_priv_test")
+	tk.MustExec("set global tidb_accelerate_user_creation_update = on")
```
fzzf678, line 592: How about add a case to verify the cache correctness when this variable is `off`

CbcWestwolf: Turn on `tidb_accelerate_user_creation_update` is just to speed up the granting process, otherwise the case will timeout.

fzzf678: For the build cache process, if we set this var `on`, is the code path same as the the 'problem path' encountered by user ？That's my concern.

CbcWestwolf: Yes, they share the same path. And in line 635 we turn `off` for test:
https://github.com/pingcap/tidb/pull/63728/files#diff-860eeb97f0492c2eeb38ecca6442094a483cb30f9afe7aac9f6ec4ce6b0d9a6bR635

CbcWestwolf: The following `flush privileges` will make another full reloading of privilege cache.

pkg/privilege/privileges/cache_test.go
```diff
@@ -576,3 +578,68 @@ func TestDBIsVisible(t *testing.T) {
...
+
+	tk.MustExec("set global tidb_accelerate_user_creation_update = off")
+	tk.MustExec("flush privileges")
+
+	for i := range userNum {
+		res1 := tk.MustQuery(fmt.Sprintf("select count(*) from mysql.columns_priv where user = '%s'", userNames[i]))
+		res2 := tk.MustQuery(fmt.Sprintf("select count(*) from information_schema.column_privileges where grantee = \"'%s'@'%%'\"", userNames[i]))
+
+		require.Equal(t, fmt.Sprintf("%d", colNum), res1.Rows()[0][0])
+		require.Equal(t, res1.Rows()[0][0], res2.Rows()[0][0])
```
fzzf678, line 643: Only the column privilege quantity is checked here, how about check the specific content here? The content and quantity should be all consistent.

CbcWestwolf: Checking column privilege quantity is enough. If a memory leak happens, the result of `select count(*) from information_schema.column_privileges where grantee ...` is 0.

PR #63759, opened 2025-09-25
stats: upgrade to use advancedSessionPool

### LINKED ISSUE #60897: Use AdvancedSessionPool in stats moudle
## Enhancement

We introduce an advanced session pool in this RFC: https://github.com/pingcap/tidb/pull/60120

It is safer and easier to use, so it is better to switch to this new session pool in the stats module.

Tristan1900:

Issue Number: close #60897

Problem Summary:

As issue describes, upgrade advancedSessionPool in stats module.
 

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/handle/util/util.go
+++ b/pkg/statistics/handle/util/util.go
@@ -93,9 +93,8 @@ func CallWithSCtx(pool syssession.Pool, f func(sctx sessionctx.Context) error, f
 			}
 			if wrapTxn {
 				return WrapTxn(sctx, f)
-			} else {
-				return f(sctx)
 			}
+			return f(sctx)
 		})
 	})
 }


pkg/statistics/handle/syncload/stats_syncload.go
```diff
@@ -327,26 +328,28 @@ func isVaildForRetry(task *statstypes.NeededItemTask) bool {
...
-			// Note: Otherwise, the session will be leaked.
-			s.statsHandle.SPool().Destroy(se)
-		}
 	}()
+
+	return s.statsHandle.SPool().WithSession(func(se *syssession.Session) error {
+		return se.WithSessionContext(func(sctx sessionctx.Context) error {
+			sctx.GetSessionVars().StmtCtx.Priority = mysql.HighPriority
+			defer func() {
+				sctx.GetSessionVars().StmtCtx.Priority = mysql.NoPriority
```
lance6716, line 343: Maybe we can add an `intest` check to session pool to assert the `Priority` and other fields are properly reset when return it to session pool.

cc @0xPoe

0xPoe: Sounds good!

0xPoe:
/hold

For https://github.com/pingcap/tidb/pull/63759#discussion_r2415687584

pkg/statistics/handle/syncload/stats_syncload.go
Tristan1900: I agree with the idea but when I add the check I realize currently there are other callers not setting priority back to 0 when returning, instead they are set to priority high.
Just want to confirm, Is that expected? If not, we probably want to reset at the session pool level instead of at callers? Please let me know!

0xPoe: I think we should usually roll back to the default priority. This change shouldn’t block this PR.
Could you please create an issue so we can fix it later?
Also, you’ll need some approvers from other modules.

Tristan1900: sounds good, thanks!
issue created for tracking https://github.com/pingcap/tidb/issues/63963

lance6716: > we probably want to reset at the session pool level instead of at callers

This is the most safe design, but less efficient IMO. Like `ResetContextOfStmt`, in future there maybe tons of fields to be reset, but a caller may only touch few of them (in some benchmarks `ResetContextOfStmt` occupies non-negligible CPU). Only the caller know what it has changed. I think it's better to make use of `intest` to check the property: caller must reset all changes before return to the pool.

Tristan1900: I understand your concerns. I might be missing some context, but do you mean adding inttest on the pool side or the caller side?
If we add it on the pool side, we would basically need to check every variable since the pool doesn’t know which variable has changed — that’s effectively equivalent to resetting all of them (though we might find a cheaper way to reset, such as keeping a default variable struct and simply cloning and swapping it).
If we add it on the caller side, then it becomes the caller’s responsibility to use inttest for verification. However, that somewhat defeats the purpose, as it doesn’t prevent cases where the caller isn’t even aware of the need to reset — which is already happening with some existing callers in the codebase.

lance6716: We should add it on the pool side. And `intest` build flag is not enabled in production so no performace drop.

PR #63779, opened 2025-09-26
*: add a variable for dynamic slow log trigger rules with global (supporting ConnID-specific) and session scope

zimulala:

Issue Number: close https://github.com/pingcap/tidb/issues/63778

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/adapter.go
+++ b/pkg/executor/adapter.go
@@ -1668,8 +1668,9 @@ func (a *ExecStmt) LogSlowQuery(txnTS uint64, succ bool, hasMoreResults bool) {
 	sessVars := a.Ctx.GetSessionVars()
 	sessVars.StmtCtx.ExecSuccess = succ
 	sessVars.StmtCtx.ExecRetryCount = uint64(a.retryCount)
-	slowItems := PrepareSlowLogItemsForRules(a.GoCtx, sessVars)
-	matchRules := ShouldWriteSlowLog(sessVars, slowItems)
+	globalRules := vardef.GlobalSlowLogRules.Load()
+	slowItems := PrepareSlowLogItemsForRules(a.GoCtx, globalRules, sessVars)
+	matchRules := ShouldWriteSlowLog(globalRules, sessVars, slowItems)
 	if (!enable || !matchRules) && !force {
 		return
 	}
--- a/pkg/executor/adapter.go
+++ b/pkg/executor/adapter.go
@@ -1670,7 +1670,13 @@ func (a *ExecStmt) LogSlowQuery(txnTS uint64, succ bool, hasMoreResults bool) {
 	sessVars.StmtCtx.ExecRetryCount = uint64(a.retryCount)
 	globalRules := vardef.GlobalSlowLogRules.Load()
 	slowItems := PrepareSlowLogItemsForRules(a.GoCtx, globalRules, sessVars)
-	matchRules := ShouldWriteSlowLog(globalRules, sessVars, slowItems)
+	var matchRules bool
+	if slowItems == nil {
+		threshold := time.Duration(atomic.LoadUint64(&cfg.Instance.SlowThreshold)) * time.Millisecond
+		matchRules = sessVars.GetTotalCostDuration() >= threshold
+	} else {
+		matchRules = ShouldWriteSlowLog(globalRules, sessVars, slowItems)
+	}
 	if (!enable || !matchRules) && !force {
 		return
 	}
--- a/pkg/executor/adapter.go
+++ b/pkg/executor/adapter.go
@@ -15,7 +15,6 @@
 package executor
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"math"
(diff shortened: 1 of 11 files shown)


pkg/executor/adapter.go
```diff
@@ -1664,161 +1663,95 @@ func resetCTEStorageMap(se sessionctx.Context) error {
...
-				TableIDs:   tableIDs,
-				IndexNames: indexNames,
-				Internal:   sessVars.InRestrictedSQL,
-			})
+
+	costTime := slowItems.TimeTotal
+	threshold := time.Duration(atomic.LoadUint64(&cfg.Instance.SlowThreshold)) * time.Millisecond
+	if costTime < threshold {
+		return
+	}
```
zimulala, line 1714: The code below does not change the logic; it just changes 1711 lines of judgment.

zimulala:
PTAL @yibin87 @XuHuaiyu

pkg/executor/adapter.go
```diff
@@ -1664,161 +1663,96 @@ func resetCTEStorageMap(se sessionctx.Context) error {
...
-			if ok {
-				continue
-			}
-			idxMap[idx] = struct{}{}
-			if buf.Len() > 1 {
-				buf.WriteByte(',')
-			}
-			buf.WriteString(idx)
+		sessVars.StmtCtx.ExecSuccess = succ
+		sessVars.StmtCtx.ExecRetryCount = uint64(a.retryCount)
```
yibin87, line 1677: Consider always record these two stmtCtx variables, or add comments to explain when these two variables contain meaningful values.
Personally, I prefer the first one, since stmtSummary or other module may use these two variables also in future.

pkg/sessionctx/variable/slow_log.go
```diff
@@ -830,106 +851,241 @@ var SlowLogRuleFieldAccessors = map[string]SlowLogFieldAccessor{
...
+		if err != nil {
+			return connID, nil, errors.Errorf("invalid slow log format, value:%s, err:%s", value, err)
+		}
+
+		if strings.EqualFold(fieldName, SlowLogConnIDStr) {
+			if !allowConnID {
+				return connID, nil, errors.Errorf("do not allow ConnID value:%s", value)
+			}
+
+			id, err := strconv.ParseUint(value, 10, 64)
```
yibin87: why not directly use fieldValue here?

pkg/executor/adapter.go
```diff
@@ -1664,161 +1663,96 @@ func resetCTEStorageMap(se sessionctx.Context) error {
...
-				User:       userString,
-				DB:         sessVars.CurrentDB,
-				TableIDs:   tableIDs,
-				IndexNames: indexNames,
-				Internal:   sessVars.InRestrictedSQL,
-			})
+
+	costTime := slowItems.TimeTotal
+	threshold := time.Duration(atomic.LoadUint64(&cfg.Instance.SlowThreshold)) * time.Millisecond
+	if costTime < threshold {
```
yibin87:
```suggestion
	if !matchRules {
```

pkg/executor/adapter.go
zimulala: No other module seems to be using it at the moment. For example, stmtSummary uses "a.retryCount" directly in a similar function. I added comments in this pr.

zimulala:
> It is a little strange that both session var and global var take effect at the same time. It would be more reasonable for only the latest session var takes effect, while global var will actually take effect after overwriting the session var.

This is because this variable is somewhat different — the global variable can specify special rules for certain sessions.
Without changing the existing mechanism for setting/getting session and global variables, it may cause a situation where the session variable is overridden by the special session rules defined in the global variable.

yibin87:
> > It is a little strange that both session var and global var take effect at the same time. It would be more reasonable for only the latest session var takes effect, while global var will actually take effect after overwriting the session var.
> 
> This is because this variable is somewhat different — the global variable can specify special rules for certain sessions. Without changing the existing mechanism for setting/getting session and global variables, it may cause a situation where the session variable is overridden by the special session rules defined in the global variable.

Actually, I think "the situation where the session variable is overridden by the special session rules defined in the global variable" seems make sense. Because it follows the same rule with other system variables.:)

zimulala:
> > > It is a little strange that both session var and global var take effect at the same time. It would be more reasonable for only the latest session var takes effect, while global var will actually take effect after overwriting the session var.
> > 
> > 
> > This is because this variable is somewhat different — the global variable can specify special rules for certain sessions. Without changing the existing mechanism for setting/getting session and global variables, it may cause a situation where the session variable is overridden by the special session rules defined in the global variable.
> 
> Actually, I think "the situation where the session variable is overridden by the special session rules defined in the global variable" seems make sense. Because it follows the same rule with other system variables.:)

What I mean is that SetGlobal may be called periodically, i.e. the rules of the later set session may also be overridden. I brought this up in the meeting before, and there's a little document. Finally, PM decided this behavior.

zimulala:
PTAL @XuHuaiyu

solotzg (approved):

pkg/sessionctx/variable/slow_log.go
```diff
@@ -830,106 +851,237 @@ var SlowLogRuleFieldAccessors = map[string]SlowLogFieldAccessor{
...
 
-			fieldName := strings.ToLower(strings.TrimSpace(match[1]))
-			value := strings.TrimSpace(match[2])
-			fieldValue, err := ParseSlowLogFieldValue(fieldName, strings.Trim(value, "\"'"))
-			if err != nil {
-				return nil, errors.Errorf("invalid slow log format, value:%s, err:%s", value, err)
+// ParseSessionSlowLogRules parses raw rules into the default (UnsetConnID) slow log rules.
+// Returns nil if no rules for UnsetConnID are found.
+func ParseSessionSlowLogRules(rawRules string) (*slowlogrule.SlowLogRules, error) {
+	globalRules, err := parseSlowLogRuleSet(rawRules, false)
```
solotzg, line 1024: Have you considered json format?

zimulala: At the time, we considered this version to be a simpler format without JSON, and we could consider it later if we needed to optimize it

XuHuaiyu (approved):

Issue #61273, opened 2025-05-22
Tracking issue: Refresh Stats Command


0xPoe:
Guide-level explanation
This command looks as follows:
```sql
REFRESH STATS <TARGETS> [FULL | LITE] [CLUSTER] 
REFRESH STATS db1.tab1, tab2, db2.*, *.*, 
```
- <TARGETS> takes a list of targets separated by comma, which can be any of the following formats.
  - tab1:  tab1 under current schema.
  - db.tab1: particular tables
  - db.* : all tables under schema db
  - `*.*`: all tables in this cluster
- [FULL | LITE] names the type of loading initial statistics. 
  - If the option is omitted, it takes the value from config [lite-init-stats](https://docs.pingcap.com/tidb/stable/tidb-configuration-file/#lite-init-stats-new-in-v710). 
  - Explicit FULL refers to [lite-init-stats](https://docs.pingcap.com/tidb/stable/tidb-configuration-file/#lite-init-stats-new-in-v710)=false
  - Explicit LITE refers to [lite-init-stats](https://docs.pingcap.com/tidb/stable/tidb-configuration-file/#lite-init-stats-new-in-v710)=true
- If [CLUSTER] is set, initialization of statistics happens for all TiDB instances. Otherwise, initialization happens to the TiDB instance connected only. 
Any users, including BR, with read permissions to the tables, can use this command to initialize the stats for the table.

0xPoe:
The Pre-RFC: https://pingcap.feishu.cn/wiki/WWh3wrBTYitbuek10wucGgM7nhh

I will draft the final RFC(Design Doc) on GitHub soon.

0xPoe:
Found a bug in the implementation:

```log
[...]

PR #63822, opened 2025-09-30
planner: check select or restore-admin privilege for the refresh cmd

0xPoe:

Problem Summary:

Do privilege checks for the refresh cmd. We require the select privilege for all tables.

So we require the select privilege for this command, and because this command is mainly for BR, we also add the restore admin permission as a special case.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/simple_test.go
+++ b/pkg/executor/simple_test.go
@@ -18,8 +18,10 @@ import (
 	"context"
 	"testing"
 
+	"github.com/pingcap/tidb/pkg/errno"
 	"github.com/pingcap/tidb/pkg/infoschema"
 	"github.com/pingcap/tidb/pkg/parser/ast"
+	"github.com/pingcap/tidb/pkg/parser/auth"
 	"github.com/pingcap/tidb/pkg/testkit"
 	"github.com/pingcap/tidb/pkg/util/dbterror/plannererrors"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


pkg/executor/simple_test.go
```diff
@@ -145,3 +147,74 @@ func TestRefreshStatsWhenDatabaseIsEmpty(t *testing.T) {
...
+	tk.MustExec("use test")
+	tk.MustExec("drop table if exists t_refresh_priv")
+	tk.MustExec("create table t_refresh_priv (a int)")
+
+	t.Run("table scope requires select", func(t *testing.T) {
+		tk.MustExec("drop user if exists 'refresh_reader'@'%'")
+		tk.MustExec("create user 'refresh_reader'@'%'")
+		t.Cleanup(func() {
+			tk.MustExec("drop user if exists 'refresh_reader'@'%'")
+		})
```
mjonss: Are these necessary? Or is not every test a new store including users etc.?

mjonss (approved):
LGTM

pkg/executor/simple_test.go
0xPoe: In theory, they are not necessary because I used a different username. I deleted it.

pkg/planner/core/planbuilder.go
```diff
@@ -4900,6 +4900,49 @@ func fillDefaultDBForRefreshStats(ctx base.PlanContext, rs *ast.RefreshStatsStmt
...
+	checker := privilege.GetPrivilegeManager(b.ctx)
+	if checker != nil {
+		activeRoles := b.ctx.GetSessionVars().ActiveRoles
+		if checker.RequestDynamicVerification(activeRoles, "RESTORE_ADMIN", false) {
+			return
+		}
+	}
+
+	user := b.ctx.GetSessionVars().User
+	for _, obj := range rs.RefreshObjects {
```
Tristan1900, line 4918: curious how many objects can it be? Could it be one million tables and have one million object and we add this logging info for one million? not sure if that's going to be meaningful and if it can cause other issues like memory pressure or log spam.

Tristan1900 (commented):
rest LGTM, just one question!

pkg/planner/core/planbuilder.go
0xPoe: No, usually we just use `*.*` for this use case. No one would try to list all one million tables in a single SQL query.

mjonss: Theoretically it could be a million, i.e. if an application would generate the command for all tables in the database, instead of using `*.*`, but then I don't think it would be an issue, even if it would take long time and use a lot of memory etc. I would say that would still be a user issue, not a TiDB issue ;)

terry1purcell (approved):

PR #63830, opened 2025-10-01
fix: close issue 63823 by not check CURRENT_DATE

### LINKED ISSUE #63823: `datetime(6)` column cannot have `default (current_date)`.
## Bug Report

```sql
create table a (a datetime(6) default (current_date));
```

Query OK

```
ERROR 1067 (42000): Invalid default value for 'a'
```

Reproduced on all of:
* v9.0.0-beta.2-pre-nightly 78eadf20801cee122e18eb449498b9b790f04e96
* v9.0.0-beta.1 7aff918dcbfa6facf2adef9ade9961c40f217421
* v8.5.3
* v8.1.2
* v7.5.7
* v7.1.6

(v6.5.12 does not support expression as default)

yihong0618:
as diff

Issue Number: close #63823

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/add_column.go
+++ b/pkg/ddl/add_column.go
@@ -653,14 +653,17 @@ func getFuncCallDefaultValue(col *table.Column, option *ast.ColumnOption, expr *
 	case ast.CurrentTimestamp, ast.CurrentDate: // CURRENT_TIMESTAMP() and CURRENT_DATE()
 		tp, fsp := col.FieldType.GetType(), col.FieldType.GetDecimal()
 		if tp == mysql.TypeTimestamp || tp == mysql.TypeDatetime {
-			defaultFsp := 0
-			if len(expr.Args) == 1 {
-				if val := expr.Args[0].(*driver.ValueExpr); val != nil {
-					defaultFsp = int(val.GetInt64())
+			// CURRENT_DATE() returns a date without time, so it's always valid for datetime/timestamp columns
+			if expr.FnName.L == ast.CurrentTimestamp {
+				defaultFsp := 0
+				if len(expr.Args) == 1 {
+					if val := expr.Args[0].(*driver.ValueExpr); val != nil {
+						defaultFsp = int(val.GetInt64())
+					}
 				}
-			}
-			if defaultFsp != fsp {
-				return nil, false, dbterror.ErrInvalidDefaultValue.GenWithStackByArgs(col.Name.O)
+				if defaultFsp != fsp {
+					return nil, false, dbterror.ErrInvalidDefaultValue.GenWithStackByArgs(col.Name.O)
+				}
 			}
 		}
 		return nil, false, nil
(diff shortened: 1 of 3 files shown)


pkg/ddl/add_column.go
```diff
@@ -653,14 +653,17 @@ func getFuncCallDefaultValue(col *table.Column, option *ast.ColumnOption, expr *
 	case ast.CurrentTimestamp, ast.CurrentDate: // CURRENT_TIMESTAMP() and CURRENT_DATE()
 		tp, fsp := col.FieldType.GetType(), col.FieldType.GetDecimal()
 		if tp == mysql.TypeTimestamp || tp == mysql.TypeDatetime {
-			defaultFsp := 0
-			if len(expr.Args) == 1 {
-				if val := expr.Args[0].(*driver.ValueExpr); val != nil {
-					defaultFsp = int(val.GetInt64())
+			// CURRENT_DATE() returns a date without time, so it's always valid for datetime/timestamp columns
+			if expr.FnName.L == ast.CurrentTimestamp {
```
lance6716, line 657: Instead of the special check of `CURRENT_DATE()`, do you think it's better that we allow `defaultFsp <= fsp` 🤔 Or maybe we should also allow storing `CURRENT_TIME(6)` into `datetime(3)`. Please check MySQL's behaviour about these cases.

yihong0618: checked mysql they are the same for tidb now
after this patch

lance6716: oh, you mean other mismatched precision case of the default expression and the column will cause error?

yihong0618: I mean after this patch the data and behavior are the same as mysql

1. no error
2. insert with default persion is the same

mysql> show tables;
+----------------+
| Tables_in_test |
+----------------+
| a              |
+----------------+
1 row in set (0.001 sec)

mysql> select * from a;
+----------------------------+
| a                          |
+----------------------------+
| 2025-10-01 00:00:00.000000 |
| 2025-10-01 00:00:00.000000 |
+----------------------------+
2 rows in set (0.001 sec)

mysql> select version()
    -> l
    -> ;
+-------+
| l     |
+-------+
| 9.4.0 |
+-------+
1 row in set (0.000 sec)

lance6716: I understand that your PR can fix the problem. But it adds a too special check which is totally skipping precision comparison for `CURRENT_DATE()`. Please if this is a more general perspective to fix it:

1. TiDB should behave the same as MySQL when column precision `==` default value expression precision (old logic should have covered this)
2. ... when column precision `>` default value expression precision (the issue, and cases like column `datetime(1)` with default `CURRENT_TIME()`)
3. ... when column precision `<` default value expression precision (cases like column `datetime(3)` with default `CURRENT_TIME(6)` )

yihong0618: copy that, will check later Happy National Day and Mid-Autumn Festival!

yihong0618: sorry for late will check it today or tomorrow

yihong0618:
closing because after some dig this one is not so simple, so waiting for core dev to solve it

PR #63846, opened 2025-10-04
*: add prompt for claude code

hawkingrei:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,89 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+### Code Organization
+
+**Package Structure:**
+
+- `/pkg/bindinfo/` - Handles all global SQL bind operations and caches SQL bind info from storage.
+- `/pkg/config/` - Configuration definitions.
+- `/pkg/ddl/` - Data Definition Language (DDL) execution logic.
+- `/pkg/distsql/` - Abstraction of distributed computing interfaces between executor and TiKV client.
+- `/pkg/domain/` - Storage space abstraction (domain/namespace), manages information schema and statistics.
+- `/pkg/errno/` - MySQL error code, message, and summary definitions.
+- `/pkg/executor/` - Execution logic for most SQL statements (operators).
+- `/pkg/expression/` - Expression-related code, including operators and built-in functions.
+- `/pkg/infoschema/` - Metadata management for SQL statements and information schema operations.
+- `/pkg/kv/` - Key-Value engine interface and public methods; storage engine adaptation layer.
+- `/pkg/lock/` - Implementation of LOCK/UNLOCK TABLES.
+- `/pkg/meta/` - SQL metadata management in storage engine; used by infoschema and DDL.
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,89 @@
...
+Standard Go tests throughout `/pkg/` packages
+
+#### How to run unit tests
+
+```bash
+# in the root directory of the repository
+make failpoint-enable
+cd pkg/<package_name>
+go test -v -run  <TestName>  -record --tags=intest
+# return to the root directory
```
lance6716: maybe ask LLM to use `pushd` and other command to switch PWD back

yihong0618:
we can add make integrationtest

hawkingrei:
> we can add make integrationtest

For personal computers, running `make integrationtest` directly takes too long, so I rarely run it directly.

CLAUDE.md
hawkingrei: Updated

CLAUDE.md
```diff
@@ -0,0 +1 @@
+AGENTS.md
```
lance6716: I'm not sure if we need to use `@AGENTS.md` to import that file

https://docs.claude.com/en/docs/claude-code/memory#claude-md-imports

yihong0618: we can just `ln` it

hawkingrei: I found that some repos still use the reference format, so I changed it to reference.

AGENTS.md
```diff
@@ -0,0 +1,93 @@
...
+```bash
+# in the root directory of the repository
+make failpoint-enable
+pushd pkg/<package_name>
+go test -v -run  <TestName>  -record --tags=intest
+popd
+make failpoint-disable
+```
+
+#### Unit Tests Specification
```
lance6716, line 76:
```suggestion
#### Unit Tests Specification

```

AGENTS.md
```diff
@@ -0,0 +1,94 @@
...
+
+Standard Go tests throughout `/pkg/` packages
+
+#### How to run unit tests
+
+```bash
+# in the root directory of the repository
+make failpoint-enable
+pushd pkg/<package_name>
+go test -v -run  <TestName>  -record --tags=intest
```
mjonss, line 67: Do we really want to use `-record` here? Don't we want it to fail if a test is changing the results? I would propose to have a separate header with 'How to update unit tests results files'.

hawkingrei: The workflow here is designed to enable AI to execute automatically and to retain the results of the run within a result set for subsequent review by a human. This is because if the agent were to execute directly, it would be very easy to lose the content of the run.

hawkingrei: The files here are intended for use by AI, not oriented towards developers.

hawkingrei: I added some prompts to instruct the AI on exactly what to do.

PR #63878, opened 2025-10-09
executor: add more tests for refresh stats command

0xPoe:

Problem Summary:

Add more tests to cover cases where the mode of the refresh stats command conflicts with the current configuration.

1. When lite-init-stats = false: lite refresh stats will not be retained for future loads. We respect the config, since there is no point in keeping lite stats in memory when lite-init-stats is false.
2. When lite-init-stats = true: full refresh stats will be retained for future loads. The purpose of a full load is to fetch all stats and use them later, so we will respect what is already in memory during future loads.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/simple_test.go
+++ b/pkg/executor/simple_test.go
@@ -235,6 +235,17 @@ func TestRefreshStatsWithFullMode(t *testing.T) {
 	statsAfterDefaultRefresh := handle.GetPhysicalTableStats(tbl1Meta.ID, tbl1Meta)
 	require.NotSame(t, statsBeforeRefresh, statsAfterDefaultRefresh)
 	require.Nil(t, statsAfterDefaultRefresh.GetIdx(1), "index stats should not be loaded in lite mode")
+
+	tk.MustExec("select * from t1 where a = 1")
+	statsAfterSelect := handle.GetPhysicalTableStats(tbl1Meta.ID, tbl1Meta)
+	require.NotSame(t, statsBeforeRefresh, statsAfterSelect, "stats versuon should not be changed after select")
+	require.NotNil(t, statsAfterSelect.GetIdx(1), "index stats will be loaded after select")
+
+	tk.MustExec("refresh stats t1")
+	statsAfterDefaultRefresh = handle.GetPhysicalTableStats(tbl1Meta.ID, tbl1Meta)
+	require.NotSame(t, statsBeforeRefresh, statsAfterDefaultRefresh)
+	require.Nil(t, statsAfterDefaultRefresh.GetIdx(1), "index stats should be removed in lite mode")
+
 	// Issue a full refresh to ensure the index stats are loaded.
 	tk.MustExec("refresh stats t1 full")
 	statsAfterFullRefresh := handle.GetPhysicalTableStats(tbl1Meta.ID, tbl1Meta)


pkg/executor/simple_test.go
```diff
@@ -209,3 +210,116 @@ func TestRefreshStatsWithRestoreAdmin(t *testing.T) {
...
+	tk.MustExec("refresh stats t1 full")
+	statsAfterFullRefresh := handle.GetPhysicalTableStats(tbl1Meta.ID, tbl1Meta)
+	require.NotSame(t, statsAfterDefaultRefresh, statsAfterFullRefresh)
+	require.NotNil(t, statsAfterFullRefresh.GetIdx(1), "index stats should be loaded in full mode")
+	require.Len(t, statsAfterFullRefresh.GetIdx(1).Buckets, 1, "buckets should be loaded in full mode")
+	indexVersionAfterFullRefresh := statsAfterFullRefresh.GetIdx(1).LastUpdateVersion
+
+	// Insert additional data and run ANALYZE again.
+	tk.MustExec("insert into t1 values (4,4), (5,5)")
+	// Analyze loads statistics using the stats currently held in memory when running in lite mode.
```
winoros: We've run `refresh stats` at line#239. Why do we say `using the stats currently held in memory when running in lite mode.

0xPoe: What I mean here is that since refresh stats loads other statistics, analyze will load the updated statistics based on the state that has already been loaded when running in lite mode.

0xPoe: I updated the comment here: [`274ac03` (#63878)](https://github.com/pingcap/tidb/pull/63878/commits/274ac03225c9409f958efd141994d1011118d565)

pkg/executor/simple_test.go
```diff
@@ -209,3 +210,116 @@ func TestRefreshStatsWithRestoreAdmin(t *testing.T) {
...
+	handle := dom.StatsHandle()
+	ctx := context.Background()
+	tbl1, err := is.TableByName(ctx, ast.NewCIStr("test"), ast.NewCIStr("t1"))
+	require.NoError(t, err)
+	tbl1Meta := tbl1.Meta()
+	statsBeforeRefresh := handle.GetPhysicalTableStats(tbl1Meta.ID, tbl1Meta)
+	tk.MustExec("refresh stats t1")
+	statsAfterDefaultRefresh := handle.GetPhysicalTableStats(tbl1Meta.ID, tbl1Meta)
+	require.NotSame(t, statsBeforeRefresh, statsAfterDefaultRefresh)
+	require.Nil(t, statsAfterDefaultRefresh.GetIdx(1), "index stats should not be loaded in lite mode")
```
AilinKid, line 237: how about query before to make sure select * from t1 use index(idx) to get idx into memory already, and see after lite refresh existed idx is still there or not?

pkg/executor/simple_test.go
```diff
@@ -209,3 +210,116 @@ func TestRefreshStatsWithRestoreAdmin(t *testing.T) {
...
+	tk.MustExec("analyze table t1 all columns with 1 topn, 2 buckets")
+
+	is := dom.InfoSchema()
+	handle := dom.StatsHandle()
+	ctx := context.Background()
+	tbl1, err := is.TableByName(ctx, ast.NewCIStr("test"), ast.NewCIStr("t1"))
+	require.NoError(t, err)
+	tbl1Meta := tbl1.Meta()
+	statsBeforeRefresh := handle.GetPhysicalTableStats(tbl1Meta.ID, tbl1Meta)
+	tk.MustExec("refresh stats t1")
```
AilinKid, line 234: default is lite init mode?

0xPoe: Yes

pkg/executor/simple_test.go
0xPoe: Good catch! I will try it.

pkg/executor/simple_test.go
0xPoe: It’s inherited from the current configuration.

pkg/executor/simple_test.go
0xPoe: The lite refresh stats will remove it. Somehow I think that makes sense, since the configuration is still in lite mode and the refresh is explicitly triggered in lite mode as well. However, I’m not sure if this is good behavior — we might be letting users shoot themselves in the foot.

0xPoe: Let’s discuss it in our stats meeting.

0xPoe:
/hold

https://github.com/pingcap/tidb/pull/63878#discussion_r2426165464

pkg/executor/simple_test.go
0xPoe: We discussed this issue in the stats weekly meeting and agreed that this is the expected behavior. When using lite init stats, the previously loaded statistics are intentionally ignored. We should document this behavior clearly.

Issue #63326, opened 2025-09-02
embed analyze action into index creation related ddl


AilinKid:
## Enhancement
for some user requirements, and for extreme scenarios, business could bear a little bit of change unstability among ddl creation or reganization of index. That's why it is necessary for collection of these undergoing index stats when ddl is public to user.

PR #63894, opened 2025-10-10
ddl: run embedded analyze only once for multi-schema change

tangenta:

Issue Number: ref https://github.com/pingcap/tidb/issues/63326

Problem Summary:

Before this PR, the embedded analyze only execute for the first sub-job for multi-schema change, other columns from other sub-jobs are not analyzed. If we fix this problem, it may execute multiple times, which is not efficient.

This PR forces the analyze to be executed in the last sub-job, so that it includes all columns & indexes changing by the whole DDL job.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,0 +1,10 @@
+{
+  "mcpServers": {
+    "gopls-mcp": {
+      "command": "gopls",
+      "args": [
+        "mcp"
+      ]
+    }
+  }
+}
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,98 @@
+# AGENTS.md
+
+This file provides guidance to agents when working with code in this repository.
+
+### Code Organization
+
+**Package Structure:**
+
+- `/pkg/bindinfo/` - Handles all global SQL bind operations and caches SQL bind info from storage.
+- `/pkg/config/` - Configuration definitions.
+- `/pkg/ddl/` - Data Definition Language (DDL) execution logic.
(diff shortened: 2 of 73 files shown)


pkg/ddl/multi_schema_change.go
```diff
@@ -367,6 +367,18 @@ func mergeAddIndex(info *model.MultiSchemaInfo) {
 	info.SubJobs = newSubJobs
 }
 
+// setNeedAnalyze sets NeedAnalyze for the last sub-job that can do embedded analyze.
+func setNeedAnalyze(parentJob *model.Job, info *model.MultiSchemaInfo) {
+	for i := len(info.SubJobs) - 1; i >= 0; i-- {
+		subJob := info.SubJobs[i]
+		proxyJob := subJob.ToProxyJob(parentJob, i)
+		if proxyJob.CanEmbeddedAnalyze() {
```
zimulala: Could we use `MayNeedReorg` instead of `CanEmbeddedAnalyze`?

tangenta: No, reorg partition "NeedReorg" but cannot do "EmbeddedAnalyze".

zimulala: I thought we didn't support "reorg partition" when multi-schema change, but now we do?

tangenta: No, but I think depending on this assumption is not good..

Issue #63856, opened 2025-10-08
Planner: Planning time increases significantly with large number of indexes


terry1purcell:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```
CREATE TABLE `t8` (
  `a` int,
  `b` int,
  `c` int,
  `d` int,
  `e` int,
  `f` int,
  `g` int,
  `h` int,
  `i` int,
  `j` int,
  `k` int,
  `l` int,
  `m` int,
  `n` int,
  `o` int,
  `p` int,
  `q` int,
  `r` int,
  `s` int,
  `t` int,
  `u` int,
  `v` int,
  `w` int,
  `x` int,
  `y` int,
  `z` int,
  `aa` int,
  `ab` int,
  `ac` int,
  `ad` int,
  `ae` int,
  `af` int,
  `ag` int,
  `ah` int,
  `ai` int,
  `aj` int,
  `ak` int,
  `al` int,
  `am` int,
  `an` int,
  `ao` int,
  `ap` int,
  `aq` int,
  `ar` int,
  `as` int,
  `at` int,
  `au` int,
  `av` int,
  `aw` int,
  `ax` int,
  `ay` int,
  `az` int,
  `ba` int,
  `bb` int,
  `bc` int,
  `bd` int,
  `be` int,
  `bf` int,
  `bg` int,
  `bh` int,
  `bi` int,
  `bj` int,
  `bk` int,
  `bl` int,
  `bm` int,
  `bn` int,
  `bo` int,
  `bp` int,
  `bq` int,
  `br` int,
  `bs` int,
  `bt` int,
  `bu` int,
  `bv` int,
  `bw` int,
  `bx` int,
  `by` int,
  `bz` int,
  `ca` int,
  `cb` int,
  `cc` int,
  `cd` int,
  `ce` int,
  `cf` int,
  `cg` int,
  `ch` int,
  `ci` int,
  `cj` int,
  `ck` int,
  `cl` int,
  `cm` int,
  `cn` int,
  `co` int,
  `cp` int,
  `cq` int,
  `cr` int,
  `cs` int,
  `ct` int,
  `cu` int,
  `cv` int,
  `cw` int,
  `cx` int,
  `cy` int,
  `cz` int,
  `da` int,
  `db` int,
  `dc` int,
  `dd` int,
  `de` int,
  `df` int,
  `dg` int,
  `dh` int,
  `di` int,
  `dj` int,
[...]

PR #63912, opened 2025-10-11
planner: Index pruning for very large number of indexes | tidb-test=pr/2633

terry1purcell:

Issue Number: ref #63856 

Problem Summary:

Add index pruning early in the planning process - to reduce the overhead associated with customers who want to create a very large number of indexes. This was motivated by one customer who had tested 250 indexes on a single table, and another customer requesting support for 500 indexes.

A new variable tidb_opt_index_prune_threshold is added to control the pruning.

This pruning occurs early in the planning process - before the predicate ranges are pushed down, and statistics are estimated for each index. The pruning must occur before this, because that is where the majority of the cost in planning is for single table queries - that is, if we wait until after this information is available, the pruning can be more precise but the performance gains are insufficient. Therefore, this early pruning uses a ranking system based upon column coverage (existence of columns in the index).

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Created a python script to create a table with 500 indexes:
```
columns = [chr(i) for i in range(ord('a'), ord('z') + 1)] + \
          ['a' + chr(i) for i in range(ord('a'), ord('z') + 1)] + \
          ['b' + chr(i) for i in range(ord('a'), ord('z') + 1)] + \
          ['c' + chr(i) for i in range(ord('a'), ord('z') + 1)] + \
          ['d' + chr(i) for i in range(ord('a'), ord('z') + 1)] + \
          ['e' + chr(i) for i in range(ord('a'), ord('z') + 1)]

index_statements = []
index_count = 0

for i in range(len(columns)):
    for j in range(i + 1, len(columns)):
        if index_count >= 500:
            break
        index_name = f"idx_{columns[i]}_{columns[j]}"
        index_statements.append(f"  KEY `{index_name}` (`{columns[i]}`, `{columns[j]}`)")
        index_count += 1

# Join the index statements with commas
indexes_sql = ",\n".join(index_statements)

# Create the full table DDL
ddl = f"CREATE TABLE `t500_indexes` (\n  " + \
      ",\n  ".join([f"`{col}` int" for col in columns]) + ",\n" + \
      indexes_sql + "\n);"

print(ddl)
```

Then tested the following 4 queries with tidb_opt_prune_index_threshold=600, 20 & 5:
```
explain select b from t500_indexes where a = 1 and (b = 12 or b = 13 or b = 15 or b = 17 or b in (0,1,2,3,4,5,6,7,8,9) or b = 11) order by c;
explain select b from t500_indexes where a = 1 and (b = 12 or b = 13 or b = 15 or b = 17 or b in (0,1,2,3,4,5,6,7,8,9) or b = 11);
explain select * from t500_indexes where (a = 1 and f = 1)  or (a = 2 and f = 1)  or (a = 3 and f = 1)  or (a = 4 and f = 1)  or (a = 5 and f = 1)  or (a = 6 and f = 1)  or (a = 7 and f = 1)  or (a = 8 and f = 1)  or (a = 9 and f = 1) or (a = 1 and f = 2)  or (a = 2 and f = 2)  or (a = 3 and f = 2)  or (a = 4 and f = 2)  or (a = 5 and f = 2)  or (a = 6 and f = 2)  or (a = 7 and f = 2)  or (a = 8 and f = 2)  or (a = 9 and f = 2);
Explain select t1.* from t500_indexes t1 join t500_indexes t2 on t1.a = t2.a where t1.d > 5 and t1.c = 5;
```

Performance results are:
- Q1 - 600 == 81 msec, 20 == 39 msec, 5 == 35 msec
- Q2 - 600 == 82 msec, 20 == 25 msec, 5 == 26 msec
- Q3 - 600 == 317 msec, 20 == 55 msec, 5 == 52 msec
- Q4 - 600 == 54 msec, 20 == 36 msec, 5 == 26 msec
These are simple examples - but no index changes occurred in the plans as the tidb_opt_index_prune_threshold was changed.

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


pkg/planner/core/stats.go
```diff
@@ -767,3 +778,348 @@ func loadTableStats(ctx sessionctx.Context, tblInfo *model.TableInfo, pid int64)
...
+					}
+					continue // Move to next index column - since a column can only appear once
+				}
+			}
+			// Check if this index column matches a join column
+			if totalJoinColumns > 0 {
+				if _, found := joinColIDs[idxColID]; found {
+					coveredJoinCount++
+					// Track consecutive matches from start of index
+					if i == consecutiveJoinCount+consecutiveWhereCount {
```
qw4990: How to understand `i == consecutiveJoinCount+consecutiveWhereCount`, is this for IndexJoin? When considering how many columns an index can match for an IndexJoin, we need to count both Join Keys and Where Predicates like `index(t2, a, b)` can match 2 columns for `where t1.a=t2.a and t2.b=1`.
Is this for that purpose?

pkg/planner/core/stats.go
```diff
@@ -767,3 +778,348 @@ func loadTableStats(ctx sessionctx.Context, tblInfo *model.TableInfo, pid int64)
...
+
+	// Build result with priority: table paths, forced indexes, perfect covering, then preferred
+	result := make([]*util.AccessPath, 0, maxIndexes)
+
+	// CRITICAL: Always include table paths - this is mandatory for correctness
+	result = append(result, tablePaths...)
+
+	// Only sort perfect covering indexes if we have more than we can use
+	// If we'll keep all of them, no need to sort
+	if len(perfectCoveringIndexes) > maxToKeep {
```
qw4990: Should be `if len(result) + len(perfrectCoveringIndexes) > maxToKeep`? If the current length of `result` is 1 and the length of `perfectIndexes` is 20 and `maxToKeep` is 20, we also need to sort `perfectIndexes` and remove the one with lowest score.

pkg/planner/core/stats.go
```diff
@@ -767,3 +778,348 @@ func loadTableStats(ctx sessionctx.Context, tblInfo *model.TableInfo, pid int64)
...
+		}
+
+		lenPerfectCoveringIndexes := len(perfectCoveringIndexes)
+		// If we have enough perfect covering indexes, don't add to preferred indexes
+		// unless the current index has more consecutive matches than the max consecutive found so far.
+		if lenPerfectCoveringIndexes >= maxToKeep &&
+			consecutiveWhereCount <= maxConsecutiveWhere &&
+			consecutiveJoinCount <= maxConsecutiveJoin {
+			continue
+		}
```
qw4990: It seems like even if we put this path into `preferred` due to `consecutive <= maxConsecutive`, this path is not going to appear in the result, since we always put `perfect` into `results` before `preferred` and `lenPerfect` is equal or larger than `maxToKeep` now.

pkg/planner/core/stats.go
terry1purcell: Yes - that is correct.

pkg/planner/core/stats.go
terry1purcell: Good point. I have changed it to instead consider "maxToKeep" as the number of additional index plans to add to existing table plans. Therefore, if "maxToKeep == 10", and we already have 1 table plan, then the result will allow 11 plans.

pkg/planner/core/stats.go
terry1purcell: Good catch. I have changed this to ">= minToKeep" - so that we can add them to preferred list, which would then be considered if we don't fill the perfect list.

terry1purcell: I also added it earlier while looping through the index columns so that it can exit early for indexes that cannot beat other indexes that have already been added.

terry1purcell:
/hold

I need to provide some performance results to post in this PR with 500 indexes

pkg/planner/core/stats.go
```diff
@@ -130,6 +130,15 @@ func deriveStats4DataSource(lp base.LogicalPlan) (*property.StatsInfo, bool, err
 	for i, expr := range ds.PushedDownConds {
 		ds.PushedDownConds[i] = expression.EliminateNoPrecisionLossCast(exprCtx, expr)
 	}
+	// Prune indexes based on WHERE and ordering columns (ORDER BY, MIN/MAX/FIRST_VALUE) if we have too many
+	threshold := ds.SCtx().GetSessionVars().OptIndexPruneThreshold
+	if len(ds.AllPossibleAccessPaths) > threshold && (len(ds.WhereColumns) > 0 || len(ds.OrderingColumns) > 0 || len(ds.JoinColumns) > 0) {
```
qw4990: Almost LGTM, but just a suggestion that could we implement this as a separate logical rule like `rule/large_number_index_pruning`? And then converge all logic of this process like collecting interesting columns, pruning indexes and scoring indexes into that rule, which should make this feature easier to maintain in the future.

pkg/planner/core/stats.go
```diff
@@ -710,3 +721,385 @@ func derivePathStatsAndTryHeuristics(ds *logicalop.DataSource) error {
...
+					if i == consecutiveWhereCount {
+						consecutiveWhereCount++
+					}
+					// Also increment consecutive join count if there has already been at least one
+					// consecutive join match, and the current index column position matches the
+					// sum of consecutive join and where matches.
+					if consecutiveJoinCount > 0 && i == consecutiveJoinCount+consecutiveWhereCount {
+						consecutiveJoinCount++
+					}
+					continue // Move to next index column - since a column can only appear once
```
qw4990: It seems this `continue` also skips the following process of handling join and ordering column, is this expected?

pkg/planner/core/stats.go
```diff
@@ -710,3 +721,385 @@ func derivePathStatsAndTryHeuristics(ds *logicalop.DataSource) error {
...
+//
+// NOTE: This is an APPROXIMATE pruning - to find "interesting" indexes, not the "best" index.
+//
+// This pruning is controlled by variable - tidb_opt_index_prune_threshold.
+// Pruning occurs before many of the index fields are populated - to avoid the cost
+// of building predicate ranges. Therefore, pruning is an approximation based upon
+// column coverage only, not the actual predicate ranges or ordering requirements.
+// The target of this pruning is for customers requiring a very large number of
+// of indexes (>100) on a table, which causes excessive planning time.
+func pruneIndexesByWhereAndOrder(ds *logicalop.DataSource, paths []*util.AccessPath, whereColumns, orderingColumns, joinColumns []*expression.Column, threshold int) []*util.AccessPath {
```
qw4990: This function seems too long 😿  could you please split this into a few smaller functions? And this could also help others understand this process better. 🙏

terry1purcell:
/hold

I need to work through the review comments from @qw4990

pkg/planner/core/stats.go
terry1purcell: Done

pkg/planner/core/rule/rule_prune_indexes.go
```diff
@@ -0,0 +1,484 @@
...
+		if _, exists := req.joinColIDs[col.ID]; !exists {
+			req.whereColIDs[col.ID] = struct{}{}
+			req.totalWhereColumns++
+		}
+	}
+
+	// Build ordering column IDs (exclude WHERE columns to avoid double-counting)
+	for _, col := range orderingColumns {
+		if _, exists := req.whereColIDs[col.ID]; !exists {
+			req.orderingColIDs[col.ID] = struct{}{}
```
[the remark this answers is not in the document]
terry1purcell: Cursor disagrees

pkg/planner/core/rule/rule_collect_plan_stats.go
```diff
@@ -189,6 +195,68 @@ func (*CollectPredicateColumnsPoint) markAtLeastOneFullStatsLoadForEachTable(
...
+		c.collectAndPruneDataSources(child, keptIndexIDs)
+	}
+}
+
+// pruneIndexesForDataSource prunes indexes for a single DataSource if needed.
+func (c *CollectPredicateColumnsPoint) pruneIndexesForDataSource(ds *logicalop.DataSource, keptIndexIDs map[int64]map[int64]struct{}) {
+	threshold := ds.SCtx().GetSessionVars().OptIndexPruneThreshold
+	if threshold < 0 {
+		return
+	}
```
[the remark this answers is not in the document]
terry1purcell: It has been changed to support -1, such that -1 disables this functionality.

pkg/planner/core/rule/rule_prune_indexes.go
```diff
@@ -0,0 +1,484 @@
...
+			req.totalOrderingColumns++
```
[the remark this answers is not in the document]
terry1purcell: This should not occur. Because if a column is both join and ordering, it can't be used to avoid a sort. So I am ignoring this recommendation.

pkg/planner/core/rule/rule_prune_indexes.go
```diff
@@ -0,0 +1,461 @@
...
```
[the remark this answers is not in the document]
terry1purcell: Previously answered. For join + ordering - you can't avoid a sort from join + ordering columns - so these should be kept separate. Ignoring this recommendation.

pkg/planner/core/rule/rule_prune_indexes.go
```diff
@@ -0,0 +1,484 @@
...
+
+	return req
+}
+
+// scoreIndexPath calculates coverage metrics for a single index path.
+func scoreIndexPath(path *util.AccessPath, req columnRequirements) indexWithScore {
+	score := indexWithScore{path: path}
+
+	for i, idxCol := range path.FullIdxCols {
+		idxColID := idxCol.ID
```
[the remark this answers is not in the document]
terry1purcell: Cursor disagrees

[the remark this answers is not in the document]
terry1purcell: cursor disagrees

pkg/planner/core/rule/rule_prune_indexes.go
```diff
@@ -0,0 +1,461 @@
...
+		// Check if this index column matches a WHERE column
+		if req.totalWhereColumns > 0 {
+			if _, found := req.whereColIDs[idxColID]; found {
+				score.whereCount++
+				if i == score.consecutiveWhereCount {
+					score.consecutiveWhereCount++
+				}
+				if score.consecutiveJoinCount > 0 && i == score.consecutiveJoinCount+score.consecutiveWhereCount {
+					score.consecutiveJoinCount++
+				}
```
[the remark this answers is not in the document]
terry1purcell: I disagree with this review comment. This code is attempting to allow where + join predicates to be counted towards the consecutive join count for an index join.

pkg/planner/core/rule/collect_column_stats_usage.go
```diff
@@ -205,6 +293,55 @@ func (c *columnStatsUsageCollector) collectPredicateColumnsForUnionAll(p *logica
...
+		// Extract and store join columns to propagate to DataSource
+		joinExprs := make([]expression.Expression, 0, len(x.EqualConditions)+len(x.OtherConditions))
+		for _, cond := range x.EqualConditions {
+			joinExprs = append(joinExprs, cond)
+		}
+		for _, cond := range x.OtherConditions {
+			joinExprs = append(joinExprs, cond)
+		}
+		c.joinColumns = expression.ExtractColumnsFromExpressions(joinExprs, nil)
+	}
```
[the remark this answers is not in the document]
terry1purcell: I implemented this suggestion but it actually broke the code. The reason is because join columns for table1 can appear as the left or right predicates associated with table1, or another table (such as table2) that table1 is joined to. So while this can make sense for ordering columns, this comment is incorrect for join columns.

pkg/planner/core/rule/rule_prune_indexes.go
```diff
@@ -0,0 +1,463 @@
...
```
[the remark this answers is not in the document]
terry1purcell: This comment is incorrect. For a nested loop join (or index join) - join columns can combine where + join columns.

IANTHEREAL:
@terry1purcell I’ve invited Pantheon-bot to review this PR. The bot is tuned to prioritize severe bugs and structural issues rather than overwhelming you with minor details. If anything about its feedback could be improved, please let us know — your input is important.

PR #63921, opened 2025-10-13
planner: PointGet will not skip the reuse chunk with enough total memory

### LINKED ISSUE #63920: Pointget will inappropriately skip the reuse chunk
## Enhancement

https://github.com/pingcap/tidb/blob/1e571855e1be7f37b861eea60ddaed5f67b671de/pkg/planner/core/optimizer.go#L1244-L1265

Even if the machine has sufficient memory, pointget will skip the reuse chunk when encountering large types.

hawkingrei:

Issue Number: close #63920 

Problem Summary:

This client has a point-get Update that leads to increased memory allocation, higher GC overhead, and a surge in CPU utilization.

The ```disableReuseChunkIfNeeded```  in ```postOptimize``` detects if there is a large-type column and causes the query to skip the reuse chunk. This is the reason for the increased memory allocation and CPU surge. Additionally, this client actually has 128GB of system memory, and using less than 50GB is not a concern. Therefore, there is no need to disable the ability to reuse chunks out of fear of OOM

### Check List

Tests 

- [x] Unit test
- [ ] Integration test

load the customer plan replayed and executor the query.

```
select @@last_sql_use_alloc;
1
```

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/optimizer.go
+++ b/pkg/planner/core/optimizer.go
@@ -1253,13 +1253,13 @@ func checkOverlongColType(sctx base.PlanContext, plan base.PhysicalPlan) (skipRe
 
 var (
 	// MaxMemoryLimitForOverlongType is the memory limit for overlong type column check.
-	// Why is it not 512 ?
+	// Why is it not 128 ?
 	// Because many customers allocate a portion of memory to their management programs,
-	// the actual amount of usable memory does not align to 512GB.
-	// TODO: We are also lacking test data for instances with less than 512 GB of memory, so we need to plan the rules here.
+	// the actual amount of usable memory does not align to 128GB.
+	// TODO: We are also lacking test data for instances with less than 128GB of memory, so we need to plan the rules here.
 	// TODO: internal sql can force to use chunk reuse if we ensure the memory usage is safe.
 	// TODO: We can consider the limit/Topn in the future.
-	MaxMemoryLimitForOverlongType = 500 * size.GB
+	MaxMemoryLimitForOverlongType = 120 * size.GB
 	maxFlenForOverlongType        = mysql.MaxBlobWidth * 2
 )
 


pkg/planner/core/optimizer.go
```diff
@@ -1232,34 +1234,80 @@ func checkOverlongColType(sctx base.PlanContext, plan base.PhysicalPlan) bool {
...
+	// Because many customers allocate a portion of memory to their management programs,
+	// the actual amount of usable memory does not align to 512GB.
+	// It is only changed by tests.
+	maxMemoryLimitForOverlongType = 500 * size.GB
+	maxFlenForOverlongType        = mysql.MaxBlobWidth * 2
+)
+
 // existsOverlongType Check if exists long type column.
-func existsOverlongType(schema *expression.Schema) bool {
+func existsOverlongType(schema *expression.Schema, pointGet bool) bool {
```
winoros, line 1272: Add some comments explaining why we need a different check for point-get.

pkg/planner/core/optimizer.go
```diff
@@ -1232,34 +1234,80 @@ func checkOverlongColType(sctx base.PlanContext, plan base.PhysicalPlan) bool {
...
+				}
+				continue
 			}
+			return true
 		}
 	}
 	return false
 }
+
+func checkOverlongType(totalFlen int) bool {
```
winoros:
```suggestion
func checkOverlongTypeForPointGet(totalFlen int) bool {
```

pkg/planner/core/optimizer.go
hawkingrei: Great, I have added.

pkg/planner/core/optimizer.go
hawkingrei: Done

pkg/planner/core/optimizer.go
```diff
@@ -1232,34 +1234,76 @@ func checkOverlongColType(sctx base.PlanContext, plan base.PhysicalPlan) bool {
...
 	}
 	return false
 }
+
+func checkOverlongTypeForPointGet(totalFlen int) bool {
+	totalMemory, err := memory.MemTotal()
+	if err != nil || totalMemory <= 0 {
+		return true
+	}
+	if totalMemory >= maxMemoryLimitForOverlongType {
```
AilinKid: a little bit hard code depend on the customer env？

AilinKid: what we want here is memory is enough and since point get is relative short, we can still reuse the chunk

hawkingrei: Yes

hawkingrei: The issue here is that we don't have enough user scenarios and data. I can only set such a threshold based on this customer's scenario and data.

YangKeao (approved):
LGTM

I'll going to improve the logic of column reuse later. It's still fine for this PR to merge.

PR #21873, opened 2020-12-18
ddl: filter the feature switches in show variables

AilinKid:
Signed-off-by: AilinKid <314806019@qq.com>

### What problem does this PR solve?

 filter the feature switches(like tidb-enable-change-multi-schema) in show variables.

### What is changed and how it works?

 filter the feature switches(like tidb-enable-change-multi-schema) in show variables.

### Check List 

Tests 

- Unit test

### Release note 

- ddl: filter the feature switches in show variables.

PR #64034, opened 2025-10-19
statistics: refresh stats concurrently

0xPoe:

Problem Summary:

Although we shouldn’t encourage users to execute this command multiple times concurrently, we need to ensure that it’s safe to do so.

In this PR, I removed the non–thread-safe initialization and replaced it with a session pool–based initialization for statistics.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/server/stat_test.go
+++ b/pkg/server/stat_test.go
@@ -120,7 +120,7 @@ func TestInitStatsSessionBlockGCCanBeCanceled(t *testing.T) {
 	require.NoError(t, err)
 
 	infoSyncer := dom.InfoSyncer()
-	// This will cause the session never be created because we do not set session manager.
+	// This prevents the session from being created because we do not set a session manager.
 	infoSyncer.SetSessionManager(nil)
 	h := dom.StatsHandle()
 	ctx, cancel := context.WithCancel(context.Background())


pkg/session/syssession/pool.go
```diff
@@ -265,6 +281,23 @@ func (p *AdvancedSessionPool) WithSession(fn func(*Session) error) error {
 		}
 	}()
 
+	// Make sure the internal session is registered to the session manager to block GC.
+	const retryInterval = 100 * time.Millisecond
```
0xPoe, line 299: /cc @lcwangchao 

Not sure if you’re interested in taking a look at this change. If you have any thoughts on it, please feel free to let me know.

I think the fact that StoreInternalSession is not guaranteed to succeed is a bad design. I’ll try to create an issue for it to explore if we can change it in the future.

mjonss:
Is the following test failure issue in mysql-test related to this PR?
```
[2025/10/27 19:25:48.011 +08:00] [INFO] [conn.go:1200] ["command dispatched failed"] [conn=2098692] [session_alias=] [connInfo="id:2098692, addr:127.0.0.1:43836 status:10, collation:utf8mb4_general_ci, user:root"] [command=Query] [status="inTxn:0, autocommit:1"] [sql="ANALYZE TABLE a,d,bb,cc,dd;"] [txn_mode=PESSIMISTIC] [timestamp=461783524398596096] [err="pessimistic lock retry limit reached\ngithub.com/pingcap/tidb/pkg/executor.(*AnalyzeExec).handleResultsErrorWithConcurrency\n\t/home/jenkins/agent/workspace/pingcap/tidb/ghpr_mysql_test/tidb/pkg/executor/analyze.go:485\ngithub.com/pingcap/tidb/pkg/executor.(*AnalyzeExec).handleResultsError\n\t/home/jenkins/agent/workspace/pingcap/tidb/ghpr_mysql_test/tidb/pkg/executor/analyze.go:419\ngithub.com/pingcap/tidb/pkg/executor.(*AnalyzeExec).Next.func2\n\t/home/jenkins/agent/workspace/pingcap/tidb/ghpr_mysql_test/tidb/pkg/executor/analyze.go:133\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\t/go/pkg/mod/golang.org/x/sync@v0.16.0/errgroup/errgroup.go:93\nruntime.goexit\n\t/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.12.linux-amd64/src/runtime/asm_amd64.s:1700"]
```

elsa0520:
Did we write in the documentation that we do not recommend users use the concurrent refresh stats function?

0xPoe:
> Did we write in the documentation that we do not recommend users use the concurrent refresh stats function?

See: [pingcap/docs@`1e76212` (#21873)](https://github.com/pingcap/docs/pull/21873/commits/1e762123960d75d70c328db1e8345ef0ec40fa76)

pkg/session/syssession/pool.go
```diff
@@ -265,6 +281,23 @@ func (p *AdvancedSessionPool) WithSession(fn func(*Session) error) error {
...
+	if forceBlockGC && !infosync.ContainsInternalSession(se.internal.sctx) {
+		for !infosync.StoreInternalSession(se.internal.sctx) {
+			// In most cases, the session manager is not set, so this step will be skipped.
+			// It is only enabled explicitly in tests through a failpoint.
+			forceBlockGCInTest := !intest.InTest
+			failpoint.Inject("ForceBlockGCInTest", func(val failpoint.Value) {
```
elsa0520: The failpoint inject can inject a function logic for test. Maybe it is better than now
```
failpoint.Inject("MockForceBlockGCInTest", func() {
    if (xxxx)
      for (xxx) {
    time.Sleep(retryInterval)
}
})


```

pkg/session/syssession/pool.go
```diff
@@ -265,6 +281,23 @@ func (p *AdvancedSessionPool) WithSession(fn func(*Session) error) error {
...
+				forceBlockGCInTest = val.(bool)
+			})
+			if !forceBlockGCInTest {
+				break
+			}
+			time.Sleep(retryInterval)
```
elsa0520: This way of writing out the normal procedure looks a bit risky, but if you're sure tidb won't be waiting here indefinitely, that's fine too.

0xPoe: Updated in [`d37acfa` (#64034)](https://github.com/pingcap/tidb/pull/64034/commits/d37acfa1297dc376c2d413227d2438128613fa1c)

pkg/session/syssession/pool.go
elsa0520: Or this is better 
```
	if intest.Intest {
		var forceBlockGCInTest bool
		failpoint.Inject("ForceBlockGCInTest", func(val failpoint.Value) {
			forceBlockGCInTest = val.(bool)
		})
		if !forceBlockGCInTest {
			break
		}
	}
```

pkg/domain/infosync/info.go
```diff
@@ -1011,9 +1011,9 @@ func DeleteInternalSession(se any) {
 	sm.DeleteInternalSession(se)
 }
 
-// ContainsInternalSessionForTest is the entry function for check whether an internal session is in Manager.
+// ContainsInternalSession is the entry function for check whether an internal session is in Manager.
 // It is only used for test.
```
mjonss, line 1015: If still only used in tests, could it be moved to a _test.go file?

pkg/executor/simple_test.go
```diff
@@ -334,3 +335,77 @@ func TestRefreshStatsWithLiteMode(t *testing.T) {
...
+				for _, sql := range sqls {
+					tkWorker.MustExec(sql)
+				}
+			}
+		}(tkWorker)
+	}
+	wg.Wait()
+
+	// Ensure we end in a full-load state deterministically.
+	tk.MustExec("REFRESH STATS *.* FULL")
```
mjonss: (minor) Why is this needed, if the last statement in `sqls` is the same?

pkg/statistics/handle/bootstrap.go
```diff
@@ -98,10 +98,10 @@ func genInitStatsMetaSQL(tableIDs ...int64) string {
 	return selectPrefix + whereClausePrefix + inListStr + whereClauseSuffix
 }
 
-func (h *Handle) initStatsMeta(ctx context.Context, tableIDs ...int64) (statstypes.StatsCache, int64, error) {
+func (h *Handle) initStatsMeta(ctx context.Context, sctx sessionctx.Context, tableIDs ...int64) (statstypes.StatsCache, int64, error) {
 	ctx = kv.WithInternalSourceType(ctx, kv.InternalTxnStats)
 	sql := genInitStatsMetaSQL(tableIDs...)
```
mjonss, line 103: Not related to PR, but I am curious, what is the max number of tableIDs that can come in this call? Could it be an issue that the SQL string becomes too long?

pkg/statistics/handle/bootstrap.go
```diff
@@ -745,26 +745,34 @@ func (h *Handle) initStatsBucketsConcurrently(cache statstypes.StatsCache, total
...
+		return se.WithSessionContext(func(sctx sessionctx.Context) error {
+			return h.initStatsLiteWithSession(ctx, sctx, tableIDs...)
+		})
+	})
+}
+
+func (h *Handle) initStatsLiteWithSession(ctx context.Context, sctx sessionctx.Context, tableIDs ...int64) (err error) {
 	defer func() {
-		_, err1 := util.Exec(h.initStatsCtx, "commit")
+		_, err1 := util.Exec(sctx, "commit")
```
mjonss, line 758: Not directly related to this PR, but should it not do "rollback" in case of error?
And if only doing SELECTs, why not default to "rollback"? Or even use auto_commit mode, unless we really need snapshot consistency of the statistics, which I assume could be the case?

pkg/domain/infosync/info.go
0xPoe: No, it is also used in the normal path. That is why I renamed it.

pkg/executor/simple_test.go
0xPoe: That makes sense. No need to do it here. It always executed in order.

pkg/statistics/handle/bootstrap.go
0xPoe: It shouldn't be too many. Because usually normal users would never use this command. And BR will only use `REFRESH *.*`.

pkg/statistics/handle/bootstrap.go
0xPoe: I think we always do it here https://github.com/0xPoe/tidb/blob/poe-patch-refresh-stats-concurrently/pkg/session/syssession/pool.go#L220. But you made a good point. We should consider rolling back here.

0xPoe: I’ll create an issue for it. I’d like to separate it from this PR.

0xPoe: https://github.com/pingcap/tidb/issues/64364

GMHDBJD (approved):
LGTM for domain changes

Issue #64008, opened 2025-10-17
Diagnosability enhacement for TiDB X


ekexium:
## Enhancement
To boost problem investigation efficiency for TiDB X, we need to enhance the diagnosability.
Currently, we would like to enhance diagnosability in two main aspects:
1. For known anomalies that may happen, we perform defensive checks. Examples are assertions, max_ts update validity checks, stale read ts checks.
2. For unknown anomaly, we need event traces to narrow down the problem area.

PR #64051, opened 2025-10-20
*: introduce the traceevent package

ekexium:

Issue Number: ref #64008

Problem Summary:

Introduce the traceevent package, to trace critical events to boost problem investigation.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Introduce the event tracing ability and a system variable `tidb_trace_event`.
```

### DIFF
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -2411,7 +2411,7 @@ func runStmt(ctx context.Context, se *session, s sqlexec.Statement) (rs sqlexec.
 		if traceevent.IsEnabled(traceevent.StmtLifecycle) {
 			stmtCtx := se.sessionVars.StmtCtx
 			sqlDigest, _ := stmtCtx.SQLDigest()
-			normalized, planDigest := stmtCtx.GetPlanDigest()
+			_, planDigest := stmtCtx.GetPlanDigest()
 			var planDigestHex string
 			if planDigest != nil {
 				planDigestHex = hex.EncodeToString(planDigest.Bytes())
@@ -2426,8 +2426,6 @@ func runStmt(ctx context.Context, se *session, s sqlexec.Statement) (rs sqlexec.
 			if err != nil {
 				fields = append(fields, zap.Error(err))
 			}
-			// Avoid logging normalized plan in finish event (it's in plan.digest event)
-			_ = normalized
 			traceevent.TraceEvent(ctx, traceevent.StmtLifecycle, "stmt.finish", fields...)
 		}
 	}()
--- a/pkg/util/traceevent/traceevent.go
+++ b/pkg/util/traceevent/traceevent.go
@@ -19,7 +19,7 @@ import (
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
-	"math/rand"
+	"math/rand/v2"
 	"strconv"
 	"strings"
 	"sync"


pkg/util/traceevent/traceevent.go
```diff
@@ -0,0 +1,493 @@
...
+	// TxnLockResolve traces lock resolution and conflict handling.
+	TxnLockResolve
+	// StmtLifecycle traces statement start/finish events.
+	StmtLifecycle
+	// StmtPlan traces statement plan digest and optimization.
+	StmtPlan
+	// KvRequest traces client-go kv request and responses
+	KvRequest
+	// UnknownClient is the fallback category of events from client-go
+	UnknownClient
```
tiancaiamao: What's this?

pkg/util/traceevent/traceevent.go
```diff
@@ -0,0 +1,493 @@
...
+	if !IsEnabled(category) {
+		return
+	}
+
+	event := Event{
+		Category:  category,
+		Name:      name,
+		Timestamp: time.Now(),
+		TraceID:   extractTraceID(ctx),
+		Fields:    copyFields(fields),
```
tiancaiamao: What's the purpose of this copy?

pkg/util/traceevent/traceevent.go
```diff
@@ -0,0 +1,493 @@
...
+	if !loggingEnabled.Load() {
+		return
+	}
+	baseFields := make([]zap.Field, 0, len(event.Fields)+3)
+	baseFields = append(baseFields, zap.String("category", getCategoryName(event.Category)))
+	baseFields = append(baseFields, zap.Time("event_ts", event.Timestamp))
+	if len(event.TraceID) > 0 {
+		baseFields = append(baseFields, zap.String("trace_id", hex.EncodeToString(event.TraceID)))
+	}
+	baseFields = append(baseFields, event.Fields...)
```
tiancaiamao: we can deliberately reserve 3 more fields (like in `copyFields`) and just append based on the old event.fields, avoiding making a new slice and copy the old data.

tiancaiamao: Or we can use a temporary buf in LogSink:

```
type LogSink struct {
    avoidAllocation []Fields
}
```

pkg/util/traceevent/traceevent.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+}
+
+// Record implements the Sink interface.
+func (r *RingBufferSink) Record(_ context.Context, event Event) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	eventCopy := event
+	eventCopy.Fields = copyFields(event.Fields)
```
tiancaiamao: Why this copy? Just store the Event should be safe, as long as we don't reuse the event.Fields ...

tiancaiamao: B.T.W, this is Go, this is not Rust ...
So there are some thing differs significantly.
Rust use RAII and borrow check, and that's cheap.

Object allocation in Go is heavy due to its GC
For example, here 

```
	r.mu.Lock()
	defer r.mu.Unlock()
```

and then allocation happen, in the fast path, it's just as fast as Rust.
But in the slow path, Go may found GC recycling can't catch up object allocating, 
Then it requires gcAssist, i.e. this user goroutine run some GC related code

Then, the lock could hold for a longer time, blocking other goroutines whenever they acquire this lock. BOOM!  it slow down the whole P9999 and it's hard to know why.

pkg/session/session.go
```diff
@@ -982,6 +1015,19 @@ func (s *session) RollbackTxn(ctx context.Context) {
 	r, ctx := tracing.StartRegionEx(ctx, "session.RollbackTxn")
 	defer r.End()
 
+	// Emit txn.rollback trace event
+	if traceevent.IsEnabled(traceevent.TxnLifecycle) {
+		startTS := s.sessionVars.TxnCtx.StartTS
+		stmtCount := uint64(s.sessionVars.TxnCtx.StatementCount)
+		isPessimistic := s.txn.IsPessimistic()
```
tiancaiamao:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x98 pc=0x103fed918]

goroutine 605 [running]:
github.com/pingcap/tidb/pkg/session.(*session).RollbackTxn(0x14001029088, {0x105b4fac0?, 0x109be3448?})
        /Users/genius/project/tidb/pkg/session/session.go:1022 +0xb8
github.com/pingcap/tidb/pkg/ddl/session.(*Pool).Put(0x1400228f8c0, {0x105bf13e0, 0x14001029088})
        /Users/genius/project/tidb/pkg/ddl/session/session_pool.go:84 +0x7c
github.com/pingcap/tidb/pkg/ddl/systable.(*manager).withNewSession(0x140000b1aa0, 0x140031f1de0)
        /Users/genius/project/tidb/pkg/ddl/systable/manager.go:74 +0xd0
github.com/pingcap/tidb/pkg/ddl/systable.(*manager).GetMinJobID(0x0?, {0x105b4fb30?, 0x140028b6230?}, 0x0?)
        /Users/genius/project/tidb/pkg/ddl/systable/manager.go:130 +0x4c
github.com/pingcap/tidb/pkg/ddl/systable.(*MinJobIDRefresher).refresh(0x0?, {0x105b4fb30?, 0x140028b6230?})
        /Users/genius/project/tidb/pkg/ddl/systable/min_job_id.go:66 +0x4c
github.com/pingcap/tidb/pkg/ddl/systable.(*MinJobIDRefresher).Start(0x14002117a70, {0x105b4fb30, 0x140028b6230})
        /Users/genius/project/tidb/pkg/ddl/systable/min_job_id.go:54 +0x38
github.com/pingcap/tidb/pkg/ddl.(*ddl).Start.func2()
        /Users/genius/project/tidb/pkg/ddl/ddl.go:840 +0x30
github.com/pingcap/tidb/pkg/util.(*WaitGroupWrapper).Run.func1()
        /Users/genius/project/tidb/pkg/util/wait_group_wrapper.go:167 +0x58
created by github.com/pingcap/tidb/pkg/util.(*WaitGroupWrapper).Run in goroutine 1
        /Users/genius/project/tidb/pkg/util/wait_group_wrapper.go:165 +0x7c
```

tiancaiamao:
I found some place discard the ctx... or maybe using context.Background

For example, `request_source=leader_internal_DistTask`

> [2025/10/24 22:38:47.187 +08:00] [INFO] [endpoint.rs:670] ["Coprocessor request entry"] [peer="Some(\"ipv4:192.168.0.226:53830\")"] [request_source=leader_internal_DistTask] [last_range_end=78FFFFFE748000FFFFFFFFFFFA5F698000000000000002013139322E3136382EFF302E3232363A3430FF3030000000000000FA] [last_range_start=78FFFFFE748000FFFFFFFFFFFA5F698000000000000002013139322E3136382EFF302E3232363A3430FF3030000000000000F9] [first_range_end=78FFFFFE748000FFFFFFFFFFFA5F698000000000000002013139322E3136382EFF302E3232363A3430FF3030000000000000FA] [first_range_start=78FFFFFE748000FFFFFFFFFFFA5F698000000000000002013139322E3136382EFF302E3232363A3430FF3030000000000000F9] [range_count=1] [region_id=114] [keyspace_id=16777214] [start_ts=461718612125417473] [trace_id=None]

pkg/util/traceevent/traceevent.go
ekexium: To tolerate errors or laggings when mapping the category from client-go type to tidb type. For example, someone updates client-go that has a new category but doesn't update the mapping function

pkg/util/traceevent/traceevent.go
ekexium: You are right this is copy redundant. Fixed.

pkg/util/traceevent/traceevent.go
ekexium: To prevent the fields from being modified by others. I guess for most of the case the fields won't change or are not mutable. But I suppose there is such a risk for example when passing in some zap.Binary. Any suggestion on this?

pkg/util/traceevent/traceevent.go
ekexium: Makes sense. PTAL

pkg/util/traceevent/traceevent.go
```diff
@@ -0,0 +1,509 @@
...
+func (*LogSink) Record(ctx context.Context, event Event) {
+	if !loggingEnabled.Load() {
+		return
+	}
+
+	// Append to reserved capacity without allocation.
+	// Field order: [event fields] [category] [timestamp] [trace_id?]
+	fields := event.Fields
+	fields = append(fields, zap.String("category", getCategoryName(event.Category)))
+	fields = append(fields, zap.Time("event_ts", event.Timestamp))
```
tiancaiamao: How about display it as a number in microsecond? https://pkg.go.dev/time#Time.UnixMicro
zap.Time will be displayed as millisecond
We already has the time information in millisecond in log, but millisecond is too corase-grained

pkg/util/traceevent/traceevent.go
```diff
@@ -0,0 +1,509 @@
...
+
+	// Update timestamp for next cooling-off check
+	lastDumpTime.Store(now)
+
+	// Perform full dump
+	logger.Info("dump flight recorder", zap.String("reason", reason), zap.Int("event_count", len(events)))
+	for _, ev := range events {
+		fields := make([]zap.Field, 0, len(ev.Fields)+4)
+		fields = append(fields, zap.String("category", getCategoryName(ev.Category)))
+		fields = append(fields, zap.Time("event_ts", ev.Timestamp))
```
tiancaiamao: And here

pkg/sessionctx/variable/sysvar.go
```diff
@@ -453,6 +454,19 @@ var defaultSysVars = []*SysVar{
 	}, GetGlobal: func(_ context.Context, s *SessionVars) (string, error) {
 		return BoolToOnOff(vardef.ProcessGeneralLog.Load()), nil
 	}},
+	// NOTE: The trace-event switch is experimental. It is subject to changes.
+	{Scope: vardef.ScopeGlobal, Name: vardef.TiDBTraceEvent, Value: vardef.DefTiDBTraceEvent, Type: vardef.TypeStr,
+		SetGlobal: func(_ context.Context, _ *SessionVars, val string) error {
+			if kerneltype.IsClassic() {
```
cfzjywxk, line 460: Should we hide it for classic version as conditional compile is not supported in tidb?

ekexium: Done

pkg/sessionctx/variable/sysvar.go
```diff
@@ -453,6 +454,19 @@ var defaultSysVars = []*SysVar{
 	}, GetGlobal: func(_ context.Context, s *SessionVars) (string, error) {
 		return BoolToOnOff(vardef.ProcessGeneralLog.Load()), nil
 	}},
+	// NOTE: The trace-event switch is experimental. It is subject to changes.
+	{Scope: vardef.ScopeGlobal, Name: vardef.TiDBTraceEvent, Hidden: kerneltype.IsClassic(), Value: vardef.DefTiDBTraceEvent, Type: vardef.TypeStr,
```
ekexium: Do you guys think instance scope is more reasonable? I suppose we rarely want to enable it for all TiDB instances. For tests we usually have 1 TiDB. @cfzjywxk @MyonKeminta @tiancaiamao

tiancaiamao: instance scope looks more reasonable @ekexium

cfzjywxk: Instance and session scopes would probably be used I think.

PR #64053, opened 2025-10-20
planner: maintain a map of columns by ID in DataSource

henrybw:

Issue Number: ref #63856 

Problem Summary: Profiling the optimizer when planning a simple `SELECT` or `EXPLAIN SELECT` query on a table with 500 indexes and 6,500 columns reveals that 26% of the time is spent in `plannercore.PrepareIdxColsAndUnwrapArrayType()`, partly due to this function being passed a slice of columns that the function has to search linearly with a for loop. Instead of scanning this slice, we should be using a map to reduce lookups of column by ID from O(N) to O(1).

* Add a `TblColsByID` map to the logical data source.
* Refactor all places in the planner that append to `ds.TblCols` directly to instead call a new `AppendTableCol()` function that also adds the column to `TblColsByID`.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/logical_plan_builder.go
+++ b/pkg/planner/core/logical_plan_builder.go
@@ -4123,7 +4123,7 @@ func addExtraPhysTblIDColumn4DS(ds *logicalop.DataSource) *expression.Column {
 		ColName:     model.ExtraPhysTblIDName,
 		OrigColName: model.ExtraPhysTblIDName,
 	}))
-	ds.TblCols = append(ds.TblCols, pidCol)
+	ds.AppendTableCol(pidCol)
 	return pidCol
 }
 
@@ -4600,6 +4600,7 @@ func (b *PlanBuilder) buildDataSource(ctx context.Context, tn *ast.TableName, as
 		Columns:                make([]*model.ColumnInfo, 0, countCnt),
 		PartitionNames:         tn.PartitionNames,
 		TblCols:                make([]*expression.Column, 0, countCnt),
+		TblColsByID:            make(map[int64]*expression.Column, countCnt),
 		PreferPartitions:       make(map[int][]ast.CIStr),
 		IS:                     b.is,
 		IsForUpdateRead:        b.isForUpdateRead,
@@ -4629,7 +4630,7 @@ func (b *PlanBuilder) buildDataSource(ctx context.Context, tn *ast.TableName, as
 			handleCols = util.NewIntHandleCols(newCol)
 		}
 		schema.Append(newCol)
-		ds.TblCols = append(ds.TblCols, newCol)
+		ds.AppendTableCol(newCol)
 	}
 	// We append an extra handle column to the schema when the handle
 	// column is not the primary key of "ds".
@@ -4648,7 +4649,7 @@ func (b *PlanBuilder) buildDataSource(ctx context.Context, tn *ast.TableName, as
 				ColName:     model.ExtraHandleName,
 				OrigColName: model.ExtraHandleName,
 			})
-			ds.TblCols = append(ds.TblCols, extraCol)
+			ds.AppendTableCol(extraCol)
 		}
 	}
 	ds.HandleCols = handleCols
--- a/pkg/planner/core/operator/logicalop/logical_datasource.go
+++ b/pkg/planner/core/operator/logicalop/logical_datasource.go
@@ -89,7 +89,8 @@ type DataSource struct {
 	UnMutableHandleCols util.HandleCols
 	// TblCols contains the original columns of table before being pruned, and it
 	// is used for estimating table scan cost.
-	TblCols []*expression.Column
+	TblCols     []*expression.Column
+	TblColsByID map[int64]*expression.Column
 	// CommonHandleCols and CommonHandleLens save the info of primary key which is the clustered index.
 	CommonHandleCols []*expression.Column
 	CommonHandleLens []int
@@ -692,4 +693,11 @@ func isIndexColsCoveringCol(sctx expression.EvalContext, col *expression.Column,
 		}
 	}
 	return false
+}
+
+// AppendTableCol appends a column to the original columns of the table before pruning,
+// accessed through ds.TblCols and ds.TblColsByID.
+func (ds *DataSource) AppendTableCol(col *expression.Column) {
+	ds.TblCols = append(ds.TblCols, col)
+	ds.TblColsByID[col.ID] = col
 }


terry1purcell (approved):

pkg/planner/core/operator/logicalop/logical_datasource.go
```diff
@@ -693,3 +694,10 @@ func isIndexColsCoveringCol(sctx expression.EvalContext, col *expression.Column,
 	}
 	return false
 }
+
+// AppendTableCol appends a column to the original columns of the table before pruning,
+// accessed through ds.TblCols and ds.TblColsByID.
+func (ds *DataSource) AppendTableCol(col *expression.Column) {
```
lance6716, line 700: Is this function needed to be exported (public)?

henrybw: Unfortunately yes, due to the way the planner's packages are organized: this is defined in `logicalop`, but used in `plannercore`. (Even though `logicalop` is a subdirectory of `plannercore`, they're still defined as separate packages.)

pkg/planner/core/operator/logicalop/logical_datasource.go
```diff
@@ -89,7 +89,8 @@ type DataSource struct {
 	UnMutableHandleCols util.HandleCols
 	// TblCols contains the original columns of table before being pruned, and it
 	// is used for estimating table scan cost.
-	TblCols []*expression.Column
+	TblCols     []*expression.Column
+	TblColsByID map[int64]*expression.Column
```
winoros, line 93: Why don't we only keep the new map struct

winoros:
What's the benefit of this(and its following prs)?
One point is that the Column slice in `expression.Schema` and the slice `Columns []*model.ColumnInfo` in `*DataSource` is a little hard to remove because they are also responding for the output order of these columns.
Will we have a siginificant benefit for the wide table cases before we can change these two main struct?

pkg/planner/core/operator/logicalop/logical_datasource.go
henrybw: The map doesn't store any information about the order of the columns, which at least once consumer of `TblCols` depends on (`physicalop.BuildIndexMergeTableScan`).

I was also trying to keep this change as minimally invasive as possible, since there are many places in the planner that use `TblCols` and would not benefit from using this map instead.

henrybw:
> What's the benefit of this(and its following prs)? One point is that the Column slice in `expression.Schema` and the slice `Columns []*model.ColumnInfo` in `*DataSource` is a little hard to remove because they are also responding for the output order of these columns. Will we have a siginificant benefit for the wide table cases before we can change these two main struct?

To be clear, I'm not suggesting that we remove these slices entirely. I am only suggesting that we add maps to replace O(N) lookups (for loops that filter out all but one column) with O(1) lookups.

As noted in the PR, the specific benefit this brings is to remove an O(N) scan in the `plannercore.PrepareIdxColsAndUnwrapArrayType` function. Profiling the time spent in the optimizer repeatedly executing, for 2 minutes, a [query on a table like this](https://github.com/pingcap/tidb/issues/63856#issue-3496225000), but with many more indexes (500) and columns (6,500), this O(N) for loop cumulatively takes over 2 seconds out of the ~25 seconds spent in the function (~8%):

Screenshot 2025-10-21 at 2 13 42 PM

Using a map instead cumulatively takes 290ms out of the ~23 seconds spent in the function (~1%):

Screenshot 2025-10-21 at 2 17 56 PM

(Yes, the repeated call to `tableInfo.Cols()` accounts for the vast majority of the time spent. I am also planning to address this in [an upcoming PR](https://github.com/pingcap/tidb/pull/64086).)

I think it is significantly beneficial that, for wide tables, using a map is almost 8x faster than using a for loop.

EDIT: I have now posted a PR to address the `tableInfo.Cols()` issue: https://github.com/pingcap/tidb/pull/64086)

pkg/planner/core/logical_plan_builder.go
```diff
@@ -4600,6 +4600,7 @@ func (b *PlanBuilder) buildDataSource(ctx context.Context, tn *ast.TableName, as
 		Columns:                make([]*model.ColumnInfo, 0, countCnt),
 		PartitionNames:         tn.PartitionNames,
 		TblCols:                make([]*expression.Column, 0, countCnt),
+		TblColsByID:            make(map[int64]*expression.Column, countCnt),
```
lance6716, line 4603: I'm not sure for a small amount of `countCnt`, the map lookup is slower than basic for-loop check. Can you check it?

henrybw: Sure. I profiled `PrepareIdxColsAndUnwrapArrayType` for two minutes while repeatedly querying a table with 16 columns and 8 indexes (the function won't even show up in the profile if there are fewer columns/indexes). The difference is negligible.

Before:
Screenshot 2025-10-22 at 6 41 23 PM

After:
Screenshot 2025-10-22 at 6 52 41 PM

lance6716 (approved):

winoros:
> > What's the benefit of this(and its following prs)? One point is that the Column slice in `expression.Schema` and the slice `Columns []*model.ColumnInfo` in `*DataSource` is a little hard to remove because they are also responding for the output order of these columns. Will we have a siginificant benefit for the wide table cases before we can change these two main struct?
> 
> To be clear, I'm not suggesting that we remove these slices entirely. I am only suggesting that we add maps to replace O(N) lookups (for loops that filter out all but one column) with O(1) lookups.
> 
> As noted in the PR, the specific benefit this brings is to remove an O(N) scan in the `plannercore.PrepareIdxColsAndUnwrapArrayType` function. Profiling the time spent in the optimizer repeatedly executing, for 2 minutes, a [query on a table like this](https://github.com/pingcap/tidb/issues/63856#issue-3496225000), but with many more indexes (500) and columns (6,500), this O(N) for loop cumulatively takes over 2 seconds out of the ~25 seconds spent in the function (~8%):
> 
> Screenshot 2025-10-21 at 2 13 42 PM
> Using a map instead cumulatively takes 290ms out of the ~23 seconds spent in the function (~1%):
> 
> Screenshot 2025-10-21 at 2 17 56 PM
> (Yes, the repeated call to `tableInfo.Cols()` accounts for the vast majority of the time spent. I am also planning to address this in [an upcoming PR](https://github.com/pingcap/tidb/pull/64086).)
> 
> I think it is significantly beneficial that, for wide tables, using a map is almost 8x faster than using a for loop.
> 
> EDIT: I have now posted a PR to address the `tableInfo.Cols()` issue: #64086)

Got it. LGTM

Issue #64049, opened 2025-10-20
improve performance in the TopN.BetweenCount


hawkingrei:
## Enhancement

<img width="836" height="243" alt="Image" src="https://github.com/user-attachments/assets/646e0073-a34a-491a-8130-411eadf51993" />

<img width="1911" height="267" alt="Image" src="https://github.com/user-attachments/assets/5b8d7b2b-6876-432e-aaa5-5cb2f346c92e" />

`LowerBound` actually traverses `topn` twice, so we can use a method to achieve obtaining two indices with a single traversal.

hawkingrei:
It's best not to write it as a recursive function.

PR #64064, opened 2025-10-21
planner: Optimize TopN.BetweenCount range search using a combined binary search pass

ghoshh:

Issue Number: ref #64049

Problem Summary:
In the TopN.BetweenCount function, previously, finding the start and end of this range required two separate binary searches, each operating over the entire length of the slice (one to find the lower bound, one for the upper).
it performs a single, combined binary search to simultaneously shrink the search area for both boundaries. Once this initial pass narrows the scope, it then launches two localized binary searches within these smaller, pre-qualified regions to precisely identify the final start and end indices of the target range.

I conducted a micro-benchmark to simulate high-frequency range queries. The test utilized a sorted slice of 10,000 integers (values 0, 2, ..., 19998) with a search range of [5000, 15000), effectively querying the middle 50% of the dataset (indices 2500 to 7500). The results demonstrate a clear improvement: the execution time dropped from 70.16 ns/op (Original Two-Pass) to 58.58 ns/op (Optimized One-Pass), yielding a performance gain of approximately 16.5%.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/util/slice/BUILD.bazel
+++ b/pkg/util/slice/BUILD.bazel
@@ -5,6 +5,9 @@ go_library(
     srcs = ["slice.go"],
     importpath = "github.com/pingcap/tidb/pkg/util/slice",
     visibility = ["//visibility:public"],
+    deps = [
+        "//pkg/util/intest",
+    ],
 )
 
 go_test(


pkg/util/slice/slice.go
```diff
@@ -49,3 +49,26 @@ func DeepClone[T interface{ Clone() T }](s []T) []T {
 	}
 	return cloned
 }
+
+// BinarySearchRange searches for target in the sorted slice x within the index range [left, right).
+// It returns the "lower bound" index idx (the smallest index i such that x[i] >= target)
+// and a boolean 'match' indicating if an exact match was found (x[idx] == target).
+// This avoids allocating a new sub-slice unlike slices.BinarySearchFunc on x[left:right].
+func BinarySearchRange[S ~[]E, E, T any](x S, target T, left, right int, cmp func(E, T) int) (int, bool) {
```
lance6716: we can use https://pkg.go.dev/slices#BinarySearch or https://pkg.go.dev/slices#BinarySearchFunc

pass the slice with `x[left:right]` to this function to avoid extra arguments

pkg/statistics/cmsketch.go
```diff
@@ -39,6 +39,7 @@ import (
 	"github.com/pingcap/tidb/pkg/util/dbterror"
 	"github.com/pingcap/tidb/pkg/util/hack"
 	"github.com/pingcap/tidb/pkg/util/intest"
+	"github.com/pingcap/tidb/pkg/util/slice"
```
lance6716: After the change of another comment, no need to import this package now

pkg/util/slice/slice.go
hawkingrei: it is the code from golang's ```slices```. The last comparison is unnecessary. You can remove it.
```
func BinarySearchFunc[S ~[]E, E, T any](x S, target T, cmp func(E, T) int) (int, bool) {
	n := len(x)
	// Define cmp(x[-1], target) < 0 and cmp(x[n], target) >= 0 .
	// Invariant: cmp(x[i - 1], target) < 0, cmp(x[j], target) >= 0.
	i, j := 0, n
	for i < j {
		h := int(uint(i+j) >> 1) // avoid overflow when computing h
		// i ≤ h < j
		if cmp(x[h], target) < 0 {
			i = h + 1 // preserves cmp(x[i - 1], target) < 0
		} else {
			j = h // preserves cmp(x[j], target) >= 0
		}
	}
	// i == j, cmp(x[i-1], target) < 0, and cmp(x[j], target) (= cmp(x[i], target)) >= 0  =>  answer is i.
	return i, i < n && cmp(x[i], target) == 0 //  <- it is unnecessary !!!
}
```

pkg/statistics/cmsketch.go
```diff
@@ -706,8 +710,8 @@ func (c *TopN) BetweenCount(sctx planctx.PlanContext, l, r []byte) (result uint6
 	if c == nil {
 		return 0
 	}
-	lIdx, _ := c.LowerBound(l)
-	rIdx, _ := c.LowerBound(r)
+	lIdx := c.LowerBoundRange(l, 0, len(c.TopN))
+	rIdx := c.LowerBoundRange(r, lIdx, len(c.TopN))
```
hawkingrei: If we find l < c.TopN[h] < r in the first loop, can we then use a divide-and-conquer approach to search for this index？

pkg/statistics/cmsketch.go
```diff
@@ -681,16 +682,15 @@ func (c *TopN) FindTopN(d []byte) int {
 	return idx
 }
 
-// LowerBound searches on the sorted top-n items,
+// GetRangeBound searches on the sorted top-n items,
 // returns the smallest index i such that the value at element i is not less than `d`.
```
lance6716: comments should be updated

pkg/statistics/cmsketch.go
```diff
@@ -681,16 +682,15 @@ func (c *TopN) FindTopN(d []byte) int {
...
-func (c *TopN) LowerBound(d []byte) (idx int, match bool) {
+func (c *TopN) GetRangeBound(r, l []byte) (ridx, ldx int) {
```
lance6716:
```suggestion
func (c *TopN) GetRangeBound(r, l []byte) (rIdx, lIdx int) {
```

pkg/store/mockstore/unistore/pd.go
```diff
@@ -441,7 +442,7 @@ func (m *mockKeyspaceManager) GetAllKeyspaces(ctx context.Context, startID uint3
 	m.mu.RLock()
 	defer m.mu.RUnlock()
 
-	startIndex, _ := slices.BinarySearchFunc(m.keyspaces, startID, func(k *keyspacepb.KeyspaceMeta, idToSearch uint32) int {
+	startIndex := slice.BinarySearchFunc(m.keyspaces, startID, func(k *keyspacepb.KeyspaceMeta, idToSearch uint32) int {
```
lance6716: why change it?

pkg/util/slice/slice.go
```diff
@@ -49,3 +49,56 @@ func DeepClone[T interface{ Clone() T }](s []T) []T {
 	}
 	return cloned
 }
+
+// BinarySearchRangeFunc performs a binary search on the slice x to find the range [t1, t2).
```
lance6716: what's the property of range [t1, t2)? Maybe `find the range [t1, t2) such that ...`

pkg/store/mockstore/unistore/pd.go
ghoshh: because the second returned value is not needed in the following operations

pkg/statistics/cmsketch.go
```diff
@@ -681,16 +682,15 @@ func (c *TopN) FindTopN(d []byte) int {
 	return idx
 }
 
-// LowerBound searches on the sorted top-n items,
-// returns the smallest index i such that the value at element i is not less than `d`.
-func (c *TopN) LowerBound(d []byte) (idx int, match bool) {
+// GetRangeBound searches on the sorted top-n items to find the index range [lIdx, rIdx)
```
lance6716: What does "corresponds to" mean? Do you mean "find longest index range [lIdx, rIdx) such that c.TopN[lIdx: rIdx] are all in the range of [l, r)"?

pkg/store/mockstore/unistore/pd.go
lance6716: I think it's better to use stardard library, because if in future it improves implementation for better performance we can automatic get it.

pkg/util/slice/slice.go
```diff
@@ -49,3 +49,57 @@ func DeepClone[T interface{ Clone() T }](s []T) []T {
 	}
 	return cloned
 }
+
+// BinarySearchRangeFunc finds the index range [i, j) in slice x that corresponds to the value range [t1, t2).
+// It returns 'i' as the smallest index where cmp(x[i], t1) >= 0, and 'j' as the smallest index where cmp(x[j], t2) >= 0.
```
lance6716, line 56: 1. t1 should be less than t2 as requirement?
2. I understand that this function just combined two binary search call into one call. So t1 vs t2 and i vs j and similar role, they are not different as one is inclusive and the other is exclusive 🤔

pkg/statistics/cmsketch.go
ghoshh: yes，I'll clarify it more clearly

pkg/util/slice/slice.go
ghoshh: 1.yes, t1 and t2 represent values and t1 must be less than t2.
2.t1 and t2 represent values while i and j represent indexes, I think they're different.

pkg/util/slice/slice.go
```diff
@@ -49,3 +49,57 @@ func DeepClone[T interface{ Clone() T }](s []T) []T {
...
+		if cmp(x[h], target) < 0 {
+			start = h + 1 // preserves cmp(x[i - 1], target) < 0
+		} else {
+			end = h // preserves cmp(x[j], target) >= 0
+		}
+	}
+	return start
+}
+
+// BinarySearchFunc performs a binary search on the slice x to find the target value.
```
lance6716: target value means equal value? please check all function comments, their meanings are not clear

pkg/util/slice/slice.go
```diff
@@ -49,3 +49,57 @@ func DeepClone[T interface{ Clone() T }](s []T) []T {
...
+		h := int(uint(i+j) >> 1) // avoid overflow when computing h
+		// i ≤ h < j
+		cmpt1 := cmp(x[h], t1)
+		if cmpt1 < 0 {
+			// x[h] < t1 < t2
+			i = h + 1 // preserves cmp(x[i - 1], target) < 0
+			continue
+		} else if cmpt1 == 0 {
+			// x[h] = t1 < t2
+			return h, BinarySearchByIndexFunc(x, t2, h, j, cmp)
```
lance6716: Please add comments that if this function can accepts duplicated elements.

pkg/statistics/cmsketch.go
```diff
@@ -681,16 +682,15 @@ func (c *TopN) FindTopN(d []byte) int {
 	return idx
 }
 
-// LowerBound searches on the sorted top-n items,
-// returns the smallest index i such that the value at element i is not less than `d`.
-func (c *TopN) LowerBound(d []byte) (idx int, match bool) {
+// GetRangeBound searches the sorted c.TopN items to find the longest index range [lIdx, rIdx)
+// such that all items c.TopN[lIdx:rIdx] fall within the value range [l, r) (l is inclusive, r is exclusive).
```
lance6716, line 686: I think this expression is wrong (I misunderstood it yesterday). The description of `BinarySearchRangeFunc` is correct

pkg/util/slice/slice.go
```diff
@@ -49,3 +49,59 @@ func DeepClone[T interface{ Clone() T }](s []T) []T {
...
+			j = h // preserves cmp(x[j], target) >= 0
+		} else if cmpt2 == 0 {
+			// t1 < x[h] = t2
+			return BinarySearchByIndexFunc(x, t1, i, h, cmp), h
+		}
+	}
+	return i, j
+}
+
+// BinarySearchByIndexFunc searches the sorted sub-slice x[start:end).Duplicates are allowed.
```
lance6716:
```suggestion
// BinarySearchByIndexFunc searches the sorted sub-slice x[start:end). Duplicates are allowed.
```

pkg/util/slice/slice.go
lance6716: this function. Not `BinarySearchByIndexFunc`

pkg/util/slice/slice.go
```diff
@@ -49,3 +49,59 @@ func DeepClone[T interface{ Clone() T }](s []T) []T {
...
+// It returns the smallest index i in [start, end] such that cmp(x[i], target) >= 0.
+func BinarySearchByIndexFunc[S ~[]E, E, T any](x S, target T, start, end int, cmp func(E, T) int) int {
```
lance6716: reuse std library

pkg/util/slice/slice.go
```diff
@@ -49,3 +49,59 @@ func DeepClone[T interface{ Clone() T }](s []T) []T {
...
+		} else if cmpt2 == 0 {
+			// t1 < x[h] = t2
+			return BinarySearchByIndexFunc(x, t1, i, h, cmp), h
+		}
+	}
+	return i, j
+}
+
+// BinarySearchByIndexFunc searches the sorted sub-slice x[start:end).Duplicates are allowed.
+// It returns the smallest index i in [start, end] such that cmp(x[i], target) >= 0.
```
lance6716, line 90: what if no such value? `x[i]` will out of bound?

ghoshh: it doesn't matter, it will return [i, i) and it won't go into the following loops

lance6716: Developers will not check source codes, they will only see function comment "It returns the smallest index i in [start, end] such that cmp(x[i], target) >= 0.". if `i` is `end`, there's no `x[i]`. Please improve your description.

pkg/util/slice/slice_test.go
```diff
@@ -39,3 +41,47 @@ func TestSlice(t *testing.T) {
...
+	checkBinarySeachByIndx(t, arr, 10, 13)   // [arr[5] < low < arr[6], arr[7]]
+	checkBinarySeachByIndx(t, arr, 11, 12)   // [arr[5], arr[6] < high < arr[7]]
+	checkBinarySeachByIndx(t, arr, 11, 13)   // [arr[5], arr[7]]
+	checkBinarySeachByIndx(t, arr, 11, 14)   // [arr[5], arr[7] < high < arr[8]]
+	checkBinarySeachByIndx(t, arr, 13, 15)   // [arr[len-2], arr[len-1]]
+	checkBinarySeachByIndx(t, arr, 10, 16)   // [arr[5] < low < arr[6], after last]
+	checkBinarySeachByIndx(t, arr, 10, 15)   // [arr[5] < low < arr[6], last]
+	checkBinarySeachByIndx(t, arr, 10, 14)   // [arr[5] < low < arr[6], arr[len-2] < high < arr[len-1]]
+	checkBinarySeachByIndx(t, arr, 130, 150) // after last
+	arr = []int{-1, 1}
```
lance6716, line 75: maybe check more lengths of duplicated value

pkg/util/slice/slice_test.go
```diff
@@ -39,3 +41,47 @@ func TestSlice(t *testing.T) {
...
+	checkBinarySeachByIndx(t, arr, -2, -2)
+	checkBinarySeachByIndx(t, arr, 2, 3)
+	checkBinarySeachByIndx(t, arr, 0, 1)
+	checkBinarySeachByIndx(t, arr, -1, 0)
+	checkBinarySeachByIndx(t, arr, -1, -1)
+	checkBinarySeachByIndx(t, arr, 1, 1)
+	arr = []int{0}
```
lance6716, line 82: check zero-length slices if function permits it, or add comments about the limitation

lance6716 (commented):
rest lgtm

pkg/util/slice/slice.go
```diff
@@ -49,3 +49,44 @@ func DeepClone[T interface{ Clone() T }](s []T) []T {
...
+		}
+		return BinarySearchByIndexFunc(x, t1, i, iEnd, cmp), BinarySearchByIndexFunc(x, t2, h+1, j, cmp)
+	}
+	// Not found, return [i, i)
+	return i, j
+}
+
+// BinarySearchByIndexFunc searches the sorted sub-slice x[start:end).
+// It returns the smallest index i in [start, end] such that cmp(x[i], target) >= 0.
+// // If no such index exists (i.e., all elements in x[start:end) are less than target), it returns end.
```
lance6716:
```suggestion
// It returns the smallest index i in [start, end) such that cmp(x[i], target) >= 0.
// If no such index exists (i.e., all elements in x[start:end) are less than target), it returns end.
```

lance6716:
```
Please run `make bazel_prepare` to update `.bazel` files, or just apply the following git diff (run `git apply -` and paste following contents with ctrl-D as ending):
diff --git a/pkg/store/mockstore/unistore/BUILD.bazel b/pkg/store/mockstore/unistore/BUILD.bazel
index c1ae5aabfc..23c49cd650 100644
--- a/pkg/store/mockstore/unistore/BUILD.bazel
+++ b/pkg/store/mockstore/unistore/BUILD.bazel
@@ -21,7 +21,6 @@ go_library(
         "//pkg/store/mockstore/unistore/tikv",
         "//pkg/tablecodec",
         "//pkg/util/codec",
-        "//pkg/util/slice",
         "//pkg/util/topsql/state",
         "@com_github_asaskevich_govalidator//:govalidator",
         "@com_github_golang_protobuf//proto",

```

@ghoshh

pkg/util/slice/slice.go
0xPoe:
```suggestion
// If no such index exists (i.e., all elements in x[start:end) are less than target), it returns end.
```

pkg/util/slice/slice_test.go
```diff
@@ -39,3 +41,64 @@ func TestSlice(t *testing.T) {
 		})
 	}
 }
+
+func checkBinarySeachByIndx(t *testing.T, arr []int, low, high int) {
```
0xPoe:
```suggestion
func checkBinarySearchByIndx(t *testing.T, arr []int, low, high int) {
```

pkg/util/slice/slice.go
```diff
@@ -49,3 +49,44 @@ func DeepClone[T interface{ Clone() T }](s []T) []T {
...
+func BinarySearchByIndexFunc[S ~[]E, E, T any](x S, target T, start, end int, cmp func(E, T) int) int {
```
0xPoe:
```suggestion
func binarySearchByIndexFunc[S ~[]E, E, T any](x S, target T, start, end int, cmp func(E, T) int) int {
```

No need to export it.

pkg/util/slice/slice.go
```diff
@@ -49,3 +49,44 @@ func DeepClone[T interface{ Clone() T }](s []T) []T {
 	}
 	return cloned
 }
+
+// BinarySearchRangeFunc finds the index range [i, j) in slice x that corresponds to the value range [t1, t2).
+// It returns 'i' as the smallest index where cmp(x[i], t1) >= 0, and 'j' as the smallest index where cmp(x[j], t2) >= 0.
+func BinarySearchRangeFunc[S ~[]E, E, T any](x S, t1, t2 T, cmp func(E, T) int) (i, j int) {
```
0xPoe, line 58: We could use `intest.Assert` to assert some basic assumptions here to avoid accidentally breaking them.

pkg/util/slice/slice.go
```diff
@@ -49,3 +49,44 @@ func DeepClone[T interface{ Clone() T }](s []T) []T {
...
+		}
+		return BinarySearchByIndexFunc(x, t1, i, iEnd, cmp), BinarySearchByIndexFunc(x, t2, h+1, j, cmp)
+	}
+	// Not found, return [i, i)
+	return i, j
+}
+
+// BinarySearchByIndexFunc searches the sorted sub-slice x[start:end).
+// It returns the smallest index i in [start, end] such that cmp(x[i], target) >= 0.
+// If no such index exists (i.e., all elements in x[start:end) are less than target), it returns end.
```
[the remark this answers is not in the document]
ghoshh: this comment is wrong.

pkg/util/slice/slice.go
```diff
@@ -49,3 +51,47 @@ func DeepClone[T interface{ Clone() T }](s []T) []T {
...
+	// No elements in range, return [i, j) where i == j (empty range).
+	return i, j
+}
+
+// binarySearchByIndexFunc searches the sorted sub-slice x[start:end).
+// It returns the smallest index i in [start, end] such that cmp(x[i], target) >= 0.
+// If no such index exists (i.e., all elements in x[start:end) are less than target), it returns end.
+func binarySearchByIndexFunc[S ~[]E, E, T any](x S, target T, start, end int, cmp func(E, T) int) int {
+	intest.Assert(start >= 0 && start <= end && end <= len(x),
+		"invalid range: start=%d end=%d len=%d", start, end, len(x))
```
winoros, line 94:
```suggestion
		"invalid range: start=%d end=%d len=%d", start, end, len(x),
	)
```

PR #64096, opened 2025-10-22
ddl: add timeout mechanism for add index analyze

GMHDBJD:

Issue Number: ref https://github.com/pingcap/tidb/issues/63326

Problem Summary:

- add AnalyzeStateTimeout
- set timeout to max(2*addIndexTime,1min)
- show analyzed timeout in admin show ddl jobs
- exaclty once analyze after create index
### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/index.go
+++ b/pkg/ddl/index.go
@@ -1302,12 +1302,19 @@ func (w *worker) queryAnalyzeStatusSince(startTS uint64, dbName, tblName string)
 	}
 	defer w.sessPool.Put(sessCtx)
 
-	exec := sessCtx.GetRestrictedSQLExecutor()
 	startTimeStr := time.Now().UTC().Format(time.DateTime)
 	if startTS > 0 {
 		startTimeStr = model.TSConvert2Time(startTS).UTC().Format(time.DateTime)
 	}
 	kctx := kv.WithInternalSourceType(w.ctx, kv.InternalTxnStats)
+
+	// set session time zone to UTC to match the time format in `startTimeStr`
+	originalTimeZone := sessCtx.GetSessionVars().TimeZone
+	sessCtx.GetSessionVars().TimeZone = time.UTC
+	defer func() {
+		sessCtx.GetSessionVars().TimeZone = originalTimeZone
+	}()
+	exec := sessCtx.GetRestrictedSQLExecutor()
 	rows, _, chkErr := exec.ExecRestrictedSQL(kctx, []sqlexec.OptionFuncAlias{sqlexec.ExecOptionUseCurSession},
 		"SELECT state FROM mysql.analyze_jobs WHERE table_schema = %? AND table_name = %? AND start_time >= %?", dbName, tblName, startTimeStr)
 	if chkErr != nil {


pkg/statistics/handle/util/util.go
```diff
@@ -100,80 +100,126 @@ func CallWithSCtx(pool syssession.Pool, f func(sctx sessionctx.Context) error, f
 }
 
 // UpdateSCtxVarsForStats updates all necessary variables that may affect the behavior of statistics.
-func UpdateSCtxVarsForStats(sctx sessionctx.Context) error {
+// UpdateSCtxVarsForStats updates all necessary variables that may affect the behavior of statistics.
```
tangenta, line 102:
```suggestion
```

pkg/statistics/handle/util/util.go
```diff
@@ -100,80 +100,126 @@ func CallWithSCtx(pool syssession.Pool, f func(sctx sessionctx.Context) error, f
...
 	}
-	sctx.GetSessionVars().AnalyzePartitionMergeConcurrency = int(ver)
-	return nil
+	sv.AnalyzePartitionMergeConcurrency = int(ver)
+
+	if sv.TimeZone == nil {
+		sv.TimeZone = time.UTC
+	}
+
+	return restore, nil
```
AilinKid: restore mechanism looks good to me, since we don't want to pollute the ddl session default variables, while stats module current require stats-approvers @0xPoe

pkg/ddl/index.go
```diff
@@ -1271,13 +1281,154 @@ func initForReorgIndexes(w *worker, job *model.Job, idxInfos []*model.IndexInfo)
...
+	}
+
+	cumulativeTimeout, found := w.ddlCtx.getAnalyzeCumulativeTimeout(job.ID)
+	if !found {
+		cumulativeTimeout = DefaultCumulativeTimeout
+		if job.RealStartTS != 0 {
+			addStart := model.TSConvert2Time(job.RealStartTS)
+			elapsed := time.Since(addStart)
+			if elapsed*2 > cumulativeTimeout {
+				cumulativeTimeout = elapsed * 2
```
AilinKid, line 1408: need a bias? current double time is possible, maybe 2.5 times ?

GMHDBJD: Judging from test result, the analyze time should be significantly lower than the add index time, so I think twice should be enough.

pkg/ddl/ddl.go
```diff
@@ -504,6 +504,76 @@ func (dc *ddlCtx) getAnalyzeDoneCh(jobID int64) chan struct{} {
...
+	dc.jobCtx.RLock()
+	defer dc.jobCtx.RUnlock()
+	ctx, exists := dc.jobCtx.jobCtxMap[jobID]
+	if !exists {
+		return time.Time{}, false
+	}
+	if ctx.analyzeStartTime.IsZero() {
+		return time.Time{}, false
+	}
+	return ctx.analyzeStartTime, true
```
tangenta: Isn't the default value of `time.Time` zero?

pkg/statistics/handle/util/util.go
```diff
@@ -100,80 +100,125 @@ func CallWithSCtx(pool syssession.Pool, f func(sctx sessionctx.Context) error, f
...
 	ver, err = strconv.ParseInt(verInString, 10, 64)
 	if err != nil {
-		return err
+		return restore, err
+	}
+	sv.AnalyzePartitionMergeConcurrency = int(ver)
+
+	if sv.TimeZone == nil {
+		sv.TimeZone = time.UTC
 	}
```
tangenta: The main change except restoring is setting the timezone to UTC, could you please check if it affects the analyze behavior? @0xPoe @AilinKid

tests/realtikvtest/addindextest4/integration_test.go
```diff
@@ -225,6 +227,88 @@ func TestAddIndexShowAnalyzeProgress(t *testing.T) {
 	require.True(t, analyzed)
 }
 
+func TestAnalyzeTimeout(t *testing.T) {
```
tangenta, line 230: How long does this test take to run?

pkg/ddl/index.go
```diff
@@ -1271,14 +1281,155 @@ func initForReorgIndexes(w *worker, job *model.Job, idxInfos []*model.IndexInfo)
...
+			}
+		}
+		w.ddlCtx.setAnalyzeCumulativeTimeout(job.ID, cumulativeTimeout)
 	}
+
 	if doneCh == nil {
+		if _, ok := w.ddlCtx.getAnalyzeStartTime(job.ID); !ok {
+			w.ddlCtx.setAnalyzeStartTime(job.ID, time.Now())
+		}
```
tangenta: Move this after `queryAnalyzeStatusSince`?

pkg/ddl/index.go
```diff
@@ -1271,14 +1281,155 @@ func initForReorgIndexes(w *worker, job *model.Job, idxInfos []*model.IndexInfo)
...
+		status, err := w.queryAnalyzeStatusSince(job.StartTS, dbName, tblName)
+		if err != nil {
+			logutil.DDLLogger().Warn("query analyze status failed", zap.Int64("jobID", job.ID), zap.Error(err))
+			status = analyzeUnknown
+		}
+
+		done, timeOut, proceed := w.analyzeStatusDecision(job, dbName, tblName, status, cumulativeTimeout)
```
tangenta: How about the failure state?

pkg/ddl/modify_column.go
```diff
@@ -1266,12 +1271,17 @@ func (w *worker) doModifyColumnIndexReorg(
 			}
 		case model.AnalyzeStateRunning:
 			// after all old index data are reorged. re-analyze it.
-			done := w.analyzeTableAfterCreateIndex(job, job.SchemaName, tblInfo.Name.L)
-			if done {
-				job.ReorgMeta.AnalyzeState = model.AnalyzeStateDone
+			done, timeOut := w.analyzeTableAfterCreateIndex(job, job.SchemaName, tblInfo.Name.L)
```
tangenta:
```suggestion
			done, timedOut := w.analyzeTableAfterCreateIndex(job, job.SchemaName, tblInfo.Name.L)
```

Make the naming consistent

tests/realtikvtest/addindextest4/integration_test.go
GMHDBJD: 32s

pkg/ddl/index.go
```diff
@@ -1271,14 +1281,156 @@ func initForReorgIndexes(w *worker, job *model.Job, idxInfos []*model.IndexInfo)
...
+	}
+	defer restore()
+
+	exec := sessCtx.GetRestrictedSQLExecutor()
+	startTimeStr := time.Now().Format(time.DateTime)
+	if startTS > 0 {
+		startTimeStr = model.TSConvert2Time(startTS).Format(time.DateTime)
+	}
+	kctx := kv.WithInternalSourceType(w.ctx, kv.InternalTxnStats)
+	rows, _, chkErr := exec.ExecRestrictedSQL(kctx, []sqlexec.OptionFuncAlias{sqlexec.ExecOptionUseCurSession}, "SHOW ANALYZE STATUS WHERE table_schema = %? AND table_name = %? AND start_time >= %?", dbName, tblName, startTimeStr)
```
0xPoe: Because when we store the start_time, we use SET start_time = CONVERT_TZ(..., '+00:00', @@TIME_ZONE), so I think we can directly query it using UTC time and read it directly from mysql.analyze_jobs. This way, we don’t need to consider time zone conversion anymore.

pkg/ddl/index.go
```diff
@@ -1271,14 +1281,156 @@ func initForReorgIndexes(w *worker, job *model.Job, idxInfos []*model.IndexInfo)
...
+// starting ANALYZE.
+func (w *worker) queryAnalyzeStatusSince(startTS uint64, dbName, tblName string) (analyzeStatus, error) {
+	sessCtx, err := w.sessPool.Get()
+	if err != nil {
+		return analyzeUnknown, err
+	}
+	defer w.sessPool.Put(sessCtx)
+
+	// internal sql may not init the analysis related variable correctly.
+	restore, err := statsutil.UpdateSCtxVarsForStats(sessCtx)
```
0xPoe: Even if we want to handle the timezone issue here, I don’t think it’s reasonable to put it in this utility function, since it’s only meant for analyze-related variables. This seems to be just a special case specific to this function.

pkg/ddl/index.go
0xPoe: I’m not sure if we really have a timezone issue here. Given that all these processes are internal details, shouldn’t we just use the server’s timezone by default? Do we actually need to consider the timezone issue here? Do you mean we’ll use the timezone from the “add index” session for all processes?

AilinKid: LGTM

AilinKid (approved):

pkg/statistics/handle/util/util.go
```diff
@@ -81,7 +81,7 @@ func CallWithSCtx(pool syssession.Pool, f func(sctx sessionctx.Context) error, f
 	defer util.Recover(metrics.LabelStats, "CallWithSCtx", nil, false)
 	return pool.WithSession(func(se *syssession.Session) error {
 		return se.WithSessionContext(func(sctx sessionctx.Context) error {
-			if err := UpdateSCtxVarsForStats(sctx); err != nil { // update stats variables automatically
+			if _, err := UpdateSCtxVarsForStats(sctx); err != nil { // update stats variables automatically
```
0xPoe: It seems we didn't handle the restore here.

pkg/statistics/handle/util/util.go
```diff
@@ -100,80 +100,118 @@ func CallWithSCtx(pool syssession.Pool, f func(sctx sessionctx.Context) error, f
 }
 
 // UpdateSCtxVarsForStats updates all necessary variables that may affect the behavior of statistics.
-func UpdateSCtxVarsForStats(sctx sessionctx.Context) error {
+// It returns a restore function that can be invoked to restore the original session variables.
+func UpdateSCtxVarsForStats(sctx sessionctx.Context) (func(), error) {
```
0xPoe: I think we’d better revert this change because it’s not related to this feature.
It’s a good idea to roll back these updates and return a clean session to the session pool.
However, the current API design feels a bit unsafe to me — when we return an error, the caller still needs to manually restore it, which is easy to forget.

So let’s revert this change for now, and I’ll create an enhancement to address this issue.

0xPoe (commented):
Thanks!

pkg/ddl/index.go
```diff
@@ -1271,14 +1281,157 @@ func initForReorgIndexes(w *worker, job *model.Job, idxInfos []*model.IndexInfo)
...
+	if err != nil {
+		return analyzeUnknown, err
+	}
+	defer restore()
+
+	exec := sessCtx.GetRestrictedSQLExecutor()
+	startTimeStr := time.Now().Format(time.DateTime)
```
0xPoe: It would be better to add some comments explaining why using the local timezone works in this case.

GMHDBJD: change to utc time

0xPoe (approved):
Thanks! :shipit: 

Some offline discussion:

```console
Dongpo: I have a few small questions, so I'll just ask them here directly. Otherwise, I'd have to go back and forth on GitHub. 
If we time out now, we won't kill the analyze process, right? We'll just wait until it finishes (I think it's indeed reasonable to let it finish). In this case, should we give users a warning, or is logging sufficient? 
Why can't we directly use the start time in analyze jobs and then directly determine if it has timed out? Why do we still need to record it in ctx itself? I feel like directly calculating the timeout in that function each time and then using a single query to check if it has timed out would suffice, right? (Of course, this is more of a nice-to-have. Simply recording time.Now doesn't cause any major issues either. It's just that in this way, both values probably don't need to be directly stored and cleared in ctx.)
Additionally, is it possible that this location could be accidentally affected by other analyzes, such as those from auto-analysis or manual analysis by users? (When you check, you might not be looking at the record of the analyze you executed. Of course, this is also a consideration for completeness, and it should not be a major issue if we make changes later.) 

GMHDBJD: At that time, the discussion was about whether the first phase would time out by using admin show ddl job, and now it is indeed the case, where timeout will be displayed 
I don't quite understand what you mean. This timeout should be the time for ddl to wait for analyze, for example, checking once every ten seconds. If analyze is completed, it returns success; otherwise, it continues to check after 10 seconds until timeout. 
As long as the analyze is completed after this ddl, it doesn't matter who does it, so it only needs to be after startts 

Dongpo:
What I mean is, since analyze_jobs already has the exact start time of the analyze task, why do we still need to record one ourselves? Can't we just use this directly? So, is it okay to simply query records greater than startts, compare its start-time with the current time, and check if it has exceeded the timeout?
If all tasks are successful, there's really no problem, as long as it's not disrupted by failed tasks. It should be fine. It's just a corner case. 
Whether to change this or not is optional; it's just a matter of implementation. 
```

PR #64097, opened 2025-10-22
statistics: ignore system tables in stats cache

### LINKED ISSUE #57176: The statistics metric cannot correctly display the stats health
## Bug Report

1. Create a tidb cluster
2. Create 1000 tables
3. Check the grafana

The stats health is correct.

image

```sql
| test    | sbtest604  |                |      89 |
| test    | sbtest875  |                |      90 |
| test    | sbtest419  |                |      89 |
| test    | sbtest746  |                |      89 |
| test    | sbtest896  |                |      89 |
+---------+------------+----------------+---------+
1000 rows in set (0.01 sec)
```

Introduce by https://github.com/pingcap/tidb/pull/56475

```diff
// UpdateStatsHealthyMetrics updates stats healthy distribution metrics according to stats cache.
func (s *StatsCacheImpl) UpdateStatsHealthyMetrics() {
	distribution := make([]int64, 9)
	uneligibleAnalyze := 0
	for _, tbl := range s.Values() {
-		distribution[4]++ // total table count
+ 		distribution[7]++ // total table count
		isEligibleForAnalysis := tbl.IsEligibleForAnalysis()
		if !isEligibleForAnalysis {
			uneligibleAnalyze++
			continue
		}
		healthy, ok := tbl.GetStatsHealthy()
		if !ok {
			continue
		}
		if healthy < 50 {
			distribution[0]++
		} else if healthy < 55 {
			distribution[1]++
		} else if healthy < 60 {
			distribution[2]++
		} else if healthy < 70 {
			distribution[3]++
		} else if healthy < 80 {
			distribution[4]++
		} else if healthy < 100 {
			distribution[5]++
		} else {
			distribution[6]++
		}
	}
	for i, val := range distribution {
		handle_metrics.StatsHealthyGauges[i].Set(float64(val))
	}
	handle_metrics.StatsHealthyGauges[8].Set(float64(uneligibleAnalyze))
}
```

master

### LINKED ISSUE #64080: Stats Healthy metric cannot display the stats health correctly
## Bug Report
Introduced by https://github.com/pingcap/tidb/pull/54822

1. Create a table
```sql
create table h (idx int, code int, typ1 int, typ2 int, update_time int, key k1(idx, typ1, typ2), key k2(idx, update_time));
```
2. Insert 500 rows
```sql
insert into h select * from (
  with recursive tt as (
    select 0 idx, 0 as code, 0 as typ1, 0 as typ2, 0 as update_time
    union all
    select mod(update_time, 100) as idx, 0 as code, 0 as typ1, 0 as typ2, update_time+1 as update_time from tt where update_time<500
  ) select * from tt
) tt;
```
3. Analyze the table:
```sql
analyze table h;
```

This is a table shown in the Stats Healthy metric within the [100, 100] range.

Everything marked unneeded.

master

0xPoe:

Issue Number: close https://github.com/pingcap/tidb/issues/64080, close https://github.com/pingcap/tidb/issues/57176

Problem Summary:

1. Avoid putting system tables into the stats cache, as it would pollute the cache.
2. Do not always place small tables in the “unneeded” range; otherwise, it may confuse users into thinking the table hasn’t been analyzed when it actually has.
3. Added a pseudo label to separate it from the unneeded label.

After these changes: total tables = pseudo tables + unneeded analyze tables + tables in healthy buckets.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [x] [Manual test](https://github.com/pingcap/tidb/pull/64097#issuecomment-3442163105) 
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
修复系统表会污染表健康度监控的问题
Fix the issue where system tables could pollute the stats health metrics
```


0xPoe (commented):

- [x] Unit tests added

- [ ] All tests pass

- [x] Bazel files updated

- [x] PR size is reasonable


/cc @time-and-fate @winoros

0xPoe:
Test locally:
1. Start the cluster with patch: `tiup playground v8.5.3 --db.binpath /Users/poe/code/tidb/bin/tidb-server`
2. Use sysbench to create 1k tables:
```console
tidb on  poe-patch-ignore-sys-tables [$] via 🐳 orbstack via 🐹 v1.23.12
😺 ❯ sysbench /opt/homebrew/share/sysbench/oltp_read_write.lua \
  --db-driver=mysql \
  --mysql-host=127.0.0.1 \
  --mysql-port=4000 \
  --mysql-user=root \
  --mysql-db=sbtest \
  --tables=1000 \
  --table-size=2000 \
  --threads=64 \
  prepare
``` 
3. Run the workload:
```sql
tidb on  poe-patch-ignore-sys-tables [$] via 🐳 orbstack via 🐹 v1.23.12
😿 ❯ sysbench /opt/homebrew/share/sysbench/oltp_read_write.lua \
  --db-driver=mysql \
  --mysql-host=127.0.0.1 \
  --mysql-port=4000 \
  --mysql-user=root \
  --mysql-db=sbtest \
  --tables=1000 \
  --table-size=2000 \
  --threads=6 \
    --time=0 \
∙ run
```
4. Check the metrics:
image

0xPoe:
/cc @D3Hunter 
I am not sure how to verify it on next-gen. Is it possible to test it locally?

0xPoe:
/hold 

Need to discuss more.

pkg/statistics/handle/metrics/metrics.go
```diff
@@ -41,8 +41,9 @@ func InitMetricsVars() {
 		metrics.StatsHealthyGauge.WithLabelValues("[70,80)"),          // 4
 		metrics.StatsHealthyGauge.WithLabelValues("[80,100)"),         // 5
 		metrics.StatsHealthyGauge.WithLabelValues("[100,100]"),        // 6
-		metrics.StatsHealthyGauge.WithLabelValues("[0,100]"),          // 7
+		metrics.StatsHealthyGauge.WithLabelValues("total"),            // 7
```
0xPoe: I will test to see what happens if the cluster upgrades from a lower version.

0xPoe: Image
This is a break change.

0xPoe: It ends up looking like this:
image

0xPoe: /cc @time-and-fate @winoros 
I’m not sure whether we should keep using the [0,100] range. It might be better to keep it as is to avoid this confusing label change.

0xPoe: I just reverted it to maintain compatibility.

0xPoe:
Tested locally again:
1. Start the cluster without the patch: `tiup playground v8.5.3 --tag test-metric1`
2. Create tables and insert data:
```console
tidb on  poe-patch-ignore-sys-tables [$] via 🐳 orbstack via 🐹 v1.23.12 took 2s
😺 ❯ sysbench /opt/homebrew/share/sysbench/oltp_read_write.lua \
  --db-driver=mysql \
  --mysql-host=127.0.0.1 \
  --mysql-port=4000 \
  --mysql-user=root \
  --mysql-db=sbtest \
  --tables=1000 \
  --table-size=2000 \
  --threads=10 \
  prepare
```
3. Run the workload:
```console
tidb on  poe-patch-ignore-sys-tables [$] via 🐳 orbstack via 🐹 v1.23.12 took 12m11s
😿 ❯ sysbench /opt/homebrew/share/sysbench/oltp_read_write.lua \
  --db-driver=mysql \
  --mysql-host=127.0.0.1 \
  --mysql-port=4000 \
  --mysql-user=root \
  --mysql-db=sbtest \
  --tables=1000 \
  --table-size=2000 \
  --threads=64 \
  --time=0 run
```
4. Check the metrics:
image
5. Start the cluster with patch: ` tiup playground v8.5.3 --db.binpath /Users/poe/code/tidb/bin/tidb-server --tag test-metric1`
6. Start the workload again
7. Check the metrics again:
image

0xPoe:
Tested on next-gen:
1. Start the minio with Docker Compose: (if the bucket creation failed, create it manually)
```yml
version: "3.9"

services:
  minio:
    image: minio/minio:latest
    restart: unless-stopped
    environment:
      MINIO_ROOT_USER: minioadmin
      MINIO_ROOT_PASSWORD: minioadmin
      MINIO_REGION: local
      MINIO_SERVER_URL: http://127.0.0.1:9000
    command: server /data --console-address ":9001"
    ports:
      - "9000:9000"
      - "9001:9001"
    volumes:
      - minio_data:/data
  minio-init:
    image: minio/mc:latest
    depends_on:
      - minio
    entrypoint: >
      /bin/sh -c "
      until mc alias set minio http://minio:9000 minioadmin minioadmin; do
        echo 'Waiting for MinIO...'
        sleep 2
      done
      mc mb --ignore-existing minio/cse-test
      "
    restart: "no"
volumes:
  minio_data:
    driver: local
```
2. Compile the TiDB, PD, and TiKV
3. Start the playground
tidb:
```yaml
keyspace-name = "SYSTEM"
tikv-worker-url = "http://localhost:20001"
[instance]
tidb_service_scope = 'dxf_service'
```
tikv:
```yaml
[storage]
reserve-space = 0
api-version = 2
enable-ttl = true

[dfs]
prefix = "tikv"
s3-endpoint = "127.0.0.1:9000"
s3-key-id = "minioadmin"
s3-secret-key = "minioadmin"
s3-bucket = "cse-test"
s3-region = "local"
```
tikv-worker:
```yaml
[dfs]
prefix = "tikv"
s3-endpoint = "127.0.0.1:9000"
s3-key-id = "minioadmin"
s3-secret-key = "minioadmin"
s3-bucket = "cse-test"
s3-region = "local"

# ia should be configured when enabled.
[ia]
mem-cap = "5GB"  # or "10%"
disk-cap = "20GB"  # or "20%"
```
pd:
```yaml
[keyspace]
pre-alloc = ["SYSTEM","keyspace1", "keyspace2"]
```
```console
tiup playground:nightly v8.5.0 --tag serverless4 \
  --db.binpath /Users/poe/code/tidb/bin/tidb-server \
  --db.config ./tidb.toml \
  --kv.binpath /Users/poe/code/cloud-storage-engine/target/debug/tikv-server \
  --kv.config ./tikv.toml \
  --pd.binpath /Users/poe/code/pd/bin/pd-server \
  --pd.config ./pd.toml \
  --tiflash 0 --db 3 --kv 3
```
4. Start the tikv worker:
```console
/Users/poe/code/cloud-storage-engine/target/debug/tikv-worker --addr 127.0.0.1:20001 --pd-endpoints http://localhost:2379 --config ./tikv-worker.toml --data-dir ~/.tiup/data/tikv-worker
```
5. Create tables and insert data:
```console
sysbench /opt/homebrew/share/sysbench/oltp_read_write.lua \
  --db-driver=mysql \
  --mysql-host=127.0.0.1 \
  --mysql-port=4000 \
  --mysql-user=root \
  --mysql-db=sbtest \
  --tables=100 \
  --table-size=2000 \
  --threads=10 \
  prepare
```
6. Check metrics
image

pkg/session/syssession/session_integration_test.go
```diff
@@ -197,7 +197,6 @@ func TestDomainAdvancedSessionPoolPutBackDirtySession(t *testing.T) {
 			require.NotNil(t, se)
 			require.True(t, se.IsInternalClosed())
 			require.False(t, se.IsOwner())
-			require.Zero(t, p.(*syssession.AdvancedSessionPool).Size())
```
0xPoe, line 200: It’s hard to check how many sessions remain because the session pool may also be used by background processes.

pkg/statistics/handle/handle.go
```diff
@@ -200,17 +202,72 @@ func (h *Handle) getStatsByPhysicalID(physicalTableID int64, tblInfo *model.Tabl
...
+	tbl.PhysicalID = physicalTableID
+
+	// TODO: Determine whether we really need to cache pseudo table stats for non-partitioned tables.
+	// If the memory overhead is manageable, we can remove this optimization.
+	shouldCachePseudo := tblInfo.GetPartitionInfo() == nil || h.Len() < 64
+	if !shouldCachePseudo {
+		return tbl, true
+	}
+
+	// HACK: Sessions borrowed from the pool cannot fetch schema metadata for local temporary tables,
```
time-and-fate: I think it's not a hack. I think this is the proper behavior, at least for now.

pkg/statistics/table.go
```diff
@@ -792,6 +792,11 @@ func (t *Table) IsEligibleForAnalysis() bool {
 	return true
 }
 
+// MeetAutoAnalyzeMinCnt checks whether the table meets the minimum count required for auto-analyze.
+func (t *Table) MeetAutoAnalyzeMinCnt() bool {
+	return t != nil && t.RealtimeCount >= AutoAnalyzeMinCnt
+}
```
time-and-fate, line 798: I think we can modify `IsEligibleForAnalysis()` to use `!MeetAutoAnalyzeMinCnt() || t.Pseudo` to reduce repetition.

pkg/statistics/handle/cache/statscache.go
```diff
@@ -357,39 +358,62 @@ func (s *StatsCacheImpl) SetStatsCacheCapacity(c int64) {
...
+	statsHealthyBucket70To80
+	statsHealthyBucket80To100
+	statsHealthyBucket100To100
+	statsHealthyBucketTotal
+	statsHealthyBucketUnneededAnalyze
+	statsHealthyBucketPseudo
+	statsHealthyBucketCount
+)
+
+var statsHealthyBucketUpperBounds = [...]int64{50, 55, 60, 70, 80, 100}
```
time-and-fate: I think we can simplify some definitions and reduce repetition here and `StatsHealthyGauges` in `metrics.go`.
We can have a unified definition in `metrics.go` For example, use a structure like this
```go
type HealthyBucketConfig struct {
    Index      int    // set to values like statsHealthyBucket0To50
    UpperBound int64
    Label      string // used by `WithLabelValues()`
}

var HealthyBucketConfigs = []HealthyBucketConfig{...}
```

pkg/session/bootstrap_test.go
```diff
@@ -2109,7 +2109,10 @@ func TestTiDBUpgradeToVer176(t *testing.T) {
 	ver, err = getBootstrapVersion(seV175)
 	require.NoError(t, err)
 	require.Less(t, int64(ver175), ver)
-	MustExec(t, seV175, "SELECT * from mysql.tidb_global_task_history")
+	// Avoid reusing the old session when checking the new table.
+	// Otherwise it may access the previous domain, which has already been closed.
```
D3Hunter, line 2113: why it works previously?

pkg/session/syssession/session_integration_test.go
```diff
@@ -74,6 +75,8 @@ func TestDomainAdvancedSessionPoolInternalSessionRegistry(t *testing.T) {
 }
 
 func TestDomainAdvancedSessionPoolPutBackDirtySession(t *testing.T) {
+	require.NoError(t, failpoint.Enable("github.com/pingcap/tidb/pkg/statistics/handle/SkipSystemTableCheck", `return(true)`))
+	defer failpoint.Disable("github.com/pingcap/tidb/pkg/statistics/handle/SkipSystemTableCheck")
```
D3Hunter: use `testfailpoint.Enable`

pkg/planner/core/casetest/dag/dag_test.go
```diff
@@ -79,13 +80,17 @@ func testDAGPlanBuilderSimpleCase(t *testing.T, testKit *testkit.TestKit, cascad
 }
 
 func TestDAGPlanBuilderSimpleCase(t *testing.T) {
+	require.NoError(t, failpoint.Enable("github.com/pingcap/tidb/pkg/statistics/handle/SkipSystemTableCheck", `return(true)`))
```
D3Hunter: here and others too

pkg/statistics/handle/handle.go
```diff
@@ -200,17 +202,72 @@ func (h *Handle) getStatsByPhysicalID(physicalTableID int64, tblInfo *model.Tabl
...
+		h.UpdateStatsCache(types.CacheUpdate{
+			Updated: []*statistics.Table{tbl},
+		})
 		return tbl, true
 	}
-	return nil, false
+
+	isSystemTable := false
+	// NOTE: You might worry that this slows down Get.
+	// It runs only once per non-partitioned table, or once per partition when the cache holds fewer than 64 entries.
```
qw4990: But if later this table is evicted from the Cache and we can't get it from the Cache, will we run this again?

qw4990: What if there are some highly-frequent queries accessing system tables, it seems like we have to run this process also frequently for these system tables, since system tables are not going to be put into the Cache.

qw4990: System tables IDs are fixed, it seems like we can just cache their table IDs in advance?

pkg/statistics/handle/handle.go
0xPoe: I have updated it as a note.

pkg/statistics/table.go
0xPoe: Good point! Updated.

pkg/session/bootstrap_test.go
0xPoe: Because the session pool was never used after the domain was closed, it was not an issue before. However, after we started using it for system table checks, it is now likely to be used.

pkg/statistics/handle/handle.go
0xPoe: > But if later this table is evicted from the Cache and we can't get it from the Cache, will we run this again?

The basic information never gets evicted.

> What if there are some highly-frequent queries accessing system tables, it seems like we have to run this process also frequently for these system tables, since system tables are not going to be put into the Cache.

In theory, we wouldn’t query system tables the same way we query user tables. However, this point might be worth testing to see how it performs.


> System tables IDs are fixed, it seems like we can just cache their table IDs in advance?


That’s a great idea, but these table IDs are not simply fixed table IDs — some of them are pre-assigned. I feel that relying solely on ID checks might easily cause other teams to miss new tables when adding them in the future. However, the filter we’re using now should be a commonly used API, so the risk of using it is relatively low.

pkg/statistics/handle/cache/statscache.go
0xPoe: Great idea! Updated!

0xPoe: Thank you!

pkg/session/syssession/session_integration_test.go
0xPoe: Updated! Thank you!

pkg/planner/core/casetest/dag/dag_test.go
0xPoe: Updated! Thank you!

0xPoe:
Tested again to make sure recent commits didn't break anything:
image

pkg/statistics/handle/handle.go
0xPoe: I ran this test for one night (20 goroutines nonstop query stats_meta table):
```go
package main

import (
	"context"
	"database/sql"
	"flag"
	"log"
	"os"
	"os/signal"
	"sync"
	"sync/atomic"
	"syscall"
	"time"

	_ "github.com/go-sql-driver/mysql"
)

func main() {
	dsn := flag.String("dsn", "root:@tcp(127.0.0.1:4000)/test", "TiDB/MySQL DSN")
	concurrency := flag.Int("concurrency", 20, "number of goroutines issuing SELECT statements")
	sleepBetween := flag.Duration("sleep", 0, "sleep between SELECT statements")
	flag.Parse()

	if *concurrency <= 0 {
		log.Fatalf("concurrency must be > 0 (got %d)", *concurrency)
	}

	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
	defer stop()

	db, err := sql.Open("mysql", *dsn)
	if err != nil {
		log.Fatalf("open database failed: %v", err)
	}
	defer db.Close()

	if err := db.PingContext(ctx); err != nil {
		log.Fatalf("ping database failed: %v", err)
	}

	var wg sync.WaitGroup
	var once sync.Once
	var firstErr atomic.Value
	var ops atomic.Int64

	runQuery := func(stmt string) error {
		rows, err := db.QueryContext(ctx, stmt)
		if err != nil {
			return err
		}
		defer rows.Close()

		cols, err := rows.Columns()
		if err != nil {
			return err
		}
		dummy := make([]any, len(cols))
		for i := range dummy {
			var holder any
			dummy[i] = &holder
		}

		for rows.Next() {
			if err := rows.Scan(dummy...); err != nil {
				return err
			}
		}
		if err := rows.Err(); err != nil {
			return err
		}
		return nil
	}

	worker := func() {
		defer wg.Done()
		for {
			select {
			case <-ctx.Done():
				return
			default:
			}
			const stmt = "SELECT * FROM mysql.stats_meta"

			if err := runQuery(stmt); err != nil {
				once.Do(func() {
					firstErr.Store(err)
					stop()
				})
				return
			}

			count := ops.Add(1)
			if count%100 == 0 {
				log.Printf("completed %d SELECT statements (last: %s)", count, stmt)
			}

			if *sleepBetween > 0 {
				select {
				case <-ctx.Done():
					return
				case <-time.After(*sleepBetween):
				}
			}
		}
	}

	for i := 0; i < *concurrency; i++ {
		wg.Add(1)
		go worker()
	}

	<-ctx.Done()
	wg.Wait()

	if errVal := firstErr.Load(); errVal != nil {
		log.Fatalf("stopped with error: %v", errVal.(error))
	}
}

```

The CPU usage:
image
The memeory usage:
image

0xPoe: It seems the impact is minimal. I don’t think anyone would query system tables repeatedly or at such a high frequency.
One of our major customers frequently uses SHOW STATS_META and SHOW STATS_HEALTHY, but these commands do not go through this code path, so they should not be affected.
Another possible way to fundamentally solve this issue is to pass the session into this function. However, that would involve quite a few changes, and I’m not sure it’s worth it.
cc: @time-and-fate @winoros If you have any thoughts, please feel free to comment.

0xPoe: I will also start a cluster with one million tables to test whether it has any negative impact.

0xPoe: image
It seems it has no significant effect on the real cluster. I couldn't even find it from the profile.

0xPoe: Another thing that makes it difficult to directly use the table ID for checking is the way IDs are assigned in the classic (non–next-gen) bootstrap path. In this mode, we don’t pre-assign IDs with the high-bit flag. When TiDB starts for the first time, doDDLWorks runs the regular CREATE TABLE ... statements defined in pkg/session/bootstrap.go, and the IDs for the mysql.* system tables are obtained from the normal global ID allocator (meta.Mutator.GenGlobalID). Because these IDs are generated in the same way as user table IDs, they don’t contain the autoid.SystemSchemaIDFlag and don’t fall within the “reserved” range.

That’s why our getStatsByPhysicalID logic still includes a fallback: if the flag check fails, it looks up the schema through the InfoSchema and relies on filter.IsSystemSchema (which checks the database name) to detect mysql, information_schema, and so on.

In the next-gen path, we create the same tables by directly injecting their metadata with IDs defined in pkg/meta/autoid/autoid.go, so the flag check alone is sufficient. The classic bootstrap path, however, still depends on the schema-name-based detection.

D3Hunter (commented):
/approve

ddl part lgtm

qw4990 (approved):

pkg/statistics/handle/handle.go
```diff
@@ -200,17 +202,71 @@ func (h *Handle) getStatsByPhysicalID(physicalTableID int64, tblInfo *model.Tabl
...
+	}
+
+	// The failpoint to skip system table check, for testing only.
+	skipSystemTableCheck := false
+	failpoint.Inject("SkipSystemTableCheck", func(val failpoint.Value) {
+		skip, ok := val.(bool)
+		if ok && skip {
+			skipSystemTableCheck = true
 		}
+	})
```
D3Hunter: maybe use `InjectCall`, pass pointer to skipSystemTableCheck as param

pkg/statistics/handle/handle.go
```diff
@@ -200,17 +214,95 @@ func (h *Handle) getStatsByPhysicalID(physicalTableID int64, tblInfo *model.Tabl
...
+// Stats healthy metrics almost never show pseudo tables, because once a DDL event is processed or the table is updated, real statistics are loaded into the cache.
+func (h *Handle) isSystemTable(physicalTableID int64, tblInfo *model.TableInfo) (bool, error) {
+	intest.Assert(tblInfo != nil, "tblInfo should not be nil for tableID %d", physicalTableID)
+	dbID := tblInfo.DBID
+	intest.Assert(dbID > 0, "invalid dbID %d for tableID %d", dbID, physicalTableID)
+	if filter.IsMemSchemaID(dbID) {
+		return true, nil
+	}
+
+	if _, ok := h.systemDBIDCache.Load(dbID); ok {
```
0xPoe, line 294: /cc @Leavrth

Is it safe to cache the system schema ID here? Do we have any chance of changing it during backup and restore?

pkg/util/filter/schema.go
```diff
@@ -37,3 +38,8 @@ func IsSystemSchema(schema string) bool {
 	return schema == DMHeartbeatSchema || schema == InspectionSchemaName ||
 		metadef.IsMemOrSysDB(schema)
 }
+
+// IsMemSchemaID checks whether schemaID is memory schema ID.
+func IsMemSchemaID(schemaID int64) bool {
```
D3Hunter: please move this into the `metadef/db.go`, which is where we want this types of methods defined, IMO, `util` pkg is a little abused right now, such as the SEM feature also resides in `util` pkg

qw4990 (approved):

0xPoe:
The latest test result:(QPS 4k, select * from mysql.stats_meta)
image

pkg/statistics/handle/handle.go
0xPoe: Offline discussion:
This is safe because system tables are renamed during backup and restore, but the schema ID never changes.

0xPoe:
/unhold

Still waiting for the stats owners to review it.

time-and-fate (approved):

0xPoe:
Just rebased. Nothing changed.

Issue #64094, opened 2025-10-22
Incorrect result for CONV with NULL parameters


sayJason:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```sql
DROP TABLE t1;
CREATE TABLE IF NOT EXISTS t1 (c1 SMALLINT UNSIGNED);
INSERT INTO t1 VALUES (23);
SELECT CONV(c1, c1, TRUNCATE(c1, NULL)) FROM t1; -- actual:{23}, expected:{NULL}
```
### 2. What did you expect to see? (Required)
NULL value should be returned.
### 3. What did you see instead (Required)
23 was returned.
### 4. What is your TiDB version? (Required)

Release Version: v8.5.3
Edition: Community
Git Commit Hash: dc2548aac79a712265e831cff2a3a896bc0a5a38
Git Branch: HEAD
UTC Build Time: 2025-07-31 13:54:43
GoVersion: go1.23.8
Race Enabled: false
Check Table Before Drop: false
Store: tikv

PR #64102, opened 2025-10-22
planner: fix TRUNCATE null handling with unsigned-flag check

### LINKED ISSUE #64095: Failed to check NULL parameters in TRUNCATE operation
## Bug Report

```sql
DROP TABLE t1;
CREATE TABLE t1 (c1 MEDIUMINT UNSIGNED PRIMARY KEY);
INSERT INTO t1 VALUES (0);
SELECT TRUNCATE(c1 = -1, -1.0 << NULL) FROM t1; -- acutal: {0}, expected: {NULL}
```
NULL should be returned.
0 was returned.

 Release Version: v8.5.3
Edition: Community
Git Commit Hash: dc2548aac79a712265e831cff2a3a896bc0a5a38
Git Branch: HEAD
UTC Build Time: 2025-07-31 13:54:43
GoVersion: go1.23.8
Race Enabled: false
Store: tikv

cpd007:

Issue Number: close #64095 #64094

Problem Summary: Failed to check NULL parameters in TRUNCATE operation

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
This PR fixes an issue where `TRUNCATE(x, d)` failed to return `NULL` when `d` was `NULL`.
The function returned `true` flag when evaluating  (`args[1]`),
skipping proper `NULL` propagation.
Adds a regression test for `SELECT TRUNCATE(c1 = -1, -1.0 << NULL)`.
```

### DIFF
--- a/pkg/expression/builtin_math.go
+++ b/pkg/expression/builtin_math.go
@@ -2154,19 +2154,13 @@ func (b *builtinTruncateIntSig) Clone() builtinFunc {
 // See https://dev.mysql.com/doc/refman/5.7/en/mathematical-functions.html#function_truncate
 func (b *builtinTruncateIntSig) evalInt(ctx EvalContext, row chunk.Row) (int64, bool, error) {
 	x, isNull, err := b.args[0].EvalInt(ctx, row)
-	if err != nil {
+	if isNull || err != nil {
 		return 0, true, err
-	}
-	if isNull {
-		return 0, true, nil
 	}
 
 	d, isNull, err := b.args[1].EvalInt(ctx, row)
-	if err != nil {
+	if isNull || err != nil {
 		return 0, true, err
-	}
-	if isNull {
-		return 0, true, nil
 	}
 
 	if mysql.HasUnsignedFlag(b.args[1].GetType(ctx).GetFlag()) {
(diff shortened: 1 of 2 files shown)


cpd007:
Screenshot 2025-10-22 at 7 31 33 PM

cpd007:
@XuHuaiyu can you please review this?

lance6716 (commented):
please avoid unneccesary changes.

cpd007:
the `check_dev_2` is having timeout failure, I am not able link it to my pr changes.

pkg/expression/builtin_math.go
```diff
@@ -2154,13 +2154,19 @@ func (b *builtinTruncateIntSig) Clone() builtinFunc {
 // See https://dev.mysql.com/doc/refman/5.7/en/mathematical-functions.html#function_truncate
 func (b *builtinTruncateIntSig) evalInt(ctx EvalContext, row chunk.Row) (int64, bool, error) {
 	x, isNull, err := b.args[0].EvalInt(ctx, row)
-	if isNull || err != nil {
-		return 0, isNull, err
+	if err != nil {
```
lance6716: we have many `if isNull || err != nil` elsewhere. In order to keep same code style, please revert this change

pkg/expression/builtin_math.go
```diff
@@ -2195,25 +2203,36 @@ type builtinTruncateUintSig struct {
...
+		return 0, true, err
+	}
+	if isNull {
+		return 0, true, nil
 	}
+
+	uintx := uint64(x)
```
lance6716: why move it? Seems the PR has no effective code change. And without any code change, the added unit test can still pass. Please add a effective test in PR.

cpd007: the true return for error case was missing for args[1] basically. moved it so that we do this op after both the args have been evaluated

pkg/expression/builtin_math.go
cpd007: done

pkg/expression/builtin_math_test.go
```diff
@@ -525,6 +525,8 @@ func TestTruncate(t *testing.T) {
 		{[]any{0, 400}, 0},
 		{[]any{0, -400}, 0},
 		{[]any{0, 3}, 0},
+		// for issue #64095
+		{[]any{0, nil}, nil},
```
lance6716, line 529: 1. Without any code change, this unit test can stil pass. Please add an effective test in this PR.
2. Where's the code change of the release note "the wrong argument (`args[1]` instead of `args[0]`)"?

cpd007: I have updated the release note, my bad, The true return for error case was missing for` args[1]` basically.

cpd007: i have also uploaded results from my local testing

lance6716: 1. Without any code change, this unit test can stil pass. So delete this line because it has no effect.
2. Please add an effective test in this PR. For example, you can search for `MustQuery` to add a SQL test to check something like `SELECT TRUNCATE(c1 = -1, -1.0 << NULL) FROM t1`

PR #64121, opened 2025-10-26
*: extract column into map in the constant propagation

### LINKED ISSUE #64128: unnecessary to extract column into slices when to constant propagation
## Bug Report

Image

In the `solve` of constant propagation, the slices of `ExtractColumn` will be converted into a map and deduplicated. Therefore, directly converting them into a map is better than converting them into slices.

hawkingrei:

Issue Number: close #64128

Problem Summary:

Image

In the solve of constant propagation, the slices of ExtractColumn will be converted into a map and deduplicated. Therefore, directly converting them into a map is better than converting them into slices.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test

before 
```
BenchmarkTPCHQ3-8   	     924	   1159361 ns/op	 1739137 B/op	   23110 allocs/op
BenchmarkTPCDSQ64-8   	     153	   7266446 ns/op	12807037 B/op	  137175 allocs/op
```

after 
```
BenchmarkTPCHQ3-8   	     966	   1149071 ns/op	 1738760 B/op	   23045 allocs/op
BenchmarkTPCDSQ64-8   	     165	   6948597 ns/op	12797896 B/op	  136460 allocs/op
```

- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/expression/constant_propagation.go
+++ b/pkg/expression/constant_propagation.go
@@ -488,12 +488,7 @@ func (s *propConstSolver) solve(keepJoinKey bool, conditions []Expression) []Exp
 func (s *propConstSolver) extractColumns(conditions []Expression) {
 	mp := GetUniqueIDToColumnMap()
 	defer PutUniqueIDToColumnMap(mp)
-	for _, cond := range conditions {
-		s.conditions = append(s.conditions, SplitCNFItems(cond)...)
-		ExtractColumnsMapFromExpressionsWithReusedMap(mp, nil, cond)
-		s.insertCols(mp)
-		clear(mp)
-	}
+	s.conditions = s.extractColumnsInternal(mp, s.conditions, conditions)
 }
 
 // PropagateConstantForJoin propagate constants for inner joins.
@@ -629,6 +624,16 @@ func (s *basePropConstSolver) dealWithPossibleHybridType(col *Column, con *Const
 		return con, true
 	}
 	return nil, false
+}
+
+func (s *basePropConstSolver) extractColumnsInternal(mp map[int64]*Column, splitConds []Expression, conds []Expression) []Expression {
+	for _, cond := range conds {
+		splitConds = append(splitConds, SplitCNFItems(cond)...)
+		ExtractColumnsMapFromExpressionsWithReusedMap(mp, nil, cond)
+		s.insertCols(mp)
+		clear(mp)
+	}
+	return splitConds
 }
 
 // pickEQCondsOnOuterCol picks constant equal expression from specified conditions.
@@ -869,16 +874,6 @@ func (s *propOuterJoinConstSolver) extractColumns(joinConds, filterConds []Expre
 	defer PutUniqueIDToColumnMap(mp)
 	s.joinConds = s.extractColumnsInternal(mp, s.joinConds, joinConds)
 	s.filterConds = s.extractColumnsInternal(mp, s.filterConds, filterConds)
-}
-
-func (s *propOuterJoinConstSolver) extractColumnsInternal(mp map[int64]*Column, splitConds []Expression, conds []Expression) []Expression {
-	for _, cond := range conds {
-		splitConds = append(splitConds, SplitCNFItems(cond)...)
-		ExtractColumnsMapFromExpressionsWithReusedMap(mp, nil, cond)
-		s.insertCols(mp)
-		clear(mp)
-	}
-	return splitConds
 }
 
 // propagateConstantDNF find DNF item from CNF, and propagate constant inside DNF.


pkg/expression/constant_propagation.go
```diff
@@ -469,10 +469,14 @@ func (s *propConstSolver) solve(keepJoinKey bool, conditions []Expression) []Exp
...
 	s.conditions = slices.Grow(s.conditions, len(conditions))
+	mp := GetUniqueIDToColumnMap()
 	for _, cond := range conditions {
 		s.conditions = append(s.conditions, SplitCNFItems(cond)...)
-		s.insertCols(ExtractColumns(cond)...)
+		ExtractColumnsMapFromExpressionsWithReusedMap(mp, nil, cond)
+		s.insertCols(mp)
+		clear(mp)
 	}
+	PutUniqueIDToColumnMap(mp)
```
windtalker: maybe use defer to put mp back?

pkg/planner/core/planbuilder.go
```diff
@@ -2101,7 +2101,7 @@ func (b *PlanBuilder) addColumnsWithVirtualExprs(tbl *resolve.TableNameW, cols *
 	}
 
 	virtualExprs := columnSelector(columns)
-	relatedCols := make(map[int64]*expression.Column, len(tblInfo.Columns))
+	relatedCols := expression.GetUniqueIDToColumnMap()
```
windtalker, line 2104: ditto

pkg/expression/constant_propagation.go
hawkingrei: The lifecycle here is clear, and the optimization steps are very time-consuming. So we should put it into the pool immediately after it has been used up.

hawkingrei: I refactor the code to use defer

pkg/planner/core/planbuilder.go
hawkingrei: Done

pkg/expression/util.go
```diff
@@ -188,6 +189,23 @@ func ExtractColumnsMapFromExpressions(filter func(*Column) bool, exprs ...Expres
...
+func GetUniqueIDToColumnMap() map[int64]*Column {
+	return uniqueIDToColumnMapPool.Get().(map[int64]*Column)
+}
+
+// PutUniqueIDToColumnMap puts map[int64]*Column map back to the pool.
+func PutUniqueIDToColumnMap(m map[int64]*Column) {
+	clear(m)
+	uniqueIDToColumnMapPool.Put(m)
+}
+
```
henrybw, line 208: Why do we need a pool of these? Are these created and destroyed frequently enough that it measurably hurts performance to not use a pool? I would be curious what the performance impact of not using a pool would be.

pkg/expression/constant_propagation.go
```diff
@@ -863,6 +864,23 @@ func (s *propOuterJoinConstSolver) solve(keepJoinKey bool, joinConds, filterCond
...
+		ExtractColumnsMapFromExpressionsWithReusedMap(mp, nil, cond)
+		s.insertCols(mp)
+		clear(mp)
+	}
+	for _, cond := range filterConds {
+		s.filterConds = append(s.filterConds, SplitCNFItems(cond)...)
+		ExtractColumnsMapFromExpressionsWithReusedMap(mp, nil, cond)
+		s.insertCols(mp)
+		clear(mp)
+	}
```
henrybw: These loops seem very similar to each other, and to the above `extractColumns` implementation. Is there any way we can refactor this so there is one implementation of this pattern of looping?

pkg/expression/util.go
hawkingrei: Memory reuse can reduce the number of objects on the heap. In fact, TiDB has always had a serious issue: it doesn’t pay much attention to memory reuse, which leads to excessive memory allocation on the heap. As a result, the GoLang GC needs to scan a large number of small objects, significantly increasing the Stop-The-World time.

hawkingrei: Another reason is that the map size here is not easy to determine, which makes the map prone to resizing. The performance of resizing a map in GoLang is relatively poor, as it involves creating a new map. Therefore, the advantage of reusing a map is that the final reused map can be utilized again, avoiding the possibility of further resizing.

pkg/expression/constant_propagation.go
hawkingrei: I have refactored this part of the code.

pkg/expression/util.go
hawkingrei: image

The phenomenon looks like this

hawkingrei:
> Overall I think this is okay, but I'm skeptical that an object pool for these maps is necessary.
> 
> How often do these maps actually get reused? In addition to making the code more complicated, I'm concerned that using a pool will lead to us hanging onto more maps than we actually need: a pool that is mostly unused seems wasteful compared to just allocating maps each time.
> 
> Do you have any measurements or performance data for how much these maps are reused?

Actually, there's a trick here: TiDB supports Pyroscope. You just need to inject Pyroscope information into the environment to continuously collect profilers and obtain detailed profiler data over a large time span.

hawkingrei:
> Overall I think this is okay, but I'm skeptical that an object pool for these maps is necessary.
> 
> How often do these maps actually get reused? In addition to making the code more complicated, I'm concerned that using a pool will lead to us hanging onto more maps than we actually need: a pool that is mostly unused seems wasteful compared to just allocating maps each time.
> 
> Do you have any measurements or performance data for how much these maps are reused?

Additionally, the benchmark data can be obtained from the benchmark results within the `tpch/tpcds` directory in `planner/core/casetest`. The `gobench` tool will display the number of allocations, and we can observe a reduction in allocations, indicating that our optimizations have taken effect.

I have already posted the gobench data in the issue description.

pkg/expression/util.go
henrybw: Makes sense. Thanks for explaining.

henrybw (approved):
Thanks! LGTM.

Issue #64158, opened 2025-10-28
The analyze process might be slowed down by the SELECT FOR UPDATE statement


0xPoe:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
1. Create a cluster and create a bunch of tables
2. Continuously update and analyze the tables

### 2. What did you expect to see? (Required)
No internal slow queues.

### 3. What did you see instead (Required)
```sql
| id                | estRows | estCost | actRows | task | access object                         | execution info                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           | operator info                      | memory | disk  |
| Batch_Point_Get_1 | 2.00    | 0.00    | 1       | root | table:stats_meta, index:tbl(table_id) | time:28.1s, loops:19, BatchGet:{num_rpc:36, total_time:19.6ms}, time_detail: {total_kv_read_wall_time: 3.03ms, tikv_wall_time: 3.64ms}, scan_detail: {total_process_keys: 18, total_process_keys_size: 792, total_keys: 54, get_snapshot_time: 202.6µs, rocksdb: {block: {cache_hit_count: 673}}}, lock_keys: {time:1.25ms, region:3, keys:3, slowest_rpc: {total: 0.001s, region_id: 4414277, store: db-tikv-17.db-tikv-peer.zxxxxxxxxxxxx.svc:20160, time_detail: {tikv_wall_time: 152.5µs}, scan_detail: {get_snapshot_time: 4.37µs, rocksdb: {key_skipped_count: 17, block: {cache_hit_count: 32}}}, }, lock_rpc:1.485808ms, rpc_count:3} | keep order:false, desc:false, lock | N/A    | N/A   |
```

### 4. What is your TiDB version? (Required)

master,v8.5.x

0xPoe:
Introduced by: https://github.com/pingcap/tidb/pull/62327

0xPoe:
Root cause: Under the RR (Repeatable Read) isolation level, even non-existent keys can be locked. As a result, the use of this fake table ID caused significant lock contention.

Issue #64062, opened 2025-10-21
flaky test raise ```pessimistic lock retry limit reached``` in the mysql test


hawkingrei:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 3. What did you see instead (Required)

```
time="2025-10-21T13:39:10+08:00" level=error msg="1 tests failed\n"

time="2025-10-21T13:39:10+08:00" level=error msg="run test [with_non_recursive_bugs] err: sql:ANALYZE TABLE a,d,bb,cc,dd;: run \"ANALYZE TABLE a,d,bb,cc,dd;\" at line 752 err Error 1105 (HY000): pessimistic lock retry limit reached"

tidb-server(PID: 1689) stopped
```

https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fghpr_mysql_test/detail/ghpr_mysql_test/41225/pipeline

zimulala:
Reproduced: https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fghpr_mysql_test/detail/ghpr_mysql_test/41469/pipeline

PR #64161, opened 2025-10-28
statistics: use negative table IDs as fake table IDs

0xPoe:

Issue Number: close https://github.com/pingcap/tidb/issues/64158 close https://github.com/pingcap/tidb/issues/64062

Problem Summary:

We need to use a proper fake ID to avoid lock contention under the RR (Repeatable Read) isolation level.
So I used the table itself to generate a better fake ID.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] [Manual test](https://github.com/pingcap/tidb/pull/64161#issuecomment-3456077960)
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/statistics/handle/storage/save.go
+++ b/pkg/statistics/handle/storage/save.go
@@ -16,14 +16,10 @@ package storage
 
 import (
 	"context"
-	"crypto/rand"
 	"encoding/json"
 	"fmt"
-	"math"
-	"math/big"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/failpoint"
@@ -121,16 +117,6 @@ func saveBucketsToStorage(sctx sessionctx.Context, tableID int64, isIndex int, h
 	return
 }
 
-// getRandomFakeID generates a random negative int64 as a fake table ID.
-// NOTE: we don't demand extremely high concurrency here, so the overhead of crypto/rand is negligible.
-// We prefer crypto/rand for its stronger randomness characteristics.
-func getRandomFakeID() int64 {
-	// Int cannot return an error when using rand.Reader.
-	id, _ := rand.Int(rand.Reader, big.NewInt(math.MaxInt64))
-	// Make sure the fake ID is negative.
-	return -id.Int64() - 1
-}
-
 // SaveAnalyzeResultToStorage saves the analyze result to the storage.
 func SaveAnalyzeResultToStorage(sctx sessionctx.Context,
 	results *statistics.AnalyzeResults, analyzeSnapshot bool) (statsVer uint64, err error) {
@@ -165,17 +151,15 @@ func SaveAnalyzeResultToStorage(sctx sessionctx.Context,
 	// txn1: lockKeys on point get (index lock)
 	// txn2: lockKeys on batch point get (row lock) — waits for txn1 for index lock
 	// txn1: lockKeys on point get (row lock) — deadlock occurs here and it's not retryable
-	// We need to use a random fake ID here to avoid heavy lock contention on the stats_meta table.
-	// Under RR (Repeatable Read) isolation level, non-existent keys are also locked.
-	start := time.Now()
-	fakeID := getRandomFakeID()
+	// Pick a fake table ID to reduce lock contention on mysql.stats_meta.
+	// Under repeatable-read, even non-existent keys are locked.
+	// Using the negative table ID keeps it separate from real table IDs and from other fake IDs.
+	fakeID := -tableID
 	tableIDStrs := []string{strconv.FormatInt(fakeID, 10), strconv.FormatInt(tableID, 10)}
 	rs, err = util.Exec(sctx, "select snapshot, count, modify_count from mysql.stats_meta where table_id in (%?) for update", tableIDStrs)
 	if err != nil {
 		return 0, err
 	}
-	// TODO: clean up the debug log later.
-	statslogutil.StatsLogger().Info("lock stats_meta", zap.Int64("tableID", tableID), zap.Duration("takeTime", time.Since(start)))
 	var rows []chunk.Row
 	rows, err = sqlexec.DrainRecordSet(ctx, rs, sctx.GetSessionVars().MaxChunkSize)
 	if err != nil {


0xPoe:
Tested locally:
1. Create one real cluster.
2. Run the following program:
```go
package main

import (
	"context"
	"database/sql"
	"flag"
	"fmt"
	"log"
	"math/rand"
	"os"
	"os/signal"
	"runtime"
	"strings"
	"sync"
	"sync/atomic"
	"syscall"
	"time"

	_ "github.com/go-sql-driver/mysql"
)

const letters = "abcdefghijklmnopqrstuvwxyz"

func main() {
	rand.Seed(time.Now().UnixNano())

	dsn := flag.String("dsn", "root:@tcp(127.0.0.1:4000)/test", "TiDB/MySQL DSN (include target database)")
	tablePrefix := flag.String("table-prefix", "bench_table", "prefix for generated tables")
	tableCount := flag.Int("table-count", 2000, "number of tables to create if missing")
	insertBatch := flag.Int("insert-batch", 1, "number of rows to insert per insert operation")
	analyzeProbability := flag.Float64("analyze-probability", 0.1, "probability [0,1] that an operation runs ANALYZE instead of INSERT")
	sleepBetweenOps := flag.Duration("sleep", 0, "sleep duration between operations")
	analyzeAllColumns := flag.Bool("analyze-all-columns", true, "use `ANALYZE TABLE ... ALL COLUMNS` when running analyze")
	createConcurrency := flag.Int("create-concurrency", runtime.NumCPU(), "number of concurrent workers creating missing tables")
	workConcurrency := flag.Int("work-concurrency", runtime.NumCPU(), "number of concurrent workers issuing insert/analyze operations")
	flag.Parse()

	if *tableCount <= 0 {
		log.Fatalf("table-count must be > 0 (got %d)", *tableCount)
	}
	if *insertBatch <= 0 {
		log.Fatalf("insert-batch must be > 0 (got %d)", *insertBatch)
	}
	if *analyzeProbability < 0 || *analyzeProbability > 1 {
		log.Fatalf("analyze-probability must be between 0 and 1 (got %f)", *analyzeProbability)
	}
	if *createConcurrency <= 0 {
		log.Fatalf("create-concurrency must be > 0 (got %d)", *createConcurrency)
	}
	if *workConcurrency <= 0 {
		log.Fatalf("work-concurrency must be > 0 (got %d)", *workConcurrency)
	}

	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
	defer stop()

	db, err := sql.Open("mysql", *dsn)
	if err != nil {
		log.Fatalf("open database failed: %v", err)
	}
	defer db.Close()

	log.Printf("ensuring %d tables with prefix %q exist...", *tableCount, *tablePrefix)
	if err := ensureTables(ctx, db, *tablePrefix, *tableCount, *createConcurrency); err != nil {
		log.Fatalf("failed to prepare tables: %v", err)
	}
	log.Printf("table preparation finished")

	log.Printf("starting workload: insert-batch=%d analyze-probability=%.2f sleep=%s", *insertBatch, *analyzeProbability, sleepBetweenOps.String())
	if err := runWorkload(ctx, db, *tablePrefix, *tableCount, *insertBatch, *analyzeProbability, *analyzeAllColumns, *sleepBetweenOps, *workConcurrency); err != nil && !isContextCanceled(err) {
		log.Fatalf("workload stopped with error: %v", err)
	}
	log.Printf("workload stopped")
}

func ensureTables(ctx context.Context, db *sql.DB, prefix string, count int, concurrency int) error {
	if concurrency > count {
		concurrency = count
	}
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()

	tableIndexes := make(chan int)
	var wg sync.WaitGroup
	var once sync.Once
	var firstErr error
	var created atomic.Int64

	createTable := func(idx int) error {
		name := tableName(prefix, idx)
		createStmt := fmt.Sprintf(
			"CREATE TABLE IF NOT EXISTS `%s` ("+
				"id BIGINT UNSIGNED NOT NULL AUTO_INCREMENT,"+
				"k BIGINT NOT NULL,"+
				"v VARCHAR(255) NOT NULL,"+
				"updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,"+
				"PRIMARY KEY (id),"+
				"KEY idx_k (k)"+
				") ENGINE=InnoDB",
			name,
		)
		if _, err := db.ExecContext(ctx, createStmt); err != nil {
			return fmt.Errorf("create table %s: %w", name, err)
		}
		total := created.Add(1)
		if total%250 == 0 || int(total) == count {
			log.Printf("created %d/%d tables", total, count)
		}
		return nil
	}

	worker := func() {
		defer wg.Done()
		for idx := range tableIndexes {
			if err := ctx.Err(); err != nil {
				return
			}
			if err := createTable(idx); err != nil {
				once.Do(func() {
					firstErr = err
					cancel()
				})
				return
			}
		}
	}

	for i := 0; i < concurrency; i++ {
		wg.Add(1)
		go worker()
	}

loop:
	for i := 0; i < count; i++ {
		select {
		case <-ctx.Done():
			break loop
		case tableIndexes <- i:
		}
	}
	close(tableIndexes)
	wg.Wait()

	if firstErr != nil {
		return firstErr
	}
	return ctx.Err()
}

func runWorkload(
	ctx context.Context,
	db *sql.DB,
	prefix string,
	count int,
	batch int,
	analyzeProbability float64,
	analyzeAllColumns bool,
	sleep time.Duration,
	concurrency int,
) error {
	if concurrency <= 0 {
		concurrency = 1
	}
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()

	var wg sync.WaitGroup
	var once sync.Once
	var firstErr error
	var ops atomic.Int64

	worker := func() {
		defer wg.Done()
		for {
			if err := ctx.Err(); err != nil {
				return
			}
			table := tableName(prefix, rand.Intn(count))
			var err error
			if rand.Float64() < analyzeProbability {
				err = analyzeTable(ctx, db, table, analyzeAllColumns)
			} else {
				err = insertRows(ctx, db, table, batch)
			}
			if err != nil {
				once.Do(func() {
					firstErr = err
					cancel()
				})
				return
			}
			total := ops.Add(1)
			if total%100 == 0 {
				log.Printf("completed %d operations (last on %s)", total, table)
			}
			if sleep > 0 {
				select {
				case <-ctx.Done():
					return
				case <-time.After(sleep):
				}
			}
		}
	}

	for i := 0; i < concurrency; i++ {
		wg.Add(1)
		go worker()
	}

	wg.Wait()
	if firstErr != nil {
		return firstErr
	}
	return ctx.Err()
}

func insertRows(ctx context.Context, db *sql.DB, table string, batch int) error {
	var sb strings.Builder
	sb.WriteString("INSERT INTO `")
	sb.WriteString(table)
	sb.WriteString("` (k, v) VALUES ")
	args := make([]any, 0, batch*2)
	for i := 0; i < batch; i++ {
		if i > 0 {
			sb.WriteString(",")
		}
		sb.WriteString("(?, ?)")
		args = append(args, rand.Int63(), randomValue(32))
	}
	if _, err := db.ExecContext(ctx, sb.String(), args...); err != nil {
		return fmt.Errorf("insert into %s: %w", table, err)
	}
	return nil
}

func analyzeTable(ctx context.Context, db *sql.DB, table string, allColumns bool) error {
	stmt := fmt.Sprintf("ANALYZE TABLE `%s`", table)
	if allColumns {
		stmt += " ALL COLUMNS"
	}
	if _, err := db.ExecContext(ctx, stmt); err != nil {
		return fmt.Errorf("analyze table %s: %w", table, err)
	}
	return nil
}

func tableName(prefix string, idx int) string {
	return fmt.Sprintf("%s_%04d", prefix, idx)
}

func randomValue(length int) string {
	var b strings.Builder
	for i := 0; i < length; i++ {
		b.WriteByte(letters[rand.Intn(len(letters))])
	}
	return b.String()
}

func isContextCanceled(err error) bool {
	return err == context.Canceled || err == context.DeadlineExceeded
}
```
3. Check slow queries:
image
image
4. Replace tidb with the patch version

5. Check the slow queries again(The slow query disappers)
image

pkg/statistics/handle/storage/save.go
```diff
@@ -151,12 +165,17 @@ func SaveAnalyzeResultToStorage(sctx sessionctx.Context,
 	// txn1: lockKeys on point get (index lock)
 	// txn2: lockKeys on batch point get (row lock) — waits for txn1 for index lock
 	// txn1: lockKeys on point get (row lock) — deadlock occurs here and it's not retryable
-	fakeID := int64(-1988)
+	// We need to use a random fake ID here to avoid heavy lock contention on the stats_meta table.
+	// Under RR (Repeatable Read) isolation level, non-existent keys are also locked.
+	start := time.Now()
+	fakeID := getRandomFakeID()
```
windtalker: it looks to me the fake id is only used to force convert the point get plan to batch point get plan, why not just use `-tableID` instead of randomFakeID

0xPoe: Great idea! Thank you! I will update it.

pkg/statistics/handle/storage/save.go
```diff
@@ -151,7 +151,10 @@ func SaveAnalyzeResultToStorage(sctx sessionctx.Context,
 	// txn1: lockKeys on point get (index lock)
 	// txn2: lockKeys on batch point get (row lock) — waits for txn1 for index lock
 	// txn1: lockKeys on point get (row lock) — deadlock occurs here and it's not retryable
-	fakeID := int64(-1988)
+	// Pick a fake table ID to reduce lock contention on mysql.stats_meta.
+	// Under repeatable-read, even non-existent keys are locked.
+	// Using the negative table ID keeps it separate from real table IDs and from other fake IDs.
```
lance6716, line 156: If there are many locks on the same table, using the same table ID still causes high contention?

/hold

free to unhold after address comments

AilinKid: how could non-existent keys are locked and even contented, >.< ! txn strange behavior

AilinKid: anyway,LGTM

0xPoe: > If there are many locks on the same table, using the same table ID still causes high contention?

Generally, this kind of situation is unlikely to happen, unless there are about 10 instances analyzing the same table at the same time.

/unhold

0xPoe: This PR fixes the issue where analyzing 10 different tables caused lock contention because they were all using the same fake table ID.

0xPoe: > how could non-existent keys are locked and even contented, >.< ! txn strange behavior

It also blows my mind. I understand that this is to address the phantom read problem — to prevent this key from being inserted by another transaction after the select, which would otherwise cause a phantom read.

PR #61634, opened 2025-06-10
*: add the new SEMv2 config and implement most of the functions

YangKeao:
### What problem does this PR solve?

Issue Number: close #61582, close #61583, close #61584, close #61585

Problem Summary:

The legacy SEM has many hard coded rules. This PR introduced a new SEM mechanism to make it possible to update the SEM rules without changing the code.

### What changed and how does it work?

1. Add a new config `security.sem-config`.
2. Define the format of sem config and add a package `/pkg/util/sem/v2` to parse this sem config.
3. Add a package as the compatibility wrapper between legacy sem and the new sem in `/pkg/util/sem/compat`.
4. Rewrite most of the legacy sem logic with the wrapper.
5. Modify most of the SEM tests to test both the legacy one and the new one.

The only exception is that the restricted SQLs are not implemented yet, because it depends on https://github.com/pingcap/tidb/pull/61402.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
Add a config `security.sem-config` to make the SEM rule configurable.
```

Issue #61581, opened 2025-06-09
SEMv2 issue track


YangKeao:
## Enhancement

We are going to introduce a new version of SEM. The main benefit of it is to provide dynamically modifiable rules rather than the hard coded one. This issue is a track issue for the development progress:

- [ ] Add SEMv2 config and parser.
- [ ] Implement `restricted_databases` and `restricted_tables`
- [ ] Implement `restricted_variables`
- [ ] Implement `restricted_privileges`
- [ ] Implement `restricted_sql`

PR #64170, opened 2025-10-29
importinto: fix some check and external-id not set after SEM v2 introduced

D3Hunter:

Issue Number: ref #61702, #61581

Problem Summary:

SEM v2 introduced in #61634, it changes some method calls, and cause some import-into related logic not working in SEM v2:
- fix the bug that `external-id` not set in SEM v2
- fix import-into SEM related logic not tested with SEM v2
- fix `disallowedOptionsForSEM` not checked in SEM v2
- some changes to SEM v2 to help tests.
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/import_into_test.go
+++ b/pkg/executor/import_into_test.go
@@ -159,7 +159,6 @@ func TestNextGenS3ExternalID(t *testing.T) {
 				})
 				tk.MustMatchErrMsg("IMPORT INTO test.t FROM 's3://bucket?EXTERNAL-ID=abc'", `(?i).*Feature 'IMPORT INTO .*external.*' is not supported when security enhanced mode is enabled`)
 			})
-
 		}
 	})
 


YangKeao:
`IMPORT INTO t FROM 's3://xxx/xxx?external-id=xxx';` is guarded by `import_with_external_id` sql rule.

For example, with SEM config:

```
{
  "version":"1","tidb_version":"8.0.0",
  "restricted_privileges": ["RESTRICTED_SQL_ADMIN"],
  "restricted_sql": {
    "rule": [
      "time_to_live",
      "alter_table_attributes",
      "import_with_external_id"
    ],
    "sql": [
      "BACKUP",
      "RESTORE",
      "ALTER RESOURCE GROUP",
      "CREATE RESOURCE GROUP"
    ]
  }
}
```

The following sql will return error:

```
mysql> IMPORT INTO t FROM 's3://xxx/xxx?external-id=xxx';
ERROR 8132 (HY000): Feature 'IMPORT INTO t FROM 's3://xxx/xxx?external-id=xxx'' is not supported when security enhanced mode is enabled
```

This design is different for SEMv1 and v2.

D3Hunter:
> `IMPORT INTO t FROM 's3://xxx/xxx?external-id=xxx';` is guarded by `import_with_external_id` sql rule.
> 
> For example, with SEM config:
> 
> ```
> {
>   "version":"1","tidb_version":"8.0.0",
>   "restricted_privileges": ["RESTRICTED_SQL_ADMIN"],
>   "restricted_sql": {
>     "rule": [
>       "time_to_live",
>       "alter_table_attributes",
>       "import_with_external_id"
>     ],
>     "sql": [
>       "BACKUP",
>       "RESTORE",
>       "ALTER RESOURCE GROUP",
>       "CREATE RESOURCE GROUP"
>     ]
>   }
> }
> ```
> 
> The following sql will return error:
> 
> ```
> mysql> IMPORT INTO t FROM 's3://xxx/xxx?external-id=xxx';
> ERROR 8132 (HY000): Feature 'IMPORT INTO t FROM 's3://xxx/xxx?external-id=xxx'' is not supported when security enhanced mode is enabled
> ```
> 
> This design is different for SEMv1 and v2.

we need `external-id` set when it's not set when sem enabled, but it's not working in SEM v2

pkg/executor/importer/import.go
```diff
@@ -690,6 +691,8 @@ func (p *Plan) initOptions(ctx context.Context, seCtx sessionctx.Context, option
 		if p.DataSourceType == DataSourceTypeQuery {
 			return plannererrors.ErrNotSupportedWithSem.GenWithStackByArgs("IMPORT INTO from select")
 		}
+	}
+	if kerneltype.IsNextGen() && sem.IsEnabled() {
```
YangKeao, line 695: Another option is to use `sql_rule` to block the SQL. However, it's also a hard code logic. I think both methods are fine.

The advantages of using `sql_rule` is that the user with `RESTRICTED_SQL_ADMIN` privilege can bypass the check, but I'm not sure whether it's needed for this function.

pkg/planner/core/planbuilder.go
```diff
@@ -4722,18 +4722,27 @@ func (b *PlanBuilder) buildImportInto(ctx context.Context, ld *ast.ImportIntoStm
...
 					return nil, err
 				}
-				ld.Path = newPath
 			}
 		}
+		// a nextgen cluster might be shared by multiple tenants, and they might
+		// share the same AWS role to access import-into source data bucket, this
+		// external ID can be used to restrict the access only to the current tenant.
+		// when SEM enabled, we need set it.
+		if kerneltype.IsNextGen() && sem.IsEnabled() && storage.IsS3(u) {
```
YangKeao, line 4740: It's also a hard code logic to set the `external-id` option. However, we don't have good abstraction on similar logic for SEMv2, so adding these codes are the only way to meet the requirement of this scenario.

pkg/executor/importer/import.go
D3Hunter: we expect this logic always enabled when SEM enabled, regardless of whether it's configed or not

pkg/planner/core/planbuilder.go
D3Hunter: we don't want this configurable, we need it enabled all the time when sem enabled

PR #64184, opened 2025-10-29
pkg/parser/ast: (Fix) Propagate error correctly

### LINKED ISSUE #64183: Correctly propagate error
The `pkg/parser/ast/dml.go` file has some functions where the returned error is not checked and the error is not bubbled up when calling "Restore()" method

There are 3 calls to `` where this happens:

https://github.com/pingcap/tidb/blob/master/pkg/parser/ast/dml.go#L3456
https://github.com/pingcap/tidb/blob/master/pkg/parser/ast/dml.go#L3463
https://github.com/pingcap/tidb/blob/master/pkg/parser/ast/dml.go#L3525

This was discovered when implementing custom types  (Not using `parser_driver` or `test_driver`) and the Restore method implementation of either the ValueExpr or the ParamMarkerExpr was returning an error.
I  this case, Restore was still returning `nil` and the generated string from Restore call was an invalid SQL query.

jfrabaute-arista:
This PR fix a problem in the parser to return errors from nested functions correctly.

Issue Number: close #64183 

Tests:
-  [X] Unit test

No side effect.

No documentation change.

### DIFF
--- a/pkg/parser/ast/dml.go
+++ b/pkg/parser/ast/dml.go
@@ -3379,7 +3379,9 @@ func (n *ShowStmt) Restore(ctx *format.RestoreCtx) error {
 			ctx.WritePlainf("%d", *n.ImportJobID)
 		} else {
 			ctx.WriteKeyWord("IMPORT JOBS")
-			restoreShowLikeOrWhereOpt()
+			if err := restoreShowLikeOrWhereOpt(); err != nil {
+				return err
+			}
 		}
 	case ShowImportGroups:
 		if n.ShowGroupKey != "" {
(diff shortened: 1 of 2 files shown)


lance6716:
please fill PR template to let CI pass

https://github.com/pingcap/tidb/blob/master/.github/pull_request_template.md?plain=1

lance6716:
@jfrabaute-arista please update PR description as the format of PR template ⬇️

https://github.com/pingcap/tidb/blob/master/.github/pull_request_template.md?plain=1

And can you add a unit test for it?

jfrabaute-arista:
> please fill PR template to let CI pass
> 
> https://github.com/pingcap/tidb/blob/master/.github/pull_request_template.md?plain=1

I added a unit test and change the PR message with the Unit test checked.

PR #64204, opened 2025-10-31
planner: maintain a map of column indices by name in DataSource

henrybw:

Issue Number: ref #63856 

Problem Summary: Profiling the time spent in the optimizer repeatedly executing, for 2 minutes, a [query on a table like this](https://github.com/pingcap/tidb/issues/63856#issue-3496225000), but with many more indexes (500) and columns (6,500), reveals that 34% of the time now is spent in `expression.IndexCol2Col()`, due to this function being passed a slice of columns that the function has to search linearly with a for loop. Instead of scanning this slice, we should be using a map to reduce lookups of column by ID from O(N) to O(1).

IndexCol2Col_pprof

Similar to #64053, this is a preliminary PR that establishes the foundation for replacing the O(N) loop with an O(1) map lookup. The goal is to eventually add a `colIdxsByName map[string]int` parameter to `expression.IndexCol2Col()` that can be supplied by the logical data source.

* Add a new `model.GetColIdxMapping()` function that builds this map from a slice of `model.ColumnInfo` structs, along with a `TestColIdxMapping()` test case for this new function.
  * There are two reasons why this function is in `model`: (1) tests in the `util/ranger` package depend on `expression.IndexInfo2PrefixCols()` and will need to build this map too,  and (2) it is next to the definition of the `model.ColumnInfo` that it operates on.
* Add a `ColIdxsByName` map to the logical data source in the planner.
* Refactor all places in the planner that append columns to `ds.Columns` and `ds.Schema()` directly to instead call a new `AppendColumn()` function that also ensures the column is tracked by the `ds.ColIdxsByName` map.
  * The only exception is `PruneColumns()`, which has been updated to also remove the pruned column from `ds.ColIdxsByName` and update the affected column indices in the map.
* Add a new `TestDataSourceColumnMaps` test case that ensures that `AppendColumn()` maintains coherency between `ds.Columns` and `ds.ColIdxsByName`, and also ensures that `AppendTableCol()` maintains coherency between `ds.TblCols` and `ds.TblColsByID`.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/meta/model/column_test.go
+++ b/pkg/meta/model/column_test.go
@@ -105,7 +105,7 @@ func TestDefaultValue(t *testing.T) {
 	require.Equal(t, mysql.TypeLonglong, extraPhysTblIDCol.GetType())
 }
 
-func TestColIdxsByName(t *testing.T) {
+func TestColIdxMapping(t *testing.T) {
 	const numCols int64 = 10
 	colInfos := make([]*ColumnInfo, 0, numCols)
 	names := make([]string, 0, numCols)


pkg/meta/model/column.go
```diff
@@ -313,3 +313,13 @@ func NewExtraPhysTblIDColInfo() *ColumnInfo {
 	colInfo.SetCollate(charset.CollationBin)
 	return colInfo
 }
+
+// GetColIdxsByName builds a map of column names to their corresponding offsets
+// into the given slice of column infos.
+func GetColIdxsByName(colInfos []*ColumnInfo) map[string]int {
```
D3Hunter:
```suggestion
func GetColIdxMapping(colInfos []*ColumnInfo) map[string]int {
```

pkg/planner/core/operator/logicalop/logical_datasource.go
```diff
@@ -195,6 +198,10 @@ func (ds *DataSource) PruneColumns(parentUsedCols []*expression.Column) (base.Lo
 				expression.GcColumnExprIsTidbShard(ds.Schema().Columns[i].VirtualExpr) {
 				continue
 			}
+			delete(ds.ColIdxsByName, ds.Columns[i].Name.L)
+			for i := i + 1; i < len(ds.Columns); i++ {
+				ds.ColIdxsByName[ds.Columns[i].Name.L] = i - 1
```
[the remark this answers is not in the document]
henrybw: Fair enough.

terry1purcell (approved):

pkg/planner/core/operator/logicalop/logical_datasource.go
```diff
@@ -52,6 +53,8 @@ type DataSource struct {
 	Columns       []*model.ColumnInfo
 	DBName        ast.CIStr
 
+	ColIdxsByName map[string]int
```
winoros: I think we can add a TODO to bring it into `expression.Schema`?

elsa0520 (approved):
LGTM

pkg/planner/core/operator/logicalop/logical_datasource.go
winoros, line 56: You can use a map `map[int]int` to store the mapping: id -> offset.
And then use the `tableInfo.Columns[IndexColumn.Offset].ID` to get the needed information.

I think the `map[string]int` will be hard to extend for other usage.

winoros: This map is used to cooperate with `expression.Schema` and other related stuff, but it's defined outside of it.
It's easy to forget to update it when pruning columns and adding columns.

winoros: As an example, I don't see that we update it during column pruning.

winoros (changes requested):

winoros:
🤔But this function's usage is limited to one specific case.
It also sounds good to me if it's possible to build it when we use it.

pkg/planner/core/operator/physicalop/physical_table_scan.go
```diff
@@ -161,6 +164,7 @@ func GetPhysicalScan4LogicalTableScan(s *logicalop.LogicalTableScan, schema *exp
 	ts := PhysicalTableScan{
 		Table:           ds.TableInfo,
 		Columns:         ds.Columns,
+		ColIdxsByName:   ds.ColIdxsByName,
```
AilinKid, line 167: why this place didn't use the clone？

pkg/planner/core/operator/logicalop/logical_index_scan.go
```diff
@@ -42,6 +42,7 @@ type LogicalIndexScan struct {
 
 	Index          *model.IndexInfo
 	Columns        []*model.ColumnInfo
+	ColIdxsByName  map[string]int
```
AilinKid, line 45: i saw the adding of ColIdxsByName into the PhysicalTableScan and LogicalIndexScan, but any usage of them?

pkg/planner/core/operator/logicalop/logical_datasource.go
henrybw: > You can use a map `map[int]int` to store the mapping: id -> offset. And then use the `tableInfo.Columns[IndexColumn.Offset].ID` to get the needed information.

Hmm, doing this would require us to also pass the table info as an argument to `IndexInfo2PrefixCols()`/`IndexInfo2Cols()`...

I'm open to the idea, though. It might mean that we wouldn't need to maintain this mapping after all, which would make things a lot simpler.

I'll try your approach and see how it goes.

pkg/planner/core/operator/logicalop/logical_index_scan.go
henrybw: This would be used when calling `expression.IndexInfo2PrefixCols()`, which would come in a follow-up PR.

pkg/planner/core/operator/physicalop/physical_table_scan.go
henrybw: I only cloned in places where `ds.Columns` was also cloned, but I might have misunderstood when it's necessary to clone. I will revisit this.

pkg/planner/core/operator/logicalop/logical_datasource.go
henrybw: > As an example, I don't see that we update it during column pruning.

We should be also updating it during column pruning (see lines 201-204 added to `logical_datasource.go`), unless I missed a spot.

That said, I think you make a good point here:

> It's easy to forget to update it when pruning columns and adding columns.

henrybw:
> 🤔But this function's usage is limited to one specific case. It also sounds good to me if it's possible to build it when we use it.

The problem with building it only when we use it is that the cost of building the map outweighs the benefits of using it, if we only build it and use it once.

That said, your suggestion to instead use the table info for finding the correlated columns might be a better approach than trying to maintain a map in the data source.

terry1purcell:
/hold

Only holding to avoid an accidental merge (since already 2 positive reviews). You can remove the hold without prior authorization from me. Thank you

hawkingrei:
I think we can add some more Go benchmark data to verify the performance improvement here.

PR #64214, opened 2025-10-31
planner: optimize decorrelate by removing redundant LIMIT when join key is unique

### LINKED ISSUE #64200: Remove redundant LIMIT 1 to enable decorrelation of correlated scalar subqueries
## Enhancement
When a correlated scalar subquery satisfies the condition of a “unique point lookup with no row expansion,” it produces at most one row per outer row, making LIMIT 1 semantically redundant. Currently, the optimizer fails to decorrelate due to the presence of LIMIT, leaving the plan as Apply + Limit, which causes significant amplification and repeated probing when the outer relation has many rows.

```
CREATE TABLE IF NOT EXISTS employees (
    id INT PRIMARY KEY,
    name VARCHAR(50),
    dept_id INT,
    salary DECIMAL(10, 2)
);
```

```
mysql> explain SELECT      e.name,     e.salary,     (SELECT e2.salary       FROM employees e2       WHERE e2.dept_id = e.dept_id limit 1) AS avg_dept_salary FROM employees e WHERE e.dept_id = 1;
+--------------------------------+----------+-----------+---------------+------------------------------------------------------------------------------+
| id                             | estRows  | task      | access object | operator info                                                                |
+--------------------------------+----------+-----------+---------------+------------------------------------------------------------------------------+
| HashJoin_21                    | 80.00    | root      |               | CARTESIAN left outer join, left side:TableReader_25                          |
| ├─HashAgg_34(Build)            | 8.00     | root      |               | group by:test.employees.dept_id, funcs:avg(test.employees.salary)->Column#16 |
| │ └─TableReader_44             | 10.00    | root      |               | data:Selection_43                                                            |
| │   └─Selection_43             | 10.00    | cop[tikv] |               | eq(1, test.employees.dept_id)                                                |
| │     └─TableFullScan_42       | 10000.00 | cop[tikv] | table:e2      | keep order:false, stats:pseudo                                               |
| └─TableReader_25(Probe)        | 10.00    | root      |               | data:Selection_24                                                            |
|   └─Selection_24               | 10.00    | cop[tikv] |               | eq(test.employees.dept_id, 1)                                                |
|     └─TableFullScan_23         | 10000.00 | cop[tikv] | table:e       | keep order:false, stats:pseudo                                               |
+--------------------------------+----------+-----------+---------------+------------------------------------------------------------------------------+
8 rows in set (0.00 sec)
```

```
mysql> explain SELECT      e.name,     e.salary,     (SELECT e2.salary       FROM employees e2       WHERE e2.dept_id = e.dept_id limit 1) AS avg_dept_salary FROM employees e WHERE e.dept_id = 1;
+----------------------------------+----------+-----------+---------------+-------------------------------------------------------------------+
| id                               | estRows  | task      | access object | operator info                                                     |
+----------------------------------+----------+-----------+---------------+-------------------------------------------------------------------+
| Projection_14                    | 10.00    | root      |               | test.employees.name, test.employees.salary, test.employees.salary |
| └─Apply_18                       | 10.00    | root      |               | CARTESIAN left outer join, left side:TableReader_21               |
|   ├─TableReader_21(Build)        | 10.00    | root      |               | data:Selection_20                                                 |
|   │ └─Selection_20               | 10.00    | cop[tikv] |               | eq(test.employees.dept_id, 1)                                     |
|   │   └─TableFullScan_19         | 10000.00 | cop[tikv] | table:e       | keep order:false, stats:pseudo                                    |
|   └─Limit_22(Probe)              | 10.00    | root      |               | offset:0, count:1                                                 |
|     └─TableReader_29             | 10.00    | root      |               | data:Limit_28                                                     |
|       └─Limit_28                 | 10.00    | cop[tikv] |               | offset:0, count:1                                                 |
|         └─Selection_27           | 10.00    | cop[tikv] |               | eq(test.employees.dept_id, test.employees.dept_id)                |
|           └─TableFullScan_26     | 10000.00 | cop[tikv] | table:e2      | keep order:false, stats:pseudo                                    |
+----------------------------------+----------+-----------+---------------+-------------------------------------------------------------------+
10 rows in set (0.00 sec)
```

King-Dylan:

Issue Number: close #64200 

Problem Summary:
When decorrelating correlated subqueries, if the join key forms a unique key constraint, the LIMIT operator in the subquery becomes redundant since the join key guarantees at most one row. This optimization removes such redundant LIMIT operators and also eliminates redundant MaxOneRow wrappers to improve query performance.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive data-driven test suite for correlated-subquery + LIMIT/OFFSET scenarios, new test fixtures, and a new test exercising decorrelate-limit behavior. Also exposed accessors to retrieve those test datasets for external tests.
* **Refactor**
  * Improved planner decorrelation and LIMIT-handling with broader pattern coverage and join-key reasoning; may produce different, often more efficient EXPLAIN plan shapes for some queries.


pkg/planner/core/rule_decorrelate.go
```diff
@@ -246,6 +246,46 @@ func (s *DecorrelateSolver) optimize(ctx context.Context, p base.LogicalPlan, gr
...
+					if li.Offset != 0 {
+						canRemove = false
+					} else {
+						// Check if join key is unique key
+						removePlan = li.Children()[0]
+						if isJoinKeyUniqueKey(apply, removePlan) {
+							canRemove = true
+						}
+					}
+				} else if proj, ok := mChild.(*logicalop.LogicalProjection); ok {
```
Reminiscent: When will this happen? Can you provide some cases for this situations?

pkg/planner/core/rule_decorrelate.go
```diff
@@ -246,6 +246,46 @@ func (s *DecorrelateSolver) optimize(ctx context.Context, p base.LogicalPlan, gr
 			apply.SetChildren(outerPlan, innerPlan)
 			return s.optimize(ctx, p, groupByColumn)
 		} else if m, ok := innerPlan.(*logicalop.LogicalMaxOneRow); ok {
+			// Check if MaxOneRow's child is Limit or TopN, and if we can remove it for LeftOuterJoin
```
Reminiscent: It seems this PR doesn't handle the `TopN` case?

pkg/planner/core/rule_decorrelate.go
```diff
@@ -470,6 +510,133 @@ func (*DecorrelateSolver) Name() string {
...
+						args := sf.GetArgs()
+						if len(args) == 2 {
+							if innerCol, ok := args[1].(*expression.Column); ok {
+								if sel.Schema().Contains(innerCol) {
+									innerJoinKeys = append(innerJoinKeys, innerCol)
+								}
+							}
+						}
+					}
+				}
```
Reminiscent: Can we make this a function? It has been used several times

pkg/planner/core/rule_decorrelate.go
```diff
@@ -470,6 +510,133 @@ func (*DecorrelateSolver) Name() string {
...
+	var extractConditions func(base.LogicalPlan)
+	extractConditions = func(p base.LogicalPlan) {
+		if sel, ok := p.(*logicalop.LogicalSelection); ok {
+			// Check conditions directly on Selection
+			for _, cond := range sel.Conditions {
+				if decExpr := apply.DeCorColFromEqExpr(cond); decExpr != nil {
+					if sf, ok := decExpr.(*expression.ScalarFunction); ok && sf.FuncName.L == ast.EQ {
+						args := sf.GetArgs()
+						if len(args) == 2 {
+							if innerCol, ok := args[1].(*expression.Column); ok {
```
Reminiscent: Can we guarantee the `args[1]` rather than `args[0]` is the column from the inner side?

pkg/planner/core/rule_decorrelate.go
```diff
@@ -470,6 +510,133 @@ func (*DecorrelateSolver) Name() string {
...
+					found = true
+					break
+				}
+			}
+			if !found {
+				allMatch = false
+				break
+			}
+		}
+		if allMatch && len(keyInfo) == len(innerJoinKeys) && len(keyInfo) > 0 {
```
Reminiscent: It seems no need to check `len(keyInfo) == len(innerJoinKeys)`. For example, the unique key is `(a, b)`. And the filter columns contain `(a, b, c)`. The `(a, b)` can guarantee the uniqueness.

pkg/planner/core/rule_decorrelate.go
```diff
@@ -470,6 +510,133 @@ func (*DecorrelateSolver) Name() string {
 	return "decorrelate"
 }
 
+// isJoinKeyUniqueKey checks if join key is unique key.
+// Returns true if the join key forms a unique key constraint.
+func isJoinKeyUniqueKey(apply *logicalop.LogicalApply, plan base.LogicalPlan) bool {
+	var hasMultiRowOperator func(base.LogicalPlan) bool
```
Reminiscent, line 595: This function needs to guarantee contains all the cases which will generate more rows. If there lacks some cases, it will generate the wring answer. For example, please add some cases related to the `unnest` function, it will generate more rows? So here should be considered more seriously.

pkg/planner/core/rule_decorrelate.go
King-Dylan: DeCorColFromEqExpr preserves the order of args[1], which is on the inner side.

pkg/planner/core/rule_decorrelate.go
King-Dylan: Done

pkg/planner/core/rule_decorrelate.go
King-Dylan: At this stage, topn is still just a LIMIT, so it doesn’t matter.

pkg/planner/core/rule_decorrelate.go
King-Dylan: Done

pkg/planner/core/rule_decorrelate.go
King-Dylan: Yes, there may be mis-deletions down the road or in certain cases. But the NoDecorrelate hint lets us sidestep the issue, even if we miss maintaining the list when new funcs are introduced.

pkg/planner/core/rule_decorrelate.go
King-Dylan: Yes, good idea~

pkg/planner/core/rule_decorrelate.go
Reminiscent: My concern is that if the middle node is topN (limit + sort), it seems can not to decorrelate.  Because sort will change the meaning? Could you add a test for this?

pkg/planner/core/rule_decorrelate.go
```diff
@@ -246,6 +246,50 @@ func (s *DecorrelateSolver) optimize(ctx context.Context, p base.LogicalPlan, gr
 			apply.SetChildren(outerPlan, innerPlan)
 			return s.optimize(ctx, p, groupByColumn)
 		} else if m, ok := innerPlan.(*logicalop.LogicalMaxOneRow); ok {
+			// Check if MaxOneRow's child is Limit or TopN, and if we can remove it for LeftOuterJoin
+			// Also handle the case where there's a Projection between MaxOneRow and Limit: MaxOneRow -> Projection -> Limit
+			if apply.JoinType == base.LeftOuterJoin {
```
hawkingrei: Can we use switch to rewrite this code? it can make code readable.

pkg/planner/core/rule_decorrelate.go
```diff
@@ -470,6 +514,133 @@ func (*DecorrelateSolver) Name() string {
...
+	}
+
+	extractConditions(plan)
+	if len(innerJoinKeys) == 0 {
+		return false
+	}
+
+	// Find the underlying DataSource to get PKOrUK
+	var findDataSource func(base.LogicalPlan) *logicalop.DataSource
+	findDataSource = func(p base.LogicalPlan) *logicalop.DataSource {
```
hawkingrei: Can we make it a unique function? Don't use an anonymous function.

pkg/planner/core/rule_decorrelate.go
King-Dylan: Since the single-row result is determined by the join key, its final result isn’t affected by any ordering. I’ve already added an ORDER BY test case in `TestDecorrelateLimitOptimization`.

pkg/planner/core/rule_decorrelate.go
King-Dylan: Ok

pkg/planner/core/rule_decorrelate.go
King-Dylan: ok

pkg/planner/core/rule_decorrelate.go
```diff
@@ -234,218 +234,267 @@ func (s *DecorrelateSolver) optimize(ctx context.Context, p base.LogicalPlan, gr
...
+					case *logicalop.LogicalProjection:
+						// Check if Projection's child is Limit: MaxOneRow -> Projection -> Limit
+						// This pattern occurs when subqueries contain some clauses like ORDER BY or HAVING clauses that require projection.
+						// Examples:
+						//   - HAVING clause: SELECT ... (SELECT AVG(...) FROM ... GROUP BY ... HAVING ... LIMIT 1)
+						//   - ORDER BY clause: SELECT ... (SELECT ... FROM ... ORDER BY ... LIMIT 1)
+						if li, ok := mChild.Children()[0].(*logicalop.LogicalLimit); ok {
+							// Limit with non-0 offset cannot be removed, but we still check for redundant MaxOneRow
+							if li.Offset != 0 {
+								canRemove = false
```
winoros: We are in `switch`, so we can break here.

pkg/planner/core/rule_decorrelate.go
```diff
@@ -470,6 +519,160 @@ func (*DecorrelateSolver) Name() string {
...
+		return true
+	case *logicalop.LogicalExpand:
+		// Expand operator splits rows for GROUPING SETS/ROLLUP
+		return true
+	}
+	return false
+}
+
+// isJoinKeyUniqueKey checks if join key is unique key.
+// Returns true if the join key forms a unique key constraint.
```
winoros: Add a TODO that this should be replaced by the functional dependency.

pkg/planner/core/rule_decorrelate.go
```diff
@@ -470,6 +519,160 @@ func (*DecorrelateSolver) Name() string {
...
+func isJoinKeyUniqueKey(apply *logicalop.LogicalApply, plan base.LogicalPlan) bool {
```
winoros: We'll reach an issue for this function: why is it not covered by the `BuildKeyInfo`?
Current `BuildKeyInfo` plays this role.

King-Dylan: When the join condition gets pushed down to the DataSource's PushedDownConds (which can happen during optimization), apply.EqualConditions becomes empty at that point. BuildKeyInfo derives uniqueness from EqualConditions, so when that list is empty it has no way to know the join key is a PK/UK — and the optimization would be incorrectly skipped.
isJoinKeyUniqueKey handles this by directly walking down the plan tree and checking both LogicalSelection.Conditions and DataSource.PushedDownConds to reconstruct the join keys, then verifying them against ds.Schema().PKOrUK.

pkg/planner/core/main_test.go
```diff
@@ -60,3 +61,11 @@ func TestMain(m *testing.M) {
 func GetRuntimeFilterGeneratorData() testdata.TestData {
 	return testDataMap["runtime_filter_generator_suite"]
 }
+
+func GetDecorrelateLimitSuiteData() testdata.TestData {
+	return testDataMap["decorrelate_limit_suite"]
+}
+
+func GetCascadesSuiteData() testdata.TestData {
```
D3Hunter: #### ⚠️ **[Major]** Uninitialized exported testdata accessor expands surface without a defined contract

**Why**
`GetCascadesSuiteData` introduces a new exported helper in the core test package, but the corresponding suite key is not initialized in `TestMain`, so the API intent is unclear and easy to misuse.

**Scope**
pkg/planner/core/main_test.go:69

**Risk if unchanged**
Future tests can call this accessor assuming it is ready and hit confusing runtime failures from missing suite data, while the unnecessary exported helper increases maintenance surface.

**Evidence**
`TestMain` loads `plan_suite_unexported`, `runtime_filter_generator_suite`, and `decorrelate_limit_suite` (pkg/planner/core/main_test.go:34-37), but `GetCascadesSuiteData` returns `testDataMap["cascades_suite"]` (pkg/planner/core/main_test.go:69-70) with no matching initialization or new caller in this diff.

**Change request**
seems no need to add this surface right now; can we remove it, or load `cascades_suite` in `TestMain` and add a caller that documents intended usage?

pkg/planner/core/rule_decorrelate.go
```diff
@@ -477,6 +532,160 @@ func (*DecorrelateSolver) Name() string {
...
+		return nil
+	}
+	innerCol, ok := args[1].(*expression.Column)
+	if !ok || !schema.Contains(innerCol) {
+		return nil
+	}
+	return innerCol
+}
+
+// findDataSource recursively finds the underlying DataSource in a logical plan tree.
```
D3Hunter: #### ⚠️ **[Major]** Unique-key guard for LIMIT removal can match the wrong table and remove MaxOneRow unsafely

**Why**
The new uniqueness check can report true even when the extracted correlated key does not belong to the same DataSource used for key validation.

**Scope**
pkg/planner/core/rule_decorrelate.go:558, pkg/planner/core/rule_decorrelate.go:580, pkg/planner/core/rule_decorrelate.go:665

**Risk if unchanged**
Decorrelate may drop `MaxOneRow`/`LIMIT` on scalar subqueries that are not actually single-row per outer row, which can change query results or suppress `subquery returns more than 1 row` behavior.

**Evidence**
`findDataSource` returns the first DataSource in the subtree, `hasMultiRowOperator` does not treat `LogicalApply` as multi-row-unsafe, and key matching uses `keyCol.ID == joinKey.ID` instead of column identity (`UniqueID`). In nested subquery/apply trees, different tables commonly reuse column IDs (for example, ID=1), so a key from another table can satisfy this check and trigger `canRemove=true`.

**Change request**
Can we tighten this guard: use `UniqueID`-based matching, require join keys to come from the same DataSource selected for PK/UK validation, and fail closed when the subtree contains `LogicalApply` (or any operator not explicitly proven row-preserving)? please add a case for nested apply/scalar-subquery under this path.

pkg/planner/core/testdata/decorrelate_limit_suite_in.json
```diff
@@ -0,0 +1,22 @@
...
+      "EXPLAIN format = 'plan_tree' SELECT e.name, e.salary, (SELECT count(e2.dept_id)  FROM employees e2  WHERE e2.id = e.id limit 1) AS distinct_dept_id FROM employees e",
+      "EXPLAIN format = 'plan_tree' SELECT e.name, e.salary,(SELECT e2.salary FROM employees e2 WHERE e2.id = e.id AND e2.dept_id IN (     SELECT dept_id     FROM employees e3     WHERE e3.id = e.id     LIMIT 1 ) LIMIT 1) AS salary_nested FROM employees e",
+      "EXPLAIN format = 'plan_tree' SELECT e.name, e.salary,(SELECT e2.salary FROM employees e2 WHERE e2.id = e.id AND EXISTS (     SELECT 1     FROM employee_notes en     WHERE en.employee_id = e2.id ) LIMIT 1) AS salary_exists FROM employees e",
+      "EXPLAIN format = 'plan_tree' SELECT e.name, e.salary, (SELECT e2.salary  FROM employees e2  WHERE e2.id = e.id  ORDER BY e2.dept_id, e2.salary DESC  LIMIT 1) AS salary_order_multi FROM employees e",
+      "EXPLAIN format = 'plan_tree' SELECT e.name, e.salary,  (SELECT DISTINCT e2.dept_id   FROM employees e2   WHERE e2.id = e.id   LIMIT 1) AS distinct_dept_id FROM employees e",
```
D3Hunter, line 16: #### ⚠️ **[Major]** Duplicate DISTINCT subquery fixture appears twice in one suite

**Why**
The same SQL scenario is defined twice with whitespace-only differences, so one case slot does not expand behavior coverage.

**Scope**
pkg/planner/core/testdata/decorrelate_limit_suite_in.json:16; pkg/planner/core/testdata/decorrelate_limit_suite_in.json:18; pkg/planner/core/testdata/decorrelate_limit_suite_out.json:163; pkg/planner/core/testdata/decorrelate_limit_suite_out.json:188; pkg/planner/core/testdata/decorrelate_limit_suite_xut.json:163; pkg/planner/core/testdata/decorrelate_limit_suite_xut.json:188

**Risk if unchanged**
Coverage blind spots remain because the suite appears to have more unique scenarios than it actually does; regressions in missing edge cases can pass unnoticed.

**Evidence**
After whitespace/case normalization, the two DISTINCT subquery entries are equivalent, and the duplicated SQL appears in both recorded output files as well.

**Change request**
this looks duplicated with the other DISTINCT case; can we keep one canonical definition and replace the duplicate with a missing edge scenario?

pkg/planner/core/rule_decorrelate.go
```diff
@@ -234,225 +234,280 @@ func (s *DecorrelateSolver) optimize(ctx context.Context, p base.LogicalPlan, gr
...
-		} else if proj, ok := innerPlan.(*logicalop.LogicalProjection); ok {
-			// After the column pruning, some expressions in the projection operator may be pruned.
-			// In this situation, we can decorrelate the apply operator.
-			if apply.JoinType == base.LeftOuterJoin {
-				if skipDecorrelateProjectionForLeftOuterApply(apply, proj) {
-					goto NoOptimize
+			case *logicalop.LogicalMaxOneRow:
+				// Check if MaxOneRow's child is Limit or TopN, and if we can remove it for LeftOuterJoin
+				// Also handle the case where there's a Projection between MaxOneRow and Limit: MaxOneRow -> Projection -> Limit
+				if apply.JoinType == base.LeftOuterJoin {
```
D3Hunter, line 259: #### ⚠️ **[Major]** LIMIT-removal rewrite is not covered by result-level regression tests

**Why**
The new decorrelation path removes `MaxOneRow`/`Limit` when `isJoinKeyUniqueKey` returns true, which is a behavior-changing rewrite for scalar subqueries. Current coverage validates only `EXPLAIN` plan shapes, so semantic compatibility regressions can pass undetected.

**Scope**
pkg/planner/core/rule_decorrelate.go:253; pkg/planner/core/rule_decorrelate.go:599; pkg/planner/core/plan_test.go:737; pkg/planner/core/testdata/decorrelate_limit_suite_in.json:5

**Risk if unchanged**
A wrong-result edge case can reach production during rolling upgrades because upgraded nodes apply the new rewrite while older nodes do not, creating version-dependent query behavior without test evidence on returned rows.

**Evidence**
The rewrite logic is introduced in the new `LogicalMaxOneRow` branch and helper (`isJoinKeyUniqueKey`), while `TestDecorrelateLimitOptimization` executes only `EXPLAIN format='plan_tree'` statements and checks recorded plans (`plan.Check(...)`) rather than query result rows.

**Change request**
add UT for it: keep the plan-shape suite, and add result-check cases that execute data queries for nullable unique keys, composite unique keys, and a non-unique-key negative path to prove LIMIT is preserved when required.

pkg/planner/core/rule_decorrelate.go
```diff
@@ -234,225 +234,280 @@ func (s *DecorrelateSolver) optimize(ctx context.Context, p base.LogicalPlan, gr
...
 				return s.optimize(ctx, p, groupByColumn)
-			}
-		} else if proj, ok := innerPlan.(*logicalop.LogicalProjection); ok {
-			// After the column pruning, some expressions in the projection operator may be pruned.
-			// In this situation, we can decorrelate the apply operator.
-			if apply.JoinType == base.LeftOuterJoin {
-				if skipDecorrelateProjectionForLeftOuterApply(apply, proj) {
-					goto NoOptimize
+			case *logicalop.LogicalMaxOneRow:
+				// Check if MaxOneRow's child is Limit or TopN, and if we can remove it for LeftOuterJoin
```
D3Hunter: #### 🟡 **[Minor]** MaxOneRow comment promises TopN handling that the code path does not implement

**Why**
The branch comment claims the decorrelate logic checks both `Limit` and `TopN`, but the actual type switch only handles `LogicalLimit` directly and `LogicalProjection` with a `LogicalLimit` child.

**Scope**
pkg/planner/core/rule_decorrelate.go:251

**Risk if unchanged**
Reviewers and future maintainers may assume TopN safety rules are already covered and skip adding required semantic checks when extending this optimizer path.

**Evidence**
Comment text at pkg/planner/core/rule_decorrelate.go:251 says "Limit or TopN", while the cases implemented in this block are `*logicalop.LogicalLimit` and `*logicalop.LogicalProjection` (pkg/planner/core/rule_decorrelate.go:258 and pkg/planner/core/rule_decorrelate.go:270).

**Change request**
please add a short why-comment that matches actual behavior: either implement a `LogicalTopN` branch with explicit constraints, or remove the TopN claim from the comment.

pkg/planner/core/rule_decorrelate.go
```diff
@@ -477,6 +532,160 @@ func (*DecorrelateSolver) Name() string {
...
+		// UNION ALL combines multiple inputs, potentially generating multiple rows
+		return true
+	case *logicalop.LogicalExpand:
+		// Expand operator splits rows for GROUPING SETS/ROLLUP
+		return true
+	}
+	return false
+}
+
+// isJoinKeyUniqueKey checks if join key is unique key.
```
D3Hunter: #### 🟡 **[Minor]** Uniqueness helper name overstates the proof it performs

**Why**
`isJoinKeyUniqueKey` reads like a general uniqueness proof, but the implementation is intentionally narrow: it only extracts equality predicates from `Selection` and `DataSource.PushedDownConds`, then checks PK/UK on the first discovered `DataSource`.

**Scope**
pkg/planner/core/rule_decorrelate.go:597

**Risk if unchanged**
A broad name can lead new callsites to over-trust the helper and apply decorrelation transforms where uniqueness was not actually established by this limited scan.

**Evidence**
Predicate extraction is limited to `LogicalSelection` and `DataSource.PushedDownConds` (pkg/planner/core/rule_decorrelate.go:624-639), and datasource resolution uses `findDataSource` first-match traversal (pkg/planner/core/rule_decorrelate.go:558-566, pkg/planner/core/rule_decorrelate.go:655).

**Change request**
prefer a more specific name; current name is confusing. Rename to reflect its scoped behavior and add a short why-comment on first-DataSource/condition-source assumptions.

pkg/planner/core/main_test.go
```diff
@@ -33,6 +33,7 @@ func TestMain(m *testing.M) {
 	flag.Parse()
 	testDataMap.LoadTestSuiteData("testdata", "plan_suite_unexported")
```
D3Hunter, line 34: #### 🟡 **[Minor]** Unused cascades suite accessor is added without corresponding suite load

**Why**
A new `GetCascadesSuiteData` accessor is introduced in this package, but `TestMain` does not load `cascades_suite`, and no new local callsite uses the accessor.

**Scope**
pkg/planner/core/main_test.go:34; pkg/planner/core/main_test.go:69

**Risk if unchanged**
This increases scope noise in a behavior PR and leaves a latent nil/empty retrieval path if the accessor is used later without noticing initialization is missing.

**Evidence**
`TestMain` loads `plan_suite_unexported`, `runtime_filter_generator_suite`, and `decorrelate_limit_suite`; the new accessor returns `testDataMap["cascades_suite"]`, which is not loaded in this file.

**Change request**
seems no need to add this accessor in this change; can we remove it, or load `cascades_suite` and add its immediate caller if this is required now?

pkg/planner/core/rule_decorrelate.go
```diff
@@ -477,6 +532,160 @@ func (*DecorrelateSolver) Name() string {
...
+	}
+	for _, child := range p.Children() {
+		if ds := findDataSource(child); ds != nil {
+			return ds
+		}
+	}
+	return nil
+}
+
+// canGenerateMultipleRows returns if the LogicalPlan can generate multiple rows from a single input row.
```
D3Hunter: #### 🟡 **[Minor]** Helper contract is conservative but named as an exact predicate

**Why**
`canGenerateMultipleRows` is documented like an exact capability check, but it intentionally returns true for all joins, including semi/anti joins called out as row-preserving.

**Scope**
pkg/planner/core/rule_decorrelate.go:572; pkg/planner/core/rule_decorrelate.go:582

**Risk if unchanged**
Future callers can assume exact semantics from the name and comments, then unknowingly gate unrelated optimization paths through this conservative over-approximation.

**Evidence**
The checklist comment says joins except semi/anti, while the implementation unconditionally returns true for `*logicalop.LogicalJoin` with a safety caveat.

**Change request**
why not use a clearer contract: either rename this helper to a conservative form (for example `mayGenerateMultipleRows`) or refine the join branch to match the current name?

pkg/planner/core/testdata/decorrelate_limit_suite_in.json
D3Hunter, line 16: #### 🟡 **[Minor]** Duplicate DISTINCT case reduces edge-case coverage and isolation

**Why**
The same `SELECT DISTINCT ... LIMIT 1` scenario is recorded twice, which consumes one slot in the new suite without expanding compatibility coverage.

**Scope**
pkg/planner/core/testdata/decorrelate_limit_suite_in.json:16; pkg/planner/core/testdata/decorrelate_limit_suite_in.json:18; pkg/planner/core/testdata/decorrelate_limit_suite_out.json:163; pkg/planner/core/testdata/decorrelate_limit_suite_out.json:188

**Risk if unchanged**
Coverage remains narrower than intended for a behavior-changing optimizer rule, increasing the chance that an untested branch regresses while all suite entries still pass.

**Evidence**
Two input SQL entries are semantically identical DISTINCT+LIMIT correlated subqueries, and both expected outputs are the same `Apply -> HashAgg` plan shape.

**Change request**
add a case for a missing negative/edge scenario by replacing one duplicate entry (for example nullable-UK behavior or projection-plus-limit path with non-unique predicates).

D3Hunter (commented):
**AI-generated review based on @D3Hunter's standards; manual review will be performed after all comments are resolved.**

### Summary

- Total findings: 11
- Inline comments: 11
- Summary-only findings (no inline anchor): 0

### Findings (highest risk first)

#### ⚠️ **[Major]** (4)

1. Uninitialized exported testdata accessor expands surface without a defined contract (`pkg/planner/core/main_test.go:69`)
2. Unique-key guard for LIMIT removal can match the wrong table and remove MaxOneRow unsafely (`pkg/planner/core/rule_decorrelate.go:558, pkg/planner/core/rule_decorrelate.go:580, pkg/planner/core/rule_decorrelate.go:665`)
3. Duplicate DISTINCT subquery fixture appears twice in one suite (`pkg/planner/core/testdata/decorrelate_limit_suite_in.json:16; pkg/planner/core/testdata/decorrelate_limit_suite_in.json:18; pkg/planner/core/testdata/decorrelate_limit_suite_out.json:163; pkg/planner/core/testdata/decorrelate_limit_suite_out.json:188; pkg/planner/core/testdata/decorrelate_limit_suite_xut.json:163; pkg/planner/core/testdata/decorrelate_limit_suite_xut.json:188`)
4. LIMIT-removal rewrite is not covered by result-level regression tests (`pkg/planner/core/rule_decorrelate.go:253; pkg/planner/core/rule_decorrelate.go:599; pkg/planner/core/plan_test.go:737; pkg/planner/core/testdata/decorrelate_limit_suite_in.json:5`)

#### 🟡 **[Minor]** (5)

1. MaxOneRow comment promises TopN handling that the code path does not implement (`pkg/planner/core/rule_decorrelate.go:251`)
2. Uniqueness helper name overstates the proof it performs (`pkg/planner/core/rule_decorrelate.go:597`)
3. Unused cascades suite accessor is added without corresponding suite load (`pkg/planner/core/main_test.go:34; pkg/planner/core/main_test.go:69`)
4. Helper contract is conservative but named as an exact predicate (`pkg/planner/core/rule_decorrelate.go:572; pkg/planner/core/rule_decorrelate.go:582`)
5. Duplicate DISTINCT case reduces edge-case coverage and isolation (`pkg/planner/core/testdata/decorrelate_limit_suite_in.json:16; pkg/planner/core/testdata/decorrelate_limit_suite_in.json:18; pkg/planner/core/testdata/decorrelate_limit_suite_out.json:163; pkg/planner/core/testdata/decorrelate_limit_suite_out.json:188`)

#### 🧹 **[Nit]** (2)

1. Several new test SQL aliases are semantically misleading (`pkg/planner/core/testdata/decorrelate_limit_suite_in.json:5`)
2. `GetCascadesSuiteData` accessor is added without loading its suite (`pkg/planner/core/main_test.go:34; pkg/planner/core/main_test.go:69`)

pkg/planner/core/testdata/decorrelate_limit_suite_in.json
```diff
@@ -0,0 +1,22 @@
+[
+  {
+    "Name": "TestDecorrelateLimitOptimization",
+    "Cases": [
+      "EXPLAIN format = 'plan_tree'  SELECT e.name, e.salary , ( select salary FROM employees e2 WHERE e2.id = e.id LIMIT 1 OFFSET 0 ) AS avg_dept_salary FROM employees e WHERE e.dept_id > 1",
```
D3Hunter, line 5: #### 🧹 **[Nit]** Several new test SQL aliases are semantically misleading

**Why**
Alias names in the new decorrelate limit suite do not consistently reflect the selected expression semantics.

**Scope**
pkg/planner/core/testdata/decorrelate_limit_suite_in.json:5

**Risk if unchanged**
Misleading aliases reduce intent readability in regression cases and make plan-diff triage harder.

**Evidence**
Examples include plain `salary` expressions labeled `avg_dept_salary` (lines 5-7) and `count(e2.dept_id)` labeled `distinct_dept_id` (line 12).

**Change request**
prefer more specific names; current aliases are confusing. Use aliases that describe actual expressions (for example `salary_by_id`, `dept_count`).

pkg/planner/core/main_test.go
D3Hunter, line 34: #### 🧹 **[Nit]** `GetCascadesSuiteData` accessor is added without loading its suite

**Why**
The helper returns `testDataMap["cascades_suite"]`, but this suite is not loaded in `TestMain`, which leaves a latent test reliability trap if the accessor is used later.

**Scope**
pkg/planner/core/main_test.go:34; pkg/planner/core/main_test.go:69

**Risk if unchanged**
Future tests may fail with confusing empty-suite behavior depending on call path, reducing determinism and debuggability of test setup.

**Evidence**
`TestMain` loads `plan_suite_unexported`, `runtime_filter_generator_suite`, and `decorrelate_limit_suite`; it does not load `cascades_suite` before exposing `GetCascadesSuiteData`.

**Change request**
do we still need this? Either load `cascades_suite` in `TestMain` or remove the accessor to keep test initialization explicit.

PR #64215, opened 2025-10-31
ranger: fix panic When the condition contains foldable expressions in prepare

### LINKED ISSUE #63914: Unexpected error `interface conversion: expression.Expression is *expression.Constant, not *expression.ScalarFunction`
## Bug Report

Hi,

The following test case triggers an unexpected error:
```
CREATE TABLE t0(c0 DOUBLE );
CREATE TABLE t1 LIKE t0;
CREATE INDEX i0 ON t1(c0 ASC);
ALTER TABLE t1  CHANGE c0 c0 DOUBLE NOT NULL ;
SET @a = 0.7481117056976476;
SET @b = NULL;
SET @c = NULL;
PREPARE prepare_query FROM 'SELECT t0.c0 FROM t0, t1 WHERE ((?)OR(((((?)<=>(t1.c0)))AND(((?)<=>(t1.c0))))))';
EXECUTE prepare_query USING @a,@b,@c; -- interface conversion: expression.Expression is *expression.Constant, not *expression.ScalarFunction
```

Furthermore, after this error, the connection will be lost. For example, if there are statements after the EXECUTE statement, will trigger `Lost connection to MySQL server during query`

No error.

```
interface conversion: expression.Expression is *expression.Constant, not *expression.ScalarFunction
```

Release Version: v9.0.0-beta.2.pre-196-g312226f869\nEdition: Community\nGit Commit Hash: 312226f869d56922135e1c177cac6c90319c3134\nGit Branch: HEAD\nUTC Build Time: 2025-09-21 08:27:14\nGoVersion: go1.23.11\nRace Enabled: false\nCheck Table Before Drop: false\nStore: unistore\nKernel Type: Classic

hawkingrei:

Issue Number: close #63914 

Problem Summary:

We usually assume that the expression entering the build range is the simplest form, but that's not always the case.
```
SELECT t0.c0 FROM t0, t1 WHERE ? OR ((? <=> t1.c0) AND (? <=> t1.c0))
SET @a = 0;
SET @b = NULL;
SET @c = NULL;
```

In this case, predicate simplification will not continue during the prepare phase. Therefore, when it comes to the execute phase, if the `((null <=> t1.c0) AND (null <=> t1.c0))` becomes a constant by ```ExtractEqAndInCondition``` and enters the build range, it will cause a panic.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable prepared-statement plan-cache behavior when queries include folded constants or NULLs; avoids incorrect ranges and improves EXPLAIN stability.

* **Tests**
  * Added extensive tests covering constant/NULL/parameter scenarios and system-var interactions; note: some test declarations were duplicated and need cleanup.

* **Documentation**
  * Added planner note describing access-condition normalization and guidance for validation.


pkg/util/ranger/detacher.go
```diff
@@ -415,9 +415,18 @@ func (d *rangeDetacher) detachCNFCondAndBuildRangeForIndex(conditions []expressi
...
+		switch sc := accessConds[eqCount].(type) {
+		case *expression.ScalarFunction:
+			if sc.FuncName.L != ast.EQ {
+				break EQCOUNTLOOP
+			}
+		case *expression.Constant:
+			// PREPARE prepare_query FROM 'SELECT t0.c0 FROM t0, t1 WHERE ? OR ((? <=> t1.c0) AND (? <=> t1.c0))';
+			// In this case, predicate simplification will not continue during the prepare phase.
+			// Therefore, when it comes to the execute phase, the first placeholder becomes a constant.
+			break EQCOUNTLOOP
```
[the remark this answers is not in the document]
hawkingrei: We should keep original behavior here.

pkg/util/ranger/detacher.go
```diff
@@ -415,9 +415,18 @@ func (d *rangeDetacher) detachCNFCondAndBuildRangeForIndex(conditions []expressi
...
+EQCOUNTLOOP:
 	for ; eqCount < len(accessConds); eqCount++ {
-		if accessConds[eqCount].(*expression.ScalarFunction).FuncName.L != ast.EQ {
-			break
+		switch sc := accessConds[eqCount].(type) {
+		case *expression.ScalarFunction:
+			if sc.FuncName.L != ast.EQ {
+				break EQCOUNTLOOP
+			}
+		case *expression.Constant:
```
qw4990: I'm afraid this could break some implicit assumption, could we just return some errors and let the Plan Cache to generate a new plan for this query, instead of reusing the old cached plan?

pkg/util/ranger/detacher.go
```diff
@@ -415,9 +415,18 @@ func (d *rangeDetacher) detachCNFCondAndBuildRangeForIndex(conditions []expressi
...
+			// PREPARE prepare_query FROM 'SELECT t0.c0 FROM t0, t1 WHERE ? OR ((? <=> t1.c0) AND (? <=> t1.c0))';
+			// In this case, predicate simplification will not continue during the prepare phase.
+			// Therefore, when it comes to the execute phase, the first placeholder becomes a constant.
+			continue
```
qw4990: Should we break or return an error here instead of continue? Since we are counting EQ in this loop, but actually this is a `Constant`, not a EQ.

time-and-fate (commented):
Please fix the failed case. There is a plan regression.
Seems the maintenance of `eqOrInCount` is incorrect.

hawkingrei:
> Please fix the failed case. There is a plan regression. Seems the maintenance of `eqOrInCount` is incorrect.

I've changed my approach to fix it. However, there's actually another issue: if we encounter an expr containing a defer function, we skip simplification, which can actually cause additional problems. We can make a note of this as a TODO.

pkg/util/ranger/detacher.go
hawkingrei: I've changed my approach to fix it. PTAL.

pkg/util/ranger/detacher.go
hawkingrei: I've changed my approach to fix it. PTAL.

PR #64224, opened 2025-10-31
ddl: Support Hybrid index creation and data import | tidb-test=ef726505f262e22c6ec9c8951273bb5ee20abefb

OliverS929:

Issue Number: ref https://github.com/pingcap/tidb/issues/64409

Problem Summary:

Adding a new type of index - hybrid index - into tidb that works on tici. Support hybrid index creation/drop and Import index data onto tici during Import Into.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/meta/model/index.go
+++ b/pkg/meta/model/index.go
@@ -228,8 +228,8 @@ type HybridInvertedSpec struct {
 // HybridSortSpec describes the order definition of the hybrid index.
 type HybridSortSpec struct {
 	Columns []*IndexColumn `json:"columns,omitempty"`
-	// Order stores true for ascending order and false for descending order.
-	Order []bool `json:"order,omitempty"`
+	// IsAsc stores, for each column, whether it is sorted in ascending order (true) or descending order (false).
+	IsAsc []bool `json:"is_asc,omitempty"`
 }
 
 // Clone clones HybridIndexInfo.
@@ -380,8 +380,8 @@ func (opt *HybridSortSpec) Clone() *HybridSortSpec {
 	}
 	cloned := &HybridSortSpec{}
 	cloned.Columns = cloneIndexColumnSlice(opt.Columns)
-	if len(opt.Order) > 0 {
-		cloned.Order = append([]bool(nil), opt.Order...)
+	if len(opt.IsAsc) > 0 {
+		cloned.IsAsc = append([]bool(nil), opt.IsAsc...)
 	}
 	return cloned
 }
--- a/pkg/ddl/index.go
+++ b/pkg/ddl/index.go
@@ -1103,7 +1103,7 @@ func buildHybridSortSpec(spec *hybridSortSpec, resolveColumn func(string) (*mode
 	}
 	return &model.HybridSortSpec{
 		Columns: columns,
-		Order:   normalized,
+		IsAsc:   normalized,
 	}, nil
 }
 
--- a/pkg/executor/show.go
+++ b/pkg/executor/show.go
@@ -2990,8 +2990,8 @@ func hybridParameterJSONForShow(tableInfo *model.TableInfo, idxInfo *model.Index
 		if len(columns) > 0 {
 			orders := make([]string, len(columns))
 			for i := range columns {
-				if i < len(clone.Sort.Order) {
-					if clone.Sort.Order[i] {
+				if i < len(clone.Sort.IsAsc) {
+					if clone.Sort.IsAsc[i] {
 						orders[i] = "asc"
 					} else {
 						orders[i] = "desc"


pkg/meta/model/index.go
```diff
@@ -172,6 +173,260 @@ type FullTextIndexInfo struct {
...
+type HybridIndexInfo struct {
+	FullText []*HybridFullTextSpec `json:"fulltext,omitempty"`
+	Vector   []*HybridVectorSpec   `json:"vector,omitempty"`
+	Inverted []*HybridInvertedSpec `json:"inverted,omitempty"`
+	Sort     *HybridSortSpec       `json:"sort,omitempty"`
+}
+
+// HybridFullTextSpec describes the configuration for a fulltext segment in a hybrid index.
+type HybridFullTextSpec struct {
+	Columns   []string                 `json:"columns"`
```
winoros: If reusing `model.IndexColumn` is not acceptable.
Use an int slice to represent `IndexColumn.Offset` may be better than storing the name here.

pkg/meta/model/index.go
```diff
@@ -172,6 +173,260 @@ type FullTextIndexInfo struct {
 	// TODO: Add other options
 }
 
+// HybridIndexInfo is the information of HYBRID index of a column.
+type HybridIndexInfo struct {
+	FullText []*HybridFullTextSpec `json:"fulltext,omitempty"`
+	Vector   []*HybridVectorSpec   `json:"vector,omitempty"`
+	Inverted []*HybridInvertedSpec `json:"inverted,omitempty"`
+	Sort     *HybridSortSpec       `json:"sort,omitempty"`
+}
```
winoros, line 182: I'm thinking of the way that we add a boolean field `OnlyFullTextCols` to represent whether this index is created by `create fulltext index ...`.
Then we will only need one IndexInfo for two types of tici indexes.

The `model.IndexInfo` will only need one `HybridIndexInfo` field. Don't need to store the `FullTextIndexInfo` anymore.

pkg/meta/model/index.go
```diff
@@ -172,6 +173,260 @@ type FullTextIndexInfo struct {
...
+// HybridInvertedSpec describes the configuration for an inverted segment in a hybrid index.
+type HybridInvertedSpec struct {
+	Columns []string       `json:"columns,omitempty"`
+	Params  map[string]any `json:"params,omitempty"`
+}
+
+// HybridSortSpec describes the order definition of the hybrid index.
+type HybridSortSpec struct {
+	Columns    []string `json:"columns,omitempty"`
+	Directions []string `json:"directions,omitempty"`
```
winoros: Can we simplify this to a boolean field?
I think the order can be either `asc` or `desc`, there's no third option.

pkg/meta/model/index.go
OliverS929: Using the offset slice was actually my initial intention before our first design discussion. However, during the meeting, I was explicitly advised to use the name slice instead, which was also the approach finalized in the last draft of the design document.

pkg/meta/model/index.go
OliverS929: Yeah, I assume we might want to revisit later whether we still want the full-text index to remain an independent implementation or be integrated into the hybrid index. However, I believe that would be a later topic for @joey-yez and @kolafish to decide on in terms of approach and product design, which is beyond the scope of this PR. Let’s hold that thought until then.

pkg/meta/model/index.go
OliverS929: Yeah, that could potentially be the case, but I’m not entirely sure about it. I’m following the design that everyone reviewed and agreed upon. Let’s hold this for a future improvement after we double-check with the rest of the project team.

pkg/meta/model/index.go
winoros: `Name` is difficult to use in the optimizer.
`ID` or `OffSet` is easier to use.

pkg/meta/model/index.go
OliverS929: Changed to boolean field in [03b30cf](https://github.com/pingcap/tidb/pull/64224/commits/03b30cf89990cb30877e7ab6967b69b246177e13).

pkg/meta/model/index.go
OliverS929: Changed in [03b30cf](https://github.com/pingcap/tidb/pull/64224/commits/03b30cf89990cb30877e7ab6967b69b246177e13).

winoros (approved):

JinheLin (approved):

wjhuang2016 (approved):

Issue #63456, opened 2025-09-11
Left outer joining pruning fails with subqueries in select list


terry1purcell:
## Enhancement

CREATE TABLE `t1` (
  `a` int,
  `b` int,
  `c` int,
  Primary KEY `ia` (`a`),
  KEY `ib` (`b`)
);

This query can prune the right table - since no columns are needed from the right table (t1b), and no duplicates are introduced by the join.
Explain `select t1a.* from t1 t1a left join t1 t1b on t1a.a = t1b.a;`

This next query will not prune the right table - since the code does not detect that the subquery in the select list doesn't need any columns from the right table:
Explain `select t1a.*, (select max(t1c.a) from t1 t1c where t1c.b = t1a.b) from t1 t1a left join t1 t1b on t1a.a = t1b.a;`

PR #64254, opened 2025-11-03
planner: Record agg subquery columns for outer join pruning

terry1purcell:

Issue Number: ref #63456

Problem Summary:

If an aggregate correlated subquery appears in the select list of a query, the columns used in that subquery are not checked to determine which table(s) they correlate to. Therefore, such queries are not eligible for outer join elimination.

This PR records the columns used in the subquery to allow detection of whether the columns allow outer join elimination to occur.

This is tied to tidb_opt_enable_no_decorrelate_in_select variable - to allow the enhancement to be enabled or disabled if problems occur. In a future release, this can be detached from this variable.

Prior PR #63381 covered exists and scalar subqueries - but missed aggegrate subqueries.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/planner/core/rule_join_elimination.go
+++ b/pkg/planner/core/rule_join_elimination.go
@@ -235,6 +235,16 @@ func (o *OuterJoinEliminator) doOptimize(p base.LogicalPlan, aggCols []*expressi
 				parentCols = append(parentCols, expression.ExtractColumns(byItem.Expr)...)
 			}
 		}
+		// Include columns required by subqueries (appear as correlated columns in child subtree)
+		// TODO: this is tied to variable tidb_opt_enable_no_decorrelate_in_select
+		// to enable outer join elimination when correlated subqueries exist in the aggregation
+		if len(x.Children()) > 0 {
+			x.SCtx().GetSessionVars().RecordRelevantOptVar(vardef.TiDBOptEnableNoDecorrelateInSelect)
+			if x.SCtx().GetSessionVars().EnableNoDecorrelateInSelect {
+				corCols := coreusage.ExtractCorrelatedCols4LogicalPlan(x.Children()[0])
+				parentCols = appendUniqueCorrelatedCols(parentCols, corCols)
+			}
+		}
 	default:
 		parentCols = append(parentCols[:0], p.Schema().Columns...)
 	}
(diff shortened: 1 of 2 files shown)


pkg/planner/core/rule_join_elimination.go
```diff
@@ -235,6 +235,16 @@ func (o *OuterJoinEliminator) doOptimize(p base.LogicalPlan, aggCols []*expressi
 				parentCols = append(parentCols, expression.ExtractColumns(byItem.Expr)...)
 			}
 		}
+		// Include columns required by subqueries (appear as correlated columns in child subtree)
+		// TODO: this is tied to variable tidb_opt_enable_no_decorrelate_in_select
```
qw4990, line 239: Why do we have to tie this to `tidb_opt_enable_no_decorrelate_in_select`? For safety to have a switch to control it or is there any particular purpose?

terry1purcell: It is only for safety - in case it causes a problem, then we have a way to switch it off. I would like to enable tidb_opt_enable_no_decorrelate_in_select by default - because decorrelation in the select list is rarely a good thing. If we do that, then tying this PR to that variable will allow us to expose the benefits - but provide a switch if any problems occur. After 1 release of this enabled -  we can remove this enhancement from being tied to this variable.

pkg/planner/core/rule_join_elimination.go
```diff
@@ -235,6 +235,16 @@ func (o *OuterJoinEliminator) doOptimize(p base.LogicalPlan, aggCols []*expressi
 				parentCols = append(parentCols, expression.ExtractColumns(byItem.Expr)...)
 			}
 		}
+		// Include columns required by subqueries (appear as correlated columns in child subtree)
```
qw4990, line 238: Emm just want to confirm why do we need to add this part of code, I commented this part of code and tested the case added into `TestOuterJoinElimination` in this PR, it seems like even without this part of code, the outer-join elimination can still work:
image

Issue #64257, opened 2025-11-04
BR scan region unable to properly fallback to leader handle


okJiang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```golang
	_ = utils.WithRetry(ctx, func() error {
		regions := make([]*RegionInfo, 0, 16)
		scanStartKey := startKey
		for {
			var batch []*RegionInfo
			if err != nil {
				batch, err = client.ScanRegions(ctx, scanStartKey, endKey, limit)
			} else {
				batch, err = client.ScanRegions(ctx, scanStartKey, endKey, limit, opt.WithAllowFollowerHandle())
			}

			if err != nil {
				err = errors.Annotatef(berrors.ErrPDBatchScanRegion.Wrap(err), "scan regions from start-key:%s, err: %s",
					redact.Key(scanStartKey), err.Error())
				return err
			}
			regions = append(regions, batch...)
			if len(batch) < limit {
				// No more region
				break
			}
			scanStartKey = batch[len(batch)-1].Region.GetEndKey()
			if len(scanStartKey) == 0 ||
				(len(endKey) > 0 && bytes.Compare(scanStartKey, endKey) >= 0) {
				// All key space have scanned
				break
			}
		}
		// if the number of regions changed, we can infer TiKV side really
		// made some progress so don't increase the retry times.
		if len(regions) != len(lastRegions) {
			backoffer.Stat.ReduceRetry()
		}
		lastRegions = regions

		if err = checkRegionConsistency(startKey, endKey, regions); err != nil {
			log.Warn("failed to scan region, retrying",
				logutil.ShortError(err),
				zap.Int("regionLength", len(regions)))
			return err
		}
		return nil
	}, backoffer)
```

When execute

```
if err != nil {
    batch, err = client.ScanRegions(ctx, scanStartKey, endKey, limit)
}
```

the err will be replace with a new error. So the next region scan will use follower-handle mode, which is unexpected.

### 2. What did you expect to see? (Required)

If encounter error, fallback to using follower handle mode.

PR #64259, opened 2025-11-04
br: improve error handling in PaginateScanRegion function

okJiang:

Issue Number: close https://github.com/pingcap/tidb/issues/64257

Problem Summary:

The scan region here is a batch. BR will retry this batch scan region.

- Expectation: If an error is reported in every retry, the whole batch scan region should be retreated to the leader handle.
- Now, when an error retry is encountered, the first scan region in the loop will be retreated, but when err is changed to nil in the first retreated scan region, the retreated will be canceled. The first scan region will be retreated, but the first time the scan region is retreated, err is changed to nil, so the retreated region is canceled, which is not as expected.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/br/pkg/restore/split/split.go
+++ b/br/pkg/restore/split/split.go
@@ -218,18 +218,19 @@ func PaginateScanRegion(
 
 	var (
 		lastRegions []*RegionInfo
-		retryErr    error
+		lastErr     error
 		backoffer   = NewWaitRegionOnlineBackoffer()
 	)
-	retryErr = utils.WithRetry(ctx, func() error {
+	_ = utils.WithRetry(ctx, func() error {
+		var err error
+		defer func() {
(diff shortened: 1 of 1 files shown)


br/pkg/restore/split/split.go
```diff
@@ -265,7 +268,7 @@ func PaginateScanRegion(
 		return nil
 	}, backoffer)
 
-	return lastRegions, err
+	return lastRegions, retryErr
```
3pointer:
```suggestion
	return lastRegions, retryErr
```

should better use last error, or the error stack will be very large.

PR #64260, opened 2025-11-04
sessionctx: change @@tidb_trace_event default value from 'off' to 'base'

tiancaiamao:

Issue Number: ref #64008

Problem Summary:

https://github.com/pingcap/tidb/pull/64051 introduce this variable, but the default values should be 'base' rather than 'off'.

OFF: disable trace event
BASE: always enable flight recorder, discard trace events or dump on required, not logging
FULL: print trace event in logging

This is still experimental SQL variable, not exposed to user yet.
The big picture there are three orthogonal combination:

- probes: this is the source of trace event generation 
- control: what kinds of event should be recorded
- sink: dump to flight recorder, or logging ...

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/sessionctx/vardef/BUILD.bazel
+++ b/pkg/sessionctx/vardef/BUILD.bazel
@@ -18,6 +18,7 @@ go_library(
         "//pkg/util/memory",
         "//pkg/util/paging",
         "//pkg/util/size",
+        "//pkg/util/traceevent",
         "@com_github_pingcap_tipb//go-tipb",
         "@org_uber_go_atomic//:atomic",
     ],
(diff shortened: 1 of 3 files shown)


cfzjywxk:
> Before changing the default value to `base` we should determine what are the possible values. One possible approach is to let this variable accept either
> 
> 1. a json structure to customize the behavior
> 2. or `off`, `base` or `full`, which are aliases of some pre-defined json configs
>    What do you think? cc @MyonKeminta @cfzjywxk

+1, it is not urgent to change the default value, to make it work properly is more prioritized IMO

tiancaiamao:
OK, use this one instead https://github.com/tiancaiamao/tidb/pull/13

PR #64289, opened 2025-11-05
planner: tolerate reasonable time lag between different TiDB nodes when updating binding cache

### LINKED ISSUE #64250: global binding cache may be incorrect after concurrent binding operations on multiple tidb nodes
## Bug Report

#### Phenomenon

After concurrent create/drop binding operations on multiple tidb nodes, the global binding cache can be wrong: some updates may be missing forever. For example, the newly created binding doesn't appear in the result of `show global bindings`.
The global binding cache on different tidb nodes may also not be able to reach consistency, and they can't recover by themselves. The only workaround is `admin reload bindings`.

#### Reproduce

This issue is a bit complicated to reproduce. In brief, you need:
1. Concurrent create/drop binding operations on multiple tidb nodes
2. Different tidb nodes don't share the same system clock
3. There should be some network latency between different nodes

I wrote a script to achieve 1: https://gist.github.com/time-and-fate/b82370627e2a55d4c54edc6309db5229 (actually written by AI).

This script connects to 3 tidb nodes, creates 100 tables, creates 200 bindings, and then waits for user input and drops all bindings at last.

To achieve 2 and 3, there are different methods:
1. Simulate some clock skew and latency in the code for **some of** the tidb nodes in the cluster
  ```diff
--- a/pkg/bindinfo/global_handle.go
+++ b/pkg/bindinfo/global_handle.go
@@ -276,7 +276,7 @@ func (h *globalBindingHandle) CreateGlobalBinding(sctx sessionctx.Context, bindi
                }
 
                for i, binding := range bindings {
-                       now := types.NewTime(types.FromGoTime(time.Now()), mysql.TypeTimestamp, 3)
+                       now := types.NewTime(types.FromGoTime(time.Now().Add(-1*time.Second)), mysql.TypeTimestamp, 3)
 
                        updateTs := now.String()
                        _, err = exec(
@@ -333,7 +333,7 @@ func (h *globalBindingHandle) dropGlobalBinding(sqlDigests []string) (deletedRow
                }
 
                for _, sqlDigest := range sqlDigests {
-                       updateTs := types.NewTime(types.FromGoTime(time.Now()), mysql.TypeTimestamp, 3).String()
+                       updateTs := types.NewTime(types.FromGoTime(time.Now().Add(-1*time.Second)), mysql.TypeTimestamp, 3).String()
                        _, err = exec(
                                sctx,
                                `UPDATE mysql.bind_info SET status = %?, update_time = %? WHERE sql_digest = %? AND update_time < %? AND status != %?`,
```
2. Deploy a real cluster that runs on different physical machines.
3. Deploy a cluster running on different virtual machines.

(If the issue still doesn't occur, you can try to disable NTP and manually modify the system clock a little.)

For the script above, there should be 200 bindings after creating, and no bindings at last.

There are fewer than 200 bindings after creating, and some remaining bindings at last.

Tested on v8.5.3. Other versions may also be affected.

### PR DESCRIPTION
**Pull request by qw4990**

Issue Number: close #64250

Problem Summary: planner: tolerate reasonable time lag between different TiDB nodes when updating binding cache

When updating binding cache, TiDB doesn't consider time lag between different nodes, which could cause inconsistency on binding cache. See more info in #64250.

Below is an simplified example, TiDB1's clock is consistent with the real time, while TiDB2's clock is 0.5s slower than the real time:

| Real Physical Time | Action of TiDB1 | Action of TiDB2 |
| -------- | -------- | ------- |
| t1 00:00:30.200000 | create the new binding, write it into `mysql.bind_info` with timestamp `00:00:30.200000` | - | 
| t2 00:00:30.300000 | update binding cache: 1) load latest binding from `mysql.bind_info` where `binding.update_time > LAST_UPDATE_TIME`, 2) set `LAST_UPDATE_TIME` to the latest `binding.update_time`, in this case its `00:00:30.200000` | - |
| t3 00:00:30.500000 | - | create a new binding with timestamp `00:00:30.000000` due to system lock issue  |
| t4 00:00:30.600000 | - | update its binding cache |
| t5 00:01:00.000000 | update binding cache, can't see the binding crated by TiDB2 ❌ | - |

In the case above, TiDB1 can never see the binding created by TiDB2 with timestamp `00:00:30.000000` because its `LAST_UPDATE_TIME = 00:00:30.200000` is larger than that binding's timestamp.

The solution is simple, set a tolerance interval (10s by default) when updating binding cache. And it's safe to load duplicated bindings, because we'll deduplicate them before actually putting them into the cache.

image

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/bindinfo/binding_cache.go
+++ b/pkg/bindinfo/binding_cache.go
@@ -68,7 +68,8 @@ func (u *bindingCacheUpdater) LoadFromStorageToCache(fullLoad bool) (err error) 
 		// timeLagToleranceSec is used to tolerate the time lag between different TiDBs.
 		// See #64250 for more details.
 		// It's safe to load duplicated bindings, because we'll deduplicate them in `pickCachedBinding`.
-		timeLagToleranceSec := 10
+		// TODO: use PD TSO to avoid time synchronization issue thoroughly.
+		const timeLagToleranceSec = 10
 		lastUpdateTime = u.lastUpdateTime.Load().(types.Time)
 		timeCondition = fmt.Sprintf("USE INDEX (time_index) WHERE update_time>DATE_SUB('%s', INTERVAL %d SECOND)",
 			lastUpdateTime.String(), timeLagToleranceSec)
--- a/pkg/bindinfo/binding_cache.go
+++ b/pkg/bindinfo/binding_cache.go
@@ -19,6 +19,7 @@ import (
 	"slices"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/dgraph-io/ristretto"
 	"github.com/pingcap/tidb/pkg/metrics"
@@ -69,10 +70,15 @@ func (u *bindingCacheUpdater) LoadFromStorageToCache(fullLoad bool) (err error) 
 		// See #64250 for more details.
 		// It's safe to load duplicated bindings, because we'll deduplicate them in `pickCachedBinding`.
 		// TODO: use PD TSO to avoid time synchronization issue thoroughly.
-		const timeLagToleranceSec = 10
+		const timeLagTolerance = 10 * time.Second
 		lastUpdateTime = u.lastUpdateTime.Load().(types.Time)
-		timeCondition = fmt.Sprintf("USE INDEX (time_index) WHERE update_time>DATE_SUB('%s', INTERVAL %d SECOND)",
-			lastUpdateTime.String(), timeLagToleranceSec)
+		lastUpdateTimeGo, err := lastUpdateTime.GoTime(time.Local)
+		if err != nil {
+			return err
+		}
+		updateTimeBoundary := types.NewTime(types.FromGoTime(lastUpdateTimeGo.Add(-timeLagTolerance)),
+			lastUpdateTime.Type(), lastUpdateTime.Fsp())
+		timeCondition = fmt.Sprintf("USE INDEX (time_index) WHERE update_time>'%s'", updateTimeBoundary.String())
 	}
 	condition := fmt.Sprintf(`%s ORDER BY update_time, create_time`, timeCondition)
 	bindings, err := readBindingsFromStorage(u.sPool, condition)
--- a/pkg/bindinfo/binding_cache.go
+++ b/pkg/bindinfo/binding_cache.go
@@ -16,6 +16,10 @@ package bindinfo
 
 import (
 	"fmt"
+	"slices"
+	"sync"
+	"sync/atomic"
+
 	"github.com/dgraph-io/ristretto"
 	"github.com/pingcap/tidb/pkg/metrics"
 	"github.com/pingcap/tidb/pkg/parser"
@@ -25,9 +29,6 @@ import (
 	"github.com/pingcap/tidb/pkg/types"
 	"github.com/pingcap/tidb/pkg/util"
 	"go.uber.org/zap"
-	"slices"
-	"sync"
-	"sync/atomic"
 )
 
 // BindingCacheUpdater maintains the binding cache and provide update APIs.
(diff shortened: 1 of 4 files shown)


pkg/bindinfo/binding_cache.go
```diff
@@ -57,14 +57,21 @@ type bindingCacheUpdater struct {
...
+	if fullLoad || lastUpdateTime.IsZero() { // avoid "update_time>'0000-00-00 00:00:00'", which is invalid
 		lastUpdateTime = types.ZeroTimestamp
 		timeCondition = ""
 	} else {
+		// If multiple TiDBs are creating bindings simultaneously and their time are not synchronized,
+		// some bindings' update_time may be earlier than the lastUpdateTime of this TiDB.
+		// timeLagToleranceSec is used to tolerate the time lag between different TiDBs.
+		// See #64250 for more details.
+		// It's safe to load duplicated bindings, because we'll deduplicate them in `pickCachedBinding`.
+		timeLagToleranceSec := 10
```
henrybw: Suggestion: This seems like it should be `const`.

henrybw (approved):
I think this is reasonable and overall an improvement.

That said, I think using node-local timestamps for synchronization is generally not a good idea. I think it would make more sense to use timestamps from the PD TSO instead, which are guaranteed to be monotonic across the cluster. (I wonder why we didn't use these in the first place for bindings...)

However, that's a bigger project that we should consider as a future improvement. For now, this seems like an acceptable compromise. LGTM.

pkg/bindinfo/binding_operator.go
```diff
@@ -76,6 +86,9 @@ func (op *bindingOperator) CreateBinding(sctx sessionctx.Context, bindings []*Bi
 
 		for i, binding := range bindings {
 			now := types.NewTime(types.FromGoTime(time.Now()), mysql.TypeTimestamp, 6)
+			if intest.InTest && mockTimeLag != 0 {
```
guo-shaoge, line 89: Maybe it's better to use failpoint here?

pkg/bindinfo/binding_cache.go
```diff
@@ -57,14 +57,21 @@ type bindingCacheUpdater struct {
...
 		lastUpdateTime = u.lastUpdateTime.Load().(types.Time)
-		timeCondition = fmt.Sprintf("USE INDEX (time_index) WHERE update_time>'%s'", lastUpdateTime.String())
+		timeCondition = fmt.Sprintf("USE INDEX (time_index) WHERE update_time>DATE_SUB('%s', INTERVAL %d SECOND)",
+			lastUpdateTime.String(), timeLagToleranceSec)
```
guo-shaoge: I think we should do this subtraction in go code.
Because all rows of the system table `bind_info` need to be computed using this filter. And if there are lots of bindings in this system table, it can cost considerable waste of computation resource in TiKV,

guo-shaoge:
other LGTM

pkg/bindinfo/binding_operator.go
qw4990: I've thought of this and finally decided to use this way for 2 reasons:
1. this way is easier to run test than failpoint because we can run this test directly instead of running `enable / disable xxFailpoint` before and after this test;
2. using `context` to pass information across different threads or goroutines is more Golang Style (I think that one of the reason why they add `context` as the first argument to many standard lib)

pkg/bindinfo/binding_cache.go
qw4990: Updated. And I also add a comment there to notify that we can use PD TSO to solve this issue more thoroughly 🙏

pkg/bindinfo/binding_cache.go
qw4990: Updated, PTAL @guo-shaoge

guo-shaoge (approved):

Issue #64274, opened 2025-11-04
Async load may panic when processing dropped indexes


0xPoe:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

https://github.com/pingcap/tidb/blob/801c4a6ec9b56e4e57e97dc50d48e69a9faee7a5/pkg/statistics/handle/storage/read.go#L624

### 2. What did you expect to see? (Required)

It should check if index is nil since we don't check it in the IsLoadNeeded func.

### 3. What did you see instead (Required)
It directly accesses the IsLoadNeeded func.

### 4. What is your TiDB version? (Required)

v7.5.7

0xPoe:
It doesn’t cause the server to exit, since we recover from the panic. However, it does generate many panic logs because we fail to properly remove that index from the “needed” map.

0xPoe:
Introduced by https://github.com/pingcap/tidb/pull/54060

0xPoe:
To reproduce on real tikv test:
```go
func TestLoadNonExistentIndexStats(t *testing.T) {
	store, dom := realtikvtest.CreateMockStoreAndDomainAndSetup(t)
	tk := testkit.NewTestKit(t, store)
	tk.MustExec("use test;")
	tk.MustExec("drop table if exists t;")
	// 1. Create a table and the statsHandle.Update(do.InfoSchema()) will load this table into the stats cache.
	tk.MustExec("create table if not exists t(a int, b int);")
	tk.MustExec("alter table t add index ia(a);")
	// 2. Insert some data and wait for the modify_count and the count is not null in the mysql.stats_meta.
	tk.MustExec("insert into t value(1,1), (2,2);")
	h := dom.StatsHandle()
	require.NoError(t, h.DumpStatsDeltaToKV(true))
	require.NoError(t, h.Update(dom.InfoSchema()))
	// 3. Try to select some data from this table by ID, it would trigger an async load.
	tk.MustExec("set tidb_opt_objective='determinate';")
	tk.MustQuery("select * from t where a = 1 and b = 1;").Check(testkit.Rows("1 1"))
	table, err := dom.InfoSchema().TableByName(model.NewCIStr("test"), model.NewCIStr("t"))
	require.NoError(t, err)
	tableInfo := table.Meta()
	droppedIndexID := tableInfo.Indices[0].ID
	// Check the histogram items
	require.Eventually(t, func() bool {
[...]

PR #64292, opened 2025-11-05
statistics: remove the unnecessary log

0xPoe:

Issue Number: close https://github.com/pingcap/tidb/issues/64274

Problem Summary:

IsLoadNeeded doesn’t check for nil internally. We should explicitly verify that it’s safe to use.

The real use case the users may have is as follows:
1. Create a small table (fewer than 1,000 rows).
2. Add a new index.
3. set tidb_opt_objective='determinate';
4. Query the table using that index.
5. The system repeatedly panics and recovers from a nil panic.

I also removed this unnecessary log this time. When I added it before, we hadn’t yet fully understood the real trigger condition of this issue — in our tests, we manually deleted histogram records to simulate it.
Now that I fully understand how it actually happens, this log is completely meaningless. This issue should be quite common once the above variable is enabled.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
修复异步加载统计信息可能遇到的 nil Panic 问题
Fix a potential nil panic issue in asynchronous statistics loading
```

### DIFF
--- a/pkg/statistics/handle/storage/read_test.go
+++ b/pkg/statistics/handle/storage/read_test.go
@@ -158,7 +158,7 @@ func TestLoadNonExistentIndexStats(t *testing.T) {
 		return false
 	}, time.Second*5, time.Millisecond*100, "Index ia should be in HistogramNeededItems")
 
-	// Verify that LoadNeededHistograms doesn't panic when the index exists in the cache
+	// Verify that LoadNeededHistograms doesn't panic when the pseudo index stats exists in the cache
 	// but doesn't have histogram data in mysql.stats_histograms yet.
 	err = util.CallWithSCtx(h.SPool(), func(sctx sessionctx.Context) error {
 		require.NotPanics(t, func() {
--- a/tests/realtikvtest/statisticstest/statistics_test.go
+++ b/tests/realtikvtest/statisticstest/statistics_test.go
@@ -348,7 +348,7 @@ func TestLoadNonExistentIndexStats(t *testing.T) {
 		return false
 	}, time.Second*5, time.Millisecond*100, "Index ia should be in HistogramNeededItems")
 
-	// Verify that LoadNeededHistograms doesn't panic when the index exists in the cache
+	// Verify that LoadNeededHistograms doesn't panic when the pseudo index stats exists in the cache
 	// but doesn't have histogram data in mysql.stats_histograms yet.
 	err = util.CallWithSCtx(h.SPool(), func(sctx sessionctx.Context) error {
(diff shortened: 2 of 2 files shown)


pkg/statistics/handle/storage/read_test.go
```diff
@@ -116,3 +121,50 @@ func TestReloadExtStatsLockRelease(t *testing.T) {
...
+	tk.MustExec("drop table if exists t;")
+	// Create table with an index. The index histogram will exist in the stats cache
+	// but won't have actual histogram data loaded yet.
+	tk.MustExec("create table if not exists t(a int, b int, index ia(a));")
+	tk.MustExec("insert into t value(1,1), (2,2);")
+	h := dom.StatsHandle()
+	require.NoError(t, h.DumpStatsDeltaToKV(true))
+	require.NoError(t, h.Update(dom.InfoSchema()))
+	// Trigger async load of index histogram by using the index in a query.
+	tk.MustExec("set tidb_opt_objective='determinate';")
```
0xPoe, line 143: When we set this variable to true, it causes the physical table to no longer use a fake physical table ID, which in turn triggers statistics loading.
```go
func (idx *Index) CheckStats() {
	// When we are using stats from PseudoTable(), all column/index ID will be -1.
	if idx.IsFullLoad() || idx.PhysicalID <= 0 {
		return
	}
	HistogramNeededItems.Insert(model.TableItemID{TableID: idx.PhysicalID, ID: idx.Info.ID, IsIndex: true})
}

```

pkg/statistics/handle/storage/read.go
```diff
@@ -621,12 +621,6 @@ func loadNeededIndexHistograms(sctx sessionctx.Context, statsCache util.StatsCac
 	// If we don't do this it might cause a memory leak.
 	// See: https://github.com/pingcap/tidb/issues/54022
 	if !ok || !index.IsLoadNeeded() {
-		if !index.IsLoadNeeded() {
```
0xPoe, line 624: The panic occurs here because I forgot to check the ok condition, and I mistakenly thought the IsLoadNeeded function handled the nil check internally.

0xPoe (commented):

- [x] Unit tests added

- [x] All tests pass

- [x] PR size is reasonable


/cc @mjonss @AilinKid @time-and-fate

AilinKid (approved):

0xPoe:
Won't fix for 6.5.

PR #64309, opened 2025-11-06
expression: replace equal condition by true when to constant propagation | tidb-test=pr/2625 

### LINKED ISSUE #64216: cannot simplify the predicate
## Bug Report

```
CREATE TABLE t1 (a int PRIMARY KEY, b int);
CREATE TABLE t2 (a int PRIMARY KEY, b int);

EXPLAIN FORMAT='brief' SELECT STRAIGHT_JOIN * FROM t1 LEFT JOIN t2 ON t1.a = t2.a WHERE t1.a = t2.a OR t1.a = t2.b;
```

In Mysql

```
+----------------------------------------------------------------------------------------------------+
| EXPLAIN                                                                                            |
+----------------------------------------------------------------------------------------------------+
| -> Nested loop inner join  (cost=0.7 rows=1) (actual time=0.172..0.172 rows=0 loops=1)             |
|     -> Table scan on t1  (cost=0.35 rows=1) (actual time=0.161..0.161 rows=0 loops=1)              |
|     -> Single-row index lookup on t2 using PRIMARY (a = t1.a)  (cost=0.35 rows=1) (never executed) |
+----------------------------------------------------------------------------------------------------+
```

```
EXPLAIN FORMAT='brief' SELECT STRAIGHT_JOIN * FROM t1 LEFT JOIN t2 ON t1.a = t2.a WHERE t1.a = t2.a OR t1.a = t2.b;
MergeJoin	3.84	root		inner join, left key:join_outer.t1.a, right key:join_outer.t2.a, other cond:or(eq(join_outer.t1.a, join_outer.t2.a), eq(join_outer.t1.a, join_outer.t2.b))
├─TableReader(Build)	3.84	root		data:Selection
│ └─Selection	3.84	cop[tikv]		or(eq(join_outer.t2.a, join_outer.t2.a), eq(join_outer.t2.a, join_outer.t2.b))
│   └─TableFullScan	4.00	cop[tikv]	table:t2	keep order:true
└─TableReader(Probe)	9.00	root		data:TableFullScan
  └─TableFullScan	9.00	cop[tikv]	table:t1	keep order:true
```

hawkingrei:

Issue Number: close #64216 

Problem Summary:

There are three scenarios here.

```
a = b and (a=b or c=d) => a=b and (a=b or c=d) and (a=a or c=d)
a = b and a in (b, c, d) => a = b and a in (b, c, d) and a in (a, c ,d)
a = b and b = c and c = d and a = d => a = b and b = c and c = d // remove a=d
```

In the previous implementation, we only passed through constant values to create a new expression. However, what we actually need is to directly replace the equivalent conditions in the expression with true, in order to avoid generating unnecessary expressions.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/expression/constant_propagation.go
+++ b/pkg/expression/constant_propagation.go
@@ -405,6 +405,7 @@ func (s *propConstSolver) propagateColumnEQ() {
 	} else {
 		s.unionSet.GrowNewIntSet(len(s.columns))
 	}
+	allVisited := true
 	for i := range s.conditions {
 		if fun, ok := s.conditions[i].(*ScalarFunction); ok && fun.FuncName.L == ast.EQ {
 			lCol, lOk := fun.GetArgs()[0].(*Column)
@@ -425,8 +426,13 @@ func (s *propConstSolver) propagateColumnEQ() {
 						RetType: types.NewFieldType(mysql.TypeTiny),
 					}
 				}
+				continue
 			}
 		}
+		allVisited = false
+	}
+	if allVisited {
+		return
 	}
 	condsLen := len(s.conditions)
 	for i, coli := range s.columns {


pkg/expression/constant_propagation.go
```diff
@@ -348,12 +413,21 @@ func (s *propConstSolver) propagateColumnEQ() {
...
 				rID := s.getColID(rCol)
-				s.unionSet.Union(lID, rID)
 				visited[i] = true
+				if s.unionSet.FindRoot(lID) != s.unionSet.FindRoot(rID) {
+					// Add the equality relation to unionSet
+					// if it has been added, we don't need to process it again.
+					// It will be deleted in the replaceEqCondtionWithTrueitionsWithConstants
+					s.unionSet.Union(lID, rID)
+				} else if lID != rID {
+					s.conditions[i] = &Constant{
```
AilinKid, line 424: here means LID and RID is already in the union set(the equiv relationship is catched, so we can safely convert this EQ(LID, RID) to be a true)

what i concern here is that, next time, if we wanna do the constant prop one more time, we should build the union set from these condition slice again, while may we still be able to catch the equiv relation ship between LID and RID?

AilinKid: looks like we only substitute for appended condition, yes ok.

hawkingrei: ```Unionset```  can capture any equivalence relations of the inferred type. for example,

```
a = b, b = c, c = d, d = e 
```

We get ```a = e ``` by the ```unionset```.

hawkingrei: Actually, it doesn't affect the logic here, since all equivalence relations are preserved, and only the redundant equivalence relations are eliminated.

PR #64322, opened 2025-11-06
expression: explore all possibilities of tryToReplaceCond as much as possible

### LINKED ISSUE #64325: it is unstable to push down the predicate
## Bug Report

```

func TestABC(t *testing.T) {
	tk.MustExec("use test")
	tk.MustExec(`CREATE TABLE t1 (a int PRIMARY KEY, b int);`)
	tk.MustExec(`CREATE TABLE t2 (a int PRIMARY KEY, b int);`)
	for i := range 100 {
		t.Log(i)
		tk.MustQuery(`EXPLAIN FORMAT='plan_tree' SELECT STRAIGHT_JOIN * FROM t1 LEFT JOIN t2 ON t1.a = t2.a WHERE t1.a IN(t2.a, t2.b);`).Check(testkit.Rows(
			`MergeJoin root  inner join, left key:test.t1.a, right key:test.t2.a, other cond:in(test.t1.a, test.t2.a, test.t2.b)`,
			`├─TableReader(Build) root  data:Selection`,
			`│ └─Selection cop[tikv]  in(test.t2.a, test.t2.a, test.t2.b)`,
			`│   └─TableFullScan cop[tikv] table:t2 keep order:true, stats:pseudo`,
			`└─TableReader(Probe) root  data:TableFullScan`,
			`  └─TableFullScan cop[tikv] table:t1 keep order:true, stats:pseudo`))
	}
}
```

```
MergeJoin root  inner join, left key:test.t1.a, right key:test.t2.a, other cond:in(test.t1.a, test.t2.a, test.t2.b)
├─TableReader(Build) root  data:Selection
│ └─Selection cop[tikv]  in(test.t2.a, test.t2.a, test.t2.b)
│   └─TableFullScan cop[tikv] table:t2 keep order:true, stats:pseudo
└─TableReader(Probe) root  data:TableFullScan
```

```
MergeJoin root  inner join, left key:test.t1.a, right key:test.t2.a, other cond:in(test.t1.a, test.t2.a, test.t2.b)
├─TableReader(Build) root  data:TableFullScan
│ └─TableFullScan cop[tikv] table:t2 keep order:true, stats:pseudo
└─TableReader(Probe) root  data:TableFullScan
```

master

hawkingrei:

Issue Number: close #64325

Problem Summary:

The case is like this.

It has a join, and the schemas of the two children are as follows.

```
right schema: [3,4]
left schema: [1,2]
```

It has an equal condition `1=3` and now we need to perform constant propagation on the following expressions.

```
1 in (3, 4)
```

The order of  1=3  will affect the changes here in the tryToReplaceCond.

When we call tryToReplaceCond to replace 3 by 1. `1 in (1,4)` has two columns, which are in the difference schema. So it cannot push down the child.

But the problem is that it will not continue to try to replace 1 by 3. If it can do it,
`3 in (3,4)` can push down into the child.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/expression/constant_propagation.go
+++ b/pkg/expression/constant_propagation.go
@@ -375,6 +375,16 @@ func (s *propConstSolver) propagateColumnEQ() {
 						s.conditions = append(s.conditions, newExpr)
 					}
 				}
+				// Why do we need to replace colj with coli again?
+				// Consider the case: col1 = col3
+				// Join Schema: left schema: {col1,col2}, right schema: {col3,col4}
+				// Conditions: col1 in (col3, col4)
+				//
+				// because the order of columns in equaility relation is not guaranteed,
+				// We may replace col3 with col1 first, it cannot push down the condition to the child.
+				// because two columns are from different side.
+				// But if we replace col1 with col3, it can be pushed down.
+				// So we need to try both directions.
 				replaced, _, newExpr = tryToReplaceCond(s.ctx, colj, coli, cond, false)
 				if replaced {
 					// TODO(hawkingrei): if it is the true expression, we can remvoe it.


pkg/expression/constant_propagation.go
```diff
@@ -371,18 +371,16 @@ func (s *propConstSolver) propagateColumnEQ() {
 				replaced, _, newExpr := tryToReplaceCond(s.ctx, coli, colj, cond, false)
 				if replaced {
 					// TODO(hawkingrei): if it is the true expression, we can remvoe it.
-					if !isConstant(newExpr) && s.vaildExprFunc != nil && !s.vaildExprFunc(newExpr) {
-						continue
+					if isConstant(newExpr) || s.vaildExprFunc == nil || s.vaildExprFunc(newExpr) {
```
AilinKid, line 374: looks like revert the inside out？
there is no quick continue, and it will fall into L378?

AilinKid: better comment what the specific case , what kind of expression you want to explore and pushdown here

hawkingrei: Done

PR #64337, opened 2025-11-07
ddl, model: fix unexpected missing analyze for multi schema change

joechenrh:

Issue Number: ref https://github.com/pingcap/tidb/issues/63366

Problem Summary:

If the last sub job in the multi schema change doesn't need reorg (i.e., no index involved), the analyze task will be skipped.

Take the following SQL as example, since `c2` has no index on it, the state is directly converted to `AnalyzeStateSkipped`, even if modifying `c1` requires reorg.

```SQL
create table t2(id int, c1 int, c2 int, index i1(c1));
alter table t2 modify column c1 char(16), modify column c2 char(16);
```

```
c1: ReorgStageModifyColumnCompleted -> AnalyzeStateSkipped
(NeedAnalyze == false)
c2: ReorgStageModifyColumnCompleted -> AnalyzeStateSkipped
(NeedAnalyze == true & len(changingIdx) == 0)
```

The problems is caused by inaccurate value for `SubJob.NeedAnalyze`. That is, **the last job with `NeedAnalyze == true` is not the last job that does reorg**. To fix this:

- Remove `NeedAnalyze` since it's not accurate.
- For sub jobs, they are converted into non-revertible state after reorg is finished.
- For multi schema change, the analyze is done in the parent job after all sub jobs are non-revertible.

This PR also do the following things
1. move a bunch of function, variables related to modify column from `ddl`  to `model` package.
2. Replace `CtxVars` with boolean value, as it's used by `MODIFY COLUMN` and store a single bool.
2. filter out temporary columns from the display information of analyze.  Previously, we may get output like: `auto analyze table all indexes, columns a, b, c,_Col$_c_0 with 256 buckets, 100 topn, 1 samplerate`

### Check List

Tests 

- [X] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/ddl/multi_schema_change_test.go
+++ b/pkg/ddl/multi_schema_change_test.go
@@ -908,8 +908,8 @@ func TestMultiSchemaChangePollJobCount(t *testing.T) {
 	})
 	// Should not test reorg DDL because the result can be unstable.
 	tk.MustExec("alter table t add column b int,  modify column a bigint, add column c char(10);")
-	require.Equal(t, 29, runOneJobCounter)
-	require.Equal(t, 9, pollJobCounter)
+	require.Equal(t, 30, runOneJobCounter)
+	require.Equal(t, 10, pollJobCounter)
 }
 
 func TestMultiSchemaChangeMDLView(t *testing.T) {


pkg/meta/model/job.go
```diff
@@ -837,11 +873,10 @@ type SubJob struct {
 	State        JobState        `json:"state"`
 	RowCount     int64           `json:"row_count"`
 	Warning      *terror.Error   `json:"warning"`
-	CtxVars      []any           `json:"-"`
+	NeedReorg    bool            `json:"-"`
 	SchemaVer    int64           `json:"schema_version"`
 	ReorgTp      ReorgType       `json:"reorg_tp"`
 	ReorgStage   ReorgStage      `json:"reorg_stage"`
-	NeedAnalyze  bool            `json:"need_analyze"`
```
fzzf678, line 874: Will the modification of `CtxVars` and  delete of `NeedAnalyze` make json unmarshal fail while updating tidb?

joechenrh: No, json unmarshal won't be affected by this deleted field. Besides, this PR should be coexist with previously PR which introduced `NeedAnalyze`.

fzzf678: About this `NeedAnalyze` and `CtxVars` field,  if we submitted the job use these 2 fields, and then update the tidb to this PR,  continue doing ddl, will unexpected things happen?

joechenrh: `CtxVars` is not persisted, it's only used when submitting the job. And it's only used by modify column, so I changed it to a boolean `NeedReorg`.

For `NeedAnalyze`, we don't rely on it with this PR, `checkNeedAnalyzeForMultiSchemaChange` just make the decision based on job type and table meta.

joechenrh:
/hold
There may be something wrong with this PR.

joechenrh:
/unhold
It's the expected behavior that the column without index is not analyzed.

PR #64341, opened 2025-11-07
*: flight recorder tracing for internal sessions

tiancaiamao:

Issue Number: ref #64008

Problem Summary:

1. Try to pass the `traceID`  for the DDL statement

~~There are all kinds of DDLs, this PR just handle DDL create table as an  example. ~~
~~Add context parameter to `CreateTable`, this interface change affected a lot code.~~
~~Before creating DDL job, the trace ID is extracted from context and attached to the job.~~
~~DDL scheduler and worker would take the DDL job out (maybe in another TiDB) and get the original trace ID~~
~~(To be frank, this seems not quite useful, every internal DDL SQL will update traceID and invalidate the original one)~~

2. In `session.ExecuteInternal`, a new trace will be created

This change could catch any internal SQLs.
And if there is not already a flight recorder binded, create a new one.
If the trace ctx is missing, I'll print the stack so we can see what's background worker is calling internal SQL
without adding the trace information.

3. Add `dump_trigger.suspicious_event.dev_debug` and `dump_trigger.suspicious_event.is_internal` dump trigger condition

The purpose of the former is that we can do this to get the missing trace for background workers:

```
curl -X POST -d @config.json http://127.0.0.1:10080/debug/traceevent
cat config.json
{
  "enabled_categories": ["*"],
  "dump_trigger": {
    "type": "suspicious_event",
    "suspicious_event": {
      "type": "dev_debug",
      "dev_debug": {
        "type": "execute_internal_trace_missing"
      }
    }
  }
}
```

And gradually add the trace for internal sessions.

`dump_trigger.suspicious_event.is_internal` is obvious, no need to explain.

4. Add trace to some of the background work.

The code pattern looks like:

```
+ trace := traceevent.NewTrace()
+ ctx := tracing.WithFlightRecorder(sm.ctx, trace)
for {
    select {
        case <-sm.ctx.Done():
            ...
            failpoint.InjectCall("beforeGetSchedulableTasks")
-           schedulableTasks, err := sm.getSchedulableTasks()
+           schedulableTasks, err := sm.getSchedulableTasks(ctx)
+           trace.DiscardOrFlush(ctx)
            ...
        }
```

Every background goroutine should comes with a new `traceevent.NewTrace()`

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [X] No need to test
  > - [X] I checked and no code files have been changed.
  
  Just adding trace, not modifying existing code logic
  
  > 
  

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -2383,6 +2383,17 @@ type isInternalAlias struct {
 	bool
 }
 
+// isInternalAlias is not intuitive, but it is defined to avoid allocation.
+// If the code is written as
+//
+//		traceevent.CheckFlightRecorderDumpTrigger(ctx, "dump_trigger.suspicious_event.is_internal", func(conf *traceevent.DumpTriggerConfig) {
+//	     	conf.Event.IsInternal = conf.Event.IsInternal
+//		})
+//
+// It's uncertain whether the Go compiler escape analysis is powerful enough to avoid allocation for the closure object.
+// isInternalAlias is defined to help the compiler, this coding style will not cause closure object allocation.
+//
+//	traceevent.CheckFlightRecorderDumpTrigger(ctx, "dump_trigger.suspicious_event.is_internal", isInternalAlias{s.isInternal()}.isInternalDumpTriggerCheck)
 func (i isInternalAlias) isInternalDumpTriggerCheck(config *traceevent.DumpTriggerConfig) bool {
 	return config.Event.IsInternal == i.bool
 }


pkg/ddl/executor.go
```diff
@@ -6625,30 +6630,38 @@ func (e *executor) genPlacementPolicyID() (int64, error) {
...
+	r, ctx := tracing.StartRegionEx(ctx, "ddl.DoDDLJobWrapper")
+	defer r.End()
+
 	job := jobW.Job
 	job.TraceInfo = &tracing.TraceInfo{
-		ConnectionID: ctx.GetSessionVars().ConnectionID,
-		SessionAlias: ctx.GetSessionVars().SessionAlias,
+		ConnectionID: sctx.GetSessionVars().ConnectionID,
+		SessionAlias: sctx.GetSessionVars().SessionAlias,
+		TraceID:      traceevent.TraceIDFromContext(ctx),
```
tiancaiamao: The `ctx` is passed all the way down to here, so the DDL job get the TraceID

```
executor/DDLExec.Next
executor/DDLExec.executeCreateTable
ddl/executor/executor.CreateTable
ddl/executor/executor.CreateTableWithInfo
ddl/executor/executor.DoDDLJobWrapper
```

pkg/ddl/job_scheduler.go
```diff
@@ -480,13 +487,21 @@ func (s *jobScheduler) mustReloadSchemas() {
 // deliveryJob deliver the job to the worker to run it asynchronously.
 // the worker will run the job until it's finished, paused or another owner takes
 // over and finished it.
-func (s *jobScheduler) deliveryJob(wk *worker, pool *workerPool, jobW *model.JobW) {
+func (s *jobScheduler) deliveryJob(ctx context.Context, wk *worker, pool *workerPool, jobW *model.JobW) {
+	r := tracing.StartRegion(ctx, "jobScheduler.deliveryJob")
+	defer r.End()
```
tiancaiamao, line 491: ddl Job scheduler load the DDL job ...  it can take the traceID out from the DDL job:


```
ddl/jobScheduler.scheduleLoop
ddl/jobScheduler.schedule
ddl/jobScheduler.loadAndDeliverJobs
ddl/jobScheduler.deliveryJob
```

tiancaiamao: The DDL worker will handle that job (in another goroutine).

The DDL worker will run several other SQLs, each SQL statement refresh the trace ID.
So the original traceID is lost ...

For internal session that execute several SQL, it's not so easy to trace the original trace ID.
Here I've tried my best.

pkg/util/traceevent/traceevent.go
```diff
@@ -269,7 +275,9 @@ func GenerateTraceID(ctx context.Context, startTS uint64, stmtCount uint64) []by
 	var rand32 uint32
 	if sink := tracing.GetSink(ctx); sink != nil {
 		if t, ok := sink.(*Trace); ok {
+			t.mu.Lock()
```
cfzjywxk, line 278: Why `Lock` is needed here?

tiancaiamao: We cannot prevent a context been pass from one goroutine to another.
And that may cause DATA RACE, so we need Lock. For example:

From goroutine1:
trace.TraceEvent(ctx, category, "event"..)

Then goroutine2:
go func(ctx context.Context) {
     session.ExecInternal(ctx) // ExecuteStmt will call GenerateTraceID and conflict with goroutine1
}

The ideal way is goroutine2 bind a new Trace

```
trace := traceevent.NewTrace()
ctx = traceevent.WithFlightRecorder(ctx, trace)
go func(ctx ) {
   session.ExecuteInternal(ctx)
}
trace.FlushOrDiscard()
```

But it's hard to guarantee all the background workers.

pkg/ddl/schematracker/dm_tracker.go
```diff
@@ -1173,7 +1174,7 @@ func (*SchemaTracker) AlterResourceGroup(_ sessionctx.Context, _ *ast.AlterResou
 // BatchCreateTableWithInfo implements the DDL interface, it will call CreateTableWithInfo for each table.
 func (d *SchemaTracker) BatchCreateTableWithInfo(ctx sessionctx.Context, schema ast.CIStr, info []*model.TableInfo, cs ...ddl.CreateTableOption) error {
 	for _, tableInfo := range info {
-		if err := d.CreateTableWithInfo(ctx, schema, tableInfo, nil, cs...); err != nil {
+		if err := d.CreateTableWithInfo(context.Background(), ctx, schema, tableInfo, nil, cs...); err != nil {
```
ekexium: Should we mark TODOs for such `context.Background()` in production code?

pkg/util/tracing/util.go
```diff
@@ -241,6 +243,10 @@ const (
 	UnknownClient
 	// General is used by tracing API
 	General
+	// DDLJob traces DDL job events.
+	DDLJob
+	// DevDebug traces development/debugging events.
+	DevDebug
```
ekexium, line 244: Let's remember to exclude it from default setting

pkg/ddl/schematracker/dm_tracker.go
tiancaiamao: You mean `context.TODO()`? It doesn't really work, just a hint we should pass down the ctx
`context.Background()` is the same ... we know if we have the correct ctx, that would be used.

pkg/util/tracing/util.go
tiancaiamao: OK, so that's do it in the next commit to update sessionctx/variable test for `@@tidb_trace_event` default value.
Otherwise, every time we need an approve from var_def/OWNER and that's quite inconvient. @ekexium

tiancaiamao: https://github.com/pingcap/tidb/pull/64511

pkg/ddl/schematracker/dm_tracker.go
ekexium: I mean will we add a `ctx` argument to `BatchCreateTableWithInfo` in the future?

pkg/ddl/executor.go
```diff
@@ -109,7 +111,7 @@ type Executor interface {
 	CreateSchema(ctx sessionctx.Context, stmt *ast.CreateDatabaseStmt) error
 	AlterSchema(sctx sessionctx.Context, stmt *ast.AlterDatabaseStmt) error
 	DropSchema(ctx sessionctx.Context, stmt *ast.DropDatabaseStmt) error
-	CreateTable(ctx sessionctx.Context, stmt *ast.CreateTableStmt) error
+	CreateTable(ctx context.Context, sctx sessionctx.Context, stmt *ast.CreateTableStmt) error
```
D3Hunter: can we pass `context.Background()` like others, so no need to pass this separate `ctx

pkg/ddl/job_scheduler.go
```diff
@@ -309,6 +310,10 @@ func (s *jobScheduler) schedule() error {
 	defer ticker.Stop()
 	s.mustReloadSchemas()
 
+	trace := traceevent.NewTrace()
+	ctx := context.Background()
```
D3Hunter: maybe use `s.schCtx`

pkg/disttask/framework/storage/task_table.go
```diff
@@ -373,6 +374,8 @@ func (mgr *TaskManager) getTopTasks(ctx context.Context, states ...proto.TaskSta
 
 // GetTaskExecInfoByExecID implements the scheduler.TaskManager interface.
 func (mgr *TaskManager) GetTaskExecInfoByExecID(ctx context.Context, execID string) ([]*TaskExecInfo, error) {
+	r := tracing.StartRegion(ctx, "TaskManager.GetTaskExecInfoByExecID")
+	defer r.End()
```
D3Hunter, line 378: seems only add tracing for this method, what about others?

pkg/session/session.go
```diff
@@ -2355,6 +2376,10 @@ func queryFailDumpTriggerCheck(config *traceevent.DumpTriggerConfig) bool {
 	return config.Event.Type == "query_fail"
 }
 
+func isInternalDumpTriggerCheck(config *traceevent.DumpTriggerConfig) bool {
+	return config.Event.Type == "is_internal"
```
D3Hunter: why the event type is called `is_internal` which more like a flag itself, not just `internal`

D3Hunter (commented):
> dump_trigger.suspicious_event.is_is_internal is obvious, no need to explain.

why we have double `is`

pkg/ddl/schematracker/dm_tracker.go
tiancaiamao: Finally we should pass ctx to all the DDL interface, including `BatchCreateTableWithInfo` and all others.
(But that's toooo much changes and will be a long continuous refactoring)

tiancaiamao:
> > dump_trigger.suspicious_event.is_is_internal is obvious, no need to explain.
> 
> why we have double `is`

A typo here, let me update the PR description.

pkg/disttask/framework/storage/task_table.go
tiancaiamao: I add some tracing for the internal tasks, but it's hard to cover all of them in one PR.
The most frequent ones should have been covered already.

pkg/session/session.go
tiancaiamao: OK, now I change it to `dump_trigger.suspicious_event.is_internal = true | false` from the old setting `dump_trigger.suspicious_event = is_internal` @D3Hunter

pkg/ddl/job_scheduler.go
tiancaiamao: done

pkg/ddl/executor.go
tiancaiamao: OK, I've reset the DDL executor interface, add a GetTraceCtx() method to sessionctx.Context

tiancaiamao:
Now I've reset the DDL interface changes.  @D3Hunter 
CreateTable uses the original signature, no extra `ctx` added.
I add a `GetTraceCtx` to `sessionctx.Context` to pass the trace ID.

tiancaiamao:
/retest

```
2025/11/19 08:05:22 http: TLS handshake error from 127.0.0.1:38128: client certificate authentication failed. The Common Name from the client certificate [client2 ] was not found in the configuration cluster-verify-cn with value: [client1]
    security_test.go:125:
                Error Trace:    pkg/util/security_test.go:125
                Error:          Error "Get \"https://127.0.0.1:44815\": write tcp 127.0.0.1:38128->127.0.0.1:44815: write: broken pipe" does not contain "tls: bad certificate"
                Test:           TestVerifyCommonNameAndRotate
^V--- FAIL: TestVerifyCommonNameAndRotate (7.71s)
```

pkg/util/traceevent/flightrecorder.go
```diff
@@ -120,6 +120,31 @@ type SuspiciousEventConfig struct {
...
+	DevDebug   *DevDebugConfig `json:"dev_debug,omitempty"`
+}
+
+// DevDebugConfig is the configuration for development debugging.
+type DevDebugConfig struct {
+	Type string
+}
+
+// compile validates the development debugging configuration.
+func (c *DevDebugConfig) compile(b *strings.Builder, mapping *dumpTriggerConfigCompiled, conf *DumpTriggerConfig) (uint64, error) {
```
D3Hunter: `dumpTriggerConfigCompiled` sounds more like a flag, not type name

pkg/util/traceevent/flightrecorder.go
```diff
@@ -120,6 +120,31 @@ type SuspiciousEventConfig struct {
...
+		return 0, fmt.Errorf("dump_trigger.suspicious_event.dev_debug missing")
+	}
+	b.WriteString(".dev_debug")
+	switch c.Type {
+	case "execute_internal_trace_missing":
+		canonicalName := b.String()
+		return mapping.addTrigger(canonicalName, conf)
+	case "send_request_trace_id_missing":
+		canonicalName := b.String()
+		return mapping.addTrigger(canonicalName, conf)
```
D3Hunter: those 2 branch seems the same

pkg/util/traceevent/flightrecorder_test.go
```diff
@@ -152,6 +152,32 @@ func testFlightRecorderConfigGoodCase(t *testing.T) {
...
+	 "enabled_categories": ["*"],
+		"dump_trigger": {
+	    "type": "suspicious_event",
+	    "suspicious_event": {
+	      "type": "dev_debug",
+	      "dev_debug": {
+	        "type": "execute_internal_trace_missing"
+	      }
+	    }
+	  }
```
D3Hunter: can you format this json for better viewing?

pkg/session/session.go
```diff
@@ -1843,6 +1848,27 @@ func (s *session) getOomAlarmVariablesInfo() sessmgr.OOMAlarmVariablesInfo {
...
+		trace := traceevent.NewTrace()
+		ctx = tracing.WithFlightRecorder(ctx, trace)
+		defer trace.DiscardOrFlush(ctx)
+
+		// A developer debugging event so we can see what trace is missing!
+		if traceevent.IsEnabled(tracing.DevDebug) {
+			traceevent.TraceEvent(ctx, tracing.DevDebug, "ExecuteInternal missing trace ctx",
+				zap.String("sql", sql),
+				zap.Stack("stack"))
+			traceevent.CheckFlightRecorderDumpTrigger(ctx, "dump_trigger.suspicious_event.dev_debug", func(config *traceevent.DumpTriggerConfig) bool {
```
D3Hunter, line 1861: it's not a `check`, it also changes internal state `trace.markBits(idx)`, i am not sure what this `bits` means, but maybe we can name this method as `traceevent.CheckAndMarkDumpTrigger`

pkg/session/session.go
```diff
@@ -1843,6 +1848,27 @@ func (s *session) getOomAlarmVariablesInfo() sessmgr.OOMAlarmVariablesInfo {
...
+				return config.Event.DevDebug.Type == "execute_internal_trace_missing"
```
D3Hunter: maybe use named constant, i see those string literal appear a few times, and other event type too

pkg/session/session.go
```diff
@@ -2353,6 +2379,14 @@ func queryFailDumpTriggerCheck(config *traceevent.DumpTriggerConfig) bool {
 	return config.Event.Type == "query_fail"
 }
 
+type isInternalAlias struct {
+	bool
+}
+
+func (i isInternalAlias) isInternalDumpTriggerCheck(config *traceevent.DumpTriggerConfig) bool {
+	return config.Event.IsInternal == i.bool
```
D3Hunter, line 2398: why we have to use a struct to do this? and it's only used in one place, maybe inline and check directly?

pkg/util/traceevent/flightrecorder.go
D3Hunter: maybe `compiledDumpTriggerConfig`

pkg/session/session.go
tiancaiamao: If we have several conditions to trigger dump, for example, A and B and C
Each A B C is represented with one bit, so the `Check` process did change that internal bit state, but only when all necessary bits are marked, it will really trigger.

`CheckAndMarkDumpTrigger` is not accurate here, it's indeed doing the check.

pkg/session/session.go
tiancaiamao: There is a incoming refactor in another PR https://github.com/pingcap/tidb/pull/64569
So don't bother here.

I use `isInternalAlias` here because I cannot trust Go compiler to do escape analysis correctly to avoid memory allocation.

For example:

```
var someValue bool
CheckFlightRecorderDumpTrigger(ctx, "dump_trigger.some.condition", func(config *DumpTriggerConfig) {
    return config.XX.YY == someValue
})
```

It's not sure the closure  func(config *DumpTriggerConfig) cause allocation here.

But writing this:

```
type isInternalAlias struct {
	bool
}

func (i isInternalAlias) isInternalDumpTriggerCheck(config *traceevent.DumpTriggerConfig) bool {
	return config.Event.IsInternal == i.bool
}

var someValue bool
mytypeAlias := isInternal{someValue}
CheckFlightRecorderDumpTrigger(ctx, "dump_trigger.some.condition", mytypeAlias.isInternalDumpTriggerCheck)
```

I'm pretty sure there is no allocation for `mytypeAlias.isInternalDumpTriggerCheck`, I manually write this for the Go compiler, it works even its  escape analysis is silly.

pkg/util/traceevent/flightrecorder.go
tiancaiamao: done

pkg/util/traceevent/flightrecorder.go
tiancaiamao: done

pkg/util/traceevent/flightrecorder_test.go
tiancaiamao: done

pkg/session/session.go
tiancaiamao: done

pkg/session/session.go
D3Hunter: maybe write a comment to explain this,  it's easy to forget and have questions

tiancaiamao: done

PR #64356, opened 2025-11-07
ddl: add auto ID rebase create table submitted by BR

joechenrh:

Issue Number: ref https://github.com/pingcap/tidb/issues/60804

Problem Summary:

In https://github.com/pingcap/tidb/issues/60804, we move auto id rebase from submitter to executor, to ensure the atomicity of create table.

In most cases, there are no problem with such changes. But BR sometimes restore db to lower version of TiDB cluster. For example, if BR with ver9.0 (or any version with #60804) backups db to TiDB with ver7.5 (or any version without #60804), rebase won't be executed by anyone.

To temporary solve this, this PR revert back part of the changes in #60804 to make restore happy again 😢.

Besides, there is a small problem with https://github.com/pingcap/tidb/pull/55847. In this PR, we copy the old value of auto id to new table. Typically, it's ok with same version of TiDB. But in the following scenario, it also introduce a problem during rename table:
- In TiDB v7.5.6, the job worker handles auto ID during rename table by copying the old value, deleting it, and then setting the new value.
- While in v9.0, auto ID is rebased when applying the info schema diff. This means that when BR v9.0 applies the new diff, the old value has already been deleted, and it will overwrite the existing value with 0.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

BR has already used this PR to do test, and it works.

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/meta/meta_autoid.go
+++ b/pkg/meta/meta_autoid.go
@@ -88,7 +88,14 @@ func (a *autoIDAccessor) CopyTo(databaseID, tableID int64) error {
 		return err
 	}
 	// There is no need to copy zero value. And in certain case, this can overwrite
-	// the existing auto ID of the target table.
+	// the existing auto ID of the target table, for example, when restoring data with
+	// a higher version of BR (≥ 8.5) into a lower version of TiDB (≤ 7.5):
+	// In older versions of TiDB, the job worker handles auto ID during rename
+	// table by copying the old value, deleting it, and then setting the new value.
+	// However, in newer versions of TiDB, auto ID is rebased when applying the info
+	// schema diff. This means that when BR applies the new diff here, the original
+	// value have already been deleted. In such cases, the existing value will be
+	// overwritten with zero, which is incorrect.
 	if curr == 0 {
 		return nil
 	}


joechenrh:
/hold

BR is now testing...

pkg/ddl/executor.go
```diff
@@ -1191,6 +1191,9 @@ func (e *executor) CreateTableWithInfo(
 		}
 
 		preSplitAndScatterTable(ctx, e.store, tbInfo, scatterScope)
+		if err := handleAutoIncID(e.getAutoIDRequirement(), jobW.Job, tbInfo); err != nil {
```
D3Hunter: can we only run this when the DDL is running with BR, you can use the mode to check, see https://github.com/pingcap/tidb/pull/64320

in this way, we can avoid rebase the id twice in normal cases

pkg/ddl/executor.go
D3Hunter, line 1: And we can also check this in the job worker, only rebase when it's run in non-BR mode

and please add comments to explain this scenario in detail, it's easy to forget

pkg/ddl/ddl.go
```diff
@@ -124,6 +124,15 @@ type CreateTableConfig struct {
 	// by BR now. By reusing IDs BR can save a lot of works such as rewriting table
 	// IDs in backed up KVs.
 	IDAllocated bool
+
+	// RebaseAutoID indicates whether to rebase auto ID for the created table for the
+	// job submitter, which is is only used by BR now.
+	// Since https://github.com/pingcap/tidb/pull/64356, we move rebase logic from
+	// submitter into DDL executor. But sometimes we use higher version of BR to backup
```
D3Hunter:
```suggestion
	// submitter into DDL executor. But sometimes we use higher version of BR to restore
```

pkg/ddl/executor.go
joechenrh: Done, store the mode in executor, so normal execution of create won't rebase twice.

Leavrth:
/unhold
br test (v7.5.6, v8.5.3) passed:
before:
```
mysql> show table mysql.stats_histograms next_row_id;
+---------+------------------+-------------+--------------------+-------------+
| DB_NAME | TABLE_NAME       | COLUMN_NAME | NEXT_GLOBAL_ROW_ID | ID_TYPE     |
+---------+------------------+-------------+--------------------+-------------+
| mysql   | stats_histograms | _tidb_rowid |                  1 | _TIDB_ROWID |
+---------+------------------+-------------+--------------------+-------------+
1 row in set (0.00 sec)
```

after:
```
mysql> show table mysql.stats_histograms next_row_id;
+---------+------------------+-------------+--------------------+-------------+
| DB_NAME | TABLE_NAME       | COLUMN_NAME | NEXT_GLOBAL_ROW_ID | ID_TYPE     |
+---------+------------------+-------------+--------------------+-------------+
| mysql   | stats_histograms | _tidb_rowid |              30001 | _TIDB_ROWID |
+---------+------------------+-------------+--------------------+-------------+
1 row in set (0.00 sec)
```

pkg/meta/meta_autoid.go
```diff
@@ -87,6 +87,11 @@ func (a *autoIDAccessor) CopyTo(databaseID, tableID int64) error {
 	if err != nil {
 		return err
 	}
+	// There is no need to copy zero value. And in certain case, this can overwrite
```
D3Hunter, line 90: what is the "certain case"?

D3Hunter:
/hold

wait for the comment, feel free to unhold

joechenrh:
/unhold

The comment has been updated

PR #64359, opened 2025-11-07
coprocessor: ensure monotonicity of ranges

ekexium:

Issue Number: ref #64008, 
This is an attempt to fix cse#3868

Problem Summary:

Ensure coprocessor ranges are sorted.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/store/copr/coprocessor.go
+++ b/pkg/store/copr/coprocessor.go
@@ -337,29 +337,119 @@ type buildCopTaskOpt struct {
 	ignoreTiKVClientReadTimeout bool
 }
 
+const (
+	rangeIssueDuplicate    = "duplicate"
+	rangeIssueOverlap      = "overlap"
+	rangeIssueContain      = "contain"
+	rangeIssueOutOfOrder   = "outOfOrder"
+	rangeIssueInvalidBound = "invalidBounds"
+	rangeIssueInfiniteTail = "infiniteTail"
+)
+
+type rangeIssueStats struct {
+	Duplicate    int `json:"duplicate,omitempty"`
+	Overlap      int `json:"overlap,omitempty"`
+	Contain      int `json:"contain,omitempty"`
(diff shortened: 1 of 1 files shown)


pkg/store/copr/coprocessor.go
```diff
@@ -335,10 +337,64 @@ type buildCopTaskOpt struct {
 	ignoreTiKVClientReadTimeout bool
 }
 
+func ensureMonotonicKeyRanges(ctx context.Context, ranges *KeyRanges) bool {
```
winoros, line 424: It's better that the error can be thrown to the upper level so that we can add more checks to know the real cause of the issue.
Whether to throw it in prod env is okay to me.
The idea that we only recover it in test env also sounds good to me.

ekexium: CI won't pass if we panic here. Some cases can produce such invalid ranges

pkg/store/copr/coprocessor.go
```diff
@@ -335,10 +337,64 @@ type buildCopTaskOpt struct {
...
+			sorted = false
+			break
+		}
+		prev = curr
+	}
+	if sorted && valid {
+		return false
+	}
+	flat := ranges.ToRanges()
+	logutil.Logger(ctx).Error("key ranges not monotonic, reorder before BatchLocateKeyRanges",
```
cfzjywxk: Would this information be collected into the trace event?

ekexium: No. If a log is printed at a level higher than DEBUG, we don't want to duplicate it as a trace event, at least for now.

ekexium:
@winoros I added more details and classfication. PTAL

PR #56923, opened 2024-10-28
executor, store: Plumb the query max execution time to timebox the PD GetRegion grpc calls

HaoW30:
### What problem does this PR solve?
This change plumbs the TiDB query max execution time to the executor point_get(bath_point_get) and coprocessor(batch_coprocessor) tasks to timebox the GetRegion PD grpc calls (and other task operations). 

Issue Number: ref #56753

Problem Summary:

### What changed and how does it work?
1. Plumbed query execution time to kv request and distsql context
2. Set context deadline based on query max execution time for point get snapshot read (point_get and batch_point_get)
3. Set context deadline based on query max execution time for coprocessor tasks (coprocessor and batch_coprocessor)

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No need to test
  > - [ ] I checked and no code files have been changed.
  > 

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #64361, opened 2025-11-07
*: unifiy the usage of max_execution_time

### LINKED ISSUE #64360: max_execution_time should only take affect on select sql
## Bug Report

According to mysql's doc, max_execution_time only works for select sql, in TiDB, it try to align this behavior, however, there is still some corner cases which max_execution_time affect non select sql, for example, in #56923, it tries to set context's timeout using `max_execution_time`, so rpc can be aware of `max_execution_time`, but it actually does not check `InSelectStmt`, which means a non select sql will be affected by this pr.

windtalker:

Issue Number: close #64360

Problem Summary:
Described in #64360
* check `InSelectStmt` in `GetMaxExecutionTime()`
* replace all the usage of `vars.MaxExecutionTime` by `vars.GetMaxExecutionTime()` during sql execution
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/pkg/executor/batch_point_get.go
+++ b/pkg/executor/batch_point_get.go
@@ -237,10 +237,11 @@ func (e *BatchPointGetExec) initialize(ctx context.Context) error {
 	var indexKeys []kv.Key
 	var err error
 	batchGetter := e.batchGetter
-	if e.Ctx().GetSessionVars().GetMaxExecutionTime() > 0 {
+	maxExecutionTime := e.Ctx().GetSessionVars().GetMaxExecutionTime()
+	if maxExecutionTime > 0 {
 		// If MaxExecutionTime is set, we need to set the context deadline for the batch get.
 		var cancel context.CancelFunc
-		ctx, cancel = context.WithTimeout(ctx, time.Duration(e.Ctx().GetSessionVars().GetMaxExecutionTime())*time.Millisecond)
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(maxExecutionTime)*time.Millisecond)
 		defer cancel()
 	}
 	rc := e.Ctx().GetSessionVars().IsPessimisticReadConsistency()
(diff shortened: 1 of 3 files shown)


pkg/executor/batch_point_get.go
```diff
@@ -237,10 +237,10 @@ func (e *BatchPointGetExec) initialize(ctx context.Context) error {
 	var indexKeys []kv.Key
 	var err error
 	batchGetter := e.batchGetter
-	if e.Ctx().GetSessionVars().MaxExecutionTime > 0 {
```
windtalker, line 240: @HaoW30 please take a look, in your pr #56923, the MaxExecutionTime will be set even for non-select sql, which is not expected because MaxExecutionTime should only take affects for select query: https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_max_execution_time

pkg/sessionctx/variable/session.go
```diff
@@ -3556,7 +3556,12 @@ func (s *SessionVars) GetRuntimeFilterMode() RuntimeFilterMode {
 }
 
 // GetMaxExecutionTime get the max execution timeout value.
```
zimulala: Do we need to explicitly mention “for SELECT statements”, similar to the [MaxExecutionTime](https://github.com/pingcap/tidb/blob/069a5086efd3a47ca0e7906e87b50a3716e18e0d/pkg/sessionctx/variable/session.go#L1224C37-L1224C57) definition?

windtalker: ok

zimulala (approved):

you06 (approved):

PR #64408, opened 2025-11-11
executor, sessiontxn/isolation: fix max_execution_time to include TSO wait time

### LINKED ISSUE #64406: Enhancement: Start max_execution_timeout timer before buildExecutor() so TSO wait is included
## Enhancement
Now, max_execution_time starts when we set process info during the executor phase (executor/adapter.go: (*ExecStmt).Exec → pi.SetProcessInfo(sql, time.Now(), cmd, maxExecutionTime)), [refer](https://github.com/pingcap/tidb/blob/master/pkg/executor/adapter.go#L643-L663).

Although, TSO fetching is initiated asynchronously (GetTimestampAsync created during Optimize, after parse/compile), the blocking wait happens later in Exec during executor construction still exists:

Exec
  → buildExecutor()
    → b.build(a.Plan) → (e.g., buildPointGet)
      → b.getSnapshot() → GetSnapshotWithStmtReadTS()
        → getStmtTS() → stmtTSFuture.Wait()  

Because the timer starts after buildExecutor(), the time spent waiting on stmtTSFuture.Wait() is not counted. This causes:
- Timeout control that does not reflect end-to-end latency under PD/TSO pressure.
- ProcessList/observability underreporting perceived latency.
- User confusion when a “slow” query doesn’t time out.

King-Dylan:

Issue Number: close #64406

Problem Summary:
Because the timer starts after buildExecutor(), the time spent waiting on stmtTSFuture.Wait() is not counted. If, in some cases, we wait a long time for TSO, the max_execution_timeout parameter will appear to have no effect.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No need to test
  > 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix an issue where the max_execution_time parameter did not include the time spent waiting for TSO, which could make the limit ineffective when TSO allocation is slow.
```

### DIFF
--- a/pkg/sessiontxn/isolation/base.go
+++ b/pkg/sessiontxn/isolation/base.go
@@ -296,6 +296,14 @@ func (p *baseTxnContextProvider) ActivateTxn() (kv.Transaction, error) {
 	}
 
 	txnFuture := p.sctx.GetPreparedTxnFuture()
+
+	// Inject delay before TSO wait for testing maxExecutionTime
+	failpoint.Inject("injectTSOWaitDelay", func(val failpoint.Value) {
+		if delayMs, ok := val.(int); ok {
+			time.Sleep(time.Duration(delayMs) * time.Millisecond)
+		}
+	})
+
 	txn, err := txnFuture.Wait(p.ctx, p.sctx)
 	if err != nil {
 		return nil, err
@@ -673,15 +681,10 @@ func newOracleFuture(ctx context.Context, sctx sessionctx.Context, scope string)
 	oracleStore := sctx.GetStore().GetOracle()
 	option := &oracle.Option{TxnScope: scope}
 
-	var innerFuture oracle.Future
 	if sctx.GetSessionVars().UseLowResolutionTSO() {
-		innerFuture = oracleStore.GetLowResolutionTimestampAsync(ctx, option)
-	} else {
-		innerFuture = oracleStore.GetTimestampAsync(ctx, option)
+		return oracleStore.GetLowResolutionTimestampAsync(ctx, option)
 	}
-
-	// Wrap the future to inject delay for testing maxExecutionTime behavior
-	return &delayedOracleFuture{inner: innerFuture}
+	return oracleStore.GetTimestampAsync(ctx, option)
 }
 
 // funcFuture implements oracle.Future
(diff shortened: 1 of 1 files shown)


bb7133:
The code change LGTM, but please add a proper 'Release Note' in the PR description.

pkg/sessiontxn/isolation/base.go
```diff
@@ -687,6 +692,22 @@ func (f funcFuture) Wait() (uint64, error) {
 	return f()
 }
 
+// delayedOracleFuture wraps an oracle.Future to inject delay before Wait() for testing
+type delayedOracleFuture struct {
```
zimulala: The `delayedOracleFuture` wrapper introduces test-only logic into the production path, adding an extra indirection for every TSO fetch.

A cleaner approach might be to inject the failpoint directly at the `Wait()` call site in `ActivateTxn()`, instead of wrapping the entire oracle. This would keep the production code simpler and better isolate test-only behavior. The code is as follows:
```
func (p *baseTxnContextProvider) ActivateTxn() (kv.Transaction, error) {
    // ...
    txnFuture := p.sctx.GetPreparedTxnFuture()
    
    // Inject delay before TSO wait for testing maxExecutionTime
    failpoint.Inject("injectTSOWaitDelay", func(val failpoint.Value) {
        if delayMs, ok := val.(int); ok {
            time.Sleep(time.Duration(delayMs) * time.Millisecond)
        }
    })
    
    txn, err := txnFuture.Wait(p.ctx, p.sctx)
    // ...
}
```

pkg/executor/adapter_test.go
```diff
@@ -540,3 +540,88 @@ func BenchmarkCheckSlowLogRulesPreAlloc(b *testing.B) {
...
+			description:      "TSO wait time (800ms) should be included, total < 1000ms",
+		},
+		{
+			name:             "TSO delay 1200ms, timeout 500ms - should timeout",
+			tsoDelayMs:       1200,
+			maxExecutionTime: 500,
+			expectTimeout:    true,
+			description:      "TSO wait time (1200ms) exceeds timeout (500ms)",
+		},
+	}
```
zimulala, line 580: The test may be flaky in CI due to:
1. Long execution time: Total sleep time is 300+800+1200 = 2.3s, which could be slower under CI load.
2. Tight margins: The case {tsoDelayMs: 800, maxExecutionTime: 1000} has only 200ms headroom, which may not be enough when CI is under heavy load.

Suggestion: Reduce sleep times and increase the gap between delay and timeout:
```
testCases := []struct {
    // ...
}{
    {
        name:             "no timeout - TSO delay within limit",
        tsoDelayMs:       50,
        maxExecutionTime: 500,  // 450ms headroom
        expectTimeout:    false,
    },
    {
        name:             "timeout - TSO delay exceeds limit", 
        tsoDelayMs:       200,
        maxExecutionTime: 50,   // clear 150ms gap
        expectTimeout:    true,
    },
}
```

zimulala (approved):

